Issue #1121, opened 2021-10-28
Application Flow: Admin Card auth - all apps


carolinemodic:
- [x] Election Manager
- [x] BSD
- [x] Precinct Scanner
- [ ] BMD
- [ ] BAS

PR #1143, opened 2021-11-01
feat(all) log authentication actions in useUserSession

carolinemodic:
Adds logging for authentication actions in the useUserSession shared hook. BMD and BAS do not currently use this hook so I will have separate PRs for those apps (either migrating them to use this or just implementing the logic separately for those apps, haven't decided yet). 

I wasn't entirely sure what "user" the user in the log should refer to when the user is changing in many of these actions (login/logout) instead of being consistent on always being the user prior to or after to the change I instead just tried to use the user if we were confident of who the user was, and "unknown" if we didn't know. So when logging an admin out, we know thats an admin taking that action so we use the "admin" user, but when inserting the admin care, before authentication, we do NOT yet know the user so the user is "unknown." When we successfully two factor the passcode then we do know that the user is an admin and can start using admin again. 

https://github.com/votingworks/vxsuite/issues/1121


libs/logging/src/types.ts
```diff
@@ -7,11 +7,17 @@ export enum LogDispositionStandardTypes {
   NotApplicable = 'na',
 }
 
-export type LoggingUserRole = CardDataTypes | 'vx-employee' | 'system';
+export type LoggingUserRole =
+  | CardDataTypes
+  | 'vx-employee'
```
benadida, line 12: maybe `vx-staff` ?

benadida (approved):
One suggestion: can we make the card types an enum or something else that has better type checking that `'admin'` strings in a few different places?

eventualbuddha:
> One suggestion: can we make the card types an enum or something else that has better type checking that `'admin'` strings in a few different places?

FWIW, string literals in TS types are still type-safe to the same degree enums are. Any random `string` is not considered acceptable when the type asks for e.g. `'admin'`.

PR #1146, opened 2021-11-02
First pass at module documentation

benadida:
very first high-level pass at our modules.

Leading me to think that we should come up with a better taxonomy that is VVSG compatible. I suggest:

- *subsystem*: independent hardware component, like the BMD, or the Precinct Scanner.
- *frontend*: a frontend process
- *service*: a backend process
- *module*: as per VVSG, a chunk of thematically consistent code.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -1,0 +1,136 @@
+# VotingWorks Modules
+
+The VotingWorks source code, which implements every component of the
+voting system, is organized into modules as detailed below.
+
+## Libraries
+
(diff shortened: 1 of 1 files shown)


mcchilders (approved):
LGTM - agree we may want to adjust some naming but this is really helpful as a start.

eventualbuddha:
My main question is: how do we ensure this is kept up to date?

benadida:
> My main question is: how do we ensure this is kept up to date?

I'm game for trying to automate this from `package.json`...

PR #1147, opened 2021-11-03
style(all): `gts-unicode-escapes`

### LINKED ISSUE #1027: GTS: Syntax → File encoding: UTF-8
> For non-ASCII characters, use the actual Unicode character (e.g. ∞). For non-printable characters, the equivalent hex or Unicode escapes (e.g. \u221e) can be used along with an explanatory comment.

https://google.github.io/styleguide/tsguide.html#file-encoding-utf-8

Part of #788

eventualbuddha:
Comes from [Google TypeScript Style Guide on "File encoding: UTF-8"](https://google.github.io/styleguide/tsguide.html#file-encoding-utf-8):

> For non-ASCII characters, use the actual Unicode character (e.g. `∞`). For non-printable characters, the equivalent hex or Unicode escapes (e.g. `\u221e`) can be used along with an explanatory comment.

Closes #1027

### DIFF
--- a/libs/eslint-plugin-vx/tsconfig.json
+++ b/libs/eslint-plugin-vx/tsconfig.json
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.shared.json",
   "compilerOptions": {
     "target": "ES2019",
-    "lib": ["ES2019"],
+    "lib": ["ES2020"],
     "module": "CommonJS",
     "moduleResolution": "Node",
     "esModuleInterop": true,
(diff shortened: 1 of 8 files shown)


carolinemodic (approved):

libs/eslint-plugin-vx/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.shared.json",
   "compilerOptions": {
     "target": "ES2019",
-    "lib": ["ES2019"],
+    "lib": ["ES2020"],
```
eventualbuddha, line 5: This was for `String#matchAll`.

PR #1148, opened 2021-11-03
feat(all) add application logging for usb state changes

carolinemodic:
Adds logging for usb-related things from the application side. Separately I'll also add system level logging when a new device is noticed. 

Since I log specific events related to mount/unmount AND I have a catchall log when the state changes some things get double logged, like after mounting there are the following three logs:
usb-drive-mount-init - when the mounting begins
usb-drive-mount-complete - when the mounting completes has a success or failure disposition
usb-drive-status-update - Status update changed from present to mounted. 

But I'm erring on the side of sometimes having excessive logs to sometimes having missed logs.

For some reason the useInterval loop in useUsbDrive sometimes runs multiple times (in the tests its consistently twice) when it first starts causing there to be multiple logs of the status changing from undefined to present, or whatever the first status change is. I'm not really sure why this happens but its somewhat out of scope of this change, and ultimately I don't think its a huge deal to have a log like that appear multiple times on occasion.

### DIFF
--- a/apps/bsd/src/contexts/AppContext.ts
+++ b/apps/bsd/src/contexts/AppContext.ts
@@ -1,16 +1,18 @@
-import { ElectionDefinition } from '@votingworks/types';
+import { LoggingUserRole } from '@votingworks/logging/src';
+import { ElectionDefinition, UserSession } from '@votingworks/types';
 import { MemoryStorage, Storage, usbstick } from '@votingworks/utils';
 import { createContext } from 'react';
 import { MachineConfig } from '../config/types';
 
 interface AppContextInterface {
   usbDriveStatus: usbstick.UsbDriveStatus;
-  usbDriveEject: () => void;
+  usbDriveEject: (currentUser: LoggingUserRole) => void;
   machineConfig: MachineConfig;
   electionDefinition?: ElectionDefinition;
   electionHash?: string;
   storage: Storage;
   lockMachine: () => void;
+  currentUserSession?: UserSession;
 }
 
 const appContext: AppContextInterface = {
@@ -21,6 +23,7 @@ const appContext: AppContextInterface = {
   electionHash: undefined,
   storage: new MemoryStorage(),
   lockMachine: () => undefined,
+  currentUserSession: undefined,
 };
 
 export const AppContext = createContext(appContext);
--- a/apps/election-manager/src/components/ExportElectionBallotPackageModalButton.tsx
+++ b/apps/election-manager/src/components/ExportElectionBallotPackageModalButton.tsx
@@ -34,10 +34,14 @@ const USBImage = styled.img`
 `;
 
 export function ExportElectionBallotPackageModalButton(): JSX.Element {
-  const { electionDefinition, usbDriveStatus, usbDriveEject } = useContext(
-    AppContext
-  );
+  const {
+    electionDefinition,
+    usbDriveStatus,
+    usbDriveEject,
+    currentUserSession,
+  } = useContext(AppContext);
   assert(electionDefinition);
+  assert(currentUserSession); // TODO(auth) should this make sure we have an admin
   const { election, electionData, electionHash } = electionDefinition;
   const electionLocaleCodes = getElectionLocales(election, DEFAULT_LOCALE);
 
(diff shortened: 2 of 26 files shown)


apps/bsd/src/contexts/AppContext.ts
```diff
@@ -1,16 +1,17 @@
-import { ElectionDefinition } from '@votingworks/types';
+import { ElectionDefinition, UserSession } from '@votingworks/types';
 import { MemoryStorage, Storage, usbstick } from '@votingworks/utils';
 import { createContext } from 'react';
 import { MachineConfig } from '../config/types';
 
 interface AppContextInterface {
   usbDriveStatus: usbstick.UsbDriveStatus;
-  usbDriveEject: () => void;
+  usbDriveEject: (currentUser: string) => void;
```
eventualbuddha, line 8: Is there a more specific type we want to use than `string`?

apps/election-manager/src/components/ExportElectionBallotPackageModalButton.tsx
```diff
@@ -34,10 +34,14 @@ const USBImage = styled.img`
...
-    AppContext
-  );
+  const {
+    electionDefinition,
+    usbDriveStatus,
+    usbDriveEject,
+    currentUserSession,
+  } = useContext(AppContext);
   assert(electionDefinition);
+  assert(currentUserSession); // TODO(auth) should this make sure we have an admin
```
eventualbuddha, line 44: Why not `assert(currentUserSession?.type === 'admin')`? What needs to happen before we can do that? This comment isn't really clear about that.

apps/election-manager/src/contexts/AppContext.ts
```diff
@@ -36,7 +36,7 @@ export interface AppContextInterface {
   saveIsOfficialResults: () => Promise<void>;
   resetFiles: (fileType: ResultsFileType) => Promise<void>;
   usbDriveStatus: usbstick.UsbDriveStatus;
-  usbDriveEject: () => Promise<void>;
+  usbDriveEject: (currentUser: string) => Promise<void>;
```
eventualbuddha, line 39: This is `string` in a bunch of places, but it should be `LoggingUserRole`, right?

libs/logging/src/logEventTypes.ts
```diff
@@ -44,6 +45,12 @@ const ApplicationStatusEventDocumentation: LogEventTypeDocumentation = {
     'Status update or message that the application took without user action.',
 };
 
+const ApplicationActionEventDocumentation: LogEventTypeDocumentation = {
+  eventType: LogEventType.ApplicationAction,
+  documentationMessage:
+    'Action taken by the votingworks application automatically when a certain condition is met. Example: When a new USB drive is detected, the application will automatically mount the USB.',
```
eventualbuddha, line 51: I think there's a good argument to be made that calling a "USB drive" just "USB" for users is okay since they often call it that. However, I think for our internal stuff we should be careful to distinguish between "USB drives" and "USB in general", i.e. an arbitrary USB device. I'd reword this like so:

```suggestion
    'Action taken by the votingworks application automatically when a certain condition is met. Example: When a new USB drive is detected, the application will automatically mount the it.',
```

Or "the drive" or "the USB drive".

libs/logging/src/logger.ts
```diff
@@ -1,6 +1,6 @@
 import makeDebug from 'debug';
 import { Dictionary } from '@votingworks/types';
-import { getDetailsForEventId, LogLine } from '.';
+import { CLIENT_SIDE_LOG_SOURCES, getDetailsForEventId, LogLine } from '.';
```
eventualbuddha, line 3: Import from the files that contain these rather than the re-exporting `index.ts`?

eventualbuddha (approved):

apps/election-manager/src/components/ExportElectionBallotPackageModalButton.tsx
carolinemodic: I'm a little confused where we are currently at with some of the auth stuff, and I think we are planning to add another role (clerk), so I just thought it was better to make a searchable TODO for these to re-evaluate when we have a clear auth structure in the apps rather then trying to figure it all out now.

PR #1167, opened 2021-11-05
feat(all) refactor to useHardware hook and add logging

carolinemodic:
Creates a shared useHardware hook that controls all of the hasXAttached boolean flags for when various devices are attached. I added one for the plustek scanner in case we want that in the future (and to log it) but it isn't used anywhere yet. The Fujitsu scanner one only checks the vendor ID and not the product ID to maintain existing behavior. 

Whenever a new device or printer is attached or removed there are logs associated with it, even if its not a device that we have a specific flag for. 

Note: I had to commit with --no-verify as there was an issue with stylelint in BAS, I'm seeing this issue in main as well. 

Future Work: Add the charger unplugged / low battery logic to this hook.

### DIFF
--- a/libs/ui/src/hooks/useHardware.test.ts
+++ b/libs/ui/src/hooks/useHardware.test.ts
@@ -1,0 +1,311 @@
+import { act, renderHook } from '@testing-library/react-hooks';
+import { Logger, LogSource, LogEventId } from '@votingworks/logging';
+import {
+  AccessibleControllerProductId,
+  AccessibleControllerVendorId,
+  FujitsuFi7160ScannerProductId,
+  FujitsuScannerVendorId,
+  MemoryHardware,
+  OmniKeyCardReaderProductId,
+  OmniKeyCardReaderVendorId,
+  PlustekScannerVendorId,
+  PlustekVTM300ScannerProductId,
+} from '@votingworks/utils';
+import { useHardware } from './useHardware';
+
+test('can connect printer as expected', async () => {
+  const hardware = new MemoryHardware();
+  const fakeLogger = new Logger(LogSource.VxBatchScanApp);
+  const logSpy = jest.spyOn(fakeLogger, 'log').mockResolvedValue();
+  const { result, rerender } = renderHook(() =>
+    useHardware({ hardware, logger: fakeLogger })
+  );
+  expect(result.current.hasPrinterAttached).toBe(false);
+  expect(logSpy).toHaveBeenCalledTimes(0);
+
+  await act(async () => await hardware.setPrinterConnected(true));
+  rerender();
+  expect(result.current.hasPrinterAttached).toBe(true);
+  expect(logSpy).toHaveBeenCalledTimes(2);
+  expect(logSpy).toHaveBeenNthCalledWith(
+    1,
+    LogEventId.PrinterConfigurationAdded,
+    'system',
+    expect.objectContaining({
+      message: expect.stringContaining('New printer configured'),
+      connected: true,
+    })
+  );
+  expect(result.current.hasCardReaderAttached).toBe(false);
+  expect(result.current.hasBatchScannerAttached).toBe(false);
+  expect(result.current.hasPrecinctScannerAttached).toBe(false);
+  expect(result.current.hasAccessibleControllerAttached).toBe(false);
+
+  await act(async () => await hardware.setPrinterConnected(false));
+  rerender();
+  expect(result.current.hasPrinterAttached).toBe(false);
+  expect(logSpy).toHaveBeenCalledWith(
+    LogEventId.PrinterConnectionUpdate,
+    'system',
+    expect.objectContaining({
+      message: expect.stringContaining('Printer'),
+      connected: false,
+    })
+  );
+
+  await act(async () => await hardware.setPrinterConnected(true));
+  rerender();
+  expect(result.current.hasPrinterAttached).toBe(true);
+  expect(logSpy).toHaveBeenCalledWith(
+    LogEventId.PrinterConnectionUpdate,
+    'system',
+    expect.objectContaining({
+      message: expect.stringContaining('Printer'),
+      connected: true,
+    })
+  );
+
+  await act(async () => await hardware.detachAllPrinters());
+  rerender();
+  expect(result.current.hasPrinterAttached).toBe(false);
+  expect(logSpy).toHaveBeenCalledWith(
+    LogEventId.PrinterConfigurationRemoved,
+    'system',
+    expect.objectContaining({
+      message: expect.stringContaining('Printer'),
+    })
+  );
+});
+
+test('can connect card reader as expected', async () => {
+  const hardware = new MemoryHardware();
+  const fakeLogger = new Logger(LogSource.VxBatchScanApp);
+  const logSpy = jest.spyOn(fakeLogger, 'log').mockResolvedValue();
+  const { result, rerender } = renderHook(() =>
+    useHardware({ hardware, logger: fakeLogger })
+  );
+  expect(result.current.hasCardReaderAttached).toBe(false);
+  expect(logSpy).toHaveBeenCalledTimes(0);
+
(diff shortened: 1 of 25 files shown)


libs/ui/src/hooks/useHardware.ts
```diff
@@ -0,0 +1,175 @@
...
+          connected: newPrinter.connected.toString(),
+        });
+      }
+    }
+    for (const oldPrinter of previousPrinters || []) {
+      const newCopyOfPrinter = allPrinters.find(
+        (printer) => printer.name === oldPrinter.name
+      );
+      if (!newCopyOfPrinter) {
+        void logger.log(LogEventId.PrinterUnattached, 'system', {
```
carolinemodic, line 120: @eventualbuddha Not sure if you are more familiar with how kiosk-browser works for this but I don't fully understand when/if this would happen. In my testing I basically always had two printers (one named for my actual HP printer and the other "VxPrinter" in the array coming from window.kiosk.printers.subscribe)and as I connect/disconnect from my VM the connection status (on both of these) will flip but they never actually get unattached and are present even when the printer is not connected.

libs/ui/src/hooks/useHardware.ts
```diff
@@ -0,0 +1,175 @@
...
+export interface UseHardwareProps {
+  hardware: Hardware;
+  logger: Logger;
+}
+
+export interface UseHardwareResult {
+  hasCardReaderAttached: boolean;
+  hasAccessibleControllerAttached: boolean;
+  hasPlustekScannerAttached: boolean;
+  hasFujitsuScannerAttached: boolean;
```
eventualbuddha, line 22: Do we want the brand names here? Or should we do something like this?

```suggestion
  hasPrecinctScannerAttached: boolean;
  hasBatchScannerAttached: boolean;
```

libs/ui/src/hooks/useHardware.ts
eventualbuddha: I assume this would happen if the printer is unconfigured. Like, you go into the system settings and remove the printer from the list.

libs/ui/src/hooks/useHardware.ts
```diff
@@ -0,0 +1,175 @@
...
+            device.productId === newDevice.productId &&
+            device.vendorId === newDevice.vendorId
+        );
+      if (!previousCopyOfDevice) {
+        void logger.log(LogEventId.DeviceAttached, 'system', {
+          message: `New ${getDeviceName(newDevice)} attached. Vendor: ${
+            newDevice.vendorId
+          } , Product: ${newDevice.productId}`,
+          productId: newDevice.productId.toString(),
+          vendorId: newDevice.vendorId.toString(),
```
eventualbuddha, line 144: These are typically displayed as hex numbers, in my experience.

```suggestion
          productId: newDevice.productId.toString(16),
          vendorId: newDevice.vendorId.toString(16),
```

libs/ui/src/hooks/useHardware.ts
```diff
@@ -0,0 +1,175 @@
...
+        previousDevices &&
+        previousDevices.find(
+          (device) =>
+            device.productId === newDevice.productId &&
+            device.vendorId === newDevice.vendorId
+        );
+      if (!previousCopyOfDevice) {
+        void logger.log(LogEventId.DeviceAttached, 'system', {
+          message: `New ${getDeviceName(newDevice)} attached. Vendor: ${
+            newDevice.vendorId
```
eventualbuddha, line 141:
```suggestion
            newDevice.vendorId.toString(16)
```

libs/ui/src/hooks/useHardware.ts
```diff
@@ -0,0 +1,175 @@
...
+    for (const oldDevice of previousDevices || []) {
+      const newCopyOfDevice = allDevices.find(
+        (device) =>
+          device.productId === oldDevice.productId &&
+          device.vendorId === oldDevice.vendorId
+      );
+      if (!newCopyOfDevice) {
+        void logger.log(LogEventId.DeviceUnattached, 'system', {
+          message: `${getDeviceName(oldDevice)} unattached. Vendor: ${
+            oldDevice.vendorId
```
eventualbuddha, line 158: Etc.

```suggestion
            oldDevice.vendorId.toString(16)
```

libs/ui/src/hooks/useHardware.test.ts
```diff
@@ -0,0 +1,310 @@
...
+  rerender();
+  expect(result.current.hasPrinterAttached).toBe(true);
+  expect(logSpy).toHaveBeenCalledTimes(2);
+  expect(logSpy).toHaveBeenNthCalledWith(
+    1,
+    LogEventId.PrinterAttached,
+    'system',
+    expect.objectContaining({
+      message: expect.stringContaining('New printer attached'),
+      connected: 'true',
```
eventualbuddha, line 36: Why a `string` value here (and for `vendorId` etc elsewhere)?

libs/ui/src/hooks/useHardware.test.ts
```diff
@@ -0,0 +1,310 @@
...
+    })
+  );
+  expect(result.current.hasCardReaderAttached).toBe(false);
+  expect(result.current.hasFujitsuScannerAttached).toBe(false);
+  expect(result.current.hasPlustekScannerAttached).toBe(false);
+  expect(result.current.hasAccessibleControllerAttached).toBe(false);
```
eventualbuddha, line 42: Grouping these into one assertion can often improve debugging since you'll get more information in the error message. Your call.

```suggestion
  expect(result.current).toEqual({
    hasCardReaderAttached: false,
    hasFujitsuScannerAttached: false,
    hasPlustekScannerAttached: false,
    hasAccessibleControllerAttached: false,
  });
```

libs/ui/src/hooks/useHardware.test.ts
```diff
@@ -0,0 +1,310 @@
...
+  expect(result.current.hasPrinterAttached).toBe(false);
+  expect(logSpy).toHaveBeenCalledWith(
+    LogEventId.PrinterUnattached,
+    'system',
+    expect.objectContaining({
+      message: expect.stringContaining('Printer'),
+    })
+  );
+});
+test('can connect card reader as expected', async () => {
```
eventualbuddha, line 79: Newline before.

```suggestion

test('can connect card reader as expected', async () => {
```

libs/ui/src/hooks/useSmartcard.ts
```diff
@@ -29,32 +28,28 @@ export interface Smartcard {
   writeLongValue(value: unknown | Uint8Array): Promise<Result<void, Error>>;
 }
 
-export type UseSmartcardResult = [
-  smartcard: Smartcard | undefined,
-  hasCardReader: boolean
-];
+export type UseSmartcardResult = [smartcard: Smartcard | undefined];
```
eventualbuddha, line 31: If it's just returning the one thing, unwrap the tuple?

```suggestion
export type UseSmartcardResult = Smartcard | undefined;
```

libs/ui/src/hooks/useSmartcard.ts
```diff
@@ -29,32 +28,28 @@ export interface Smartcard {
...
-  hasCardReaderAttached: false,
 };
 
 /**
  * React hook for getting the current smartcard data.
  *
  * @example
  *
- * const [smartcard, hasCardReader] = useSmartcard({ card, hardware })
+ * const { hasCardReaderAttached } = useSmartcard({ hardware })
```
eventualbuddha, line 49:
```suggestion
 * const { hasCardReaderAttached } = useHardware({ hardware, logger })
```

eventualbuddha (approved):

libs/ui/src/hooks/useHardware.test.ts
carolinemodic: I had typed the logging function to only take strings, but that seems silly now so I'll update to also accept boolean/numbers.

PR #1176, opened 2021-11-08
style(all): use `snake_case` filenames

### LINKED ISSUE #1022: GTS: Syntax → Identifiers → Imports
> Module namespace imports are lowerCamelCase while files are snake_case, which means that imports correctly will not match in casing style.

https://google.github.io/styleguide/tsguide.html#identifiers

Part of #788

eventualbuddha:
Look at the `feat(codemods)` commit to see the codemod that did most of the work.

Closes #1022


carolinemodic (approved):
FYI I just merged a PR so you might need to rerun the codemod. From looking at the codemod I was worried there might be places we have an acronym all caps in a filename like SEMS which would be bad to translate to s_e_m_s but I couldn't find any examples of that happening (although it is difficult lol)

PR #1188, opened 2021-11-10
style(all): require explicit object types

### LINKED ISSUE #1076: GTS: Type System → Structural Types vs Nominal Types
> When providing a structural-based implementation, explicitly include the type at the declaration of the symbol (this allows more precise type checking and error reporting).

https://google.github.io/styleguide/tsguide.html#structural-types-vs-nominal-types

Part of #788

eventualbuddha:
Enforces this [from GTS](https://google.github.io/styleguide/tsguide.html#structural-types-vs-nominal-types):

> TypeScript's type system is structural, not nominal. That is, a value matches a type if it has at least all the properties the type requires and the properties' types match, recursively.
>
> Use structural typing where appropriate in your code. Outside of test code, use interfaces to define structural types, not classes. In test code it can be useful to have mock implementations structurally match the code under test without introducing an extra interface.
>
> When providing a structural-based implementation, explicitly include the type at the declaration of the symbol (this allows more precise type checking and error reporting).

Closes #1076


apps/election-manager/src/__snapshots__/app.test.tsx.snap
```diff
@@ -144,7 +144,7 @@ exports[`printing ballots, print report, and test decks 1`] = `
       >
         Election Hash: 
         <strong>
-          6bdcd6d885
+          61cf15c77e
```
carolinemodic, line 147: why did this change?

carolinemodic (approved):

apps/election-manager/src/__snapshots__/app.test.tsx.snap
eventualbuddha: Because of this: https://github.com/votingworks/vxsuite/pull/1188/files#diff-3bbfed999ed23f50545ca95915cabc176aed8a7cef389532d0ecf6b7105da868R34-R36.

PR #1209, opened 2021-11-15
Add lint rule for bad workspace imports

carolinemodic:
Add a lint rule to catch bad imports of workspace libraries. 

I wasn't sure if hmpb-interpreter should be an exception to this rule or if its exports should just be fixed so it can fall in line with this but it was easy enough to just export the two missing things a the top-level, but happy to revert that if that is not how you want to handle hmpb-interpreter @eventualbuddha.

### DIFF
--- a/libs/eslint-plugin-vx/src/rules/no_import_workspace_subfolders.ts
+++ b/libs/eslint-plugin-vx/src/rules/no_import_workspace_subfolders.ts
@@ -8,7 +8,7 @@ export default createRule({
   meta: {
     docs: {
       description:
-        'When import libraries from the VotingWorks workspace, do not include subfolders like /src in the import',
+        'When importing libraries from the VotingWorks workspace, do not include subfolders like /src in the import',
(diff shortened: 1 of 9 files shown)


libs/eslint-plugin-vx/src/rules/no_import_workspace_subfolders.ts
```diff
@@ -0,0 +1,51 @@
...
+import { createRule } from '../util';
+
+const VOTINGWORKS_WORKSPACE_PREFIX = '@votingworks';
+
+export default createRule({
+  name: 'no-import-workspace-subfolders',
+  meta: {
+    docs: {
+      description:
+        'When import libraries from the VotingWorks workspace, do not include subfolders like /src in the import',
```
eventualbuddha, line 11:
```suggestion
        'When importing libraries from the VotingWorks workspace, do not include subfolders like /src in the import',
```

eventualbuddha (approved):

PR #1215, opened 2021-11-15
fix(bsd) fix setupProxy filename

carolinemodic:
I think react-scripts needs this to be named setupProxy and not setup_proxy. With it in snake cause pnpm build and make run worked fine but when I tried to run with pnpm start I got an exception about setupProxy.js not existing.


eventualbuddha (approved):
I thought I had corrected this. Thanks for catching!

PR #1228, opened 2021-11-17
feat(bsd) add logs for advanced screen actions

carolinemodic:
Adds logs for all of the actions on the advanced options screen of BSD.

### DIFF
--- a/libs/logging/src/log_event_ids.ts
+++ b/libs/logging/src/log_event_ids.ts
@@ -64,6 +64,14 @@ export enum LogEventId {
   ConvertingResultsToSemsFormat = 'converting-to-sems',
   TestDeckPrinted = 'test-deck-printed',
   TestDeckTallyReportPrinted = 'test-deck-tally-report-printed',
+  // VxBatch specific user action logs
+  TogglingTestMode = 'toggle-test-mode-init',
+  ToggledTestMode = 'toggled-test-mode',
+  ClearingBallotData = 'clear-ballot-data-init',
+  ClearedBallotData = 'clear-ballot-data-complete',
+  OverridingMarkThresholds = 'override-mark-threshold-init',
+  OverrodeMarkThresholds = 'override-mark-thresholds-complete',
+  DownloadedScanImageBackup = 'download-backup-scan-images,',
 }
 
 export interface LogDetails {
@@ -87,7 +95,7 @@ const ElectionUnconfiguredEvent: LogDetails = {
   defaultMessage:
     'Application has been unconfigured from the previous election.',
   documentationMessage:
-    'The user has unconfigured current machine to remove the current election definition.',
+    'The user has unconfigured current machine to remove the current election definition, and all other data.',
 };
 
 const MachineBootInitEvent: LogDetails = {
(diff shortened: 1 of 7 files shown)


apps/bsd/src/app_root.tsx
```diff
@@ -96,6 +96,10 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
     adjudication: { remaining: 0, adjudicated: 0 },
     scanner: ScannerStatus.Unknown,
   });
+  const currentNumberOfBallots = useMemo(
```
eventualbuddha, line 99: Why `useMemo` here?

apps/bsd/src/app_root.tsx
```diff
@@ -142,6 +146,10 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
     setElectionJustLoaded(true);
   }
 
+  const currentUserType = useMemo(() => currentUserSession?.type ?? 'unknown', [
```
eventualbuddha, line 149: And here, why `useMemo`?

apps/bsd/src/screens/advanced_options_screen.tsx
```diff
@@ -36,7 +43,10 @@ export function AdvancedOptionsScreen({
   markThresholds,
   electionDefinition,
 }: Props): JSX.Element {
-  const { lockMachine } = useContext(AppContext);
+  const { lockMachine, logger, currentUserSession } = useContext(AppContext);
+  const currentUserType = useMemo(() => currentUserSession?.type ?? 'unknown', [
```
eventualbuddha, line 47: Same question here re: `useMemo`. I won't spam any more of the same question after this.

eventualbuddha (approved):

libs/logging/src/log_event_ids.ts
```diff
@@ -291,6 +299,56 @@ const SmartcardProgrammedOverrideWriteProtection: LogDetails = {
...
+  eventId: LogEventId.OverrodeMarkThresholds,
+  eventType: LogEventType.UserAction,
+  documentationMessage:
+    'User has finished overriding the thresholds of when to count marks seen by the scanning module. Success or failure is indicated by the disposition. New mark thresholds specified in the keys `marginal` `definite` and, if defined, `writeInText`.',
+};
+const DownloadedScanImageBackup: LogDetails = {
+  eventId: LogEventId.DownloadedScanImageBackup,
+  eventType: LogEventType.UserAction,
+  documentationMessage:
+    'User downloaded a backup file of the scanned ballot image files and Cvrs. Success or failure indicated by disposition.',
```
eventualbuddha, line 349: Consistently use `CVR`?

```suggestion
    'User downloaded a backup file of the scanned ballot image files and CVRs. Success or failure indicated by disposition.',
```

apps/bsd/src/app_root.tsx
carolinemodic: Well I realized I forget to add this to the dependency array for the zeroData useCallback so I'll fix that but since this (and currentUserType) are used as dependencies in the useCallback calls I am using useMemo so they don't get redefined (and therefore cause the useCallback function to get redefined) on every render. 

It's not added in this PR but I'm also planning to add a log which would log every time this variable changes, which would be a useEffect with both this and currentUserType in the dependency array.

eventualbuddha: But isn't this (and the other case) using a primitive value? If it is the same primitive value it'll be `===` and memoizing won't do anything. Is an ESLint rule complaining about this?

carolinemodic: Ah that makes sense, I know I saw an eslint at some point in the past suggesting I use useMemo or something but I forget the exact situation and if it was a primitive value or not so then I just got in the habit of useMemo-ing things after that. I just tried removing this one and didn't see the eslint error.

PR #1270, opened 2021-11-23
feat(em) tweak logging

carolinemodic:
Tweaks two things about logging in election-manager
- Renamed recompting-tally and recomputed-tally IDs to recompute-tally-init and recompute-tally-complete for consistency with the other ID names
- Changes the logs when there are no CVR files found on the usb drive in the import modal from an error to a success log, this isn't really an error, and after doing a similar log in bsd I decided that both should still treat this case as a "success" and just log there were no files.


frontends/election-manager/src/components/import_cvrfiles_modal.tsx
```diff
@@ -197,27 +197,20 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
...
     } catch (err) {
       if (err.message.includes('ENOENT')) {
         // No files found
         setFoundFiles([]);
         setCurrentState(ModalState.INIT);
         await logger.log(LogEventId.CvrFilesReadFromUsb, currentUserType, {
-          message: 'No CVR files automatically found on usb drive',
-          disposition: 'failure',
+          message:
+            'No CVR files automatically found on usb drive. User allowed to import manually.',
```
eventualbuddha, line 212:
```suggestion
            'No CVR files automatically found on USB drive. User allowed to import manually.',
```

eventualbuddha (approved):

PR #1281, opened 2021-11-30
feat(em) add export logs modal

carolinemodic:
Adds a simple export logs modal to election manager. 

https://user-images.githubusercontent.com/14897017/144110628-f2752391-4beb-4d5f-a172-d75b21a7ac05.mov

@beausmith currently this just mimicks our other save dialogs but in the future will need to also export in other formats (primarily the CDF format, but we may want others) so would love any thoughts you have on how to extend the design to that. We might also want to allow to export other log files (syslog for example) but we can currently do this in tty2 and this would really only be something we want votingworks employees doing so its probably fine ignoring that for now. 

https://github.com/votingworks/vxsuite/issues/1260

Note: will require followup task in vxsuite-complete-system to update the permissions given to kiosk-browser.

Ideally this would be in the shared UI repo but that is blocked on Modal being there.

### DIFF
--- a/frontends/election-manager/src/components/export_logs_modal.test.tsx
+++ b/frontends/election-manager/src/components/export_logs_modal.test.tsx
@@ -1,0 +1,202 @@
+import React from 'react';
+
+import { fireEvent, waitFor } from '@testing-library/react';
+import {
+  advanceTimersAndPromises,
+  fakeKiosk,
+  fakeUsbDrive,
+} from '@votingworks/test-utils';
+import { usbstick } from '@votingworks/utils';
+
+import { LogEventId, Logger, LogSource } from '@votingworks/logging';
+import { ExportLogsModal } from './export_logs_modal';
+import { renderInAppContext } from '../../test/render_in_app_context';
+
+const { UsbDriveStatus } = usbstick;
+
+const fileSystemEntry: KioskBrowser.FileSystemEntry = {
+  name: 'file',
+  path: 'path',
+  type: 1,
+  size: 1,
+  atime: new Date(),
+  ctime: new Date(2021, 0, 1),
+  mtime: new Date(),
+};
+
+test('renders loading screen when usb drive is mounting or ejecting in export modal', () => {
+  const usbStatuses = [UsbDriveStatus.present, UsbDriveStatus.ejecting];
+
+  for (const status of usbStatuses) {
+    const closeFn = jest.fn();
+    const { getByText, unmount } = renderInAppContext(
+      <ExportLogsModal onClose={closeFn} />,
+      {
+        usbDriveStatus: status,
+      }
+    );
+    getByText('Loading');
+    unmount();
+  }
+});
+
+test('renders no log file found when usb is mounted but no log file on machine', async () => {
+  jest.useFakeTimers();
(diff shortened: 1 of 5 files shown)


frontends/election-manager/src/components/export_logs_modal.tsx
```diff
@@ -0,0 +1,283 @@
...
+enum ModalState {
+  ERROR = 'error',
+  SAVING = 'saving',
+  DONE = 'done',
+  INIT = 'init',
+}
+
+export function ExportLogsModal({ onClose }: Props): JSX.Element {
+  const { usbDriveStatus, currentUserSession, logger } = useContext(AppContext);
+  assert(currentUserSession); // TODO(auth) should this check for a specific user type
```
benadida, line 34: yes, in the future we'll want an admin session of some kind here, and we may want to reauth in case the card has been removed.

benadida (approved):
good for now, though I think we want to put this and probably other features on some admin / advanced screen that requires re-auth or different auth in the future.

frontends/election-manager/src/components/export_logs_modal.test.tsx
```diff
@@ -0,0 +1,202 @@
...
+    getByAltText('Insert USB Image');
+
+    fireEvent.click(getByText('Cancel'));
+    expect(closeFn).toHaveBeenCalled();
+
+    unmount();
+  }
+});
+
+test('renders save modal when usb is mounted but and log file on machine', async () => {
```
eventualbuddha, line 107:
```suggestion
test('renders save modal when usb is mounted and log file on machine', async () => {
```

frontends/election-manager/src/components/export_logs_modal.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+export interface Props {
+  onClose: () => void;
+}
+
+enum ModalState {
+  ERROR = 'error',
+  SAVING = 'saving',
+  DONE = 'done',
+  INIT = 'init',
```
eventualbuddha, line 29: Any reason for these to be shouty?

```suggestion
  Error = 'error',
  Saving = 'saving',
  Done = 'done',
  Init = 'init',
```

frontends/election-manager/src/components/export_logs_modal.tsx
```diff
@@ -0,0 +1,283 @@
...
+          <Prose>
+            <h1>Logs Saved</h1>
+            <p>
+              Log file successfully saved{' '}
+              {savedFilename !== '' && (
+                <span>
+                  as <strong>{savedFilename}</strong>
+                </span>
+              )}{' '}
+              directly on the inserted USB drive.
```
eventualbuddha, line 162: This is only true when you click "Save", not "Save As".

eventualbuddha (approved):

frontends/election-manager/src/components/export_logs_modal.tsx
carolinemodic: Yeah, but kiosk-browser still only has permissions to save on usb drives so its impossible to save it elsewhere (in production), it will not show the folder if you save it in a folder on the usb though so I'm going to remove the word "directly"

PR #1294, opened 2021-12-02
Export Logs in BSD

carolinemodic:
Copies the export logs modal from election-manager to BSD. I also realized that it will overwrite a file with an identical name so I added the current timestamp to the default filename in both versions of the dialog.


benadida (approved):

frontends/bsd/src/components/export_logs_modal.tsx
```diff
@@ -0,0 +1,288 @@
...
+          defaultPath: defaultFilename,
+        });
+
+        if (!fileWriter) {
+          throw new Error('could not begin download; no file was chosen');
+        }
+
+        await fileWriter.write(results);
+        filenameLocation = fileWriter.filename;
+        await fileWriter.end();
```
eventualbuddha, line 108: This feels like we're missing a `copyFile` method.

frontends/bsd/src/components/export_logs_modal.tsx
```diff
@@ -0,0 +1,288 @@
...
+      } else {
+        assert(typeof usbPath !== 'undefined');
+        const pathToFile = path.join(usbPath, defaultFilename);
+        assert(window.kiosk);
```
eventualbuddha, line 112: Why do we have to do this again?

eventualbuddha (approved):

PR #1296, opened 2021-12-02
fix(em): update on unmounted component

eventualbuddha:
`UnconfiguredScreen` sometimes would call `setIsLoading(false)` in the  `catch` clause of `updateStatus` after the component had already been unmounted. This commit fixes it by canceling the component's async operations on unmount.


benadida (approved):

benadida:
approved on code, but of course failing tests to review.

carolinemodic (approved):

eventualbuddha:
Unsure why cypress tests are failing. Not worth investigating right now.

PR #1299, opened 2021-12-04
feat(scan) adds logs to fujitsu scanner

carolinemodic:
Adds logs to the fujitsu scanner, @eventualbuddha curious to your thoughts on this approach of wrapping around the debug, I figured it would be super annoying to not be able to turn off these logs in module-scan, but they need to always print in production so this way in dev you can use the normal ways of controlling them with debug to turn on/off. 

Future work is to do a similar thing in the plustek class, and probably also other places of module-scan.

### DIFF
--- a/libs/logging/src/logger.ts
+++ b/libs/logging/src/logger.ts
@@ -26,7 +26,8 @@ export class Logger {
   async log(
     eventId: LogEventId,
     user: LoggingUserRole,
-    logData: LogData = {}
+    logData: LogData = {},
+    outerDebug?: debug.Debugger
   ): Promise<void> {
     const eventSpecificDetails = getDetailsForEventId(eventId);
     const {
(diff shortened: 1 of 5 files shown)


services/scan/src/scanners/fujitsu.ts
```diff
@@ -157,9 +200,14 @@ export class FujitsuScanner implements Scanner {
       endBatch: async (): Promise<void> => {
         if (!done) {
           done = true;
-          debug(
-            'scanimage [pid=%d] stopping scan by closing stdin',
-            scanimage.pid
+          void this.logger.log(
```
carolinemodic, line 203: I can't comment on the actual line but ... @eventualbuddha curious if you think I should wrap the debug on line 178 in a call to logger.log, I didn't really understand what was happening there well enough to know what the log should be so I just left it for now.

eventualbuddha: I don't think that needs to be in the user-facing logs. It's just saying that it's sending `\n\n` to the stdin of the `scanimage` process to trigger the scanning of two pages (one sheet), which works because `scanimage` was spawned with the `--batch-prompt` option.

libs/logging/src/logger.ts
```diff
@@ -43,6 +44,14 @@ export class Logger {
       disposition,
       ...additionalData,
     };
+    // If the caller is passing in a debug instance, and we are not in production log to the debugger rather then through the normal logging pipeline.
+    // This is to make logs more manageable in development, so a developer can toggle what logs to view with the normal debug namespaces.
+    /* istanbul ignore next - figure out how to test this */
+    if (outerDebug && process.env.NODE_ENV !== 'production') {
+      outerDebug(logLine);
```
eventualbuddha, line 51: I'm unsure how to feel about this. Part of me feels it should just be that the `logger` used is one that uses `debug` or doesn't depending on the environment, rather than mixing the two. I don't feel strongly enough to push back on it in this PR, but I do think we should watch how this works for us and change if it becomes unwieldy.

eventualbuddha (approved):

PR #1313, opened 2021-12-13
feat(logging) add method to convert to CDF format

carolinemodic:
Adds function to convert a log file in our base json format to the CDF format specified by: https://github.com/usnistgov/ElectionEventLogging


libs/logging/src/log_event_ids.ts
```diff
@@ -555,6 +557,18 @@ const ExportLogFileFound: LogDetails = {
...
+  eventId: LogEventId.LogConversionToCdfComplete,
+  eventType: LogEventType.UserAction,
+  documentationMessage:
+    'The user has converted the log file to a CDF format for export. Success or failure indicated by disposition.',
+};
+const LogConversionToCdf: LogDetails = {
+  eventId: LogEventId.LogConversionToCdf,
+  eventType: LogEventType.UserAction,
+  documentationMessage:
+    'Message during the conversion of logs to the CDF format. If it is an error the disposition will be failure.',
```
benadida, line 570: is this the start of CDF conversion? Maybe it should be more explicit if so?, at least in documentation message, possibly also in naming.

benadida (approved):

libs/logging/src/types.ts
```diff
@@ -41,5 +46,103 @@ export interface LogLine extends Dictionary<string> {
...
+  disposition: CdfEventDispositionType;
+  sequence: string;
+  timestamp: string;
+  type: LogEventType;
+  details?: string;
+  description?: string;
+  hash?: string;
+  otherDisposition?: string;
+  severity?: string;
+  userId: string;
```
eventualbuddha, line 129: These are `lowerCamelCase` whereas the [CDF JSON schema](https://github.com/usnistgov/ElectionEventLogging/blob/master/NIST_V1_election_event_logging.json) uses `UpperCamelCase`. So these won’t actually be compatible with something expecting CDF logging data, right?

carolinemodic: Yeah will fix, I got thrown off by this not matching the style-guide rules. I wish I had an example CDF file :'( 

Do you think from that that things like "Device" need to be prefixed with "ElectionLogging." as well?

libs/logging/src/logger.test.ts
```diff
@@ -101,3 +111,267 @@ test('logging from a client side app without sending window.kiosk does NOT log t
...
+  test('builds device and election info properly', () => {
+    const logger = new Logger(LogSource.VxAdminApp);
+    const cdfLogContent = logger.buildCDFLog(
+      electionMinimalExhaustiveSampleDefintion,
+      '',
+      '12machine34',
+      'thisisacodeversion',
+      'admin'
+    );
+    const cdfLog = JSON.parse(cdfLogContent);
```
eventualbuddha, line 131: Using `JSON.parse` here means TS treats `cdfLog` and everything derived from it as `any`, which means refactoring on the type this is supposed to have does not work. For example, I tried to rename `CdfDevice#id` to `CdfDevice#Id` and line 141 was not updated because TS doesn't know that `id` is the same as this one.

carolinemodic: do you think I should use zod here then? That felt like overkill to me just for the tests when we don't need to parse in the actual code but I don't feel strongly.

eventualbuddha: Why don't we need to parse it in the actual code?

carolinemodic: Well I guess if we end up needing to import logs then we will (this is in the VVSG spec but Monica seems to think that it doesn't make sense that we would need to do that and told me to ignore it for now), but otherwise we are never reading a CDF log file in the actual code, just converting to it as an output.

libs/logging/src/logger.ts
```diff
@@ -58,4 +87,92 @@ export class Logger {
...
+        message: 'The current application is not able to export logs.',
+        result: 'Log file not converted to CDF format.',
+        disposition: 'failure',
+      });
+      throw new Error('Can only export CDF logs from a frontend app.');
+    }
+
+    const allEvents: CdfEvent[] = [];
+    const logs = rawLogFileContents.split('\n').filter((l) => l !== '');
+    // eslint-disable-next-line no-debugger
```
eventualbuddha, line 109:
```suggestion
```

libs/logging/src/logger.ts
```diff
@@ -1,22 +1,51 @@
 import makeDebug from 'debug';
-import { Dictionary } from '@votingworks/types';
+import {
+  Dictionary,
+  ElectionDefinition,
+  safeParse,
+  safeParseJson,
+} from '@votingworks/types';
+import { z } from 'zod';
+import { strict as assert } from 'assert';
```
eventualbuddha, line 9:
```suggestion
import { assert } from '@votingworks/utils';
```

libs/logging/src/logger.ts
```diff
@@ -58,4 +87,92 @@ export class Logger {
...
+        disposition: 'failure',
+      });
+      throw new Error('Can only export CDF logs from a frontend app.');
+    }
+
+    const allEvents: CdfEvent[] = [];
+    const logs = rawLogFileContents.split('\n').filter((l) => l !== '');
+    // eslint-disable-next-line no-debugger
+    let idx = 0;
+    for (const log of logs) {
```
eventualbuddha, line 111:
```suggestion
    for (const [idx, log] of logs.entries()) {
```

Or, even better, stream the log lines.

carolinemodic:
updated to use autogen'd types library

Issue #1314, opened 2021-12-13
replace `react-scripts` with `vite`


eventualbuddha:
## Why?
- **It's fast.** Vite (pronounced "veet") uses [esbuild](https://esbuild.github.io), which is potentially orders of magnitude faster than with babel + webpack. Add TypeScript type checking on top of that and the difference is even more huge.
- **It's lightweight.** CRA includes the kitchen sink and battery-includes everything. This is good for getting started, but bad for upgrades and customization. Vite includes only what you need for HMR development and fast builds. Testing/linting is separate.

## The Plan
[Proposal](https://docs.google.com/document/d/1GxHfXOruz_fmBlhZpWPGqbJGd1TJSzvgTfgkZrVWghI/edit#)
I'd like to evaluate it for use with vxsuite. I've done some initial exploration and it seems doable, but will require some restructuring. For example:
- probably shouldn't use NodeJS builtins polyfills like for `assert` or `crypto`. there are ways to polyfill but we hopefully won't need to
- stop using TS project references, a la https://github.com/mmazzarolo/tangerine-monorepo
- more that I haven't uncovered yet

eventualbuddha:
#1669 was related to this work.

eventualbuddha:
I'm going to count this as done. We no longer use `react-scripts`, but we do still have some of the leftovers from `eject`ing. We can clean that stuff up later.

PR #1316, opened 2021-12-13
feat(codemods): add `use_utils_assert` codemod

eventualbuddha:
Transforms uses of `assert` that can be converted automatically to use `@votingworks/utils` instead.

Refs #1314

### DIFF
--- a/codemods/src/use_utils_assert.ts
+++ b/codemods/src/use_utils_assert.ts
@@ -1,0 +1,103 @@
+/**
+ * Changes imports of `assert` to use `@votingworks/utils` instead.
+ *
+ * ```ts
+ * // broken
+ * import assert from 'assert';
+ *
+ * // fixed
+ * import { assert } from '@votingworks/utils';
+ *
+ *
+ * // broken
+ * import { strict as assert } from 'assert';
+ *
+ * // fixed
+ * import { assert } from '@votingworks/utils';
+ *
+ *
+ * // broken
+ * import { ok } from 'assert';
+ *
+ * // fixed
+ * import { assert } from '@votingworks/utils';
+ * ```
+ *
+ * Run it like so:
+ *
+ * ```sh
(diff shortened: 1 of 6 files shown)


codemods/src/use_utils_assert.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Run it like so:
+ *
+ * ```sh
+ * $ pnpx codemod -p codemods/src/use_utils_assert.ts --printer prettier frontends/ libs/ services/
+ * ```
+ */
+
+import { PluginItem, NodePath } from '@babel/core';
+import * as t from '@babel/types';
+import assert from 'assert';
```
carolinemodic, line 35: how meta

codemods/src/use_utils_assert.ts
```diff
@@ -0,0 +1,103 @@
...
+ * ```ts
+ * // broken
+ * import assert from 'assert';
+ *
+ * // fixed
+ * import { assert } from '@votingworks/utils';
+ *
+ *
+ * // broken
+ * import { strict as assert } from 'assert';
```
carolinemodic, line 13: The test cases imply to me that this case can't be fixed by the codemod so I'm confused by it being used as an example here, am I reading the tests wrong? That said I think this is the most common way we import from assert (or at least what I usually do) so it seems worth supporting if we can.

carolinemodic (approved):

codemods/src/use_utils_assert.ts
eventualbuddha: This example is supported. I specifically allow `strict` and `ok` named imports.

codemods/src/use_utils_assert.ts
eventualbuddha: Heh, yeah. Fun stuff 😉

Issue #1238, opened 2021-11-17
Investigate adopting CDF semantics/structure


mcchilders:
**Task:** Investigate whether adopting part/all of the semantics/structure of the election CDFs would be worthwhile, given that they already contain all the attributes/relationships that are needed to accommodate basically any contest, nationwide. Timeboxed, preferably.

**Goal:** Should we adopt any pieces of the CDF semantics/structure internally? If so, which parts? If not, why not?
Bonus points: start mapping our current data structure to the CDFs, so that we have a head start on translation.

I'll paste this from the notes on the results reporting CDF, which reflects the primary value I think we'd get out of adopting it:

 _The JSON and XML schemas associated with this specification are derived from a UML model that defines the types, structure, and interrelationships of geopolitical geography across the U.S. This model was designed to accommodate multiple types of contests and their many variations, and to provide the capability to report on these contests from higher aggregate levels down to very fine levels of detail, including:_

_Reporting by precinct, splits of precincts, voting centers, and other geographic localities;_
_Reporting by various ballot type, e.g., absentee, election day, early voting; and_
_Reporting by by finer levels of detail such as device type and specific devices, batch, and dropbox._ 

CDFs are here (the election reporting and CVR CDFs are the immediately relevant ones):
https://www.nist.gov/itl/voting/interoperability

Things to keep in mind:
- We know we have to be able to intake data/export data in these formats for 2.0 cert, so we will have to map/translate these fields to our existing ones regardless. The reasoning behind investigating whether it makes sense to adopt some/all of these internally is that our current representation is very MS-driven, and doesn't contain everything we'll need to work nationally (either data or relationships between data), whereas the CDF already does - and it was crafted to handle edge cases without requiring a ton of cruft (see the "other" options on almost every field). So, making sure we at least 1) represent the same data and 2) have the same relationships/hierarchies between that data in our system allows us to take advantage of all that work and "future-proof" our system before it goes into cert.
- The potential downside of doing all this now is that there's a chance we won't need/want most of this for a long time, until we have more clients, if ever. However, if we're going to certify our system, it would be a bummer to have to recert before we could serve a client that does need some or all of these things.
- There are a number of 2.0 certification requirements that reflect "things all the major voting systems can do" that we don't have yet and will have to add, that the CDFs *do already.* So there are some freebies there (in terms of not having to figure out the best way to model/represent that data in relationship to everything else, not in terms of implementation.)
[...]

PR #1318, opened 2021-12-15
feat(cdf): add schema builder and logging CDF

eventualbuddha:
Adds a new library to convert CDF schemas written with XSD XML (which themselves are derived from a UMD representation) into TypeScript enums, interfaces, and Zod schemas.

Uses this library to build types for the `ElectionEventLogging` CDF specification and keep it up to date as desired.

Refs #1238


.circleci/config.yml
```diff
@@ -191,6 +191,42 @@ jobs:
...
+      - run:
+          name: Lint
+          command: |
+            pnpm --dir libs/cdf-schema-builder lint
+      - run:
+          name: Test
+          command: |
+            pnpm --dir libs/cdf-schema-builder test
+
+  test-libs-cdf-types-election-event-logging:
```
carolinemodic, line 212: curious why you are making this specific to the file type as opposed to a more general cdf types library? 

It would be more annoying for us but I could see an argument for them to be cdf-specific, and have it live as a separate repo outside of vxsuite as it could be theoretically useful to someone else. But I'm not sure its worth the overhead of that.

carolinemodic (approved):

.circleci/config.yml
carolinemodic: After I wrote this I realized there could be type name conflicts between the generated files and its probably easier to avoid that this way.

eventualbuddha: That's part of it. NIST breaks it into multiple parts (i.e. logging, reporting, etc) so I followed suit.

PR #1319, opened 2021-12-16
build(all): upgrade TS to 4.4

eventualbuddha:
The primary change that affects us is that `catch` error bindings are now `unknown`, not `any`. To deal with that I am asserting that the errors are of type `Error`.

This also updates the `@typescript-eslint` packages to go along with the update.

### DIFF
--- a/libs/ui/src/hooks/use_usb_drive.ts
+++ b/libs/ui/src/hooks/use_usb_drive.ts
@@ -1,5 +1,5 @@
 import useInterval from 'use-interval';
-import { usbstick } from '@votingworks/utils';
+import { assert, usbstick } from '@votingworks/utils';
 import { useCallback, useState } from 'react';
 import makeDebug from 'debug';
 import { LogEventId, Logger, LoggingUserRole } from '@votingworks/logging';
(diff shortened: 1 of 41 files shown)


libs/ui/src/hooks/use_usb_drive.ts
```diff
@@ -55,11 +55,11 @@ export function useUsbDrive({ logger }: UsbDriveProps): UsbDrive {
         await logger.log(LogEventId.UsbDriveEjected, currentUser, {
           disposition: 'failure',
           message: 'USB Drive failed when attempting to eject.',
-          error: error.message,
+          error: (error as Error).message,
```
jonahkagan, line 58: why a diff pattern here?

jonahkagan (approved):

libs/cdf-types-election-event-logging/data/err.xml
```diff
@@ -0,0 +1,2348 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
eventualbuddha: Nope.

libs/ui/src/hooks/use_usb_drive.ts
eventualbuddha: Had done this everywhere, then changed. I'll fix.

eventualbuddha:
This has a number of problems that are difficult to reconcile. Will try again later.

PR #1320, opened 2021-12-17
feat(utils): add various cryptography tools

eventualbuddha:
Some stuff runs in node and some in the browser, making supporting both a bit annoying and requiring polyfills when we import e.g. `crypto` in the browser. This consolidates SHA256, base64, and seeded/unseeded random number generation in one place.

Because this touches random/crypto stuff in a non-trivial way I'd like to have an approval from @benadida or @umbernhard on that part of it. Anyone else can review for general coding correctness.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -194,7 +194,7 @@ importers:
       dompurify: 2.2.6
       fetch-mock: 9.11.0
       history: 4.10.1
-      http-proxy-middleware: 1.0.6
+      http-proxy-middleware: 1.0.6_debug@4.3.2
       lodash.camelcase: 4.3.0
       luxon: 1.26.0
       mockdate: 3.0.2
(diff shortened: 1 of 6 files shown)


pnpm-lock.yaml
```diff
@@ -194,7 +194,7 @@ importers:
       dompurify: 2.2.6
       fetch-mock: 9.11.0
       history: 4.10.1
-      http-proxy-middleware: 1.0.6
+      http-proxy-middleware: 1.0.6_debug@4.3.2
```
benadida, line 197: did you mean to change the version to a `debug` one?

benadida (commented):
What's the reasoning for doing this byte by byte in the generator? There's a chance, depending on implementation of the internal randomness generation, that this could be much slower.

eventualbuddha:
> What's the reasoning for doing this byte by byte in the generator? There's a chance, depending on implementation of the internal randomness generation, that this could be much slower.

That’s just what ended up being the easiest way to do it. Backing up a bit, my goal was to allow what I did in #1321 where I could set up tests in such a way that the random values generated would be deterministic. However, I realized shortly after I woke up this morning that this _also_ applies in production with the only difference being that it’ll differ on each instantiating of `AppRoot`. This is still bad since it means every ballot ID would be the same.

I’m open to suggestions on how to achieve the test consistency goal while still preserving real randomness in production. This clearly isn’t quite right.

PR #1321, opened 2021-12-17
refactor(bmd): use `randomBase64` from `/utils`

eventualbuddha:
And provides a deterministic random generator in tests so we can take snapshots without worrying about the values changing from one run to the next.

Builds on #1320


frontends/bmd/src/app_root.tsx
```diff
@@ -1397,7 +1399,9 @@ export function AppRoot({
                 votes: votes ?? blankBallotVotes,
               }}
             >
-              <Ballot />
+              <RandomContext.Provider value={{ generator: randomGenerator() }}>
```
eventualbuddha, line 1402: This ensures we're using real random values from the browser's `crypto` API in real code.

frontends/bmd/test/test_utils.tsx
```diff
@@ -55,6 +57,7 @@ export function render(
     forceSaveVote = jest.fn(),
     userSettings = { textSize: GLOBALS.TEXT_SIZE as TextSizeSetting },
     votes = {},
+    random = mersenne(0),
```
eventualbuddha, line 60: Creates a seeded generator in tests that always generates the same values. So for example, every ballot ID generated in test will be the same.

carolinemodic (approved):

frontends/bmd/src/app_root.tsx
jonahkagan: Why use a context instead of just a module/function mock?

PR #1322, opened 2021-12-20
feat(logging) add helper to build documentation file

carolinemodic:
Adds helper to build documentation file for CDF output of logs. This could be pre-generated per device but since there is a generated time in the output I'm not totally sure if that makes sense, and its not particularly time consuming to just generate on demand ultimately.


libs/logging/src/log_source.ts
```diff
@@ -0,0 +1,24 @@
+export enum LogSource {
+  System = 'system',
+  VxAdminApp = 'vx-admin',
+  VxAdminServer = 'vx-admin-server',
+  VxBatchScanApp = 'vx-batch-scan',
+  VxBatchScanSever = 'vx-batch-scan-server',
```
eventualbuddha, line 6: I see that this was just moved rather than added in this commit, but I wonder if it should be using this?
```suggestion
  VxBatchScanFrontend = 'vx-batch-scan-frontend',
  VxBatchScanService = 'vx-batch-scan-service',
```

libs/logging/src/logger.test.ts
```diff
@@ -396,3 +398,89 @@ describe('test cdf conversion', () => {
...
+      'VxAdmin 1.0',
+      'codeversion'
+    );
+    const structuredDataResult = safeParseJson(
+      cdfDocumentationContent,
+      ElectionEventLogDocumentationSchema
+    );
+    expect(structuredDataResult.isOk()).toBeTruthy();
+    const structuredData = structuredDataResult.ok();
+    assert(structuredData);
```
eventualbuddha, line 417: This will throw the error in the case that the result is not ok, whereas as written it will just assert `undefined`.
```suggestion
    const structuredData = structuredDataResult.unsafeUnwrap();
```

eventualbuddha (approved):

PR #1323, opened 2021-12-23
feat(eslint-plugin-vx): upgrade TS & related deps

eventualbuddha:
This is a smaller version of #1319 that only upgrades the `eslint-plugin-vx` package's dependencies. This one is tricky because typescript isn't just a build-time dependency.


eventualbuddha:
This has again become a large change due to the sprawl of interrelated dependencies. Mainly, upgrading everything except `react-scripts` means the versions of everything `react-scripts` wants to use differ from the ones we have installed. Trying to upgrade to react-scripts v5 runs into another problem: it doesn't polyfill nodejs modules like `crypto` by default anymore, and it wasn't clear to me how to do configure it to do so. In any event, this became too large so I'm closing again and will try again with an even more focused change.

PR #1324, opened 2021-12-23
feat(eslint-plugin-vx): upgrade `@typescript-eslint/*` deps

eventualbuddha:
They changed how they map a TS AST to an ESTree AST, and also the metadata for a rule. Another even more restricted attempt at #1323.


jonahkagan (approved):
Looks good assuming you work out the test failure

PR #1327, opened 2022-01-05
fix(em) fixes bug when exporting ballot package

carolinemodic:
In https://github.com/votingworks/vxsuite/commit/ae047790b7216c1a6d84b8cc1e128215c89fded1 we migrated to a new version of assert and accidently changed this assert from an assert.notEqual to an assert checking equality. Noticed that I couldn't export any ballot packages because of this throwing and then found the mis-translation from that PR. This updates it to check that it is not equal as intended. 

cc @eventualbuddha


jonahkagan (approved):
Do you think there are any other asserts that got flipped?

carolinemodic:
I did another skim through that PR and it all looked fine, I forget if there were other similar conversion PRs but I'll take a look.

PR #1328, opened 2022-01-07
feat(em) add advanced screen with log export as cdf option

carolinemodic:
Adds an advanced screen to election manager with options to export the logs file and export the logs file in CDF format. I include the advanced screen where there is no election but don't allow the log export in CDF as the CDF format requires an election defined. 

Video: 

https://user-images.githubusercontent.com/14897017/148472024-5233a760-25b5-44a5-bcd3-91c6f43507da.mov

Will migrate the new version of ExportLogsModal to bsd in a followup PR (note: this is not in the shared components library as it is blocked on Modal being there).

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -66,7 +66,10 @@ jest.mock('./components/hand_marked_paper_ballot');
 beforeEach(() => {
   // we don't care about network errors logged to the console, they're crowding things
   jest.spyOn(console, 'error').mockImplementation(() => {}); // eslint-disable-line @typescript-eslint/no-empty-function
-
+  Object.defineProperty(window, 'location', {
+    writable: true,
+    value: { assign: jest.fn() },
+  });
   window.location.href = '/';
   const mockKiosk = fakeKiosk();
   window.kiosk = mockKiosk;
@@ -138,7 +141,6 @@ test('create election works', async () => {
   const { getByText, getAllByText, queryAllByText, getByTestId } = render(
     <App card={card} hardware={hardware} />
   );
-
   await screen.findByText('Create New Election Definition');
   fireEvent.click(getByText('Create New Election Definition'));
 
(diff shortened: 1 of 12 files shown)


frontends/election-manager/src/app.test.tsx
```diff
@@ -66,7 +66,10 @@ jest.mock('./components/hand_marked_paper_ballot');
 beforeEach(() => {
   // we don't care about network errors logged to the console, they're crowding things
   jest.spyOn(console, 'error').mockImplementation(() => {}); // eslint-disable-line @typescript-eslint/no-empty-function
-
+  Object.defineProperty(window, 'location', {
```
carolinemodic, line 69: The line do reset the window location wasn't actually doing anything without this. Returning a Switch instead of just UnconfiguredScreen in ElectionManager made it so that the URL from the previous test was getting incorrectly carried over to the next test without having this reset working. In manual testing I didn't notice any difference in routing.

libs/logging/src/types.ts
```diff
@@ -5,8 +5,9 @@ import {
 } from '@votingworks/types';
 import { z } from 'zod';
 import { DeviceType } from '@votingworks/cdf-types-election-event-logging';
-import { LogEventId, LogEventType } from '.';
+import { LogEventId } from './log_event_ids';
```
carolinemodic, line 8: There was a weird bug without these changes where the z.nativeEnum call was created a enum parser with no possible enum values defined. For some reason this only happened when importing to election-manager and not when the tests in libs/logging where running. I'm guessing it was some kind of circular dependency problem.

frontends/election-manager/src/components/export_logs_modal.tsx
```diff
@@ -81,17 +89,38 @@ export function ExportLogsModal({ onClose }: Props): JSX.Element {
     }
     void checkLogFile();
   }, [currentUserSession, logger]);
-  const defaultFilename = generateLogFilename('vx-logs');
+  const defaultFilename = generateLogFilename('vx-logs', logFileType);
```
benadida, line 92: Maybe worth having the filename say "cdf" somewhere, in addition to `.json`.

libs/utils/src/filenames.ts
```diff
@@ -246,8 +252,17 @@ export function generateBatchResultsDefaultFilename(
...
   logFileName: string,
+  fileType: LogFileType,
   time: Date = new Date()
 ): string {
   const timeInformation = moment(time).format(TIME_FORMAT_STRING);
-  return `${logFileName}${SUBSECTION_SEPARATOR}${timeInformation}.log`;
+  switch (fileType) {
+    case LogFileType.Raw:
+      return `${logFileName}${SUBSECTION_SEPARATOR}${timeInformation}.log`;
+    case LogFileType.Cdf:
```
benadida, line 262: probably this is where the `-cdf` or something would live.

benadida (approved):

beausmith (approved):
UX/UI/copy looks great!

PR #1332, opened 2022-01-11
feat(logging) Scripts to generate documentation as cdf and markdown

carolinemodic:
When I build the code for the documentation for logs in CDF format before I assumed it would be included in the CDF export from the apps. However after talking to Monica I realized this is something we will just provide separately for each machine. So I moved that code a bit and made a script to generate those files when we need them. I also added the ability to generate documentation in markdown format, primarily for our internal use but it may be useful more broadly. That generated file is stored in the repo for reference and there is a test to make sure it stays up to date.

### DIFF
--- a/libs/logging/VotingWorksLoggingDocumentation.md
+++ b/libs/logging/VotingWorksLoggingDocumentation.md
@@ -1,0 +1,320 @@
+
+# VotingWorks Log Documentation
+## Event Types
+Types are logged with each log line to categorize the log.
+### user-action
+**Description:** A log that results from a user taking an action, i.e. an election admin uploading an election definition to a machine.
+### application-status
+**Description:** Status update or message that the application took without user action.
+### system-action
+**Description:** A log that results from the system taking some action, i.e. the machine booting.
+### system-status
+**Description:** A log that results from the system updating on the status of a process it is running, i.e. completion of machine shutdown or boot.
+### application-action
+**Description:** Action taken by the votingworks application automatically when a certain condition is met. Example: When a new USB drive is detected, the application will automatically mount it.
+## Event IDs
+IDs are logged with each log to identify the log being written.
+### election-configured
+**Type:** [user-action](#user-action)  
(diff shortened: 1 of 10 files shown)


libs/logging/VotingWorksLoggingDocumentation.md
```diff
@@ -0,0 +1,320 @@
+
+# VotingWorks Log Documentation
+## Event Types
+Types are logged with each log line to categorize the log.
+  ### user-action
```
benadida, line 5: some extraneous spaces here for the first title?

libs/logging/VotingWorksLoggingDocumentation.md
```diff
@@ -0,0 +1,320 @@
...
+**Description:** Status update or message that the application took without user action.
+### system-action
+**Description:** A log that results from the system taking some action, i.e. the machine booting.
+### system-status
+**Description:** A log that results from the system updating on the status of a process it is running, i.e. completion of machine shutdown or boot.
+### application-action
+**Description:** Action taken by the votingworks application automatically when a certain condition is met. Example: When a new USB drive is detected, the application will automatically mount it.
+## Event IDs
+IDs are logged with each log to identify the log being written.
+  ### election-configured
```
benadida, line 17: yep, looks like it's a pattern.

libs/logging/src/log_documentation.ts
```diff
@@ -0,0 +1,83 @@
...
+    LogEventType
+  ).map((eventType) => getDocumentationForEventType(eventType));
+  const allEventIdsForDevice: LogDetails[] = Object.values(LogEventId).map(
+    (eventId) => getDetailsForEventId(eventId)
+  );
+  return `
+# VotingWorks Log Documentation
+## Event Types
+Types are logged with each log line to categorize the log.
+  ${allEventTypes
```
benadida, line 25: this is probably where the extra spaces are coming from.

benadida (approved):

Issue #1330, opened 2022-01-10
Script to help Drew QA card readers


mcchilders:
A starting point for this is here:
https://github.com/votingworks/vxsuite/blob/main/services/smartcards/testCardReader.py

Desired user experience:
* start the script
* be prompted to connect a card reader
* connect a card reader
* if no card in card reader, prompted to insert a card
* start the test of reading / writing various blocks of data
* declare result, then prompt to disconnect card reader
* once disconnected, prompt to connect card reader and start again.

This way, the process of testing a bunch of card readers can be purely focused on plugging / unplugging readers and cards, never touching the keyboard in between.

PR #1348, opened 2022-01-13
feat(smartcard): Smartcard QA script

jonahkagan:
Task: #1330 

Beef up the existing card test script to QA multiple cards in a row.

### DIFF
--- a/services/smartcards/testCardReader.py
+++ b/services/smartcards/testCardReader.py
@@ -1,36 +1,100 @@
-
-import json, hashlib, secrets, time, sys
+import re
+import json
+import hashlib
+import secrets
+import time
+import sys
 from smartcards.core import CardInterface
+import smartcard
 
-# generate random bytes for short value and long value
-short_bytes = secrets.token_bytes(61)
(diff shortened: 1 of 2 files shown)


services/smartcards/testCardReader.py
```diff
@@ -1,36 +1,100 @@
...
+    print("Card reader connected")
+
+
+def wait_for_card():
+    if not CardInterface.card:
+        print("Insert card")
+    while True:
+        time.sleep(REFRESH_INTERVAL)
+        if CardInterface.card:
+            break
```
benadida, line 29: Some cards are write-protected with a bit that we set and look for, and in that case the tests will fail. This minimal change just exits if the card is write-protected:

```suggestion
            if not CardInterface.card.write_enabled:
                print("card is write-protected, please use a different card.")
                sys.exit(1)
            break
```

benadida (approved):

services/smartcards/testCardReader.py
jonahkagan: added 8a0b05c7

Issue #1349, opened 2022-01-13
When a smart card is inserted incorrectly, tell the user


benadida:
## Summary

When a smartcard is inserted into a card reader on any of our devices, if it's inserted in the wrong orientation, the system behaves as if no card was inserted at all, thus giving the user no visible feedback of the action or how to correct it.

As it turns out, we can tell at a low-level that a card was inserted but the chip cannot be powered up, which is almost certainly because the card was inserted in the wrong orientation. It could also happen if a non-chip card were inserted, say an old mag-stripe card, but that scenario seems unlikely and it's not the end of the world if the error message is not perfect in that case.

## Requirements

On precinct-scanner where it's most urgent, when a card is inserted and the chip cannot be powered up, display an error screen that says:

```
Card cannot be read.
Make sure the chip on the card is facing away from you.
```

On BMD, opposite message:

```
Card cannot be read.
Make sure the chip on the card is facing you.
```

## Non-Requirements

No need to present a different error message if they are using a completely incorrect card.

## Tasks

* [x] add appropriate information in the smartcard service
* [x] show the error screen in precinct-scanner
* [x] show the error screen in BMD

beausmith:
Why not use more simple and direct language for all the screens:

> Card is Backwards
> Remove the card, rotate, and insert again.

PR #1351, opened 2022-01-18
feat(smartcards): Detect and expose card connection errors

jonahkagan:
Task: #1349 

When a card is inserted backwards (or is otherwise unreadable), a connection error occurs. We catch that error and expose that it happened in the smartcards API via a new `status` field.

We also update consumers of the API to consume the new field.

### DIFF
--- a/services/smartcards/smartcards/core.py
+++ b/services/smartcards/smartcards/core.py
@@ -8,7 +8,7 @@ from flask import Flask, send_from_directory, request
 from smartcard.System import readers
 from smartcard.util import toHexString, toASCIIBytes, toASCIIString
 
-from .card import VXCardObserver
+from .card import CardStatus, VXCardObserver
 from .mockcard import MockCard
 
 MockInstance = MockCard().update_from_environ()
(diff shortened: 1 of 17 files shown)


services/smartcards/smartcards/core.py
```diff
@@ -23,8 +23,12 @@ def card_reader():
     is_connected = CardInterface.is_reader_connected()
     return json.dumps({"connected": is_connected})
 
+
 @app.route('/card/read')
 def card_read():
+    if CardInterface.has_connection_error():
+        return json.dumps({"present": True, "connectionError": True})
```
benadida, line 30: I'm wondering if we might want to not set `present` to `True` in this case, and constrain the semantics of that field to be "there's a readable card present", not just "there's a piece of plastic present.

benadida (commented):
This looks great! I have a question for consideration, but I'm comfortable moving forward either way and I think my question might be answered more easily once we look at consumers of the API.

services/smartcards/smartcards/core.py
eventualbuddha: Check out the relevant types on the client side: https://github.com/votingworks/vxsuite/blob/4d3dbb213d24c2f362ab280f027701c225dcdc33/libs/utils/src/types.ts#L100-L166

I'd probably aim toward what @benadida said and use `present: false` with another optional field that indicates a card connection error.

jonahkagan: Ok thanks, that makes sense.

I considered adding a `status` field to the API to indicate the disjoint states a card could be in (e.g. `present`, `no_card`, `error`), but decided that it would probably be a large change for consumers of the API. But it's worth noting that now we have one of those weird APIs where there are multiple flags that can only pair up in certain ways. Just something to think about for API design in the future - using a flag is not the most extensible strategy.

jonahkagan:
Ok this PR now updates the smartcards service API and all of its consumers. I'm curious to know:
- How much can I rely on existing test coverage to prove to myself that I didn't break anything?
- Are there any new tests you think I need to add that I didn't add already?

jonahkagan:
Ok wait now tests are failing that were passing before. Please hold!

jonahkagan:
Ok, just a typo. Fixed it. Ready for review!

benadida (approved):
lgtm, this is definitely the kind of change that I recommend doing a good bit of manual testing on with actual card reader and cards. Assuming you've done that, good to go.

PR #1353, opened 2022-01-18
fix(smartcards): Clear card correctly in testCardReader.py

jonahkagan:
The card module expects that the short value isn't an empty string when writing a long value, which was causing the testCardReader.py script to crash. I considered changing that assumption, but then I looked at clearCard.py, and realized that "clearing" a card according to this script means writing an empty object, not an empty string. So I updated testCardReader.py accordingly.

I also changed it to override write protection automatically, since if the script fails at any point after writing a long value, it will have write-protected the card, and you won't be able to run the script again with that card unless you override write protection.


benadida (approved):
I'm n

PR #1355, opened 2022-01-19
feat(precinct-scanner): Show message when card inserted backwards

jonahkagan:
- Update the useSmartcard hook to consume the new smartcard status
- Use the status to determine when to show a new screen with a message about the card being backwards

TODO: write tests

### DIFF
--- a/libs/utils/src/types.ts
+++ b/libs/utils/src/types.ts
@@ -97,26 +97,27 @@ export interface Storage {
   clear(): Promise<void>;
 }
 
-export interface CardAbsentApi {
-  present: false;
+export interface CardApiNotReady {
+  status: 'no_card' | 'error';
 }
-export const CardAbsentApiSchema: z.ZodSchema<CardAbsentApi> = z.object({
-  present: z.literal(false),
-});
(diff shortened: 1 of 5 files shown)


libs/utils/src/types.ts
```diff
@@ -97,26 +97,27 @@ export interface Storage {
   clear(): Promise<void>;
 }
 
-export interface CardAbsentApi {
-  present: false;
+export interface CardApiNotReady {
+  status: 'no_card' | 'error';
```
benadida, line 101: suggest something slightly more descriptive than `error`, something specific to the card (as opposed to, say, the reader itself, which we might want to add later.) Something like `error_card_unreadable` or something like it.

benadida (approved):
one small suggestion but overall this is the right path.

libs/utils/src/types.ts
jonahkagan: I was thinking we could add an `errorType` field if we wanted to surface multiple errors later.

benadida: I can live with that!

PR #1356, opened 2022-01-20
chore(smartcards): Add format/check-format commands to Makefile

jonahkagan:
Run `make check-format` in CI to ensure that the code stays formatted.
Ideally this would occur in a pre-commit hook, but this is a simpler
change for now.


eventualbuddha (approved):
Yay 😁

PR #1359, opened 2022-01-21
feat(ps): guide poll workers after open/close polls

beausmith:
This screen share shows the flow: open polls, scan a ballot, close polls, export results to usb…
https://user-images.githubusercontent.com/28444/150447871-f6d0ad0f-805e-45c7-995e-92fbdeddf43c.mov

### DIFF
--- a/frontends/precinct-scanner/src/app.test.tsx
+++ b/frontends/precinct-scanner/src/app.test.tsx
@@ -269,13 +269,17 @@ test('admin and pollworker configuration', async () => {
   fireEvent.click(await screen.findByText('Open Polls for All Precincts'));
   await screen.findByText('Print Report and Open Polls');
   fireEvent.click(await screen.findByText('Cancel'));
+
+  // Disconnect Printer to test saving report to card
   await act(async () => {
     await hardware.setPrinterConnected(false);
   });
   fireEvent.click(await screen.findByText('Open Polls for All Precincts'));
   fireEvent.click(await screen.findByText('Save Report and Open Polls'));
   await screen.findByText('Saving to Card');
-  await screen.findByText('Close Polls for All Precincts');
+  await screen.findByText(
+    'Insert the poll worker card into a VxMark to print the report.'
+  );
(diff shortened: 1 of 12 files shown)


beausmith:
I'll get to the tests soon! But please review for general idea.

mattroe:
Most important change IMO is prompting pollworkers with a green button to export usb immediately after closing. 

One aspect that isn't obvious from testing the current flow with election commissioners yesterday is that there is no indication you need to take the pollworker card to VxMark to actually print the reports. I don't think we should treat this as self-evident and feel like an explicit cue to VxMark after opening/closing would be valuable.

ginvdr:
Comments:
1. Open Polls Screen - turn Open Polls for All Precincts button green
2. Polls Close Screen - turn Close Polls for All Precincts button green
3. On USB Drive Screen - turn Export button green (do this in both places)
4. Polls are Closed Screen - after they export the CVRs, turn the Export Results to USB Drive button gray?  Or some other indicator that they've completed it

CVR export seems much easier now. Agree with Roe that a prompt to remove the card and take to VxMark is probably helpful.

benadida:
@beausmith important to point out that the flow you're showing is when the printer is connected. In the setups @ginvdr and @mattroe are thinking about (Ms), there's no printer connected to VxScan, we're writing data to the card for VxMark printing. So worth seeing how that flow goes and in particular how we prompt sending.

Otherwise, I like these changes and agree with Ginny's comments.

beausmith:
Thx. For all this feedback. Will incorporate and post another video today.

eventualbuddha (approved):
Approving for code style etc. Others can take the business logic as it's already being discussed in this PR.

eventualbuddha:
[The warning in the precinct-scanner tests](https://app.circleci.com/pipelines/github/votingworks/vxsuite/2939/workflows/a1626685-9dce-47bb-87cf-a7b97bf67e3a/jobs/47565?invite=true#step-108-35) should be fixed by #1390.

beausmith:
Merged #1390 to main. Rebased and waiting on tests.

PR #1364, opened 2022-01-26
feat(bmd): Card backwards screen

jonahkagan:
Task: #1349

### DIFF
--- a/frontends/bmd/src/pages/card_error_screen.tsx
+++ b/frontends/bmd/src/pages/card_error_screen.tsx
@@ -1,0 +1,32 @@
+import React, { useEffect } from 'react';
+
+import { Main, MainChild } from '@votingworks/ui';
+
+import { Prose } from '../components/prose';
+import { Screen } from '../components/screen';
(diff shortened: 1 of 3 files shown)


frontends/bmd/src/pages/card_error_screen.tsx
```diff
@@ -0,0 +1,32 @@
...
+  useEffect(triggerAudioFocus, []);
+
+  return (
+    <Screen white>
+      <Main>
+        <MainChild center>
+          <Prose textCenter id="audiofocus">
+            <h1>Card is backwards</h1>
+            <p>Remove the card, turn it around, and insert it again.</p>
+          </Prose>
```
jonahkagan, line 27: I followed the screen pattern of bmd here, which was different from precinct-scanner. I also didn't include a graphic since the graphics didn't seem to be set up in bmd like they are in precinct-scanner, and the other bmd error screens don't have graphics.

carolinemodic (approved):

Issue #1363, opened 2022-01-26
update product naming in product


benadida:
- [x] Election Manager becomes VxAdmin
- [x] Ballot Scanner becomes VxCentralScan
- [x] VxMark stays VxMark, and notably no longer says "VxMark + VxPrint", just says VxMark. If there's no attached printer (so the underlying mode is VxMark only), then it might say "VxMark with standalone printer."
- [x] Precinct scanner is VxScan
- [x] bas remains VxEncode
- [x] bmd in print-only mode remains VxPrint.

carolinemodic:
FYI @jonahkagan I tried to use the new names when setting up the logging stuff but predicted the wrong names for the central/precinct scanner: https://github.com/votingworks/vxsuite/blob/main/libs/logging/src/log_source.ts

carolinemodic:
Oh you already saw that haha nevermind

PR #1367, opened 2022-01-26
feat(all): Update Batch Scanner name to VxCentralScan

jonahkagan:
Task: #1363 

Notably, I also updated the logs, which were using the outdated name
VxBatch.

### DIFF
--- a/frontends/election-manager/src/components/export_election_ballot_package_modal_button.tsx
+++ b/frontends/election-manager/src/components/export_election_ballot_package_modal_button.tsx
@@ -343,7 +343,7 @@ export function ExportElectionBallotPackageModalButton(): JSX.Element {
           <h1>Download Complete</h1>
           <p>
             You may now eject the USB drive. Use the exported ballot package on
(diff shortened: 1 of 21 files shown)


frontends/election-manager/src/components/export_election_ballot_package_modal_button.tsx
```diff
@@ -343,7 +343,7 @@ export function ExportElectionBallotPackageModalButton(): JSX.Element {
           <h1>Download Complete</h1>
           <p>
             You may now eject the USB drive. Use the exported ballot package on
-            this USB drive to configure the Ballot Scanner.
+            this USB drive to configure VxScan or VxCentralScan.
```
jonahkagan, line 346: I was guessing here. Is this right? Or should it just say VxCentralScan?

carolinemodic: yeah this is right

carolinemodic (approved):
thank you for updating the logs!

PR #1371, opened 2022-01-27
Remove duplicate ballots when importing CVRs

carolinemodic:
Automatically dedups ballots by "ballotId" when importing to election manager.

Flow
note that this example has 1 large file with the same 3000 CVR entries repeated several times, this is impossible in real life but was the easiest way to create a larger file for testing. 

https://user-images.githubusercontent.com/14897017/151267079-b06dabd1-6a45-4cbf-aa3e-a41a36d6413f.mov


frontends/election-manager/src/utils/cast_vote_record_files.ts
```diff
@@ -188,6 +201,77 @@ export class CastVoteRecordFiles {
 
     return result;
   }
```
jonahkagan, line 203: It seems like `addAllFromFileSystemEntries` isn't used anymore, and `addFromFileSystemEntry` is only called by this function. Can all that code be removed?

frontends/election-manager/src/utils/cast_vote_record_files.ts
```diff
@@ -188,6 +201,77 @@ export class CastVoteRecordFiles {
...
+      );
+    } catch (error) {
+      return new CastVoteRecordFiles(
+        this.signatures,
+        this.files,
+        this.duplicateFilenames,
+        mapAdd(this.parseFailedErrors, () => fileData.name, error.message),
+        this.deduplicatedCastVoteRecords
+      );
+    }
```
jonahkagan, line 273: it seems like this error case is already handled within `addFromFileContent`, unless i'm reading it incorrectly

frontends/election-manager/src/utils/cast_vote_record_files.ts
jonahkagan: Similarly, `add` and `addAll` are only used in one place in the modal to add a single file (besides being used in tests), so they might be candidates for removal as well

(I started looking into this because I was getting confused trying to understand the interface of this module, and realized it has a lot of entry points and could probably be reduced to a few key functions)

jonahkagan (commented):
I'm not sure if I'm familiar enough with this code to give an approval, but hopefully some fresh eyes are helpful!

benadida (approved):
Jonah's comments seem worth addressing, either now or as followup TODOs, otherwise this lgtm.

frontends/election-manager/src/utils/cast_vote_record_files.ts
```diff
@@ -117,28 +113,46 @@ export class CastVoteRecordFiles {
...
+      string,
+      CastVoteRecord
     >
   ) {}
 
   /**
    * Import from exported localStorage string
    */
   static import(stringifiedCvrFiles: string): CastVoteRecordFiles {
+    // TODO use zod schemas for constructing/deconstructing this.
```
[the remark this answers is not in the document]
carolinemodic: later

frontends/election-manager/src/utils/cast_vote_record_files.ts
```diff
@@ -188,6 +201,77 @@ export class CastVoteRecordFiles {
...
+            file.name,
+            parsedFileInfo?.timestamp || new Date(file.mtime),
+            parsedFileInfo?.machineId || '',
+            parsedFileInfo?.isTestModeResults ?? false,
+            election
+          );
+          results.push(parsedFile);
+        }
+      } catch (error) {
+        continue;
```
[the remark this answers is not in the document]
carolinemodic: I'll add a comment but if there is an error reading the file or it isn't a valid CVR file it might throw. In those cases then its not a valid file to import either so we can just ignore it and not include it in the import wizard.

PR #1374, opened 2022-01-27
feat(bmd/precinct-scanner): Add rotate card image to card backwards screen

jonahkagan:
Screen Shot 2022-01-27 at 1 48 21 PM


frontends/bmd/src/pages/card_error_screen.tsx
```diff
@@ -21,6 +27,11 @@ export function CardErrorScreen({
     <Screen white>
       <Main>
         <MainChild center>
+          <RotateCardImage
+            aria-hidden
+            src="/images/rotate-card.svg"
+            alt="rotate-card"
+          />
```
beausmith, line 34: Nit: I'd like to do this with less code because this is going to become a pattern… but it can be optimized later.

Preferably, move the image (content) into the `Prose` block because it is part of the "prose". Wrap it in a `<p>` tag for default "prose" margin spacing. 

But to do the centering, I guess you'd need some CSS as there isn't anything in the Prose component for centering images yet! So the centering could be inline on the image: `style={{ width: '300px', margin: '0 auto' }}`

beausmith (approved):
One nit… but this is good to go without it!

beausmith:
Ah, one consistency thing I see in the screenshot (from a previous PR) … When using sentence case, end the sentence with punctuation. If title case, not punctuation is necessary.

```diff
- Card is backwards
+ Card is backwards.
```

frontends/bmd/src/pages/card_error_screen.tsx
jonahkagan: I tried this but it put the image too high up due to the h1 top margin. Though the centering did work automatically. So gonna leave as is this time around

jonahkagan:
Updated to title case, thanks for catching that

PR #1393, opened 2022-02-03
refactor(bmd/ps): Extract shared clock setting components

jonahkagan:
Task: #445 

Since the UI for opening the clock setting modal was slightly different
between bmd/ps, I extracted three different components:
- PickDateAndTimeModal
- SetClockButton (which opens the modal and interacts with kiosk browser)
- CurrentDateAndTime (for displaying the current clock setting)

These can be composed in different ways as desired in the various frontends.


benadida (approved):
I like the composability!

PR #1394, opened 2022-02-03
Abstract ElectionInfoBar from precinct scanner to shared components and use in VxAdmin and VxCentralScan

carolinemodic:
Abstracts the ElectionInfoBar foot from precinct-scanner (which include the code version) into the shared UI library and uses it from VxAdmin and VxCentralScan.


benadida (approved):
love the clean separate commits, made review a breeze.

Issue #1343, opened 2022-01-11
scan: prototype scanning existing NH ballot


eventualbuddha:
Sample Ballots:
- [milford-nh-ballot.pdf](https://github.com/votingworks/vxsuite/files/7850444/milford-nh-ballot.pdf)
- [barnstead-nh-ballot.pdf](https://github.com/votingworks/vxsuite/files/7850468/barnstead-nh-ballot.pdf)
- [claremont-ward-3-nh-ballot.pdf](https://github.com/votingworks/vxsuite/files/7850469/claremont-ward-3-nh-ballot.pdf)

PR #1395, opened 2022-02-03
refactor: rename `hmpb-interpreter`

eventualbuddha:
We're likely to have different ballot interpreters for different types of ballots. Until now we've just had our own automatically-generated ballot. We're now building support for ballots with timing marks, which is completely different from the approach we've taken with our own ballots. Therefore I believe it makes sense to create different libraries to support different ballots when the differences between them are large, as with ours vs. timing-mark ballots.

- Rename `libs/hmpb-interpreter` to `libs/ballot-interpreter-vx`.
- Update all references (just a global search & replace).
- Re-order some sorted lists.

Refs #1343


jonahkagan (approved):
One less acronym, I like it!

benadida (approved):
I like it, though double checking that you want to get rid of hand-marked in the name... is this for BMD and hand-marked ballots?

benadida:
> > I like it, though double checking that you want to get rid of hand-marked in the name... is this for BMD and hand-marked ballots?
> 
> That's a good question. I think it does make sense to go down that road, but it'll require more refactoring to make this name make complete sense. How do we feel about it being a little inconsistent for now?

I worry a tad that it might be confusing... but I'm not that worried about it. I can live with the current name.

PR #1407, opened 2022-02-08
feat(scan): add initial NH ballot interpreter

eventualbuddha:
This is not complete, but it has the basics for finding timing marks in a NH state ballot.


libs/ballot-interpreter-nh/src/accuvote.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  decodeFrontTimingMarkBits,
+  findTimingMarks,
+  scanForTimingMarksByScoringBlocks,
+} from './accuvote';
+import { withSvgDebugger } from './debug';
+import { readGrayscaleImage } from './images';
+import { decodeBottomRowTimingMarks } from './timing_marks';
+import { FrontMarksMetadataSchema } from './types';
+
+test('hudson p1 front', async () => {
```
eventualbuddha, line 14: This is essentially the integration test for this library. Scans a real (template) image and locates the marks, decodes the barcodes, etc.

libs/ballot-interpreter-nh/src/accuvote.ts
```diff
@@ -0,0 +1,519 @@
...
+    bitsRightToLeft[5] * (1 << 3) +
+    bitsRightToLeft[6] * (1 << 4) +
+    bitsRightToLeft[7] * (1 << 5) +
+    bitsRightToLeft[8] * (1 << 6) +
+    bitsRightToLeft[9] * (1 << 7) +
+    bitsRightToLeft[10] * (1 << 8) +
+    bitsRightToLeft[11] * (1 << 9) +
+    bitsRightToLeft[12] * (1 << 10) +
+    bitsRightToLeft[13] * (1 << 11) +
+    bitsRightToLeft[14] * (1 << 12);
```
jonahkagan, line 415: I found this quite confusing, mostly because I'm not familiar with the bit operators. Took me a while to figure out you were just converting binary to base 10. Seems like it could potentially be extracted into a helper function `bitsToNumber(bits: Bit[]): number` which might make it clearer

jonahkagan (commented):
Nice work!

eventualbuddha:
> Nice work!

Thanks for looking at it, @jonahkagan!

libs/ballot-interpreter-nh/src/accuvote.ts
eventualbuddha: Thanks for the suggestion, here's the function for it: https://github.com/votingworks/vxsuite/pull/1407/commits/a78cda92e93643919f4215cf629fbbabb1d1a83d#diff-23718685541b5b070e874cc94031f2f8d8456496e76d37c6e538c5601dcab6d8R309-R323.

eventualbuddha:
Closing in favor of #1604

PR #1413, opened 2022-02-09
feat(scan): add `PUT /config/package`

eventualbuddha:
This allows configuring the scan service by `PUT`ting a ballot package zip file directly. This is useful for local testing, but will also be used for NH demo.

### DIFF
--- a/libs/utils/src/ballot_package.ts
+++ b/libs/utils/src/ballot_package.ts
@@ -133,10 +133,11 @@ async function readJsonEntry(zipfile: ZipFile, entry: Entry): Promise<unknown> {
 }
 
 async function readBallotPackageFromZip(
-  zipfile: ZipFile,
+  source: Buffer,
   fileName: string,
   fileSize: number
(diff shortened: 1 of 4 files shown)


libs/utils/src/ballot_package.ts
```diff
@@ -133,10 +133,11 @@ async function readJsonEntry(zipfile: ZipFile, entry: Entry): Promise<unknown> {
 }
 
 async function readBallotPackageFromZip(
```
benadida, line 135: should the name change to `FromBuffer` or `FromZipBuffer`?

benadida (approved):

libs/utils/src/ballot_package.ts
eventualbuddha: > should the name change to `FromBuffer` or `FromZipBuffer`?

Probably better, yes. Will do another PR (guess I shouldn't use auto-merge 😄).

PR #1420, opened 2022-02-10
refactor(scan): extract `DbClient`

eventualbuddha:
## Overview
`Store` has been responsible for both the business logic and the database access jobs. This commit extracts the database access job into a new class, `DbClient`, and renames its functions to be less namespaced. For now I retained most of the `db*` methods in `Store` to keep the changes to existing files isolated to `store.ts`. At a future time we should remove them and provide more semantic access to the database to any callers using them. This commit also creates new tests for `DbClient`.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tests added for `DbClient`, existing tests cover refactored code.

## Checklist

### DIFF
--- a/services/scan/src/db_client.test.ts
+++ b/services/scan/src/db_client.test.ts
@@ -1,0 +1,142 @@
+import * as fs from 'fs-extra';
+import * as tmp from 'tmp';
+import { DbClient } from './db_client';
+
+test('file database client', async () => {
+  const dbFile = tmp.fileSync();
+  const client = await DbClient.fileClient(dbFile.name);
+
+  await client.reset();
+  await fs.access(dbFile.name);
+
+  expect(client.getDatabasePath()).toBe(dbFile.name);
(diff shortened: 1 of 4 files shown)


services/scan/src/db_client.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  expect(await client.all('select * from muppets')).toEqual([
+    { name: 'Kermit' },
+    { name: 'Fozzie' },
+  ]);
+  expect(await client.one('select * from muppets')).toEqual({ name: 'Kermit' });
+  expect(
+    await client.one('select * from muppets where name != ?', 'Kermit')
+  ).toEqual({ name: 'Fozzie' });
+});
```
jonahkagan, line 65: I checked the coverage and it looks like there are still a number of uncovered lines in db_client. Might be a good time to address them

jonahkagan (approved):

PR #1423, opened 2022-02-11
fix(bsd): interpret blank single-page ballots appropriately

benadida:


carolinemodic (approved):
should fix the test but its about write in adjudication which isn't relevant for m13 right?

Issue #1417, opened 2022-02-09
VxMark: Hardware diagnostic report


jonahkagan:
## High-Level Requirement

BMD should have ability to print a hardware diagnostic report that lists:
* power status (plugged in and battery status)
* printer status (plugged in and toner + paper levels where possible)
* accessible controller & audio status (as per latest active diagnostic test run)

## Details

* On the admin screen, offer a "Hardware Diagnostic Report" button.
* Display, on screen, power status, printer status, and latest result from accessible controller & audio active diagnostic test, including when this was run.
* Offer a "print" button which, when tapped, prints the report including the data that is displayed on screen, plus machine ID, what it is configured for (election hash and name), and current time.

adghayes:
I'm just going to close this because we don't care about VxMark's diagnostic report, as it is not a part of the certified system.

PR #1436, opened 2022-02-15
refactor(ui/hooks): Change use_hardware to return full device info

jonahkagan:
## Overview

Task: #1417 
Preliminary work to set up the data we need for a hardware diagnostic screen.

- Rename use_hardware to use_devices
- Return full device info (instead of just whether or not the device is connected)
- Also add battery status polling to this hook
- Refactor useSmartcard to consume the new output of this hook
- Updates bmd and precinct-scanner to consume the changed hook.

## Demo Video or Screenshot
N/A
## Testing Plan 
Updated unit tests
## Checklist

### DIFF
--- a/frontends/bmd/src/app_root.tsx
+++ b/frontends/bmd/src/app_root.tsx
@@ -514,9 +514,13 @@ export function AppRoot({
     []
   );
 
-  const devices = useDevices({ hardware, logger });
-  const { battery } = devices.computer;
-  const hasPrinterAttached = devices.printer !== undefined;
+  const {
+    cardReader,
+    printer: printerInfo,
+    accessibleController,
+    computer,
+  } = useDevices({ hardware, logger });
+  const hasPrinterAttached = printerInfo !== undefined;
   const previousHasPrinterAttached = usePrevious(hasPrinterAttached);
 
   const ballotStyle =
@@ -1178,7 +1182,7 @@ export function AppRoot({
     initializedFromStorage,
   ]);
 
-  if (!devices.cardReader) {
+  if (!cardReader) {
     return (
       <SetupCardReaderPage
         useEffectToggleLargeDisplay={useEffectToggleLargeDisplay}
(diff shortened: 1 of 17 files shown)


frontends/bmd/src/app_root.tsx
```diff
@@ -1181,17 +1132,13 @@ export function AppRoot({
     void updateStorage();
     startCardShortValueReadPolling();
     startLongValueWritePolling();
-    startHardwareStatusPolling();
     return /* istanbul ignore next - triggering keystrokes issue - https://github.com/votingworks/bmd/issues/62 */ () => {
       stopCardShortValueReadPolling();
-      stopHardwareStatusPolling();
```
jonahkagan, line 1187: I'm not sure if there was any particular reason to start/stop battery status polling here in particular. It seems like it should be ok to just do it in a hook on mount, but if there was a reason to do it here, it would be good to think about.

frontends/bsd/src/app_root.tsx
```diff
@@ -650,7 +653,7 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
                 <ScanButton
                   onPress={scanBatch}
                   disabled={isScanning}
-                  isScannerAttached={hasBatchScannerAttached}
+                  isScannerAttached={!!batchScanner}
```
carolinemodic, line 656: @eventualbuddha Should there be an eslint rule for this? I thought we preferred having a more explicit expression here like `batchScanner != undefined` but I'm forgetting if thats part of the code style guide or not.

libs/ui/src/hooks/use_devices.ts
```diff
@@ -30,66 +41,49 @@ function getDeviceName(device: KioskBrowser.Device) {
...
-      });
+      .subscribe(setAllPrinters);
+
     return () => {
       hardwareStatusSubscription.unsubscribe();
       printerStatusSubscription.unsubscribe();
     };
   }, [hardware]);
 
+  useInterval(
```
carolinemodic, line 79: I don't think we need to do any logging for the battery but just a heads up for when you add things like printer paper checks to this I think we do need to log when the printer runs out of paper and things like that.

carolinemodic (approved):

libs/ui/src/hooks/use_devices.ts
jonahkagan: noted, thanks!

PR #1440, opened 2022-02-17
refactor: improve compressed tally types

eventualbuddha:
## Overview

The tests for the compressed tallies were very hard to read because of the opaque lists of numbers, and to a lesser extent the nonspecific type `number[]` used to represent them.

To improve this I added comments to indicate what each number is for. And the types are now numeric tuples with labeled entries, making it easier to see what field corresponds to what.

## Demo Video or Screenshot
Improved type information:
image

## Testing Plan 
Automated testing.

## Checklist

### DIFF
--- a/frontends/bmd/src/pages/poll_worker_screen.test.tsx
+++ b/frontends/bmd/src/pages/poll_worker_screen.test.tsx
@@ -1,20 +1,25 @@
 import React from 'react';
 import {
   asElectionDefinition,
-  electionMinimalExhaustiveSampleDefintion,
+  electionMinimalExhaustiveSampleDefinition,
 } from '@votingworks/fixtures';
 import {
+  CandidateContestWithoutWriteInsCompressedTally,
+  CandidateContestWithWriteInsCompressedTally,
   CompressedTally,
   ContestId,
   Dictionary,
   Election,
+  MsEitherNeitherContestCompressedTally,
   safeParseElection,
+  YesNoContestCompressedTally,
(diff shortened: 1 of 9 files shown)


frontends/bmd/src/pages/poll_worker_screen.test.tsx
```diff
@@ -580,7 +826,7 @@ test('printing precinct scanner report works as expected with all precinct speci
   });
 
   // Check that the expected results are on the tally report for Precinct 1 Mammal Party
-  const precinct1MammalReports = await screen.getAllByTestId(
+  const precinct1MammalReports = screen.getAllByTestId(
```
eventualbuddha, line 829: This was an unrelated opportunistic change to remove some unnecessary `await` expressions (`getAllByTestId` does not return a `Promise`).

jonahkagan (approved):
Seems clearer!

I don't totally know what compressed tallies are for but it makes me wonder if we could represent the compressed tally in a more readable way (e.g. objects with the labels as keys) and then just use `Object.values` whenever we actually need to compress and store it. But I'm lacking a lot of context so this may not actually be a useful idea

eventualbuddha:
Something like that may work. @carolinemodic and @benadida have more context on this.

carolinemodic:
@jonahkagan We keep the tallies in a more readable form most of the time, they are only converted to/from compressed tallies when we're storing them on the smartcards where we need to use as little space as possible.

PR #1447, opened 2022-02-22
refactor(scan): remove `contest` from mark info

eventualbuddha:
## Overview

_(There was a ticket to do this at some point, but I can't find it.)_ Adding the `contest` and `option` objects to the mark info creates a lot of cruft in the database and in responses from the scan service. This PR replaces them with `contestId` and `optionId`.

## Demo Video or Screenshot
n/a

## Testing Plan 
In addition to automated testing, I did manual testing of bsd and precinct-scanner with the Union County April 6, 2021 election.

## Checklist

### DIFF
--- a/libs/ballot-interpreter-vx/src/hmpb/votes.ts
+++ b/libs/ballot-interpreter-vx/src/hmpb/votes.ts
@@ -1,58 +1,52 @@
 import {
-  AnyContest,
-  Candidate,
-  CandidateContest,
+  CandidateId,
   CandidateVote,
-  MsEitherNeitherContest,
+  ContestId,
+  Election,
   VotesDict,
-  YesNoContest,
+  WriteInCandidate,
   YesNoOption,
+  YesNoOptionId,
   YesNoVote,
 } from '@votingworks/types';
-import { inspect } from 'util';
+import { assert, find, throwIllegalValue } from '@votingworks/utils';
 
 export function addVote(
+  election: Election,
   votes: VotesDict,
-  contest: CandidateContest,
-  candidate: Candidate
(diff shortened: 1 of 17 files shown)


libs/types/src/election.ts
```diff
@@ -98,14 +98,14 @@ export const DistrictsSchema = z
   });
 
 // Candidates
-export type WriteInId = Id;
-export const WriteInIdSchema: z.ZodSchema<WriteInId> = z
+export type WriteInId = `__write-in` | `__write-in-${string}`;
```
jonahkagan, line 101: I think we also do some of the form `write-in__<candidate name>` for BMD ballots where we know the text of the write in. i found this in services/scan/src/build_cast_vote_record.test.tsx

libs/ballot-interpreter-vx/src/hmpb/votes.ts
```diff
@@ -1,58 +1,52 @@
...
-  contest: AnyContest,
-  candidateOrYesNoOrEitherNeither: Candidate | 'yes' | 'no' | YesNoOption
+  contestId: ContestId,
+  option:
+    | WriteInCandidate
+    | 'yes'
+    | 'no'
+    | YesNoOption
+    | CandidateId
+    | YesNoOptionId
```
jonahkagan, line 30: I'm a little confused why you can use either `'yes' | 'no'` or `YesNoOption` or `YesNoOptionId` - seems like some of those should be the same? I noticed a change lower down that set the type of `optionId` to `'yes' | 'no'`

jonahkagan (approved):

PR #1471, opened 2022-03-03
refactor(scan): separate template interpretation

eventualbuddha:
## Overview

[](https://github.com/eventualbuddha)I plan to make it so the scan service does not need PDFs in order to do its job. Instead, it will have layout information that is generated by interpreting the rendered PDF as we do now, only that will happen in election-manager. That layout information may also have oval images to compare against–I haven't figured that part out yet.

For now, this makes `Interpreter#addTemplate` require a `BallotPageLayoutWithImageData` object rather than allowing an `ImageData` to be passed. It still has an `interpretTemplate` method to find the layout from an image, but it forwards to a standalone function that we'll be able to use from election-manager in the future.

This PR also splits up the takes-forever-to-run `interpreter.test.ts` file so that the parts can run in parallel.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated testing.

## Checklist


libs/ballot-interpreter-vx/src/layout.ts
```diff
@@ -0,0 +1,140 @@
...
+    imageData,
+    metadata,
+    flipped,
+    detectQrCode,
+  });
+
+  debug('using metadata for template: %O', normalized.metadata);
+
+  const contests = findContestOptions([
+    ...map(
```
jonahkagan, line 119: any reason not to use built-in Array.prototype.map here? just curious

jonahkagan (commented):
I don't have enough context to comment on the overall architecture decision, but the code changes look good to me

libs/ballot-interpreter-vx/src/layout.ts
eventualbuddha: No. There may have been at one point, as this `map` works with (and returns) iterables instead of arrays. In theory that allows you to work with larger collections without the memory overhead, but the N here is so small that it's surely not worth it.

benadida (approved):
YES, excellent direction.

PR #1503, opened 2022-03-07
fix(scan): make `render-pages` script work again

eventualbuddha:
## Overview

The primary issue was that the path needed to be corrected to use an underscore instead of a dash. Secondarily I switched to using `esbuild-runner` instead of `ts-node` since it's way faster.

I'm using this script to update some fixtures.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tried it manually on a couple PDFs.

## Checklist

### DIFF
--- a/esbuild-runner.config.js
+++ b/esbuild-runner.config.js
@@ -1,0 +1,7 @@
+module.exports = {
+  type: "transform",
+  esbuild: {
+    // Any esbuild build or transform options go here
+    target: "es2019",
+  },
+};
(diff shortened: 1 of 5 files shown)


benadida (approved):

jonahkagan (approved):
Cool, hadn't heard of `esbuild` before but faster sounds great. Are you testing this out to see if we can use it more broadly?

eventualbuddha:
> Cool, hadn't heard of `esbuild` before but faster sounds great. Are you testing this out to see if we can use it more broadly?

Yes. It's _almost_ a drop-in replacement but has some differences in operation that make it take some work. Will report back more.

PR #1510, opened 2022-03-08
chore(scan): use `esbuild-runner` for `vx` CLI

eventualbuddha:
## Overview

This CLI only worked when the project was built and required a rebuild for changes to the `.ts` files to be reflected. This commit switches to translating to `.js` on the fly using `esbuild-runner`. This would have been a simple change that only touched `bin/ballot-interpreter-vx.js`, but `esbuild` works a little bit differently from `tsc`. In particular, it translates ES imports/exports to CommonJS in a way that makes circular imports not work in some cases where they would with `tsc`. Since the `help` command was referring to `index` and vice versa, this code failed to run correctly under `esbuild-runner`. To fix it I changed the structure to no longer have cycles. While I was in there I improved the tests a bit and stopped throwing exceptions for non-exceptional things like invalid command arguments.

## Demo Video or Screenshot
Output of `layout` command run on one of the fixtures:
image

## Testing Plan 
Automated testing, plus a bunch of manual testing.

## Checklist

### DIFF
--- a/libs/ballot-interpreter-vx/src/cli/index.ts
+++ b/libs/ballot-interpreter-vx/src/cli/index.ts
@@ -1,71 +1,11 @@
-import { throwIllegalValue } from '@votingworks/utils';
-import * as helpCommand from './commands/help';
-import * as interpretCommand from './commands/interpret';
-import * as layoutCommand from './commands/layout';
+import { err, ok, Result } from '@votingworks/types';
+import { find } from '@votingworks/utils';
+import { get as getCommands } from './commands';
+import { GlobalOptions } from './types';
 
-export interface Command<O> {
-  name: string;
-  description: string;
-  parseOptions(globalOptions: GlobalOptions): Promise<O>;
-  run(
-    options: O,
-    stdin: typeof process.stdin,
-    stdout: typeof process.stdout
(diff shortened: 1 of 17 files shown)


libs/ballot-interpreter-vx/src/cli/index.ts
```diff
@@ -140,40 +45,34 @@ export async function main(
...
 
-      case 'interpret':
-        return await interpretCommand.run(
-          commandOptions.options,
-          stdin,
-          stdout
-        );
+  if (globalOptionsResult.isErr()) {
+    stderr.write(`error: ${globalOptionsResult.err().message}\n`);
+    return -1;
```
jonahkagan, line 53: never seen a negative exit code before. is this a thing?

jonahkagan (approved):

libs/ballot-interpreter-vx/src/cli/index.ts
eventualbuddha: It becomes 255. I've been using it to mean "I couldn't understand what you told me to do", i.e. a 400 in HTTP-speak.

```sh
❯ sh -c 'exit -1'
❯ echo $?
255
```

Issue #1476, opened 2022-03-04
VxScan not scanning when powered on in wrong order


mcchilders:
VxScan not scanning when powered on in wrong order - ballots were not being scanned when the power cord is plugged in after the tablet is turned on where VxScan just doesn't suck them in at all. I was able to reproduce this on the scanner I was using when this came up, but we should investigate it further. If this is something out of our control to fix (I vaguely remember us discussing this before), we need to at least document the procedure for poll workers to make sure they can correct this (which is solved by just restarting the tablet).

Perhaps consider onscreen instruction?

mattroe:
I couldn't reproduce this on the precinct scanner I have at home.

benadida:
also cannot repro, blaming this on some bad scanners we had, and closing.

PR #1515, opened 2022-03-08
feat(precinct-scanner): Scanner Not Detected screen

jonahkagan:
## Overview
Task: #1476 

When the precinct scanner is disconnected, show a screen to the user and
don't poll for scanner status. Hopefully this will be a useful reminder
and also prevent making hanging requests to the scan service (which is
non-responsive when the scanner is disconnected - though likely we'll
want to fix that as well).

## Demo Video or Screenshot

Screen Shot 2022-03-08 at 3 32 00 PM

## Testing Plan 
Added automated test, also manually tested.

## Checklist

### DIFF
--- a/frontends/precinct-scanner/src/screens/setup_scanner_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/setup_scanner_screen.tsx
@@ -1,30 +1,18 @@
-import React, { useEffect } from 'react';
-import { fontSizeTheme, Main, MainChild, Prose, Screen } from '@votingworks/ui';
+import React from 'react';
+import { CenteredScreen, CenteredLargeProse } from '../components/layout';
 
-function doNothing() {
-  // do nothing
-}
-
-interface Props {
-  useEffectToggleLargeDisplay?: () => void;
-}
-
-export function SetupScannerScreen({
-  useEffectToggleLargeDisplay = doNothing,
-}: Props): JSX.Element {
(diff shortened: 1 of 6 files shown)


frontends/precinct-scanner/src/screens/setup_scanner_screen.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect } from 'react';
```
jonahkagan, line 1: this is just a copy of the card reader not detected screen with the text changed

frontends/precinct-scanner/src/screens/setup_scanner_screen.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { fontSizeTheme, Main, MainChild, Prose, Screen } from '@votingworks/ui';
+
+function doNothing() {
+  // do nothing
+}
+
+interface Props {
+  useEffectToggleLargeDisplay?: () => void;
```
carolinemodic, line 9: this is a VxMark thing I think so you can remove it

carolinemodic (approved):
Nice!

frontends/precinct-scanner/src/screens/setup_scanner_screen.tsx
jonahkagan: thanks!

PR #1523, opened 2022-03-09
docs(smartcards): remove mention of smartcard fixtures

jamesturk:
The documentation for running the smartcard service recommends using the local fixtures, which seem to have drifted out of date with the real schema.  Opening this PR to start a discussion of what the full fix here is.

I'm not sure how others are using mockCardReader.py, but it seems from a bit of discussion today that perhaps these fixtures are outdated altogether?  If so, I can extend this PR to remove the --fixtures flag & data entirely and document the --admin/--pollworker flags exclusively, but didn't want to take that step without further discussion and clarification.

### DIFF
--- a/services/smartcards/README.md
+++ b/services/smartcards/README.md
@@ -45,8 +45,7 @@ make coverage
 ## Mock a Smart Card
 
 Once you're running the server, you can enable a mock card reader with fixture
-data as in the examples below. Supply your own election definition, or use one
-of the existing [`fixtures/`](./fixtures).
+data as in the examples below. 
 
 ### Using your own election definition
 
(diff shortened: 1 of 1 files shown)


eventualbuddha:
> So what I'd propose (if I'm understanding the history here correctly):
> 
> * remove the --fixtures flag feature and fixtures here, having docs point at using an election from the libs/fixtures folder instead
> * add a replacement --empty flag to compliment --admin, --pollworker, etc. and replace the only fixture that might have remained useful
> * Update docs accordingly.

This sounds right to me.

eventualbuddha (approved):
LGTM so far. I'm happy to re-review once you're done with the rest.

Issue #1416, opened 2022-02-09
vxadmin: printing test deck prints too many ballots and some malformed ones in some cases


benadida:
## Description
Test deck printing may be broken, possibly in multi-precinct elections (so didn't show up in demo election test)

It appears that multiple print jobs are being sent, not just one for the single precinct. Maybe the logic that prints all the test decks is being mistakenly activated for single precinct. But in any case it's always printing the selected precinct that was requested, multiple times.

## Repro
* Use 2021-06-08 Warren County election definition
* print a single precinct test deck
* printing goes on and on, and eventually prints a malformed ballot (front and back are same)

carolinemodic:
@benadida Was this happening every time for you? I've tested this with the election you mentioned and can't repro, I haven't tried a fully locked down machine yet but when I run through the production builds in vxsuite-complete-system & kiosk-browser in my VM it doesn't repro in either Debian or Ubuntu 20.

benadida:
repro'ed this with 2021-06-08 Ms county 75 (Warren), try printing test deck for single precincts: Auditorium prints the test deck twice. Plumbers/Pipefitters prints test deck twice and then a malformed ballot.

PR #1528, opened 2022-03-09
feat(utils): make memory battery status more realistic

eventualbuddha:
## Overview

Refs #1416 

When polling for battery status in kiosk-browser, the real value returned by `readBatteryStatus` is always a new object. This currently causes `useDevices` to trigger a re-render of its calling component whenever battery status is updated, which is currently every 3s.

However, when runnning in Chrome we're using `MemoryHardware`, a fake hardware interface. Its `readBatteryStatus` method was returning the same object every time unless the battery level changed. This does not mimick the real one, and therefore was a difference in behavior between Chrome and kiosk-browser that can make bugs appear in one and not the other.

This commit fixes the issue by ensuring that the object returned by `MemoryHardware#readBatteryStatus` is always different, just like with the real one. This makes reproducing #1416 in Chrome easy.

## Demo Video or Screenshot
n/a

## Testing Plan 
n/a

## Checklist

### DIFF
--- a/libs/utils/src/Hardware/memory_hardware.ts
+++ b/libs/utils/src/Hardware/memory_hardware.ts
@@ -124,7 +124,9 @@ export class MemoryHardware implements Hardware {
    * Reads Battery status
    */
   async readBatteryStatus(): Promise<KioskBrowser.BatteryInfo | undefined> {
-    return this.batteryStatus;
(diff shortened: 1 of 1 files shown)


jonahkagan (approved):
This is smart. Can you add an explanatory comment? Otherwise I worry it will just get changed back.

PR #1530, opened 2022-03-09
fix(em): prevent accidental/broken prints of test decks

### LINKED ISSUE #1416: vxadmin: printing test deck prints too many ballots and some malformed ones in some cases

eventualbuddha:
## Overview

Fixes #1416

We're using `React.memo` to prevent re-rendering `TestDeckBallots`, but this is explicitly against the React docs: https://reactjs.org/docs/react-api.html#reactmemo
> This method only exists as a performance optimization. Do not rely on it to “prevent” a render, as this can lead to bugs.

We need to figure out a better way to handle renders that happen at times we don't expect, not by preventing them but by being resilient to them.

## Demo Video or Screenshot
https://user-images.githubusercontent.com/1938/157502915-05b550aa-33da-48fa-9fe3-47b76ced5946.mov

## Testing Plan 
- [x] Manual testing with single precinct in Chrome
- [x] Manual testing with 13-precinct test deck in Chrome
- [x] Manual testing on real hardware (@benadida)

## Checklist

### DIFF
--- a/frontends/election-manager/src/screens/print_test_deck_screen.tsx
+++ b/frontends/election-manager/src/screens/print_test_deck_screen.tsx
@@ -78,6 +78,8 @@ function TestDeckBallots({
 // What happens if we remove `React.memo`? See https://github.com/votingworks/vxsuite/issues/1416.
 // We need to figure out a better way to handle renders that happen at times we
 // don't expect, not by preventing them but by being resilient to them.
+//
+// https://github.com/votingworks/vxsuite/issues/1531
 const TestDeckBallotsMemoized = React.memo(TestDeckBallots);
 
 export function PrintTestDeckScreen(): JSX.Element {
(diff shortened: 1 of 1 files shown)


jonahkagan (approved):
Would be nice to have a link to the ticket for a long-term fix in the comment once that ticket gets created.

eventualbuddha:
> Would be nice to have a link to the ticket for a long-term fix in the comment once that ticket gets created.

Yeah, I was hoping @benadida would beat me to it.

Issue #1409, opened 2022-02-08
VxScan and VxCentralScan are in live mode right after configuratin, whereas VxMark starts up in test mode


benadida:
They should be consistent, and they should be in test mode by default right after configuration.

carolinemodic:
Both VxScan and VxCentralScan configure their mode through the scanning service: https://github.com/votingworks/vxsuite/tree/main/services/scan . We should make it so that the default mode that gets configured when an election is first set (which happens through a PATCH to election/config: https://github.com/votingworks/vxsuite/blob/main/services/scan/src/server.ts#L119 ) is that it is in test mode.

PR #1532, opened 2022-03-09
fix(scan): default scan service to test mode

jamesturk:
## Overview

[#1409](https://github.com/votingworks/vxsuite/issues/1409)

## Checklist


eventualbuddha (commented):
This looks like the right place to make the change. Seems some tests assumed it was live by default and need to be updated to change from test to live.

jamesturk:
Yep, working on those tests now 👍

eventualbuddha (approved):

PR #1542, opened 2022-03-10
feat(em): automatically remove write protection when creating admin cards

### LINKED ISSUE #1466: VxAdmin: Make programming admin cards easier
Initial Feedback: https://votingworks.slack.com/archives/CJU9MSC6S/p1646269484826299

Remove the "Enable overwrite protection" step in VxAdmin when programming admin cards and just automatically do that when you are programming an admin card. 

We could just do it automatically or add an extra step to the modal "This card has been previously programmed as an admin card, are you sure you want to rewrite it?" 

As a first step, for this ticket, let's do the simple thing:
- when programming an admin card, automatically override write protection
- don't change any of the buttons, so that if someone really wants to, they can override write protection on an admin card and make it a poll worker card, but let's not make that the default.

jamesturk:
## Overview
automatically calls overrideWriteProtection before writing admin cards,fixes #1466

## Checklist

### DIFF
--- a/frontends/election-manager/src/screens/smartcards_screen.tsx
+++ b/frontends/election-manager/src/screens/smartcards_screen.tsx
@@ -128,6 +128,9 @@ export function SmartcardsScreen(): JSX.Element {
     setIsProgrammingCard(false);
   }
   async function programAdminCard(passcode: string) {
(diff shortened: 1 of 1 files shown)


carolinemodic (approved):
yay!

PR #1548, opened 2022-03-11
chore(ps) make save cvrs to usb function

carolinemodic:
## Overview

Moves the logic to save the CVR export onto the usb drive it's own function. If you are now in kiosk-browser it will just download the file locally. You can call this function with `openFilePickerDialog` set to true to open the file picker dialog (i..e Save As) or false to autosave to the default location. 

## Testing Plan 
Added test for new function, manually exported results. 

## Checklist
- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. - We should add logging to this but skipping for m13 purposes. 

### DIFF
--- a/frontends/precinct-scanner/src/utils/save_cvr_export_to_usb.ts
+++ b/frontends/precinct-scanner/src/utils/save_cvr_export_to_usb.ts
@@ -1,0 +1,76 @@
+import { ElectionDefinition } from '@votingworks/types';
+import {
+  generateFilenameForScanningResults,
+  SCANNER_RESULTS_FOLDER,
+  usbstick,
+  generateElectionBasedSubfolderName,
+} from '@votingworks/utils';
+import path from 'path';
+import fileDownload from 'js-file-download';
+import * as scan from '../api/scan';
+import { MachineConfig } from '../config/types';
+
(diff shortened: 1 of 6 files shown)


carolinemodic:
@beausmith let me know if this is what you need!

frontends/precinct-scanner/src/utils/save_cvr_export_to_usb.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Saves the current results from the scanning service to the usb drive, if present.
+ * @param openFilePickerDialog Toggles whether to open the file picker dialog and allow the user to customize the filename / location to save.
+ */
+export async function saveCvrExportToUsb(
+  electionDefinition: ElectionDefinition,
+  machineConfig: MachineConfig,
+  scannedBallotCount: number,
+  isTestMode: boolean,
+  openFilePickerDialog: boolean
```
eventualbuddha, line 22: It's hard to understand what all these arguments are when they're just positional. Consider making this an object with options.

eventualbuddha (approved):

Issue #1432, opened 2022-02-13
Software Updates in the field


carolinemodic:
## Summary
Currently the process to update software on machines is a difficult process that can only be done by Drew and/or other engineers on the team. We need something that at the very least non-engineers on the team (Charlie, Ginny, etc.) can do, and ideally customers can even do someday. 

We've got m13b equipment in the field with the ability to reboot from USB using a superadmin card. We need to complete the feature for m14 so that we can actually perform the update using "reboot from USB."

## Requirements
- We can update m13b machines in the field to m14 with minimal technical skill.

## Non-Requirements
- we don't expect customers to be able to do this yet, though one particular capable customer may be.

## Tasks
m13B 
- [x] https://github.com/votingworks/vxsuite/issues/1401 
- [x] Add "superadmin" or "vxadmin" card that shows a basic screen with a button to reboot from usb (and export logs? and unconfigure?) 

m14 
- [x] https://github.com/votingworks/vxsuite/issues/1431
- [ ] #1862
- [ ] Process for getting USBs in the field and doing the install work

mcchilders:
@carolinemodic are the security concerns (making sure you're updating with official/unmodified Vx software, regenerating anything needed for updated system auth code, etc.) included in this card?

mattroe:
closing in favor of the roadmap issue https://github.com/votingworks/vxsuite-roadmap/issues/9

PR #1554, opened 2022-03-12
feat(all) add superadmin screen to reboot from usb

carolinemodic:
## Overview

[Software Updates in the field](https://github.com/votingworks/vxsuite/issues/1432)
Adds the ability to reboot from USB in the UI. When calling `prepareToBootFromUsb` window.kiosk may return false, in this case the machine usually needs to be rebooted with the usb inserted before it can work. 

Note I left the control of the outer Screen elements to the individual app @beausmith  we should figure out what the recommended practice is there. This means the buttons are different sizes in each app based on how they do sizing, etc. In BMD the way it makes things bigger is a little different and also causes the modal to be huge which doesn't happen in the other apps. 

## Demo Video or Screenshot
From my VM in each app, this shows the flow when there is no usb, then it mounts, but it is not in the bootable list yet. When it is in the bootable list the machine will automatically reboot after the main "Reboot from USB" button is pressed. 

https://user-images.githubusercontent.com/14897017/157999621-9553dfc0-71e9-4341-a2a6-20529cacb4f4.mov

https://user-images.githubusercontent.com/14897017/157999631-3639a3ef-1470-400c-9c3d-2c6b9d60c5ea.mov

https://user-images.githubusercontent.com/14897017/157999627-dc74c70a-d7af-4c56-b9e6-bde6e108445e.mov

https://user-images.githubusercontent.com/14897017/157999630-d263d523-d8a1-415c-bef3-e2718241e8ac.mov

Happy path on real hardware: 

https://user-images.githubusercontent.com/14897017/158033865-49cc02e7-ae31-4c95-b345-d7f0339a3e4d.MOV

## Testing Plan 
Tested as shown above in VM, added test cases to get to full coverage in /ui and /bmd. Tested on real hardware but not fully locked down (but I did test the kiosk browser pieces locked down seperately) 

## Checklist


carolinemodic:
Note @benadida one thing to note about how the superadmin card is implemented: In VxAdmin and VxCentralScan cards are ignored if they are inserted when a user session is active (so that you can program cards in VxAdmin) so you have to lock the machine before entering the superadmin card for it to work as expected.

libs/ui/src/reboot_from_usb_button.tsx
```diff
@@ -0,0 +1,147 @@
...
+          <React.Fragment>
+            <Button primary onPress={reboot}>
+              Reboot
+            </Button>
+            <Button onPress={onClose}>Close</Button>
+          </React.Fragment>
+        }
+      />
+    );
+  }
```
eventualbuddha, line 124: This case doesn't make sense to me. Is it trying to say that the currently inserted USB drive is not bootable? What will pressing "Reboot" do? Just reboot back into the app? If so, why even allow rebooting here? I think this needs to be clarified.

eventualbuddha (commented):
Did a quick pass, looks reasonable. I don't get what the warning state of the modal is supposed to be, though.

libs/ui/src/reboot_from_usb_button.tsx
carolinemodic: It depends on the hardware but most of them only update the devices that `efibootmgr` will list as bootable when the machine boots. So if you insert a bootable usb stick after the machine has booted you actually need to reboot once before it will work.

Issue #1452, opened 2022-02-25
Add totp code to all components


carolinemodic:
## Summary
Right now the totp code only exists on VxScan we should add it VxMark, VxAdmin and VxCentralScan (priority is in that order)

carolinemodic:
Where we are currently fetching this code in VxScan: https://github.com/votingworks/vxsuite/blob/ce7f4e1eb32f0098d57ad83c7e79655faa05869d/frontends/precinct-scanner/src/screens/poll_worker_screen.tsx#L203 
And displaying it: https://github.com/votingworks/vxsuite/blob/ce7f4e1eb32f0098d57ad83c7e79655faa05869d/frontends/precinct-scanner/src/screens/poll_worker_screen.tsx#L445 

Should go on the pollworker screen for VxMark.

benadida:
closing this as we're going to move to a different feature.

PR #1555, opened 2022-03-14
feat(bmd): add TOTP code to poll worker screen

jamesturk:
## Overview
adds TOTP to VxMark (part of #1452)

## Demo Video or Screenshot

Wanted to confirm this is the right screen & placement.  Wasn't entirely clear on where TOTP codes are used so far.

Screen Shot 2022-03-14 at 9 57 20 AM

If this is correct, I'll update to also include other components.

## Checklist

### DIFF
--- a/frontends/bmd/src/config/globals.ts
+++ b/frontends/bmd/src/config/globals.ts
@@ -19,3 +19,4 @@ export enum YES_NO_VOTES {
 }
 export const WRITE_IN_CANDIDATE_MAX_LENGTH = 40;
 export const QUIT_KIOSK_IDLE_SECONDS = 5 * 60; // 5 minutes
+export const POLLING_INTERVAL_FOR_TOTP = 2000;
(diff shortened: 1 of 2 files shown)


jamesturk:
Also, right now tests are failing because there is no coverage of the conditional within:

```
useInterval(
    async () => {
      const totpResult = await window.kiosk?.totp?.get();
      if (totpResult) {  // NO COVERAGE IN HERE
        const codeChunks = totpResult.code.match(/.{1,3}/g);
        if (codeChunks) setSystemAuthenticationCode(codeChunks.join('·'));
      }
    },
    POLLING_INTERVAL_FOR_TOTP,
    true
  );
```

Any advice on how kiosk-only features are typically handled for situations like this?   (cc @carolinemodic)

carolinemodic:
@jamesturk Yep you can set window.kiosk to a fakeKiosk object we have created and then mock out the responses to certain calls. Here is an example: https://github.com/votingworks/vxsuite/blob/main/frontends/precinct-scanner/src/screens/poll_worker_screen.test.tsx#L54 
Although that one has slightly different syntax then what I would usually expect, this is how I usually do it: https://github.com/votingworks/vxsuite/blob/main/frontends/bmd/src/pages/test_ballot_deck_screen.test.tsx#L65

beausmith (changes requested):
Request: Can you place the TOTP in the `<VersionsData>` component such that — in the screenshot — it shows up under the line "Software Version: dev"?

PR #1557, opened 2022-03-14
fix(scan): handle intentionally blank pages better

eventualbuddha:
## Overview

If a ballot sheet has contests on only one side then it will have an intentionally blank back page. This page isn't actually blank, but has a sentence explaining that it is being left blank intentionally. It also has a footer with the usual election information and QR code.

Additionally, we have some code that attempts to determine whether a page is blank (in the sense of an empty white page) before trying to look for a QR code. This is partly an optimization and partly to make sure nothing weird ends up on the other side of a BMD ballot.

Putting these together, though, causes a problem. The "blank page" detection works by counting the foreground (dark) vs background (light) pixels and, if the percentage of foreground pixels is less than 0.7% then we consider it truly blank and do not bother looking for a QR code. The threshold of 0.7% is very close to the value for an average "blank" sheet back page of letter size. For legal size, the foreground percent can be even lower.

This hasn't really been a problem because we infer "missing" QR codes when they aren't read properly. However, we can do better. This commit looks in the areas that we would look for QR codes and determines if any of them looks dark enough. If they are dark enough, we treat the page as non-blank and proceed to look for a QR code.

## Demo Video or Screenshot
n/a

## Testing Plan 
- [x] test with BMD ballots
- [x] test with HMPB ballots

## Checklist

### DIFF
--- a/services/scan/src/util/luminosity.ts
+++ b/services/scan/src/util/luminosity.ts
@@ -1,4 +1,5 @@
 import { otsu } from '@votingworks/ballot-interpreter-vx';
+import { Rect } from '@votingworks/types';
 import makeDebug from 'debug';
 
 const debug = makeDebug('scan:threshold');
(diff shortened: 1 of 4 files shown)


services/scan/src/util/luminosity.ts
```diff
@@ -19,35 +20,50 @@ export interface LuminosityCounts {
...
   let foreground = 0;
-  for (let i = 0; i < data.length; i += 1) {
-    if (data[i] < threshold) {
-      foreground += 1;
+  for (let { x } = bounds; x < bounds.x + bounds.width; x += 1) {
+    for (let { y } = bounds; y < bounds.y + bounds.height; y += 1) {
+      const pixel = data[(y * width + x) * channels];
+      if (x === 1126 && y === 1476) {
+        debug(`pixel at (${x}, ${y}) is foreground: ${pixel}`);
+      }
```
benadida, line 40: is this your favorite pixel?

benadida (approved):

PR #1562, opened 2022-03-14
fix(scan): make store.batchStatus query include batches w/deleted sheets

### LINKED ISSUE #1561: store.batchStatus() does not include batches with deleted sheets
## Description
`store.batchStatus()` does not return a given batch if all of its sheets have been deleted.

## Repro
The easiest way to repro this is to pull [this branch](https://github.com/votingworks/vxsuite/tree/tara/1561-fix-batch-status), and revert my changes to `store.ts`. The [test case I added in `store.test.ts`](https://github.com/votingworks/vxsuite/commit/68f77cb84d735810f33fe0fd2889be67ac7731ec#diff-b73dd18746392573b2161fd86538ca76949243aafc8782e0147f10bb371b3c23R183) will fail.

## Expected behavior
I believe (I may be wrong) that the expected behavior of `store.batchStatus()` is that a batch should still be included in the results even all of its sheets have been deleted.

taravancil:
## Overview
Fixes #1561

## Testing Plan 
Added a test case for `store.batchStatus()`

## Checklist

### DIFF
--- a/services/scan/src/store.test.ts
+++ b/services/scan/src/store.test.ts
@@ -4,6 +4,7 @@ import {
   BallotMetadata,
   BallotType,
   CandidateContest,
+  BallotPageMetadata,
   BallotPageLayout,
   YesNoContest,
 } from '@votingworks/types';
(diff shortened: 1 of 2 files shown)


eventualbuddha (approved):
Code looks fine. I'd want @benadida to give a 👍 on whether this is what we want product-wise.

taravancil:
@eventualbuddha thanks for the quick review! Definitely agree that @benadida should confirm that this is the intended behavior.

benadida:
As per synchronous conversation, I'm very confused as to why the first query behaved the way it did, but the new query is definitely better, and the resulting behavior is, IMO, correct. 👍

PR #1565, opened 2022-03-14
fix(all): remove placeholder empty election hashes

eventualbuddha:
## Overview

We had these all over the place, primarily in tests. We also had them when reading the old-style URL-based QR codes on HMPBs, because they don't contain a partial election hash. So this also changes that code to expect an election definition to return the hash properly. That code will hopefully be removed soon anyway.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated testing.

## Checklist


services/scan/src/cli/retry-scan/index.test.ts
```diff
@@ -143,11 +143,9 @@ test('query with sheet ids', () => {
   const inputWorkspace = await createWorkspace(dirSync().name);
   const { store } = inputWorkspace;
 
-  store.setElection({
-    election: fixtures.election,
-    electionData: JSON.stringify(fixtures.election),
-    electionHash: '02f807b005e006da160b',
-  });
+  store.setElection(fixtures.electionDefinition);
+  store.setTestMode(false);
```
eventualbuddha, line 147: Interestingly, this was needed because of the change in default from live to test mode in the scan service. It didn't fail in CI because this test doesn't run in CI due to resource constraints.

benadida: huh, any chance we can run these more intensive tests in a non-blocking CI environment or something, maybe every now and then?

benadida: /cc @carolinemodic to think about this issue, not blocking this PR of course.

benadida (approved):

Issue #1384, opened 2022-02-01
scan and centralscan: don't allow user to clear election data until they export the backup with full image scans


benadida:
There is a button in the admin screen of VxCentralScan (frontends/bds in vxsuite) labelled "Export Backup..." that exports a zip package containing full image exports of all the ballots scanned, the cvr file, etc. We want to make sure that users do this before they clear all of the election data on the machine. Feel free to work with Beau on the design / user flows, but also take ownership of how you think this should work. A few possibilities: 
- Disable the "Delete election data" button with a sentence over it as to why until the backup is exported
- When you click "Delete election data" have a modal pop up that says you must first export backup (and maybe let you do that from the modal?) 

We will also want to do the same thing in VxScan (the precinct scanner found in frontends/scan in vxsuite). In VxScan we should update the language on the "Unconfigure" button to match VxCentralScan "Delete Election Data from VxScan…". 

In order to do this we will need to somehow track what scans have been exported, you will need to do this in the backend of the scanning service in services/scan in vxsuite. This backend is used by both the VxScan and VxCentralScan frontends.

Nice to Have: 
In VxCentralScan we haven't updated "Export Backup" to use the same nicer, export to usb modal flow that the regular "Export" button has, and that either Export option has in VxScan. VxScan also saved the backup automatically into a folder called scanner-backups we should use the same logic/naming in both places (and perhaps include the machine id in the filename?). This work to clean up the UI of VxCentralScan here can be split out as a separate task.

Point of contact on UI Design / Flows : Beau
Point of contact for engineering help, particularly module-scan : Brian
Point of contact for eng help with the save file to usb modals : Caroline

mattroe:
@benadida this should be for VxScan too not just VxCentralScan

carolinemodic:
I have confirmed that the export still works as expected, other then one slight issue I'm seeing with the automatic export in VxScan (doing a custom save works fine still), will look into that.

carolinemodic:
Oh VxScan is working I just didn't realize it was getting saved to a folder in scanner-backups on the usb drive. 
We should actually update VxCentralScan to save to the same location that VxScan is. But perhaps both should have the machine ID in the filename.

PR #1580, opened 2022-03-15
Require export before delete in VxScan

taravancil:
## Overview
#1384 

This does several things:

* Adds a table `backups` that tracks timestamps of given backup keys (currently `Cvrs` and `Batches`)
* Adds `store.getCanUnconfigure()`, which determines whether the machine can be unconfigured
* Updates `DELETE /config/election` and `POST /scan/zero` to require `store.getCanUnconfigure()` to return true.
* Updates the VxScan admin UI to disable the "Delete Election Data..." button until CVRs and batches have been backed up. The button is disabled again when an export is invalidated by either adding or deleting a batch or adding new sheets. The button is always enabled in test mode.
  * _Note that a machine with no batches will allow deleting election data. This seems to me like the right thing to do, but I am open to other opinions_

The work to implement this in VxCentralScan will happen in a separate ticket.

## Demo Video or Screenshot
Uploading Screen Recording 2022-04-14 at 5.53.33 AM.mov…

## Testing Plan
I have added several tests and manually QAed this flow.

## Checklist

### DIFF
--- a/frontends/bsd/src/app_root.tsx
+++ b/frontends/bsd/src/app_root.tsx
@@ -640,6 +640,10 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
               hasBatches={status.batches.length > 0}
               isTestMode={isTestMode}
               toggleTestMode={toggleTestMode}
+              hasExportedBackupForAllBatches={status.batches.every(
+                (b) => b.exportedAt
+              )}
+              exportedAt={status.batches[0] ? status.batches[0].exportedAt : ''}
               setMarkThresholdOverrides={setMarkThresholdOverrides}
               markThresholds={markThresholds}
               isTogglingTestMode={isTogglingTestMode}
--- a/frontends/bsd/src/app_root.tsx
+++ b/frontends/bsd/src/app_root.tsx
@@ -641,7 +641,7 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
               isTestMode={isTestMode}
               toggleTestMode={toggleTestMode}
               canUnconfigure={status.batches.every((b) => b.exportedAt)}
-              exportedAt={status.batches[0] ? status.batches[0].exportedAt : ''}
+              exportedAt={status.batches[0]?.exportedAt}
               setMarkThresholdOverrides={setMarkThresholdOverrides}
               markThresholds={markThresholds}
               isTogglingTestMode={isTogglingTestMode}
--- a/frontends/bsd/src/app_root.tsx
+++ b/frontends/bsd/src/app_root.tsx
@@ -640,9 +640,7 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
               hasBatches={status.batches.length > 0}
               isTestMode={isTestMode}
               toggleTestMode={toggleTestMode}
-              hasExportedBackupForAllBatches={status.batches.every(
-                (b) => b.exportedAt
-              )}
+              canUnconfigure={status.batches.every((b) => b.exportedAt)}
               exportedAt={status.batches[0] ? status.batches[0].exportedAt : ''}
               setMarkThresholdOverrides={setMarkThresholdOverrides}
               markThresholds={markThresholds}
--- a/frontends/bsd/src/app_root.tsx
+++ b/frontends/bsd/src/app_root.tsx
@@ -95,6 +95,7 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
   const [isTestMode, setTestMode] = useState(false);
   const [isTogglingTestMode, setTogglingTestMode] = useState(false);
   const [status, setStatus] = useState<GetScanStatusResponse>({
+    canUnconfigure: true,
     batches: [],
     adjudication: { remaining: 0, adjudicated: 0 },
     scanner: ScannerStatus.Unknown,
@@ -640,7 +641,7 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
               hasBatches={status.batches.length > 0}
               isTestMode={isTestMode}
               toggleTestMode={toggleTestMode}
-              canUnconfigure={status.batches.every((b) => b.exportedAt)}
+              canUnconfigure={status.canUnconfigure}
               exportedAt={status.batches[0]?.exportedAt}
               setMarkThresholdOverrides={setMarkThresholdOverrides}
               markThresholds={markThresholds}
--- a/frontends/bsd/src/app_root.tsx
+++ b/frontends/bsd/src/app_root.tsx
@@ -642,7 +642,6 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
               isTestMode={isTestMode}
               toggleTestMode={toggleTestMode}
               canUnconfigure={status.canUnconfigure}
-              exportedAt={status.batches[0]?.exportedAt}
               setMarkThresholdOverrides={setMarkThresholdOverrides}
               markThresholds={markThresholds}
               isTogglingTestMode={isTogglingTestMode}
--- a/frontends/bsd/src/screens/admin_actions_screen.tsx
+++ b/frontends/bsd/src/screens/admin_actions_screen.tsx
@@ -124,6 +124,7 @@ export function AdminActionsScreen({
                   isTestMode={isTestMode}
                   isTogglingTestMode={isTogglingTestMode}
                   toggleTestMode={toggleTestMode}
+                  canUnconfigure={canUnconfigure}
                 />
               </p>
               <p>
--- a/frontends/bsd/src/screens/admin_actions_screen.tsx
+++ b/frontends/bsd/src/screens/admin_actions_screen.tsx
@@ -1,8 +1,7 @@
 import { ElectionDefinition, MarkThresholds } from '@votingworks/types';
-import { shortDateTime } from '@votingworks/utils';
 import React, { useCallback, useEffect, useState, useContext } from 'react';
 import { LogEventId } from '@votingworks/logging';
-import { LogFileType } from '@votingworks/utils';
+import { shortDateTime, LogFileType } from '@votingworks/utils';
 import { Loading, Modal, SetClockButton } from '@votingworks/ui';
 import { Button } from '../components/button';
 import { LinkButton } from '../components/link_button';
--- a/frontends/bsd/src/screens/admin_actions_screen.tsx
+++ b/frontends/bsd/src/screens/admin_actions_screen.tsx
@@ -1,4 +1,5 @@
 import { ElectionDefinition, MarkThresholds } from '@votingworks/types';
+import { shortDateTime } from '@votingworks/utils';
 import React, { useCallback, useEffect, useState, useContext } from 'react';
 import { LogEventId } from '@votingworks/logging';
 import { LogFileType } from '@votingworks/utils';
@@ -8,6 +9,7 @@ import { LinkButton } from '../components/link_button';
 import { Main, MainChild } from '../components/main';
 import { MainNav } from '../components/main_nav';
 import { Prose } from '../components/prose';
+import { Text } from '../components/text';
 import { Screen } from '../components/screen';
 import { ToggleTestModeButton } from '../components/toggle_test_mode_button';
 import { SetMarkThresholdsModal } from '../components/set_mark_thresholds_modal';
@@ -21,6 +23,8 @@ interface Props {
   hasBatches: boolean;
   isTestMode: boolean;
   isTogglingTestMode: boolean;
+  hasExportedBackupForAllBatches: boolean;
+  exportedAt?: string;
   toggleTestMode: () => Promise<void>;
   setMarkThresholdOverrides: (markThresholds?: MarkThresholds) => Promise<void>;
   markThresholds?: MarkThresholds;
@@ -35,6 +39,8 @@ export function AdminActionsScreen({
   isTestMode,
   isTogglingTestMode,
   toggleTestMode,
+  hasExportedBackupForAllBatches,
+  exportedAt,
   setMarkThresholdOverrides,
   markThresholds,
   electionDefinition,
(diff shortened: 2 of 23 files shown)


libs/utils/src/date.ts
```diff
@@ -73,3 +73,14 @@ export function getDaysInMonth(year: number, month: number): DateTime[] {
   }
   return days;
 }
+
+function z2(number: number) {
+  return number.toString().padStart(2, '0');
+}
+
+export function shortDateTime(iso8601Timestamp: string) {
```
eventualbuddha, line 81: `eslint` says:
> Missing return type on function. `eslint@typescript-eslint/explicit-module-boundary-types`

libs/utils/src/date.ts
```diff
@@ -73,3 +73,14 @@ export function getDaysInMonth(year: number, month: number): DateTime[] {
...
+  const d = new Date(iso8601Timestamp);
+  return `${d.getFullYear()}-${z2(d.getMonth() + 1)}-${z2(
```
eventualbuddha, line 83: Maybe just use `Intl.DateTimeFormat` for this?

frontends/bsd/src/app_root.tsx
```diff
@@ -640,6 +640,10 @@ export function AppRoot({ card, hardware }: AppRootProps): JSX.Element {
               hasBatches={status.batches.length > 0}
               isTestMode={isTestMode}
               toggleTestMode={toggleTestMode}
+              hasExportedBackupForAllBatches={status.batches.every(
+                (b) => b.exportedAt
+              )}
+              exportedAt={status.batches[0] ? status.batches[0].exportedAt : ''}
```
eventualbuddha, line 646: Could be abbreviated:
```suggestion
              exportedAt={status.batches[0]?.exportedAt}
```
No need for the `''` as `undefined` should suffice.

services/scan/src/store.ts
```diff
@@ -626,6 +642,7 @@ export class Store {
       batchId
     ) as { count: number };
     this.client.run('delete from batches where id = ?', batchId);
+    this.invalidateBatchesExportedAt();
```
eventualbuddha, line 645: Does this really invalidate anything? Any previous backups would still have all the other batches (plus this one). I'm unsure how this ought to work cc @benadida.

frontends/bsd/src/screens/admin_actions_screen.tsx
```diff
@@ -21,6 +23,8 @@ interface Props {
   hasBatches: boolean;
   isTestMode: boolean;
   isTogglingTestMode: boolean;
+  hasExportedBackupForAllBatches: boolean;
+  exportedAt?: string;
```
eventualbuddha, line 27:
```suggestion
  exportedAt?: Iso8601Timestamp
```

services/scan/src/server.ts
```diff
@@ -162,6 +162,21 @@ export function buildApp({ store, importer }: AppOptions): Application {
   app.delete<NoParams, DeleteElectionConfigResponse>(
     '/config/election',
     async (_request, response) => {
+      const batches = await store.batchStatus();
+      if (!batches.every((b) => b.exportedAt)) {
```
eventualbuddha, line 166: This construct is used a few times in this PR. Consider making something equivalent to this a method on `Store`.

eventualbuddha (commented):
A good start! Do we also want to prevent unconfigure until CVRs have been exported? cc @benadida

benadida:
> A good start! Do we also want to prevent unconfigure until CVRs have been exported? cc @benadida

Yes, that's a good call. A few enhancements here that should have been part of the ticket from the start, my bad for not specifying them:

- we should require CVR export.
- we should block "Delete Ballot Data" in the same way, and we can put that button at the bottom so it's grouped together with the big red button. In fact I'd say both those buttons should be red.
- we should NOT block if we're in test mode – it's just test ballots and the warning is fine.

@taravancil the UX / text is good by me. And sorry for the added scope here, I should have scoped this ticket better.

services/scan/src/store.ts
```diff
@@ -432,6 +446,7 @@ export class Store {
         frontFinishedAdjudicationAt ?? null,
         backFinishedAdjudicationAt ?? null
       );
+      this.invalidateBatchesExportedAt();
```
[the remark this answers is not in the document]
taravancil: That is how it worked before d2fe24b0223dc1277930a73dc0a919dbffaf06bb, but as I was working on invalidating the export when _deleting a batch_, I realized the approach of marking individual batches with `exported_at=null` would not work in that case, because each remaining batch would still have the `exported_at` column set.

As an alternative approach, we could use a single flag `exported_at` instead of `exported_at` on each batch. That is effectively how `batches.exported_at` works in this current state, anyway.

I guess before going any further, I should also confirm that deleting a batch should invalidate the export. What do you think @eventualbuddha? If not, I can revert d2fe24b0223dc1277930a73dc0a919dbffaf06bb and go back to individually invalidating batches.

services/scan/src/store.ts
taravancil: Yeah, I was unsure about whether deleting a batch should invalidate an export. As mentioned [here](https://github.com/votingworks/vxsuite/pull/1580#discussion_r827341485), if this should not invalidate the backup, we can revert d2fe24b0223dc1277930a73dc0a919dbffaf06bb.

taravancil:
> we should require CVR export.

@benadida checking my understanding, the resulting .zip when I click "Export Backup..." includes a `cvrs.jsonl` file. ([source](https://github.com/votingworks/vxsuite/blob/7d6b73ba5eb117cd28c1ac817b8d89ec765938e7/services/scan/src/backup.ts#L73,L75)). Is that sufficient or is there more to do here?

benadida:
> > we should require CVR export.
> 
> @benadida checking my understanding, the resulting .zip when I click "Export Backup..." includes a `cvrs.jsonl` file. ([source](https://github.com/votingworks/vxsuite/blob/7d6b73ba5eb117cd28c1ac817b8d89ec765938e7/services/scan/src/backup.ts#L73,L75)). Is that sufficient or is there more to do here?

Not sufficient, cause the election admin will use those two files differently.

frontends/bsd/src/screens/admin_actions_screen.tsx
```diff
@@ -145,8 +151,22 @@ export function AdminActionsScreen({
...
+              )}
               <p>
-                <Button danger onPress={toggleIsConfirmingUnconfigure}>
+                <Button
+                  danger
+                  disabled={!hasExportedBackupForAllBatches}
+                  onPress={toggleIsConfirmingUnconfigure}
+                >
                   Delete Election Data from VxCentralScan…
                 </Button>
```
beausmith, line 171: The text above the button is equally close to the button above it too. We should more closely associate the text to the button it's related to.

I haven't run the code below to ensure it works, but it should:
- Place the text to the right of the button.
- Update copy to be more specific. (Though now realizing that "Export Backup" should probably be changed to something more descriptive which doesn't contain two closely related verbs/tems. 🙃 )
- Use warning style text.
- Use a warning icon.

I've removed the italic style as I'd like to save that for adding emphasis to content next to non-italicized content.

```suggestion
                <Button
                  danger
                  disabled={!hasExportedBackupForAllBatches}
                  onPress={toggleIsConfirmingUnconfigure}
                >
                  Delete Election Data from VxCentralScan…
                </Button>{' '}
                {!hasExportedBackupForAllBatches && (
                  <Text as="span" warning warningIcon>
                    You must "Export Backup" before you may delete election data.
                  </Text>
                )}
                {hasExportedBackupForAllBatches && exportedAt && (
                  <Text as="span" muted>
                    Election data was last backed up at {shortDateTime(exportedAt)}
                  </Text>
                )}
```

frontends/bsd/src/screens/admin_actions_screen.tsx
```diff
@@ -145,8 +151,22 @@ export function AdminActionsScreen({
...
+              {!hasExportedBackupForAllBatches && (
+                <Text italic muted>
+                  Export a backup in order to delete election data.
+                </Text>
+              )}
+              {hasExportedBackupForAllBatches && exportedAt && (
+                <Text italic muted>
+                  Election data backed up at {shortDateTime(exportedAt)}
+                </Text>
+              )}
```
beausmith, line 163: Related to other comment.

```suggestion
```

beausmith (approved):
This looks good. I've got a suggestion for the visual display. LMK if you feel otherwise.

services/scan/src/server.ts
taravancil: I've added `Store.getCanUnconfigure()`, and then also added `canUnconfigure` to `GetScanStatusResponse`.

taravancil:
@benadida called out that manually invalidating the backup when we add/delete batches/sheets is brittle, because it requires us to remember to invalidate backups if/when we add new features. He made a good suggestion that we could compare export timestamps to the max of `batches.ended_at`, `sheets.created_at`, `sheets.deleted_at` to decide if a backup is up-to-date. I like this idea, but it still requires manually invalidating the export when we delete a batch.

We could get around this by setting `deleted_at` on batches instead of actually deleting batches (like we do for sheets), but before I move forward with that implementation, I wanted to pause and see if 1. we think it's worth doing that work, or if we are willing to move forward with manually invalidating backups 2. if there is a more simple/elegant way to do this that I'm not thinking of.

taravancil:
I still need to implement this feature on VxScan, and I'm also waiting on some final design feedback. I'll leave it up to reviewers if they want to do another round of review now or wait until this is out of draft status.

benadida:
> @benadida called out that manually invalidating the backup when we add/delete batches/sheets is brittle, because it requires us to remember to invalidate backups if/when we add new features. He made a good suggestion that we could compare export timestamps to the max of `batches.ended_at`, `sheets.created_at`, `sheets.deleted_at` to decide if a backup is up-to-date. I like this idea, but it still requires manually invalidating the export when we delete a batch.
> 
> We could get around this by setting `deleted_at` on batches instead of actually deleting batches (like we do for sheets), but before I move forward with that implementation, I wanted to pause and see if 1. we think it's worth doing that work, or if we are willing to move forward with manually invalidating backups 2. if there is a more simple/elegant way to do this that I'm not thinking of.

Unless we fully delete all ballot data, I think it makes sense to only mark things as deleted, rather than to actually delete them from the database. So I would suggest we do that, which ... does expand the scope here, but it is scope that is making our system better!

eventualbuddha:
I agree that we should do soft deletes of batches. In fact, I'd suggest doing that as a precursor to this task that could be its own PR.

taravancil:
> I agree that we should do soft deletes of batches. In fact, I'd suggest doing that as a precursor to this task that could be its own PR.

Good idea. I'll file a ticket.

frontends/bsd/src/screens/admin_actions_screen.tsx
beausmith: Oof. Perhaps place a `<br />` before the `<span>`? so that  the text is under the button and more tightly associated with the button? (sorry for the delay in response)

services/scan/src/store.ts
```diff
@@ -375,6 +381,95 @@ export class Store {
...
+      key
+    ) as { value: Iso8601Timestamp } | undefined;
+    return row?.value;
+  }
+
+  /**
+   * Returns whether the appropriate backups have been completed and it is safe
+   * to unconfigure a machine/zero out data. Always returns true in test mode.
+   */
+  getCanUnconfigure(): boolean {
```
taravancil, line 417: I think this is the most important piece for reviewers to spend time on.

Specifically, if someone better at SQL is interested in helping me capture this heuristic in a single query, that would be awesome!

services/scan/src/store.ts
```diff
@@ -375,6 +381,95 @@ export class Store {
...
+    ) as { maxCvrsCreatedAt: Iso8601Timestamp };
+    const { maxCvrsDeletedAt } = this.client.one(
+      'select max(deleted_at) as maxCvrsDeletedAt from sheets'
+    ) as { maxCvrsDeletedAt: Iso8601Timestamp };
+    const { maxBatchesStartedAt } = this.client.one(
+      'select max(started_at) as maxBatchesStartedAt from batches'
+    ) as { maxBatchesStartedAt: Iso8601Timestamp };
+    const { maxBatchesDeletedAt } = this.client.one(
+      'select max(deleted_at) as maxBatchesDeletedAt from batches'
+    ) as { maxBatchesDeletedAt: Iso8601Timestamp };
```
benadida, line 446: You could do this in two SQL statements, e.g.:

`select max(created_at) as maxCVrsCreatedAt, max(deleted_at) as maxCvrsDeletedAt from sheets`

services/scan/src/store.ts
```diff
@@ -375,6 +381,95 @@ export class Store {
...
+
+    // Filter out null values
+    const cvrsInvalidatingEvents = [maxCvrsCreatedAt, maxCvrsDeletedAt].filter(
+      Boolean
+    );
+    const batchesInvalidatingEvents: Iso8601Timestamp[] = [
+      maxBatchesStartedAt,
+      maxBatchesDeletedAt,
+    ].filter(Boolean);
```
benadida, line 455: possibly simplify by putting this logic in the `reduce`, e.g.:

`(max, curr) => ((max && max > curr) ?  max : curr),`

services/scan/src/store.ts
benadida: there's a way to do this in two simple SQL queries, I think, which, depending on your pov, might be ugly, or might be a nice encapsulation of the whole thing.

`select max(max(created_at, deleted_at)) <= ? from sheets`

`select max(max(created_at, deleted_at)) <= ? from batches`

If both are true, you're good to go.

benadida (approved):
This looks correct to me with one minor comment and one minor concern.

The minor comment: I think the code for `getCanUnconfigure` might be simplifiable a good bit, see my suggestion.

The minor concern: we're hammering sqlite on every `status` call which happens often, so doing a number of SQL queries every time gives me pause... it's not that much data so it's probably fine, but maybe worth a followup performance test ticket? (Also maybe worth an index on the timestamps in the `sheets` table.)

I'm good to move forward, though would suggest that followup performance check ticket, which is broader than just this ticket anyways since there is other code that, on `status`, is hammering sqlite.

services/scan/src/store.ts
taravancil: Good idea! Done.

services/scan/src/store.ts
taravancil: That doesn't handle if `curr` is `null`, so I'm going to keep using the `.filter` approach`. I think the intention that way is more clear, and since `n` is small I don't feel too bad about looping multiple times. We can and probably should revisit this in the followup perf ticket I'll file.

PR #1586, opened 2022-03-15
Fix various state updates after unmount errors etc

eventualbuddha:
## Overview

In working on a task in `election-manager` I found that my errors were being swallowed. This turned out to be due to `app.test.tsx` ignoring all calls to `console.error`. Upon removing the code that did that, I found there were several update-state-after-unmount errors occurring. This PR re-enables `console.error` and fixes the errors.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated testing.

## Checklist

### DIFF
--- a/libs/ui/src/hooks/use_devices.ts
+++ b/libs/ui/src/hooks/use_devices.ts
@@ -10,6 +10,7 @@ import { map } from 'rxjs/operators';
 import { LogEventId, Logger } from '@votingworks/logging';
 import useInterval from 'use-interval';
 import { usePrevious } from '..';
+import { useCancelablePromise } from './use_cancelable_promise';
 
 export const LOW_BATTERY_THRESHOLD = 0.25;
 export const BATTERY_POLLING_INTERVAL = 3000;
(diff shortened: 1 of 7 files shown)


libs/ui/src/hooks/use_devices.ts
```diff
@@ -78,7 +80,7 @@ export function useDevices({ hardware, logger }: Props): Devices {
 
   useInterval(
     async () => {
-      setBattery(await hardware.readBatteryStatus());
+      setBattery(await makeCancelable(hardware.readBatteryStatus()));
```
jonahkagan, line 83: I would have thought that `useInterval` would clear the interval on unmount. I don't think `hardware.readBatteryStatus()` would take long enough to start before unmount and finish after, especially in test. So I'm surprised this needs the `makeCancelable`. Do you have any insight into why it was needed?

eventualbuddha: It's true that `useInterval` will not _call_ its callback after unmount, but the call to `hardware.readBatteryStatus` necessarily resolves only after the current tick. Since the component may be unmounted in this tick it is possible, _especially_ in a test, that the call to `setBattery` happens after unmount.

jonahkagan: Makes sense

jonahkagan (approved):

PR #1585, opened 2022-03-15
feat(em): export HMPB layout with ballot package

eventualbuddha:
## Overview

Refs #1558

Pulls `ballot-interpreter-vx` into `election-manager` in order to analyze the layout of the ballots during export. This required getting PDF.js working in kiosk-browser, which required serving `pdf.worker.js` in a way that PDF.js could load it. This required modifying the proxy for `election-manager` to support loading from `pdfjs-dist`'s directory. Otherwise, it was pretty straightforward. In a future commit, I'll modify the `scan` service to accept the layout information from the ballot package.

## Demo Video or Screenshot
```
❯ unzip -l ~/Desktop/choctaw-county_2020-general-election_e5a8525177__2022-03-14_11-26-27.zip 
Archive:  /home/brian/Desktop/choctaw-county_2020-general-election_e5a8525177__2022-03-14_11-26-27.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
    25849  2022-03-14 18:26   election.json
    12856  2022-03-14 18:26   manifest.json
    19846  2022-03-14 18:26   election-e5a8525177-precinct-east-weir-id-6525-style-1-English-live-layout.json
   216552  2022-03-14 18:26   election-e5a8525177-precinct-east-weir-id-6525-style-1-English-live.pdf
    19844  2022-03-14 18:26   election-e5a8525177-precinct-east-weir-id-6525-style-1-English-test-layout.json
   216149  2022-03-14 18:26   election-e5a8525177-precinct-east-weir-id-6525-style-1-English-test.pdf
    19846  2022-03-14 18:26   election-e5a8525177-precinct-panhandle-id-6532-style-1-English-live-layout.json
   216369  2022-03-14 18:26   election-e5a8525177-precinct-panhandle-id-6532-style-1-English-live.pdf
…snip…
---------                     -------
  6183436                     54 files
```

```json5
[
  {
    "pageSize": {
      "width": 1224,
      "height": 2016
    },
    "metadata": {
      "ballotStyleId": "4",
      "electionHash": "e5a852517758f69040540c22b072bb4e7c41fabd10ef648e1c02dac9a46b8394",
      "ballotType": 0,
      "precinctId": "6538",
      "locales": {
        "primary": "en-US"
      },
      "isTestMode": false,
      "pageNumber": 1
    },
    "contests": [
      {
        "bounds": {
          "x": 429,
          "y": 43,
          "width": 366,
          "height": 1384
        },
        "corners": [
          {
            "x": 429,
[...]

PR #1588, opened 2022-03-16
Prepare for ballot package changes

eventualbuddha:
## Overview

In trying to get the tests working for https://github.com/votingworks/vxsuite/pull/1585 I discovered that the way `app.test.tsx` was testing the ballot package export was broken, but only if there were a series of files being added to the package that were non-empty. `zip-stream` appears to operate synchronously when the files added to the zip stream are empty, but asynchronously when there is data to actually compress. Because we'd mocked `printToPDF` to resolve to an empty buffer, the file data has so far been empty. Writing the layout information for each ballot was not empty, and so we triggered the asynchronous behavior of `zip-stream`. This meant that when we tried to check that the download had completed by looking at the calls to `kiosk.log` we would not find it.

I tried a variety of things to get it to wait long enough, but in the end the only thing that worked was to call `jest.advanceTimersByTime(0)` _while fake timers were disabled_. This, of course, made jest issue a warning but it did work. My theory as to why is that when the application is started and interacted with using fake timers, that the `zip-stream` library was caching a reference to the fake `setTimeout` somehow and that, despite calling `jest.useRealTimers` later, it still held on to the fake timer which would still respond to `jest.advanceTimersByTime`. I didn't dig deep enough to prove this, but decided to just cut my losses and remove the ballot package testing from `app.test.tsx` since it's already pretty well tested by its component test. This removes the `jest.useRealTimers` call that was mucking things up, too. In general we should _not_ mix real and fake timers in a single test or, even better, in a single test file.

This PR also includes a few other minor fixes I made along the way.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated testing.

## Checklist

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -13,12 +13,7 @@ import {
 } from '@testing-library/react';
 import fetchMock from 'fetch-mock';
 import { electionWithMsEitherNeitherWithDataFiles } from '@votingworks/fixtures';
-import {
-  sleep,
-  MemoryStorage,
-  MemoryCard,
-  MemoryHardware,
-} from '@votingworks/utils';
+import { MemoryStorage, MemoryCard, MemoryHardware } from '@votingworks/utils';
 import {
   advanceTimersAndPromises,
   fakeKiosk,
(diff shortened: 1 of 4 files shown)


jonahkagan (approved):
Sounds like a tough situation!

PR #1589, opened 2022-03-16
chore(deps): Add fsevents as optionalDependency to fix prod build

jonahkagan:
## Overview
There was a prod build error that had been ignored for a long time. fsevents is an optional dependency of some vxsuite subdependencies. fsevents is only used on Mac, not Linux, so it should have been skipped, but for some reason the prod build script was trying to install it. This explicitly names it as a top-level optional dependency, which seems to make it get skipped correctly.

## Demo Video or Screenshot
N/A

## Testing Plan 
Manually tested

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,5 +16,8 @@
     "pnpm": "^5.18.10",
     "prettier": "^2.3.2",
     "sort-package-json": "^1.50.0"
+  },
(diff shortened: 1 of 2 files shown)


jonahkagan:
Seems like we didn't need this fix

Issue #1464, opened 2022-03-03
VxAdmin Increase width of Import Cvrs Modal


carolinemodic:
The import cvrs modal holds a lot of information in a small modal, we should make it take up more of the screen (I think there is a maxWidth parameter or something we can give the Modal component). VxAdmin is found in frontends/election-manager.

carolinemodic:
context: https://votingworks.slack.com/archives/CJU9MSC6S/p1646235325102169

carolinemodic:
After talking with @taravancil offline we realized that you need to run this through kiosk-browser with a properly formatted usb drive to test so it will be easier for me to take this one.

PR #1601, opened 2022-03-17
fix(em) fix bug with import cvr modal width

carolinemodic:
## Overview

https://github.com/votingworks/vxsuite/issues/1464
This is actually a bug that was introduced when we got rid of the Modal css files. @beausmith let me know if you think this approach is ok of having a ModalWidth variable to expand the width in specific cases. 

## Demo Video or Screenshot
Before:
Screen Shot 2022-03-17 at 11 07 22 AM

After:

Screen Shot 2022-03-17 at 11 33 09 AM

## Testing Plan 
Loaded and saw wider modal, loaded other modals and they were the same size as before.

## Checklist

### DIFF
--- a/frontends/election-manager/src/components/import_cvrfiles_modal.tsx
+++ b/frontends/election-manager/src/components/import_cvrfiles_modal.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import path from 'path';
 import moment from 'moment';
 
-import { Modal, Table, TD } from '@votingworks/ui';
+import { Modal, ModalWidth, Table, TD } from '@votingworks/ui';
 import {
   assert,
   generateElectionBasedSubfolderName,
@@ -46,10 +46,6 @@ const LabelText = styled.span`
 
 const TestMode = styled.span`
   color: #ff8c00;
-`;
-
-const ImportCvrsModalContent = styled.div`
-  max-width: 45rem;
 `;
 
 enum ModalState {
@@ -340,7 +336,7 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
           <Prose>
             <h1>No USB Drive Detected</h1>
             <p>
-              <UsbImage src="usb-drive.svg" alt="Insert USB Image" />
+              <UsbImage src="assets/usb-drive.svg" alt="Insert USB Image" />
               Please insert a USB drive in order to import CVR files from the
               scanner.
             </p>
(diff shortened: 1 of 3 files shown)


frontends/election-manager/src/components/import_cvrfiles_modal.tsx
```diff
@@ -340,7 +336,7 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
           <Prose>
             <h1>No USB Drive Detected</h1>
             <p>
-              <UsbImage src="usb-drive.svg" alt="Insert USB Image" />
```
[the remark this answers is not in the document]
carolinemodic: oh wait the new path is the right one though, I think I had this change in VsCode from another PR I already merged and forgot to save this file... so we should update this.

libs/ui/src/modal.test.tsx
```diff
@@ -31,13 +31,56 @@ describe('Modal', () => {
...
+        role="alertdialog"
+        tabindex="-1"
+      >
+        <div
+          class="sc-hKwCoD RDMJL"
+        >
+          Do you want to do the thing?
+        </div>
+      </div>
+    `);
```
jonahkagan, line 49: I personally don't love using snapshots for HTML content due to the styled-components classnames, which seem to change often. Was there something specific you were trying to test for here?

libs/ui/src/modal.tsx
```diff
@@ -86,6 +97,7 @@ interface Props {
   actions?: ReactNode;
   onAfterOpen?: () => void;
   onOverlayClick?: () => void;
+  modalWidth?: ModalWidth;
```
jonahkagan, line 100: could also just take a `style` prop and apply it to the content div - might be more flexible for future style tweaks

jonahkagan (commented):
(just passing through and thought i'd give some thoughts since i caused this bug in the first place)

libs/ui/src/modal.test.tsx
carolinemodic: Yes I agree, but libs/ui has a 100% code coverage requirement so it was forcing me to add a test for the Wide modal case, because the css change is underneath the @media(min-width: 480) {} block, it doesn't automatically work with the expect().toHaveStyle() function and I did some googling and found a few threads of people saying its impossible to test things like that. However it does generate a different css classname in the snapshot tests so given the urgency with m13b and the relative unimportance of testing this well that seemed like the easiest thing to do for now to get this change in.

jonahkagan: Oh that's annoying

PR #1620, opened 2022-03-22
fix(bmd): Fix Text component icon props

jonahkagan:
## Overview

When passing `warningIcon={false}` and `voteIcon={true}`, the vote icon
wouldn't be rendered. I found this issue in a number of text components in different frontends and libs/ui. 

## Testing Plan 
Added regression test in libs/ui, since that's probably the component consolidate around in the future. Also found a great new library to improve testing with styled-components.

## Checklist


eventualbuddha (approved):
This looks like it was my fault.

PR #1631, opened 2022-03-24
feat(em): convert from NH definition

eventualbuddha:
## Overview

[](https://github.com/eventualbuddha)Allows running VxAdmin with `VX_CONVERTER=nh`, which replaces the MS SEMS election definition converter with one for NH. Doing so will present a form with file fields for the XML and PDF files needed to convert to our election format. Once both have been entered, the conversion will happen and the election will be loaded.

Unfortunately, this doesn't handle errors very well right now since the existing infrastructure didn't. We should come back to that and make it capable of showing errors that occurred during conversion.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/1938/159815594-ffaf55c7-9935-4420-afcf-a3ed6a935c5e.mov

## Testing Plan 
Manually tested with one election definition. Added some automated testing.

## Checklist

### DIFF
--- a/frontends/election-manager/prodserver/setupProxy.js
+++ b/frontends/election-manager/prodserver/setupProxy.js
@@ -49,7 +49,7 @@ module.exports = function (app) {
       machineId: process.env.VX_MACHINE_ID || '0000',
       codeVersion: process.env.VX_CODE_VERSION || 'dev',
       bypassAuthentication: asBoolean(process.env.BYPASS_AUTHENTICATION),
-      converter: process.env.VX_CONVERTER || 'ms',
+      converter: process.env.VX_CONVERTER || 'ms-sems',
     });
   });
 
--- a/frontends/election-manager/prodserver/setupProxy.js
+++ b/frontends/election-manager/prodserver/setupProxy.js
@@ -49,6 +49,7 @@ module.exports = function (app) {
       machineId: process.env.VX_MACHINE_ID || '0000',
       codeVersion: process.env.VX_CODE_VERSION || 'dev',
       bypassAuthentication: asBoolean(process.env.BYPASS_AUTHENTICATION),
+      converter: process.env.VX_CONVERTER || 'ms',
     });
   });
 
--- a/frontends/election-manager/src/screens/unconfigured_screen.tsx
+++ b/frontends/election-manager/src/screens/unconfigured_screen.tsx
@@ -259,7 +259,7 @@ export function UnconfiguredScreen({
             ) : (
               <p key={file.name}>
                 <FileInputButton
-                  // accept=".txt"
+                  accept={file.accept}
                   buttonProps={{
                     fullWidth: true,
                   }}
--- a/frontends/election-manager/src/screens/unconfigured_screen.tsx
+++ b/frontends/election-manager/src/screens/unconfigured_screen.tsx
@@ -4,10 +4,15 @@ import styled from 'styled-components';
 import { parseElection } from '@votingworks/types';
 
 import { Modal, useCancelablePromise } from '@votingworks/ui';
-import { ConverterClient, VxFile } from '../lib/converter_client';
+import { assert } from '@votingworks/utils';
+import {
+  ConverterClient,
+  getElectionDefinitionConverterClient,
+  VxFile,
+} from '../lib/converters';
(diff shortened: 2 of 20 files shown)


frontends/election-manager/src/screens/unconfigured_screen.tsx
```diff
@@ -65,8 +74,12 @@ export function UnconfiguredScreen(): JSX.Element {
   );
   const [isLoading, setIsLoading] = useState(true);
   const [vxElectionFileIsInvalid, setVxElectionFileIsInvalid] = useState(false);
-  const [client] = useState(new ConverterClient('election'));
-  const [isConvertSems, setIsConvertSems] = useState(false);
+  const [client, setClient] = useState<ConverterClient>();
```
jonahkagan, line 77: does this need to be state? it looks like `client` is just derived from props and then never set again

frontends/election-manager/src/screens/unconfigured_screen.tsx
```diff
@@ -229,19 +247,19 @@ export function UnconfiguredScreen(): JSX.Element {
...
           <p> Select the following files from a USB drive, etc.</p>
-          {inputConversionFiles.map((file: VxFile) =>
+          {inputConversionFiles.map((file) =>
             file.path ? (
               <Loaded key={file.name}>{`Loaded ${file.name}`}</Loaded>
             ) : (
               <p key={file.name}>
                 <FileInputButton
-                  accept=".txt"
+                  // accept=".txt"
```
jonahkagan, line 262: restore or remove?

frontends/election-manager/prodserver/setupProxy.js
```diff
@@ -49,6 +49,7 @@ module.exports = function (app) {
       machineId: process.env.VX_MACHINE_ID || '0000',
       codeVersion: process.env.VX_CODE_VERSION || 'dev',
       bypassAuthentication: asBoolean(process.env.BYPASS_AUTHENTICATION),
+      converter: process.env.VX_CONVERTER || 'ms',
```
jonahkagan, line 52: add to README?

frontends/election-manager/src/screens/unconfigured_screen.tsx
```diff
@@ -229,19 +247,19 @@ export function UnconfiguredScreen(): JSX.Element {
     );
   }
 
-  if (isConvertSems && inputConversionFiles.length > 0) {
+  if (isUsingConverter && inputConversionFiles.length > 0) {
     return (
       <NavigationScreen mainChildCenter>
         <Prose textCenter>
-          <h1>Convert from SEMS files</h1>
+          <h1>Convert from {client?.getDisplayName()} files</h1>
```
jonahkagan, line 254: I noticed in the video that uploading the first file gives some nice visual feedback ("Loaded NH Card something something" in green) and then upon uploading the second file, there's no feedback and it flashes through one "Loading" screen and then the "Election loading..." modal. Two thoughts:
1. I think it would be nice to have confirmation of the second file and then get to click a button to start the actual processing of the files
2. Any way to remove the flashed "Loading" screen before the modal kicks in?

These suggestions may be out of scope but figured I'd raise them.

frontends/election-manager/src/screens/unconfigured_screen.tsx
eventualbuddha: It does for the NH one because the state is stored within the class. We didn't have to for the MS one because the state was actually stored server side.

frontends/election-manager/src/screens/unconfigured_screen.tsx
eventualbuddha: These are definitely problems with the current implementation, but they existed before so I'm going to leave it as-is for now. However, I created https://github.com/votingworks/vxsuite/issues/1633 to ensure this feedback isn't lost.

frontends/election-manager/src/screens/unconfigured_screen.tsx
jonahkagan: as we discussed, this seems ok for now. might be worth exploring separating out the code that needs the converter into a subcomponent in order to clarify where `client` is actually needed so we don't need to assert whether it's defined or not

jonahkagan (approved):

PR #1651, opened 2022-03-29
fix(scan): confirm delete batch on enter

eventualbuddha:
## Overview

Cause the 'Yes, Delete Batch' button to be focused when showing the modal so that pressing Enter or Space will confirm the delete. Escape already works to dismiss it, but with this change you can also press Tab to focus the Cancel button and select it with Enter or Space.

I only made this change in one place to ask whether this is the way we want to do this. In general I am supportive of using built-in browser features, and this seems to fit the desired behavior. I'm open to alternative suggestions, though.

## Demo Video or Screenshot
https://user-images.githubusercontent.com/1938/160702635-a9cb50da-e4b4-4dab-9aca-8721502d7bfe.mov

## Testing Plan 
Manually tested.

## Checklist

### DIFF
--- a/frontends/bsd/src/screens/dashboard_screen.tsx
+++ b/frontends/bsd/src/screens/dashboard_screen.tsx
@@ -170,6 +170,7 @@ export function DashboardScreen({
                 danger
                 onPress={confirmDeleteBatch}
                 disabled={isDeletingBatch}
+                autoFocus
               >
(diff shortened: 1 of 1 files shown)


benadida (approved):
I'm good with this. Given that tapping/clicking delete the first time is not likely to be done with the enter key, I don't think this will lead to erroneous deleting of batches.

@beausmith shout if you disagree!

PR #1656, opened 2022-03-30
refactor(scan): simplify write-in reason info type

eventualbuddha:
## Overview

Uses a named union type to reduce some of the duplication in this file.

## Demo Video or Screenshot
n/a

## Testing Plan 
n/a

## Checklist

### DIFF
--- a/frontends/bsd/src/screens/write_in_adjudication_screen.tsx
+++ b/frontends/bsd/src/screens/write_in_adjudication_screen.tsx
@@ -12,7 +12,6 @@ import {
   ContestId,
   ElectionDefinition,
   Rect,
-  UnmarkedWriteInAdjudicationReasonInfo,
   WriteInAdjudicationReasonInfo,
   WriteInMarkAdjudication,
 } from '@votingworks/types';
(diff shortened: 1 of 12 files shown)


frontends/bsd/src/screens/write_in_adjudication_screen.tsx
```diff
@@ -45,6 +45,9 @@ import { Prose } from '../components/prose';
 import { Screen } from '../components/screen';
 import { AppContext } from '../contexts/app_context';
 
+type AnyWriteInAdjudicationReasonInfo =
+  | WriteInAdjudicationReasonInfo
+  | UnmarkedWriteInAdjudicationReasonInfo;
```
jonahkagan, line 50: How about this? I think if I were using these types I might assume accidentally that `WriteInAdjudicationReasonInfo` is the main type

```suggestion
type WriteInAdjudicationReasonInfo =
  | MarkedWriteInAdjudicationReasonInfo
  | UnmarkedWriteInAdjudicationReasonInfo;
```

benadida (commented):
lgtm, leaving this as comment so you get a chance to address @jonahkagan's suggestion without auto-merge :)

eventualbuddha:
> lgtm, leaving this as comment so you get a chance to address @jonahkagan's suggestion without auto-merge :)

Disabled auto-merge. Feel free to approve if you think this would be good after doing the refactor @jonahkagan suggested.

benadida (approved):
cool, agree with @jonahkagan's comments, and otherwise lgtm

frontends/bsd/src/screens/write_in_adjudication_screen.tsx
eventualbuddha: I changed this to make a broader global change like you suggested.

PR #1659, opened 2022-03-30
fix(scan): ensure write-in adjudication works front & back

eventualbuddha:
## Overview

There was an outstanding bug with write-in adjudication where if both the front and back had write-ins to adjudicate _and_ the front had more contests with write-ins than the back then it would crash. Even if the front had the same or fewer it still wouldn't behave correctly, and would instead start at the index of the last contest on the front when adjudicating the back.

The problem was that we were managing the `selectedContestId` at the wrong level of the component hierarchy. This changes it to manage the selected contest ID at the `WriteInAdjudicationScreen` level so that it can reset it properly when finishing a contest rather than letting a stale value from the previous page's contest list stay in state.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/1938/160939334-9527a1e5-0fef-45eb-86ea-ec5a87b987b4.mov

## Testing Plan 
Tested manually, updated automated tests.

## Checklist

### DIFF
--- a/frontends/bsd/src/screens/write_in_adjudication_screen.tsx
+++ b/frontends/bsd/src/screens/write_in_adjudication_screen.tsx
@@ -10,7 +10,7 @@ import {
   BallotPageContestLayout,
   BallotPageLayout,
   CandidateContest,
-  Contest,
+  ContestId,
   ElectionDefinition,
   InterpretedHmpbPage,
   Rect,
@@ -29,7 +29,6 @@ import { assert, find } from '@votingworks/utils';
 import React, {
   useCallback,
   useContext,
-  useEffect,
   useLayoutEffect,
   useRef,
   useState,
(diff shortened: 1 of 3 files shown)


frontends/bsd/src/screens/write_in_adjudication_screen.tsx
```diff
@@ -641,6 +640,7 @@ export function WriteInAdjudicationScreen({
         prev
       )
     );
+    setSelectedContestId(undefined);
```
eventualbuddha, line 643: Most of the rest of the changes are in support of this line.

jonahkagan (commented):
I think the control flow here will be simpler if you remove the `WriteInAdjudicationByContest` component and move its contents into `WriteInAdjudicationScreen`. I'm noticing that there's a lot of stuff that needs to get passed into the child component to manage navigation between contests that might be accomplished more easily in the parent.

Issue #1658, opened 2022-03-30
NH: make write-in adjudication work with timing mark ballots


eventualbuddha:
## Summary
Write-in adjudication works by showing the cropped & zoomed part of the ballot corresponding to a write-in area on the ballot. To do this it depends on knowing the layout of the ballot as stored in the database. We do not have this with timing mark ballots, because they use a different mechanism.

## Requirements
The write-in adjudication screen should at least roughly show the correct cropped area when adjudicating write-ins.

## Non-Requirements
- The backend for layout handling does not need to be unified.
- We do not need to know the layout of the _scanned_ ballot or otherwise normalize it so long as the write-in area is shown mostly correctly.

## Tasks
- [ ] Determine the expected layout of a perfectly scanned timing mark ballot and use that when adjudicating write-ins.

PR #1664, opened 2022-03-31
fix(scan): make write-in adjudication work for timing mark ballots

eventualbuddha:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/1658

This is only an approximate fix. It generates the expected layout for the timing mark ballots, but does not base it on the real interpreted ballot. Thus, it's possible for the coordinates to be somewhat off, especially if a slightly incorrect scaling factor was used. In the future, we should generate a layout based on the `interpret` output for timing mark ballots, which should include positional information for the ovals at minimum.

There were actually a couple other bugs that this PR fixes in multiple commits. It's probably best to look one commit at a time.

## Demo Video or Screenshot
image

## Testing Plan 
Manually tested with a couple different timing mark ballots with write-ins. I was able to get all the way through adjudication and the ballot was accepted.

## Checklist

### DIFF
--- a/frontends/bsd/src/screens/write_in_adjudication_screen.tsx
+++ b/frontends/bsd/src/screens/write_in_adjudication_screen.tsx
@@ -409,14 +409,43 @@ function ContestAdjudication({
 }
 
 export interface Props {
+  /**
+   * Database ID of the sheet being adjudicated.
+   */
   sheetId: string;
+
+  /**
+   * Which side of the sheet being adjudicated.
+   */
   side: Side;
+
+  /**
+   * URL of the image to the whole ballot page being adjudicated.
+   */
   imageUrl: string;
+
+  /**
+   * All write-ins flagged for adjudication on this page.
+   */
   writeIns: ReadonlyArray<
     WriteInAdjudicationReasonInfo | UnmarkedWriteInAdjudicationReasonInfo
   >;
+
+  /**
+   * Layout of the ballot page being adjudicated. The number of contests in this
+   * layout must match the number of contests in `allContestIds`.
+   */
   layout: BallotPageLayout;
-  contestIds: readonly ContestId[];
+
+  /**
+   * Contest IDs for every contest on this page, even those without write-ins.
+   */
+  allContestIds: readonly ContestId[];
+
+  /**
+   * Callback for when every contest with write-ins has been adjudicated and
+   * the user has chosen to save their adjudication.
+   */
   onAdjudicationComplete?(
     sheetId: string,
     side: Side,
@@ -430,7 +459,7 @@ export function WriteInAdjudicationScreen({
   imageUrl,
   writeIns,
   layout,
-  contestIds,
+  allContestIds,
   onAdjudicationComplete,
 }: Props): JSX.Element {
   const { electionDefinition, storage } = useContext(AppContext);
@@ -443,11 +472,10 @@ export function WriteInAdjudicationScreen({
   const [selectedContestId, setSelectedContestId] = useState<ContestId>();
   const [isSaving, setIsSaving] = useState(false);
 
-  const contestsWithWriteIns = useMemo(
-    () => new Set([...writeIns].map(({ contestId }) => contestId)),
+  const contestsWithWriteInsIds = useMemo(
+    () => uniq([...writeIns].map(({ contestId }) => contestId)),
     [writeIns]
   );
-  const contestsWithWriteInsCount = contestsWithWriteIns.size;
   const styleForContest = useCallback(
     (contestId: ContestId): React.CSSProperties => {
       return contestId === selectedContestId
@@ -502,13 +530,15 @@ export function WriteInAdjudicationScreen({
     await onAdjudicationComplete?.(sheetId, side, adjudications);
   }, [adjudications, onAdjudicationComplete, setWriteInPresets, sheetId, side]);
 
-  const selectedContestIdOrDefault = selectedContestId ?? contestIds[0];
-  const selectedContestIndex = contestIds.indexOf(selectedContestIdOrDefault);
-  const isFirstContestSelected = selectedContestIndex === 0;
+  const selectedContestIdOrDefault =
+    selectedContestId ?? contestsWithWriteInsIds[0];
+  const isFirstContestSelected =
+    selectedContestIdOrDefault === contestsWithWriteInsIds[0];
   const isLastContestSelected =
-    selectedContestIndex === contestsWithWriteInsCount - 1;
+    selectedContestIdOrDefault ===
+    contestsWithWriteInsIds[contestsWithWriteInsIds.length - 1];
 
-  const contestsWithWriteInsIndex = [...contestsWithWriteIns].findIndex(
+  const contestsWithWriteInsIndex = [...contestsWithWriteInsIds].findIndex(
     (id) => id === selectedContestIdOrDefault
   );
   const contest = find(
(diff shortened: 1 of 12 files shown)


libs/ballot-interpreter-nh/src/layout.ts
```diff
@@ -0,0 +1,298 @@
...
+
+      contests.push({
+        options,
+        bounds,
+        corners: [
+          { x: bounds.x, y: bounds.y },
+          { x: bounds.x + bounds.width - 1, y: bounds.y },
+          { x: bounds.x, y: bounds.y + bounds.height - 1 },
+          { x: bounds.x + bounds.width - 1, y: bounds.y + bounds.height - 1 },
+        ],
```
jonahkagan, line 301: If `corners` is just a function of `bounds`, why specify both?

services/scan/src/store.ts
```diff
@@ -898,7 +899,15 @@ export class Store {
 
   getBallotLayoutsForMetadata(
     metadata: BallotPageMetadata
-  ): BallotPageLayout[] {
+  ): BallotPageLayout[] | undefined {
+    const electionDefinition = this.getElectionDefinition();
+    if (electionDefinition?.election.gridLayouts) {
```
jonahkagan, line 904: this case could use an explanatory comment

services/scan/src/store.ts
```diff
@@ -898,7 +899,15 @@ export class Store {
 
   getBallotLayoutsForMetadata(
     metadata: BallotPageMetadata
-  ): BallotPageLayout[] {
+  ): BallotPageLayout[] | undefined {
```
jonahkagan, line 902: why not pass through the possible error?

frontends/bsd/src/screens/write_in_adjudication_screen.tsx
```diff
@@ -502,13 +533,15 @@ export function WriteInAdjudicationScreen({
     await onAdjudicationComplete?.(sheetId, side, adjudications);
   }, [adjudications, onAdjudicationComplete, setWriteInPresets, sheetId, side]);
 
-  const selectedContestIdOrDefault = selectedContestId ?? contestIds[0];
-  const selectedContestIndex = contestIds.indexOf(selectedContestIdOrDefault);
-  const isFirstContestSelected = selectedContestIndex === 0;
+  const selectedContestIdOrDefault =
+    selectedContestId ?? contestsWithWriteInsIds[0];
```
jonahkagan, line 537: do you know why we have a default here instead of using a default in the `useState` for `selectedContestId`? It seems once we're done adjudicating all of the contests this component will get unmounted, so it won't need a default anymore

frontends/bsd/src/screens/write_in_adjudication_screen.tsx
```diff
@@ -529,8 +563,12 @@ export function WriteInAdjudicationScreen({
   );
 
   const goPrevious = useCallback(() => {
-    setSelectedContestId(contestIds[selectedContestIndex - 1]);
-  }, [contestIds, selectedContestIndex]);
+    setSelectedContestId(
+      contestsWithWriteInsIds[
+        contestsWithWriteInsIds.indexOf(selectedContestIdOrDefault) - 1
```
jonahkagan, line 568:
```suggestion
        contestsWithWriteInsIndex - 1
```

frontends/bsd/src/screens/write_in_adjudication_screen.tsx
```diff
@@ -543,15 +581,19 @@ export function WriteInAdjudicationScreen({
           setIsSaving(false);
         }
       } else {
-        setSelectedContestId(contestIds[selectedContestIndex + 1]);
+        setSelectedContestId(
+          contestsWithWriteInsIds[
+            contestsWithWriteInsIds.indexOf(selectedContestIdOrDefault) + 1
```
jonahkagan, line 586:
```suggestion
            contestsWithWriteInsIndex + 1
```

libs/ballot-interpreter-nh/src/layout.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    ballotStyleId: electionDefinition.election.ballotStyles[0]!.id,
+    precinctId: electionDefinition.election.precincts[0]!.id,
+    ballotType: BallotType.Standard,
+    isTestMode: true,
+    electionHash: electionDefinition.electionHash,
+    locales: { primary: 'en-US' },
+  }).unsafeUnwrap();
+
+  await (
+    await import('fs')
```
jonahkagan, line 49: why late import here?

jonahkagan (commented):
Is it possible to add regression tests for the two bug fixes?

libs/ballot-interpreter-nh/src/layout.ts
eventualbuddha: The `BallotPageLayout` type applies not only to templates, where this is true, but also to scanned images where it may not be.

frontends/bsd/src/screens/write_in_adjudication_screen.tsx
eventualbuddha: I needed a way to make the component automatically move to the first contest of the back page after completing the front, so I set the selected contest ID to `undefined` (L529) so that we can get the first one automatically when the new `allContestIds` prop comes through for the back side. If I had done `const [selectedContestId, setSelectedContestId] = useState(contestsWithWriteInsIds[0])` instead, then I'd have no good way to clear the stale state of the front page's last selected contest when adjudicating the front was complete.

libs/ballot-interpreter-nh/src/layout.test.ts
eventualbuddha: This was some debug code. Removing.

services/scan/src/store.ts
eventualbuddha: This was from an earlier iteration and is no longer needed.

frontends/bsd/src/screens/write_in_adjudication_screen.tsx
jonahkagan: Ah interesting. It's not totally clear as the reader, since you can't easily see that the component will be rerendered with a different sheet on completion of adjudication. I would suggest:
- at a minimum adding a comment
- potentially trying to colocate the selected sheet and selected contest state so that the dependency can be seen clearly
- add a key prop to this component in order to reset the state when changing sheets (see https://twitter.com/sebmarkbage/status/1262937670444974081). I think there's a valid argument for saying this is a "master/detail" situation

jonahkagan: More on that here: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key

eventualbuddha: Thanks for digging into this. Using a `key` is certainly a better approach, which I've applied in 97522ab8c3ac4c65952c97fa500e3eb94ecc664f.

jonahkagan: Awesome. Yeah I've run into problems like this a few times in Arlo and have been trying to get my head around how to solve them. Still feels tricky for me

jonahkagan (approved):
One tiny nit left. Thanks for working through all the feedback. I think things ended up in a much clearer place!

frontends/bsd/src/screens/write_in_adjudication_screen.tsx
```diff
@@ -498,27 +531,27 @@ export function WriteInAdjudicationScreen({
...
+    selectedContestId === contestsWithWriteInsIds[0];
   const isLastContestSelected =
-    selectedContestIndex === contestsWithWriteInsCount - 1;
+    selectedContestId ===
+    contestsWithWriteInsIds[contestsWithWriteInsIds.length - 1];
 
-  const contestsWithWriteInsIndex = [...contestsWithWriteIns].findIndex(
-    (id) => id === selectedContestIdOrDefault
+  const selectedContestIdOrDefaultIndex = contestsWithWriteInsIds.indexOf(
+    selectedContestId
```
jonahkagan, line 544: I think this variable is equivalent to `selectedContestIndex` and can be removed

eventualbuddha:
> Thanks for working through all the feedback. I think things ended up in a much clearer place!

Thank you! I agree, definitely better.

PR #1665, opened 2022-03-31
feat(bmd): Accessible Controller Diagnostic

jonahkagan:
## Overview

Task: #1417 

Adds a section for the accessible controller to the pollworker System Diagnostics screen, showing:
- Whether the controller is connected or not
- A button to start an interactive diagnostic test
- Results from the last test, if there are any

Note that currently test results only persist as long as the System Diagnostics screen is open. We may want to persist them longer in the future if we get feedback on that from users.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/530106/161169933-14e5d359-a23e-4fb2-b1c1-83f7c5558ef1.mov

localhost_3000_(Lenovo IdeaPad Flex 14 (FHD))

## Testing Plan 
Added automated tests

## Checklist
  - will add logging in a follow up PR

### DIFF
--- a/frontends/bmd/src/pages/accessible_controller_diagnostic_screen.test.tsx
+++ b/frontends/bmd/src/pages/accessible_controller_diagnostic_screen.test.tsx
@@ -1,0 +1,158 @@
+import React from 'react';
+import { act, render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import MockDate from 'mockdate';
+import { DateTime } from 'luxon';
+import { fakeTts } from '../../test/helpers/fake_tts';
+import { AriaScreenReader } from '../utils/ScreenReader';
+import {
+  AccessibleControllerDiagnosticScreen,
+  AccessibleControllerDiagnosticProps,
+} from './accessible_controller_diagnostic_screen';
+
+const now = DateTime.fromISO('2022-03-23T11:23:00.000Z');
+
+function renderScreen(
+  props: Partial<AccessibleControllerDiagnosticProps> = {}
+) {
+  return render(
+    <AccessibleControllerDiagnosticScreen
+      onComplete={jest.fn()}
+      onCancel={jest.fn()}
+      screenReader={new AriaScreenReader(fakeTts())}
+      {...props}
+    />
+  );
+}
+
+describe('Accessible Controller Diagnostic Screen', () => {
+  beforeEach(() => {
+    MockDate.set(now.toISO());
+  });
+
+  it('yields a success result when all steps are completed', async () => {
+    const mockTts = fakeTts();
+    const screenReader = new AriaScreenReader(mockTts);
+    const onComplete = jest.fn();
+    renderScreen({ onComplete, screenReader });
+
+    screen.getByText('Accessible Controller Test');
+
+    screen.getByText(/Step 1 of 6/);
+    screen.getByText('Press the up button.');
+    // Try out pressing an incorrect button to make sure we actually detect the
+    // right button.
+    act(() => void userEvent.keyboard('{ArrowDown}'));
+    screen.getByText(/Step 1 of 6/);
+    // Then press the up button.
+    // We have to wrap key presses in act to avoid a warning. This may be due to
+    // the fact that the keyDown listener is attached to the document instead of
+    // a React component.
+    act(() => void userEvent.keyboard('{ArrowUp}'));
+
+    await screen.findByText(/Step 2 of 6/);
+    screen.getByText('Press the down button.');
+    act(() => void userEvent.keyboard('{ArrowDown}'));
+
+    await screen.findByText(/Step 3 of 6/);
+    screen.getByText('Press the left button.');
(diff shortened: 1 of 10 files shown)


frontends/bmd/src/pages/accessible_controller_diagnostic_screen.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const buttons = [
+    ['Up', 'ArrowUp'],
+    ['Down', 'ArrowDown'],
+    ['Left', 'ArrowLeft'],
+    ['Right', 'ArrowRight'],
+    ['Select', 'Enter'],
+  ];
+  const steps = [
+    ...buttons.map(([buttonName, buttonKey]) => (
```
eventualbuddha, line 174: Why not just build the one we need for the current step?

jonahkagan: the sound diagnostic is also a step though, so it should only be rendered when `step` is 5

eventualbuddha: Oh, whoops. I misread the code. I might try to have a data structure that's just a little more info than the `buttons` one is that says what type it is etc, but I realize that's also introducing similar complexity in another direction. So this is fine.

frontends/bmd/src/pages/poll_worker_screen.tsx
```diff
@@ -339,211 +342,212 @@ export function PollWorkerScreen({
     );
   }
 
-  const pollworkerMainScreen = (
```
jonahkagan, line 342: this diff is mostly undoing a change I had made when I added the diagnostics screen. previously, I had only rendered the sidebar in PollWorkerScreen and then had the main panel render either the poll worker main screen or the diagnostics screen. That didn't work anymore with the full-screen accessible controller diagnostic, so now the diagnostics screen has to render the sidebar itself. this could probably be handled better with an app-wide router, but i think this works for what we've got

frontends/bmd/src/pages/accessible_controller_diagnostic_screen.tsx
```diff
@@ -0,0 +1,288 @@
...
+      margin-bottom: 1em;
+    }
+  }
+`;
+
+interface AccessibleControllerIllustrationProps {
+  highlight?: ButtonName | 'Headphones';
+}
+
+function AccessibleControllerIllustration({
```
carolinemodic, line 47: maybe make this its own file? Are we able to make an svg file or will that not work because of the parameters in the svg?

frontends/bmd/src/pages/accessible_controller_diagnostic_screen.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  it('yields a success result when all steps are completed', async () => {
+    const mockTts = fakeTts();
+    const screenReader = new AriaScreenReader(mockTts);
+    const onComplete = jest.fn();
+    renderScreen({ onComplete, screenReader });
+
+    screen.getByText('Accessible Controller Test');
+
+    function expectToHaveIllustrationHighlight(
```
carolinemodic, line 41: nice

frontends/bmd/src/pages/diagnostics_screen.test.tsx
```diff
@@ -20,20 +26,32 @@ function expectToHaveWarningIcon(element: HTMLElement) {
   expect(element).toHaveClass('dikopi');
 }
 
+function renderScreen(props: Partial<DiagnosticsScreenProps> = {}) {
```
carolinemodic, line 29: I like this pattern

carolinemodic (approved):
Can be a different PR but it would be great to have logging for this :). You could log on each step's success (Down button works, Left button works, etc.) but I think it would also be fine to just log an "init" message when the user starts the test and then a "success" or "failure" after the entire diagnostic with the result. (And I would do a similar pattern for the refreshing printer status as well, an "init" message when the button is clicked and success/failure with the result)

frontends/bmd/src/pages/accessible_controller_diagnostic_screen.tsx
jonahkagan: I first tried making an SVG file but couldn't figure out how to parameterize it without adding a complicated step to the build (there's a tool called SVGR that automatically turns an SVG file into a React component, but no basic TS way to import an SVG as tags, only as an image). I can move it to a separate file

jonahkagan: Actually now that I try moving it, I'm realizing it relies on the ButtonName type from this module, so I think it's actually just easier to leave it in the same file for now.

jonahkagan:
Yeah I'm gonna add logging for the entire hardware diagnostic screen once I finish building each section. I want to get a handle on what the whole feature looks like so I can see if there's a way to make the logging standardized for all of the pieces

Issue #1453, opened 2022-02-28
Remove "pi" backup pin in VxAdmin


carolinemodic:
Right now we have a fallback in VxAdmin where if the user enters pi "314159" there is an option to remove the election definition. This is a fallback in case the machines need to set up and no one knows the pin. We are going to solve this a better way moving forward with a backup admin card sent to customers so this "feature" can be removed.

PR #1666, opened 2022-03-31
Remove pi backup PIN from VxAdmin

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/1453

VxAdmin supports a special backup PIN. If an admin enters pi (314159), they'll be given an option to remove the election definition. This is a fallback for the case that machines need to be set up, and no one knows the PIN. Moving forward, we'll be solving this with a backup admin card sent to customers, so we can remove this special backup PIN.

### Before

https://user-images.githubusercontent.com/12616928/161155214-e87d2543-e052-4f67-a6c5-6ae6e0f8ef29.mov

### After

https://user-images.githubusercontent.com/12616928/161155210-bbe4fb92-38f5-41f0-9cb3-8821711002c3.mov

## Testing

- [x] Manually verified that the backup PIN no longer works (demoed above)
- [x] Manually verified that a valid PIN still works
- [x] Verified that the pre-existing unit test for the backup PIN now fails (before removing the test)

## Checklist

### DIFF
--- a/frontends/election-manager/src/screens/unlock_machine_screen.test.tsx
+++ b/frontends/election-manager/src/screens/unlock_machine_screen.test.tsx
@@ -64,29 +64,3 @@ test('authentication', async () => {
 
   expect(screen.queryByText('Invalid code. Please try again.')).toBeNull();
 });
-
-test('Unconfigure VxAdmin from secret menu', async () => {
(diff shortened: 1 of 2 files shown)


carolinemodic (approved):
giphy

Issue #1505, opened 2022-03-08
NH: support ballots with bubbles on the right of the name


benadida:

eventualbuddha:
Here's a [PDF](https://github.com/votingworks/vxsuite/files/8393517/election-d0696d9f5c-precinct-hudson-id-town-id-12101-precinct-id--style-card-number-54-English-live-absentee.pdf) that has the bubbles on the right and an annotated image showing where it found the contests & targets:

![election-d0696d9f5c-precinct-hudson-id-town-id-12101-precinct-id--style-card-number-54-English-live-absentee-p1-layout](https://user-images.githubusercontent.com/1938/161162806-377bf096-6c96-4692-a265-c09bf635741e.jpg)

PR #1667, opened 2022-03-31
feat(scan): support scanning right-side targets

eventualbuddha:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/1505

Changes `findTargets` to accept a `targetMarkPosition` which determines whether it starts looking offset from the left or the right side of the contest box.

Updates the `layout` CLI command to find the targets and annotate them in the output. It also now takes an option to specify whether to look on the right or left for the targets.

Adds a basic test for scanning a filled-in ballot with the right-side targets.

## Demo Video or Screenshot
| Template | Scan |
|-|-|
| image | 20220304_163426-ballot-0001 |

## Testing Plan 
Tested manually with VxCentralScan and the ballot shown above. Added the same ballot as an automated test to `ballot-interpreter-vx`.

## Checklist


libs/ballot-interpreter-vx/src/hmpb/find_targets.ts
```diff
@@ -26,7 +31,13 @@ export function* findTargets(
   const minAspectRatio = aspectRatio - aspectRatioTolerance;
   const maxAspectRatio = aspectRatio + aspectRatioTolerance;
 
-  const x = bounds.x + Math.round(inset + expectedWidth / 2);
+  const x =
+    bounds.x +
+    Math.round(
+      targetMarkPosition === BallotTargetMarkPosition.Left
```
benadida, line 37: is there any chance that this causes an explicit `null` to default `BallotTargetMarkPosition.Right`, and thus should this conditional be checking that it equals `Right` and otherwise defaulting to `Left`?

I see the default parameter value, so maybe not, but I liked how your other conditionals were explicitly checking for `Right` and otherwise doing the `Left` thing.

libs/ballot-interpreter-vx/src/layout.ts
```diff
@@ -110,7 +110,14 @@ export async function interpretTemplate({
       ({ bounds, corners }) => ({
         bounds,
         corners,
-        targets: [...reversed(findTargets(normalized.imageData, bounds))],
+        targets: [
+          ...reversed(
+            findTargets(normalized.imageData, bounds, {
+              targetMarkPosition:
+                electionDefinition.election.ballotLayout?.targetMarkPosition,
```
benadida, line 117: I know there's a default in `findTargets`, but now I'm wondering if the default should be set here, at a higher level, too.

benadida (approved):
couple of small comments to consider, but not blocking, lgtm.

libs/ballot-interpreter-vx/src/hmpb/find_targets.ts
eventualbuddha: I agree that we should switch the conditional to check for `Right` as an extra measure of safety and for consistency. That said, passing `null` here would cause type checking to fail because an optional field allows `undefined` but not `null`. Here's a simplified example ([playground](https://www.typescriptlang.org/play?#code/KYOwrgtgBACg9gZwJYBclxFA3gKCvqAGWADMUoBeKAcgBtSVqAaPAgJSQHMALcq6gE5dezHAF8cOEmBABjNBigATOAEEAKtyQhOACixQADolTpMVeMgUgAdMTJQxlbGICU2VvlkYEcejdo4PQNjKzNHVwBucUkVDS0dXSicOM1tYKMTawAuKHBaWgjo1ISM0NMMXMsK2w4ecjdIoA)):

```ts

enum Position {
    Left = 'left',
    Right = 'right',
}

function doAThing({ position = Position.Left } = {}) {
    console.log({ position });
}

doAThing();
doAThing({ position: null });
doAThing({ position: Position.Right });
```

You get this error on the `null` line:

```
Type 'null' is not assignable to type 'Position | undefined'.
```

libs/ballot-interpreter-vx/src/layout.ts
eventualbuddha: One approach would be to require the value to be provided to `findTargets`, and the `?? BallotTargetMarkPosition.Left` would be here. What do you think?

Issue #442, opened 2021-04-10
Tabulation reports should include number of seats for each contest


benadida:

eventualbuddha:
Where do we put it on the report? @beausmith

beausmith:
@arsalansufi - Here is my guidance…

- [ ] Logic: when a Candidate Contest has more than 1 seat, append the number of seats to the contest title using: `<Text as="span" small noWrap>({contest.seats} seats)</Text>`
- [ ] Ensure there is a space between the end of the contest title and the span.

The finished result should look like this:

<img width="1134" alt="image" src="https://user-images.githubusercontent.com/28444/160925881-c455a8e1-a044-42db-85a0-6cbed1725d29.png">

My goals here:
- use typographic solution for simplicity.
- attempt to not cause the content to wrap (hence not placing # seats in the "ballots/overvotes/undervotes" line.
- associate the seats value with the title of the contest.

arsalansufi:
Thanks Beau!

PR #1678, opened 2022-04-08
Add number of seats for contests with >1 seat to tabulation reports

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/442

This PR adds the number of seats for contests with >1 seat to tabulation reports for clarity.

## Demo screenshots

_Contest with two seats_

two-seats

_Contest with two seats, multiline header_

two-seats-multiline-1

_Contest with one seat (unchanged, no added text)_

one-seat

## Testing plan 

- [x] Tested manually (as shown above)
- [x] Added snapshot and unit tests

## Checklist

### DIFF
--- a/libs/ui/src/contest_tally.test.tsx
+++ b/libs/ui/src/contest_tally.test.tsx
@@ -1,0 +1,103 @@
+import React from 'react';
+import {
+  CastVoteRecord,
+  Election,
+  ExternalTally,
+  Tally,
+} from '@votingworks/types';
+import { computeTallyWithPrecomputedCategories } from '@votingworks/utils';
+import {
+  electionMultiPartyPrimaryWithDataFiles,
+  electionSample2WithDataFiles,
+} from '@votingworks/fixtures';
+import { render } from '@testing-library/react';
+
+import { ContestTally } from './contest_tally';
+
+function parseCvrsFileContents(cvrsFileContents: string): CastVoteRecord[] {
+  const lines = cvrsFileContents.split('\n');
+  return lines
+    .filter((line) => line.length > 0)
+    .map((line) => JSON.parse(line) as CastVoteRecord);
+}
+
+function constructTally(cvrsFileContents: string, election: Election): Tally {
+  const castVoteRecords = parseCvrsFileContents(cvrsFileContents);
+  const { overallTally } = computeTallyWithPrecomputedCategories(
+    election,
+    new Set(castVoteRecords),
+    []
+  );
+  return overallTally;
+}
+
+describe('ContestTally', () => {
+  let election: Election;
+  let electionTally: Tally;
+  let externalTallies: ExternalTally[];
+
+  beforeEach(() => {
+    election = electionSample2WithDataFiles.electionDefinition.election;
+    electionTally = constructTally(
+      electionSample2WithDataFiles.cvrDataSmall1,
+      election
+    );
+    externalTallies = [];
+  });
+
+  it('Renders', () => {
+    const { container } = render(
+      <ContestTally
+        election={election}
+        electionTally={electionTally}
+        externalTallies={externalTallies}
+      />
+    );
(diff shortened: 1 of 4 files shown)


arsalansufi:
Design q for you @beausmith 

I followed the designs that you specified in https://github.com/votingworks/vxsuite/issues/442 for the multiline header case. The spacing looks fine when some of the larger title text spills onto to the second line (left), but the spacing feels a bit funny when only the smaller seat text spills onto the second line (right), e.g. the space between the seat text and "N ballots cast" feels small. Curious if you want to make any adjustments.

two-seats-multiline-1 two-seats-multiline-2

carolinemodic (commented):
Code LGTM but I'll let @beausmith sign off on the design

libs/ui/src/contest_tally.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { ContestTally } from './contest_tally';
+
+function parseCvrsFileContents(cvrsFileContents: string): CastVoteRecord[] {
+  const lines = cvrsFileContents.split('\n');
+  return lines
+    .filter((line) => line.length > 0)
+    .map((line) => JSON.parse(line) as CastVoteRecord);
+}
+
+function constructTally(cvrsFileContents: string, election: Election): Tally {
```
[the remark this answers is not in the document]
arsalansufi: Yeah I was thinking about this! At the bare minimum, let me move this `constructTally` function to `@votingworks/test-utils`

beausmith (approved):
Thanks for this note. I measured and they distance is the same between the lines on both the screenshots you shared. I do agree that it "feels" a little different though. Let's push this as is and revisit if we find that we get more feedback about it.

arsalansufi:
Sounds good!

libs/ui/src/contest_tally.test.tsx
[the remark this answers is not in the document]
arsalansufi: Changed my mind slightly and moved `parseCvrsFileContents` to `@votingworks/test-utils` while leaving `constructTally` in this file, since `constructTally` is now just a combination of one helper from `@votingworks/test-utils` and another from `@votingwork/utils` 

https://github.com/votingworks/vxsuite/pull/1678/commits/817e02c737fc7282fc9fc17b8085947f18ddf164

libs/ui/src/contest_tally.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const castVoteRecords = parseCvrsFileContents(cvrsFileContents);
+  const { overallTally } = computeTallyWithPrecomputedCategories(
+    election,
+    new Set(castVoteRecords),
+    []
+  );
+  return overallTally;
+}
+
+describe('ContestTally', () => {
```
eventualbuddha, line 23: For the most part our tests have been using the top-level `test` without any `describe` nesting. We still have some older ones that use this pattern, but we generally haven't been creating new ones.

arsalansufi: Got it! I've historically liked having the ability to pair `describe`s with `beforeEach`s  for sub-conditions and corresponding setup. So I'm curious how we would feel about this "primarily flat but nested when helpful" structure:

```
beforeEach(() => {
  // Common setup
});

test(...);

test(...);

describe('When pigs fly', () => {
  beforeEach(() => {
    // Additional setup for a subset of tests
    allowPigsToFly();
  });

  test(...);

  test(...);
});
```

eventualbuddha: Personally I prefer to either copy the shared setup into each `test` or to split the file into multiple test files.

jonahkagan: I didn't know about this convention and been using `describe` as well. Seems like a good topic for an eng meeting

PR #1686, opened 2022-04-13
Add Code Owners file for review assignment 

carolinemodic:
Adds an initial codeowners file to vxsuite to automate PR assignment. This can grow and evolve overtime, anyone is welcome to add additional rules to route things to themselves as they see fit. Obviously I added more specific rules for myself as I knew best what I wanted to be included on reviews for but wasn't sure about what other people would want so those are left more vague.


arsalansufi (approved):
LGTM from my end!

jonahkagan (approved):

Issue #1682, opened 2022-04-12
Unified smartcard auth interface


jonahkagan:
Build a new interface to abstract all smartcard and authentication logic from the React apps.
 
**Implementation plan**
- [x] Build a prototype of this interface
- [x] Migrate bmd to the new interface
- [ ] ~~Check that we like it by implementing the card diagnostic~~ this feature was cut
- [x] Migrate the rest of the apps to the new interface

carolinemodic:
@jonahkagan is this ready to mark as closed?

jonahkagan:
Yes!

PR #1702, opened 2022-04-15
refactor(bmd): Authenticate with useSmartcardAuth (superadmin, admin, pollworker)

jonahkagan:
## Overview

Task: #1682

Adds a basic first version of the useSmartcardAuth hook that only supports the features needed to authenticate superadmin, admin, and pollworker users on the BMD. Then refactors BMD to use the hook.

I'll come back and do the voter auth in a separate PR.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Updated existing tests
- Added tests for useSmartcardAuth

## Checklist


frontends/bmd/src/pages/poll_worker_screen.tsx
```diff
@@ -73,18 +77,70 @@ export interface PollworkerScreenProps {
   screenReader: ScreenReader;
   printer: Printer;
   togglePollsOpen: () => void;
-  tallyOnCard?: PrecinctScannerCardTally;
-  clearTalliesOnCard: () => Promise<void>;
```
jonahkagan, line 77: The only major refactor I made to BMD was to move the reading/clearing of the precinct tally from the pollworker card into the PollworkerScreen component instead of doing it in AppRoot. I thought this would be a bit clearer since the card is guaranteed to be ready for reading when the pollworker is authenticated.

libs/ui/src/hooks/use_smartcard_auth.ts
```diff
@@ -0,0 +1,272 @@
...
+}: UseSmartcardAuthArgs): SmartcardAuth {
+  const [{ card, auth }, dispatch] = useReducer(
+    smartcardAuthReducer(allowedUserRoles, electionDefinition),
+    {
+      card: { status: 'no_card' },
+      auth: { status: 'logged_out', reason: 'no_card' },
+    }
+  );
+  // Every card read/write must obtain a lock to prevent concurrent operations.
+  const cardLock = useLock();
```
jonahkagan, line 228: I'm interested in getting feedback on this idea. There was similar logic in useSmartcard, but I noticed that it only blocked concurrent writes and allowed reads during writes. Maybe that's fine? I'm not totally sure. Only allowing one card operation at a time seemed like a surefire way to ensure no concurrency issues.

jonahkagan: Thinking about it more, I'm wondering if concurrency protection is even necessary at all. I think consumers won't make concurrent long value reads/writes. The only concurrent requests I could imagine happening would be the short value reads from the useInterval. But the smartcard service will only process one request at a time, so I think those requests will get linearized by the server anyway.

benadida: commenting just on this thing for now: having multiple layers of protection against concurrent requests is not a bad thing. I can imagine one day we change the backend service in some subtle way, and end up with concurrent requests processing... so I like the extra protection if it's not too onerous.

jonahkagan: Thanks, that's helpful. If it's just defensive and not solving an actual current problem, then I think it can be simplified. I changed it to better match the existing behavior of the locking in useSmartcard to just error on concurrent writes: 432c8725.

benadida (commented):
just doing a high-level architecture review, looks good to me. Leaving specific code review to others.

frontends/bmd/src/app_root.tsx
```diff
@@ -539,6 +474,12 @@ export function AppRoot({
   const hasPrinterAttached = printerInfo !== undefined;
   const previousHasPrinterAttached = usePrevious(hasPrinterAttached);
 
+  const auth = useSmartcardAuth({
+    cardApi: card,
+    electionDefinition: optionalElectionDefinition,
+    allowedUserRoles: ['superadmin', 'admin', 'pollworker', 'voter'],
```
carolinemodic, line 480: slight preference to have this array defined as a global config or something

libs/ui/src/hooks/use_smartcard_auth.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    /* istanbul ignore next - compile time check for completeness */
+    default:
+      throwIllegalValue(auth, 'status');
+  }
+}
+
+// Some useful type guards for checking who's logged in
+
+export function isSuperadminAuth(
```
carolinemodic, line 272: I like these helpers

carolinemodic (approved):

Issue #1710, opened 2022-04-19
Augment VxAdmin test decks with VxMark test decks and update expected tally reports accordingly


arsalansufi:
Augment VxAdmin test decks with VxMark test decks and update expected tally reports accordingly

Issue #1617, opened 2022-03-21
libs/ui has suprising behavior with snapshot tests


carolinemodic:
Needs a good repro but I've noticed several times where I will add a new component to libs/ui or update a component, and in the frontend apps, snapshot tests that rely on completely different, unchanged components from libs/ui will regenerate the class ids causing a frustrating and annoying number of snapshot tests to change when making a very small PR.

adghayes:
Repro'd this last week in #2793. 

Based on research I've done, we don't have a way to modify the murmur hash generated class names from `styled-components`. We can add a prefix, filename, or component name with `babel-plugin-styled-components`, but not affect the hash portion of the classname.

If we want to avoid this issue, I think that we would have to make a little snapshot utility that wrapped the regular snapshot but stripped out the `styled-components` classname. We could make it easier to target those classnames by using the Babel plugin mentioned above to prefix them all (from `libs/ui`).

I think the risks for this would be low (we're not looking at those classnames anyway!) but wondering if the complexity would be worth it. I lean toward yes, because the snapshot churn in frontends other than the one you're working in can be very confusing.

adghayes:
I haven't heard about this being an issue in a while and, per my comment two years ago, I didn't see a good fix. So I'm going to close it unless @carolinemodic or @kofi-q still deem this a problem.

PR #1736, opened 2022-04-22
Move PrintedBallot to shared location for use in VxAdmin test decks

arsalansufi:
## Overview

A first step for issue https://github.com/votingworks/vxsuite/issues/1710

This PR moves VxMark's `PrintedBallot` component (and relevant downstream code) to a shared location for use in VxAdmin so that we can augment VxAdmin's test deck with the contents of VxMark's test deck, part of the larger L&A revamp.

Note that, while we've currently moved `PrintedBallot` to `@votingworks/ui`, longer term, this probably belongs in a package that's higher level than `@votingworks/ui`, e.g. `@votingworks/ballots`. The higher-level grouping just isn't clear yet.

## Demo video or screenshot

N/A

## Testing plan

_Automated testing_

- [x] Updated snapshot tests, making sure that all updates were expected and not regressions
- [x] Added additional tests to keep `frontends/bmd`, `@votingworks/ui`, and `@votingworks/utils` at 100% test coverage

_Manual testing_

Generated printed ballots with the following contents, before and after this PR, verifying that nothing had changed:

- [x] Printed ballot with candidate contest
- [x] Printed ballot with yes-no contest
- [x] Printed ballot with MS either-neither contest
- [x] Printed ballot with seal
- [x] Printed ballot without seal

## Checklist

- [ ] ~I have added a screenshot and/or video to this PR to demo the change~ N/A
- [ ] ~I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates~ N/A


frontends/bmd/src/pages/print_page.test.tsx
```diff
@@ -1,20 +1,25 @@
...
+  const original = jest.requireActual<Record<string, unknown>>(
+    '@votingworks/utils'
+  );
+  // Mock random string generation so that snapshots match, while leaving the rest of the module
+  // intact
+  return {
+    ...original,
+    randomBase64: () => 'CHhgYxfN5GeqnK8KaVOt1w',
+  };
 });
```
arsalansufi, line 23: I tried a number of alternatives to the above to mock `randomBase64` while leaving the rest of `@votingworks/utils` intact. For example:

```ts
return {
  ...original,
  randomBase64: mockOf(randomBase64).mockReturnValue(
    'CHhgYxfN5GeqnK8KaVOt1w'
  ),
};
```

But this resulted in the following error:

```
The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.
Invalid variable access: randomBase64
Note: This is a precaution to guard against uninitialized mock variables. If it is ensured that the mock is required lazily, variable names prefixed with `mock` (case insensitive) are permitted.
```

I tried renaming variables as per the note in the error message. I also tried using `jest.doMock` instead of `jest.mock` (the latter hoists mock code above the relevant import while the former doesn't). But encountered various other issues.

I landed on what I did as the simplest working solution.

libs/fixtures/src/index.test.ts
```diff
@@ -34,6 +34,8 @@ test('has various election definitions', () => {
       "electionSampleDefinition",
       "electionSampleLongContent",
       "electionSampleLongContentDefinition",
+      "electionSampleNoSeal",
+      "electionSampleNoSealDefinition",
```
arsalansufi, line 38: Copied from `frontends/bmd` ([frontends/bmd/data/electionSampleNoSeal.json](https://github.com/votingworks/vxsuite/blob/main/frontends/bmd/data/electionSampleNoSeal.json) and [frontends/bmd/src/data/electionSampleNoSeal.json.ts](https://github.com/votingworks/vxsuite/blob/main/frontends/bmd/src/data/electionSampleNoSeal.json.ts)) for use in `@votingworks/ui` tests (next commit)

arsalansufi:
@beausmith You were automatically added as a reviewer given changes to `@votingworks/ui`, but this is a purely technical change with no visual implications so feel free to skip

libs/ui/src/printed_ballot.tsx
```diff
@@ -237,6 +240,9 @@ interface Props {
   votes: VotesDict;
 }
 
+/**
+ * Renders a printed ballot, generated by a ballot-marking device
+ */
 export function PrintedBallot({
```
arsalansufi, line 247: This does make me wonder whether I should rename this component now that it lives outside of `frontends/bmd`, to distinguish it from a printed HMPB. Maybe `BMDPrintedBallot`?

arsalansufi: Went forth with that change! https://github.com/votingworks/vxsuite/pull/1736/commits/ce66a5c93cb4a70e644dd898daac083113bdefb1

frontends/bmd/src/__snapshots__/app_contest_candidate_no_party.test.tsx.snap
```diff
@@ -479,7 +479,7 @@ exports[`Single Seat Contest 1`] = `
     class="c1"
   >
     <main
-      class="sc-bkkfTU enzCGe"
+      class="sc-cxpRKc iyrZoO"
```
arsalansufi, line 482: This simple rename seems to have caused a lot of snapshot updates, as documented in issue https://github.com/votingworks/vxsuite/issues/1617

arsalansufi, line 482: Looks like this component rename affected the class names in a large number of snapshots, as documented in issue https://github.com/votingworks/vxsuite/issues/1617

libs/types/src/election.ts
```diff
@@ -724,6 +724,10 @@ export type YesNoVote =
   | readonly ['no', 'yes']
   | readonly [];
 export type YesOrNo = Exclude<YesNoVote[0] | YesNoVote[1], undefined>;
+export const DisplayTextForYesOrNo: Record<YesOrNo, string> = {
+  yes: 'Yes',
+  no: 'No',
+};
```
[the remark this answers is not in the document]
jonahkagan: The display text for yes/no votes seems more like a UI concern than a typing concern. Is there a place this would make sense to live in the ui library?

frontends/bmd/src/components/printed_ballot.tsx
```diff
@@ -135,7 +132,12 @@ function CandidateContestResult({
   election,
   precinctId,
   vote = [],
-}: CandidateContestResultInterface): JSX.Element {
+}: {
+  contest: CandidateContest;
+  election: Election;
+  precinctId: PrecinctId;
+  vote: CandidateVote;
+}): JSX.Element {
```
jonahkagan, line 143: Would it make sense to inline these interfaces in this file but name them `CandidateContestResultProps`, etc? That's the usual convention I've seen for component prop typing (as opposed to inlining the object type)

libs/ui/src/printed_ballot.tsx
jonahkagan: Yeah I was thinking the same thing right before encountering this comment!

libs/ui/src/printed_ballot.tsx
```diff
@@ -160,6 +160,7 @@ function CandidateContestResult({
           </Text>{' '}
           {candidate.partyId &&
             `/ ${
+              /* istanbul ignore next: This .find should always return something */
               election.parties.find((p) => p.id === candidate.partyId)?.name
```
jonahkagan: I think libs/utils/find provides a version of `find` that always returns something or throws an error

libs/ui/src/printed_ballot.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+    },
+  },
+  {
+    description: 'PrintedBallot renders when no seal',
+    electionDefinition: electionSampleNoSealDefinition,
+    ballotStyleId: '5',
+    precinctId: '21',
+    votes: {},
+  },
+];
```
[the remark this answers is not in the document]
jonahkagan: When I read these snapshot tests, I don't really feel like I can understand well what they are testing. The snapshots are too large and too obscure. Is it possible to write explicit test assertions to help show what differences we're looking for in each case? (That may be out of scope for this change, but if you did get a good sense of it while working here, this might be a good time to document it via tests)

jonahkagan (approved):
Thanks for the nice commit-by-commit PR! Left a few small thoughts. I don't think any are blocking so approving

libs/types/src/election.ts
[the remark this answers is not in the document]
arsalansufi: Good call! It could probably go in [`libs/ui/src/globals.ts`](https://github.com/votingworks/vxsuite/blob/main/libs/ui/src/globals.ts) instead

libs/ui/src/printed_ballot.test.tsx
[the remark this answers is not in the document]
arsalansufi: Mm really good point. Lemme make some tweaks!

beausmith (approved):
LGTM. The only suggestion I have is to rename "bmd_printed_ballot.tsx" to "bmd_paper_ballot.tsx". 

This will make the filenames for printed ballots more consistent, as the HMPB filename is "hand_marked_paper_ballot.tsx".

libs/ui/src/printed_ballot.test.tsx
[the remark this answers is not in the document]
arsalansufi: Taken care of in https://github.com/votingworks/vxsuite/pull/1736/commits/3323d27858ed67879280884a6a1b0cabc9d2ff45

PR #1767, opened 2022-05-03
lint(async): enable `@typescript-eslint/require-await`

### LINKED ISSUE #1759: lint: enable `@typescript-eslint/require-await`
We use `async` in places where it's not needed. We should consider enabling `@typescript-eslint/require-await` to disallow such uses of `async`. Looking into this a bit, we seem to have a few places where we use `async` without an `await` where we cannot simply remove the `async`:
1. **Implementing an interface:** If the method requires returning `Promise<…>`, but we don't actually need to `await` anything (such as in `MemoryStorage`), we've used `async` functions as a convenient way to wrap the return value in a `Promise`. The alternative is to explicitly use `return Promise.resolve(…)`. I'm not sure that's actually better.
1. **Forcing a `process.nextTick`:** Sometimes we need to wait in a test, and making something `async` that doesn't really need to be can force it to wait for the next tick, which _is_ needed when async operations happen as a result of e.g. `fireEvent.click`.

eventualbuddha:
## Overview

Fixes a bunch of violations, including those that cascaded from removing the `async` on functions making the calls of those functions no longer requiring `await` etc.

I marked a bunch of `async` without `await` as ignored because they were implementing an interface either as a mock or fake version of something. Others were ignored because of async issues in tests that it was easier to ignore for now.

Closes #1759 

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated.

## Checklist

### DIFF
--- a/frontends/election-manager/src/components/save_file_to_usb.tsx
+++ b/frontends/election-manager/src/components/save_file_to_usb.tsx
@@ -32,7 +32,7 @@ export enum FileType {
 
 export interface Props {
   onClose: () => void;
-  generateFileContent: () => Promise<Uint8Array | string>;
+  generateFileContent: () => Uint8Array | string | Promise<Uint8Array | string>;
   defaultFilename: string;
   fileType: FileType;
   promptToEjectUsb?: boolean;
--- a/frontends/election-manager/src/components/save_file_to_usb.tsx
+++ b/frontends/election-manager/src/components/save_file_to_usb.tsx
@@ -7,6 +7,7 @@ import { assert, usbstick, throwIllegalValue } from '@votingworks/utils';
 import { Modal, UsbControllerButton } from '@votingworks/ui';
 
 import { LogEventId } from '@votingworks/logging';
+import { PromiseOr } from '@votingworks/types';
 import { AppContext } from '../contexts/app_context';
 import { Button } from './button';
 import { Prose } from './prose';
@@ -32,7 +33,7 @@ export enum FileType {
 
 export interface Props {
   onClose: () => void;
-  generateFileContent: () => Uint8Array | string | Promise<Uint8Array | string>;
+  generateFileContent: () => PromiseOr<Uint8Array | string>;
   defaultFilename: string;
   fileType: FileType;
   promptToEjectUsb?: boolean;
--- a/libs/ballot-interpreter-vx/src/cli/types.ts
+++ b/libs/ballot-interpreter-vx/src/cli/types.ts
@@ -8,7 +8,7 @@ export interface Command {
     stdin: NodeJS.ReadableStream,
     stdout: NodeJS.WritableStream,
     stderr: NodeJS.WritableStream
-  ): Promise<number>;
+  ): number | Promise<number>;
 }
 
 export interface GlobalOptions {
--- a/libs/ballot-interpreter-vx/src/cli/types.ts
+++ b/libs/ballot-interpreter-vx/src/cli/types.ts
@@ -1,3 +1,5 @@
+import { PromiseOr } from '@votingworks/types';
+
 export interface Command {
   name: string;
   description: string;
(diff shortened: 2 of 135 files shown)


frontends/election-manager/src/components/save_file_to_usb.tsx
```diff
@@ -32,7 +32,7 @@ export enum FileType {
 
 export interface Props {
   onClose: () => void;
-  generateFileContent: () => Promise<Uint8Array | string>;
+  generateFileContent: () => Uint8Array | string | Promise<Uint8Array | string>;
```
jonahkagan, line 35: This confused me. How do you call a function that may or may not return a promise? Seems like it might be easier to just have it always return a promise

libs/ballot-interpreter-vx/src/cli/types.ts
```diff
@@ -8,7 +8,7 @@ export interface Command {
     stdin: NodeJS.ReadableStream,
     stdout: NodeJS.WritableStream,
     stderr: NodeJS.WritableStream
-  ): Promise<number>;
+  ): number | Promise<number>;
```
jonahkagan, line 11: same question as above here

jonahkagan (approved):
Wow there were a lot of those! Thanks for enabling this rule, seems like it will definitely be useful. Just had one question

frontends/election-manager/src/components/save_file_to_usb.tsx
eventualbuddha: It's basically saying "I take a function that can be either synchronous or asynchronous". Perhaps this would be clearer with a type like this?

```ts
type MaybePromise<T> = T | Promise<T>;

generateFileContent: () => MaybePromise<Uint8Array | string>;
```

jonahkagan: Yeah I like that idea.

Also found this in MDN docs which I didn't know before:
> If the value of the expression following the await operator is not a Promise, it's converted to a [resolved Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve).

So effectively the return value of this function will end up being wrapped in a Promise, since the caller will have to `await` it.

I don't know if it's simpler to do the `MaybePromise` approach or just have a `Promise` return type and then either disable the lint rule for non-Promise functions or explicitly return a Promise. What do you think would be the clearest to readers (since functionally they're all the same)?

eventualbuddha: I'm not sure which is better. I'm inclined toward the `MaybePromise` approach for situations where you really don't know whether the result will be async or not. For example, it'd be annoying if the callback to Jest's `test` had to always return a promise just because it can sometimes return a promise. For the mocking/faking cases, I think it makes sense to make the interface use async but then disable the lint rule for those specific cases, since mocks are not a typical use of the API.

I'll try out the `MaybePromise` thing and see how it feels.

jonahkagan: Sounds good!

eventualbuddha: I ended up calling it `PromiseOr<T>`. What do you think?

jonahkagan: I usually read `Container<T>` as "Container of T" in my mind, so it reads a bit weird to me: "Promise Or of T". I liked `MaybePromise` (or `OptionalPromise` or something in that vein)

eventualbuddha: I'd support either of those alternatives. One advantage `PromiseOr` has is that it'll auto-complete when typing `Promise`, which can aid discovery. But that's not a huge deal. I'm not sure whether `OptionalPromise` would be confused with `Optional`, i.e. would you expect `type OptionalPromise<T> = Promise<T> | undefined`? `MaybePromise` has a similar problem. Or maybe I'm overthinking it 😜

jonahkagan: We're definitely deep in the weeds haha. I agree `OptionalPromise` is more likely to be confused with `Optional`

Issue #1772, opened 2022-05-03
Update `Modal` component to add a fullscreen variant


taravancil:
As discussed with @beausmith , the transcription screen for write-in adjudication will be implemented as a fullscreen modal. This ticket should add a fullscreen variant to the existing `Modal` component.

PR #1774, opened 2022-05-03
Add fullscreen modal variant

taravancil:
## Overview
#1772 

## Demo Video or Screenshot
Non-fullscreen modal
Screen Shot 2022-05-03 at 3 28 40 PM

Fullscreen modal
Screen Shot 2022-05-03 at 3 39 48 PM

## Testing Plan 
Manually QAed the new fullscreen variant and other modals. I added a test to `modal.test.tsx` to satisfy our coverage requirements, but we don't get much real utility out of it. A screenshot test would be awesome here, but afaict we don't have those yet and that is out-of-scope for this issue.

## Checklist

- [x] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates]

### DIFF
--- a/libs/ui/src/modal.tsx
+++ b/libs/ui/src/modal.tsx
@@ -14,6 +14,7 @@ export enum ModalWidth {
 }
 
 interface ReactModalContentInterface {
+  fullscreen?: boolean;
   modalWidth?: ModalWidth;
 }
 const ReactModalContent = styled('div')<ReactModalContentInterface>`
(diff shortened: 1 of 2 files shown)


carolinemodic (approved):
defer to Beau on the design but code lgtm!

libs/ui/src/modal.tsx
```diff
@@ -32,15 +33,20 @@ const ReactModalContent = styled('div')<ReactModalContentInterface>`
   -webkit-overflow-scrolling: touch;
   @media (min-width: 480px) {
     position: static;
-    border-radius: 0.25rem;
-    max-width: ${({ modalWidth = ModalWidth.Standard }) => modalWidth};
+    border-radius: ${({ fullscreen }) => (fullscreen ? '0' : '0.25rem')};
```
beausmith, line 36: Unless the value of `0` is necessary, it's preferred to return `undefined`.

libs/ui/src/modal.tsx
```diff
@@ -50,7 +56,7 @@ const ReactModalOverlay = styled.div`
   z-index: 999; /* Should be above all default UI */
   background: rgba(0, 0, 0, 0.75);
   @media (min-width: 480px) {
-    padding: 0.5rem;
+    padding: ${({ fullscreen }) => (fullscreen ? '0' : '0.5rem')};
```
beausmith, line 59: Same as above comment re: `undefined`

beausmith (commented):
A couple post-merge comments re CSS strategy. If we return `0` it should be overriding another style. If not, return `undefined` to make clear that passing the style is unnecessary.

PR #1803, opened 2022-05-05
feat(election-manager): Use Prose from lib/ui. Add HMPBProse.

beausmith:
Related https://github.com/votingworks/vxsuite/issues/1778


frontends/election-manager/src/components/prose.tsx
```diff
@@ -1,89 +0,0 @@
...
-  textCenter?: boolean;
-}
-
-export const Prose = styled('div')<Props>`
-  margin: ${({ textCenter }) => (textCenter ? 'auto' : undefined)};
-
-  /* width: 100%; */
-  max-width: ${({ maxWidth = true }) => (maxWidth ? '66ch' : undefined)};
-  text-align: ${({ textCenter }) => (textCenter ? 'center' : undefined)};
-  line-height: ${({ density }) => (density !== 0 ? '1.1' : '1.3')};
```
beausmith, line 16: This PR does have one "breaking" change: the default line-height in VxAdmin UI is changing from `1.3` to `1.2`.

This was never intended to be 1.3 in the UI of the app, only in the printed HMPB. The value of `1.3` is preserved for the HMPB.

libs/ui/src/hand_marked_paper_ballot_prose.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+    </p>
+    <hr />
+  </React.Fragment>
+);
+describe('renders HandMarkedPaperBallotProse', () => {
+  test('with defaults', () => {
+    const { container } = render(
+      <HandMarkedPaperBallotProse>{proseContent}</HandMarkedPaperBallotProse>
+    );
+    expect(container.firstChild).toMatchSnapshot();
```
jonahkagan, line 35: I looked through these snapshots and I couldn't tell the difference - they seem to be identical. Is it possible to add more specific test assertions for the differences between these cases instead of using snapshots?

beausmith: Replaced all the snapshots for the `Main` component with verification of logic: https://github.com/votingworks/vxsuite/pull/1819/files#diff-f57a44aa306887ba87f92fa4265dfcaec9d6858fa45ae9ab8dcf4a6a8c2fe1b5

Will apply here next week.

beausmith: Just pushed fresh test code.

jonahkagan (approved):

PR #1804, opened 2022-05-06
remove ellipses from buttons

### LINKED ISSUE #1748: Remove ellipsis in various button components
We have inconsistent uses of ellipses through the product, notably in the VxCentralScan admin screen, and should remove them for consistency and simplicity. 

https://votingworks.slack.com/archives/CJU9MSC6S/p1651082417519689

adghayes:
## Overview
Closes #1748

Removes ellipses in the text of buttons across apps. Ellipses were left in:
- Temporary status button text (e.g. "Exporting...")
- "Save As..." buttons
- "Select a precinct for this device..." dropdown in VxMark
- "Select File..." buttons

## Demo Video or Screenshot
VxCentralScan Admin Screen:
bsd_admin_after

VxAdmin Tally Screen
tally_after

## Checklist

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -10,6 +10,7 @@ import {
   getByText as domGetByText,
   getAllByRole as domGetAllByRole,
   act,
+  within,
 } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import fetchMock from 'fetch-mock';
@@ -455,9 +456,10 @@ test('tabulating CVRs', async () => {
   );
 
   getByText('View Unofficial Full Election Tally Report');
-  fireEvent.click(getByText('Mark Tally Results as Official…'));
-  getByText('Mark Unofficial Tally Results as Official Tally Results?');
   fireEvent.click(getByText('Mark Tally Results as Official'));
+  getByText('Mark Unofficial Tally Results as Official Tally Results?');
+  const modal = await screen.findByRole('alertdialog');
+  fireEvent.click(within(modal).getByText('Mark Tally Results as Official'));
   expect(mockKiosk.log).toHaveBeenCalledWith(
     expect.stringContaining(LogEventId.MarkedTallyResultsOfficial)
   );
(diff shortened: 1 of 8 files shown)


adghayes:
Synced with @beausmith and we agreed to leave in the ellipses in the places indicated in the overview.

frontends/election-manager/src/app.test.tsx
```diff
@@ -455,9 +456,10 @@ test('tabulating CVRs', async () => {
   );
 
   getByText('View Unofficial Full Election Tally Report');
-  fireEvent.click(getByText('Mark Tally Results as Official…'));
-  getByText('Mark Unofficial Tally Results as Official Tally Results?');
   fireEvent.click(getByText('Mark Tally Results as Official'));
+  getByText('Mark Unofficial Tally Results as Official Tally Results?');
+  const modal = await screen.findByRole('alertdialog');
```
arsalansufi, line 461: Ooh interesting, so as far as I know, `findByRole` doesn't actually return the found element. It just throws an error when it can't find something. To actually get and use the found element, you typically have to use one of the `getBy` queries: https://testing-library.com/docs/queries/about#types-of-queries

Would be curious to understand why the test is still passing!

arsalansufi: Actually, I take that back! Looks like it does return the element -- that's what the returned promise resolves to in the success case. (It wasn't completely clear in the docs what the returned promise resolved to)

arsalansufi (approved):
LGTM!

frontends/election-manager/src/app.test.tsx
adghayes: My reading of the documentation was that `findBy...` returns a promise of the element. I thought this was the case based on the chart in the docs:
Screen Shot 2022-05-09 at 11 59 07 AM


Also, the Arlo example also uses `findBy...`: https://github.com/votingworks/arlo/blob/e2a58b5a557ef09c0fffd77385a9f3ed72e87f62/client/src/components/AuditBoard/AuditBoardView.test.tsx#L269-L281

Is there a difference between how it's being used in the Arlo case and this one?

adghayes: Ok, great! I was't sure about the comparative advantages of one or the other, so I emulated the way Arlo did it.

Issue #1786, opened 2022-05-04
upgrade `react-scripts`/CRA to v5


eventualbuddha:
## Summary
[`react-scripts` v5](https://github.com/facebook/create-react-app/releases/tag/v5.0.0) changes a fair amount. For our purposes it seems the biggest changes are:
- updating to `webpack` v5, which does not bundle NodeJS built-in modules/globals (e.g. `process`, `events`, `util`, `Buffer`, etc) by default
- unpinning dependency versions (e.g. `eslint`, `jest`, etc), which lets us upgrade at least within a major version

## Tasks
- [ ] upgrade any packages that conflict with `react-scripts` v5
- [ ] either stop depending on NodeJS built-in modules/globals OR use `react-app-rewired` or similar to change the webpack config to bring that feature back
  - I'm inclined toward the former because that work is necessary for #1314 anyway

eventualbuddha:
#1781 was related to this

eventualbuddha:
We're not doing this. Doing #1314 instead.

PR #1814, opened 2022-05-11
feat(frontends): separate feature flags from machine config

eventualbuddha:
## Overview
- Allows overriding machine config using `.env` files, making persistent local changes easier to manage.
- Hides "Write-Ins" tabe by default. While this feature is in development it'll be hidden by default, but will be easy to persistently enable locally using `.env` files.
- Move `converter` from machine config to a feature flag. This is something that, for now, is part of the build and will not be changed once the machine is built.

As a result of these changes, we can no longer bypass authentication in production builds. Currently, write-in adjudication will also be removed from production builds. This can be changed when the feature becomes ready to ship to a customer by removing the `NODE_ENV` check.

I'd considered doing this using [`import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta), but it's not easy to augment that with webpack v4 and we cannot use webpack v5 until we upgrade to react-scripts v5 (#1786).

## Demo Video or Screenshot
n/a

## Testing Plan 
Did a fair amount of manual testing and updated cypress to account for the fact that it can no longer bypass authentication.

## Checklist

### DIFF
--- a/frontends/election-manager/src/config/features.ts
+++ b/frontends/election-manager/src/config/features.ts
@@ -1,0 +1,57 @@
+import { unsafeParse } from '@votingworks/types';
+import { asBoolean } from '@votingworks/utils';
+import { ConverterClientType, ConverterClientTypeSchema } from './types';
+
+/**
+ * Determines whether authentication is enabled.
+ *
+ * To disable authentication, add this line to `frontends/election-manager/.env.local`:
+ *
+ *     REACT_APP_VX_BYPASS_AUTHENTICATION=true
+ *
+ * To enable it, remove the line or comment it out. Restarting the server is required.
+ *
+ * @see https://create-react-app.dev/docs/adding-custom-environment-variables/
+ */
+export function isAuthenticationEnabled(): boolean {
+  return (
+    process.env.NODE_ENV !== 'development' ||
+    !asBoolean(process.env.REACT_APP_VX_BYPASS_AUTHENTICATION)
+  );
+}
+
+/**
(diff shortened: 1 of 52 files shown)


carolinemodic (approved):
Awesome! For the future one thing I'm curious about is how we manage features that impact all the apps, it seems like we have to individually define the feature flags for each app in this PR but I imagine in a lot of situations a given flag might be relevant to all the apps so that adds a bit of annoying overhead. Similarly if you want to enable a flag across all apps, is there a way to do that without modifying 4 files? I'm thinking mostly about how to make it easy to turn flags on/off in VxDev

eventualbuddha:
Right now it's definitely manual/annoying. But since the data just lives in `.env` (or, for VxDev, probably `.env.local`) files that are easy to edit by machine, we can easily script things like that in the future.

frontends/election-manager/src/config/features.ts
```diff
@@ -0,0 +1,57 @@
...
+    !asBoolean(process.env.REACT_APP_VX_BYPASS_AUTHENTICATION)
+  );
+}
+
+/**
+ * Determines whether write-in adjudication is enabled.
+ *
+ * To enable write-in adjudication, add this line to `frontends/election-manager/.env.local`:
+ *
+ *     REACT_APP_VX_WRITE_IN_ADJUDICATION=true
```
carolinemodic, line 28: just noticed that this comment doesn't have the "ENABLE" thats in the actual code check

Issue #1793, opened 2022-05-05
Consolidate how write-in values are represented in CVRs


taravancil:
Currently we use a few different conventions: `write-in-{id}`, `__write-in-{id}`, and I believe also `write-in-{typed-value}`. We should consider representing write-in values consistently.

PR #1815, opened 2022-05-12
Represent write-in values consistently

taravancil:
## Overview
#1793

This updates how we represent write-in values to use a consistent format: `write-in-{id|typedValue}`.

I've branched off of the existing WIA work in order to include some unmerged tests from #1809, so there are commits here that do not need to be reviewed.

This is a large diff, but it is almost entirely the result of three different "find and replace"s and running `pnpm build:resources` in `libs/fixtures`. The commits to review:

- a0533cb4d366077274d9c027e34da737e0301361 Replace `__write-in-{id}` with `write-in-{id}`, updates fixtures
- b921ed624f4470455c1fa97a2f28b0ae9d1d3b83 Replace `__writein-{id}` with `write-in-{id}`
- 1a693642f4dc5ae993b757df7e2a8ca5da2d6095 Replace `write-in__{typedValue}` with `write-in-{typedValue}`
- a381b9d67fb859933ff21568c12598db796e5878 Update `normalizeWriteInId` util

I think the most important thing to review is the format I chose: `write-in-{id|typedValue}`. Additionally, if anyone can think of code paths I might have missed with find and replace, that would be super helpful.

## Demo Video or Screenshot
N/A

## Testing Plan 
Automated.

## Checklist


eventualbuddha:
> I think the most important thing to review is the format I chose: `write-in-{id|typedValue}`. Additionally, if anyone can think of code paths I might have missed with find and replace, that would be super helpful.

The first thing I thought of here is that we probably want to choose a string that cannot appear as a candidate ID. We could make it so candidate IDs may not start with `write-in`, which would cover the case you chose. What do you think?

libs/types/src/election.ts
```diff
@@ -98,16 +98,16 @@ export const DistrictsSchema = z
...
+  // TODO: Remove this in favor of `write-in-${string}` or some other unified
   // format for BMD and HMPB write-in IDs.
-  | `write-in__${string}`;
+  | `write-in-${string}`;
 export const WriteInIdSchema = z
   .string()
   .nonempty()
   .refine(
-    (id) => /^(__write-in(-.+)?|write-in__(.+))$/.test(id),
+    (id) => /^(write-in(-.+)?|write-in-(.+))$/.test(id),
```
eventualbuddha, line 110:
```suggestion
    (id) => /^write-in(-.+)?$/.test(id),
```

eventualbuddha (approved):
Looks good, with a couple minor tweaks.

libs/types/src/election.ts
```diff
@@ -98,16 +98,16 @@ export const DistrictsSchema = z
...
 export type WriteInId =
-  | `__write-in`
-  | `__write-in-${string}`
-  // TODO: Remove this in favor of `__write-in-${string}` or some other unified
+  | `write-in`
+  | `write-in-${string}`
+  // TODO: Remove this in favor of `write-in-${string}` or some other unified
   // format for BMD and HMPB write-in IDs.
-  | `write-in__${string}`;
+  | `write-in-${string}`;
```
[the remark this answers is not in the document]
taravancil: Ha, I just caught this myself while working on the CandidateId schema changes, working on it 😄

libs/types/src/election.ts
```diff
@@ -121,13 +116,18 @@ export interface Candidate {
...
+export const CandidateSchema: z.ZodSchema<Candidate> = z
+  .object({
+    _lang: TranslationsSchema.optional(),
+    id: CandidateIdSchema,
+    name: z.string().nonempty(),
+    partyId: PartyIdSchema.optional(),
+    isWriteIn: z.boolean().optional(),
+  })
+  .refine(
+    ({ id, isWriteIn }) => isWriteIn || !id.startsWith('write-in'),
```
eventualbuddha, line 128: Seems like this should be `isWriteIn === id.startsWith('write-in')`, right? Then it will fail in either wrong case.

taravancil: Hm, I _think_ my test cases cover both cases, and this change makes those tests fail.

taravancil: Good catch! Added another test case and fixed in 0b000fee7d7090dfb95ae949ed46144ed4910666

PR #1818, opened 2022-05-12
feat(em): add overvote to hmpb deck

### LINKED ISSUE #1711: Augment VxAdmin test decks with pre-marked overvoted HMBP
Augment VxAdmin test decks with pre-marked overvoted HMBP

adghayes:
## Overview
Closes #1711. Appends an overvote ballot to the HMPB test deck printer from VxAdmin for L&A. 

## Demo Video or Screenshot
Screen Shot 2022-05-12 at 3 16 11 PM

## Testing Plan 

Updated snapshot. @arsalansufi it's worth noting that the snapshots used for HMPB ballots are mocks that include the precinct, election, and ballot style, but don't actually include the votes. Same as in #1808. So we are testing that an additional ballot of the correct ballot style is being added, but not that it is properly marked. 

## Checklist

### DIFF
--- a/frontends/election-manager/src/utils/election.ts
+++ b/frontends/election-manager/src/utils/election.ts
@@ -264,3 +264,45 @@ export function generateBlankBallots({
 
   return ballots;
 }
+
+export function generateOvervoteBallot({
+  election,
+  precinctId,
+}: {
+  election: Election;
+  precinctId: PrecinctId;
+}): Dictionary<string | VotesDict> | undefined {
+  const precinctBallotStyles = election.ballotStyles.filter((bs) =>
+    bs.precincts.includes(precinctId)
+  );
+
+  const ballot: Dictionary<string | VotesDict> = { precinctId };
+  const votes: VotesDict = {};
+
+  for (const ballotStyle of precinctBallotStyles) {
+    const contests = election.contests.filter(
+      (c) =>
+        ballotStyle.districts.includes(c.districtId) &&
+        ballotStyle.partyId === c.partyId
+    );
+
+    for (const contest of contests) {
+      if (contest.type === 'yesno') {
+        votes[contest.id] = ['yes', 'no'];
+      } else if (contest.type === 'ms-either-neither') {
+        votes[contest.eitherNeitherContestId] = ['yes', 'no'];
+      } else if (
+        contest.type === 'candidate' &&
+        contest.candidates.length > contest.seats
+      ) {
+        votes[contest.id] = contest.candidates.slice(0, contest.seats + 1);
+      } else {
+        continue;
+      }
+
+      ballot.ballotStyleId = ballotStyle.id;
+      ballot.votes = votes;
+      return ballot;
+    }
+  }
+}
--- a/frontends/election-manager/src/__snapshots__/app.test.tsx.snap
+++ b/frontends/election-manager/src/__snapshots__/app.test.tsx.snap
@@ -282,6 +282,19 @@ exports[`L&A (logic and accuracy) flow 1`] = `
       6522
       .
     </div>
+    <div>
+      <h1>
+        Mocked HMPB
+      </h1>
+      Election: 
+      Mock General Election Choctaw 2020
+      <br />
+      Ballot Style 
+      5
+      , precinct 
+      6522
+      .
+    </div>
   </div>
   <div />
 </div>
--- a/frontends/election-manager/src/screens/print_test_deck_screen.tsx
+++ b/frontends/election-manager/src/screens/print_test_deck_screen.tsx
@@ -27,6 +27,7 @@ import { HandMarkedPaperBallot } from '../components/hand_marked_paper_ballot';
 import {
   generateTestDeckBallots,
   generateBlankBallots,
+  generateOvervoteBallot,
 } from '../utils/election';
 
 interface TestDeckBallotsParams {
@@ -44,6 +45,9 @@ function TestDeckBallots({
 }: TestDeckBallotsParams) {
   const ballots = generateTestDeckBallots({ election, precinctId });
   ballots.push(...generateBlankBallots({ election, precinctId, numBlanks: 2 }));
+
+  const overvoteBallot = generateOvervoteBallot({ election, precinctId });
+  if (overvoteBallot) ballots.push(overvoteBallot);
 
   let numRendered = 0;
   function onRendered() {


frontends/election-manager/src/utils/election.ts
```diff
@@ -264,3 +264,45 @@ export function generateBlankBallots({
...
+  const votes: VotesDict = {};
+
+  for (const ballotStyle of precinctBallotStyles) {
+    const contests = election.contests.filter(
+      (c) =>
+        ballotStyle.districts.includes(c.districtId) &&
+        ballotStyle.partyId === c.partyId
+    );
+
+    for (const contest of contests) {
```
arsalansufi, line 289: While it's reasonably clear reading the code that we're generating a minimally overvoted ballot, worth adding a comment noting that! Maybe with a (super brief) explanation as to why we made that call

frontends/election-manager/src/utils/election.ts
```diff
@@ -264,3 +264,45 @@ export function generateBlankBallots({
...
+        votes[contest.id] = contest.candidates.slice(0, contest.seats + 1);
+      } else {
+        continue;
+      }
+
+      ballot.ballotStyleId = ballotStyle.id;
+      ballot.votes = votes;
+      return ballot;
+    }
+  }
```
arsalansufi, line 307: > Updated snapshot. @arsalansufi it's worth noting that the snapshots used for HMPB ballots are mocks that include the precinct, election, and ballot style, but don't actually include the votes. Same as in https://github.com/votingworks/vxsuite/pull/1808. So we are testing that an additional ballot of the correct ballot style is being added, but not that it is properly marked.

Ooh good to know! Let's test this function through additional means then. A set of unit tests targeted at this helper, covering the various contest cases as well as edge cases (no contest to validly overvote, etc.), could work well!

In some of our other frontend packages (frontends/bmd), we have automated code coverage checks that verify that every branch of code is covered by a test. While we don't have that set up in frontends/election-manager, I still think the "cover every branch" mindset is helpful, in this case writing tests that cover all the branches within this helper

frontends/election-manager/src/utils/election.ts
```diff
@@ -264,3 +264,45 @@ export function generateBlankBallots({
...
+        contest.candidates.length > contest.seats
+      ) {
+        votes[contest.id] = contest.candidates.slice(0, contest.seats + 1);
+      } else {
+        continue;
+      }
+
+      ballot.ballotStyleId = ballotStyle.id;
+      ballot.votes = votes;
+      return ballot;
```
arsalansufi, line 305: Echoing back my understanding of the code: If there's no contest to validly overvote, I _think_ we'll end up returning the last ballot style with no votes - is that correct / what we want? Based on discussion in https://github.com/votingworks/vxsuite/issues/1711, I believe we want to return nothing at all in that case

arsalansufi (commented):
Thanks again for driving the discussion in https://github.com/votingworks/vxsuite/issues/1711 - it's given me lots of clarity too! :smile:

frontends/election-manager/src/utils/election.ts
```diff
@@ -264,3 +264,45 @@ export function generateBlankBallots({
...
   return ballots;
 }
+
+export function generateOvervoteBallot({
+  election,
+  precinctId,
+}: {
+  election: Election;
+  precinctId: PrecinctId;
+}): Dictionary<string | VotesDict> | undefined {
```
arsalansufi, line 274: Worth adding a comment above the function noting when we expect to return `undefined` (I believe when there's no contest to validly overvote)

frontends/election-manager/src/utils/election.test.ts
```diff
@@ -16,3 +18,77 @@ test('getBallotPath allows digits in file names', () => {
     'election-d34db33f-precinct-north-springfield-id-21-style-77-English-live-absentee.pdf'
   );
 });
+
+describe('generateOvervoteBallot adds an overvote conflict when possible', () => {
```
arsalansufi, line 22: Very minor: People often like to have the combination of their `describe` and `test` / `it` commands read with continuity. Shortening this `describe` to `describe('generateOvervoteBallot', () => {` would accomplish that:

```
generateOvervoteBallot
  minimally overvotes an initial candidate contest
  etc.
```

I think the `adds an overvote conflict when possible` is actually already implied through your descriptive `test` strings!

frontends/election-manager/src/utils/election.test.ts
```diff
@@ -16,3 +18,77 @@ test('getBallotPath allows digits in file names', () => {
...
+
+    expect(Object.keys(votes)).toEqual([overvoteContestId]);
+    expect(votes[overvoteContestId]?.length).toEqual(2);
+  });
+
+  test('overvotes an initial yes-no contest', () => {
+    const election = _.cloneDeep(electionSample);
+
+    // remove all but the yes-no contests
+    _.remove(election.contests, (c) => c.type !== 'yesno');
```
arsalansufi, line 43: While I know we reached the consensus that lodash is fine for us to use in these tests, especially for `cloneDeep`, it's still nice to avoid using it when easily possible. This `remove` can be easily replaced with a `filter`:

```ts
election.contests = election.contests.filter(c => c.type === 'yesno')
```

arsalansufi, line 43: While I know we reached the consensus that lodash is fine for us to use in these tests, especially for `cloneDeep`, it's still nice to avoid using it when easily possible. This `remove` can be easily replaced with a `filter`:

```ts
election.contests = election.contests.filter(c => c.type === 'yesno')
```

frontends/election-manager/src/utils/election.test.ts
```diff
@@ -16,3 +18,77 @@ test('getBallotPath allows digits in file names', () => {
...
+      election: electionSample,
+      precinctId,
+    });
+    expect(overvoteBallot).toBeDefined();
+
+    const votes = overvoteBallot!.votes as VotesDict;
+    const overvoteContestId = electionSample.contests[0].id;
+
+    expect(Object.keys(votes)).toEqual([overvoteContestId]);
+    expect(votes[overvoteContestId]?.length).toEqual(2);
```
arsalansufi, line 36: Did you consider boiling this down to a single assertion of what's in votes?

```test
expect(votes).toEqual(/* Exactly what's in votes */)
```

I can see arguments in favor of both approaches! The current approach makes the qualities of the state you want clear (e.g. 2 votes). The only reason I lean toward the single exact assertion is that the current code could pass if we overvoted a yes-no contest instead of a candidate contest since `['yes', 'no']` is also of length `2`

arsalansufi, line 36: Did you consider boiling this down to a single assertion of what's in votes?

```test
expect(votes).toEqual(/* Exactly what's in votes */)
```

I can see arguments in favor of both approaches! The current approach makes the qualities of the state you want clear (e.g. 2 votes). The only reason I lean toward the single exact assertion is that the current code could pass if we overvoted a yes-no contest instead of a candidate contest since `['yes', 'no']` is also of length `2`

arsalansufi (approved):
This looks great! Thanks for iterating on the tests!

frontends/election-manager/src/utils/election.test.ts
arsalansufi, line 22: Very minor: People often like to have the combination of their `describe` and `test` / `it` commands read with continuity. Shortening this `describe` to `describe('generateOvervoteBallot', () => {` would accomplish that:

```
generateOvervoteBallot
  minimally overvotes an initial candidate contest
  etc.
```

I think the `adds an overvote conflict when possible` is actually already implied through your descriptive `test` strings!

frontends/election-manager/src/utils/election.test.ts
adghayes: The problem I encountered is that the properties on the `Election` type are all `readonly`, so I can't change the array reference. I had to filter the arrays in place, which brought me back to lodash. If there's another way around the `readonly` flag, I would do that.

frontends/election-manager/src/utils/election.test.ts
adghayes: Good catch! While the current approach might be a little easier to read without knowing the fixture data, I think I should simplify as you suggest to avoid passing a yes/no situation.

frontends/election-manager/src/utils/election.test.ts
adghayes: Cool, I'll edit the descriptions in the next commit.

frontends/election-manager/src/utils/election.test.ts
arsalansufi: Ah got it! In that case, this seems totally reasonable

arsalansufi (approved):
Updates LGTM as well!

frontends/election-manager/src/utils/election.test.ts
arsalansufi: Retroactive thought: We actually can avoid lodash `cloneDeep` by using object and array spreads! https://github.com/votingworks/vxsuite/pull/1890/commits/fe843ac4dd7bb860dfdbe410baad8671d20992a8

PR #1819, opened 2022-05-13
feat(ui): Simplify Main/MainChild logic

beausmith:
The `flex-direction` styles were not properly cascading. This commit simplifies the logic to use a `MainChild` with every `Main` component. 

A new `MainChildFlexRow` is created to support a row of containers, such as in the Write-In Adjudication Transcription screen.

The `Main` and `MainChild` components will get cleaned up further to simplify their usage in future commits.


jonahkagan (approved):
Nice work with the tests!

PR #1840, opened 2022-05-17
refactor(types): split out API types

eventualbuddha:
## Overview

Importing from a sub-path of a package doesn't play well with some tooling, especially when those packages are build-less as when using vite.

Refs #1314

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated testing, manual running of `services/scan` with `bsd`.

## Checklist


codemods/src/types_to_api.ts
```diff
@@ -0,0 +1,100 @@
+/**
```
eventualbuddha, line 1: This file did most of the changes to `.ts` files.

jonahkagan (approved):
Nice, I actually kind of like the readability of having the service name as a module prefix when accessing types. It makes it clear that these types are coming from a service.

PR #1842, opened 2022-05-18
feat(frontends, ui): Simpler Screen component

beausmith:
Related to [#44](https://github.com/votingworks/vxsuite-roadmap/issues/44)


libs/ui/src/screen.tsx
```diff
@@ -1,22 +1,17 @@
 import styled from 'styled-components';
 
 interface Props {
-  flexDirection?: 'row' | 'row-reverse' | 'column' | 'column-reverse';
-  voterMode?: boolean;
   white?: boolean;
+  navLeft?: boolean;
+  navRight?: boolean;
```
eventualbuddha, line 6: Since these are mutually exclusive, should it be something like `nav="left"` or `nav="right"` instead?

```suggestion
  nav?: 'left' | 'right'
```

eventualbuddha (approved):

libs/ui/src/screen.tsx
beausmith: You're probably right. There's something I also like about the simple boolean nature of these. But I'll probably change it.

Issue #1479, opened 2022-03-04
More obvious "test mode" banners in VxMark/VxScan


mcchilders:
We currently have banners at the top of every screen when a VxMark/VxScan is in test mode, but we've heard feedback that people are missing them (and leaving machines in test mode). Let's make these obvious/obnoxious so they are hard to miss!

https://votingworks.slack.com/archives/CJU9MSC6S/p1646356787928509

PR #1847, opened 2022-05-18
feat(precinct-scanner): More prominent Testing Mode banner.

beausmith:
Resolves https://github.com/votingworks/vxsuite/issues/1479


carolinemodic (approved):
With moving the tally/cvr stuff around in VxAdmin we should think about how to add this orange paradigm to that to make it clearer when you are looking at test tallies and such.

beausmith:
Good call @carolinemodic. As VxAdmin doesn't have a "testing mode" toggle… but when you import test CVRs — Boom! — you've got a testing banner.

Issue #1738, opened 2022-04-25
Incorporate expected tally reports into VxAdmin test decks (instead of having them be printed separately)


arsalansufi:
Incorporate expected tally reports into VxAdmin test decks (instead of having them be printed separately)

And bring the button to print the expected tally report for all precincts out to the top-level L&A page such that the layout matches that described in [this proposal](https://docs.google.com/document/d/1M7n9fR7VE-qmcEqpW_lFVW3BCFZBwkbhppv9Q2_qJ2Y/edit#heading=h.42t8143wtvw)

Update: Moved the second piece of this task (button to print the expected tally report for all precincts) to its own ticket: https://github.com/votingworks/vxsuite/issues/1869

adghayes:
@arsalansufi my thoughts on steps as part of this change:
- **Refactor to Create "Test Deck Tally Report" component.** There is already an Election Manager Tally Report component, but it appears we have opted not to use that component for the test deck tallies while we do use it for the zero report. I assume this is because the Test Deck Tally Report, unlike the zero report, is _only for testing purposes_, so we wouldn't want to confuse it with any other "Unofficial Results." Props: election, electionTally and precinct
- **Append Test Deck Tally Report to L&A Test Deck.** Seems like we are crossing the threshold from "Test Deck" to "L&A Package" so I think we should now rename some of the key components in `print_test_deck_screen.tsx` to make this clearer. 
- **Create "Print Full Test Deck Tally Report Button" component.** Can model this off of the zero report button component.
- **Place New Button Component on L&A screen**, removing current link to tally report sub-page
- **Delete `test_deck_screen.tsx`** which, despite it's name, is for printing tally reports
- **Update Testing**

adghayes:
Note that currently, the Test Deck Tally page offers an option to save the report to PDF. That will be lost with this change. However, I don't think that matters because it's hard to imagine why exactly you would use that feature for test deck tallies.

adghayes:
@arsalansufi & @adghayes discussed - the above plan of action is acceptable, and we plan to break it out into two different commits / merges: the first to make the TestDeckTally component and use it in the LNA package and the second to add the top-level button.

adghayes:
@arsalansufi I just pushed my latest changes to the branch #1851 - haven't rebased since you might do that later anyway. Notes:

-  Overall printing, timing, & formatting seem to work in single and all precincts case
- `print_test_deck_screen.test.tsx` test failing due to weird timeouts
[...]

PR #1851, opened 2022-05-19
Add Tally Report to L&A Package

adghayes:
## Overview
Removes the dedicated test deck tally report pages and prints the test deck tally report alongside the HMPB test deck, together comprising an "L&A Package." Wraps up the first part of #1738 (that ticket also requires creating a top-level L&A button to print the full election test deck tally report). 

## Testing Plan

L&A testing flow updated to account for new "L&A Package" format.

## Checklist


arsalansufi:
Just pushed some proof of concept test fixes! The key update that I had to make was splitting the two timer advances (33,000 --> 3,000 for tally report + 30,000 for HMPBs)

adghayes:
Gotcha, thanks for figuring that out. Next time we sync I'd like to understand how you figured to do that. 

If those tests are fixed - do you think this is ready to merge after cleaning up the git history, or does it need something else? Don't want to keep you blocked for the VxMark work. After this is merged, I can add the top-level button to print the overall report.

arsalansufi:
Yes I can def walk you through how I arrived at those changes! We can even sync tomorrow. Let me go ahead and do a proper review now! In terms of next task, let's actually prioritize the kiosk-browser tray-targeting investigation next, since that's one of the bigger remaining unknowns in this project

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -215,14 +278,21 @@ export function PrintTestDeckScreen(): JSX.Element {
...
-      )}
+      {printIndex !== undefined &&
+        (printIndex.sectionIndex === 0 ? (
+          <OneSidedMaterials
+            election={election}
+            precinctId={precinctIds[printIndex.precinctIndex]}
+            onAllRendered={onAllOneSidedMaterialsRendered}
+          />
+        ) : (
+          <TwoSidedMaterialsMemoized
```
arsalansufi, line 289: Given this discussion in Slack (https://votingworks.slack.com/archives/CEL6D3GAD/p1652996981036299), we may actually want to define these components not in terms of one-sided/two-sided but in terms of contents, e.g. `<HmpbMaterials>`, `<BmdMaterials>`, `<TallyReports>`

frontends/election-manager/src/app.test.tsx
```diff
@@ -339,8 +339,10 @@ test('L&A (logic and accuracy) flow', async () => {
     exact: false,
   });
   expect(printer.print).toHaveBeenCalledTimes(2);
```
arsalansufi, line 341: Now that we're also printing tally reports, I _think_ we expect one more print call here. If we move this check after the `await waitFor` below, we should be able to check for 3 printer calls

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -130,59 +176,76 @@ export function PrintTestDeckScreen(): JSX.Element {
...
+      if (!printIndex) {
         return;
       }
 
       await printer.print({ sides: 'two-sided-long-edge' });
       await logger.log(LogEventId.TestDeckPrinted, currentUserType, {
         disposition: 'success',
-        message: `Test Deck printed for precinct id: ${precinctIds[precinctIndex]}`,
-        precinctId: precinctIds[precinctIndex],
+        message: `L&A package printed for precinct id: ${
```
arsalansufi, line 208: Sticking to our approach of valid intermediate states, we may now want to update our button titles, too, e.g."Print Test Decks" --> "Print L&A Package"

Likewise, we can remove the "Print Test Deck Tally Reports" button from the L&A page

frontends/election-manager/src/components/test_deck_tally_report.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { filterTalliesByParty } from '@votingworks/utils';
+
+const PrintableTallyReport = styled.div`
+  @page {
+    margin: 0;
+  }
+  section {
+    padding: 0.5in;
+    page-break-after: always;
+  }
```
arsalansufi, line 22: Just for my understanding: How did you land on these style additions? Put another way, where are these styles applied for the current test deck tally reports?

arsalansufi (commented):
Amazing job figuring this out!

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -117,72 +163,97 @@ export function PrintTestDeckScreen(): JSX.Element {
...
+      }
+
+      await printer.print({ sides: 'one-sided' });
+      await logger.log(LogEventId.TestDeckTallyReportPrinted, currentUserType, {
+        disposition: 'success',
+        message: `Test deck tally report printed as part of L&A package for precinct id: ${
+          precinctIds[printIndex.precinctIndex]
+        }`,
+        precinctId: precinctIds[printIndex.precinctIndex],
+      });
```
arsalansufi, line 196: Nice! Thanks for adding this log

arsalansufi (approved):
LGTM!

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -19,26 +21,65 @@ import { Loading } from '../components/loading';
...
+import { TestDeckTallyReport } from '../components/test_deck_tally_report';
 
 import {
   generateTestDeckBallots,
   generateBlankBallots,
   generateOvervoteBallot,
 } from '../utils/election';
 
-interface TestDeckBallotsParams {
+interface PrecinctTallyReportParams {
```
arsalansufi, line 32: Minor comment: I see you were just following the existing interface naming convention! But common practice for React components is to suffix the interface name with `Props`, so `PrecincyTallyReportProps`, `HandMarkedPaperBallotsProps`

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -117,72 +163,97 @@ export function PrintTestDeckScreen(): JSX.Element {
...
+          precinctIds[printIndex.precinctIndex]
+        }`,
+        precinctId: precinctIds[printIndex.precinctIndex],
+      });
+
+      await makeCancelable(sleep(numPages * 3000));
+      setPrintIndex({
+        precinctIndex: printIndex.precinctIndex,
+        component: 'hmpb',
+      });
```
arsalansufi, line 202: Order-wise, I think we actually want the tally report to come after the HMPBs, the thinking being that you scan the test deck ballots and then verify that the tallies match the expected tallies

adghayes: I put it last based on [Ginny's comment in the design doc](https://docs.google.com/document/d/1M7n9fR7VE-qmcEqpW_lFVW3BCFZBwkbhppv9Q2_qJ2Y/edit?disco=AAAAY2nnZfQ). I like it at the front because, since we're not doing spacers, it serves as a signal that this is the next L&A package when printing all of them. But I don't have a strong opinion, and definitely see your point. 

I think the critical piece of ordering is that the VxMark test deck comes before the VxAdmin test deck, since the VxAdmin is padded with the "extra" ballots at the end and it's good to have the two proper test decks be consecutive.

arsalansufi: Ah nice! I'd missed that comment and like Ginny's reasoning for putting the tally report first

arsalansufi:
Great points - I'll think about some of these as I add BMD ballots!

In terms of the timings, I, too, have been a little skeptical of the current timings and agree that we should tweak after testing on physical hardware

Issue #1861, opened 2022-05-20
Coordinate split printing between HP printer trays for L&A documents


adghayes:
## Summary

We want to print all L&A materials for each precinct together as an "L&A Package." In the case where ballots are legal size, however, the materials can't all print out of the same tray because a tray can only be loaded with one type of paper at a time. We can, however, split printing across multiple trays. 

## Hardware Feasibility Research

You can control the source tray (`InputSlot`) with CUPS, and this is already a possible option (`paperSource`) in `kiosk-browser`'s `print` function. If you don't specify a tray, the default `InputSlot` is `Auto` for which the printer decides based on hardware state (in general, the lowest numbered tray that is available and has paper). If do you specify a tray and paper is not available in that tray, the printer will prompt you to add paper or press a button to use a next available tray. We would have to train users to not press the button and instead load paper in those situations.

You can target the optional Tray 3 (`Tray3`) **_if it is set up properly_**.  First, the printer itself must detect Tray 3. If you attach Tray 3 while the printer is powered on, it will not be detected and is not usable. Tray 3 must be attached before the printer is powered on. Second, `Tray3` must be included in the CUPS filters derived from the PPD file. I'm not really sure how that works, but I think the printer provides the PPD file. CUPS updates the PPD file whenever the printer is reconnected or power-cycled. So power cycling the printer with the tray attached _should be_ sufficient to enable targeted printing from Tray 3.

We need a way to check that `Tray3` is in the CUPS filters before we attempt split-tray printing for L&A packages. That information is currently not exposed in `getPrinterInfo.` Our list of paper sources [is hardcoded](https://github.com/votingworks/kiosk-browser/blob/98b0a1a2d506c9f1e76a7c7a31a7362d2a1433f4/src/ipc/print.ts#L19) so does not reflect state. The currently available trays are listed in `/etc/cups/ppd/[printer].ppd` but may be available elsewhere. We need to expose the list on `kiosk`. @eventualbuddha or @benadida I suspect you have more insight here.

Once we do that, we can check for the presence of Tray 3 from `election-manager` and split jobs accordingly. If we were to skip a check, then jobs sent to `Tray3` would actually be treated as `Auto` and print out of Tray 2. 

## Tasks
- `kiosk-browser`: Expose the options for the InputSlot CUPS filters in `kiosk-browser`
- `vxsuite`: Prevent printing L&A materials if the extra tray is required and it is not available & split the jobs appropriately when printed

adghayes:
[...]

PR #1863, opened 2022-05-23
Augment VxAdmin test decks with VxMark test decks

arsalansufi:

# Overview

Issue link: https://github.com/votingworks/vxsuite/issues/1710

This PR augments the VxAdmin test decks (currently only hand-marked paper ballots) with the VxMark test decks (BMD paper ballots) and updates expected test deck tally reports accordingly (multiplying expected tallies by 2).

Election officials will now be able to print everything they need for L&A from VxAdmin!

We still have a number of additional improvements to make. I've taken note of a few where relevant in GitHub comments.

# Demo video

Demo video for a single precinct. Prints the precinct tally report, BMD test deck, and HMPB test deck (including two blank ballots and one overvoted ballot).

https://user-images.githubusercontent.com/12616928/169854037-e7cadc97-fa00-4b56-898a-906156bfe600.mov

When "All Precincts" is clicked, VxAdmin will print the above for all precincts.

Note that, when using kiosk-browser, users won't see any browser print dialogs. The browser will automatically proceed from print job to print job.

# Testing

- [x] Manually tested single-precinct case
- [x] Manually tested all-precincts case
- [x] Updated tests and snapshots

## Checklist

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -68,6 +68,16 @@ const EITHER_NEITHER_SEMS_DATA =
 
 jest.mock('./components/hand_marked_paper_ballot');
 jest.mock('./utils/pdf_to_images');
+jest.mock('@votingworks/utils', (): typeof import('@votingworks/utils') => {
+  const original: typeof import('@votingworks/utils') =
+    jest.requireActual('@votingworks/utils');
+  // Mock random string generation so that snapshots match, while leaving the rest of the module
+  // intact
+  return {
+    ...original,
+    randomBallotId: () => 'Asdf1234Asdf12',
+  };
+});
 
 let mockKiosk!: jest.Mocked<KioskBrowser.Kiosk>;
 
@@ -331,12 +341,12 @@ test('L&A (logic and accuracy) flow', async () => {
   );
   expect(screen.getAllByText('0').length).toBe(40);
 
-  // Test printing L&A Package
+  // Test printing L&A package
   userEvent.click(screen.getByText('L&A'));
   userEvent.click(screen.getByText('Print L&A Packages'));
   userEvent.click(screen.getByText('District 5'));
 
-  // L&A Package: Tally Report
+  // L&A package: Tally report
   await screen.findByText('Printing L&A Package: District 5', {
     exact: false,
   });
@@ -347,9 +357,9 @@ test('L&A (logic and accuracy) flow', async () => {
     )
   );
   expect(container).toMatchSnapshot();
-  jest.advanceTimersByTime(3000);
+  jest.advanceTimersByTime(5000);
 
-  // L&A Package: HMBP Test Deck
+  // L&A package: BMD test deck
   await screen.findByText('Printing L&A Package: District 5', {
     exact: false,
   });
@@ -358,6 +368,25 @@ test('L&A (logic and accuracy) flow', async () => {
     expect(mockKiosk.log).toHaveBeenCalledWith(
       expect.stringContaining(LogEventId.TestDeckPrinted)
     )
+  );
+  expect(mockKiosk.log).toHaveBeenCalledWith(
+    expect.stringContaining('BMD paper ballot test deck')
+  );
+  expect(container).toMatchSnapshot();
+  jest.advanceTimersByTime(30000);
+
+  // L&A package: HMPB test deck
+  await screen.findByText('Printing L&A Package: District 5', {
+    exact: false,
+  });
+  expect(printer.print).toHaveBeenCalledTimes(4);
+  await waitFor(() =>
+    expect(mockKiosk.log).toHaveBeenCalledWith(
+      expect.stringContaining(LogEventId.TestDeckPrinted)
+    )
+  );
+  expect(mockKiosk.log).toHaveBeenCalledWith(
+    expect.stringContaining('Hand-marked paper ballot test deck')
   );
   expect(container).toMatchSnapshot();
 });
--- a/frontends/election-manager/src/screens/print_test_deck_screen.test.tsx
+++ b/frontends/election-manager/src/screens/print_test_deck_screen.test.tsx
@@ -43,8 +43,7 @@ test('Printing all L&A packages sorts precincts', async () => {
     'Southwest Ackerman',
     'West Weir',
   ];
-  for (let i = 0; i < precinctsInAlphabeticalOrder.length; i += 1) {
-    const precinct = precinctsInAlphabeticalOrder[i];
+  for (const [i, precinct] of precinctsInAlphabeticalOrder.entries()) {
     const printText = `Printing L&A Package (${i + 1} of 13): ${precinct}.`;
     await screen.findByLabelText(printText);
     jest.advanceTimersByTime(5000);
--- a/frontends/election-manager/src/screens/print_test_deck_screen.test.tsx
+++ b/frontends/election-manager/src/screens/print_test_deck_screen.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
-import { fireEvent, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { fakeKiosk, fakePrinterInfo } from '@votingworks/test-utils';
+import { screen } from '@testing-library/react';
 
 import { PrintTestDeckScreen } from './print_test_deck_screen';
 import { renderInAppContext } from '../../test/render_in_app_context';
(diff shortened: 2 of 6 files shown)


frontends/election-manager/src/utils/election.test.ts
```diff
@@ -29,7 +29,7 @@ describe('generateOvervoteBallot', () => {
     });
     expect(overvoteBallot).toBeDefined();
 
-    const votes = overvoteBallot!.votes as VotesDict;
+    const { votes } = overvoteBallot!;
```
arsalansufi, line 32: This casting is no longer necessary because of the `Ballot` type I define in `frontends/election-manager/src/utils/election.ts` below

frontends/election-manager/src/utils/election.ts
```diff
@@ -269,23 +274,20 @@ export function generateBlankBallots({
...
-// In cases where it is not possible to overvote a ballot style, returns undefined.
+// In cases where it is not possible to overvote a ballot style, returns null.
 export function generateOvervoteBallot({
   election,
   precinctId,
 }: {
   election: Election;
   precinctId: PrecinctId;
-}): Dictionary<string | VotesDict> | undefined {
+}): Ballot | null {
```
arsalansufi, line 284: I have a slight preference for `null` over `undefined` as an explicitly assigned empty value

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -99,14 +99,14 @@ function HandMarkedPaperBallots({
       {ballots.map((ballot, i) => (
         <HandMarkedPaperBallot
           key={`ballot-${i}`} // eslint-disable-line react/no-array-index-key
-          ballotStyleId={ballot.ballotStyleId as string}
+          ballotStyleId={ballot.ballotStyleId}
```
arsalansufi, line 146: This casting is no longer necessary, again because of the new `Ballot` type def

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -141,7 +141,7 @@ export function PrintTestDeckScreen(): JSX.Element {
   const currentUserType = currentUserSession.type;
   const { election, electionHash } = electionDefinition;
   const [precinctIds, setPrecinctIds] = useState<string[]>([]);
-  const [printIndex, setPrintIndex] = useState<PrintIndex | undefined>();
+  const [printIndex, setPrintIndex] = useState<PrintIndex | null>();
```
arsalansufi, line 189: Once again, I have a slight preference for `null` over `undefined` as an explicitly assigned empty value

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -210,43 +209,43 @@ export function PrintTestDeckScreen(): JSX.Element {
...
+        await makeCancelable(sleep(numBallots * TWO_SIDED_PAGE_PRINT_TIME_MS));
         setPrintIndex({
           precinctIndex: printIndex.precinctIndex + 1,
-          component: 'tally',
+          component: 'TallyReport',
         });
       } else {
-        await makeCancelable(sleep(3000));
-        setPrintIndex(undefined);
+        await makeCancelable(sleep(numBallots * TWO_SIDED_PAGE_PRINT_TIME_MS));
```
arsalansufi, line 295: I believe we were previously forgetting to multiply by `numBallots` here

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -64,6 +74,40 @@ function PrecinctTallyReport({
...
+  }, [ballots, onAllRendered]);
+
+  return (
+    <div className="print-only">
+      {ballots.map((ballot, i) => (
+        <BmdPaperBallot
+          ballotStyleId={ballot.ballotStyleId}
+          electionDefinition={electionDefinition}
+          isLiveMode={false}
+          key={`ballot-${i}`} // eslint-disable-line react/no-array-index-key
```
arsalansufi, line 102: I'd like to avoid using an array index as a key here and in other existing code in this file but want to think through how to best approach this, e.g. hash ballot.votes to create a unique identifier, etc.

https://github.com/votingworks/vxsuite/issues/1864

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -197,12 +242,35 @@ export function PrintTestDeckScreen(): JSX.Element {
...
+
+      const precinctId = precinctIds[printIndex.precinctIndex];
+      await printer.print({ sides: 'one-sided' });
+      await logger.log(LogEventId.TestDeckPrinted, currentUserType, {
+        disposition: 'success',
+        message: `BMD paper ballot test deck printed as part of L&A package for precinct ID: ${precinctId}`,
+        precinctId,
+      });
+
+      await makeCancelable(sleep(numBallots * ONE_SIDED_PAGE_PRINT_TIME_MS));
```
arsalansufi, line 265: We're planning to test with real hardware to determine whether our print time estimates are accurate. This will become especially important as we introduce the ability to print to different trays depending on paper size

https://github.com/votingworks/vxsuite/issues/1865
https://github.com/votingworks/vxsuite/issues/1861

frontends/election-manager/src/app.test.tsx
```diff
@@ -359,6 +369,25 @@ test('L&A (logic and accuracy) flow', async () => {
...
+  });
+  expect(printer.print).toHaveBeenCalledTimes(4);
+  await waitFor(() =>
+    expect(mockKiosk.log).toHaveBeenCalledWith(
+      expect.stringContaining(LogEventId.TestDeckPrinted)
+    )
+  );
+  expect(mockKiosk.log).toHaveBeenCalledWith(
+    expect.stringContaining('Hand-marked paper ballot test deck')
+  );
```
arsalansufi, line 390: The snapshots throughout this test provide solid coverage, but I do plan to return to these tests to add explicit assertions to make what we're testing clearer, e.g. assertions for expected tally counts, etc.

https://github.com/votingworks/vxsuite/issues/1867

frontends/election-manager/src/screens/print_test_deck_screen.tsx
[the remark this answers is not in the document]
arsalansufi: Ah that makes a lot of sense! I'll switch back and add a comment noting our preference to free up the UI from the modal earlier

frontends/election-manager/src/utils/election.ts
eventualbuddha: We've generally preferred `undefined` over `null` in `vxsuite`. If you think it'd be better to prefer `null`, that seems like a good discussion to have. For now, mind leaving "unset" values as `undefined`?

frontends/election-manager/src/screens/print_test_deck_screen.tsx
eventualbuddha: One advantage of using `undefined` is that you can just do this:

```suggestion
  const [printIndex, setPrintIndex] = useState<PrintIndex>();
```

frontends/election-manager/src/screens/print_test_deck_screen.test.tsx
```diff
@@ -22,110 +23,36 @@ test('Printing all L&A packages sorts precincts', async () => {
...
+    'Hebron',
+    'Kenego',
+    'Panhandle',
+    'Reform',
+    'Sherwood',
+    'Southwest Ackerman',
+    'West Weir',
+  ];
+  for (let i = 0; i < precinctsInAlphabeticalOrder.length; i += 1) {
+    const precinct = precinctsInAlphabeticalOrder[i];
```
eventualbuddha, line 47: This is fine, but I find this more readable:

```suggestion
  for (const [i, precinct] of precinctsInAlphabeticalOrder.entries()) {
```

frontends/election-manager/src/utils/election.ts
```diff
@@ -20,6 +19,12 @@ import { LANGUAGES } from '../config/globals';
 
 import { sortBy } from './sort_by';
 
+export interface Ballot {
+  ballotStyleId: string;
+  precinctId: string;
```
eventualbuddha, line 24:
```suggestion
  ballotStyleId: BallotStyleId;
  precinctId: PrecinctId;
```

frontends/election-manager/src/utils/election.ts
arsalansufi: Ooh good to know! My preference isn't strong, and I'm happy to switch back to `undefined` for consistency

frontends/election-manager/src/screens/print_test_deck_screen.test.tsx
arsalansufi: Nice! I prefer what you've suggested too

frontends/election-manager/src/screens/print_test_deck_screen.tsx
arsalansufi: Update: There was a reason for this! https://github.com/votingworks/vxsuite/pull/1863#discussion_r879666027

Switched back with tweaks for clarity: https://github.com/votingworks/vxsuite/pull/1863/commits/6802328453e09b05ab0cbf58c8cbb874a51df626

frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -67,7 +75,41 @@ function PrecinctTallyReport({
...
+  precinctId: string;
+}
+
+function BmdPaperBallots({
+  electionDefinition,
+  onAllRendered,
+  precinctId,
+}: BmdPaperBallotsProps): JSX.Element {
+  const { election } = electionDefinition;
+  const ballots = generateTestDeckBallots({ election, precinctId });
```
arsalansufi, line 90: Btw @adghayes I tried moving ballot generation out of these components and to the top-level `PrintTestDeckScreen` as you considered in https://github.com/votingworks/vxsuite/pull/1851#issuecomment-1133058527! With `PrecinctTallyReport`, `BmdPaperBallots`, and `HandMarkedPaperBallots` all taking `ballots` props as inputs

But this interfered with the temporary `React.memo` workaround. Happy to dive into further details in person!

adghayes: Interesting!

arsalansufi:
@adghayes This is now ready for a follow-up look! Addressed all comments ✅

adghayes:
I'm not sure if I am sanctioned to review PRs for merge yet, but it looks good to me. I think the final print delay was previously 3 seconds rather than 5 seconds, but we should revisit that when we revisit all the other print timings so I don't think it's worth making another change here.

arsalansufi:
> I think the final print delay was previously 3 seconds rather than 5 seconds, but we should revisit that when we revisit all the other print timings so I don't think it's worth making another change here.

Yeah, I switched to the 5 seconds somewhat arbitrarily but agree that we can revisit all the timings holistically! Brian left a great suggestion on https://github.com/votingworks/vxsuite/issues/1865 to alternatively watch the print jobs

> I'm not sure if I am sanctioned to review PRs for merge yet, but it looks good to me. 

Chatting with Caro about this today!

frontends/election-manager/src/screens/print_test_deck_screen.test.tsx
```diff
@@ -22,110 +23,35 @@ test('Printing all L&A packages sorts precincts', async () => {
...
+    'Fentress',
+    'French Camp',
+    'Hebron',
+    'Kenego',
+    'Panhandle',
+    'Reform',
+    'Sherwood',
+    'Southwest Ackerman',
+    'West Weir',
+  ];
```
eventualbuddha, line 45: Is there a benefit to hardcoding this over `[...election.precincts.map(({ name }) => name)].sort()`?

eventualbuddha (approved):

PR #1876, opened 2022-05-24
Fix occasional test deck misordering

arsalansufi:
# Overview

Issue link: https://github.com/votingworks/vxsuite/issues/1868

This was a super interesting one. Drew and I were occasionally seeing misordered test decks get generated on our laptops, e.g. ballot styles were interleaved, the first HMPB in the test deck had bubbles other than the first in each contest filled in, etc.

After some investigation, we pinpointed this to an issue with our Paged.js-processing of HMPBs. While the ballots are [originally rendered in the correct order](https://github.com/votingworks/vxsuite/blob/3e1689bdb1db2e7bf9c4e874ab74662d78fd9275/frontends/election-manager/src/screens/print_test_deck_screen.tsx#L144-L157), there's no guarantee that the initial order will be maintained as [each is processed by Paged.js](https://github.com/votingworks/vxsuite/blob/3e1689bdb1db2e7bf9c4e874ab74662d78fd9275/frontends/election-manager/src/components/hand_marked_paper_ballot.tsx#L623-L627). Every ballot processes itself, unaware of the other ballots, and renders its processed self to a shared target element ([the element referenced by `printBallotRef`](https://github.com/votingworks/vxsuite/blob/3e1689bdb1db2e7bf9c4e874ab74662d78fd9275/frontends/election-manager/src/components/hand_marked_paper_ballot.tsx#L626)).

In the screenshot below, for example, you can see that what was originally ballot 22 is the first ballot among the processed ballots.

pagedjs

To fix this, we now render every processed ballot to a unique target element instead of a shared one. These target elements are created and cleared on the fly as needed.

Below is a (surprisingly relaxing 😌) video demonstrating this in action. Note how processing begins with ballot 22 and loops back to ballot 0 - before this fix, this would've resulted in a misordered test deck!

https://user-images.githubusercontent.com/12616928/170114229-e7e25d1c-8705-4654-8622-c472d623d347.mov

# Testing

- [x] Verified that automated tests still pass (with small expected tweaks to snapshots/mocks)
- [x] Manually tested the single-precinct case
- [x] Manually tested the all-precincts case
- [x] Verified that target elements were created _and cleared_ on the fly, via Chrome dev tools

# Checklist

- [ ] ~I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates~ N/A

### DIFF
--- a/frontends/election-manager/src/__snapshots__/app.test.tsx.snap
+++ b/frontends/election-manager/src/__snapshots__/app.test.tsx.snap
@@ -2931,7 +2931,26 @@ exports[`L&A (logic and accuracy) flow 3`] = `
       .
     </div>
   </div>
-  <div />
+  <div>
+    <div
+      id="ballot-0"
+    />
+    <div
+      id="ballot-1"
+    />
+    <div
+      id="ballot-2"
+    />
+    <div
+      id="ballot-3"
+    />
+    <div
+      id="ballot-4"
+    />
+    <div
+      id="ballot-5"
+    />
+  </div>
 </div>
 `;
 
--- a/frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
+++ b/frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
@@ -532,6 +532,7 @@ export interface HandMarkedPaperBallotProps {
   ballotId?: BallotId;
   votes?: VotesDict;
   onRendered?(props: Omit<HandMarkedPaperBallotProps, 'onRendered'>): void;
+  targetElementId?: string;
 }
 
 export function HandMarkedPaperBallot({
@@ -546,6 +547,7 @@ export function HandMarkedPaperBallot({
   ballotId,
   votes,
   onRendered,
+  targetElementId,
 }: HandMarkedPaperBallotProps): JSX.Element {
   const layoutDensity = getBallotLayoutDensity(election);
   assert(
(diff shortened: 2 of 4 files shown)


adghayes:
@arsalansufi what's the reason for / advantage of the current implementation instead of just wrapping each `HandMarkedPaperBallot` in a `div` here? https://github.com/votingworks/vxsuite/blob/4e047d6a90d56b0befb10b6d901e121965c8f489/frontends/election-manager/src/screens/print_test_deck_screen.tsx#L157-L161

benadida (approved):
I don't have an issue with this implementation, though I would +1 @adghayes 's question: is there a much simpler way to do this by just adding a `<div>` in the markup itself?

Approve to give you flexibility to merge since I don't see a problem with this implementation, but nudge to consider the simpler solution if it does, in fact, work.

arsalansufi:
Excellent question @adghayes and @benadida! I should've left some notes proactively about why I ultimately didn't take that approach, since you're absolutely right that it's the intuitive approach.

I tried a few different permutations of divs rendered through React either in `HandMarkedPaperBallot` or one-level higher in `HandMarkedPaperBallots` (plural) but couldn't get them to work.

For example, when I tried wrapping `HandMarkedPaperBallot` in a new div and rendering to that, the resulting ballots were styled incorrectly.

improperly-rendered-ballot

I didn't get to an exact root cause, but my sense was that the React render lifecycle was interrupting Paged.js processing.

When I moved the divs up one-level to `HandMarkedPaperBallots` (plural), same issue.

I ultimately concluded that the divs needed to be rendered 1) separate from `HandMarkedPaperBallot` and 2) in advance of `HandMarkedPaperBallot`. To meet these criteria , I rendered the divs within the isolated div that HMPBs are currently rendered to (`printBallotRef`). And I did so in a way that guarantees that the divs exist before `HandMarkedPaperBallot` tries rendering to them.

All this said, I could perhaps still meet these criteria while also rendering through React instead of JS. Let me do a final timeboxed exploration to see if I can strike a happy medium! I'll also add a brief comment noting the constraints I've encountered.

benadida:
> Excellent question @adghayes and @benadida! I should've left some notes proactively about why I ultimately didn't take that approach, since you're absolutely right that it's the intuitive approach.

I'm convinced! My goal is simplicity, and if this is the simplest thing that *works* now, let's do it and move forward.

PR #1877, opened 2022-05-25
Full Test Deck Tally Report Button

### LINKED ISSUE #1869: Bring the button to print the expected tally report for all precincts out to the top-level L&A page

adghayes:
## Overview
Closes #1869. Adds a button to the L&A page which prints a test deck tally report for all precincts. The totals in the report are 4 times the totals of a single test deck, because it is used to compare with results after scanning two precinct test decks at each precinct and two full test decks at the central scanner.

## Demo Video or Screenshot

Screen Shot 2022-05-24 at 5 35 59 PM

full_test_deck_tally_report

## Testing Plan 

Updated L&A flow to test printing of the report and its totals.

## Checklist
- [ ] ~~I have added JSDoc comments to any newly introduced exports~~

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -9,6 +9,7 @@ import {
   getByTestId as domGetByTestId,
   getByText as domGetByText,
   getAllByRole as domGetAllByRole,
+  getAllByText as domGetAllByText,
   act,
   within,
 } from '@testing-library/react';
@@ -340,7 +341,7 @@ test('L&A (logic and accuracy) flow', async () => {
   const storage = await createMemoryStorageWith({
     electionDefinition: eitherNeitherElectionDefinition,
   });
-  const { container } = render(
+  const { container, getByTestId } = render(
     <App card={card} hardware={hardware} printer={printer} storage={storage} />
   );
   jest.advanceTimersByTime(2001); // Cause the usb drive to be detected
@@ -349,12 +350,15 @@ test('L&A (logic and accuracy) flow', async () => {
   // Test printing zero report
   userEvent.click(screen.getByText('L&A'));
   userEvent.click(screen.getByText(ZERO_REPORT_BUTTON_TEXT));
+  await waitFor(() => {
+    const zeroReport = getByTestId('zero-report');
+    expect(domGetAllByText(zeroReport, '0').length).toBe(40);
+  });
   await screen.findByText('Printing');
   expect(printer.print).toHaveBeenCalledTimes(1);
   expect(mockKiosk.log).toHaveBeenCalledWith(
     expect.stringContaining(LogEventId.TallyReportPrinted)
   );
-  expect(screen.getAllByText('0').length).toBe(40);
 
   // Test printing L&A package
   userEvent.click(screen.getByText('L&A'));
@@ -404,6 +408,34 @@ test('L&A (logic and accuracy) flow', async () => {
     expect.stringContaining('Hand-marked paper ballot test deck')
   );
   expect(container).toMatchSnapshot();
+
+  // Test printing full test deck tally
+  const expectedTallies: { [tally: string]: number } = {
+    '104': 6,
+    '52': 12,
+    '36': 2,
+    '24': 2,
+    '12': 3,
+    '8': 1,
+    '4': 1,
+    '0': 10,
+  };
+
+  userEvent.click(screen.getByText('L&A'));
+  userEvent.click(screen.getByText('Print Full Test Deck Tally Report'));
+  await waitFor(() => {
+    const fullTestDeckTallyReport = getByTestId('full-test-deck-tally-report');
+    for (const [tally, times] of Object.entries(expectedTallies)) {
+      expect(domGetAllByText(fullTestDeckTallyReport, tally).length).toEqual(
+        times
+      );
+    }
+  });
+  await screen.findByText('Printing');
+  expect(printer.print).toHaveBeenCalledTimes(5);
+  expect(mockKiosk.log).toHaveBeenCalledWith(
+    expect.stringContaining(LogEventId.TestDeckTallyReportPrinted)
+  );
 });
 
 test('L&A features are available after test results are loaded', async () => {
--- a/frontends/election-manager/src/components/full_test_deck_tally_report_button.tsx
+++ b/frontends/election-manager/src/components/full_test_deck_tally_report_button.tsx
@@ -1,0 +1,70 @@
+import React, { useContext, useCallback } from 'react';
+import { assert, tallyVotesByContest } from '@votingworks/utils';
+import { LogEventId } from '@votingworks/logging';
+import { Tally, VotingMethod } from '@votingworks/types';
+
+import { AppContext } from '../contexts/app_context';
+import { PrintButton } from './print_button';
+import { TestDeckTallyReport } from './test_deck_tally_report';
+import { generateTestDeckBallots } from '../utils/election';
+
+export function FullTestDeckTallyReportButton(): JSX.Element {
+  const { currentUserSession, electionDefinition, logger } =
+    useContext(AppContext);
+
+  assert(currentUserSession);
+  const currentUserType = currentUserSession.type;
+
+  assert(electionDefinition);
+  const { election } = electionDefinition;
+
+  const ballots = generateTestDeckBallots({ election });
+  const votes = ballots.map((b) => b.votes);
+
+  // Full test deck tally is 4 times a single test deck tally, because it counts scanning
+  // 2 test decks (BMD + HMPB) in 2 places (VxScan + VxCentralScan)
+  const testDeckTally: Tally = {
+    numberOfBallotsCounted: ballots.length * 4,
+    castVoteRecords: new Set(),
+    contestTallies: tallyVotesByContest({
+      election,
+      votes: [...votes, ...votes, ...votes, ...votes],
+    }),
+    ballotCountsByVotingMethod: { [VotingMethod.Unknown]: ballots.length },
+  };
+
+  const afterPrint = useCallback(() => {
+    void logger.log(LogEventId.TestDeckTallyReportPrinted, currentUserType, {
+      disposition: 'success',
+      message: `User printed the full test deck tally report`,
+    });
+  }, [logger, currentUserType]);
+
+  const afterPrintError = useCallback(
+    (errorMessage: string) => {
+      void logger.log(LogEventId.TestDeckTallyReportPrinted, currentUserType, {
+        disposition: 'failure',
+        errorMessage,
+        message: `Error printing the full test deck tally report: ${errorMessage}`,
+        result: 'User shown error.',
+      });
+    },
+    [logger, currentUserType]
+  );
+
+  const fullTestDeckTallyReport = (
+    <TestDeckTallyReport election={election} electionTally={testDeckTally} />
+  );
+
+  return (
+    <PrintButton
+      afterPrint={afterPrint}
+      afterPrintError={afterPrintError}
+      sides="one-sided"
+      printTarget={fullTestDeckTallyReport}
+      printTargetTestId="full-test-deck-tally-report"
+    >
+      Print Full Test Deck Tally Report
+    </PrintButton>
+  );
+}
--- a/frontends/election-manager/src/components/print_button.test.tsx
+++ b/frontends/election-manager/src/components/print_button.test.tsx
@@ -1,6 +1,13 @@
 import React from 'react';
-import { act, fireEvent, waitFor } from '@testing-library/react';
+import {
+  act,
+  fireEvent,
+  waitFor,
+  getByText as domGetByText,
+  waitForElementToBeRemoved,
+} from '@testing-library/react';
 import { fakeKiosk, fakePrinterInfo } from '@votingworks/test-utils';
+import { deferred } from '@votingworks/utils';
 
 import { PrintButton } from './print_button';
 import { fakePrinter } from '../../test/helpers/fake_printer';
@@ -70,4 +77,49 @@ test('if connected printers, show printing modal', async () => {
 
   expect(printer.print).toBeCalled();
   expect(afterPrint).toBeCalled();
+});
+
+test('if passed a printTarget, render for printing', async () => {
+  const mockKiosk = window.kiosk! as jest.Mocked<KioskBrowser.Kiosk>;
+  mockKiosk.getPrinterInfo.mockResolvedValue([
+    fakePrinterInfo({ connected: false }),
+    fakePrinterInfo({ name: 'VxPrinter', connected: true }),
+  ]);
+
+  const printer = fakePrinter();
+  const printTarget = <p>Print Target</p>;
+  const printTargetTestId = 'print-target';
+  const { getByTestId, getByText, queryByText } = renderInAppContext(
+    <PrintButton
+      sides="two-sided-long-edge"
+      printTarget={printTarget}
+      printTargetTestId={printTargetTestId}
+    >
+      Print Now
+    </PrintButton>,
+    { printer }
+  );
+
+  const printDeferred = deferred<void>();
+  printer.print.mockReturnValueOnce(printDeferred.promise);
+  fireEvent.click(getByText('Print Now'));
+
+  // Check that printTarget is rendered before printing
+  await waitFor(() => {
+    const renderedPrintTarget = getByTestId('print-target');
+    domGetByText(renderedPrintTarget, 'Print Target');
+    expect(printer.print).not.toBeCalled();
+  });
+
+  // Check that printTarget is still rendered after printing
+  await waitFor(() => {
+    expect(printer.print).toBeCalled();
+    getByText('Print Target');
+  });
+
+  // Allow `printer.print` to resolve
+  printDeferred.resolve();
+
+  // Check that printTarget is removed after printing
+  await waitForElementToBeRemoved(() => queryByText('Print Target'));
 });
--- a/frontends/election-manager/src/components/print_button.tsx
+++ b/frontends/election-manager/src/components/print_button.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useState } from 'react';
+import React, { useContext, useState, useEffect, useCallback } from 'react';
 
 import {
   Modal,
@@ -10,6 +10,7 @@ import { Button, StyledButtonProps } from './button';
 import { Loading } from './loading';
 import { PrintOptions } from '../config/types';
 import { AppContext } from '../contexts/app_context';
+import { PrintableArea } from './printable_area';
 
 interface ConfirmModal {
   content: React.ReactNode;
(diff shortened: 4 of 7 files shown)


frontends/election-manager/src/components/zero_report_print_button.tsx
```diff
@@ -36,29 +36,44 @@ export function ZeroReportPrintButton(): JSX.Element {
...
-  function logZeroReportPrintSuccess() {
+  function disablePrinting() {
+    document.getElementById('zero-report')?.classList.add('no-print');
+  }
+
+  function enablePrinting() {
+    document.getElementById('zero-report')?.classList.remove('no-print');
+  }
+
+  useEffect(() => {
```
adghayes, line 47: Using useEffect to disable printing immediately after render seemed like the simplest to disable printing without complicating the `PrintableArea` component further, although that is an option if we want to avoid `useEffect`.

frontends/election-manager/src/app.test.tsx
```diff
@@ -339,7 +340,8 @@ test('L&A (logic and accuracy) flow', async () => {
   expect(mockKiosk.log).toHaveBeenCalledWith(
     expect.stringContaining(LogEventId.TallyReportPrinted)
   );
-  expect(screen.getAllByText('0').length).toBe(40);
+  const zeroReport = document.querySelector('#zero-report') as HTMLElement;
+  expect(domGetAllByText(zeroReport, '0').length).toBe(40);
```
adghayes, line 344: Now that another report is rendered on the same screen, we have to isolate the zero report to test its contents.

frontends/election-manager/src/app.test.tsx
```diff
@@ -389,6 +391,33 @@ test('L&A (logic and accuracy) flow', async () => {
...
+    '52': 12,
+    '36': 2,
+    '24': 2,
+    '12': 3,
+    '8': 1,
+    '4': 1,
+    '0': 10,
+  };
+  for (const [tally, times] of Object.entries(expectedTallies)) {
+    expect(domGetAllByText(fullTestDeckTallyReport, tally).length).toEqual(
```
adghayes, line 417: Extending the approach from testing the contents of the zero report, this seemed preferable to a snapshot as we are trying to minimize snapshots.

frontends/election-manager/src/components/full_test_deck_tally_report_button.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  return (
+    <React.Fragment>
+      <PrintButton
+        afterPrint={afterPrint}
+        afterPrintError={afterPrintError}
+        beforePrint={enablePrinting}
+        sides="one-sided"
+      >
+        Print Full Test Deck Tally Report
```
adghayes, line 76: Can certainly change copy if desired.

frontends/election-manager/src/app.test.tsx
```diff
@@ -339,7 +340,8 @@ test('L&A (logic and accuracy) flow', async () => {
   expect(mockKiosk.log).toHaveBeenCalledWith(
     expect.stringContaining(LogEventId.TallyReportPrinted)
   );
-  expect(screen.getAllByText('0').length).toBe(40);
+  const zeroReport = document.querySelector('#zero-report') as HTMLElement;
```
eventualbuddha, line 343: Can we give it a `data-testid` and do something like this:

```suggestion
  const zeroReport = screen.getByTestId('zero-report');
```

frontends/election-manager/src/app.test.tsx
```diff
@@ -389,6 +391,33 @@ test('L&A (logic and accuracy) flow', async () => {
...
+  // Test printing full test deck tally
+  userEvent.click(screen.getByText('L&A'));
+  userEvent.click(screen.getByText('Print Full Test Deck Tally Report'));
+  await screen.findByText('Printing');
+  expect(printer.print).toHaveBeenCalledTimes(5);
+  expect(mockKiosk.log).toHaveBeenCalledWith(
+    expect.stringContaining(LogEventId.TestDeckTallyReportPrinted)
+  );
+  const fullTestDeckTallyReport = document.querySelector(
+    '#full-test-deck-tally-report'
```
eventualbuddha, line 404: Ditto `data-testid` here.

frontends/election-manager/src/app.test.tsx
```diff
@@ -612,7 +641,8 @@ test('tabulating CVRs', async () => {
   expect(mockKiosk.log).toHaveBeenCalledWith(
     expect.stringContaining(LogEventId.TallyReportPrinted)
   );
-  expect(getAllByText('0').length).toBe(40);
+  const zeroReport = document.querySelector('#zero-report') as HTMLElement;
```
eventualbuddha, line 644: And again.

frontends/election-manager/src/components/full_test_deck_tally_report_button.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { useContext, useEffect } from 'react';
```
eventualbuddha, line 1: This file is reasonably well covered by `app.test.tsx`, but I'm not sure how much of the actual behavior it's testing. For example, the tests seem to pass even if I comment out the bodies of `disablePrinting` and `enablePrinting`. Perhaps add a unit test that can test those more specific things?

frontends/election-manager/src/components/full_test_deck_tally_report_button.tsx
```diff
@@ -0,0 +1,86 @@
...
+      election,
+      votes: [...votes, ...votes, ...votes, ...votes],
+    }),
+    ballotCountsByVotingMethod: { [VotingMethod.Unknown]: ballots.length },
+  };
+
+  function disablePrinting() {
+    document
+      .getElementById('full-test-deck-tally-report')
+      ?.classList.add('no-print');
```
eventualbuddha, line 37: This is an effect, and so should be happening inside a `useEffect`. That said, this is more appropriately modeled using state.

frontends/election-manager/src/components/zero_report_print_button.tsx
eventualbuddha: I'd rather add the appropriate behavior to `PrintableArea` rather than spreading the responsibility to other components like this and `FullTestDeckTallyReport`. We don't need to drop down to raw DOM for this.

eventualbuddha (changes requested):
I added some comments before I really understood the whole set of changes in this PR. It seems like a primary goal is to be able to control whether a child `PrintableArea` component has the `no-print` class or not. This can be accomplished using normal props and state rather than this raw DOM add/remove via element IDs. I suggest we take another pass at it.

eventualbuddha:
I played around with this a bit on a branch. Here's where I landed with a "seems to work" `PrintableArea`:

```ts
import React, { useEffect, useRef } from 'react';
import ReactDom from 'react-dom';

interface Props {
  noPrint?: boolean;
  children?: React.ReactNode;
  'data-testid'?: string;
}

/**
 * <PrintableArea> renders the provided children in a standalone div with the .print-only class
 * name (standalone meaning via a React portal to ensure that a parent's CSS styles don't prevent
 * printability)
 */
export function PrintableArea({
  'data-testid': dataTestId,
  noPrint,
  children,
}: Props): JSX.Element {
  const printableContainerRef = useRef(document.createElement('div'));

  useEffect(() => {
    const printableContainer = printableContainerRef.current;
    if (dataTestId) {
      printableContainer.dataset.testid = dataTestId;
    }

    printableContainer.classList.add(
      'print-only',
      ...(noPrint ? ['no-print'] : [])
    );

    document.body.appendChild(printableContainer);
    return () => {
      document.body.removeChild(printableContainer);
    };
  }, [dataTestId, noPrint]);

  return ReactDom.createPortal(children, printableContainerRef.current);
}
```

adghayes:
@eventualbuddha I am not having success with this version of `PrintableArea`. If you are, can you push up your branch so I can see changes you made in the parent components?

When I try to print these documents, I just get blank pages. Watching logging in the console, it does appear that the print job is sent before the state has time to update and trickle down `PrintableArea`

Screen Shot 2022-05-25 at 10 13 18 AM

arsalansufi:
Update after in-person discussion: We're now no longer planning on using the `PrintButton` component. We'll orchestrate rendering and printing manually with effects to accomplish what we need to in a "React-friendly" way

adghayes:
@arsalansufi @eventualbuddha as I started digging into this, it seemed like it made more sense to alter and use `PrintButton` because `PrintButton` does a fair amount in terms of providing printing logic, a "Printing..." modal, and an error modal. With the latest changes, `PrintButton` can accept a JSX element that it is responsible for rendering and printing as an effect. I believe it works reliably because a) effects propagate up the React ancestor tree, so when `PrintButton` its effect is called, the children to be printed are already rendered and b) using a portal does not change this. 

I have manually tested all the print buttons in the app. Because the printed element is only rendered for a moment, automated tests had to be reordered to `waitFor` the printed component at the right time, but those are all working. Let me know if there needs to be additional test on `PrintButton`.

If this approach is acceptable, I'll clean up the git history before merging.

frontends/election-manager/src/components/print_button.tsx
```diff
@@ -66,7 +72,35 @@ export function PrintButton({
...
+  ]);
+
+  useEffect(() => {
+    async function printOnTargetRendered() {
+      await print();
+    }
+
+    if (renderPrintTarget) {
+      void printOnTargetRendered();
+    }
```
arsalansufi, line 95: Curious: Why did you have to wrap `print` in `printOnTargetRendered`? I imagine there's a reason why you couldn't just write:

```ts
if (renderPrintTarget) {
  print();
}
```

arsalansufi, line 95: Curious: Why did you have to wrap `print` in `printOnTargetRendered`? I imagine there's a reason why you couldn't just write:

```ts
if (renderPrintTarget) {
  print();
}
```

frontends/election-manager/src/components/zero_report_print_button.tsx
```diff
@@ -54,27 +53,30 @@ export function ZeroReportPrintButton(): JSX.Element {
...
+
   return (
     <React.Fragment>
       <PrintButton
         afterPrint={logZeroReportPrintSuccess}
         afterPrintError={logZeroReportPrintError}
         sides="one-sided"
+        printTarget={zeroReport}
+        printTargetTestId="zero-report"
       >
```
arsalansufi, line 73: Really cool to see how you've combined `PrintButton` and `PrintableArea`, making the print interface even more intuitive!

arsalansufi, line 73: Really cool to see how you've combined `PrintButton` and `PrintableArea`, making the print interface even more intuitive!

frontends/election-manager/src/components/full_test_deck_tally_report_button.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { election } = electionDefinition;
+
+  const ballots = generateTestDeckBallots({ election });
+  const votes = ballots.map((b) => b.votes);
+  const testDeckTally: Tally = {
+    numberOfBallotsCounted: ballots.length * 4,
+    castVoteRecords: new Set(),
+    contestTallies: tallyVotesByContest({
+      election,
+      votes: [...votes, ...votes, ...votes, ...votes],
```
arsalansufi, line 28: > votes, Votes, VOTES, VOTES!!

How my mind reads this line 😆

Mind adding a comment here noting why we count 4 sets of votes, as you documented in the PR description?

> The totals in the report are 4 times the totals of a single test deck, because it is used to compare with results after scanning two precinct test decks at each precinct and two full test decks at the central scanner.

Also just curious if all counties do L&A in this way, e.g. what if a county doesn't plan on using precinct scanners or a central scanner?

arsalansufi, line 28: > votes, Votes, VOTES, VOTES!!

How my mind reads this line 😆

Mind adding a comment here noting why we count 4 sets of votes, as you documented in the PR description?

> The totals in the report are 4 times the totals of a single test deck, because it is used to compare with results after scanning two precinct test decks at each precinct and two full test decks at the central scanner.

Also just curious if all counties do L&A in this way, e.g. what if a county doesn't plan on using precinct scanners or a central scanner?

frontends/election-manager/src/components/print_button.test.tsx
```diff
@@ -71,3 +77,43 @@ test('if connected printers, show printing modal', async () => {
...
+
+    // Check that printTarget is still rendered after printing
+    await waitFor(async () => {
+      expect(printer.print).toBeCalled();
+      getByTestId('print-target');
+
+      // Check that printTarget is eventually removed
+      await waitForElementToBeRemoved(() => queryByTestId('print-target'));
+    });
+  });
```
arsalansufi, line 118: I _believe_ you shouldn't have to nest `waitFor`s here. And you can also replace some `waitFor`s with `findBy`s, which wait automatically

```ts
const renderedPrintTarget = await findByTestId('print-target');
domGetByText(renderedPrintTarget, 'Print Target');
expect(printer.print).not.toBeCalled();

await waitFor(() => expect(printer.print).toBeCalled());
getByTestId('print-target');
await waitForElementToBeRemoved(() => queryByTestId('print-target'));
```

Correct me if I'm missing something!

arsalansufi, line 118: I _believe_ you shouldn't have to nest `waitFor`s here. And you can also replace some `waitFor`s with `findBy`s, which wait automatically

```ts
const renderedPrintTarget = await findByTestId('print-target');
domGetByText(renderedPrintTarget, 'Print Target');
expect(printer.print).not.toBeCalled();

await waitFor(() => expect(printer.print).toBeCalled());
getByTestId('print-target');
await waitForElementToBeRemoved(() => queryByTestId('print-target'));
```

Correct me if I'm missing something!

arsalansufi (commented):
This new approach looks great to me! Though a bit complex, it feels "React-friendly." Left some small comments and questions

frontends/election-manager/src/components/print_button.tsx
adghayes: The function passed to `useEffect` [can't be asynchronous](https://devtrium.com/posts/async-functions-useeffect), it turns out, but you can call an asynchronous function within.  I noticed a couple other places in the codebase where the pattern was used, e.g. https://github.com/votingworks/vxsuite/blob/ed7aace62c0bce7afb9e097edd6a00cb325f20dd/frontends/bmd/src/app_root.tsx#L958-L968

frontends/election-manager/src/components/full_test_deck_tally_report_button.tsx
adghayes: All our counties with precinct scanners do L&A this way, but if it's a county with a central system only (we only have one customer like this now) then the totals will be off. Their total should be 2x rather than 4x. It seems like something we can adjust later alongside a number of other changes to make if someone doesn't have a precinct system.

adghayes: I'll add a comment in!

frontends/election-manager/src/components/print_button.test.tsx
adghayes: It wasn't working for me if I didn't nest the `waitFor`s. The `printTarget`'s div had already been removed from the page! Maybe I can remove some nesting with `findBy`'s though. I will try.

frontends/election-manager/src/components/print_button.tsx
```diff
@@ -119,6 +153,11 @@ export function PrintButton({
           actions={<Button onPress={donePrintingError}>Okay</Button>}
         />
       )}
+      {printTarget && renderPrintTarget && (
```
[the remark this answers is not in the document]
adghayes: Sounds good!

frontends/election-manager/src/components/print_button.tsx
adghayes: This has me wondering if I could just do `void print();`... I will see what the linter says.

frontends/election-manager/src/components/print_button.tsx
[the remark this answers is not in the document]
adghayes: Renamed.

frontends/election-manager/src/components/print_button.test.tsx
adghayes: Wasn't able to unnest the `waitFor` statements, although I refactored and commented a little bit. If you simply `find` the `div` then, on the next assertion, the `div` will already have been printed and its children unmounted. Nesting the `waitFor` enables us to catch the moment where all the conditions are true in sequence.

eventualbuddha: This worked for me:

```ts
  const printDeferred = deferred<void>();
  printer.print.mockReturnValueOnce(printDeferred.promise);
  fireEvent.click(getByText('Print Now'));

  // Check that printTarget is rendered before printing
  await waitFor(() => {
    const renderedPrintTarget = getByTestId('print-target');
    domGetByText(renderedPrintTarget, 'Print Target');
    expect(printer.print).not.toBeCalled();
  });

  // Check that printTarget is still rendered after printing
  await waitFor(() => {
    expect(printer.print).toBeCalled();
    getByText('Print Target');
  });

  // Allow `printer.print` to resolve
  printDeferred.resolve();

  // Check that printTarget is removed after printing
  await waitFor(() => {
    expect(queryByText('Print Target')).not.toBeInTheDocument();
  });
```

frontends/election-manager/src/components/print_button.tsx
```diff
@@ -66,7 +72,31 @@ export function PrintButton({
...
+    title,
+  ]);
+
+  useEffect(() => {
+    if (shouldRenderPrintTarget) {
+      void print();
+    }
+  }, [shouldRenderPrintTarget, print]);
+
+  async function startPrint() {
```
eventualbuddha, line 94: This should be wrapped in `useCallback`.

frontends/election-manager/src/components/printable_area.tsx
```diff
@@ -3,17 +3,21 @@ import ReactDom from 'react-dom';
 
 interface Props {
   children?: React.ReactNode;
+  dataTestId?: string;
```
eventualbuddha, line 6: You can allow the typical style for this by defining the prop like so:

```suggestion
  'data-testid'?: string;
```

Then you can do this:

```ts
export function PrintableArea({ 'data-testid': dataTestId, children }: Props): JSX.Element {
```

eventualbuddha (approved):
Thanks for addressing my earlier feedback. There's a couple more small things but otherwise LGTM.

frontends/election-manager/src/components/print_button.tsx
adghayes: You mean that `startPrint` should be wrapped in `useCallback`, since it has dependencies on props?

eventualbuddha: No, because it is passed to a JSX prop.

PR #1880, opened 2022-05-25
fix: make `useDevices` et al work again

eventualbuddha:
## Overview

They were broken by votingworks/kiosk-browser#120. The fix for that bug, votingworks/kiosk-browser#121, changed the API a bit: `kiosk.devices` is no longer an RxJS observable, but a custom simpler object with a similar but smaller API.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tested manually in Election Manager + kiosk-browser with votingworks/kiosk-browser#121.

## Checklist

### DIFF
--- a/frontends/bmd/package.json
+++ b/frontends/bmd/package.json
@@ -118,7 +118,6 @@
     "react-idle-timer": "4.2.12",
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.1",
-    "rxjs": "^6.6.6",
     "setimmediate": "^1.0.5",
     "styled-components": "^4.4.1",
     "typescript": "4.6.3",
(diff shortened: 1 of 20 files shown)


taravancil (approved):

eventualbuddha:
> is it necessary to repeat `observable.ts` in `utils` and `test-utils`?

It was the easiest way to deal with it. They both need a `BehaviorSubject` for different kiosk-browser mocking stuff (one for tests, one for demo). `utils` depends on `test-utils`, but only in `devDependencies` so `utils` can't use `test-utils` stuff outside tests and `test-utils` cannot depend on `utils` or it'd create a cycle.

The harder way to deal with it is to create another package that both of them use, but that seemed overkill for this. There are a handful of things that are getting duplicated because of this, though, like `assert`. After we move to vite we may have more flexibility here as we will no longer have build-time dependency cycles.

PR #1884, opened 2022-05-25
fix(ui): correct `setResultsReportingUrl` race condition

eventualbuddha:
## Overview

It's unlikely to happen, but it's possible that the `useEffect` that signs the VxQR results payload will set the results reporting URL to a stale value if two back-to-back renders have `kiosk.sign` calls that return out of order.

We were wrapping the `kiosk.sign` result `Promise` in `makeCancelable`, but that only prevents calling `setResultsReportingUrl` if the component _is unmounted entirely_. It does not stop an earlier render from clobbering the URL update from a re-render.

The fix is to track whether the current render is stale. `useEffect` will call the returned function both on unmount _and on re-render_. We use that to mark a render as stale and do not update the URL if it is stale by the time the `kiosk.sign` call resolves.

If this is something we find ourselves doing often it may make sense to create a hook or other utility to help consolidate the logic.

I discovered this while working on #1314 as the state update was triggering a test warning, though that warning doesn't seem to happen on `main`.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated testing. Specifically, I added a test that fails without the `isCurrentRender` mitigation:

```
 FAIL  src/precinct_scanner_tally_qrcode.test.tsx
  ✓ renders WITHOUT results reporting when there are CVRs but polls are open (13 ms)
  ✓ renders with results reporting when there are CVRs and polls are closed (99 ms)
  ✓ renders with results reporting when there are CVRs and polls are closed in testing mode (33 ms)
  ✓ renders with unsigned results reporting when there is no kiosk (28 ms)
  ✕ renders the correct signature on re-renders (47 ms)

  ● renders the correct signature on re-renders

    expect(received).toContain(expected) // indexOf

    Expected substring: "SECONDFAKESIGNATURE"
    Received string:    "https://results.voting.works/?p=b52e9f4728bb34e7ff483ee769439d61def51a0abdcd82c9328f10bf138896f8.DEMO-0000.1.1632049500.W1swLDAsMCwwLDAsMCwwLDAsMCwwXSxbMCwwLDAsMCwwLDAsMCwwLDAsMCwwXSxbMCwwLDAsMCwwLDAsMCwwLDBdLFswLDAsMCwwLDAsMCwwLDAsMCwwLDAsMCwwLDAsMCwwLDAsMCwwLDBdLFswLDAsMCwwLDAsMCwwLDAsMCwwLDAsMCwwXSxbMCwwLDAsMCwwLDBdLFswLDAsMCwwLDBdLFswLDAsMCwwLDAsMCwwXSxbMywwLDEsMSwwLDAsMCwwLDAsMCwwLDAsMCwwLDBdLFswLDAsMCwwLDBdLFswLDAsMCwwLDAsMF0sWzAsMCwwLDAsMCwwXSxbMCwwLDAsMCwwLDAsMCwwLDAsMF0sWzAsMCwwLDAsMF0sWzAsMCwwLDAsMF0sWzAsMCwwLDAsMF0sWzAsMCwwLDAsMF0sWzAsMCwwLDAsMF0sWzAsMCwwLDAsMF0sWzAsMCwwLDAsMF0sWzAsMCwwLDAsMF0sWzAsMCwwLDAsMF1d&s=FIRSTFAKESIGNATURE"

      246 |   // despite the fact that the first one was resolved later
      247 |   const { value } = screen.getByTestId('qrcode').dataset;
    > 248 |   expect(value).toContain('SECONDFAKESIGNATURE');
          |                 ^
      249 | });
      250 |

      at Object.<anonymous> (src/precinct_scanner_tally_qrcode.test.tsx:248:17)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 4 passed, 5 total
Snapshots:   0 total
Time:        3.329 s, estimated 4 s
Ran all test suites matching /src\/precinct_scanner_tally_qrcode\.test\.tsx/i.
```

## Checklist

### DIFF
--- a/libs/ui/src/precinct_scanner_tally_qrcode.test.tsx
+++ b/libs/ui/src/precinct_scanner_tally_qrcode.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, screen, waitFor } from '@testing-library/react';
+import { act, render, screen, waitFor } from '@testing-library/react';
 import {
   electionSample,
   electionSampleDefinition,
@@ -12,6 +12,7 @@ import {
 import {
   calculateTallyForCastVoteRecords,
   compressTally,
+  deferred,
 } from '@votingworks/utils';
 import { fakeKiosk, mockOf } from '@votingworks/test-utils';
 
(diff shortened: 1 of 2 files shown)


benadida (approved):

jonahkagan:
@eventualbuddha I think the [`useLock`](https://github.com/votingworks/vxsuite/blob/main/libs/ui/src/hooks/use_lock.ts) hook that I just created might be useful for situations like this.

eventualbuddha:
@jonahkagan I think the way that's being used in `PollWorkerScreen` actually would prevent the _subsequent_ update from rendering at all, rather than the preventing an out-of-order resolution to two `sign` calls from using the wrong one. We always want the most recent one, dropping any work from a previous render.

jonahkagan:
Ah yeah that makes sense. I wasn't reading closely enough. I'm surprised this issue hasn't happened elsewhere. Seems like something that could happen easily with any async request that's based on props that can change.

eventualbuddha:
It probably is happening elsewhere, but at a low enough frequency that we don't notice?

PR #1892, opened 2022-05-26
fix(lint): make linting work properly again

eventualbuddha:
## Overview

I'm not sure why, but linting has not been working 100% for a little while. It has been working to some degree because our prettier formatting is done through eslint, but some other rules (like `prefer-readonly`) were not being enforced. This PR fixes that, possibly by fixing the `vx/gts-spread-like-types` rule.

We've been seeing false positives with this rule somtimes. Turns out it's a difference in the TypeScript version from when we upgraded from 4.2.x to 4.6.x, but I didn't catch it during the upgrade because I forgot to upgrade the linter package's copy of TypeScript. This fixes it so it should work in both versions, though we only really care about 4.6.x.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated tests, navigating through various projects in VS Code to ensure the ESLint integration is working well.

## Checklist

### DIFF
--- a/frontends/election-manager/src/lib/converters/nh_converter_client.ts
+++ b/frontends/election-manager/src/lib/converters/nh_converter_client.ts
@@ -21,11 +21,11 @@ const VxElectionDefinitionFile: VxFile = {
 
 /* eslint-disable @typescript-eslint/require-await */
 export class NhConverterClient implements ConverterClient {
-  private inputFiles: Map<VxFile, File | undefined> = new Map([
+  private readonly inputFiles: Map<VxFile, File | undefined> = new Map([
     [NhCardDefinitionFile, undefined],
     [NhCardBallotFile, undefined],
   ]);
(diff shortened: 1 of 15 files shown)


benadida (commented):
This is a pretty big change given the addition of all the `useCallback`s right? Should we do some extensive manual testing before merging? Or is there a reason not to be too concerned?

eventualbuddha:
> This is a pretty big change given the addition of all the `useCallback`s right? Should we do some extensive manual testing before merging? Or is there a reason not to be too concerned?

For reasons I still don't understand, the `react/jsx-no-bind` errors stopped showing up after I rebased on `main` again. I do think we'll want to get to that in the future, but I removed the fixes for those for now, leaving only the (much safer) `prefer-readonly` fixes.

jonahkagan (approved):
Nice! The `gts-spread-like-types` rule failures were so confusing. Glad you looked into them.

Issue #1931, opened 2022-06-06
VxMark: admin card written by election manager can't program BMD


benadida:
It keeps saying that the election definition doesn't match. @jonahkagan thinks it has something to do with how we're computing the hashes.

mattroe:
@benadida are you still experiencing this issue? I haven't seen it all in my testing lately.

benadida:
This has been fixed by #1934

PR #1934, opened 2022-06-06
fix(election-manager): encode election data properly when writing admin card

jonahkagan:
## Overview
Task: #1931

When encoding admin cards, the browser seems to convert LF to CRLF line endings when making a multi-part form request. This causes election hash mismatches. To prevent this, we use x-www-url-encoded encoding in the POST submission.

## Demo Video or Screenshot
N/A

## Testing Plan 
Manually tested

## Checklist
- [ ] ~~I have added JSDoc comments to any newly introduced exports~~

### DIFF
--- a/frontends/election-manager/src/screens/smartcards_screen.tsx
+++ b/frontends/election-manager/src/screens/smartcards_screen.tsx
@@ -132,7 +132,6 @@ export function SmartcardsScreen(): JSX.Element {
       message: 'Programming an admin card...',
       programmedUserType: 'admin',
     });
-    const formData = new FormData();
     setIsProgrammingCard(true);
     setIsPromptingForAdminPasscode(false);
     const shortValue = JSON.stringify({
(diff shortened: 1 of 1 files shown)


eventualbuddha:
A possible alternate solution:

```diff
diff --git a/frontends/election-manager/src/screens/smartcards_screen.tsx b/frontends/election-manager/src/screens/smartcards_screen.tsx
index 2d2505be0..0e09a6758 100644
--- a/frontends/election-manager/src/screens/smartcards_screen.tsx
+++ b/frontends/election-manager/src/screens/smartcards_screen.tsx
@@ -141,7 +141,10 @@ export function SmartcardsScreen(): JSX.Element {
       p: passcode,
     });
     formData.append('short_value', shortValue);
-    formData.append('long_value', electionData);
+    formData.append(
+      'long_value',
+      new Blob([electionData], { type: 'application/octet-stream' })
+    );
     const response = await fetch('/card/write_short_and_long', {
       method: 'post',
       body: formData,
```

benadida:
Before we push forward, I'm wary of ... what caused this regression? Are we actually writing the wrong data to the card?

Could this be... a kiosk-browser / electron / chromium behavior change?

benadida:
I'm still confused as to what caused this regression. In any case, I agree with @jonahkagan's diagnosis, and have a bit more data as to what's going on: `FormData` uses `multipart/form-data` as an encoding type, and that just sends the raw text, but somewhere in the HTTP pipeline, new lines are rewritten. If a different encoding type were used, specifically ,`x-www-form-url-encoded`, then this shouldn't happen. I imagine this behavior changed, probably to be more up to HTTP spec, with our Electron upgrade, but I'm not sure.

I'm a little wary of @jonahkagan's proposed solution, because if later the card data legitimately includes `\r\n` then we will be in trouble.

I think @eventualbuddha's approach of changing the way the form is submitted so that the content bytes are truly respected is the better path. I don't know for sure if that code works, @eventualbuddha, but assuming it does at preserving the exact bytes, let's do that.

benadida:
taking over this PR as per discussion in standup today. Using @eventualbuddha's solution and testing it.

benadida:
Turns out @eventualbuddha's solution doesn't make flask happy... not sure if we'd have to parse the form differently.

Instead, I am manually putting together a url-encoded POST body.

eventualbuddha (approved):
I'm guessing the octet-stream thing didn't work?

benadida:
> I'm guessing the octet-stream thing didn't work?

Not without modifying the server side, which felt a lot riskier.

PR #1954, opened 2022-06-10
fix(scan): make some improvements to NH definition conversion

eventualbuddha:
## Overview

Introduces the Amherst 7/12/22 election definition and begins work to convert it properly. Pre-work for #1952 and #1953.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated testing.

## Checklist


libs/ballot-interpreter-nh/src/convert.ts
```diff
@@ -168,72 +217,198 @@ export function readGridFromElectionDefinition(
...
+  MissingBallotSize = 'MissingBallotSize',
+  MissingCandidateName = 'MissingCandidateName',
+  MissingElectionDate = 'MissingElectionDate',
+  MissingElectionId = 'MissingElectionId',
+  MissingElectionName = 'MissingElectionName',
+  MissingOfficeName = 'MissingOfficeName',
+  MissingPrecinctId = 'MissingPrecinctId',
+  MissingTimingMarkMetadata = 'MissingTimingMarkMetadata',
+  MissingTownId = 'MissingTownId',
+  MissingTownName = 'MissingTownName',
```
benadida, line 240: wondering if we'll find this level of granularity in types useful, or onerous in the long run.

consider maybe `MissingRequiredData` and then a detailed error message that says what's missing.

Not a blocker, ok with the current approach, only a suggestion for consideration.

benadida (approved):

libs/ballot-interpreter-nh/src/convert.ts
eventualbuddha: Yep, that's a good point. I consolidated most of those down to one error type.

PR #1962, opened 2022-06-15
docs(ui): Add documentation for Prose component

beausmith:
Related: https://github.com/votingworks/vxsuite-roadmap/issues/44

### DIFF
--- a/libs/ui/src/prose.tsx
+++ b/libs/ui/src/prose.tsx
@@ -1,6 +1,22 @@
 import styled from 'styled-components';
 import { Theme } from './themes';
 
+// Prose!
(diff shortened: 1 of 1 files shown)


libs/ui/src/prose.tsx
```diff
@@ -1,6 +1,22 @@
 import styled from 'styled-components';
 import { Theme } from './themes';
 
+// Prose!
```
carolinemodic, line 4: Prose ultimately is a <p> element right? Can we explain that and how you shouldn't nest prose's inside of one another?

carolinemodic (changes requested):

carolinemodic (approved):

PR #1965, opened 2022-06-15
fix(types): remove deprecated election properties

eventualbuddha:
## Overview

We've been meaning to rename `sealURL` to `sealUrl` to comply with our naming rules. We also replaced `adjudicationReasons` with `precinctScanAdjudicationReasons` and `centralScanAdjudicationReasons` a while back. This commit removes both `sealURL` and `adjudicationReasons` while still allowing those values to appear in an election definition. This allows loading older election definitions while still presenting them with the current interface.

## Demo Video or Screenshot
n/a

## Testing Plan 
Manually tried loading elections in Election Manager and BMD.

## Checklist


services/scan/src/workers/json_serialization.ts
```diff
@@ -205,10 +205,11 @@ export function deserialize(
 
         default:
           throwIllegalValue(data, '__dataType__');
+          break;
       }
+      break;
     }
 
-    // eslint-disable-next-line no-fallthrough
```
eventualbuddha, line 211: This change was kinda unrelated. @jonahkagan changed it recently so TS will not delete unreachable code, so the `break;` statements can be added above to make ESLint happy. TS knows they're unreachable, but ESLint doesn't.

benadida (approved):

PR #1993, opened 2022-06-21
Add "Remove Card" screen to VxAdmin & VxCentralScan

### LINKED ISSUE #1910: Implement success screen that instructs admins to remove their card after successful auth on VxAdmin and VxCentralScan

adghayes:
## Overview
Closes #1910. Adds a screen to prompt users to remove their card after authentication.

## Demo Video or Screenshot

Screen Shot 2022-06-19 at 8 15 02 PM

## Testing Plan
- [x] Snapshot test for `RemoveCardPage` component in `libs/ui`
- [x] Updated "authentication works" tests in VxAdmin & VxCentralScan
- [x] Updated authentication helpers in VxAdmin, VxCentralScan, and `integration-testing/bsd` to include removing card

## Checklist


adghayes:
@arsalansufi some notes about the implementation:
- Icon assets were placed in the frontends rather than in `libs/ui` because I found that images in `libs/ui` actually are not set up to be served out of the frontends. 
- The `RemoveCardPage` component was created in `libs/ui`. I didn't make the whole screen there, just the design described in #1910.  Although we could pass in props to have the whole screen in `libs/ui`, it seemed like better encapsulation to keep the ElectionInfoBar in the frontend where it is used across screens. In a later refactor, after the new auth flows are in place, we could improve the way `ElectionInfoBar` is used repetitively (and at times inconsistently) across the apps. 
- I am not keeping `libs/ui/index.ts` alphabetical. When I tried to do so initially, it generated some really weird test failures. The re-ordering of `styled` components leads to the dynamically generated class names being different, so snapshots were failing across apps (in this case, wherever there was a `Screen` component).

libs/ui/src/remove_card_page.tsx
```diff
@@ -0,0 +1,24 @@
...
+  height: 20vw;
+`;
+
+export function RemoveCardPage(): JSX.Element {
+  return (
+    <Main padded centerChild>
+      <RemoveCardImage src="remove-card.svg" alt="Remove Card Icon" />
+      <Prose textCenter theme={fontSizeTheme.medium} maxWidth={false}>
+        <h1>Successfully Authenticated.</h1>
+        <p>Remove Card.</p>
```
arsalansufi, line 20: To match the text styling conventions used elsewhere, e.g.

styling

let's remove the period after the "Successfully Authenticated" title and sentence case "Remove card."

libs/ui/src/remove_card_page.tsx
```diff
@@ -0,0 +1,24 @@
...
+  margin-right: auto;
+  margin-bottom: 5px;
+  margin-left: auto;
+  height: 20vw;
+`;
+
+export function RemoveCardPage(): JSX.Element {
+  return (
+    <Main padded centerChild>
+      <RemoveCardImage src="remove-card.svg" alt="Remove Card Icon" />
```
arsalansufi, line 17: I know we haven't been consistent about this in the frontends' `public/` directories, but I see we've begun trying to group public items. Let's accordingly move these images to [`public/assets/`](https://github.com/votingworks/vxsuite/tree/main/frontends/election-manager/public/assets)

frontends/bsd/src/app_root.tsx
```diff
@@ -238,6 +240,14 @@ export function AppRoot({
     lastScannedSheetIdx,
   ]);
 
+  useEffect(() => {
+    if (currentUserSession?.authenticated && smartcard.status === 'no_card') {
+      setRemoveCardScreenShown(true);
```
arsalansufi, line 245: Interesting - this logic feels flipped to me, since it seems that we should show the card-removal prompt when you're authenticated and there _is_ a card plugged in (not when you're authenticated and there isn't a card plugged in)

I'm likely missing something!

frontends/bsd/src/app_root.tsx
```diff
@@ -566,6 +576,20 @@ export function AppRoot({
     );
   }
 
+  if (smartcard.status !== 'no_card' && !removeCardScreenShown) {
```
arsalansufi, line 579: Curious why we have to check `smartcard.status` here when `smartcard.status` is in a sense captured in `removeCardScreenShown`. Also curious why we display the card-removal card screen when `removeCardScreenShown` is false - I'd expect the opposite

arsalansufi (commented):
Left some questions as I'm trying to understand the logic for when to display the screen! And am relatedly wondering if there may be some way to simplify 🤔

libs/ui/src/remove_card_page.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Main } from './main';
+import { Prose } from './prose';
+import { fontSizeTheme } from './themes';
+
+const RemoveCardImage = styled.img`
+  margin-right: auto;
+  margin-bottom: 5px;
```
arsalansufi, line 9: Design nit: I've always felt that the lock icon on the lock screen is a bit too close to the "[Machine Name] is Locked" text. I see you're just copying [that styling](https://github.com/votingworks/vxsuite/blob/ff31293e85377fb79e5cc760b40b522bc31e9da6/frontends/election-manager/src/screens/machine_locked_screen.tsx#L14) here, but what do you think of increasing that spacing a bit, both here and elsewhere?

arsalansufi:
> I am not keeping libs/ui/index.ts alphabetical. When I tried to do so initially, it generated some really weird test failures. The re-ordering of styled components leads to the dynamically generated class names being different, so snapshots were failing across apps (in this case, wherever there was a Screen component).

Ah yes, this is a known issue (with an unknown explanation): https://github.com/votingworks/vxsuite/issues/1617

frontends/bsd/src/app_root.tsx
adghayes: I think this is a result of poor variable naming. By `removeCardScreenShown` I meant "the remove card screen has already been shown," not "the remove card screen is shown." I will rename to something clearer.

frontends/bsd/src/app_root.tsx
adghayes: Again, there is a poor variable naming issue here.

libs/ui/src/remove_card_page.tsx
adghayes: I played with it a bit when I made the component but didn't notice a difference / care too much, so sided with consistency. I would be pro- changing it in both places, because I also felt like it was slightly cramped.

adghayes:
@arsalansufi one thing to call out about the state here - we want to show the card removal screen only after initial auth. There will be cases where we need to insert an admin card during an authenticated session and _not_ show the card removal screen, like when we're programming an admin card. Otherwise, the state could be as simple as 'no_card' = show screen. Thus, we have to track if we've shown the screen already or not. Which is why there is a (poorly named) state variable and an effect hook.

libs/ui/src/remove_card_page.tsx
adghayes: Let's save the change the lock screen for when we refactor those screens generally, and I'll increase the space on the remove card screen now.

adghayes:
@beausmith @arsalansufi the way the top of the hand does not have white fill in this icon looks a little strange to me. Is it something you want to fix?

Screen Shot 2022-06-21 at 11 36 12 PM

libs/ui/src/remove_card_page.tsx
arsalansufi: Sounds great!

frontends/bsd/src/app_root.tsx
arsalansufi: Ah paired with your comment below (https://github.com/votingworks/vxsuite/pull/1993#issuecomment-1162528287), I understand now!

frontends/bsd/src/app_root.tsx
```diff
@@ -566,6 +576,20 @@ export function AppRoot({
     );
   }
 
+  if (smartcard.status !== 'no_card' && !cardRemovedAfterAuth) {
```
arsalansufi, line 579: The new `cardRemovedAfterAuth` variable name has made everything much clearer - nice!

arsalansufi (approved):
LGTM!

arsalansufi:
> beausmith arsalansufi the way the top of the hand does not have white fill in this icon looks a little strange to me. Is it something you want to change?

Nice catch! I agree that the white fill looks a bit strange. Since Beau may not be easy to reach right now, we can always just open an issue with the "needs_design" label for this

PR #1996, opened 2022-06-21
first pass at livecheck prototype

benadida:
The livecheck QR code in the poll worker screen, behind a feature flag.


frontends/precinct-scanner/src/screens/poll_worker_screen.test.tsx
```diff
@@ -147,3 +124,43 @@ describe('shows Export Results button only when polls are closed and more than 0
...
 });
+
+describe('shows Livecheck button only when enabled', () => {
+  test('enable livecheck', async () => {
+    // eslint-disable-next-line @typescript-eslint/require-await
+    await act(async () => {
+      renderScreen({
+        scannedBallotCount: 5,
+        isPollsOpen: true,
+        enableLivecheck: true,
```
carolinemodic, line 135: are we never testing the default behavior of calling isLiveCheckEnabled?

benadida: good callout. I thought about this some more and decided that adding `enableLiveCheck` as a prop to the poll worker screen kind of overcomplicates the feature flag approach, so I took that out and instead mocked the `isLiveCheckEnabled` call.

frontends/precinct-scanner/src/config/features.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ *     REACT_APP_VX_ENABLE_LIVECHECK=true
+ *
+ * @see https://create-react-app.dev/docs/adding-custom-environment-variables/
+ */
+export function isLiveCheckEnabled(): boolean {
+  return (
+    (process.env.NODE_ENV === 'development' ||
+      asBoolean(process.env.REACT_APP_VX_DEV)) &&
+    asBoolean(process.env['REACT_APP_VX_ENABLE_LIVECHECK'])
```
eventualbuddha, line 18: This should be added to `env.d.ts` so you can use autocomplete and dot syntax.

Issue #1987, opened 2022-06-16
ignore NH XML ballot size


eventualbuddha:
Apparently the XML may say the ballot is 8.5X14, even if the actual paper is 8.5X11. This is because some tool they're using automatically makes it 8.5X14 if the ballot is long enough, but then someone manually goes and makes it fit in 8.5X11 again. ¯\_(ツ)_/¯

PR #1999, opened 2022-06-21
fix(scan): use the paper size of the NH PDF

eventualbuddha:
## Overview

We already didn't fail on a mismatch, but now we'll use the PDF image size rather than the XML ballot size since the XML is apparently sometimes wrong.

Refs #1987 

## Demo Video or Screenshot
n/a

## Testing Plan 
This was blocking the ability to scan the Amherst ballot. It still doesn't work, but for different reasons.

## Checklist

### DIFF
--- a/libs/ballot-interpreter-nh/test/fixtures/amherst-2022-07-12/election.json
+++ b/libs/ballot-interpreter-nh/test/fixtures/amherst-2022-07-12/election.json
@@ -1,6 +1,6 @@
 {
   "ballotLayout": {
-    "paperSize": "legal",
+    "paperSize": "letter",
     "targetMarkPosition": "right"
   },
   "ballotStyles": [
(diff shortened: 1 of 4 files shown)


libs/ballot-interpreter-nh/test/fixtures/amherst-2022-07-12/election.json
```diff
@@ -1,6 +1,6 @@
 {
   "ballotLayout": {
-    "paperSize": "legal",
+    "paperSize": "letter",
```
benadida, line 3: I'm a little confused as to why this became `letter`. Assuming there's a clear explanation and I'm just not seeing it, approved :)

eventualbuddha: Because the actual PDF is letter size but the XML says legal size, and we were incorrectly using legal size. The PR fixes that issue and updates this fixture.

Issue #2007, opened 2022-06-23
vxmark: "Tally Report on Card" modal reappears immediately after printing tally report


benadida:
To make it go away, you have to remove and reinsert the poll worker card.

It seems like it is immediately coming back even before we've had a chance to erase the tally from the card.

PR #2015, opened 2022-06-23
fix(bmd): Don't load tally from pollworker card if already loaded

jonahkagan:
## Overview
Task: #2007 

There was a loop happening because the pollworker screen would load the
tally once, then print it, then load it again before clearing it from
the card (due to re-renders triggering the useEffect hook multiple
times).

To fix, we put in a simple guard to make sure we only load the data if
we haven't already loaded it.

## Testing Plan 
Manually tested

## Checklist


benadida (approved):

benadida:
YES, thank you for fixing this so quickly.

PR #1925, opened 2022-06-02
refactor(bmd): Use useSmartcardAuth for voters

jonahkagan:
## Overview
Task: #1682 

_I suggest commit-by-commit reviewing_

Continues the refactor of BMD to use the useSmartcardAuth hook for voters. There are two types of voter auth:
- Voter cards
- Cardless voters, which are activated by pollworkers

Also in this PR, I removed local storage of voter sessions in development, since it would have been pretty complicated to refactor into this new approach, and it's not a crucial feature.

I generally tried to keep the changes as minimal as possible, but I do think there's still a lot of possibility to make app_root.tsx clearer now that a lot of the complexity has been factored out.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Existing automated tests
- New unit tests for useSmartcardAuth

## Checklist
- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.
- [ ] I have added JSDoc comments to any newly introduced exports

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #2021, opened 2022-06-27
refactor(bmd): Use useSmartcardAuth for voters (2nd attempt)

### LINKED ISSUE #2008: vxmark: BMD prints many copies of the ballot
We have a regression, where the ballot appears to be printing non-stop.

jonahkagan:
## Overview
Reopens #1925, which got reverted, and fixes #2008 

## Demo Video or Screenshot
N/A

## Testing Plan 
- Manually tested printing flow

## Checklist


frontends/bmd/src/pages/print_page.tsx
```diff
@@ -46,8 +47,11 @@ export function PrintPage(): JSX.Element {
     'precinctId is required to render PrintPage'
   );
   const printerTimer = useRef(0);
+  const printLock = useLock();
 
   const printBallot = useCallback(async () => {
+    /* istanbul ignore if */
+    if (!printLock.lock()) return;
```
arsalansufi, line 54: Curious why this lock wasn't needed when using the hold smartcard/auth hooks - were we seeing fewer re-renders with the prior hooks?

arsalansufi (approved):
Updates LGTM!

frontends/bmd/src/pages/print_page.tsx
jonahkagan: Great question. The entire app re-renders every 3 seconds when the battery status is refreshed. That's been the case for a while. However, these re-renders weren't getting propagated down to this particular component (PrintPage) because they previously didn't update the BallotContext that this component accesses. This is because in the old code, all of the state and data put in the context was aggressively memoized to prevent it from changing on re-renders. 

The new useSmartcardAuth hook intentionally _does not_ memoize its result, because memoization is supposed to only be used to improve performance, not to ensure correctness. A component should be able to be re-rendered at any time without causing errant side-effects.

I'm hoping that this can help us flush out all of the places we're relying on memoization for correctness. My hunch is that we don't actually need to memoize much - we don't do much expensive rendering (besides ballots).

PR #2030, opened 2022-06-28
doc(scan): Add build instruction to README

jonahkagan:
## Overview
I found that when I tried to run `pnpm dev` in `services/scan` without first building the scan service, I would get an error about the plustek module not being built.

### DIFF
--- a/services/scan/README.md
+++ b/services/scan/README.md
@@ -11,6 +11,7 @@ then run the service like so:
 
 ```sh
 # in services/scan
+pnpm build:watch
 pnpm dev
 ```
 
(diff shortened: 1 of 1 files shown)


services/scan/README.md
```diff
@@ -11,6 +11,7 @@ then run the service like so:
 
 ```sh
 # in services/scan
+pnpm build:watch
```
eventualbuddha, line 14: Maybe background it if you're running it like this?

```suggestion
pnpm build:watch &
```

eventualbuddha (approved):

PR #2031, opened 2022-06-28
refactor(ps): Use useInsertedSmartcardAuth for authentication

jonahkagan:
## Overview
Task: #1682 

Refactors precinct-scanner to use the useInsertedSmartcardAuth hook.

The precinct scanner previously had passcode checking for admin cards, so I added that to the useInsertedSmartcardAuth hook. Notably, it allowed for admin cards that do not have a passcode set to skip this check. I did not implement this functionality (effectively removing it).

Adding passcode checking to the hook breaks BMD, since BMD doesn't have the requisite screens yet. I will add those in a separate PR.
 

## Demo Video or Screenshot
N/A

## Testing Plan 
- Existing automated tests
- New unit tests for the hook
- Some manual testing

## Checklist


libs/ui/src/hooks/smartcard_auth/use_inserted_smartcard_auth.ts
```diff
@@ -140,7 +149,15 @@ function smartcardAuthReducer(allowedUserRoles: UserRole[], scope: AuthScope) {
                 scope
               );
               if (userResult.isOk()) {
-                return { status: 'logged_in', user: userResult.ok() };
+                const user = userResult.ok();
+                if (previousState.auth.status === 'logged_out') {
+                  // TODO either need to make this configurable or add passcode screen to BMD
```
arsalansufi, line 154: Can probably remove this TODO since you've prepared a PR (https://github.com/votingworks/vxsuite/pull/2032) for the latter (which I think was the right call since it's what we want from a product perspective)!

frontends/precinct-scanner/src/screens/unlock_admin_screen.test.tsx
```diff
@@ -45,21 +40,15 @@ test('authentication', async () => {
...
+    expect(auth.checkPasscode).toHaveBeenNthCalledWith(1, '012345')
   );
+  screen.getByText('- - - - - -');
+});
 
+test('If passcode is incorrect, error message is shown', () => {
+  const auth = Inserted.fakeCheckingPasscodeAuth({
+    wrongPasscodeEntered: true,
+  });
+  render(<UnlockAdminScreen auth={auth} />);
```
arsalansufi, line 52: Any reason why you chose to test this case by mocking an auth already in the wrong-passcode state as opposed to actually entering an invalid code?

PR #2036, opened 2022-06-30
feat(hooks): Add logging to useInsertedSmartcardAuth and useDippedSmartcardAuth

jonahkagan:
## Overview
Task: #1682 

Adds auth logging to the useInsertedSmartcardAuth hook (which means adding logging to BMD and precinct-scanner) and useDippedSmartcardAuth (which means adding logging to election-manager and bsd). I based this on the logging already in place for useUserSession, but simplified the log events a bit to fit the new hooks.

_Review by commits._

## Demo Video or Screenshot
N/A

## Testing Plan 
Updated hook unit tests and frontend unit tests

## Checklist


jonahkagan:
@carolinemodic I'd love an initial review before I go further and also add logging to useDippedSmartcardAuth (which should will fix all of the broken tests). I'd like to get your feedback on the logging approach.

libs/logging/src/log_documentation.test.ts
```diff
@@ -28,7 +28,7 @@ describe('test cdf documentation generation', () => {
     expect(structuredData.DeviceModel).toBe('VxAdmin 1.0');
     expect(structuredData.GeneratedDate).toBe('2020-07-24T00:00:00.000Z');
     expect(structuredData.EventTypeDescription).toHaveLength(5);
-    expect(structuredData.EventIdDescription).toHaveLength(55);
+    expect(structuredData.EventIdDescription).toHaveLength(53);
```
carolinemodic, line 31: I shouldn't have made the tests like this lol

libs/ui/src/hooks/smartcard_auth/use_inserted_smartcard_auth.ts
```diff
@@ -379,3 +383,117 @@ export function useInsertedSmartcardAuth({
...
+        }
+      }
+      return;
+    }
+
+    case 'logged_in': {
+      if (auth.status === 'logged_out') {
+        await logger.log(LogEventId.AuthLogout, previousAuth.user.role, {
+          disposition:
+            auth.reason === 'no_card'
```
carolinemodic, line 453: what are the reasons where this ends in a failure?

carolinemodic (commented):
mostly makes sense to me one question

libs/ui/src/hooks/smartcard_auth/use_inserted_smartcard_auth.ts
jonahkagan: Generally, cases where the user is logged in and then something happens that forces them to get logged out. I think currently the only case is when a voter card expires.

jonahkagan: Do you think that `failure` is the right disposition for that case? Still wrapping my head around things

frontends/bmd/src/pages/unlock_admin_screen.test.tsx
```diff
@@ -46,7 +46,7 @@ test('authentication', async () => {
 
 test('If passcode is incorrect, error message is shown', () => {
   const auth = Inserted.fakeCheckingPasscodeAuth({
-    wrongPasscodeEnteredAt: true,
+    wrongPasscodeEnteredAt: new Date(),
```
jonahkagan, line 49: should have been in previous commit, whoops

libs/logging/VotingWorksLoggingDocumentation.md
```diff
@@ -42,25 +42,17 @@ IDs are logged with each log to identify the log being written.
...
-**Description:** Admin smartcard inserted, the user will be prompted for passcode to complete authentication.  
-**Machines:** All
-### user-session-activation
-**Type:** [user-action](#user-action)  
-**Description:** A user attempted to authenticate as a new user role, disposition and message clarify the user roles and success/failure.  
-**Machines:** All
-### user-logged-out
-**Type:** [user-action](#user-action)  
-**Description:** User logged out of the current session.  
+**Description:** A user attempted to log in. Disposition is success if they logged in, failure if not. An optional reason may be provided.
```
arsalansufi, line 55: Nice! The new event types feel much more intuitive to me

libs/ui/src/hooks/smartcard_auth/use_inserted_smartcard_auth.ts
```diff
@@ -386,3 +382,113 @@ export function useInsertedSmartcardAuth({
...
+
+    case 'logged_in': {
+      if (auth.status === 'logged_out') {
+        await logger.log(LogEventId.AuthLogout, previousAuth.user.role, {
+          disposition: LogDispositionStandardTypes.Success,
+          message: `User logged out: ${auth.reason}.`,
+          reason: auth.reason,
+        });
+      } else if (auth.status === 'logged_in') {
+        if (auth.user.role !== previousAuth.user.role) {
```
arsalansufi, line 454: Is this capturing the cardless voter session case? (It's the only case I can think of where you go straight from being logged in as one user type to another user type)

libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.ts
```diff
@@ -79,16 +76,24 @@ function smartcardAuthReducer(
...
-                const userResult = validateCardUser(action.card);
-                if (userResult.isOk()) {
-                  const user = userResult.ok();
+                const user = parseUserFromCard(action.card);
+                const validationResult = validateCardUser(user);
+                if (validationResult.isOk()) {
+                  assert(
+                    user &&
+                      (user.role === 'superadmin' || user.role === 'admin')
+                  );
```
arsalansufi, line 79: Is this assertion here because we only ever expect to use dipped smartcard auth for super admins and admins?

libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.test.ts
```diff
@@ -222,26 +311,50 @@ describe('useDippedSmartcardAuth', () => {
...
+
     cardApi.insertCard('invalid card data');
     jest.advanceTimersByTime(CARD_POLLING_INTERVAL);
     await waitForNextUpdate();
     expect(result.current).toMatchObject({
       status: 'logged_out',
       reason: 'invalid_user_on_card',
     });
+
+    expect(logger.log).toHaveBeenCalledTimes(2);
```
arsalansufi, line 325: Is the first log call here unrelated to auth and unimportant to this test?

libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.ts
```diff
@@ -256,3 +241,111 @@ export function useDippedSmartcardAuth({
...
+    status: 'logged_out',
+    reason: 'machine_locked',
+    bootstrapAuthenticatedAdminSession: /* istanbul ignore next */ () =>
+      undefined,
+  },
+  auth: DippedSmartcardAuth.Auth
+) {
+  switch (previousAuth.status) {
+    case 'logged_out': {
+      if (auth.status === 'logged_in') {
```
arsalansufi, line 257: With dipped smartcard auth, is the transition from `logged_out` straight to `logged_in` possible? Or will you always go from `remove_card` to `logged_in` for successful login? Just wondering whether this case is really necessary

arsalansufi (approved):
LGTM!

libs/ui/src/hooks/smartcard_auth/use_inserted_smartcard_auth.ts
jonahkagan: Yes, you are correct!

libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.test.ts
jonahkagan: No, it would be the same log line, since this test case tests two different card values that cause the same error. I think I just left out the second log call because it wasn't testing any different code path.

libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.ts
jonahkagan: Good question! Had me stumped for a second there. 100% code coverage guarantees that this case must be used for something...

I tried commenting it out and it looks like the test for bootstrapping admin auth fails, so I think that's the only case (which means once you remove bootstrapping, this case can indeed be removed)

libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.test.ts
arsalansufi: Ah yes I see now!

libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.ts
arsalansufi: Ah yes, that makes sense! Thanks for looking into that!

PR #2037, opened 2022-06-30
fix(scan-nh): update timing mark detection to handle partially-cropped ballot images

### LINKED ISSUE #2035: update timing mark detection to handle partially-cropped ballot images

eventualbuddha:
## Overview
Closes #2035 

The old approach of finding timing marks by scoring blocks has a major downside: if the blocks are sufficiently cropped they will fall below the threshold and no longer appear as timing marks. This happens regularly when using the precinct scanner. Reducing the threshold introduces too much noise. Though it's possible we could have inferred the missing ones on one side by looking at the other, there's no guarantee we'll see the ones on that side properly either.

Instead, we now look for the gaps between timing marks and determine the lines of blocks based on that. This is quite reliable on the left and right, but less so on the top and bottom. I haven't done extensive testing yet and I'm sure we'll need to iterate, but this seems both faster and less sensitive to distortions.

## Demo Video or Screenshot
| Step 1: Left Gaps | Step 2: Right Gaps | Step 3: Top Gaps | Step 4: Marks & Grid |
|-|-|-|-|
|01-front-left side|02-front-right side|03-front-top side|08-front|

## Testing Plan 
Tested with a few scans of the Hudson & Amherst ballots. More manual testing will follow, but this PR shouldn't block on that. Try it yourself like so:

```
❯ ./bin/interpret test/fixtures/amherst-2022-07-12/{election.json,scan-unmarked-{front,back}.jpeg}                          
test/fixtures/amherst-2022-07-12/scan-unmarked-front.jpeg:
Governor
⬜️ ( 2.86%) Josiah Bartlett
⬜️ ( 1.74%) Hannah Dustin
⬜️ ( 2.73%) John Spencer
⬜️ ( 2.24%) Write-In #1

United States Senator
⬜️ ( 2.24%) John Langdon
⬜️ ( 2.86%) William Preston
⬜️ ( 2.11%) Write-In #1

…
```

## Checklist


libs/ballot-interpreter-nh/src/accuvote.ts
```diff
@@ -102,7 +47,7 @@ export const TemplateBallotCardGeometry8pt5x11: BallotCardGeometry = {
   }),
   ovalSize: { width: 15, height: 10 },
   /* Converted from the documented size in inches: 3/16" x 1/16" */
-  timingMarkSize: { width: 37.5, height: 12.5 },
+  timingMarkSize: { width: 13.5, height: 4.5 },
```
eventualbuddha, line 50: This was a copy-paste error that is now fixed.

PR #2042, opened 2022-06-30
fix(frontends): do not assume we're at the root

eventualbuddha:
## Overview

A bunch of asset URLs were set relative to the current URL rather than the root of the domain. This means we'd end up with a broken image if you reload the page when at a path other than `/`.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/1938/176764545-adc9746c-0f3c-4474-a897-592b75304a73.mov

## Testing Plan
Tested as above in the video. 

## Checklist

### DIFF
--- a/frontends/bsd/src/components/election_configuration.tsx
+++ b/frontends/bsd/src/components/election_configuration.tsx
@@ -223,7 +223,7 @@ export function ElectionConfiguration({
           <Main padded>
             <Prose>
               <h1>No Election Ballot Package Files Found</h1>
-              <Image src="assets/usb-drive.svg" alt="Insert USB Image" />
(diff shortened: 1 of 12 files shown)


frontends/election-manager/src/screens/ballot_screen.tsx
```diff
@@ -338,7 +338,7 @@ export function BallotScreen(): JSX.Element {
         </Prose>
         <BallotPreviewHeader>
           <h4>Front Pages</h4>
-          <h4>Back Pages</h4>
+          <h4>BACK Pages</h4>
```
jonahkagan, line 341: was this intentional?

eventualbuddha: Hah! No.

benadida (approved):

PR #2049, opened 2022-06-30
migrate cvrs to admin service from local storage

carolinemodic:
This is obviously written in a rush pretty grossly. I didn't think that deeply about a lot of choices here, and tests are massively broken, but this seems to work from manual testing. I also commented out the OvervoteCombinationScreen as we don't use it and I didn't feel like making it work with this. 

Note this PR is just to merge to the nh-demo branch.


benadida (approved):
good for branch :)

PR #2058, opened 2022-07-01
update logic for finding layout info for contest

carolinemodic:
I think this does what @eventualbuddha described to find the right contest layout bounds, but I'm not completely sure. It seems to work with our ballots on very limited testing but definitely doesn't work right on the NH ballots.


carolinemodic:
merging to nh-demo

Issue #893, opened 2021-09-17
Abstract locked/unlocked screens into shared components/hooks


carolinemodic:
Maybe also a shared custom hook for userCurrentUserSession?

adghayes:
Jotting down some eventual implementation thoughts from working on #1910... The Locked, Invalid Card, Keypad, Remove Card, and if desirable, also the Setup Card Reader page could all be abstracted individually and then into a single, full auth flow component in `libs/ui`. Would help remove 3 screens from each app, remove duplication of things like `ElectionInfoBar` in so many places, and deduplicate auth flow testing.

carolinemodic:
This is done

PR #2079, opened 2022-07-05
restyle + refactor auth modals

### LINKED ISSUE #2022: Usage of white and transparent color in card assets looks a bit off
See https://github.com/votingworks/vxsuite/pull/1993#issuecomment-1162739340

adghayes:
## Overview
Closes #2022. Implements @beausmith's UI principle that purely action oriented screens (setup card reader, keypad, invalid card, remove card) are full screen modals and should all a) have a white background and b) have no election footer. 

Also does some work toward #893, abstracting the auth screens into shared UI components. When changing the screens, instead of changing it in both `bsd` and `election-manager`, I refactored to get the components from `libs/ui`

## Demo Video or Screenshot

Before:
[auth-ui-before.webm](https://user-images.githubusercontent.com/37960853/177234601-034e6710-89c2-408e-9e3a-fb53a1c2accc.webm)

After:
[auth-ui-tweak.webm](https://user-images.githubusercontent.com/37960853/177234616-04e7afb1-5366-4894-8b38-c107b6b832fb.webm)

## Testing Plan 
- [x] Updated tests in `frontends` for new copy
- [x] Moved `UnlockMachineScreen` tests to `libs/ui`
- [x] Added sanity check tests for new `libs/ui` components (not snapshots)

## Checklist


adghayes:
I should have split the refactor and the restyle into different commits but... alas. Luckily it isn't too much to review.

libs/ui/src/remove_card_screen.tsx
```diff
@@ -0,0 +1,31 @@
...
+export function RemoveCardScreen(): JSX.Element {
+  return (
+    <Screen white>
+      <Main centerChild>
+        <Prose textCenter theme={fontSizeTheme.medium}>
+          <p>
+            <RemoveCardImage
+              aria-hidden
+              src="assets/remove-card.svg"
+              alt="Remove Card Icon"
```
arsalansufi, line 22: Since we already have text with clear instructions, this image can probably be considered a decorative one from an accessibility standpoint, in which case we should make the `alt` tag an empty string, `alt=""`. Much more on this here! https://www.w3.org/WAI/tutorials/images/decorative

libs/ui/src/remove_card_screen.tsx
```diff
@@ -0,0 +1,31 @@
...
+            />
+          </p>
```
arsalansufi, line 24: Since this `p` just contains an image and no text, let's remove the `p` wrapping (and add styles as needed to `RemoveCardImage`)

libs/ui/src/invalid_card_screen.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Screen } from './screen';
+import { Main } from './main';
+import { Prose } from './prose';
+import { fontSizeTheme } from './themes';
+
+export function InvalidCardScreen(): JSX.Element {
+  return (
+    <Screen white>
+      <Main centerChild>
+        <Prose textCenter theme={fontSizeTheme.medium} maxWidth={false}>
```
arsalansufi, line 11: Any reason we apply `maxWidth={false}` to this screen but not the `RemoveCardScreen`?

libs/ui/src/unlock_machine_screen.tsx
```diff
@@ -23,16 +22,19 @@ const NumberPadWrapper = styled.div`
...
   text-align: center;
   font-family: monospace;
   font-size: 1.5em;
   font-weight: 600;
 `;
 
-export function UnlockMachineScreen(): JSX.Element {
-  const { auth, electionDefinition, machineConfig } = useContext(AppContext);
+interface Props {
+  auth: DippedSmartcardAuth.CheckingPasscode;
```
arsalansufi, line 34: I believe we can also expect to receive an `InsertedSmartcardAuth.CheckingPasscode` auth object here. For example, on BMDs, admins will still use inserted smartcard auth instead of dipped smartcard auth. Likewise, poll worker cards will soon be able to have PINs (for VVSG2 cert).

(cc @jonahkagan)

libs/ui/src/invalid_card_screen.tsx
adghayes: It looks like the `InvalidCardScreen` text will wrap if `maxWidth` is true. But the `RemoveCardScreen` text is too short to wrap even if `maxWidth` is true.

adghayes:
Some funky test failures...

`precinct-scanner` test coverage not passing because the removal of the `unlock_admin_screen.tsx` page brought the average function coverage down. This seems like a case where our test coverage thresholds are not achieving anything. Do I just go find somewhere else unrelated to add a test to bring the coverage up?

`integration-testing-election-manager` is failing because of an assertion timeout. I suspect this would succeed if it ran again. Those tests are always so slow, I think we should look into making them faster.

libs/ui/src/invalid_card_screen.tsx
arsalansufi: Got it!

arsalansufi:
> precinct-scanner test coverage not passing because the removal of the unlock_admin_screen.tsx page brought the average function coverage down. This seems like a case where our test coverage thresholds are not achieving anything. Do I just go find somewhere else unrelated to add a test to bring the coverage up?

If the average test coverage percentage just dropped because of the removal of a fully covered file, I think it's okay to decrease the coverage threshold! Since we haven't actually regressed or increased the work to get to 100% test coverage

frontends/bsd/src/app.test.tsx
```diff
@@ -444,7 +444,7 @@ test('authentication works', async () => {
   fireEvent.click(screen.getByText('6'));
 
   // 'Remove Card' screen is shown after successful authentication.
-  await screen.findByText('Remove card.');
+  await screen.findByText('Remove card', { exact: false });
```
jonahkagan, line 447: One way I've seen inexact matching done in the testing library docs is to use a regex: `await screen.findByText(/Remove card/)`

I would personally lean towards just checking for the actual full text that's expected on the screen, but that's just me!

jonahkagan (commented):
Love this refactor!

frontends/bsd/src/app.test.tsx
adghayes: I think of the full text as copy that could change (and did change between the original screen and this PR) and I wanted to avoid testing copy. I believe this copy is final, however, so checking the actual full text makes sense to me to avoid unintended consequences of inexact matching, so I switched it.

adghayes:
@arsalansufi @jonahkagan this should be good to merge pending your approval.

arsalansufi (approved):
LGTM!

PR #2116, opened 2022-07-08
fix(em): close the kiosk filewriter when done saving zip file.

### LINKED ISSUE #2115: nh-demo production image: VxAdmin does not eject USB and ballot package is not written to disk

benadida:
fixes #2115

We saw this in dev but didn't take it seriously enough. It looks like the Zip / Vite refactoring for `election-manager` closed the outer zip writer, but not the inner `Kiosk.FileWriter` instance. That left the writer open, and then we couldn't eject USB drives anymore, but only when saving the ballot package.

Now we keep track of the `FileWriter` and call `end()` when needed.


arsalansufi (approved):
LGTM!

Issue #2124, opened 2022-07-10
VxScan: design for new screens


mattroe:
New screens as defined in: https://docs.google.com/document/d/1PJ_lRSP2eINez62AuoDt42ISCK3zbewPkLtsnL4qFkc/edit#

mattroe:
Closing as covered by https://github.com/votingworks/vxsuite/pull/2129. will open an issue later for design & copy polish

Issue #2055, opened 2022-07-01
VxScan frozen on insert after invalid ballot scan


carolinemodic:
Today I have been able to somewhat consistently reproduce the issue in the video below, where VxScan is frozen and won’t pull in ballots from the “Scan Your Ballot Screen”

Reproduction Steps:
• Insert invalid ballot (blank piece of paper)
• On rejection, pull it out and re-insert it after 1 - 2 seconds
    ◦ If you re-insert too quickly, it will stay on the “Scanning Error” Screen
    ◦ If you re-insert too slowly, it will scan the ballot normally
    ◦ If you re-insert sometime in between, it will do the following:
        ▪︎ Pull paper very slightly in, push it out, pull it in
        ▪︎ Scan it, but the screen will not change to “Scanning…“, it will stay on “Scan Your Ballot”
        ▪︎ Once rejected, VxScan will be wedged in this state
I can consistently wedge it within 25 tries or so.

@caroline I’m not sure if you want to dedicate any troubleshooting time to this if the whole scanner flow is going to be reworked? But documenting here so we can at least re-test later

Issue automatically created from slack feedback reported by drew 
https://votingworks.slack.com/archives/CJU9MSC6S/p1656697896700599?thread_ts=1656697896.700599&cid=CJU9MSC6S

jonahkagan:
@adghayes I'm pretty sure this is resolved now, so gonna close, but let me know if you can still repro this

Issue #1958, opened 2022-06-15
VxScan: change behavior when we fail to eject ballot


carolinemodic:
## Description
Some more discussion/thought is going to be needed on what the "right" flow is here. See
https://votingworks.slack.com/archives/C03KGSP78MQ/p1655260069583929 for a description of the current behavior/codepath. 

## Desired Behavior
Continue to reject the paper out the front but do NOT count the ballot as tabulated in this case, show a scanning error or something on screen to prompt the user to try to scan again, this should be the first and only time the scan is actually counted. 

Separately the logs have surfaced that we don't handle the PLKSS_ERRCODE_PAPER_STATUS_JAM enum value which we should fix. 

Logs: https://drive.google.com/file/d/1W2idECNYyLzcmPLuKNvlj9aRZ9k1197L/view?usp=sharing the relevant repro starts on line 353,352.

Issue #1839, opened 2022-05-17
VxScan: review tight for loop in plustek scanner status checking


benadida:
I also noticed that the scan service aggressively polls for plustek scanners (there's just a `while` loop with no delay or anything). That's the most aggressive polling I've seen anywhere in the system so far.

_Originally posted by @jonahkagan in https://github.com/votingworks/vxsuite/issues/1168#issuecomment-1061204234_

PR #2129, opened 2022-07-11
Precinct scanner state machine

jonahkagan:
## Overview
Tasks: https://github.com/votingworks/vxsuite/issues/1958, https://github.com/votingworks/vxsuite/issues/2124, https://github.com/votingworks/vxsuite/issues/1824, https://github.com/votingworks/vxsuite/issues/1839, https://github.com/votingworks/vxsuite/issues/2055

_Review by commits_

Adds a state machine implementation of precinct scanning to services/scan and updates the precinct scanner frontend to use it.

This PR gets the basic functionality working, but does not add new tests to cover it. The goal is to get this into `main` quickly so we can start building on top of it. Outstanding comments from this PR will be addressed as follow ups.

## Demo Video or Screenshot
Too much to screenshot - recommend playing around with it

## Testing Plan 
- Update existing tests
- Manual testing


eventualbuddha:
👾 👾 👾

eventualbuddha:
```
                                                                            
                                              ░░░░░░                        
                                            ▒▒░░░░░░░░                      
                                            ▒▒░░░░░░░░                      
                ████████████                ▒▒░░░░░░░░                      
              ██▒▒▒▒▒▒▒▒▒▒▒▒██              ▒▒░░░░░░░░          ░░░░        
            ██▒▒░░░░░░░░░░░░▒▒██            ▒▒░░░░░░░░        ░░    ░░      
          ██▒▒░░░░░░░░░░░░░░░░▒▒██          ▒▒░░░░░░░░      ░░        ░░    
        ██▒▒░░░░▒▒▒▒▒▒▒▒▒▒▒▒░░░░▒▒██        ▒▒░░░░░░░░      ░░        ░░    
        ██░░▓▓                ▓▓░░▒▒██      ▒▒░░░░░░░░        ░░    ░░      
    ▓▓██▒▒▓▓                    ▓▓▒▒██▓▓    ▒▒░░░░░░░░          ░░░░        
  ▓▓░░██▓▓░░  ░░          ░░▓▓    ▓▓▓▓░░▓▓  ▒▒░░░░░░░░                      
  ▓▓░░░░▓▓░░▓▓          ░░▓▓░░▓▓  ▓▓░░░░▓▓██▒▒░░░░░░░░                      
    ▓▓░░▓▓▓▓░░        ░░▓▓░░░░▓▓░░▓▓░░▓▓██▒▒▒▒░░░░░░░░                      
      ▓▓░░▓▓░░░░░░▓▓▓▓▓▓░░░░░░░░▓▓  ▓▓▒▒░░░░▒▒░░░░░░░░                      
      ▓▓░░▓▓▓▓▓▓▓▓▓▓░░░░░░░░████▓▓░░▓▓░░░░░░░░░░░░░░                        
      ▓▓░░▓▓  ░░██░░░░░░░░██░░  ▓▓░░▓▓▒▒▒▒▒▒██░░                            
        ▓▓▓▓░░░░  ░░░░░░░░  ░░░░▓▓▓▓  ██████▒▒░░▓▓░░▓▓                      
          ██░░░░░░░░░░░░░░░░░░░░██        ▓▓▓▓▒▒▒▒▒▒▓▓▓▓                    
          ██▓▓░░░░░░░░░░░░░░░░▓▓▓▓██    ▓▓  ▓▓▓▓  ▓▓▓▓  ▓▓                  
        ██▒▒▓▓▓▓░░░░░░░░░░░░▓▓▓▓░░▓▓██        ▓▓▒▒▓▓                        
      ██▒▒░░▓▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓░░░░░░▓▓▓▓▓▓██░░░░░░██                      
    ██▒▒░░░░▓▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓░░░░░░░░░░░░░░░░░░░░██                      
  ██░░░░░░▓▓▓▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓██▓▓██████████░░░░░░██                      
  ██░░░░░░██▒▒▓▓▓▓▓▓░░  ▓▓▓▓▓▓▒▒▓▓██          ██████                        
  ██░░░░░░▓▓▒▒░░▒▒▓▓░░░░▓▓▒▒░░▒▒██                                          
    ████████▒▒░░░░▓▓▒▒▒▒▓▓░░░░▒▒██                                          
          ████▒▒▓▓░░░░░░░░▓▓▒▒████                                          
        ██▓▓▓▓████████████████▓▓▓▓██                                        
        ██▓▓▓▓▓▓██        ██▓▓▓▓▓▓██                                        
          ██████            ██████                                          
```

services/scan/src/plustek_machine.ts
```diff
@@ -0,0 +1,428 @@
...
+  | { type: 'INTERPRETATION_INVALID' }
+  | { type: 'REVIEW_CAST' }
+  | { type: 'REVIEW_RETURN' }
+  | { type: 'SET_INTERPRETATION_MODE'; mode: InterpretationMode };
+
+async function connectToPlustek(): Promise<ScannerClient> {
+  const plustekClient = await createClient(DEFAULT_CONFIG);
+  // console.log(plustekClient);
+  if (plustekClient.isOk()) return plustekClient.ok();
+  throw plustekClient.err();
```
eventualbuddha, line 58: This is just `return plustekClient.unsafeUnwrap();`

services/scan/src/precinct_scanner_app.ts
```diff
@@ -0,0 +1,454 @@
...
+    electionDefinition,
+    ballotImagesPath: workspace.ballotImagesPath,
+    testMode: store.getTestMode(),
+    layouts,
+  });
+
+  machine.configure(store, interpreter);
+}
+
+export function buildPrecinctScannerApp(
```
jonahkagan, line 48: @eventualbuddha probably my biggest question is how to organize the shared bits of the scan service that both central scanner and precinct scanner use. To start, I just copy/pasted all the routes used by the precinct scanner and tweaked them to not use `Importer` (instead inlining the relevant pieces of `Importer`). Should I leave things this way? Seems like a lot of duplicated code. Part of me thinks these could eventually be two separate services, in which case splitting them seems ok. But then again there's a lot of shared logic (though I guess most of it is just glue code/request validation). Curious to hear your thoughts

services/scan/src/precinct_scanner_app.ts
```diff
@@ -0,0 +1,454 @@
...
+  app.get('/*', (request, response) => {
+    const url = new URL(`http://${request.get('host')}${request.originalUrl}`);
+    url.port = '3000';
+    response.redirect(301, url.toString());
+  });
+
+  // Reload configuration from store on startup
+  if (store.getElectionDefinition()) {
+    void configureMachine(machine, workspace);
+  }
```
jonahkagan, line 451: @eventualbuddha does this seem right?

services/scan/src/simple_interpreter.ts
```diff
@@ -0,0 +1,211 @@
...
+      const [frontQrcodeOutput, backQrcodeOutput] =
+        qrcodeWorker.normalizeSheetOutput(
+          electionDefinition,
+          await mapSheet(
+            [frontPath, backPath],
+            qrcodeWorker.detectQrcodeInFilePath
+          )
+        );
+      return ok(
+        await mapSheet(
```
jonahkagan, line 110: @eventualbuddha are we good with `mapSheet` in prod or were you thinking of moving back to a worker model? I didn't look closely at the worker pool - is it using multiple processes?

services/scan/src/plustek_machine.ts
```diff
@@ -0,0 +1,809 @@
...
+      invoke: {
+        src: connectToPlustek,
+        onDone: {
+          target: 'checking_initial_paper_status',
+          actions: assign({ client: (_context, event) => event.data }),
+        },
+        onError: 'error_disconnected',
+      },
+    },
+    waiting_for_configuration: {
```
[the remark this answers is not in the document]
eventualbuddha: Assuming we'll want to handle calibration through the state machine, I'd say "scan service start" makes sense as the place to connect to plustek.

services/scan/src/simple_interpreter.ts
eventualbuddha: The `mapSheet` helper itself I think is useful and could be applied in a lot of places.

As for the worker pool, here are my thoughts on that:
- maybe it's fast enough without it? @benadida to consider this
- while having our own worker pool implementation has made some things easier, like making `Result` be sendable across processes, maybe we should be using [Piscina](https://github.com/piscinajs/piscina) or similar. There are probably gremlins in here as regards our use of TypeScript in dev and compiled JS in production as that's something I've had to explicitly support.

services/scan/src/precinct_scanner_app.ts
eventualbuddha: I think a reasonable compromise for now would be to have them in the same service but distinct `Application`s. That way the shared code is relatively easy to manage and we just start the application that makes sense. Longer term perhaps we do make them distinct services, but that'll require an investment in factoring out the shared code like `Store` etc.

services/scan/src/plustek_machine.ts
```diff
@@ -0,0 +1,809 @@
...
+
+  const sheetId = uuid();
+  const result = await interpreter.interpret(sheetId, scannedSheet);
+  const interpretedSheet = result.unsafeUnwrap();
+  const [front, back] = interpretedSheet;
+  const frontType = front.interpretation.type;
+  const backType = back.interpretation.type;
+  const interpretation: Interpretation = { sheetId, sheet: interpretedSheet };
+
+  if (frontType === 'BlankPage' && backType === 'BlankPage') {
```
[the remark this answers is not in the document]
eventualbuddha: We should try to consolidate this stuff, but it might take a lot of effort to do so properly. IMO we should not have separate `PageInterpretation` objects for each page. Instead, we should have a `SheetInterpretation` that combines them together, marks/votes/errors etc.

Also, this particular logic looks wrong. A `BlankPage` means a literally white page with essentially nothing on it. A `BlankBallot` is a ballot without any marks. Confusing, I know 😢

services/scan/src/plustek_machine.ts
```diff
@@ -0,0 +1,809 @@
...
+        onError: {
+          target: 'error_accepting',
+          actions: assign((_context, event) => ({ error: event.data })),
+        },
+      },
+    },
+    accepted: {
+      // Delay on this state for 5s to show accepted screen, then go to no_paper
+      // (unless we get a different status in the meantime, e.g. ready to scan)
+      after: { 5000: 'no_paper' },
```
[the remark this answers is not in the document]
eventualbuddha: I think your idea is reasonable. I'd try that approach.

services/scan/src/plustek_machine.ts
```diff
@@ -0,0 +1,809 @@
...
+      after: { 1000: 'no_paper' },
+    },
+    // Paper has been rejected and is held in the front, waiting for removal.
+    rejected: {
+      invoke: pollPaperStatus,
+      on: {
+        SCANNER_READY_TO_SCAN: { target: 'rejected', internal: true },
+        SCANNER_NO_PAPER: 'no_paper',
+      },
+    },
```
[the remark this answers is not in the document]
eventualbuddha: I'd be inclined to try the mini state machine and run it by @benadida to gut check it. If it feels too obtuse to either of you, try a more traditional approach like a builder for this particular pattern.

services/scan/src/precinct_scanner_app.ts
eventualbuddha: Yeah, I think so. The thing to be careful of with `void …` to essentially "background" a `Promise`-returning function is that you should be quite sure that it won't throw under normal circumstances. In future (unspecified) versions of NodeJS an unhandled promise rejection will crash the process.

Will this get called again somehow if the machine doesn't boot up already configured?

services/scan/src/scanners/fujitsu.ts
```diff
@@ -128,7 +128,7 @@ export class FujitsuScanner implements Scanner {
       scannedFiles.push(path);
       if (scannedFiles.length % 2 === 0) {
         results.resolve(
-          Promise.resolve(scannedFiles.slice(-2) as SheetOf<string>)
+          Promise.resolve(scannedFiles.slice(-2) as unknown as SheetOf<string>)
```
eventualbuddha, line 131: I ran into this issue locally when I tried it out. What I'd probably do instead is something like this:

```ts
    let frontPagePath: string | undefined;
    let backPagePath: string | undefined;
    new StreamLines(scanimage.stdout).on('line', (line: string) => {
      const path = line.trim();
      void this.logger.log(
        LogEventId.FujitsuScanImageScanned,
        'system',
        {
          message: `scanimage [pid=${scanimage.pid}] reported a scanned file: ${path}`,
          disposition: 'success',
        },
        debug
      );

      if (!frontPagePath) {
        frontPagePath = path;
      } else if (!backPagePath) {
        backPagePath = path;
      } else {
        results.resolve(Promise.resolve([frontPagePath, backPagePath]));
        frontPagePath = undefined;
        backPagePath = undefined;
      }
    });
```

services/scan/src/plustek_machine.ts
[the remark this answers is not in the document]
jonahkagan: Ok yeah that makes sense, I like that vision. Might expand my scope a bit too much - we'll see.

Good callout on the `BlankPage` - thanks. Seems like `BlankPage` on both sides isn't actually a real case that could happen then?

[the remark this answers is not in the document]
eventualbuddha: It'll happen if someone puts in a fully blank sheet, but that's it.

services/scan/src/plustek_machine.ts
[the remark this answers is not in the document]
jonahkagan: Ah right, the one thing I hadn't gotten to yet. Totally forgot to factor that in. I think in the app you can't calibrate before configuration, so it's still possible that we could wait for configuration to connect. Thoughts?

services/scan/src/precinct_scanner_app.ts
jonahkagan: Ok that sounds good. I'll continue down this path then.

services/scan/src/plustek_machine.ts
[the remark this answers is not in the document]
eventualbuddha: Ah, you're right. That's fine.

services/scan/src/precinct_scanner_app.ts
jonahkagan: Oh ok interesting. Any thoughts on how to make it safe but non-blocking?

This function also gets called in the `doneTemplates` endpoint (which I think signifies the end of configuration)

eventualbuddha: Mostly just make sure it handles anything that might be thrown by things it calls, except for `assert`s. Probably means you'll want to think about what this should do if loading configuration fails. Have some sort of configuration error state?

services/scan/src/plustek_machine.ts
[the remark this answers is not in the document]
benadida: we might want to enable calibration before configuration, as i can see an election admin wanting to do that in one fell swoop before configuration, so if it's just as easy, let's build that capability. If it's not, I'm okay not adding it.

services/scan/src/plustek_machine.ts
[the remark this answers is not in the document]
benadida: could this be done with the same states but a context variable?

services/scan/src/simple_interpreter.ts
benadida: I like the idea of trying this first without parallelizing / workers.

benadida: the parallelizing is likely more important for central scanner anyways.

services/scan/src/plustek_machine.ts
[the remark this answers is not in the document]
jonahkagan: Did this, seems to work well!

services/scan/src/scanners/fujitsu.ts
jonahkagan: Ended up solving like this:

```ts
      scannedFiles.push(path);
      if (scannedFiles.length % 2 === 0) {
        const [frontPath, backPath] = scannedFiles.slice(-2);
        results.resolve(Promise.resolve([frontPath, backPath]));
      }
```

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+  }
+}
+
+export interface Context {
+  store?: Store;
+  interpreter?: SimpleInterpreter;
+  client?: ScannerClient;
+  scannedSheet?: SheetOf<string>;
+  interpretation?: InterpretationResultEvent;
+  error?: Error;
```
benadida, line 60: might we want to have all 3 of these bundled as a single context field `scanResult` so that that field is updated as a single operation? Maybe this is too paranoid, but I can imagine `scannedSheet` and `interpretation` getting out of sync.

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+
+export type InterpretationResultEvent = Scan.InterpretationResult & {
+  interpretation: Interpretation;
+};
+
+type CommandEvent =
+  | { type: 'SCAN' }
+  | { type: 'ACCEPT' }
+  | { type: 'RETURN' }
+  | { type: 'START_OVER' }
```
benadida, line 95: nit: `START_OVER` is not clear as a command compared to others. Feels almost like a catch-all. Maybe it isn't, just commenting on naming.

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+  if (error instanceof ClientDisconnectedError) {
+    return { type: 'SCANNER_DISCONNECTED' };
+  }
+  switch (error) {
+    // This error code happens when the paper gets blocked and tries to scan for
+    // a long time (or if you just hold onto the paper while scanning).
+    case ScannerError.SaneStatusInval:
+      return { type: 'SCANNER_JAM' };
+    case ScannerError.SaneStatusIoError:
+      return { type: 'SCANNER_DISCONNECTED' };
```
benadida, line 151: are we sure we want to treat these errors as disconnected? In my playing with the scanner, this error appeared to be recoverable,

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+
+  const sheetId = uuid();
+  const result = await interpreter.interpret(sheetId, scannedSheet);
+  const interpretedSheet = result.unsafeUnwrap();
+  const [front, back] = interpretedSheet;
+  const frontType = front.interpretation.type;
+  const backType = back.interpretation.type;
+  const interpretation: Interpretation = { sheetId, sheet: interpretedSheet };
+
+  if (
```
benadida, line 234: all the interpretation stuff below this line that looks at front and back, seems like it belongs in `Interpreter`, not in the state machine.

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+  // Optimistically record the interpretation in the store. If accept fails
+  // (e.g. due to paper jam), we delete the interpretation from the store. If
+  // the Plustek crashes during accept, it will automatically accept on reboot.
+  const { sheetId, sheet } = interpretation.interpretation;
+  storeInterpretedSheet(store, sheetId, sheet);
+  debug('Stored sheet before accepting: %s', sheetId);
+  debug('Accepting');
+  const acceptResult = await client.accept();
+  debug('Accept result: %o', acceptResult);
+  acceptResult.unsafeUnwrap();
```
benadida, line 350: is there a missing `return` here?

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+            error: new PrecinctScannerError('paper_in_back_on_startup'),
+          }),
+        },
+      },
+    },
+    no_paper: {
+      entry: assign({
+        error: undefined,
+        scannedSheet: undefined,
+        interpretation: undefined,
```
benadida, line 461: maybe this negates my concern above about all of these being set at once.

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+    returning: {
+      invoke: {
+        src: reject,
+        onDone: 'checking_returning_completed',
+        onError: 'error_jammed',
+      },
+    },
+    // After rejecting, before the plustek grabs the paper to hold it, it sends
+    // NO_PAPER status for a bit before sending READY_TO_SCAN. So we need to
+    // wait for the READY_TO_SCAN.
```
benadida, line 576: 😍

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+      invoke: pollPaperStatus,
+      on: {
+        SCANNER_BOTH_SIDES_HAVE_PAPER: {
+          target: 'error_both_sides_have_paper',
+          internal: true,
+        },
+        // For now, if the front paper is removed, just reject the back paper,
+        // since we don't have context on how we got here and what was supposed
+        // to happen.
+        SCANNER_READY_TO_EJECT: 'rejecting',
```
benadida, line 670: This is probably overly conservative, as the main way this will happen is if you put in a second sheet right before the first sheet is accepted, so removing the second sheet from intake should then cause the first sheet to resume being accepted. But, we can optimize this later. This isn't wrong per se, just overly conservative.

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -0,0 +1,854 @@
...
+          error instanceof PrecinctScannerError ? error.type : 'plustek_error',
+      };
+    },
+
+    scan: () => {
+      if (!machineService.state.matches('ready_to_scan')) {
+        return err(
+          new Error('Scanner must be in ready_to_scan state in order to scan')
+        );
+      }
```
benadida, line 793: is this repetitive? Shouldn't the state machine be able to say "hey, I don't know how to scan right now." ?

services/scan/src/simple_interpreter.ts
```diff
@@ -0,0 +1,211 @@
...
+ * Stores an interpreted sheet of ballot paper in the {@link Store}, returning
+ * the effective sheet ID.
+ */
+export function storeInterpretedSheet(
+  store: Store,
+  sheetId: Id,
+  sheet: SheetOf<PageInterpretationWithFiles>
+): Id {
+  // TODO instead of one batch per ballot, use one batch per scanning session
+  // (e.g. from polls open to polls close)
```
benadida, line 202: before you do that, keep in mind that we have code that removes an incomplete batch on startup of the scanning software.

services/scan/jest.config.js
```diff
@@ -18,10 +18,10 @@ module.exports = {
...
   coverageThreshold: {
     global: {
-      statements: 90,
-      branches: 73,
-      functions: 90,
-      lines: 89,
+      statements: 78,
+      branches: 65,
+      lines: 78,
+      functions: 76,
```
benadida: 😨

services/scan/src/precinct_scanner_app.ts
```diff
@@ -0,0 +1,466 @@
...
+          {
+            type: 'invalid-value',
+            message: `expected content type to be application/octet-stream, got ${request.header(
+              'content-type'
+            )}`,
+          },
+        ],
+      });
+      return;
+    }
```
benadida, line 99: are the config calls shared between central and precinct? Hopefully they can be?

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -643,14 +316,55 @@ export function AppRoot({
...
 
-  function dismissError() {
-    /* istanbul ignore next */
-    if (timeoutToInsertScreen) {
-      window.clearTimeout(timeoutToInsertScreen);
+  const togglePollsOpen = useCallback(() => {
+    dispatchAppState({ type: 'togglePollsOpen' });
+  }, []);
+
+  const toggleTestMode = useCallback(async () => {
```
benadida, line 323: TODO: move away from `toggle` APIs and have them be idempotent, i.e. `setTestMode(true/false)`.

benadida (approved):

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Good to know. Do you have any ideas for a clearer name? It's used for both going back to "no_paper" from "accepted" and "calibrated" so it is doing double duty a bit.

Maybe something like "WAIT_FOR_PAPER" or "RESET_TO_NO_PAPER"?

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Oh interesting I'm not sure. That was just the error that happened when I turned off the Plustek. I'm not sure how to distinguish a recoverable `SaneStatusIoError` from that case. Do you remember what triggered it when it was recoverable?

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Agreed, that's on my todo list

mattroe:
This might be a new record for number of issues closed in a PR :)

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Technically no because we don't actually do anything with the results. `unsafeUnwrap` just makes sure we throw an error if there is one. But I will add the `return` for consistency with the others.

(Makes me realize that one of the downsides of having `Result` type is that whenever you interface with a library that expects thrown errors you have to unwrap)

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Good question. `error` is used for errors in other cases that aren't related to scanning, so that feels like it needs to be separate (or get refactored into two separate error fields).

`scannedSheet` and `interpretation` are separate steps, so they aren't set at the same time. `scannedSheet` is input for the interpreter. I do feel a little weird about the current `Context` type in general because it's just a blob of all the state fields. I've been considering using a union type for the various states it can actually be in, which might resolve some of these concerns.

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Yep, I agree completely. My goal on this first pass was to make sure we never ended up in an inconsistent state.

I think we should for sure come back and improve the user experience here. There is a fancy [history](https://xstate.js.org/docs/guides/history.html#history-state-configuration) feature in xstate that would allow this to be easily changed, but I was/am hesitant to bring in fancier state machine features if we don't have to.

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Yes, though the error isn't the most clear. I put this in to help abstract away the state machine. But maybe I'll try removing it to see what happens.

services/scan/src/precinct_scanner_state_machine.ts
benadida: `WAIT_FOR_PAPER` is great.

services/scan/src/precinct_scanner_state_machine.ts
benadida: I've seen it happen usually right as you try to scan. In your initial BallotQuest code, you went to checkpaperstatus on that error, and checkpaperstatus usually was able to get a status and recover from there.

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Oh now I remember why I put this in. `send` won't fail when this happens, so the API request will always be successful. I think maybe that's ok, since there will be an error status within the state machine, which is probably better anyway since we handle that smartly on the frontend (vs failed API requests, which I don't think we handle well).

services/scan/src/simple_interpreter.ts
jonahkagan: Hmm, yeah it will probably take some care to make this safe. It's not high on my todo list - prob only really important when we get to RLAs

services/scan/src/precinct_scanner_app.ts
jonahkagan: They aren't shared currently. The major obstacle to sharing them is that the central scanner API uses `Importer`, whereas this uses `store` directly. It will take some refactoring/cleverness to make them reusable.

Given that most of the actual logic of these endpoints is encapsulated in the store, I'm not sure there's _that_ much value to figuring out how to reuse them if it takes a lot of work. Something to investigate

frontends/precinct-scanner/src/app_root.tsx
jonahkagan: Yes I agree, toggle scares me. Was trying to minimize my changes when possible so resisted changing this one

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: 8ca47de57

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Hmm interesting. Is there any way I can easily repro this? I'm hesitant to change this logic without being able to test that my changes actually work.

One idea I have to fix it is to retry getting paper status a few times on error before sending this disconnected event.

benadida: it can be repro'ed sometimes by holding the ballot back a little when it's trying to grab it / scan it.

To be clear, I'm fine with not fixing this now! We can optimize this later.

jonahkagan:
Ok, I'm planning to merge this as-is and leave comments open for things that I will address in future PRs.

PR #2135, opened 2022-07-13
Add new methods to Card interface to support smartcard programming

arsalansufi:

## Overview

Issue link: Part of https://github.com/votingworks/vxsuite/issues/1919, https://github.com/votingworks/vxsuite/issues/1920, and https://github.com/votingworks/vxsuite/issues/1921

This PR adds new methods to the `Card` interface to support smartcard programming. Smartcard programming currently bypasses this interface and communicates with the services/smartcards API directly, which isn't ideal.

The PR also updates the `WebServiceCard` interface to use x-www-form-urlencoded encoding instead of multipart/form-data (where relevant) since the latter can convert LF line endings to CRLF line endings, resulting in election hash mismatches. See https://github.com/votingworks/vxsuite/pull/1934 for more context.

## Testing 

- [x] Updated automated tests
- [x] Manually used the new methods to successfully program cards
- [ ] Looking into manual testing of existing consumers of `WebServiceCard.writeLongUint8Array` (the method affected by the multipart/form-data to x-www-form-urlencoded switch)

## Checklist

### DIFF
--- a/libs/utils/src/Card/memory_card.ts
+++ b/libs/utils/src/Card/memory_card.ts
@@ -1,6 +1,6 @@
 import { ok, Optional, Result, safeParseJson } from '@votingworks/types';
 import { z } from 'zod';
-import { Card, CardSummary } from '../types';
+import { Card, CardSummary, ShortAndLongValues } from '../types';
 
 /* eslint-disable @typescript-eslint/require-await */
 
@@ -101,6 +101,25 @@ export class MemoryCard implements Card {
     }
 
     this.longValue = Uint8Array.from(value);
+  }
+
+  /**
+   * Writes new short and long values to the card.
+   */
+  async writeShortAndLongValues({
+    shortValue,
+    longValue,
+  }: ShortAndLongValues): Promise<void> {
+    await this.writeShortValue(shortValue);
+    await this.writeLongUint8Array(new TextEncoder().encode(longValue));
+  }
+
+  /* istanbul ignore next: A no-op but necessary to satisfy the Card interface */
+  /**
+   * Overrides card write protection.
+   */
+  async overrideWriteProtection(): Promise<void> {
+    return Promise.resolve();
   }
 
   /**
--- a/libs/utils/src/Card/web_service_card.test.ts
+++ b/libs/utils/src/Card/web_service_card.test.ts
@@ -57,7 +57,11 @@ it('reads binary data from /card/read_long_b64', async () => {
 
 it('writes short value using /card/write', async () => {
   fetchMock.post('/card/write', (url: string, mockRequest: MockRequest) => {
-    expect(url).toBe('/card/write');
+    expect(url).toEqual('/card/write');
+    expect(mockRequest.headers).toEqual({
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    });
     expect(mockRequest.body).toEqual('abc');
     return { success: true };
   });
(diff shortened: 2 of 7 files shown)


libs/utils/src/Card/web_service_card.ts
```diff
@@ -29,8 +37,9 @@ export class WebServiceCard implements Card {
   async readLongObject<T>(
     schema: z.ZodSchema<T>
   ): Promise<Result<Optional<T>, SyntaxError | z.ZodError>> {
-    const response = await fetch('/card/read_long');
-    const { longValue } = await response.json();
+    const { longValue } = (await fetchJson(
+      '/card/read_long'
+    )) as LongValueResponse;
```
arsalansufi, line 47: The `fetchJson` helper checks status codes via the `ok` field on responses (which is set to false for 4xxs and 5xxs):

https://github.com/votingworks/vxsuite/blob/42069db8d35f25a239f0e7880e9d137215e5dbc4/libs/utils/src/fetch_json.ts#L13-L15

The base `fetch` doesn't throw errors for 4xxs and 5xxs. So before this PR, 4xxs and 5xxs would result in us accessing properties of undefined objects, etc.

libs/utils/src/Card/web_service_card.ts
```diff
@@ -93,13 +93,12 @@ export class WebServiceCard implements Card {
...
-
-    formData.append('long_value', longValueBase64);
-
     const { success } = (await fetchJson('/card/write_long_b64', {
       method: 'post',
-      body: formData,
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+      },
+      body: `long_value=${encodeURIComponent(longValueBase64)}`,
```
arsalansufi, line 106: As noted in the commit description, see https://github.com/votingworks/vxsuite/pull/1934 for more context!

libs/utils/src/Card/web_service_card.ts
```diff
@@ -104,4 +109,37 @@ export class WebServiceCard implements Card {
...
+    })) as SuccessIndicationResponse;
+    if (!success) {
+      throw new Error('Failed to write short and long values');
+    }
+  }
+
+  /**
+   * Overrides card write protection.
+   */
+  async overrideWriteProtection(): Promise<void> {
```
arsalansufi, line 137: I went back and forth on whether to implicitly override write protection within write methods or to have consumers explicitly do so. I landed on the latter as a safer, non-breaking change

libs/utils/src/Card/memory_card.ts
```diff
@@ -103,6 +103,25 @@ export class MemoryCard implements Card {
...
+    await this.writeLongUint8Array(new TextEncoder().encode(longValue));
+  }
+
+  /* istanbul ignore next: A no-op but necessary to satisfy the Card interface */
+  /**
+   * Overrides card write protection.
+   */
+  async overrideWriteProtection(): Promise<void> {
+    return Promise.resolve();
+  }
```
arsalansufi, line 123: I _could_ implement a write-protection mechanism for `MemoryCard` and have it mimic the (fairly complex) circumstances under which our web service cards automatically enable and disable write protection, but that didn't feel worth the trouble, especially when we're considering revisiting the write-protection mechanism anyway

libs/utils/src/Card/web_service_card.test.ts
```diff
@@ -103,6 +127,82 @@ it('writes binary data using /card/write_long_b64', async () => {
...
+        Accept: 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+      });
+      const body = mockRequest.body as string;
+      const values = body.split('&');
+      expect(values.length).toEqual(2);
+      expect(
+        values[0]?.startsWith('short_value=') &&
+          values[1]?.startsWith('long_value=')
+      ).toEqual(true);
```
arsalansufi, line 153: I know the short value doesn't have to come before the long value, and this test will break if the string concatenation in `writeShortAndLongValues` ever changes that order, but it didn't feel worth the complexity to support both orderings in this test

libs/utils/src/Card/web_service_card.ts
jonahkagan: Nice! That seems much better. My only concern is that based on that code snippet you linked, it looks like we don't surface the actual error code or message anywhere.

libs/utils/src/Card/web_service_card.ts
jonahkagan: Seems easy enough to change later if we desire

libs/utils/src/Card/web_service_card.test.ts
jonahkagan: I probably would have just checked for an exact expected string in the body haha, so I think you've already gone above and beyond.

jonahkagan (approved):

libs/utils/src/Card/web_service_card.ts
arsalansufi: Good point! I updated `fetch_json` to at least include the status code in its error messages: https://github.com/votingworks/vxsuite/pull/2135/commits/72c7b99897f0994bdf99da9c23caf327f584be89

PR #2138, opened 2022-07-13
Add write-ins to test decks

### LINKED ISSUE #1513: VxAdmin: add write in selection option to L&A test deck
Before an election the election workers run L&A, Logic & Accuracy testing to test that everything with the voting machine system works accurately. This includes printing a test deck with all possible contest options chosen at least once. So a ballot with 2 contests, each with 2 candidates running, will print 2 ballots in the test deck (per precinct). One with the first candidate voted for in both contests, and one with the second. Currently we do not print any ballots in the test deck with the write in option selected. We should add that. We will need to fill in the bubble, but we don't need to add any fake text to the line about who to vote for as it will just count as a write in. (This may change in the future with write-in adjudication). It would potentially look nicer if there was fake handwriting of "Write in" or something though.

adghayes:
## Overview
Closes #1513. Adds mock write-ins to the test decks, such that every possible ballot position is now included in test decks.

## Demo Video or Screenshot

Screen Shot 2022-07-14 at 6 38 02 PM

## Testing Plan
- Unit tests for test deck generation helpers
- Unit test (finally!) for test deck generation
- Updated snapshots, test timings, and expected tallies
- Manual testing

## Checklist


frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
```diff
@@ -60,6 +60,25 @@ function hasVote(
   );
 }
 
+function hasWriteInAtPosition(
+  vote: CandidateVote | undefined,
```
arsalansufi, line 64: I'm on the fence here as to whether we should handle undefined votes in this function or expect consumers to check that a vote is defined before calling this helper

I lean toward the latter since `hasWriteInAtPosition` is asking whether a vote has a write-in at a certain position, so it feels a bit funny to ask that question for a non-existent vote. At the same time, I see the convenience of putting the check in this function

Same question holds for for `writeInNameAtPosition`

Curious what you think!

frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
```diff
@@ -60,6 +60,25 @@ function hasVote(
...
+  position: number
+): boolean {
+  return (
+    vote?.some((candidate) => candidate.writeInPosition === position) ?? false
+  );
+}
+
+function writeInNameAtPosition(
```
arsalansufi, line 72: Since `writeIn` can also be read as a verb, what do you think of renaming this `getWriteInNameAtPosition` to make it clear that we're just retrieving a value and not actually writing in a name?

frontends/election-manager/src/utils/election.ts
```diff
@@ -174,6 +176,37 @@ export function getContestsForPrecinct(
...
 }
 
+export function generateWriteIn(position: number): WriteInCandidate {
+  return {
+    id: 'write-in',
+    isWriteIn: true,
+    name: 'WRITE-IN',
+    writeInPosition: position,
+  };
+}
```
arsalansufi, line 186: Since this isn't a real write-in candidate and just a prefilled one for test deck generation, I think we should somehow make that distinction clear, e.g. by renaming `generateWriteIn` `generateTestDeckWriteIn`. I realize that's imperfect, though, since calling `generateTestDeckWriteIn` within `candidateAtPosition` could be confusing

Happy to bounce ideas!

arsalansufi (commented):
Some initial thoughts!

frontends/election-manager/public/fonts/edu-tas-beginner.css
```diff
@@ -0,0 +1,6 @@
+@font-face {
+  font-family: 'Edu TAS Beginner';
+  font-weight: 400 700;
```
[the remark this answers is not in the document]
adghayes: The font file I'm using here, from Google Fonts, is a "Variable Font File" so it includes both normal and bold and everything in between. Based on what I have seen online, I believe the `font-weight: 400 700;` syntax sets up the range of font weights for which the font can be used, so if you use `font-weight: bold;` anywhere, it will be bold. I have confirmed that works.

frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
```diff
@@ -466,8 +493,9 @@ export function CandidateContestChoices({
   targetMarkPosition,
 }: CandidateContestChoicesProps): JSX.Element {
   const { t } = useTranslation();
-  const writeInCandidates = vote?.filter((c) => c.isWriteIn);
-  const remainingChoices = [...Array.from({ length: contest.seats }).keys()];
+  const writeInItemKeys = [...Array.from({ length: contest.seats }).keys()].map(
```
eventualbuddha, line 496: Possibly easier to understand?
```suggestion
  const writeInItemKeys = take(contest.seats, integers()).map(
```

libs/types/src/election.ts
```diff
@@ -123,6 +124,7 @@ export const CandidateSchema: z.ZodSchema<Candidate> = z
     name: z.string().nonempty(),
     partyIds: z.array(PartyIdSchema).optional(),
     isWriteIn: z.boolean().optional(),
+    writeInPosition: z.number().nonnegative().optional(),
```
eventualbuddha, line 127:
```suggestion
    writeInPosition: z.number().int().nonnegative().optional(),```

File: frontends/election-manager/src/utils/election.ts [hunk 7]
```diff
@@ -174,6 +176,37 @@ export function getContestsForPrecinct(
...
+  return 2;
+}
+
+export function candidateAtPosition(
+  contest: CandidateContest,
+  position: number
+): Candidate {
+  assert(position < numBallotPositions(contest)); // safety check
+  if (position < contest.candidates.length) {
+    return contest.candidates[position];
```
eventualbuddha, line 203: I don't think this will work with [`CandidateContest#rotation`](https://github.com/votingworks/vxsuite/blob/7380af20d365f142f5ac32da536430faf135577f/libs/types/src/election.ts#L199).

frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
```diff
@@ -493,25 +521,21 @@ export function CandidateContestChoices({
...
-            </span>
-          </BubbleMark>
-        </Text>
-      ))}
       {contest.allowWriteIns &&
-        remainingChoices.map((k) => (
-          <WriteInItem key={k} data-write-in>
-            <BubbleMark position={targetMarkPosition}>
+        writeInItemKeys.map((key, position) => (
+          <WriteInItem key={key} data-write-in>
```
eventualbuddha, line 526: Where is the `data-write-in` attribute used?

frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
arsalansufi: Ah nice, especially given that the existing `hasVote` also uses this pattern, this seems totally reasonable

libs/types/src/election.ts
```diff
@@ -115,6 +115,7 @@ export interface Candidate {
   readonly name: string;
   readonly partyIds?: readonly PartyId[];
   readonly isWriteIn?: boolean;
+  readonly writeInPosition?: number;
```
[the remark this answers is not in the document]
adghayes: I agree we should reconsider the types, and I'm glad you looked at it this PR for that reason. Unless you're kidding and I don't get it, I will rename `writeInIndex` so that when we refactor it's clear these are referring to similar things :trollface:. Shall I make a ticket to reconsider and then refactor these types? I think that should be done but punted past m14. 

I was also confused how the `Candidate` and `WriteInCandidate` types are related, and how the default `writeInCandidate` is actually of type `Candidate`

frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
adghayes: I do not know. Nowhere, as far as I can tell? Same goes for other data labels, like `data-candidate` in this file.

libs/types/src/election.ts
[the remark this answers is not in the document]
eventualbuddha: > Shall I make a ticket to reconsider and then refactor these types? I think that should be done but punted past m14.

Yes please 🙏

frontends/election-manager/src/utils/election.ts
eventualbuddha: No, I'm fine with this as long as rotation does work. Probably should have a test to that effect.

frontends/election-manager/src/utils/election.ts
adghayes: Thanks for the name switch suggestion!

The resulting could be a little confusing, but not too confusing. And I'm renaming the function `candidateAtPosition` to `getTestDeckCandidateAtIndex` to be maximally specific, although a mouthful. Sound better?

These functions are written specifically for test decks so I agree it's good to name them specifically for test decks too.

frontends/election-manager/src/utils/election.ts
eventualbuddha: Oh, I forgot it's only for BMDs 🤦 Carry on.

frontends/election-manager/src/utils/election.ts
arsalansufi: Nice! I agree that naming specifically for test decks through and through makes sense. And I also think that verbose names are okay for the sake of clarity!

arsalansufi (approved):
LGTM!

PR #2150, opened 2022-07-18
fix(scan): copy images if we cannot link them

eventualbuddha:
## Overview

We link instead of copy to save space, but if the source and destination are on different partitions or physical disks we cannot link them. If the link fails, just do a copy.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated testing.

### DIFF
--- a/services/scan/src/util/save_images.ts
+++ b/services/scan/src/util/save_images.ts
@@ -10,6 +10,18 @@ interface SaveImagesResult {
   normalized: string;
 }
 
+/**
+ * Links {@link src} to {@link dest} if both paths are on the same file system,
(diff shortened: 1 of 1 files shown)


benadida (approved):

eventualbuddha:
I changed this to using `move` instead because we might as well free up the original, but then realized that would mess with the way our fixtures and `MOCK_SCANNER_FILES` work. So I reverted back to the link-or-copy setup I had before ([Slack](https://votingworks.slack.com/archives/CEL6D3GAD/p1658173304906909?thread_ts=1657579298.722339&cid=CEL6D3GAD)).

PR #2165, opened 2022-07-20
refactor(precinct-scanner): Clean up card auth

jonahkagan:
## Overview
Two small changes related to card auth that I ran across:
- There are two unused app state fields probably related to an older iteration of card auth
- When writing the tally to the pollworker card, we used the `Card` interface directly instead of using the new smartcard auth interface

## Demo Video or Screenshot
N/A

## Testing Plan 
- Existing automated tests


arsalansufi (approved):
:100:

Issue #1959, opened 2022-06-15
Investigate spurious overvotes seen in m13


carolinemodic:

eventualbuddha:
Do we have the backups in Drive somewhere?

mattroe:
raising this discussion in #m13

eventualbuddha:
We don't have a backup from M13 that shows this, but I managed to reproduce it after just 3 tries locally with the Famous Names election (Lincoln Municipal General Election):

![localhost_3000_(Lenovo 14_)](https://user-images.githubusercontent.com/1938/178618180-544fcc5a-b9aa-4a7f-b449-a997f12ed30f.png)

Both contests in question are on the back. Here's the images:

| | Front | Back |
|-|-|-|
| **Original** | ![PIC-1657669701-001](https://user-images.githubusercontent.com/1938/178617835-cc5172d2-ed5c-4378-9886-9cda784bdceb.jpg) | ![PIC-1657669701-002](https://user-images.githubusercontent.com/1938/178617900-0ac8d470-48bf-4cb9-ae05-3c67569c0920.jpg) |
| **Normalized** | ![PIC-1657669701-001-0d14020e-4047-485b-919e-49412e56190d-normalized](https://user-images.githubusercontent.com/1938/178618026-7f2ed0c4-d005-4015-9b22-6ad16d10d063.jpg) | ![PIC-1657669701-002-0d14020e-4047-485b-919e-49412e56190d-normalized](https://user-images.githubusercontent.com/1938/178618045-91a6e943-2e23-4921-b5f7-7b2590c447c7.jpg) |


Here are the votes:

```json
{
    "board-of-alderman": [
        {
            "id": "steve-jobs",
            "name": "Steve Jobs",
            "partyIds": [
                "1"
            ]
        },
        {
            "id": "vincent-van-gogh",
            "name": "Vincent Van Gogh",
            "partyIds": [
                "1"
            ]
        },
        {
            "id": "pablo-picasso",
            "name": "Pablo Picasso",
            "partyIds": [
                "1"
            ]
        },
        {
            "id": "wolfgang-amadeus-mozart",
            "name": "Wolfgang Amadeus Mozart",
            "partyIds": [
                "2"
            ]
        },
        {
            "id": "write-in-0",
            "name": "Write-In",
            "isWriteIn": true
        },
        {
            "id": "write-in-1",
            "name": "Write-In",
            "isWriteIn": true
        },
        {
            "id": "write-in-2",
[...]

PR #2174, opened 2022-07-20
Fix/vx scan/stretching again

eventualbuddha:
## Overview
_I recommend reviewing commit-by-commit._

This change reverts the change that rejected ballots that were too stretched. Instead, we simply allow the "jiggling" to find the likely target to move up to 5px instead of just 2px. This was enough for the few tests I did, but is hopefully conservative enough not to introduce a new failure mode.

This is a more proper fix for https://github.com/votingworks/vxsuite/issues/1959.

## Demo Video or Screenshot
| Before | After |
|-|-|
| before | after |
| before-zoomed | after-zoomed |

## Testing Plan 
Automated tests, plus some manual testing.


eventualbuddha:
For posterity, here's an example of the debug images generated when processing the templates:
00-targets

PR #2176, opened 2022-07-20
refactor(precinct-scanner): Move getCvrsFromExport function

jonahkagan:
## Overview
This function didn't need to be a callback defined in AppRoot for any reason. In fact, it had extra guarding for the no-election-definition case even though it will only be called after configuration. I moved it into the pollworker screen component. More steps towards simplifying AppRoot so I can refactor it without breaking things.
## Demo Video or Screenshot
N/A

## Testing Plan 
- Existing automated tests

### DIFF
--- a/frontends/precinct-scanner/src/screens/poll_worker_screen.test.tsx
+++ b/frontends/precinct-scanner/src/screens/poll_worker_screen.test.tsx
@@ -6,6 +6,7 @@ import MockDate from 'mockdate';
 import React from 'react';
 import { InsertedSmartcardAuth } from '@votingworks/types';
 import { mocked } from 'ts-jest/utils';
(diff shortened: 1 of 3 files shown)


frontends/precinct-scanner/src/screens/poll_worker_screen.test.tsx
```diff
@@ -18,6 +19,7 @@ MockDate.set('2020-10-31T00:00:00.000Z');
 beforeEach(() => {
   jest.useFakeTimers();
   window.location.href = '/';
+  fetchMock.post('/scan/export', {});
```
jonahkagan, line 22: copied this from app.test.tsx

eventualbuddha (approved):

Issue #1615, opened 2022-03-21
Upgrade to cypress v10.2.0


carolinemodic:

eventualbuddha:
I tried the `Dockerfile` in this comment: https://github.com/cypress-io/cypress/issues/4478#issuecomment-1068916303. It did successfully build, though it's a newer version of cypress than we're using. In theory we can build an older version using something similar, or update our cypress version to use the latest. The question then remains how we want to set this up for local development. Could just be a build script that installs and configures it, though that's a decent amount of work.

carolinemodic:
This is solved by https://github.com/cypress-io/cypress/pull/22252, which I have tested locally, you can check out the pre-release binary following the steps in that PR. It will walk you through some automated conversion steps to upgrade to cypress v10 in order to work with the most recent code. If you are running in election-manager there are other issues with the plugins file where you have to change the imports to requires...

carolinemodic:
This is now released as of cypress v10.2.0

PR #2180, opened 2022-07-21
test(bmd): upgrade cypress

eventualbuddha:
## Overview
Also stops depending on the `/#demo` URL, which means we can remove `DemoApp` after this lands.

Refs #1615

## Demo Video or Screenshot

https://user-images.githubusercontent.com/1938/180103242-0bc79fa9-6a68-422c-b725-98c695bdea09.mov

## Testing Plan 
Automated testing.


eventualbuddha:
Looks like [I need to make sure `smartcards` has all of its dependencies](https://app.circleci.com/pipelines/github/votingworks/vxsuite/4769/workflows/ef28e984-fd63-47fe-90f1-549cda06bab9/jobs/100889?invite=true#step-110-69) set up first. @carolinemodic, how do you feel about this being in `bmd` rather than in a new `integration-testing` folder? I'm not tied to this approach, it was just the easiest way to go.

carolinemodic:
I think eventually we'll want to move over to the integration testing folder but it doesn't need to be tied to this PR. We still need to upgrade the cypress version in those tests too right?

eventualbuddha:
Yeah, this doesn't touch those. But they should be pretty straightforward. Some of the changes in this PR were actually made by cypress.

eventualbuddha:
This is ready to review if anyone wants to take a look. Automerge is on.

carolinemodic (approved):

Issue #1862, opened 2022-05-23
VxSuite: make reboot from USB fully resilient


benadida:
Currently, "Reboot from USB" on our machines may not be able to detect all USB sticks.

For upgrading m13b to m14, we have the fallback that we can use the grub prompt. We will *not* have this once we have the full secure-boot setup, where grub is no longer part of the picture. That means our "reboot from USB" functionality needs to be particularly resilient so we can always make it work.

The work here may involve simply specifying exactly how to set up a USB stick that works for "Reboot from USB", or improve "Reboot from USB" in how it detects USB sticks to boot from, or a combination thereof.

We also want "Reboot into BIOS".

benadida:
Done by:
https://github.com/votingworks/kiosk-browser/pull/127
and
https://github.com/votingworks/kiosk-browser/pull/126

benadida:
Actually, one more PR to merge, leaving this open for now:
https://github.com/votingworks/vxsuite/pull/2190

benadida:
OK, now done :)

PR #2190, opened 2022-07-23
Reboot to BIOS

benadida:
As part of #1862, add an option to reboot into BIOS on the superadmin screen.

### DIFF
--- a/frontends/bmd/src/pages/superadmin_screen.test.tsx
+++ b/frontends/bmd/src/pages/superadmin_screen.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { fireEvent, screen } from '@testing-library/react';
+import { fireEvent, screen, waitFor } from '@testing-library/react';
 import { fakeKiosk } from '@votingworks/test-utils';
 
 import { usbstick } from '@votingworks/utils';
@@ -11,7 +11,7 @@ import { SuperAdminScreen } from './superadmin_screen';
 
 const logger = new Logger(LogSource.VxBallotMarkingDeviceService);
 
-it('the right buttons and reset calls quit', () => {
+it('the right buttons and reset calls quit', async () => {
   render(
     <SuperAdminScreen
       usbDriveStatus={usbstick.UsbDriveStatus.absent}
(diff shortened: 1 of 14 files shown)


libs/test-utils/src/fake_kiosk.ts
```diff
@@ -99,6 +99,8 @@ export function fakeKiosk({
     sign: jest.fn(),
     log: jest.fn(),
     reboot: jest.fn(),
+    // eslint-disable-next-line vx/gts-identifiers
+    rebootToBIOS: jest.fn(),
```
jonahkagan, line 103: Nit: I know that it's weird to follow the camel case convention for acronyms, but the rest of the code does it (e.g. see `Usb` below and `Ipp` elsewhere), so I don't know that there's a strong reason to make an exception here.

frontends/bmd/src/pages/superadmin_screen.test.tsx
```diff
@@ -20,14 +20,26 @@ it('the right buttons and reset calls quit', () => {
...
 
   // test without kiosk
   fireEvent.click(screen.getByText('Reset'));
 
   // test with kiosk
   window.kiosk = fakeKiosk();
-  fireEvent.click(screen.getByText('Reset'));
-
-  expect(window.kiosk.quit).toHaveBeenCalledTimes(1);
+  act(() => {
```
jonahkagan, line 31: I think `act` is unnecessary here. Was there a specific reason for adding it?

jonahkagan (approved):

libs/test-utils/src/fake_kiosk.ts
benadida: grumble... you are right, even as I find this particular decision annoying :)

frontends/bmd/src/pages/superadmin_screen.test.tsx
benadida: I'm pretty sure there was a warning until I added it, but I'll check.

benadida: huh, I was convinced it asked for it, but turns out it didn't, you are right. Removing.

PR #2202, opened 2022-07-26
fix(ps): sync cvr export to disk

### LINKED ISSUE #1981: VxScan: no CVR saved on polls closed
This issue was observed during L&A in Warren and reproduced on a locked-down production imaged VxScan.

CVR files are not saved to USB disk after polls are closed, but they are still saved if you either 1) power down the machine before ejection or 2) export CVRs through the admin card screen.

### Repro steps
1. Scan ballots
2. Insert poll worker card and close polls
3. Remove USB
4. No CVR files will be saved

adghayes:
## Overview
Closes #1981. When saving CVRs - whether on polls closed or when exporting from the Admin menu - forces the machine to finish writing the file to disk by calling the `sync` command. This prevents the CVR file from not being saved or being saved improperly. It does slightly increase the delay when closing polls, by a couple of seconds, but that tradeoff is easily worth it.

Depends on [votingworks/kiosk-browser#129](https://github.com/votingworks/kiosk-browser/pull/129) to introduce the `kiosk.syncUsbDrive` API. 

Sidebar: the precinct scanner poll worker flow lacks any logging. Not sure if that's intentional, but I didn't touch it in this PR. 

## Testing Plan
- Manual testing on hardware
- Unit test for `usbstick` util

## Checklist

### DIFF
--- a/libs/utils/src/usbstick.ts
+++ b/libs/utils/src/usbstick.ts
@@ -61,3 +61,15 @@ export async function doUnmount(): Promise<void> {
   await window.kiosk.unmountUsbDrive(device.deviceName);
   return await sleep(FLUSH_IO_DELAY_MS);
 }
+
+// Triggers linux 'sync' command which forces any cached file data to be
+// flushed to the removable drive. Used to prevent incomplete file transfers.
+export async function doSync(): Promise<void> {
+  const device = await getDevice();
+  const mountPoint = device?.mountPoint;
+  if (!mountPoint) {
+    return;
+  }
+  assert(window.kiosk);
+  await window.kiosk.syncUsbDrive(mountPoint);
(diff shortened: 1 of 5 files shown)


libs/utils/src/usbstick.ts
```diff
@@ -61,3 +61,13 @@ export async function doUnmount(): Promise<void> {
   await window.kiosk.unmountUsbDrive(device.deviceName);
   return await sleep(FLUSH_IO_DELAY_MS);
 }
+
+export async function doSync(): Promise<void> {
```
jonahkagan, line 65: Can you add a comment explaining what this does and why it's necessary?

jonahkagan (approved):

PR #2218, opened 2022-07-27
fix(scan): Ignore extraneous commands to precinct scanner state machine

jonahkagan:
## Overview
This prevents race conditions when the frontend has an outdated scanner status and tries to send a command. E.g. if the frontend thinks the scanner is in `ready_to_scan` state but the paper got quickly removed and it's back to `no_paper`, it might still try to send a `SCAN` command.

This also allows the frontend to not have to worry about sending commands multiple times, which makes the frontend logic much simpler.

## Demo Video or Screenshot
N/A

## Testing Plan 
Manual testing

### DIFF
--- a/frontends/precinct-scanner/src/app.test.tsx
+++ b/frontends/precinct-scanner/src/app.test.tsx
@@ -417,7 +417,7 @@ test('admin and pollworker configuration', async () => {
   await authenticateAdminCard();
 
   // Calibrate scanner
-  fetchMock.postOnce('/scanner/calibrate', { body: { status: 'ok' } });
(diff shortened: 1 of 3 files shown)


benadida (approved):

cjerdonek:
Would it be better at least to log the command that's not being handled? I say this because I wonder if there are any scenarios where ignoring the command could wind up hiding / masking something not behaving correctly.

jonahkagan:
Yeah, haven't set up logging for any of this new code yet. Will definitely think about how to cover this case. Thanks!

PR #2220, opened 2022-07-27
fix(scan): Interpret blank BMD pages correctly for precinct scanner

jonahkagan:
## Overview
There was a special case to catch blank pages of a BMD ballot (i.e. the back of the ballot) in the interpret_vx worker that wasn't included in the new simple interpreter used by the precinct scanner. I moved the logic into the interpreter itself to make sure both central and precinct scanners can use the same logic.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Scanned a BMD ballot
- Added a new test

### DIFF
--- a/services/scan/src/interpreter.ts
+++ b/services/scan/src/interpreter.ts
@@ -63,7 +63,7 @@ interface BallotImageData {
 export async function getBallotImageData(
   file: Buffer,
   filename: string,
-  detectQrcodeResult: qrcodeWorker.Output
+  detectQrcodeResult: qrcodeWorker.NonBlankPageOutput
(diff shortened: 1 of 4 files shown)


services/scan/src/interpreter.ts
```diff
@@ -63,7 +63,7 @@ interface BallotImageData {
 export async function getBallotImageData(
   file: Buffer,
   filename: string,
-  detectQrcodeResult: qrcodeWorker.Output
+  detectQrcodeResult: qrcodeWorker.NonBlankPageOutput
```
jonahkagan, line 66: this function only gets called on L238, so I was able to narrow the type after adding the new guard on L234

benadida (approved):

PR #2221, opened 2022-07-27
Save All Ballot PDFs

### LINKED ISSUE #2146: VxAdmin: make it easier for customers to save ballot PDFs to send to third party printers
## Summary
Right now, there are two ways to export PDFs of ballots to send to a third party printer and they both have usability issues:
1. Exporting ballot package: the zip of all ballot styles includes test & live mode ballots, as well as extraneous information for customers such as the layout .json files.
2. Saving ballot PDFs from the ballot proofing/printing view: this requires that a user save ballot PDFs individually by ballot style, which can be overly onerous/error prone requiring dozens of individual downloads.

The priority of this issue depends on whether customers print ballots with a third party printer or on-demand via VxAdmin for m14 (which is also dependent on m14 paper weight testing).

## Requirements
An administrator should be able to export to a USB drive the PDFs of all ballot styles for a given ballot mode in a one-click manner.

adghayes:
## Overview
Closes #2146. Introduces `Export Ballot PDFs` button on the Ballots tab that allows an election administrator to export all the PDFs of a given type & mode at once. The export also includes a `.csv` metadata file that they could edit and use to send a job to their printer.

## Demo Video or Screenshot
[export-ballot-pdfs-rev2.webm](https://user-images.githubusercontent.com/37960853/181384963-c036bbbf-70b5-407d-8fd0-ffabb29d32b1.webm)

## Testing Plan
- Unit testing of new utility function
- Button / Modal Component tests

There are two additional things that would be nice to test but the effort is out of the scope of this PR, unless someone has a better suggestion:
1. Check that the ballot PDFs exported as part of this function match the printed ballots. Ideally we would be generating the ballots for each in a shared place, but that would require a deep dive into changing how we print / PDFize our documents in general. And very ideally, we would be checking that these PDFs work are accepted by our scanning service.
2. Check that the zip archive contains the file entries it is supposed to. Doing this would require an actual file download, meaning Cypress, but Cypress can't use the feature because AFAIK we can't mock USB drive detection in Cypress.

## Checklist

### DIFF
--- a/frontends/election-manager/src/components/export_ballot_pdfs_button.test.tsx
+++ b/frontends/election-manager/src/components/export_ballot_pdfs_button.test.tsx
@@ -1,0 +1,162 @@
+import { screen, within } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { fakeLogger, LogEventId } from '@votingworks/logging';
+import { fakeKiosk, fakeUsbDrive } from '@votingworks/test-utils';
+import { assert, usbstick } from '@votingworks/utils';
+import React from 'react';
+import { fakeFileWriter } from '../../test/helpers/fake_file_writer';
+import { renderInAppContext } from '../../test/render_in_app_context';
+import { ExportBallotPdfsButton } from './export_ballot_pdfs_button';
+
+const { UsbDriveStatus } = usbstick;
+
+jest.mock('../components/hand_marked_paper_ballot');
+
+beforeEach(() => {
+  const mockKiosk = fakeKiosk();
+  mockKiosk.getUsbDrives.mockResolvedValue([fakeUsbDrive()]);
+  const fileWriter = fakeFileWriter();
+  mockKiosk.saveAs = jest.fn().mockResolvedValue(fileWriter);
+  mockKiosk.writeFile = jest.fn().mockResolvedValue(fileWriter);
+  window.kiosk = mockKiosk;
+});
+
+afterEach(() => {
+  delete window.kiosk;
+});
+
+test('button renders properly when not clicked', () => {
+  renderInAppContext(<ExportBallotPdfsButton />);
+
+  expect(screen.queryByText('Export Ballot PDFs')).toHaveProperty(
+    'type',
+    'button'
+  );
+  expect(screen.queryByRole('alertdialog')).toBeNull();
+});
+
+test('modal renders insert usb screen appropriately', async () => {
+  const usbStatuses = [
+    UsbDriveStatus.absent,
+    UsbDriveStatus.recentlyEjected,
+    UsbDriveStatus.notavailable,
+  ];
+
+  for (const usbStatus of usbStatuses) {
+    const { unmount } = renderInAppContext(<ExportBallotPdfsButton />, {
+      usbDriveStatus: usbStatus,
+    });
+    userEvent.click(screen.getByText('Export Ballot PDFs'));
+
+    const modal = await screen.findByRole('alertdialog');
+    within(modal).getByText('No USB Drive Detected');
+    within(modal).getByAltText('Insert USB Image');
+    userEvent.click(within(modal).getByText('Cancel'));
+
+    expect(screen.queryByRole('alertdialog')).toBeNull();
+    unmount();
+  }
+});
+
+test('modal renders loading screen when usb drive is mounting or ejecting', async () => {
+  const usbStatuses = [UsbDriveStatus.present, UsbDriveStatus.ejecting];
+
+  for (const usbStatus of usbStatuses) {
+    const { unmount } = renderInAppContext(<ExportBallotPdfsButton />, {
+      usbDriveStatus: usbStatus,
+    });
+    userEvent.click(screen.getByText('Export Ballot PDFs'));
+    const modal = await screen.findByRole('alertdialog');
+    within(modal).getByText('Loading');
+
+    unmount();
+  }
+});
+
+test('modal happy path flow works', async () => {
+  const logger = fakeLogger();
+  const ejectFunction = jest.fn();
+  renderInAppContext(<ExportBallotPdfsButton />, {
+    logger,
+    usbDriveStatus: UsbDriveStatus.mounted,
+    usbDriveEject: ejectFunction,
+  });
+
+  userEvent.click(screen.getByText('Export Ballot PDFs'));
+  const modal = await screen.findByRole('alertdialog');
+  userEvent.click(within(modal).getByText('Export'));
+  await within(modal).findByText('Generating Ballot PDFs…');
+  await within(modal).findByText('Finishing Export…');
+  await within(modal).findByText('Export Complete');
+
+  assert(window.kiosk);
+  expect(window.kiosk.makeDirectory).toHaveBeenCalledTimes(1);
+  expect(window.kiosk.writeFile).toHaveBeenCalledTimes(1);
+  expect(window.kiosk.writeFile).toHaveBeenCalledWith(
+    'fake mount point/ballot-pdfs/ballot-pdfs-election-b1e83122e8-live.zip'
+  );
+  expect(logger.log).toHaveBeenCalledWith(
+    LogEventId.FileSaved,
+    'admin',
+    expect.objectContaining({ disposition: 'success' })
+  );
+
+  expect(modal).toBeInTheDocument();
+  userEvent.click(within(modal).getByText('Eject USB'));
+  expect(ejectFunction).toHaveBeenCalledTimes(1);
+  userEvent.click(within(modal).getByText('Close'));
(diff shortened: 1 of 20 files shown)


adghayes:
I recommend commit-by-commit review on this PR.

frontends/election-manager/src/utils/generate_pdf_export_metadata_csv.ts
```diff
@@ -0,0 +1,63 @@
...
+  isAbsentee: boolean;
+  ballotLocales: BallotLocales;
+}): string {
+  const { election, electionHash } = electionDefinition;
+  const ballotStyleList = sortBallotStyleDataByPrecinct(
+    election,
+    getBallotStylesData(election)
+  );
+
+  let finalDataString = headers.join(', ');
```
arsalansufi, line 43: I believe CSVs aren't expected to have spaces after the commas that separate entries ([relevant link](https://discuss.codecademy.com/t/in-a-csv-file-does-spacing-matter-before-and-after-commas/353056))

frontends/election-manager/src/components/export_ballot_pdfs_button.tsx
```diff
@@ -0,0 +1,375 @@
...
+  text-align: center;
+`;
+
+enum ModalState {
+  BEGIN_EXPORT = 'begin export',
+  GENERATE_FILES = 'generate files',
+  END_EXPORT = 'end export',
+  DONE = 'done',
+  ERROR = 'error',
+}
```
arsalansufi, line 60: Same comment here as earlier about unions of string literals vs. enums. I used a union of string literals for a very similar purpose (modal states) back when we were working on L&A!

https://github.com/votingworks/vxsuite/blob/3df0503a5e13fcca49f3b4036bbb92787025393b/frontends/election-manager/src/screens/print_test_deck_screen.tsx#L333-L337

frontends/election-manager/src/components/export_ballot_pdfs_button.tsx
```diff
@@ -0,0 +1,375 @@
...
+  const { electionDefinition, usbDriveStatus, usbDriveEject, auth, logger } =
+    useContext(AppContext);
+  assert(electionDefinition);
+  assert(isAdminAuth(auth) || isSuperadminAuth(auth));
+  const userRole = auth.user.role;
+  const { election, electionHash } = electionDefinition;
+
+  const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
+  const [modalState, setModalState] = useState(ModalState.BEGIN_EXPORT);
+  const [modalError, setModalError] = useState<Error | undefined>(undefined);
```
arsalansufi, line 74: I believe you can simplify this line to:

```ts
const [modalError, setModalError] = useState<Error>();
```

It'll infer from this that the type can be `Error` or `undefined`

frontends/election-manager/src/components/export_ballot_pdfs_button.tsx
```diff
@@ -0,0 +1,375 @@
...
+  const [archive, setArchive] = useState(new DownloadableArchive());
+  const [ballotMode, setBallotMode] = useState(BallotMode.Official);
+  const [isAbsentee, setIsAbsentee] = useState(false);
+  const [ballotIndex, setBallotIndex] = useState(0);
+
+  const defaultArchiveFilename = useMemo<string>(
+    () => getBallotArchiveFilename(electionHash, ballotMode, isAbsentee),
+    [electionHash, ballotMode, isAbsentee]
+  );
```
arsalansufi, line 83: For simpler computations like the one in `getBallotArchiveFilename`, I lean toward not using `useMemo`. `useMemo` has overhead of its own that can end up negating performance benefits for simple computations, and it can also make code harder to read

frontends/election-manager/src/components/export_ballot_pdfs_button.tsx
```diff
@@ -0,0 +1,375 @@
...
+          <h1>Export Failed</h1>
+          <p>An error occurred: {modalError && modalError.message}.</p>
+        </Prose>
+      );
+      actions = <LinkButton onPress={closeModal}>Close</LinkButton>;
+      break;
+    }
+
+    default:
+      // nothing to do
```
arsalansufi, line 357: To make sure that you don't forget to cover any states in the `switch` `case`, adding a `throwIllegalValue` call here will make TS ensure that you've covered everything:

https://github.com/votingworks/vxsuite/blob/3df0503a5e13fcca49f3b4036bbb92787025393b/frontends/election-manager/src/components/smartcard_modal/status_message.tsx#L45-L48

frontends/election-manager/src/components/export_ballot_pdfs_button.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+    UsbDriveStatus.recentlyEjected,
+    UsbDriveStatus.notavailable,
+  ];
+
+  for (const usbStatus of usbStatuses) {
+    const { unmount, getByText, queryAllByAltText, queryAllByTestId } =
+      renderInAppContext(<ExportBallotPdfsButton />, {
+        usbDriveStatus: usbStatus,
+      });
+    fireEvent.click(getByText('Export Ballot PDFs'));
```
arsalansufi, line 48: Same comment here as earlier about `userEvent` vs. `fireEvent`

frontends/election-manager/src/components/export_ballot_pdfs_button.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+test('modal renders insert usb screen appropriately', async () => {
+  const usbStatuses = [
+    UsbDriveStatus.absent,
+    UsbDriveStatus.recentlyEjected,
+    UsbDriveStatus.notavailable,
+  ];
+
+  for (const usbStatus of usbStatuses) {
+    const { unmount, getByText, queryAllByAltText, queryAllByTestId } =
```
arsalansufi, line 44: Instead of using the query methods returned by the `render` call, it's recommended to use the general query methods on the `screen` object ([relevant link](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen))

frontends/election-manager/src/components/export_ballot_pdfs_button.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+      renderInAppContext(<ExportBallotPdfsButton />, {
+        usbDriveStatus: usbStatus,
+      });
+    fireEvent.click(getByText('Export Ballot PDFs'));
+    await waitFor(() => getByText('No USB Drive Detected'));
+    expect(queryAllByAltText('Insert USB Image')).toHaveLength(1);
+    expect(queryAllByTestId('modal')).toHaveLength(1);
```
arsalansufi, line 51: Instead of using a test ID (which is viewed as a last resort when no other queries will find what you're looking for), you can usually detect modals using a role query:

https://github.com/votingworks/vxsuite/blob/3df0503a5e13fcca49f3b4036bbb92787025393b/frontends/election-manager/src/components/smartcard_modal/smartcard_modal.test.tsx#L231

frontends/election-manager/src/components/export_ballot_pdfs_button.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+    fireEvent.click(getByText('Cancel'));
```
arsalansufi, line 53: Related to the above, once I find a modal and want to interact with items in it, I like using `within` for clarity:

https://github.com/votingworks/vxsuite/blob/3df0503a5e13fcca49f3b4036bbb92787025393b/frontends/election-manager/src/components/smartcard_modal/smartcard_modal.test.tsx#L232

arsalansufi (commented):
This looks great! Thank you for the detailed walkthrough video!

frontends/election-manager/src/config/types.ts
```diff
@@ -36,6 +36,12 @@ export const PrintableBallotType = {
 export type PrintableBallotType =
   typeof PrintableBallotType[keyof typeof PrintableBallotType];
 
+export enum BallotMode {
+  Official = 'live',
+  Test = 'test',
+  Sample = 'sample',
+}
```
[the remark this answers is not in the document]
adghayes: Thank you for the suggestion on unions, I was wondering which one would be better! 

I think I may want to make an exception here on the renaming. The literal value has to be 'live' in filenames, otherwise we might break the ballot package interpretation in `bsd`. However, we use `Official` in the UI so I think that it makes the code clearer. And with the enum, if we wanted to auto generate button text off of the enum, we could take the key value.

Thoughts?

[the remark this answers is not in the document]
arsalansufi: Got it! I had a feeling there might be a reason for the divergence - in that case, keeping `Official = 'live'` seems reasonable to me!

PR #2228, opened 2022-07-28
feat(scan): Reconnect to Plustek after accept

jonahkagan:
## Overview
The goal is to avoid crashes from long-running plustekctl processes.

## Demo Video or Screenshot
N/A

## Testing Plan 
Manual test


jonahkagan:
Not pursuing this route, it doesn't seem to help with crashes we've seen

PR #2232, opened 2022-07-29
BMD Text to Speech via kiosk-browser

adghayes:
## Checklist


adghayes:
Text-to-speech works again with this PR, but I need some feedback! I have not written testing yet.

@benadida I'm looking for feedback on a few parts of this change in particular:

- In the newly implemented `KioskTextToSpeech` I have to do a little bit of complicated promise management. Because `SpeechSynthesis` is not managing the queue of utterances anymore, we have to do it ourselves. But in order to keep `speak`'s behavior such that it only resolves when speaking is done, we have to maintain a queue of promises and resolve them only after that particular speaking is done. Got the idea from [this article](https://medium.com/@karenmarkosyan/how-to-manage-promises-into-dynamic-queue-with-vanilla-javascript-9d0d1f8d4df5). Let me know if it seems somehow problematic.
- Because muting now requires a call to `kiosk-browser` which is asynchronous, all actions that trigger muting are now asynchronous as well. I did some type rearrangements to make that work. An alternative is to make the `kiosk-browser` API synchronous, but I didn't want to synchronously wait for a child process.
- I've removed most of the SpeechSynthesis content. It seems like we were working hard to let that load properly, and getting rid of it removes a good deal of code. Let me know if anything should be kept.
- Originally there was a [call to `mute` in `App`'s renderer function](https://github.com/votingworks/vxsuite/blob/e052702bffb624b9ebce577d5970b42403815bff/frontends/bmd/src/app.tsx#L56), I'm not sure if that had a particular motivation.

frontends/bmd/src/utils/ScreenReader/kiosk_text_to_speech.ts
```diff
@@ -0,0 +1,142 @@
...
+  ): Promise<void> {
+    if (this.isMuted()) {
+      return;
+    }
+
+    if (now) {
+      await this.stop();
+    }
+
+    await this.enqueueSpeech(utterance);
```
benadida, line 37: I see why you chose to do this, which is to keep supporting the idea that `speak()` supports the idea of queuing up things to speak... but I'm torn because this adds a good bit of complexity and state for a feature that I don't think we ever use, nor do I anticipate we will want to use it – I can't think of a use case in our context.

If we want to support this feature, I think it's best supported in `kiosk-browser`, where we would pass the `now` argument, and implement it there. And for now, I would suggest we simply error out on `now=false` and say "not supported yet." We leave ourselves the API opportunity in case we do find a use case for it, but we don't take on the complexity and associated bugs (there will be bugs!)

benadida (changes requested):
- recommend simplifying the enqueuing, I don't think we need it.
- can we keep the other speech synthesis, and use it when we don't have `window.kiosk`? This will make testing changes to audio content in a web browser easier.

adghayes:
Closed in favor of #2246

PR #2233, opened 2022-07-29
test(scan): Basic tests for precinct scanner service

jonahkagan:
## Overview
Some basic tests for the precinct scanner service. The tests use the API as the interface and use the Plustek client mock to simulate the Plustek unit.

Currently they use real interpretation, which makes them very slow.


services/scan/src/precinct_scanner_app.test.ts
```diff
@@ -0,0 +1,745 @@
+import { MockScannerClient, ScannerClient } from '@votingworks/plustek-sdk';
```
jonahkagan, line 1: This is the main change. Github collapsed it by default so commenting to ensure you don't miss it

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -560,9 +562,17 @@ function buildMachine(createPlustekClient: CreatePlustekClient) {
...
+            actions: clearLastScan,
+          },
           // If the paper didn't get dropped, it's an error
-          SCANNER_READY_TO_EJECT: 'error_accepting',
+          SCANNER_READY_TO_EJECT: {
+            target: 'error_accepting',
+            actions: assign({
+              error: new PrecinctScannerError('paper_in_back_after_accept'),
+            }),
+          },
```
jonahkagan, line 575: making sure the API returns a reason for this error (before it was `undefined`)

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -560,9 +562,17 @@ function buildMachine(createPlustekClient: CreatePlustekClient) {
         on: {
           WAIT_FOR_PAPER: 'no_paper',
           SCANNER_NO_PAPER: { target: 'accepted', internal: true },
-          SCANNER_READY_TO_SCAN: 'ready_to_scan',
+          SCANNER_READY_TO_SCAN: {
+            target: 'ready_to_scan',
+            actions: clearLastScan,
```
jonahkagan, line 567: making sure we don't keep the interpretation from the last scan hanging around

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -795,15 +805,35 @@ export function createPrecinctScannerStateMachine(
         }
       })();
       const { error, interpretation } = state.context;
+      // Remove interpretation details
```
jonahkagan, line 808: the frontend doesn't actually need to know the raw interpretation data, just the `type` and associated `reason`/`reasons`

benadida (approved):

PR #2235, opened 2022-07-29
fix(plustek): turn off swdeskew

benadida:
which we suspect has a race condition causing occasional segfaults.


mattroe (approved):

benadida:
Further testing in a production build appears to confirm, so far, that this solves the issues, without any obvious regressions.

PR #2238, opened 2022-08-01
feat(quickresults): make QuickResults configurable in election definition

benadida:
Two needs met here:
- we want to be able to turn off VxQR
- we want to make the VxQR results reporting server configurable

These two needs can be elegantly met by an optional configuration parameter in the election definition that points to a quickresults reporting server. If that option is blank, VxQR is turned off by default. It can easily be turned on by supplying `https://results.voting.works`, our current results reporting server.

This gives us the VxQR configurability we want, without needing a build-time env variable.


libs/types/src/election.ts
```diff
@@ -565,6 +566,7 @@ export const ElectionSchema: z.ZodSchema<Election> = z
       .array(z.lazy(() => AdjudicationReasonSchema))
       .optional(),
     precincts: PrecinctsSchema,
+    quickResultsReportingUrl: z.string().nonempty().optional(),
```
jonahkagan, line 569: Would it be useful to use `z.url()`?

libs/ui/src/precinct_scanner_tally_qrcode.tsx
```diff
@@ -58,7 +58,7 @@ export function PrecinctScannerTallyQrCode({
         }
 
         setResultsReportingUrl(
-          `https://results.voting.works/?p=${encodeURIComponent(
+          `${election.quickResultsReportingUrl}/?p=${encodeURIComponent(
```
jonahkagan, line 61: Any concern here about the URL having a trailing slash?

jonahkagan (commented):
Overall approach seems fine to me! Got some tests failing still

libs/ui/src/precinct_scanner_tally_qrcode.tsx
benadida: I don't think I want us to try to be too smart about correcting this... maybe we issue a warning?

libs/types/src/election.ts
benadida: could be yes! I'll look into it.

libs/ui/src/precinct_scanner_tally_qrcode.tsx
benadida: or maybe we make it part of `zod` constraints?

benadida:
@jonahkagan I had to add quite a bit in terms of fixtures for testing purposes, so throwing it back to you for a quick look, and if you have any other thoughts about the trailing slash.

libs/fixtures/src/data/electionMinimalExhaustiveSampleWithReportingUrl/electionMinimalExhaustiveSample.json.ts
```diff
@@ -0,0 +1,61 @@
+/* Generated by res-to-ts. DO NOT EDIT */
```
jonahkagan, line 1: Would it be possible to just change this fixture to have results reporting turned on and use a different fixture for results reporting turned off? Or did you need to add a new election fixture just to distinguish these two cases?

libs/ui/src/precinct_scanner_tally_qrcode.tsx
jonahkagan: I like the idea of making it part of the zod validation if that's easy!

libs/fixtures/src/data/electionMinimalExhaustiveSampleWithReportingUrl/electionMinimalExhaustiveSample.json.ts
benadida: I didn't want to do that and risk changing the hash of that election, with all the potential downstream effects.

jonahkagan: Gotcha, that makes sense, though I would hope our tests aren't dependent on the exact hash of an election definition. That would make them particularly brittle.

The reason I ask is because it seems like it could get pretty unmanageable to add a different fixture for every time we want to test one field change of an election definition.

libs/ui/src/precinct_scanner_tally_qrcode.tsx
benadida: done!

libs/fixtures/src/data/electionMinimalExhaustiveSampleWithReportingUrl/electionMinimalExhaustiveSample.json.ts
benadida: ok, I will explore if I can make this change safely

benadida: sadly, it does cause other tests to fail, because we include the election hash in a number of directory structures. I don't know that that makes our tests brittle so much as we have inherent constraints given election hashes.

On the whole, my sense is we're better off adding that new election with reporting URL, since it is not longer the default, but I welcome pushback.

jonahkagan: My only opinion is that we should be able to make changes to test fixtures without breaking tests (unless we actually make a logical change that breaks the behavior that the test is testing). I think that should be possible even with strict election hash checking.

Maybe you don't want to get blocked on that in this PR, which is reasonable, and I'm happy to approve.

jonahkagan (approved):

libs/fixtures/src/data/electionMinimalExhaustiveSampleWithReportingUrl/electionMinimalExhaustiveSample.json.ts
benadida: @jonahkagan I don't think we have that capability right now, given how tightly everything is bound to the election definition. It would be a significant effort to change this. I'll file a ticket for us to consider:

https://github.com/votingworks/vxsuite/issues/2247

jonahkagan: Sounds good! Thanks

PR #2245, opened 2022-08-02
refactor(scan): Combine page interpretations in SimpleInterpreter

jonahkagan:
## Overview
Instead of combining the interpretations for the two pages of a ballot in the state machine module, we do it behind the interface of the interpreter. This will allow easier mocking of the interpreter in tests.

I also changed the types so that the state machine event type isn't reused as the interpretation result type. Instead, I created a new type to represent the interpretation of a ballot sheet.

## Demo Video or Screenshot
N/A

## Testing Plan 
Existing tests

### DIFF
--- a/services/scan/src/simple_interpreter.ts
+++ b/services/scan/src/simple_interpreter.ts
@@ -1,5 +1,7 @@
 import { interpret as interpretNh } from '@votingworks/ballot-interpreter-nh';
 import {
+  AdjudicationReason,
+  AdjudicationReasonInfo,
   BallotPageLayoutWithImage,
   ElectionDefinition,
   Id,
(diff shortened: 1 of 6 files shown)


benadida (approved):

services/scan/src/simple_interpreter.ts
```diff
@@ -24,14 +28,119 @@ export interface CreateInterpreterOptions {
...
+  }
+
+  if (frontType === 'UnreadablePage' || backType === 'UnreadablePage') {
+    return {
+      type: 'InvalidSheet',
+      reason: 'unreadable',
+    };
+  }
+
+  // TODO what does this case actually mean?
```
eventualbuddha, line 89: This means we found the QR code that tells us it's a HMPB, but we were not able to actually interpret the page layout.

Issue #1484, opened 2022-03-04
VxMark: headphone volume control


mcchilders:
## Background

For any number of reasons, the headphone volume on the VxMark might be wrong. Right now, we don't provide any means of adjusting it. We should, and we have a button on the accessible controller that we can use.

## Expected Flow

- voter presses the single volume button on the accessible controller
- volume increases by a notch, with 9 levels from 20db to 100db in 10db increments
- on dispay, show "Volume 4/9" (in some kind of modal display that fades away after 3-4s.)
- audio interface says "volume set to 4 out of 9."
- if the volume is already at 9, then pressing the button cycles back to 1/9.

## Suggested Architecture

- do some calibration of volume on our headphones + hardware to see where 20db and 100db land.
- add `kiosk.volume.get()` and `kiosk.volume.set()` 
- intercept the appropriate function key that corresponds to the volume-button press and do the right thing using the APIs above.

mattroe:
Relevant VVSG requirement for reference:
> 7.1-K – Audio settings
> 2. The default volume for each voting session is set between 60 and 70 dB SPL.
> 3. The volume is adjustable from a minimum of 20 dB SPL up to a maximum of 100 dB SPL, in increments no greater than 10 dB.

adghayes:
I'm going to close this for now because we are meting _most_ of the requirements here. We're missing:
- Display feedback about audio level
- Decibel calibration:
> I don't have the volumes mapped to decibel levels. I'm yet quite sure how to do that, or if the choice of headphones affects that. But the VVSG range is, I think, an absolute maximum and minimum - not the maximum and minimum we have to use. We can't go outside of the range they specify. Because the VVSG maximum - 100dB - is equivalent to a jet plane taking off or a jackhammer. Perhaps I am misunderstanding something.

Since we are planning to revisit this to add a settings menu of sorts anyway, I am closing this issue and allowing future issues to be represented by other tickets. cc @mattroe @benadida

PR #2246, opened 2022-08-02
Fix BMD Audio + Add Volume Button

### LINKED ISSUE #2117: VxMark: no audio in headphones
## Description
There is no audio when using headphones on a locked down production VxMark [using this image](https://s3.console.aws.amazon.com/s3/object/votingworks-machine-images?region=us-west-2&prefix=2022-07-nh/2022-07-08-vxmark.img.lz4) for the NH demo.

adghayes:
## Overview
Closes #2117 and makes an initial pass at #1484. Split into three commits:
1. Updates `KioskBrowser` interfaces to include the `speak` APIs introduced in [votingworks/kiosk-broswer#130](https://github.com/votingworks/kiosk-browser/pull/130)
2. Adds `KioskTextToSpeech` class which uses the new `kiosk-browser` APIs, and is active when `window.kiosk` is present. This fixes BMD audio and closes #2117.
3. Adds a preliminary volume control feature. The single volume button on the audio controller will change the volume up and down, cycling. An initial pass at #1484.

I don't have the volumes mapped to decibel levels. I'm yet quite sure how to do that, or if the choice of headphones affects that. But the VVSG range is, I think, an absolute maximum and minimum - not the maximum and minimum we have to use. We can't go outside of the range they specify. Because the VVSG maximum - 100dB - is equivalent to a jet plane taking off or a jackhammer. Perhaps I am misunderstanding something. 

In any case, for this first pass, I think we should just settle on reasonable defaults. In this implementation, there are 10 possible audio levels (1 of 10 to 10 of 10) mapping to an underlying audio setting of 28% to 100% volume. That underlying range is what roughly I think a user might need, and it led to pretty-ish numbers.

## Testing Plan 
- Unit tests for `KioskTextToSpeech`
- Test for the keypress detection in `app.test.tsx`

## Checklist

### DIFF
--- a/frontends/bmd/src/app.test.tsx
+++ b/frontends/bmd/src/app.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, screen } from '@testing-library/react';
+import { fireEvent, screen, waitFor } from '@testing-library/react';
 import { fakeKiosk, makePollWorkerCard } from '@votingworks/test-utils';
 import { MemoryCard, MemoryHardware, MemoryStorage } from '@votingworks/utils';
 import fetchMock from 'fetch-mock';
@@ -8,11 +8,13 @@ import {
   setStateInStorage,
 } from '../test/helpers/election';
 import { fakeMachineConfigProvider } from '../test/helpers/fake_machine_config';
+import { fakeTts } from '../test/helpers/fake_tts';
 import { advanceTimersAndPromises } from '../test/helpers/smartcards';
 import { render } from '../test/test_utils';
 import { App } from './app';
 import { PrecinctSelectionKind, MarkAndPrint } from './config/types';
 import { electionSampleDefinition } from './data';
+import { AriaScreenReader } from './utils/ScreenReader';
 
 beforeEach(() => {
   jest.useFakeTimers();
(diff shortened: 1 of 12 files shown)


frontends/bmd/src/app.tsx
```diff
@@ -76,11 +76,12 @@ export function App({
     }
   }, [internalHardware, screenReader]);
 
-  /* istanbul ignore next - need to figure out how to test this */
-  const onKeyPress = useCallback(
+  const onKeyDown = useCallback(
```
adghayes, line 79: Have to use KeyPress to detect keys that do not produce a character, like `F17` from the Accessible Controller.

frontends/bmd/src/app.test.tsx
```diff
@@ -72,6 +74,38 @@ it('uses kiosk storage when in kiosk-browser', async () => {
...
+  // check that 'r' toggles the screen reader
+  expect(screenReader.toggle).toHaveBeenCalledTimes(0);
+  fireEvent.keyDown(screen.getByText('Device Not Configured'), { key: 'r' });
+  await waitFor(() => {
+    expect(screenReader.toggle).toHaveBeenCalledTimes(1);
+  });
+
+  // check that 'F17' changes volume
+  expect(screenReader.changeVolume).toHaveBeenCalledTimes(0);
+  fireEvent.keyDown(screen.getByText('Device Not Configured'), { key: 'F17' });
```
benadida, line 105: If there's any way to not depend on the a particular piece of text here that is irrelevant to the action, that will make this test a little bit more robust if we choose to change the layout / wording for the "Device Not Configured" screen. Maybe add something with a `test-id`, and use `getByTestId` ?

Not strictly required to merge, just a suggestion.

frontends/bmd/src/app.tsx
benadida: are we sure this change doesn't affect behavior of the accessible controller for other actions?

benadida (approved):
looks good! One small suggestion, take it or leave it, and one double check that the key event change didn't screw anything else up.

frontends/bmd/src/app.test.tsx
adghayes: Good call, just found I can use `getByRole('main')` which is a degree more resilient.

adghayes:
I am almost certain that the switch the `keyDown` does not affect any behavior. First of all, [`keyPress` has been deprecated ](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event) in favor of `keyDown` and `keyUp`. The behavior of the accessible controller's other buttons is not handled by our own custom events, but by the browser itself, so it shouldn't affect that.

frontends/bmd/src/utils/ScreenReader/kiosk_text_to_speech.ts
```diff
@@ -0,0 +1,112 @@
...
+  constructor() {
+    assert(window.kiosk, 'KioskTextToSpeech requires window.kiosk');
+  }
+
+  /**
+   * Directly triggers speech of text. Resolves when speaking is finished
+   * or cancelled.
+   */
+  async speak(text: string, { now = true }: SpeakOptions = {}): Promise<void> {
+    assert(now, 'method "speak"  with "options.now = false" is not supported');
```
eventualbuddha, line 35: As @benadida mentioned in his earlier comment on this, I don't think we ever actually use `now = false`, so maybe we just strip it out altogether?

Issue #1975, opened 2022-06-16
Remove Official/Test/Sample toggle from super admin Draft Ballots tab and only allow printing of draft ballots


arsalansufi:
A "draft ballot" looks just like a sample ballot but has all instances of the word "sample" replaced by "draft" (including the watermark)

PR #2249, opened 2022-08-02
Distinguish Draft Ballots tab from Ballots tab

arsalansufi:

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/1975

This PR distinguishes the system admin (super admin) Draft Ballots tab from the election manager (admin) Ballots tab. On the Draft Ballots tab, you can only print draft ballots.

## Demo Screenshots

_Draft Ballots tab on left vs. Ballots tab on right_

draft-ballots-1 ballots-1
draft-ballots-2 ballots-2

## Testing Plan 

- [x] Added an automated test
- [x] Tested manually

## Checklist

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -474,17 +474,17 @@ test('printing ballots and printed ballots report', async () => {
   fireEvent.click(getByText('Absentee'));
   fireEvent.click(getByText('Test'));
   fireEvent.click(getByText('Official'));
-  fireEvent.click(getByText('Print 1 Official', { exact: false }));
+  fireEvent.click(getByText('Print 1', { exact: false }));
   await waitFor(() => getByText('Printing'));
   expect(printer.print).toHaveBeenCalledTimes(1);
   expect(mockKiosk.log).toHaveBeenCalledWith(
     expect.stringContaining(LogEventId.BallotPrinted)
   );
-  fireEvent.click(getByText('Print 1 Official', { exact: false }));
+  fireEvent.click(getByText('Print 1', { exact: false }));
   await waitFor(() => getByText('Printing'));
   expect(printer.print).toHaveBeenCalledTimes(2);
   fireEvent.click(getByText('Precinct'));
-  fireEvent.click(getByText('Print 1 Official', { exact: false }));
+  fireEvent.click(getByText('Print 1', { exact: false }));
   await waitFor(() => getByText('Printing'));
   expect(printer.print).toHaveBeenCalledTimes(3);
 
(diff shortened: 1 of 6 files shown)


frontends/election-manager/src/components/election_manager.tsx
```diff
@@ -159,10 +159,7 @@ export function ElectionManager(): JSX.Element {
             <DefinitionContestsScreen allowEditing={false} />
           </Route>
           <Route exact path={routerPaths.ballotsList}>
-            <BallotListScreen />
-          </Route>
-          <Route exact path={routerPaths.printedBallotsReport}>
-            <PrintedBallotsReportScreen />
```
arsalansufi, line 165: It isn't possible to print a printed ballots report as a system admin, so this route isn't needed

frontends/election-manager/src/screens/ballot_screen.tsx
```diff
@@ -241,14 +276,10 @@ export function BallotScreen(): JSX.Element {
...
-              ) : (
-                <strong>
-                  {ballotMode === BallotMode.Test ? 'Test' : 'Sample'}
-                </strong>
-              )}{' '}
-              {isAbsentee ? <strong>Absentee</strong> : 'Precinct'}{' '}
+              <strong>
+                {ballotModeToReadableString(ballotMode)}{' '}
+                {isAbsentee ? 'Absentee' : 'Precinct'}
+              </strong>{' '}
```
arsalansufi, line 282: Slight change here - rather than only bolding certain ballot modes/types, I decided it couldn't hurt to always bold the type of ballot being printed. Makes the code a bit simpler, too (FYI @beausmith)

arsalansufi:
@adghayes I know you're in the bakery today, but I thought this would be a reasonable PR for you to review and approve! Especially since you worked on this code recently

frontends/election-manager/src/screens/ballot_screen.tsx
```diff
@@ -271,9 +302,11 @@ export function BallotScreen(): JSX.Element {
...
             </LinkButton>
           </p>
-          <p>
-            Ballot Package Filename: <Monospace>{filename}</Monospace>
-          </p>
+          {!draftMode && (
+            <p>
+              Ballot Package Filename: <Monospace>{filename}</Monospace>
+            </p>
+          )}
```
jonahkagan, line 309: Do you know why we show this filename on this screen? The action to export the ballot package is on the parent screen.

frontends/election-manager/src/config/types.ts
```diff
@@ -40,6 +40,7 @@ export enum BallotMode {
   Official = 'live',
   Test = 'test',
   Sample = 'sample',
+  Draft = 'draft',
```
jonahkagan, line 43: Maybe some comments explaining these modes could be helpful at this point. I'm starting to get confused 😄

frontends/election-manager/src/screens/ballot_screen.tsx
adghayes: Appreciate this simplification.

frontends/election-manager/src/screens/ballot_screen.tsx
adghayes: I just looked in the git blame and it looks like the filename display was shown before we added the feature to save individual ballots as PDFs. So, at the time, the Ballot Package was the only place to access the PDF. Now that we have a few other, easier ways, I think it should be removed.

frontends/election-manager/src/screens/ballot_list_screen.tsx
```diff
@@ -20,12 +20,15 @@ import { ExportBallotPdfsButton } from '../components/export_ballot_pdfs_button'
 
 const Header = styled.div`
   display: flex;
-  flex-direction: row-reverse;
```
adghayes, line 23: @arsalansufi what does this change do?

frontends/election-manager/src/app.test.tsx
```diff
@@ -474,17 +474,17 @@ test('printing ballots and printed ballots report', async () => {
   fireEvent.click(getByText('Absentee'));
   fireEvent.click(getByText('Test'));
   fireEvent.click(getByText('Official'));
-  fireEvent.click(getByText('Print 1 Official', { exact: false }));
+  fireEvent.click(getByText('Print 1', { exact: false }));
```
adghayes, line 477: @arsalansufi why be less specific here?

adghayes (approved):
LGTM. Couple clarification questions just to understand the fine details. I think based on my other comment that we could pull out the ballot package filename, but that doesn't need to be as part of this PR / block you.

frontends/election-manager/src/config/types.ts
arsalansufi: Excellent call! Added in https://github.com/votingworks/vxsuite/pull/2249/commits/7520221ef50d53b0d2d7b15b81b2f71de47be2cf

frontends/election-manager/src/screens/ballot_screen.tsx
arsalansufi: Thanks for finding that explanation @adghayes! Given that context, I also lean toward removing. @beausmith Any concerns with removing? (I won't block this PR on that call)

frontends/election-manager/src/app.test.tsx
arsalansufi: Yes good q! Because I added bolding to this case (https://github.com/votingworks/vxsuite/pull/2249#discussion_r935867860), the text is now broken up in a way that even `{ exact: false }` can't capture

An alternative that would be even more preferable is:

```ts
getByRole('button', name: { 'Print 1 Official Absentee Ballot' })
```

Unfortunately, for a reason I haven't yet been able to pin down, these kinds of retrievals sometimes fail. When I replaced all cases of `Print 1 Official` in this test with the above, the first worked and subsequent didn't, despite the element existing in the DOM

frontends/election-manager/src/screens/ballot_list_screen.tsx
arsalansufi: So this `flex-direction: row-reverse` was making it such that the children of the `Header` were rendered in the reverse of the order in which they were declared. To get the Precinct/Style toggle to show up to the left of the export buttons, we accordingly had to declare the export buttons first.

That's fine! But a little unintuitive. I just got rid of the `flex-direction: row-reverse` and declared the Precinct/Style toggle first.

`flex-direction: row-reverse` makes more sense when you're designing responsive layouts, and you, for example, want items to appear in a completely different layout / arrangement on mobile vs. desktop.

PR #2251, opened 2022-08-02
fix(scan): Count ballots after accepting

jonahkagan:
## Overview
Now that Plustek will hold the paper on reboot, if the scanner is powered off during interpretation/accepting, we can safely reject the ballot on reboot. This means we don't need to count a ballot until after we have confirmed it was accepted.

## Demo Video or Screenshot
N/A

## Testing Plan 
Existing tests, plus some basic manual testing

### DIFF
--- a/services/scan/src/precinct_scanner_app.test.ts
+++ b/services/scan/src/precinct_scanner_app.test.ts
@@ -200,7 +200,6 @@ test('configure and scan hmpb', async () => {
   await expectStatus(app, {
     state: 'accepting',
     interpretation,
-    ballotsCounted: 1,
   });
   await waitForStatus(app, {
     state: 'accepted',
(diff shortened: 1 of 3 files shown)


benadida (approved):
I'm good with this, but I am worried that we now open the door to a different error: paper is accepted and crash happens before storing the CVR.

Ultimately, only way to solve this is with a modest write-ahead log. Let's keep that in mind as a future fix.

jonahkagan:
@ben I hear that. I agree we should consider adding it. Maybe make a task to track it?

benadida:
> @ben I hear that. I agree we should consider adding it. Maybe make a task to track it?

https://github.com/votingworks/vxsuite/issues/2254

Issue #2203, opened 2022-07-26
Make smartcard PIN generation non-random for testing/demos


mattroe:
Automatically generated random smartcard PINs are unnecessarily difficult to manage when testing or demoing VxSuite. 

When in dev/demo mode, VxAdmin should automatically set all smartcard PINs to `000000`. This can be implemented through a feature flag similar to how we disable card reader checks entirely: https://github.com/votingworks/vxsuite/pull/2075

PR #2255, opened 2022-08-02
Introduce flag to make generated smartcard PINs all zeros for local development and demos

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2203

Random PIN generation = Good for security, tricky for local development and demos 😅

This PR introduces a very literally named `REACT_APP_VX_ENABLE_ALL_ZERO_SMARTCARD_PIN_GENERATION`  feature flag to make generated smartcard PINs all zeros (000000) instead of random. The flag (like most of our flags) can't be enabled in production.

## Testing Plan 

- [x] Added automated tests
- [x] Manually tested with the flag on
- [x] Manually tested with the flag off

## Checklist

- [ ] ~I have added JSDoc comments to any newly introduced exports~ Updated existing JSDoc comments

### DIFF
--- a/libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.ts
+++ b/libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.ts
@@ -22,7 +22,10 @@ import {
 import deepEqual from 'deep-eql';
 import { useEffect, useReducer } from 'react';
 import useInterval from 'use-interval';
-import { areVvsg2AuthFlowsEnabled } from '../../config/features';
+import {
+  areAllZeroSmartcardPinsEnabled,
+  areVvsg2AuthFlowsEnabled,
+} from '../../config/features';
 import { useLock } from '../use_lock';
 import { usePrevious } from '../use_previous';
 import {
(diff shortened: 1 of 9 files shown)


libs/ui/src/hooks/smartcard_auth/use_dipped_smartcard_auth.ts
```diff
@@ -170,12 +173,14 @@ function smartcardAuthReducer(scope: DippedSmartcardAuthScope) {
 
       case 'check_passcode': {
         assert(previousState.auth.status === 'checking_passcode');
+        const isPasscodeCorrect = areAllZeroSmartcardPinsEnabled()
+          ? action.passcode === '000000'
+          : action.passcode === previousState.auth.user.passcode;
```
jonahkagan, line 178: why is it necessary to flag here as well if the card itself has an all-zero PIN?

arsalansufi: Good q! My thinking was that, whenever the flag is disabled, you may create cards with real non-zero PINs. When this flag is enabled, you should ideally be able to get into any machine, even using those cards with real non-zero PINs, by entering 000000. But I realize that this adds complexity.

@mattroe Given the use cases you're imagining, any thoughts here? Is it useful / important to be able to get into machines using cards with real PINs generated when the flag was not enabled by entering 000000?

mattroe: @arsalansufi It's OK if one needs to program a card on VxAdmin with this flag set before being able to login with 000000. I think it's pretty unlikely that cards programmed with random PINs would be used in this circumstance as we are now enforcing EM-card election-hash checking that forces one to program for the given election definition.

Overall, it's unlikely that I would need to login as 000000 for a random PIN card.

arsalansufi: Awesome, thank you for the fresh eyes @jonahkagan and thank you for confirming @mattroe!

> I think it's pretty unlikely that cards programmed with random PINs would be used in this circumstance as we are now enforcing EM-card election-hash checking

This is a great point. Removed the modified PIN checking logic and kept only the modified PIN generation logic ✅

jonahkagan (approved):

arsalansufi:
Just realized I should've moved the flag from `libs/ui` to `frontends/election-manager` since it's only used there now - will handle that in a separate PR

Issue #2242, opened 2022-08-01
VxScan: ballot count issue when inserting second sheet immediately after first


mattroe:
Slack discussion and issue description: https://votingworks.slack.com/archives/CEL6D3GAD/p1659134671317349?thread_ts=1659132432.040419&cid=CEL6D3GAD

PR #2256, opened 2022-08-02
fix(scan): Handle insertion of second ballot while first ballot is accepting

jonahkagan:
## Overview
Task: https://github.com/votingworks/vxsuite/issues/2242

If a second ballot is inserted while the first ballot is in the middle of being accepted (a very short window, but possible), the first ballot is successfully dropped but plustek returns a jam error. By ignoring that error, we can correctly record that the ballot got accepted.

Then, we also add a delay to make sure that the user has time to see the "ballot counted" screen before the second ballot starts scanning.

## Demo Video or Screenshot
Recommend trying it out yourself

## Testing Plan 
- Updated automated tests
- Manually tested

### DIFF
--- a/services/scan/src/precinct_scanner_state_machine.ts
+++ b/services/scan/src/precinct_scanner_state_machine.ts
@@ -98,6 +98,7 @@ export type Event =
 
 interface Delays {
   DELAY_RECONNECT: number;
+  DELAY_ACCEPTED_READY_FOR_NEXT_BALLOT: number;
   DELAY_ACCEPTED_RESET_TO_NO_PAPER: number;
 }
 
@@ -464,8 +465,11 @@ function buildMachine(createPlustekClient: CreatePlustekClient) {
           invoke: {
             src: accept,
             onDone: 'checking_accepting_completed',
+            // In some cases, Plustek can return an error even if the paper got
+            // accepted, so we need to check paper status to determine what to do
+            // next. We still record the error for debugging purposes.
             onError: {
-              target: 'rejecting',
+              target: 'checking_accepting_completed',
               actions: assign({ error: (_context, event) => event.data }),
             },
           },
@@ -474,6 +478,9 @@ function buildMachine(createPlustekClient: CreatePlustekClient) {
           invoke: pollPaperStatus,
           on: {
             SCANNER_NO_PAPER: 'accepted',
+            // If there's a paper in front, that means the ballot in back did get
+            // dropped but somebody quickly inserted a new ballot in front, so we
+            // should count the first ballot as accepted.
             SCANNER_READY_TO_SCAN: 'accepted',
             // If the paper didn't get dropped, it's an error
             SCANNER_READY_TO_EJECT: {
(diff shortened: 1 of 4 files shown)


services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -476,10 +487,20 @@ function buildMachine(createPlustekClient: CreatePlustekClient) {
         on: {
           WAIT_FOR_PAPER: 'no_paper',
           SCANNER_NO_PAPER: { target: 'accepted', internal: true },
-          SCANNER_READY_TO_SCAN: {
-            target: 'ready_to_scan',
-            actions: clearLastScan,
+          SCANNER_READY_TO_SCAN: { target: 'accepted', internal: true },
+        },
+        // TODO should this timeout be controlled by the frontend
+        after: { 3000: 'accepted_and_waiting_for_paper' },
```
jonahkagan, line 493: @benadida I'm curious to hear your thoughts on the best way to implement this. For the 5s timeout to move from `accepted` back to `no_paper`, I added the `WAIT_FOR_PAPER` command so the frontend could control when that happens. Now, we have three cases after accept:
- (immediately) don't scan new ballots, show accepted screen
- (3s later) scan new ballots, show accepted screen
- (5s later) scan new ballots, show insert ballot screen

Should both transitions be controlled by the frontend? Or just one? Or neither?

I'm starting to think maybe these could just be config for the state machine rather than frontend concerns. Seems like a bit of overkill to have multiple API calls for these transitions when the delay will always be the same. Plus, there's no architectural reason the frontend needs to be in control of this (i.e. if an admin card was inserted before one of these delays, nothing bad would happen)

benadida: I like your approach of putting all of these in the state machine. The only exception is the actual decision to start scanning a ballot, which I think you moved to the frontend for permissioning sanity, which makes sense. I would not move anything more to the front-end, I would add other states, as you suggest, for these things.

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -483,14 +494,22 @@ function buildMachine(createPlustekClient: CreatePlustekClient) {
...
+              after: {
+                DELAY_ACCEPTED_READY_FOR_NEXT_BALLOT: 'ready_for_next_ballot',
+              },
             },
+            ready_for_next_ballot: {
+              on: { SCANNER_READY_TO_SCAN: '..ready_to_scan' },
+            },
+          },
+          after: {
+            DELAY_ACCEPTED_RESET_TO_NO_PAPER: 'no_paper',
```
jonahkagan, line 511: I'd been trying up until this point to avoid using nested states, but I think it's the clearest way to implement this situation with the two delays.

Curious how it is reading it - there are other places where nested states could also be used (e.g. when we use a series of states to complete one operation and check that it succeeded - e.g. rejecting)

benadida: This reads pretty clean to me!

benadida (approved):

services/scan/src/precinct_scanner_state_machine.ts
jonahkagan: Glad to hear it!

PR #2265, opened 2022-08-03
refactor(scan): Move accepted timeout to the scan service 

jonahkagan:
## Overview
Previously, we had let the frontend decide when to timeout the "accepted" state by hitting the `/wait-for-paper` endpoint. Simplify by moving this to the scan service as a configurable param to the state machine. (This also preps for adding another delay to the accepted state for when to start scanning newly inserted ballots in #2256)

Also makes the other delays configurable while we're at it.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Update existing tests


services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -30,6 +30,8 @@ import {
 import { SheetOf } from './types';
```
jonahkagan, line 30: Recommend reviewing this file with "hide whitespace" turned on

benadida (approved):

Issue #2200, opened 2022-07-25
Miscellaneous UX polish


arsalansufi:
- [x] Remove unprogram confirmation modal
- [x] Bold and increase size of PIN text
- [x] Center modal content vertically
- [x] Add padding between software update buttons on VxAdmin sys admin Settings screen (covered in https://github.com/votingworks/vxsuite/pull/2264)
- [x] Other changes captured in https://github.com/votingworks/vxsuite/compare/beau/wip-card-programing
- [ ] ~Reprogram Card button (https://votingworks.slack.com/archives/CEL6D3GAD/p1659043972751609)~ (considered but skipping after discussion with Beau - or at least lowering in priority for m14)

Issue #2181, opened 2022-07-21
VxSuite: Use new user role names


arsalansufi:
Super Admin --> System Administrator
Admin --> Election Manager

~For this issue, let's just focus on names as displayed in UIs. No need to update all internal variable names yet since that's more involved, though once this issue is closed, we should open a separate issue for that change.~

Update: We now are planning on updating everything, user-facing and internal, at once to avoid confusion.

PR #2159, opened 2022-07-19
Complete new smartcard programming UIs

arsalansufi:
_Commit by commit reviewing recommended_

## Overview

Issue link(s): https://github.com/votingworks/vxsuite/issues/1919, https://github.com/votingworks/vxsuite/issues/1920, https://github.com/votingworks/vxsuite/issues/1921

This PR completes the new smartcard programming UIs!

Apologies in advance for the hefty PR! Though it's a lot of lines, the majority of the PR is tests. I'd initially planned to implement each screen (card details, election card programming, and super admin card programming) separately but found that it was helpful to think about them holistically since they share information, e.g. successfully programming a card on a card programming screen drops you on the card details screen with a success message, etc. I've tried to make reviewing as easy as possible by chunking the PR and leaving comments in GitHub to guide reviewing but obviously take your time and let me know if you have any questions!

## Demo screenshots and videos

There are a lot of states that this UI can be in. Below are a video and screenshots that capture many of them, and the tests cover even more states.

https://user-images.githubusercontent.com/12616928/179774513-c93755ce-c21e-4446-86e5-6c0392cdd293.mov

### Smartcard details screens

_Super admin card_

super-admin

_Admin card_

admin

_Poll worker card_

poll-worker

_Admin card programmed for a prior election_

admin-prior-election

### Smartcard programming screens

_Election card programming_

election-cards

_Super admin card programming_

super-admin-cards

_Election card programming when no election definition_

no-election-definition

### Smartcard action states

_Program success_

program-success

_PIN reset success_

pin-reset-success

_Unprogram success_

unprogram-success

_Unprogram confirmation modal_

unprogram-confirmation-modal

_Progress modal_

progress-modal

_Error_

program-error

## Testing

- [x] Added tests covering all possible states (I'm pretty sure but will double check 😝)
- [x] Tested all possible states manually

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced (Covered in a prior lower-level PR: https://github.com/votingworks/vxsuite/pull/2140)
- [x] I have added JSDoc comments to any newly introduced exports
- [x] I have added a screenshot and/or video to this PR to demo the change
- [x] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #2267, opened 2022-08-03
Polish new smartcard programming flows given feedback

arsalansufi:

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2200

This PR polishes the new smartcard programming flows given various pieces of UI feedback ✨

Below are a number of after screenshots. To get a sense of what the UI looked like before, you can look at https://github.com/votingworks/vxsuite/pull/2159. That said, the commits in this PR should make all of the individual changes clear and easy to review!

Note that I'm still using old user role names in this PR. The switch to the new user role names is tracked in a separate ticket (https://github.com/votingworks/vxsuite/issues/2181)

## Demo Screenshots

Grab a cup of tea, sit back, relax, and scroll through screenshots for days 😌

### Card Details

_Super admin card_

super-admin

_Admin card_

admin

_Poll worker card_

poll-worker

_Admin card from prior election_

admin-prior-election

_Poll worker card from prior election_

poll-worker-prior-election

### Card Programming

_Create election card_

create-election-card

_Create super admin card_

create-super-admin-card

### When Machine Has No Election Definition

_Super admin card when machine has no election definition (unaffected)_

super-admin

_Admin card when machine has no election definition_

admin

_Poll worker card when machine has no election definition_

poll-worker

_Create election card when machine has no election definition_

create-election-card

### Success States

_Create super admin card, success_

create-super-admin-card

_Create admin card, success_

create-admin-card

_Create poll worker card, success_

create-poll-worker-card

_Reset card PIN, success_

reset-card-pin

_Unprogram card, success_

unprogram-card

### In Progress States

_Create card, in progress_

create-card

_Reset card PIN, in progress_

reset-card-pin

_Unprogram card, in progress_

unprogram-card

### Error States

_Create super admin card, error_

create-super-admin-card

_Create election card, error_

create-election-card

_Reset card PIN, error_

reset-card-pin

_Unprogram card, error_

unprogram-card

## Testing Plan 

- [x] Tested all relevant cases manually (see above!)
- [x] Updated automated tests, which already cover all of the states captured above

## Checklist


frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
```diff
@@ -1,12 +1,23 @@
...
+import { electionToDisplayString } from './elections';
 import { generatePin } from './pins';
 import { SmartcardActionStatus, StatusMessage } from './status_message';
 
+interface HeadingProps {
+  marginTop: string;
+}
+
+const Heading = styled.h1<HeadingProps>`
+  /* stylelint-disable-next-line declaration-no-important */
```
arsalansufi, line 17: Necessary to override the high-specificity `Prose` styles - will add a clarifying comment!

frontends/election-manager/src/components/smartcard_modal/status_message.tsx
```diff
@@ -74,7 +74,9 @@ export function StatusMessage({
         throwIllegalValue(action);
       }
     }
-    return <Modal content={<Loading as="p">{text}</Loading>} />;
+    return (
+      <Modal centerContent content={<Loading as="strong">{text}</Loading>} />
```
arsalansufi, line 78: As noted in the commit message, I'm holding off of making in-progress text larger until we switch to new
user role names, which are longer. I'll test larger text then

frontends/election-manager/src/components/smartcard_modal/smartcard_modal.test.tsx
```diff
@@ -641,15 +615,13 @@ test('Error handling', async () => {
     {
       cardData: makeAdminCard(electionHash),
       buttonToPress: 'Unprogram Card',
-      confirmButtonToPress: 'Yes',
-      expectedProgressText: 'Deleting all data on card',
+      expectedProgressText: 'Unprogramming Admin card',
```
jonahkagan, line 621: I find these sentences with just one uncapitalized word ("card") to look a bit weird. I would suggest capitalizing to "Admin Card" (it will also better match every other place we say "Admin Card" on the screen)

frontends/election-manager/src/components/smartcard_modal/smartcard_modal.tsx
```diff
@@ -63,7 +58,5 @@ export function SmartcardModal(): JSX.Element | null {
       />
     );
   }
-  return (
-    <Modal content={<ModalContents>{contents}</ModalContents>} fullscreen />
-  );
+  return <Modal centerContent content={contents} fullscreen />;
```
jonahkagan, line 61: Love the look of this change!

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
```diff
@@ -103,70 +111,72 @@ export function CardDetailsView({
...
+      )}
+      {(actionStatus || possibleActions.size > 0) && <HorizontalRule />}
+
+      {showingSuccessOrErrorMessage ? (
+        <Text bold>Remove card to continue.</Text>
+      ) : (
+        <p>
+          Remove card to{' '}
+          {possibleActions.size > 0 ? 'cancel' : 'leave this screen'}.
+        </p>
```
jonahkagan, line 175: This screen is getting a bit complex. It's hard to puzzle out what will actually be rendered. I wonder if it's possible to simplify this logic at all.

Are there a few high level cases that could be broken out into separate `return` blocks (even if there's some repetition)?

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
```diff
@@ -61,31 +75,42 @@ export function ProgramElectionCardView({
...
+      {electionDefinition && (
+        <p>{electionToDisplayString(electionDefinition.election)}</p>
+      )}
+
+      <HorizontalRule />
       {electionDefinition ? (
-        <p>Admin and Poll Worker cards only work for the current election.</p>
+        <p>
+          <Button disabled={!electionDefinition} onPress={programAdminCard}>
+            Admin Card
```
jonahkagan, line 95: Maybe change to "Create Admin Card" to match the language of the other buttons (which all have a verb)

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
```diff
@@ -61,31 +75,42 @@ export function ProgramElectionCardView({
...
+          </Button>{' '}
+          or{' '}
```
jonahkagan, line 97: The "or" was a bit jarring when I saw this screen. I feel like having two buttons next to each other is an implicit "or" already and we don't need to actually write the word

beausmith:
The screenshots "when machine has no election definition" kinda don't make sense to me. Perhaps they are useful to know what is on a card when there is no election definition? Why is there no button to "Unprogram" the card? What am I missing?

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
beausmith: Why aren't we using the negative-margin on the status message as I had committed?

This implementation is a bit odd as it adds styles to one element when another element is introduced.

Additionally, we should avoid using `!important` as this will likely cause us major headaches in the future.

jonahkagan (commented):
Looking really good! Thanks for the screenshots.

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
beausmith: @jonahkagan - With the title of this fullscreen modal being "Create New Election Card", I was treating the buttons as choices of what type of card to create. The title is not a full question, but felt clear enough. Having "Create" next to each type seemed a bit verbose.

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
beausmith: @jonahkagan - Yeah, this is less common in our app. And eventually you won't need to make this choice once we know what type of card this is.

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
```diff
@@ -103,70 +111,72 @@ export function CardDetailsView({
...
-    electionDisplayString = 'N/A';
+  }
+
+  const possibleActions = new Set<SmartcardActionStatus['action']>();
+  if (
+    'passcode' in programmedUser &&
+    (role === 'superadmin' ||
+      // If the card is from a prior election, no need to display PIN resetting. Unprogramming is
+      // the only meaningful action in this case
+      doesCardElectionHashMatchMachineElectionHash)
```
beausmith, line 125: The other election may not be in the past, it may be a parallel election. Thus I think we do need ability to Reset PIN.

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
```diff
@@ -103,70 +111,72 @@ export function CardDetailsView({
...
+        </p>
+      )}
+      {(actionStatus || possibleActions.size > 0) && <HorizontalRule />}
+
+      {showingSuccessOrErrorMessage ? (
+        <Text bold>Remove card to continue.</Text>
+      ) : (
+        <p>
+          Remove card to{' '}
+          {possibleActions.size > 0 ? 'cancel' : 'leave this screen'}.
```
beausmith, line 174: When we do translation/internationalization it is necessary to have full sentences to translate as sentence structure changes in different languages. This is why my commit had separated these into two distinct phrases rather than one phrase with conditional logic inside it. 

If I'm doing a string search for a phase that may in a test, it's also nice to not have to think about what part of the sentence may be a conditional, and just be able to search for the full phrase.

frontends/election-manager/src/components/smartcard_modal/status_message.tsx
```diff
@@ -74,34 +78,36 @@ export function StatusMessage({
...
+        ) : (
+          <React.Fragment>New card created.</React.Fragment>
         )}
       </Text>
     );
   }
 
   if (action === 'PinReset' && programmedUser && 'passcode' in programmedUser) {
+    // No need to include the programmed user role in these messages since it's so prominent
+    // elsewhere on the screen
```
beausmith, line 101: Is this comment necessary as it mentions something that is no longer on the screen?

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
beausmith: Another solution for the styling of the status message and accounting for the Prose spacing is to place the message in a separate Prose block before the current one. Doing so will eliminate the margin between these elements and then we can simply add the desired margin instead of overriding any margins.

frontends/election-manager/src/components/smartcard_modal/program_super_admin_card_view.tsx
```diff
@@ -1,10 +1,20 @@
...
 
+interface HeadingProps {
+  marginTop: string;
+}
+
+const Heading = styled.h1<HeadingProps>`
+  /* stylelint-disable-next-line declaration-no-important */
+  margin-top: ${(props) => props.marginTop} !important;
+`;
+
```
beausmith, line 17: Would like to handle this in another way. I fear this is getting fragile and working against the spirit of the Prose component.

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
```diff
@@ -94,87 +98,81 @@ export function CardDetailsView({
...
+    possibleActions.add('Unprogram');
   }
 
   return (
-    <Prose textCenter>
-      <h2>Card Details</h2>
-      {/* An empty div to maintain space between the header and subsequent p. TODO: Consider adding
-        a `maintainSpaceBelowHeaders` prop to `Prose` */}
-      <div />
+    <Prose textCenter theme={fontSizeTheme.medium}>
```
beausmith, line 137: Why was this changed from large?

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
beausmith: I agree with Jonah. Don't know if it helps, but when determining how to write programming logic, my goal is to make it easy for another programmer to read. Often this other programmer is me coming back to the code having lost all previous context. :ton

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
arsalansufi: Yes good q! I initially switched to a modified top margin on the heading because 1) I generally find positive margins more intuitive than negative ones (don't feel strongly about this, though) and 2) I wanted to be able to remove the margin when `StatusMessage` is displaying an in-progress modal (more on this below).

That said, you raise some really good points about not using `!important` (the need to use it was a red flag for me, too).

Working on some updates! I like the idea of using two separate `Prose` blocks. I'm also looking into better distinguishing status messages for success and error messages from in-progress modals. They're really two separate things, but both captured in the single `StatusMessage` component at the moment. Separating them out will help eliminate challenge 2.

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: Good callout! I agree that the branching within this component has gotten high and that this component could benefit from higher-level branching. Will tweak 🛠️

> when determining how to write programming logic, my goal is to make it easy for another programmer to read. Often this other programmer is me coming back to the code having lost all previous context

Yes, this is a given! I've probably just gotten too familiar with this code over the past few days 😝

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
arsalansufi: So just to double check @beausmith, we still want to keep the "or" here, then?

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: Ah yes I know you've pointed this out before! Will update and keep in mind for future changes

frontends/election-manager/src/components/smartcard_modal/status_message.tsx
```diff
@@ -74,34 +78,36 @@ export function StatusMessage({
   }
 
   if (action === 'Program' && programmedUser) {
+    // No need to include the programmed user role in these messages since it's so prominent
+    // elsewhere on the screen
```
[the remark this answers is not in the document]
arsalansufi: I only noted this because other messages in this component, error and in-progress, do include user roles. And I could see someone thinking we just forgot to include them here when it was actually a conscious choice

frontends/election-manager/src/components/smartcard_modal/program_super_admin_card_view.tsx
arsalansufi: As noted in https://github.com/votingworks/vxsuite/pull/2267#discussion_r937894753, agreed!

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: Large was spilling outside the bounds of my 14-inch MacBook screen, but I may not be testing on the right resolution. What's the resolution of the ThinkPads that we run VxAdmin on?

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: Yes, this crossed my mind. That said, this change is larger in scope than meets the eye.

When we reset the PIN today, we actually just fully re-program the card, election definition and all:

https://github.com/votingworks/vxsuite/blob/a4b355a3942cca8ac2ce9465f6eae34f8ba88d61/frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx#L76-L84

We haven't yet exposed an interface to just update the PIN, while leaving other contents, like a previous election def, intact. Lots of ways to go about this, none too difficult, but will require testing, etc.

Since we haven't yet built multi-election support on VxAdmin, I chose not to prioritize this for m14. That said, the current code comment is a bit misleading. Let me update the comment to shed some light on future directions. I'll also document the PIN reset updates in a follow-up task.

arsalansufi:
> The screenshots "when machine has no election definition" kinda don't make sense to me. Perhaps they are useful to know what is on a card when there is no election definition? Why is there no button to "Unprogram" the card? What am I missing?

Let's chat about these in person!

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
arsalansufi: https://github.com/votingworks/vxsuite/pull/2267/commits/0c443ac0c24b03350597322b7489abe5144f52bb + https://github.com/votingworks/vxsuite/pull/2267/commits/c986a9beda806953b218d4dd58316d2330d1a70b?w=1

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: ^ Got the resolution of the ThinkPads but also collectively decided with @beausmith that medium text is fine. What's really most important is that the PIN text is larger relative to other text, which it still is at `1.5em`

frontends/election-manager/src/components/smartcard_modal/smartcard_modal.test.tsx
arsalansufi: https://github.com/votingworks/vxsuite/pull/2267/commits/df6babff8f28bf2527832277de026c74f4aa7f26

frontends/election-manager/src/components/smartcard_modal/program_election_card_view.tsx
arsalansufi: Chatted with @beausmith and we're going to keep this as is for now

Down the road, the screen real estate piece will become less of an issue as we lock cards down to the user roles printed on them and will only need to display one option. We can add "Create" back in, then

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: https://github.com/votingworks/vxsuite/pull/2267/commits/0982d1392f52aeabfc5b13ae1f885ad0dc685757

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
```diff
@@ -103,70 +111,72 @@ export function CardDetailsView({
...
-        on the machine since cards can't be programmed in this state */}
-      {(role === 'admin' || role === 'pollworker') && electionDefinition && (
-        <React.Fragment>
-          <p>
+      {!showingSuccessOrErrorMessage && possibleActions.size > 0 && (
+        <p>
+          {possibleActions.has('PinReset') && (
+            <Button onPress={resetCardPin}>Reset Card PIN</Button>
+          )}
+          {possibleActions.size > 1 && ' '}
```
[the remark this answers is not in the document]
arsalansufi: https://github.com/votingworks/vxsuite/pull/2267/commits/47faf5eb524d3aa8dc2cc096509e3d9a999af08e

frontends/election-manager/src/components/smartcard_modal/status_message.tsx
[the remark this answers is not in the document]
arsalansufi: Despite the above, thought some more about this one and agree that this isn't necessary! https://github.com/votingworks/vxsuite/pull/2267/commits/0769554e8bb31e5a44876b96a1ff5955abe77c45

It may even lose relevance as we potentially shorten other status messages to not include user role either

frontends/election-manager/src/components/smartcard_modal/status_message.tsx
arsalansufi: Covered in discussion on https://github.com/votingworks/vxsuite/pull/2267#discussion_r937741413

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: Okay! Landed on this: https://github.com/votingworks/vxsuite/pull/2267/commits/c983f0ea3b0a01fb90efe24e17194c746c29b357

Hopefully makes what's displayed under different circumstances easier to glean

I know I bias toward avoiding repetition, so if this doesn't go far enough, and you'd still prefer some earlier returns, more than happy to tweak further

arsalansufi:
@beausmith @jonahkagan I'm working on updated screenshots, but otherwise, this is ready for a follow-up look!

jonahkagan (approved):

PR #2268, opened 2022-08-03
L&A Screen Instructions

beausmith:
## Demo Video or Screenshot
image
image

### DIFF
--- a/frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
+++ b/frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
@@ -13,24 +13,43 @@ export function LogicAndAccuracyScreen(): JSX.Element {
   return (
     <NavigationScreen>
       <Prose maxWidth={false}>
-        <h1>L&amp;A Materials</h1>
+        <h1>L&A Testing Documents</h1>
         {isLiveMode ? (
           <p>
-            L&amp;A materials are not available after live CVRs have been
-            imported.
+            L&A testing documents are not available after official election CVRs
+            have been imported.
           </p>
         ) : (
           <React.Fragment>
             <p>
-              Go to the “Tally” tab and print the “Unofficial Full Election
-              Tally Report”. This report is referred to as the “Zero Report”
-              because — before CVRs have been imported — all tallies should be
-              zero.
+              Print the following reports and ballot packages in preparation for
+              L&A testing.
+            </p>
+            <h2>1. Unofficial Full Election Tally Report</h2>
+            <p>
+              Print the Full Election Tally Report to document that no votes
+              have been tallied before L&A testing begins.
+            </p>
+            <p>
+              <em>
+                Go to the Reports tab and select the “Unofficial Full Election
+                Tally Report” button at the top of the page to print the report.
+              </em>
+            </p>
+            <h2>2. Precinct L&A Packages</h2>
+            <p>
+              Each Precinct L&A Package contains marked test ballots, unmarked
+              test ballots, and a tally report with expected results.
             </p>
             <p>
               <LinkButton to={routerPaths.testDecks}>
-                Print L&amp;A Packages
+                List Precinct L&A Packages
               </LinkButton>
+            </p>
+            <h2>3. Test Deck Tally Report for All Precincts</h2>
+            <p>
+              This report has the results that are expected after scanning all
+              the test ballots.
             </p>
             <p>
               <FullTestDeckTallyReportButton />
--- a/frontends/election-manager/src/screens/print_test_deck_screen.tsx
+++ b/frontends/election-manager/src/screens/print_test_deck_screen.tsx
@@ -18,6 +18,7 @@ import {
   Prose,
   isAdminAuth,
   isSuperadminAuth,
+  HorizontalRule,
 } from '@votingworks/ui';
 
 import { AppContext } from '../contexts/app_context';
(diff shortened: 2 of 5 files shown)


frontends/election-manager/src/screens/print_test_deck_screen.tsx
```diff
@@ -574,15 +575,35 @@ export function PrintTestDeckScreen(): JSX.Element {
...
-            Select desired precinct for <strong>{election.title}</strong>.
+            Print all or select a precinct to print the L&A Package for that
+            precinct.
           </p>
+          <HorizontalRule />
+          <p>Each Precinct L&A Package contains:</p>
+          <ol>
+            <li>
+              The Precinct Tally Report is the expected results from scanning
+              the marked ballots for this precinct.
```
arsalansufi, line 589: This being a full sentence reads a bit funny to me since we're also leading into it with "Each Precinct L&A Package contains".

I end up reading it as "Each Precinct L&A Package contains The Precinct Tally Report is the expected results from scanning the marked ballots for this precinct."

What do you think of changing this item to "The Precinct Tally Report. This is the expected result from scanning the marked ballots for this precinct." or something along those lines? This will also make this list item better match the other list items.

frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
```diff
@@ -13,25 +13,43 @@ export function LogicAndAccuracyScreen(): JSX.Element {
...
           </p>
         ) : (
           <React.Fragment>
             <p>
-              Go to the “Tally” tab and print the “Unofficial Full Election
-              Tally Report”. This report is referred to as the “Zero Report”
-              because — before CVRs have been imported — all tallies should be
-              zero.
+              Print the following reports and ballots in preparation for L&A
+              Testing.
```
arsalansufi, line 25: We're capitalizing "L&A Testing" here but not below when we say "Print this tally report to document that no votes have been tallied before L&A testing begins."

Let's stick with one for consistency (not including titles of course). I lean toward lower case "testing"

frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
```diff
@@ -13,25 +13,43 @@ export function LogicAndAccuracyScreen(): JSX.Element {
   return (
     <NavigationScreen>
       <Prose maxWidth={false}>
-        <h1>L&amp;A Materials</h1>
+        <h1>L&A Testing Documents</h1>
         {isLiveMode ? (
           <p>
-            L&amp;A materials are not available after live CVRs have been
-            imported.
+            L&A materials are not available after live CVRs have been imported.
```
arsalansufi, line 19: Should this read "L&A testing documents are not available..." for consistency with the "L&A Testing Documents" title? Don't feel strongly about this!

frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
```diff
@@ -13,25 +13,43 @@ export function LogicAndAccuracyScreen(): JSX.Element {
...
               <LinkButton to={routerPaths.testDecks}>
-                Print L&amp;A Packages
+                List Precinct L&A Packages
               </LinkButton>
             </p>
+            <h2>3. Test Deck Tally Report for All Precincts</h2>
+            <p>
+              This report has the results that are expected after scanning all
+              the ballots.
+            </p>
```
arsalansufi, line 52: I learned during the L&A project that the tally report includes double the counts of all the ballots since we scan everything on a precinct scanner and a central scanner (@adghayes taught me this!). Do we need to mention that anywhere?

This page is already getting quite verbose, so I understand if that doesn't need to be mentioned or is implied

arsalansufi (commented):
Nice! Left a few thoughts

frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
beausmith: lowercased.

frontends/election-manager/src/screens/print_test_deck_screen.tsx
beausmith: Changing to Ginny's suggestion of "A Precinct Tally Report — the expected results of the precinct."

frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
beausmith: Good catch. Changing to "L&A testing documents are not available after official election CVRs have been imported."

frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
beausmith: I know that this was an issue in the past, however I'm not sure this is still the case. 

@ginvdr - Can you clarify?

arsalansufi (approved):
LGTM!

frontends/election-manager/src/screens/logic_and_accuracy_screen.tsx
ginvdr: I don't think we need to add instruction about that @arsalansufi but it's a good question to raise.  I'll make sure it's in our documentation and covered in training.

arsalansufi: Awesome, I figured as much!

Issue #2241, opened 2022-08-01
VxScan: partial ballot insert leads to indefinite "please wait" screen


mattroe:

jonahkagan:
Closed by #2303

Issue #2240, opened 2022-08-01
VxScan: one yellow-orange Plustek light


mattroe:
Slack discussion: https://votingworks.slack.com/archives/CEL6D3GAD/p1659132867252809?thread_ts=1659132432.040419&cid=CEL6D3GAD

1. Need to determine what this status is
2. Depending on status determination, decide what the user-facing flow should be, but it probably shouldn't be "Internal Connection Problem"

mattroe:
Thread discussing similar behavior as seen on latest main: https://votingworks.slack.com/archives/CEL6D3GAD/p1659570987125819

jonahkagan:
Closed by https://github.com/votingworks/vxsuite/pull/2303

PR #2275, opened 2022-08-05
fix(scan): Safe scanning retries

jonahkagan:
## Overview
Tasks: https://github.com/votingworks/vxsuite/issues/2241, https://github.com/votingworks/vxsuite/issues/2240

Addresses a few different issues when attempting to scan:
- When Plustek fails to grab the ballot, we cap the number of times we will retry the scan command and then show an error to remove the ballot.
- We only will retry scanning on known errors (error feeding, error no paper) or when the paper status indicates the paper didn't get scanned. This protects us against both cases from https://github.com/votingworks/vxsuite/issues/2240, in which case the scan command returns a "jam" error even though there's not actually a jam. In the case of any unexpected error from the scan command, we'll treat it like any other unexpected error.

## Testing Plan 
TODO


jonahkagan:
Closing in favor of https://github.com/votingworks/vxsuite/pull/2303

PR #2276, opened 2022-08-05
feat(ps): Design and copy review.

beausmith:
## Demo Video or Screenshot
1. Run app.
2. Go to localhost:3000/preview

### DIFF
--- a/frontends/precinct-scanner/src/screens/insert_ballot_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/insert_ballot_screen.tsx
@@ -23,7 +23,7 @@ export function InsertBallotScreen({
       {showNoChargerWarning && (
         <Text warning small center>
           <strong>No Power Detected.</strong> Please ask a poll worker to plug
-          in the power cord for this machine.
+          in the power cord.
         </Text>
       )}
       <InsertBallot />
--- a/frontends/precinct-scanner/src/screens/invalid_card_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/invalid_card_screen.tsx
@@ -11,7 +11,8 @@ export function InvalidCardScreen(): JSX.Element {
     <ScreenMainCenterChild infoBar={false}>
       <TimesCircle />
       <CenteredLargeProse>
-        <h1>Invalid Card, please remove.</h1>
+        <h1>Invalid Card</h1>
+        <p>Remove the card to continue.</p>
       </CenteredLargeProse>
     </ScreenMainCenterChild>
   );
--- a/frontends/precinct-scanner/src/screens/scan_error_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/scan_error_screen.tsx
@@ -17,29 +17,33 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
   const errorMessage = (() => {
     if (!error) return undefined;
     switch (error) {
-      // Invalid interpretation
+      // Invalid ballot interpretations
       case 'invalid_test_mode':
-        return isTestMode ? 'Live Ballot detected.' : 'Test ballot detected.';
+        return isTestMode
+          ? 'Live ballot detected. Scanner is in test mode.'
+          : 'Test ballot detected. Scanner is in live mode.';
       case 'invalid_election_hash':
-        return 'Scanned ballot does not match the election this scanner is configured for.';
+        return 'The ballot does not match the election this scanner is configured for.';
       case 'invalid_precinct':
-        return 'Scanned ballot does not match the precinct this scanner is configured for.';
+        return 'The ballot does not match the precinct this scanner is configured for.';
       case 'unreadable':
-        return 'There was a problem reading this ballot. Please try again.';
       case 'unknown':
-        return undefined;
-      // Precinct scanner error
+        return 'There was a problem reading this ballot. Please scan again.';
+      // Precinct scanner errors
       case 'scanning_failed':
+        return 'Ballot not fully inserted. Remove ballot to continue.';
       case 'both_sides_have_paper':
+        return 'Scanning interrupted. Remove ballot to continue.';
       case 'paper_in_front_on_startup':
+        return 'Scanner detected an unexpected ballot in the tray. Remove ballot to continue.';
       case 'paper_in_back_on_startup':
       case 'paper_in_back_after_accept':
-        return 'Take your ballot out of the tray and try again.';
+        return 'Scanner detected an unexpected ballot at the back of the scanner. Remove ballot to continue.';
       case 'scanning_timed_out':
       case 'unexpected_paper_status':
       case 'unexpected_event':
       case 'plustek_error':
-        return undefined;
+        return 'The scanner experienced an error and needs to be reset. Please turn off and unplug the scanner from the power outlet. Plug it back in and turn it on again.';
       default:
         throwIllegalValue(error);
     }
@@ -48,9 +52,11 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
     <ScreenMainCenterChild infoBar={false}>
       <TimesCircle />
       <CenteredLargeProse>
-        <h1>Scanning Error</h1>
+        <h1>Ballot Not Counted</h1>
         <p>{errorMessage}</p>
-        <Text italic>Ask a poll worker for help.</Text>
+        <Text small italic>
+          Ask a poll worker for help.
+        </Text>
       </CenteredLargeProse>
     </ScreenMainCenterChild>
   );
--- a/frontends/precinct-scanner/src/screens/scan_jam_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/scan_jam_screen.tsx
@@ -11,8 +11,11 @@ export function ScanJamScreen(): JSX.Element {
     <ScreenMainCenterChild infoBar={false}>
       <TimesCircle />
       <CenteredLargeProse>
-        <h1>Scanner Jammed</h1>
-        <Text italic>Ask a poll worker for help.</Text>
+        <h1>Ballot Not Counted</h1>
+        <p>The ballot is jammed in the scanner.</p>
+        <Text small italic>
+          Ask a poll worker for help.
+        </Text>
       </CenteredLargeProse>
     </ScreenMainCenterChild>
   );
--- a/frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
+++ b/frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
@@ -46,10 +46,14 @@ test('overvote', () => {
 
   screen.getByRole('heading', { name: 'Too Many Votes' });
   screen.getByText(
-    new RegExp(`You voted for too many choices for ${contest.title}.`)
+    new RegExp(
+      `There are too many votes marked in the contest for: ${contest.title}.`
+    )
   );
-  userEvent.click(screen.getByRole('button', { name: 'Count my ballot' }));
-  userEvent.click(screen.getByRole('button', { name: 'Yes, count my ballot' }));
+  userEvent.click(screen.getByRole('button', { name: 'Cast Ballot As Is' }));
+  userEvent.click(
+    screen.getByRole('button', { name: 'Yes, Cast Ballot As Is' })
+  );
   expect(fetchMock.done()).toBe(true);
 });
 
@@ -59,9 +63,11 @@ test('blank ballot', () => {
     adjudicationReasonInfo: [{ type: AdjudicationReason.BlankBallot }],
   });
 
-  screen.getByRole('heading', { name: 'Blank Ballot' });
-  screen.getByText('Your ballot does not have any votes.');
-  userEvent.click(screen.getByRole('button', { name: 'Count blank ballot' }));
+  screen.getByRole('heading', {
+    name: 'Review Your Ballot',
+  });
+  screen.getByText('No votes were found when scanning this ballot.');
+  userEvent.click(screen.getByRole('button', { name: 'Cast Ballot As Is' }));
   userEvent.click(
     screen.getByRole('button', { name: 'Yes, count blank ballot' })
   );
@@ -87,9 +93,11 @@ test('undervote no votes', () => {
   });
 
   screen.getByRole('heading', { name: 'Review Your Ballot' });
-  screen.getByText(new RegExp(`You did not vote for ${contest.title}.`));
-  userEvent.click(screen.getByRole('button', { name: 'Count my ballot' }));
-  userEvent.click(screen.getByRole('button', { name: 'Yes, count my ballot' }));
+  screen.getByText(new RegExp(`No votes detected for: ${contest.title}.`));
+  userEvent.click(screen.getByRole('button', { name: 'Cast Ballot As Is' }));
+  userEvent.click(
+    screen.getByRole('button', { name: 'Yes, Cast Ballot As Is' })
+  );
   expect(fetchMock.done()).toBe(true);
 });
 
@@ -115,10 +123,14 @@ test('undervote by 1', () => {
 
   screen.getByRole('heading', { name: 'Review Your Ballot' });
   screen.getByText(
-    new RegExp(`You can vote for more people for ${contest.title}.`)
+    new RegExp(
+      `You may vote for more candidates in the contests for: ${contest.title}.`
+    )
   );
-  userEvent.click(screen.getByRole('button', { name: 'Count my ballot' }));
-  userEvent.click(screen.getByRole('button', { name: 'Yes, count my ballot' }));
+  userEvent.click(screen.getByRole('button', { name: 'Cast Ballot As Is' }));
+  userEvent.click(
+    screen.getByRole('button', { name: 'Yes, Cast Ballot As Is' })
+  );
   expect(fetchMock.done()).toBe(true);
 });
 
@@ -141,11 +153,13 @@ test('multiple undervotes', () => {
   screen.getByRole('heading', { name: 'Review Your Ballot' });
   screen.getByText(
     new RegExp(
-      `You can vote for more people for ${contests[0].title} and ${contests[1].title}.`
+      `You may vote for more candidates in the contests for: ${contests[0].title} and ${contests[1].title}.`
     )
   );
-  userEvent.click(screen.getByRole('button', { name: 'Count my ballot' }));
-  userEvent.click(screen.getByRole('button', { name: 'Yes, count my ballot' }));
+  userEvent.click(screen.getByRole('button', { name: 'Cast Ballot As Is' }));
+  userEvent.click(
+    screen.getByRole('button', { name: 'Yes, Cast Ballot As Is' })
+  );
   expect(fetchMock.done()).toBe(true);
 });
 
@@ -158,8 +172,10 @@ test('unreadable', () => {
     ],
   });
 
-  screen.getByRole('heading', { name: 'Review Your Ballot' });
-  userEvent.click(screen.getByRole('button', { name: 'Count my ballot' }));
-  userEvent.click(screen.getByRole('button', { name: 'Yes, count my ballot' }));
+  screen.getByRole('heading', { name: 'Scanning Failed' });
+  userEvent.click(screen.getByRole('button', { name: 'Cast Ballot As Is' }));
+  userEvent.click(
+    screen.getByRole('button', { name: 'Yes, Cast Ballot As Is' })
+  );
   expect(fetchMock.done()).toBe(true);
 });
--- a/frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
@@ -6,6 +6,7 @@ import {
   AdjudicationReasonInfo,
   OvervoteAdjudicationReasonInfo,
   UndervoteAdjudicationReasonInfo,
+  AnyContest,
 } from '@votingworks/types';
 import { Button, Modal, Prose, Text } from '@votingworks/ui';
 import { assert, find, integers, take } from '@votingworks/utils';
@@ -31,34 +32,41 @@ function OvervoteWarningScreen({
   overvotes,
 }: OvervoteWarningScreenProps): JSX.Element {
   const [confirmTabulate, setConfirmTabulate] = useState(false);
-  const contestNames = overvotes.map(
-    (overvote) =>
+  const contestNames = overvotes
+    .map((overvote) =>
       find(
         electionDefinition.election.contests,
         (contest) => contest.id === overvote.contestId
-      ).title
-  );
+      )
+    )
+    .reduce<AnyContest[]>(
+      (a, c) => (a.findIndex((o) => o.title === c.title) >= 0 ? a : [...a, c]),
+      []
+    )
+    .map((c) => c.title);
 
   return (
     <ScreenMainCenterChild infoBar={false}>
       <ExclamationTriangle />
       <CenteredLargeProse>
         <h1>Too Many Votes</h1>
-        <Text small>
-          You voted for too many choices for {toSentence(contestNames)}. To fix
-          this, take your ballot and ask a poll worker for a new ballot.
+        <Text>
+          There are too many votes marked in the{' '}
+          {pluralize('contest', contestNames.length)} for:{' '}
+          {toSentence(contestNames)}.
         </Text>
         <p>
-          <Button primary large onPress={scanner.returnBallot}>
-            Return my ballot
+          <Button primary onPress={scanner.returnBallot}>
+            Return Ballot
+          </Button>{' '}
+          or{' '}
+          <Button onPress={() => setConfirmTabulate(true)}>
+            Cast Ballot As Is
           </Button>
         </p>
-        <p>
-          <Button small onPress={() => setConfirmTabulate(true)}>
-            Count my ballot
-          </Button>
-        </p>
-        <Text italic>Ask a poll worker if you need help.</Text>
+        <Text italic small>
+          Ask a poll worker if you need help.
+        </Text>
       </CenteredLargeProse>
       {confirmTabulate && (
         <Modal
@@ -66,7 +74,7 @@ function OvervoteWarningScreen({
             <Prose textCenter>
               <h1>Are you sure?</h1>
               <p>
-                Your votes for {pluralize('contest', overvotes.length, true)}{' '}
+                Your votes in {pluralize('contest', contestNames.length, true)}{' '}
                 will not be counted.
               </p>
             </Prose>
@@ -74,7 +82,7 @@ function OvervoteWarningScreen({
           actions={
             <React.Fragment>
               <Button primary onPress={scanner.acceptBallot}>
-                Yes, count my ballot
+                Yes, Cast Ballot As Is
               </Button>
               <Button onPress={() => setConfirmTabulate(false)}>Cancel</Button>
             </React.Fragment>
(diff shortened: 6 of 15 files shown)


frontends/precinct-scanner/src/screens/insert_ballot_screen.tsx
```diff
@@ -23,7 +23,7 @@ export function InsertBallotScreen({
       {showNoChargerWarning && (
         <Text warning small center>
           <strong>No Power Detected.</strong> Please ask a poll worker to plug
```
mattroe, line 25: I don't think this state or `setup_power_page` is ever shown in the new app experience. This copy is better though, so let's use it in `setup_scanner_screen`, which is shown whenever power is disconnected.

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -46,7 +48,7 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
     <ScreenMainCenterChild infoBar={false}>
       <TimesCircle />
       <CenteredLargeProse>
-        <h1>Scanning Error</h1>
+        <h1>Ballot Not Counted</h1>
         <p>{errorMessage}</p>
         <Text italic>Ask a poll worker for help.</Text>
```
mattroe, line 53: We should also probably prompt to remove ballot?

> Remove ballot and ask a poll worker for help.

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -19,13 +19,15 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
...
       case 'invalid_election_hash':
-        return 'Scanned ballot does not match the election this scanner is configured for.';
+        return 'The ballot does not match the election this scanner is configured for.';
       case 'invalid_precinct':
-        return 'Scanned ballot does not match the precinct this scanner is configured for.';
+        return 'The ballot does not match the precinct this scanner is configured for.';
       case 'unreadable':
-        return 'There was a problem reading this ballot. Please try again.';
+        return 'There was a problem reading this ballot. Please scan again.';
       case 'unknown':
```
mattroe, line 31: Based on the latest implementation, I believe this specific `unknown` state is non-recoverable by removing the ballot, so we may want to prompt a poll worker to restart the tablet. No need to change this yet, until we finalize the implementation as we have a bunch of pending fixes that may make this point moot.

frontends/precinct-scanner/src/screens/scan_jam_screen.tsx
```diff
@@ -11,7 +11,7 @@ export function ScanJamScreen(): JSX.Element {
     <ScreenMainCenterChild infoBar={false}>
       <TimesCircle />
       <CenteredLargeProse>
-        <h1>Scanner Jammed</h1>
```
mattroe, line 14: This the most important screen to convey that a ballot was not counted explicitly.

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
```diff
@@ -158,8 +166,10 @@ test('unreadable', () => {
     ],
   });
 
-  screen.getByRole('heading', { name: 'Review Your Ballot' });
-  userEvent.click(screen.getByRole('button', { name: 'Count my ballot' }));
-  userEvent.click(screen.getByRole('button', { name: 'Yes, count my ballot' }));
+  screen.getByRole('heading', { name: 'Can’t Interpret Ballot' });
```
mattroe, line 169: I have never seen this screen. Does it actually occur? We should be going to a scan error screen in this case and not giving the voter an opportunity to cast an uninterpretable ballot.

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
```diff
@@ -115,10 +119,12 @@ test('undervote by 1', () => {
 
   screen.getByRole('heading', { name: 'Review Your Ballot' });
   screen.getByText(
-    new RegExp(`You can vote for more people for ${contest.title}.`)
+    new RegExp(`You may vote for more candidates for: ${contest.title}.`)
```
mattroe, line 122: The undervote screen uniquely provides an `or` option between the two buttons. Was this intentional? Let's make this consistent between other review screens.

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
```diff
@@ -59,9 +61,9 @@ test('blank ballot', () => {
     adjudicationReasonInfo: [{ type: AdjudicationReason.BlankBallot }],
   });
 
-  screen.getByRole('heading', { name: 'Blank Ballot' });
-  screen.getByText('Your ballot does not have any votes.');
-  userEvent.click(screen.getByRole('button', { name: 'Count blank ballot' }));
+  screen.getByRole('heading', { name: 'No Votes Detected' });
```
mattroe, line 64: Nit: keep consistent language between `detected` and `found`. Suggest we simplify header with `No Votes Found`

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -118,26 +125,27 @@ function UndervoteWarningScreen({
...
           <Text small>
-            You did not vote for {toSentence(blankContestNames, ', ', ' or ')}.
+            No votes detected for: {toSentence(blankContestNames, ', ', ' or ')}
+            .
           </Text>
         )}
         {partiallyVotedContestNames.length > 0 && (
           <Text small>
-            You can vote for more people for{' '}
+            You may vote for more candidates for:{' '}
```
mattroe, line 134: It's not always going to be a candidate as it could be a ballot measure. We should use more generic language.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -226,20 +237,23 @@ function BlankBallotWarningScreen(): JSX.Element {
...
         <p>
           <Button primary large onPress={scanner.returnBallot}>
-            Return my ballot
+            Return Ballot
           </Button>
         </p>
         <p>
           <Button small onPress={() => setConfirmTabulate(true)}>
-            Count my ballot
+            Accept Ballot As Is
```
mattroe, line 256: We're using `Count Ballot As Is` in other places. Let's make this consistent. I prefer count over accept.

mattroe (changes requested):

frontends/precinct-scanner/src/screens/invalid_card_screen.tsx
```diff
@@ -11,7 +11,8 @@ export function InvalidCardScreen(): JSX.Element {
     <ScreenMainCenterChild infoBar={false}>
       <TimesCircle />
       <CenteredLargeProse>
-        <h1>Invalid Card, please remove.</h1>
+        <h1>Invalid Card</h1>
+        <p>Remove card the to continue.</p>
```
jonahkagan, line 15: Words are out of order in this sentence

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -19,13 +19,15 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
     switch (error) {
       // Invalid interpretation
       case 'invalid_test_mode':
-        return isTestMode ? 'Live Ballot detected.' : 'Test ballot detected.';
+        return isTestMode
+          ? 'Live Ballot detected. Scanner is in test mode.'
```
jonahkagan, line 23: uncapitalize "Ballot" for consistency

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
jonahkagan: So, this error occurs when either front or back is interpreted as 'UninterpretedHmpbPage'. I don't know what that actually means.

In any case, VxScan _will_ recover when you remove the ballot in this case.

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
```diff
@@ -59,9 +61,9 @@ test('blank ballot', () => {
     adjudicationReasonInfo: [{ type: AdjudicationReason.BlankBallot }],
   });
 
-  screen.getByRole('heading', { name: 'Blank Ballot' });
-  screen.getByText('Your ballot does not have any votes.');
```
jonahkagan, line 63: I think "Your ballot does not have any votes" is plainer language. It uses active voice and also the possessive "your ballot", both of which are recommended in the plain language guidelines. ([use of pronouns](https://www.plainlanguage.gov/guidelines/audience/address-the-user/)) ([active voice](https://www.plainlanguage.gov/guidelines/conversational/use-active-voice/))

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -31,31 +32,37 @@ function OvervoteWarningScreen({
...
+    .map((c) => c.title);
 
   return (
     <ScreenMainCenterChild infoBar={false}>
       <ExclamationTriangle />
       <CenteredLargeProse>
         <h1>Too Many Votes</h1>
         <Text small>
-          You voted for too many choices for {toSentence(contestNames)}. To fix
-          this, take your ballot and ask a poll worker for a new ballot.
```
jonahkagan, line 49: same comments about using pronouns and active voice

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -31,31 +32,37 @@ function OvervoteWarningScreen({
...
-          </Button>
-        </p>
-        <p>
-          <Button small onPress={() => setConfirmTabulate(true)}>
-            Count my ballot
+          <Button primary onPress={scanner.returnBallot}>
+            Return Ballot
+          </Button>{' '}
+          <Button onPress={() => setConfirmTabulate(true)}>
+            Count Ballot As Is
```
jonahkagan, line 65: I think "as is" could be replaced by something more common that could communicate the same idea. Maybe "Count Ballot Anyway" or just "Count Ballot"

I'm having a hard time putting my finger on why "as is" feels a little weird to me

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
jonahkagan: Are there ballot measures that allow you to vote for more than one choice? This case only shows for vote-for-n contests where n>1

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -118,26 +125,27 @@ function UndervoteWarningScreen({
...
           </Button>
         </p>
-        <p>
-          <Button small onPress={scanner.returnBallot}>
-            Return my ballot
-          </Button>
-        </p>
-        <Text italic>Ask a poll worker if you need help.</Text>
+        <Text italic>
+          Marking votes in every contest is not required.
```
jonahkagan, line 145: Maybe to make this more "plain," something like:

> You don't have to vote for every contest.

or 

> Your votes will be counted even if you leave some contests blank.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -154,7 +162,7 @@ function UndervoteWarningScreen({
                 )}
                 {partiallyVotedContestNames.length > 0 && (
                   <span>
-                    You can vote for more people in{' '}
+                    You may still vote in{' '}
```
jonahkagan, line 165: I think "can vote for more people" is more specific about what they can actually still do than "may still vote"

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
beausmith: So should I update this state to add a returned message "Please remove your ballot."?

image

beausmith: Perhaps no action needed here if we update the last sentence per Roe's suggestion: https://github.com/votingworks/vxsuite/pull/2276/files#r938938765

jonahkagan: I would like to know why this happens. @benadida do you have any idea what 'UninterpretedHmpbPage' means? As far as I can tell from diving into the code, it's whenever there's an error interpreting a HMPB, but I'm not sure what those errors might be.

I'm tempted to just treat this the same as "unreadable" and ask them to try again. I'm sure they'll eventually give up if it doesn't work a few times.

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
jonahkagan: It's not guaranteed that the ballot will be held in the front in all of these cases. E.g. `plustek_error` could happen at any time.

Since this screen handles both invalid interpretations and other scanning errors, I'm not sure it even makes sense to always show this message. I suggest we move this message to be a part of the `switch` above so that we can determine when to show something about asking a poll worker or not.

Also, I'm realizing that other pages say "Ask a poll worker if you need help," which might be more appropriate in some cases when there's a mitigation that the voter can take (e.g. removing ballot and trying again). I think "Ask a poll worker for help" is better for cases when there's no voter mitigation available.

Happy to help distinguish which cases are which if that's helpful!

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
beausmith: It was intentional when the buttons were on one line, but I see that the overvote screen didn't have this. fixed.

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
beausmith: @jonahkagan - Your thoughts here?

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
benadida: That's right, it is things like can't find the expected contest boxes, either not right shape or number. Could be because of some marginal marks confusing things.

This is also likely the error that we'll use if we detect too weird a skew.

This should result at least in having the voter try again once or twice.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: I know that "As Is" seems a little weird, but is it clear?

I feel like it's not as awkward when said aloud. Maybe because of the way it looks when spelled?

I believe we do want to make the voter pause here, so I don't mind if "As Is" causes the voter to pause and make the correct selection here.

I'm also wondering if "Count" is the wrong word to use because the ballot will not be "counted" if there is an overvote. Perhaps the label should use "Accept" as in "Accept Ballot Anyway" or "Accept Ballot As Is"

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
jonahkagan: I don't know how to make it happen either.

Here are the possible adjudication reasons:
```ts
export enum AdjudicationReason {
  UninterpretableBallot = 'UninterpretableBallot',
  MarginalMark = 'MarginalMark',
  Overvote = 'Overvote',
  Undervote = 'Undervote',
  BlankBallot = 'BlankBallot',
}
```

It looks like Overvote, Undervote, and BlankBallot are handled by other screens, so that leaves just UninterpretableBallot and MarginalMark.

As far as I can tell, UninterpretableBallot only occurs [when we can't correctly identify the contests](https://github.com/votingworks/vxsuite/blob/ad22e6c64fce8e4bcbbd17f2656d40a947f1a009/libs/utils/src/hmpb/ballot_adjudication_reasons.ts#L28), but I'm not actually sure why that happens and why we don't just return an `UnreadablePage` or `UninterpretedHmpbPage` interpretation.

MarginalMark I also don't know much about.

Maybe @eventualbuddha or @benadida can shed some light

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
mattroe: Thanks for clarifying that. We should be ok in that case.

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
benadida: `MarginalMark` is for future use, not currently in use.

`UninterpretableBallot`... I'm not entirely sure! Is it important to determine now? If so I can dig.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
jonahkagan: Might have to test this one on some voters to get to the bottom of it.

Re: the word "count" - I was thinking about this a bit when editing these screens. I opted for "Count" since we have similar language elsewhere (e.g. "ballots counted" and "ballot not counted", etc). So I wanted it to be consistent. I also think "count" is a very plain word - plainer than "accept."

That being said, I think the word that really comes to mind here to describe what's happening is "cast." I'm wondering if "cast" feels plain enough or if it's too jargon-y. We could potentially use "cast" to refer to ballots and "count" to refer to specific votes.

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
jonahkagan: I think @mattroe should weigh in on how important this is to figure out now.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: @mattroe - I had the same thought at first, then realized that only candidate contests can have more than one vote. yesno contests and mississippi either neither don't apply.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: Thanks. I'm leaning towards using two different phrases to say the same thing in two separate ways with the goal of ensuring that the message is communicated.

More active voice on the page:
> Your votes will count, even if you leave some contests blank.

Confirmation modal:
> You are not required to vote in every contest.

beausmith: Upon further editing, I'm leaning towards a slightly more generic phrase in both locations:

> Your votes will count, even if you leave some blank.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: Agreed. Implementing with "candidates" instead of "people".

> You can vote for more candidates in 2 contests.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: Can do… however in this case nothing can be "counted" because the document being scanned is not a ballot. Hence "accept" rather than "count".

Does this change your mind?

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
beausmith: Thanks. Updating `unknown` and `unreadable` to display the same message:

> There was a problem reading this ballot. Please scan again.

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: I think "cast" should be find in the context of voting:
- A voter "casts" their ballot.
- The scanner "counts" ballots and votes.

I'm changing all to "Cast Ballot As Is" instead of "count" or "accept".

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: Changing to "Cast Ballot As Is"

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
mattroe: Not important to figure out now. I suggest we present the generic scanning error screen in the case of this improbable uninterpretable ballot case.

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
beausmith: All screens should ideally inform/empower the voter to resolve the issue themself before asking a poll worker for help.

If the voter removes their ballot, the error screen will no longer be displayed. So they will have to scan again before they can show a poll worker what the error is.

Jonah is adding the additional previews. We can determine where we should to explicitly prompt the voter to remove their ballot once these are available.

frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx
beausmith: Uninterpretable Ballot state has been updated to look like this:

image

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -17,27 +17,31 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
...
+        return 'The ballot does not match the precinct this scanner is configured for.';
       case 'unreadable':
-        return 'There was a problem reading this ballot. Please try again.';
       case 'unknown':
-        return undefined;
-      // Precinct scanner error
+        return 'There was a problem reading this ballot. Please scan again.';
+      // Precinct scanner errors
       case 'both_sides_have_paper':
+        return 'Scanner detected unexpected ballots in both the front and back. Remove both ballots to continue.';
```
jonahkagan, line 34: So this case is a bit complex:
- When a ballot is inserted in the front while there's already a ballot in the scanner, we show the scanner busy screen
- After the front ballot is removed, we'll try to continue scanning. In some cases, we can't continue, so the back ballot will be rejected and held in front, at which point this message will show.

So I might just say something like:
> Remove the ballot and try again.

Or if we really want to explain what happened, maybe something like:
> Scanning interrupted. Remove the ballot and try again.

But I don't think the voter necessarily will care or need to know in this case.

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -17,27 +17,31 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
...
       case 'paper_in_front_on_startup':
+        return 'Scanner detected an unexpected ballot in the tray. Remove ballot to continue.';
       case 'paper_in_back_on_startup':
       case 'paper_in_back_after_accept':
-        return 'Take your ballot out of the tray and try again.';
+        return 'Scanner detected an unexpected ballot at the back of the scanner. Remove ballot to continue.';
       case 'unexpected_paper_status':
+        return 'Scanner detected an unexpected ballot in scanner. Remove ballot to continue.';
```
jonahkagan, line 41: This case should be handled equivalently to `unexpected_event` and `plustek_error` - they have slightly different causes but all basically mean that Plustek did something weird that we didn't expect.

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -17,27 +17,31 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
...
       case 'unexpected_event':
       case 'plustek_error':
-        return undefined;
+        return 'The scanner experienced an error and needs to be reset. Please turn off and unplug the scanner from the power outlet. Plug it back in and turn it on again.';
```
jonahkagan, line 44: We're still working out what actually will be needed to recover from these cases. Currently, we're experimenting with the scan service automatically disconnecting and reconnecting to Plustek to try to mitigate this without any need for user action (with a 5s delay).

If we stick with that, then maybe this should just have a spinner or something? And then have a timeout in the frontend where if this screen shows for more than 5s we switch to something like the error message you have here.

Now that I'm thinking about it, I wonder if just asking for a manual reset might be clearer than having this automatic recovery. Worth thinking through

beausmith: So then, is this safe to commit in the meantime?

jonahkagan: Sure!

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -111,33 +119,56 @@ function UndervoteWarningScreen({
...
+    partiallyVotedContestNamesDisplayCount
+  );
+  if (partiallyVotedContestNamesDisplayRemainder) {
+    partiallyVotedContestNamesDisplay.push(
+      `${partiallyVotedContestNamesDisplayRemainder} more ${pluralize(
+        'contest',
+        partiallyVotedContestNamesDisplayRemainder
+      )}`
+    );
+  }
```
jonahkagan, line 142: does this logic need to apply to `blankContestNames` as well?

frontends/precinct-scanner/src/screens/setup_power_page.tsx
```diff
@@ -12,9 +12,9 @@ export function SetupPowerPage(): JSX.Element {
           <h1>
             No Power Detected <NoWrap>and Battery is Low</NoWrap>
           </h1>
-          <p>
-            Please ask a poll worker to plug-in the power cord for this machine.
-          </p>
+          <Text italic>
+            Please ask a poll worker to plug-in the power cord.
+          </Text>
```
jonahkagan, line 17: Should we just remove this screen since we are requiring power to be plugged in all the time whether or not battery is low?

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: Yes. Will create shared function.

frontends/precinct-scanner/src/screens/setup_power_page.tsx
beausmith: Yes. Can do. Seems like a separate PR though? I'll queue it up after this one.

jonahkagan: agreed!

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -17,29 +17,33 @@ export function ScanErrorScreen({ error, isTestMode }: Props): JSX.Element {
...
+        return 'The ballot does not match the precinct this scanner is configured for.';
       case 'unreadable':
-        return 'There was a problem reading this ballot. Please try again.';
       case 'unknown':
-        return undefined;
-      // Precinct scanner error
+        return 'There was a problem reading this ballot. Please scan again.';
+      // Precinct scanner errors
       case 'scanning_failed':
+        return 'Scanning Failed. Remove ballot to continue.';
```
jonahkagan, line 34: This case happens when the scanner repeatedly tries to pull the ballot in but it doesn't work (e.g. if you hold onto the ballot and don't let it scan). Eventually we give up trying to scan it.

I think we can recommend removing the ballot and trying again.

beausmith: How about this?
> Ballot not fully inserted. Remove ballot to continue.

FWIW, I have removed all the "and try again" language because I feel that inserting a ballot and trying again is implicit the use of a ballot scanner. Additionally, once the voter follows the direction "Remove ballot to continue" the screen updates to be the "Insert Ballot" screen.

I'm going to commit this assuming you'll agree. However I expect you'll tell me if otherwise! 😄

jonahkagan: Sounds good!

jonahkagan (approved):

mattroe (approved):

PR #2283, opened 2022-08-06
Print All Ballot Styles At Once

### LINKED ISSUE #2175: VxAdmin: enable printing of all ballot styles for a given ballot type
Warren County m13 feedback:
> VxAdmin – Print test ballots for proofing – need a way to print all precincts, with one click, instead of printing one at a time. When working down the list of precincts, when I go back, I am at the top of the page again, and have to scroll back down.

VxAdmin should enable a user to print one copy of all ballot styles for a given ballot type [official/test/sample] in a one-click manner.

Printing of all ballot styles should be available for Election Managers, but System Administrators should only be able to print draft ballots.

adghayes:
_Recommend commit-by-commit review_

## Overview
Closes #2175. Adds feature to the ballot list screen to allow printing all ballot styles for a given ballot type at once.

## Demo Video or Screenshot
Admin Happy Path:
[print-all-ballots-flow.webm](https://user-images.githubusercontent.com/37960853/183234868-d59431b8-3aa8-4177-9f97-cf5bee3b4dd8.webm)

Superadmin Screenshot:
print-all-draft-ballots

No Printer Handling:
[print-all-ballots-no-printer.webm](https://user-images.githubusercontent.com/37960853/183234893-4c203c67-59bf-44b7-8a14-cd92ac93f7b4.webm)

## Testing Plan 
- New test suite for new component
- Manual testing

## Checklist

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -30,7 +30,6 @@ import {
   fakePrinterInfo,
   fakeUsbDrive,
   makeAdminCard,
-  makeSuperadminCard,
   mockOf,
 } from '@votingworks/test-utils';
 import {
@@ -55,6 +54,10 @@ import {
 import { MachineConfig } from './config/types';
 import { VxFiles } from './lib/converters';
 import { createMemoryStorageWith } from '../test/util/create_memory_storage_with';
+import {
+  authenticateWithAdminCard,
+  authenticateWithSuperAdminCard,
+} from '../test/util/authenticate';
 
 const EITHER_NEITHER_CVR_DATA = electionWithMsEitherNeitherFixtures.cvrData;
 const EITHER_NEITHER_CVR_FILE = new File([EITHER_NEITHER_CVR_DATA], 'cvrs.txt');
@@ -146,39 +149,6 @@ afterEach(() => {
   delete window.kiosk;
   MockDate.reset();
 });
-
-async function authenticateWithAdminCard(card: MemoryCard) {
-  // Machine should be locked
-  await screen.findByText('VxAdmin is Locked');
-  card.insertCard(
-    makeAdminCard(eitherNeitherElectionDefinition.electionHash, '123456')
-  );
-  await screen.findByText('Enter the card security code to unlock.');
-  userEvent.click(screen.getByText('1'));
-  userEvent.click(screen.getByText('2'));
-  userEvent.click(screen.getByText('3'));
-  userEvent.click(screen.getByText('4'));
-  userEvent.click(screen.getByText('5'));
-  userEvent.click(screen.getByText('6'));
-  await screen.findByText('Remove card to continue.');
-  card.removeCard();
-  await screen.findByText('Lock Machine');
-}
-
-async function authenticateWithSuperAdminCard(card: MemoryCard) {
-  await screen.findByText('VxAdmin is Locked');
-  card.insertCard(makeSuperadminCard());
-  await screen.findByText('Enter the card security code to unlock.');
-  userEvent.click(screen.getByText('1'));
-  userEvent.click(screen.getByText('2'));
-  userEvent.click(screen.getByText('3'));
-  userEvent.click(screen.getByText('4'));
-  userEvent.click(screen.getByText('5'));
-  userEvent.click(screen.getByText('6'));
-  await screen.findByText('Remove card to continue.');
-  card.removeCard();
-  await screen.findByText('Lock Machine');
-}
 
 test('create election works', async () => {
   const card = new MemoryCard();
@@ -350,7 +320,7 @@ test('L&A (logic and accuracy) flow', async () => {
   const { container, getByTestId } = render(
     <App card={card} hardware={hardware} printer={printer} storage={storage} />
   );
-  await authenticateWithAdminCard(card);
+  await authenticateWithAdminCard(card, eitherNeitherElectionDefinition);
 
   userEvent.click(screen.getByText('L&A'));
 
@@ -438,7 +408,7 @@ test('L&A features are available after test results are loaded', async () => {
   });
   render(<App card={card} hardware={hardware} storage={storage} />);
 
-  await authenticateWithAdminCard(card);
+  await authenticateWithAdminCard(card, eitherNeitherElectionDefinition);
 
   // Confirm that test results are loaded
   userEvent.click(screen.getByText('Tally'));
@@ -463,7 +433,7 @@ test('printing ballots and printed ballots report', async () => {
     <App storage={storage} printer={printer} card={card} hardware={hardware} />
   );
   jest.advanceTimersByTime(2000); // Cause the usb drive to be detected
-  await authenticateWithAdminCard(card);
+  await authenticateWithAdminCard(card, eitherNeitherElectionDefinition);
 
   fireEvent.click(getByText('Reports'));
   await screen.findByText('0 official ballots');
@@ -523,7 +493,7 @@ test('tabulating CVRs', async () => {
     <App storage={storage} card={card} hardware={hardware} printer={printer} />
   );
   jest.advanceTimersByTime(2000); // Cause the usb drive to be detected
-  await authenticateWithAdminCard(card);
+  await authenticateWithAdminCard(card, eitherNeitherElectionDefinition);
 
   fireEvent.click(getByText('Reports'));
   expect(getByTestId('total-ballot-count').textContent).toEqual('100');
@@ -704,7 +674,7 @@ test('tabulating CVRs with SEMS file', async () => {
     <App storage={storage} card={card} hardware={hardware} />
   );
   jest.advanceTimersByTime(2000);
-  await authenticateWithAdminCard(card);
+  await authenticateWithAdminCard(card, eitherNeitherElectionDefinition);
 
   fireEvent.click(getByText('Tally'));
   getByText('External Results (sems-results.csv)');
@@ -814,7 +784,7 @@ test('tabulating CVRs with SEMS file and manual data', async () => {
   const { getByText, getByTestId, getAllByText, queryAllByText } = render(
     <App storage={storage} card={card} hardware={hardware} />
   );
-  await authenticateWithAdminCard(card);
+  await authenticateWithAdminCard(card, eitherNeitherElectionDefinition);
 
   fireEvent.click(getByText('Tally'));
   expect(getByTestId('total-cvr-count').textContent).toEqual('200');
@@ -997,7 +967,7 @@ test('changing election resets sems, cvr, and manual data files', async () => {
     <App storage={storage} card={card} hardware={hardware} />
   );
 
-  await authenticateWithAdminCard(card);
+  await authenticateWithAdminCard(card, eitherNeitherElectionDefinition);
 
   fireEvent.click(getByText('Reports'));
   await screen.findByText('0 official ballots');
@@ -1052,7 +1022,7 @@ test('clearing all files after marking as official clears SEMS, CVR, and manual 
       converter="ms-sems"
     />
   );
-  await authenticateWithAdminCard(card);
+  await authenticateWithAdminCard(card, eitherNeitherElectionDefinition);
 
   fireEvent.click(getByText('Reports'));
   await screen.findByText('0 official ballots');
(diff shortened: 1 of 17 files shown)


adghayes:
After completing the feature, I encountered a weird bug where infrequently, one of the ballots would not print and instead a blank page would print. I couldn't identify a gap in the code but presumably - somewhere between a ballot being rendered and it being printed - there is a re-render which leaves it blank. This could during asynchronous gaps in the `print` function or in `hand_marked_paper_ballot`'s post-render flow, or it could be in this component.

In in any case, I traced a lot of spurious re-renders to a callback function `addPrintedBallots` that this component gets from `AppContext` and uses to log official ballots were printed. That function was being created on every render, so I replaced it with a `useCallback`. Dependencies forced me to  make a number of other functions `useCallback`s, so I just made all the callbacks I found in `app_root.tsx` into `useCallback`s. This reduced the re-renders I was seeing and I have not been able to reproduce the issue since.

`useCallback` is just syntactic sugar for `useMemo`, which is not supposed to be used to prevent re-renders. So while this may be common practice, it is not ideal. However, the underlying problem and the challenge of printing sequences of documents seems not worth diving into right now until we re-investigate printing writ large.

However, the functions passed through `AppContext` should be `useCallback` anyway. We have a lint rule which requires that functions passed as props be made with `useCallback`. Passing through `AppContext` is a sort of back door that skirts the lint rule, and reducing the re-renders is a useful optimization.

frontends/election-manager/src/components/print_ballot_button_text.tsx
```diff
@@ -0,0 +1,45 @@
...
+  Election,
+  getElectionLocales,
+} from '@votingworks/types';
+import pluralize from 'pluralize';
+import { DEFAULT_LOCALE } from '../config/globals';
+import { BallotMode, ballotModeToReadableString } from '../config/types';
+import { getHumanBallotLanguageFormat } from '../utils/election';
+
+interface Props {
+  ballotCopies: number;
```
arsalansufi, line 13: Very minor: I like to name numerical props like this `numBallotCopies` or `numberOfBallotCopies` for clarity. Not a huge deal because we've already typed it as a number

frontends/election-manager/src/app.test.tsx
```diff
@@ -148,36 +151,10 @@ afterEach(() => {
 });
 
 async function authenticateWithAdminCard(card: MemoryCard) {
-  // Machine should be locked
-  await screen.findByText('VxAdmin is Locked');
-  card.insertCard(
-    makeAdminCard(eitherNeitherElectionDefinition.electionHash, '123456')
+  return authenticateWithAdminCardNoDefinition(
```
arsalansufi, line 154: Interesting - I see we're wrapping in this way so that the changes to this test file are minimal.

That said, renaming `authenticateWithAdminCard` from `../test/util` `authenticateWithAdminCardNoDefinition` could be confusing since that function does require and use an election definition.

A clearer alternative might be to have the util function take an optional election definition that defaults to `eitherNeitherElectionDefinition`:

```ts
export async function authenticateWithAdminCard(
  card: MemoryCard,
  electionDefinition: ElectionDefinition = eitherNeitherElectionDefinition
): Promise<void>
```

Then, no wrapping should be necessary in this file

frontends/election-manager/src/components/print_all_ballots_button.tsx
```diff
@@ -0,0 +1,362 @@
...
+      );
+      actions = <LinkButton onPress={closeModal}>Cancel</LinkButton>;
+      break;
+
+    case 'options':
+      mainContent = (
+        <Prose>
+          <h1>Print All Ballot Styles</h1>
+          <p>Select the ballot type and number of copies to print:</p>
+          <CenteredOptions>
```
arsalansufi, line 263: Instead of using custom styles, is there a prop higher up that we could rely on, e.g. the `Modal` `centerContent` prop or the `Prose` `textCenter` prop?

frontends/election-manager/src/components/print_all_ballots_button.tsx
```diff
@@ -0,0 +1,362 @@
...
+  useEffect(() => {
+    if (hasPrinterAttached) {
+      if (modalState === 'no-printer') {
+        setModalState('options');
+      }
+    } else if (!hasPrinterAttached) {
+      if (modalState === 'options') {
+        setModalState('no-printer');
+      }
+    }
```
arsalansufi, line 122: This might be a bit easier to parse if we collapse the nested if statements, e.g.

```ts
if (hasPrinterAttached && modalState === 'no-printer') {
  ...
} else if (!hasPrinterAttached && modalState === 'options') {
  ...
}
```

frontends/election-manager/src/components/print_all_ballots_button.tsx
```diff
@@ -0,0 +1,362 @@
...
+    [election]
+  );
+
+  useEffect(() => {
+    if (draftMode) {
+      setBallotMode(BallotMode.Draft);
+    } else {
+      setBallotMode(BallotMode.Official);
+    }
+  }, [draftMode]);
```
arsalansufi, line 111: I believe this is to reset the ballot mode between closes and re-opens of the modal, is that right? If so, might make sense to just reset the ballot mode in `closeModal` instead. (I may be missing something here!)

frontends/election-manager/src/components/print_all_ballots_button.tsx
```diff
@@ -0,0 +1,362 @@
...
+              election={election}
+            />
+          </Button>
+          <LinkButton onPress={closeModal}>Cancel</LinkButton>
+        </React.Fragment>
+      );
+      break;
+
+    case 'printing': {
+      assert(typeof ballotIndex === 'number');
```
arsalansufi, line 308: The purpose of this assertion might be clearer if you instead use:

```ts
assert(ballotIndex !== undefined);
```

I could see someone wondering why this is necessary when the state value has a `number` type

frontends/election-manager/src/components/print_all_ballots_button.test.tsx
```diff
@@ -0,0 +1,248 @@
...
+beforeEach(() => {
+  jest.useFakeTimers();
+});
+
+async function authenticateWithAdminCard(card: MemoryCard) {
+  return authenticateWithAdminCardNoDefinition(
+    card,
+    electionMinimalExhaustiveSampleDefinition
+  );
+}
```
arsalansufi, line 44: Same comment here as earlier about the `authenticateWithAdminCardNoDefinition` function name being a bit confusing (though the fix here may look different than the fix in `app.test.tsx` since we're using a different election definition)

arsalansufi (commented):
Looks great! A few comments

frontends/election-manager/src/app_root.tsx
```diff
@@ -254,20 +253,26 @@ export function AppRoot({
...
+    [setStorageKeyAndLog]
+  );
 
-  async function addPrintedBallot(printedBallot: PrintedBallot) {
-    const ballots = await getPrintedBallots();
-    ballots.push(printedBallot);
-    await savePrintedBallots(ballots);
-    setPrintedBallots(ballots);
-  }
+  const addPrintedBallot = useCallback(
```
arsalansufi, line 267: > In in any case, I traced a lot of spurious re-renders to a callback function addPrintedBallots that this component gets from AppContext and uses to log official ballots were printed. That function was being created on every render, so I replaced it with a useCallback. Dependencies forced me to make a number of other functions useCallbacks, so I just made all the callbacks I found in app_root.tsx into useCallbacks. This reduced the re-renders I was seeing and I have not been able to reproduce the issue since.

Glad that this fixed our downstream issues for now! And this seems totally reasonable as a temporary fix for m14. That said, I could definitely see someone down the road removing the `useCallback` wrappings in this file, not realizing that they're relied on downstream.

Could we apply a temporary fix further downstream, like we do for test deck printing?

https://github.com/votingworks/vxsuite/blob/80984f19c14e39e8303fee2972ee33f420a01ecb/frontends/election-manager/src/screens/print_test_deck_screen.tsx#L246-L258

arsalansufi:
Also just wanted to give the heads up that https://github.com/votingworks/vxsuite/pull/2284 will likely cause some merge conflicts if I merge it before you merge this - more than happy to help address those if the need comes up!

frontends/election-manager/src/app_root.tsx
adghayes: I tried memoizing the component and it didn't do anything. I didn't devote the time into digging into why! I think it's because in this case, the prop value passed to `React.memo` does in fact change (the callback from `AppContext`) so it is re-rendered anyway. I think we have to memoize further upstream.

That being said, I understand the concern. Upstream memoization is a recipe for someone hitting this same issue later.

frontends/election-manager/src/components/print_all_ballots_button.tsx
adghayes: When you close the modal in admin mode, you don't know whether the next time it will be opened in admin or superadmin mode. So I don't think it can be reset in `closeModal`?

The purpose is not to reset between closes, but to reset between userRole switches.

frontends/election-manager/src/components/print_ballot_button_text.tsx
adghayes: Going to leave this one as is, just to avoid changing variable names in other files.

frontends/election-manager/src/components/print_all_ballots_button.tsx
adghayes: It seemed weird to me to wrap a `Prose` in a `Prose`, but that would do the trick. Would you prefer that?

frontends/election-manager/src/app.test.tsx
adghayes: I ended up just using the helper with no default. I think for people using this helper in the future, it will be easier to force them to specify an election definition rather than risk potential confusion of why their tests are failing due to the wrong card.

adghayes: Also the real reason is I have a vendetta against the `eitherNeitherElectionDefinition` :)

arsalansufi: LOL that makes a lot of sense to me! To always have an explicit specification

frontends/election-manager/src/components/print_all_ballots_button.tsx
arsalansufi: Ah yes, I totally agree that a `Prose` in a `Prose` doesn't make sense - I was actually wondering if we could apply `textCenter` to the existing outer `Prose`

That said, I'm remembering now that the header and "Select the ballot type" prompt are currently left aligned. If we're okay with centering everything (I think that would look okay but understand if that's not what we want), then that should work! If we intentionally want those elements left aligned, then the current solution is fine!

frontends/election-manager/src/app_root.tsx
arsalansufi: Got it! In that case, let's just open a ticket to document that we want to find a more robust solution after m14

frontends/election-manager/src/components/print_all_ballots_button.tsx
arsalansufi: Got it! The current approach makes sense then

In the case of a user role switch, I imagine that this component would completely unmount and state would be cleared, making this `useEffect` unnecessary. (I could be wrong here.) But it also doesn't hurt to have it

arsalansufi (approved):
Responded to all your comments and while some small changes may follow, LGTM!

frontends/election-manager/src/components/print_all_ballots_button.tsx
adghayes: For no particularly good reason (other than perhaps being consistent with the other modals in the app) I prefer the left-align, so I will leave it as is.

frontends/election-manager/src/components/print_all_ballots_button.tsx
adghayes: I hit that issue and thought I needed it, but I think that was because of a different issue. I tested and it is unnecessary on latest, so I'll remove it!

frontends/election-manager/src/components/print_all_ballots_button.tsx
arsalansufi: > other than perhaps being consistent with the other modals in the app

^ I think that's a great reason!

PR #2285, opened 2022-08-08
fix(bmd): Reduce post-ballot printing instructions timeout and make timeout actually work

### LINKED ISSUE #1823: VxMark: "Done" screen is shown too long
When a ballot is printed, the final instructions are shown for a very long and should timeout after a short period of time.

Suggest 30 seconds as the timeout.

taravancil:
## Overview
Closes #1823

This drops the timeout for showing post-ballot-printing instructions from 60s to 30s and removes `hidePostVotingInstructions` as a dependency in the effect. `hidePostVotingInstructions` was updating repeatedly, I _believe_ because it has a dependency on `auth`.

## Checklist


arsalansufi (approved):

mattroe:
@taravancil did you confirm that the timeout is actually being applied? per my comment in https://github.com/votingworks/vxsuite/issues/1823, I have not been seeing the Done screen cleared at all even after the longer timeout when testing on the latest VxMark in dev.

frontends/bmd/src/app_root.tsx
```diff
@@ -434,7 +434,8 @@ export function AppRoot({
     return () => {
       clearTimeout(PostVotingInstructionsTimeout.current);
     };
-  }, [showPostVotingInstructions, hidePostVotingInstructions]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [showPostVotingInstructions]);
```
arsalansufi, line 438: Ooh very interesting - how exactly does ignoring updates to `hidePostVotingInstructions` make the timeout actually work?

taravancil: `hidePostVotingInstructions` was updating repeatedly, I believe because it has a dependency on `auth`. So the effect was being cleaned up and the timeout was being cleared repeatedly, thus never hitting the 30s threshold.

arsalansufi: Got it! Maybe worth adding either a brief explanatory comment or a test to make sure someone doesn't accidentally add the dependency back in

arsalansufi (approved):
One last thought, then LGTM!

Issue #2230, opened 2022-07-28
VxScan: blank ballots are considered undervotes when both are adjudication reasons


mattroe:
## Description
Blank ballots are considered undervotes when both cases are precinct scanner adjudication reasons.

## Repro
1. Program VxScan with an election definition where both `Undervote` and `BlankBallot` are included in `precinctScanAdjudicationReasons`
2. Scan a blank ballot
3. The undervote review screen is shown with all contests listed

## Expected behavior
The blank ballot review screen should be shown.

PR #2291, opened 2022-08-08
fix(scan): Fix blank ballot adjudication reasons

jonahkagan:
## Overview
Fixes: https://github.com/votingworks/vxsuite/issues/2230
I made a mistake by assuming that a blank ballot would only have one adjudication reason: BlankBallot. Instead, it has Undervotes for each contest followed by a BlankBallot at the end.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Manually tested
- Updated existing integration tests that were just wrong

### DIFF
--- a/services/scan/src/precinct_scanner_app.test.ts
+++ b/services/scan/src/precinct_scanner_app.test.ts
@@ -1,7 +1,6 @@
 import { MockScannerClient, ScannerClient } from '@votingworks/plustek-sdk';
 import {
   AdjudicationReason,
-  AdjudicationReasonInfo,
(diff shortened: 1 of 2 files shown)


benadida (approved):

eventualbuddha:
> I made a mistake by assuming that a blank ballot would only have one adjudication reason: BlankBallot. Instead, it has Undervotes for each contest followed by a BlankBallot at the end.

The reason for this, fwiw, is that you might have the `Undervote` adjudication reason enabled but not `BlankBallot`.

Issue #1749, opened 2022-04-28
flaky test: `test-integration-testing-election-manager` Tallies for candidate contests compute end to end as expected


eventualbuddha:
[<img width="1035" alt="image" src="https://user-images.githubusercontent.com/1938/165813001-75429a54-36b2-480b-b466-b5f3c118d4fd.png">](https://app.circleci.com/pipelines/github/votingworks/vxsuite/3659/workflows/63389bb8-17a6-4a3f-bd22-ecb36671118a/jobs/66993?invite=true#step-110-167)

eventualbuddha:
Another instance: https://app.circleci.com/pipelines/github/votingworks/vxsuite/3668/workflows/848bdffc-1d0a-49cc-8d30-de6a76ff73ab/jobs/67225

eventualbuddha:
Another instance: https://app.circleci.com/pipelines/github/votingworks/vxsuite/3668/workflows/dbf978ff-6802-431a-978c-eae5696668bf/jobs/67253?invite=true#step-110-167

jonahkagan:
Another: https://app.circleci.com/pipelines/github/votingworks/vxsuite/5129/workflows/588c4039-eecf-467a-bc91-dd127481811e/jobs/112477

PR #2292, opened 2022-08-08
Bump timeouts in VxAdmin integration test to minimize flakiness

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/1749

Some things are meant to be flaky 😋

pie

Others are not 🙅

failure-1

failure-2

There are at least two different points in this VxAdmin integration test that occasionally (actually fairly frequently) time out. The default Cypress timeout is 4 seconds, so I've doubled to 8 seconds at these two points.

I haven't been able to repro the error locally, so I won't close the linked issue until some time passes and no one sees this test fail in CI!


jonahkagan (approved):
Thanks for slicing into this one!

PR #2293, opened 2022-08-08
scanners: encourage user to wait for backup to finish

### LINKED ISSUE #2250: Scanners: make backups safer
Backups from VxScan and VxCentralScan can take a long time - several minutes. Exporting them is not currently very safe. 

In VxScan, you are offered the option to eject after the download is complete, but the ejection does not actually ensure that the file is saved to disk (#2248). We regularly see a delay much longer than 10s for the file to completely download to disk. 

In VxCentralScan we don't have a backup modal at all so, in addition to the ejection, you are not even explicitly prompted to eject. 

If we want to be safe, we need to at least complete #2248 and, ideally, edit the backup flow so that the backup `sync`s / flushes to disk while we are still telling the user we are backing up. If not, they will be confused why "Ejecting..." is taking so long.

adghayes:
## Overview
Closes #2250. The current backup flow makes it easy for the user to take the USB drive out too early. In this change, we call `sync` on the mounted USB drive before allowing the user to eject, ensuring that data is actually flushed to disk. In `bsd`, I introduce a very simple modal to prevent the user from exiting or clicking "Eject" in the corner while exporting a backup. Currently the only indicator is the disabled "Export Backup" button.

## Demo Video or Screenshot

Screen Shot 2022-08-08 at 3 58 46 PM (2)

## Testing Plan
- Updated tests in `bsd` and `precinct-scanner`
- Manual Testing (in VM with attached hardware)

Have not manually tested this on full hardware yet, which is important, but I think better to merge and test with a production image on Wednesday. We've used the `sync` API for CVRs in VxScan so I believe it is a safe operation on hardware.

## Checklist

### DIFF
--- a/frontends/bsd/src/screens/admin_actions_screen.test.tsx
+++ b/frontends/bsd/src/screens/admin_actions_screen.test.tsx
@@ -48,12 +48,17 @@ test('clicking "Export Backup" shows progress', async () => {
     backupButton.click();
     expect(backup).toHaveBeenCalledTimes(1);
 
-    // Verify progress message is shown.
-    await waitFor(() => screen.getByText('Exporting…'));
+    // Verify progress modal is shown.
+    await waitFor(() => {
+      const modal = screen.getByRole('alertdialog');
+      within(modal).getByText('Exporting Backup');
+      screen.getByText('Exporting…');
+    });
 
     // Trigger backup finished, verify back to normal.
     resolve();
     await waitFor(() => screen.getByText('Export Backup'));
+    expect(screen.queryAllByRole('alertdialog').length).toBe(0);
   });
 });
 
--- a/frontends/bsd/src/app_root.tsx
+++ b/frontends/bsd/src/app_root.tsx
@@ -391,6 +391,11 @@ export function AppRoot({ card, hardware, logger }: AppRootProps): JSX.Element {
 
   const backup = useCallback(async () => {
     await download('/scan/backup');
+    if (window.kiosk) {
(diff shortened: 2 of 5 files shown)


frontends/bsd/src/screens/admin_actions_screen.test.tsx
```diff
@@ -48,12 +48,17 @@ test('clicking "Export Backup" shows progress', async () => {
     backupButton.click();
     expect(backup).toHaveBeenCalledTimes(1);
 
-    // Verify progress message is shown.
-    await waitFor(() => screen.getByText('Exporting…'));
+    // Verify progress modal is shown.
+    await waitFor(() => {
+      const modal = screen.getByRole('alertdialog');
```
arsalansufi, line 53: Instead of wrapping all three of these lines in a `waitFor`, you could instead rely on `findBy` queries:

```ts
const modal = await screen.findByRole('alertdialog');
```

arsalansufi (approved):
LGTM!

frontends/bsd/src/screens/admin_actions_screen.test.tsx
adghayes: The `isExporting` state comes and goes so quickly, that if you use `findBy` queries, the latter ones will not pass because the component state will have already changed. Wrapping it in a `waitFor` catches a moment where all assertions are true.

arsalansufi: Ah very interesting, makes sense!

PR #2294, opened 2022-08-08
chore(ps): Add previews for all error cases

jonahkagan:
## Overview
Making sure that we can see all the possible error screens when using preview mode.


beausmith (approved):
LGTM.

Issue #2237, opened 2022-07-29
VxScan: app crashes if admin card is removed while on calibration success modal


mattroe:
## Description
VxScan crashes if an admin card is removed after calibration, but before clicking "Close" in the modal. A white screen is presented that is recoverable by restarting the tablet.

This does not occur if you close the modal before removing the card.

## Repro
1. Calibrate VxScan
2. After calibration, do not click close in the success modal.
3. Remove admin card
4. See crash

mattroe:
This specific behavior is a less common flow, but marking as high priority as it might occur in other modal states.

mattroe:
Confirmed this is specific to the calibration flow, so lowering priority, but @taravancil don't stop if this is already in progress.

PR #2298, opened 2022-08-09
fix(ps): Don't crash if card removed during calibration

jonahkagan:
## Overview
Fixes: https://github.com/votingworks/vxsuite/issues/2237

Previously, we asserted that the scanner would never be in the `calibrating` state while in voter mode. That's not true - if the election manager removes their card during calibration and the polls are open, it's possible. Since this isn't something we really expect to happen, we just handle it by rendering a blank screen (which will go away when calibration finishes). If we decide we need a screen here, we can come back and add it later. The goal of this PR is to not crash.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/530106/183534955-52309a36-8ec1-47a8-9f18-c6bf98c7ccba.mov

## Testing Plan 
- Manual test
- Added regression test

### DIFF
--- a/libs/api/src/services/scan/index.ts
+++ b/libs/api/src/services/scan/index.ts
@@ -659,7 +659,6 @@ export const PrecinctScannerStateSchema = z.enum([
   'rejecting',
   'rejected',
   'calibrating',
-  'calibrated',
   'jammed',
   'both_sides_have_paper',
   'error',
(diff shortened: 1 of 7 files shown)


libs/api/src/services/scan/index.ts
```diff
@@ -659,7 +659,6 @@ export const PrecinctScannerStateSchema = z.enum([
   'rejecting',
   'rejected',
   'calibrating',
-  'calibrated',
```
jonahkagan, line 662: this case was made obsolete in #2263

benadida (approved):

PR #2301, opened 2022-08-09
fix(interpret): error out if insufficient bubble-template overlap

benadida:
With particularly bad skew, in particular skew that changes as the page is scanned, our deskewing process can fail to sufficiently deskew. We already have a "bubble jiggle" process to align the marked ballot bubble with the template, and we've increased that jiggle previously to 7 pixels.

However, we can't rule out that worse skew will occur.

Rather than increase the jiggle some more, which could have some deleterious effects and which gets harder to test as it becomes quite rare, we want these rare situations to simply cause an error that requires a rescan. This won't happen very often, so it's okay.

The approach we take here is simple: if we cannot find sufficiently good alignment between the marked ballot and its template for a single bubble target area, even with the jiggle, then we are clearly in trouble, and we should error out and have the voter re-scan.

Experimentally, the threshold has been determined to be 85 pixels of non-overlap at our current pixel density. If there are more than 85 pixels in the template that we cannot align with the marked ballot, then we declare an error.


libs/ballot-interpreter-vx/src/interpreter/index.ts
```diff
@@ -699,6 +752,16 @@ export class Interpreter {
       }
     }
 
+    // if we couldn't find a close enough alignment, the skew is too much and we bail
+    if (
+      !bestMatchNewTemplatePixels ||
```
jonahkagan, line 757: Are meaning to throw an error if `bestMatchNewTemplatePixels` is `0` or `undefined` or both? Just making sure because it can be dangerous to use `!` on a number

jonahkagan (approved):

libs/ballot-interpreter-vx/src/interpreter/index.ts
benadida: great catch! only if undefined, I will correct. It's unlikely to ever be zero, but... yeah

PR #2303, opened 2022-08-09
feat(scan): Timeouts, retries, and automatic reconnect

jonahkagan:
## Overview
A bundle of resiliency improvements for the precinct scanner service:
- When Plustek fails to grab the ballot, we cap the number of times we will retry the scan command and then show an error to remove the ballot.
- We only will retry scanning on known errors (error feeding, error no paper) or when the paper status indicates the paper didn't get scanned.
- Sometimes the `scan` command will lag for a long time (up to 20s in some cases, triggering an internal Plustek timeout, or even infinitely if the scanner is in some sort of broken state). I've also seen some lag on the `get paper status` command. We timeout these commands after 10s.
- After an unexpected error (e.g. unexpected paper status, timeouts), try to close our connection to Plustek and then reconnect after a cooling off period.

### DIFF
--- a/libs/plustek-sdk/src/mocks.ts
+++ b/libs/plustek-sdk/src/mocks.ts
@@ -8,6 +8,7 @@ import {
   PropertyAssigner,
   interpret,
 } from 'xstate';
+import { send } from 'xstate/lib/actions';
 import { waitFor } from 'xstate/lib/waitFor';
 import { ScannerError } from './errors';
 import { PaperStatus } from './paper_status';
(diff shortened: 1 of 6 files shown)


libs/plustek-sdk/src/mocks.ts
```diff
@@ -484,6 +492,15 @@ export class MockScannerClient implements ScannerClient {
           debug('scan failed, jam');
           return err(ScannerError.PaperStatusJam);
         }
+        if ((this.machine.state.value as string) === 'ready_to_scan') {
+          debug('scan failed, error feeding');
+          /* istanbul ignore next - randomness makes this hard to cover */
```
benadida, line 497: just mock `Math.random` ! ;)

j/k, it's fine.

benadida (approved):

PR #2305, opened 2022-08-09
feat(all): Explicitly use Helvetica Neue and Noto Emoji

beausmith:
Election Manager was using "Helvetica Neue" as a web-font. Other apps referred to it but were defaulting to Arial or similar font on machines which did not have "Helvetica Neue" installed.

The distribution versions of Linux do not have the fonts we want and on Debian it did not contain an emoji font.

This commit adds Helvetica Neue and Noto Emoji fonts to all apps (skipping `bas`).

Here are the two icons which were previously not working on Debian:
image


libs/ui/src/precinct_scanner_polls_report.tsx
```diff
@@ -63,6 +63,7 @@ const SignatureLine = styled.div`
   width: 50%;
   min-height: 4em;
   &::before {
+    font-family: 'Noto Emoji', sans-serif;
```
beausmith, line 66: This may be overly specific and un-necessary… but I don't know how to emulate a Poll Worker card with a polls closed report on it.

frontends/bmd/public/fonts/noto-emoji.css
```diff
@@ -0,0 +1,41 @@
...
+  src: local(''), url('noto-emoji-v26-emoji-regular.woff2') format('woff2');
+}
+
+/* noto-emoji-600 - emoji */
+@font-face {
+  font-family: 'Noto Emoji';
+  font-weight: 600;
+  font-style: normal;
+  src: local(''), url('noto-emoji-v26-emoji-600.woff2') format('woff2');
+}
```
arsalansufi, line 25: The most minor nit that you're welcome to ignore: We could move this below `/* noto-emoji-500 - emoji */` such that the font faces are ordered in terms of increasing weight (here and in the other `noto-emoji.css` files)

arsalansufi (approved):
Down the road, could be nice to find a way to not have to duplicate these font files across the frontends and share them in `libs/ui` but for m14 LGTM!

adghayes (commented):
Agree it would be nice to load these files from `libs/ui` - the same applies to image assets we use across apps - but punt for m14.

@beausmith have you smoke tested the BMD ballots with the new font?

frontends/bmd/public/fonts/noto-emoji.css
beausmith: 👍  Fixed. I appreciate the nit. This was not on purpose. I used a tool which auto-generated this so that I didn't have to create it by hand. Not sure why the tool ordered them this way.

beausmith:
@adghayes @arsalansufi - Yes, agreed, would be ideal to put all this static content into lib/ui. I tried a couple ways to do this but didn't have success. 

@adghayes - Yes, I did smoke test ballots from a few different Election Definitions. All look good. FWIW, the ballots were always displaying with this font on my machine in the past due to the `font-family` font stack. So now we'll be sure that everyone sees the same font all the time.

Issue #2260, opened 2022-08-03
VxMark: poll worker / admin screens show a11y control outlines


mattroe:
Most obviously shown on the poll worker PIN input screen.

arsalansufi:
Definitely more to revisit here but settled on a quick and simple solution for m14 (more context in https://github.com/votingworks/vxsuite/pull/2308)

PR #2308, opened 2022-08-10
Hide focus outlines on VxMark PIN screen, at least for now

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2260

We have extra visible focus outlines on VxMark for accessible controller navigation. These outlines are applied globally, but there are certain circumstances where we don't need them, most notably on the System Administrator / Election Manager PIN screen. In fact, we've received feedback that the outlines make that screen feel broken. There's an ongoing discussion about where else these outlines are and aren't needed, but for m14, let's just disable the outlines on the VxMark PIN screen.

## Screenshots

Before on the left, after on the right

outlines-before outlines-after

## Testing

- [x] Tested manually
- [x] Added rudimentary unit test

## Checklist

### DIFF
--- a/frontends/bmd/src/pages/unlock_admin_screen.test.tsx
+++ b/frontends/bmd/src/pages/unlock_admin_screen.test.tsx
@@ -51,3 +51,27 @@ test('If passcode is incorrect, error message is shown', () => {
   render(<UnlockAdminScreen auth={auth} />);
   screen.getByText('Invalid code. Please try again.');
 });
+
+test('Hides focus outlines when necessary', () => {
+  const auth = Inserted.fakeCheckingPasscodeAuth();
+  const hideFocusOutlinesCssClass = '.hide-focus-outlines';
+
+  let renderResult = render(
+    <UnlockAdminScreen areFocusOutlinesVisible={false} auth={auth} />
+  );
+  expect(document.querySelector(hideFocusOutlinesCssClass)).toBeInTheDocument();
+  renderResult.unmount();
+
+  renderResult = render(
+    <UnlockAdminScreen areFocusOutlinesVisible auth={auth} />
+  );
+  expect(
+    document.querySelector(hideFocusOutlinesCssClass)
+  ).not.toBeInTheDocument();
+  renderResult.unmount();
+
+  render(<UnlockAdminScreen auth={auth} />);
+  expect(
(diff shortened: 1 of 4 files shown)


frontends/bmd/src/pages/unlock_admin_screen.test.tsx
```diff
@@ -51,3 +51,27 @@ test('If passcode is incorrect, error message is shown', () => {
...
 });
+
+test('Hides focus outlines when necessary', () => {
+  const auth = Inserted.fakeCheckingPasscodeAuth();
+  const hideFocusOutlinesCssClass = '.hide-focus-outlines';
+
+  let renderResult = render(
+    <UnlockAdminScreen areFocusOutlinesVisible={false} auth={auth} />
+  );
+  expect(document.querySelector(hideFocusOutlinesCssClass)).toBeInTheDocument();
```
arsalansufi, line 62: This was a tricky thing to test from a purely user-oriented perspective since we're not changing whether or not something is focused, just the styling.

Given the constraint that we have to test styling, I could've alternatively checked the value of the `outline` style, but this isn't possible without somehow rendering the component with the styles from `App.css` applied, also tricky

arsalansufi:
Update: I misinterpreted one of the goals of this task. We want the focus outlines to show up only after a button on the accessible controller has been pressed, a stricter check than whether or not the accessible controller is simply attached. Modifying accordingly!

arsalansufi:
So I did some digging after hearing that the real issue was that the focus outlines triggered right away on the PIN screen, even before accessible controller input (while on other screens, the focus outlines only appear after accessible controller input).

I have limited hardware for testing right now, so I did my best to emulate in the browser, with the "desktop touch" setting enabled in Chrome and my keyboard as a proxy for an accessible controller. With this setup, I _wasn't_ able to repro focus outlines appearing right away on the PIN screen. The focus outlines only appeared after keyboard press or occasionally after clicks in some of the empty spaces on the page but _not_ clicks on buttons.

I've updated this PR to simply always hide focus outlines on the VxMark PIN screen as a potential option for m14. Curious what you think about this vs. just leaving the page completely as is for m14 @mattroe!

mattroe:
Hiding the focus outlines for now on this screen is totally fine for m14. We can cross the poll worker a11y bridge when we do our precinct system a11y work in the future.

PR #2309, opened 2022-08-10
fix(scan): filter out phantom QR codes 

benadida:
These phantom QR codes tend to happen on the back of BMD ballots when there's noise.

When that happens, the error can be quite disconcerting, for example: invalid election.

To address this, we filter out QR codes that do not include the proper prefixes we expect. That's roughly 1 in 2^16 chance that a phantom QR code would have the prefixes we're looking for. We can eventually do better, by seeing if the QR code content fully parses. Or we can eventually make our fixed prefix a bit longer. But, this simple improvement is likely enough in practice.


jonahkagan (approved):

services/scan/src/interpreter.test.ts
```diff
@@ -12,15 +12,19 @@ import {
...
 import * as choctaw2020Fixtures from '../test/fixtures/2020-choctaw';
 import * as stateOfHamiltonFixtures from '../test/fixtures/state-of-hamilton';
+import * as msDemoFixtures from '../test/fixtures/election-b0260b4e-mississippi-demo';
 import { Interpreter, sheetRequiresAdjudication } from './interpreter';
+import { createInterpreter } from './simple_interpreter';
 import { pdfToImages } from './util/pdf_to_images';
 import { detectQrcodeInFilePath } from './workers/qrcode';
 
 const sampleBallotImagesPath = join(__dirname, '..', 'sample-ballot-images/');
+const interpreterOutputPath = join(__dirname, '..', 'test-output-dir/');
```
eventualbuddha, line 26: Any reason not to just create a temp dir?

benadida: no good reason! Probably a better solution.

PR #2312, opened 2022-08-10
feat(scan): Record rejected/returned ballots in database

### LINKED ISSUE #2300: scan: ensure all ballot scans are stored in db
We currently do not store ballot scan data for reviewed ballots that are removed by a voter. We may not be storing ballot scan data for other cases as well.

We should store all ballot scans for proper auditing, debugging, and general analysis purposes.

jonahkagan:
## Overview
Closes #2300 

If a ballot is interpreted but not cast, we still want to record the interpretation in the db to ensure that we know what happened.

## Demo Video or Screenshot
N/A

## Testing Plan 
Updated tests

### DIFF
--- a/services/scan/src/precinct_scanner_app.test.ts
+++ b/services/scan/src/precinct_scanner_app.test.ts
@@ -132,7 +132,7 @@ async function waitForStatus(
 ) {
   await waitForExpect(async () => {
     await expectStatus(app, status);
-  });
+  }, 1_000);
 }
 
 async function createApp(
@@ -156,7 +156,7 @@ async function createApp(
     delays
   );
   const app = buildPrecinctScannerApp(precinctScannerMachine, workspace);
-  return { app, mockPlustek };
+  return { app, mockPlustek, workspace };
 }
 
 const famousNamesPath = join(
@@ -289,7 +289,7 @@ const needsReviewInterpretation: Scan.SheetInterpretation = {
 };
 
 test('ballot needs review - return', async () => {
-  const { app, mockPlustek } = await createApp();
+  const { app, mockPlustek, workspace } = await createApp();
   await configureApp(app, { addTemplates: true });
 
   await mockPlustek.simulateLoadSheet(ballotImages.unmarkedHmpb);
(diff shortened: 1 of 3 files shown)


services/scan/src/simple_interpreter.ts
```diff
@@ -316,19 +316,50 @@ export function createInterpreter(
...
-/**
- * Stores an interpreted sheet of ballot paper in the {@link Store}, returning
- * the effective sheet ID.
- */
-export function storeInterpretedSheet(
+function storeInterpretedSheet(
   store: Store,
-  sheetId: Id,
-  sheet: SheetOf<PageInterpretationWithFiles>
+  sheetId: string,
```
carolinemodic, line 321: why is this just a string now?

services/scan/src/simple_interpreter.ts
```diff
@@ -316,19 +316,50 @@ export function createInterpreter(
...
+export function storeAcceptedSheet(
+  store: Store,
+  sheetId: string,
+  interpretation: SheetInterpretation
+): void {
+  storeInterpretedSheet(store, sheetId, interpretation);
+  // If we're storing an accepted sheet that needed review that means it was
+  // "adjudicated" (i.e. the voter said to count it without changing anything)
+  if (interpretation.type === 'NeedsReviewSheet') {
+    store.adjudicateSheet(sheetId, 'front', []);
```
carolinemodic, line 345: isn't a sheet already just one side of the ballot, why are there both the front and back calls here?

services/scan/src/precinct_scanner_app.test.ts
```diff
@@ -302,15 +302,29 @@ test('ballot needs review - return', async () => {
...
+  await expectStatus(app, {
+    state: 'returning',
+    interpretation,
+    canUnconfigure: false,
+  });
   await waitForStatus(app, {
     state: 'returned',
     interpretation,
-    ballotsCounted: 0,
+    canUnconfigure: false,
```
carolinemodic, line 313: I'm confused why the items in this struct are changing in the context of this PR?

carolinemodic (approved):
A few questions but overall lgtm

services/scan/src/simple_interpreter.ts
jonahkagan: because I messed up

services/scan/src/simple_interpreter.ts
jonahkagan: Apparently a "sheet" is the whole piece of paper while a "page" is one side

services/scan/src/precinct_scanner_app.test.ts
jonahkagan: `waitForStatus` takes a partial expected status object and fills in the rest with defaults.

In this particular case `ballotsCounted: 0` is already the default, so I removed it for consistency and to make this less confusing overall (though I admit it's confusing if you don't know about the default thing).

 `canUnconfigure` has a default expected value in `waitForStatus` - `false` if `ballotCount == 0`, `true` otherwise. It's a crude approximation of the logic of `canUnconfigure`. I'm planning to move `canUnconfigure` out of this endpoint entirely because it's only used on the admin screen and isn't really part of the scanner status IMO, so I think it's fine to leave this a bit messy for now.

PR #2319, opened 2022-08-11
feat(scan): Basic logging for precinct scanner state machine

### LINKED ISSUE #2314: VxScan: production logging

jonahkagan:
## Overview
Closes #2314 

Adds basic logging to the precinct scanner state machine, logging all machine events and transitions, which basically captures the complete behavior of the machine.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Added some basic checks to the tests
- Some manual testing

### DIFF
--- a/libs/logging/src/test_utils.ts
+++ b/libs/logging/src/test_utils.ts
@@ -3,6 +3,6 @@ import { LogSource } from './log_source';
 
 export function fakeLogger(logSource: LogSource = LogSource.System): Logger {
   const logger = new Logger(logSource);
-  jest.spyOn(logger, 'log').mockResolvedValue();
+  logger.log = jest.fn().mockResolvedValue(undefined);
   return logger;
 }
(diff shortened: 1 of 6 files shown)


libs/logging/src/test_utils.ts
```diff
@@ -3,6 +3,6 @@ import { LogSource } from './log_source';
 
 export function fakeLogger(logSource: LogSource = LogSource.System): Logger {
   const logger = new Logger(logSource);
-  jest.spyOn(logger, 'log').mockResolvedValue();
+  logger.log = jest.fn().mockResolvedValue(undefined);
```
jonahkagan, line 6: With the spy approach, I was running into errors from `console.log` being called after tests completed (I think because the state machine kept running for some reason). It seemed like the spying wasn't actually replacing the real implementation of `logger.log`, so I decided to just change to a straight up mock.

carolinemodic (approved):

PR #2320, opened 2022-08-11
Ballot Bag Capacity Stop

### LINKED ISSUE #2280: VxScan: ballot bag capacity stop
## Description
When a VxScan ballot bag reaches capacity, no new ballots should be scanned until a poll worker confirms the ballot bag has been replaced and sealed.

## Requirements
- No new ballots are scanned after capacity is met and a new ballot bag full error screen should be shown. 
- The ballot bag full screen should still show the ballot bag count for poll workers to document for ballot accounting purposes.
- A poll worker can confirm the ballot bag has been swapped to continue voting and advance to the insert your ballot screen.

## Suggested Implementation (needs discussion)
- The screen in this state should read: *Ballot Bag Full* _Insert poll worker card to continue scanning after replacing ballot bag._
- Poll worker inserts card to access a button "Swap Ballot Bag" with a confirmation modal to confirm bag was swapped and sealed to begin voting.
- Insert your ballot screen is shown after card removal.

adghayes:
## Overview
Closes #2280. Introduces a "Ballot Bag is Full" modal which stops scanning every time the ballot bag hits a pre-configured threshold. The Poll Worker is prompted to insert their PW card and confirm they have switched the bag.

## Demo Video or Screenshot
[replace-ballot-bag-rev2.webm](https://user-images.githubusercontent.com/37960853/184051004-eda28cc9-cc8b-4b22-9504-5f7c153389ff.webm)

## Testing Plan 
Because of the card interaction, easiest to test this at the `App` level. Added an extensive test in `app.test.tsx` that gives the new component full test coverage.

## Checklist


adghayes:
Haven't added any logging. Let me know if we feel that is necessary for this PR, or if that could be handled when we do a pass through `precinct-scanner` later because it lacks logging generally.

benadida:
ballot bag capacity of `2` feels a little low, though, as per video demo. ;)

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  useEffect(() => {
+    if (modalState === 'no_auth' && pollWorkerAuthenticated) {
+      setModalState('confirming');
+    } else if (modalState === 'confirming' && !pollWorkerAuthenticated) {
+      setModalState('no_auth');
+    } else if (modalState === 'confirmed' && !pollWorkerAuthenticated) {
+      onComplete();
+    }
+  }, [modalState, onComplete, pollWorkerAuthenticated]);
```
jonahkagan, line 31: I think this is one of those cases where `useEffect` is not actually needed. I think in general it's good to try to reserve `useEffect` for side effects, not for managing UI state.

A sketch of how to mostly eliminate it:

```ts
const [confirmed, setConfirmed] = useState(false)

if (!confirmed && !pollWorkerAuthenticated) {
   // render ballot bag full message
}

if (!confirmed && pollWorkerAuthenticated) {
  // render confirmation screen
}

if (confirmed) {
  // render remove card screen
}
```
Then, the confirm button handler can be `setConfirmed(true)`.

You still need a `useEffect` to listen for the final card removal:
```ts
useEffect(() => {
  if (confirmed && !pollworkerAuthenticated)
    onComplete()
})
```

What do you think?

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -188,6 +200,12 @@ function appReducer(state: State, action: AppAction): State {
         ...state,
         isPollsOpen: !state.isPollsOpen,
       };
+    case 'ballotBagReplaced':
+      return {
+        ...state,
+        ballotBagLastReplacedAt:
+          state.ballotBagLastReplacedAt + BALLOT_BAG_CAPACITY,
```
jonahkagan, line 207: This seems a bit tricky. Yes, it's true that at the time this action is sent, the  current ballot count  equals the last replaced count plus the capacity, but that takes a bit of reasoning out. I think a slightly clearer way would be to send the current ballot count as part of the action, e.g.:

```ts
{ type: 'ballotBagReplaced', currentBallotCount: number }
 ```

File: frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx [hunk 3]
```diff
@@ -0,0 +1,110 @@
...
+              Has the full ballot bag has been replaced with an empty ballot
+              bag?
+            </p>
+            <p>
+              <Button primary onPress={() => setModalState('confirmed')}>
+                Yes, Resume Voting
+              </Button>
+            </p>
+            <Text small italic>
+              Remove card if you’re not ready to resume voting.
```
jonahkagan, line 60: This instruction confused me when I watched the video. If I were a poll worker and I wasn't ready to resume voting, wouldn't I just not take any action? Why do they need to remove the card?

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
```diff
@@ -0,0 +1,110 @@
...
+  }, [modalState, onComplete, pollWorkerAuthenticated]);
+
+  const mainContent = (() => {
+    switch (modalState) {
+      case 'no_auth':
+        return (
+          <CenteredLargeProse>
+            <h1>The Ballot Bag is Full</h1>
+            <p>
+              A Poll Worker must replace the full ballot bag with an empty
```
jonahkagan, line 40: In general, I think we don't capitalize "poll worker"

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
```diff
@@ -0,0 +1,110 @@
...
+              ballot bag.
+            </p>
+            <p>Insert a Poll Worker Card after the ballot bag is replaced.</p>
```
jonahkagan, line 43: A suggestion for making these instructions more straightforward:

> Ask a poll worker to replace the ballot bag. Insert a poll worker card to continue voting.

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
```diff
@@ -0,0 +1,110 @@
...
+      onComplete();
+    }
+  }, [modalState, onComplete, pollWorkerAuthenticated]);
+
+  const mainContent = (() => {
+    switch (modalState) {
+      case 'no_auth':
+        return (
+          <CenteredLargeProse>
+            <h1>The Ballot Bag is Full</h1>
```
jonahkagan, line 38: I think our other headings in VxScan don't use full sentences and articles (e.g.  "Ballot Not Counted" or "Too Many Votes"). To stick with that pattern, maybe just "Ballot Bag Full" here?

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -451,6 +478,16 @@ export function AppRoot({
     return <InsertUsbScreen />;
   }
 
+  if (needsToReplaceBallotBag && scannerStatus.state !== 'accepted') {
```
jonahkagan, line 481: waiting for the accepted screen to clear is a really nice touch!

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
adghayes: Good call, thanks. I think the Effect was necessary when we were able to step back from the confirmation screen with the "cancel" button, but def not with the current flow.

adghayes: The only clause that might have to remain in the useEffect is - we have to trigger the callback when the poll worker card is removed & confirmed. That seems like a side effect to me?

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
```diff
@@ -0,0 +1,110 @@
...
+          </CenteredLargeProse>
+        );
+      /* istanbul ignore next - compile time check for completeness */
+      default:
+        throwIllegalValue(modalState);
+    }
+  })();
+
+  return (
+    <Modal
```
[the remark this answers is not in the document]
beausmith: Agreed, solid reasoning. If we do this and the margins cause content to reflow, I'd like to take another pass at copy/etc.

frontends/precinct-scanner/src/app_root.tsx
adghayes: Sounds good.

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -88,6 +90,7 @@ interface ScannerConfigState {
 
 interface FrontendState {
   isPollsOpen: boolean;
+  ballotBagLastReplacedAt: number;
```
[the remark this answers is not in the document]
adghayes: OK!

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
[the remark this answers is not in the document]
adghayes: Makes sense to me.

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
jonahkagan: Yeah, generally the approach I try to use is to trigger state updates based on user actions (e.g. button press). That's easy when the user action is on screen. With hardware actions, you can't do it as easily due to the way we currently track hardware actions via a changing value. So using a `useEffect` there is a good workaround.

That being said, I'm noodling on how we might design a hardware interface that could let us register action handlers similar to on-screen actions. Let me know if you have ideas!

adghayes: Clearly a longer term convo but...

What if we had Listener Components, and hardware actions at the top level would trigger events on all listener components currently on the screen?

I don't know enough to know why that would be a better paradigm, the current paradigm seems acceptable to me.

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
beausmith: I want the poll worker to know that they can remove the card in the case that they need to before replacing the bag. I want the poll worker to have confidence and know what their options are. 

It would be odd if a poll worker inserted their card, then left their card in the machine and walked away to get a new ballot bag.

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
beausmith: I like the simplified copy, but it doesn't communicate that the poll worker should replace the bag before inserting their card to resume voting.

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
jonahkagan: The risk with the current approach is that a `useEffect` might run at any time, not just on the change hardware status. Using the dependency array to get around this only sometimes works (e.g. it doesn't work when the dependency is an object that isn't memoized), so it's tricky to reason about. And React intends the dependency array to only be used for optimization, not correctness.

In contrast, with a button handler, you can trivially rely on it only getting called when the button gets pressed. Very easy to reason about!

I like your idea! I was thinking along similar lines!

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
beausmith: Agreed, let's make it lower case.

frontends/precinct-scanner/src/components/replace_ballot_bag_modal.tsx
beausmith: Yeah, this would be more consistent. I had changed to this sentence to make it a bit more friendly/approachable in the case that a voter was the first to see it. Lets change it for consistency!

adghayes:
Latest flow, although would prefer not to have any more UX iterations at this point:

[ballot-bag-full-rev3.webm](https://user-images.githubusercontent.com/37960853/184265735-372ddee4-6ab3-4bff-b228-8fee6261033c.webm)

beausmith (approved):
LGTM from a design perspective.

PR #2321, opened 2022-08-11
feat(ps): Remove obsolete setup power screen.

beausmith:
From [@jonahkagan's comment](https://github.com/votingworks/vxsuite/pull/2276#discussion_r942616520):
image


arsalansufi (approved):

frontends/precinct-scanner/src/screens/insert_ballot_screen.tsx
```diff
@@ -20,16 +20,16 @@ export function InsertBallotScreen({
...
       <InsertBallot />
       <CenteredLargeProse>
         <h1>Insert Your Ballot Below</h1>
         <p>Scan one ballot sheet at a time.</p>
+        {showNoChargerWarning && (
+          <Text warning small center>
+            <strong>No Power Detected.</strong> Please ask a poll worker to plug
+            in the power cord.
+          </Text>
+        )}
```
jonahkagan, line 32: Trying to think through when this message would show up based on the wiring of VxScan. My understanding is the power wiring is like this:

```
wall outlet --------+------- computer
                    |
                    +------- plustek
```

So if the power cord is unplugged from the wall, we'll see that the battery is not charging, but the Plustek will also be turned off, so we'll just show the `SetupScannerScreen`.

That would mean that this case where we're on the `InsertBallotScreen` but the battery isn't charging is not possible. Is that true?

beausmith: Sorry, I didn't follow the logic. I was trusting that you knew what you were asking for. 🤷‍♂️

jonahkagan: Yeah, i don't think there's anything wrong with this PR. I am just suggesting that we can also remove this case.

beausmith: I'm gonna punt on this for now being that we're at code freeze, it's my day off, there is not harm of leaving as is.

PR #2326, opened 2022-08-11
feat(scan): Kill unresponsive plustekctl

jonahkagan:
## Overview
In some cases, when `plustekctl` becomes unresponsive, it will not respond to the `quit` command that we try to send to disconnect. That will block a new `plustekctl` process from connecting to the scanner (seemingly, the old process is still holding some lock).

In this branch, we add a new command `kill` to the plustek-sdk library to send a `SIGTERM` to `plustekctl`. When we try disconnecting after an unexpected error or timeout, we wait a bit and then give up and send the kill signal.

This branch also makes a few changes to the frontend:
- When we're recovering from an error, show the scanning progress screen. More often than not, these types of errors occur during scanning, and continuing to show that screen makes for a seamless UX. We could probably get fancier and find a way to keep the state machine on whatever state its on when the error occurs, but that would be a bigger change.
- Make sure actual unrecoverable errors show a message to restart the scanner, with simplified copy from @mattroe.

## Demo Video or Screenshot
Try it out

## Testing Plan 
- Manual testing
- Automated test updates
- Bakery testing


benadida (approved):
Ahah you're just sending a sigterm, i thought it was going to be a harsher kill. Lgtm.

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -37,19 +39,19 @@ export function ScanErrorScreen({
...
+        return 'Remove ballot to continue.';
+      case 'paper_status_timed_out':
       case 'scanning_timed_out':
       case 'unexpected_paper_status':
       case 'unexpected_event':
       case 'plustek_error':
-        return 'The scanner experienced an error and needs to be reset. Please turn off and unplug the scanner from the power outlet. Plug it back in and turn it on again.';
+        // These cases require restart, so we don't need to show an error
+        // message, since that's handled below.
+        return undefined;
```
jonahkagan, line 54: @beausmith heads up I made some copy simplifications here. A few reasons:
- The scan service now distinguishes between recoverable and non-recoverable errors better
- The error cases above that had specific reasons the ballot was rejected weren't quite lining up with the reality of what was happening with the scanner, so being a bit more ambiguous makes for a smoother flow

Hopefully in the future we can be more precise, but I think this is the best way forward for now

PR #2327, opened 2022-08-11
-feat(bmd): remove test deck printing from BMD

### LINKED ISSUE #1874: VxMark: remove test decks
Since they now live in VxAdmin! https://github.com/votingworks/vxsuite/issues/1710

benadida:
fixes #1874


arsalansufi (approved):
Looks like a few more tests are in need of updating but otherwise LGTM!

PR #2340, opened 2022-08-12
feat(bsd): clarify adjudication screen

### LINKED ISSUE #443: VxCentralScan: adjudication screen should be clarified
Right now, our screen tries to be aligned with the Ms process, but because that process isn't consistent, it's not helpful. Instead, we should have the text describe the exact action that the *scanner* is taking, not the ones that the adjudication board will take. Specifically, using the term "duplicate ballot" is confusing.

So, the text should read:

----

## Overvote

The last ballot was not tabulated.

[Send to resolution board]

or

[Tabulate as is]

----

Instead of an upper-right-hand corner confirmation, clicking either button should pop up a modal.

In the case of send to resolution board, the modal should read:

----

You have chosen to send this ballot to the resolution board.

Please remove this ballot from the stack.

[Cancel] [I've removed the ballot]

----

In the case of "Tabulate as Is":

----

You have chosen to tabulate this ballot as is.

[Cancel] [Tabulate as is]

----

beausmith:
## Overview
Closes #443

## Demo Video or Screenshot
image
image
image

### DIFF
--- a/frontends/bsd/src/components/ballot_sheet_image.tsx
+++ b/frontends/bsd/src/components/ballot_sheet_image.tsx
@@ -77,7 +77,7 @@ export function BallotSheetImage({
         src={imageUrl}
         alt="front"
         onLoad={recalculateScale}
-        style={{ maxWidth: '100%', maxHeight: '87vh' }}
+        style={{ maxWidth: '100%', maxHeight: '82vh' }}
       />
       {layout &&
         contestIds &&
--- a/frontends/bsd/src/screens/__snapshots__/ballot_eject_screen.test.tsx.snap
+++ b/frontends/bsd/src/screens/__snapshots__/ballot_eject_screen.test.tsx.snap
@@ -30,43 +30,40 @@ exports[`calls out live ballot sheets in test mode 1`] = `
         </div>
         <div
           class="sc-pFZIQ fNwxur"
-        >
-          <button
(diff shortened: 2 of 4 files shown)


benadida:
One of the changes I'd made was to remove the explicit mention of a resolution board, because it varies from state to state in name and process. Plus, are we actually sending to resolution board in our product? I suggest sticking to "remove ballot from count" or something like that.

Otherwise looks great.

benadida (approved):
I have a slight preference for not using terminology like "resolution board" which differs from state to state and which assumes more than needed. But, I can live with it.

mattroe:
Agree with @benadida on the desire to generalize terminology when we can. 

`Remove to Adjudicate`?
`Remove to Review`?

beausmith:
Agree on generic language. "Resolution Board" was in the issue copy the @benadida wrote, so I assumed this was what was wanted. Happy to update this to @mattroe's suggestion "Remove to Adjudicate"

benadida:
> Agree on generic language. "Resolution Board" was in the issue copy the @benadida wrote, so I assumed this was what was wanted. Happy to update this to @mattroe's suggestion "Remove to Adjudicate"

HAH HAH, I am hoisted on my own petard! ;)

OK, cool, glad we agree, sorry for the misdirection.

PR #2354, opened 2022-08-12
fix(ps): Don't show scanning screen on accept after review

jonahkagan:
## Overview
Accepting a ballot after review only takes about half a second. Previously, we showed the "Please wait, checking your ballot" screen, since it used the same code path as a ballot that didn't need review.

To fix this, we add a new state `accepting_after_review` to the state machine (which uses the same implementation as `accepting` - finally figured out how to reuse state machine logic).

The frontend is updated to disable the buttons in the modal for that half second, which provides enough visual feedback that the button press actually did something without being too disruptive.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/530106/184453020-343fc88f-8f17-4e0f-9b79-036eefb6d82c.mov

## Testing Plan 
- Updated existing tests
- Manually tested with ballots of each review type

## Checklist

### DIFF
--- a/frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
@@ -30,6 +30,42 @@ import { AppContext } from '../contexts/app_context';
 import { toSentence } from '../utils/to_sentence';
 import { useSound } from '../hooks/use_sound';
 
+interface ConfirmModalProps {
+  content?: React.ReactNode;
+  onConfirm: () => void;
+  onCancel: () => void;
+}
+
+function ConfirmModal({ content, onConfirm, onCancel }: ConfirmModalProps) {
+  const [confirmed, setConfirmed] = useState(false);
+
+  return (
+    <Modal
+      theme={fontSizeTheme.large}
+      modalWidth={ModalWidth.Wide}
+      content={content}
+      actions={
+        <React.Fragment>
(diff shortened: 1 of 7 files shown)


services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -117,11 +119,11 @@ function connectToPlustek(createPlustekClient: CreatePlustekClient) {
   };
 }
 
-async function disconnectFromPlustek({ client }: Context) {
+async function closePlustekClient({ client }: Context) {
```
jonahkagan, line 122: unrelated name change

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -197,20 +222,13 @@ function UndervoteWarningScreen({
                   </span>
                 )}
               </p>
-              <Text italic>
+              <Text italic small>
```
carolinemodic, line 225: why are we also making the text smaller?

benadida (approved):

carolinemodic (approved):

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
jonahkagan: It better matches the hierarchy of text on the screen before opening the modal

PR #2378, opened 2022-08-15
Allow sys admins to unconfigure VxCentralScan without backing up

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2365

This PR allows system admins to unconfigure VxCentralScan without backing the machine up first. Some additional context from Roe in Slack:

> the original goal of this feature in SA mode was for us in demo'ing to easily reset a machine / unbork a card state

> now this does call into question whether we want to force a sysadmin to take the same pre-unconfigure steps as an election manager (exporting backups). I think we can trust the sysadmin to be responsible as this feature is not going to be the common unconfigure flow.

## Testing

- [x] Updated unit tests
- [x] Tested manually

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ N/A
- [ ] ~I have added JSDoc comments to any newly introduced exports~ N/A
- [ ] ~I have added a screenshot and/or video to this PR to demo the change~ N/A
- [x] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

Issue #2365, opened 2022-08-15
vxcentralscan + vxscan: unconfigure via systemadmin card doesn't work


benadida:
maybe because backup hasn't been saved?

PR #2379, opened 2022-08-15
Allow sys admins to unconfigure VxScan without backing up

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2365

This PR allows system admins to unconfigure VxScan without backing the machine up first. This is the same change as https://github.com/votingworks/vxsuite/pull/2378 but for VxScan instead of VxCentralScan.

## Testing

- [x] Updated unit tests
- [ ] Tested manually (in-progress)

## Checklist

- [ ] ~I have added a screenshot and/or video to this PR to demo the change~

### DIFF
--- a/frontends/precinct-scanner/src/api/config.test.ts
+++ b/frontends/precinct-scanner/src/api/config.test.ts
@@ -32,6 +32,14 @@ test('DELETE /config/election to delete election', async () => {
   await config.setElection(undefined);
 });
 
+test('DELETE /config/election ?ignoreBackupRequirement query param', async () => {
+  fetchMock.deleteOnce(
+    '/config/election?ignoreBackupRequirement=true',
+    JSON.stringify({ status: 'ok' })
+  );
+  await config.setElection(undefined, { ignoreBackupRequirement: true });
+});
+
 test('DELETE /config/election to delete election with bad API response', async () => {
   fetchMock.deleteOnce(
     '/config/election',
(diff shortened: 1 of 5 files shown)


arsalansufi:
Still working on manual testing but wanted to get this up for review since I don't expect to have to make any changes!

carolinemodic (approved):

frontends/precinct-scanner/src/api/config.test.ts
```diff
@@ -32,6 +32,14 @@ test('DELETE /config/election to delete election', async () => {
   await config.setElection(undefined);
 });
 
+test('DELETE /config/election ?ignoreBackupRequirement query param', async () => {
+  fetchMock.deleteOnce(
+    '/config/election?ignoreBackupRequirement=true',
+    JSON.stringify({ status: 'ok' })
+  );
+  await config.setElection(undefined, { ignoreBackupRequirement: true });
```
benadida, line 40: what are we testing here? Is there any way for this test to fail? Maybe just missing an assert?

benadida (approved):

frontends/precinct-scanner/src/api/config.test.ts
arsalansufi: Good q! I was following the pattern of the test above this one:

https://github.com/votingworks/vxsuite/blob/a3f1c3378d83adcd70a2113558f48490f4fff18b/frontends/precinct-scanner/src/api/config.test.ts#L30-L33

The test I've added _will_ fail if an unmatched call is made, i.e. a call to a URL other than `/config/election?ignoreBackupRequirement=true`. That said, looks like the test _won't_ fail if no call at all is made - we could definitely add an assert to make sure a call is made.

We should in fact make that change in many places in the codebase. I can open a tech improvement ticket to fill in this gap across vxsuite.

PR #2381, opened 2022-08-15
test(cypress): fix incorrect election hashes

eventualbuddha:
## Overview

The `cypress-file-upload` package, by going through `cy.fixture`, didn't handle JSON files the way we expected. It appears to [parse JSON files for us in a way we don't want](https://github.com/cypress-io/cypress/blob/e0bd6ac2aaf8d00b9233fffefed8f6ed2484bf45/packages/server/lib/fixture.js#L31-L35
). Cypress 9.3.0 added `selectFile`, which lets us pass the data we want as a `Buffer`. I migrated from `attachFile` to `selectFile` everywhere per [the guide](https://docs.cypress.io/guides/references/migration-guide#Migrating-from-cypress-file-upload-to-selectFile).

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated.

## Checklist


integration-testing/bsd/cypress/e2e/configuration.cy.ts
```diff
@@ -46,16 +46,19 @@ describe('BSD and services/Scan', () => {
 
   it('BSD can be configured with services/scan with an election JSON file', () => {
     cy.contains('Load Election Configuration');
-    cy.get('input[type="file"]').attachFile('election.json');
+    cy.get('input[type="file"]').selectFile('cypress/fixtures/election.json', {
+      force: true,
```
eventualbuddha, line 50: These `force: true` options were required because we use the "hidden input with a button or label" technique, so cypress sees the input as covered or inaccessible and will error by default without this.

benadida:
wow! Great find.

arsalansufi (approved):
Awesome find! LGTM!

PR #2387, opened 2022-08-16
refactor(scan): Add route prefix to distinguish central scanner and precinct scanner APIs

jonahkagan:
## Overview
Now that the scan service provides two separate APIs for central scanner and precinct scanner, we may be running into caching problems because they reuse some of the same endpoints. This PR simply adds a prefix to each API to ensure they are disjoint.

I also changed a couple precinct scanner API endpoints to be a bit more consistent.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Updated tests
- Did a basic manual test of each app

## Checklist


frontends/precinct-scanner/src/api/hmpb.ts
```diff
@@ -82,7 +82,10 @@ export function addTemplates(pkg: BallotPackage): AddTemplatesEvents {
           );
         }
 
-        await fetch('/scan/hmpb/addTemplates', { method: 'POST', body });
+        await fetch('/precinct-scanner/config/addTemplates', {
```
arsalansufi, line 85: Just double checking that the renaming of this path was intentional. I assume this is one of the cases you're referring to when you say the following in the PR description:

> I also changed a couple precinct scanner API endpoints to be a bit more consistent.

If this is intentional and this better fits under `/precinct-scanner/config`, should we move it to the `src/api/config.ts` file?

frontends/precinct-scanner/src/api/hmpb.ts
```diff
@@ -96,5 +99,5 @@ export function addTemplates(pkg: BallotPackage): AddTemplatesEvents {
 }
 
 export async function doneTemplates(): Promise<void> {
-  await fetch('/scan/hmpb/doneTemplates', { method: 'POST' });
+  await fetch('/precinct-scanner/config/doneTemplates', { method: 'POST' });
```
arsalansufi, line 102: Same comment here as above

arsalansufi (commented):
One question!

frontends/precinct-scanner/src/api/hmpb.ts
jonahkagan: Yeah it was intentional. I was trying to restrain myself from further renaming/refactoring, but I think you're right that I left things kind of halfway changed here. I will update `src/api` to match.

The other thing I would like to do is make the casing consistent - currently we use a mix of hyphens and camelCase. But I think that could be a separate PR

jonahkagan: 5cc9a9c6e

arsalansufi (approved):
LGTM!

PR #2403, opened 2022-08-17
fix(ps): Detect Plustek disconnection with scan service, not kiosk browser

### LINKED ISSUE #2402: VxScan: internal connection problem screen requiring restart
slack issue discussion: https://votingworks.slack.com/archives/C03KGSP78MQ/p1660695566667109

jonahkagan:
## Overview
Fixes #2402 

Currently, we have two ways of telling if the Plustek scanner is disconnected: kiosk browser's USB device list and the scan service scanner status. Kiosk browser can't differentiate between a powered off/unplugged scanner and a crashed scanner, while the scan service can. Here we change the frontend to only check the scan service status and ignore kiosk browser in order to have more precise error messages when the Plustek is not reachable.

When the Plustek has crashed, the scan service will return an "unrecoverable_error" status, at which point we show an error message with instructions to restart (this is already in place).

## Demo Video or Screenshot
N/A

## Testing Plan 
- Since I can't repro a Plustek crash, I simulated one and checked that we show the right screen. I also manually tested that powering off or unplugging the Plustek USB cord works.
- Added a new frontend unit test

## Checklist

### DIFF
--- a/frontends/precinct-scanner/src/app_root.tsx
+++ b/frontends/precinct-scanner/src/app_root.tsx
@@ -255,7 +255,6 @@ export function AppRoot({
     cardReader,
     computer,
     printer: printerInfo,
-    precinctScanner,
   } = useDevices({
     hardware,
     logger,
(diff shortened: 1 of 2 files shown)


carolinemodic (approved):
As long as we have tested that powered off / unconnected still works as expected I'm good with this, especially with doing more scan testing on rc3

PR #2436, opened 2022-08-19
Address flaky tests by adding/extending waits

arsalansufi:
@jonahkagan recently noticed some flaky tests that I believe I've also noticed. While I wasn't able to repro the failures locally, this PR adds/extends waits to make the failures less likely in CI.


frontends/election-manager/src/components/export_election_ballot_package_modal_button.test.tsx
```diff
@@ -138,7 +138,9 @@ test('Modal renders export confirmation screen when usb detected and manual link
   within(modal).getByText(/Optionally, you may pick a custom save location./);
 
   fireEvent.click(within(modal).getByText('Custom'));
-  await waitFor(() => within(modal).getByText('Ballot Package Saved'));
+  await within(modal).findByText('Ballot Package Saved', undefined, {
+    timeout: 2000,
```
arsalansufi, line 142: The default `timeout` is 1000ms

jonahkagan (approved):
Thanks for tackling this!

Issue #2048, opened 2022-06-30
Remove useSmartcard hook


jonahkagan:
The useSmartcard hook is now replaced by useDippedSmartcardAuth and useInsertedSmartcardAuth. It's still used in one fronted, however: bas. Once we add the card programming interface to the new auth hooks, we can replace useSmartcard in bas and then remove the old hook.

arsalansufi:
Copying from https://github.com/votingworks/vxsuite/pull/2444:

> While we initially planned to update VxEncode to use the new smartcard hooks so that we could remove the old `useSmartcard` hook, VxEncode has some unique needs not currently covered by the new smartcard hooks, namely programming of voter cards by poll workers. Instead of updating `useDippedSmartcardAuth` to support this need (currently, only system admins and election managers can use dipped smartcard auth, and only system admins can program cards), I've just moved the old `useSmartcard` hook from `libs/ui` to `frontends/bas` to guarantee it isn't mistakenly used in other frontends.

Updating the title of this ticket accordingly

PR #2444, opened 2022-08-22
Move deprecated useSmartcard hook from libs/ui to frontends/bas

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2048

We've replaced the `useSmartcard` hook in `libs/ui` with `useDippedSmartcardAuth` and `useInsertedSmartcardAuth`. The only frontend that still uses `useSmartcard` is `frontends/bas` (i.e. VxEncode), a frontend that we aren't deploying at the moment and that we may remove from the codebase completely.

While we initially planned to update VxEncode to use the new smartcard hooks so that we could remove the old `useSmartcard` hook, VxEncode has some unique needs not currently covered by the new smartcard hooks, namely programming of voter cards by poll workers. Instead of updating `useDippedSmartcardAuth` to support this need (currently, only system admins and election managers can use dipped smartcard auth, and only system admins can program cards), I've just moved the old `useSmartcard` hook from `libs/ui` to `frontends/bas` to guarantee it isn't mistakenly used in other frontends.

If we do ever choose to bring VxEncode back, we can update `useDippedSmartcardAuth` then.

## Testing

- [x] Automated tests still pass

## Checklist


jonahkagan (approved):
Nice, seems like a good short term compromise

PR #2448, opened 2022-08-22
fix(scan): Log error fields correctly

jonahkagan:
## Overview
By default, `JSON.stringify` will log an instance of `Error` as `{}` since the `message` and `stack` fields are not enumerable. Here, we explicitly pick out those two fields for logging to ensure we get the info we need about the error.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Manually tested
- Updated automated tests

### DIFF
--- a/services/scan/src/precinct_scanner_state_machine.ts
+++ b/services/scan/src/precinct_scanner_state_machine.ts
@@ -878,7 +878,7 @@ function setupLogging(
         // Make sure we log the important fields of an error
         .map(([key, value]) =>
           key === 'error' && value instanceof Error
-            ? [key, { message: value.message, stack: value.stack }]
+            ? [key, { ...value, message: value.message, stack: value.stack }]
             : [key, value]
         )
         .map(([key, value]) => [
--- a/services/scan/src/precinct_scanner_state_machine.ts
+++ b/services/scan/src/precinct_scanner_state_machine.ts
@@ -875,6 +875,12 @@ function setupLogging(
         .map(([key, value]) =>
           key === 'interpretation' ? [key, value?.type] : [key, value]
         )
+        // Make sure we log the important fields of an error
+        .map(([key, value]) =>
+          key === 'error' && value instanceof Error
+            ? [key, { message: value.message, stack: value.stack }]
+            : [key, value]
+        )
         .map(([key, value]) => [
           key,
           value === undefined ? 'undefined' : value,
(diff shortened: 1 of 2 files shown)


services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -865,6 +865,12 @@ function setupLogging(
         .map(([key, value]) =>
           key === 'interpretation' ? [key, value?.type] : [key, value]
         )
+        // Make sure we log the important fields of an error
+        .map(([key, value]) =>
+          key === 'error' && value instanceof Error
+            ? [key, { message: value.message, stack: value.stack }]
```
eventualbuddha, line 871: Won't this then ignore any other properties on `value`? Maybe do `{ message: value.message, stack: value.stack, ...value }`.

```
> e = new Error('hi')
Error: hi
    at REPL31:1:5
    at Script.runInThisContext (node:vm:129:12)
    at REPLServer.defaultEval (node:repl:572:29)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:902:10)
    at REPLServer.emit (node:events:525:35)
    at REPLServer.emit (node:domain:489:12)
    at REPLServer.Interface._onLine (node:readline:491:10)
    at REPLServer.Interface._line (node:readline:868:8)
> e.foo = 1
1
> ({ message: e.message, stack: e.stack, ...e })
{
  message: 'hi',
  stack: 'Error: hi\n' +
    '    at REPL31:1:5\n' +
    '    at Script.runInThisContext (node:vm:129:12)\n' +
    '    at REPLServer.defaultEval (node:repl:572:29)\n' +
    '    at bound (node:domain:433:15)\n' +
    '    at REPLServer.runBound [as eval] (node:domain:444:12)\n' +
    '    at REPLServer.onLine (node:repl:902:10)\n' +
    '    at REPLServer.emit (node:events:525:35)\n' +
    '    at REPLServer.emit (node:domain:489:12)\n' +
    '    at REPLServer.Interface._onLine (node:readline:491:10)\n' +
    '    at REPLServer.Interface._line (node:readline:868:8)',
  foo: 1
}
```

jonahkagan: Yeah, I like that idea. I'm also thinking about moving this logic into the logger itself so that we don't get bitten by this again in the future

jonahkagan: Realized that it's a bigger change to move this into the logger, since right now the logger doesn't accept nested data, so abandoning that

carolinemodic (approved):

PR #2451, opened 2022-08-22
m14-release-branch: fix(scan): Reconnect to plustek after any unexpected scanning error

jonahkagan:
## Overview
Previously, if we got an unexpected scanning error and the Plustek was still responsive, we would either automatically retry scanning or reject and let the voter retry, depending on the paper status. Now, if we get an unexpected scanning error, we handle it like any other unexpected error, reconnecting to Plustek. This will hopefully make us more resilient to scanning errors where we don't know the root cause.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Various manual scan error tests
- Updated automated tests

### DIFF
--- a/services/scan/src/precinct_scanner_app.test.ts
+++ b/services/scan/src/precinct_scanner_app.test.ts
@@ -859,7 +859,9 @@ test('insert second ballot while first ballot is returning', async () => {
 });
 
 test('jam on scan', async () => {
-  const { app, mockPlustek } = await createApp();
+  const { app, mockPlustek } = await createApp({
+    DELAY_RECONNECT_ON_UNEXPECTED_ERROR: 500,
+  });
   await configureApp(app);
 
   await mockPlustek.simulateLoadSheet(ballotImages.completeBmd);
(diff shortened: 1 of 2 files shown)


services/scan/src/precinct_scanner_app.test.ts
```diff
@@ -859,17 +859,20 @@ test('insert second ballot while first ballot is returning', async () => {
...
 
   mockPlustek.simulateJamOnNextOperation();
   await post(app, '/scanner/scan');
-  await waitForStatus(app, { state: 'jammed' });
-
-  await mockPlustek.simulateRemoveSheet();
+  await waitForStatus(app, {
+    state: 'recovering_from_error',
+    error: 'plustek_error',
+  });
```
jonahkagan, line 875: If we get a jam error returned from the scan command, we treat it as an unexpected error now. I tried to reproduce this manually and couldn't do it - I think scan only returns a jam error after a long time trying to ingest the paper, but we now handle that with a timeout. So this case really would be "unexpected" now.

eventualbuddha (approved):

Issue #2459, opened 2022-08-22
Remove need for ballot package 


carolinemodic:
Some context: https://votingworks.slack.com/archives/CEL6D3GAD/p1661200961493069

eventualbuddha:
Likely solution will:
1. include layout(s) in final election definition instead of separate files in the ballot package zip
2. include a base64-encoded image of an empty oval at 200dpi (or whatever we're using for scanning) in the election definition
3. for VX ballots, use layout and oval image from (1) and (2) to do interpretation
4. for accuvote ballots, use `gridLayouts` from election definition that already exists
5. stop exporting the zip, instead just the election definition JSON

Bonus points: standardize `gridLayouts` and new layouts from (1) into a list with proper metadata.

adghayes:
@eventualbuddha, why not just move our process for generating ballot images and layouts from VxAdmin to both VxScan and VxCentralScan? 

Working on #2760 has made me think it wouldn't be too much lift, taking not much more than two weeks, if that:

1. Move some things to shared libraries - `HandMarkedPaperBallot`, some `BallotStyle` utilities, and the hook from `bmd` to load an election definition from an Election Manager card (2 - 3 days)
2. Create a generator function in `libs/ui` to generate the ballot configs, ballots, and layouts. Will rely on the incoming `printElementToPdf` utility also in `libs/ui` and not involve rendering within the app. (1 - 2 days).
3. Use the new generator function as part of the configuration sequence in the scanners, rather than reading from a ballot package (3 - 5 days). 
4. Clean up old ballot package code

It seems like you may be thinking of a bigger change that involves what the scan service actually needs to interpret ballots?

adghayes:
I understand the concern, I'm just not sure if how much to weight that concern. What sort of slight differences in machine configuration could affect the way the ballot is laid out? In the slack discussion, you mentioned you had tried generating the ballots in the scan service and they were slightly different. But if they are all generated in a React frontend, with shared `libs/ui` code, outside of the app root, what configuration differences could affect the layout? Screen size?

I think we should have reasonable guarantees that we are generating ballots in a single environment or set of conditions, but not necessarily that it needs to be a single application.

The layouts in the election definition makes sense (the oval thing strikes me as a bit weird) but sounds like more work.

eventualbuddha:
Possible things that could effect it that I can think of right now:
- installed/configured fonts
- screen resolution and pixel density
- Chromium/Electron version
[...]

Issue #2458, opened 2022-08-22
WIA: ability to adjudicate name-only marks (no oval)


mattroe:
Will vary state by state and would likely need to be configurable.

PR #2461, opened 2022-08-22
chore(scan): remove write-in area scoring

eventualbuddha:
## Overview

We don't have a customer that needs this right now and it makes #2459 more complicated because it requires us to have the ballot template images when interpreting. If we want to bring this back I suggest we encode the area to look for a write-in mark within the layout of the ballots and assume the entire area is white and/or make the threshold configurable to account for the inclusion of some black lines or text. Issue to re-evaluate this later: #2458.

Support for this as an adjudication reason was removed in #2076.
[Slack discussion for further context](https://votingworks.slack.com/archives/CEL6D3GAD/p1661200605817059)

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated.

## Checklist


jonahkagan (approved):
Yay for removing unused features!

PR #2464, opened 2022-08-22
fix(ps): persist new frontend state

adghayes:
## Overview
Fixes bug identified in testing (https://votingworks.slack.com/archives/C03KGSP78MQ/p1661208205049509) where newly added frontend state (the last time the ballot bag was replaced, and whether the sounds are muted) were not persisted across restarts. 

## Testing Plan 
- Manual Testing


adghayes:
Once this is merged, I will make a follow-up ticket to make sure all frontend state is saved using some Typescript magic, rather than this ad hoc approach.

carolinemodic (approved):

jonahkagan (approved):
Looks good. Can you add/update tests as a follow up?

PR #2465, opened 2022-08-23
chore(em): Use shared jest config in election manager 

jonahkagan:
## Overview
It's the only jest config that doesn't use the shared config, which means it doesn't get the default test timeout setting.


frontends/election-manager/package.json
```diff
@@ -209,6 +209,7 @@
     "stylelint-config-prettier": "^8.0.1",
     "stylelint-config-styled-components": "^0.1.1",
     "stylelint-processor-styled-components": "^1.10.0",
+    "ts-jest": "^28.0.8",
```
jonahkagan, line 212: This package provides the `preset` specified in the shared jest config

eventualbuddha (approved):

PR #2466, opened 2022-08-23
refactor(ps): better type frontend states

adghayes:


frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -354,14 +343,16 @@ export function AppRoot({
 
   useEffect(() => {
     async function storeAppState() {
-      // only store app state if we've first initialized from the stored state
+      // Only store app state if we've first initialized from the stored state
       if (initializedFromStorage) {
-        // TODO: If possible, store all FrontendState with types rather than specify each field
-        await storage.set(stateStorageKey, {
+        // We use an explicit type here to reminder ourselves to store any
```
eventualbuddha, line 348:
```suggestion
        // We use an explicit type here to remind ourselves to store any
```

eventualbuddha (approved):
Looks like the right approach to me.

adghayes:
@jonahkagan take a look to give some thoughts about the testing approach please!

frontends/precinct-scanner/src/app.test.tsx
```diff
@@ -1459,3 +1156,66 @@ test('replace ballot bag flow', async () => {
...
+  });
+  card.removeCard();
+
+  // Confirm ballot bag status is saved to storage
+  fetchMock.get(
+    '/precinct-scanner/scanner/status',
+    {
+      body: scannerStatus({
+        state: 'no_paper',
+        ballotsCounted: BALLOT_BAG_CAPACITY * 2,
```
jonahkagan, line 1197: Are you trying to hit the next threshold by doubling the number of ballots here? Or are you just trying to surpass the current threshold? If just trying to surpass, I think `BALLOT_BAG_CAPACITY + 1` would make that clearer

frontends/precinct-scanner/src/app.test.tsx
```diff
@@ -99,17 +93,35 @@ const getMarkThresholdOverridesConfigNoMarkThresholdOverridesResponseBody: Scan.
...
+
+beforeEach(() => {
+  jest.useFakeTimers();
+
+  card = new MemoryCard();
+  storage = new MemoryStorage();
+  hardware = MemoryHardware.build({
+    connectCardReader: true,
+    connectPrecinctScanner: true,
+  });
```
jonahkagan, line 118: Loved that you factored out all this repeated code!

In general I try to avoid having global shared objects between tests, because I find that sometimes tests can get intertwined and mess with each other in weird ways. Seems like this is working fine right now, so you don't necessarily need to change it, but I lean towards a pattern of trying to keep as much local to a test function as possible.

I factoring out repeated setup functionality by creating a function to do that work and return the results, e.g.:

```ts
function setupFixtures() {
  return {
      card: new MemoryCard(),
      storage: new MemoryStorage(),
      ... etc ...
  }
}
```
and then I destructure the result in each test:
```ts
const { card, storage } = setupFixtures()
```

Or sometimes I even include the rendering of the component in that setup function!

jonahkagan (approved):
Thanks for cleaning all this up. Looks great!

frontends/precinct-scanner/src/app.test.tsx
adghayes: Here I'm trying to hit the next threshold.

jonahkagan: Ah ok I get it now. Had to read the test a bit closer

frontends/precinct-scanner/src/app.test.tsx
adghayes: I see, that seems like a safer pattern. I'm going to leave it for now but in the future plan to do what you suggest.

PR #2467, opened 2022-08-23
test(scan): Optimize precinct scanner app tests

jonahkagan:
## Overview
- Reduce some of the delays
- Mock interpretation where possible

Reduces the total test time from ~140s to ~70s on my machine


arsalansufi (approved):
Nice!

PR #2477, opened 2022-08-25
refactor(admin): extract `useElectionManagerStore` hook

eventualbuddha:
## Overview

This hook encapsulates all the loading & persisting of data to the storage mechanism, which for now is local/kiosk storage. I attempted to make some of the operations play better with an eventual HTTP API such as having "add" functions rather than manually getting the old value, appending the new thing, and overwriting the whole thing. Some were left alone because they'd have involved a larger refactor than I was comfortable with in one PR.

This ended up revealing a few unused bits like the VxAdmin demo app, so I removed those too.

## Demo Video or Screenshot
n/a

## Testing Plan 
I relied heavily on automated testing for this. I also manually tested the basic configure/unconfigure flows, loading of CVRs, and viewing tally reports in kiosk-browser and Chrome.

## Checklist


frontends/election-manager/src/contexts/app_context.ts
```diff
@@ -33,19 +33,16 @@ export interface AppContextInterface {
   printBallotRef?: RefObject<HTMLElement>;
   saveCastVoteRecordFiles: SaveCastVoteRecordFiles;
   saveElection: SaveElection;
-  setCastVoteRecordFiles: React.Dispatch<
```
jonahkagan, line 36: What if the store was accessible through a separate context? It might be clearer to know which data/functions interact with the store vs which don't. Currently it gets obscured by adding functions to this app context. Might be a bigger change than you want to make for this PR, but wanted to at least get your thoughts.

frontends/election-manager/src/hooks/use_election_manager_store.ts
```diff
@@ -0,0 +1,697 @@
...
+    }
+  }, [logger, storage]);
+
+  const getElectionDefinitionQuery = useQuery<ElectionDefinition | null>(
+    [electionDefinitionStorageKey],
+    async () => {
+      return (await loadElectionDefinition()) ?? null;
+    }
+  );
+  const electionDefinition = getElectionDefinitionQuery.data ?? undefined;
```
jonahkagan, line 229: Why the pattern of having the `useQuery` return `TData | null` and then replacing `null` with `undefined` here?

frontends/election-manager/src/hooks/use_election_manager_store.ts
```diff
@@ -0,0 +1,697 @@
...
+      disposition: 'success',
+      storageKey: configuredAtStorageKey,
+      configuredAt,
+    });
+
+    return configuredAt;
+  }, [logger, storage]);
+
+  const getConfiguredAtQuery = useQuery<string | null>(
+    ['configuredAt'],
```
jonahkagan, line 262: I noticed that sometimes you use the storage key as the query key and sometimes you just use a string. Maybe pick one and use consistently?

frontends/election-manager/src/hooks/use_election_manager_store.ts
```diff
@@ -0,0 +1,697 @@
...
+      });
+    } catch (error) {
+      assert(error instanceof Error);
+      await logger.log(LogEventId.SaveToStorage, currentUserRoleRef.current, {
+        message: 'Failed clearing all current data in storage.',
+        disposition: 'failure',
+        error: error.message,
+      });
+    }
+    void queryClient.invalidateQueries([electionDefinitionStorageKey]);
```
jonahkagan, line 440: I don't remember `queryClient.invalidateQueries` being async from my previous experience with react-query. Are you sure the types are working?

frontends/election-manager/src/hooks/use_election_manager_store.ts
```diff
@@ -0,0 +1,697 @@
...
+      void logger.log(
+        LogEventId.ElectionUnconfigured,
+        currentUserRoleRef.current,
+        {
+          disposition: LogDispositionStandardTypes.Success,
+          previousElectionHash: previousElection.electionHash,
+        }
+      );
+    }
+    // we set a new election definition, reset everything
```
jonahkagan, line 424: Is this comment accurate? It doesn't seem like "resetting" implies that there was a new election definition set

frontends/election-manager/src/hooks/use_election_manager_store.ts
```diff
@@ -0,0 +1,697 @@
...
+    }
+  );
+
+  const markResultsOfficial = useCallback(async () => {
+    await markResultsOfficialMutation.mutateAsync();
+  }, [markResultsOfficialMutation]);
+
+  const addPrintedBallotMutation = useMutation(
+    async (newPrintedBallot: PrintedBallot) => {
+      const oldPrintedBallots = await loadPrintedBallots();
```
jonahkagan, line 560: isn't this already cached? is there a reason to load it again?

frontends/election-manager/src/app_root.tsx
```diff
@@ -160,70 +92,23 @@ export function AppRoot({
...
-          message: `Failed to save ${logDescription} to storage.`,
-          storageKey,
-          error: error.message,
-          disposition: 'failure',
-        });
-      }
-    },
-    [currentUserRole, logger, storage]
-  );
+  store.setCurrentUserRole(currentUserRole);
```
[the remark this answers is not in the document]
eventualbuddha: Unfortunately I can't because `currentUserRole` depends on `auth` which depends on `electionDefinition` which depends on `store`.

frontends/election-manager/src/app_root.tsx
```diff
@@ -387,167 +166,49 @@ export function AppRoot({
       disposition: 'success',
       totalBallots,
     });
-  }, [fullElectionTally, fullElectionExternalTallies, currentUserRole, logger]);
+  }, [
+    currentUserRole,
+    fullElectionTally.overallTally.numberOfBallotsCounted,
+    logger,
+    store.fullElectionExternalTallies,
+  ]);
```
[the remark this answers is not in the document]
eventualbuddha: Probably. I initially left this alone because the recomputation was happening here, but I moved that into the store so this should move as well.

frontends/election-manager/src/contexts/app_context.ts
eventualbuddha: I was thinking of just putting the `store` into the context, or even going through the work of making it so other components could `useElectionManagerStore` to get access to it, but that felt like too much to do in this PR. I agree we should move in that direction, though.

frontends/election-manager/src/app_root.tsx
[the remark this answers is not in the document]
jonahkagan: Ah I see now

frontends/election-manager/src/hooks/use_election_manager_store.ts
eventualbuddha: Yeah, I'll pick one.

frontends/election-manager/src/hooks/use_election_manager_store.ts
eventualbuddha: From the [docs](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientinvalidatequeries):

```tsx
await queryClient.invalidateQueries(['posts'], {
  exact,
  refetchType: 'active',
}, { throwOnError, cancelRefetch })
```

frontends/election-manager/src/hooks/use_election_manager_store.ts
eventualbuddha: That comment got carried around from when that was the only way a reset happened. Will remove.

frontends/election-manager/src/hooks/use_election_manager_store.ts
eventualbuddha: I can't just use the data from the `printedBallots` query since its data may not be available yet. Do you know of a better way to handle this?

frontends/election-manager/src/hooks/use_election_manager_store.ts
jonahkagan: Ah, I guess it's slightly incompatible with our conventions around `null` and `undefined`. An explanatory comment might be helpful for future reference

frontends/election-manager/src/hooks/use_election_manager_store.ts
jonahkagan: TIL!

I was always following the examples from the [guide](https://tanstack.com/query/v4/docs/guides/invalidations-from-mutations) which don't away the promise...

frontends/election-manager/src/hooks/use_election_manager_store.ts
jonahkagan: Hmm I see. I guess it's more straightforward when the backend API supports adding one new item.

I guess you could assert that the printed ballots data must be loaded before calling this mutation (which I would imagine is the case where it's used). Not sure it's a worthwhile optimization

jonahkagan: Looks like you did handle a similar case slightly differently below in `addFullElectionExternalTallyMutation`

frontends/election-manager/src/hooks/use_election_manager_store.ts
eventualbuddha: Will add a comment.

frontends/election-manager/src/hooks/use_election_manager_store.ts
eventualbuddha: I'll unify them to both load when doing this. As you say, when we have an API that supports adding directly we can remove this.

frontends/election-manager/src/hooks/use_election_manager_store.ts
eventualbuddha: Comment added.

frontends/election-manager/src/hooks/use_election_manager_store.ts
eventualbuddha: Picked the `…StorageKey` variables.

frontends/election-manager/src/app_root.tsx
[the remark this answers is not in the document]
eventualbuddha: Actually, this is based on something that's computed in `AppRoot` so I'm going to leave it here for now.

jonahkagan (approved):

PR #2481, opened 2022-08-25
Back up either original or normalized scan images depending on scanned ballot count

arsalansufi:
_Commit by commit reviewing recommended_

## Overview

We've been hitting file size limits exporting backups to FAT32 formatted USB drives, which have a 4GB file size limit. As a short-term mitigation for m14, after a certain number of ballots have been scanned, we'll only export normalized scan images (much smaller than original scan images). Before that number of ballots have been scanned, we'll only export original scan images (most useful to us for debugging). Today, we export both normalized and original images.

Through testing, we've landed on different thresholds for VxScan and VxCentralScan.

## Testing Plan 

- [x] Updated services/scan tests
- [x] Updated frontends/precinct-scanner tests
- [x] Tested VxScan manually
- [x] Tested VxCentralScan manually

## Checklist

### DIFF
--- a/services/scan/src/backup.ts
+++ b/services/scan/src/backup.ts
@@ -7,9 +7,41 @@ import Database from 'better-sqlite3';
 import { fileSync } from 'tmp';
 import ZipStream from 'zip-stream';
 import { FULL_LOG_PATH } from '@votingworks/logging';
+import { throwIllegalValue } from '@votingworks/utils';
 import { Store } from './store';
 
 const debug = makeDebug('scan:backup');
+
+/**
+ * Specifies what to include in the backup
+ *
+ * @property scanImagesToInclude Specifies which scan images to include in the backup
+ */
+export interface BackupOptions {
+  scanImagesToInclude: 'all' | 'normalizedOnly' | 'originalOnly';
+}
+
+const defaultBackupOptions: BackupOptions = {
+  scanImagesToInclude: 'all',
+};
+
+/**
+ * Parses a query param as a BackupOptions['scanImagesToInclude']
+ */
+export function parseScanImagesToIncludeQueryParam(
+  param: qs.ParsedQs[string]
+): BackupOptions['scanImagesToInclude'] {
+  switch (param) {
+    case 'all':
+      return 'all';
+    case 'normalizedOnly':
+      return 'normalizedOnly';
+    case 'originalOnly':
+      return 'originalOnly';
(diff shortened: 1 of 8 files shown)


services/scan/src/backup.ts
```diff
@@ -84,10 +115,18 @@ export class Backup {
...
+        await this.addFileEntry(sheet.front.normalized);
+        await this.addFileEntry(sheet.back.original);
+        await this.addFileEntry(sheet.back.normalized);
+      } else if (options.scanImagesToInclude === 'normalizedOnly') {
+        await this.addFileEntry(sheet.front.normalized);
+        await this.addFileEntry(sheet.back.normalized);
+      } else if (options.scanImagesToInclude === 'originalOnly') {
+        await this.addFileEntry(sheet.front.original);
+        await this.addFileEntry(sheet.back.original);
+      }
```
eventualbuddha, line 129:
```suggestion
      } else {
        throwIllegalValue(options.scanImagesToInclude);
      }
```

eventualbuddha (commented):
Looks like you're on the right track.

services/scan/src/precinct_scanner_app.ts
```diff
@@ -459,7 +459,10 @@ export async function buildPrecinctScannerApp(
     }
   );
 
-  app.get('/precinct-scanner/backup', (_request, response) => {
+  app.get('/precinct-scanner/backup', (request, response) => {
+    const scanImagesToInclude = parseScanImagesToIncludeQueryParam(
+      request.query['scanImagesToInclude']
+    );
```
jonahkagan, line 465: Any particular reason to have a query parameter and determine this on the frontend vs just doing it all on the backend?

carolinemodic (approved):

services/scan/src/precinct_scanner_app.ts
arsalansufi: Mm a really great question - thinking about this some more, looking at the sheet counts on the server would be a really clean and simple way to do this

I'm going to merge this PR as is to have a mitigation in place and to unblock others, but I'll explore the alternative and potentially get a PR up for the alternative tomorrow morning

arsalansufi: @jonahkagan I did revisit this and tweaked the implementation in https://github.com/votingworks/vxsuite/pull/2484!

PR #2484, opened 2022-08-26
Contain space-optimized backup logic to the scan service

arsalansufi:
## Overview

We quickly prepared a patch in https://github.com/votingworks/vxsuite/pull/2481 to avoid file size limits when exporting backups to FAT32 formatted USB drives. @jonahkagan pointed out that the logic in that PR, which spans both the backend scan service and VxScan/VxCentralScan frontends, could be contained to just the backend scan service.

This PR tweaks the implementation of the optimization accordingly. In addition to cleanliness/simplicity, this tweak also has the nice benefit of counting failed scans toward the "switch to normalized images only" threshold. That said, I don't see an urgent need to incorporate this tweak into our m14 images.

## Testing

- [x] Updated scan service tests
- [x] Tested VxScan with a modified threshold manually
- [x] Tested VxCentralScan with a modified threshold manually

## Checklist

### DIFF
--- a/services/scan/src/backup.ts
+++ b/services/scan/src/backup.ts
@@ -7,7 +7,6 @@ import Database from 'better-sqlite3';
 import { fileSync } from 'tmp';
 import ZipStream from 'zip-stream';
 import { FULL_LOG_PATH } from '@votingworks/logging';
-import { throwIllegalValue } from '@votingworks/utils';
 import { Store } from './store';
 
 const debug = makeDebug('scan:backup');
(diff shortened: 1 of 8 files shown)


services/scan/src/backup.ts
```diff
@@ -7,40 +7,21 @@ import Database from 'better-sqlite3';
...
-    case 'all':
-      return 'all';
-    case 'normalizedOnly':
-      return 'normalizedOnly';
-    case 'originalOnly':
-      return 'originalOnly';
-    default:
-      return defaultBackupOptions.scanImagesToInclude;
-  }
+  saveOnlyOriginalImagesThenOnlyNormalizedImagesAfterNumSheets?: number;
```
arsalansufi, line 24: Long variable name I know 😅 but the most accurate descriptor I could come up with

adghayes (approved):
lgtm!

PR #2486, opened 2022-08-27
fix(ps) save plustek images as workspace ballot image path

carolinemodic:
## Overview

We are automatically saving plustek images to /tmp, in prod /tmp only has 1GB of space so every ~300 ballots we run out of space and require a reboot, this fixes this by moving to a directory in the ballot-images workspace. Conveniently we already just link files when we put them in ballot-images so it doesn't actually increase the space we are using at all. 

## Testing Plan 
Tested with and without the proper directory already existing, tested exporting a backup and made sure images got exported properly. 

## Checklist

### DIFF
--- a/services/scan/src/precinct_scanner_app.test.ts
+++ b/services/scan/src/precinct_scanner_app.test.ts
@@ -208,6 +208,7 @@ async function createApp(
   }
   const precinctScannerMachine = createPrecinctScannerStateMachine(
     createPlustekClient,
+    workspace,
     logger,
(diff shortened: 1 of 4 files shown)


carolinemodic:
There are so many conflicts between the release branch and main this PR is just for the release branch I'll leave the change on main to @eventualbuddha or @jonahkagan to do on Monday but I just want to get this for testing this weekend.

eventualbuddha (approved):

PR #2488, opened 2022-08-29
fix(ps): show infobar on polls closed

### LINKED ISSUE #2425: VxScan: the polls closed screen doesn’t have the election info bar
VxScan: The polls closed screen doesn’t have the election info bar. It’s the first screen I saw after configuring the machine, so I was hoping to confirm it was configured correctly. In contrast, when the polls are open, the election info bar is shown.

Issue automatically created from slack feedback reported by jonah 
https://votingworks.slack.com/archives/CJU9MSC6S/p1660774237911019?thread_ts=1660774237.911019&cid=CJU9MSC6S

adghayes:
Closes #2425. The Election Info Bar should have been on this page - and technically was - but wasn't able to display because the screen wasn't being passed appropriate app context.

### DIFF
--- a/frontends/precinct-scanner/src/app_root.tsx
+++ b/frontends/precinct-scanner/src/app_root.tsx
@@ -570,11 +570,22 @@ export function AppRoot({
 
   if (!isPollsOpen) {
     return (
-      <PollsClosedScreen
-        isLiveMode={!isTestMode}
-        showNoChargerWarning={!computer.batteryIsCharging}
-        scannedBallotCount={scannerStatus.ballotsCounted}
-      />
+      <AppContext.Provider
+        value={{
+          electionDefinition,
+          currentPrecinctId,
+          currentMarkThresholds,
+          machineConfig,
+          auth,
+          isSoundMuted,
+        }}
+      >
+        <PollsClosedScreen
+          isLiveMode={!isTestMode}
+          showNoChargerWarning={!computer.batteryIsCharging}
+          scannedBallotCount={scannerStatus.ballotsCounted}
+        />
+      </AppContext.Provider>
     );
   }
 
--- a/frontends/precinct-scanner/src/screens/polls_closed_screen.test.tsx
+++ b/frontends/precinct-scanner/src/screens/polls_closed_screen.test.tsx
@@ -43,6 +43,12 @@ describe('PollsClosedScreen', () => {
     await screen.findByText('No Power Detected.');
   });
 
+  test('does not show "No Power Detected" when not called for', async () => {
+    renderScreen(false);
+    await screen.findByText('Polls Closed');
+    expect(screen.queryAllByText('No Power Detected.').length).toBe(0);
+  });
+
   test('shows ballot count', async () => {
     renderScreen(false);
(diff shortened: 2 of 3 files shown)


frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -570,11 +570,22 @@ export function AppRoot({
...
+          auth,
+          isSoundMuted,
+        }}
+      >
+        <PollsClosedScreen
+          isLiveMode={!isTestMode}
+          showNoChargerWarning={!computer.batteryIsCharging}
+          scannedBallotCount={scannerStatus.ballotsCounted}
+        />
+      </AppContext.Provider>
```
jonahkagan, line 588: Maybe move this down into the `voterScreen` variable so it gets the context automatically?

jonahkagan (approved):
Possible to add a regression test?

frontends/precinct-scanner/src/app_root.tsx
adghayes: That would be confusing to me personally, since it's not a voter screen.

I think it would make sense to wrap in this component in `AppContext`, as this is now the fourth time in the file we've had to explicitly do it. I don't really want do that refactor for this PR, however, because I think the current approach is fine and maybe more readable. Although I could write a helper to wrap in `AppContext` if you think that would be helpful.

adghayes:
@jonahkagan I just created a test file for the screen and covered a few things, including the election info bar. Please give it a look and let me know if you think that is sufficient for testing.

frontends/precinct-scanner/src/app_root.tsx
jonahkagan: I think it's fine for now

frontends/precinct-scanner/src/screens/polls_closed_screen.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { render, screen } from '@testing-library/react';
+import { Inserted } from '@votingworks/test-utils';
+import { electionSampleDefinition } from '@votingworks/fixtures';
+import { AppContext } from '../contexts/app_context';
+import { PollsClosedScreen } from './polls_closed_screen';
+
+const TEST_BALLOT_COUNT = 50;
+const MACHINE_ID = '0003';
+
+function renderScreen(showNoChargerWarning: boolean) {
```
jonahkagan, line 11: Maybe change this to take `props: Partial<PollsClosedScreenProps> = {}` as its argument and then call like `renderScreen({ showNoChargerWarning: false })`. This is a pattern I've seen us using elsewhere. Otherwise it's hard to know what the boolean means just from looking at the call site of `renderScreen`.

jonahkagan (approved):
Looks great! Thanks for adding those tests!!

frontends/precinct-scanner/src/screens/polls_closed_screen.test.tsx
adghayes: We love some well-targeted Typescript, thanks for the suggestion.

PR #2493, opened 2022-08-30
Fix: Disable Buttons to Touch

adghayes:
## Overview
Currently our buttons components, when disabled, can still be activated by touch. This is a fix for that.

## Video of Problem
[button-no-disable.webm](https://user-images.githubusercontent.com/37960853/187503185-84b22774-5b43-4391-a514-388ec6c2b6ef.webm)

## Testing Plan
- New tests in `libs/ui`
- New tests in `frontends/bsd`
- Manual Testing
- Smoke test of EM after switching to shared component

## Notes
It was fairly easy to switch VxAdmin to the shared component because the style differences were not significant. 
- `rem`-> `em`, which has no impact because no buttons in VxAdmin are within components with a custom font size
- `big` prop renamed to `large`, which has no impact because the prop isn't being used
- `text-align` isn't getting explicitly set to a default of `center` anymore, but it doesn't seem to impact the UI
- `outline: none;` isn't getting explicitly set without the `noFocus` prop, but it doesn't seem to impact the UI with touch or click

VxCentral has some actual differences in style so @beausmith I'm going to make a low priority tech improvement ticket. I just applied the same fix in `bsd` as I did in `libs/ui` and added minimal testing there.


jonahkagan (approved):

beausmith (approved):
🙏 Nice find and fix Drew!

PR #2495, opened 2022-08-30
fix(em): make test deck tally report look same as (un)official

### LINKED ISSUE #1552: VxAdmin: Test Deck Expected Results Report should have identical layout to normal tally report
Right now, three visible differences between the Test Deck Expected Results Tally Report and a real tally report:
- the "Ballots By Voting Method" table is missing in the expected results
- the font size appears to be a tad bigger for Expected vs. real.
- the title bar is different, bigger font, on expected.

We should make Expected look like the current real tally report, with the only difference being in the words in the title of the report.

adghayes:
## Overview
Closes #1552. The test deck tally report was using a different component than our actual reports. This standardizes them, so the test deck tally report is using the same underlying component and will change as our main report changes. Specifically, it makes the title look the same and adds the voting method table. Took the chance to abstract a little logic into the `TestDeckTallyReport` component.

### New Test Deck Report Look
[full-election-test-new.pdf](https://github.com/votingworks/vxsuite/files/9455722/full-election-test-new.pdf)

### New (and Old) Regular Report Look
[full-election-unofficial-same.pdf](https://github.com/votingworks/vxsuite/files/9455723/full-election-unofficial-same.pdf)

## Testing Plan 
- Manual Testing
- Need to update a snapshot

## Checklist

### DIFF
--- a/frontends/election-manager/src/__snapshots__/app.test.tsx.snap
+++ b/frontends/election-manager/src/__snapshots__/app.test.tsx.snap
@@ -281,7 +281,7 @@ exports[`L&A (logic and accuracy) flow 1`] = `
     </div>
   </div>
   <div
-    class="sc-hiCivh bWrhyG"
+    class="sc-hiCivh bWrhyG print-only"
   >
     <section
       class="sc-kLwgWK fnXcLU"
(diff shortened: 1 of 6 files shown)


arsalansufi (approved):
Awesome, LGTM!

PR #2496, opened 2022-08-30
Reduce Unmount Delay & Sync On Unmount

### LINKED ISSUE #2248: "Ejecting..." Delay on File Sync, not 10s
## Summary

Currently, when we unmount a USB drive, `kiosk-browser` waits 10s before confirming it is unmounted. This is an annoying and unnecessary delay, and we're not actually confirming that the disk is safe to remove before telling the user they can remove it. 

## Requirements

Instead of 10s delay, wait for a `sync` and wait 1s before prompting user to remove USB drive. I lean towards making this the default behavior at the `kiosk-browser` level rather than managing the logic in our app, in which case there are no changes to `vxsuite` that need to be made. But we could also manage it in `vxsuite`.

adghayes:
## Overview
Closes #2248. Instead of waiting a static 10 seconds to unmount USBs (hoping that that 10s will ensure the data is flushed to disk), trigger a `sync` before each unmount and wait a minimum of one seconds so the user has some feedback.

## Testing Plan 
- Updated automated testing where necessary
- Manual testing of ejecting the USB in different apps. It's a little challenging to test a longer delay because all of our files are small enough to eject within the 1 second delay, and our largest file exports, the backups, are set up so that they will already be synced by the time the user is prompted to eject.

## Checklist

### DIFF
--- a/libs/ui/src/hooks/use_usb_drive.test.tsx
+++ b/libs/ui/src/hooks/use_usb_drive.test.tsx
@@ -21,8 +21,8 @@ async function waitForStatusUpdate(): Promise<void> {
   await advanceTimersAndPromises(POLLING_INTERVAL_FOR_USB / 1000);
 }
 
-async function waitForIoFlush(): Promise<void> {
-  await advanceTimersAndPromises(usbstick.FLUSH_IO_DELAY_MS / 1000);
+async function waitForUnmount(): Promise<void> {
+  await advanceTimersAndPromises(usbstick.MIN_TIME_TO_UNMOUNT_USB / 1000);
 }
 
 let logger: Logger;
@@ -170,7 +170,7 @@ test('full lifecycle with USBControllerButton', async () => {
   expect(kiosk.unmountUsbDrive).toHaveBeenCalled();
   await waitForStatusUpdate();
   await advanceTimersAndPromises();
-  await waitForIoFlush();
+  await waitForUnmount();
 
   await waitForStatusUpdate();
   screen.getByText('Ejected');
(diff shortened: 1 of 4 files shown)


eventualbuddha (commented):
Seems like a reasonable approach. What testing have you done?

adghayes:
I've only tested the `Eject USB` button on VxAdmin to confirm it could work, there are at least a few more places to give it a try. The underlying `sync` API with kiosk-browser was tested when we added it for the precinct scanner polls closed CVR export, this is just using that API elsewhere.

The main risk I see in this is, what happens if `sync` hangs? `sync` should hang if a file transfer is not complete, but I'm wondering if there are cases where it hangs and it shouldn't. In my own use of the command I've never seen `sync` hang.  I wanted your input particularly in case there was something about `sync` that we should be worried about.

eventualbuddha:
I don't know of any cases to be worried about, but I am not super familiar with `sync`. Maybe @umbernhard has thoughts?

adghayes:
The one caveat I do know about `sync` is that it will only flush what's in the cache. If a file is not written to the cache yet, then `sync` will exit before the download is complete. For example we will be fine if we do:
```
await download();
await usbEject();
```
But we might run into trouble if we did:
```
void download();
await usbEject();
```
We wouldn't have a reason to do that example specifically, but if there are cases where we prompt the user to eject before a file has finished downloading (which, AFAIK, is nowhere in the apps) we could have an incomplete transfer. I think this is an edge case for us, since we always wait on downloads for UI state anyway, like to go to the final state of a modal.

umbernhard:
sync blocks on writing dirty blocks in the filesystem to disk. So if the data isn't there is won't know to wait for the write. You want to make sure you filesystem is aware of the whole file before calling (as @adghayes pointed out). Other than that, it should be pretty foolproof.

adghayes:
OK, then this PR should be good to go. I've confirmed that ejection works on VxAdmin and we not ending up with ill-formed files (I checked a ballot package and ballot pdf export on VxAdmin, in dev). I mostly feel comfortable with this change since we've thoroughly tested the same with the VxScan CVR export in solving #1981.

eventualbuddha (approved):

PR #2501, opened 2022-08-30
refactor(admin): introduce `backend` abstraction 

eventualbuddha:
## Overview
This is a continuation of https://github.com/votingworks/vxsuite/pull/2477. That PR added `react-query` and `useElectionManagerStore` to encapsulate it. Previously, `useElectionManagerStore` took a `storage` prop to control how the data was stored and retrieved to make testing and encapsulation easier. This PR replaces `storage` with `backend` and introduces a storage-based backend that it still uses for now, but will be replaced with an admin service-based `backend`. Additionally, a `memory`-based `backend` is introduced to make testing easier.

## Demo Video or Screenshot
n/a

## Testing Plan 
Mostly updated and added automated tests. Also did manual testing of basic config/tally/unconfig flows.

## Checklist


frontends/election-manager/src/app_root.tsx
```diff
@@ -44,8 +44,11 @@ import {
 } from './config/types';
 import { useElectionManagerStore } from './hooks/use_election_manager_store';
 import { getExportableTallies } from './utils/exportable_tallies';
+import { ElectionManagerStoreBackend } from './lib/backends/types';
 
 export interface Props {
+  logger: Logger;
+  backend: ElectionManagerStoreBackend;
   storage: Storage;
```
jonahkagan, line 52: I think you can remove `storage` now that it's not used

frontends/election-manager/src/app.tsx
```diff
@@ -1,38 +1,50 @@
...
-const defaultStorage = window.kiosk
-  ? new KioskStorage(window.kiosk)
-  : new LocalStorage();
-
 export function App({
+  backend: backendFromProps,
   hardware = getHardware(),
   card = new WebServiceCard(),
-  storage = defaultStorage,
+  storage,
```
jonahkagan, line 27: does `App` need both `backend` and `storage`? Or can it just take `backend`?

frontends/election-manager/src/lib/backends/memory_backend.ts
```diff
@@ -0,0 +1,148 @@
...
+} from '@votingworks/types';
+import { PrintedBallot } from '../../config/types';
+import { ElectionManagerStoreBackend } from './types';
+import { CastVoteRecordFiles } from '../../utils/cast_vote_record_files';
+
+/**
+ * An in-memory backend for ElectionManagerStore. Useful for tests or an
+ * ephemeral session.
+ */
+export class ElectionManagerStoreMemoryBackend
```
jonahkagan, line 15: Why mock at the level of the backend vs the actual storage itself? Seems like we'll get more thorough testing out of it if we mock storage (and then after switching to a server backend, mock fetch)

frontends/election-manager/src/app_root.tsx
```diff
@@ -79,7 +78,7 @@ export function AppRoot({
 
   const store = useElectionManagerStore({
     logger,
-    storage,
+    backend,
   });
```
jonahkagan, line 82: Consider using a context provider to make the store available throughout the app. Maybe something for a later PR

frontends/election-manager/src/app.tsx
```diff
@@ -32,34 +29,20 @@ const defaultStorage = window.kiosk
 const queryClient = new QueryClient();
 
 export function App({
-  hardware,
+  hardware = getHardware(),
```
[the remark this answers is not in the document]
eventualbuddha: Extracted as #2503.

frontends/election-manager/src/lib/backends/memory_backend.ts
eventualbuddha: Mainly because, when we do switch to fetching from the backend service, I don't want to have `fetchMock.*` calls all over the place. I'd rather be able to have my own mock that can be more focused. And the `backend` object itself can be heavily unit tested, which I think should provide pretty solid integration assurance.

frontends/election-manager/src/app_root.tsx
eventualbuddha: Were you thinking it would be like this?

```ts
// app.tsx
function App() {
  const logger = …;
  const backend = …;

  const store = useElectionManagerStore({ backend, logger });

  return (
    <StoreContext.Provider value={{ store }}>
      …
    </StoreContext.Provider>
  );
}

// my_component.tsx
function MyComponent() {
  const { store } = useContext(StoreContext);
}
```

Or like this?

```ts
// app.tsx
function App() {
  const logger = …;
  const backend = …;

  return (
    <StoreContext.Provider value={{ backend, logger }}>
      …
    </StoreContext.Provider>
  );
}

// use_election_manager_store.ts
function useElectionManagerStore() {
  const { logger, backend } = useContext(StoreContext);

  return …;
}

// my_component.tsx
function MyComponent() {
  const store = useElectionManagerStore();
}
```

jonahkagan: I was thinking along the lines of the first example, but with a little more sugar:

```ts
// my_component.tsx
function MyComponent() {
  const store = useStore();
}
```

Another option entirely would be to create custom hooks for each query/mutation in the store (which is the pattern we use in Arlo). Just throwing that on the table for consideration, I don't necessarily think it's better.

frontends/election-manager/src/lib/backends/memory_backend.ts
jonahkagan: That makes sense. `fetchMock` can definitely be annoying. I built a replacement in Arlo because I couldn't deal with it haha.

I think the downside that I've seen with building these custom mocks is that you then have to ensure that the mock behaves as expected, which requires thinking and/or tests.

Tradeoffs...

frontends/election-manager/src/app.tsx
eventualbuddha: Yes, you're right. This just expanded the scope of the change but I think you're right that it doesn't make sense to leave it in this intermediate state.

frontends/election-manager/src/app.test.tsx
```diff
@@ -134,7 +135,9 @@ test('create election works', async () => {
   const card = new MemoryCard();
   const hardware = MemoryHardware.buildStandard();
   const { getByText, queryAllByText, getByTestId } = render(
-    <App card={card} hardware={hardware} />
+    <QueryClientProvider client={new QueryClient()}>
```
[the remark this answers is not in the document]
eventualbuddha: Done.

frontends/election-manager/src/app_root.tsx
eventualbuddha: Done.

frontends/election-manager/src/app.tsx
eventualbuddha: Done.

frontends/election-manager/src/app_root.tsx
eventualbuddha: Will think about this more for a follow-up.

jonahkagan (approved):
Nice, I like how this turned out!

PR #2504, opened 2022-08-31
feat(em): card backwards screen on card programming

### LINKED ISSUE #2417: VxAdmin: cards backwards screen when doing card programming
VxAdmin: Might be nice to have feedback that card is backwards when doing card programming

Issue automatically created from slack feedback reported by jonah 
https://votingworks.slack.com/archives/CJU9MSC6S/p1660758972167679

adghayes:
## Overview
Closes #2417. Shows a "Card is Backwards" screen if card is inserted backwards during card programming.

## Demo Video or Screenshot

[card-backwards.webm](https://user-images.githubusercontent.com/37960853/187802451-75e2c2da-1be7-4ce0-bcd0-5d68c529fd7e.webm)

## Testing Plan 
- Added test in `libs/ui` where the hook changed slightly
- Added test in `frontends/election-manager` to test the screen comes up.

## Checklist

### DIFF
--- a/libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
+++ b/libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
@@ -31,19 +31,20 @@ export interface RemoveCard {
   readonly user: SystemAdministratorUser | ElectionManagerUser;
 }
 
+export type ProgrammableCard =
+  | CardSummaryNotReady
+  | ({ status: 'ready' } & CardProgramming & CardStorage);
+
 export interface SystemAdministratorLoggedIn {
   readonly status: 'logged_in';
   readonly user: SystemAdministratorUser;
-  readonly card:
-    | CardSummaryNotReady['status']
-    | (CardStorage & CardProgramming);
+  readonly programmableCard: ProgrammableCard;
   readonly logOut: () => void;
 }
 
 export interface ElectionManagerLoggedIn {
   readonly status: 'logged_in';
   readonly user: ElectionManagerUser;
-  readonly card: CardSummaryNotReady['status'] | CardStorage;
   readonly logOut: () => void;
 }
 
--- a/libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
+++ b/libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
@@ -1,3 +1,4 @@
+import { CardSummaryNotReady } from '../card';
 import {
   ElectionManagerUser,
   CardProgramming,
@@ -30,19 +31,19 @@ export interface RemoveCard {
   readonly user: SystemAdministratorUser | ElectionManagerUser;
 }
 
-export type CardSummaryNotReadyStatus = 'no_card' | 'error';
-
 export interface SystemAdministratorLoggedIn {
   readonly status: 'logged_in';
   readonly user: SystemAdministratorUser;
-  readonly card: CardSummaryNotReadyStatus | (CardStorage & CardProgramming);
+  readonly card:
+    | CardSummaryNotReady['status']
+    | (CardStorage & CardProgramming);
   readonly logOut: () => void;
 }
 
 export interface ElectionManagerLoggedIn {
   readonly status: 'logged_in';
   readonly user: ElectionManagerUser;
-  readonly card: CardSummaryNotReadyStatus | CardStorage;
+  readonly card: CardSummaryNotReady['status'] | CardStorage;
   readonly logOut: () => void;
 }
 
--- a/libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
+++ b/libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
@@ -30,17 +30,19 @@ export interface RemoveCard {
   readonly user: SystemAdministratorUser | ElectionManagerUser;
 }
 
+export type CardSummaryNotReadyStatus = 'no_card' | 'error';
+
 export interface SystemAdministratorLoggedIn {
   readonly status: 'logged_in';
   readonly user: SystemAdministratorUser;
-  readonly card?: CardStorage & CardProgramming;
+  readonly card: CardSummaryNotReadyStatus | (CardStorage & CardProgramming);
   readonly logOut: () => void;
 }
 
 export interface ElectionManagerLoggedIn {
   readonly status: 'logged_in';
   readonly user: ElectionManagerUser;
-  readonly card?: CardStorage;
+  readonly card: CardSummaryNotReadyStatus | CardStorage;
   readonly logOut: () => void;
 }
 
--- a/libs/ui/src/hooks/smartcard_auth/auth_helpers.test.ts
+++ b/libs/ui/src/hooks/smartcard_auth/auth_helpers.test.ts
@@ -39,14 +39,12 @@ const allowedUserRoles: UserRole[] = [
 
 function assertAndGetPostDipCard(
   result: RenderResult<DippedSmartcardAuth.Auth>
-): CardStorage & CardProgramming {
+): CardProgramming & CardStorage {
   assert(isSystemAdministratorAuth(result.current));
-  const { card } = result.current;
-  expect(card).not.toEqual('no_card');
-  assert(card !== 'no_card');
-  expect(card).not.toEqual('error');
-  assert(card !== 'error');
-  return card;
+  const { programmableCard } = result.current;
+  expect(programmableCard.status).toEqual('ready');
+  assert(programmableCard.status === 'ready');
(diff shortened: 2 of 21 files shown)


adghayes:
@jonahkagan I was wondering about how the handle the type `'no_card' | 'error'`. I define a little type in `libs/ui`, but it already exists in `libs/utils` as `CardSummaryNotReady.status`. If I wanted to re-use it, I could add the `utils` library as a dependency and bring it in. But I'm not sure why all those types are in the `utils` library anyway, and not in `types`, so I could move them over. End of the day, not sure how much it matters.

libs/ui/src/hooks/smartcard_auth/auth_helpers.test.ts
```diff
@@ -30,6 +37,18 @@ const allowedUserRoles: UserRole[] = [
   'cardless_voter',
 ];
 
+function assertAndGetPostDipCard(
+  result: RenderResult<DippedSmartcardAuth.Auth>
+): CardStorage & CardProgramming {
+  assert(isSystemAdministratorAuth(result.current));
+  const { card } = result.current;
+  expect(card).not.toBe('no_card');
```
arsalansufi, line 45: I've started to always use `.toEqual` instead of a mix of `.toEqual` and `.toBe`. Though my opinions aren't as strong as the [title of this article](https://dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca), it raises some good points about the benefits of always using `.toEqual`!

libs/ui/src/hooks/smartcard_auth/auth_helpers.test.ts
```diff
@@ -219,11 +238,10 @@ describe('Card interface', () => {
     assert(isSystemAdministratorAuth(result.current));
 
     // Insert an unprogrammed card
-    expect(result.current.card).not.toBeDefined();
+    expect(result.current.card).toBe('no_card');
```
arsalansufi, line 241: Same comment as above

libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
```diff
@@ -30,17 +30,19 @@ export interface RemoveCard {
   readonly user: SystemAdministratorUser | ElectionManagerUser;
 }
 
+export type CardSummaryNotReadyStatus = 'no_card' | 'error';
```
arsalansufi, line 33: Great question about the organization of types across `libs/utils` and `libs/types` - I'm tempted to say that we should move the card types from `libs/utils/src/types.ts` to a `libs/types/src/card.ts` file

Trying to import the types from `libs/utils` into `libs/types` will result in a circular dependency, since `libs/utils` also imports `libs/types`. I think of `libs/types` as one of the most fundamental, low-level packages in VxSuite. Other VxSuite packages should depend on it, but it should depend on no other VxSuite packages

If you'd like to handle this in a separate PR, totally fine by me!

arsalansufi (commented):
Looks great!

libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
```diff
@@ -30,17 +30,19 @@ export interface RemoveCard {
...
-  readonly card?: CardStorage & CardProgramming;
+  readonly card: CardSummaryNotReadyStatus | (CardStorage & CardProgramming);
   readonly logOut: () => void;
 }
 
 export interface ElectionManagerLoggedIn {
   readonly status: 'logged_in';
   readonly user: ElectionManagerUser;
-  readonly card?: CardStorage;
+  readonly card: CardSummaryNotReadyStatus | CardStorage;
```
jonahkagan, line 45: Why does the election manager type need to change? I wouldn't expect it to change, since only system admins have programming ability.

libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
```diff
@@ -30,17 +30,19 @@ export interface RemoveCard {
   readonly user: SystemAdministratorUser | ElectionManagerUser;
 }
 
+export type CardSummaryNotReadyStatus = 'no_card' | 'error';
+
 export interface SystemAdministratorLoggedIn {
   readonly status: 'logged_in';
   readonly user: SystemAdministratorUser;
-  readonly card?: CardStorage & CardProgramming;
+  readonly card: CardSummaryNotReadyStatus | (CardStorage & CardProgramming);
```
jonahkagan, line 38: I'm a little wary of changing the type of `card` for dipped auth but not inserted auth. I've been trying to keep them matching to reduce cognitive overhead. Maybe instead we can add an intersecting card status field and change all the `card` types in all the auth hooks?

e.g.

```ts
// For user types with programming and storage
readonly card: { status: 'no_card' | 'error' } | { status: 'ready' } & CardStorage & CardProgramming
// For user types with just storage
readonly card: { status: 'no_card' | 'error' } | { status: 'ready' } & CardStorage
```

or if we want to get fancy and make it more concise:

```ts
type Card<TCardActions> = { status: 'no_card' | 'error' } | { status: 'ready' } & TCardActions
// For user types with programming and storage
readonly card: Card<CardStorage & CardProgramming>
// For user types with just storage
readonly card: Card<CardStorage>
```

libs/ui/src/hooks/smartcard_auth/auth_helpers.test.ts
jonahkagan: Ooh good call, I am down to adopt that convention. Wonder if there's an easy lint rule for it

eventualbuddha: I don't think [eslint-plugin-jest](https://github.com/jest-community/eslint-plugin-jest#readme) has that, but we could write our own.

arsalansufi: Nice, glad we're in alignment! Made an issue to track: https://github.com/votingworks/vxsuite/issues/2510

libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
adghayes, line 45: I thought I saw a comment about why we changed this, but now I can't find it @jonahkagan. I don't understand why there is the possibility of this being anything but undefined (or `no_card`) because if Election Managers don't do smartcard programming then we shouldn't be caring about any cards that are inserted while they're logged in anyway.

adghayes: Why is there a 'card' attribute here at all?

jonahkagan: The `card` attribute is there to access card storage (aka "long value") (not sure if that's actually used for anything in the product by Election Managers today).

jonahkagan: (Also, yes, I had commented here but deleted it in favor of my other comment above)

libs/types/src/smartcard_auth/dipped_smartcard_auth.ts
adghayes: This doesn't make sense to me yet. For `InsertedAuth`, why would it be possible to have status be `no_card`?

Right now it makes sense to me that card has a different meaning in each of these cases. My instinct is to:
- Leave `card` for InsertedSmartCard the same as it is
- Remove `card` for DippedSmartCard for `ElectionManagerAuth`, unless we have plans to do card programming or reading in Election Manager mode
- Rename `card` for DippedSmartCard for `SystemAdministratorAuth` to make clearer that it is not the card that was used to log in but is a different card. This was confusing me at first.

adghayes: I am thinking I will still take your suggestion on cleaning up the type to avoid a union of a string and an object, but that type would only be in `Dipped`. 

I don't have ideas for renaming. `cardProgramming` works for me, especially because I don't think `CardStorage` is used anywhere in the apps where we use `Dipped` so that could be removed so simplify the type further. 

Defer to @arsalansufi in case there's more context or more coming down the pipe.

adghayes: I took out the `card` for Election Manager, but didn't pare down the one for Sys Admin because it looks like we're using CardStorage if only for testing and it does make sense to check long and short values have been cleared.

arsalansufi: Ah yes good catch!

arsalansufi (approved):
Really like where we've landed! Thanks for all the thoughtful questions and iteration!

PR #2514, opened 2022-09-01
refactor(admin): store elections in the database

eventualbuddha:
## Overview

`election-manager` now requires that the `admin` service is running to do anything interesting. This also removes all the CVR file management in the Admin service because it doesn't actually work. I'll add it back as I transition the frontend to use it.

## Demo Video or Screenshot
n/a

## Testing Plan 
Manual testing of creating and removing elections.

## Checklist


services/admin/schema.sql
```diff
@@ -20,12 +28,16 @@ create table cvrs (
...
 	filename text,
 	timestamp text,
 	imported_cvr_count integer,
 	duplicated_cvr_count integer,
 	scanner_ids text,
 	precinct_ids text,
-	contains_test_mode_cvrs boolean
+	contains_test_mode_cvrs boolean,
+	foreign key (election_id) references elections(id)
+	  on update cascade
```
jonahkagan, line 41: what does `on update cascade` do here?

services/admin/src/server.ts
```diff
@@ -176,26 +242,41 @@ export function buildApp({ store }: { store: Store }): Application {
...
     (_, response) => {
       response.json(store.getAdjudicationCountsGroupedByContestId());
     }
   );
 
-  app.get<NoParams>('/admin/write-ins/adjudications/counts', (_, response) => {
-    response.json(store.getAdjudicationCountsByContestIdAndTranscribedValue());
-  });
+  app.get<ElectionsNamespaceParams>(
+    '/admin/elections/:electionId/adjudications/grouped-by/contest-id-and-transcribed-value',
```
jonahkagan, line 253: Nit: I would expect something like `grouped-by` to be a query parameter in a normal REST API

services/admin/src/server.ts
```diff
@@ -176,26 +242,41 @@ export function buildApp({ store }: { store: Store }): Application {
...
+      );
+    }
+  );
 
-  app.get('/admin/write-ins/adjudications/:contestId/', (request, response) => {
-    const { contestId } = request.params;
-    response.json(store.getAdjudicationsByContestId(contestId));
-  });
+  app.get<ContestNamespaceParams>(
+    '/admin/elections/:electionId/contests/:contestId/adjudications',
```
jonahkagan, line 264: After designing an API for Arlo that used nested resources like this, I ended up finding it kind of overly verbose and annoying. I now lean more towards less nesting (e.g. `/admin/contests/:contestId/adjudications` or `/admin/elections/:electionId/adjudications?contest_id=<id>`). Not a big deal if you prefer this but thought I'd mention it (I know this is prime bike-shed territory 😄 )

libs/types/src/election.ts
```diff
@@ -670,11 +670,22 @@ export interface ElectionDefinition {
...
+  .superRefine((electionDefinition, ctx) => {
+    const { electionData, electionHash } = electionDefinition;
+    const electionDataHash = sha256(electionData);
+    if (electionDataHash !== electionHash) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        path: ['electionHash'],
+        message: `Election data hash '${electionDataHash}' does not match election hash '${electionHash}'.`,
+      });
+    }
```
jonahkagan, line 688: nice, this is cool

frontends/election-manager/src/lib/backends/admin_backend.ts
```diff
@@ -0,0 +1,110 @@
...
+} from '@votingworks/types';
+import { fetchJson } from '@votingworks/utils';
+import { Admin } from '@votingworks/api';
+import { ElectionManagerStoreStorageBackend } from './storage_backend';
+
+/** @visibleForTesting */
+export const activeElectionIdStorageKey = 'activeElectionId';
+
+export class ElectionManagerStoreAdminBackend extends ElectionManagerStoreStorageBackend {
+  private async loadActiveElectionId(): Promise<Id | undefined> {
```
jonahkagan, line 16: I'm a bit confused - why is there some local storage state in this backend? I expected the backend to just be a stateless interface for making fetch requests.

eventualbuddha: I modeled it as the client storing the current election that it's working with, but (long term) that's it. I toyed with having a `active` flag on `elections`, but decided it was better to just have the client manage it. In the future we'll probably want to present UI to select and switch elections, and this approach doesn't encode any server-side assumptions about that.

services/admin/src/server.ts
eventualbuddha: I buy this.

services/admin/src/server.ts
eventualbuddha: Yeah, I wasn't sure about how I wanted to handle this. I think these API endpoints probably won't survive long anyway.

services/admin/schema.sql
eventualbuddha: Were the election ID to change, I think this means that update would cascade to the `election_id` values that reference it.

services/admin/src/server.ts
eventualbuddha: I ended up removing these API endpoints altogether.

services/admin/src/server.ts
eventualbuddha: These API endpoints are now removed, but I'll keep this in mind as I add new ones back in.

frontends/election-manager/src/lib/backends/admin_backend.test.ts
```diff
@@ -0,0 +1,147 @@
...
+        createdAt: '2021-01-01T00:00:00.000Z',
+        updatedAt: '2021-01-01T00:00:00.000Z',
+      },
+      {
+        id: 'test-election-2',
+        electionDefinition: electionMinimalExhaustiveSampleDefinition,
+        createdAt: '2022-01-01T00:00:00.000Z',
+        updatedAt: '2022-01-01T00:00:00.000Z',
+      },
+    ]),
```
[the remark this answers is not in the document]
eventualbuddha: Done.

frontends/election-manager/src/lib/backends/admin_backend.ts
beausmith: Assuming the computer's clock is correct, we could use the election date in the election definition to determine which elections are in the past.

eventualbuddha: > Assuming the computer's clock is correct, we could use the election date in the election definition to determine which elections are in the past.

True, but there might be more than one in the future. Especially if you have one that's a "sandbox" or something.

jonahkagan: Ah interesting! I can totally see the need for having UI state for "this is the election I'm currently working with" and wanting that state to be remembered across restarts.

I don't think this will be the only piece of UI state with these characteristics, so it's worth thinking about a pattern we can use a bit more broadly.

I think local storage is a very sensible place to put that state, even though there's no real difference since we have a single-client/single-server architecture. It seems nice to keep the UI state/user preferences separate from the app domain state.

Maybe it would be clearer to not have this managed in the backend adapter but instead in the app itself so that we can easily reason about it and add future UI state to it as well.

services/admin/schema.sql
jonahkagan: Interesting. Doesn't seem like something we need to support but doesn't seem to cause any harm

jonahkagan (approved):

PR #2515, opened 2022-09-02
test(scan): Fix flaky precinct scanner test

jonahkagan:
## Overview
This test was flaky because sometimes the precinct scanner state machine would get to the `returned` state before the second ballot was inserted. To prevent the chance of that happening, I removed a check for `returning` state that delayed insertion of the second ballot. It's not an important check since it's covered by other tests.

### DIFF
--- a/services/scan/src/precinct_scanner_app.test.ts
+++ b/services/scan/src/precinct_scanner_app.test.ts
@@ -1063,12 +1063,6 @@ test('insert second ballot while first ballot is returning', async () => {
   await waitForStatus(app, { state: 'needs_review', interpretation });
 
   await post(app, '/precinct-scanner/scanner/return');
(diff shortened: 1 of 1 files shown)


eventualbuddha (approved):

eventualbuddha:
Seems like this may still be an issue: [logs](https://app.circleci.com/pipelines/github/votingworks/vxsuite/5707/workflows/9cca9f5e-5ac0-4b9c-98cb-a69dde5a24d2/jobs/131120?invite=true#step-108-152).

eventualbuddha:
[Another one](https://app.circleci.com/pipelines/github/votingworks/vxsuite/5724/workflows/df5b47ea-ac28-4679-a625-79f78c825a98/jobs/131662?invite=true#step-108-145), I think?

PR #2517, opened 2022-09-02
add pnpm start options to scan service

carolinemodic:
I thought these would be helpful so people don't have the remember the name of the environment variable


beausmith (approved):
😁👍

PR #2521, opened 2022-09-02
Drew - Require Scan Precinct

adghayes:
Fixes bug on main where no ballots will scan in precinct scanner on all precincts mode: https://votingworks.slack.com/archives/CEL6D3GAD/p1661984695650299. 

@eventualbuddha @jonahkagan I realize I may have navigated the types wrong, and need to redo a lot of this. I extended our `PrecinctId` type to have an explicit `AllPrecinctsId` subtype. But now I'm seeing there's duplicated typing between `Precinct` and `PrecinctSelection`. I think I should go back and, reverting the `PrecinctId` type, type everything in the scan service and frontends as `PrecinctSelection`. Want to get your confirmation that's the right way to go before digging in again.

### DIFF
--- a/frontends/bmd/src/pages/admin_screen.tsx
+++ b/frontends/bmd/src/pages/admin_screen.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from 'react';
 
-import { ElectionDefinition } from '@votingworks/types';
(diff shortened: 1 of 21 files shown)


adghayes:
Went ahead and tried the alternate approach in #2530.

arsalansufi:
Closing this since https://github.com/votingworks/vxsuite/pull/2530 has since been merged! 🧹

PR #2530, opened 2022-09-03
Precinct Selection Scan Fix + Refactor

adghayes:
Fixes bug on main where no ballots will scan in precinct scanner on all precincts mode: https://votingworks.slack.com/archives/CEL6D3GAD/p1661984695650299.

Standardizes the `PrecinctSelection` type across apps, and refactors in other apps where appropriate. I recommend reviewing by app:
- `libs/types`: some additions and edits to to `PrecinctSelection` types
- `libs/ui`: use type in `ElectionInfoBar`
- `services/scan`: use and store `PrecinctSelection` rather than `PrecinctId`
- `frontends/precinct-scanner`: use `PrecinctSelection` rather than `PrecinctId`
- `frontends/bmd`: use type from `libs/types` rather than the frontend types


libs/types/src/precinct_selection.ts
```diff
@@ -1,22 +1,65 @@
...
+  };
+}
+
+export const ALL_PRECINCTS_SELECTION: AllPrecinctsSelection = {
+  kind: 'AllPrecincts',
+};
+
+export const ALL_PRECINCTS_NAME = 'All Precincts';
+
+export function getPrecinctSelectionName(
```
arsalansufi, line 48: This might actually better fit in `libs/utils` - if we put it there, we can also make us of the `find` util instead of manually recreating that logic. The same could be said of `getSinglePrecinctSelection`

services/scan/src/importer.ts
```diff
@@ -136,6 +137,8 @@ export class Importer {
    */
   configure(electionDefinition: ElectionDefinition): void {
     this.workspace.store.setElection(electionDefinition);
+    // Central scanner only uses all precinct mode, set on every configure
+    this.workspace.store.setPrecinctSelection(ALL_PRECINCTS_SELECTION);
```
arsalansufi, line 141: Since this added logic is specific to the central scanner, we may want to move it up to `services/scan/src/central_scanner_app.ts` instead of keeping it in this lower, machine-agnostic layer

services/scan/src/store.ts
```diff
@@ -57,7 +58,7 @@ export enum ConfigKey {
   MarkThresholdOverrides = 'markThresholdOverrides',
   // @deprecated
   SkipElectionHashCheck = 'skipElectionHashCheck',
-  CurrentPrecinctId = 'currentPrecinctId',
+  PrecinctSelectionKey = 'precinctSelection',
```
arsalansufi, line 61: For consistency with the other keys in this enum, we may want to shorten this to `PrecinctSelection = 'precinctSelection'`. The `Key` in `ConfigKey` should make it clear that this is a key, e.g. `ConfigKey.PrecinctSelection`

services/scan/src/workers/interpret_vx.ts
```diff
@@ -55,6 +59,7 @@ export async function configure(store: Store): Promise<void> {
   debug('creating a new interpreter');
   interpreter = new Interpreter({
     electionDefinition,
+    precinctSelection: ALL_PRECINCTS_SELECTION,
```
arsalansufi, line 62: Is there a reason why we're using `ALL_PRECINCTS_SELECTION` here instead of `store.getPrecinctSelection()`?

libs/ui/src/election_info_bar.tsx
```diff
@@ -48,9 +43,15 @@ export function ElectionInfoBar({
         <NoWrap as="strong">{electionDefinition.election.title}</NoWrap> —{' '}
         <NoWrap>{electionDate}</NoWrap>
         <Text as="div" small>
-          {showPrecinctInfo && (
+          {precinctSelection && (
             <React.Fragment>
-              <NoWrap>{precinct?.name ?? 'All Precincts'}</NoWrap> —{' '}
+              <NoWrap>
+                {getPrecinctSelectionName(
```
arsalansufi, line 49: A slightly more intuitive name for this helper might be `getSelectedPrecinctName` or `precinctSelectionToReadableString`

frontends/bmd/src/pages/admin_screen.tsx
```diff
@@ -57,12 +59,9 @@ export function AdminScreen({
...
 
     if (precinctId === ALL_PRECINCTS_OPTION_VALUE) {
-      updateAppPrecinct({ kind: PrecinctSelectionKind.AllPrecincts });
+      updateAppPrecinct(ALL_PRECINCTS_SELECTION);
     } else if (precinctId) {
-      updateAppPrecinct({
-        kind: PrecinctSelectionKind.SinglePrecinct,
-        precinctId,
-      });
+      updateAppPrecinct(getSinglePrecinctSelection(precinctId));
```
arsalansufi, line 64: I don't feel strongly about this! `constructSinglePrecinctSelection` might be a clearer name than `getSinglePrecinctSelection`, which reads as if we're retrieving something from a store

arsalansufi (commented):
Thank you for making this consistent across VxSuite! Just a few thoughts!

libs/types/src/precinct_selection.ts
adghayes: Will do!

services/scan/src/importer.ts
adghayes: I originally had it in the higher layer but it had to be in a few places - both on initial setup and any reconfigures - to work properly.

The main point to make is that, I believe, `importer.ts` is not actually machine-agnostic. It is only used for VxCentralScan. I think @jonahkagan told me this. Will let him or @eventualbuddha confirm.

services/scan/src/store.ts
adghayes: I tried this but there's a variable name conflict in the file if you do that, will take another look to see if there's another way.

services/scan/src/workers/interpret_vx.ts
adghayes: This is a really helpful catch @arsalansufi, I think this was a mistake.

services/scan/src/importer.ts
adghayes: I was confused by this at first and if `importer.ts` is machine specific now, we might rename the file to reflect that.

services/scan/src/workers/interpret_vx.ts
arsalansufi: Ah nice!

services/scan/src/importer.ts
arsalansufi: Ah interesting - I like the idea of renaming then! Maybe `central_scanner_importer.ts` / `CentralScanImporter`

frontends/bmd/src/pages/admin_screen.tsx
adghayes: Renamed to `singlePrecinctSelectionFor`.

services/scan/src/workers/interpret_vx.ts
adghayes: Fixed.

libs/ui/src/election_info_bar.tsx
adghayes: The current name makes the most sense to me, as `Selected` feels like it implies a selection was made when that was not always the case and `ReadableString` is less clear than `Name` only because `Name` is how we've been identifying these strings in code to date.

services/scan/src/store.ts
adghayes: Renamed the key by importing the type as:
```
  PrecinctSelection as PrecinctSelectionType,
```

services/scan/src/importer.ts
adghayes: https://github.com/votingworks/vxsuite/issues/2532

libs/ui/src/election_info_bar.tsx
arsalansufi: Ah good point that the precinct isn't actually selected on VxCentralScan - the current name is okay then!

arsalansufi (approved):
LGTM!

services/scan/src/importer.ts
jonahkagan: Yep, importer is now only used by the central scanner. Renaming it makes sense!

services/scan/src/central_scanner_app.ts
```diff
@@ -197,21 +197,21 @@ export async function buildCentralScannerApp({
     response.json({ status: 'ok' });
   });
 
-  app.get<NoParams, Scan.GetCurrentPrecinctConfigResponse>(
+  app.get<NoParams, Scan.GetPrecinctSelectionConfigResponse>(
```
jonahkagan, line 200: Does the central scanner use these endpoints at all? As far as I can tell, no. And I don't see why it would need them since it's not configurable by precinct. They can probably be removed (along with the corresponding functions in frontends/bsd/src/api/config.ts)

services/scan/src/importer.ts
```diff
@@ -236,17 +241,18 @@ export class Importer {
       backInterpretation
     );
 
-    debug('currentPrecinctId=%s', currentPrecinctId);
-    if (currentPrecinctId) {
+    debug('precinctSelection=%o', precinctSelection);
+    if (precinctSelection.kind !== 'AllPrecincts') {
```
jonahkagan, line 245: Isn't this case now impossible?

services/scan/src/precinct_scanner_app.ts
```diff
@@ -248,15 +250,15 @@ export async function buildPrecinctScannerApp(
...
       return;
     }
-    store.setCurrentPrecinctId(bodyParseResult.ok().precinctId);
+    store.setPrecinctSelection(bodyParseResult.ok().precinctSelection);
     await configureInterpreter(interpreter, workspace);
     response.json({ status: 'ok' });
   });
 
-  app.delete<NoParams, Scan.DeleteCurrentPrecinctConfigResponse>(
+  app.delete<NoParams, Scan.DeletePrecinctSelectionConfigResponse>(
```
jonahkagan, line 258: Is this endpoint used anymore? I noticed you removed it from the tests

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
```diff
@@ -126,6 +137,12 @@ export function ElectionManagerScreen({
     await unconfigure();
   }
 
+  const initialPrecinctValue = precinctSelection
+    ? precinctSelection.kind === 'AllPrecincts'
+      ? ALL_PRECINCTS_OPTION_VALUE
+      : precinctSelection.precinctId
+    : '';
```
jonahkagan, line 144: This is a bit of a confusing name. The `Select` is a controlled component, so its value will always be whatever is passed into the `value` prop. So this variable isn't the "initial" value but rather just the value. I might name it `precinctSelectionValue` or something like that

frontends/bmd/src/pages/admin_screen.tsx
```diff
@@ -103,7 +101,7 @@ export function AdminScreen({
                     Select a precinct for this device…
                   </option>
                   <option value={ALL_PRECINCTS_OPTION_VALUE}>
-                    {AllPrecinctsDisplayName}
+                    All Precincts
```
jonahkagan, line 104: Didn't you make a constant for this in the precinct selection utils?

services/scan/src/workers/interpret_vx.ts
jonahkagan: @adghayes were the tests passing with this mistake? If so, that might be a bit concerning

jonahkagan (approved):
Nice work! The refactoring makes things a lot cleaner. Left a few other ideas for cleaning up more. If possible, it would make it easier to review if you can separate out the refactoring into a separate PR next time (or separate commits).

frontends/bmd/src/pages/admin_screen.tsx
adghayes: Yes, I just didn't bother importing it here. Can do so.

services/scan/src/central_scanner_app.ts
adghayes: No, we don't use them, wasn't sure if I should remove them. I will go ahead and do so.

services/scan/src/precinct_scanner_app.ts
adghayes: It is being used, yes, I'll have to check where I removed it from the tests.

services/scan/src/importer.ts
adghayes: Yes, this is a place where we still have precinct-specific logic in the central scanner. Like with the endpoint, we can remove them.

services/scan/src/precinct_scanner_app.ts
jonahkagan: I guess a more relevant question would be: is there a use case for clearing the VxScan precinct selection back to an unconfigured state? I can't really see why a user would want to do that personally.

adghayes: When you unconfigure the machine, you want it to return to a no precinct state. The precinct selection makes no sense outside of the context of the election definition, so if the election definition is removed, the precinct selection should be too.

Are you suggesting that we delete this endpoint and then just remove the precinct selection on zeroing out the scanner? That makes more sense to me, now that I'm thinking about it.

services/scan/src/precinct_scanner_app.test.ts
```diff
@@ -120,6 +119,14 @@ function postTemplate(
     .expect(200, { status: 'ok' });
 }
 
+async function setAppPrecinct(app: Application, precinctId?: PrecinctId) {
+  await put(app, '/precinct-scanner/config/precinct', {
+    precinctSelection: precinctId
+      ? singlePrecinctSelectionFor(precinctId)
+      : ALL_PRECINCTS_SELECTION,
+  });
+}
```
jonahkagan, line 128: I realized we don't have a test that confirms that unconfiguring the scanner using the `DELETE /precinct-scanner/config/election` endpoint also clears the precinct config. Reading the code, I _think_ it works, but might be worth checking/testing if you haven't already, since the frontend doesn't explicitly clear the precinct config on unconfigure.

services/scan/src/precinct_scanner_app.ts
jonahkagan: Yeah, and I think that's actually what we already do. Just left another comment about it haha

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
adghayes: Renamed.

frontends/bmd/src/pages/admin_screen.tsx
adghayes: Switched.

services/scan/src/precinct_scanner_app.test.ts
adghayes: I don't think we have a test that covers the` DELETE /precinct-scanner/config/election` endpoint at all. Will do some manual testing.

adghayes: We should also of course write a test, but I need to close this out and need to do manual testing anyway.

services/scan/src/importer.ts
adghayes: Removed.

services/scan/src/central_scanner_app.ts
adghayes: Removed.

adghayes:
@jonahkagan I addressed most of your feedback. In manual testing, I found that my changes had broken the configuration process for VxScan. It was because `configureInterpreter` now requires a precinct selection, but was being called without one having been made. I changed the logic of when we configure the interpreter in `precinct_scanner_app.ts`. The central scanner hits the `/doneTemplates` endpoint after all templates have been uploaded which triggers a configure. That's what we were doing in `precinct-scanner`, but now that you have to select a precinct, you can't configure the interpreter until after a precinct selection. All automated testing was passing before I made this change.

To your previous question, automated testing was passing before I fixed the mistake that @arsalansufi pointed out. I have not dug into why that was the case yet. 

I'm not going to merge this until I get a LGTM or another close look on your part. This turned out, in general, to be a much deeper dive than expected.

services/scan/src/precinct_scanner_app.ts
```diff
@@ -91,7 +91,7 @@ export async function buildPrecinctScannerApp(
   const { store } = workspace;
 
   // Load interpreter configuration from the store on startup
-  if (store.getElectionDefinition()) {
+  if (store.getElectionDefinition() && store.getPrecinctSelection()) {
```
jonahkagan, line 94: Since there are now a bunch of repeated checks that the precinct selection is configured before configuring the interpreter, we could move them inside the `configureInterpreter` function (along with the check for election definition).

jonahkagan (approved):

services/scan/src/precinct_scanner_app.ts
adghayes: I like that, will do one last refactor. Seems like a potential foot gun later if we don't.

adghayes:
Did some manual testing on VxScan (testing different precinct configurations, and reconfigure) and VxCentralScan (testing scanning, and reconfigure). Looks good. Merging.

PR #2531, opened 2022-09-06
feat(image-utils): add `pdfToImages`

eventualbuddha:
## Overview

Consolidates the existing two implementations into one in `image-utils`. I also plan to use it in a third package, so it seemed prudent to extract it to a shared location.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tested generating a ballot package in VxAdmin in kiosk-browser to ensure its browser-based usage still works, as that one tended to be more fragile.

## Checklist


services/scan/src/precinct_scanner_app.ts
```diff
@@ -1,6 +1,6 @@
 import { OkResponse, Scan } from '@votingworks/api';
-import * as streams from 'memory-streams';
-import { Buffer } from 'buffer';
```
umbernhard, line 3: I'm assuming the reorg here is just due to linting?

umbernhard (approved):

services/scan/src/precinct_scanner_app.ts
eventualbuddha: I used the VS Code "Organize Imports" command to sort them. I prefer to just have an automatically sorted list of imports, though I don't think we've really codified that and different packages do things a bit differently.

PR #2537, opened 2022-09-08
refactor(admin): make store updates more granular

eventualbuddha:
## Overview

When making changes to CVRs, i.e. loading a new one or clearing the list, we previously would get the list of all the CVRs and build a new one from it, then set the whole thing again. Instead, we now have distinct operations for the actions we want to perform: `add` and `clear`.

This same thing applies to the external tallies, with the addition that we now enforce that there is only ever at most one of each external tally source type.

The next step here is to add API endpoints for the CVRs in the admin service corresponding to the `add` and `clear` actions.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tested manually in VxAdmin with the exhaustive minimal sample election:
- [ ] loading CVRs, unique and duplicative
- [ ] clearing CVRs
- [ ] adding manually entered data
- [ ] adding external (SEMS) data

## Checklist

### DIFF
--- a/frontends/election-manager/src/components/import_cvrfiles_modal.tsx
+++ b/frontends/election-manager/src/components/import_cvrfiles_modal.tsx
@@ -73,7 +73,7 @@ function throwBadStatus(s: never): never {
 export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
   const {
     usbDriveStatus,
-    saveCastVoteRecordFiles,
+    addCastVoteRecordFile,
     castVoteRecordFiles,
     electionDefinition,
     auth,
@@ -97,7 +97,9 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
       fileData,
       election
     );
-    await saveCastVoteRecordFiles(newCastVoteRecordFiles);
+    await addCastVoteRecordFile(
+      new File([fileData.fileContent], fileData.name)
+    );
 
     if (newCastVoteRecordFiles.duplicateFiles.includes(fileData.name)) {
       setCurrentState(ModalState.DUPLICATE);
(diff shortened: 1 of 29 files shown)


frontends/election-manager/src/components/import_cvrfiles_modal.tsx
```diff
@@ -97,7 +97,9 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
       fileData,
       election
     );
-    await saveCastVoteRecordFiles(newCastVoteRecordFiles);
+    await addCastVoteRecordFile(
```
carolinemodic, line 100: is your plan to remove the newCastVoteRecordFiles part of this? And this addCastVoteRecordFile function will return an error if there is an error or if its a duplicate file?

frontends/election-manager/src/lib/backends/storage_backend.ts
```diff
@@ -157,30 +159,39 @@ export class ElectionManagerStoreStorageBackend
...
+
+    const loadCastVoteRecordFilesResult =
+      (await this.loadCastVoteRecordFiles()) ?? CastVoteRecordFiles.empty;
+
+    const newCastVoteRecordFiles = await loadCastVoteRecordFilesResult.add(
+      newCastVoteRecordFile,
+      loadElectionResult.electionDefinition.election
+    );
+
+    await this.setStorageKeyAndLog(
```
carolinemodic, line 178: so storage backend is still just storing things in local storage? Do we need to add an entirely new backend type for saving in the database or will this just be updated to talk to admin-service at some point?

carolinemodic (approved):

frontends/election-manager/src/lib/backends/storage_backend.ts
eventualbuddha: The `ElectionManagerStoreAdminBackend` extends this one and has bits overridden to handle talking to the admin service.

frontends/election-manager/src/components/import_cvrfiles_modal.tsx
eventualbuddha: Yes, eventually the `addCastVoteRecordFile` will return a value that can be used to replace `newCastVoteRecordFiles`.

Issue #2518, opened 2022-09-02
Load .env files in scan-service


carolinemodic:
We added[ this feature flag](https://github.com/votingworks/vxsuite/blob/main/services/scan/src/config/features.ts#L15) for exporting write-in images but its hard to use, particularly in VxDev, because the scan service doesn't load environment variables from .env and .env.local files like the frontend apps do.

Issue #2442, opened 2022-08-20
Improve management of environment variables 


carolinemodic:
Now that we are using environment variables for feature flagging and for managing workarounds to make local development simpler there are already a quickly growing list of environment variables. It is hard to remember which environment variables  exist, what they all do, the right way to spell them, etc. 

This task may be as simple as adding better documentation for the environment variables but I think there should at the very least be a forcing function to make sure it gets updated when new env variables are added. Maybe we have to add them to a file somewhere that autogenerates documentation similar to the logging repo. We could also write a small script that asks you whether or want to enable or disable various flags and then generates the right .env file for you. 

One thing this is a pain point is how you have to have these files in each application's repo and it's also hard to remember where the right place for the right flag is.

carolinemodic:
@eventualbuddha Would it be possible to just have one shared env file people can maintain at the root level of vxsuite that all of the apps know to read in from? I feel like its a little confusing and hard to manage right now which flags are available in different apps, and some like the disabling card reader check have the feature config code living in a shared directory like libs/ui and then can apply to multiple different frontends.  

I guess I'm imagining if we have just one .env file then we could put all of the code to define different feature flags in one shared locations (maybe utils or a new sub repo) and there could be a little script in there that could auto-generate a .env file for you after asking you what things you want to enable, or at least autogenerate documentation of the available flags. 

This would also make VxDev's interaction with feature flags a bit simpler as VxDev would just have to copy its config file with flags to one spot.

eventualbuddha:
The simplest way to have just one `.env` file would be to use symlinks. Or we could change the frontends to look up the directory tree to keep one at the root.

PR #2538, opened 2022-09-09
Load env file from root of vxsuite and in scan service

carolinemodic:
## Overview
Addresses https://github.com/votingworks/vxsuite/issues/2518 and takes a first step towards https://github.com/votingworks/vxsuite/issues/2442

This will cause all of the frontend react apps, and the scan-service to load environment variables from .env files both at the repo level AND at the root of vxsuite. A file in the subrepo for a particular app will supersede one at the root level. I think this will make environment variable management easier by being able to have a file just at the root level of vxsuite, that will be particularly convenient for VxDev. Next I want to colocate all of the feature flagging configuration in a shared library with a script that can autogenerate a .env file and place it at the root so that people know where to look for what environment variables can be set and configure as they want to. VxDev will then also be able to run this script. 

## Testing Plan 
Placed env files at both the local subrepo level and the root of vxsuite, running frontends with both pnpm start and running production builds saw the variables loaded with the subrepo overriding anything at the root level. 

## Checklist

### DIFF
--- a/services/scan/src/index.ts
+++ b/services/scan/src/index.ts
@@ -2,16 +2,45 @@
 import { MockScannerClient, ScannerClient } from '@votingworks/plustek-sdk';
 import { Logger, LogSource, LogEventId } from '@votingworks/logging';
 import { ok, Result } from '@votingworks/types';
+import fs from 'fs';
+import * as dotenv from 'dotenv';
+import * as dotenvExpand from 'dotenv-expand';
 import {
   MOCK_SCANNER_FILES,
   MOCK_SCANNER_HTTP,
   MOCK_SCANNER_PORT,
   VX_MACHINE_TYPE,
+  NODE_ENV,
 } from './globals';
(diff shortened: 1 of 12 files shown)


services/scan/src/index.ts
```diff
@@ -2,17 +2,46 @@
...
+  // results for everyone
+  NODE_ENV !== 'test' ? `${dotEnvPath}.local` : '',
+  `${dotEnvPath}.${NODE_ENV}`,
+  dotEnvPath,
+  NODE_ENV !== 'test' ? `../../${dotEnvPath}.local` : '',
+  `../../${dotEnvPath}`,
+].filter(Boolean);
+
+// Load environment variables from .env* files. Suppress warnings using silent
+// if this file is missing. dotenv will never modify any environment variables
```
jonahkagan, line 35: I don't see any code related to `silent` - is this comment outdated?

jonahkagan (approved):

PR #2539, opened 2022-09-09
Move feature flag configuration into shared location

carolinemodic:
## Overview

Moves all of the feature flags into one shared location so that its easier to find the complete list of different flags that you can use and to make it easier to have flags that may impact more than one app inside of vxsuite. Also adds a script to auto-generate a .env file, this will be useful in VxDev but I'm not entirely sure the best way for VxDev to interface with it. For now I added some pnpm options at the root to configure a root level .env.local file. I'm not really sure the best way to call the script open to suggestion @eventualbuddha . 

## Testing Plan 
Ran script with various different input arguments, setting prompts to different options and verified the expected .env file was generated. 
Manually tested the write in adjudication tab feature flagging was working as expected and relied on tests for the other flags relocated. 

## Checklist


.env
```diff
@@ -0,0 +1,4 @@
+REACT_APP_VX_ENABLE_WRITE_IN_ADJUDICATION=TRUE
```
carolinemodic, line 1: This is a file generated with all the default options, I'm not sure the best way to keep this up to date but we should do something there longer term.

frontends/precinct-scanner/src/screens/poll_worker_screen.test.tsx
```diff
@@ -126,7 +139,7 @@ describe('shows Export Results button only when polls are closed and more than 0
 
 describe('shows Livecheck button only when enabled', () => {
   test('enable livecheck', async () => {
-    mocked(isLiveCheckEnabled).mockReturnValue(true);
+    mocked(isFeatureFlagEnabled).mockReturnValue(true);
```
jonahkagan, line 142: This testing approach seems like it could be a bit unpredictable. E.g. if another feature flag is added that influences this component, this mock will secretly change its behavior.

Maybe we can have a test helper to mock feature flags. E.g.:

```ts
const featureFlags = mockFeatureFlags()

beforeEach(() => featureFlags.reset())

test(..., () => {
  featureFlags.enable(featureFlagName)
```

libs/utils/src/scripts/generate_env_file.ts
```diff
@@ -0,0 +1,102 @@
...
+  outputPath?: string;
+  isVxDev?: boolean;
+  help?: boolean;
+}
+
+const flagOptions: Map<string, boolean> = new Map();
+function question(flagName: string, enable: boolean) {
+  return new Promise<void>((resolve) => {
+    rl.question(
+      `Enable ${flagName}? (${enable ? 'Y/n' : 'y/N'}):`,
```
jonahkagan, line 23: Should we suggest a default value here and allow just using the default?

jonahkagan (approved):

frontends/precinct-scanner/src/screens/poll_worker_screen.test.tsx
carolinemodic: Yes I was thinking we should have some kind of test helper but couldn't quite figure out how it should work.

libs/utils/src/scripts/generate_env_file.ts
carolinemodic: There is a default, so it will either tell you to enter Y/n (which defaults to yes) or y/N (which defaults to no)

jonahkagan: Ah ok I didn't understand that's what the different capitalization meant

frontends/precinct-scanner/src/screens/poll_worker_screen.test.tsx
carolinemodic: https://github.com/votingworks/vxsuite/issues/2548

PR #2544, opened 2022-09-12
feat(ui): Select supports primary prop

beausmith:
image


beausmith:
These tests are passing locally. May have to "git clean -dfX"…

jonahkagan:
Could you add a screenshot of what this change looks like?

beausmith:
@jonahkagan - updated!

frontends/bas/src/App.css
```diff
@@ -31,6 +31,14 @@ legend {
   display: block;
 }
 
+select option {
+  background-color: #ffffff;
+  color: #000000;
+  &:disabled {
+    color: #999999;
+  }
+}
```
jonahkagan, line 40: What are these styles for?

jonahkagan (approved):

frontends/bas/src/App.css
beausmith: These styles ensure that when the options are displayed, the background is always white, the disabled options are grey, and the enabled options are black. 

When adding the primary style, the kiosk browser version displayed their background as green when the Select was primary. 🤦

jonahkagan: Yikes! Is there any reason to define them here vs in the Select component? Seems like splitting them up means you have to look in multiple places to understand the styles, but maybe you have a reason?

beausmith: Thanks a good point/question. I appreciate your questions. I guess I consider this a global reset that I want to apply everywhere a select is used. If a new Component is created which contains a <select> tag, I want that select tag to have these styles too. However, it's mostly the white background that I want to ensure is consistent across browsers. The definition of grey and black text colors were perhaps an overreach and can be moved to the Select component. I'll fix this. Please LMK if I can clarify anything further.

jonahkagan: That makes sense - I had a feeling this file was a "reset" file. Sounds good to me!

beausmith: I take it back; these do need to be "reset" css as if they are not specified, the `<option>` element styles inherit from the parent `<select>` element when they should not (see below). This reset can be easily overridden in a styled-component. TL;DR: I'm going to leave these styles in the reset css files for each app.
https://user-images.githubusercontent.com/28444/190001086-5e1b7853-d0cf-4ade-ad5c-ac966da376ae.mov

beausmith: Bummed that these videos don't display inline as they do in the above description. 🤷‍♂️

jonahkagan: Sounds good!

PR #2545, opened 2022-09-12
refactor(admin): store CVRs in `admin` backend

eventualbuddha:
## Overview

Adds endpoints to import CVR files and process them. Adds a join table for `cvr_files` and `cvrs` so that duplicate CVRs are linked to both files. Changes the frontend to use the new endpoints for adding and clearing files, but also preserves the existing storage for now to keep the tally code the same.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tested with VxAdmin in Chrome adding and removing CVR files, including duplicates etc. Haven't really tested thoroughly with kiosk browser and the import table.

## Checklist

### DIFF
--- a/frontends/election-manager/src/lib/backends/admin_backend.ts
+++ b/frontends/election-manager/src/lib/backends/admin_backend.ts
@@ -8,6 +8,9 @@ import {
 import { fetchJson } from '@votingworks/utils';
 import { Admin } from '@votingworks/api';
 import { ElectionManagerStoreStorageBackend } from './storage_backend';
+import { AddCastVoteRecordFileResult } from './types';
+
+const CVR_FILE_ATTACHMENT_NAME = 'cvrFile';
 
 /** @visibleForTesting */
 export const activeElectionIdStorageKey = 'activeElectionId';
@@ -87,6 +90,57 @@ export class ElectionManagerStoreAdminBackend extends ElectionManagerStoreStorag
     await this.storage.set(activeElectionIdStorageKey, parsedResponse.id);
 
     return parsedElectionDefinition;
+  }
+
+  async addCastVoteRecordFile(
+    newCastVoteRecordFile: File
+  ): Promise<AddCastVoteRecordFileResult> {
+    await super.addCastVoteRecordFile(newCastVoteRecordFile);
+
+    const activeElectionId = await this.loadActiveElectionId();
+
+    if (!activeElectionId) {
+      throw new Error('no election configured');
+    }
+
+    const formData = new FormData();
+    formData.append(CVR_FILE_ATTACHMENT_NAME, newCastVoteRecordFile);
+
+    const addCastVoteRecordFileResponse = (await fetchJson(
+      `/admin/elections/${activeElectionId}/cvr-files`,
+      {
+        method: 'POST',
+        body: formData,
+      }
+    )) as Admin.PostCvrFileResponse;
+
+    if (addCastVoteRecordFileResponse.status !== 'ok') {
+      throw new Error(
+        `could not add cast vote record file: ${addCastVoteRecordFileResponse.errors
+          .map((e) => e.message)
+          .join(', ')}`
+      );
+    }
+
+    return {
+      wasExistingFile: addCastVoteRecordFileResponse.wasExistingFile,
+      newlyAdded: addCastVoteRecordFileResponse.newlyAdded,
+      alreadyPresent: addCastVoteRecordFileResponse.alreadyPresent,
+    };
+  }
+
+  async clearCastVoteRecordFiles(): Promise<void> {
+    await super.clearCastVoteRecordFiles();
+
+    const activeElectionId = await this.loadActiveElectionId();
+
+    if (!activeElectionId) {
+      throw new Error('no election configured');
+    }
+
+    await fetchJson(`/admin/elections/${activeElectionId}/cvr-files`, {
+      method: 'DELETE',
+    });
   }
 
   async reset(): Promise<void> {
--- a/services/admin/schema.sql
+++ b/services/admin/schema.sql
@@ -7,37 +7,43 @@ create table elections (
 );
 
 create table adjudications (
-	id varchar(36) primary key,
-	contest_id text,
-	transcribed_value text,
-	cvr_id varchar(36),
-	foreign key (cvr_id) references cvrs(id)
-		on update cascade
-		on delete cascade
+  id varchar(36) primary key,
+  contest_id text not null,
+  transcribed_value text,
+  cvr_id varchar(36) not null,
+  foreign key (cvr_id) references cvrs(id)
+    on update cascade
+    on delete cascade
 );
 
 create table cvrs (
-	id varchar(36) primary key,
-	ballot_id varchar(36) unique,
-	imported_by_file varchar(36),
-	data text,
-	foreign key (imported_by_file) references cvr_files(id)
-	  on update cascade
-	  on delete cascade
+  id varchar(36) primary key,
+  election_id varchar(36) not null,
+  ballot_id varchar(36) not null,
+  data text not null,
+  created_at timestamp not null default current_timestamp,
+  updated_at timestamp not null default current_timestamp
 );
 
 create table cvr_files (
-	id varchar(36) primary key,
-	election_id varchar(36) not null,
-	signature text,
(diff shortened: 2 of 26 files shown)


frontends/election-manager/src/lib/backends/admin_backend.ts
```diff
@@ -89,6 +92,57 @@ export class ElectionManagerStoreAdminBackend extends ElectionManagerStoreStorag
     return parsedElectionDefinition;
   }
 
+  async addCastVoteRecordFile(
```
carolinemodic, line 95: Where are we determining which store/backend we're using?

carolinemodic (approved):

services/admin/src/store.ts
```diff
@@ -96,4 +132,361 @@ export class Store {
...
+   * Adds a CVR file entry record and returns its ID. If a CVR file entry with
+   * the same contents has already been added, returns the ID of that record and
+   * associates `cvrFileId` with it.
+   */
+  addCastVoteRecordFileEntry(
+    electionId: Id,
+    cvrFileId: Id,
+    data: string
+  ): Result<{ id: Id; isNew: boolean }, AddCastVoteRecordError> {
+    const cvr = safeParseJson(data).unsafeUnwrap() as CastVoteRecord;
```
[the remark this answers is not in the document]
eventualbuddha: Yeah, will do that in a follow up.

services/admin/schema.sql
```diff
@@ -7,37 +7,43 @@ create table elections (
...
-	id varchar(36) primary key,
-	contest_id text,
-	transcribed_value text,
-	cvr_id varchar(36),
-	foreign key (cvr_id) references cvrs(id)
-		on update cascade
-		on delete cascade
+  id varchar(36) primary key,
+  contest_id text not null,
+  transcribed_value text,
```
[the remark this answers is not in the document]
eventualbuddha: Yeah, something like that. I'll add that in the follow-up PR.

frontends/election-manager/src/lib/backends/admin_backend.ts
eventualbuddha: https://github.com/votingworks/vxsuite/blob/4478ce90d99388af32d63eb3c654e0fd6a9b4d79/frontends/election-manager/src/index.tsx#L16

carolinemodic: gotcha, I missed that the call to super here is how we are also storing it in the old location

eventualbuddha:
Added another commit which adds the API endpoints for most of the adjudication flow. See [this test](https://github.com/votingworks/vxsuite/blob/ad55d0b9c105c97f4479d024475faf33e9dd91f5/services/admin/src/server.writeins.test.ts#L20-L145) for a fairly complete example for their usage.

services/admin/src/store.ts
```diff
@@ -489,4 +536,276 @@ export class Store {
...
+        );
+
+        const writeInRecord: Admin.WriteInRecord = {
+          id: row.id,
+          status: adjudication
+            ? 'adjudicated'
+            : typeof row.transcribedValue === 'string'
+            ? 'transcribed'
+            : 'pending',
+          castVoteRecordId: row.castVoteRecordId,
```
carolinemodic, line 670: Do you think the UI should just grab the actual CVR to get the image data from the electron storage, or will we need another endpoint to get the CVR data for an ID (or just return that in this endpoint, but since you are getting all of the write in rows for a contest that seems like it would explode the response too much maybe idk)

services/admin/src/server.ts
```diff
@@ -140,6 +155,145 @@ export function buildApp({ store }: { store: Store }): Application {
...
+            type: 'ValidationError',
+            message: parseQueryResult.err().message,
+          },
+        ],
+      });
+      return;
+    }
+
+    const { contestId } = parseQueryResult.ok();
+    store.getWriteInRecords({ electionId });
```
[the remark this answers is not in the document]
eventualbuddha: Mistake. Will fix.

services/admin/src/store.ts
eventualbuddha: This API endpoint is mostly meant to be used with a `limit` query param, likely of 1. So it could include that data.

carolinemodic: I'm a little confused how the limit works actually. If there are 5 different CVRs with a write in for a given contest, you can filter down to those 5 with the contestId param, and then limit to 1 but don't we need a paging param to then be able to get the other write ins?

services/admin/src/store.ts
```diff
@@ -489,4 +536,276 @@ export class Store {
       );
     });
   }
+
+  /**
+   * Gets a summary of the write-in adjudication status.
+   */
+  getWriteInAdjudicationSummary({
```
carolinemodic, line 543: (for a future pr) it might be useful for reporting to get a summary that is just the adjudicated values & counts, without the level of detail of the transcribed values, but the frontend could also just figure that out from this API

carolinemodic (approved):
This is awesome!

services/admin/src/store.ts
eventualbuddha: You'd also use `status=pending`, so as long as you're transcribing then the head of that response will always be different.

services/admin/schema.sql
```diff
@@ -6,38 +6,62 @@ create table elections (
...
 create table cvrs (
-	id varchar(36) primary key,
-	ballot_id varchar(36) unique,
-	imported_by_file varchar(36),
-	data text,
-	foreign key (imported_by_file) references cvr_files(id)
-	  on update cascade
-	  on delete cascade
+  id varchar(36) primary key,
+  election_id varchar(36) not null,
```
benadida, line 39: Probably want foreign key on `election_id` and `on delete cascade`.

services/admin/src/store.ts
```diff
@@ -96,4 +135,677 @@ export class Store {
...
+  }): Result<
+    {
+      id: Id;
+      wasExistingFile: boolean;
+      newlyAdded: number;
+      alreadyPresent: number;
+    },
+    AddCastVoteRecordError
+  > {
+    this.assertElectionExists(electionId);
```
benadida, line 182: this is not needed as the foreign key constraint will fail the row insert if the referenced `election_id` doesn't exist.

services/admin/src/store.ts
```diff
@@ -96,4 +135,677 @@ export class Store {
...
+          a.transcribedValue === row.transcribedValue
+      );
+
+      return {
+        contestId: row.contestId,
+        transcribedValue: row.transcribedValue ?? undefined,
+        writeInCount: row.writeInCount,
+        writeInAdjudication: adjudication,
+      };
+    });
```
benadida, line 588: this could be done with an additional join in the database.

benadida (approved):
some suggestions, none of them required, easily done in a followup PR.

services/admin/schema.sql
```diff
@@ -6,38 +6,62 @@ create table elections (
...
-	foreign key (cvr_id) references cvrs(id)
-		on update cascade
-		on delete cascade
+create table write_in_adjudications (
+  id varchar(36) primary key,
+  election_id varchar(36) not null,
+  contest_id text not null,
+  transcribed_value text not null,
+  adjudicated_value text,
+  adjudicated_option_id text,
```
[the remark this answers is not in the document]
eventualbuddha: That's true if we're thinking we might change to mysql or postgresql in the future. For SQLite, though, they're the same:

> Note that numeric arguments in parentheses that following the type name (ex: "VARCHAR(255)") are ignored by SQLite - SQLite does not impose any length restrictions (other than the large global [SQLITE_MAX_LENGTH](https://www.sqlite.org/limits.html#max_length) limit) on the length of strings, BLOBs or numeric values.

For now I think I'll leave it as I'd rather not spend time coming up with suitable lengths only to have SQLite ignore them.

services/admin/schema.sql
```diff
@@ -6,38 +6,62 @@ create table elections (
...
+  created_at timestamp not null default current_timestamp,
+  foreign key (election_id) references elections(id)
+    on update cascade
```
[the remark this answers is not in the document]
eventualbuddha: Probably not. I can remove it.

PR #2547, opened 2022-09-12
feat(ui): Table has expanded style and TH component

beausmith:
image


jonahkagan:
Could you add a screenshot?

beausmith:
@jonahkagan - screenshot posted.

libs/ui/src/table.tsx
```diff
@@ -17,8 +18,12 @@ export const Table = styled.table<TableProps>`
...
   & td {
     border-bottom: 1px solid ${borderColor};
-    padding: ${({ condensed }) =>
-      condensed ? '0.15rem 0.25rem' : '0.25rem 0.5rem'};
+    padding: ${({ condensed, expanded }) =>
+      condensed
+        ? '0.15rem 0.25rem'
+        : expanded
+        ? '0.25rem 1rem'
+        : '0.25rem 0.5rem'};
```
jonahkagan, line 26: I'm not totally sure what the context for the expanded/condensed tables are, but it might be nice to keep the horizontal padding equal on all styles so that the text always lines up with other tables on the page, regardless of style. But this might not be a real concern, so feel free to ignore

jonahkagan (approved):

libs/ui/src/table.tsx
beausmith: Yeah, this was a contrived example for a screenshot. At present don't use different tables on the same screen… and I don't think we will. 🤞

PR #2552, opened 2022-09-13
build(node): install NodeJS from nodejs.org

eventualbuddha:
## Overview
There are differences among the 16.x versions of NodeJS that are relevant to our development. For example, `Intl.DateTimeFormat` yields slightly different results between 16.14.2 and 16.17.0. [We can't easily install a specific version from nodesource](https://github.com/nodesource/distributions/issues/1427#issuecomment-1207640371
), so instead this commit changes the `setup-dev` script to just download the right tarball and unpack it into `/usr/local`.

## Demo Video or Screenshot
n/a

## Testing Plan 
Ran it locally on an ARM VM, seemed to work:
```
❯ ./script/setup-dev 
[sudo] password for brian: 

❯ /usr/local/bin/node -v
v16.14.2
```

## Checklist


beausmith (approved):

eventualbuddha:
@benadida I know you're more of a fan of doing things through `apt` et al. What do you think?

benadida:
> @benadida I know you're more of a fan of doing things through `apt` et al. What do you think?

My two reasons for preferring `apt` are:
- easier to update and such
- digital signature check on downloads.

I think the first point is not important for our use case.

The second point *is* useful. It would be nice to rely on something extra on top of SSL.

Good news is, we're downloading a specific version of node. So, we could get the hash of the file we expect and stash that hash in this installer file (or in a side file). If we can't reliably use `apt` for what we're trying to do, then I would suggest we do a follow-on task to figure out how to add some end-to-end integrity check on the node download.

eventualbuddha:
> If we can't reliably use `apt` for what we're trying to do, then I would suggest we do a follow-on task to figure out how to add some end-to-end integrity check on the node download.

Okay, I added a `sha256sum --check`. @benadida?

benadida (approved):

PR #2557, opened 2022-09-14
feat(admin): add `useWriteInQuery`

eventualbuddha:
## Overview

This will be used to fetch write-ins for the transcription/adjudication UI.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated, since I haven't been able to get the "Write Ins" screen from `caro/write_in_tab_data` working properly.

## Checklist


frontends/election-manager/src/hooks/use_write_ins_query.ts
```diff
@@ -0,0 +1,37 @@
...
+  contestId,
+  status,
+  enabled = true,
+}: Props = {}): WriteInsQuery {
+  const { backend } = useContext(ServicesContext);
+
+  return useQuery(
+    getWriteInsQueryKey({ contestId, status }),
+    () => backend.loadWriteIns({ contestId, status }),
+    { enabled }
```
carolinemodic, line 35: what does enabled do?

frontends/election-manager/src/lib/backends/storage_backend.ts
```diff
@@ -426,4 +431,27 @@ export class ElectionManagerStoreStorageBackend
       return parsedElectionDefinition;
     }
   }
+
+  setWriteIns(writeIns?: readonly Admin.WriteInRecord[]): Promise<void> {
+    if (writeIns) {
```
carolinemodic, line 436: this code should never be run right? so why do we need to also implement this?

carolinemodic (approved):

frontends/election-manager/src/lib/backends/storage_backend.ts
eventualbuddha: Correct, but I wanted it for validation purposes. It'll be removed eventually.

frontends/election-manager/src/hooks/use_write_ins_query.ts
eventualbuddha: Determines whether the query is active and should fetch/refetch etc. Since hooks have to be called every time, this is a way of making the query conditional. See https://react-query-v3.tanstack.com/guides/dependent-queries.

frontends/election-manager/src/lib/backends/storage_backend.ts
carolinemodic: got it, definitely feels like a good reason to finish the refactor sooner rather then later as it feels like a lot of overhead to have to pointlessly implement everything here too. I just put in the function but threw a not implemented error to get around the validation issue in my branch.

eventualbuddha: To be clear, I meant _validation of the concept_, not type checking.

PR #2558, opened 2022-09-14
feat(admin): make write-in transcription work

eventualbuddha:
## Overview

This simplifies a few things in `WriteInScreen` and `WriteInTranscriptionScreen` around data management by removing the `useEffect` + `setSomeState` pattern. We can just compute things on the fly from the query data and we'll do something to make it faster if it's slow.

Responsibility for the current transcription "page" now belongs to `WriteInTranscriptionScreen`. In changing to react-query and `useWriteInsQuery` we reload the write-in data every time it changes. This isn't really ideal, but will probably work for now. As a result of that the `useEffect` that was trying to reset `paginationIdx` to 0 was being triggered on every transcription. Instead, we now let the transcription screen handle it and use `key=${contestId}` to ensure the new `offset` is reset in the event we somehow switch contests without unmounting the component first.

Transcription is now handled by a react-query mutation, wrapped in `useTranscribeWriteInMutation`. This ensures that the necessary data is reloaded when transcription happens.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/1938/190278778-99afbf6c-d82e-499a-a405-67dafa046cd7.mov

## Testing Plan 
Tested manually:
- [x] adding new transcription value
- [x] selecting an existing transcription value

## Checklist

### DIFF
--- a/frontends/election-manager/src/screens/write_ins_screen.tsx
+++ b/frontends/election-manager/src/screens/write_ins_screen.tsx
@@ -1,147 +1,69 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useEffect, useMemo, useState } from 'react';
 import styled from 'styled-components';
 
 import { Button, Modal, Prose, Table, TD } from '@votingworks/ui';
 import {
-  Adjudication,
   CandidateContest,
   ContestId,
   getPartyAbbreviationByPartyId,
 } from '@votingworks/types';
 
+import { collections, groupBy } from '@votingworks/utils';
+import { Admin } from '@votingworks/api';
 import { NavigationScreen } from '../components/navigation_screen';
 import { WriteInsTranscriptionScreen } from './write_ins_transcription_screen';
 import { AppContext } from '../contexts/app_context';
 import { WriteInsAdjudicationScreen } from './write_ins_adjudication_screen';
+import { useWriteInsQuery } from '../hooks/use_write_ins_query';
+import { useTranscribeWriteInMutation } from '../hooks/use_transcribe_write_in_mutation';
 
 const ContentWrapper = styled.div`
   display: inline-block;
 `;
 
-const initialTemporaryTranscriptions = [
-  {
-    id: 'd1',
-    contest_id: 'zoo-council-mammal',
-    transcribed_value: 'Jumpy',
-    adjudicated_value: '',
-    cvr_id: 'asdf',
-  },
-  {
-    id: 'd2',
-    contest_id: 'zoo-council-mammal',
-    transcribed_value: 'Elephant',
-    adjudicated_value: '',
-    cvr_id: 'asdf',
-  },
-  {
-    id: 'd3',
-    contest_id: 'zoo-council-mammal',
-    transcribed_value: 'Long Trunk',
-    adjudicated_value: '',
-    cvr_id: 'asdf',
-  },
-  {
-    id: 'd4',
-    contest_id: 'zoo-council-mammal',
-    transcribed_value: 'Jumpy',
-    adjudicated_value: '',
-    cvr_id: 'asdf',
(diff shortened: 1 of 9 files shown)


frontends/election-manager/src/screens/write_ins_screen.tsx
```diff
@@ -1,147 +1,70 @@
...
+      (writeIns) =>
+        collections.map(
+          groupBy(writeIns, ({ status }) => status),
+          (group) => group.size
+        )
+    );
+    setWriteInCountsByContest(map);
+  }, [writeInsQuery.data]);
 
   // When we start adjudicating a new contest, reset pagination index and fetch adjudications
```
carolinemodic, line 56: comment is out of date

frontends/election-manager/src/screens/write_ins_screen.tsx
```diff
@@ -155,131 +78,17 @@ export function WriteInsScreen(): JSX.Element {
...
-        ) {
-          return {
-            ...t,
-            adjudicated_value: adjudicatedValue,
-          };
-        }
-        return t;
-      })
-    );
+    console.log('updateTranscriptions', transcribedValue, adjudicatedValue);
```
carolinemodic, line 91: remove

carolinemodic: oh actually I'm assuming this is just temp until you implement this part so I guess its fine

frontends/election-manager/src/screens/write_ins_screen.tsx
```diff
@@ -155,131 +78,17 @@ export function WriteInsScreen(): JSX.Element {
...
   }
 
-  const transcriptionContestIds = [
-    ...new Set(temporaryTranscriptions.map((t) => t.contest_id)),
-  ];
-  const contestsWithWritins = election.contests.filter((c) =>
-    transcriptionContestIds.includes(c.id)
-  );
-
-  function sortWithEmptyStringLast(a: string, b: string) {
```
carolinemodic, line 165: FYI I think there is some magic beau was doing in this code where if you adjudicate transcibeOptionA as transcibeOptionB it adjudicates BOTH of those not just the first one, he also might have been doing some sorting magic in here we will need to think about when implementing the wiring up of the adjudication stuff. Although I don't think any of that is mission critical.

frontends/election-manager/src/screens/write_ins_screen.tsx
```diff
@@ -1,147 +1,70 @@
...
   getPartyAbbreviationByPartyId,
 } from '@votingworks/types';
 
+import { collections, groupBy } from '@votingworks/utils';
+import { Admin } from '@votingworks/api';
 import { NavigationScreen } from '../components/navigation_screen';
 import { WriteInsTranscriptionScreen } from './write_ins_transcription_screen';
 import { AppContext } from '../contexts/app_context';
 import { WriteInsAdjudicationScreen } from './write_ins_adjudication_screen';
+import { useWriteInsQuery } from '../hooks/use_write_ins_query';
```
carolinemodic, line 17: love that we don't have to boilerplate feed all these things through the context now.

carolinemodic (approved):
I was also just refetching everything whenever we make a change, I think thats fine for now. We should test it once we have ballot images in there though, but that could just be a separate fetch happening from the transcriptions screen if its an issue.

frontends/election-manager/src/screens/write_ins_screen.tsx
eventualbuddha: Restating myself to ensure I understand: if you transcribe some things as “Bob” and some others as “Robert” and then you adjudicate “Bob” as “Robert”, then both “Bob” and “Robert” are adjudicated as “Robert”.

PR #2559, opened 2022-09-15
fix feature flagging in production builds

carolinemodic:
## Overview

See https://votingworks.slack.com/archives/CEL6D3GAD/p1663178111465069
For some reason in production builds, and specifically on real hardware not in VMs, trying to access indexed values on process.env was throwing type errors. On my surface go it was saying process was undefined and on my thinkpads it was saying env was undefined, but if I defined a variable on the env.d.ts file I could access it fine so this was the only way I could figure out how to fix it (I tried adding an arbitrary `[key: string]: string | undefined` but that didn't work). I'm annoyed that this adds 2 more locations where you have to add a bit of boilerplate when adding a feature flag so if people have suggestions on other things to try let me know.  

## Testing Plan 
Ran on VxDev things worked as expected 

## Checklist

### DIFF
--- a/libs/utils/src/env.d.ts
+++ b/libs/utils/src/env.d.ts
@@ -3,5 +3,9 @@ declare namespace NodeJS {
     NODE_ENV: 'development' | 'production' | 'test';
     REACT_APP_VX_DEV?: string;
     REACT_APP_VX_DISABLE_CARD_READER_CHECK?: string;
+    REACT_APP_VX_ENABLE_WRITE_IN_ADJUDICATION?: string;
+    REACT_APP_VX_ENABLE_ALL_ZERO_SMARTCARD_PIN_GENERATION?: string;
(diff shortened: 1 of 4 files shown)


benadida (approved):
No better suggestion right now.

eventualbuddha (approved):
As for why this happens, `vite.config.ts` only specifically replaces `process.env.X`, where `X` is a value defined in an `.env` file at the time of build. So `process.env[flag]` is left alone:

```js
// production build output
function isFeatureFlagEnabled(flag) {
  const flagInfo = getFlagDetails(flag);
  return (flagInfo.allowInProduction || false || isVxDev()) && asBoolean(process.env[flag]);
}
```

I suspect the difference with VM vs real hardware actually boils down to a configuration difference. And adding the explicit named properties for `ProcessEnv` is nice for development, but ignored by Rollup when building for production so it shouldn't matter for the final build output.

carolinemodic:
Interesting!

PR #2560, opened 2022-09-15
allow wia flag to be enabled in prod

carolinemodic:
We will need this for the images for the NH demo


benadida (approved):

eventualbuddha (approved):
Snapshots need updating.

PR #2561, opened 2022-09-15
feat(admin) display ballot images in wia

carolinemodic:
## Overview

Displays ballot images in transcription screen. The logic to figure out the crops for these is weird, I mostly just blindly copied what we were doing before but fixed the type errors in it. We should improve it and test it better in the long term but for now this is good enough. 

## Demo Video or Screenshot

https://user-images.githubusercontent.com/14897017/190517028-4dcf3188-2207-49f5-8c42-a9d0273b2471.mov

## Testing Plan 
Manual testing with NH election, see above

## Checklist


frontends/election-manager/src/lib/backends/admin_backend.ts
```diff
@@ -191,4 +191,25 @@ export class ElectionManagerStoreAdminBackend extends ElectionManagerStoreStorag
 
     return response;
   }
+
+  async loadWriteInImage(
+    writeInId: string
+  ): Promise<Admin.WriteInImageEntry[]> {
+    console.log('loading write in image');
```
eventualbuddha, line 198:
```suggestion
```

script/setup-dev
```diff
@@ -51,7 +51,7 @@ TMPDIR=$(mktemp -d -t node-tmp-XXXXXX)
   curl -sSLO ${DOWNLOAD_URL}
   set -x
   sha256sum --check "${DIR}/hashes/$(basename "${DOWNLOAD_URL}").sha256"
-  cat "${TMPFILE}" | sudo tar -xz -C /usr/local --strip-components=1
+  sudo tar -xz -C /usr/local --strip-components=1
```
eventualbuddha, line 54: The old version looks like it failed to define `TMPFILE`, and this version is not specifying a file or a pipe as input to `tar`. Also, unrelated to this PR.

services/admin/src/server.ts
```diff
@@ -293,6 +303,146 @@ export function buildApp({ store }: { store: Store }): Application {
     );
   });
 
+  /* istanbul ignore next */
+  app.get<
+    { electionId: Id; writeInId: Id },
+    Admin.GetWriteInImageResponse,
+    Admin.GetWriteInImageRequest
+  >(
+    '/admin/elections/:electionId/cvr-file/:writeInId/write-in-image',
```
eventualbuddha, line 312: Per @jonahkagan I haven't been nesting things like this. Also, it should be `cvrFileId`, right?

```suggestion
    '/admin/cvr-files/:cvrFileId/write-in-image',
```

services/admin/src/server.ts
```diff
@@ -293,6 +303,146 @@ export function buildApp({ store }: { store: Store }): Application {
...
+      try {
+        const ballotStyle = getBallotStyle({
+          ballotStyleId: cvr._ballotStyleId,
+          election,
+        });
+        if (
+          // eslint-disable-next-line no-underscore-dangle
+          cvr._layouts === undefined ||
+          // eslint-disable-next-line no-underscore-dangle
+          cvr._ballotImages === undefined
```
eventualbuddha, line 354: Add these exceptions here: https://github.com/votingworks/vxsuite/blob/81e01023da007a12525f76f5f95c8b272a3577ba/libs/eslint-plugin-vx/src/configs/recommended.ts#L110

services/admin/src/server.ts
```diff
@@ -293,6 +303,146 @@ export function buildApp({ store }: { store: Store }): Application {
...
+          contestIdx -= currentLayout.contests.length;
+        }
+        currentLayout = layouts[currentLayoutOptionIdx];
+        if (!currentLayout) {
+          throw new Error('unexpected types');
+        }
+        const contestLayout =
+          layouts[currentLayoutOptionIdx]?.contests[contestIdx];
+
+        // Options are laid out from the bottom up, so we reverse write-ins to get the correct bounds
```
eventualbuddha, line 396: This is only true for NH ballots. If we merge this, I think we need a comment or link to an issue on this endpoint saying what needs to change in the future.

eventualbuddha (approved):

script/setup-dev
carolinemodic: oh yeah this was erroring for me so I was playing with it and forgot to remove

services/admin/src/server.ts
carolinemodic: well I'm also using the election id as I need election information, but the other id is for the write in id, I should rename the endpoint though I made it cvr-files when I was feeding through the cvrId and then realized I needed the writeInId (as I need the contestId and optionId)

carolinemodic: oh I guess I could get the electionId from the writeInId too duh

services/admin/src/server.ts
carolinemodic: https://github.com/votingworks/vxsuite/issues/2564

PR #2562, opened 2022-09-15
feat(EM): Add report for write-in votes.

beausmith:
## Demo Video or Screenshot
https://user-images.githubusercontent.com/28444/190520075-e23ce780-c188-4a99-8944-8c2ad72b5e34.mov


frontends/election-manager/src/screens/tally_writein_report_screen.tsx
```diff
@@ -0,0 +1,351 @@
...
+      ['Fred Wilson', 4],
+      ['Charles Foo', 4],
+      ['Fanny May', 17],
+    ]),
+  ],
+  [
+    'State-Representatives-Hillsborough-District-34-b1012d38',
+    new Map([['Fanny May', 31]]),
+  ],
+]);
```
beausmith, line 95: Fake Data!

frontends/election-manager/src/components/election_manager_writein_tally_report.tsx
```diff
@@ -0,0 +1,202 @@
...
+    const precinctIds =
+      precinctIdFromProps === 'all'
+        ? election.precincts.map((p) => p.id)
+        : [precinctIdFromProps];
+
+    const writeInMetadataFooter = (
+      <React.Fragment>
+        <br />
+        <Text small as="span">
+          Note: Write-in votes adjudicated to official candidates are included
```
carolinemodic, line 71: I feel like it should either be "and not this report" or "and are not in this report"

frontends/election-manager/src/screens/tally_writein_report_screen.tsx
```diff
@@ -0,0 +1,351 @@
...
+      )})`;
+    }
+    return '';
+  }, [batchId, fullElectionTally, election]);
+  const reportDisplayTitle = useMemo(() => {
+    if (precinctName) {
+      return `${statusPrefix} Precinct Write-In Tally Report for ${precinctName}`;
+    }
+    if (scannerId) {
+      return `${statusPrefix} Scanner Write-In Tally Report for Scanner ${scannerId}`;
```
carolinemodic, line 209: We're not actually supporting this broken down in all of these ways right so we can probably remove some of this? (I know you just copied this)

frontends/election-manager/src/utils/write_ins.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return collections.map(writeInsByContestAndCandidate, (byCandidate) =>
+    collections.map(byCandidate, (entries) =>
+      collections.reduce(
+        entries,
+        (sum, entry) => sum + entry.writeInCount ?? 0,
+        0
+      )
+    )
+  );
+}
```
eventualbuddha, line 46: This file doesn't need to be a `.tsx` file. Just `.ts` will do.

frontends/election-manager/src/screens/tally_writein_report_screen.tsx
beausmith: Will we ever want this?

carolinemodic: yeah idk, I had started to remove it in my PR on top of this and then had the same thought, I'm not sure it really will make sense ever to break down in those ways? But we can leave for now since its simpler at this point to leave it haha

carolinemodic (approved):

frontends/election-manager/src/screens/tally_writein_report_screen.tsx
carolinemodic: I'm guessing we will want by party eventually but not the others

PR #2565, opened 2022-09-16
include official write ins in tally reports

carolinemodic:
## Overview
Note that I'm not seeing adjudicatedOptionId always set properly but assuming it is this will include those tallies in our regular tally reports and subtract them from the write-ins row. I don't love the way I did this, or the lack of testing, we should revisit this after the demo, and also make it a configurable thing since I imagine some jurisdictions won't want this. 

## Testing Plan 
Manually adjudicated as real candidates, saw the tally report adjust accordingly. 

## Checklist

### DIFF
--- a/frontends/election-manager/src/components/election_manager_tally_report.tsx
+++ b/frontends/election-manager/src/components/election_manager_tally_report.tsx
@@ -23,6 +23,7 @@ import { filterTalliesByParams, find } from '@votingworks/utils';
 import React, { forwardRef } from 'react';
 
 import { filterExternalTalliesByParams } from '../utils/external_tallies';
+import { modifyTallyWithWriteInInfo } from '../lib/votecounting';
 
 export type TallyReportType = 'Official' | 'Unofficial' | 'Test Deck';
 
@@ -38,6 +39,7 @@ export interface Props {
   precinctId?: string;
   scannerId?: string;
   votingMethod?: VotingMethod;
+  officialCandidateWriteIns?: Map<string, Map<string, number>>; // Contest -> Candidate ID -> Count
 }
 
 export const ElectionManagerTallyReport = forwardRef<HTMLDivElement, Props>(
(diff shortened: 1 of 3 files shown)


frontends/election-manager/src/lib/votecounting.ts
```diff
@@ -406,6 +409,57 @@ function processCastVoteRecord({
   return newCvr;
 }
 
+export function modifyTallyWithWriteInInfo(
+  tally: Tally,
+  writeIns: Map<string, Map<string, number>>
+): Tally {
+  console.log('start of call', writeIns);
```
benadida, line 416:
```suggestion
```

benadida (approved):

frontends/election-manager/src/components/election_manager_tally_report.tsx
```diff
@@ -38,6 +39,7 @@ export interface Props {
   precinctId?: string;
   scannerId?: string;
   votingMethod?: VotingMethod;
+  officialCandidateWriteIns?: Map<string, Map<string, number>>; // Contest -> Candidate ID -> Count
```
eventualbuddha, line 42:
```suggestion
  officialCandidateWriteIns?: Map<ContestId, Map<CandidateId, number>>;
```

frontends/election-manager/src/lib/votecounting.ts
```diff
@@ -406,6 +409,56 @@ function processCastVoteRecord({
   return newCvr;
 }
 
+export function modifyTallyWithWriteInInfo(
+  tally: Tally,
+  writeIns: Map<string, Map<string, number>>
```
eventualbuddha, line 414: Ditto here. Maybe give it a type alias?

frontends/election-manager/src/lib/votecounting.ts
```diff
@@ -406,6 +409,56 @@ function processCastVoteRecord({
   return newCvr;
 }
 
+export function modifyTallyWithWriteInInfo(
```
eventualbuddha, line 412: Would be nice to have a really basic test, at least.

eventualbuddha (approved):

PR #2566, opened 2022-09-16
fix(em): remove isLoading state for election def loading / conversion

benadida:
That state is not useful, and causing re-rendering challenges.


carolinemodic (approved):
yolo

PR #2568, opened 2022-09-16
fix(em): tweak language for ballot screen, and driver for displaying special instructions

benadida:


frontends/election-manager/src/screens/ballot_list_screen.tsx
```diff
@@ -69,18 +68,16 @@ export function BallotListScreen(): JSX.Element {
         <Header>
           <Prose>
             <p>This election uses custom ballots not produced by VxAdmin.</p>
-            <p>
-              The Ballot Package is still used to configure VxScan, the ballot
-              scanner.
-            </p>
+            <p>The Ballot Package is still used to configure VxScan.</p>
```
mattroe, line 71: nit: I would prefer just providing instructions here, rather than describing how the ballot package is still used as a user may not know how a ballot package is used in a different context.

suggest: `Save the Ballot Package to USB to configure VxCentralScan or VxScan.`

I'm throwing in VxCentralScan just to tease the concept as I want in in NH :)

mattroe (approved):

PR #2575, opened 2022-09-18
chore(lint): upgrade linting packages

eventualbuddha:
## Overview

Upgrades essentially all the linting tooling and fixes or disables any new lint violations that occurred as a result.

## Demo Video or Screenshot
n/a

## Testing Plan 
n/a

## Checklist


libs/eslint-plugin-vx/src/configs/react.ts
```diff
@@ -31,6 +31,9 @@ export = {
     ],
     'react/jsx-filename-extension': ['error', { extensions: allExtensions }],
     'react/jsx-fragments': ['error', 'element'],
+    'react/jsx-no-bind': 'off',
+    'react/jsx-no-constructed-context-values': 'off',
+    'react/jsx-no-useless-fragment': ['error', { allowExpressions: true }],
```
eventualbuddha, line 36: Allows e.g. `<React.Fragment>{foo}</React.Fragment>`.

libs/ui/src/election_info_bar.tsx
```diff
@@ -24,7 +24,7 @@ export function ElectionInfoBar({
   precinctSelection,
 }: Props): JSX.Element {
   if (!electionDefinition) {
-    return <React.Fragment />;
+    return <React.Fragment>{null}</React.Fragment>;
```
eventualbuddha, line 27: I don't love this, but `null` is not a `JSX.Element`, so we'd have to change all the `JSX.Element` return type annotations to be `React.ReactNode` instead.

libs/eslint-plugin-vx/src/configs/react.ts
jonahkagan: Does this mean we'll no longer get warnings when we have an unnecessary `React.Fragment`?

libs/ui/src/election_info_bar.tsx
jonahkagan: Can we just change the annotations as needed?

If I encountered this particular line of code I'd probably be confused.

jonahkagan (approved):

libs/eslint-plugin-vx/src/configs/react.ts
eventualbuddha: It's an auto-fixable rule in the case where the fragment has exactly one element child; that's what fixed most of the violations in this commit. Enabling `allowExpressions: true` does mean that, in the event the `<React.Fragment>` is not needed when used with an expression instead of an element, the rule will allow it when it arguably shouldn't. I haven't seen any cases of this in vxsuite, though. We _do_ have some cases of using `<React.Fragment>` with expressions, which is why I enabled this exception.

jonahkagan: Ah gotcha, so that's for computing dynamic fragments? I see how that could be useful

libs/ui/src/election_info_bar.tsx
eventualbuddha: Actually, I tried a few different return type values and none of them made TypeScript happy:

- `React.ReactChild`: cannot return `null`, nor can the component be used as a JSX component.
- `React.ReactNode`: can return `null`, but cannot use the component as JSX component.
- `React.ReactElement`: cannot return `null`, but the component can be used as a JSX component.

I introduced a `Nothing` component in a new commit that should make this less confusing, anyway.

jonahkagan: Yeah, that's been my experience as well. I thought maybe I was wrong and `React.ReactNode` was working because you were using it here.

I do think returning `null` is idiomatic for React, so maybe we just make a new type? Something like: `type ReactComponent = React.ReactElement | null`? Or `type ReactComponent = ReturnType<React.FunctionalComponent>`?

PR #2580, opened 2022-09-20
Better converter fix

benadida:
The previous fix caused a UX lag when converting that could be confusing. The `isLoading` state is still useful. So brought it back, and moved the state transitions to strictly when they're needed, so it doesn't happen on every load.

There's likely still some further refactoring we could do here, just to simplify the `updateStatus` approach. Leaving that to later work.

### DIFF
--- a/frontends/election-manager/src/screens/unconfigured_screen.tsx
+++ b/frontends/election-manager/src/screens/unconfigured_screen.tsx
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 
 import { safeParseElection } from '@votingworks/types';
 import { electionFamousNames2021Fixtures } from '@votingworks/fixtures';
-import { Button, Modal, Prose, useMountedState } from '@votingworks/ui';
+import { Button, Prose, useMountedState } from '@votingworks/ui';
 import { assert } from '@votingworks/utils';
 
 import {
@@ -61,7 +61,6 @@ export function UnconfiguredScreen(): JSX.Element {
   const { converter, saveElection } = useContext(AppContext);
 
   const [isUploading, setIsUploading] = useState(false);
-  const [showSuccess, setShowSuccess] = useState(false);
 
   const [inputConversionFiles, setInputConversionFiles] = useState<VxFile[]>(
     []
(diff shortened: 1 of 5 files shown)


frontends/election-manager/src/screens/unconfigured_screen.tsx
```diff
@@ -79,16 +79,12 @@ export function UnconfiguredScreen(): JSX.Element {
     history.push(routerPaths.electionDefinition);
   }
 
-  const saveElectionAndShowSuccess = useCallback(
-    (electionJson: string) => {
+  const saveElectionSafely = useCallback(
```
eventualbuddha, line 82: Not a blocker but… why "safely"?

eventualbuddha (approved):

frontends/election-manager/src/screens/unconfigured_screen.tsx
benadida: I can rename it if that's not obvious, I meant because it validates the data model.

eventualbuddha: Ah. Just a little weird since it then calls `unsafeUnwrap`.
is it safe

benadida: oooh, I think I misread that! So that's ... not validating the schema? Something else is?

eventualbuddha: See, it is confusing! It is validating the data and returning a `Result`, which we then `unsafeUnwrap` which will throw if the validation failed. The "safe" part refers to the fact that the function doesn't `throw`.

benadida: ahah! OK so I'm doing it right but maybe using the word `safe` is just confusing here and I'll use something clearer, like `validateAndSaveElection`

frontends/election-manager/src/screens/unconfigured_screen.tsx
```diff
@@ -209,29 +214,14 @@ export function UnconfiguredScreen(): JSX.Element {
     void updateStatus();
   }, [updateStatus]);
 
-  if (isUploading) {
+  if (isUploading || isLoading) {
     return (
       <NavigationScreen>
```
beausmith, line 219: Add the `centerContent` prop here to center the “Loading…” text in the content area.

beausmith (approved):
One visual nit! Lgtm.

PR #2581, opened 2022-09-22
feat(admin): split write-in adjudication options into groups

eventualbuddha:
## Overview

When adjudicating, the table now shows the options in groups:
- Official Candidates
- Write-In Candidates (from previous adjudications, if any)
- Original Transcription (lists the transcription as-is)

Additionally, instead of using the `write-in-summary` endpoint and doing a bunch of work to transform the data on the client, this commit introduces the `write-in-adjudication-table` endpoint that returns exactly the data needed for the table to render and perform the mutations it can do. 

This involved a decent amount of refactoring as well, so some of the logic for manipulating the data into the right format for the frontend have moved into a "view" within the `/api` library.

## Demo Video or Screenshot
https://user-images.githubusercontent.com/1938/191811992-1ad8b32a-e4d1-4af2-a133-7160206441de.mov

## Testing Plan 
Extensively tested the adjudication screen manually with a single test election.

## Checklist


services/admin/src/server.ts
```diff
@@ -379,12 +380,58 @@ export function buildApp({ store }: { store: Store }): Application {
...
+      })
     );
   });
 
+  app.get<
+    Admin.GetWriteInAdjudicationTableUrlParams,
+    Admin.GetWriteInAdjudicationTableResponse,
+    Admin.GetWriteInAdjudicationTableRequest
+  >(
+    '/admin/elections/:electionId/contests/:contestId/write-in-adjudication-table',
```
carolinemodic, line 398: I thought we were trying to avoid nesting params?

frontends/election-manager/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -49,96 +43,45 @@ export interface Props {
...
-    }
-  }
-
-  const adjudicationQueuePhrase = `${format.count(
-    pendingAdjudications.length
-  )} ${pluralize('transcriptions', pendingAdjudications.length)} to adjudicate`;
+  const writeInAdjudicationTableQuery = useWriteInAdjudicationTableQuery({
+    contestId: contest.id,
+  });
+  const { isLoading } = writeInAdjudicationTableQuery;
```
carolinemodic, line 53: Should this generally be how we handling loading with react-query? I think I have not been doing this.

carolinemodic (approved):

services/admin/src/server.ts
eventualbuddha: Where it's possible to infer the election ID I haven't been including it in the URL, yeah. Here both the election ID and contest ID must be specified, so I put them both in the URL.

frontends/election-manager/src/screens/write_ins_adjudication_screen.tsx
eventualbuddha: I'm unsure. Mostly I've just cared about whether `data` is loaded or not, but in this case I wanted a Boolean value so I pulled `isLoading` since it was already there.

PR #2583, opened 2022-09-23
fix(admin): use non-array index key

### LINKED ISSUE #1864: Avoid using array index as key in test deck ballot generation
HMPB test deck ballots: https://github.com/votingworks/vxsuite/blob/499fc20eb2f2de3c419a6350a73d253bdc58b89e/frontends/election-manager/src/screens/print_test_deck_screen.tsx#L163

BMD test deck ballots: https://github.com/votingworks/vxsuite/blob/499fc20eb2f2de3c419a6350a73d253bdc58b89e/frontends/election-manager/src/screens/print_test_deck_screen.tsx#L104

Note that we can't just use something like `ballot-${ballot.precinctId}-${ballot.ballotStyleId}` because, in a test deck, there are multiple ballots for each precinct + ballot style combo.

eventualbuddha:
## Overview

The pair of `precinctId` and `ballotStyleId` should be unique per ballot. Closes #1864.

## Demo Video or Screenshot
n/a

## Testing Plan 
n/a

## Checklist

### DIFF
--- a/frontends/election-manager/src/screens/print_test_deck_screen.tsx
+++ b/frontends/election-manager/src/screens/print_test_deck_screen.tsx
@@ -96,12 +96,12 @@ function BmdPaperBallots({
 
   return (
     <div className="print-only">
-      {ballots.map((ballot, i) => (
+      {ballots.map((ballot) => (
         <BmdPaperBallot
           ballotStyleId={ballot.ballotStyleId}
           electionDefinition={electionDefinition}
           isLiveMode={false}
-          key={`ballot-${i}`} // eslint-disable-line react/no-array-index-key
+          key={`ballot-${ballot.precinctId}-${ballot.ballotStyleId}`}
           precinctId={ballot.precinctId}
           votes={ballot.votes}
           onRendered={onRendered}


benadida (approved):

benadida:
arg @eventualbuddha I was a little too fast on this – are we sure this is safe? In the test deck we print multiple ballots per precinct and ballot style, since it's a test deck. I'm thinking this might be unsafe.

eventualbuddha:
I tested it, though I was not totally scientific about what I was looking for. I didn't see any warnings about duplicate keys, though.

arsalansufi:
I did some testing and was able to repro an error about duplicate keys:

errors

So agree with the call to revert this.

Also worth noting that this PR tweaks the keys for BMD ballots but leaves HMPB ballots as is:

https://github.com/votingworks/vxsuite/blob/9f451870504d19509ba661a8fa1a8d95f49b6e5b/frontends/election-manager/src/screens/print_test_deck_screen.tsx#L163

I'll re-open the original issue (https://github.com/votingworks/vxsuite/issues/1864) and add some notes to it accordingly!

PR #2601, opened 2022-09-27
fix bug with tally report not loading adjudications on first load

carolinemodic:
## Overview

When the tally report screen loads it calls react query for the write in summary information. It immediately calls it twice, the first set of results is stale and writeInSummaryQuery.isStale is true, long term the better fix here is probably to be using that variable and show a loading state or something while isState is true, but its immediately re-called and then given new data. However since the counts are not in this useEffect that shows the preview on the screen it is not redrawn with the new data when it gets loaded, like the write in tally report does, so the new data never appears. This was the most minimal change to fix that but I defer to @eventualbuddha on if we should be using isStale or doing something else long term as the better solution here. 

## Testing Plan 
Adjudicated write-ins immediately loaded tally report and saw the proper tallies. 

## Checklist


benadida (approved):

eventualbuddha:
Seems like the kind of thing we should do with a callback rather than a `useEffect` that _might_ fire at the right time.

PR #2617, opened 2022-09-29
Print Additional Polls Reports

### LINKED ISSUE #1497: VxScan / VxMark: ability to print more copies of polls-open / polls-close report
This might be necessary for any number of reasons, e.g. hard printer failure like paper jam, etc. Or just another copy is desired.

Once we allow this, we need to think about the difference between when a report was *generated* and when it was *printed*

### LINKED ISSUE #2604: VxScan: print additional reports
NH feedback that they need more tally reports at end of night from the tabulator. Manual says to print 4 accuvote reports, but we should likely add the ability to print a non-fixed number like VxAdmin.

adghayes:
## Overview
Closes #1497. Closes #2604. Those are duplicates.

Adds feature to both VxMark and VxScan to print additional polls open or close reports, one at a time. I spun up the simplest UX that made sense to me, please let me know if/how to edit.

### VxMark Flow

[vxmark_additional_report_flow.webm](https://user-images.githubusercontent.com/37960853/193161402-6a048e16-0e2e-44a4-81d2-452d7ac61524.webm)

### VxScan Screenshots

polls_open_add

polls_closed_add

printing_report

### Report Changes

In order to not get confused about report purposes, I removed distinctions between reports for posting vs. official filing:

[current_polls_closed.pdf](https://github.com/votingworks/vxsuite/files/9679672/current_polls_closed.pdf)

[new_polls_opened.pdf](https://github.com/votingworks/vxsuite/files/9679674/new_polls_opened.pdf)

## Testing Plan 
- Updated Existing RTL Tests
- Manual Testing

Have not added logging since the poll worker flow currently has no logging at all and that is a larger project I didn't want to jump into. 

## Checklist


frontends/bmd/src/pages/poll_worker_screen.tsx
```diff
@@ -172,65 +164,123 @@ function PrecinctScannerTallyReportModal({
...
-          }
-        />
-      )}
+      {/* 
+        This modal is torn down in some way which react-modal doesn't expect
+        and as a result, can leave the app hidden (in terms of aria). 
+        https://votingworks.slack.com/archives/CEL6D3GAD/p1633723756081200 
+      */}
+      <Modal
+        ariaHideApp={false}
```
adghayes, line 280: My changes to this page somehow reintroduced the issue where the accessibility controller (or the arrow keys) would not longer work to navigate on VxMark after this modal was opened.

This `ariaHideApp={false}` was added as a result of the original issue (https://votingworks.slack.com/archives/CEL6D3GAD/p1633723756081200), but at some point was removed. I tried experimenting with different permutations of the modal - but based on @eventualbuddha's comments in that thread, it may be that we're rarely closing modals gracefully and in this unique case react-modal can't handle it? In any case, this fix works and we are not prioritizing accessibility on the poll worker screens right now.

Props to better RTL practices, like using `userEvent` and `getByRole`, which caught this in automated testing!

libs/types/package.json
```diff
@@ -44,6 +44,7 @@
   },
   "devDependencies": {
     "@types/jest": "^27.0.3",
+    "@types/kiosk-browser": "workspace:*",
```
adghayes, line 47: Let me know if there is a problem with `@types/kiosk-browser` being a dependency for `libs/types`. I think it makes a little more sense for our hardware types to live in `libs/types` too, not `libs/utils`.

adghayes (commented):
I suggest reviewing this commit-by-commit.

adghayes:
@arsalansufi, adding you as an approver because you are working tomorrow :grin:

frontends/bmd/src/pages/poll_worker_screen.tsx
adghayes: cc @benadida

libs/types/package.json
benadida: I don't see an issue with this. @arsalansufi @eventualbuddha do you?

benadida (approved):

libs/types/package.json
arsalansufi: I also don't see any issue with this!

> I think it makes a little more sense for our hardware types to live in libs/types too, not libs/utils.

Also agree with this :100:

frontends/bmd/src/pages/poll_worker_screen.tsx
adghayes: @benadida sounds good, I created a ticket to track (#2618) and link to that instead.

@arsalansufi, I'm not sure that's an acceptable comment format for JSX? I tried it and the linter didn't like it and the code wouldn't run. I added an extra line to specify we're talking about that prop though.

arsalansufi: Ah! You probably have to use the standard backslashes when commenting a prop:

```jsx
<Modal
  // This modal is torn down in a way that react-modal doesn't expect and, as
  // a result, can leave the app hidden (in terms of accessibility). We set
  // this prop to false to disable the app hiding altogether
  ariaHideApp={false}
```

Issue #1836, opened 2022-05-17
Single sheet polls/open closed reports


benadida:
A few things about our polls open/close report:
- there's a lot of stuff printing: signature page, tally sheet, VxQR QR code.
- in primaries, we print separate tally sheets, but not separate signature pages.

Let's solve both of these problems at once by aiming for a polls open report that is, in most cases, just one sheet. In the case of primaries, one sheet per party. The polls open report should include:

- a top banner with the title ("Polls Open"), machine information, compact space for signatures, and ideally a VxQR code in the upper right-hand corner or the like
- the rest of the sheet shows the tally, much like today, only lower by a couple of inches to leave room for the top banner.

Thus, in a primary, a polls open report would be 4 sheets:
- 1 sheet for R, 1 sheet for D --> to be posted on the wall
- 1 sheet for R, 1 sheet for D --> to be filed for records.

mattroe:
For New Hampshire specifically, the main requirement is to have the signatures on the tally sheet itself.

adghayes:
@mattroe I’m thinking about how to consolidate the precinct reports, a few design questions:
- Should / can we drop the seal? Nice to have but would free up space for contests.
- Can we drop the “Unofficial” vs. “Official”? I don’t understand why we have this on our top-level signature page at all, as precinct results can’t be “Official” in the certified sense.
- Signatures at top or bottom of page? Signature at bottom makes more sense, but in the case where tally reports span multiple pages then the print formatting becomes more complicated (signature on every page or only on last page?) cc @benadida 

Other things to move, not controversial:
- Signature Space, somewhere
- TEST vs. not test
- Machine ID
- Report saved vs. report create times (do both if different)

adghayes:
If we want to put signatures on the bottom of the page but, in the case of multi-page reports, only on the last page, that gets us into PagedJS territory. Which is definitely out of scope for this change. So the options are:

- Signature at top of page
- Signatures at bottom of page, but will appear on every page for multi-page reports
- 
This is a bit of an edge case because I don’t think we’re expecting multi-page reports for the closest elections.

adghayes:
@benadida is the VxQR code going to be the same across all pages of the report if there are multiple pages?

adghayes:
@mattroe are we comfortable dropping the seal?

mattroe:
@adghayes yes, we can drop the seal -- it is not a hard requirement.

[...]

PR #2621, opened 2022-10-01
Consolidate Signature Page and Polls Tally Report

adghayes:
## Overview
Completes most of #1836, but does not add the VxQR code to the single-page report. Consolidates our polls report header page (with signature space) and the tally report. 

### New Report Look

[conslidated-report-rev1.pdf](https://github.com/votingworks/vxsuite/files/9690826/conslidated-report-rev1.pdf)

### Old Report For Reference

[unconsolidated-report-main.pdf](https://github.com/votingworks/vxsuite/files/9690827/unconsolidated-report-main.pdf)

## Testing Plan
- Manual Testing
- Updated RTL Tests

## Checklist


adghayes:
@mattroe we discussed putting the signatures at the bottom of each page but... I didn't realize that wasn't so simple. For each report (not each page), we can easily put the signature section above or below the tally results. I think top is better because on the bottom there's a possibility that the signature section ends up orphaned on the next page. 

It makes the language a little weird, "the undersigned" is no longer under anything. I'm not sure if that language is in state law, or just copying Accuvote. 

Now looking just for layout feedback.

adghayes:
To whoever reviews - I did a little change before the report consolidation that is actually more closely related to #2617. Because we can now print additional reports, the time printed might diverge from the time polls were closed. So I made sure we can specify both separately (and include both in the new report). I also renamed or separated cases of `timeSaved` to `timePollsToggled` because that is what matters and is more relevant on the report IMO.

benadida (approved):
suggest squashing snapshot and test updates into main commit so we don't have a broken commit in the main branch.

adghayes:
@benadida, sounds good. Currently, "Squash and merge" is the only merge option we have into main. "Rebase and merge" is technically an option but doesn't work with commit signing.

benadida:
> @benadida, sounds good. Currently, "Squash and merge" is the only merge option we have into main. "Rebase and merge" is technically an option but doesn't work with commit signing.

Does squash and merge even work? I think you have to squash them on your machine first.

adghayes:
cc @mattroe, to peek the reports

mattroe:
Just printed out to review a physical copy. My thoughts:
- The smallest font size on this page is quite hard to read in print. This applies to the timestamp, cert signature text, section, and cast/overvote/undervote. I think we should increase this a bit.
- The scanner ID could use it's own line in larger font. It gets lost in this.
- Signature area should have a little bit more padding. A small signature easily goes over the other text area.

adghayes:
@mattroe, I don't really want to revamp the pre-existing formatting of the report, but agree the signature area could use more clarity. Here's my suggestion:

**Change**:
- Cert Signature Text --> Larger Font
- Signature Space --> More Padding

**Don't Change**:
- Other small fonts

Question - what is the user story where the Scanner ID is important to whoever is reading the report, in which case it will be important to be more prominent? I made it small because I think it won't matter in most cases when reading this report.

adghayes:
I am interested in keeping the report compact to avoid multi-page reports as much as possible.

mattroe:
To be clear, this first version meets the core requirements for NH. Feel free to just take a quick pass at the certification signature section and move on. I have some additional thoughts on that below.

Re: font size, we can also just follow-up on overall design improvements for the report later. I was simply recognizing how small some of the fonts are on this page when reviewing the new content.

Re: scanner ID, thanks for pushing, you're most likely correct that this field will not be used often other than when multiple tabulators are used in a given precinct. My thought was that it would not surrender too much vertical real estate to move from the timestamp line.

Re: certification copy adjustment, you may want to see what it looks like to style "Certification Signatures" in bold and the sentence in normal text like you did for the scanner ID. I recognize the copy was in bold on the original report style, but I think we should keep the key (bold): value (unstyled) font design consistent in the report header.

adghayes:
@mattroe does the signature section on this look a little better? If so, will merge and move on.

[that-new-new-report.pdf](https://github.com/votingworks/vxsuite/files/9701377/that-new-new-report.pdf)

mattroe:
Yep, move on!

PR #2624, opened 2022-10-03
feat(em): breakout manual results in reports

### LINKED ISSUE #2615: VxAdmin: tally reports need to distinguish between manually entered and scanned counts
slack discussion: https://votingworks.slack.com/archives/CEL6D3GAD/p1664403030207379

adghayes:
## Overview
Closes #2615. Although we discussed breaking out all external results (with manual data being only a type of external results), it made sense to me from a product perspective to do just manual data and I found it wasn't much more difficult.

### Regular

Screen Shot 2022-10-03 at 2 33 42 PM

### With Manual Results Broken Out

Screen Shot 2022-10-03 at 2 34 13 PM

## Testing Plan
- Rewrote testing for the `contest_tally` component
- Added some assertions in integration tests to test for the manual data notice on top of the reports
- Manual testing, as always

## Checklist


adghayes:
cc @carolinemodic for code review

cc @carolinemodic @benadida @mattroe @beausmith for input on how exactly to format this in reports, if the above is not satisfactory.

adghayes:
@carolinemodic I rename a number of props to make the code more understandable to myself, but please let me know if I've made anything more confusing.

frontends/election-manager/src/components/election_manager_tally_report.tsx
```diff
@@ -138,15 +162,17 @@ export const ElectionManagerTallyReport = forwardRef<HTMLDivElement, Props>(
                       election={election}
                     />
                   </Prose>
+                  {manualTallyForReport && <ManualResultsNotice />}
```
carolinemodic, line 165: Personally I think it might be a little bit clearer to just have a flag here to figure out if there is a manualTally to determine whether or not to show the notice and then still pass all the externalTallies into ContestTally and have ContestTally do the logic of splitting out the manual tally inside of it, so that the caller doesn't always have to do that gymnastics, but I don't feel that strongly if you think its easier this way.

carolinemodic (approved):
Code lgtm. I defer to the others on the design, personally I think having the parenthesis after the main results makes more sense (so 1150 (5) instead of (5) 1150 ) but whatever Roe and Beau think is best.

mattroe:
@adghayes and I just discussed this synchronously. The main gap in this pass is we are missing the ability to easily see the machine-counted tally for any given selection, which is going to be the primary value compared to during a hand tally audit after the election. @adghayes is going to take a second pass to explicitly distinguish between 1) manual count 2) machine count and 3) total count.

frontends/election-manager/src/components/election_manager_tally_report.tsx
adghayes: I'm remembering now - I did it this way because the tally reports are already doing the gymnastics of filtering the results for the contest (which seems fine to me). So at the report level, we're filtering the `FullExternalExternalTallies` into `ExternalTallies`. `ExternalTallies` don't carry the source information with them, so I couldn't filter on those to identify the manual ones. I could edit the `ExternalTally` type to include the source. Or leave it as-is. Thoughts?

adghayes:
cc @mattroe, let me know how this looks!

### Before Manual Data
Screen Shot 2022-10-04 at 9 08 10 AM

### With Manual Data
Screen Shot 2022-10-04 at 9 07 56 AM

adghayes:
In playing around with it, I felt that:
- Vertical bars work best in the contests
- Dots work better for the metadata
- Just letting the metadata wrap seems a little less obnoxious to me then essentially making a mini-table on top of the main table, which I tried and looked funny

mattroe:
@adghayes has all the info needed for easy hand tally comparison so we're good on the main requirement. As discussed, the main goal is to support NH and we'll likely do a larger design overhaul of this report in the future where we can improve this presentation / make presentation more robust for larger vote counts / etc.

A few thoughts:
1. I would prefer consistency between dots and vertical lines as delimiters, but defer to your judgment from playing around with it.
2. Question: in the current implementation, if there is manually entered data for one contest, but not all, is the manually entered data shown as `0` or not at all?
3. One small copy nit suggestion: `*This report contains manually entered results. Vote counts are ordered as: scanned | manual | total `

adghayes:
1. I'll take another look
2. Not shown at all. Current setup will never show a manual count of `0`. Can stick with that (option A) or alter that in one of two ways - B) if a contest has any manual votes, show the breakdown for every number in that contest or C) if the report has any manual votes, show the breakdown for every number on the report. 
3. Update made

adghayes:
@mattroe and I discussed and decided that if a report contains any manual data, then the whole thing will be formatted in manual data mode. So reports without manual data will look exactly the same, and reports with manual data will look significantly different.

Here's my latest pass:

Screen Shot 2022-10-04 at 9 08 10 AM

adghayes:
Here's where we've landed!

Screen Shot 2022-10-04 at 5 57 01 PM
Screen Shot 2022-10-04 at 5 57 09 PM

[manual-reporting-latest.pdf](https://github.com/votingworks/vxsuite/files/9711778/manual-reporting-latest.pdf)

Re-requesting review from @carolinemodic because a good deal has changed.

carolinemodic (approved):
code lgtm, lets merge this and then when @beausmith is back see if he wants to take a design pass

PR #2632, opened 2022-10-05
use concurrently to spin up election-manager easily

adghayes:
Try out using [`concurrently` ](https://www.npmjs.com/package/concurrently)to spin up our dev environments more easily and not open 5 terminals to run one app. Changes:

- `services/admin` - `pnpm dev` now also watches for file changes and does hot-reloading
- `frontends/election-manager` - `pnpm start` no also watches for file changes in `libs`. `pnpm start:all`, the exciting part of this PR, spins up the `smartcard` service, `converter-sems`, and the admin service all at once, in a single terminal.

I'm hoping to give people a chance to try this and then, if it is useful, to apply the same to the other apps!


adghayes:
@eventualbuddha I added in two things since that other PR:
- incorporating the `watch` in the admin service
- spinning up `sems` as well. can edit this later if necessary

eventualbuddha (approved):

PR #2633, opened 2022-10-05
feat(BMD): Design pass of open/close polls report header

beausmith:
A design/cleanup pass to @adghayes' PR: https://github.com/votingworks/vxsuite/pull/2621

## Before
image

## After
image


frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
```diff
@@ -257,6 +257,7 @@ export function PollWorkerScreen({
     };
 
     assert(isPollWorkerAuth(auth));
+    console.log({ fullTallyInformation });
```
adghayes, line 260: Wayward `console.log`

libs/ui/src/precinct_scanner_tally_report.test.tsx
```diff
@@ -56,9 +56,17 @@ test('renders as expected for all precincts in a general election', () => {
...
+  expect(electionTitle.parentElement).toHaveTextContent(
+    'General Election: Tuesday, November 3, 2020, Franklin County, State of Hamilton'
+  );
+  const eventDate = screen.getByText('Polls Closed:');
+  expect(eventDate.parentNode).toHaveTextContent(
+    'Polls Closed: Sep 19, 2021, 11:05 AM'
+  );
+  const printedAt = screen.getByText('Report Printed:');
+  expect(printedAt.parentElement).toHaveTextContent(
+    'Report Printed: Sep 19, 2021, 11:06 AM'
```
adghayes, line 69: I realize I forgot to add test assertions for the `Scanner ID:` section, which should be added for consistency.

libs/ui/src/precinct_scanner_tally_report.tsx
```diff
@@ -21,68 +21,67 @@ import {
...
-  width: 30%;
-  padding-bottom: 1px;
-  &::before {
-    font-family: 'Noto Emoji', sans-serif;
-    font-size: 1.5em;
-    content: '⨉';
+const Signatures = styled.div`
+  display: flex;
+  & > div {
+    flex: 1;
```
adghayes, line 53: This is neat, I didn't know you could do this.

libs/ui/src/precinct_scanner_tally_report.tsx
```diff
@@ -135,22 +134,28 @@ export function PrecinctScannerTallyReport({
           <LogoMark />
           <Header>
             <h1>{reportTitle}</h1>
-            <h2>{electionTitle}</h2>
             <p>
-              {electionDate}, {election.county.name}, {election.state}
+              <strong>{electionTitle}:</strong> {electionDate},{' '}
+              {election.county.name}, {election.state}
+            </p>
```
adghayes, line 140: In cases where electionTitle includes the party name and is longer, I think this line could wrap. But in my opinion I don't think that's worth making any changes.

adghayes (commented):
I think the look is cleaner and much improved. I noticed a `console.log` and a test assertion we're missing, but otherwise LGTM.

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
adghayes: I wonder why we're not linting for this anymore. @eventualbuddha any idea?

libs/ui/src/precinct_scanner_tally_report.test.tsx
beausmith: Done.

libs/ui/src/precinct_scanner_tally_report.tsx
beausmith: Cool. Glad that I could share something new. 

Note: One benefit of this implementation is that we can easily change the number of signature lines by simply adding/removing child div elements. We could have a election config value for `pollsReportCertificationSignatures` where `0` disables the feature, and a value N adds N signature lines.

libs/ui/src/precinct_scanner_tally_report.tsx
beausmith: Yes, wrapping here is fine. It will look better if it doesn't wrap, but it will still communicate what it needs to if the content does wrap.

adghayes (approved):
LGTM!

Issue #2593, opened 2022-09-26
WIA: prevent editing after marking as official


mattroe:

carolinemodic:
After you mark results as official in VxAdmin you can no longer import CVRs, or change anything however you can still change things in the write in adjudication flow, we should change it to make it so that write in adjudication transcription and adjudication is disabled once results are marked as official.

PR #2638, opened 2022-10-06
WIA: disable adjudication after finalizing results

kofi-q:
## Overview
Closes https://github.com/votingworks/vxsuite/issues/2593
- Disable action buttons on the WIA screen after results have been marked as official.
- Additionally, adding messaging in the header for the screen to explain why editing has been disabled.

## Demo Video or Screenshot
wia_disabled_state

https://user-images.githubusercontent.com/264902/194391008-326b6aae-18bf-45af-88d9-533cf44dcf43.mov

## Testing Plan 
- Added unit test for the new case to verify disabled buttons and presence of explanation message
- Sanity check with local runthrough

## Checklist

### DIFF
--- a/frontends/election-manager/src/screens/write_ins_screen.test.tsx
+++ b/frontends/election-manager/src/screens/write_ins_screen.test.tsx
@@ -34,12 +34,12 @@ test('Tally results already marked as official', async () => {
 
   await screen.findByText(/No further changes may be made/);
 
-  const transcribeButtons = await screen.findAllByText(/Transcribe \d/);
+  const transcribeButtons = screen.queryAllByText(/Transcribe \d/);
   for (const transcribeButton of transcribeButtons) {
     expect(transcribeButton).toBeDisabled();
   }
 
-  const adjudicateButtons = await screen.findAllByText('Adjudicate');
+  const adjudicateButtons = screen.queryAllByText('Adjudicate');
   for (const adjudicateButton of adjudicateButtons) {
     expect(adjudicateButton).toBeDisabled();
   }
--- a/frontends/election-manager/src/screens/write_ins_screen.test.tsx
+++ b/frontends/election-manager/src/screens/write_ins_screen.test.tsx
@@ -17,6 +17,34 @@ test('No CVRs loaded', async () => {
   expect(screen.queryByText('Transcribe')).not.toBeInTheDocument();
 });
 
+test('Tally results already marked as official', async () => {
+  const backend = new ElectionManagerStoreMemoryBackend({
+    electionDefinition,
+  });
+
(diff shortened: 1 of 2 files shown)


kofi-q:
cc @mattroe for sanity check and any tweaks to the newly added copy

frontends/election-manager/src/screens/write_ins_screen.tsx
```diff
@@ -27,8 +27,13 @@ const ContentWrapper = styled.div`
   }
 `;
 
+const ResultsFinalizedNotice = styled.p`
+  color: rgb(71, 167, 75);
```
jonahkagan, line 31: Where did this color come from? I'm wondering if we have any colors defined or other patterns to follow already that we could use here.

frontends/election-manager/src/screens/write_ins_screen.test.tsx
```diff
@@ -17,6 +17,34 @@ test('No CVRs loaded', async () => {
...
+
+  renderInAppContext(<WriteInsScreen />, {
+    backend,
+    electionDefinition,
+    isOfficialResults: true,
+  });
+
+  await screen.findByText(/No further changes may be made/);
+
+  const transcribeButtons = await screen.findAllByText(/Transcribe \d/);
```
jonahkagan, line 37: Nit: I think you can use `screen.getAllByText` here, since you don't need to wait for any async DOM updates. In general, I try to only use the async testing-library queries when needed in order to indicate to readers that something is updating in the DOM, but that's just a personal style choice so take it or leave it

jonahkagan (approved):
I think you may have linked the wrong task in the PR description, because it didn't seem to correspond to this change. But code looks generally good! Just had a couple small notes

kofi-q:
Bah, good catch - definitely pasted the wrong link!

frontends/election-manager/src/screens/write_ins_screen.tsx
kofi-q: Actually meant to ask if we have any centrally defined colors I might be able to reference - copied this one from the "primary" color on our button component, but would prefer a named variable or something if we have one

frontends/election-manager/src/screens/write_ins_screen.test.tsx
kofi-q: Ah, thanks - I like that as a guiding principle. I also need to get more familiar with the react testing tools in general, so tips very welcome

frontends/election-manager/src/screens/write_ins_screen.tsx
jonahkagan: Yeah this might be a good q for @beausmith

PR #2641, opened 2022-10-07
VxScan: Automatically Set Precinct if Only One

### LINKED ISSUE #2636: VxScan: do not require configuration if election definition only has one precinct

adghayes:
## Overview
Closes #2636. If an election definition has only one precinct, then that individual precinct is set as the VxScan's precinct after the ballot package is done loading. This will bypass the "No Precinct Selected" screen. 

In the case of having only one precinct, I've removed the "All Precincts" option from VxScan.

## Demo Video or Screenshot
Not very much to show.

## Testing Plan 
- Added test for augmented `addTemplates` method
- Added test for `ElectionManagerScreen` in case of only one precinct

## Checklist


adghayes:
Reviewers: review the two commits as one change. The first one is a relic of a different approach which I decided against (but has more integration testing).

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
```diff
@@ -156,10 +158,16 @@ export function ElectionManagerScreen({
             onChange={changeAppPrecinctSelection}
             large
           >
-            <option value="" disabled>
-              Select a precinct for this device…
-            </option>
-            <option value={ALL_PRECINCTS_OPTION_VALUE}>All Precincts</option>
+            {!precinctSelectionValue && (
```
jonahkagan, line 161: What does this change do?

frontends/precinct-scanner/src/api/config.ts
```diff
@@ -249,6 +254,12 @@ export function addTemplates(pkg: BallotPackage): AddTemplatesEvents {
         });
       }
 
+      // Set the precinct if there is only one precinct
+      if (pkg.electionDefinition.election.precincts.length === 1) {
+        const onlyPrecinct = pkg.electionDefinition.election.precincts[0];
+        await setPrecinctSelection(singlePrecinctSelectionFor(onlyPrecinct.id));
+      }
```
jonahkagan, line 261: How did you choose to add this logic to `addTemplates`? It seems a bit unrelated to the templates logic.

A few alternative ideas:
- Rename this function to better represent what it's doing holistically
- Move this logic to somewhere else (maybe where `addTemplates` is called)
- Move this logic to the backend (since that's the general direction we're moving in)

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
adghayes: Hides the `Select a precinct for this device...` text when the precinct is already set, because it's just a disabled option in the list. A bit unrelated, and happy to reverse.

frontends/precinct-scanner/src/api/config.ts
adghayes: I chose to add it to to `addTemplates` because it's also where the election definition is set, so it seemed appropriate to set the precinct here as well. Originally (in the first commit) I had it where `addTemplates` is called. I don't think it was much clearer, and it was way harder to test. `addTemplates` should probably be renamed.

I prefer the idea of moving it to the backend, I'll look at that.

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
jonahkagan: Cool, that seems fine

jonahkagan (approved):

PR #2649, opened 2022-10-10
Add Manual Write-In Results

### LINKED ISSUE #2585: WIA: solution for non-official candidate hand-adjudicated (overvoted) ballots

adghayes:
## Overview
Closes #2585. Adds the ability to manually add write-in candidates and their vote counts via our manually entered data flow.

## Demo Video or Screenshot
[manual-write-ins.webm](https://user-images.githubusercontent.com/37960853/195188836-dde635da-f480-4ab1-b51c-79f9fadafc40.webm)

## Testing Plan
- Manual and Automated Testing

## Checklist

### DIFF
--- a/frontends/election-manager/src/app_write_in_flows.test.tsx
+++ b/frontends/election-manager/src/app_write_in_flows.test.tsx
@@ -43,7 +43,6 @@ afterEach(() => {
 });
 
 test('manual write-in data end-to-end test', async () => {
-  // Set up an an existing adjudicated value
   const backend = new ElectionManagerStoreMemoryBackend({
     electionDefinition: electionMinimalExhaustiveSampleDefinition,
   });
@@ -53,6 +52,22 @@ test('manual write-in data end-to-end test', async () => {
       'partial1.jsonl'
     )
   );
+  const card = new MemoryCard();
+  const hardware = MemoryHardware.buildStandard();
+  renderRootElement(<App card={card} hardware={hardware} />, { backend });
+
+  // Navigate to manual data entry
+  await authenticateWithElectionManagerCard(
+    card,
+    electionMinimalExhaustiveSampleDefinition
+  );
+  userEvent.click(screen.getByText('Tally'));
+  userEvent.click(screen.getByText('Add Manually Entered Results'));
+  userEvent.click(screen.getByText('Edit Precinct Results for Precinct 1'));
+
+  // Navigate away, adjudicated a write-in, and return - to ensure the
+  // page does not use stale data
+  userEvent.click(await screen.findByText('Cancel'));
   const writeIn = (
     await backend.loadWriteIns({
       contestId: 'zoo-council-mammal',
@@ -64,17 +79,6 @@ test('manual write-in data end-to-end test', async () => {
     'Chimera',
     'Chimera'
   );
-  const card = new MemoryCard();
-  const hardware = MemoryHardware.buildStandard();
-
-  renderRootElement(<App card={card} hardware={hardware} />, { backend });
-
-  await authenticateWithElectionManagerCard(
-    card,
-    electionMinimalExhaustiveSampleDefinition
-  );
-  userEvent.click(screen.getByText('Tally'));
-  userEvent.click(screen.getByText('Add Manually Entered Results'));
   userEvent.click(screen.getByText('Edit Precinct Results for Precinct 1'));
 
   // Enter some official results for Precinct 1
--- a/frontends/election-manager/src/app_write_in_flows.test.tsx
+++ b/frontends/election-manager/src/app_write_in_flows.test.tsx
@@ -1,0 +1,249 @@
+import React from 'react';
+import { screen, within } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import {
+  electionMinimalExhaustiveSampleDefinition,
+  electionMinimalExhaustiveSampleFixtures,
+} from '@votingworks/fixtures';
+import { MemoryCard, MemoryHardware, typedAs } from '@votingworks/utils';
+import { fakeKiosk, hasTextAcrossElements } from '@votingworks/test-utils';
+import fetchMock from 'fetch-mock';
+import { renderRootElement } from '../test/render_in_app_context';
+import { authenticateWithElectionManagerCard } from '../test/util/authenticate';
+import { App } from './app';
+import { ElectionManagerStoreMemoryBackend } from './lib/backends';
+import { VxFiles } from './lib/converters';
+import { MachineConfig } from './config/types';
+
+let mockKiosk!: jest.Mocked<KioskBrowser.Kiosk>;
+
+beforeEach(() => {
+  mockKiosk = fakeKiosk();
+  window.kiosk = mockKiosk;
+  fetchMock.reset();
+  fetchMock.get(
+    '/convert/tallies/files',
+    typedAs<VxFiles>({
+      inputFiles: [{ name: 'name' }, { name: 'name' }],
+      outputFiles: [{ name: 'name' }],
+    })
+  );
+  fetchMock.get(
+    '/machine-config',
+    typedAs<MachineConfig>({
+      machineId: '0000',
+      codeVersion: 'TEST',
+    })
+  );
+  fetchMock.delete('/admin/write-ins/cvrs', { body: { status: 'ok ' } });
+});
+
+afterEach(() => {
+  delete window.kiosk;
+});
+
+test('manual write-in data end-to-end test', async () => {
+  // Set up an an existing adjudicated value
+  const backend = new ElectionManagerStoreMemoryBackend({
+    electionDefinition: electionMinimalExhaustiveSampleDefinition,
+  });
+  await backend.addCastVoteRecordFile(
+    new File(
+      [electionMinimalExhaustiveSampleFixtures.partial1CvrFile.asBuffer()],
+      'partial1.jsonl'
+    )
+  );
+  const writeIn = (
+    await backend.loadWriteIns({
+      contestId: 'zoo-council-mammal',
+    })
+  )[0];
+  await backend.transcribeWriteIn(writeIn.id, 'Chimera');
+  await backend.adjudicateWriteInTranscription(
+    'zoo-council-mammal',
+    'Chimera',
+    'Chimera'
+  );
+  const card = new MemoryCard();
+  const hardware = MemoryHardware.buildStandard();
+
+  renderRootElement(<App card={card} hardware={hardware} />, { backend });
+
+  await authenticateWithElectionManagerCard(
+    card,
+    electionMinimalExhaustiveSampleDefinition
+  );
+  userEvent.click(screen.getByText('Tally'));
+  userEvent.click(screen.getByText('Add Manually Entered Results'));
+  userEvent.click(screen.getByText('Edit Precinct Results for Precinct 1'));
+
+  // Enter some official results for Precinct 1
+  await screen.findByText('Manually Entered Precinct Results:');
+  userEvent.type(
+    screen.getByTestId('zoo-council-mammal-zebra-input').closest('input')!,
+    '7'
+  );
+  userEvent.type(
+    screen.getByTestId('zoo-council-mammal-kangaroo-input').closest('input')!,
+    '3'
+  );
+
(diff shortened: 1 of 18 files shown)


frontends/election-manager/src/contexts/app_context.ts
```diff
@@ -48,6 +49,8 @@ export interface AppContextInterface {
   updateExternalTally: (
     newExternalTally: FullElectionExternalTally
   ) => Promise<void>;
+  manualTallyVotingMethod: VotingMethod;
+  setManualTallyVotingMethod: (votingMethod: VotingMethod) => void;
```
adghayes, line 53: I moved the default voting method for manual data up into app state. This means that, unfortunately, that data can be in two places - in the app state and as the voting method for the external tally. 

The original set up of having multiple entry points to create the manual tally, and also creating an empty tally when the voting method was toggled, was not ideal.

frontends/election-manager/src/app_write_in_flows.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+    })
+  );
+  fetchMock.delete('/admin/write-ins/cvrs', { body: { status: 'ok ' } });
+});
+
+afterEach(() => {
+  delete window.kiosk;
+});
+
+test('manual write-in data end-to-end test', async () => {
```
adghayes, line 45: I chose to break out this test from the `app.test.tsx` file because it is getting unwieldy, but we can move it back in it preferred.

frontends/election-manager/src/components/election_manager_writein_tally_report.tsx
```diff
@@ -76,8 +89,13 @@ export const ElectionManagerWriteInTallyReport = forwardRef<
...
     return (
       <TallyReport className="print-only" ref={ref}>
-        {ballotStylePartyIds.map((partyId) =>
-          precinctIds.map((precinctId) => {
+        {ballotStylePartyIds.map((partyId) => {
+          const writeInCountsFilteredByParty = filterWriteInCountsByParty(
+            writeInCounts,
+            election,
+            partyId
+          );
```
adghayes, line 97: There was previously a bug where all contests were appearing on both party pages. In other words, we were not filtering contests by party.

frontends/election-manager/src/screens/manual_data_import_precinct_screen.test.tsx
```diff
@@ -335,68 +444,17 @@ test('can enter data for yes no contests as expected', async () => {
   );
 });
 
-test('loads prexisting manual data to edit', () => {
-  const talliesByPrecinct = getEmptyExternalTalliesByPrecinct(
-    electionSampleDefinition.election
```
adghayes, line 340: This test broke because it wasn't using properly typed data, which now matters. In order to make it less cumbersome, I rewrote it to be simpler, use `buildExternalTally`, and not check all values / precincts.

frontends/election-manager/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -117,6 +195,53 @@ function getNumericalValueForTally(tally: number | EmptyValue): number {
   return tally;
 }
 
+// Convert internal structure of contest data that allows for empty strings, to the regular
+// type by mapping any empty string values to zeros.
+function convertContestTallies(
```
adghayes, line 200: Note this function was just moved to reorganize the file, not added.

frontends/election-manager/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -155,80 +315,70 @@ export function ManualDataImportPrecinctScreen(): JSX.Element {
...
     (p) => p.id === currentPrecinctId
   );
-  if (currentPrecinct === undefined) {
-    return (
-      <Prose>
-        Error: Could not find precinct {currentPrecinctId}.{' '}
-        <LinkButton to={routerPaths.manualDataImport}>Back to Index</LinkButton>
-      </Prose>
-    );
-  }
```
adghayes, line 168: Moved this down to avoid breaking hook rules.

frontends/election-manager/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -380,22 +681,38 @@ export function ManualDataImportPrecinctScreen(): JSX.Element {
...
+
           return (
             <ContestData key={contest.id}>
               <Text small>{contest.section}</Text>
               <h3>{contestTitle}</h3>
               <Table borderTop condensed>
                 <tbody>
                   {contest.type === 'candidate' &&
-                    getAllPossibleCandidatesForCandidateContest(contest).map(
+                    getCandidatesFromContestTally(contestTally).map(
```
adghayes, line 695: We're now populating this form with candidates from the _results_, which we created to include pre-adjudicated write-ins and could include manual write-ins. Previously we would populate from the candidates in the contest definition.

frontends/election-manager/src/utils/external_tallies.test.ts
```diff
@@ -41,50 +39,6 @@ const presidentcontest = electionSample.contests.find(
   (c) => c.id === 'president'
 ) as CandidateContest;
 
-// Note this helper uses 'getEmptyContestTallies' and 'getTotalNumberOfBallots' util functions so should not be used to test those implementations.
-function buildExternalTally(
-  election: Election,
```
adghayes, line 46: Moved to a separate helper file.

carolinemodic:
@adghayes heads up I think the video is wrong

carolinemodic:
I realize this is irrelevant for NH but what happens if there is more then one precinct, 

You add "Matt Row" as a write-in for Precinct 1, and then go to Precinct 2 will Matt already be there are a write-in? Or will you need to add it again?

frontends/election-manager/src/app_write_in_flows.test.tsx
carolinemodic: No seperate is better IMO

frontends/election-manager/src/app_write_in_flows.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+  // Add a write-in to appear in the results
+  userEvent.click(within(zooMammalCouncil).getByText('Add Write-In Candidate'));
+  userEvent.type(
+    within(zooMammalCouncil).getByTestId('zoo-council-mammal-write-in-input'),
+    'Rapidash'
+  );
+  userEvent.click(within(zooMammalCouncil).getByText('Add'));
+  userEvent.type(
+    within(zooMammalCouncil).getByTestId(
+      'zoo-council-mammal-write-in-(Rapidash)-manual-input'
```
carolinemodic, line 111: :)

adghayes:
If you add it for one precinct, it will add it to all the other precincts. If you remove it from one precinct, it will be removed from all the other precincts. This has some test coverage in the end-to-end test in `app_write_in_flows.test.tsx`.

frontends/election-manager/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -155,80 +315,70 @@ export function ManualDataImportPrecinctScreen(): JSX.Element {
...
-          ...optionTally,
-          tally: optionTally.tally === '' ? 0 : optionTally.tally,
-        };
+
+  const writeInSummaryQuery = useWriteInSummaryQuery({ status: 'adjudicated' });
+  // Get empty tallies with previously adjudicated candidate names, only
+  // when none already exist at initial page load
+  useEffect(() => {
+    if (talliesByPrecinct) return;
+    if (!writeInSummaryQuery.data) return;
```
carolinemodic, line 345: I forget exactly the right thing to check for but we were seeing a bug with the tally reports where this would get fetched twice right away and the first time would be stale data, so I would just make sure to test adding a new adjudication value, and then going to this screen first thing and making sure that value shows up.

frontends/election-manager/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -404,22 +721,33 @@ export function ManualDataImportPrecinctScreen(): JSX.Element {
                         </ContestDataRow>
                       )
                     )}
+                  {contest.type === 'candidate' && contest.allowWriteIns && (
+                    <AddWriteInRow
+                      addWriteInCandidate={(name) =>
+                        addWriteInCandidate(contest.id, name)
```
carolinemodic, line 727: so what is the candidate ID we are giving these write-ins? It should still probably start with __write-in or something right, so we know its a write-in?

frontends/election-manager/src/utils/write_ins.ts
```diff
@@ -34,3 +50,142 @@ export function getWriteInCountsByContestAndCandidate(
...
+  }
+  return filteredCounts;
+}
+
+// ExternalTally write-in candidate id creation
+export function getAdjudicatedWriteInCandidateId(
+  name: string,
+  manual: boolean
+): string {
+  return `write-in-(${name})${manual ? '-manual' : ''}`;
```
carolinemodic, line 76: oh I see this is where the ID comes from

frontends/election-manager/src/screens/manual_data_import_precinct_screen.test.tsx
```diff
@@ -118,46 +122,72 @@ test('can enter data for candidate contests as expected', async () => {
...
+
+  // Contest total ballots now should be updated
+  await waitFor(() => {
+    expect(screen.getByTestId('president-numBallots').textContent).toEqual(
+      '100'
+    );
   });
 
-  // A contest that does not allow write ins has no write in row.
-  expect(queryAllByTestId('president-write-in').length).toBe(0);
```
carolinemodic, line 159: why did you remove this check?

frontends/election-manager/src/screens/manual_data_import_precinct_screen.test.tsx
```diff
@@ -188,67 +218,137 @@ test('can enter data for candidate contests as expected', async () => {
...
+    </Route>,
+    {
+      route: '/tally/manual-data-import/precinct/precinct-1',
+      electionDefinition: electionMinimalExhaustiveSampleDefinition,
+      backend,
+    }
   );
+
+  // Adjudicated value should be in the form and not be removable
+  await screen.findByText('Chimera (write-in)');
```
carolinemodic, line 344: nice

carolinemodic (approved):
Whew I knew that the manual data stuff was thrown together by me but I forgot how messy it was haha

frontends/election-manager/src/screens/manual_data_import_precinct_screen.tsx
adghayes: I had the same issue in development, and that's why I added the `writeInSummaryQuery.isFetchedAfterMount` check. I agree it would be good to bake into the test for the screen.

frontends/election-manager/src/utils/write_ins.ts
adghayes: Does that ID format seem OK?

frontends/election-manager/src/screens/manual_data_import_precinct_screen.test.tsx
adghayes: Now, none of the tables have a write-in row by default, since you can add them. The difference would be that a contest that does not allow write-ins has no button to "Add Write-In Candidate," which I believe I cover in a new test.

It opens up a larger question - what if the user wants to just represent Write-Ins generically in their manual data and not specify names? Currently they can just add a "Generic" write-in candidate, but for NH this is a non-issue.

frontends/election-manager/src/screens/manual_data_import_precinct_screen.tsx
adghayes: After thinking about it a bit, I think this is covered by the integration test, actually, so I will leave it for now.

adghayes: Edited end-to-end test to check this.

frontends/election-manager/src/app_write_in_flows.test.tsx
adghayes: :)

Issue #2631, opened 2022-10-04
nh-accuvote: hide precinct vs. absentee counts


mattroe:
We don't have a mechanism of distinguishing precinct vs. absentee ballot counts when scanning accuvote ballots and therefore are misrepresenting on tally reporting absentee vs. precinct ballots.

When in accuvote-converter mode, we should not show:
- Precinct vs. absentee on tally reports - in both VxAdmin and VxScan reports
- The reports by voting method section on the reports page

carolinemodic:
@eventualbuddha Do you mean adding something to the election definition for can distinigush ballot type? 

@kofi-q For context here in VxAdmin there are various reports, you can see on the reports tab that all have a box at the top showing the total number of ballots that are "precinct" ballots or "absentee" ballots, we just want to remove that box entirely for all of the reports. Separately on the reports page in VxAdmin there are reports and a table that break down the tallies by precinct/absentee voting method and we should hide that table. 

VxScan also prints a polls closed and polls open report that has the same box (and general layout) as the VxAdmin report so we will want to remove the box there too, figuring out how to test that part might be a bit trickier so I would start with the VxAdmin pieces.

eventualbuddha:
Doesn't have to be in the election definition. Can still ultimately be checking for `gridLayouts`, but I just want to have a function called something like `canDistinguishPrecinctAndAbsenteeBallots` to make it so we can change how we make that decision later.

PR #2652, opened 2022-10-12
[VxAdmin|VxScan] Hide tally by voting method if unable to distinguish

kofi-q:
## Overview
https://github.com/votingworks/vxsuite/issues/2631:
- Hiding tallies broken down by voting method whenever we're unable to distinguish voting methods for an election.
- Adding `canDistinguishPrecinctAndAbsenteeBallots` util (currently just a proxy for absence of `gridLayouts` in the election definition).

### [VxAmdin] Reports Screen:
Screenshot 2022-10-12 at 10 16 33 AM Screenshot 2022-10-12 at 10 16 01 AM

### [VxAdmin] Tally Report Table:
Screenshot 2022-10-12 at 9 43 40 AM Screenshot 2022-10-12 at 9 45 11 AM

### [VxScan] Polls Open/Closed Report:
Screenshot 2022-10-12 at 12 17 15 PM Screenshot 2022-10-12 at 12 16 04 PM

## Testing Plan 
- Unit test coverage and local verification

## Checklist

### DIFF
--- a/libs/ui/src/tally_report_summary.test.tsx
+++ b/libs/ui/src/tally_report_summary.test.tsx
@@ -2,20 +2,12 @@ import React from 'react';
 import { render, getByText as domGetByText } from '@testing-library/react';
 
 import { Tally, VotingMethod } from '@votingworks/types';
-import { electionSampleDefinition } from '@votingworks/fixtures';
-import { mockOf } from '@votingworks/test-utils';
-import { canDistinguishVotingMethods } from '@votingworks/utils';
+import {
+  electionSampleDefinition,
+  electionGridLayoutNewHampshireAmherstFixtures,
+} from '@votingworks/fixtures';
 
 import { TallyReportSummary } from './tally_report_summary';
-
-jest.mock('@votingworks/utils', (): typeof import('@votingworks/utils') => ({
-  ...jest.requireActual('@votingworks/utils'),
-  canDistinguishVotingMethods: jest.fn(),
-}));
-
-beforeEach(() => {
-  mockOf(canDistinguishVotingMethods).mockReturnValue(true);
-});
 
 test('Renders with data source table and voting method table when all data provided', () => {
   const ballotCounts: Tally['ballotCountsByVotingMethod'] = {
@@ -77,8 +69,6 @@ test('Hides the other row in the voting method table when empty', () => {
 });
 
 test('Is empty element if voting methods cannot be distinguished for election', () => {
-  mockOf(canDistinguishVotingMethods).mockReturnValue(false);
-
   const ballotCounts: Tally['ballotCountsByVotingMethod'] = {
     [VotingMethod.Absentee]: 1200,
     [VotingMethod.Precinct]: 1045,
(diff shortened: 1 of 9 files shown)


kofi-q:
heads-up: looks like i've got one more manual verification to do for the precinct scanner change - will add screenshots for that as one once i get it working locally

libs/ui/src/tally_report_summary.test.tsx
```diff
@@ -55,7 +70,27 @@ test('Hides the other row in the voting method table when empty', () => {
...
       totalBallotCount={3579}
       ballotCountsByVotingMethod={ballotCounts2}
+      election={electionSampleDefinition.election}
     />
   );
   expect(queryAllByText2('Other').length).toBe(0);
 });
+
+test('Is empty element if voting methods cannot be distinguished for election', () => {
+  mockOf(canDistinguishPrecinctAndAbsenteeBallots).mockReturnValue(false);
```
jonahkagan, line 80: Why does this need to be mocked here? My guess is that it's because `electionSampleDefinition.election` doesn't have `gridLayouts`. I think there's probably an election fixture that does have `gridLayouts` that you could use instead.

libs/utils/src/elections.ts
```diff
@@ -0,0 +1,11 @@
+import { Election } from '@votingworks/types';
+
+/**
+ * Determines if we can distinguish between voting methods for a given election.
+ * See https://github.com/votingworks/vxsuite/issues/2631 for added context.
+ */
+export function canDistinguishPrecinctAndAbsenteeBallots(
```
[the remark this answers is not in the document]
kofi-q: Oh thanks, that's a better name

libs/ui/src/tally_report_summary.test.tsx
kofi-q: Just isolating this component from the internal logic of the `canDistinguish` utility, since that could (in theory) change in the future

eventualbuddha: Yep: https://github.com/votingworks/vxsuite/blob/e75f1d7d12e3e9bfdfa7ca43518ac07dbdaacfb5/libs/fixtures/src/data/electionGridLayoutNewHampshireHudson/election.json.ts.

jonahkagan: Ah gotcha. That's an interesting approach @kofi-q. On one hand it helps avoid testing the internal logic of `canDistinguish`, on the other hand the test now depends on internal implementation details of the component, which is something we've generally tried to avoid. I'm not sure which concern is more important

kofi-q: Ohh that makes sense - yeah fine I'm either way, depending on where the team leans, philosophically. I think this is mostly residual habit from a previous job, so not tied to it - will switch over to varying the election data

jonahkagan: That actually reminds me that we did try to align a little while back on a testing philosophy. I'll send you the doc

kofi-q:
Managed to test and verify all the relevant user flows now (and finally got an intro to running in kiosk-browser 🙂) - LMK if you notice anything else that needs double-checking

beausmith:
@kofi-q: When printing reports outside Kiosk Browser, please update your print prefs to:
- Margins: `none`
- Background Graphics: checked/enabled

Screen Shot 2022-10-12 at 14 00 45

jonahkagan (approved):
Looks good!

PR #2659, opened 2022-10-12
Environment Flag to Skip PIN Entry

adghayes:
Adds a flag, `REACT_APP_VX_SKIP_PIN_ENTRY` that skips pin entry during authentication.

Each commit (small) I recommend reviewing separately. The second is refactoring BMD to use the shared component, so the flag works in BMD (and we have less code!).


libs/ui/src/env.d.ts
```diff
@@ -1,7 +1,6 @@
 declare namespace NodeJS {
   export interface ProcessEnv {
-    readonly NODE_ENV: 'development' | 'production' | 'test';
-    readonly REACT_APP_VX_DEV?: string;
-    readonly REACT_APP_VX_DISABLE_CARD_READER_CHECK?: string;
+    NODE_ENV: 'development' | 'production' | 'test';
+    REACT_APP_VX_SKIP_PIN_ENTRY?: string;
```
benadida, line 4: I realize you're removing `readonly` for testing purposes, but I'm not sure that's the right tradeoff. I think in other tests we've chosen to mock the feature flag function, not `process.env` directly.

@eventualbuddha what do you think?

libs/ui/src/env.d.ts
```diff
@@ -1,7 +1,6 @@
 declare namespace NodeJS {
   export interface ProcessEnv {
-    readonly NODE_ENV: 'development' | 'production' | 'test';
-    readonly REACT_APP_VX_DEV?: string;
-    readonly REACT_APP_VX_DISABLE_CARD_READER_CHECK?: string;
```
benadida, line 5: did you mean to remove these?

benadida (commented):
over looks good, except for this question I have around mocking `process.env` that I'd love @eventualbuddha's take on.

libs/ui/src/env.d.ts
adghayes: FWIW, I chose to mock it this way because [in other tests we have mocked `process.env` directly](https://github.com/votingworks/vxsuite/blob/2b7cbb1eb674cdf9b8e902cf6560761dfacbe5df/libs/utils/src/environment_flag.test.ts#L30).

libs/ui/src/env.d.ts
adghayes: Yes, per this slack discussion: https://votingworks.slack.com/archives/CEL6D3GAD/p1665608644655979. They're only necessary when referenced explicitly, as in the tests now.

libs/ui/src/env.d.ts
eventualbuddha: I think I agree with @benadida here. We're changing `process.env` directly in that test because the code under test is the thing that accesses `process.env`. There are some other places we do it, but we also have places where we're [mocking the `isFeatureFlagEnabled` function](https://github.com/votingworks/vxsuite/blob/7b6a705f25164e145e9153df84e20c2be29f208c/frontends/precinct-scanner/src/screens/scan_warning_screen.test.tsx#L81-L85). I think neither of these approaches are great, really. I think I'd prefer something that can be controlled with a React context that we could use like this:

```ts
// in a component
const isSkipPinEntryEnabled = useFeatureFlag(EnvironmentFlagName.SKIP_PIN_ENTRY);

useEffect(() => {
  if (isSkipPinEntryEnabled) {
    auth.checkPasscode(auth.user.passcode);
  }

// …

// in a test
renderInAppContext(
  <SomeComponent />,
  {
    featureFlags: {
      [EnvironmentFlagName.SKIP_PIN_ENTRY]: true
    }
  }
);
```

But we don't need to do that in this PR. I'd prefer changing this to do the mocking as I linked to above, but I don't think it's required.

libs/ui/src/unlock_machine_screen.tsx
```diff
@@ -55,6 +59,10 @@ export function UnlockMachineScreen({ auth }: Props): JSX.Element {
   }, []);
 
   useEffect(() => {
+    if (isFeatureFlagEnabled(EnvironmentFlagName.SKIP_PIN_ENTRY)) {
+      auth.checkPasscode(auth.user.passcode);
```
eventualbuddha, line 70: Should we `return` after this?

eventualbuddha (approved):

libs/ui/src/unlock_machine_screen.tsx
adghayes: The rest of the effect won't run because there is no security pin to check, so it shouldn't matter, but I think it's a good idea in case there's edge cases I can't think of.

libs/ui/src/env.d.ts
adghayes: Made these `readonly` again.

libs/ui/src/env.d.ts
adghayes: I switched to the other way to mock.

PR #2667, opened 2022-10-13
fix(scan-nh): improve bubble matching

eventualbuddha:
## Overview

Previously we'd start from where we thought the bubble should be and look in an area around that to find the location that gave the best mark score. The problem with this is it disincentivizes actually lining up with the bubble when it's empty or nearly so. Instead, we now compute two scores: one for how well we match the bubble and one for how filled the area is. We take the fill score for the area that had the best match, regardless of whether there might be higher or lower fill scores elsewhere. This means we should get a lot closer to 0% scores for truly empty bubbles and better align on bubbles with faint marks, which may actually reduce their scores a bit. Because of this, I suggest we change our `definite` mark threshold to 8% from 12% (based on extensive manual testing).

## Demo Video or Screenshot
| Before | After |
|-|-|
| blank ballot - before | blank ballot - after |

## Testing Plan 
Tested manually extensively with @mattroe and @carolinemodic. Added and improved automated testing.

## Checklist


libs/ballot-interpreter-nh/src/interpret/interpret_oval_marks.ts
```diff
@@ -1,17 +1,225 @@
...
+export function scoreOvalMark(
+  imageData: ImageData,
+  ovalTemplate: ImageData,
+  ovalTopLeftPoint: Point,
+  geometry: BallotCardGeometry,
+  threshold: number,
+  debug: Debugger
+): Omit<InterpretedOvalMark, 'gridPosition'> {
+  const maximumOffset = 7;
+  const outlinedOvalTemplate = outline(binarize(ovalTemplate, threshold));
```
benadida, line 60: instead of a scanned bubble template that we then binarize every time... should we maybe use the rendered graphic of the ballot PDF, cropped accordingly, rather than a scan of the bubble, which will inevitably carry some noise / skew? Not a big deal, just a thought.

benadida (approved):

libs/ballot-interpreter-nh/src/interpret/interpret_oval_marks.ts
eventualbuddha: That's actually part of the point. The scans will have that stuff, so I wanted to have an image that is as close as possible to what we're actually going to see coming from the scanner. I'm not saying this is perfect, but I think it's better than trying to pull something from the PDF.

PR #2668, opened 2022-10-13
Build Script to Build All Dependencies

adghayes:
Modeled after @eventualbuddha's #2640, this PR creates a general `build-dev` script and uses it to build all the frontends and their services. Like `run-dev` it uses the existing vx config package.json key to determine which services need to be built along with a frontend.

`pnpm build` by default builds all the services (which normally is quite fast) and `pnpm build:core` builds only the frontend and its dependencies, to parallel `pnpm run`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ more information about VotingWorks.
 ## Development
 
 Building VxSuite for development requires git, [NodeJS](https://nodejs.org/)
-v16.14.2 and [pnpm](https://pnpm.js.org) v5 (or use `corepack` to automatically use the right package manager).
+v16.14.2 and [pnpm](https://pnpm.js.org) v5 (or use `corepack` to automatically
+use the right package manager).
 
 Most of the code is written in TypeScript. We follow the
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -273,8 +271,16 @@ help you contribute.
...
+
+Feature flags are defined in the libs/utils subrepo
+[here](https://github.com/votingworks/vxsuite/blob/main/libs/utils/src/environment_flag.ts).
+To configure which flags you are using you can set them in an `.env.local` file.
+This file can live either at the root of vxsuite or in the subrepo you want that
+flag to apply to. Flag values set in subrepo files will override those set at
+the root. `.env.local` file will also override a default `.env` file at either a
+subrepo or root level. To generate a `.env.local` file properly run the
+`pnpm configure-dev` command. If you are using VxDev you can run the
+`pnpm configure-vxdev-env` command BEFORE running the update code program.
```
adghayes, line 283: I think maybe my linter locally is moving things around. Perhaps we don't have standard linting for `.md`.

PR #2670, opened 2022-10-13
Mock Smartcard Tweaks

adghayes:
Makes two tweaks to the `mockCardReader.py` script for easier development, as separate commits.

First, allows the `mockCardReader.py` to accept a `--dip` flag which will result in the card getting removed shortly after being inserted. Only makes sense to use alongside the `SKIP_PIN_ENTRY` flag. 

Second, adds the ability to mock long values on poll worker cards. This is important in developing report printing flows on VxMark. I've added fixtures for one election definition only.


services/smartcards/writePollWorkerCard.py
```diff
@@ -12,13 +12,24 @@
 election_bytes = f.read()
 f.close()
 
+long_value = None
+
+if len(sys.argv) > 2:
```
eventualbuddha, line 17: Can we either add a comment explaining what the expected args are and what they do, or handle `--help`/`-h` args to print expected usage?

services/smartcards/mockCardReader.py
```diff
@@ -232,6 +238,8 @@ def disable():
         elif arg == "--ballot-style":
             i += 1
             ballot_style_id = sys.argv[i]
+        elif arg == "--dip":
```
eventualbuddha, line 241: Add `--dip` to the usage info above, along with an explanation and example.

eventualbuddha (approved):
Needs a couple tweaks, but LGTM after.

services/smartcards/writePollWorkerCard.py
adghayes: Comment!

Issue #2661, opened 2022-10-12
Update default threshold to X% for nh-accuvote election definitions 


carolinemodic:
X is most likely 8 but to be confirmed by @mattroe and @carolinemodic

PR #2674, opened 2022-10-14
update nh mark thresholds

carolinemodic:
Updates mark thresholds for NH to 8% as discussed 

https://github.com/votingworks/vxsuite/issues/2661


mattroe (approved):
What does `marginal` currently do out of curiosity?

carolinemodic:
I don't think it currently does anything

PR #2681, opened 2022-10-15
VxScan: Stream CVR Exports

adghayes:
## Overview
Does two things:
1. Makes CVR exports streaming (from the backend to the frontend to the download) to accommodate larger CVRs
2. Sets up option to get CVRs without inline images for tally purposes, to speed up polls closing

Now that we are downloading the file from the backend, we should set it's file name via the content disposition. However, currently we don't have access to the machine ID (which needs to be in the filename) in the backend. That's easy to do, but to do it right, we should abstract out the machine config provider to `utils` which I didn't want to bother with on the fly.

## Testing Plan 
- Edited a bunch of tests, added a few
- Manual testing _in kiosk-brower_
- Note: I have not tested this in the browser. We have some logic for downloads in the browser for dev purposes, and I have not validated those, although they should be intact

## Checklist

### DIFF
--- a/frontends/precinct-scanner/src/api/scan.test.ts
+++ b/frontends/precinct-scanner/src/api/scan.test.ts
@@ -11,10 +11,10 @@ test('calibrate success', async () => {
   expect(fetchMock.done()).toBe(true);
 });
 
-test('getExport returns CVRs on success', async () => {
+test('getExportWithoutImages returns CVRs on success', async () => {
   const fileContent = electionWithMsEitherNeitherFixtures.cvrData;
   fetchMock.postOnce('/precinct-scanner/export', fileContent);
-  const cvrsFileString = await scan.getExport();
+  const cvrsFileString = await scan.getExportWithoutImages();
   const lines = cvrsFileString.split('\n');
   const cvrs = lines.flatMap((line) =>
     line.length > 0 ? (JSON.parse(line) as CastVoteRecord) : []
@@ -22,12 +22,20 @@ test('getExport returns CVRs on success', async () => {
   expect(cvrs).toHaveLength(100);
 });
 
+test('getExportWithoutImages specifies to skip images in request', async () => {
+  fetchMock.postOnce(
+    { url: '/precinct-scanner/export', body: { skipImages: true } },
+    'anything'
+  );
+  await scan.getExportWithoutImages();
+});
+
 test('getExport throws on failure', async () => {
   fetchMock.postOnce('/precinct-scanner/export', {
     status: 500,
     body: { status: 'error' },
   });
-  await expect(scan.getExport()).rejects.toThrowError(
+  await expect(scan.getExportWithoutImages()).rejects.toThrowError(
     'failed to generate scan export'
   );
 });
--- a/frontends/precinct-scanner/src/components/export_results_modal.tsx
+++ b/frontends/precinct-scanner/src/components/export_results_modal.tsx
@@ -167,6 +167,7 @@ export function ExportResultsModal({
                 <Button
                   data-testid="manual-export"
                   onPress={() => exportResults(true)}
+                  disabled // Not currently supported
                 >
                   Save
                 </Button>
--- a/frontends/precinct-scanner/src/components/export_results_modal.tsx
+++ b/frontends/precinct-scanner/src/components/export_results_modal.tsx
@@ -129,7 +129,12 @@ export function ExportResultsModal({
   }
 
   if (currentState === ModalState.SAVING) {
-    return <Modal content={<Loading />} onOverlayClick={onClose} />;
+    return (
+      <Modal
+        content={<Loading>Saving CVRs</Loading>}
+        onOverlayClick={onClose}
+      />
+    );
   }
 
   if (currentState !== ModalState.INIT) {
--- a/frontends/precinct-scanner/src/utils/save_cvr_export_to_usb.test.ts
+++ b/frontends/precinct-scanner/src/utils/save_cvr_export_to_usb.test.ts
@@ -1,7 +1,6 @@
 import { electionMinimalExhaustiveSampleFixtures } from '@votingworks/fixtures';
 import { fakeKiosk, fakeUsbDrive } from '@votingworks/test-utils';
 import fetchMock from 'fetch-mock';
-import fileDownload from 'js-file-download';
 import MockDate from 'mockdate';
 import { fakeFileWriter } from '../../test/helpers/fake_file_writer';
 import { MachineConfig } from '../config/types';
@@ -9,12 +8,27 @@ import { MachineConfig } from '../config/types';
 import { saveCvrExportToUsb } from './save_cvr_export_to_usb';
 
 MockDate.set('2020-10-31T00:00:00.000Z');
-jest.mock('js-file-download');
 
 const machineConfig: MachineConfig = {
   machineId: '0003',
   codeVersion: 'TEST',
 };
+
+let kiosk = fakeKiosk();
+
+beforeEach(() => {
+  kiosk = fakeKiosk();
+  kiosk.getUsbDrives.mockResolvedValue([fakeUsbDrive()]);
+  kiosk.writeFile.mockResolvedValue(
+    fakeFileWriter() as unknown as ReturnType<KioskBrowser.Kiosk['writeFile']>
+  );
+  kiosk.saveAs.mockResolvedValue(fakeFileWriter());
+  window.kiosk = kiosk;
+});
+
+afterEach(() => {
+  delete window.kiosk;
+});
 
 test('throws error when scan service errors', async () => {
   fetchMock.postOnce('/precinct-scanner/export', {
@@ -31,12 +45,12 @@ test('throws error when scan service errors', async () => {
       openFilePickerDialog: false,
     })
   ).rejects.toThrowErrorMatchingInlineSnapshot(
-    `"failed to generate scan export"`
+    `"Unable to get CVR file: FetchFailed (status=Internal Server Error)"`
   );
 });
 
 test('throws error when there is no usb mounted in kiosk mode', async () => {
-  window.kiosk = fakeKiosk();
+  kiosk.getUsbDrives.mockResolvedValue([]);
   fetchMock.postOnce(
     '/precinct-scanner/export',
     electionMinimalExhaustiveSampleFixtures.cvrData
@@ -60,11 +74,6 @@ test('calls kiosk saveAs when opening file picker dialog', async () => {
     '/precinct-scanner/export',
     electionMinimalExhaustiveSampleFixtures.cvrData
   );
-  const mockKiosk = fakeKiosk();
-  mockKiosk.getUsbDrives.mockResolvedValue([fakeUsbDrive()]);
-  const fileWriter = fakeFileWriter();
-  mockKiosk.saveAs = jest.fn().mockResolvedValue(fileWriter);
-  window.kiosk = mockKiosk;
   await saveCvrExportToUsb({
     electionDefinition:
       electionMinimalExhaustiveSampleFixtures.electionDefinition,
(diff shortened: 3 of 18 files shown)


frontends/precinct-scanner/src/utils/save_cvr_export_to_usb.test.ts
```diff
@@ -73,9 +82,8 @@ test('calls kiosk saveAs when opening file picker dialog', async () => {
     isTestMode: false,
     openFilePickerDialog: true,
   });
-  expect(window.kiosk.saveAs).toHaveBeenCalledWith({
-    defaultPath:
-      'fake mount point/cast-vote-records/sample-county_example-primary-election_0dabcacc5d/machine_0003__0_ballots__2020-10-31_00-00-00.jsonl',
+  expect(kiosk.saveAs).toHaveBeenCalledWith({
+    defaultPath: 'machine_0003__0_ballots__2020-10-31_00-00-00.jsonl',
```
adghayes, line 86: the default "Save As" path for CVRs is no longer within a specific directory, but just at the root of the USB drive. seems unimportant

services/scan/src/precinct_scanner_app.ts
```diff
@@ -440,13 +453,26 @@ export async function buildPrecinctScannerApp(
...
+      debug(`exporting CVRs ${skipImages ? 'without' : 'with'} inline images`);
+
+      const cvrFilename = generateFilenameForScanningResults(
+        // TODO: Move machine config provider to shared utilities and access
+        // actual machine config, with dev overrides, here instead
+        '0000',
+        getBallotsCounted(store),
+        store.getTestMode(),
+        new Date()
+      );
```
adghayes, line 467: this is really a placeholder, and could be removed, because the frontend is overriding the filename in the content-disposition

adghayes: maybe it's bad to have it in for now because if the `TODO` isn't done and and the endpoint is relied on later it could give CVR filenames without correct machine ID

services/scan/src/store.test.ts
```diff
@@ -716,6 +716,102 @@ test('exportCvrs does not export ballot images when feature flag turned off', as
   );
 });
 
+test('exportCvrs does not export ballot images when skipImages is true', async () => {
+  mockOf(isFeatureFlagEnabled).mockImplementation(() => true);
+
+  const buildCastVoteRecordMock = jest.spyOn(
```
adghayes, line 722: test which is, for the most part, shamelessly copied

services/scan/src/store.ts
```diff
@@ -946,6 +949,8 @@ export class Store {
         writeStream.write('\n');
       }
     }
+
+    writeStream.end();
```
adghayes, line 953: didn't end up being important after I landed on using the `Passthrough` stream, but still makes sense to close the stream

frontends/precinct-scanner/src/api/scan.test.ts
```diff
@@ -11,23 +11,31 @@ test('calibrate success', async () => {
...
   );
   expect(cvrs).toHaveLength(100);
 });
 
+test('getExportWithoutImages specifies to skip images in request', async () => {
+  fetchMock.postOnce(
+    { url: '/precinct-scanner/export', body: { skipImages: true } },
+    'anything'
+  );
+  await scan.getExportWithoutImages();
```
eventualbuddha, line 30: This doesn't appear to actually validate anything about the request.

frontends/precinct-scanner/src/utils/save_cvr_export_to_usb.ts
```diff
@@ -52,26 +50,40 @@ export async function saveCvrExportToUsb({
...
+      },
+    });
+  }
+  if (!result.isOk()) {
+    const error = result.err();
+    let errorMessage = '';
+    switch (error.kind) {
+      case DownloadErrorKind.FetchFailed:
+      case DownloadErrorKind.FileMissing:
+        errorMessage = `Unable to get cvr file: ${error.kind} (status=${error.response.statusText})`;
```
eventualbuddha, line 75:
```suggestion
        errorMessage = `Unable to get CVR file: ${error.kind} (status=${error.response.statusText})`;
```

frontends/precinct-scanner/src/api/scan.test.ts
adghayes: Wouldn't the test fail if `fetchMock` was not matched?

libs/api/src/services/scan/index.ts
```diff
@@ -502,20 +502,24 @@ export const DoneTemplatesResponseSchema: z.ZodSchema<DoneTemplatesResponse> =
...
- *
  * @url /scan/export
  * @method POST
  */
-export type ExportRequest = never;
+export type ExportRequest =
+  | undefined
+  | {
+      skipImages: boolean;
+    };
```
eventualbuddha, line 512: Maybe this?

```ts
export type ExportRequest = Optional<{ skipImages: boolean }>;
```

frontends/precinct-scanner/src/api/scan.test.ts
adghayes: `fetchMock` expecting a request meeting a certain filter was, in my mind, what was being validated.

services/scan/src/precinct_scanner_app.test.ts
```diff
@@ -1390,3 +1390,20 @@ test('stops completely if plustekctl freezes and cant be killed', async () => {
...
+
+    await configureApp(app);
+    await request(app)
+      .post('/precinct-scanner/export')
+      .set('Accept', 'application/json')
+      .set('Content-Type', 'application/json')
+      .send({ skipImages: true })
+      .expect(200);
+
+    expect(spyExportCvrs).toHaveBeenCalledWith(expect.anything(), true);
```
eventualbuddha, line 1407: It's a little hard to tell what the `true` here refers to. Maybe make it an options object instead of just a bare argument?

libs/api/src/services/scan/index.ts
adghayes: Done.

services/scan/src/precinct_scanner_app.ts
```diff
@@ -440,13 +453,26 @@ export async function buildPrecinctScannerApp(
...
+
+      response
+        .header('Content-Type', 'text/plain; charset=utf-8')
+        .header('Content-Disposition', `attachment; filename="${cvrFilename}"`);
+      const cvrStream = new PassThrough();
```
eventualbuddha, line 472: Why not `store.exportCvrs(response, skipImages)`? Why do we have the `PassThrough` stream?

adghayes: `exportCvrs` is expecting a Writable interface, but we also need it to be a Readable interface to pipe it into the response. The `Passthrough` just gives us both. Do you know an easier way?

frontends/precinct-scanner/src/api/scan.test.ts
eventualbuddha: Ah, you're right. The `body` of the first argument does match the request body.

adghayes: Do you think expecting `fetchMock.done()` would make it any more clear?

services/scan/src/precinct_scanner_app.test.ts
adghayes: Will do.

frontends/precinct-scanner/src/api/scan.test.ts
eventualbuddha: Nah, we do this "just call a thing and if it doesn't blow up it passes" elsewhere.

services/scan/src/precinct_scanner_app.ts
```diff
@@ -440,13 +453,26 @@ export async function buildPrecinctScannerApp(
...
+      cvrStream.pipe(response);
+      store.exportCvrs(cvrStream, { skipImages });
```
eventualbuddha, line 474: Should we make this change with the central scanner app as well?

adghayes: Yeah, definitely, but I think we can punt that for NH because NH would be using the central scanner. Is that right @mattroe?

eventualbuddha: I believe that's the case. I just thought if it's easy let's do it now. Your call.

adghayes: https://github.com/votingworks/vxsuite/issues/2682

services/scan/src/precinct_scanner_app.ts
eventualbuddha: The `response` is writable. I just changed it to this and it appears to work with kiosk-browser and pass the tests.

services/scan/src/precinct_scanner_app.ts
adghayes: @carolinemodic do you want to merge this in imminently? In which case I'll put off the change and testing for central scanner.

carolinemodic: Yeah lets do the central scanner in a separate PR as thats not used for NH

services/scan/src/precinct_scanner_app.ts
adghayes: Oh, that makes a lot of sense. Will change to be so.

eventualbuddha (commented):
As a side effect of these changes we seem to have lost the ability to download the file in Chrome. I suggest we fix that or completely remove that as an option.

services/scan/src/store.ts
```diff
@@ -946,6 +952,8 @@ export class Store {
...
```
eventualbuddha, line 956: One thing to note is that we're not really using the writable stream API correctly. This can lead to excessive buffering and potentially a much slower export operation due to the memory usage/GC/swapping going on. See https://enlear.academy/nodejs-backpressuring-in-streams-52638f505e1b for what we should be doing differently, specifically waiting for a `drain` event after each `write` that returns `false`.

adghayes:
Working on fixing the Chrome download.

services/scan/src/store.ts
adghayes: Interesting. I'm not sure exactly how it will look - I tried just waiting on the writeStream to drain but we can't do that because we have an ongoing SQL query that we're iterating through, so may have to change the way we're working through the data. Not going to touch that for now.

eventualbuddha (approved):
LGTM once the Chrome download is fixed or disabled.

frontends/precinct-scanner/src/components/export_results_modal.tsx
```diff
@@ -167,6 +167,7 @@ export function ExportResultsModal({
                 <Button
                   data-testid="manual-export"
                   onPress={() => exportResults(true)}
+                  disabled // Not currently supported
```
adghayes, line 170: underlying feature is disabled but also disabling the button

services/scan/src/precinct_scanner_app.ts
```diff
@@ -459,7 +460,7 @@ export async function buildPrecinctScannerApp(
       const cvrFilename = generateFilenameForScanningResults(
         // TODO: Move machine config provider to shared utilities and access
         // actual machine config, with dev overrides, here instead
-        '0000',
+        'NO-ID',
```
adghayes, line 463: Just a clearer non-id if anyone ever ends up using this.

services/scan/src/precinct_scanner_app.ts
adghayes: Had to use Passthrough here, otherwise it didn't work!

services/scan/src/store.ts
```diff
@@ -947,9 +947,10 @@ export class Store {
           : undefined
       );
 
+      // TODO: We should be waiting for the writeStream to drain before
+      // writing more data to it
       if (cvr) {
-        writeStream.write(JSON.stringify(cvr));
-        writeStream.write('\n');
+        writeStream.write(`${JSON.stringify(cvr)}\n`);
```
adghayes, line 953: I don't know if this does anything but it seems better to have fewer rather than more write events.

adghayes:
There are two issues with the way this PR that we will eventually have to circle back to:
1. Downloading CVRs via Chrome no longer works. This can be enabled by changing the endpoint from`POST` to `GET`
2. For some reason, the CVR export doesn't return headers until it's done collecting data, even if we `flushHeaders` this may have to do with the ways in which we're `pipe`ing into the response. For now, it means that the "Custom" location export works it doesn't prompt for the location until after the file is complete.

PR #2692, opened 2022-10-17
fix(precinct-scanner): export CVRs in order of random ballot ID for the precinct scanner

benadida:


services/scan/src/store.ts
```diff
@@ -820,7 +820,7 @@ export class Store {
   /**
    * Exports all CVR JSON data to a stream.
    */
-  exportCvrs(writeStream: Writable): void {
+  exportCvrs(writeStream: Writable, orderBySheetId = false): void {
```
eventualbuddha, line 823: This is going to conflict with @adghayes's change: #2681.

eventualbuddha (approved):

umbernhard (approved):

services/scan/src/store.ts
```diff
@@ -569,6 +569,7 @@ export class Store {
         backFinishedAdjudicationAt ?? null
       );
     } catch (error) {
+      console.log(error);
```
eventualbuddha, line 572:
```suggestion
```

eventualbuddha (approved):

cjerdonek:
Am I right that this is needed because the database may preserve remnants of the insertion order as an implementation detail? (Otherwise, you wouldn't need to impose an order on the `SELECT`.)

If this is the case, could an admin still reconstruct the insertion order from raw queries? If so, is there a way to configure the database to forget (or "scramble") the information of the insertion order on insert?

benadida:
You are correct in your analysis, @cjerdonek. We're not worried about raw queries because on our locked down production machines, no one ever gets to make those. Only interface is via the export.

There are more advanced techniques we will want to consider in the future if someone were to make a raw copy of the hard drive, which again is fairly tricky to do on production machines, so not an immediate concern.

cjerdonek:
Sounds good. Thanks, @benadida!

PR #2693, opened 2022-10-17
fix(admin): handle larger CVR file import

eventualbuddha:
## Overview

CVR files that have a lot of write-in images can be pretty large, i.e. 1GB+. Attempting to read this in as a string is a really bad idea and NodeJS will simply not allow it as it exceeds the maximum length of a string in V8.

This commit fixes the issue by:
a) passing in the file path when importing rather than the contents
b) computing the sha256 hash via a file read stream
c) reading the CVR file one line at a time

We end up reading the file twice (once for the hash, once for the CVR entry import), but overall this is both more efficient and faster even for strings that do fit in memory.

**NOTE:** This does _not_ change the client-side CVR handling which still doesn't seem happy with CVR files this large.

## Demo Video or Screenshot
https://user-images.githubusercontent.com/1938/196287850-24a74875-9a08-48ca-87a5-f5d46345f184.mov

## Testing Plan 
Tested with @carolinemodic's 1.3GB 967-entry CVR file with a bunch of write in images.

## Checklist

### DIFF
--- a/libs/api/src/services/admin/endpoints.ts
+++ b/libs/api/src/services/admin/endpoints.ts
@@ -17,7 +17,7 @@ import {
   OkResponseSchema,
 } from '../../base';
 import {
-  CastVoteRecordFileMetadata,
+  CastVoteRecordFileRecord,
   CastVoteRecordFileRecordSchema,
   ElectionRecord,
   ElectionRecordSchema,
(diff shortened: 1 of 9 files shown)


kofi-q (approved):
LGTM! (but worth a sanity check from someone with more context 🙃)

benadida (approved):

PR #2695, opened 2022-10-17
Portrait Precinct Scanner UI

### LINKED ISSUE #2684: [VxScan] "No USB Drive Detected" buttons are in the wrong order
image

They should be "Cancel" left, "Save" right.

beausmith:
## Overview
A number of fixes and tweaks to make the current VxScan app work in portrait orientation.

## Preview Screen Portrait/Landscape
| landscape | portrait |
| --- | --- |
| image | image |

## Ballot Bag Full
| bag is full | card inserted | button pressed |
| --- | --- | --- |
| image | image | image |

## Update copy to be read from the voter's perspective
image
image

## More Flexible Wrapping Election Info Bar - Insert Ballot Screen Before & After
| - | - |
| --- | --- |
| before | image |
| after | image |
| after long title | image |

## More Flexible Wrapping Election Info Bar - Election Manager Screen Before & After
| - | - |
| --- | --- |
| before | image |
| after | image |
| after long title | image |

## Preview Screen Fix
This preview previously didn't have the testing mode banner at the top: 
image

### Landscape - Before & After
image

| Portrait - Before | Portrait - After |
| --- | --- |
| image | image |

Resolves #2684


mattroe:
@beausmith I'm still seeing the "old" review screen button presentation when I emulate a vertical device in my browser. Other than that, everything else LGTM visually.

beausmith:
@mattroe - How are you emulating a vertical device? Here is how I do it: 
image

mattroe:
@beausmith false alarm, I must have had something wrong in my env setup. looks as expected to me know in the new button layout.

frontends/precinct-scanner/src/screens/scan_error_screen.tsx
```diff
@@ -116,7 +116,7 @@ export function InvalidTestModeBallotPreview(): JSX.Element {
 export function InvalidLiveModeBallotPreview(): JSX.Element {
   return (
     <ScanErrorScreen
-      isTestMode
+      isTestMode={false}
```
arsalansufi, line 119: Nice catch!

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
```diff
@@ -37,6 +38,20 @@ import { AppContext } from '../contexts/app_context';
 import { toSentence } from '../utils/to_sentence';
 import { useSound } from '../hooks/use_sound';
 
+const ResponsiveButtonParagraph = styled.p`
+  @media (orientation: portrait) {
```
arsalansufi, line 42: Any reason we can't apply these styles to both portrait and landscape? Or alternatively, could we tweak the styles here such that they work for both portrait and landscape?

arsalansufi (commented):
Amazing! Thanks for breaking this up into clear commits! Just one q

frontends/precinct-scanner/src/screens/scan_warning_screen.tsx
beausmith: We can apply this styles to both orientations, however if we apply these styles to the screen the buttons will become as wide as the widest text block on the page:
image

These styles were added to "tweak" the styles so that the button layout looked good on portrait as they already looked good on landscape.

What is the problem that you're hoping to solve?

arsalansufi: Got it!

Not at all a problem to have media queries that target specific orientations! In this case, the desired layouts in portrait vs. landscape just struck me as the same, so it seemed possible to accomplish what we wanted with the same styles in all orientations. Good to know about the widening challenge specific to landscape, though, which I now see was the reason that we needed to apply these styles behind a media query.

arsalansufi (approved):
LGTM! 🖼️

PR #2696, opened 2022-10-18
Improve Performance + Safety of Large CVR Export

adghayes:
Small tweaks to follow #2681. Three commits:
1. Safely waits for writeStream to be ready to consume data before writing to it
2. Passes electionDefinition as an argument to stop hitting the DB (this the time of the `exportCvrs` method from about 8 to 2 seconds when not exporting write-ins)
3. Gets rid of intermediary Passthrough stream.

This fixes the problem where the file-picker doesn't come up write away AND it means the `exportCvrs` method is no longer blocking.

### DIFF
--- a/services/scan/src/store.ts
+++ b/services/scan/src/store.ts
@@ -940,14 +940,20 @@ export class Store {
           {
             interpretation: frontInterpretation,
             contestIds: isHmpbPage(frontInterpretation)
-              ? this.getContestIdsForMetadata(frontInterpretation.metadata)
+              ? this.getContestIdsForMetadata(
+                  frontInterpretation.metadata,
+                  electionDefinition
+                )
               : undefined,
             markAdjudications: frontAdjudications,
           },
           {
             interpretation: backInterpretation,
             contestIds: isHmpbPage(backInterpretation)
-              ? this.getContestIdsForMetadata(backInterpretation.metadata)
+              ? this.getContestIdsForMetadata(
+                  backInterpretation.metadata,
+                  electionDefinition
+                )
               : undefined,
             markAdjudications: backAdjudications,
           },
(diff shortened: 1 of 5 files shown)


services/scan/src/store.ts
```diff
@@ -1045,17 +1060,21 @@ export class Store {
...
+    return this.getBallotPageLayoutsForMetadata(
+      metadata,
+      electionDefinition
+    ).find((layout) => layout.metadata.pageNumber === metadata.pageNumber);
   }
 
   getBallotPageLayoutsForMetadata(
-    metadata: BallotMetadata
+    metadata: BallotMetadata,
+    electionDefinitionFromProps?: ElectionDefinition
```
arsalansufi, line 1074: Instead of call this `electionDefinitionFromProps` (since props are more of a React-component-specific term), we could eliminate the need for a different name by specifying the fallback up here:

```ts
getBallotPageLayoutsForMetadata(
  metadata: BallotMetadata,
  electionDefinition: ElectionDefinition = this.getElectionDefinition(),
): BallotPageLayout[] {
```

arsalansufi (approved):
One comment then LGTM!

services/scan/src/store.ts
adghayes: I tried and was stuck in linting purgatory:

Screen Shot 2022-10-18 at 8 41 14 AM


Screen Shot 2022-10-18 at 8 41 36 AM


Screen Shot 2022-10-18 at 8 42 15 AM

@arsalansufi I think the ideal solution would be to cache values like this rather than hit the database and parse it every time, long-term.

eventualbuddha: This should work, @adghayes:

```ts
  getBallotPageLayoutsForMetadata(
    metadata: BallotMetadata,
    electionDefinition = this.getElectionDefinition()
  ): BallotPageLayout[] {
```

PR #2701, opened 2022-10-18
fix(admin): clear uploads after request

eventualbuddha:
## Overview

Previously we were storing uploads in a temporary file in `/tmp`, but they were never getting cleaned up. Eventually the `/tmp` partition would run out of disk space and we'd be unable to upload anything new, and other operations that used `/tmp` would start to fail as well.

Now, we upload to the `uploads` folder within the current workspace. After the request finishes we delete the uploaded file.

## Demo Video or Screenshot
Watching the `uploads` directory during the upload of a 1.3GB CVR file:
```
while true; do du -a ~/code/vxsuite/services/admin/dev-workspace/uploads/ | sort -rn | head -n 5; sleep 1; done
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
4	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
90388	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
90384	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/8ceb0164eb7b57a59a414772a3ebdb83
302964	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
302960	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/8ceb0164eb7b57a59a414772a3ebdb83
508548	/home/brian/code/vxsuite/services/admin/dev-workspace/uploads/
[...]

PR #2702, opened 2022-10-18
fix(scan): clear uploads after request

eventualbuddha:
## Overview

Similar as #2701, but for `services/scan`. Also clears uploads on startup, and adds the same for `services/admin`.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tested manually similarly to how I did it in #2701.

## Checklist

### DIFF
--- a/services/scan/src/central_scanner_app.ts
+++ b/services/scan/src/central_scanner_app.ts
@@ -1,29 +1,30 @@
 import { Scan } from '@votingworks/api';
-import express, { Application } from 'express';
-import multer from 'multer';
-import { Buffer } from 'buffer';
 import {
-  safeParseElectionDefinition,
-  safeParse,
-  safeParseJson,
-  BallotPageLayoutSchema,
   BallotPageLayout,
+  BallotPageLayoutSchema,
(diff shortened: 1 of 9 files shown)


services/scan/src/central_scanner_app.ts
```diff
@@ -142,27 +148,35 @@ export async function buildCentralScannerApp({
...
+                message: 'expected file field to be named "package"',
+              },
+            ],
+          });
+          return;
+        }
 
-      const pkg = await readBallotPackageFromBuffer(await readFile(file.path));
+        const pkg = await readBallotPackageFromBuffer(
+          await readFile(file.path)
```
jonahkagan, line 166: I noticed you switched to `fs.readFile` in precinct scanner app, but not here. Should we do that here too?

jonahkagan (approved):

services/scan/src/central_scanner_app.ts
eventualbuddha: https://github.com/votingworks/vxsuite/pull/2707

PR #2709, opened 2022-10-19
fix(ps): backup cvr stream blockage

adghayes:
In #2696, we made it so we only write to the CVR stream when it is ready to write (it drains). This improves performance when we're writing to a `response`.

However, in many cases, we are writing to stream which is not being piped anywhere. In other words, it is not being drained, so there is no `drain` event. Thus, it hangs. When we use streams like this (e.g. `new WritableStream()`) we need to set the `highWaterMark` to something that is larger than the expected data size so that it never blocks.

In this PR, I both set the `highWaterMark` and `pipe` the export stream directly into the zip file stream (via a `Passthrough`) which improves efficiency.

Also left some useful debugging in there.

### DIFF
--- a/services/scan/src/backup.ts
+++ b/services/scan/src/backup.ts
@@ -1,12 +1,12 @@
 import { Buffer } from 'buffer';
 import makeDebug from 'debug';
 import { createReadStream, existsSync } from 'fs-extra';
-import { WritableStream } from 'memory-streams';
 import { basename } from 'path';
 import Database from 'better-sqlite3';
 import { fileSync } from 'tmp';
 import ZipStream from 'zip-stream';
 import { FULL_LOG_PATH } from '@votingworks/logging';
+import { PassThrough } from 'stream';
 import { Store } from './store';
 
 const debug = makeDebug('scan:backup');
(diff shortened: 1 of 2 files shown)


adghayes:
@carolinemodic I never got this to work locally but I am 90% sure that is unrelated, and is instead related to the way I'm trying to replicate your `dev-workspace`. I know that, through logs, the issue identified above is fixed.

services/scan/src/backup.ts
```diff
@@ -85,24 +85,30 @@ export class Backup {
...
 
+    debug('adding election.json to backup...');
     await this.addEntry('election.json', electionDefinition.electionData);
+    debug('added election.json to backup');
 
-    const cvrStream = new WritableStream();
-    await this.store.exportCvrs(cvrStream);
-    await this.addEntry('cvrs.jsonl', cvrStream.toBuffer());
+    debug('adding CVRs to backup...');
+    const cvrStream = new PassThrough({ highWaterMark: 100000000 }); // 100MB
```
eventualbuddha, line 93: I tested this without the `highWaterMark` option and it worked, at least on a small backup.

eventualbuddha: If we keep this, we can at least use the thousands separator:
```suggestion
    const cvrStream = new PassThrough({ highWaterMark: 100_000_000 }); // 100MB
```

benadida:
Assuming `highWaterMark` is necessary... is 100MB enough?

adghayes:
I'll remove the `highWaterMark` because it's not necessary.

If we weren't using a `Passthrough` stream (which handles piping under the hood for us, I believe), then we would need to set the `highWaterMark` to something like several gigabytes.

As a follow-up, I think we should rewrite the CVR export to return a Readable Stream rather than accept a Writable stream. We would have issues with who we provide as the receiver. Per the node docs:

> If the data to be written can be generated or fetched on demand, it is recommended to encapsulate the logic into a [Readable](https://nodejs.org/api/stream.html#class-streamreadable) and use [stream.pipe()](https://nodejs.org/api/stream.html#readablepipedestination-options). However, if calling write() is preferred, it is possible to respect backpressure and avoid memory issues using the ['drain'](https://nodejs.org/api/stream.html#event-drain) event:

eventualbuddha (approved):

Issue #1926, opened 2022-06-02
PIN Uniqueness


arsalansufi:
Smartcard PINs are randomly generated today, but are not unique, which is a requirement per the NIST security standard.

Potential steps:
- [ ] Set up services/admin endpoints for unique PIN generation (could use a SQLite persisted bloom filter)
- [ ] Have UI use these endpoints

arsalansufi:
Copying from that Slack thread the relevant investigation that kicked off this realization:

VVSG points to NIST SP 800-63B:
> 11.3.2-B: The voting system must, at minimum, meet the password complexity requirements within the latest version of NIST SP 800-63B Digital Identity Guidelines standards.

In [NIST SP 800-63B](https://nvlpubs.nist.gov/nistpubs/specialpublications/nist.sp.800-63b.pdf), Java Cards with PINs fall under `5.1.8 Multi-Factor Cryptographic Software`:
> The multi-factor software cryptographic authenticator is _something you have_, and it SHALL be activated by either _something you know_ or _something you are_.

NIST then says:
> Multi-factor software cryptographic authenticators encapsulate one or more secret keys **unique** to the authenticator and accessible only through the input of an additional factor, either a memorized secret or a biometric.

"unique" refers to the secret key on the Java Card, not the PIN. The PIN is the "memorized secret." NIST then says about the memorized secret:
> Any memorized secret used by the authenticator for activation SHALL be a randomly-chosen numeric value at least 6 decimal digits in length or other memorized secret meeting the requirements of Section 5.1.1.2

The above doesn't mention uniqueness nor does Section 5.1.1.2, which just says:
> Memorized secrets that are randomly chosen by the CSP (e.g., at enrollment) or by the verifier (e.g., when a user requests a new PIN) SHALL be at least 6 characters in length and SHALL be generated using an approved random bit generator [SP 800-90Ar1].
> ...
> When processing requests to establish and change memorized secrets, verifiers SHALL compare the prospective secrets against a list that contains values known to be commonly-used, expected, or compromised.

PR #2714, opened 2022-10-19
[services/admin] Add PIN generation endpoint

kofi-q:
## Overview
https://github.com/votingworks/vxsuite/issues/1926:
Adding `/admin/pins/generate-card-pin` endpoint (currently unused) that returns a randomly generated PIN via the `@votingworks/utils` `generatePin` method.

Copied over dotenv parsing code from `services/scan` - will look into consolidating later, so we're not duplicating that code so often.

Future PRs will add storage for already generated PINs and checks for uniqueness.

### With "all zeroes" dev flag disabled:
Screenshot 2022-10-19 at 2 29 45 PM

### With "all zeroes" dev flag enabled:
Screenshot 2022-10-19 at 2 44 21 PM

## Testing Plan 
- Added (mostly stub) unit test for the new endpoint
- Verified locally with manual HTTP requests

## Checklist


libs/api/src/base.ts
```diff
@@ -8,12 +8,9 @@ export const OkResponseSchema: z.ZodSchema<OkResponse> = z.object({
   status: z.literal('ok'),
 });
 
-export interface ErrorsResponse {
-  status: 'error';
-  errors: Array<{ type: string; message: string }>;
-}
-
-export const ErrorsResponseSchema: z.ZodSchema<ErrorsResponse> = z.object({
+export const ErrorsResponseSchema = z.object({
```
kofi-q, line 11: had to take out the template param and let type inference do its thing to get the `discriminatedUnion` working - but, let me know if this pattern of deriving the response type from the schema breaks anything (functional or ideological)

libs/api/src/services/admin/endpoints.ts
```diff
@@ -574,3 +574,35 @@ export type GetWriteInImageResponse = WriteInImageEntry[] | ErrorsResponse;
...
+ * @method POST
+ */
+export type GenerateCardPinRequest = z.TypeOf<
+  typeof GenerateCardPinRequestSchema
+>;
+
+/**
+ * Schema for {@link GenerateCardPinResponse}.
+ */
+export const GenerateCardPinResponseSchema = z.discriminatedUnion('status', [
```
eventualbuddha, line 594: I wasn't aware of [`discriminatedUnion`](https://github.com/colinhacks/zod#discriminated-unions)! Sounds like it solves one of the main gripes I have with zod's error messages in that they so often have a bunch of bogus info about the variants that don't match the discriminator.

libs/api/src/services/admin/endpoints.ts
```diff
@@ -574,3 +574,35 @@ export type GetWriteInImageResponse = WriteInImageEntry[] | ErrorsResponse;
...
+  ErrorsResponseSchema,
+]);
+
+/**
+ * @url /admin/pins/generate-card-pin
+ * @method POST
+ */
+export type GenerateCardPinResponse = z.TypeOf<
+  typeof GenerateCardPinResponseSchema
+>;
```
eventualbuddha, line 608: I haven't used the inferred types from the zod schema because the type values shown by TS when hovering or in error messages loses the type names. See [this TS playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcDkyEAJvgNwBQFApgB6SxwDGEAdgM7wCiUOU7AJWrtIHagGUmAC2ogAhnAC8KAHQQARgCtqTGAAoEFOHE5yYAV3YAuVQBtgMalDm29+J33wBKADRG4HtDWqnK8cgCeeshqWjr6SDDhYNQ20ZxQwKwA5nq+cCDC7HJZKarpmTle6F6+FGhelDT00PCJyXA8fILCouzUSqoAKknUAPKYADxt1BCYHbxBQiJsfZIy8gB8jQD0AFS7Rrtwa7IKmNCIAAL2rADWcADi1KxOZtQAwqHEAAqZS73UNAqQ7bJoMeAsDjwJ4vZyOT5QH5-HorCTSU4DaLmVjANh6ADa-miGm0ugM-mMpgswWi9kczlc+Agt28fmMxjAmVSKnK2VybOqAs6ixRYhO8j8AF0GlQ9gc4EdLuYoLY4Ns5MQQJltpyONsSrC3gBaJhfI26w5wS4FGBSEhwb6jcSDEFglpwaaPZ6veFfX6sf6ozEqYbJcYTfzTWZew2+xH+wNi9HyChbWX7S1KlVqjVa1g6zLsfXeuHUE1mi0Kq02u3EB1Ol0K0F0cEekZwACCthgMJ9Hz9yOWYiU-gAPogKeyqZYbEyWZR2ezdTZeVkF+y0GP5l1E30dhmq+KzhcENdMvcuz2S28EUiA6K+kDXS33ZDOJ3u73S7eEw+0esFGULEcTxQljGJWIyUMRdpxpFQ6VeRlmVZSdlzKGAMj5WpjHqIUFn4Xd-1OKUZQoYgdFsUJ+jfeAoD-GwvxvAd7yHPcKG2bZFy49kAD0AH4qHIphKLo5gVngFwYEImxL0YuM70IygOO47j+KAA) which compares the displayed type of this form vs the original form. I agree that this would be better if not for this problem:

image

services/admin/src/server.pins.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  };
+});
+
+let app: Application;
+let workspace: Workspace;
+
+beforeEach(() => {
+  mockOf(generatePin).mockReset();
+
+  workspace = createWorkspace(dirSync().name);
```
eventualbuddha, line 24: With our recent battles with `/tmp` disk space usage I've been trying to be more aware of when we're not cleaning things up. Add `fs.rm(workspace.path, { recursive: true })` to an `afterEach`?

services/admin/src/server.pins.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  app = buildApp({ workspace });
+});
+
+afterAll(() => {
```
eventualbuddha, line 28: Does this need to be `afterAll`? What is `restoreAllMocks` restoring?

services/admin/src/server.ts
```diff
@@ -568,6 +568,20 @@ export function buildApp({ workspace }: { workspace: Workspace }): Application {
     }
   });
 
+  app.post<
+    NoParams,
+    Admin.GenerateCardPinResponse,
+    Admin.GenerateCardPinRequest
+  >('/admin/pins/generate-card-pin', (_request, response) => {
```
eventualbuddha, line 575: I'd do this like so to be more standard REST and as PINs should be unique per-election:

```suggestion
  >('/admin/:electionId/pins', (_request, response) => {
```

services/admin/src/server.pins.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  jest.restoreAllMocks();
+});
+
+describe('POST admin/pins/generate-card-pin', () => {
```
eventualbuddha, line 32: Doesn't hurt to have it, but we typically don't use `describe` unless we really want to separate things within a file. Even then, it's probably best to split into separate files.

eventualbuddha (commented):
Looks like a good start, though there are a few things it's probably worth changing.

services/admin/src/server.ts
kofi-q: ah, i didn't think to ask about whether we need to scope uniqueness to a single election! how does that tie in with system admin cards being (IIRC) election-agnostic?

kofi-q: cc @arsalansufi, in case this came up in earlier discussions about unique PINs

arsalansufi: Ah yes great q! The original intent was to have the uniqueness be scoped to a single election. (Or rather, that's all we need for VVSG2. There's no requirement that PINs be unique across elections, though maybe @mattroe can confirm.)

In terms of how this interplays with sys admin cards being election-agnostic, the ability to program election manager and poll worker cards is only made visible once a sys admin has configured the machine for a specific election, so we'll have an election ID available. (Zooming out, the sys admin configures the election definition on VxAdmin and programs election manager cards for that specific election. Election managers then use their cards to configure the other machines.)

eventualbuddha: If we're going to need to call this endpoint outside the context of an election for sysadmin cards, then I can see why we'd want to leave it as you had it. Product-wise as long as we're okay with PINs being unique across all elections on a machine then this works as-is. @mattroe?

arsalansufi: > If we're going to need to call this endpoint outside the context of an election for sysadmin cards

Ah yes of course - forgot about that case

kofi-q:
Following clarifying team discussion, https://github.com/votingworks/vxsuite/issues/1926 (PIN uniqueness) is not a real requirement for VVSG cert, so closing this out! ✂️

PR #2718, opened 2022-10-20
fix additional report state

adghayes:
Fixes issue where "Additional Reports" are the opposite of what they should be - Opened where it should be Closed, vice versa.


carolinemodic (approved):
as a followup might be good to add a test for this if possible

benadida (approved):

benadida:
Could use a test, but good for now.

adghayes:
Agree. A lot of this will be updated with `sus-polls` so will make sure to cover there.

PR #2719, opened 2022-10-20
print correct report in more states

adghayes:
#2718 fixes the issue but could fail as a fix if the execution order / state updates are different. this is an improvement.


carolinemodic (approved):
tested locally on my VxDev

Issue #1531, opened 2022-03-09
Rethink Printing across all of our apps 


benadida:
We have struggled with various bugs and uncertainties in our printing strategies. A number of different ideas around printing have been brainstormed. We should spend some time trying different options and determining what the most robust and easy to develop with strategy for printing is and implement it across our various printing features. 

--- Details from a previous ticket --- 
Printing ballots in VxAdmin (and likely VxMark) depends on preventing React re-renders. It shouldn't.
Best example is in printing the test deck in VxAdmin.

We used PagedJS to render a bunch of ballots, and we are counting on that not being re-rendered before triggering a print. It's a bit tricky to solve this because PagedJS uses the DOM, so it's not obvious how to make that asynchronous work survive cleanly across re-renders.

eventualbuddha:
Some discussion on the topic: https://votingworks.slack.com/archives/CEL6D3GAD/p1646834507614739.

carolinemodic:
Related bug that we should aim to solve as a part of this ticket: https://github.com/votingworks/vxsuite/issues/2299

carolinemodic:
We will also want to rethink the print time estimates in tackling this: https://github.com/votingworks/vxsuite/issues/1865

adghayes:
@carolinemodic I think that the two other tickets closed in favor of this ticket are separate and not solved by, for example, the solution that @jonahkagan laid out recently. If you are thinking of this as a larger ticket, then grouping them is understandable.

The double print issue in #2299 is actually a Linux / CUPS level behavior, and won’t be affected by how we trigger printing in the app. The current ticket might prevent *other* cases of double-printing. The timing issues in #1865 would require getting print job ID’s from lp (instead of lpr) and tracking those in the app somehow. Definitely related, but different.

carolinemodic:
Yeah I'm thinking of this as a larger/broader ticket. 
Related: https://github.com/votingworks/vxsuite/issues/2289

carolinemodic:
Potential idea and demo from @jonahkagan https://votingworks.slack.com/archives/CEL6D3GAD/p1660580505853769

adghayes:
It seems like we've established an approach in #2730, so here is the task list for this project AFAICT:

- [ ] #2730
- [ ] #2758 (< 1 day)
- [ ] #2759 (~ 4 days)
- [ ] #2760 (~ 3 days)
- [ ] #2761 (3 - 6 days)

adghayes:
[...]

PR #2721, opened 2022-10-20
Print Element Utility + First Use Case

adghayes:
Begins #1531 by creating a tool to print a `JSX` element directly. Currently, we employ two approaches:

- have document persistently rendered to page, out of view, and print on button push
- render document and print as a side effect

With this tool, you pass it a `JSX` element or a function which returns a `JSX` element with a callback. It will handle placing the element on the page, printing it at the appropriate time, and removing it from the page.

In preparation for votingworks/vxsuite-roadmap#37, I used the new tool for printing polls open and closed reports on VxMark and VxScan.


adghayes:
Opening PR on another branch.

PR #2726, opened 2022-10-20
fix(wia): adjust contest index based on the previous page's number of contests

benadida:
In locating the proper portions of the ballot for write-in adjudication, we go through each page's layout and find the current contest's index on its current page. To do that, we subtract from the index the contests that appear on prior pages.

There was a bug where we were subtracting the number of contests on the *current* page, rather than the previous page. This wasn't noticed before because we happened to be testing on an election that had the same number of contests on the front and back pages.

### DIFF
--- a/libs/ballot-interpreter-nh/src/layout.ts
+++ b/libs/ballot-interpreter-nh/src/layout.ts
@@ -309,10 +309,20 @@ export function generateBallotPageLayouts(
       });
     }
 
+    // ensure the layouts are returned in the contest-order defined by the election definition
+    const contestsInOrder: BallotPageContestLayout[] = election.contests
+      .map((electionContest) =>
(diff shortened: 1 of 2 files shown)


arsalansufi (approved):
Nice catch!

eventualbuddha (approved):
I think this logic might have been copied from elsewhere, and if so it'd need fixing there in a follow up. @carolinemodic?

carolinemodic:
I just copied the logic from the nh-demo branch, where it was in the frontend instead of the backend, but now it only exists in the backend so you shouldn't need to update anywhere else.

benadida:
> I think this logic might have been copied from elsewhere, and if so it'd need fixing there in a follow up. @carolinemodic?

I didn't see anything like it anywhere else, at least not with the same variable names.

Issue #2727, opened 2022-10-20
Improve tests for WIA issues


benadida:
We fixed two issues without regression tests because those tests would require publishing ballots that are not yet public information. Once they are public information, we can write these tests a lot more easily.

Specifically:
- the bug in `services/admin` that was improperly computing `contestIdx` in the write-in-adjudication image loading.
- the bug in`ballot-interpreter-nh` that was returning contest layout information in potentially the wrong order.

PR #2728, opened 2022-10-20
fix(ballot-interpreter-nh): ensure the contest layout bounds are generated in right order

benadida:
Sometimes, `generateBallotPageLayouts` outputs contest layouts in an order different from the election definition. That messes up write-in adjudication downstream.

I considered alternate fixes:
- fixing the election definition converter so that it would always yield grid layouts in an order that would yield the right contest layout downstream, but that seemed a bit flimsy.
- changing the data structure to a map keyed on `contestId`, but that seems like a large, risky, and unnecessary change when we already often rely on contest ordering identical to the election definition.

### DIFF
--- a/libs/ballot-interpreter-nh/src/layout.ts
+++ b/libs/ballot-interpreter-nh/src/layout.ts
@@ -309,10 +309,20 @@ export function generateBallotPageLayouts(
       });
     }
 
+    // ensure the layouts are returned in the contest-order defined by the election definition
+    const contestsInOrder: BallotPageContestLayout[] = election.contests
+      .map((electionContest) =>
+        contests.find((c) => c.contestId === electionContest.id)
+      )
+      .filter(
+        (contestLayout): contestLayout is BallotPageContestLayout =>
+          contestLayout !== undefined
+      );
+
     layouts.push({
       metadata: { ...metadata, pageNumber: i + 1 },
       pageSize: geometry.canvasSize,
-      contests,
+      contests: contestsInOrder,
     });
   }
 


benadida:
hmmm, ok, I might have a simple and more functional tweak to this... will do that soon.

benadida:
ok, more functional now.

jonahkagan (approved):
Can you add a regression test?

benadida:
Regression test coming in a little bit #2727

libs/ballot-interpreter-nh/src/layout.ts
```diff
@@ -309,10 +309,20 @@ export function generateBallotPageLayouts(
...
 
+    // ensure the layouts are returned in the contest-order defined by the election definition
+    const contestsInOrder: BallotPageContestLayout[] = election.contests
+      .map((electionContest) =>
+        contests.find((c) => c.contestId === electionContest.id)
+      )
+      .filter(
+        (contestLayout): contestLayout is BallotPageContestLayout =>
+          contestLayout !== undefined
+      );
```
eventualbuddha, line 320: Another solution ⛳ 
```ts
    const contestsInOrder = [...contests].sort(
      (a, b) =>
        election.contests.findIndex(({ id }) => id === a.contestId) -
        election.contests.findIndex(({ id }) => id === b.contestId)
    );
```

benadida: ahah nice, that is more elegant :) for a future patch!

PR #2717, opened 2022-10-20
refactor(admin): move `isOfficialResults` to backend

eventualbuddha:
## Overview

Stores the flag in the `elections` table and adds an endpoint to allow setting it. Automatically resets the flag when deleting CVR files. Moves anything related to the election metadata out of `useElectionManagerStore` and into separate hooks.

Refs #2713

## Demo Video or Screenshot
Behavior is unchanged.

## Testing Plan
Manually tested:
- loading a new election
- resetting the election
- loading CVRs
- marking results as official and seeing reports be "Official"
- clearing results and seeing reports revert to "Unofficial"

## Checklist
- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.
- [ ] I have added JSDoc comments to any newly introduced exports

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #2729, opened 2022-10-20
refactor(admin): move printed ballots to backend

eventualbuddha:
## Overview

Follow-up to https://github.com/votingworks/vxsuite/pull/2717.

- replaces `AppContext` with hooks for adding/getting printed ballot records
- adds `services/admin` endpoints for managing printed ballot records & associated `react-query` hooks
- removes duplicate `BallotLocales` interface
- tracks all ballots printed, not just official ballots, but only includes official ballots in reports

## Demo Video or Screenshot
n/a

## Testing Plan 
Manual testing of printing ballots, verifying the UI updates appropriately, and the counts are correct.

## Checklist


frontends/election-manager/src/lib/backends/backends.test.ts
```diff
@@ -253,6 +285,55 @@ function makeAdminBackend(): ElectionManagerStoreAdminBackend {
...
+      return {
+        body: typedAs<Admin.GetPrintedBallotsResponse>({
+          status: 'ok',
+          printedBallots: await dbEntry.memoryBackend.loadPrintedBallots(
+            parseQueryResult.ok()
+          ),
+        }),
+      };
+    })
+    .post('glob:/admin/elections/*/printed-ballots', async (url, request) => {
```
kofi-q, line 319: general question on this pattern: as an uninitiated engineer, it seems like unnecessary overhead to model server-side logic (and db data/side effects) in these tests vs the minimum of verifying that the client makes the expected HTTP requests. what's the team philosophy/context behind doing it this way?

kofi-q (approved):
🚀

frontends/election-manager/src/lib/backends/backends.test.ts
eventualbuddha: I'm not sure we have a strong/consistent philosophy on this. In general my preference is to use simplified "real" versions of any complicated component such as a "backend" (see also `MemoryStorage` and `MemoryCard`). My rationale for doing so is that it is often convenient to have such versions of the component and it often promotes better testing methodology. The alternatives also often involve opaque fixtures whose values are hardcoded mysteries.

However, I recognize that _this_ case may be taking that too far. It was easy at first, but it's become more difficult to maintain. Ditching the "storage backend" in #2717 improved things in that regard. However, I don't think we actually plan to have a demo version of this running in a browser somewhere for the general public to try, so it may be "unnecessary overhead" as you put it. I'd welcome trying out something with less overhead as long as it also avoids the "hardcoded mysteries" problem.

kofi-q: ah, the `MemoryStorage`/`MemoryCard` examples are helpful! i was going to say that larger fakes like that have been useful in my experience for more behavioural testing that involves complex interactions with another component or user input, vs simple unit tests, so that's in line with what i'm seeing with those.

agreed that this particular case (at least post-refactor) could be simplified, since we can test each unit independently, but i'll see if that thought holds after i've spent a bit more time in that code!

Issue #2687, opened 2022-10-17
VxScan Ballot Batching


adghayes:
See the [Tech Spec](https://docs.google.com/document/d/15mSk4QPQUqsLlHFwF5Dw058OsprmoJixwxm7C2GmGGk/edit?usp=sharing) for the overall https://github.com/votingworks/vxsuite-roadmap/issues/37.

Fix batching on the precinct-scanner so that everything is one batch until a) polls are suspended or b) the ballot bag is replaced.

PR #2751, opened 2022-10-27
Pause Polls on VxScan

### LINKED ISSUE #2683: Allow Polls Suspension on VxScan
See the [Tech Spec](https://docs.google.com/document/d/15mSk4QPQUqsLlHFwF5Dw058OsprmoJixwxm7C2GmGGk/edit?usp=sharing) for the overall [roadmap item](https://github.com/votingworks/vxsuite-roadmap/issues/37).

Implement polls suspension on VxScan, including:
- Creating Polls State type(s)
- Managing polls state on the "backend"
- Updating poll worker screens
- Updating idle screens

This chunk of work will not update the printed reports or reports saved to the tally, which is tracked in a separate ticket.

adghayes:
## Overview
Closes #2683. Changes two main things about the election flow:
1. After polls are closed, polls cannot be re-opened. Poll Workers will not be able to re-open the polls and Election Managers will not be able to re-open the polls without reconfiguring first. System Administrators have an option to reset polls from closed to paused.
2. Adds polls paused (or "suspended" according to VVSG language) state. In this state, voting cannot take place but polls _can_ be re-opened.

We also place more limits on when election managers can change the precinct on a device. Formerly, they could change the precinct at any time. Now, they can change it freely while polls are not open. While polls are open but no ballots are not cast, they can still change the precinct but it will require a modal confirmation, reset the polls to unopened, and reprint the polls report.

Some things that are _not_ in this change:
- The new polls transition types (paused, unpaused) will _not_ be represented on the printed reports or in VxMark in any way. That will come in the next two PRs (VxMark and reports respectively).
- Testing for the previous bullet, since it doesn't exist yet

## Demo Video or Screenshot
Poll Worker Flow:
[all_states_in_flow.webm](https://user-images.githubusercontent.com/37960853/198367053-1c91aea3-db32-4786-8108-5abad0e252e2.webm)

Sys Admin Reset Flow:
[reset_polls_paused.webm](https://user-images.githubusercontent.com/37960853/198367073-9438d51c-2b4f-45a1-89ea-a00fc89b6f30.webm)

Election Manager Precinct Change Flow:
[change_precinct.webm](https://user-images.githubusercontent.com/37960853/198367157-b270dec1-37c4-473e-a49b-ce5ecc58cdb8.webm)

## Testing Plan
I refactored a lot of our testing in the frontend and wrote a good bit of new testing. Some backend and utility testing too. Manual testing.

## Checklist

### DIFF
--- a/frontends/precinct-scanner/src/app_root.tsx
+++ b/frontends/precinct-scanner/src/app_root.tsx
@@ -356,7 +356,7 @@ export function AppRoot({
     await config.setPollsState(newPollsState);
   }, []);
 
-  const resetPollsToClosed = useCallback(async () => {
+  const resetPollsToPaused = useCallback(async () => {
     await updatePollsState('polls_paused');
   }, [updatePollsState]);
 
@@ -460,7 +460,7 @@ export function AppRoot({
           }
           showResetPollsToPausedButton
           resetPollsToPaused={
-            pollsState === 'polls_closed_final' ? resetPollsToClosed : undefined
+            pollsState === 'polls_closed_final' ? resetPollsToPaused : undefined
           }
           isMachineConfigured={Boolean(electionDefinition)}
           usbDriveStatus={usbDriveDisplayStatus}
--- a/frontends/precinct-scanner/src/app_root.tsx
+++ b/frontends/precinct-scanner/src/app_root.tsx
@@ -4,11 +4,11 @@ import makeDebug from 'debug';
 
 import {
   Card,
-  OptionalElectionDefinition,
   Provider,
   MarkThresholds,
   ElectionDefinition,
   PrecinctSelection,
+  PollsState,
 } from '@votingworks/types';
 import {
   useCancelablePromise,
@@ -41,7 +41,7 @@ import * as scanner from './api/scan';
 import { usePrecinctScannerStatus } from './hooks/use_precinct_scanner_status';
 import { ElectionManagerScreen } from './screens/election_manager_screen';
 import { InvalidCardScreen } from './screens/invalid_card_screen';
-import { PollsClosedScreen } from './screens/polls_closed_screen';
+import { PollsNotOpenScreen } from './screens/polls_not_open_screen';
 import { PollWorkerScreen } from './screens/poll_worker_screen';
 import { InsertBallotScreen } from './screens/insert_ballot_screen';
 import { ScanErrorScreen } from './screens/scan_error_screen';
@@ -79,12 +79,12 @@ interface HardwareState {
 interface ScannerConfigState {
   electionDefinition?: ElectionDefinition;
   precinctSelection?: PrecinctSelection;
+  pollsState: PollsState;
   currentMarkThresholds?: MarkThresholds;
   isTestMode: boolean;
 }
 
 interface StoredFrontendState {
-  isPollsOpen: boolean;
   ballotCountWhenBallotBagLastReplaced: number;
   isSoundMuted: boolean;
 }
@@ -111,10 +111,10 @@ const initialScannerConfigState: Readonly<ScannerConfigState> = {
   isTestMode: false,
   currentMarkThresholds: undefined,
   precinctSelection: undefined,
+  pollsState: 'polls_closed_initial',
 };
 
 const initialStoredFrontendState: Readonly<StoredFrontendState> = {
-  isPollsOpen: false,
   ballotCountWhenBallotBagLastReplaced: 0,
   isSoundMuted: false,
 };
@@ -139,16 +139,12 @@ type AppAction =
     }
   | { type: 'resetPollsToClosed' }
   | {
-      type: 'updateElectionDefinition';
-      electionDefinition: OptionalElectionDefinition;
-    }
-  | {
       type: 'refreshConfigFromScanner';
       scannerConfig: ScannerConfigState;
     }
   | { type: 'updatePrecinctSelection'; precinctSelection: PrecinctSelection }
   | { type: 'updateMarkThresholds'; markThresholds?: MarkThresholds }
-  | { type: 'togglePollsOpen' }
+  | { type: 'updatePollsState'; pollsState: PollsState }
   | { type: 'toggleIsSoundMuted' }
   | { type: 'ballotBagReplaced'; currentBallotCount: number }
   | { type: 'setMachineConfig'; machineConfig: MachineConfig };
@@ -171,17 +167,10 @@ function appReducer(state: State, action: AppAction): State {
         ...action.storedFrontendState,
         initializedFromStorage: true,
       };
-    case 'updateElectionDefinition':
-      return {
-        ...state,
-        electionDefinition: action.electionDefinition,
-        isPollsOpen: false,
-        ballotCountWhenBallotBagLastReplaced: 0,
-      };
     case 'resetPollsToClosed':
       return {
         ...state,
-        isPollsOpen: false,
+        pollsState: 'polls_closed_initial',
         ballotCountWhenBallotBagLastReplaced: 0,
       };
     case 'refreshConfigFromScanner': {
@@ -195,17 +184,16 @@ function appReducer(state: State, action: AppAction): State {
       return {
         ...state,
         precinctSelection: action.precinctSelection,
-        isPollsOpen: false,
       };
     case 'updateMarkThresholds':
       return {
         ...state,
         currentMarkThresholds: action.markThresholds,
       };
-    case 'togglePollsOpen':
+    case 'updatePollsState':
       return {
         ...state,
-        isPollsOpen: !state.isPollsOpen,
+        pollsState: action.pollsState,
       };
     case 'toggleIsSoundMuted':
       return {
@@ -242,7 +230,7 @@ export function AppRoot({
     isTestMode,
     precinctSelection,
     currentMarkThresholds,
-    isPollsOpen,
+    pollsState,
     initializedFromStorage,
     machineConfig,
     isSoundMuted,
@@ -282,6 +270,7 @@ export function AppRoot({
     const newPrecinctSelection = await makeCancelable(
       config.getPrecinctSelection()
     );
+    const newPollsState = await makeCancelable(config.getPollsState());
     const newMarkThresholds = await makeCancelable(config.getMarkThresholds());
 
     const scannerConfig: ScannerConfigState = {
@@ -289,6 +278,7 @@ export function AppRoot({
       isTestMode: newIsTestMode,
       currentMarkThresholds: newMarkThresholds,
       precinctSelection: newPrecinctSelection,
+      pollsState: newPollsState,
     };
 
     dispatchAppState({
@@ -346,7 +336,6 @@ export function AppRoot({
         // We use an explicit type here to remind ourselves to store any
         // new fields we add to StoredFrontendState.
         const frontendStateToStore: StoredFrontendState = {
-          isPollsOpen,
           ballotCountWhenBallotBagLastReplaced,
           isSoundMuted,
         };
@@ -357,15 +346,19 @@ export function AppRoot({
     void storeAppState();
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
-    isPollsOpen,
     ballotCountWhenBallotBagLastReplaced,
     initializedFromStorage,
     isSoundMuted,
   ]);
 
-  const togglePollsOpen = useCallback(() => {
-    dispatchAppState({ type: 'togglePollsOpen' });
+  const updatePollsState = useCallback(async (newPollsState: PollsState) => {
+    dispatchAppState({ type: 'updatePollsState', pollsState: newPollsState });
+    await config.setPollsState(newPollsState);
   }, []);
+
+  const resetPollsToClosed = useCallback(async () => {
+    await updatePollsState('polls_paused');
+  }, [updatePollsState]);
 
   const toggleTestMode = useCallback(async () => {
     await config.setTestMode(!isTestMode);
@@ -398,6 +391,7 @@ export function AppRoot({
       precinctSelection: newPrecinctSelection,
     });
     await config.setPrecinctSelection(newPrecinctSelection);
+    await refreshConfig();
   }
 
   async function updateMarkThresholds(markThresholds?: MarkThresholds) {
@@ -418,7 +412,12 @@ export function AppRoot({
   const voterMode = auth.status === 'logged_out' && auth.reason === 'no_card';
   useEffect(() => {
     async function automaticallyScanAndAcceptBallots() {
-      if (!(isPollsOpen && voterMode && !needsToReplaceBallotBag)) return;
+      if (
+        !(pollsState === 'polls_open' && voterMode && !needsToReplaceBallotBag)
+      ) {
+        return;
+      }
+
       if (scannerStatus?.state === 'ready_to_scan') {
         await scanner.scanBallot();
       } else if (scannerStatus?.state === 'ready_to_accept') {
@@ -458,6 +457,10 @@ export function AppRoot({
           }
           unconfigureMachine={() =>
             unconfigureServer({ ignoreBackupRequirement: true })
+          }
+          showResetPollsToPausedButton
+          resetPollsToPaused={
+            pollsState === 'polls_closed_final' ? resetPollsToClosed : undefined
           }
           isMachineConfigured={Boolean(electionDefinition)}
           usbDriveStatus={usbDriveDisplayStatus}
@@ -511,6 +514,7 @@ export function AppRoot({
           updatePrecinctSelection={updatePrecinctSelection}
           scannerStatus={scannerStatus}
           isTestMode={isTestMode}
+          pollsState={pollsState}
           toggleLiveMode={toggleTestMode}
           setMarkThresholdOverrides={updateMarkThresholds}
           unconfigure={unconfigureServer}
@@ -556,11 +560,10 @@ export function AppRoot({
       >
         <PollWorkerScreen
           scannedBallotCount={scannerStatus.ballotsCounted}
-          isPollsOpen={isPollsOpen}
-          togglePollsOpen={togglePollsOpen}
+          pollsState={pollsState}
+          updatePollsState={updatePollsState}
           hasPrinterAttached={!!printerInfo}
           isLiveMode={!isTestMode}
-          usbDrive={usbDrive}
         />
       </AppContext.Provider>
     );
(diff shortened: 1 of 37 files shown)


libs/utils/src/precinct_selection.ts
```diff
@@ -32,3 +32,18 @@ export function getPrecinctSelectionName(
...
+  if (precinctSelectionOne.kind === 'AllPrecincts') {
+    return precinctSelectionTwo.kind === 'AllPrecincts';
+  }
+
+  if (precinctSelectionTwo.kind === 'AllPrecincts') {
+    return false;
+  }
+
+  return precinctSelectionOne.precinctId === precinctSelectionTwo.precinctId;
+}
```
adghayes, line 49: Not sure if there's a more idiomatic name I should use for this type of function, like `compare`.

libs/api/src/services/scan/index.ts
```diff
@@ -274,6 +276,54 @@ export type DeletePrecinctSelectionConfigResponse = OkResponse;
...
+ */
+export type PutPollsStateConfigResponse = OkResponse | ErrorsResponse;
+
+/**
+ * @url /config/polls
+ * @method PUT
+ */
+export const PutPollsStateConfigResponseSchema: z.ZodSchema<PutPollsStateConfigResponse> =
+  z.union([OkResponseSchema, ErrorsResponseSchema]);
+
```
adghayes, line 326: Yes please, I'll have two orders of boilerplate.

services/scan/src/precinct_scanner_app.ts
```diff
@@ -551,9 +598,21 @@ export async function buildPrecinctScannerApp(
...
+          status: 'error',
+          errors: [
+            {
+              type: 'polls-closed',
+              message: 'cannot scan ballots while polls are closed',
+            },
+          ],
+        });
+        return;
+      }
```
adghayes, line 615: I'm not sure if this is too limiting - to disabled scanning when polls are closed. Any reservations about this? Seems like a security in depth way to lock down "no voting while polls closed" requirements.

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -138,17 +138,13 @@ type AppAction =
       storedFrontendState: StoredFrontendState;
     }
   | { type: 'resetPollsToClosed' }
-  | {
-      type: 'updateElectionDefinition';
-      electionDefinition: OptionalElectionDefinition;
-    }
```
adghayes, line 144: We're not using this, just removed it.

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -398,6 +391,7 @@ export function AppRoot({
       precinctSelection: newPrecinctSelection,
     });
     await config.setPrecinctSelection(newPrecinctSelection);
+    await refreshConfig();
```
adghayes, line 394: Because we may change other things, like polls state, when changing precinct selection, we need to refresh config after.

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
```diff
@@ -509,14 +597,6 @@ export function PollWorkerScreen({
...
       </Prose>
       <ScannedBallotCount count={scannedBallotCount} />
-      {isExportingResults && (
-        <ExportResultsModal
-          onClose={() => setIsExportingResults(false)}
-          usbDrive={usbDrive}
-          isTestMode={!isLiveMode}
-          scannedBallotCount={scannedBallotCount}
-        />
-      )}
```
adghayes, line 519: @mattroe has been suggesting / asking / arguing to remove the export results option from the poll worker flow for a while, and I decided to do it here. I didn't want to update existing coverage, and we didn't have great coverage on it anyway.

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -357,16 +346,20 @@ export function AppRoot({
...
     ballotCountWhenBallotBagLastReplaced,
     initializedFromStorage,
     isSoundMuted,
   ]);
 
-  const togglePollsOpen = useCallback(() => {
-    dispatchAppState({ type: 'togglePollsOpen' });
+  const updatePollsState = useCallback(async (newPollsState: PollsState) => {
+    dispatchAppState({ type: 'updatePollsState', pollsState: newPollsState });
+    await config.setPollsState(newPollsState);
```
carolinemodic, line 356: We should log whenever this updates. If you look at the table on page 253 of VVSG it gives some specific language about logging opening and closing of polls and also needing to log "The “zero totals” check conducted before opening
the polling place or counting a precinct centrally"

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
```diff
@@ -149,36 +156,53 @@ export function ElectionManagerScreen({
...
+                  <option key={precinct.id} value={precinct.id}>
+                    {precinct.name}
+                  </option>
+                ))}
+            </Select>
+          </p>
+        ) : (
+          precinctSelection && (
+            <p>
+              <ChangePrecinctButton
```
carolinemodic, line 194: Is it possible to reuse this component for the initial selection of a precinct as well? It feels like we are duplicating a lot in this component and in the code above here.

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
```diff
@@ -56,14 +60,24 @@ import * as scan from '../api/scan';
...
-  CLOSE_POLLS_FLOW__COMPLETE = 'close polls flow: complete',
-  EITHER_FLOW__REPRINTING = 'either polls flow: reprinting',
+type PollWorkerFlowState =
+  | 'open_polls_prompt'
+  | 'close_polls_prompt'
+  | 'polls_transition_processing'
+  | 'polls_transition_complete'
+  | 'reprinting_report';
+
+// TODO: Remove this. This is a temporary conversion of the new type for
```
carolinemodic, line 70: I assume this is for a future PR?

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
carolinemodic: I'm not sure I understand the desire of removing it but I guess I defer to @mattroe

libs/types/src/polls.ts
```diff
@@ -0,0 +1,27 @@
+import * as z from 'zod';
+
+export type PollsState =
```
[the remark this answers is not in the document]
carolinemodic: I might be missing some decision that you all have made here about the best practice though.

libs/api/src/services/scan/index.ts
eventualbuddha: All the boilerplate you could want:
image

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -459,6 +458,10 @@ export function AppRoot({
           unconfigureMachine={() =>
             unconfigureServer({ ignoreBackupRequirement: true })
           }
+          showResetPollsToPausedButton
```
carolinemodic, line 461: why do we have this boolean if its just always true?

frontends/precinct-scanner/test/helpers/mock_polls_state.ts
```diff
@@ -0,0 +1,25 @@
+import { PollsState } from '@votingworks/types';
+import fetchMock from 'fetch-mock';
+
+export function mockPollsState(pollsState: PollsState): void {
```
carolinemodic, line 4: love these helpers

carolinemodic (changes requested):

services/scan/src/precinct_scanner_app.ts
carolinemodic: I like this

frontends/precinct-scanner/src/app_root.tsx
adghayes: It's not true on VxCentralScan, where I think this screen is also used.

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
adghayes: Yup!

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
adghayes: Could be, I think we might want to reuse it in the BMD too. I will take a look.

frontends/precinct-scanner/src/app_root.tsx
adghayes: We are lacking _a lot_ of logging on precinct scanner and I held off on some of it because I know we'll have to do a general pass later. I think it may be cleaner to do all that (polls state change, precinct state change (your comment above), report printed, saved to card) in a separate PR?

frontends/precinct-scanner/src/app_root.tsx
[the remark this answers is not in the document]
adghayes: I don't think it's a bug right now, since right now nothing else can change when the precinct gets updated.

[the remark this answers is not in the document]
adghayes: See my other comment about logging, I'd like to do a logging pass in VxScan as a separate PR.

libs/types/src/polls.ts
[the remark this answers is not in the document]
adghayes: You will get a Typescript error if you have a typo in the string. @arsalansufi  has been suggesting union types over enums for best practice, let me see if I can find the article...

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
adghayes: It's not a feature that we have a reasonable user story for.

services/scan/src/store.ts
```diff
@@ -70,6 +72,7 @@ export enum ConfigKey {
   // @deprecated
   SkipElectionHashCheck = 'skipElectionHashCheck',
   PrecinctSelection = 'precinctSelection',
+  PollsState = 'pollsState',
```
jonahkagan, line 75: The polls state feels like a different class of data than the rest of the config values. Those are used to set up ("configure") the machine, while this is more a piece of operating state that has to do with using the machine (it's not a setting).

What do you think about storing it separately and using separate API routes? Does this seem too pedantic? The motivation would mostly just be clarity.

libs/types/src/polls.ts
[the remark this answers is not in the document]
adghayes: Typescript error:
Screen Shot 2022-10-27 at 1 42 23 PM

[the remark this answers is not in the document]
adghayes: https://github.com/votingworks/vxsuite/pull/2221#discussion_r932457382

services/scan/src/store.ts
adghayes: It does seem too pedantic in isolation, but I'm thinking about how as part of this project I plan to move the last bits of frontend state (ballot replaced count and `isSoundsMuted`). Since we're going to have more of these values, I think it makes sense to add the clarity of breaking it out. 

What do you think about all settings being under a single API endpoint that you update in a `Partial` fashion? Just trying to limit code, but maybe that's the wrong motivation.

jonahkagan: Yeah, updating settings with a single endpoint seems fine.

Here's my take on how I would categorize the different bits of state:
- "Config" refers to setting up the machine to work in the right context (e.g. election definition, templates, precinct selection) (we might want to rename this to avoid confusion with machine config - machine id, etc)
- "Settings" are user settings that modify machine behavior, but don't have to do with the election (e.g. muting)
- "Operating state" tracks what's happening while the machine is being used (e.g. test mode, polls state) (this category maybe needs a better name)

Does that seem intuitive/clear to you, or would you think about it differently?

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
adghayes: Refactored.

services/scan/src/store.ts
adghayes: This seems intuitive and clear to me, mostly. My only confusion is that, one way I've thought about config in the past is that it's something that affects how we interpret ballots. and "Test Mode" does affect how we read ballots. So on that piece, it's a big unclear. 

Given this is looking like a bigger change than I expected, I would prefer to do this in a different PR and merge the current polls open state as is, change later.

eventualbuddha: FWIW I want to kill the `configs` table and replace it with specific tables/columns/etc for the data we want to store.

frontends/precinct-scanner/src/components/change_precinct_button.tsx
```diff
@@ -81,43 +141,21 @@ export function ChangePrecinctButton({
...
                 danger
                 onPress={confirmPrecinctChange}
-                disabled={areEqualPrecinctSelections(
-                  initialPrecinctSelection,
-                  pendingPrecinctSelection
-                )}
+                disabled={
+                  !unconfirmedPrecinctSelection ||
+                  areEqualPrecinctSelections(
+                    assertDefined(appPrecinctSelection),
```
adghayes, line 155: I added this `assertDefined` utility because I think it's valuable to be able to indicate, in code, that something should be defined, just like to do with `assert`. It is essentially re-introducing the bang operator, and `assert` is preferable where possible. But these cases, using the existence of a prop as a condition for rendering the element may create the false impression that the element is not meant to be rendered under those conditions.

In @jonahkagan's words, better to fail loudly. cc @arsalansufi who discussed this with me.

services/scan/src/store.ts
adghayes: @eventualbuddha where do you envision just one-off settings, like `isSoundMuted` living? All in their own column / table?

libs/types/src/polls.ts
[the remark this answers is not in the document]
carolinemodic: Ah I see ok, I think typescript has improved how it deals with this since the last time I tried haha

frontends/precinct-scanner/src/app_root.tsx
carolinemodic: Yeah there are definitely a lot of other areas where we are missing logging, I think its probably more efficient in general to just add things when we're in that code as its a little hard later sometimes to remember exactly what the error cases are or whats happening, and doing all of the logging later is a task no one is going to want to do. But we don't have to block on it

services/scan/src/store.ts
eventualbuddha: Assuming we have an `elections` table similar to VxAdmin, these settings could either live on that table (if they're tied to an election) or in a `settings` table that's essentially a singleton table.

adghayes: I would prefer to separate out the reshuffling of how we store / access state from this PR. I propose I leave everything as-is for now, and I will revisit when I work on #2687, which is when I'll move the rest of the state back. Between now and then, we can sync for 15 min and sketch out how we'd like to lay out state?

frontends/precinct-scanner/src/components/change_precinct_button.tsx
eventualbuddha: I've written a utility like this at various times in my various TypeScript projects. It's better than a type cast/assertion (whether `as` or `!`) because it will fail _here_ and not at some later point where the value being used is not as expected. It's sometimes worse in terms of readability, but on the whole I think it's good.

adghayes: Sounds good. I ended up not needing it right here but have left it in this PR because I can anticipate needing it again.

frontends/precinct-scanner/src/app_root.tsx
carolinemodic: semi seperately should we also not show the button at all if the polls aren't in polls_closed_final?

adghayes: Right now it's just disabled if the polls aren't in `polls_closed_final`, I think that's fine for now. I actually originally had it such that it was hidden if polls were not in `polls_closed_final` but React / the testing didn't like how it was being removed the the page (trying to close the modal after the higher level state had already removed it from the page).

frontends/precinct-scanner/src/app_tally_report_paths.test.tsx
```diff
@@ -1273,3 +1238,178 @@ test('saving to card: polls closed, general election, single precinct', async ()
...
+  mockPollsStateChange('polls_open');
+  await screen.findByText('Polls are open.');
+
+  await expectPrint((printedElement) => {
+    printedElement.getByText('TEST Polls Opened Report for Center Springfield');
+    // TODO: Write testing once polls paused reports are designed vxsuite#2686
+  });
+});
+
+test('saving to card: polls unpaused', async () => {
```
carolinemodic, line 1366: AFAICT we don't have any testing right now for the polls paused straight to poll closed transition, probably worth adding a printing and save to card test for that flow.

carolinemodic (approved):
One suggested test I thought off after watching the videos but otherwise this lgtm

frontends/precinct-scanner/src/app_tally_report_paths.test.tsx
adghayes: Yeah, I skipped that that one since it's not a use case we expect very much. But, having a test is better :) Will add then merge.

eventualbuddha:
From the first video:

> it doesn't indicate right on this screen that you are coming from a paused state since that should be clear from the previous screen

I wonder about using the phrase "Open Polls" to describe what's happening here. To me, opening the polls is something that happens _at the start of voting_. And it appears we don't get a polls opened report when we do this "Open Polls". That _feels_ weird to me. I'm imagining a poll worker thinking "I opened the polls this morning and got a report, but I opened the polls again in the afternoon and didn't get a report. Why not? Should I have?". I'd argue we should say more precisely what we're doing: "Open Polls" / "Pause Polls" / "Unpause Polls" (or "Resume Polls"?) / "Close Polls".

I of course defer to @mattroe and @ginvdr on this, just wanted to register my discomfort with that.

adghayes:
@eventualbuddha I wondered the same thing, and we could definitely change it.

My thinking was that poll workers will rarely be doing this in the middle of the day. The main use case for this feature is a vote center model, where you'd start voting at the beginning of each day. If it's a new day of voting, "Open Polls" makes sense to me, even if polls were opened yesterday. The current plan is that the report will say "Polls Opened Report", but it will have an appearance different from the first day polls opened report. And it will say on the report that it was opened from a paused state.

Will wait for @mattroe and/or @ginvdr to weigh in.

adghayes:
Larger context - there will be a report with every single polls state change. The one's for pausing and unpausing will only have ballot counts, not results, per VVSG.

jonahkagan:
I had the same wondering about terminology as @eventualbuddha!

PR #2755, opened 2022-10-31
feat(scan): support loading raw image files

eventualbuddha:
## Overview

Such files are a temporary workaround for the Custom scanner not working with JPEG. This is only a problem because we're using the same approach as `plustekctl` which writes the files to disk, but long term we should consider loading the scanned images directly in memory.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tested manually with the files generated by the [`plustekctl` for Custom](https://github.com/votingworks/plustekctl/compare/customctl).

## Checklist

### DIFF
--- a/services/scan/src/interpreter.ts
+++ b/services/scan/src/interpreter.ts
@@ -21,14 +21,11 @@ import {
   BallotPageMetadata,
   BallotType,
   ElectionDefinition,
-  err,
   getContestsFromIds,
   InterpretedBmdPage,
   MarkThresholds,
-  ok,
   PageInterpretation,
   PrecinctSelection,
-  Result,
 } from '@votingworks/types';
 import {
   adjudicationReasonDescription,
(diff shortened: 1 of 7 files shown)


services/scan/src/interpreter.ts
```diff
@@ -241,21 +233,20 @@ export class Interpreter {
...
-    if (result.isErr()) {
-      timer.end();
-      return { interpretation: result.err() };
+    if (!detectQrcodeResult.qrcode) {
+      return {
+        interpretation: {
+          type: 'UnreadablePage',
+          reason: 'No QR code found',
+        },
+      };
```
jonahkagan, line 242: Did we previously handle this case elsewhere? If so, can we remove the check there?

jonahkagan (approved):

services/scan/src/interpreter.ts
eventualbuddha: Good catch, fixed in 919d1bb.

PR #2765, opened 2022-11-03
chore(scan): remove adjudication

eventualbuddha:
## Overview

We used to have adjudication features in VxCentralScan, i.e. for adjudicating overvotes/undervotes/etc. We took the UI for that out a while back but never fully removed the backend implementation. Since this is unlikely to ever be a product requirement, or if it is it won't look like this, it's better to just remove it.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated testing.

## Checklist


frontends/bsd/src/screens/ballot_eject_screen.tsx
```diff
@@ -153,63 +146,59 @@ export function BallotEjectScreen({
...
       }
     }
-  }, [reviewInfo, logger, userRole]);
+  }
 
   useEffect(() => {
     void (async () => {
       const frontAdjudicationComplete =
-        !!frontMarkAdjudications ||
+        !!isFrontAdjudicationDone ||
```
jonahkagan, line 198: I think the `!!` here is unnecessary now

services/scan/schema.sql
```diff
@@ -35,6 +35,7 @@ create table sheets (
   -- When adjudication is finished, these values are updated to now.
   front_finished_adjudication_at datetime,
   back_finished_adjudication_at datetime,
+  finished_adjudication_at datetime,
```
jonahkagan, line 30: I'm a little confused - why is there still a concept of adjudication if we removed the feature?

jonahkagan (approved):

services/scan/schema.sql
eventualbuddha: We still have a concept of “this ballot has a problem” like an over vote that the user can decide something about (accept or reject). So we need this value to indicate when it was accepted. I'm happy to update the language if you can think of something better.

jonahkagan: Ah ok, I think that still sounds like adjudication to me! And [NIST](https://pages.nist.gov/ElectionGlossary/#:~:text=adjudication,ballots%20to%20reflect%20voter%20intent.) agrees:
Screen Shot 2022-11-03 at 11 17 35 AM

Maybe I was just thrown off by the PR title

PR #2767, opened 2022-11-03
VxScan Ballot Batching + Remove Persisted Frontend State

### LINKED ISSUE #1520: Determine how to group "batches" for VxScan
Currently the precinct-scanner considers every ballot its own batch, that is confusing in VxAdmin where you can see the ballots by Batch and Scanner. That view was added in. case someone runs a RLA. We should figure out with @mcchilders how we should actually be making batches out of the VxScan results.

adghayes:
## Overview
Closes #1520 and #2687. Rather than batching every ballot on the precinct scanner as its own batch, batches all ballots together until a new batch is started by a) pausing and unpausing polls or b) replacing the ballot bag.

Took this change as an opportunity to move forward on our architecture plan and remove all persistent state from the frontend. Refactored some of the endpoints / API calls and testing utilities as a result. 

## Testing Plan
Manually tested running the scanner, pausing/unpausing polls, replacing the ballot bag, and power cycling (restarting services and app). Confirmed that pausing/unpausing and replacing the ballot bag started a new batch, but power cycling did not.

## Checklist


services/scan/src/end_to_end.test.ts
```diff
@@ -42,27 +42,18 @@ afterEach(async () => {
 });
 
 test('going through the whole process works', async () => {
-  // Do this first so interpreter workers get initialized with the right value.
   await request(app)
-    .patch('/central-scanner/config/skipElectionHashCheck')
-    .send({ skipElectionHashCheck: true })
```
adghayes, line 48: Not sure why we needed to do this first, the tests pass if we do it after setting the election.

services/scan/src/end_to_end.test.ts
```diff
@@ -42,27 +42,18 @@ afterEach(async () => {
...
 
-  {
-    // try export before configure
-    const response = await request(app)
-      .post('/central-scanner/scan/export')
-      .set('Accept', 'application/json')
-      .expect(200);
-    expect(response.text).toBe('');
-  }
-
```
adghayes, line 61: Why would we test export before configure?

services/scan/src/importer.test.ts
```diff
@@ -162,6 +163,7 @@ test('restoreConfig reconfigures the interpreter worker', async () => {
     workers.Input,
     workers.Output
   >(workerCall);
+  workspace.store.setElection(electionSampleDefinition);
```
adghayes, line 166: In a a few places, I have to set the election in tests because now we have the constraint that configs depend on an election existing.

services/scan/schema.sql
```diff
@@ -1,3 +1,18 @@
...
+  marginal_mark_threshold_override real,
+  definite_mark_threshold_override real,
+  cvrs_backed_up_at datetime,
+  scanner_backed_up_at datetime,
+  created_at timestamp not null default current_timestamp
+);
+
 create table batches (
   batch_number integer primary key autoincrement,
   id varchar(36) unique,
```
adghayes, line 18: Could put foreign keys on `batches` to `election`, but if we really think it's a singleton table, then that seems like a waste.

services/scan/src/importer.ts
```diff
@@ -563,9 +563,8 @@ export class Importer {
   /**
    * Reset all the data, both in the store and the ballot images.
    */
-  async doZero(): Promise<void> {
+  doZero(): void {
     this.workspace.zero();
-    await this.setMarkThresholdOverrides(undefined);
```
adghayes, line 568: I'm not sure why we're resetting the mark threshold overrides in `doZero`. We shouldn't reset the thresholds when we switch between test modes, or clear ballot data. On unconfigure they will be cleared anyway. Removed.

services/scan/src/store.ts
```diff
@@ -620,19 +675,28 @@ export class Store {
...
 
+  resetVotingState(): void {
+    if (this.hasElection()) {
+      this.client.run(
+        'update election set cvrs_backed_up_at = null, scanner_backed_up_at = null'
+      );
+      this.setPollsState('polls_closed_initial');
+    }
+  }
+
```
adghayes, line 686: The equivalent of resetting the "session."

services/scan/src/end_to_end.test.ts
adghayes: If we are concerned about this, I think we should probably `400` on this.

eventualbuddha: Agreed it should be 400. And why test? To make sure it does send the right response code.

services/scan/schema.sql
eventualbuddha: You can do cascade deletion with a foreign key. See [this example](https://github.com/votingworks/vxsuite/blob/ea1602678a988eba239c69237a696303a6924ff0/services/admin/schema.sql#L17-L18). And yes, I do think we should add these foreign keys.

services/scan/schema.sql
```diff
@@ -1,3 +1,18 @@
+create table election (
+  -- enforce singleton table
```
eventualbuddha, line 2: Do we actually want to enforce this? If so, then I guess I feel less certain about my foreign key comment from earlier.

services/scan/src/store.ts
```diff
@@ -223,73 +204,158 @@ export class Store {
     this.client.reset();
   }
 
+  /**
+   * Gets the current election definition.
```
eventualbuddha, line 208:
```suggestion
   * Gets whether an election is currently configured.
```

services/scan/src/store.ts
```diff
@@ -223,73 +204,158 @@ export class Store {
...
+   */
+  hasElection(): boolean {
+    return Boolean(this.client.one('select * from election'));
```
eventualbuddha, line 211:
```suggestion
    return Boolean(this.client.one('select id from election'));
```

services/scan/src/store.ts
```diff
@@ -223,73 +204,158 @@ export class Store {
...
    * Gets the current election definition.
    */
   getElectionDefinition(): ElectionDefinition | undefined {
-    const electionDefinition: ElectionDefinition | undefined = this.getConfig(
-      ConfigKey.Election,
+    const electionRow = this.client.one(
+      'select election_definition from election'
+    ) as { election_definition: string } | undefined;
+
+    const rawElectionDefinition = electionRow?.election_definition;
```
eventualbuddha, line 222:
```suggestion
    const electionRow = this.client.one(
      'select election_definition as electionDefinition from election'
    ) as { electionDefinition: string } | undefined;

    const rawElectionDefinition = electionRow?.election_definition;
```

services/scan/src/store.ts
```diff
@@ -223,73 +204,158 @@ export class Store {
...
+      this.client.run('delete from election');
+    } else {
+      if (this.hasElection()) {
+        this.setElection(undefined);
+      }
+      this.client.run(
+        'insert into election (election_definition) values (?)',
+        JSON.stringify(electionDefinition)
+      );
+    }
```
eventualbuddha, line 262: This can be simpler, but also I think we should just store the election data itself. Then above you'd use `safeParseElectionDefinition` to load it.
```suggestion
    this.client.run('delete from election');
    if (electionDefinition) {
      this.client.run(
        'insert into election (election_definition) values (?)',
        electionDefinition.electionData
      );
    }
```

services/scan/src/store.ts
```diff
@@ -223,73 +204,158 @@ export class Store {
...
   }
 
   /**
    * Gets the current test mode setting value.
    */
   getTestMode(): boolean {
-    // service should default to test mode if not set
-    return this.getConfig(ConfigKey.TestMode, true, z.boolean());
+    const electionRow = this.client.one('select is_test_mode from election') as
+      | { is_test_mode: number }
```
eventualbuddha, line 270:
```suggestion
    const electionRow = this.client.one('select is_test_mode as isTestMode from election') as
      | { isTestMode: number }
```

services/scan/src/store.ts
```diff
@@ -223,73 +204,158 @@ export class Store {
...
    * Gets the current override values for mark thresholds if they are set.
    * If there are no overrides set, returns undefined.
    */
   getMarkThresholdOverrides(): Optional<MarkThresholds> {
-    return this.getConfig(
-      ConfigKey.MarkThresholdOverrides,
-      MarkThresholdsSchema
-    );
+    const electionRow = this.client.one(
+      'select marginal_mark_threshold_override, definite_mark_threshold_override from election'
```
eventualbuddha, line 338: Alias to camel case as above.

services/scan/src/store.ts
```diff
@@ -223,73 +204,158 @@ export class Store {
...
+          marginal_mark_threshold_override?: number;
+          definite_mark_threshold_override?: number;
+        }
+      | undefined;
+
+    if (!electionRow) {
+      return undefined;
+    }
+
+    if (electionRow.definite_mark_threshold_override) {
```
eventualbuddha, line 350:
```suggestion
    if (typeof electionRow.definite_mark_threshold_override === 'number') {
```

services/scan/schema.sql
adghayes: SGTM.

services/scan/src/store.ts
```diff
@@ -299,21 +365,27 @@ export class Store {
...
-  }
-
   /**
    * Sets the current override values for mark thresholds. A value of undefined
    * will remove overrides and cause thresholds to fallback to the default values
    * in the election definition.
    */
   setMarkThresholdOverrides(markThresholds?: MarkThresholds): void {
-    this.setConfig(ConfigKey.MarkThresholdOverrides, markThresholds);
+    if (this.hasElection()) {
```
eventualbuddha, line 374: Maybe short-circuit on `!this.hasElection()` to avoid indenting the whole thing?

services/scan/src/store.ts
```diff
@@ -322,100 +394,72 @@ export class Store {
    * default).
    */
   getPrecinctSelection(): Optional<PrecinctSelectionType> {
-    return this.getConfig(ConfigKey.PrecinctSelection, PrecinctSelectionSchema);
+    const electionRow = this.client.one(
+      'select precinct_selection from election'
```
eventualbuddha, line 398: Camel case.

services/scan/src/store.ts
```diff
@@ -322,100 +394,72 @@ export class Store {
...
+    return precinctSelectionParseResult.ok();
   }
 
   /**
    * Sets the current precinct `scan` is accepting ballots for. Set to
    * `undefined` to accept from all precincts (this is the default).
    */
   setPrecinctSelection(precinctSelection?: PrecinctSelectionType): void {
-    this.setConfig(ConfigKey.PrecinctSelection, precinctSelection);
+    if (this.hasElection()) {
```
eventualbuddha, line 425: Ditto on short circuit.

services/scan/src/store.ts
```diff
@@ -322,100 +394,72 @@ export class Store {
...
-      schema = defaultValueOrSchema as z.ZodSchema<T>;
-    }
-
-    if (typeof row === 'undefined') {
-      debug('returning default value for config %s: %o', key, defaultValue);
-      return defaultValue;
-    }
-
-    const result = safeParseJson(row.value, schema);
+    const electionRow = this.client.one('select polls_state from election') as
```
eventualbuddha, line 443: Camel case.

services/scan/src/store.ts
```diff
@@ -322,100 +394,72 @@ export class Store {
...
     }
 
-    const value = result.ok();
-    let inspectedResult = inspect(value, false, 2, true);
-    if (inspectedResult.length > 200) {
-      inspectedResult = `${inspectedResult.slice(0, 199)}…`;
-    }
-    debug('returning stored value for config %s: %s', key, inspectedResult);
-    return value;
+    return electionRow.polls_state;
```
eventualbuddha, line 452: Validate it at all?

services/scan/src/store.ts
```diff
@@ -322,100 +394,72 @@ export class Store {
...
-    }
-
-    if (typeof row === 'undefined') {
-      debug('returning default value for config %s: %o', key, defaultValue);
-      return defaultValue;
-    }
-
-    const result = safeParseJson(row.value, schema);
+    const electionRow = this.client.one('select polls_state from election') as
+      | { polls_state: PollsStateType }
```
eventualbuddha, line 444:
```suggestion
      | { polls_state: string }
```

services/scan/src/store.ts
```diff
@@ -447,35 +491,46 @@ export class Store {
...
+   * Gets the timestamp for the last cvr export
    */
-  private getBackupTimestamp(key: BackupKey): Iso8601Timestamp | undefined {
-    const row = this.client.one<[Iso8601Timestamp]>(
-      'select value from backups where key = ?',
-      key
-    ) as { value: Iso8601Timestamp } | undefined;
-    return row?.value;
+  private getCvrBackupTimestamp(): Iso8601Timestamp | undefined {
+    const row = this.client.one('select cvrs_backed_up_at from election') as
```
eventualbuddha, line 530: One last time, then I'm going to just assume you know: camel case.

services/scan/src/store.ts
```diff
@@ -620,19 +675,28 @@ export class Store {
...
+    this.resetVotingState();
   }
 
   getBallotFilenames(
     sheetId: string,
     side: Scan.Side
   ): { original: string; normalized: string } | undefined {
     const row = this.client.one<[string]>(
-      `
+      `
```
eventualbuddha, line 699:
```suggestion
      `
```

services/scan/src/store.ts
```diff
@@ -620,19 +675,28 @@ export class Store {
     );
   }
 
+  resetVotingState(): void {
+    if (this.hasElection()) {
+      this.client.run(
+        'update election set cvrs_backed_up_at = null, scanner_backed_up_at = null'
+      );
+      this.setPollsState('polls_closed_initial');
```
eventualbuddha, line 683: This is unlikely to fail, but consider doing this as a transaction?

eventualbuddha (commented):
Overall approach looks sane.

services/scan/schema.sql
adghayes: I think that we do. Right now the product expectation is that all devices but VxAdmin are only ever configured for one election at a time. Might as well enforce that expectation and get the benefit of never accidentally creating two elections. Easy to remove later if necessary.

And yes, this is why I was ambivalent about the foreign keys. I will leave them off for now.

services/scan/src/store.ts
adghayes: It seems unnecessary in that typescript is validating the value entering the DB, and it only has one entry point, so it should always be correct. Is that myopic?

Also, if we want to validate it, why not validate it at the DB level with a `CHECK` in the schema?

services/scan/src/store.ts
adghayes: For this case, can just do it in one statement. Maybe we will add more here later and it will need to be a transaction.

services/scan/src/store.ts
adghayes: Safe assumption.

services/scan/src/store.ts
adghayes: Never mind, transactions are easy :)

services/scan/src/store.ts
adghayes: I am assuming this is related to your previous comment - if we type it as `string`, then typescript expects some kind of validation.

services/scan/src/precinct_scanner_app_other.test.ts
```diff
@@ -0,0 +1,393 @@
...
+  });
+
+  // Wait for transition back to no paper
+  await waitForStatus(app, {
+    state: 'no_paper',
+    ballotsCounted: initialBallotsCounted + 1,
+  });
+}
+
+test("setting the election also sets precinct if there's only one", async () => {
```
adghayes, line 320: I wasn't sure what to name this file, but it seemed like `precinct_scanner_app.test` was getting too big, and there was a distinction between the testing of all the scan behavior and then the testing of all the other backend functions. If you agree we can start splitting this testing, I'll extract the helpers above.

services/scan/src/precinct_scanner_app_scanning.test.ts
```diff
@@ -545,16 +525,14 @@ test('invalid ballot rejected', async () => {
...
-    canUnconfigure: false,
   });
   await waitForStatus(app, {
     state: 'rejected',
     interpretation,
-    canUnconfigure: false,
   });
 
   await mockPlustek.simulateRemoveSheet();
-  await waitForStatus(app, { state: 'no_paper', canUnconfigure: false });
```
adghayes, line 557: Previously the behavior was that you couldn't unconfigure after any batch had been started, even if the batch was empty or only contained rejected (deleted) sheets. Since now, the batch starts with polls open, that means that you could not unconfigure after polls open. Behavior changed to that you can unconfigure if there are no counted (accepted) ballots.

services/scan/src/server.ts
```diff
@@ -114,6 +114,9 @@ export async function start({
         importer: resolvedImporter,
         workspace: resolvedWorkspace,
       }));
+
+    // cleanup incomplete batches from before
+    resolvedWorkspace.store.cleanupIncompleteBatches();
```
adghayes, line 119: Moving this to the central scanner only because otherwise it would delete the ongoing back with the polls session on VxScan.

services/scan/src/store.ts
```diff
@@ -486,45 +645,49 @@ export class Store {
...
-      maxBatchesStartedAt: Iso8601Timestamp;
+
+    // Deleting non-empty batches will update the CVR count
+    const { maxBatchesDeletedAt } = this.client.one(`
+      select
+        max(batches.deleted_at) as maxBatchesDeletedAt
+      from batches inner join sheets
+      on sheets.batch_id = batches.id
+      where sheets.deleted_at is null
+    `) as {
```
adghayes, line 678: I update the logic a bit of what invalidates the backup. For example, I think deleting an empty batch should not invalidate a backup. Logic is laid out in the test file.

services/scan/src/store.ts
```diff
@@ -441,39 +553,86 @@ export class Store {
...
+   * Returns the id of the an unfinished batch if there is one
+   */
+  getOngoingBatchId(): Optional<string> {
+    const ongoingBatchRow = this.client.one(
+      'select id from batches where ended_at is null'
+    ) as { id: string } | undefined;
+
+    return ongoingBatchRow?.id;
+  }
+
```
adghayes, line 566: Ongoing Batch = batch associated with current polls session

services/scan/src/precinct_scanner_app.ts
```diff
@@ -118,22 +109,15 @@ export async function buildPrecinctScannerApp(
...
+    '/precinct-scanner/config',
+    (_request, response) => {
+      response.json({
+        electionDefinition: store.getElectionDefinition(),
+        precinctSelection: store.getPrecinctSelection(),
+        markThresholdOverrides: store.getMarkThresholdOverrides(),
+        isSoundMuted: store.getIsSoundMuted(),
+      });
     }
   );
```
adghayes, line 122: I'm grouping all the config values together as one API request to simplify logic and reduce code on the backend and frontend.

adghayes:
@jonahkagan I'd like you to take a glance at this (not a full review) since I mostly followed your thinking in terms of data separation. See `app_root`, `precinct_scanner_app`, and `store`. Although we may store it together in the DB, the two groups of state that the backend provides to the frontend (outside of scanner status) are the _config_ and the details of the current _election session_. I'm trying to standardize on the language "election session" to think about what gets reset when switching modes.

No more frontend state! 

This is not polished - needs some refactoring on the backend and a lot of rewritten testing on the frontend (with some new test utilities).

services/scan/src/precinct_scanner_state_machine.ts
```diff
@@ -254,12 +254,13 @@ function storeInterpretedSheet(
...
-  const batchId = store.addBatch();
-  const addedSheetId = store.addSheet(sheetId, batchId, interpretation.pages);
-  store.finishBatch({ batchId });
+  const ongoingBatchId = store.getOngoingBatchId();
+  assert(typeof ongoingBatchId === 'string');
+  const addedSheetId = store.addSheet(
+    sheetId,
+    ongoingBatchId,
+    interpretation.pages
+  );
```
adghayes, line 263: Adding ballots to a batch rather than each ballot having its own batch!

services/scan/src/precinct_scanner_app.ts
```diff
@@ -275,41 +296,29 @@ export async function buildPrecinctScannerApp(
...
+    '/precinct-scanner/electionSession',
     (_request, response) => {
-      const pollsState = store.getPollsState();
-      response.json({ status: 'ok', pollsState });
+      response.json({
+        isTestMode: store.getTestMode(),
+        pollsState: store.getPollsState(),
+        ballotCountWhenBallotBagLastReplaced:
+          store.getBallotCountWhenBallotBagLastReplaced(),
+      });
```
jonahkagan, line 311: I like how you've packaged up the election session!

The abstraction defined between app and store feels a bit murky to me. Who manages the rules about election sessions, the app or the store? Currently, the app assembles the election session and decides what changes merit resetting it, but the store knows what resetting a session means. The downside of this is that the logic for maintaining the integrity of the session is slightly spread out, which makes it more likely we'll mess it up at some point.

A few directions I could imagine moving in:
1. The store doesn't know anything about election sessions and only puts data into slots in the election table on the db
2. The store offers CRUD methods for election sessions, the app decides when election sessions get created/reset based on other changes (e.g. changing config)
3. Option 2, plus the store enforces rules about when sessions are created/reset

There are tradeoffs to each approach. I'm not really sure what the vision for the "store" is, and would be curious to hear @eventualbuddha's thoughts as well.

Without thinking deeply about it, I'd probably pick option 2, which would treat election session as it's own data type (even though secretly the store is storing it in the election table).

adghayes: @jonahkagan thanks for taking a look at this! I definitely can see the murkiness you’re talking about. I think Option 1 wouldn’t actually prevent the sort of footguns i think you’re worried about. Option 2 could, but it seems to me like it might be over investing in this new concept of the election session. 

The election session is created initially from that point on is only ever reset. If the app were just calling for the reset, and the store were managing the reset, that would be a fine division of function. But the murkiness is where the app is also packaging the election session for the frontend API. I think this is a tolerable amount of murkiness for now, since for now these concepts are just buckets of data that we’re separating for readability rather than use case.

jonahkagan: Well, there's a bit more murkiness than that. The store has methods to get/set individual parts of the election session as well. So you could conceivably reset one without changing the others if you didn't see the specific reset method. I think things start to get dangerous when there are multiple ways to do things.

I agree with you that the logic isn't super complicated today, but establishing a clear foundation with defined abstraction lines between different modules will help it stay clear as it develops over time.

adghayes: The app has to be able to get/set fields of the election session somehow, right? Do you just want a clearer API, like store.electionSession.method()?

I think we have a good indication from product that this is not an abstraction that will be built upon. I don’t foresee having multiple election sessions. Or doing anything other than resetting them. It’s just a matter deciding what is reset when switching modes and not. Perhaps myopic, sure.

Moreover, I think we’re thinking a little too much about this. It doesn’t strike me as very important. I’d like to hear from @eventualbuddha and move on. I’m happy with a maximalist approach, where election sessions are packaged and distinguished in the store and even separated in the DB on a separate table. I’m happy with the current approach. I just want to push back against investing time into abstractions that I don’t see being foundational. I’ll defer to y’all. 

I’m curious about your thoughts on packaging things together for the API between frontend in backend, as I’ve drafted here. It made sense to me to make 1 or 2 requests rather than however many pieces of data we have.

jonahkagan: If you think it's not a useful abstraction, that's totally fine with me. I would say in that case just remove it. I'm mainly pushing back against having a "halfway" abstraction.

adghayes: I don't think that the abstraction is useful in that it's not helpful for future work that we can anticipate. The goal was to introduce clarity, in response to your comment [here](https://github.com/votingworks/vxsuite/pull/2751#discussion_r1007352267). But it sounds like this "halfway" abstraction creates as many additional potential issues as it does clarity.

In that case, I will remove it, and just leave all "operating state" like polls state and ballot bag last replaced count behind the `/config` endpoint.

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
```diff
@@ -179,7 +179,8 @@ export function PollWorkerScreen({
       setCurrentTally(tally);
     }
     void calculateTally();
-  }, [election, scannedBallotCount, precinctSelection]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
adghayes, line 182: I saw in tests how this effect may run multiple times and reload CVRs, when in reality we only want to run it when the poll worker screen loads. The CVRs from the backend cannot update while the poll worker screen is open (no ballots can be cast).

frontends/precinct-scanner/src/contexts/app_context.ts
```diff
@@ -21,7 +21,7 @@ export interface AppContextInterface {
 const appContext: AppContextInterface = {
   electionDefinition: undefined,
   precinctSelection: undefined,
-  currentMarkThresholds: undefined,
+  markThresholdOverrides: undefined,
```
adghayes, line 24: Renamed this to be clearer.

libs/api/src/services/scan/index.ts
```diff
@@ -234,92 +234,124 @@ export const GetPrecinctSelectionConfigResponseSchema: z.ZodSchema<GetPrecinctSe
 
 /**
  * @url /config/precinct
- * @method PUT
+ * @method PATCH
  */
-export interface PutPrecinctSelectionConfigRequest {
+export interface PatchPrecinctSelectionConfigRequest {
```
adghayes, line 239: Standardizing on `PATCH` for config changes, some of my past PRs had introduced a meaningless variation.

services/scan/src/store.ts
eventualbuddha: Data in a database can get out of sync with the code that uses it. For example, say we had this:

```ts
interface Person {
  firstName: string;
  lastName: string
}
```

And we stored such a value in the database as a JSON blob. Then, sometime later, we changed the code like so:

```ts
interface Person {
  givenName: string;
  familyName: string;
}
```

We could do `const person: Person = JSON.parse(row);` and say that the data should be correct because when we inserted the data it was a TypeScript-validated `Person` and now we're loading it as a `Person` so it's fine, right? But since the `Person` type changed we can no longer make any guarantees. That's why, in general, when we load data from external sources I like to validate it even if we "own" the data and put it there ourselves.

Depending on your desired behavior you can throw an exception, ignore the invalid data, migrate it to the new format, or something else. But if you never check you'll never know it's wrong until something blows up or misbehaves later.

adghayes: In this particular case you called out, I ended up adding the validation. Your explanation makes sense to me!

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
eventualbuddha: Did the previous dependencies change and trigger the effect again? If it does get triggered again, does that cause bugs?

libs/api/src/services/scan/index.ts
eventualbuddha: I'm hoping we can autogenerate a bunch of this in the future and probably ditch meaningful HTTP methods. That said, here's what MDN says about [`PATCH`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) vs `PUT`:

> A `PATCH `request is considered a set of instructions on how to modify a resource. Contrast this with [`PUT`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT); which is a complete representation of a resource.

So if the URL you're `PATCH`ing represents the complete representation of the resource, you should be using `PUT`. If you really wanted to use `PATCH` you'd have a `/config` endpoint and your HTTP request body would have something like `{ "precinctSelection": … }`.

services/scan/src/util/workspace.ts
```diff
@@ -62,8 +62,8 @@ export function createWorkspace(root: string): Workspace {
     scannedImagesPath,
     uploadsPath,
     store,
-    zero() {
-      store.zero();
+    resetElectionSession() {
```
eventualbuddha, line 65: @jonahkagan is this how you envisioned a "session"?

eventualbuddha (commented):
Took a first pass. Will have to come back later to look more deeply.

frontends/precinct-scanner/src/api/config.ts
```diff
@@ -34,29 +33,29 @@ async function patch<Body extends string | ArrayBuffer | unknown>(
...
+//     headers: {
+//       'Content-Type': isJson ? 'application/json' : 'application/octet-stream',
+//     },
+//   });
+//   const body: OkResponse | ErrorsResponse = await response.json();
+
+//   if (body.status !== 'ok') {
+//     throw new Error(`PUT ${url} failed: ${JSON.stringify(body.errors)}`);
+//   }
+// }
```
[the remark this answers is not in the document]
adghayes: Sure!

libs/api/src/services/scan/index.ts
adghayes: I was just trying to standardize. Based on your info, it seems like they should all be `PUT`. But I don't know how important that switch is, so I'll leave it out here unless you insist.

adghayes:
I ran through a manual testing flow and confirmed the batching looks good. 

However, with the new way in which all `config` values are bunched under a common API request, I found a new issue on unconfigure. Because the `GET /precinct-scanner/config` endpoint returns the config with `undefined` values - and those are stripped out of the JSON by express / `stringify`, they don't end up in the frontend and overwrite the truthy config values after unconfigure. UI behavior appears like - on the "Delete Election Data" flow, the modal seems to be stuck on `Loading...`, because normally we use the hiding of that screen to hide the modal.

So @eventualbuddha, I'm not sure what the best way to fix this is. Thoughts?
1. The `GET` response should use null values rather than undefined, and we convert between null and undefined on both server and client
2. Just add a couple lines in the `dispatch` action to overwrite with undefined before using the values from the server (so if they are missing, they are interpreted as undefined)
3. Use `null` on the frontend and backend rather than undefined (bigger change).

services/scan/src/util/workspace.ts
jonahkagan: In terms of what data would be grouped together, yes, more or less. I imagined it as a first-class data type, but @adghayes and I discussed that in another thread and decided to not break out that type for now

services/scan/src/store.ts
```diff
@@ -283,101 +414,82 @@ export class Store {
...
+    }
+
+    if (!precinctSelection) {
+      this.client.run('update election set precinct_selection = null');
+    } else {
+      this.client.run(
+        'update election set precinct_selection = ?',
+        JSON.stringify(precinctSelection)
+      );
+    }
```
eventualbuddha, line 456:
```suggestion
    this.client.run(
      'update election set precinct_selection = ?',
      precinctSelection ? JSON.stringify(precinctSelection) : null
    );
```

services/scan/src/store.ts
```diff
@@ -404,90 +516,142 @@ export class Store {
...
-    return row?.value;
+  getCvrsBackupTimestamp(): Iso8601Timestamp | undefined {
+    const row = this.client.one(
+      'select cvrs_backed_up_at as cvrsBackedUpAt from election'
+    ) as { cvrsBackedUpAt: Iso8601Timestamp } | undefined;
+    return row?.cvrsBackedUpAt;
+  }
+
+  getBallotsCounted(): number {
+    const r = this.client.one(`
```
eventualbuddha, line 585: Use `row` like we do elsewhere.
```suggestion
    const row = this.client.one(`
```

services/scan/src/store.ts
```diff
@@ -404,90 +516,142 @@ export class Store {
...
-
-    if (!batchesLastUpdatedAt) {
-      return true;
-    }
 
-    const isBackupUpToDate =
-      scannerBackedUpAt >= cvrsLastUpdatedAt &&
-      scannerBackedUpAt >= batchesLastUpdatedAt;
-    return isBackupUpToDate;
+    return scannerBackedUpAt >= cvrsLastUpdatedAt;
```
eventualbuddha, line 654: I realize all this was here before, but it makes me anxious that all this date math is being done with `string`s. Can we use `DateTime` from `luxon` or `Date` instead? Using strings assumes that all the values are in the timezones. That should be true here, but it feels like an unnecessarily fragile assumption. If not in this PR, can you make a follow-up ticket?

eventualbuddha (approved):

services/scan/src/precinct_scanner_app.ts
```diff
@@ -324,36 +352,46 @@ export async function buildPrecinctScannerApp(
...
-
   app.patch<
     NoParams,
-    Scan.PatchMarkThresholdOverridesConfigResponse,
-    Scan.PatchMarkThresholdOverridesConfigRequest
+    Scan.PatchBallotCountWhenBallotBagLastReplacedResponse,
+    Scan.PatchBallotCountWhenBallotBagLastReplacedRequest
   >(
-    '/precinct-scanner/config/markThresholdOverrides',
+    '/precinct-scanner/config/ballotCountWhenBallotBagLastReplaced',
```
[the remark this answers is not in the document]
adghayes: I thought of that but didn't know if there might be some use case in the future where we want to specify? I can't think of any.

services/scan/src/store.ts
adghayes: I actually hit this issue when I tried to use something other than `current_timestamp` during testing. I think it makes sense, and I will include it in this PR, but will have you take a quick look.

adghayes: I've started using `DateTime`, but did you know what we're storing our dates (from `current_timestamp`) in non-timezoned GMT strings that don't even necessarily fit the `ISO8601` spec? E.g. `2022-11-16 18:18:54`. We should probably be storing with timezones, right?

services/scan/src/precinct_scanner_app.ts
```diff
@@ -366,11 +404,26 @@ export async function buildPrecinctScannerApp(
         return;
       }
 
-      store.setMarkThresholdOverrides(
-        bodyParseResult.ok().markThresholdOverrides
-      );
-      await updateInterpreterConfig(interpreter, workspace);
+      // Start a new batch
+      const ongoingBatchId = store.getOngoingBatchId();
+      assert(typeof ongoingBatchId === 'string');
```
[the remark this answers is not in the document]
adghayes: Thinking about this... If this assertion fails, it really is a failure of server state and an "Internal Server Error" rather than a "Bad Request." But if this function throws an error and crashes, does the whole backend crash?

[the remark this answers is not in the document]
adghayes: Based on how we are managing state on the server, the assertion _should_ be true and it really is a "Internal Server Error" rather than a "Bad Request." So I actually think the current behavior is appropriate? There's not a request I can think of that would trigger the assertion to fail, only changes to server logic.

services/scan/src/precinct_scanner_app.ts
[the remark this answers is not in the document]
adghayes: Changed this, agree it is better this way.

services/scan/src/store.ts
eventualbuddha: Oh no, really? grrr 😠 sqlite

adghayes: @kofi-q just changed it in the admin service: https://github.com/votingworks/vxsuite/pull/2789#discussion_r1025510235

Issue #2716, opened 2022-10-19
VxAdmin: Consolidate all CVR information in the admin service backend


carolinemodic:
We are currently duplicating CVR information in both the local storage of kiosk browser and in the backend admin service, this is causing a number of weird bugs in edge cases of the import file modal. We should relocate everything to the admin-service backend to ease confusion and have a single source of truth, and make sure all import edge cases are properly handled in the process. 

Related import bugs: 
https://github.com/votingworks/vxsuite/issues/2715
https://github.com/votingworks/vxsuite/issues/2680

PR #2771, opened 2022-11-08
[VxAdmin] Prevent importing mixed 'live' and 'test' CVRs

kofi-q:
## Overview

- Added validation code to the "add CVR file" backend API to prevent imports of both "live" and "test" CVRs for the same election.
- Added a new `fetchWithSchema` util to use instead of `fetchWithJson` - the latter throws an exception for any non-200 response, which didn't play well with trying to propagate errors from the backend.
  - Not sure if the location of that util fits with our code organisation vision, so let me know if it should live elsewhere.
- Fixed bug with hanging transactions whenever an error occurs while adding a single CVR entry during a file import.

Error propagation is not in a great state at the moment, WRT what gets displayed to the user, so it'll be worth a future pass to clean that up a bit.

## Demo Video or Screenshot
https://user-images.githubusercontent.com/264902/200440995-92e98556-4dd6-455f-9122-16d48f43a5a7.mov

Screenshot 2022-11-07 at 4 25 22 PM Screenshot 2022-11-07 at 4 25 46 PM

## Testing Plan 
- Added unit tests for new error paths
- Tested locally to verify behaviour

## Checklist


libs/api/src/client_utils.ts
```diff
@@ -0,0 +1,41 @@
+import { safeParseJson } from '@votingworks/types';
+import * as z from 'zod';
+
+import { ErrorsResponse, OkResponse } from './base';
+
+/**
+ * A Fetch wrapper for API clients that accepts an API response schema for
+ * parsing and validating server responses.
+ */
+export async function fetchWithSchema<T extends OkResponse | ErrorsResponse>(
```
carolinemodic, line 10: nice

carolinemodic (approved):
Nice!

services/admin/src/server.ts
```diff
@@ -175,7 +175,9 @@ export function buildApp({ workspace }: { workspace: Workspace }): Application {
             errors: [
               {
                 type: result.err().kind,
-                message: JSON.stringify(result.err()),
+                message:
+                  result.err().userFriendlyMessage ||
```
jonahkagan, line 179: Maybe we leave user-friendly copy to the frontend, since that seems like more of a UI concern?

services/admin/src/store.ts
```diff
@@ -277,16 +285,55 @@ export class Store {
...
+
         for await (const line of lines) {
           if (line.trim() === '') {
             continue;
           }
 
-          // TODO(https://github.com/votingworks/vxsuite/issues/2716): add error checking against
-          // attempts to mix "live" and "test" CVRs for the same election.
+          // Parse CVR data for validation.
+          const cvr = safeParseJson(line).unsafeUnwrap() as CastVoteRecord;
```
jonahkagan, line 299: If we're parsing here, should we use the parsed data in later functions? I think I see that `addCastVoteRecordFileEntry` also parses the line, for one. Not sure if there are others. Maybe not an important optimization, but for large CVR files it might make a difference.

services/admin/src/store.ts
```diff
@@ -374,6 +421,22 @@ export class Store {
       return err({ kind: 'BallotIdRequired' });
     }
 
+    // Ensure CVR mode matches any previously imported CVRs:
```
jonahkagan, line 424: Is this check different from the one above? It looks the same to me (and thus unnecessary), but maybe I'm missing something

services/admin/src/store.ts
```diff
@@ -277,16 +285,55 @@ export class Store {
...
 
-          // TODO(https://github.com/votingworks/vxsuite/issues/2716): add error checking against
-          // attempts to mix "live" and "test" CVRs for the same election.
+          // Parse CVR data for validation.
+          const cvr = safeParseJson(line).unsafeUnwrap() as CastVoteRecord;
+          nextLineCvrMode = cvr._testBallot
+            ? Admin.CvrFileMode.Test
+            : Admin.CvrFileMode.Official;
+
+          // Ensure CVR mode matches previous file imports, if any:
```
jonahkagan, line 304: Do we need to check this for every line? Maybe we could just check this once at the end? (Again, probably not a necessary optimization, but it's in the critical path)

libs/api/src/client_utils.ts
jonahkagan, line 10: I might check with @eventualbuddha to see if we have an existing pattern for passing errors from the backend. I haven't worked with this codebase at all, so I'm not sure off the top of my head. I poked around a little and found [this test case](https://github.com/votingworks/vxsuite/blob/b891f3dd43b688fe9d1b3e561e227d5670bf8dbd/frontends/election-manager/src/lib/backends/admin_backend.test.ts#L114-L132) that looks like it shows how an error gets passed from backend to frontend.

Also in terms of code organization, I'm not sure, but maybe putting it in the same place as `fetchJson` would make sense?

services/admin/src/store.test.ts
```diff
@@ -271,6 +271,155 @@ test('add a CVR file entry without a ballot ID', async () => {
   );
 });
 
+test('add a CVR file entry matching an existing ballot ID with different data', async () => {
```
jonahkagan, line 274: Tests look great!

services/admin/src/server.ts
jonahkagan: Maybe the two different error cases handled here actually need two different error types?

services/admin/src/store.ts
```diff
@@ -277,16 +285,55 @@ export class Store {
...
+                'they mix live and test ballots',
+            });
+          }
+
+          lastLineCvrMode = nextLineCvrMode;
+
           const result = this.addCastVoteRecordFileEntry(electionId, id, line);
 
           if (result.isErr()) {
+            this.client.run('rollback transaction');
```
jonahkagan, line 336: Nice catch! I wonder if there's a way to ensure we always rollback the transaction on error (other than just being diligent)

libs/api/src/client_utils.ts
kofi-q: oh, thanks for that test case - that makes me wonder if maybe the intention is to just always return HTTP-200, even when we're returning errors. that reduces the necessity of this util, although this would still give us the schema/type safety.

@eventualbuddha - should api-generated errors be returned with status 200?

kofi-q: and re: co-locating with `fetchJson`, i initially started doing that and hit a circular reference issue when trying to reference the response types from `@votingworks/api` - maybe those types should just be in `@votingworks/types`? or we'd have to spin up a `@votingworks/api-client` lib?

there's possibly some crossover with what y'all are looking into WRT architecture changes?

services/admin/src/server.ts
kofi-q: yeah, good point, nice to have fresh eyes on this - i iterated a couple times on direction and hadn't taken a step back to simplify

services/admin/src/store.ts
kofi-q: yeah, this was slightly frustrating, but since this is a separate function, i thought we'd need a guard against trying to add a single CVR that didn't match the current file mode through some other code path down the line (i.e. not in the context of a full file import) - but, maybe that's not something we need to support and might be worth just merging this code with the `addCvrFile` method?

services/admin/src/store.ts
kofi-q: ah thanks, agreed - had made that change at some point and can't remember why i reverted!

services/admin/src/store.ts
kofi-q: that's worth looking into. long shot, but hoping there's a way to set sqlite transactions to time out from inactivity and roll back - at the very least, we can re-structure the code a bit

services/admin/src/store.ts
kofi-q: true, this feels like it could just go at the end

services/admin/src/store.ts
jonahkagan: I would support shrinking the surface of the public API since the helper is not called externally. It could still be a separate function in the file if it's helpful for readability

libs/api/src/client_utils.ts
jonahkagan: Yeah this feels like it will be exactly what we're gonna look at with the architecture changes. Though, as I said, I haven't worked with this code at all so your guess is as good as mine right now. I'll let @eventualbuddha shed some light

services/admin/src/store.ts
jonahkagan: Yeah, maybe we can have some sort of `withTransaction(doTransaction: (client) => void)` function that knows to rollback on `err` results and exceptions...

services/admin/src/server.ts
kofi-q: k, just spent some time refactoring to allow the client to interpret the error codes and it ended up ballooning out of scope (and getting into the realm of "changes that might be rendered obsolete by upcoming architecture changes") - will hold off for now until we have clarity on direction for our client-server communication story.

added a separate error code for `InvalidCvrFileMode` though, to disambiguate between the two error cases

services/admin/src/store.ts
kofi-q: ah yeah, i realised that the sub-method would have to re-stringify the CVR for storage anyway which is why i reverted earlier - but, with some refactoring, was able to avoid the double parsing and it's hopefully looking a little cleaner now

services/admin/src/store.ts
kofi-q: really like this - gonna follow up with another PR to implement!

kofi-q (commented):
will wait post-election to see where the dust settles on the discussion about server error propagation, but all other straightforward feedback addressed - thanks for being thorough!

libs/api/src/client_utils.ts
eventualbuddha: Up to now I've been trying to follow some basic HTTP principles:
- use REST & HTTP methods to say what you want to happen
- use status codes to indicate success (200, arguably 201 but at least 200) and failure (400, 404, 405, etc)
- use content-type to indicate the content … type, and it should almost always be `application/json`

So for this PR I'd say _don't_ expect that error responses will have 200 status codes. In the future I'm thinking of abstracting away most of the HTTP layer with some sort of RPC setup, but we're not there yet and may never be.

libs/api/src/client_utils.test.ts
```diff
@@ -0,0 +1,123 @@
...
+import { ErrorsResponse, ErrorsResponseSchema, OkResponse } from './base';
+import { fetchWithSchema } from './client_utils';
+
+type TestResponse =
+  | OkResponse<{
+      foo: 'bar';
+    }>
+  | ErrorsResponse;
+
+const TestSchema: z.ZodSchema<TestResponse> = z.union([
```
eventualbuddha, line 14:
```suggestion
const TestResponseSchema: z.ZodSchema<TestResponse> = z.union([
```

libs/api/src/client_utils.ts
```diff
@@ -0,0 +1,41 @@
...
+    throw new Error(
+      `invalid JSON received: '${responseText}' | status code: ${response.status}`
+    );
+  }
+
+  const responseJson = jsonParseResult.ok();
+  const parseResult = responseSchema.safeParse(responseJson);
+  if (!parseResult.success) {
+    throw new Error(
+      `invalid response received: ${parseResult.error} | status code: ${response.status}`
```
eventualbuddha, line 36: I've generally tried to use our wrappers: `safeParse` et al.
```suggestion
  const parseResult = safeParse(responseSchema, responseJson);
  if (parseResult.isErr()) {
    throw new Error(
      `invalid response received: ${parseResult.err()} | status code: ${response.status}`
```

services/admin/src/store.ts
```diff
@@ -277,25 +286,78 @@ export class Store {
...
+
+          if (!cvr._ballotId) {
+            return err({ kind: 'BallotIdRequired' });
+          }
+
+          // TODO(https://github.com/votingworks/vxsuite/issues/2716): add error checking
+          // to prevent importing CVRs for the wrong election.
+
+          containsTestBallots = containsTestBallots || cvr._testBallot;
+          containsLiveBallots = containsLiveBallots || !cvr._testBallot;
```
eventualbuddha, line 310:
```suggestion
          containsTestBallots ||= cvr._testBallot;
          containsLiveBallots ||= !cvr._testBallot;
```

eventualbuddha (approved):
Great to see progress on moving to the backend 🚀

services/admin/src/server.ts
jonahkagan: Ok, sounds good

jonahkagan (approved):

jonahkagan:
Thanks for revising! Looks great!

PR #2773, opened 2022-11-09
[VxAdmin] Move votecounting util to libs/utils

kofi-q:
## Overview
Related to https://github.com/votingworks/vxsuite/issues/2716

Moving the `votecounting.ts` util from `frontends/election-manager` to `libs/utils` to enable moving logic server-side (particularly the CVR parsing/validation logic).

Ran into an issue with `parseCvrsAndAssertSuccess` being imported directly from `votecounting.test.ts` in other test files - tried moving it to `libs/test-utils`, but we end up with a circular reference between that and `libs/utils`, so just ended up duplicating it to the two other test files that use it. Let me know you if think it's worth investing time into consolidating those.

Also, not sure why this change triggered detection of the obsolete test snapshot files, but had to remove those to get tests passing.

## Testing Plan 
- Local run to verify import-from-usb-drive flow isn't broken
- Relying on existing tests for more comprehensive regression guards

## Checklist


libs/utils/src/votecounting.ts
```diff
@@ -339,7 +339,13 @@ function makePairs<T>(inputArray: T[]): Array<Pair<T>> {
   const pairs = [];
   for (let i = 0; i < inputArray.length; i += 1) {
     for (let j = i + 1; j < inputArray.length; j += 1) {
-      pairs.push({ first: inputArray[i], second: inputArray[j] });
+      const first = inputArray[i];
+      const second = inputArray[j];
+      if (!first || !second) {
```
kofi-q, line 344: this was failing compilation after moving the file, apparently because the build config varies slightly between frontends/election-manager and libs/utils

libs/utils/jest.config.js
```diff
@@ -7,4 +7,14 @@ module.exports = {
   ...shared,
   testEnvironment: 'jsdom',
   setupFilesAfterEnv: ['<rootDir>/src/setupTests.ts'],
+  coverageThreshold: {
+    ...shared.coverageThreshold,
+    // Moved from frontends/election-manager with pre-existing low coverage:
+    ['src/votecounting.ts']: {
```
eventualbuddha, line 13: Unnecessary computed property.
```suggestion
    'src/votecounting.ts': {
```

libs/utils/src/votecounting.ts
eventualbuddha: It's the `noUncheckedIndexedAccess` enabled in the shared config that's disabled in `frontends/election-manager`.

eventualbuddha (approved):

Issue #2680, opened 2022-10-14
VxAdmin: loading a cvr for the wrong election is not rejected


mattroe:
Tested on NH elections. Steps to reproduce:

1. Program VxAdmin for Test Ballot
2. Import a CVR for Ashland
3. Total CVR count is shown as 0, reports are shown as 0, but the CVR is loaded
4. Write-in tab is populated with adjudication available. No write-in images are shown though.

Expected behavior: CVR should be rejected for not matching the election.

carolinemodic:
Adding that if the election is only slightly wrong, same contests but different scan thresholds for example, it will import the tallies of the wrong cvr file as well. This can only happen if you manually select the file.

mattroe:
closing in favor of https://github.com/votingworks/vxsuite/issues/2716

PR #2774, opened 2022-11-10
[VxAdmin] Block CVR file imports on failed validation

kofi-q:
## Overview

- Now using the `parseCvr()` util (used in the client) in the server
- Failing CVR file imports whenever any validation errors are found
- Showing generic messaging to the user on most errors, but adding special handling for the following, given the issue mentioned in https://github.com/votingworks/vxsuite/issues/2680:
  - Mismatched election - CVR contains contests not found in the election definition
  - Invalid vote - a yes/no or candidate choice has an invalid value
- Also updated `addCastVoteRecordFile()` to use a modified `DbClient.transaction` to avoid the chance of forgetting to roll back transactions when returning errors

## Demo Video or Screenshot

https://user-images.githubusercontent.com/264902/201155601-359d0b8d-f3bc-4035-bd1f-fa98bacef448.mov

Screenshot 2022-11-09 at 5 38 01 PM

Screenshot 2022-11-09 at 5 33 01 PM

Screenshot 2022-11-09 at 5 32 33 PM

## Testing Plan 
- Added test cases for the new error paths
- Verified locally

## Checklist

### DIFF
--- a/libs/db/src/client.test.ts
+++ b/libs/db/src/client.test.ts
@@ -122,7 +122,7 @@ test('read/write', () => {
   );
 });
 
-test('transactions', () => {
+test('transactions', async () => {
   const client = Client.memoryClient();
 
   client.exec(
@@ -134,7 +134,8 @@ test('transactions', () => {
     count: 1,
   });
 
-  expect(() =>
+  // Should roll back on exception:
+  await expect(
     client.transaction(() => {
       client.run('insert into muppets (name) values (?)', 'Fozzie');
       expect(client.one('select count(*) as count from muppets')).toEqual({
@@ -142,16 +143,54 @@ test('transactions', () => {
       });
       throw new Error('rollback');
     })
-  ).toThrow('rollback');
+  ).rejects.toThrow('rollback');
   expect(client.one('select count(*) as count from muppets')).toEqual({
     count: 1,
   });
 
-  client.transaction(() => {
+  // Should commit by default, if no exceptions occur:
+  await client.transaction(() => {
     client.run('insert into muppets (name) values (?)', 'Fozzie');
   });
   expect(client.one('select count(*) as count from muppets')).toEqual({
     count: 2,
+  });
+
+  // Should roll back if shouldCommit test returns false:
+  await expect(
+    client.transaction(
+      () => {
+        client.run('insert into muppets (name) values (?)', 'Gonzo');
+        expect(client.one('select count(*) as count from muppets')).toEqual({
+          count: 3,
+        });
+        return 'this is a result';
+      },
+      (result) => {
+        expect(result).toEqual('this is a result');
+        return false;
+      }
+    )
+  ).resolves.toEqual('this is a result');
+  expect(client.one('select count(*) as count from muppets')).toEqual({
+    count: 2,
+  });
+
+  // Should commit if shouldCommit test returns true:
+  await expect(
+    client.transaction(
+      async () => {
+        client.run('insert into muppets (name) values (?)', 'Gonzo');
+        return await Promise.resolve('another result');
+      },
+      (result) => {
+        expect(result).toEqual('another result');
+        return true;
+      }
+    )
+  ).resolves.toEqual('another result');
+  expect(client.one('select count(*) as count from muppets')).toEqual({
+    count: 3,
   });
 });
 
--- a/libs/db/src/client.test.ts
+++ b/libs/db/src/client.test.ts
@@ -134,8 +134,8 @@ test('transactions', async () => {
     count: 1,
   });
 
-  // Should roll back on exception:
-  await expect(
+  // Should roll back on synchronous exception:
+  expect(() =>
     client.transaction(() => {
       client.run('insert into muppets (name) values (?)', 'Fozzie');
       expect(client.one('select count(*) as count from muppets')).toEqual({
@@ -143,21 +143,35 @@ test('transactions', async () => {
       });
       throw new Error('rollback');
     })
+  ).toThrow('rollback');
+  expect(client.one('select count(*) as count from muppets')).toEqual({
+    count: 1,
+  });
+
+  // Should roll back on async exception:
+  await expect(() =>
+    client.transaction(() => {
+      client.run('insert into muppets (name) values (?)', 'Fozzie');
+      expect(client.one('select count(*) as count from muppets')).toEqual({
+        count: 2,
+      });
+      return Promise.reject(new Error('rollback'));
+    })
   ).rejects.toThrow('rollback');
   expect(client.one('select count(*) as count from muppets')).toEqual({
     count: 1,
   });
 
   // Should commit by default, if no exceptions occur:
-  await client.transaction(() => {
+  client.transaction(() => {
     client.run('insert into muppets (name) values (?)', 'Fozzie');
   });
   expect(client.one('select count(*) as count from muppets')).toEqual({
     count: 2,
   });
 
-  // Should roll back if shouldCommit test returns false:
-  await expect(
+  // Should roll back if `shouldCommit` test returns false:
+  expect(
     client.transaction(
       () => {
         client.run('insert into muppets (name) values (?)', 'Gonzo');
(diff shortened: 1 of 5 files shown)


libs/db/src/client.test.ts
```diff
@@ -134,25 +134,64 @@ test('transactions', () => {
...
+    count: 2,
+  });
+
+  // Should commit if shouldCommit test returns true:
+  await expect(
+    client.transaction(
+      async () => {
+        client.run('insert into muppets (name) values (?)', 'Gonzo');
+        return await Promise.resolve('another result');
+      },
```
eventualbuddha, line 185: This makes odd use of `async`/`await` and `Promise.resolve`. I get that you're doing it because of the linter yelling at you, but maybe `client.transaction` should work with either sync or async callback functions? See [`mapSheet`](https://github.com/votingworks/vxsuite/blob/370fa214a1a41def2f01ee157cc28f8b9f130d16/services/scan/src/types.ts#L13-L41) for an example of this.

services/admin/src/store.test.ts
```diff
@@ -412,6 +415,99 @@ test('add a CVR file with mixed live and test CVRs', async () => {
...
+  store.addElection(electionDefinition.electionData);
+
+  const { cvrFile } = primaryElectionSampleFixtures;
+
+  const result = await store.addCastVoteRecordFile({
+    electionId: 'not-a-real-election-id',
+    filePath: cvrFile.asFilePath(),
+  });
+
+  expect(result.isErr()).toBe(true);
```
eventualbuddha, line 431: Maybe not necessary? The following line would fail whenever this would fail, and probably with a better error message.

services/admin/src/store.ts
```diff
@@ -206,6 +221,45 @@ export class Store {
...
 
+  private convertCvrParseErrorsToApiError(
+    cvrParseResult: ParseCastVoteRecordResult
+  ): AddCastVoteRecordError {
+    // Find errors for which we want surface specific messaging to
+    // the user:
+    // [TODO: there's room for simplification here - currently leveraging
+    // older validation code in `parseCvrs` as a first step.]
+    for (const errorMessage of cvrParseResult.errors) {
+      if (errorMessage.includes('not in the election definition')) {
```
eventualbuddha, line 232: As your comment indicates, this isn't great. It's quite fragile, but I'm okay with it for now because there are tests. We should fix this soon, though.

services/admin/src/store.ts
```diff
@@ -296,21 +343,22 @@ export class Store {
             continue;
           }
 
-          // Parse CVR data for validation.
-          const cvr = safeParseJson(line).unsafeUnwrap() as CastVoteRecord;
+          const [cvrParseResult] = [...parseCvrs(line, election)];
```
eventualbuddha, line 346: Do we ever want to call `parseCvrs` with more than one line? Maybe simplify that function since we're doing the line streaming here. Alternatively, use an iterable-friendly approach?
```suggestion
          const cvrParseResult = take(1, parseCvrs(line, election));
```

services/admin/src/store.ts
```diff
@@ -360,23 +405,23 @@ export class Store {
...
         wasExistingFile: !!existing,
         newlyAdded,
         alreadyPresent,
       });
-    } catch (error) {
-      if (inTransaction) {
-        this.client.run('rollback transaction');
+    };
+
+    function shouldCommit(result: AddCastVoteResult) {
```
eventualbuddha, line 416: Return type annotation?
```suggestion
    function shouldCommit(result: AddCastVoteResult): boolean {
```

services/admin/src/store.ts
```diff
@@ -223,22 +277,15 @@ export class Store {
...
+  }): Promise<AddCastVoteResult> {
+    const election = this.getElection(electionId)?.electionDefinition.election;
+    if (!election) {
+      return err({ kind: 'InvalidElectionId' });
+    }
+
     const sha256Hash = await sha256File(filePath);
 
-    try {
+    const transactionFn = async (): Promise<AddCastVoteResult> => {
```
eventualbuddha, line 288: Use function declaration style?
```suggestion
    async function transactionFn(): Promise<AddCastVoteResult> {
```

services/admin/src/store.ts
```diff
@@ -360,23 +405,23 @@ export class Store {
...
-        this.client.run('rollback transaction');
+    };
+
+    function shouldCommit(result: AddCastVoteResult) {
+      if (analyzeOnly) {
+        return false;
       }
-      throw error;
+
+      return result.isOk();
```
eventualbuddha, line 421: Maybe this reads easier?
```suggestion
      return !analyzeOnly && result.isOk();
```

Or even inline it?
```ts
return await this.client.transaction(transactionFn, (result) => !analyzeOnly && result.isOk());
```

services/admin/src/server.ts
```diff
@@ -206,9 +206,9 @@ export function buildApp({ workspace }: { workspace: Workspace }): Application {
     { electionId: Id },
     Admin.DeleteCvrFileResponse,
     Admin.DeleteCvrFileRequest
-  >('/admin/elections/:electionId/cvr-files', (request, response) => {
+  >('/admin/elections/:electionId/cvr-files', async (request, response) => {
     const { electionId } = request.params;
-    store.deleteCastVoteRecordFiles(electionId);
+    await store.deleteCastVoteRecordFiles(electionId);
```
eventualbuddha, line 211: If `transaction` is made sync or async depending on the callback then this would no longer need to be async, right?

eventualbuddha (approved):
Some things that could be different, nothing blocking though.

services/admin/src/store.ts
kofi-q: ha! funny thing is, i had it exactly that way before and decided to expand it for readability over conciseness - i can go either way

libs/db/src/client.test.ts
kofi-q: was really only doing this in the test to verify that it works with async functions as well - will update `client.transaction` to the overloaded pattern for the other reasons pointed out, but this test  case might need to remain unchanged for coverage purposes

services/admin/src/store.test.ts
kofi-q: ah, true - blind copy-paste from old tests 😅

services/admin/src/store.ts
kofi-q: i assumed that would cause the references to `this` within the function to be overridden, but if we have a strong preference for function dec, i could maybe wrap it in a call and pass in the parent reference?

services/admin/src/store.ts
kofi-q: yeah, definitely opportunities for drastic simplification here - at the moment, we're in the in-between where it's being used both client and server-side, so post-refactoring, we can get surgical with `parseCvrs` and other older parts of the code. not even a fan of the fact that it conflates parsing with data validation, but that's a gripe for another day 🙃

libs/db/src/client.test.ts
eventualbuddha: Maybe just ditch the `async` and `await` and do `return Promise.resolve('another result');`?

services/admin/src/store.ts
eventualbuddha: I would like to hear more about what you think a better model for parsing/data validation would be. Are you primarily talking about splitting the JSON parsing from the object validation, or something different?

services/admin/src/store.ts
eventualbuddha: Ah, yes it would. Didn't notice the `this` in there. Our linters prefer function declaration but allow for arrow when `this` is used.

services/admin/src/store.ts
eventualbuddha: I don't feel strongly. I do like that the way you had it keeps the name which serves as a hint as to its function.

services/admin/src/store.ts
kofi-q: yeah, thinking along the lines of keeping things modular and separating the concerns of parsing a CVR object from a raw string vs validating that the values are valid for a given election.
contrived case: `parseCvrs` doesn't easily lend itself to validating an existing CVR object, for example, without first stringifying

PR #2662, opened 2022-10-12
[Internal] Scan Diagnostic Tool

eventualbuddha:
## Overview

Adds a very basic scan diagnostic tool to view scans and some information about them.

## Demo Video or Screenshot
skewed ballot - bubble matching

## Testing Plan 
Did a fair amount of manual testing by using it over the last week or so, and added some very basic automated testing. Currently the coverage threshold is 10%, though I hope to bring that up over time.

## Checklist
- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.
- [ ] I have added JSDoc comments to any newly introduced exports

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #2777, opened 2022-11-10
refactor(scan): extract CVR export from `Store`

eventualbuddha:
## Overview

There's a lot of code in exporting CVRs that isn't really the responsibility of `Store`, like resizing images. This commit extracts that functionality to make `Store` leaner and more able to be extracted from `services/scan` for other users like in https://github.com/votingworks/vxsuite/pull/2662.

Additionally, this replaces the pattern of "passing a writable stream" with "returns a readable stream" in the CVR export code. The core functions now just deal with async iterators and use `Readable.from` to convert those to streams, then `pipeline` to pipe those streams to writable streams like HTTP responses or file writers.

## Demo Video or Screenshot
n/a

## Testing Plan 
Tested exporting backup files and CVR files with 30 scanned ballots in BSD.

## Checklist


adghayes:
Taking a look!

services/scan/src/cvrs/export.ts
```diff
@@ -0,0 +1,184 @@
...
+}: {
+  store: Store;
+  skipImages?: boolean;
+  orderBySheetId?: boolean;
+}): NodeJS.ReadableStream {
+  return Readable.from(
+    mapAsync(
+      exportCastVoteRecords({ store, skipImages, orderBySheetId }),
+      (cvr) => `${JSON.stringify(cvr)}\n`
+    )
```
adghayes, line 182: My understanding of generating a [Readable from an iterator](https://nodejs.org/api/stream.html#streamreadablefromiterable-options) is that this will will mean items will become available in the stream as they are generated, as opposed to a Readable from an `Buffer` or `string`, correct?

Is there anything that we need to do to indicate that the Readable is closed and will no longer emit data once we've worked through all the sheets? Or is that being handled somewhere already?

services/scan/src/cvrs/export.test.ts
```diff
@@ -0,0 +1,590 @@
...
+      },
+    },
+  ]);
+  store.adjudicateSheet(sheetId);
+
+  stream = new streams.WritableStream();
+  await pipeline(exportCastVoteRecordsAsNdJson({ store }), stream);
+  expect(stream.toString()).toEqual(
+    expect.stringContaining(stateOfHamilton.election.precincts[0].id)
+  );
```
adghayes, line 263: I thought you had concerns about being able to test a `Readable` CVR export, but this seems pretty simple?

services/scan/src/precinct_scanner_app.ts
```diff
@@ -524,10 +526,13 @@ export async function buildPrecinctScannerApp(
...
-      response.on('end', () => {
-        store.setCvrsAsBackedUp();
-      });
+
+      await pipeline(
+        exportCastVoteRecordsAsNdJson({ store, skipImages }),
+        response
+      );
+
+      store.setCvrsAsBackedUp();
```
adghayes, line 535: Nice.

services/scan/src/store.ts
```diff
@@ -859,18 +821,11 @@ export class Store {
...
-  ): Promise<void> {
-    const electionDefinition = this.getElectionDefinition();
-
-    if (!electionDefinition) {
-      throw new Error('no election configured');
-    }
-
+  *forEachResultSheet(options?: {
+    orderBySheetId?: boolean;
+  }): Generator<ResultSheet> {
```
adghayes, line 828: Do we need testing for this, or do you consider it tested via the `export.test.ts` testing?

services/scan/src/importer.ts
```diff
@@ -537,7 +539,10 @@ export class Importer {
       return;
     }
 
-    await this.workspace.store.exportCvrs(writeStream);
+    await pipeline(
+      exportCastVoteRecordsAsNdJson({ store: this.workspace.store }),
+      writeStream
+    );
```
adghayes, line 545: We could just raise this up into `central_scanner_app` to match `precinct_scanner_app`, I'm not sure if that consistency is valuable.

services/scan/src/backup.ts
```diff
@@ -90,9 +90,10 @@ export class Backup {
     debug('added election.json to backup');
 
     debug('adding CVRs to backup...');
-    const cvrStream = new PassThrough();
-    void this.store.exportCvrs(cvrStream);
-    await this.addEntry('cvrs.jsonl', cvrStream);
+    await this.addEntry(
+      'cvrs.jsonl',
+      exportCastVoteRecordsAsNdJson({ store: this.store })
+    );
```
adghayes, line 96: Beautiful.

adghayes (commented):
Looks good to me! A few clarifying questions.

services/scan/src/cvrs/export.ts
eventualbuddha: Looking at the [implementation of `Readable.from`](https://github.com/nodejs/node/blob/badc631d778bd13043770a34091ead49291a73d9/lib/internal/streams/from.js#L88-L89), you can see that when the iterator is done it pushes `null` to the stream, which should end it properly. For a generator function `next` will automatically return `{ done: true }` when the function exits, so this should be sufficient without any explicit cleanup.

services/scan/src/cvrs/export.test.ts
eventualbuddha: Previously I was trying to test the internals of the stream because I was trying to build a `Readable` by hand. Now I'm just using the included utilities like `pipeline` because I'm trusting that `Readable.from` is doing the right thing.

services/scan/src/store.ts
eventualbuddha: I think it's reasonable to have at least some basic testing in `store.test.ts`. I'll add that.

services/scan/src/importer.ts
eventualbuddha: I'm gonna leave this alone for now.

adghayes (approved):

PR #2778, opened 2022-11-10
[VxAdmin] Add react query for analyzing CVR file contents before import

kofi-q:
## Overview

Adding a `use_cvr_file_analysis_query` hook  which leverages the existing `analyzeOnly` option flag in the `POST /admin/elections/:id/cvr-files` server endpoint.

This is in preparation for replacing the client-side file analysis in the `ImportCvrFilesModal` with server calls to get information about files discovered on the USB drive, although there's still a bit of work to do in follow-up PRs to return additional data we can use for that.

## Testing Plan 
- Backfilled a couple test cases for the `admin_backend.addCastVoteRecordFile()` function and added a case for the `analyzeOnly` path.
- Not hooked up to any user flows yet, so will test locally later as this gets fleshed out

## Checklist


kofi-q:
cc @eventualbuddha since you know this code best

but no rush, either of you, since i'm signing off for the week in a bit anyway

eventualbuddha (approved):

PR #2782, opened 2022-11-16
[VxAdmin] Remove client-side CVR pre-loading logic

kofi-q:
## Overview

- Removing client-side logic around pre-loading/parsing CVR files found on the USB drive to detect duplicate CVRs prior to importing.
- This simplifies the code a bit as we move CVR data management to the backend and should improve performance in cases where there are multiple large CVR files on the USB drive.
- Side-effects:
  - We no longer show duplicate CVR counts in the import modal (these are still surfaced in the confirmation modal after import)
  - All information displayed in the CVR table are now based solely on information from the filename (as was the case before the pre-loading logic was added)

More context around the decision in [this slack thread](https://votingworks.slack.com/archives/CEL6D3GAD/p1668541124423049)

### Before:
Screenshot 2022-11-16 at 9 55 05 AM

### After:
Screenshot 2022-11-15 at 4 53 02 PM

Screenshot 2022-11-15 at 4 53 14 PM

## Testing Plan 
- Updated existing tests
- Verified locally

## Checklist

### DIFF
--- a/frontends/election-manager/src/components/import_cvrfiles_modal.tsx
+++ b/frontends/election-manager/src/components/import_cvrfiles_modal.tsx
@@ -143,7 +143,14 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
   async function importSelectedFile(
     fileData: CastVoteRecordFilePreprocessedData
   ) {
-    await importCvrFile(new File([fileData.fileContent], fileData.name));
+    assert(window.kiosk);
+    // TODO: To avoid unnecessarily loading potentially large CVR files into
+    // memory, we might want to add a file streaming API to kiosk-browser, or
+    // update the backend API to support importing files with just a file path
+    // instead of file attachments.
+    const fileContent = await window.kiosk.readFile(fileData.path, 'utf-8');
+
+    await importCvrFile(new File([fileContent], fileData.name));
   }
 
   const processCastVoteRecordFileFromFilePicker: InputEventFunction = async (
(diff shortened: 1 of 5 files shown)


frontends/election-manager/src/utils/cast_vote_record_files.ts
```diff
@@ -186,38 +186,24 @@ export class CastVoteRecordFiles {
    * Parses the CVR files from `files` to determine the number of CVR entries
    * in the file and the number of those entries that are duplicates.
    */
```
jonahkagan, line 188: This comment seems out of date now

frontends/election-manager/src/utils/cast_vote_record_files.ts
```diff
@@ -186,38 +186,24 @@ export class CastVoteRecordFiles {
...
-    files: KioskBrowser.FileSystemEntry[],
-    election: Election
-  ): Promise<CastVoteRecordFilePreprocessedData[]> {
+  parseAllFromFileSystemEntries(
+    files: KioskBrowser.FileSystemEntry[]
+  ): CastVoteRecordFilePreprocessedData[] {
     const results: CastVoteRecordFilePreprocessedData[] = [];
     for (const file of files) {
       try {
         const importedFile = [...this.files].find((f) => f.name === file.name);
```
jonahkagan, line 195: nit: could change this from `find` to `some` to get a boolean back right away

frontends/election-manager/src/components/import_cvrfiles_modal.tsx
```diff
@@ -143,7 +143,14 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
   async function importSelectedFile(
     fileData: CastVoteRecordFilePreprocessedData
   ) {
-    await importCvrFile(new File([fileData.fileContent], fileData.name));
+    assert(window.kiosk);
+    // TODO: To avoid unnecessarily loading potentially large CVR files into
+    // memory, we might want to add a file streaming API to kiosk-browser, or
+    // update the backend API to support importing files with just a file path
+    // instead of file attachments.
```
jonahkagan, line 150: Agreed, I think it makes sense for the frontend to only know about file paths - it doesn't need to load file contents

jonahkagan (approved):
Looks great!

frontends/election-manager/src/components/import_cvrfiles_modal.tsx
eventualbuddha: In fact, I'd go so far as to say the backend should be the one that reads the USB drive and gives the frontend just enough info to render the table and communicate the selections back to the server. Maybe that's file paths, but maybe not even that!

Issue #2360, opened 2022-08-14
Reevaluate how to transfer write in images from scanning to VxAdmin


carolinemodic:
For the nh-demo we quickly implemented the ballot image transfer embedded within the CVR files. We should reevaulate this approach for the longer term, and do some load testing to see how big of files we can handle with this, or other, approaches and make sure we are comfortable with that load for a nh pilot.

eventualbuddha:
What numbers are we expecting for this? Put another way, what number of write-in images should we be testing with? Are we only transferring ballot images for sheets with write-ins or for everything?

mattroe:
@eventualbuddha realistically between 1,000-5,000 and probably closer to the smaller end of the scale

mattroe:
Current plan is to revisit this when we address VVSG-CVR/CDF work.

adghayes:
The switch to CDF format does not require that we change how we're transferring write-in images. CDF allows them to be embedded just as we're doing now. This means that the work can be decoupled, and can happen in either order.

To help me with contest - what are the advantages are there in not embedding? cc @eventualbuddha @carolinemodic.
- Avoids file size limitations
- Allows one to manually inspect ballot images in CVR 

There are disadvantages with switching, namely the complexity that all our CVRs, fixtures and in the wild, are now directories rather than single files. @mattroe do you know if that's the case with competitors?

PR #2791, opened 2022-11-17
feat(scan): save backups directly from the backend

eventualbuddha:
## Overview

Previously we would stream the data through the client and either download it (Chrome) or write it directly to a file (`kiosk-browser`). Now we directly write to the USB drive, ensuring that we can only write to paths matching `/media/**`.

This also adds the ability to split large files into 4GB chunks when writing to USB drives given the assumption that we're working with FAT32-formatted drives.

The file export functionality, including splitting, lives in a new library in part because `utils` is a large grab bag and adding file system-specific code makes loading `utils` in clients difficult (we have to stub `fs` and related modules).

Refs #2360 

## Demo Video or Screenshot
No changes to the UI other than removing the "Custom" option for where to save the backup.

## Testing Plan
Tested with artificially small maximum file sizes (10 MB) and verified that joining the files together with `cat` did yield a usable `.zip` file.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -413,6 +413,28 @@ jobs:
       - store_test_results:
           path: libs/cvr-fixture-generator/reports/
 
+  test-libs-data:
+    executor: nodejs
+    resource_class: xlarge
+    steps:
+      - checkout-and-install
+      - run:
+          name: Build
+          command: |
+            pnpm --dir libs/data build
+      - run:
+          name: Lint
+          command: |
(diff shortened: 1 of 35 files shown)


eventualbuddha:
@benadida, I ended up leaving the file splitting code in there. I think there's a chance we will use it in the future even if we end up going with a "many files in various directories" approach for backups and CVRs.

benadida (approved):
remove zip.ts

reviewed sync with @eventualbuddha

PR #2792, opened 2022-11-17
[VxAdmin] Add numCvrsImported to the GetCvrFilesResponse schema

kofi-q:
## Overview

Adding a `numCvrsImported` field to the `GET .../cvr-files` API response - used in the client when listing imported CVR files in the client.

Holds the count of *new* CVRs that were added when the file was imported (excluding duplicates from previous imports).

## Demo Video or Screenshot
Screenshot 2022-11-17 at 3 49 48 PM

## Testing Plan 
- Updated existing tests
- Verified API response locally

## Checklist


eventualbuddha (approved):
I read through the SQL statement carefully enough to convince myself it is plausibly correct, but SQL is not a strength of mine. I suggest @benadida take a quick look.

services/admin/src/store.ts
```diff
@@ -637,22 +637,39 @@ export class Store {
...
+          min(cvr_files.created_at) as created_at
+        from cvr_file_entries
+        join cvr_files on cvr_files.id = cvr_file_entries.cvr_file_id
+        where
+          cvr_files.election_id = ?
+        group by cvr_file_entries.cvr_id
+      )
+      group by id
       order by export_timestamp desc
     `,
```
benadida, line 666: Two thoughts:

1/ I'm confused as to why sqlite is NOT giving a warning / error on selecting fields that are neither in the group-by nor the result of aggregate operation. This query would fail in Postgres. Specifically, if you're selecting `filename`, then you must `group by filename` or select `max(filename)` or something like that.

2/ that query can be simplified with either a subquery or a double group by. Likely a subquery is easiest in this case, like this:

```
select
        cvr_file_id as id,
        filename,
        export_timestamp as exportTimestamp,
        (select count(*)
          from cvr_file_entries
          where cvr_file_id = cvr_files.id)
          as numCvrsImported,
        precinct_ids as precinctIds,
        scanner_ids as scannerIds,
        sha256_hash as sha256Hash,
        datetime(min(cvr_files.created_at), 'localtime') as createdAt
from cvr_files
      where election_id = ?
```

Notice the only change from the old query is the addition of this subquery as a new field in the select:
```
        (select count(*)
          from cvr_file_entries
          where cvr_file_id = cvr_files.id)
          as numCvrsImported,
```

kofi-q: was also surprised sqlite didn't require all the columns to be grouped/aggregated, but apparently, it's an implementation detail of sqlite that it's happy to fill in the non-grouped columns with data from a randomly chosen row!

the proposed subquery approach ends up counting `cvr_file_entries` rows that should be attributed to a different `cvr_files ` row - the challenge we have here is that for each `cvr_id`, we need to find the associated `cvr_files` row (of the potential multiple) with the earliest timestamp - so the `min` function needs to happen after we've joined `cvrs` and `cvr_files`.

kofi-q: update for anyone reading back on this thread:

had a chance to huddle with Ben and landed on some tweaks that clean up the subquery a bit and avoid having to list out the columns twice 🚀

PR #2794, opened 2022-11-17
Adam/clean runthrough

amcmanus:
## Overview
This improves the initial (and ongoing) dev experience related to getting vxsuite components up and running. After `script/bootstrap` has completed, every service and frontend component is ready for use, barring any weird edge-cases I haven't yet encountered.

This PR also changes how sudo is granted to a developer account. Adding to the `sudo` group may be convenient, but it's a bad practice we should avoid. Explicitly creating `/etc/sudoers.d/username` configs is cleaner and makes auditing/troubleshooting much more clear.

The bulk of this PR is related to the bootstrap process. Previously, all it did was run `make install` within the various `services/frontends` directories. It left developers in an in-between state that was confusing and not very well documented. With this change to the bootstrap process, developers can now run `pnpm start` in any configured frontend without worrying about related services or build steps. For consistency, I've updated every Makefile to include a `bootstrap` directive appropriate for it. That said, there are more opportunities for improvement here. To simplify things (for now), I added multiple package installs to the bootstrap script rather than updating the Makefile `install` directive everywhere since it would result in excessive duplication. Longer term, I think we should identify system-level packages and make them a separate configuration requirement. 

## Testing Plan 

Build a fresh VM from scratch. Confirm you can start up any frontend you wish. 

## Checklist
- [ n/a] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.
- [n/a ] I have added JSDoc comments to any newly introduced exports

- [n/a ] I have added a screenshot and/or video to this PR to demo the change

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ sudo whoami
 ```
 
 Automatically install and build all dependencies in the vxsuite repo with the 
-following command
+following command:
 
 ```sh
 ./script/bootstrap
--- a/README.md
+++ b/README.md
@@ -86,8 +86,16 @@ node -v # this should return 16.x.x
 pnpm -v # this should return 5.x.x
 ```
 
-Automatically install all dependencies in the vxsuite repo with the following
-command
+NOTE: The initial bootstrap will take some time. You may be prompted for your
+`sudo` password during the run. To minimize the chance of it happening, you can
+reset the timer by running a quick sudo command before the bootstrap:
+
+```sh
+sudo whoami
+```
+
+Automatically install and build all dependencies in the vxsuite repo with the 
+following command
 
 ```sh
 ./script/bootstrap
--- a/README.md
+++ b/README.md
@@ -42,13 +42,12 @@ the following commands.
 
 ```sh
 su - # this will prompt for the root password
-usermod -aG sudo <USERNAME> # use your user account username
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -86,8 +86,16 @@ node -v # this should return 16.x.x
...
+NOTE: The initial bootstrap will take some time. You may be prompted for your
+`sudo` password during the run. To minimize the chance of it happening, you can
+reset the timer by running a quick sudo command before the bootstrap:
+
+```sh
+sudo whoami
+```
+
+Automatically install and build all dependencies in the vxsuite repo with the 
+following command
```
eventualbuddha, line 98:
```suggestion
following command:
```

eventualbuddha (approved):
Great! Thanks for fixing this.

README.md
```diff
@@ -42,13 +42,12 @@ the following commands.
 
 ```sh
 su - # this will prompt for the root password
-usermod -aG sudo <USERNAME> # use your user account username
+echo "USERNAME ALL=(ALL:ALL) ALL" > /etc/sudoers.d/USERNAME # use your user account username
```
benadida, line 45: @amcmanus what's the insight behind this one? Is this just a better practice?

PR #2797, opened 2022-11-18
Fix Precinct Change Lag on VxScan

adghayes:
## Overview
Fix issues identified in [slack thread](https://votingworks.slack.com/archives/CEL6D3GAD/p1668803903643469). On changing the precinct, the UI lags and easily creates confusion for the election manager. And with pressing the button multiple times, possibly weird states.

The problem was introduced when we decided to limit when the precinct can be changed to only when there are 0 ballots. As a result of that, the backend has to do a non-trivial amount of work, and it takes longer. It is essentially the same change as switching modes.

So, I added a progress modal when you change the precinct.

## Demo Video or Screenshot

Since I made this video I changed the modal language from "Changing Precinct" to "Loading" to match the mode switch.
[precinct-change-progress.webm](https://user-images.githubusercontent.com/37960853/202814421-33ea20dd-3e0d-4d39-aa14-65a26e640160.webm)

### DIFF
--- a/frontends/precinct-scanner/src/app.test.tsx
+++ b/frontends/precinct-scanner/src/app.test.tsx
@@ -282,7 +282,7 @@ test('election manager and poll worker configuration', async () => {
   // Change mode as Election Manager
   mockTestModeChange(false);
   userEvent.click(await screen.findByText('Live Election Mode'));
-  await screen.findByText('Loading');
+  await screen.findByText('Switching to Live Election Mode');
   await advanceTimersAndPromises(1);
   expect(logger.log).toHaveBeenCalledWith(
     LogEventId.AuthLogin,
@@ -400,7 +400,7 @@ test('election manager and poll worker configuration', async () => {
     await screen.findByText('Delete All Election Data from VxScan')
   );
   userEvent.click(await screen.findByText('Yes, Delete All'));
-  await screen.findByText('Loading');
+  await screen.findByText('Deleting Election Data');
   await waitFor(() =>
     expect(
       fetchMock.calls('./precinct-scanner/config/election', {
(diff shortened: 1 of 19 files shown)


adghayes:
@jonahkagan and I have discussed getting rid of optimistic updating in VxScan, but not doing that here.

@beausmith I saw some weird behavior in development where, with the select dropdown, sometimes the change event would fire twice if there was a re-render triggered by the `onPress` callback. I avoided the issue by both disabling the select dropdown and short-circuiting the event handler if the precinct is being changed. I still am not quite sure of the underlying issue, though. Have you seen something like this?

adghayes:
Updated copy based on @mattroe's suggestion:

[screen-capture (12).webm](https://user-images.githubusercontent.com/37960853/202822496-a4694590-cc57-4b78-a588-ea8ece829278.webm)

adghayes:
The reason that changing precinct (and switching live mode) takes so long is that, on the backend, anytime we change any configuration value of the interpreter, we reload and process all the templates from store. AFAICT, this is unnecessary, so we could make a different change and just have these actions be a lot faster.

jonahkagan:
@adghayes can you point me to the slow code in question and we can think through whether we can optimize it easily?

adghayes:
In precinct scanner interpreter:

https://github.com/votingworks/vxsuite/blob/e277bf40347b48b7f703c4799c50e093ba21bae7/services/scan/src/precinct_scanner_interpreter.ts#L291-L293


In precinct scanner app: 

https://github.com/votingworks/vxsuite/blob/e277bf40347b48b7f703c4799c50e093ba21bae7/services/scan/src/precinct_scanner_app.ts#L71-L74

I think we could easily optimize it by allowing a partial config update to the interpreter config, which would allow us to change mark thresholds, mode, or precinct without reloading the templates as we are doing in the app.

jonahkagan:
Yeah, that makes sense. The goal of all of this "configuring the interpreter" is to cache data in memory that would take a long time to load from the db on every scan. I think you're right that the layouts are the only piece of data that really needs to be cached.

It might be nice to actually do that caching in the store, since the store would be a good place to know (a) what takes a long time to load and therefore needs to be cached, and (b) what mutations invalidate the cache for a specific piece of data.

adghayes:
closing this in favor of #2806

Issue #1660, opened 2022-03-30
Upgrade to React 18


eventualbuddha:
## Summary
React 18 is out! We should upgrade once the dust settles a bit.
- Announcement: https://reactjs.org/blog/2022/03/29/react-v18.html
- Upgrade guide: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

## Requirements
- `@testing-library/react` has polished support for React 18, not hacks or workarounds.
- Each library and frontend depending on React should depend on React 18.

## Non-Requirements
- We should not use any of the new features in React 18, such as `<Suspense>` or other concurrent rendering, if we don't absolutely have to in order to make things work. We should do the minimal changes required to complete the upgrade before considering such changes.

## Tasks
- [ ] upgrade libraries (e.g. `libs/ui`)
- [ ] upgrade frontends (e.g. `frontends/bmd`)

carolinemodic:
@eventualbuddha did we do this in the node upgrade or is this still pending?

adghayes:
I took a stab the past couple days at upgrading React 18 and found that this is a massive and tricky project. I'm going to cut my losses for now and work on more critical changes, but I want to write up all the challenges to save the next person some time.

Upgrading to React 18 requires a number of other major or minor updates to meet peer dependencies

- `@testing-library/react` v14
  - `@testing-library/render-hook` is removed as `renderHook` is now a part of mainline RTL
  - `@testing-library/user-event` v14
- `react-router-dom`, `react-modal`, `react-textarea-autosize`, etc. minor updates
- `stylelint` minor update, which introduces breaking changes that need to be addressed
- corresponding type package upgrades

The changes to `React` itself are minimal and don't require almost any changes, only tweaking how we render our apps at the root of each app and in the print root for the print utilities. But the effects for testing are enormous. 
- `userEvent` becomes an entirely `async` API
- because of React 18's improved `setState` batching, many of our loading states become inaccessible during tests. it's no longer possible to `await screen.findByText('Loading')` in many places because it's never rendered within the test environment
- although the goal of the update is to reduce annoying test warnings, in some places it introduces new warnings where we were doing some actions triggering state updates outside of the UI and not wrapping in `act`
- fake timers get all messed up, but I couldn't identify exactly how

[...]

PR #2801, opened 2022-11-22
Fix Test Warnings in VxAdmin

adghayes:
With the introduction of React Query, we are getting a lot test warnings about updates to unmounted components. This can make it hard to read the test output and can obscure more concerning warnings.

The cause seems to be that React Query is fetching and making state updates after tests have ended. RTL good practice would be to wait for changes on the screen before ending or continuing the test. But in some cases, there are no changes on the screen, because the data is simply refetching or it is fetching data for which we have a default value (so there is no change on screen). I'll comment to point out some of the different cases.

I don't necessarily intend to merge this PR, but I was curious to see what it took to sort out the warnings and to start a discussion around it.

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -730,7 +730,8 @@ test('tabulating CVRs', async () => {
 
   const reportPreview2 = getByTestId('report-preview');
   expect(within(reportPreview2).getAllByText('0').length).toBe(40);
-  // useQuery's in AppRoot are refetching data, and there's no change to wait on
+  // useQuery's in AppRoot are refetching data, and there's no change to wait on.
+  // TODO: Remove after upgrade to React 18, which does not warn in this case.
   await advanceTimersAndPromises();
 });
 
@@ -1109,7 +1110,7 @@ test('changing election resets sems, cvr, and manual data files', async () => {
   await screen.findByText('Currently tallying live ballots.');
   // We're waiting on a query for isOfficialResults. It has a default value,
   // so there is no change on the page to wait for before test ends.
-  // Await promises to avoid test warning.
+  // TODO: Remove after upgrade to React 18, which does not warn in this case.
   await advanceTimersAndPromises();
 });
 
@@ -1259,7 +1260,7 @@ test('election manager UI has expected nav', async () => {
   await screen.findByRole('heading', { name: 'L&A Testing Documents' });
   // We're waiting on a query for the file mode (live vs. test).
   // It has a default value, so there is no change on the page when loaded.
-  // Await promises to avoid test warning.
+  // TODO: Remove after upgrade to React 18, which does not warn in this case
   await advanceTimersAndPromises();
   userEvent.click(screen.getByText('Tally'));
   await screen.findByRole('heading', {
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -9,6 +9,7 @@ import {
   getAllByRole as domGetAllByRole,
   act,
   within,
+  waitForElementToBeRemoved,
 } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import fetchMock from 'fetch-mock';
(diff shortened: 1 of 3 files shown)


frontends/election-manager/src/app.test.tsx
```diff
@@ -494,7 +495,7 @@ test('printing ballots and printed ballots report', async () => {
   userEvent.click(screen.getByRole('button', { name: 'Continue' }));
 
   // Printing should now continue normally
-  await waitFor(() => getByText('Printing'));
```
adghayes, line 497: In this printing modal case, we simply weren't waiting for the modal to close before moving on. Not a React Query issue.

frontends/election-manager/src/app.test.tsx
```diff
@@ -726,6 +729,8 @@ test('tabulating CVRs', async () => {
   );
   const printableArea3 = getByTestId('printable-area');
   expect(within(printableArea3).getAllByText('0').length).toBe(40);
+  // useQuery's in AppRoot are refetching data, and there's no change to wait on
+  await advanceTimersAndPromises();
```
adghayes, line 733: Unclear to me what was being fetched and why, but warnings indicated it was one of the hooks in `AppRoot`.

frontends/election-manager/src/app.test.tsx
```diff
@@ -1086,6 +1091,11 @@ test('changing election resets sems, cvr, and manual data files', async () => {
 
   fireEvent.click(await screen.findByText('Tally'));
   getByText('No CVR files loaded.');
+  await screen.findByText('Currently tallying live ballots.');
+  // We're waiting on a query for isOfficialResults. It has a default value,
+  // so there is no change on the page to wait for before test ends.
+  // Await promises to avoid test warning.
+  await advanceTimersAndPromises();
```
adghayes, line 1098: Case where we have a default value and there's nothing we could wait on.

frontends/election-manager/src/app.test.tsx
```diff
@@ -1232,6 +1242,10 @@ test('election manager UI has expected nav', async () => {
   await screen.findAllByText('View Ballot');
   userEvent.click(screen.getByText('L&A'));
   await screen.findByRole('heading', { name: 'L&A Testing Documents' });
+  // We're waiting on a query for the file mode (live vs. test).
+  // It has a default value, so there is no change on the page when loaded.
+  // Await promises to avoid test warning.
+  await advanceTimersAndPromises();
```
adghayes, line 1248: This can happen in the middle of a test as well, not just at the end.

frontends/election-manager/src/screens/manual_data_import_precinct_screen.test.tsx
```diff
@@ -36,8 +42,13 @@ test('displays error screen for invalid precinct', () => {
...
   );
-  getByText('Error: Could not find precinct 12345.');
-  getByText('Back to Index');
+  await screen.findByText('Error: Could not find precinct 12345.');
+  screen.getByText('Back to Index');
+  // There's no change in the invalid precinct case after the writeInQuery
+  // fires, so we just have to wait to avoid a test warning.
+  await act(async () => {
+    await sleep(1);
+  });
```
adghayes, line 51: In this case, `advancePromises` didn't do the trick. I actually had to sleep. Could probably do the same by using fake timers and using `advanceTimersAndPromises`

frontends/election-manager/src/screens/write_ins_screen.test.tsx
```diff
@@ -9,6 +10,14 @@ import { WriteInsScreen } from './write_ins_screen';
 const { electionDefinition, cvrData } = electionMinimalExhaustiveSampleFixtures;
 const abbreviatedCvrData = cvrData.split('\n').slice(0, 20).join('\n');
 
+afterEach(async () => {
+  // Several tests on this page create test warnings because hooks run after
+  // the end of the test, and there is no specific change on the page to check
+  await act(async () => {
+    await sleep(1);
+  });
```
adghayes, line 18: First three tests in this file all had the same problem.

adghayes:
@kofi-q my understanding from our discussion is that, if we want to avoid warnings in the present, we may want to merge these little fixes. I added comments with `TODO`'s that indicate to remove random waits after we complete #1660 and upgrade to React 18.

kofi-q (approved):
thanks for digging into this!

PR #2803, opened 2022-11-23
State architecture proof of concept

jonahkagan:


apps/vx-scan/backend/src/server.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  },
+};
+
+const api = grout.createApi({
+  getConfig: grout.query(async (): Promise<Config | undefined> => {
+    return store.config;
+  }),
+
+  updateConfig: grout.mutation(async (newConfig: Config): Promise<void> => {
```
benadida, line 40: can we add `mutationWithSideEffect` or something like it that maps to POST? I'm thinking of something like `printBallot`, that isn't idempotent, vs. `saveState` which is.

PR #2804, opened 2022-11-23
[VxAdmin] Add API and client for fetching CVRs for the current election

kofi-q:
## Overview

- Adding a `GET .../cvrs` endpoint to the admin service for fetching all CVRs for a given election.
- Also adding the relevant client code and react-query hook for interacting with the server endpoint.
- Will be used for generating tallies in the client, in lieu of the CVRs from local storage

## Demo Video or Screenshot
Screenshot 2022-11-23 at 8 30 24 AM

## Testing Plan 
- New test cases for the new API handler and client code
- Manual verification that API call returns expected data

## Checklist

### DIFF
--- a/frontends/election-manager/src/lib/backends/memory_backend.ts
+++ b/frontends/election-manager/src/lib/backends/memory_backend.ts
@@ -150,6 +150,18 @@ export class ElectionManagerStoreMemoryBackend
     );
   }
 
+  getCvrs(): Promise<CastVoteRecord[]> {
+    if (!this.electionDefinition) {
+      throw new Error('Election definition must be configured first');
+    }
+
+    if (!this.castVoteRecordFiles) {
+      return Promise.resolve([]);
+    }
+
+    return Promise.resolve([...this.castVoteRecordFiles.castVoteRecords]);
+  }
+
   loadCastVoteRecordFiles(): Promise<CastVoteRecordFiles | undefined> {
     return Promise.resolve(this.castVoteRecordFiles);
   }
--- a/services/admin/src/server.ts
+++ b/services/admin/src/server.ts
@@ -125,6 +125,10 @@ export function buildApp({ workspace }: { workspace: Workspace }): Application {
     }
   );
 
+  // TODO(https://github.com/votingworks/vxsuite/issues/2613): This endpoint
+  // can be removed once we've moved tally computation to the server - it's
+  // currently only used as a stopgap while we migrate all app state to the
+  // server.
   app.get<{ electionId: Id }, Admin.GetCvrsResponse>(
     '/admin/elections/:electionId/cvrs',
     (request, response) => {
@@ -135,6 +139,12 @@ export function buildApp({ workspace }: { workspace: Workspace }): Application {
             (entry) =>
               safeParseJson(entry.data).unsafeUnwrap() as CastVoteRecord
           )
+          .map((cvr) => ({
+            ...cvr,
+            // Strip out ballot images to keep the response size low, since
+            // they're not needed client-side.
+            _ballotImages: undefined,
+          }))
       );
     }
   );
--- a/services/admin/src/server.ts
+++ b/services/admin/src/server.ts
@@ -3,12 +3,14 @@ import { LogEventId, Logger, LogSource } from '@votingworks/logging';
 import {
   BallotPageLayout,
   CandidateContest,
+  CastVoteRecord,
   getBallotStyle,
   getContests,
   Id,
   InlineBallotImage,
   Rect,
   safeParse,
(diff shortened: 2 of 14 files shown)


services/admin/src/server.ts
```diff
@@ -121,6 +123,20 @@ export function buildApp({ workspace }: { workspace: Workspace }): Application {
...
+        store
+          .getCastVoteRecordEntries(request.params.electionId)
+          .map(
+            (entry) =>
+              safeParseJson(entry.data).unsafeUnwrap() as CastVoteRecord
+          )
+      );
+    }
+  );
+
```
adghayes, line 139: My understanding is that this will return all the CVRs for an election. Which, on VxAdmin, could theoretically be several GBs worth of data. I'm not sure what our limitations are, but I'm wondering if we hit a scaling problem here eventually, especially considering write-in data. cc @eventualbuddha

frontends/election-manager/src/lib/backends/memory_backend.ts
```diff
@@ -150,6 +150,18 @@ export class ElectionManagerStoreMemoryBackend
...
+      throw new Error('Election definition must be configured first');
+    }
+
+    if (!this.castVoteRecordFiles) {
+      return Promise.resolve([]);
+    }
+
+    return Promise.resolve([...this.castVoteRecordFiles.castVoteRecords]);
+  }
+
```
adghayes, line 164: In theory, should we be writing tests for `MemoryBackend`? I'm not suggesting you edit this PR to do so (I don't want to have to write tests for `MemoryBackend` 😄 ) but wondering if it would be good practice to. Or perhaps testing our test utilities is a rabbit hole.

adghayes (approved):
LGTM! A couple questions I'm thinking about but no changes suggested.

frontends/election-manager/src/lib/backends/memory_backend.ts
kofi-q: 😅 yeah, I think we would ideally make sure there's test coverage for this.
I'm sort of hoping that once the dust settles on the new architecture push, we're ok with moving over to spinning up a real server (with the in-memory sqlite instance) for our functional tests and using that instead of duplicating server behaviour here. It does feel a bit annoying to maintain `MemoryBackend`, but we can evaluate the test performance hit of using a real server (which I'm guessing won't be much)

services/admin/src/server.ts
kofi-q: Hm, if we're talking multiple GBs of data, then it might be worth blocking this project on moving tally computation to the server first 😩

Do either of you have a sense of the upper limit of the data we're talking about here? Piecing together conversations I've overheard since joining, it sounds like that might be due to including image data in the CVR JSON? Or did I misunderstand that?

frontends/election-manager/src/lib/backends/memory_backend.ts
kofi-q: cc @carolinemodic

As mentioned in eng planning, I'm thinking the options here are:
1. [S] Punt on reading raw CVR data from the backend and schedule some future time for moving tally computation to the backend.
2. [M] Invest in making sure this temporary API can handle large lists of CVRs (e.g. add support for paging)
3. [L?] Opportunistically move tally computation to the backend as part of this project and avoid investing in the temporary API.

Maybe (2) is a good compromise here, so we don't stay in the state of having two sources of truth for raw CVR data?

kofi-q: oops, thanks! 😅
will move to the right thread

services/admin/src/server.ts
kofi-q: cc @carolinemodic

As mentioned in eng planning, I'm thinking the options here are:

[S] Punt on reading raw CVR data from the backend and schedule some future time for moving tally computation to the backend.
[M] Invest in making sure this temporary API can handle large lists of CVRs (e.g. add support for paging)
[L?] Opportunistically move tally computation to the backend as part of this project and avoid investing in the temporary API.
Maybe (2) is a good compromise here, so we don't stay in the state of having two sources of truth for raw CVR data?

kofi-q: Got a suggestion from @carolinemodic to strip out the image data from the `_ballotImages` field before returning the CVRs, in order to keep the response size low - this seems like a great stopgap before we eventually move tally computation server-side.

Did some more digging on http `content-length` limits and it does seem like some browsers have limits, but not chrome, AFAICT. All the same, I'm also going to add the express `compression` module, which reduces the ~730kB response in the PR description to 23kB (for 3000 CVRs)

PR #2806, opened 2022-11-28
Fix Precinct Change Lag on VxScan Backend

adghayes:
## Overview
Instead of #2797, in which I introduced loading screens to handle the slow precinct change, here we optimize on the backend to make the operation faster. This also makes switching modes faster (see video), although switch modes could take longer than in the video if there were more ballots scanned.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/37960853/204388692-dac9580c-6d5d-44c8-b8dd-24581fb959a1.mov

### DIFF
--- a/frontends/precinct-scanner/src/screens/election_manager_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/election_manager_screen.tsx
@@ -111,7 +111,11 @@ export function ElectionManagerScreen({
       openToggleLiveModeWarningModal();
     } else {
       setIsLoading(true);
+      const minimumDelay = new Promise((resolve) => {
+        setTimeout(resolve, 2000);
+      });
       await toggleLiveMode();
+      await minimumDelay;
       setIsLoading(false);
     }
   }
--- a/services/scan/src/precinct_scanner_app.ts
+++ b/services/scan/src/precinct_scanner_app.ts
@@ -31,34 +31,6 @@ const debug = makeDebug('scan:precinct-scanner:app');
 
 type NoParams = never;
 
-/**
- * Loads all of the relevant configuration from the workspace store and
- * configures the interpreter. Should be called anytime one of these config
- * values is changed in order to update the interpreter's config.
- */
-async function updateInterpreterConfig(
-  interpreter: PrecinctScannerInterpreter,
-  workspace: Workspace
-) {
-  const { store } = workspace;
-  const electionDefinition = store.getElectionDefinition();
-  const precinctSelection = store.getPrecinctSelection();
-  if (!electionDefinition || !precinctSelection) {
-    interpreter.unconfigure();
-  } else {
-    const layouts = await store.loadLayouts();
-    assert(layouts);
-    interpreter.configure({
-      electionDefinition,
-      precinctSelection,
-      layouts,
-      testMode: store.getTestMode(),
-      markThresholdOverrides: store.getMarkThresholdOverrides(),
-      ballotImagesPath: workspace.ballotImagesPath,
-    });
-  }
-}
-
 export async function buildPrecinctScannerApp(
   machine: PrecinctScannerStateMachine,
   interpreter: PrecinctScannerInterpreter,
@@ -67,8 +39,8 @@ export async function buildPrecinctScannerApp(
 ): Promise<Application> {
   const { store } = workspace;
 
-  // Try loading interpreter configuration from the store on startup
-  await updateInterpreterConfig(interpreter, workspace);
+  // Trigger the layouts to be cached in the store
+  await store.loadLayouts();
 
   const app: Application = express();
   const upload = multer({
@@ -101,7 +73,7 @@ export async function buildPrecinctScannerApp(
     NoParams,
     Scan.PatchElectionConfigResponse,
     Scan.PatchElectionConfigRequest
-  >('/precinct-scanner/config/election', async (request, response) => {
+  >('/precinct-scanner/config/election', (request, response) => {
     const { body } = request;
 
     if (!Buffer.isBuffer(body)) {
@@ -144,7 +116,6 @@ export async function buildPrecinctScannerApp(
       store.setPrecinctSelection(
         singlePrecinctSelectionFor(electionDefinition.election.precincts[0].id)
       );
-      await updateInterpreterConfig(interpreter, workspace);
     }
     response.json({ status: 'ok' });
   });
@@ -179,7 +150,7 @@ export async function buildPrecinctScannerApp(
     NoParams,
     Scan.PatchPrecinctSelectionConfigResponse,
     Scan.PatchPrecinctSelectionConfigRequest
-  >('/precinct-scanner/config/precinct', async (request, response) => {
+  >('/precinct-scanner/config/precinct', (request, response) => {
     const bodyParseResult = safeParse(
       Scan.PatchPrecinctSelectionConfigRequestSchema,
       request.body
@@ -210,7 +181,6 @@ export async function buildPrecinctScannerApp(
 
     store.setPrecinctSelection(bodyParseResult.ok().precinctSelection);
     workspace.resetElectionSession();
-    await updateInterpreterConfig(interpreter, workspace);
     response.json({ status: 'ok' });
   });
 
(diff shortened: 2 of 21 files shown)


services/scan/src/store.ts
```diff
@@ -167,10 +171,11 @@ export class Store {
   }
 
   /**
-   * Resets the database.
+   * Resets the database and any cached data in the store.
    */
   reset(): void {
     this.client.reset();
+    this.cachedLayouts = [];
```
jonahkagan, line 178: Are there any other times we need to invalidate the cache? E.g. when a new election is configured?

services/scan/src/precinct_scanner_app.ts
```diff
@@ -78,7 +46,7 @@ async function updateInterpreterConfig(
   if (!electionDefinition || !precinctSelection) {
     interpreter.unconfigure();
   } else {
-    const layouts = await loadLayouts(store);
+    const layouts = await store.loadLayouts();
```
jonahkagan, line 49: Now that layouts are cached, we can do all of this "configuring the interpreter" on each scan, instead of caching a reference to the interpreter. All of the rest of the data should be very quick to load from the db. This will remove the scariest part of this module - calling `updateInterpreterConfig` any time an API request modifies any of the interpreter args.

Currently we're in a bit of a halfway state where we have two layers of caching - one in the store, and then one in this file. Doesn't necessarily need to happen in this PR (and I'm happy to do it myself if you'd like, since I'll be working in this file soon), but I do think it will be a big win.

Thoughts?

services/scan/src/store.ts
```diff
@@ -75,6 +78,7 @@ function dateTimeFromNoOffsetSqliteDate(noOffsetSqliteDate: string): DateTime {
  */
 export class Store {
   private constructor(private readonly client: DbClient) {}
+  private cachedLayouts: BallotPageLayoutWithImage[] = [];
```
jonahkagan, line 81: I noticed in the video that this caching made it so fast that the screen flashes when you switch modes. I think that's a very short loading modal, right? Any tweak we could make to avoid the flash?

Some ideas:
- Only start the loading modal after 500ms of waiting
- Experimentally determine a threshold of ballots scanned for which we think a loading modal is necessary
- Add an artificial minimum delay on the frontend so the modal always shows for at least a second or two

jonahkagan (commented):
Looks great so far! Just a few thoughts

services/scan/src/store.ts
adghayes: When are the HMPB templates cleared from the database? Only when this `reset` function is called, which happens on unconfigure. So currently this is the only place we need to invalidate the cache. 

We only configure a new election after unconfigure. Separate from this caching question, maybe we should error out if we try to set the election without unconfiguring first.

services/scan/src/store.ts
adghayes: I think an artificial delay sounds easiest. Doing the modal after 500ms could be a problem if the operation finishes at 501ms, and the experimentally determined threshold may be sensitive to later changes we make. I'm not even sure if it will get much slower with more ballots, because delete operations in the DB are fast. Not sure about the filesystem though.

adghayes: I'm going to see if it actually slows down with lots of ballots.

services/scan/src/precinct_scanner_app.ts
adghayes: The `updateInterpreterConfig` thing is a real footgun and I agree, we should follow your suggestion. I'll try it here.

services/scan/src/store.ts
jonahkagan: Ok great, that was my guess that this was the only place, but wanted to make sure. I do like the idea of putting in more guardrails somehow so that we make sure we don't mess up and forget to invalidate the cache if we change this down the road.

services/scan/src/precinct_scanner_app.ts
adghayes: I moved it in the latest commit, and tested to see that it adds an average of 5 milliseconds before each scan. Maybe we could speed that up by hitting the database fewer times. But does that seem acceptable?

services/scan/src/store.ts
jonahkagan: Good call

services/scan/src/precinct_scanner_app.ts
jonahkagan: What's the overall time of the scan? How big of a percentage of that time is 5ms?

adghayes: The scan is a little more than 5 seconds for simple ballots, could be longer. So 5ms is like 0.1%. Seems OK?

services/scan/src/store.ts
adghayes: Added a minimum 2 second delay.

services/scan/src/precinct_scanner_app.ts
jonahkagan: Yeah I'd say that's more than ok

services/scan/src/precinct_scanner_app.ts
```diff
@@ -99,8 +39,8 @@ export async function buildPrecinctScannerApp(
 ): Promise<Application> {
   const { store } = workspace;
 
-  // Try loading interpreter configuration from the store on startup
-  await updateInterpreterConfig(interpreter, workspace);
+  // Trigger the layouts to be cached in the store
+  await store.loadLayouts();
```
jonahkagan, line 43: Do we need to do this here? Doesn't it happen automatically when the store is initialized now?

services/scan/src/precinct_scanner_app.ts
```diff
@@ -618,6 +548,22 @@ export async function buildPrecinctScannerApp(
...
+      assert(precinctSelection);
+      assert(layouts);
+      interpreter.configure({
+        electionDefinition,
+        precinctSelection,
+        layouts,
+        testMode: store.getTestMode(),
+        markThresholdOverrides: store.getMarkThresholdOverrides(),
+        ballotImagesPath: workspace.ballotImagesPath,
+      });
```
jonahkagan, line 565: I was imagining going one level deeper here. If you look at interpreter, it's basically a very thin wrapper around the interpret functions that caches the config. I think we can delete that wrapper and simply make a function `interpret` that loads the config from the store and passes it to the interpret functions. Then the state machine can call that interpret function (it has access to the store already).

This works though, so I'm open to stopping here if you feel strongly.

jonahkagan (approved):

services/scan/src/precinct_scanner_app.ts
adghayes: oops! meant to remove that.

services/scan/src/precinct_scanner_app.ts
adghayes: I don't feel strongly, but I want to close this, so will leave as-is.

frontends/precinct-scanner/src/screens/election_manager_screen.tsx
```diff
@@ -111,7 +111,11 @@ export function ElectionManagerScreen({
       openToggleLiveModeWarningModal();
     } else {
       setIsLoading(true);
+      const minimumDelay = new Promise((resolve) => {
+        setTimeout(resolve, 2000);
+      });
```
eventualbuddha, line 116: Could have been this:
```suggestion
      const minimumDelay = sleep(2000);
```

services/scan/src/store.ts
```diff
@@ -1084,6 +1098,8 @@ export class Store {
       JSON.stringify(metadata),
       JSON.stringify(layouts)
     );
+    // cache our layouts so they can be immediately used by the interpreter
+    this.cachedLayouts.push(...layoutsWithImages);
```
eventualbuddha, line 1102: While it happens that `addHmpbTemplate` is called sequentially with each layout in the same order that `getHmpbTemplates` would load them, nothing actually guarantees that's the case. If the order changed for any reason it would result in interpretation bugs.

services/scan/test/helpers/mock_layouts.ts
```diff
@@ -0,0 +1,47 @@
...
+    contests: [],
+  };
+}
+
+export function getMockImageData(width = 1, height = 1): ImageData {
+  return {
+    data: Uint8ClampedArray.of(0, 0, 0, 0),
+    width,
+    height,
+  };
```
eventualbuddha, line 24:
```suggestion
  return createImageData(width, height);
```

Issue #2800, opened 2022-11-21
Non-zero totals


mattroe:
VVSG 1.1.3-B:

> The voting system must not enter the voting mode until all steps necessary to isolate test data from election data have been performed successfully and all vote counters have been zeroed. An attempt to open polls with non-zero counters: must be recorded in the audit log and an election worker must be clearly notified of the event

adghayes:
We are not implementing a similar warning for the BMD currently per @mattroe, although would definitely have to do so for a combined marking + casting device.

adghayes:
We have implemented on VxScan in #2809, but we need to remember to implement this on VxMarkScan (or whatever) if we get to that. Leaving open for however you want to track @carolinemodic @mattroe.

adghayes:
@mattroe based on my reading of 1.1.3-B, I believe we need to add a warning to VxMarkScan if polls are opened with non-zero vote total. Even though VxMarkScan is not tallying votes, it is still a voting device which enters a voting mode so I believe this applies.

VxMark can be avoided because it's not included in certification.

@mattroe let me know if you agree or disagree with my interpretation, depending on which I will implement this or not.

adghayes:
Implemented where relevant, because it was deemed not necessary for VxMarkScan (now VxMark).

PR #2809, opened 2022-11-29
Non-Zero Total Warning on VxScan

adghayes:
## Overview
To meet VVSG 2.0 1.1.3-B as part of #2800. If we're only doing the warning on VxScan, it closes that issue. If we have to do it on VxMark as well, will make a separate PR. Please review copy below, knowing that this copy will almost certainly never be seen in production. 

I also added logging to the regular polls closed/open/paused actions in a second commit, as this was something missed in the first pass.

## Demo Video or Screenshot

Screen Shot 2022-11-29 at 9 49 29 AM

## Testing Plan

Automated testing and added this screen to the Preview App for easy viewing.

## Checklist

### DIFF
--- a/frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
+++ b/frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
@@ -43,6 +43,7 @@ import {
   PollsTransition,
   Optional,
 } from '@votingworks/types';
+import { LogEventId } from '@votingworks/logging';
 import {
   CenteredLargeProse,
   ScreenMainCenterChild,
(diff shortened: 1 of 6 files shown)


frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
```diff
@@ -349,6 +385,13 @@ export function PollWorkerScreen({
     }
     setCurrentPollsTransitionTime(timePollsTransitioned);
     updatePollsState(getPollsTransitionDestinationState(pollsTransition));
+    await logger.log(
```
carolinemodic, line 388: I realize I'm being a little silly here but can we add a message in the polls_opened case that states that there are zero ballots scanner, or add as a seperate param the ballotsScanned count. Just so we make sure we are extra explicit in the success case that we have checked this.

carolinemodic: I don't think it would hurt to include the ballotsScanned count on all poll transition logs so that might be easiest.

carolinemodic (approved):
1 nit but otherwise lgtm

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
adghayes: Done

PR #2812, opened 2022-11-29
[VxAdmin] Update client to use CVR file metadata from the server

kofi-q:
## Overview
Closes https://github.com/votingworks/vxsuite/issues/2716:

- Replacing all references to CVR data in client-side local storage with queries for server-side data.
- Removing all logic around writing to and reading CVR data from local storage.
- Leaving the `CastVoteRecords` class mostly intact, since it's used with our `memory_backend` test utility.

## Demo Video or Screenshot
https://user-images.githubusercontent.com/264902/204634905-846ff00f-13c9-4243-9e34-934cade05b7b.mov

## Testing Plan 
- Updated existing tests to account for the changes
- Manual functional verification, clicking through the various user flows that read/write CVR data and also checking for console errors/warnings:
  - Tally screen + import/delete modals
  - Write-ins screen
  - Reports screen and tally report

## Checklist

### DIFF
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -1107,6 +1107,10 @@ test('changing election resets sems, cvr, and manual data files', async () => {
     fireEvent.click(getByText('Load Demo Election Definition'));
   });
 
+  await backend.configure(
+    electionFamousNames2021Fixtures.electionDefinition.electionData
+  );
+
   userEvent.click(screen.getByText('Lock Machine'));
   await authenticateWithElectionManagerCard(
     card,
--- a/frontends/election-manager/src/app.test.tsx
+++ b/frontends/election-manager/src/app.test.tsx
@@ -583,7 +583,9 @@ test('tabulating CVRs', async () => {
     expect.anything()
   );
 
-  fireEvent.click(getByText('Mark Tally Results as Official'));
+  const markOfficialButton = getByText('Mark Tally Results as Official');
+  await waitFor(expect(markOfficialButton).toBeEnabled);
+  fireEvent.click(markOfficialButton);
   getByText('Mark Unofficial Tally Results as Official Tally Results?');
   const modal = await screen.findByRole('alertdialog');
   fireEvent.click(within(modal).getByText('Mark Tally Results as Official'));
@@ -711,7 +713,12 @@ test('tabulating CVRs', async () => {
 
   // Clear results
   fireEvent.click(getByText('Tally'));
-  fireEvent.click(getByText('Clear All Tallies and Results'));
+
+  const clearTalliesButton = await screen.findByText(
+    'Clear All Tallies and Results'
+  );
+  expect(clearTalliesButton).toBeEnabled();
+  fireEvent.click(clearTalliesButton);
   fireEvent.click(getByText('Remove All Data'));
   await waitFor(() => expect(getByText('No CVR files loaded.')));
   expect(logger.log).toHaveBeenCalledWith(
@@ -767,7 +774,7 @@ test('tabulating CVRs with SEMS file', async () => {
   );
 
   fireEvent.click(getByText('Tally'));
-  getByText('External Results (sems-results.csv)');
+  await screen.findByText('External Results (sems-results.csv)');
 
   fireEvent.click(getByText('Reports'));
   getByText('External Results (sems-results.csv)');
@@ -847,7 +854,7 @@ test('tabulating CVRs with SEMS file', async () => {
 
   // Test removing the SEMS file
   fireEvent.click(getByText('Tally'));
-  fireEvent.click(getByText('Remove External Results File'));
+  fireEvent.click(await screen.findByText('Remove External Results File'));
   fireEvent.click(getByText('Remove External Files'));
   await waitFor(() =>
     expect(getByTestId('total-cvr-count').textContent).toEqual('100')
@@ -886,7 +893,8 @@ test('tabulating CVRs with SEMS file and manual data', async () => {
   );
 
   fireEvent.click(getByText('Tally'));
-  expect(getByTestId('total-cvr-count').textContent).toEqual('200');
+  await advanceTimersAndPromises(5); // Allow async queries to resolve first.
+  expect(await screen.findByTestId('total-cvr-count')).toHaveTextContent('200');
 
   fireEvent.click(getByText('Add Manually Entered Results'));
   getByText('Manually Entered Precinct Results');
@@ -914,7 +922,7 @@ test('tabulating CVRs with SEMS file and manual data', async () => {
     expect(getByTestId('total-ballots-entered').textContent).toEqual('100');
   });
   fireEvent.click(getByText('Back to Tally'));
-  expect(getByTestId('total-cvr-count').textContent).toEqual('300');
+  expect(await screen.findByTestId('total-cvr-count')).toHaveTextContent('300');
 
   const fileTable = getByTestId('loaded-file-table');
   const manualRow = domGetByText(
@@ -953,7 +961,7 @@ test('tabulating CVRs with SEMS file and manual data', async () => {
 
   // Now edit the manual data
   fireEvent.click(getByText('Tally'));
-  fireEvent.click(getByText('Edit Manually Entered Results'));
+  fireEvent.click(await screen.findByText('Edit Manually Entered Results'));
 
   // Existing data is still there
   const district5Row = getByText('District 5').closest('tr')!;
@@ -987,7 +995,7 @@ test('tabulating CVRs with SEMS file and manual data', async () => {
   });
 
   fireEvent.click(getByText('Back to Tally'));
-  expect(getByTestId('total-cvr-count').textContent).toEqual('400');
+  expect(await screen.findByTestId('total-cvr-count')).toHaveTextContent('400');
   const fileTable2 = getByTestId('loaded-file-table');
   const manualRow2 = domGetByText(
     fileTable2,
@@ -1024,7 +1032,7 @@ test('tabulating CVRs with SEMS file and manual data', async () => {
 
   // Remove the manual data
   fireEvent.click(getByText('Tally'));
-  fireEvent.click(getByText('Remove Manual Data'));
+  fireEvent.click(await screen.findByText('Remove Manual Data'));
 
   getByText('Do you want to remove the manually entered data?');
   const modal = await screen.findByRole('alertdialog');
@@ -1106,7 +1114,7 @@ test('changing election resets sems, cvr, and manual data files', async () => {
   );
 
   fireEvent.click(await screen.findByText('Tally'));
-  getByText('No CVR files loaded.');
+  await screen.findByText('No CVR files loaded.');
   await screen.findByText('Currently tallying live ballots.');
   // We're waiting on a query for isOfficialResults. It has a default value,
   // so there is no change on the page to wait for before test ends.
@@ -1165,7 +1173,9 @@ test('clearing all files after marking as official clears SEMS, CVR, and manual 
   );
 
   fireEvent.click(getByText('Unofficial Full Election Tally Report'));
-  fireEvent.click(getByText('Mark Tally Results as Official'));
+  const markOfficialButton = getByText('Mark Tally Results as Official');
+  await waitFor(expect(markOfficialButton).toBeEnabled);
+  fireEvent.click(markOfficialButton);
   const modal = await screen.findByRole('alertdialog');
   fireEvent.click(within(modal).getByText('Mark Tally Results as Official'));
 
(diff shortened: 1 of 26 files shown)


frontends/election-manager/src/components/import_cvrfiles_modal.tsx
```diff
@@ -73,14 +75,10 @@ function throwBadStatus(s: never): never {
...
-    usbDriveStatus,
-    castVoteRecordFiles,
-    electionDefinition,
-    auth,
-    logger,
-  } = useContext(AppContext);
+  const { usbDriveStatus, electionDefinition, auth, logger } =
+    useContext(AppContext);
   const addCastVoteRecordFileMutation = useAddCastVoteRecordFileMutation();
+  const cvrFiles = useCvrFilesQuery().data || [];
```
jonahkagan, line 81: Is it safe to use a default here? Or should we wait for data to load?

frontends/election-manager/src/screens/tally_screen.tsx
```diff
@@ -152,7 +154,11 @@ export function TallyScreen(): JSX.Element {
               Load CVR Files
             </Button>{' '}
             <Button
-              disabled={!castVoteRecordFiles.wereAdded || isOfficialResults}
+              disabled={
+                cvrFilesQuery.isFetching ||
```
jonahkagan, line 158: Is there a reason you're using `isFetching` instead of `isLoading`?

frontends/election-manager/src/utils/cast_vote_record_files.ts
```diff
@@ -185,19 +185,20 @@ export class CastVoteRecordFiles {
   /**
    * Parses the filenames for the given CVR {@link files} for information about
    * the CVRs they contain.
+   *
+   * TODO: After a few rounds of refactoring, this is now mostly just a proxy to
+   * {@link parseCvrFileInfoFromFilename} - should consolidate at some point.
```
jonahkagan, line 190: Is the rest of this module still used now that the data is loaded from the backend?

frontends/election-manager/src/app.test.tsx
```diff
@@ -580,7 +580,9 @@ test('tabulating CVRs', async () => {
     expect.anything()
   );
 
-  fireEvent.click(getByText('Mark Tally Results as Official'));
+  const markOfficialButton = getByText('Mark Tally Results as Official');
+  await waitFor(expect(markOfficialButton).toBeEnabled);
```
eventualbuddha, line 584: This took me a while to parse mentally. Also, in general it's not guaranteed to be safe to do this as `toBeEnabled` could be a class prototype method and calling it this way would not properly bind `this`. I gather that isn't the case as it appears to be working, but it still feels weird.

eventualbuddha (commented):
Yay 🥳

frontends/election-manager/src/components/import_cvrfiles_modal.tsx
kofi-q: hm, now that i think about it, probably better to block actions on waiting for data to load, since we decide which import buttons to enable based on previously imported files

frontends/election-manager/src/app.test.tsx
kofi-q: ah, good point! will wrap in a function

frontends/election-manager/src/components/import_cvrfiles_modal.tsx
jonahkagan: Ok, I noticed there are a number of other places that use this pattern. I didn't leave comments on all of them, but same question applies

frontends/election-manager/src/screens/tally_screen.tsx
jonahkagan: I definitely agree we should consider this state.

What do you think about using the approach we discussed earlier? Handle `isLoading` at the top-level of the component for the initial fetch and then make sure that mutations that invalidate the cache wait for the `queryClient.invalidateQueries` promise to resolve.

frontends/election-manager/src/utils/cast_vote_record_files.ts
jonahkagan: okey doke

frontends/election-manager/src/screens/tally_screen.tsx
kofi-q: i worry about a component making too many assumptions about how and when data ends up in certain states (if that makes sense). if we only handle `isLoading`, we miss the case where the query's in `isFetching` state,  and the data after the fetch could be different from what was in the cache before. from our vantage point, we know it's unlikely because of the way the product works, but not handling that state makes the component less resilient against some potential future change down the line.

so, less concerned about this specific case and more so about adopting a consistent pattern that helps us avoid unexpected edge cases, tough-to-diagnose bugs, or even just weird UI flicker.

agreed that we should go through and do a sweeping change to make all our mutations wait on invalidations though, but that's just one of a few ways for this to end up in the `isFetching` state, since just navigating back and forth will do that

jonahkagan: Yeah, I think I see what you're getting at. I agree it seems like something we should figure out a consistent pattern for. Maybe we can discuss sync tomorrow?

It would be helpful to have a list of all the cases we need to consider. So far, here's what I think we've surfaced:
- Initial load when component mounts
- Refetch after invalidation from a mutation
- Navigating to a diff screen and coming back (q: in what ways is this different from initial load?)

Are there others you have in mind? I know there are also the refetchOnWindowFocus and refetchOnReconnect params.

Some relevant docs I've been reading to help think about this (which you may have already seen):
- https://tanstack.com/query/v4/docs/guides/queries#why-two-different-states
- https://tanstack.com/query/v4/docs/guides/background-fetching-indicators

kofi-q: yeah, sync tomorrow sounds good! maybe we can stay on after standup and anyone else interested can linger?
i'll give the different cases some thought ahead of time - thanks!

frontends/election-manager/src/hooks/use_election_manager_store.ts
```diff
@@ -127,7 +109,12 @@ export function useElectionManagerStore(): ElectionManagerStore {
     await queryClient.invalidateQueries(
       getCurrentElectionMetadataResultsQueryKey()
     );
-    await queryClient.invalidateQueries(getCvrFilesQueryKey());
+    await queryClient.invalidateQueries(getCvrFilesQueryKey(), {
+      // TODO(kofi): This is a pattern we're potentially adopting across the
+      // repo and will be make sweeping updates for soon. This specific case is
+      // needed to enable a test to pass after a recent change.
+      refetchType: 'all',
```
kofi-q, line 116: added this to fix a test that removes the "current" election and then navigates to the "Tally" screen to verify that the previously loaded CVR files are gone.

that particular screen tries to map data in the CVRs to data in the current election, which are out of sync when the screen is loaded and that was causing an exception. very unlikely in practice, but it does surface a potential edge case we should handle in general.

frontends/election-manager/src/components/import_cvrfiles_modal.tsx
kofi-q: thanks - audited the other cases and made sure we're handling appropriately

frontends/election-manager/src/index.tsx
```diff
@@ -18,7 +18,10 @@ import { ServicesContext } from './contexts/services_context';
 const queryClient = new QueryClient({
   defaultOptions: {
     mutations: { networkMode: 'always' },
-    queries: { networkMode: 'always' },
+    queries: {
+      networkMode: 'always',
+      staleTime: Infinity,
```
kofi-q, line 23: went ahead and updated the `staleTime` default in this PR, to avoid handling the `isFetching` state

frontends/election-manager/src/components/confirm_removing_file_modal.tsx
```diff
@@ -28,27 +29,31 @@ export function ConfirmRemovingFileModal({
...
   let fileTypeName = '';
   let singleFileRemoval = true;
   switch (fileType) {
     case ResultsFileType.CastVoteRecord: {
-      const { fileList } = castVoteRecordFiles;
+      if (isLoading) {
+        return <Loading />;
+      }
+
+      const fileList = cvrFilesQuery.data || [];
```
jonahkagan, line 48: Do you still need a default value here to get around `data` possibly being undefined? If so, that's probably because checking `query.isLoading` isn't sufficient, we also need to check `query.isError` (or, use `query.isSuccess` to check both). Even if we're going to ignore errors, I think it would be good to do it explicitly.

frontends/election-manager/src/hooks/use_election_manager_store.ts
jonahkagan: Sounds good, glad you surfaced it!

frontends/election-manager/src/index.tsx
jonahkagan: Great! Do you think it's worth leaving a comment explaining the reasoning or should we wait until we decide how to standardize this to document it?

frontends/election-manager/src/components/confirm_removing_file_modal.tsx
jonahkagan: (And noting there are a few other places with similar patterns)

frontends/election-manager/src/index.tsx
kofi-q: yeah, that makes sense - adding a comment

re: standardization: thinking we should have a shared `QueryClient` config that all apps import and use (and extend as needed), or better yet, something along the lines of a shared factory function for creating the query client

jonahkagan: I like both of those ideas!

frontends/election-manager/src/screens/tally_report_screen.tsx
```diff
@@ -255,7 +257,12 @@ export function TallyReportScreen(): JSX.Element {
           {location.pathname === routerPaths.tallyFullReport && (
             <p>
               <Button
-                disabled={!castVoteRecordFiles.wereAdded || isOfficialResults}
+                disabled={
+                  cvrFilesQuery.isLoading ||
+                  !cvrFilesQuery.data ||
```
jonahkagan, line 262: Another place to handle error explicitly

frontends/election-manager/src/screens/reports_screen.tsx
```diff
@@ -183,6 +184,11 @@ export function ReportsScreen(): JSX.Element {
     </p>
   );
 
+  const canSaveResults =
+    !cvrFilesQuery.isLoading &&
+    !!cvrFilesQuery.data &&
```
jonahkagan, line 189: One more place to handle error explicitly. Could just say `cvrFilesQuery.isSuccess && cvrFilesQuery.data.length > 0`

frontends/election-manager/src/screens/tally_screen.tsx
```diff
@@ -63,9 +64,10 @@ export function TallyScreen(): JSX.Element {
       .join(', ');
   }
 
-  const castVoteRecordFileList = castVoteRecordFiles.fileList;
+  const cvrFilesQuery = useCvrFilesQuery();
+  const castVoteRecordFileList = cvrFilesQuery.data ?? [];
```
jonahkagan, line 68: Here also need to handle loading/error (or remove default)

frontends/election-manager/src/screens/write_ins_screen.tsx
```diff
@@ -137,7 +138,7 @@ export function WriteInsScreen(): JSX.Element {
       );
     }
 
-    if (!castVoteRecordFiles.wereAdded) {
+    if (cvrFilesQuery.data?.length === 0) {
```
jonahkagan, line 141: This is a short-hand way to handle loading/error. I've experimented with it a bit and don't mind it, but I think I'm leaning more towards explicit handling (or at least a consistent pattern with other uses of the query)

jonahkagan (commented):
Hey sorry to send you back with more comments but I found a few more places to clean up. Let me know if this is too much nitpicking - my goal is just to make sure we're not missing anything accidentally.

Also, for all of these places where we're ignoring error, would be good to maybe mark them down so we can come back and handle the errors later or something.

frontends/election-manager/src/screens/tally_report_screen.tsx
kofi-q: bah, thanks - my search regex wasn't broad enough to include these! will take a look.

FWIW, i think most of these cases don't need specific user-facing error handling, since the cvrFiles query failing isn't show-stopping for those flows, but will audit and see if any should be handled differently

kofi-q: (also that query failing probably means something that affects the entire app is going on)

jonahkagan (approved):
What a PR! Thanks for sticking with it and hashing all that out

frontends/election-manager/src/screens/tally_report_screen.tsx
jonahkagan: Ok sounds good

kofi-q:
> What a PR! Thanks for sticking with it and hashing all that out

Haha for sure! Thanks for pushing - I'd rather we sort out the kinks up front, so others don't have to later on

PR #2816, opened 2022-11-29
Setup separate vx-scan app

jonahkagan:
## Overview
Task: https://github.com/votingworks/vxsuite/issues/2813

_Review by commits_

- Creates a new apps/vx-scan directory
- Moves frontends/precinct-scanner to apps/vx-scan/frontend
- Copies services/scan to apps/vx-scan/backend
- Updates a bunch of scripts and tooling

Follow up PRs will:
- Delete unnecessary precinct scanner code from services/scan
- Delete unnecessary central scanner code from apps/vx-scan/backend
- Rename logs and such with more appropriate names

## Demo Video or Screenshot
N/A

## Testing Plan 
- Ran commands in the package.json for both the frontend and backend
- Poked around and tried to run every relevant script I could find in the repo (but I certainly may have missed something)

## Checklist


apps/vx-scan/backend/package.json
```diff
@@ -15,9 +15,7 @@
     "lint": "pnpm type-check && eslint .",
     "lint:fix": "pnpm type-check && eslint . --fix",
     "pre-commit": "lint-staged",
-    "start": "node ./build/index.js",
-    "start:central": "VX_MACHINE_TYPE=bsd node ./build/index.js",
-    "start:precinct": "VX_MACHINE_TYPE=precinct-scanner node ./build/index.js",
+    "start": "VX_MACHINE_TYPE=precinct-scanner node ./build/index.js",
```
carolinemodic, line 18: we should remove the need for this env var when we remove the central scan code from this subdir

carolinemodic (approved):
generally lgtm but I defer to @eventualbuddha on some of the build config stuff. 

Note that this will break vxsuite-complete-system as it is assuming that precinct-scanner lives in frontends/precinct-scanner so you will also want to make a PR to the build/run scripts in complete-system to make sure that those continue to work so that VxDev can continue to work, this will be important for Julia probably to use VxScan next week so we shouldn't wait to long to make sure that is working.

.circleci/config.yml
```diff
@@ -779,11 +801,12 @@ workflows:
   test:
     jobs:
       - test-codemods
+      - test-apps-vx-scan-frontend
+      - test-apps-vx-scan-backend
```
eventualbuddha, line 805: Alphabetize?

eventualbuddha (approved):
Feels like we may want to make `pnpm start` in `apps/vx-scan` work as it's unclear why that's the `frontend`'s job.

jonahkagan:
> Feels like we may want to make pnpm start in apps/vx-scan work as it's unclear why that's the frontend's job.

@eventualbuddha interesting idea. I wasn't quite sure how to structure the contents of `apps/vx-scan` so I just went with the thing that mapped easily to our current structure. I'm open to other ideas if you have some (though might defer to a follow up, depending on the complexity)

eventualbuddha:
> > Feels like we may want to make pnpm start in apps/vx-scan work as it's unclear why that's the frontend's job.
> 
> @eventualbuddha interesting idea. I wasn't quite sure how to structure the contents of `apps/vx-scan` so I just went with the thing that mapped easily to our current structure. I'm open to other ideas if you have some (though might defer to a follow up, depending on the complexity)

Yeah, do it in a follow-up. I think the simplest thing is to just put the `start` script in a bare-bones `package.json` at that level. Either that or have some other means of starting it that doesn't directly depend on the (P)NPM infrastructure (e.g. `make`, a `bash` script, etc).

.circleci/config.yml
jonahkagan: a025e6d3b

PR #2817, opened 2022-11-29
feat(bmd): add environment variable for screen orientation

beausmith:
## Overview
Adding a new env variable to explicitly specify which orientation the BMD screen is expected to be in.


beausmith:
I don't understand why tests are failing. This PR seems to have nothing to do with the tests that are failing.

eventualbuddha:
> I don't understand why tests are failing. This PR seems to have nothing to do with the tests that are failing.

Try rebasing? Looks like you're going to need to anyway to incorporate #2813.

frontends/bmd/.env
```diff
@@ -1 +1,2 @@
 SKIP_PREFLIGHT_CHECK=true
+REACT_APP_VX_SCREEN_ORIENTATION=portrait
```
carolinemodic, line 2: add this to the root level .env file not here

carolinemodic: I'm deleting this file in a PR I just merged, but putting it in the root-level .env will make it easier to control this flag in VxDev.

carolinemodic (approved):
You will need to rebase this. 

I had imagined this as a new boolean feature flags in libs/utils/environment_variable but this works too and it does kind of make sense as part of the machine config as it is named so... works for now I'm not really sure yet long term how we should manage the machine config variables.

adghayes:
The integration tests are failing because, in the `beforeEach` section of each test, it tries to start a voting session. However, when it goes to the poll worker screen, the options to select a ballot style are not displayed because you can only select a ballot style in `MarkAndPrint` mode, but the tests are running in `MarkOnly`.

I'm still poking around, but I currently don't understand how these tests are passing on `main` because the same problem should apply. I'm not sure

adghayes:
@beausmith I just pushed a commit to fix.

The integration tests were failing because the tests were running with app mode `MarkOnly` rather than `MarkAndPrint`, which is necessary for our `beforeEach` blocks to work setting up the voting session with a poll worker card: https://github.com/votingworks/vxsuite/blob/b671e57f4e5934256fe20e706ada886192f7ab3b/frontends/bmd/cypress/support/e2e.js#L92-L95

In integration tests, the `machineConfig` comes from the values provided in `setupProxy.js`: https://github.com/votingworks/vxsuite/blob/b671e57f4e5934256fe20e706ada886192f7ab3b/frontends/bmd/prodserver/setupProxy.js#L17-L27 

You didn't update those to include the `screenOrientation` (which would be hard to remember, because it's not type-checked because it sits outside of `src`) so `machine_config.ts` was failing to parse the received `machineConfig`. So it defaulted to the defaults set in the app, which are actually `MarkOnly`

Tests are passing, but we should make two changes eventually:
- Change how we do machine config. Doesn't make sense to have this whole API setup when we could instead do it synchronously and share types across libraries and access the machine config from any service or library. I will share some thoughts on this in Slack and make a ticket.
- After the request failed in this test, it defaulted to `MarkOnly` since that is the default in other parts of `bmd`. Those should probably just default to `MarkAndPrint`, although that wouldn't have addressed root cause here.

PR #2818, opened 2022-11-30
fix broken full test deck tally report

### LINKED ISSUE #2772: Full Test Deck Tally Report has no contests on `main`
Screen Shot 2022-11-08 at 6 53 27 PM

adghayes:
## Overview
Closes #2772 by fixing full test deck tallies for primaries. The problem was that the `ElectionManagerTallyReport` component was expecting / looking for pre-computed party sub-tallies, and we weren't providing them for the test deck tally. Added testing for regressions.

If you look at the PDF, there may be some CSS shading issue with the ballot count breakdown table header. Looking into that.

## Demo Video or Screenshot
[fixed_full_test_deck_report.pdf](https://github.com/votingworks/vxsuite/files/10126036/fixed_full_test_deck_report.pdf)

### DIFF
--- a/frontends/election-manager/src/components/full_test_deck_tally_report_button.test.tsx
+++ b/frontends/election-manager/src/components/full_test_deck_tally_report_button.test.tsx
@@ -1,0 +1,60 @@
+import { screen, within } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import {
+  electionFamousNames2021Fixtures,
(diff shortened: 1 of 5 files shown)


adghayes:
Thanks @arsalansufi!

beausmith:
@adghayes - the shading issues are because of an option that is not enabled on your local machine. Look for a checkbox to "print background images" or something like that; it is necessary to print any backgrounds.

Issue #1622, opened 2022-03-22
Convert jest snapshot tests to use jest-styled-components


jonahkagan:
Snapshot tests for components using styled-components are really annoying because the generated class names change whenever you change styles. The jest-styled-components library fixes this problem.

https://github.com/styled-components/jest-styled-components

A few ways we could approach this:
- Add an import for jest-styled-components in all of our tests that have snapshots and then update all the snapshots
- Figure out a way to auto-import it using the jest config
- Convert tests that rely on snapshots to use more specific test assertions instead and use `toHaveStyleRule` when they want to actually test for specific styles

Example PR: https://github.com/votingworks/vxsuite/pull/1620

eventualbuddha:
How mechanistic is the change? Could it be automated?

jonahkagan:
Just tried to set it up in bmd, and it didn't work due to outdated jest and styled-components versions. I tried upgrading them but it was non-trivial since we'd also need to upgrade node.

So @eventualbuddha I guess in theory it's automatable, but might take some pre-work.

taravancil:
Just noting that this is probably unblocked by #1735

carolinemodic:
This may solve https://github.com/votingworks/vxsuite/issues/1617

adghayes:
I did a little bit of digging. Good news: `jest-styled-components` can be easily set up in a package by simply adding the import the the `setupTests.ts` file. I haven't tried it in all the frontends but it works in `libs/ui` and `libs/election-manager`. On the individual package level, it does as it says it will - converts ambiguous, shifting class names to serialized class names with inspectable style rules. It would make our `libs/ui` snapshots much more useful, and I think we should at least implement it there ASAP (very low lift). 

Bad news: this is just a testing utility, so there's no impact on exports from a package. That means that there are still the `styled-components` generated class names on exports from `libs/ui`. So if you change a style on on a `libs/ui` component, it will change the class name in the snapshots in all the `frontends`, rather than serialize and inject style rules. Perhaps that's desired behavior, in that it will let us know where our `libs/ui` style changes are impacting the frontends?

That means that this also has no impact on [#1617](https://github.com/votingworks/vxsuite/issues/1617). However, I think that [#1617](https://github.com/votingworks/vxsuite/issues/1617) has mostly to do with `styled-components` generating classNames based partially on the _order of the components_ in the index file. Still a problem, but in most cases can be avoided by adding new components to the end of the index file.

PR #2819, opened 2022-11-30
libs/ui testing improvements

### LINKED ISSUE #872: [ui] fix test coverage
https://github.com/votingworks/vxsuite/pull/871 began enforcing 100% test coverage again, but had to disable checking for a few recently-added files. We should address that by adding test coverage for these files and removing them from the exclusion list.

adghayes:
## Overview
Closes #872 by getting to 100% test coverage and tests the water of #1622 by using `jest-styled-components` in `libs/ui`. In order to complete #1622, we'll have to apply `jest-styled-components` in the frontends which is temporarily blocked on #2816. Applying to the frontends should entail minor changes to the snapshots and be quick.

I do think these new snapshots are far more useful.

I took at pass at all our existing testing in `libs/ui` and made some tweaks:
- Adding some assertions where they make more sense than snapshots, or in addition to snapshots
- Convert inline snapshots to regular snapshots, to standardize


libs/ui/src/election_info_bar.test.tsx
```diff
@@ -16,30 +17,51 @@ test('Renders ElectionInfoBar without precinct information by default', () => {
       mode="admin"
     />
   );
+  screen.getByText('General Election');
+  screen.getByText('Nov 3, 2020');
+  screen.getByText('Franklin County,');
+  screen.getByText('State of Hamilton');
+  screen.getByText(hasTextAcrossElements('Software VersionDEV'));
+  screen.getByText(hasTextAcrossElements('Machine ID0000'));
```
beausmith, line 25: Maybe we should add a space in here to increase developer readability?

libs/ui/src/loading.test.tsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 
 import { Loading } from './loading';
 
 test('Renders Loading with defaults', () => {
   const { container } = render(<Loading />);
+  screen.getByText('Loading');
   expect(container.firstChild).toMatchSnapshot();
```
beausmith, line 9: Should we remove this snapshot?

libs/ui/src/modal.test.tsx
```diff
@@ -33,22 +33,7 @@ describe('Modal', () => {
...
-        tabindex="-1"
-      >
-        <div
-          class="sc-hKwCoD bNJKkA"
-        >
-          Do you want to do the thing?
-        </div>
-      </div>
-    `);
+    expect(modal).toMatchSnapshot();
```
beausmith, line 36: Should we remove this snapshot?

libs/ui/src/modal.test.tsx
```diff
@@ -65,44 +50,14 @@ describe('Modal', () => {
...
-        tabindex="-1"
-      >
-        <div
-          class="sc-hKwCoD gghqgk"
-        >
-          Do you want to do the thing?
-        </div>
-      </div>
-    `);
+    expect(modal).toMatchSnapshot();
```
beausmith, line 53: Should we remove this snapshot?

libs/ui/src/modal.test.tsx
```diff
@@ -65,44 +50,14 @@ describe('Modal', () => {
...
-        tabindex="-1"
-      >
-        <div
-          class="sc-hKwCoD cZPwiQ"
-        >
-          Do you want to do the thing?
-        </div>
-      </div>
-    `);
+    expect(modal).toMatchSnapshot();
```
beausmith, line 60: Should we remove this snapshot?

libs/ui/src/progress_ellipsis.test.tsx
```diff
@@ -5,11 +5,5 @@ import { ProgressEllipsis } from './progress_ellipsis';
...
 test('renders ProgressEllipsis', () => {
   const { container } = render(<ProgressEllipsis />);
-  expect(container).toMatchInlineSnapshot(`
-    <div>
-      <span
-        class="sc-bdvvaa etEPtU"
-      />
-    </div>
-  `);
+  expect(container).toMatchSnapshot();
```
beausmith, line 8: This should be removed & replaced by a more specific test. (not necessarily for this PR)

libs/ui/src/reboot_from_usb_button.test.tsx
```diff
@@ -18,16 +18,7 @@ test('renders without a USB drive as expected.', async () => {
...
-    <div>
-      <button
-        class="sc-gsDJrp eHDFnB"
-        type="button"
-      >
-        Reboot from USB
-      </button>
-    </div>
-  `);
+  expect(container).toMatchSnapshot();
```
beausmith, line 21: This snapshot can be removed.

libs/ui/src/reboot_to_bios_button.test.tsx
```diff
@@ -15,16 +15,7 @@ test('renders as expected.', async () => {
...
-    <div>
-      <button
-        class="sc-gsDJrp eHDFnB"
-        type="button"
-      >
-        Reboot to BIOS
-      </button>
-    </div>
-  `);
+  expect(container).toMatchSnapshot(``);
```
beausmith, line 18: This snapshot can be removed.

libs/ui/src/setup_card_reader_page.test.tsx
```diff
@@ -1,27 +1,29 @@
...
+import { render, screen } from '@testing-library/react';
 
 import { SetupCardReaderPage } from './setup_card_reader_page';
 
 describe('renders SetupCardReaderPage', () => {
   test('with no useEffect trigger as expected', () => {
     const { container } = render(<SetupCardReaderPage />);
+    screen.getByText('Card Reader Not Detected');
+    screen.getByText('Please ask a poll worker to connect card reader.');
+    expect(container.firstChild).toMatchSnapshot();
```
beausmith, line 11: This looks like a duplicate of the next line. Both snapshots can be removed.

libs/ui/src/text.test.tsx
```diff
@@ -114,31 +113,6 @@ describe('renders Text', () => {
...
-            And I’m a paragraph with a
-          </span>
-          <br />
-          <span>
-            line break.
-          </span>
-        </p>
-      </div>
-    `);
+    expect(container).toMatchSnapshot();
```
beausmith, line 116: This should be replaced with test assertions which look for the content to be broken into specific container tags. But doesn't have to be changed for this PR.

beausmith (approved):
I support removing snapshots because most people don't know how to determine if a change to a snapshot is expected or not. Ideally more explicit checks for specific expected text or expected styles are added where appropriate.

libs/ui/src/election_info_bar.test.tsx
adghayes: Replacing it with this, which is a bit more readable: 
```js
  const versionLabel = screen.getByText('Software Version');
  expect(versionLabel.parentElement?.lastChild).toHaveTextContent('DEV');

  const machineIdLabel = screen.getByText('Machine ID');
  expect(machineIdLabel.parentElement?.lastChild).toHaveTextContent('0000');

  const electionIdLabel = screen.getByText('Election ID');
  expect(electionIdLabel.parentElement?.lastChild).toHaveTextContent(/^.{10}$/);
  ```

File: libs/ui/src/loading.test.tsx (hunk 2 above)
Comment (adghayes) [reply libs/ui/src/loading.test.tsx:9]: You are probably right in that most of us don't handle changes to snapshots correctly, but I like the ease of capturing all the styles in a snapshot.

My thought is to capture defaults in a snapshot, and then variations (like fullscreen) with specific assertions. That way a change to a default value (or one affecting all tests) doesn't cause all behavior to fail, just the one snapshot. And we have more explicit checks on the variations.

I don't plan to do that for all the files here, but having up to one snapshot per component seems reasonable to me.

File: libs/ui/src/loading.test.tsx [hunk 11]
```diff
@@ -14,5 +15,6 @@ test('Renders Loading fullscreen with tag and label', () => {
       Printing
     </Loading>
   );
+  screen.getByText('Printing');
   expect(container.firstChild).toMatchSnapshot();
```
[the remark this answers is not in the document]
adghayes: Yes, will write alternate testing.

libs/ui/src/modal.test.tsx
adghayes: yes, this is covered by style assertions.

libs/ui/src/setup_card_reader_page.test.tsx
adghayes: Removing duplicate, but this is a case where I don't really see the harm in having a snapshot. Keeping one for now. If I were to go and edit this component, I think I would find it useful to be able to see styles changing.

libs/ui/src/test_mode.test.tsx
```diff
@@ -5,15 +5,5 @@ import { TestMode } from './test_mode';
...
-      <div
-        class="sc-bdvvaa hoKchV"
-      >
-        <div>
-          Machine is in Testing Mode
-        </div>
-      </div>
-    </div>
-  `);
+  expect(container).toMatchSnapshot();
```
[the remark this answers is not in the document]
adghayes: I think we should have that assertion (added), but here I also don't see the harm in having a snapshot to visualize style changes if we don't have style assertions.

libs/ui/src/text.test.tsx
adghayes: I'll improve it in this PR.

PR #2828, opened 2022-12-02
Add support for `customctl`

eventualbuddha:
## Overview

Allows running VxScan with `PLUSTEKCTL_PATH=/path/to/customctl` to work with the Custom America A4 scanner.

## Demo Video or Screenshot
```sh
# Runs like so:
PLUSTEKCTL_PATH=$HOME/code/plustekctl/customctl pnpm start
```

## Testing Plan 
Tested manually with the `customctl` branch of the `plustekctl` repo with an IdeaPad Flex running vxsuite and the UI running in Chrome on macOS. Used the Ashland 11/8/22 election.

## Checklist

### DIFF
--- a/libs/types/src/result.ts
+++ b/libs/types/src/result.ts
@@ -50,6 +50,21 @@ class Ok<T> {
   }
 
   /**
+   * Returns the contained value.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  unsafeExpect(message: string): T {
+    return this.value;
+  }
+
+  /**
+   * Throws the given error message.
+   */
+  unsafeExpectErr(message: string): never {
+    throw new Error(message);
+  }
+
+  /**
    * Provides a custom inspect function for `Ok` values so that `console.log`
    * and the node REPL will pretty-print the wrapper and value.
    */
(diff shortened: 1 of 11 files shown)


libs/types/src/result.ts
```diff
@@ -49,6 +49,21 @@ class Ok<T> {
...
+  unsafeExpect(message: string): T {
+    return this.value;
+  }
+
+  /**
+   * Throws the given error message.
+   */
+  unsafeExpectErr(message: string): never {
+    throw new Error(message);
+  }
```
jonahkagan, line 65: Why would you want to use these instead of `unsafeUnwrap` and `unsafeUnwrapErr`?

jonahkagan (approved):

libs/types/src/result.ts
eventualbuddha: They act more like `assert` with an error message instead of just throwing a value that is potentially less descriptive. The Rust `Result` has `unwrap` and `expect`, and the convention is that `unwrap` is really only for quick and dirty code while you're working on it. If you would use an `assert` in other circumstances, then `expect` is better. Otherwise you should actually handle the error properly.

jonahkagan: Ok, gotcha. I can see the use of that. Reading the code that uses them, I was pretty confused. Expect makes me think about tests. I think we might want to work on the names a bit.

Maybe `const width = widthResult.assertOk('width is valid');` would mesh more with how I generally think about things in TS.

eventualbuddha: https://github.com/votingworks/vxsuite/pull/2847

Issue #2796, opened 2022-11-18
Improve usage of env feature flagging 


carolinemodic:
We used to control feature flags by setting them up in each individual service or frontend, and having a default .env file in that subdirectory of vxsuite with default values. Then devs could also add a .env.local file to override it and enable flags on or off. This was confusing for two reasons: 
1) Many identical feature flags are used in more then one service/frontend, there is no central place to see all of the options you have with the flags, and its hard to manage setting up a ton of different env files properly. 
2) In VxDev people want to set up flags and have it apply to everything and continue to apply as you update the code (which wipes any changes in vxsuite/) so we copy a .env.local file from outside of vxsuite after updating the code. This file needing to be set in every single subdirectory of vxsuite is annoying.

So in the PRs linked below I did a few things to improve feature flagging, but unfortunately caused some other problems in the process:
- I added the concept of a root level vxsuite .env file. It is loaded and is overridden by the subdirectory .env file so that if you want to make a sub-service/app config change you can. However we still have some default env files at the subrepo level that override the root level file which has made it impossible to set things properly in vxdev if they only copy to the root level. I attempted to change the ordering but could not figure out a way to do what I wanted: https://votingworks.slack.com/archives/CEL6D3GAD/p1664233394439899 . I ultimately never removed those subdirectory .env files either as there is one non-feature flag value in VxAdmin's default .env file to set the convertor to either ms-sems or nh-accuvote. I wasn't sure how this should fit into the env file workings and forgot about it. 
- I moved all of the feature flagging setup in vxsuite to define flags, etc. into a shared location in libs/utils. I also potentially got too fancy and created a script to autogenerate a .env.local after answering command line prompts about what you want to set. My goal was to be able to call into this script from a VxDev script/program so you could configure your flags for VxDev without needing to open a text file in a mysterious location. However I never it got it all working properly and I think the script is already broken and unusable in vxsuite. I made this script and the shared code only handle boolean feature flags but was torn on how to / whether to also handle text environment flags like the convertor example in VxAdmin. 

https://github.com/votingworks/vxsuite/pull/2538
https://github.com/votingworks/vxsuite/pull/2539

PR #2829, opened 2022-12-02
Clean up loose ends with environment files 

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/2796
This PR does a couple of small clean ups to improve how we're handling environment variables and files

- Fixes the pnpm call to run the generation script to create a .env.local file
- Moves the logic for the VxAdmin converter environment variable to the shared utility. Since this is the only string variable we currently have and we want it strongly typed I didn't fully generalize this yet. But by moving it here we are able to include that variable in the script that generates .env.local files.  In the process of doing this I renamed some of the boolean variable functions/types, feel free to suggest better names. 
- Updates the default .env file at the root level to be all of the default values when you run the env file generation script (in the future we should figure out how to automate this being updated or have a test to enforce its updated cc @amcmanus ) 
- Removes the now completely redundant and unnecessary subrepo default .env files. 

## Testing Plan 
Verified that I can still set the convertor with .env or .env.local files properly. Ran the generation script with valid and invalid inputs. 

## Checklist


.circleci/config.yml
```diff
@@ -116,6 +116,7 @@ jobs:
       - run:
           name: Build
           command: |
+            cp .env frontends/election-manager/.env
```
carolinemodic, line 119: I feel like there must be a better solution to this problem but for some reason circleCI doesn't seem to be looking at the root-level .env file, which only matters for getting the ms-sems converter value in VxAdmin, this fixed it but I don't know why I need to do this.

.env
```diff
@@ -2,3 +2,7 @@ REACT_APP_VX_ENABLE_WRITE_IN_ADJUDICATION=TRUE
 REACT_APP_VX_ENABLE_ALL_ZERO_SMARTCARD_PIN_GENERATION=TRUE
 REACT_APP_VX_DISABLE_CARD_READER_CHECK=FALSE
 REACT_APP_VX_ENABLE_LIVECHECK=TRUE
+REACT_APP_VX_DISALLOW_CASTING_OVERVOTES=FALSE
```
umbernhard, line 5: this is a huge nitpick but would it be possible to make this `ALLOW_CASTING_OVERVOTES` so it wouldn't be a double negative?

.circleci/config.yml
umbernhard: FWIW it looks like this file did exist but is deleted by this PR.

umbernhard (approved):
Looks good to me, a few nitpicks here and there.

.circleci/config.yml
carolinemodic: yeah but it shouldn't be necessary

.env
carolinemodic: out of scope of this pr

PR #2848, opened 2022-12-07
feat(image-utils): dithering

eventualbuddha:
## Overview

Adds monochrome dithering for use with monochrome displays or printers.

https://shihn.ca/posts/2020/dithering/

## Demo Video or Screenshot
| Original | Dithered |
|-|-|
| mona-lisa | mona-lisa-dithered |
| ca-seal | ca-seal-dithered |

## Testing Plan 
Automated, plus the manual ones shown above.

## Checklist


libs/image-utils/src/image_data.ts
```diff
@@ -355,3 +355,39 @@ export function toRgba(
...
+
+  for (let offset = 0, size = input32.length; offset < size; offset += 1) {
+    const px = input32[offset] as number;
+    const r = px & 0xff;
+    const g = (px >>> 8) & 0xff;
+    const b = (px >>> 16) & 0xff;
+    const a = (px >>> 24) & 0xff;
+
+    // Luminosity grayscale formula.
+    const luminosity = (0.21 * r + 0.72 * g + 0.07 * b) | 0;
```
adghayes, line 388: I agree we should not use gamma correction for the performance hit, although I am curious if it would make it better. I stubbed it out but haven't tried it yet:
https://github.com/votingworks/vxsuite-print-scan/blob/cf803658f913c63cf925d2633754f09ba7e6bfe8/services/print-scan/src/printer_scanner/printing.ts#L18-L56

libs/image-utils/src/dither.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return dither(toGrayscaleResult.ok());
+  }
+
+  const channelCount = 1;
+  const { data, width, height } = input;
+  const src = Float32Array.from(data);
+  const dst = new Uint8ClampedArray(width * height * channelCount);
+  const output = createImageData(dst, width, height);
+  const threshold = 128;
```
adghayes, line 35: Have you played with other thresholds? I have found that, when just doing template mapping, 128 is not the right threshold and it should be higher.

adghayes (approved):
I am curious how this will do in terms of performance on certain hardware.

adghayes (changes requested):
When I tried this with a ballot image from the command line, I got these results:

Source:
ballot

Destination:
ballot-dither

Do you know what might be happening here?

libs/image-utils/src/dither.ts
eventualbuddha: I added `threshold` as an option in 21b890b and made the default use `otsu`. The results don't change that much, though:

| 50 | 100 | 200 | otsu (110) |
|-|-|-|-|
|mona-lisa-dithered50|mona-lisa-dithered100|mona-lisa-dithered200|mona-lisa-dithered|

adghayes (approved):

Issue #2814, opened 2022-11-29
Convert VxScan backend to use Grout


jonahkagan:

jonahkagan:
Status update: all of the easy API methods have been converted as of #2854. There are still two endpoints that need more extensive changes:
1. Exporting CVRs - We should export CVRs to USB on the backend and have a separate endpoint for summary data needed for building tally reports on the frontend.
2. Loading the ballot package templates - We should do this all on the backend.

jonahkagan:
Also need to clean up the VxScan types to be in their own module (currently they are in a shared module with VxCentralScan)

PR #2851, opened 2022-12-07
Introducing Grout

jonahkagan:
## Overview
Task: #2814 

Adds the beginning of a new library, Grout, that creates type-safe RPC glue code. I recommend reading the [README](https://github.com/votingworks/vxsuite/pull/2851/files#diff-257d9336911e0f18657e5e688a7fdcadc7082d5c34d037815f2506dca3cc2e43) first to see if that makes sense before diving into the code.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Automated tests

## Checklist
- [ n/a] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ n/a] I have added a screenshot and/or video to this PR to demo the change
- [ n/a] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

### DIFF
--- a/libs/grout/README.md
+++ b/libs/grout/README.md
@@ -54,23 +54,12 @@ const api = grout.createApi({
 
 const app = express();
 
-// When registering the API with an Express app, you must use a separate
-// Router. That way you can isolate the Grout API routes from other routes in
-// your app.
-const router = express.Router();
-
-// You must have the router use the `express.text()` middleware with type
-// `application/json` exactly as shown below.
-router.use(express.text({ type: 'application/json' }));
-
-// Then, simply register the API routes.
-grout.registerRoutes(router, api);
-
-// Finally, mount the router on your app. You can use any path prefix you want -
-// here we use `/api`. It's important that you don't use any other body-parsing
-// middleware upstream of the router - e.g. don't call `app.use(express.json())`
-// before this.
-app.use('/api', router);
+// When registering the API with an Express app, Grout creates an Express Router
+// to isolate the Grout API routes from other routes in your app. You can mount
+// this router using any path prefix you want - here we use `/api`. It's
+// important that you don't use any other body-parsing middleware upstream of
+// the router - e.g. don't call `app.use(express.json())` before this.
+app.use('/api', buildRouter(api, express));
 
 // Don't forget to export the API type - we'll need it for the client.
 export type MyApi = typeof api;
--- a/libs/grout/README.md
+++ b/libs/grout/README.md
@@ -102,7 +102,9 @@ await apiClient.getAllPeeple(); // => TS error: Property getAllPeeple does not e
 await apiClient.getPersonByName({ nam: 'Bob' }); // => TS error: Argument of type '{ nam: string; }' is not assignable to parameter of type '{ name: string; }'.
 await apiClient.updatePersonAge({ name: 'Bob', age: '1' }); // => TS error: Argument of type '{ name: 'Bob'; age: '1'; }' is not assignable to parameter of type '{ name: string; age: number; }'.
 
-// Since we used the Result type, known error cases must be handled explicitly.
+// Since we used the Result type to handle a known error case, we should
+// explicitly check for errors. This is a strongly recommended pattern to help
+// us avoid forgetting to handle known errors.
 const updateResult = await apiClient.updatePersonAge({ name: 'Bob', age: -1 });
 if (updateResult.isErr()) {
   console.error(updateResult.error.message); // => 'Age must be at least 1.'
@@ -116,6 +118,22 @@ try {
 } catch (error) {
   if (error instanceof grout.ServerError) {
     console.error(error.message);
+  }
+}
+```
+
+### tsconfig settings
+
+Grout works out of the box with our default `tsconfig.json` settings. However,
+you can enable VS Code's "Go to Definition" feature by turning on
+`compilerOptions.declarationMap` in your server-side `tsconfig.build.json`. This
+will allow you to Cmd+Click on a method name in the client code and jump to the
+server-side implementation.
+
+```json
+{
+  "compilerOptions": {
+    "declarationMap": true
   }
 }
 ```
--- a/libs/grout/README.md
+++ b/libs/grout/README.md
@@ -87,7 +87,7 @@ import type { MyApi } from './server';
 
 // We need to set the base URL for the client. Make sure to include the path
 // prefix (in our case, /api).
-const baseUrl = 'http://localhost:1234/api';
+const baseUrl = '/api';
 
 // Create the client using MyApi as a type parameter.
 const apiClient = grout.createClient<MyApi>({ baseUrl });
--- a/libs/grout/README.md
+++ b/libs/grout/README.md
@@ -44,8 +44,8 @@ const api = grout.createApi({
     name: string;
     age: number;
   }): Promise<Result<void, Error>> {
-    if (age <= 0) {
-      return err(new Error('Age must be at least 1.'));
+    if (age < 0) {
+      return err(new Error('Age must be at least 0.'));
     }
     await store.updatePersonAge(input.name, input.age);
     return ok();
@@ -107,7 +107,7 @@ await apiClient.updatePersonAge({ name: 'Bob', age: '1' }); // => TS error: Argu
 // us avoid forgetting to handle known errors.
 const updateResult = await apiClient.updatePersonAge({ name: 'Bob', age: -1 });
 if (updateResult.isErr()) {
-  console.error(updateResult.error.message); // => 'Age must be at least 1.'
+  console.error(updateResult.error.message); // => 'Age must be at least 0.'
 }
 
 // If there's an unexpected server error (e.g. a crash or runtime exception),
--- a/libs/grout/README.md
+++ b/libs/grout/README.md
@@ -1,0 +1,160 @@
+# Grout
+
+A library to create type-safe RPC glue code.
+
+Grout lets you define a server API as a plain TS object with async methods.
+Based on this API definition, Grout will create Express route handlers on the
+server side and a type-safe client on the client side, abstracting away the
+details of the underlying HTTP requests.
+
+## Installation
+
+Grout must be installed as a dependency in both your client and server projects.
+
+## Usage
+
+### Server
+
+```ts
+import * as grout from '@votingworks/grout';
+
+interface Person {
+  name: string;
+  age: number;
+}
+
+// Create an API definition by building an object with async methods.
+const api = grout.createApi({
+  // Each method should return a Promise. The data returned should be simple TS
+  // values - not fancy class instances. Values that can be easily serialized to
+  // JSON. More details on this below.
+  async getAllPeople(): Promise<Person[]> {
+    return store.getAllPeople(); // Assume this accesses the database, etc.
+  },
+
+  // Methods can take input, but it must be packaged into a single object.
+  // Think of it like using named parameters, or React component props.
+  async getPersonByName(input: { name: string }): Promise<Person | undefined> {
+    return store.getPersonByName(input.name);
+  },
+
+  // For known errors (e.g. invalid input), use a Result type so that the client
+  // forced to handle them explicitly.
+  async updatePersonAge(input: {
+    name: string;
+    age: number;
+  }): Promise<Result<void, Error>> {
+    if (age <= 0) {
+      return err(new Error('Age must be at least 1.'));
+    }
+    await store.updatePersonAge(input.name, input.age);
+    return ok();
+  },
+});
+
+const app = express();
+
+// When registering the API with an Express app, you must use a separate
+// Router. That way you can isolate the Grout API routes from other routes in
+// your app.
+const router = express.Router();
+
+// You must have the router use the `express.text()` middleware with type
+// `application/json` exactly as shown below.
+router.use(express.text({ type: 'application/json' }));
+
+// Then, simply register the API routes.
+grout.registerRoutes(router, api);
+
+// Finally, mount the router on your app. You can use any path prefix you want -
+// here we use `/api`. It's important that you don't use any other body-parsing
+// middleware upstream of the router - e.g. don't call `app.use(express.json())`
+// before this.
+app.use('/api', router);
+
+// Don't forget to export the API type - we'll need it for the client.
+export type MyApi = typeof api;
+```
+
+### Client
+
+```ts
+import * as grout from '@votingworks/grout';
+
+// First, we need to import the API type from the server. This will *not* import
+// any of the actual server code - just the type definition.
+import type { MyApi } from './server';
+
+// We need to set the base URL for the client. Make sure to include the path
+// prefix (in our case, /api).
+const baseUrl = 'http://localhost:1234/api';
+
+// Create the client using MyApi as a type parameter.
+const apiClient = grout.createClient<MyApi>({ baseUrl });
+
+// Now we can call the API methods we defined as normal functions.
+await apiClient.getAllPeople(); // => [{ name: 'Alice', age: 99 }, ...]
+await apiClient.getPersonByName({ name: 'Bob' }); // => { name: 'Bob', age: 42 }
+await apiClient.updatePersonAge({ name: 'Bob', age: 43 }); // => ok()
+
+// If we make a mistake, we'll get a type error, just like with normal typed functions
+await apiClient.getAllPeeple(); // => TS error: Property getAllPeeple does not exist
+await apiClient.getPersonByName({ nam: 'Bob' }); // => TS error: Argument of type '{ nam: string; }' is not assignable to parameter of type '{ name: string; }'.
+await apiClient.updatePersonAge({ name: 'Bob', age: '1' }); // => TS error: Argument of type '{ name: 'Bob'; age: '1'; }' is not assignable to parameter of type '{ name: string; age: number; }'.
+
+// Since we used the Result type, known error cases must be handled explicitly.
+const updateResult = await apiClient.updatePersonAge({ name: 'Bob', age: -1 });
+if (updateResult.isErr()) {
+  console.error(updateResult.error.message); // => 'Age must be at least 1.'
+}
+
+// If there's an unexpected server error (e.g. a crash or runtime exception),
+// the Promise returned by the client method will be rejected, so we still need
+// to handle exceptions.
+try {
+  await apiClient.getAllPeople();
+} catch (error) {
+  if (error instanceof grout.ServerError) {
+    console.error(error.message);
+  }
+}
+```
+
+## HTTP Transport Details
+
+Grout uses HTTP POST requests for all RPC method calls.
+
+Input is sent using a JSON request body and output is returned as a JSON
+response, both of which are serialized using Grout's serialization format. This
+allows requests to be human-readable while still supporting richer types than
+plain JSON.
+
+Grout supports the following data types:
+
+- `undefined`
+- `null`
+- `boolean`
+- `number`
+- `string`
+- `Array`
+- `Object` (plain objects)
+- `Error`
+- `Result` (from @votingworks/types)
+
+Data can compose and nest as much as you like using the above types.
+
+## Development
+
+See package.json for a full list of development commands. Here are a few common
+ones:
+
+```sh
+# Install dependencies
+pnpm install
+
+# Run tests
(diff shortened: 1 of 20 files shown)


libs/grout/src/util.ts
```diff
@@ -0,0 +1,36 @@
...
+  return typeof value === 'string';
+}
+
+export function isArray(value: unknown): value is unknown[] {
+  return Array.isArray(value);
+}
+
+export function isObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null;
+}
```
arsalansufi, line 21: Worth noting that this function will return true when passed an array, which might be misleading given the typeguard

libs/grout/README.md
```diff
@@ -0,0 +1,160 @@
...
+  },
+
+  // Methods can take input, but it must be packaged into a single object.
+  // Think of it like using named parameters, or React component props.
+  async getPersonByName(input: { name: string }): Promise<Person | undefined> {
+    return store.getPersonByName(input.name);
+  },
+
+  // For known errors (e.g. invalid input), use a Result type so that the client
+  // forced to handle them explicitly.
```
arsalansufi, line 42: For my understanding, what exactly do you mean when you say "forced to handle them explicitly"? Do you mean that TS will warn you if you don't (I don't think that's the case)? Or that you can't rely on some global error handler like the react-query error handler?

libs/grout/README.md
```diff
@@ -0,0 +1,160 @@
...
+// If there's an unexpected server error (e.g. a crash or runtime exception),
+// the Promise returned by the client method will be rejected, so we still need
+// to handle exceptions.
+try {
+  await apiClient.getAllPeople();
+} catch (error) {
+  if (error instanceof grout.ServerError) {
+    console.error(error.message);
+  }
+}
```
arsalansufi, line 120: I take it that, in practice, we'll be relying on a global error handler like the react-query error handler for these errors instead of wrapping every `apiClient` call in a try-catch - is that thinking correct?

libs/grout/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+### Client
+
+```ts
+import * as grout from '@votingworks/grout';
+
+// First, we need to import the API type from the server. This will *not* import
+// any of the actual server code - just the type definition.
+import type { MyApi } from './server';
```
arsalansufi, line 86: Thanks for all the sample code and comments in this README! Very helpful format!

libs/grout/src/client.ts
```diff
@@ -0,0 +1,92 @@
...
+      return async (input?: unknown) => {
+        const inputJson = serialize(input);
+
+        debug(`Call: ${methodName}(${inputJson})`);
+
+        const response = await fetch(methodUrl(methodName, options.baseUrl), {
+          method: 'POST',
+          body: serialize(input),
+          headers: { 'Content-type': 'application/json' },
+        });
```
arsalansufi, line 69: I see that some failures in this function will be thrown as ServerErrors while others will be thrown as plain Errors, e.g. if this `await fetch` rejects, it'll be thrown as a plain Error. Do we want this function to consistently throw ServerErrors? Or is there a meaningful distinction between the cases that result in ServerErrors vs. plain Errors?

arsalansufi (commented):
This looks great! Still reviewing but thought I'd send these first few comments over

libs/grout/README.md
jonahkagan: Since `Result` is a wrapper and is always either `ok(someVal)` or `err(someErr)`, you always have to check which one it is and unwrap to get the result. This is basically a strongly recommended pattern that @eventualbuddha came up with so that we can't forget to handle known errors (rather than use exceptions/a global error handler).

From a UX perspective, the idea is to surface known errors to users near the point of action that induced the error and reserve a global error handler for unexpected errors.

libs/grout/README.md
jonahkagan: Yes, I'm planning to look into setting up a React error boundary for unexpected errors.

libs/grout/src/client.ts
jonahkagan: Good catch. I just missed that one because it was implicit. I think we should catch and wrap that in a `ServerError`

libs/grout/README.md
arsalansufi: :100: Thank you for the explanation!

libs/grout/src/util.ts
jonahkagan: ea098eef0

libs/grout/src/client.ts
jonahkagan: b73c70421

libs/grout/src/server.ts
```diff
@@ -0,0 +1,176 @@
...
+ * ensures that consumers of the method will be forced to handle that error case
+ * explicitly.
+ */
+export function registerRoutes(api: AnyApi, router: express.Router): void {
+  if (!router.stack.some((layer) => layer.name === 'textParser')) {
+    throw new GroutError(
+      'Grout requires the router to use the text body parser middleware with content type application/json.' +
+        " Add it like so: app.use(express.text({ type: 'application/json' }))."
+    );
+  }
```
arsalansufi, line 130: Cool that you found a way to validate this!

libs/grout/src/server.ts
```diff
@@ -0,0 +1,176 @@
...
+    // more complexity for little practical gain.
+    router.post(path, async (request, response) => {
+      try {
+        debug(`Call: ${methodName}(${request.body})`);
+        if (!isString(request.body)) {
+          throw new GroutError(
+            'Request body was parsed as something other than a string.' +
+              ' Make sure you havent added any other body parsers upstream' +
+              ' of the Grout router - e.g. app.use(express.json()).' +
+              ` Body: ${JSON.stringify(request.body)}`
```
arsalansufi, line 150: Will this constraint make it hard to migrate apps to Grout gradually?

libs/grout/README.md
jonahkagan: Thinking through this more though, it's actually still very possible to forget to handle the `Err` case with a mutation that returns `Result<void, Error>`, since we don't need to check the `Ok` case to get an output value. @eventualbuddha do you have any thoughts about how we might prevent this kind of mistake?

libs/grout/src/server.ts
```diff
@@ -0,0 +1,176 @@
...
+          );
+        }
+
+        const input = deserialize(request.body);
```
arsalansufi, line 154: Could we accomplish all our validation / transformation needs while still using a JSON body parser and `res.json`? The current approach makes sense to me! Just sanity checking that we really have to use a text body parser and text responses

libs/grout/README.md
eventualbuddha: We actually have an ESLint rule for this but it's broken! I'm fixing it now. It'll look like this when it's working:

image

libs/grout/src/server.ts
jonahkagan: In my initial migration of VxScan, I'm just moving all the other routes to a separate router as a first step so there's no body-parsing middleware on the top-level app. It's a pretty easy change to make.

libs/grout/src/server.ts
jonahkagan: Yes, we could, but it would require an extra pass over the parsed JSON data, which could slow things down (though not sure it's a necessary optimization at this stage - I was trying to be a bit more performance minded than usual since this will be used in a lot of different places). I also just thought it would be nice to use the built in replacing functionality of JSON.stringify.

Do you think there's an advantage to using Express's JSON serialization?

libs/grout/src/client.ts
```diff
@@ -0,0 +1,92 @@
...
+        debug(`Response status code: ${response.status}`);
+
+        if (
+          !response.headers.get('Content-type')?.includes('application/json')
+        ) {
+          throw new ServerError('Response is not JSON');
+        }
+
+        if (!response.ok) {
+          const message = (await response.json())?.message;
```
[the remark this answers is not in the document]
jonahkagan: Ooh good point. Will handle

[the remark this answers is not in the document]
jonahkagan: Though we did already check that content-type was application/json above, so is there really a realistic case where this wouldn't be JSON?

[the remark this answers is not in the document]
jonahkagan: I figured just wrapping all of this in a try-catch couldn't hurt. What do you think of this? f9995296d

libs/grout/README.md
jonahkagan: Awesome!

libs/grout/src/serialization.ts
```diff
@@ -0,0 +1,164 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
eventualbuddha, line 1: This duplicates a lot of code that is used for the worker pool data transfer. Worth unifying?

libs/grout/README.md
eventualbuddha: How do you envision these imports actually working? Would the type just live in `api` and be imported by both the server and client?

libs/grout/src/client.ts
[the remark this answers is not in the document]
arsalansufi: > Though we did already check that content-type was application/json above, so is there really a realistic case where this wouldn't be JSON?

Oh right! Forgot about that check - in that case, it shouldn't be possible as far as I can tell

But wrapping all the fetch logic still makes sense to me

libs/grout/README.md
jonahkagan: Here's an example that I think I got working in VxScan:
- Exporting the type from the backend: https://github.com/votingworks/vxsuite/pull/2854/files#diff-041bc090c8c2462dccf83998ef49612c01e865111f24d78b8cda709aabe6a6abR60
- Importing the type on the frontend: https://github.com/votingworks/vxsuite/pull/2854/files#diff-c10dee518b69cf839da28bbee8f6e170d9717f29ea1898cf5ec2c4388f3b7463R2

libs/grout/src/serialization.ts
jonahkagan: Ooooh didn't see that. Yeah it's quite similar! I'm inclined to wait and see what happens with the worker pool code before investing in it, since it might change due to the perf work you're starting. What do you think?

I was thinking about how Grout might be used in a server->worker context as I was building it, and I don't think there's anything that would prevent it from working. Though also not sure how much benefit it would bring.

libs/grout/src/util.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function isObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && !isArray(value) && value !== null;
+}
+
+export function isPlainObject(
+  value: unknown
+): value is Record<string, unknown> {
+  return isObject(value) && value.constructor === Object;
```
eventualbuddha, line 26: A safer way to do this:
```suggestion
  return isObject(value) && Object.prototype.toString.call(value) === '[object Object]';
```
If we somehow get an object from another sandbox then its `Object` and our `Object` won't match up.

libs/grout/src/util.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function isFunction(value: unknown): value is Function {
```
eventualbuddha, line 30: Could use `(...args: unknown[]) => unknown` instead.

eventualbuddha (approved):

libs/grout/src/serialization.ts
eventualbuddha: Fine to leave it for now.

libs/grout/src/serialization.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  function expectToBeRejected(value: any) {
+    expect(() => serialize(value)).toThrow(/Cannot serialize value to JSON/);
+  }
+
+  expectToBeRejected(NaN);
+  expectToBeRejected(Infinity);
+  expectToBeRejected(-Infinity);
+  expectToBeRejected(() => 1);
+  expectToBeRejected({ method: () => 1 });
+  // We could build in support for Dates if we wanted to, but starting out safe by rejecting.
```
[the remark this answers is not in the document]
jonahkagan: Yeah, I'm kind of liking keeping them as ISO strings til you actually need to do something with them.

libs/grout/src/serialization.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/**
+ * A Tagger defines how to detect and tag a particular kind of value that we
+ * want to define custom tagged serialization for.
+ *
+ * Grout could potentially be extended to allow custom taggers in the future,
+ * but for now we just support a few useful types.
+ */
+// eslint-disable-next-line vx/gts-type-parameters, vx/gts-identifiers
```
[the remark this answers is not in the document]
jonahkagan: Ok, I was wondering about this. What do you think about the convention of prefixing type variables with `T` to indicate that they are a variable? Is that just a holdover that I should abandon? I can't even remember where I got it from at this point

libs/grout/src/serialization.ts
```diff
@@ -0,0 +1,164 @@
...
+ *    __grout_type: 'undefined',
+ *    __grout_value: 'undefined'
+ *  }
+ */
+interface TaggedValue {
+  __grout_type: string;
+  __grout_value: JsonBuiltInValue;
+}
+
+function isTaggedValue(value: any): value is TaggedValue {
```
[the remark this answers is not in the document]
jonahkagan: I'll give it a shot!

libs/grout/src/util.ts
jonahkagan: Awesome, thanks!

libs/grout/README.md
```diff
@@ -0,0 +1,178 @@
...
+  },
+
+  // For known errors (e.g. invalid input), use a Result type so that the client
+  // forced to handle them explicitly.
+  async updatePersonAge(input: {
+    name: string;
+    age: number;
+  }): Promise<Result<void, Error>> {
+    if (age <= 0) {
+      return err(new Error('Age must be at least 1.'));
```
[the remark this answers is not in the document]
jonahkagan: hehe fair point

libs/grout/README.md
```diff
@@ -0,0 +1,178 @@
...
+router.use(express.text({ type: 'application/json' }));
+
+// Then, simply register the API routes.
+grout.registerRoutes(router, api);
+
+// Finally, mount the router on your app. You can use any path prefix you want -
+// here we use `/api`. It's important that you don't use any other body-parsing
+// middleware upstream of the router - e.g. don't call `app.use(express.json())`
+// before this.
+app.use('/api', router);
```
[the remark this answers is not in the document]
jonahkagan: Yeah, I did start with that. But then grout has to depend on express. I was a bit wary of ending up with mismatched express versions between grout and its consumers. I considered peer dependencies but was intimidated because they seem finicky. So that's how I landed here. But I'm very willing to reconsider and would love to hear your thoughts about the tradeoffs.

libs/grout/src/client.ts
```diff
@@ -0,0 +1,103 @@
...
+export class ServerError extends Error {}
+
+/**
+ * Creates a Grout RPC client based on the type of an API definition. You should
+ * import the API definition type using `import type` to avoid importing the
+ * server code.
+ *
+ * API methods can be called just like any regular function. For example:
+ *
+ *  import type { MyApi } from 'my-api-server-package';
```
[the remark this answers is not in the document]
jonahkagan: I tried this but it broke stylelint...

```ts
// src/api/api.ts
import { type Api } from '@votingworks/vx-scan-backend';
```
 
```sh
# stylelint output
src/api/api.ts
 2:14  ✖  Unexpected token, expected "," (2:14)
```

Maybe a newer version of stylelint would do better...

libs/grout/src/util.ts
jonahkagan: Hmm, looks like this approach thinks a `Result` is a plain object.

```
> const { ok } = require('@votingworks/types')
undefined
> Object.prototype.toString.call(ok())
'[object Object]'
```

I think the goal of this function is to differentiate plain objects from instances of classes. Any other ideas?

jonahkagan: Maybe `value.constructor.name === 'Object'`?

jonahkagan: After some researching on SO, I came up with this:

```ts
export function isPlainObject(
  value: unknown
): value is Record<string, unknown> {
  return (
    isObject(value) &&
    Object.getPrototypeOf(value) === Object.prototype &&
    value.constructor.name === Object.name
  );
}
```

libs/grout/src/serialization.ts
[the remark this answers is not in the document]
jonahkagan: Worked! f433bd077

libs/grout/src/serialization.ts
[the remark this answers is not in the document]
jonahkagan: (Changed here: f433bd077, but still curious to hear your thoughts)

libs/grout/README.md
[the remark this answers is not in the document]
jonahkagan: f433bd077

libs/grout/src/util.ts
jonahkagan: f433bd077

libs/grout/src/server.ts
arsalansufi: Ah nice, that makes sense!

libs/grout/src/server.ts
arsalansufi: Got it! The main advantage that I was seeing was not needing code to prevent the use of `app.use(express.json())` and also not having to be so careful about how we send responses, e.g. being able to use `res.json` instead of:

```ts
response.set('Content-type', 'application/json');
response.status(200).send(jsonResult);
```

But those are both minor concerns

arsalansufi (approved):
LGTM! Excited to see this in use!

libs/grout/README.md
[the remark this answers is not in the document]
eventualbuddha: Maybe a `peerDependencies` entry would fix that.

libs/grout/src/serialization.ts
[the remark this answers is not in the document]
eventualbuddha: I personally prefer either single letter type variables or descriptive names. The style you reference is common too, and Morgan and I had a disagreement about it when it was becoming the standard in Arlo. I don't care that much as long as we can square it with our lint rules.

libs/grout/README.md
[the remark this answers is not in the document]
jonahkagan: I just remembered that the other reason I did it this way was to avoid frontends having to have express as a dependency. Another solution to that would be breaking grout into two separate packages - one for frontend and one for backend. Not sure if there's an easy way to do that other than creating two libs (maybe a custom build process?)

On the other hand maybe it doesn't matter and express can just be a peer dep for frontends as well, just seems a little weird.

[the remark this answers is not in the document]
eventualbuddha: I think the inversion is fine, I just wonder if we can make it _one_ thing you have to do rather than this series of steps.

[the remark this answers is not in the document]
jonahkagan: Maybe just passing in the `express` object would solve that.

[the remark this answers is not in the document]
jonahkagan: This approach seems to work well: 461034c82. Thanks for pushing on this point!

eventualbuddha (approved):

libs/grout/src/server.ts
jonahkagan: Yeah, I can see the point about being able to use `express.json` for the whole app being easier. Since we have a way to detect that and provide a clear error message, I'm not super worried about it for now. But would like to see if it's a pain in practice. Since I've already got it working this way, I'm going to leave it for now, but will keep an eye on it.

libs/grout/src/client.ts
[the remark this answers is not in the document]
jonahkagan: Skipping this for now since it's not a high priority fix, but would be down to revisit in the future!

Issue #2682, opened 2022-10-17
WIA: Central Scanner


adghayes:
Currently central scanner cannot export CVRs with large amounts of write ins. We need to make a few small changes for that to be possible, like in #2681.

adghayes:
This ticket would be a good opportunity to revisit the issues coming out of #2681, see https://github.com/votingworks/vxsuite/pull/2681#issuecomment-1281640629

PR #2852, opened 2022-12-08
[VxCentralScan] Stream CVR exports directly to file

kofi-q:
## Overview
Closes https://github.com/votingworks/vxsuite/issues/2682:

Updating CVR file export feature to stream CVR data directly to file to avoid issues when dealing with large amounts of data (potentially from batches with lots of write-in images included).
- Changed the `POST .../export` endpoint to just accept a directory path and filename and write directly to the filesystem instead of downloading to the client first and writing via kiosk-browser.
- Added a `GET` version of the endpoint that streams the data back to the client as an attachment, to support the  case where the user manually selects an export location.
  - We could eventually do away with the `GET` endpoint altogether if we implement a server-backed file picker instead of the native one that the kiosk-browser triggers.

### Mainline Flow (kiosk-browser, default USB drive directory):
https://user-images.githubusercontent.com/264902/206526621-75bb2114-a6ba-44bb-93ca-82f2951e4109.mov

### Export in kiosk-browser to custom directory:
https://user-images.githubusercontent.com/264902/206526850-e0ad6a70-1f95-48fc-a9e3-2bae83f17656.mov

### Export in browser
https://user-images.githubusercontent.com/264902/206526976-db3fe387-34c2-412c-9ad1-38661d6b1b4f.mov

## Testing Plan 
- Updated existing tests
- Added test for new `GET .../export` endpoint
- Manual test to verify functionality

## Checklist


services/scan/src/central_scanner_app.ts
```diff
@@ -472,7 +473,57 @@ export async function buildCentralScannerApp({
...
+          ],
+        });
+        return;
+      }
+
+      const { directoryPath, filename } = request.body;
+
+      await fs.mkdir(directoryPath, { recursive: true });
+      const file = await fs.open(path.join(directoryPath, filename), 'w');
+      const fileStream = file.createWriteStream();
```
eventualbuddha, line 494: This doesn't appear to restrict what directories it is allowed to write to. Did you see [`exportToUsbDrive`](https://github.com/votingworks/vxsuite/blob/b671e57f4e5934256fe20e706ada886192f7ab3b/libs/data/src/exporter.ts#L124)? That ensures the path is one that is allowed to be written to in case the client sends a malicious request. Check out the [`/central-scanner/scan/backup-to-usb-drive` endpoint](https://github.com/votingworks/vxsuite/blob/b671e57f4e5934256fe20e706ada886192f7ab3b/services/scan/src/central_scanner_app.ts#L615-L630). Also, note that I renamed that endpoint because I explicitly didn't want the client to be allowed to choose where it went. Maybe do that here, too?

frontends/bsd/src/util/download.ts
```diff
@@ -71,6 +78,16 @@ async function kioskDownload(
...
 
+function browserDownload(url: string) {
+  const anchorElement = document.createElement('a');
+  anchorElement.href = url;
+  anchorElement.download = '';
+  anchorElement.style.display = 'none';
+  document.body.appendChild(anchorElement);
+  anchorElement.click();
+  document.body.removeChild(anchorElement);
+}
```
eventualbuddha, line 89: Oh browser, you so weird.

services/scan/src/central_scanner_app.test.ts
```diff
@@ -55,6 +56,13 @@ beforeEach(async () => {
   app = await buildCentralScannerApp({ importer, workspace });
 });
 
+afterEach(async () => {
+  await fs.rm(workspace.path, {
+    force: true,
+    recursive: true,
+  });
```
eventualbuddha, line 63: 🙏

eventualbuddha (changes requested):
Needs some thinking around how we do export to USB. Happy to discuss synchronously if anything isn't clear.

services/scan/src/central_scanner_app.ts
kofi-q: Ah, thanks for the pointers! Hadn't looked at the backup endpoint - will update to leverage `exportToUsbDrive`

apps/vx-scan/backend/src/util/exporter.ts
```diff
@@ -7,6 +7,7 @@ import { SCAN_ALLOWED_EXPORT_PATTERNS } from '../globals';
 export function buildExporter(): Exporter {
   const exporter = new Exporter({
     allowedExportPatterns: SCAN_ALLOWED_EXPORT_PATTERNS,
+    getUsbDrives,
```
kofi-q, line 10: Open to pushback here, but I needed a way to make the exporter easier to test with. LMK if we should use a different pattern here - didn't see any existing tests that depend on `Exporter` (besides the exporter tests themselves, which are able to mock `getUsbDrives` directly

eventualbuddha (approved):

PR #2853, opened 2022-12-09
Enforce `vx/no-floating-results`

eventualbuddha:
## Overview

This lint rule was broken because the `getName()` was always returning `undefined` in the cases I tried, so I included a fallback that seems to work.

## Demo Video or Screenshot
image

## Testing Plan 
Tested locally in all the packages to find the violations of the rule and fix them.

## Checklist


libs/eslint-plugin-vx/src/rules/no_floating_results.ts
```diff
@@ -15,7 +15,9 @@ interface Options {
 function isResultType(checker: ts.TypeChecker, node: ts.Node): boolean {
   // TODO: consider the actual declaration location?
   const type = checker.getTypeAtLocation(node);
-  return type.getSymbol()?.getName() === 'Result';
+  return (
+    (type.getSymbol()?.getName() ?? type.aliasSymbol?.getName()) === 'Result'
+  );
```
jonahkagan, line 20: Maybe add a comment with the context of why both are needed

jonahkagan (approved):
Excited to have this rule working!

PR #2854, opened 2022-12-09
Convert VxScan API to Grout

jonahkagan:
## Overview
Task: #2814 

_Review by commits_

Converts the VxScan API to use Grout, more or less one endpoint at a time. Skips two endpoints that aren't readily converted to JSON (exporting CVRs and loading ballot templates), which I'll convert in separate PRs, since they'll require larger changes.

## Demo Video or Screenshot
N/A

## Testing Plan
- Manually tested each feature as I went
- Updated backend tests
- Updated frontend tests

## Checklist
- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.
- [x] I have added JSDoc comments to any newly introduced exports

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #2856, opened 2022-12-14
Add mockFunction to test-utils and createMockClient to grout

jonahkagan:
## Overview
Task: #2814 

An approach for mocking Grout RPC calls using a stricter function mock than `jest.fn()`.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Automated tests
- Using this in #2854 

## Checklist


libs/test-utils/src/mock_function.test.ts
```diff
@@ -0,0 +1,218 @@
...
+      Actual: <none>"
+    `);
+  });
+
+  it('assertComplete errors if there are unexpected calls', () => {
+    const addMock = mockFunction<typeof add>('add');
+    addMock.expectCallWith(1, 2).returns(3);
+    expect(addMock(1, 2)).toBe(3);
+    expect(() => addMock(1, 3)).toThrow();
+    expect(() => addMock.assertComplete()).toThrowErrorMatchingInlineSnapshot(`
```
arsalansufi, line 165: So even though `addMock(1, 3)` will already throw and print an error, we'll get a more holistic summary from `addMock.assertComplete()` - nice!

libs/test-utils/src/mock_function.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  const mock: MockFunction<Func> = (...input: Parameters<Func>) => {
+    state.actualCalls.push({ input });
+    const expectedCall = state.expectedCalls[state.actualCalls.length - 1];
+    if (!expectedCall) {
+      const message = `Unexpected call to mock function: ${formatFunctionCall(
+        name,
+        input
+      )}`;
+      // Log to console since exceptions can be swallowed depending on how the mock function is used
```
arsalansufi, line 101: I know this means that, in some cases, we'll double log, but I see why this is necessary 👍

libs/test-utils/src/mock_function.ts
```diff
@@ -0,0 +1,180 @@
...
+ *  // The actual input must match the expected input
+ *  addMock.expectCallWith(1, 2).returns(3);
+ *  addMock(1, 3); // -> throws an error because the input doesn't match
+ *
+ *  // Every expected call must be used
+ *  addMock.expectCallWith(1, 2).returns(3);
+ *  addMock.assertComplete(); // -> throws an error because we didn't use the expected call
+ *
+ * Recommendations:
+ * - Always call mockFunction.assertComplete() to ensure that all expected calls were used
```
arsalansufi, line 81: Down the road, could be nice to find a way to make it such that writers of tests don't have to remember to do this - I could definitely see someone forgetting to do so. No simple ideas are coming to mind for this

arsalansufi (approved):
Looks great!

I'm always a bit skeptical of reinventing the wheel. I've seen cases like this where the handmade alternative starts off simple, but additional needs inevitably come up (like input matchers, as discussed at yesterday's standup), and the handmade alternative becomes harder and harder to maintain.

That said, I do see the problems that we're trying to solve here and am excited to try this out! Would still be great to keep an eye out for Jest tooling that gives us the stricter checking that we want out of the box (some kind of Jest "strict" mode? 😄)

libs/test-utils/src/mock_function.test.ts
```diff
@@ -0,0 +1,218 @@
...
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  function add(num1: number, num2: number): number {
+    throw new Error('Not implemented');
+  }
+
+  it('creates a mock function that returns a value', () => {
+    const addMock = mockFunction<typeof add>('add');
```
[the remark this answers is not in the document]
jonahkagan: Good call. I build it this way specifically with the grout Proxy use case in mind, where we have a type for the function and a name as the string, but the function itself doesn't exist. I guess I could make a fake function with the name and type and use the API you propose. I'm trying to think about whether the average use case for `mockFunction` will have the original function on hand to pass in...

Other mocking libraries make the string name an optional argument, but I decided to require it to force good practices that would make debugging easier.

libs/test-utils/src/mock_function.ts
jonahkagan: Yeah I noodled on this for a while and couldn't think of anything either. The best idea I had was making it so you could only instantiate a mock function by calling a wrapper function like this:

```ts
withMockFunction(mockFunction => ... use the mock function ... )
```

Then the wrapper can run `assertCompleted` logic after the callback finishes. This is the approach used by `withMockFetch` in Arlo.

I opted for not doing this because it doesn't mesh well with jest's philosophy about using `before` and `after` for setup/cleanup. Even though I don't like that philosophy because it encourages shared global state between tests, I was trying to play nice. Definitely willing to reconsider this decision though!

Luckily `assertCompleted` really only matters for catching extra expected mock calls, so if you forget to call it the worst thing that will happen is that you'll have some cruft in your test.

arsalansufi: Ah yes, that makes the motivation for the wrapper function approach in Arlo clearer!

I also agree that 1) we should try to stay in line with Jest principles and 2) the implications of missing `assertCompleted` aren't that bad, so I'm okay punting on this

Implementing some kind of shared global state to make things more automatic also definitely increases the risk I mentioned in my top-level comment of the handmade alternative becoming harder and harder to maintain

libs/test-utils/src/mock_function.test.ts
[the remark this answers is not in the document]
arsalansufi: > I'm trying to think about whether the average use case for mockFunction will have the original function on hand to pass in...

Ah good point! I don't feel strongly about this alternative over the current approach

libs/test-utils/src/mock_function.ts
jonahkagan: I decided to move this exception trapping functionality to the mock grout client, since the main reason we need it is that we have a bad habit of not wrapping api call promises in try/catch. I don't think it makes as much sense for a general mock function to not raise an exception. b0047f544

libs/test-utils/src/mock_function.ts
jonahkagan: Yeah, one thing I actually like a lot about pytest is that they let you run code after the test by using a generator for the fixture factory function. The fixture gets yielded and then after the test runs, control returns to the fixture factory function (example: https://github.com/votingworks/arlo/blob/39a8bf585ddf40558201e7e75779e7d2cbf199e8/server/tests/test_background_tasks.py#L31-L44). I wish there were something like that for jest

libs/test-utils/src/mock_function.test.ts
jonahkagan: Yes, I've found it quite useful so far!

PR #2862, opened 2023-01-04
refactor(vx-scan): Configure from ballot package on backend

jonahkagan:
## Overview
Task: https://github.com/votingworks/vxsuite/issues/2814

Converts VxScan to load the configuration ballot package from a USB drive on the backend. This removes some complex file transfer endpoints and allows us to easily use Grout for a new endpoint to trigger configuration. I left the USB detection and triggering of configuration on the frontend for now, since the scope of this refactor is just to use Grout.

## Demo Video or Screenshot

The UX is slightly changed, since we no longer show incremental progress as we load each ballot template.

Configuration flow
https://user-images.githubusercontent.com/530106/210628017-fc80297e-da9f-45e7-8c45-568499765760.mov

When there's no ballot package found on the USB drive

https://user-images.githubusercontent.com/530106/210628413-3552f712-8213-474b-8d24-5cb73a43e247.mov

## Testing Plan 
- Manual testing
- Updated automated tests

## Checklist


apps/vx-scan/backend/test/helpers/app_helpers.ts
```diff
@@ -40,38 +47,63 @@ import { createWorkspace, Workspace } from '../../src/util/workspace';
...
+type MockFileTree = MockFile | MockDirectory;
+type MockFile = Buffer;
+interface MockDirectory {
+  [name: string]: MockFileTree;
+}
+
+function writeMockFileTree(destinationPath: string, tree: MockFileTree): void {
+  if (Buffer.isBuffer(tree)) {
+    // Sleep 1ms to ensure that each file is created with a distinct timestamp
+    execSync('sleep 0.01');
```
eventualbuddha, line 59: Seems like this should be in the loop, maybe?

apps/vx-scan/backend/src/app.ts
```diff
@@ -35,14 +36,117 @@ const debug = rootDebug.extend('app');
...
 ) {
   const { store } = workspace;
+
   return grout.createApi({
+    async configureFromBallotPackageOnUsbDrive(): Promise<
+      Result<void, ConfigurationError>
+    > {
+      assert(!store.getElectionDefinition());
+      const [usbDrive] = await usb.getUsbDrives();
+      assert(usbDrive?.mountPoint !== undefined, 'No USB drive mounted');
```
eventualbuddha, line 67: This could just be a normal error, right?

```suggestion
      if (!usbDrive?.mountPoint) {
        return err('no_usb_drive_mounted');
      }
```

apps/vx-scan/backend/src/app.ts
```diff
@@ -35,14 +36,117 @@ const debug = rootDebug.extend('app');
...
+              scale: 2,
+            }
+          )) {
+            const ballotPageLayout = find(
+              ballotTemplate.layout,
+              (l) => l.metadata.pageNumber === pageNumber
+            );
+            layoutsWithImages.push({ ballotPageLayout, imageData: page });
+          }
+          return { ...ballotTemplate, layoutsWithImages };
```
eventualbuddha, line 120: This is potentially a lot of data to load into memory all at once. Our letter-size ballots are stored in the ballot package as 1224x1584 JPEGs. Loading them into memory using `canvas` takes about 7.4MB per page, or 14.8MB per ballot card. Some of our MS elections have up to 26 ballot styles with 1-card ballots, which is 14.8MB * 26 = 384.8MB. Better to process a single card at a time if you can manage it.

eventualbuddha: Thinking about this more, this means we now have two places in the app where we're rendering the PDFs: one here and one in `Store`. It'd be great to change the API of `Store` such that it doesn't allow this partial loading of templates and we instead just reset them, add the PDFs and layout info to the db, then load from the db so we always use the same code path? Something like that, anyway.

apps/vx-scan/backend/src/app.ts
```diff
@@ -35,14 +36,117 @@ const debug = rootDebug.extend('app');
 
 type NoParams = never;
 
+/**
+ * An interface for interacting with USB drives. We inject this into the app so
+ * that we can easily mock it in tests.
+ */
```
arsalansufi, line 42: 👍

apps/vx-scan/frontend/src/screens/unconfigured_election_screen.tsx
```diff
@@ -28,158 +19,50 @@ export function UnconfiguredElectionScreen({
...
   useEffect(() => {
-    async function attemptToLoadBallotPackageFromUsb() {
-      if (usbDriveStatus !== usbstick.UsbDriveStatus.mounted) {
-        setErrorMessage('');
-        setIsLoadingBallotPackage(false);
-        setTotalTemplates(0);
+    async function configure() {
+      setError(undefined);
+      if (usbDriveStatus !== usbstick.UsbDriveStatus.mounted) return;
+      const result = await apiClient.configureFromBallotPackageOnUsbDrive();
```
arsalansufi, line 28: Nice! Love how much simpler this has made the client

apps/vx-scan/backend/test/helpers/app_helpers.ts
```diff
@@ -206,28 +252,49 @@ export const ballotImages = {
   ],
 } as const;
 
+// Loading of HMPB templates is slow, so in some tests we want to skip it by
+// removing the templates from the ballot package.
+export function createBallotPackageWithoutTemplates(
```
arsalansufi, line 257: Could put this in `@votingworks/test-utils` since it may come in handy for other frontends (VxCentralScan?) - but also fine with leaving it here if that feels preemptive

arsalansufi (approved):
Looks great!

apps/vx-scan/backend/src/app.ts
arsalansufi: ^ I did notice that this error will be surfaced by Grout as a 500 instead of a properly handled error result object, but figured that was fine (and perhaps even intentional) since we only ever call this endpoint once a USB is mounted, so this error is never expected. That said, I could imagine a race condition where the USB is mounted and then quickly removed, triggering this case on the server

apps/vx-scan/backend/src/app.ts
```diff
@@ -35,14 +36,111 @@ const debug = rootDebug.extend('app');
...
+              (l) => l.metadata.pageNumber === pageNumber
+            );
+            layoutsWithImages.push({ ballotPageLayout, imageData: page });
+          }
+          return { ...ballotTemplate, layoutsWithImages };
+        })
+      );
+
+      // eslint-disable-next-line @typescript-eslint/no-shadow
+      store.withTransaction((store) => {
```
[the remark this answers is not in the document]
jonahkagan: Yeah, that's fair, I can drop it.

apps/vx-scan/backend/test/helpers/app_helpers.ts
jonahkagan: Would the goal be to make sure it also creates directories with a unique timestamp? I don't need that property for the current usage, but seems reasonably defensive, so I will change.

apps/vx-scan/backend/src/app.ts
jonahkagan: It was intentional to signify "this should never happen." I agree there's a possible race condition, but I wasn't sure if it was possible/worth handling. In general I imagine the same issue applies during the actual reading from the USB - if it's removed, things will explode. So that makes me think there's not a huge advantage to catching the USB removed right before L67 vs right after L67.

This does raise another question for me: how we can gracefully handle the USB being removed during the configuration process? Looking at the old code I removed, there was a catch-all handler that showed an "Unknown error" message. I'm planning to implement something like that for the whole app once I convert the frontend to react-query.

apps/vx-scan/backend/src/app.ts
jonahkagan: Thanks for pointing this out. The old code did this one at a time, so I think it makes sense to preserve that behavior given the memory implications.

As far as the changes to the store, that sounds fine to me. Do you think it's worth doing now, or is this all likely to change anyway? I know you had mentioned in the past that we might change how we do configuration

apps/vx-scan/backend/test/helpers/app_helpers.ts
jonahkagan: Yeah, I could see it being useful (along with the usb mocking as well). I'm leaning towards leaving them here for now because I've been finding it kind of confusing to track down helpers in many different packages so I'm averse to doing it unless necessary.

apps/vx-scan/backend/src/app.ts
jonahkagan: Is this what you were imagining: 6eadf3690?

eventualbuddha: I think it's okay to do the smaller change for now.

jonahkagan: Well, I already got the bigger change done haha

eventualbuddha: Hah, thanks! The only critique I'd add is something that extends deep into existing code and isn't _that_ big a deal. But since you asked, I'd replace `setHmpbTemplates(ballotTemplates: BallotPackageEntry[])` with `setHmpbTemplates(ballotTemplates: Iterable<BallotPackageEntry>)`, then make the caller pass some sort of iterator so we don't have to load all the PDFs into memory at once. However, the existing code already does that and the PDFs are much smaller than the rendered images (O(100KB) each or about ~3MB for the 26 ballot style case), so it's probably fine.

I just wanted to bring this up because I think until recently we have been pretty unconcerned with memory and performance. We're starting to change that and I think it's worth pushing on things like this so instead of the default being "oh I am processing N of a thing so I need an array" maybe moving to "oh I am processing N of a thing so I need an iterable or a stream, or maybe an array if I know it'll always be fairly small".

jonahkagan: That makes sense. Agree it's a bit too big of a change to implement here right now, but I'm glad to get a better understanding of the general constraints. I'll add a note about memory perf to this code so we don't forget.

PR #2871, opened 2023-01-05
refactor(vx-scan): Get CVRs for tally from separate endpoint

jonahkagan:
## Overview
Task: #2814 

Instead of using the streaming CVR export endpoint, we add a new endpoint just for loading the CVRs as a single JSON blob (since Grout doesn't support streaming). We don't currently need streaming in this case since images aren't included. Eventually, all of the tally calculation can be moved to the backend, but this is a minimal refactoring step that removes one consumer of the CVR export endpoint. Coming up, I'll refactor the actual exporting of CVRs to USB to happen on the backend.

## Demo Video or Screenshot
N/A
## Testing Plan 
- Some cursory manual testing

## Checklist


arsalansufi (approved):

adghayes (approved):
LGTM! My first time reviewing `grout` code, really cool.

apps/vx-scan/backend/src/app.ts
```diff
@@ -237,6 +241,17 @@ function buildApi(
       return result.isErr() ? result : ok();
     },
 
+    async getCastVoteRecordsForTally(): Promise<CastVoteRecord[]> {
```
eventualbuddha, line 244: I suggest we mark this as `@deprecated` so we don't expand its usage.

```suggestion
    /**
     * @deprecated we should build the tally in the backend instead
     */
    async getCastVoteRecordsForTally(): Promise<CastVoteRecord[]> {
```

image

jonahkagan: Will do!

PR #2876, opened 2023-01-05
docs: add instructions for signing commits

eventualbuddha:
## Overview

Uses Git's newish support for signing commits with SSH keys. Configures an SSH agent run by 1Password to provide both SSH auth and commit signing.

[Rendered](https://github.com/votingworks/vxsuite/blob/docs/1password-commit-signing/docs/commit_signing.md)

## Demo Video or Screenshot
image

## Testing Plan 
Documented as I set it up on my machine. I'd like at least one other reviewer to try it and fix any steps that don't work for them before merging.

### DIFF
--- a/docs/commit_signing.md
+++ b/docs/commit_signing.md
@@ -1,0 +1,76 @@
+# Configuring Git commit signing with 1Password
+
+## Overview
+
+This document describes how to configure Git commit signing with 1Password. Git
+v2.34 introduced support for signing commits with SSH keys. Combined with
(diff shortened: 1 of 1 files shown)


benadida (approved):
I haven't followed the steps myself yet, so just reviewing for general security: this looks pretty great, I like it, and it's a good direction for us to standardize on, even.

Suggestion for documentation: make it clearer what's in the host, and what's in the VM.

amcmanus (approved):

PR #2879, opened 2023-01-05
refactor(vx-scan): Move types from libs/api into apps/vx-scan/backend

jonahkagan:
## Overview
Task: #2814 
There were a bunch of types that were used by the scan service but only applied to the precinct scanner. Now that VxScan is its own contained app, those types can live within the backend and be imported by the frontend.

There were also a few shared scanning types that I moved into libs/types. Now libs/api/scan only contains types used by VxCentralScan, so when we convert it to the new app-based architecture, we can similarly move those types into the app.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Existing tests

## Checklist

### DIFF
--- a/apps/vx-scan/backend/src/app.ts
+++ b/apps/vx-scan/backend/src/app.ts
@@ -1,4 +1,3 @@
-import { Scan } from '@votingworks/api';
 import * as grout from '@votingworks/grout';
 import { LogEventId, Logger } from '@votingworks/logging';
 import {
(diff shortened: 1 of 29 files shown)


apps/vx-scan/backend/src/app.ts
```diff
@@ -222,9 +221,8 @@ function buildApi(
       store.setBallotCountWhenBallotBagLastReplaced(store.getBallotsCounted());
     },
 
-    async backupToUsbDrive(): Promise<Result<void, Scan.BackupError>> {
-      const result = await backupToUsbDrive(store, usb);
-      return result.isErr() ? result : ok();
+    async backupToUsbDrive(): Promise<Result<void, ExportDataError>> {
+      return await backupToUsbDrive(store, usb);
```
jonahkagan, line 225: I moved this line into `backupToUsbDrive` so it would match `exportCastVoteRecordsToUsbDrive`

eventualbuddha (approved):

Issue #2858, opened 2022-12-16
Grout v1 cleanup


jonahkagan:
- [x] Support RPC method implementations that aren't async (i.e. don't return a promise)
- [x] Automatically use node-fetch when not in a browser environment
- ~[ ] Upgrade stylelint so we can use `import { type Api } ...` instead of `import type { Api }`~
- [x] Figure out best way to organize package(s) for consumers to import easily (client/server/mock client)
- [x] Log a server-side stack trace on unexpected errors (500s)

PR #2884, opened 2023-01-06
refactor(grout): Use cross-fetch ponyfill

jonahkagan:
## Overview
Task: #2858 

This ensures that Grout works in browser and server environments seamlessly.

## Checklist


libs/grout/src/client.ts
```diff
@@ -1,12 +1,20 @@
 /* eslint-disable vx/gts-identifiers */
 /* eslint-disable vx/gts-type-parameters */
+import nodeFetch from 'node-fetch';
```
eventualbuddha, line 3: Is this elided on the client, or are we pulling in all of `node-fetch` in the browser?

jonahkagan: I think the way I have it now, it will be imported in the browser. Do you have any ideas on how to leave it out?

eventualbuddha: I think we'd have to stub it in any of the frontends that pull in `grout`. Or maybe it's not actually an issue and we should try building with this configuration and see if `node-fetch` gets pulled in.

jonahkagan: Is the goal just to avoid pulling in unnecessary code? Or do you think maybe the build won't work due to node-only dependencies? Or something else?

eventualbuddha: Both.

jonahkagan: Finally tested this out. node-fetch did cause issues in the browser (it assumes on node's Readable is in scope), so I switched to [cross-fetch](https://github.com/lquixada/cross-fetch), which is designed for this purpose, and that worked.

eventualbuddha (approved):

PR #2885, opened 2023-01-06
VxAdmin USB Formatting Feature

### LINKED ISSUE #1746: VxAdmin: USB formatting function to wipe/properly format a USB drive
vxadmin feature request: a USB formatting function to wipe/properly format a USB drive

Issue automatically created from slack feedback reported by roe 
https://votingworks.slack.com/archives/CJU9MSC6S/p1651095069214139

adghayes:
## Overview
Closes #1746. Adds feature to format USB drives as blank, single-partition FAT32 drives. Only accessible by System Administrator on "Settings" screen.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/37960853/211114948-e7b49f1a-0ed1-41a7-a050-56d3b71b240d.mov

## Testing Plan 
- [x] Automated Testing
- [x] Manual Testing with USBs of various incorrect formats, the correct format, etc.  Confirmed that USB drives can be formatted and then used with our software. Note: the feature will _not_ format a USB drive which has _no_ partitions on it, because that USB drive is not detected with our `kiosk-browser` APIs. That would require some deeper changes in kiosk-browser.
- [x] `vxsuite-complete-system` testing to make sure it all fits together. I'm blocked on this without my equipment ATM. 

## Checklist


libs/ui/src/hooks/use_usb_drive.ts
```diff
@@ -1,14 +1,48 @@
...
+export interface PostFormatUsbOptions {
+  action: 'mount' | 'eject';
+}
+
 export type UsbDriveStatus =
   | 'absent'
+  | 'bad_format'
   | 'mounting'
   | 'mounted'
   | 'ejecting'
```
adghayes, line 48: We could also have `formatting` state but I did not want to handle that in all the frontends.

adghayes (commented):
Self-review, looks sane so far.

frontends/election-manager/src/components/format_usb_modal.tsx
```diff
@@ -0,0 +1,178 @@
...
+      setErrorMessage((error as Error).message);
+    }
+  }, [usbDrive, userRole]);
+
+  // Handle USB drive being removed during stages of the  process
+  useEffect(() => {
+    if (usbDrive.status === 'absent') {
+      setState('init');
+    }
+  }, [state, usbDrive.status]);
```
jonahkagan, line 46: To avoid needing this effect, I might put the formatting flow and its associated state in a sub-component that only renders when the usb drive is mounted.

(See [docs](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) for more on why adjusting state based on props is considered an anti-pattern)

frontends/election-manager/src/components/format_usb_modal.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+type ModalState = 'init' | 'confirm' | 'formatting' | 'done' | 'error';
+
+export function FormatUsbModal({ onClose }: FormatUsbModalProps): JSX.Element {
+  const { usbDrive, auth } = useContext(AppContext);
+  assert(isSystemAdministratorAuth(auth) || isElectionManagerAuth(auth));
+  const userRole = auth.user.role;
+
+  const [state, setState] = useState<ModalState>('init');
+  const [errorMessage, setErrorMessage] = useState('');
```
jonahkagan, line 28: One best practice I've learned is to model state that changes together as one state object. In this case, there is an implicit invariant that `errorMessage` is only non-empty when `state` is not `'error'`. We can build that invariant into the data structure to help enforce it:

```ts
type ModalState =
  | { state: 'init' }
  | { state: 'confirm' }
  ...
  | { state: 'error', message: string }
```

In this particular case, it's not so complicated so it's not the most important, but thought I'd mention it because I think it's a good habit in general.

frontends/election-manager/src/components/format_usb_modal.tsx
```diff
@@ -0,0 +1,178 @@
...
+  }, [usbDrive, userRole]);
+
+  // Handle USB drive being removed during stages of the  process
+  useEffect(() => {
+    if (usbDrive.status === 'absent') {
+      setState('init');
+    }
+  }, [state, usbDrive.status]);
+
+  if (state === 'error') {
```
jonahkagan, line 48: `switch` on `state` to ensure completeness?

frontends/election-manager/src/components/format_usb_modal.tsx
```diff
@@ -0,0 +1,178 @@
...
+                  ? 'The format of the inserted USB drive is not VotingWorks compatible. Would you like to format the USB drive?'
+                  : 'The format of the inserted USB drive is already VotingWorks compatible. Would you like to re-format the USB drive?'}
+              </p>
+            </Prose>
+          }
+          onOverlayClick={onClose}
+          actions={
+            <React.Fragment>
+              <Button primary onPress={() => setState('confirm')}>
+                Format USB
```
jonahkagan, line 155: Maybe "Format USB Drive" to be consistent with the rest of the modal

frontends/election-manager/src/components/format_usb_modal.tsx
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  if (state === 'done') {
+    return (
+      <Modal
+        content={
+          <Prose>
+            <h1>USB Drive Formatted</h1>
+            <p>
+              USB drive successfully re-formatted. It is now ready to use with
```
jonahkagan, line 70: I think "reformat" is general spelled without the hyphen. If you agree we should remove the hyphen, there are a few other places that would also need to be updated

libs/ui/src/hooks/use_usb_drive.ts
jonahkagan: Yeah, that seems fine. It shouldn't be possible to get into that state other than in the code that runs the format operation.

jonahkagan (approved):
Looks good overall! A few small notes.

Would have been helpful as a reviewer to have the refactor to create the `usbDrive` object separated from the implementation of this feature

frontends/election-manager/src/components/format_usb_modal.tsx
adghayes: Agree, removed in four places.

frontends/election-manager/src/components/format_usb_modal.tsx
adghayes: I'll leave the buttons as-is, I prefer to keep the buttons less wordy but that is just a personal preference.

frontends/election-manager/src/components/format_usb_modal.tsx
adghayes: Changed.

frontends/election-manager/src/components/format_usb_modal.tsx
adghayes: Changed.

frontends/election-manager/src/components/format_usb_modal.tsx
adghayes: Changed.

PR #2886, opened 2023-01-10
feat(ui): Change listener hooks

jonahkagan:
## Overview

Adds two new hooks to `libs/ui` for registering a change listener to run on external state changes. See hook documentation in the code for full explanation. The goal is to make a safer alternative to using a `useEffect` hook to try to listen to an external state change.

Uses one of these hooks in VxScan as an example.

## Demo Video or Screenshot
N/A

## Testing Plan
- Automated tests

## Checklist

### DIFF
--- a/apps/vx-scan/frontend/src/screens/unconfigured_election_screen.tsx
+++ b/apps/vx-scan/frontend/src/screens/unconfigured_election_screen.tsx
@@ -1,8 +1,11 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 // eslint-disable-next-line vx/gts-no-import-export-type
 import type { ConfigurationError } from '@votingworks/vx-scan-backend';
 import { throwIllegalValue } from '@votingworks/utils';
-import { UsbDriveStatus } from '@votingworks/ui';
+import {
+  UsbDriveStatus,
+  useExternalStateChangeListener,
+} from '@votingworks/ui';
 import {
   CenteredLargeProse,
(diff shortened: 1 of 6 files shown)


adghayes (approved):
My understanding is that the real benefit of this will be in responding to state maintained in the backend and retrieved through queries.

For the stop-gap approach, the ergonomics / readability, at least in the single example, seem to me to be about the same as `useEffect`. How is it "safer" than `useEffect`?

jonahkagan:
> How is it "safer" than useEffect?

Instead of relying on the comparison semantics of the `useEffect` dependency array, it uses deep equality checks on one specific value. So it's guaranteed to only run if and only if that value changes.

An example:

```ts
function ExportModal({ usbDriveStatus }) {
  const exportCvrs = useMutation(...)

  useEffect(() => {
    (async () => {
      if (usbDriveStatus === 'mounted') {
        await exportCvrs.mutateAysnc()
      }
    })()
  }, [exportCvrs])
}
```

This code has a subtle bug. `exportCvrs` will change when the mutation starts and ends (due to its `status` field changing) which will cause the mutation to be executed multiple times. There are ways to work around this, but using the hook you can write the exact same effect handler without the bug:

```ts
function ExportModal({ usbDriveStatus }) {
  const exportCvrs = useMutation(...)

  useExternalStateChangeListener(usbDriveStatus, async (newUsbDriveStatus) => {
    if (usbDriveStatus === 'mounted') {
      await exportCvrs.mutateAysnc()
    }
  })
}
```

Issue #484, opened 2021-05-06
upgrade to pnpm v7


eventualbuddha:
Something about pnpm v6 is incompatible with vxsuite. Nothing in the [v6 release notes](https://github.com/pnpm/pnpm/releases/tag/v6.0.0) obviously causes it.

eventualbuddha:
Yes, especially since they're now on v7 😬

PR #2891, opened 2023-01-12
Ugrade pnpm to v7

kofi-q:
## Overview
Closes https://github.com/votingworks/vxsuite/issues/484

Upgrading pnpm from v5 to v7, which necessitated the following:
- Updated pnpm version references in the package.json files.
- Also removed the no-longer-needed `--` arg separator for `pnpm run` invocations in the project scripts - this was a breaking change from v6 -> v7.
- Migrated `pnpmfile.js` to `.pnpmfile.cjs` as required by the breaking change from v5 -> v6.
  - Also updated an outdated reference there: `eslint-module-utils` -> `@typescript-eslint/utils`
- Enabled the `auto-install-peers` .npmrc flag - after the upgrade, I was seeing pnpm install errors about missing peer dependencies. Couldn't think of any obvious issues this might cause, but please flag if any occur to you.
  - Maybe, it's better if we go through and explicitly add peer dependencies to package.json where they are missing?
- Updated some fetch-mock invocations in a couple tests that started failing. Not 100% sure what made that necessary - possibly a different fetch-mock version getting locked after the upgrade - but, changes were minor.
- Updated one functional test which was expecting to be able to check content on the screen after a user action that would trigger a route change - not sure why that ever worked, but seemed brittle anyway.
- Made a few type inference hacks to code in `codemods` - post-upgrade, was seeing type errors from a mismatch between the return type of an AST parsing method and the input type of the `traverse` method. Couldn't get things to work after trying a few different package version changes in package.json, so just went with explicit `any` typings for now.
  - Doesn't look like `codemods` is imported anywhere anyway?

## Testing Plan 
- Hoping circleci passing gives us a bit of confidence
- Also ran each frontend locally to make sure we haven't introduced any obvious runtime issues.

## Checklist


.npmrc
```diff
@@ -6,4 +6,6 @@ hoist = false
 
 # We _do_ try to hoist `@types` packages so that TS will not end up with type
 # errors from comparing types from different versions of the same package.
-public-hoist-pattern=['@types/*']
\ No newline at end of file
+public-hoist-pattern=['@types/*']
+
+auto-install-peers=true
```
jonahkagan, line 11: What would it look like to install peer deps explicitly? Is it possible to see which peer deps are currently being auto installed? I have a slight leaning towards more explicitness when it comes to package configuration, since it can already be confusing.

jonahkagan (approved):
Awesome that you got this to work

.npmrc
kofi-q: Yeah, this nudge is useful - very much leaning toward explicit imports as well, so I'm going to test that out locally and see if we can avoid this config change

eventualbuddha (approved):

.npmrc
kofi-q: Spent way too long trying to untangle the peer dependency mess, but the repo is in a bit of a brittle state and it was a can of worms trying to get it working. After digging a little more into the breaking changes pnpm introduced between v6 and v7, it looks like the previous behaviour was similar to setting `auto-install-peers` to true, so I'm going to leave it in place for now and see if we can get stricter in the future.

jonahkagan: Ok, thanks for digging into that!

PR #2892, opened 2023-01-13
USB Mocking

adghayes:
## Overview

Gives us the ability to mock USB drives. Dependent on https://github.com/votingworks/kiosk-browser/pull/142 to work properly. View the README for a simple overview of the scripts.

## Testing Plan 

None, so far. Could write a test script but I think it's worth letting this harden a bit first and see how it's used.


script/bootstrap
```diff
@@ -5,6 +5,11 @@ set -euo pipefail
 # pixman-1 is effectively a dependency in all pnpm calls, go ahead and install it 
 sudo apt install -y libsane libpng-dev libjpeg-dev libx11-dev libpixman-1-dev libcairo2-dev libpango1.0-dev libgif-dev build-essential
 
+
+# Create mount point for kiosk-browser to mount USB drives
+sudo mkdir -p /media/vx/usb-drive
+sudo chown -R $USER /media/vx
+
```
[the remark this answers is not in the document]
adghayes: This is not particular to the mocking scripts, this has to do with the `kiosk-browser` change.

libs/usb-mocking/README.md
```diff
@@ -0,0 +1,38 @@
+# USB Drive Mocking
+
+The `bash` scripts in this library are useful for mocking USB drives in
+`vxsuite`. The mock consists of a simple loop device backed by an `.img` file.
+The contained filesystem persists and its contents can be separately inspected
```
carolinemodic, line 5: how do you separately inspect and manipulate the contents? Is is just in /media/vx/usb-drive ? If so can you add that info to the readme

carolinemodic (approved):
So cool! Would be good to get @amcmanus 's eyes on this but I don't think you need to block on that to merge.

amcmanus (approved):

libs/usb-mocking/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+The current iteration of these scripts only supports one virtual USB drive image
+at a time. You can reset and resize whenever necessary.
+
+### `insert.sh` - Insert Drive
+
+"Inserts" the virtual USB drive. Does not mount the drive.
+
+```bash
+sudo bash insert.sh
```
jonahkagan, line 16: Any way to remove the `sudo bash` part at the beginning?

libs/usb-mocking/insert.sh
```diff
@@ -0,0 +1,20 @@
...
+    echo "No virtual USB drive found. Initializing a virtual USB drive with default size of 64MB..."
+    bash $DIR/initialize.sh -s 64
+    echo "To create a larger virtual USB drive, run 'sudo bash initalize.sh -s <size>' where <size> is the desired size in megabytes."
+fi
+
+# Attach the loop device
+DEVICE_BASENAME=$( basename $( losetup -fP --show $DIR/usb-mock.img ) )
+
+# Use a delay to allow the OS to detect the filesystems on the loop device partition
+sleep 1
```
jonahkagan, line 15: Is there a way to know when the OS has detected the filesystems? E.g. a command you could run to check? If so, maybe you could loop here to wait for that

libs/usb-mocking/initialize.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Create an image for the virtual USB drive
+rm -f $VIRTUAL_DRIVE_IMAGE
+dd if=/dev/zero bs=1M count=$size of=$VIRTUAL_DRIVE_IMAGE status=none
+
+# Configure the virtual USB drive
+DEVICE=$( losetup -fP --show $VIRTUAL_DRIVE_IMAGE ) # Attach loop device
+sfdisk -q $DEVICE <<< type=c # Partition with single FAT32 partition
+mkfs.fat -F 32 -n VIRTUAL-USB "$DEVICE"p1 > /dev/null # Initialize FAT32 filesystem
+losetup -d $DEVICE # Detach loop device
```
eventualbuddha, line 30: Quotes around variables to ensure spaces don't mess things up. For example:
```bash
VIRTUAL_DRIVE_IMAGE="~/a/path/with spaces"
rm -f $VIRTUAL_DRIVE_IMAGE
# ^ equivalent to `rm -f ~/a/path/with; rm -f spaces`,
#   which because of the `-f` will just fail silently
```

```suggestion
rm -f "$VIRTUAL_DRIVE_IMAGE"
dd if=/dev/zero bs=1M count="$size" of="$VIRTUAL_DRIVE_IMAGE" status=none

# Configure the virtual USB drive
DEVICE=$( losetup -fP --show "$VIRTUAL_DRIVE_IMAGE" ) # Attach loop device
sfdisk -q "$DEVICE" <<< type=c # Partition with single FAT32 partition
mkfs.fat -F 32 -n VIRTUAL-USB "${DEVICE}p1" > /dev/null # Initialize FAT32 filesystem
losetup -d "$DEVICE" # Detach loop device
```

libs/usb-mocking/insert.sh
adghayes: Yeah, that's doable!

libs/usb-mocking/README.md
adghayes: I'm not sure how do do that. Because creating loop devices always require `sudo` permissions, `sudo` has to be somewhere. I can make them executable so instead it's `sudo ./insert.sh`, but that's about all I know how to do at this point.

jonahkagan: How about wrapper scripts that call `sudo` for you? (I have no idea if this is a good idea or if it's better to explicitly call `sudo`)

jonahkagan: Ok cool, I would be in favor of a one-time setup personally as long as we feel ok about it from a security perspective. But not a huge deal - mostly just pushing on this to make the interface as easy to use as possible

libs/usb-mocking/insert.sh
```diff
@@ -0,0 +1,42 @@
...
+
+if [ -e "/dev/disk/by-id/usb-mock-part1" ]; then
+    echo "There is currently a virtual USB drive inserted. Remove it before re-initializing the virtual USB drive."
+    exit 1
+fi
+
+if [ -f "$DIR/usb-mock.img" ]; then
+    echo "Virtual USB drive image found..."
+else
+    echo "No virtual USB drivei image found. Initializing a virtual USB drive image with default size of 64MB..."
```
jonahkagan, line 13: typo? "drivei" -> "drive"

libs/usb-mocking/README.md
adghayes: I think it's a good idea but I'm going to punt on it for now. I think, based on other future feedback, we may want a single entry point (single script) which could serve as a wrapper. 

Also, I might be wrong about the `/etc/sudoers` file. Need to think about that. It can allow us to run without a password but not necessarily run without `sudo`. Maybe @eventualbuddha or @amcmanus have a suggestion? In any case, going to punt for now.

libs/usb-mocking/README.md
adghayes: added

PR #2893, opened 2023-01-13
`no-jest-to-be` Lint Rule

### LINKED ISSUE #2510: Create a custom lint rule requiring use of .toEqual in tests instead of .toBe
Relevant GitHub comment thread: https://github.com/votingworks/vxsuite/pull/2504#discussion_r960750873
Relevant article: https://dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca

adghayes:
## Overview
Closes #2510. This is not a high-priority issue (in fact, it's not even prioritized to low-priority) but I wanted to understand our linting apparatus by making a lint rule.

## Demo Video or Screenshot

Screen Shot 2023-01-13 at 1 34 56 PM

## Testing Plan 
- Tests for rule added

### DIFF
--- a/libs/eslint-plugin-vx/src/rules/no_jest_to_be.ts
+++ b/libs/eslint-plugin-vx/src/rules/no_jest_to_be.ts
@@ -1,0 +1,45 @@
+import { AST_NODE_TYPES, TSESLint, TSESTree } from '@typescript-eslint/utils';
+import { createRule } from '../util';
+
+const rule: TSESLint.RuleModule<'noJestToBe', readonly unknown[]> = createRule({
+  name: 'no-jest-to-be',
+  meta: {
+    docs: {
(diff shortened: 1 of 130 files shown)


libs/eslint-plugin-vx/src/rules/no_jest_to_be.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+    schema: [],
+    type: 'problem',
+  },
+  defaultOptions: [],
+
+  create(context) {
+    return {
+      MemberExpression(node: TSESTree.MemberExpression) {
+        if (
```
eventualbuddha, line 25: Ensure we don't match things like this: `expect(…)[toBe]()`, though such a thing will surely never be written in vxsuite.
```suggestion
        if (
          !node.computed &&
```

eventualbuddha (approved):

PR #2894, opened 2023-01-18
refactor(vx-scan): Cleanup after react-query refactor

jonahkagan:
## Overview
 
_Review by commits_

A variety of cleanup. The main themes are:
- Using react-query idiomatically (using query/mutation hooks directly by components that need them, rather than prop drilling)
- Handling errors well (one big commit dedicated to this has more info)

## Demo Video or Screenshot
N/A

## Testing Plan
- Automated tests

## Checklist


apps/vx-scan/frontend/src/components/error_boundary.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <CenteredLargeProse>
+            <h1>Something went wrong</h1>
+            <Text>Ask a poll worker to restart the scanner.</Text>
+          </CenteredLargeProse>
+        </ScreenMainCenterChild>
+      );
+    }
+
+    const { children } = this.props;
+    return children;
```
jonahkagan, line 81: The lint rules require this unnecessary destructuring

libs/grout/test-utils/src/mock_client.test.ts
```diff
@@ -41,7 +41,7 @@ test('catches exceptions from mock function failures and logs them', async () =>
 });
 
 test('doesnt catch intentional exceptions from mock functions', () => {
-  const mockClient = createMockClient<typeof api>();
+  const mockClient = createMockClient<typeof api>({ catchErrors: true });
```
arsalansufi, line 44: While the test description makes the expected behavior clear, `catchErrors: true` sounds like a setting for catching all errors, intentional or not. `catchUnexpectedErrors: true` (or something like that) might be clearer

apps/vx-scan/frontend/src/app_unhappy_paths.test.tsx
```diff
@@ -242,7 +242,6 @@ test('App shows warning message to connect to power when disconnected', async ()
   // Remove pollworker card
   card.removeCard();
   await screen.findByText('Insert Your Ballot Below');
-  return;
```
arsalansufi, line 239: 🙃 Good find!

apps/vx-scan/frontend/src/screens/election_manager_screen.tsx
```diff
@@ -49,12 +49,12 @@ export function ElectionManagerScreen({
   isTestMode,
```
arsalansufi, line 45: While I see that components are now consistently pulling react-query mutations themselves, how have you been thinking about react-query data? Specifically when to pull it in directly vs. when to pass it through props or context? I know passing data through props can let you ensure that the data is loaded, avoiding unnecessary is-loading checks

apps/vx-scan/frontend/src/components/calibrate_scanner_modal.tsx
```diff
@@ -23,12 +23,13 @@ export function CalibrateScannerModal({
...
   const [calibrationState, setCalibrationState] =
     useState<CalibrationState>('ready');
-  const makeCancelable = useCancelablePromise();
 
-  async function onCalibrate() {
+  function onCalibrate() {
     setCalibrationState('calibrating');
-    const success = await makeCancelable(calibrateMutation.mutateAsync());
-    setCalibrationState(success ? 'calibrated' : 'failed');
+    calibrateMutation.mutate(undefined, {
```
arsalansufi, line 29: Nice - I like the recommendation of preferring `.mutate` over `.mutateAsync`

Out of scope for this PR of course but would be cool if we could write a lint rule to enforce the use of `.mutate` (or `.mutateAsync` wrapped in a try-catch)!

apps/vx-scan/frontend/src/preview_dashboard.tsx
```diff
@@ -119,32 +119,29 @@ export function PreviewDashboard({
...
+    if (file) {
+      const json = await file.text();
+      const result = safeParseElectionDefinition(json);
+      if (result.isOk()) {
+        setElectionDefinitions((prev) => [...prev, result.ok()]);
+        setElectionDefinition(result.ok());
       }
-    }, []);
+    }
+  };
```
arsalansufi, line 144: 🧹✨

arsalansufi (approved):
LGTM!

apps/vx-scan/frontend/src/screens/election_manager_screen.tsx
jonahkagan: I have been leaning a bit towards pushing down queries and mutations as low in the component hierarchy as possible for a few reasons:
- Colocation makes it easier to read, especially for simple read/write components that update a single setting - you don't have to trace the source of the data back up the component hierarchy.
- Duplicated data loading in subcomponents is not a problem due to caching.
- It's nice when writing a component-level test to have the data loading and updating both happen within the component so you can easily simulate a full read/write/re-read cycle. This encourages writing tests for smaller components rather than the app-level tests we currently have (many of which just test one part of a specific sub-component, but are set up at the app level because that's where all the plumbing used to live)

However, this is explicitly the opposite of the general React guidance of lifting up shared state. So I'm not 100% convinced it's the way to go, but it's the way I've been leaning lately.

jonahkagan: We discussed and decided that it's usually good to write test that capture the entire scope of a feature, so locating the query for data used in a feature at the same level in the component hierarchy as the feature makes sense. We also should try not to write too many tests for subcomponents in order to give ourselves more flexibility about how we organize the component hierarchy over time.

This reminds me of the idea that integration tests are generally the best bang for your buck from this article: https://kentcdodds.com/blog/write-tests#mostly-integration
>Integration tests strike a great balance on the trade-offs between confidence and speed/expense.

apps/vx-scan/frontend/src/components/calibrate_scanner_modal.tsx
jonahkagan: Yes, a lint rule would be great! I'll save that for after getting feedback from the team on this approach

libs/grout/test-utils/src/mock_client.test.ts
jonahkagan: Will change!

apps/vx-scan/frontend/src/screens/election_manager_screen.tsx
jonahkagan: Will incorporate in follow up PR to remove app context

Issue #2815, opened 2022-11-29
Use react-query to manage state in VxScan frontend


jonahkagan:
- [x] initial refactor
- [x] cleanup to more idiomatic react-query code
- [x] add universal error handling
- [x] remove app context pattern, locate query at feature level (e.g. role-based screen)

PR #2895, opened 2023-01-18
refactor(vx-scan): Control sounds using query change listener

jonahkagan:
## Overview
Task: #2815 

Rather than playing sounds when screens mount using useEffect, add in a logic handler that plays sounds according to scanner status transitions. This decouples sounds from how/when we mount and render screens, making them easier to grok and less error-prone.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Manual testing

## Checklist


apps/vx-scan/frontend/src/utils/use_sound.ts
```diff
@@ -0,0 +1,6 @@
+import useSoundLib from 'use-sound';
+
+export function useSound(sound: 'success' | 'warning' | 'error'): () => void {
+  const [playSound] = useSoundLib(`/sounds/${sound}.mp3`);
+  return playSound;
+}
```
jonahkagan, line 6: This file got moved. I also moved the mute logic out of this hook into the voter screen where the sounds get played

PR #2898, opened 2023-01-19
refactor(vx-scan): Remove AppContext

jonahkagan:
## Overview
Task: #2815 

_Review by commits_

Removes the AppContext and instead uses either:
- Query data passed down through props
- Local component query hook calls

I wasn't too strict about which of the two approaches I used for now, since the main motivation for using one over the other would be to make tests easier to write, so I mainly just followed what would work well with the existing tests.

## Demo Video or Screenshot
N/A

## Testing Plan 
- Updated existing tests
- Some manual testing
- TODO test the preview app

## Checklist


arsalansufi (approved):
Thanks for the easy-to-review PR!

mr-clean-lgtm

PR #2918, opened 2023-01-24
Handwritten CDF version of famous names election

jonahkagan:
Based on https://github.com/votingworks/vxsuite/blob/jonah/ballot-def-cdf-research/libs/fixtures/data/electionFamousNames2021/election.json. Only includes one contest for brevity.

### DIFF
--- a/libs/fixtures/data/electionFamousNames2021/cdf-election.ts
+++ b/libs/fixtures/data/electionFamousNames2021/cdf-election.ts
@@ -1,0 +1,244 @@
+const cdfElection = {
+  '@type': 'BallotDefinition.BallotDefinition',
+
+  // Definition of the election (mainly candidates, contests, and ballot styles)
+  Election: [
+    {
+      '@type': 'BallotDefinition.Election',
+      ElectionScopeId: 'state-hamilton', // The GPUnit for this election
+      StartDate: '2023-02-01',
+      EndDate: '2023-02-01',
+      Type: 'general', // Other types: primary, runoff, special, etc.
+      Name: {
+        '@type': 'InternationalizedText',
+        Text: [
+          {
+            '@type': 'BallotDefinition.LanguageString',
+            Language: 'en',
+            Content: 'Lincoln Municipal General Election',
+          },
+        ], // Election name in other languages would go here
+      },
+
+      // Definition of all candidates in the election across contests
+      Candidate: [
+        {
+          '@type': 'BallotDefinition.Candidate',
+          '@id': 'sherlock-holmes',
+          BallotName: {
+            '@type': 'InternationalizedText',
+            Text: [
+              {
+                '@type': 'BallotDefinition.LanguageString',
+                Language: 'en',
+                Content: 'Sherlock Holmes',
+              },
+            ],
+          },
+          PartyId: 'party-0',
+        },
+        {
+          '@type': 'BallotDefinition.Candidate',
+          '@id': 'thomas-edison',
+          BallotName: {
+            '@type': 'InternationalizedText',
+            Text: [
+              {
+                '@type': 'BallotDefinition.LanguageString',
+                Language: 'en',
+                Content: 'Thomas Edison',
+              },
+            ],
+          },
+          PartyId: 'party-1',
+        },
+      ],
+
+      // Definition of contests
+      Contest: [
+        {
+          '@type': 'BallotDefinition.CandidateContest',
+          '@id': 'mayor',
+          ElectionDistrict: 'district-1',
+          // TODO what is the analogous field for "section"?
+          Name: 'Mayor', // Contest name, not for ballot
+          BallotTitle: {
+            // Display name for contest on ballot
+            '@type': 'InternationalizedText',
+            Text: [
+              {
+                '@type': 'BallotDefinition.LanguageString',
+                Language: 'en',
+                Content: 'Mayor',
+              },
+            ],
+          },
+          VotesAllowed: 1,
+          // TODO there's no field analogous to "allowWriteIns" in CDF
+          // Contest options reference the candidates defined above by id
+          ContestOption: [
+            {
+              '@type': 'BallotDefinition.CandidateOption',
+              CandidateIds: ['sherlock-holmes'],
+            },
+            {
+              '@type': 'BallotDefinition.CandidateOption',
+              CandidateIds: ['thomas-edison'],
+            },
+          ],
+        },
+      ],
+
+      // Definition of ballot styles
+      BallotStyle: [
+        {
+          '@type': 'BallotDefinition.BallotStyle',
+          '@id': 'ballot-style-1',
+          // TODO do we also reference districts here? what's the relationship between districts and precincts?
+          GpUnit: ['precinct-23', 'precinct-22'],
+        },
+      ],
+    },
+  ],
+
+  // Definition of parties
+  Party: [
+    {
+      '@type': 'BallotDefinition.Party',
+      '@id': 'party-0',
+      // TODO there's no way to distinguish between "name" and "fullName"
+      // This Name goes on the ballot
+      Name: {
+        '@type': 'InternationalizedText',
+        Text: [
+          {
+            '@type': 'BallotDefinition.LanguageString',
+            Language: 'en',
+            Content: 'Democrat',
+          },
+        ],
+      },
+      Abbreviation: {
+        '@type': 'InternationalizedText',
+        Text: [
+          {
+            '@type': 'BallotDefinition.LanguageString',
+            Language: 'en',
+            Content: 'D',
+          },
+        ],
+      },
+    },
+  ],
+
+  // Definition of "geo-political units" (e.g. states, counties, districts, precincts, etc)
+  GpUnit: [
+    {
+      '@type': 'BallotDefinition.ReportingUnit',
+      '@id': 'state-hamilton',
+      Type: 'state',
+      Name: {
+        '@type': 'InternationalizedText',
+        Text: [
+          {
+            '@type': 'BallotDefinition.LanguageString',
+            Language: 'en',
+            Content: 'State of Hamilton',
+          },
+        ],
+      },
+      ComposingGpUnit: ['county-franklin'],
+    },
+    {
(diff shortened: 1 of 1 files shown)


libs/fixtures/data/electionFamousNames2021/cdf-election.ts
```diff
@@ -0,0 +1,244 @@
...
+        },
+      ],
+
+      // Definition of contests
+      Contest: [
+        {
+          '@type': 'BallotDefinition.CandidateContest',
+          '@id': 'mayor',
+          ElectionDistrict: 'district-1',
+          // TODO what is the analogous field for "section"?
```
jonahkagan, line 63: I wasn't sure what the "section" field in the VX contest definition represented, so it wasn't clear how to represent that in CDF

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
```diff
@@ -0,0 +1,244 @@
...
+            Text: [
+              {
+                '@type': 'BallotDefinition.LanguageString',
+                Language: 'en',
+                Content: 'Mayor',
+              },
+            ],
+          },
+          VotesAllowed: 1,
+          // TODO there's no field analogous to "allowWriteIns" in CDF
```
jonahkagan, line 77: Ideas for how to represent the "allowWriteIns" setting? What do we use it for?

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
```diff
@@ -0,0 +1,244 @@
...
+        },
+      ],
+
+      // Definition of ballot styles
+      BallotStyle: [
+        {
+          '@type': 'BallotDefinition.BallotStyle',
+          '@id': 'ballot-style-1',
+          // TODO do we also reference districts here? what's the relationship between districts and precincts?
+          GpUnit: ['precinct-23', 'precinct-22'],
```
jonahkagan, line 98: In the VX election def, ballot styles are associated with precincts and districts. Why is that? Do we need to preserve that in CDF?

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
```diff
@@ -0,0 +1,244 @@
...
+      ],
+    },
+  ],
+
+  // Definition of parties
+  Party: [
+    {
+      '@type': 'BallotDefinition.Party',
+      '@id': 'party-0',
+      // TODO there's no way to distinguish between "name" and "fullName"
```
jonahkagan, line 109: In VX election def, we have a "name" and "fullName" field for a party. What are they used for? In CDF, we can only have one by default

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
```diff
@@ -0,0 +1,244 @@
...
+        '@type': 'InternationalizedText',
+        Text: [
+          {
+            '@type': 'BallotDefinition.LanguageString',
+            Language: 'en',
+            Content: 'Franklin County',
+          },
+        ],
+      },
+      // TODO what's the relationship between counties and districts?
```
jonahkagan, line 166: GpUnits in CDF are flexible and can model many different hierarchies. VX election def seems more constrained. How should we approach that?

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
```diff
@@ -0,0 +1,244 @@
...
+  GeneratedDate: '2023-01-23T00:00:00Z',
+  Issuer: 'VotingWorks',
+  IssuerAbbreviation: 'VX',
+  VenderApplicationId: 'VxDefine', // App that generated this file
+  Version: '1.0.0', // CDF version
+  SequenceStart: 1, // Report number (out of a series of reports - don't think this applies for this use case but is required)
+  SequenceEnd: 1, // Total number of reports in series
+
+  // Extra fields not in CDF
+  // TODO should these be specified in this file?
```
jonahkagan, line 227: Do we want to extend the CDF to include these fields?

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
adghayes: I think the way that we're handling `section` is not great. Despite the name, in the code all that `section` really means is a subtitle for the race. @mattroe, Beau, and I have done a lot of bike shedding about what we should use that subtitle for in our MS ballots. And you'll see in almost all cases in our fixtures, `section` is the same as the associated district name. But code and design-wise, it's just a subtitle. And unlike most subtitles, it goes _above_ the title for each contest.

I think we need some data on what titles and subtitles really need to go above races on ballots in states we want to enter. As far as creating a ballot from a CDFs, we could do a few things:
- Use `BallotSubTitle` if exists, have no subtitle otherwise
- Always use the name of the associated GpUnit
- Use `BallotSubTitle` if exists, the name of the GpUnit otherwise

As far as converting one of our definitions to CDF... I think it's hard because we haven't ever been clear on what exactly `section` is. But for now it could be BallotSubTitle or completely fine to drop IMO.

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
adghayes: `allowWriteIns` tells us whether to put write in bubbles and spaces in a given contest when it's laid out on the ballot. We include the same number of bubbles as VotesAllowed.

With the CDF, maybe we need to lay out `CandidateOption`s with `isWriteIn` as `true`?

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
adghayes: In CDF-land, precincts and districts are all `GpUnit`s, and I've also been wondering how we handle. This seems like one of the thornier issues. I wonder if we want to move to a similar model internally, but that would be a huge change. After all, say you have a mail-in state. Not all ballots are associated with a precinct. But in our system, all ballots are always associated with a precinct. 

We use `districts` to determine which contests are on a ballot style, because we don't have any idea of composing `GpUnits`. For us, `precincts` is basically a reporting split and a label in the bottom the ballot - it doesn't give any information about what contests to put on a ballot. So, we would need to maintain the data somehow or change how we handle this internally.

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
adghayes: "Democratic" vs "Democratic Party", I think we originally used name but have moved to `fullName` as it supports localization better (e.g. "Democratic Party" --> "Partido Demócrata"). We still are using `name` in places but we probably want to drop it entirely. @benadida probably has most context here.

I think we should just use full name as we move to CDF.

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
adghayes: Since we use these fields to layout the ballot and interpret scans, I think we have to include them somehow.

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
mattroe: It doesn't surprise me that there isn't an analogous field represented in the CDF because I haven't seen a field like our the intended use of our section on any other ballot. We don't even use the section in the way Beau intended when we convert from SEMS.

On all of our VotingWorks ballots in MS, which are converted from SEMS, the "section" on the ballot is most similar to the election district value and in many cases is duplicative of the district name in the election definition. As Beau explained it to me, this was not the original intention of this field - it was intended to be analogous to "level of government" such as "federal", "state", "local" etc. which is metadata uncommonly found on the ballots. I would be happy moving away from the section concept entirely and in our default ballot design aligning it with district as we do in MS today.

mattroe: I wrote my note before I saw @adghayes's comment, but overall I think we're on the same page. It's just a subtitle that we in practice use most commonly to display the "district"

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
mattroe: VVSG requires that we support up to three levels of geographic hierarchy and we only have the hardcoded two. The desired levels are not going to be consistent across customers and elections, which is a problem we've already run into in NH and MS. I think we're going to need to move to a flexible model where we aren't constrained to the district/precinct hierarchy only.

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
mattroe: IIRC we can put the seal somewhere in the physical layout?

For now, we need to include them as this is the only way we can configure, but longterm we may want to move adjudication/threshold configuration into the app.

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
mattroe: I agree that the geographies are probably the trickiest bit here that we should probably discuss, but as mentioned in another comment, I think we're going to want to move to a model where we aren't constrained to the fix district/precinct hierarchy only. For reporting, we'll need to be able to cut by all geographic associations.

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
jonahkagan: Yeah I thought about that workaround. Seems reasonable

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
jonahkagan: Sounds good to me

libs/fixtures/data/electionFamousNames2021/cdf-election.ts
jonahkagan: Great context. I like the idea of just using the GpUnit name if that's what we're using `section` for today. Easy to add support for `BallotSubTitle` later on.

PR #2919, opened 2023-01-24
Set up Grout BFF for VxMark with Machine Config endpoint

carolinemodic:
## Overview
Sets up a basic BFF for VxMark with the machine config endpoint moved there. This ended up being a more involved change then I thought it would with all of the build config issues, integration testing, etc. I tried to split things up by commit but there are a few fairly large commits in here. Some things to note: 

- For some reason if I exported the interface object types like MarkOnly or MarkAndPrint from vx-mark-backend it would throw a crazy util.inherits is not implemented error when used in the frontend. I couldn't figure this out but moving the types to the shared types lib fixed it so I just did that. 
- We should move integration testing for bmd to it's own integration testing folder, I can do that as a followup didn't want to add more here. The integration testing is working through somewhat brute force right now. 
- I will need to make a followup PR to update complete-system 

## Testing Plan 
Used the app, saw machine config populated correctly and rest of VxMark functionally working. Ran tests. 

## Checklist


libs/logging/src/log_documentation.test.ts
```diff
@@ -89,7 +89,7 @@ describe('test cdf documentation generation', () => {
 });
 
 describe('test markdown documentation generation', () => {
-  test('generated documentation is up to date you need to run `pnpx esr --cache scripts/generate_documentation.ts` if this fails', () => {
+  test('generated documentation is up to date you need to run `pnpx esbuil-runner --cache scripts/generate_documentation.ts` if this fails', () => {
```
arsalansufi, line 92: I think we're missing a `d` at the end of `esbuil` here

apps/vx-mark/frontend/src/app_root.tsx
```diff
@@ -757,6 +718,15 @@ export function AppRoot({
     storage,
     initializedFromStorage,
   ]);
+  if (!machineConfigQuery.isSuccess) {
+    return (
+      <UnconfiguredScreen
+        hasElectionDefinition={Boolean(optionalElectionDefinition)}
+      />
```
arsalansufi, line 725: I'm wondering whether this is the right screen to display for this case since it's really just an ephemeral loading state and not a true unconfigured state. Something like [VxScan's loading configuration screen](https://github.com/votingworks/vxsuite/blob/main/apps/vx-scan/frontend/src/screens/loading_configuration_screen.tsx) might make more sense

arsalansufi (commented):
Thank you for taking this on!!

apps/vx-mark/backend/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+You generally should not need to run the backend directly. Instead, run the
+frontend, which will automatically run the backend.
+
+```sh
+cd apps/vx-mark/frontend
+pnpm start
+```
+
+The commands below assume you'll be running them as described above.
```
jonahkagan, line 17: Remove this line since there are no "commands below"?

apps/vx-mark/backend/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "@votingworks/types": "workspace:*",
+    "debug": "^4.3.1",
+    "dotenv": "^16.0.2",
+    "dotenv-expand": "^9.0.0",
+    "express": "^4.18.0",
+    "fs-extra": "^9.0.1",
+    "js-sha256": "^0.9.0",
+    "multer": "^1.4.2",
+    "uuid": "^8.3.2",
+    "zod": "3.14.4"
```
jonahkagan, line 49: I would imagine many of these deps (e.g. `multer`) and dev deps (e.g. `node-fetch`) aren't necessary  unless I'm missing something. Was this list copied from another app? Maybe start with an empty list and add deps as needed?

apps/vx-mark/frontend/src/app_root.tsx
jonahkagan: I think rendering `null` is also not a bad choice for queries that we expect to load extremely quickly like this one

apps/vx-mark/frontend/src/components/error_boundary.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * A React error boundary component that catches any errors that occur in its
+ * children. Note that may only be used to catch errors that occur in the render
+ * phase of the React lifecycle. It should not be used to catch errors that
+ * occur in event handlers or other async code, which should still use
+ * try/catch.
+ *
+ * The one exception is errors that occur in react-query hooks, which may be
+ * propagated to the error boundary using the `useErrorBoundary` setting.
+ */
+export class ErrorBoundary extends React.Component<Props, State> {
```
jonahkagan, line 22: This component is probably a good candidate for factoring out to `libs/ui`

apps/vx-mark/frontend/src/apimachine_config.test.tsx
```diff
@@ -1,113 +1,32 @@
...
-  const machineConfig: Provider<MachineConfig> = {
-    // eslint-disable-next-line @typescript-eslint/require-await
-    async get(): Promise<MachineConfig> {
-      // @ts-expect-error - we're mocking an API failure
-      return undefined;
-    },
-  };
-
-  // Render app which gets machineConfig in componentDidMount
+test('machineConfig is fetched from /machine-config by default', async () => {
```
jonahkagan, line 21: Maybe change the name of this test since /machine-config isn't a thing anymore?

apps/vx-mark/frontend/src/apimachine_config.test.tsx
```diff
@@ -1,113 +1,32 @@
...
-      machineConfig={machineConfig}
       reload={jest.fn()}
+      apiClient={apiMock.mockApiClient}
     />
   );
   await advanceTimersAndPromises();
-
-  // No expect?
-  // Unfortunately, the only thing this test does is provide code-coverage
-  // for the `else` block of `setMachineConfig`
```
jonahkagan: Is any of the machine config rendered on the screen for us to check?

apps/vx-mark/frontend/src/app.test.tsx
```diff
@@ -9,47 +9,87 @@ import {
...
+  apiMock.expectGetMachineConfig({ machineId: '0002', codeVersion: '3.14' });
 
-  render(<App hardware={hardware} />);
+  render(<App hardware={hardware} apiClient={apiMock.mockApiClient} />);
 
   await screen.findByText('Card is Backwards');
   expect(fetchMock.done()).toEqual(true);
 });
 
+it('will throw an error when using default api', async () => {
```
jonahkagan, line 51: What is this test for? I don't quite understand it

apps/vx-mark/frontend/src/app_refresh.test.tsx
```diff
@@ -74,12 +80,15 @@ it('Refresh window and expect to be on same contest', async () => {
   });
 
   unmount();
+  apiMock.mockApiClient.assertComplete();
+  apiMock.mockApiClient.reset();
```
jonahkagan, line 84: Why reset here? Is it to help make debugging easier by separating phases of the test?

apps/vx-mark/frontend/src/config/types.ts
```diff
@@ -6,78 +6,14 @@ import {
...
-      return PrintOnly;
-    case MarkOnly.key:
-      return MarkOnly;
-    case MarkAndPrint.key:
-      return MarkAndPrint;
-    default:
-      throw new Error(`unknown app mode: ${key}`);
-  }
-}
+import { MachineConfig } from '@votingworks/vx-mark-backend';
```
jonahkagan, line 16: `import type` here?

apps/vx-mark/backend/src/app_config.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    codeVersion: 'dev',
+    screenOrientation: 'portrait',
+  });
+});
+
+test('defaults to MarkAndPrint on unknown app mode', async () => {
+  const originalEnv = process.env;
+  process.env = {
+    ...originalEnv,
+    VX_APP_MODE: 'Alpaca',
```
jonahkagan, line 54: 🦙

apps/vx-mark/frontend/package.json
```diff
@@ -82,10 +84,12 @@
     "@types/styled-components": "^5.1.7",
     "@votingworks/ballot-encoder": "workspace:*",
     "@votingworks/fixtures": "workspace:*",
+    "@votingworks/grout": "workspace:*",
     "@votingworks/logging": "workspace:*",
     "@votingworks/types": "workspace:*",
     "@votingworks/ui": "workspace:*",
     "@votingworks/utils": "workspace:*",
+    "@votingworks/vx-mark-backend": "workspace:*",
```
jonahkagan, line 92: We should be able to move this to dev deps and only use `import type` to pull in types from the backend

jonahkagan (commented):
This is awesome!

apps/vx-mark/backend/package.json
carolinemodic: yeah I copied from vx-scan backend and tried to strip out the stuff that seemed obviously not useful but probably left some unnecessary things in, will see what I can remove.

jonahkagan: This tool might help: https://www.npmjs.com/package/depcheck

apps/vx-mark/frontend/src/app.test.tsx
carolinemodic: its really just to get code coverage for the case when the appClient param is not defined in App

apps/vx-mark/frontend/src/app_refresh.test.tsx
carolinemodic: I thought I was getting errors without this but now I'm not so... will remove

apps/vx-mark/frontend/src/config/types.ts
carolinemodic: I don't think I can do import type because MachineConfig is an interface

eventualbuddha: `interface` and `type` are nearly equivalent, and you can `import type` with an `interface`.

carolinemodic: oh.. when I try to add type eslint just autoremoves it so I assumed it didn't work with interfaces but I'll play around more

arsalansufi: I think it'll work if you disable the relevant eslint rule!

https://github.com/votingworks/vxsuite/blob/f0147e2eee46d5fe54be6617ecc62499b1b7bf42/apps/vx-scan/frontend/src/app.tsx#L9-L10

I also recently got tripped up by the auto-removal

carolinemodic: oh duh thanks

carolinemodic:
@jonahkagan this should be ready for rereview

apps/vx-mark/frontend/src/app.test.tsx
jonahkagan: Ah gotcha. Maybe leave a comment explaining that?

jonahkagan (approved):

Issue #837, opened 2021-09-13
figure out how to host images in a shared way across apps


carolinemodic:
So we can have all the seals in one spot instead of duplicated across bsd and precinct-scanner

carolinemodic:
@kofi-q I believe you're project will handle the seals issue here at least

PR #2921, opened 2023-01-24
Install storybook.js for shared component development/demos

kofi-q:
## Overview
Installing [storybook.js](https://storybook.js.org/) in libs/ui as a way for us live demo shared components in development without having to render them in another app first.

Notes:
- Had to install the storybook v7 beta for better pnpm and vite support. The current v6 release requires a bit more hacking to get working with pnpm workspaces and vite.
- Support for pnpm workspaces is still not fully baked in storybook and I ended up having to install the `@storybook/*` dependencies and related packages in the root workspace for everything to work properly. Hopefully, we can eventually move the dependencies into libs/ui once v7 officially launches.
  - Let me know if there are any potential issues/concerns with having these deps at the root level in the meantime.
- Added a few sample "story" files for testing. Noticed storybook doesn't play well with pure `styled-components` component exports, so I'll probably be migrating files that use that pattern to export components with explicit props instead. That's preferable anyway, to keep the shared components opinionated and standardized.
- Excluded `*.stories.ts[x]` files from a couple eslint rules that don't apply (we can think of these files as test files, in a sense).

## Demo Video or Screenshot

https://user-images.githubusercontent.com/264902/214414842-7a0b82b9-69a3-43ef-a44b-9314dfe36f2d.mov

## Testing Plan 
- Added a few `*.stories.tsx` files to test that demos are getting loaded and displayed properly.
- Added a basic unit test for the new `AppBase` component introduced

## Checklist


libs/ui/.storybook/manager-head.html
```diff
@@ -0,0 +1 @@
+<link rel="shortcut icon" href="/favicon.ico" />
```
adghayes, line 1: It's a minute problem for another day, but this favicon is super old. We should replace it across the apps.

libs/ui/.storybook/main.ts
```diff
@@ -0,0 +1,76 @@
...
+          // Create aliases for all workspace packages, i.e.
+          //
+          //   {
+          //     '@votingworks/types': '…/libs/types/src/index.ts',
+          //     '@votingworks/utils': '…/libs/utils/src/index.ts',
+          //      …
+          //   }
+          //
+          // This allows re-mapping imports for workspace packages to their
+          // TypeScript source code rather than the built JavaScript.
```
adghayes, line 56: Does this mean that `storybook` can reflect changes in codes without rebuilding the packages?

libs/ui/src/prose.tsx
```diff
@@ -95,3 +97,7 @@ export const Prose = styled('div')<ProseProps>`
     border-top: 0.1em solid #666666;
   }
 `;
+
+export function Prose(props: ProseProps): JSX.Element {
+  return <ProseContainer {...props} />;
+}
```
adghayes, line 103: Could you elaborate more about your comment in the PR description about switching these to `React` components? Specifically "That's preferable anyway, to keep the shared components opinionated and standardized."

adghayes (approved):
The setup looks good and sane to me!

libs/ui/.storybook/main.ts
kofi-q: Yeah, this will allow it to live-update any changes in the dependency packages as well - copied this setup from the vite configs we have in the other frontend packages

libs/ui/.storybook/manager-head.html
kofi-q: Ah, good to know! Makes me think I can eventually move a lot of the base html scaffolding into AppBase as well (or at least some common html template), then stuff like the favicon can just be updated in one place

libs/ui/.storybook/preview.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const parameters: Parameters = {
+  actions: { argTypesRegex: '^on[A-Z].*' },
+  controls: {
+    matchers: {
+      color: /(background|color)$/i,
+      date: /Date$/,
+    },
+  },
+};
```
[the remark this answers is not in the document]
kofi-q: Ah, this was boilerplate generated by storybook - it basically matches any prop name that matches the pattern on the right to the control type on the left (so a `textColor` prop will cause storybook to render a visual color selector control instead of just a text field for the color, and `startDate` will render a calendar control)

Glad you asked - makes me realize a comment will be useful here

libs/ui/src/prose.tsx
kofi-q: Sure thing - a `styled.div` will expose all possible props for `<div>`, whereas if we only export the wrapper component, we can limit what props are accepted to just what's relevant for rendering a piece of prose text, reducing and simplifying the API surface area.

That also makes sure that if we want to add something like a a11y props, we have to explicitly add them within this component (or expose additional props that control which a11y props are passed to the div) rather than allowing one-off a11y settings to be added in one frontend and not the other (if that makes sense)

libs/ui/.storybook/manager-head.html
adghayes: Relatedly: https://github.com/votingworks/vxsuite/issues/837.

kofi-q: Ohhh, nice - I have thoughts 😁

PR #2934, opened 2023-01-25
Sketching election ADT

jonahkagan:
## Checklist

### DIFF
--- a/libs/ballot-encoder/src/index.test.ts
+++ b/libs/ballot-encoder/src/index.test.ts
@@ -2,18 +2,18 @@ import {
   electionSampleLongContentDefinition as electionDefinition,
   electionWithMsEitherNeitherDefinition,
 } from '@votingworks/fixtures';
+import { buildVotes } from '@votingworks/test-utils';
 import {
   BallotIdSchema,
   BallotType,
   BallotTypeMaximumValue,
   CompletedBallot,
-  getContests,
+  ContestVotes,
+  ElectionADT,
   HmpbBallotPageMetadata,
-  isVotePresent,
   unsafeParse,
-  vote,
-  VotesDict,
 } from '@votingworks/types';
+import { BallotMeasureContest } from '@votingworks/types/src/election_adt';
 import '../test/expect';
 import { BitWriter, toUint8, Utf8Encoding } from './bits';
 import {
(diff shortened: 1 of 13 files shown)


libs/ballot-encoder/src/index.test.ts
```diff
@@ -84,13 +86,15 @@ test('can detect our legacy formats', () => {
 });
 
 test('encodes & decodes with Uint8Array as the standard encoding interface', () => {
-  const { election, electionHash } = electionDefinition;
-  const ballotStyle = election.ballotStyles[0]!;
-  const precinct = election.precincts[0]!;
+  const { electionHash } = electionDefinition;
```
adghayes, line 89: Did you ever look into providing the hash from the ADT? We'd have to build the ADT off of `electionData` rather than the object, but it could take out a step.

And any reason why you'd can't define the ADT election at the top level apart from any particular test, and reuse, if you don't want to repeat this conversion?

adghayes (commented):
I left some brief thoughts on `votes` and fixtures.

I'm not sure how to make these changes incrementally. The `votes` change is a great example, because lots of other data structures depend on that. I wonder if it actually might be easier to strip down the existing `CandidateVote` and just use that. In either case, I think we might have to alter `Votes` in a separate PR. Let's sync up!

libs/types/src/index.ts
```diff
@@ -6,6 +6,8 @@ export * from './cast_vote_record';
 export * from './converter_client';
 export * from './dom';
 export * from './election';
+export * as ElectionADT from './election_adt';
```
eventualbuddha, line 9: I wondered why this was being allowed. Turns out it's a bug: https://github.com/votingworks/vxsuite/pull/2995.

libs/types/src/election_adt.ts
```diff
@@ -0,0 +1,257 @@
...
+  precinctScanAdjudicationReasons?: readonly AdjudicationReason[];
+
+  // Localization
+  /** @deprecated */
+  _lang?: Translations;
+  /** @deprecated */
+  ballotStrings?: BallotStrings;
+}
+
+export function buildElectionFromVxf(electionData: VxElectionData): Election {
```
eventualbuddha, line 132: We (currently) call `electionData` the raw JSON string. Do you plan to rename that to avoid ambiguity?

jonahkagan:
Not pursuing the ADT for now

PR #2936, opened 2023-01-25
Introduce `message-coder` library

eventualbuddha:
## Overview

This grew out of the work on the Custom driver. The idea is that we want to encode structured JavaScript data in the binary form that the Custom scanner accepts, and also to decode responses into structured JavaScript data. 

## Demo Video or Screenshot
From the [README](https://github.com/votingworks/vxsuite/blob/feat/libs/message-coder/libs/message-coder/README.md):
```ts
import {
  CoderType,
  fixedString,
  literal,
  message,
  uint8,
} from '@votingworks/message-coder';

enum Genre {
  Blues = 0x00,
  ClassicRock = 0x01,
  Country = 0x02,
  Dance = 0x03,
  Disco = 0x04,
  Funk = 0x05,
  Grunge = 0x06,
  HipHop = 0x07,
  Jazz = 0x08,
  // ...
}

// define a coder for ID3v1 tags
const id3v1 = message({
  header: literal('TAG'),
  title: fixedString(30),
  artist: fixedString(30),
  album: fixedString(30),
  year: fixedString(4),
  comment: fixedString(28),
  zeroByte: literal(0x00),
  track: uint8(),
  genre: uint8(Genre),
});
type id3v1 = CoderType<typeof id3v1>;

// encode a tag
const tag: id3v1 = {
  title: 'The Title',
  artist: 'The Artist',
  album: 'The Album',
  year: '2000',
  comment: 'The Comment',
  track: 1,
  genre: 2,
};

console.log('Tag:', tag);
console.log('Encoded:', id3v1.encode(tag));

// open an mp3 file and read the last 128 bytes,
// then decode it as an ID3v1 tag
const file = fs.readFileSync('song.mp3');
const tagBytes = file.slice(file.length - 128);
console.log('Decoded from song.mp3:', id3v1.decode(tagBytes));
```

## Testing Plan 
Added a bunch of tests, and have been using it in the Custom driver library.

## Checklist


libs/message-coder/README.md
```diff
@@ -0,0 +1,67 @@
...
+  title: fixedString(30),
+  artist: fixedString(30),
+  album: fixedString(30),
+  year: fixedString(4),
+  comment: fixedString(28),
+  zeroByte: literal(0x00),
+  track: uint8(),
+  genre: uint8(Genre),
+});
+type id3v1 = CoderType<typeof id3v1>;
```
eventualbuddha, line 42: Some useful things to know about this:

1. `message` coders know how to ignore `literal` and `padding` coder properties, so the inferred type of `id3v1` is `MessageCoder<{ title: string; artist: string; album: string; year: string; comment: string; track: number; genre: Genre }>`, which does not include `header` or `zeroByte`.
2. `CoderType` extracts the value that a given `Coder` type encodes (similar to `z.TypeOf` in zod), and that type is given a name that matches the coder value as a convention that I think I like.

libs/message-coder/src/types.ts
```diff
@@ -0,0 +1,148 @@
...
+  result: Result<T, CoderError>,
+  fn: (value: T) => U
+): Result<U, CoderError>;
+
+/**
+ * Maps a `Result` to a new `Result` by applying a function to a contained `Ok`
+ * value and optionally wrapping the result in an `Ok`, leaving an `Err` value
+ * untouched.
+ */
+export function mapResult<T, U, E>(
```
eventualbuddha, line 120: This is here rather than in `result.ts` because I'm not sure yet whether we want this. Also, I spent a while trying to get the types right and found it difficult. I found [`@badrap/result`](https://github.com/badrap/result) which we may want to switch to at some point since it seems to get all the tricky type stuff right. cc @jonahkagan

libs/message-coder/src/message_coder.ts
```diff
@@ -0,0 +1,123 @@
...
+ */
+type CoderPartsOperand<T> = T extends MessageCoderParts<infer U> ? U : never;
+
+/**
+ * Gets the type of a coder from its parts. Excludes any coders that don't
+ * contribute to the JS object.
+ */
+type CoderFromParts<T> = Coder<
+  CoderPartsOperand<FilterOut<T, LiteralCoder | PaddingCoder>>
+>;
```
eventualbuddha, line 35: This is the utility type that does the filtering of literals and padding properties.

libs/message-coder/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@votingworks/message-coder",
```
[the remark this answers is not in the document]
eventualbuddha: The other stuff @adghayes was working on could benefit from this, as could any other future direct-to-USB drivers we write in the future. One potential place that could use this right now is `ballot-encoder` where it'd replace `BitReader` and `BitWriter`.

libs/message-coder/src/fixed_string.ts
```diff
@@ -0,0 +1,103 @@
...
+  constructor(
+    private readonly byteLength: ByteLength,
+    private readonly includeTrailingNulls = false
+  ) {}
+
+  bitLength(): BitLength {
+    return toBitLength(this.byteLength);
+  }
+
+  encode(value: string): Result<Buffer, CoderError> {
```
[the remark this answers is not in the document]
eventualbuddha: This is sort of a special case as it's easier to let the `Buffer` class do the heavy lifting here. I could just duplicate the `encode` implementation in `encodeInto` if that'd be clearer.

libs/message-coder/README.md
eventualbuddha: I think that'd be fine, though there may be some benefit in keeping them the same. So that'd be `Id3v1` for both, which I'm fine with.

libs/message-coder/src/uint16_coder.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  encodeInto(
+    value: Uint16,
+    buffer: Buffer,
+    bitOffset: BitOffset
+  ): EncodeResult {
+    return mapResult(this.validateValue(value), () =>
+      this.encodeUsing(buffer, bitOffset, (byteOffset) =>
+        buffer.writeUInt16LE(value, byteOffset)
```
adghayes, line 28: FWIW, the paper handler driver sometimes takes input in `UInt16` little endian and sometime big endian, it's not consistent.

libs/message-coder/package.json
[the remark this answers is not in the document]
adghayes: If I started working on the `print-scan` work again, I don't think I would want to use this library.

The data structures I was encoding there were usually commands or simple numbers / enumerations. Using simple helpers worked fine / seems preferable to me. For example:

```js
  async setPrintingSpeed(
    printingSpeed: PrintingSpeed
  ): Promise<USBOutTransferResult> {
    return this.transferOutGeneric([
      ...SET_PRINTING_SPEED,
      PRINTING_SPEED_CODES[printingSpeed],
    ]);
  }
```

I could be convinced that the additional typing and defining structures is helpful for reading the code or being safer, but I don't expect to iterate on the drivers like other parts of the code base. So if it works, that's good enough for me. I'm less concerned about strictly and clearly defining the form of each command in code - I'll let the driver command manual do that.

On the other hand, I did copy over a good bit of code from `ballot-encoder` and there's a lot of that which could be shared.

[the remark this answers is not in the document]
adghayes: On second thought, I think there are a couple commands with hefty content where this could be helpful. So I could see using it for just those heavier endpoints.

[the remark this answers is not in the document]
adghayes: E.g. https://github.com/votingworks/vxsuite-print-scan/blob/main/services/print-scan/src/paper_handler/driver/scanner_capability.ts#L5

libs/message-coder/src/fixed_string.ts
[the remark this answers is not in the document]
jonahkagan: Ah ok gotcha. I think an explanatory comment would suffice.

libs/message-coder/README.md
```diff
@@ -0,0 +1,67 @@
...
+  header: literal('TAG'),
+  title: fixedString(30),
+  artist: fixedString(30),
+  album: fixedString(30),
+  year: fixedString(4),
+  comment: fixedString(28),
+  zeroByte: literal(0x00),
+  track: uint8(),
+  genre: uint8<Genre>(Genre),
+});
```
jonahkagan, line 41: Is the encoded data dependent on the order of fields in the message object? Is that important?

eventualbuddha:
> FWIW, the paper handler driver sometimes takes input in `UInt16` little endian and sometime big endian, it's not consistent. 

Good to know. We can update it in the future to support that use case.

libs/message-coder/README.md
eventualbuddha: Yes, the order is important. I added a JSDoc comment to `message` that explains it in more detail.

libs/message-coder/src/base_coder.ts
```diff
@@ -0,0 +1,45 @@
...
+  abstract encodeInto(
+    value: T,
+    buffer: Buffer,
+    bitOffset: BitOffset
+  ): EncodeResult;
+  abstract decodeFrom(buffer: Buffer, bitOffset: BitOffset): DecodeResult<T>;
+
+  encode(value: T): Result<Buffer, CoderError> {
+    const buffer = Buffer.alloc(toByteLength(this.bitLength(value)));
+    return mapResult(this.encodeInto(value, buffer, 0), () => buffer);
```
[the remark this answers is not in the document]
eventualbuddha: Replaced with your version.

libs/message-coder/src/fixed_string.ts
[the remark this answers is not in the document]
eventualbuddha: Comment added.

libs/message-coder/src/uint16_coder.ts
eventualbuddha: Good to know. We can add support for big endian as needed.

libs/message-coder/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": ["plugin:vx/recommended"],
+  "rules": {
+    "no-bitwise": "off"
```
adghayes, line 4: 🥳

libs/message-coder/README.md
jonahkagan: Ah ok awesome, that fills in a key gap for my understanding of this whole library. My default mental model treats objects as unordered maps, and I've trained myself to _not_ rely on field order in JS. Do you think I should be updating my mental model or is this a special case?

libs/message-coder/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "license": "AGPL-3.0",
+  "author": "VotingWorks Eng <eng@voting.works>",
+  "main": "build/index.js",
+  "scripts": {
+    "build": "tsc --build tsconfig.build.json",
+    "clean": "rm -rf build tsconfig.tsbuildinfo tsconfig.build.tsbuildinfo",
+    "lint": "pnpm type-check && eslint .",
+    "lint:fix": "pnpm type-check && eslint . --fix",
+    "pre-commit": "lint-staged",
+    "test": "is-ci test:ci test:watch",
```
adghayes, line 15: Unrelated to the content of this PR, I have been confused about this series of commands. `is-ci test:ci test:watch`. It seems like three different commands slammed together, how is this syntax working?

libs/message-coder/src/types.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Descriptive alias for `number`.
+ */
+export type ByteLength = number;
+
+/**
+ * Descriptive alias for `number`.
+ */
+export type Uint2 = number;
```
adghayes, line 36: Why did you not list the possible values for the smaller types, as you did in the `ballot-encoder` libs `bits` helpers?

Not listing them is great as far as I'm concerned because it avoids type assertions that convolute the code.

libs/message-coder/src/types.ts
```diff
@@ -0,0 +1,153 @@
...
+import { Buffer } from 'buffer';
+
+/**
+ * Errors that can occur when encoding or decoding.
+ */
+export enum CoderError {
+  SmallBuffer = 'SmallBuffer',
+  InvalidValue = 'InvalidValue',
+  UnsupportedOffset = 'UnsupportedOffset',
+}
```
adghayes, line 11: Nit, but I've really come to prefer string unions over enums. Easier to type and doesn't require imports. I don't care what it is for this PR or this specific case, but if there are reasons you explicitly prefer `enum` here, what is it? I think we need to outline some best practices internally.

libs/message-coder/src/bits.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  expect(toByteLength(4)).toEqual(1);
+  expect(toByteLength(7)).toEqual(1);
+  expect(toByteLength(8)).toEqual(1);
+  expect(toByteLength(9)).toEqual(2);
+
+  fc.assert(
+    fc.property(fc.integer(0, 0xffff), (value) => {
+      expect(toByteLength(value)).toEqual(Math.ceil(value / 8));
+    })
+  );
```
adghayes, line 25: What are the advantages of `fast-check` and why do you use it in this library?

libs/message-coder/src/message_coder.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  // encode a tag
+  const tag: id3v1 = {
+    title: 'The Title',
+    artist: 'The Artist',
+    album: 'The Album',
+    year: '2000',
+    comment: 'The Comment',
+    track: 1,
+    genre: 2,
+  };
```
adghayes, line 96: Correct me if I'm wrong - the logic in `message_coder.ts` means that the order of the message data does _not_ matter, it will be included in the encoded message in the correct order regardless.

libs/message-coder/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "test:ci": "jest --ci --coverage --reporters=default --reporters=jest-junit --maxWorkers=7",
+    "test:coverage": "jest --coverage",
+    "test:watch": "TZ=UTC jest --watch",
+    "type-check": "tsc --build"
+  },
+  "dependencies": {
+    "@votingworks/types": "workspace:*",
+    "@votingworks/utils": "workspace:*",
```
adghayes, line 23: I think you're only using `@votingworks/utils` for `typedAs`. I don't think @jonahkagan currently has that moved in #2940 but I think it would make sense to move it in #2940. CC @jonahkagan.

libs/message-coder/src/types.ts
```diff
@@ -0,0 +1,153 @@
+import { isResult, ok, Result } from '@votingworks/types';
+import { Buffer } from 'buffer';
+
+/**
+ * Errors that can occur when encoding or decoding.
+ */
+export enum CoderError {
```
adghayes, line 7: If you wanted coder errors all you had to do was call.

adghayes (commented):
@eventualbuddha LGTM, I think this is great abstraction and I'm looking forward to having it at my disposal in `print-scan` work.

Before I approve - what are the big `node-custom/data` files doing in this PR? I don't see them being used as fixtures anywhere.

libs/message-coder/README.md
adghayes: I think this is a special case.

When you later use a `MessageCoder` to _encode_ data, the order of that data does not matter. It conforms to your mental model, and I think one of the main benefits of the abstraction that @eventualbuddha has created here is that you can use the normal mental model when encoding and not worry about order.

USB commands are sent with strictly ordered data, however, and we're just using object notation to represent ordered data. JS allows us to do that because the order in which key-value pairs are defined is respected in browsers and `node`, although looking online and testing in my REPL I see there is an edge cases with numeric keys. I don't think we need to worry about it, but for reference:
```
> const x = { hello: 3, world: 6, 5: 7 }
undefined
> x
{ '5': 7, hello: 3, world: 6 }
```

libs/message-coder/package.json
jonahkagan: Sure, I can do that. I erred on the side of not including anything I was unsure about in libs/basics on first pass, but this seems reasonable. Feel free to request other things you think should be moved as well

libs/message-coder/src/types.ts
eventualbuddha: Mostly I felt they were not adding to the safety of the code much and were cumbersome to work with, so I just made them all `number`.

adghayes: I think Typescript lacks the ability to narrow such a type effectively. For example, `export type UInt1 = 0 | 1`. If I filter on `x >= 0 && x < 2` the linter still won't accept it as `UInt1`. But if typescript ever gets there, may be worth it!

libs/message-coder/src/types.ts
eventualbuddha: I think they're fine for string enums. For numeric enums I don't think they work so well. For example ([playground](https://www.typescriptlang.org/play?target=7#code/PQ18ZXAUCAEBXAzgSwHYHN4AMCm6iAtjvAGYD2ATvIUXlagMbwBuAhgDaJ7LwhxoQ4RBgwCxeACVeABwrpkeAKJUq1AMIUAJnngBvGPH4AqE0eMn4AdXVZy1IuwAu8CmXjOAFnvaKA7gzk6kSePvDITH7oDAB08ABC7NrwfsiBVLEW8CawxgBiAPJSALIAggAqAPrKUlLF8AC88AAMADRixsBm2VYAylHoMTRpGXgpAEQAkugcnKgpTBRETujaE1nGOXnwUwByAGplADJTACJVGoUl5XtnTa0AHgAc7Z2m5lv9g8OpAQzjeDTWZcBbwABSFAARrszhtejt9kdThdwYUElVzg8Wi8AIwdAC+YjIiHQTGcqAU8FkjHQzlU6ioAAolroAFzSOQKJQMzQ6PAASg5rAoYMMxiWigonDwsU4FEwTIA5LV6lI2Uq2pzkPJFCo1HzdABtVl4AC6AoA3DAiYJ4D4qHopvBBkglNrdTyDVQtLpPBR4Jg8K5vHp5shXO5qRRkGgoTK2FweMgYDSMPTvUyZDrufrGb7ZUVSpUanVilaxGIRNXq3bVoguLgJCQHDQ6AxmInuLx+LAa-3oGJJRGPbneT7+QAmB7ij69Gx2bCUKhOSMeUN-dJBMghMJ6SLROKJZKbjKbSw7Ivlaqq4oct4WbqfSzwAaHkb-R2TGZzMFLFZ+Os57bBYSInOclzXLcZz3i8D5dD0Xyvj8QSjAC34gvMKSQjC5zwl8iKHOBqLopiMFPM8+I2qkfDDs41owM4ACesh6Nmnp5oaeDTs0TLMaxUbsWO3oFpOApGgA1ngTFRvxeCCVyerjqJZoMSSZIUlSaZ0uOk4svyHJCUpIlTkKbCiiks4yq42ioDqnDsExHIRrS2AAD5IGseBkBg4wMcY6SoM4TBePA+m6AKBjZC67DukZXr5lOsRXiWt7qtFxi2fZjkPEqKU3mWUhKtaWxbFCjrsBJ-lbFEcWKQlXGTrEYEopBNxlHcbIZfAWWyA5TG5S1EFXO1dzFd15V4JV1USrFbH1ZxE66E1Q0kRi5xdaVmV2X1OXNEqq1VGi61nONW3wJN00WLaErctKsryoqKqFRqWq9f1Fa2ggzoRqgnCcPa7CsHozgBig805sZiXLf6gbBnu8Dhmu0axqg8Z6HMyYdAgyABs6gxKq4QMWX88CoEQ8hUK4qAeM6-h+ETiCg-+fXBngqa0hmjJ6fFi2iclxTXqWaoVkAA)):

```ts
/////////////////////////////////////
// using `enum` for numeric values //
/////////////////////////////////////

enum ResponseErrorCode {
  /**
   * Wrong format of the answer from the scanner. Bad answer.
   */
  FORMAT_ERROR = 0,

  /**
   * Scanner answered "Invalid command".
   */
  INVALID_COMMAND = 0x80,

  /**
   * Scanner answered "Invalid Job ID".
   */
  INVALID_JOB_ID = 0x81,
}

function printError(code: ResponseErrorCode): void {
  console.log('ERROR:', ResponseErrorCode[code]);
}

// here I can use ResponseErrorCode to get the list of possible values
printError(ResponseErrorCode.FORMAT_ERROR);



//////////////////////////////////////
// manual `enum` for numeric values //
//////////////////////////////////////

const ResponseErrorCode2 = {
  /**
   * Wrong format of the answer from the scanner. Bad answer.
   */
  FORMAT_ERROR: 0,

  /**
   * Scanner answered "Invalid command".
   */
  INVALID_COMMAND: 0x80,

  /**
   * Scanner answered "Invalid Job ID".
   */
  INVALID_JOB_ID: 0x81,
} as const;

type ResponseErrorCode2 = (typeof ResponseErrorCode2)[keyof typeof ResponseErrorCode2];

function printError2(code: ResponseErrorCode2): void {
  let display: string | undefined;

  switch (code) {
    case ResponseErrorCode2.FORMAT_ERROR:
      display = 'FORMAT_ERROR';
      break;

    case ResponseErrorCode2.INVALID_COMMAND:
      display = 'INVALID_COMMAND';
      break;

    case ResponseErrorCode2.INVALID_JOB_ID:
      display = 'INVALID_JOB_ID';
      break;
  }

  console.log('ERROR:', display);
}

// I still have to use ResponseErrorCode2 to get the list of possible values,
// so I can't avoid an import if I want autocomplete
printError2(ResponseErrorCode2.FORMAT_ERROR);
```

The manual version retains most of the benefits of the built-in `enum`, but lacks the reverse mapping from value to key.

libs/message-coder/package.json
eventualbuddha: I think it would make sense for it to be in `@votingworks/basics`, yeah.

libs/message-coder/src/message_coder.test.ts
eventualbuddha: The order of the values in a decoded message does not matter, that's true. What matters is the order of the properties when you define the message coder.

libs/message-coder/src/bits.test.ts
eventualbuddha: It generates a bunch of examples to check and, if it finds a counterexample, will do the work to reduce it to the simplest failing example it can. It helps check the edge cases that you might forget to try.

libs/message-coder/package.json
jonahkagan: Done in #2940

libs/message-coder/src/types.ts
adghayes: When numeric values are involved, I think an enum makes most sense.

In this particular case, I am thinking that instead of:
```
export enum CoderError {
  SmallBuffer = 'SmallBuffer',
  InvalidValue = 'InvalidValue',
  UnsupportedOffset = 'UnsupportedOffset',
}
```
You could have:
```
export type CoderError = 'SmallBuffer' | 'InvalidValue' | 'UnsupportedOffset'
```
At one point the typescript linting and autocomplete didn't support this fully, but now it does.

libs/message-coder/package.json
eventualbuddha: [`is-ci` takes the names of `npm` scripts as the first two arguments](https://github.com/YellowKirby/is-ci-cli/blob/6a11a14bec65dcbbb6ff52fdb468827157ee1be3/cli.js#L25), then calls `npm run …` with the appropriate one.

eventualbuddha:
> Before I approve - what are the big `node-custom/data` files doing in this PR? I don't see them being used as fixtures anywhere.

Those were accidentally added. Removed now!

adghayes (approved):
Looks like there's an build error in `lib/message-coder` but once fixed LGTM!

libs/message-coder/package.json
eventualbuddha: Updated to use `@votingworks/basics`.

libs/message-coder/src/types.ts
eventualbuddha: Switched to the string union style for this particular enum.

libs/message-coder/README.md
eventualbuddha: Per [this SO post](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order), ES3 did have unordered `Object` properties. In ES2015 that changed to make them be ordered, but with the exception @adghayes pointed out.

PR #2938, opened 2023-01-26
Remove `contest.section`

### LINKED ISSUE #2923: Remove contest.section from VXF election def

adghayes:
## Overview
Closes #2923. Removes `contest.section` from the VotingWorks election definition to bring the format closer to the CDF for easier transition. In most places, `contest.section` is simply replaced by `contest.district.name`. There are a few places where we were checking whether `contest.section === contest.title` and hiding the `section` if so. We are no longer doing that, which will affect the appearance of some reports. Should we want to hide `district.name` on those reports, we should have a clearer and more explicit setting / option for that.


apps/vx-scan/backend/src/vx_interpreter.ts
```diff
@@ -223,11 +223,9 @@ export class Interpreter {
     if (!this.skipElectionHashCheck) {
       const actualElectionHash =
         decodeElectionHash(detectQrcodeResult.qrcode.data) ?? 'not found';
-      const expectedElectionHash = this.electionDefinition.electionHash.slice(
-        0,
-        ELECTION_HASH_LENGTH
+      const expectedElectionHash = sliceElectionHash(
+        this.electionDefinition.electionHash
       );
```
adghayes, line 228: Adding this point in allows easy mocking.

frontends/bsd/src/screens/ballot_eject_screen.test.tsx
```diff
@@ -9,6 +9,14 @@ import { act } from 'react-dom/test-utils';
...
 import { BallotEjectScreen } from './ballot_eject_screen';
 
+jest.mock('@votingworks/types', () => {
+  return {
+    ...jest.requireActual('@votingworks/types'),
+    // mock election hash so snapshots don't change with every change to the election definition
+    getDisplayElectionHash: () => '0000000000',
+  };
+});
+
```
adghayes, line 19: ### Election Hash Testing Strategy 5
Calculate the 10 character display hash in a helper function form `libs/ui` and mock it so that it does not change in every snapshot.

frontends/election-manager/src/components/hand_marked_paper_ballot.tsx
```diff
@@ -998,9 +1010,9 @@ export function HandMarkedPaperBallot({
               <Contest
                 key={contest.id}
                 density={layoutDensity}
-                section={dualPhraseWithSlash(
-                  contest.section,
-                  localeContestsById?.[contest.id]?.section,
+                districtName={dualPhraseWithSlash(
+                  getContestDistrictName(election, contest),
+                  localeDistrictsById?.[contest.districtId]?.name,
```
adghayes, line 1015: I maintained this dual phrase stuff but not sure if we want to keep it in future PRs @jonahkagan.

frontends/election-manager/src/screens/write_ins_screen.tsx
```diff
@@ -188,13 +189,8 @@ export function WriteInsScreen(): JSX.Element {
...
                       <Text as="span" muted={!hasWriteIns}>
-                        {contest.section === contest.title ? (
-                          contest.title
-                        ) : (
-                          <React.Fragment>
-                            {contest.section}, <strong>{contest.title}</strong>
-                          </React.Fragment>
-                        )}
+                        {getContestDistrictName(election, contest)},{' '}
+                        <strong>{contest.title}</strong>
```
adghayes, line 193: One place where we would normally exclude showing `contest.section` if it matched `contest.title` if they were the same. 

That flexibility was nice but it seems like a poor and opaque pattern, so better to remove it and figure something else out later if we want it.

frontends/election-manager/src/screens/write_ins_transcription_screen.tsx
```diff
@@ -281,11 +282,7 @@ export function WriteInsTranscriptionScreen({
                 <br />
                 <strong>{adjudicationId.substring(0, 4)}</strong>
               </TranscriptionId>
-              <Text>
-                {contest.section !== contest.title
-                  ? contest.section
-                  : 'Contest'}
-              </Text>
+              <Text>{getContestDistrictName(election, contest)}</Text>
```
adghayes, line 285: Another place with the "hide of section equals title" logic.

integration-testing/election-manager/cypress/e2e/configuration.cy.ts
```diff
@@ -20,7 +20,6 @@ describe('Election Manager and Module Converter MS SEMS configuration', () => {
       { force: true }
     );
     cy.contains('Special Election for Senate 15');
-    cy.contains('0d610ab44c');
```
adghayes, line 23: ### Election Hash Testing Strategy 6
Don't test it! I don't quite know what the value in testing it is, when the data of the election is tested elsewhere in the UI. Simply removed.

libs/ballot-interpreter-nh/src/convert.test.ts
```diff
@@ -62,8 +62,8 @@ test('converting the Hudson ballot', async () => {
     issues: [],
   });
 
-  expect(JSON.stringify(convertResult.election, null, 2)).toEqual(
-    electionGridLayoutNewHampshireHudsonFixtures.electionDefinition.electionData
+  expect(convertResult.election).toMatchObject(
+    electionGridLayoutNewHampshireHudsonFixtures.electionDefinition.election
```
adghayes, line 66: Some testing issues were also due to spacing differences, which can be resolved by not comparing strings but rather comparing objects.

libs/ballot-interpreter-vx/test/helpers/replace_snapshot_values.ts
```diff
@@ -0,0 +1,59 @@
+// TODO: Import this helper from `@votingworks/test-utils`. There is some
+// dependency issue currently preventing this.
```
adghayes, line 2: See #2945. I wanted to unblock myself. Will add a reference to the issue into the code.

libs/test-utils/src/replace_snapshot_values.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Takes a plain object and returns the same object with the specified key-value
+ * pairs replaced. Useful for stripping out values from snapshots that would. Acts
+ * recursively on the object. Useful for updating matchers on values that might
+ * change often and do not need to be checked exactly.
+ *
+ * @param entity The object in which to replace values
+ * @param replacements A dictionary of keys to replace with their respective values
+ */
+export function replaceSnapshotValues(
+  entity: unknown[],
```
adghayes, line 19: I couldn't get this to work using `PlainObject`, will take another look.

libs/types/src/schema.test.ts
```diff
@@ -800,7 +800,7 @@ test('safeParseElectionDefinition computes the election hash', () => {
   expect(
     t.safeParseElectionDefinition(electionData).unsafeUnwrap().electionHash
   ).toMatchInlineSnapshot(
-    `"d5366378eeccc2fd38953e6e34c3069dea0dca4b7a8f5c789f3d108dc1807d3c"`
+    `"89d5b1e06c888634c318593eda29331adb9a74011178cdd7c9555cca06838e7d"`
```
adghayes, line 803: In some places where we're actually computing the election hash, we should expect it to change and include that in our testing.

libs/ui/src/contest_tally.tsx
```diff
@@ -319,9 +320,7 @@ export function ContestTally({
         return (
           <Contest key={`div-${contest.id}`} dim={!talliesRelevant}>
             <Prose maxWidth={false} data-testid={`results-table-${contest.id}`}>
-              {contest.section !== contest.title && (
-                <Text small>{contest.section}</Text>
-              )}
+              <Text small>{getContestDistrictName(election, contest)}</Text>
               <h3>
```
adghayes, line 324: This is probably the biggest UX change in the whole PR. Now for all definitions, a district name subtitle will appear above every contest tally in a report. If we want to hide that, we should probably have a better pattern than checking if `contest.section === contest.title`.

services/converter-ms-sems/tests/test_SEMSinput.py
```diff
@@ -28,23 +31,23 @@
...
+
         result = main(main_file, candmap_file)
 
         expected_result_file = open(expected_file, "r")
         expected_result = expected_result_file.read()
 
-        assert result.strip() == expected_result.strip()
-
-    
+        assert json.loads(result) == json.loads(expected_result)
```
adghayes, line 53: Test on objects, not strings, which it seems like `Python` makes easy?

services/scan/src/central_scanner_app.test.ts
```diff
@@ -329,8 +330,13 @@ test('PUT /config/package', async () => {
...
     )
     .expect(200, { status: 'ok' });
+
+  const ballotPackage = await readBallotPackageFromBuffer(
+    electionFamousNames2021Fixtures.ballotPackage.asBuffer()
+  );
+
   expect(importer.configure).toBeCalledWith(
-    electionFamousNames2021Fixtures.electionDefinition
+    expect.objectContaining(ballotPackage.electionDefinition)
```
adghayes, line 339: ### Election Hash Testing Strategy 8
Compare against the election definition in the fixture ballot package, not the separate election definition, as the two have now diverged. 

@jonahkagan I'm wondering why the election parser isn't failing though? Maybe it doesn't care about extraneous fields. I can replace this fixture if you recommend, but I'd like to avoid replacing it for all upcoming changes.

services/scan/src/interpreter.test.ts
```diff
@@ -299,9 +301,603 @@ test('interprets marks on an upside-down HMPB', async () => {
     ).interpretation as InterpretedHmpbPage
   ).toMatchInlineSnapshot(`
     Object {
-      "actualElectionHash": "602c9b551d08a348c3e1",
-      "expectedElectionHash": "965aa0b918b9bab9a2a4",
-      "type": "InvalidElectionHashPage",
```
adghayes, line 304: This seems like a snapshot that got changed inaccurately / accidentally in a past PR.

services/converter-ms-sems/tests/test_SEMSinput.py
jonahkagan: Makes sense to me, unless there's some reason we were intentionally testing string equality (e.g. because we want it to hash consistently?)

services/scan/src/central_scanner_app.test.ts
jonahkagan: Hmm, I'm not totally sure I grok this at first glance. Where would you expect a failure?

apps/vx-scan/backend/src/vx_interpreter.test.ts
```diff
@@ -294,16 +305,19 @@ test('interprets marks on an upside-down HMPB', async () => {
...
-    ).interpretation as InterpretedHmpbPage
+    replaceSnapshotValues(
+      (
+        await interpreter.interpretFile({
+          ballotImagePath,
+          detectQrcodeResult: await detectQrcodeInFilePath(ballotImagePath),
+        })
+      ).interpretation as InterpretedHmpbPage,
+      { expectedElectionHash: expect.anything() }
+    )
```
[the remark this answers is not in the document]
jonahkagan: My other concern with `replaceSnapshotValues` is that it will enable us to rely on snapshots more, rather than dissuading us from using snapshots (which I think is a better tendency overall)

services/scan/src/central_scanner_app.test.ts
adghayes: The election definition within the ballot package still has `contest.section` in it. So one might think that on parsing that election definition, there would be a `Zod` error because it has fields not in the schema. But I've never been sure if `Zod` cares about extraneous fields.

frontends/election-manager/src/components/export_ballot_pdfs_button.test.tsx
```diff
@@ -85,7 +93,7 @@ test('modal happy path flow works', async () => {
   expect(window.kiosk.makeDirectory).toHaveBeenCalledTimes(1);
   expect(window.kiosk.writeFile).toHaveBeenCalledTimes(1);
   expect(window.kiosk.writeFile).toHaveBeenCalledWith(
-    'fake mount point/ballot-pdfs/ballot-pdfs-election-b1e83122e8-live.zip'
+    'fake mount point/ballot-pdfs/ballot-pdfs-election-0000000000-live.zip'
```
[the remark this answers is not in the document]
adghayes: That's an easy one

services/scan/src/central_scanner_app.test.ts
jonahkagan: > By default Zod object schemas strip out unrecognized keys during parsing. You can disallow unknown keys with .strict() . If there are any unknown keys in the input, Zod will throw an error.

https://github.com/colinhacks/zod#strict

apps/vx-scan/backend/src/vx_interpreter.test.ts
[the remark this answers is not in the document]
adghayes: In this case it may be easier, but not in all cases. I am not always sure what the test is really covering, so I can't convert snapshots to assertions. I will replace snapshots in this PR with assertions where easier, but I don't think I can fully remove the need for `replaceSnapshotValues`. I agree with you that we should lean away from snapshots but I'm not concerned `replaceSnapshotValues` will make that harder.

frontends/bsd/src/screens/ballot_eject_screen.test.tsx
adghayes: Removed here.

libs/ballot-interpreter-nh/src/convert.test.ts
adghayes: There is a comment above this:

```ts  
  // uncomment this to update the fixture
  // require('fs').writeFileSync(
  //   require('path').join(
  //     __dirname,
  //     '../../fixtures/data/electionGridLayoutNewHampshireHudson/election.json'
  //   ),
  //   JSON.stringify(convertResult.election, null, 2),
  //   'utf8'
  // );
```

I think it is simply assuming that you update the fixture as such, when I actually updated the fixture manually (so `prettier` applied).

Serialization does affect election hashes, though, because when we parse election definitions we're not stripping out whitespace or new lines.

adghayes: @eventualbuddha is there a reason we don't strip out spacing when we get the `electionData` or when we compute the `electionHash`?

services/converter-ms-sems/tests/test_SEMSinput.py
adghayes: I don't know... @eventualbuddha? :)

apps/vx-mark/frontend/src/pages/print_page.test.tsx
```diff
@@ -10,6 +10,17 @@ import {
...
+  '@votingworks/ballot-encoder',
+  (): typeof import('@votingworks/ballot-encoder') => {
+    return {
+      ...jest.requireActual('@votingworks/ballot-encoder'),
+      // mock encoded ballot so BMD ballot QR code does not change with every change to election definition
+      encodeBallot: () => new Uint8Array(),
+    };
+  }
+);
+
```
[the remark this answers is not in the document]
adghayes: Good catch, we are losing some information. We could regain testing two by making assertions against the arguments passed to `encodeBallot`. I'll try that/

I don't think we want to go as far as decoding the QR code. It's generated by a dependency and decoding it in a test (from SVG, which we don't have a good way of doing) seems like more work than its worth to sanity check a dependency. I think we should just trust that the QR code encodes data correctly. And trust the `ballot-encoder` lib and its testing that values are being encoded to bits correctly.

jonahkagan (approved):
Thanks for investing in all these improvements! Conditionally approving once the outstanding questions about those couple of tests are resolved

services/converter-ms-sems/tests/test_SEMSinput.py
eventualbuddha: I'm unsure, but I'm guessing we didn't want it to produce different strings given the same input because that would lead to different hashing as @jonahkagan says.

libs/ballot-interpreter-nh/src/convert.test.ts
eventualbuddha: I think this gets back to the canonicalization problem that @benadida talks about. The claim is that it's hard to canonicalize JSON (e.g. by stripping whitespace) in a way that is consistent across implementations. On the other hand, it's relatively easy to compute the same SHA256 hash from the same UTF-8 bytes. I think this is another instance of wanting to ensure that we always produce the same data with the same hash given the same input.

benadida: It's more than convenience, there are so many bugs that can occur when you're relying on authenticity of a data structure that assumes canonicalization. Consider things like duplicate keys in a JSON string, does that get canonicalized in the same way as it is interpreted by the app logic?

Really important to always have the original bytes of any file that requires authentication (which is going to be all the inputs and outputs.)

adghayes: OK, thanks for the background! This explains to me why we don't try to canonicalize within the app. Rather, we deal in the original bytes of the file.

For the two tests being flagged here, however, I'm still under the impression that I don't need to test on the exact formatting of the outputted election definitions, as strings. One test is in `ballot-interpreter-nh`'s `convert` module and one is in `converter-ms-sems`. These modules output an election definition in bytes, and then the application and our system use it from then onward as unchanging bytes. What does matter is that the app does not try to canonicalize, it just takes the file as is. The exact spacing and formatting of that output file does not matter as long as it can be parsed. Am I missing something?

Issue #2743, opened 2022-10-25
Contrast settings


mattroe:
- High Contrast (default): Black on White
- High Contrast: Cyan on Black (Save yellow to communicate “warning”) 
- Low Contrast: White on Dark Gray (#757575)

mattroe:
closing this issue as the main work is done it's all now just polish/cleanup

PR #2939, opened 2023-01-26
[libs/ui] Add theming support to AppBase and use in all apps

kofi-q:
## Overview
Related to: https://github.com/votingworks/vxsuite/issues/2743

Adding scaffolding for UI themes to libs/ui:
- Added a styled-components `ThemeProvider` to the recently added `AppBase` component
- `AppBase` now accepts `colorMode` and `sizeMode` props which are used to set the appropriate theme. These default to `"legacy"`, which will preserve the current existing styles for the shared components.
- Wrapping all `<AppRoot>` components in the various apps with `<AppBase>` to enable support for themed components.
  -  This also allows us to centralize the global stylesheets and remove the mostly duplicated `App.css` files in each app package.

The themes don't do much at the moment - will be fleshing out the styles in follow-up PRs. This just allows us to update teh shared components in smaller PR chunks without breaking any existing UIs.

### Apps after wrapping in `<AppBase>`:

Screenshot 2023-01-25 at 18 16 25 Screenshot 2023-01-25 at 18 15 26 Screenshot 2023-01-25 at 18 05 13 Screenshot 2023-01-25 at 18 04 06 Screenshot 2023-01-25 at 17 16 56

### BMD with barebones `"contrastHighDark"` theme:
Screenshot 2023-01-26 at 10 55 00

## Testing Plan 
- Tested all apps with and without the new `<AppBase>` wrapper to make sure the global styles were equivalent.
- Added unit tests for the new code paths within libs/ui

## Checklist


kofi-q:
FYI - had to rebase with the VxMark refactor in main, so things have moved around a bit!

libs/types/src/index.ts
```diff
@@ -16,5 +16,6 @@ export * from './printing';
 export * from './result';
 export * from './smartcard_auth';
 export * from './tallies';
+export * from './ui_theme';
```
adghayes, line 19: Any particular reason you put this in `libs/types`? It's nice to move types there but it all seems UI dependent. Just curious.

libs/ui/src/global_styles.tsx
```diff
@@ -1,16 +1,141 @@
...
+ * TODO: Copied from old App.css files in the frontend packages - could probably
+ * use some cleanup and de-duping with the normalize.css styles we're already
+ * importing.
+ *
+ * TODO: Hardcode base64-encoded versions of our font files and reference here,
+ * so that everything's centralized and we don't have to have duplicate
+ * copies in each app's package.
  */
-export const GlobalStyles = createGlobalStyle`
+export const GlobalStyles = createGlobalStyle<GlobalStlyesProps>`
```
adghayes, line 21: Love this centralization. I noticed a typo here - `GlobalStylyesProps`. Do you have spell-checker enabled in your environment? That would have at least notified you.

libs/ui/src/global_styles.tsx
```diff
@@ -1,16 +1,141 @@
...
+  }
+  @media print {
+    html {
+      background: #ffffff;
+
+      /* Adjust printed ballot font-size */
+      /* stylelint-disable-next-line declaration-no-important */
+      font-size: ${(p) => p.legacyPrintFontSizePx ?? 16}px !important;
+    }
+  }
```
adghayes, line 49: I wonder if this will break styles on our printing. I suggest you take a document from one of the apps and print (in Chrome) a copy from `main` and a copy from this branch. 

Printed documents are rendered outside of the app as an appended child to the document body, see `/libs/ui/src/print_element.tsx`. Previously, the `App.css` file would be applied everywhere in the body. But my understanding is that these styles will now only apply within the apps. 

If that's the case, I think that's good, but we'll have to move some of these print related styles into the `PrintOnly` styled component wrapping in `libs/ui/src/print_element.tsx`. That is at least one way of doing it. Passing in the legacy base font sizes, I'm not sure how best to do that. It will be outside of the `ThemeProvider` so we can't snatch it from the theme. Ideas?

And, if I'm misunderstanding this, and these styles still apply to the printed stuff, great.

frontends/bas/package.json
```diff
@@ -15,8 +15,8 @@
     "lint:fix": "pnpm type-check && eslint . --fix && pnpm stylelint:run:fix",
     "start": "pnpm -w run-dev bas",
     "start:core": "pnpm -w run-dev bas --core-only",
-    "stylelint:run": "stylelint 'src/**/*.{js,jsx,ts,tsx}' && stylelint 'src/**/*.css' --config .stylelintrc-css.js",
-    "stylelint:run:fix": "stylelint 'src/**/*.{js,jsx,ts,tsx}' --fix && stylelint 'src/**/*.css' --config .stylelintrc-css.js --fix",
+    "stylelint:run": "stylelint 'src/**/*.{js,jsx,ts,tsx}'",
+    "stylelint:run:fix": "stylelint 'src/**/*.{js,jsx,ts,tsx}' --fix",
```
adghayes, line 19: I would also remove the `.stylelintrc-css.cs` files.

adghayes (commented):
@kofi-q, exciting to see the styling confusing get consolidated. We need to validate printing and possibly add some data passing for print styles before we merge this.

frontends/bas/package.json
adghayes: FYI (not important to this PR, but you may hit it) is that the `--fix` option for `stylelint` is broken in `css-in-js`. There's not a good fix in the ecosystem yet, but it looks like one is evolving: https://github.com/styled-components/styled-components/issues/3607#issuecomment-1374344911. I'm tracking that thread and once something looks solid I'll make a ticket for anyone to complete.

kofi-q: Ah, didn't know about the stylelint issue - thanks for tracking!

Also, good call re: .stylelintrc-css files

libs/ui/src/global_styles.tsx
kofi-q: Yeah, it's a good nudge - the `createGlobalStyles` feature hoists all the global styles to the `<head>` section of the html file, so it should apply to everything in the document, in theory, including the print content.

I'll verify with a few test prints, just in case - forgot to test that path!

Love that we already have the `PrintOnly` component and the other print helpers! Was contemplating adding something like that when I noticed the print only styles in `App.css` - will see if I can move stuff over to use the component while I work through the design refresh.

libs/ui/src/global_styles.tsx
kofi-q: Bah, thanks! You know, I thought I did have a spell checker enabled - need to make sure that's working properly

libs/types/src/index.ts
kofi-q: No solid reason - I haven't fully developed an intuition around our code organization WRT types, but thought there could be a case where we want to reference theme types or the color enum from a package that we don't want depending on libs/ui (e.g. in the unlikely case, we reference a color or theme from an election config or something similar)

adghayes (approved):
LGTM, I didn't know what `createGlobalStyles` was doing.

kofi-q:
Verified that changing stuff in `GlobalStyles` affects the print layout, so it's looking good - found a typo in the print font size for election-manager, though. Fixed that and double-checked that I had the right font size set for all the apps 🚀

PR #2940, opened 2023-01-26
refactor(libs): Extract `basics` lib

jonahkagan:
## Overview
Pull out generic utilities from libs/utils and libs/types that aren't VX-specific into their own library to help avoid dependency cycles.

## Checklist


libs/basics/README.md
```diff
@@ -0,0 +1,10 @@
+# basics
+
+An extended TS standard library of generic utility functions and types.
+
+This library should only include utilities that are not specific to VotingWorks
+products, but would be generally applicable to any TypeScript project (similar
+to a library like lodash). In other words, it should be considered as an
+extension of the standard TS libraries. This means it shouldn't have very many
```
eventualbuddha, line 8: Maybe even mandate that it has no workspace dependencies so that everyone can import it without fear of cycles?

eventualbuddha (commented):
LGTM so far.

libs/basics/README.md
jonahkagan: Good call, updated

.circleci/config.yml
```diff
@@ -904,6 +926,7 @@ workflows:
       - test-libs-ballot-encoder
       - test-libs-ballot-interpreter-nh
       - test-libs-ballot-interpreter-vx
+      - test-libs-basics
```
adghayes, line 929: Make sure to have @carolinemodic add this to the required tests on Github after merging.

libs/basics/package.json
```diff
@@ -0,0 +1,60 @@
...
+    ],
+    "*.+(js|jsx|ts|tsx)": [
+      "eslint --quiet --fix"
+    ],
+    "package.json": [
+      "sort-package-json"
+    ]
+  },
+  "dependencies": {
+    "util": "^0.12.4"
```
adghayes, line 34: This is only used in `setupTests` AFAICT. Is there any reason we couldn't move it to `devDependencies` it, and make one principle of this repo that there are no `dependencies`?

adghayes (approved):
@jonahkagan I reviewed what you included in the switch and what is left in `libs/utils` and I agree with your erring on the side of not moving things to `libs/basics`, and didn't notice anything else I think should move. I do think `typedAs` would be best to live in `libs/basics` but I'm approving this PR in case it is easier for you or to do it in a follow-up.

libs/basics/package.json
jonahkagan: It's also used in result.ts unfortunately

PR #2942, opened 2023-01-26
Custom A4 Scanner driver library

eventualbuddha:
## Overview

WIP implementation of a driver for the Custom A4 scanner. Been testing with the `./bin/demo` script.

### Known Issues
- [ ] only side A scanning works; side A/B just hangs waiting for more data
- [ ] tests are essentially non-existent
- [ ] locking is probably not done quite right
- [ ] if the scanner stalls we don't really have the ability to recover without a power cycle. the `node-usb` `clearHalt` implementation throws an exception

## Demo Video or Screenshot
TBD

## Testing Plan 
TBD

## Checklist


jonahkagan (commented):
Some initial questions

libs/node-custom/src/cli/demo/index.ts
```diff
@@ -0,0 +1,99 @@
...
+          `LL ${SensorStatus[status.sensorInputLeftLeft]}, ` +
+          `CL ${SensorStatus[status.sensorInputCenterLeft]}, ` +
+          `CR ${SensorStatus[status.sensorInputCenterRight]}, ` +
+          `RR ${SensorStatus[status.sensorInputRightRight]}`
+      );
+
+      if (status.isTicketOnEnterA4 && !status.isMotorOn) {
+        watcher.stop();
+        console.log('Scanning…');
+        const scanResult = await scanner.scan(ScanSide.A_AND_B);
```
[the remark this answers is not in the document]
eventualbuddha: I have some (probably buggy) locking in place that should prevent such race conditions. What should be happening, I think, is that the `Scanner#scan` promise will only resolve once the full scan is complete. There will be many underlying commands sent to the scanner in the process, but nothing callable from `Scanner`'s public API should be able to interfere with the scan in progress. So yes, the status does have `isScanInProgress` but the current API never allows that to be `true`.

Based on your experience with Plustek, is this desirable?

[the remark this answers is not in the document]
jonahkagan: I'm not sure. With Plustek, there were times where the `scan` command hung indefinitely and there was no way to interrupt it. That made me wish that all of the commands were synchronous/immediate and that I could control all of the status checking and error handling based on whatever physical effect happened as a result of the command.

That being said, since we have lower level control here, I think the abstractions that you're building could still give us that flexibility.

Thinking about the context in which this code will be used in our apps, I can say that it's useful to always be able to query for the current status of the scanner (or at least be notified on changes).

Some of this probably comes down to where we want the state-machine-esque logic to live. Is it in the driver or the application code? One lesson I took away from Plustek was that grouping as much of that together as possible was useful (whereas with most code I would instead try to build multiple layers of abstraction).

Random collection of thoughts, hope some of that is useful!

libs/node-custom/README.md
```diff
@@ -0,0 +1,23 @@
+# node-custom
```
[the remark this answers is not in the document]
adghayes: Separate but related thought about naming - there may end up being more than one "Custom" driver in the monorepo, so I'd suggest the title be more specific to the product (A4 scanner).

libs/node-custom/pnpm-lock.yaml
```diff
@@ -0,0 +1,298 @@
+lockfileVersion: 5.3
```
adghayes, line 1: Is there a reason this lib has its own `pnpm-lock.yaml`?

libs/node-custom/src/lock.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * A simple lock implementation.
+ */
+export class Lock {
+  private locked = false;
+  private readonly resolvers: Array<() => void> = [];
```
adghayes, line 6: I created something similar to this for `print-scan` so I think it would be useful to have in `utils` or `basics`.

libs/node-custom/src/usb_device.ts
```diff
@@ -0,0 +1,265 @@
...
+    [read: number, write: number],
+    ErrorCode
+  > {
+    try {
+      const readEndpointNumber =
+        this.device.configurations[0]?.interfaces[INTERFACE_INDEX]?.alternate
+          .endpoints[READ_ENDPOINT_INDEX]?.endpointNumber;
+      const writeEndpointNumber =
+        this.device.configurations[0]?.interfaces[INTERFACE_INDEX]?.alternate
+          .endpoints[WRITE_ENDPOINT_INDEX]?.endpointNumber;
```
adghayes, line 244: Are these endpoint numbers not constant? Why not hardcode them?

libs/node-custom/src/mutex.ts
```diff
@@ -0,0 +1,30 @@
+import { Optional } from '@votingworks/types';
+
+/**
+ * A simple mutex implementation.
+ */
+export class Mutex {
+  private locked = false;
```
adghayes, line 7: Is there a distinction between `Mutex` and `Lock`?

libs/node-custom/pnpm-lock.yaml
eventualbuddha: It shouldn't. Will delete.

libs/node-custom/src/mutex.ts
eventualbuddha: This is sort of what I meant in the description how the locking is messed up. It's kind of in transition right now. `Lock` waits when `acquire` is called, `Mutex` just fails if it's busy. The latter emulates how the official Custom driver works. I haven't decided which approach to take.

libs/node-custom/README.md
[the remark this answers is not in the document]
eventualbuddha: Ended up going with `custom-scanner`. 👍

libs/node-custom/src/cli/demo/index.ts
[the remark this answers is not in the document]
eventualbuddha: `scan` will wait and block other calls until it's done. This isn't true of all the others: `move` does not wait until the move has finished. I'll make sure the JS Doc comments explain this.

libs/node-custom/src/cli/demo/index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        if (scanResult.isErr()) {
```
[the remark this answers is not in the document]
eventualbuddha: The alignment is better both because the status information seems to be more consistent and because we can better control how all this happens.

libs/node-custom/src/usb_device.ts
eventualbuddha: I've generally tried to give names to the magic numbers in our codebase so we know what they're for.

eventualbuddha:
Superseded by https://github.com/votingworks/vxsuite/pull/2973.

Issue #2946, opened 2023-01-30
Stop Using `election.json` Fixtures outside of `libs/fixtures`


adghayes:
## Summary

We currently have a lot of older fixtures in places like `bmd`. The fewer fixtures we can have across the code base the better. We'd like to remove them from the following:

- [x] `apps/vx-mark/frontend`
- [ ] `libs/ballot-interpreter-vx`
- [x] `frontends/bas`
- [x] `frontends/bsd`
- [x] `frontends/election-manager`
- [ ] `integration-testing/bsd`
- [ ] `apps/vx-scan/backend`?
- [ ] `services/scan`?
- [ ] `converter-ms-sems`?

The SEMS converter may be a special case, and the scan services may be thornier and partially represented by #1524. The scan libs have the extra challenge of image fixtures, which we currently have no means to generate as needed. We can only generate ballots via a frontend, and can only generate ballots with specific features (stretched, low mark threshold) literally manually.

## Requirements

Despite the list above, I would consider this closed if we checked off all the boxes above without question marks.

## Non-Requirements

List items above with question marks.

## Tasks

See list above.

adghayes:
Closing this due to inactivity.

PR #2951, opened 2023-01-31
Remove bmd fixtures

adghayes:
## Overview
Removes fixtures from `apps/vx-mark/frontend` as part of #2946, instead uses fixtures only from `@votingworks/fixtures`


apps/vx-mark/frontend/src/__snapshots__/app_contest_multi_seat.test.tsx.snap
```diff
@@ -537,94 +537,6 @@ exports[`Single Seat Contest 1`] = `
...
-                    class="sc-bdvvaa eSpfPW"
-                  >
-                    <strong>
-                      Amanda Marracini
-                    </strong>
-                    <br />
-                    Labor
-                  </p>
-                </div>
-              </button>
```
adghayes, line 627: Looks like the newer version of this contest has fewer options ¯\_(ツ)_/¯

apps/vx-mark/frontend/src/app_contest_ms_either_neither.test.tsx
```diff
@@ -1,328 +0,0 @@
-import React from 'react';
```
adghayes, line 1: Just deleting this file since it will be deleted later for #2949

apps/vx-mark/frontend/jest.config.js
```diff
@@ -12,6 +12,14 @@ module.exports = {
...
     '!src/contexts/ballot_context.ts',
   ],
+  coverageThreshold: {
+    global: {
+      statements: 96,
+      branches: 87,
+      functions: 96,
+      lines: 96,
+    },
+  },
```
adghayes, line 22: Dropping the test coverage for the either neither work, should go back to 100% once you remove it @jonahkagan and you can remove this.

adghayes:
@jonahkagan removing VxMark fixtures was pretty straightforward. VxMark had three closely related definitions:

- `electionSample` (seal url)
- `electionSampleWithSeal` (seal svg)
- `electionSampleNoSeal`

And `libs/fixtures` had two:
- `electionSample` (seal svg)
- `electionSampleNoSeal`

So, there is no longer a fixture with a seal url. The main place this matters is in `libs/ui` for the BMD ballot component, and it is tested there. The definition used that has a seal url is... `electionWithMsEitherNeither` :). This seal is a great example of a place where we'd like to process a single election definition after the fact rather than have multiple.

In the case of removing the testing for this PR, however, I think there are no issues. We weren't actually testing `sealUrl` and `seal` separately in the only component that it matters, `ElectionInfo`, so no testing lost AFAICT.

apps/vx-mark/frontend/jest.config.js
jonahkagan: Will do

jonahkagan (approved):
Nice, seemed relatively painless

PR #2952, opened 2023-01-31
refactor(converter-ms-sems): Use snapshot testing

jonahkagan:
## Overview
Instead of having manually created snapshot files, use a snapshot library. Otherwise it's impossible to change any logic without having to manually recreate expected test results.

## Checklist


adghayes (approved):

services/converter-ms-sems/tests/test_SEMSoutput.py
```diff
@@ -1,49 +1,51 @@
...
+
+        TESTS = [
+            {
+                'election': get_sample_file('general-election.json'),
+                'cvrs': '10_cvrs.csv',
+                'tallies': '10_tallies.json',
+            },
+            {
+                'election': election_file.name,
+                'cvrs': '10_8-26-2020-cvrs.txt',
```
adghayes, line 31: Where are we using these `cvrs` files?

jonahkagan: Uhh, that's weird. They were there before but also unused. I guess I can remove?

adghayes: If it doesn't break anything, I think so. The test files in are in our current CVR file format, which we'll be changing anyway, so no use keeping around.

I didn't think we convert directly from our CVRs - we convert from an overall tally when we export results - so I'm not sure what those are doing there.

jonahkagan: Maybe an artifact from an earlier version of the script. Removed in 2da763a53

PR #2957, opened 2023-02-01
Remove `bas`, Ballot Activation System

adghayes:
Per [slack discussion](https://votingworks.slack.com/archives/CEL6D3GAD/p1675209119208789) we will be removing `frontends/bas`, the Ballot Activation System. `bas` was used for activating ballots on voter cards which would then be inserted into a Ballot Marking Device. We are not currently using voter cards for ballot activation, however, so `bas` has been unused for some time. Although there are many use cases for `bas`, they are well into the future at which point it will make more sense to rewrite `bas` from scratch.


adghayes:
@carolinemodic we'll need to remove branch protection for `test-frontends-bas`.

carolinemodic (approved):
explosion

PR #2960, opened 2023-02-01
Move VxAdmin auth to the backend

arsalansufi:
_Recommend reviewing by commit_

## Overview

Issue links: https://github.com/votingworks/vxsuite/issues/2906, https://github.com/votingworks/vxsuite/issues/2900

It's happening! We're finally moving auth to the backend as per the plan laid out in this [VxSuite Backend Auth Spec](https://docs.google.com/document/d/1_iIopPJ6HhILPUzuD2hZUpizMRhcQ4vD20HBVbIv8Mc/edit?usp=sharing). This PR sets up dipped smart card auth on the backend and updates VxAdmin to use it instead of frontend auth.

## Testing

- [x] Updated test suites
- [x] Manually tested VxAdmin auth and card programming, including edge case scenarios (e.g. when there's no election def)

## Checklist


libs/ui/src/unlock_machine_screen.tsx
```diff
@@ -39,43 +43,67 @@ const EnteredCode = styled.div`
...
+      setCurrentPasscode(passcode);
+      if (passcode.length === SECURITY_PIN_LENGTH) {
+        if ('checkPasscode' in auth) {
+          auth.checkPasscode(passcode);
+        } else {
+          assert(checkPinFromProps !== undefined);
+          await checkPinFromProps(passcode);
+        }
+        setCurrentPasscode('');
+      }
```
arsalansufi, line 79: Your favorite kind of change @jonahkagan! Moved PIN checking logic out of a `useEffect` and into the relevant event handler

Actually had to do this to address a bug where the PIN was being checked twice, the second check resulting in an error on the backend because we were no longer in the `CheckingPin` state

libs/ui/src/hooks/auth/use_dipped_smart_card_auth.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  useEffect(() => {
+    async function logOut() {
+      await authApiClient.logOut();
+      setHasBeenReset(true);
+    }
+    void logOut();
+    // Log out once on mount to avoid inconsistencies between the backend and the frontend when the
+    // frontend restarts and the backend does not (e.g. during frontend development)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
arsalansufi, line 28: I was initially super excited about the fact that we'd no longer have to auth over and over every time the page refreshed during frontend development since auth state is now managed on the backend and only cleared when the backend restarts.

But I ultimately ran into various (development-specific) errors when the frontend restarted but the backend didn't. Here's an example that actually came up in Cypress integration tests. If you auth as an election manager (and don't log out) and then open an incognito window to simulate Cypress starting a new test in a new browser, you'll hit an assertion error [here](https://github.com/votingworks/vxsuite/blob/77bd5851ed42380c7937aad34f022ab908354ab7/frontends/election-manager/src/screens/ballot_list_screen.tsx#L41) because you're still authenticated on the backend but an election definition hasn't been defined yet in the new browser.

Still thinking through a few more edge cases but I've thus far found logging out on mount to be a simple way to keep things in sync

libs/ui/src/hooks/auth/use_dipped_smart_card_auth.ts
```diff
@@ -0,0 +1,44 @@
+import useInterval from 'use-interval';
+import { DippedSmartCardAuth } from '@votingworks/types';
+import { useEffect, useState } from 'react';
+
+const AUTH_STATUS_POLLING_INTERVAL = 100;
```
arsalansufi, line 5: I lowered the polling interval from 200 to 100 since we temporarily have two layers of polling: VxAdmin frontend ➡️ VxAdmin backend, and VxAdmin backend ➡️ card reader. And if they're offset in just the right way, the times can add up.

In my JavaCard implementation of backend auth, I'm hoping to avoid the second layer (VxAdmin backend ➡️ card reader) by using listeners

frontends/election-manager/src/app_root.tsx
```diff
@@ -218,6 +213,10 @@ export function AppRoot({
     [logger, currentUserRole, clearCastVoteRecordFilesMutation, store]
   );
 
+  if (!currentElection.isSuccess) {
+    return null;
+  }
```
arsalansufi, line 218: This actually fixed one of those "frontend restarted but backend didn't" bugs I referenced earlier. But unfortunately didn't fix them all

frontends/election-manager/test/helpers/api.ts
```diff
@@ -0,0 +1,28 @@
+import type { Api } from '@votingworks/admin'; // eslint-disable-line vx/gts-no-import-export-type
+import { createMockClient, MockClient } from '@votingworks/grout-test-utils';
+import { DippedSmartCardAuth } from '@votingworks/types';
+
+export type MockApiClient = Omit<MockClient<Api>, 'getAuthStatus'> & {
+  // Because this is polled so frequently, we opt for a standard jest mock instead of a
+  // libs/test-utils mock since the latter requires every call to be explicitly mocked
```
arsalansufi, line 7: One limitation of the new mocking approach - not a big deal and easy enough to work around!

frontends/election-manager/src/components/navigation_screen.tsx
```diff
@@ -111,7 +113,7 @@ export function NavigationScreen({
             {(isSystemAdministratorAuth(auth) ||
               isElectionManagerAuth(auth)) && (
               <React.Fragment>
-                <Button onPress={() => auth.logOut()} small>
+                <Button onPress={() => apiClient.logOut()} small>
```
arsalansufi, line 116: I know I'm not yet handling potential event handler errors as discussed during your presentation @jonahkagan, but I wanted to keep this PR from growing too much in scope

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
```diff
@@ -102,7 +97,7 @@ export function CardDetailsView({
       role,
       status: 'InProgress',
     });
-    const result = await card.unprogramUser();
+    const result = await apiClient.unprogramCard();
```
arsalansufi, line 100: I did explore the idea that we discussed in the spec @jonahkagan of updating the methods to take an `authStatus` of a specific type, e.g. `SystemAdministratorLoggedIn`, to ensure that we're in the right state before making calls like this. But I found the resulting code to be complicated and for little gain

libs/utils/src/Card/web_service_card.ts
```diff
@@ -24,12 +24,21 @@ interface SuccessIndicationResponse {
...
   /**
    * Reads basic information about the card, including whether one is present,
    * what its short value is and whether it has a long value.
    */
   async readSummary(): Promise<CardSummary> {
-    return unsafeParse(CardSummarySchema, await fetchJson('/card/read'));
+    return unsafeParse(
+      CardSummarySchema,
+      await fetchJson(`${this.baseUrl}/card/read`)
+    );
```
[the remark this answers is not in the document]
eventualbuddha: I think @jonahkagan used [`cross-fetch`](https://www.npmjs.com/package/cross-fetch) successfully?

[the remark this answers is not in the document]
arsalansufi: Ah sorry I misspoke! cross-fetch is what I tried using (since that’s what I saw in Grout)

libs/ui/src/hooks/auth/use_dipped_smart_card_auth.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface AuthApiClient {
+  getAuthStatus: () => Promise<DippedSmartCardAuth.AuthStatus>;
+  logOut: () => Promise<void>;
+}
+
+export function useDippedSmartCardAuth(
```
jonahkagan, line 12: Could this hook be replaced by a `useQuery`?

frontends/election-manager/src/app_root.tsx
jonahkagan: Hah, I had left a comment in response to your comment about that suggesting exactly this fix. Would it be worth talking through the remaining issues?

frontends/election-manager/src/components/navigation_screen.tsx
jonahkagan: Thanks for calling this out. I'm assuming you have a plan to revisit?

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
jonahkagan: Sounds good. Maybe we can have some runtime assertions within the auth lib to help enforce those constraints

frontends/election-manager/test/helpers/api.ts
jonahkagan: Makes sense. I've been noodling on this and I have another idea for how to work around it: In test, configure the polling interval to be something really large, then use fake timers to advance whenever we are testing something that depends on an auth status change.

But happy to try this approach for now and see how it goes. We could potentially build some sort of support for this type of mocking into the mock client if we like it.

jonahkagan: Also, I'd be curious to hear how things went with the new mocking approach. Did you run into any issues? Did it make anything easier?

libs/ui/src/hooks/auth/use_dipped_smart_card_auth.ts
jonahkagan: And in general, are you planning to use `useMutation` for the auth methods?

jonahkagan (commented):
Looks great! I don't have any major issues/questions, but a few small things to think through.

libs/utils/src/Card/web_service_card.ts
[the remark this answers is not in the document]
arsalansufi: > I'm assuming the use of `fetchJson` is temporary while we phase out the smartcard service

Correct!

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -0,0 +1,26 @@
+import { DippedSmartCardAuth, ElectionDefinition } from '@votingworks/types';
+
+/**
+ * The API for a dipped smart card auth instance, "dipped" meaning that the card needs to be
+ * inserted and removed from the card reader to complete authentication
+ */
+export interface DippedSmartCardAuthApi {
+  getAuthStatus: () => DippedSmartCardAuth.AuthStatus;
```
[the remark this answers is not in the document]
arsalansufi: The layers 😆

services/admin/src/index.ts
```diff
@@ -1,7 +1,37 @@
...
+// Load environment variables from .env* files. Suppress warnings using silent
+// if this file is missing. dotenv will never modify any environment variables
+// that have already been set.  Variable expansion is supported in .env files.
+// https://github.com/motdotla/dotenv
+// https://github.com/motdotla/dotenv-expand
+for (const dotenvFile of dotenvFiles) {
+  if (fs.existsSync(dotenvFile)) {
+    dotenvExpand.expand(dotenv.config({ path: dotenvFile }));
+  }
+}
```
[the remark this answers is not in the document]
arsalansufi: This crossed my mind as well. Went ahead and opened an issue: https://github.com/votingworks/vxsuite/issues/2964

frontends/election-manager/src/components/navigation_screen.tsx
arsalansufi: Yes! Tracking miscellaneous cleanup items in https://github.com/votingworks/vxsuite/issues/2965

frontends/election-manager/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: Yeah that makes sense! Will be sure to include those in the `DippedSmartCardAuthWithJavaCard` and `InsertedSmartCardAuthWithJavaCard` implementations

frontends/election-manager/test/helpers/api.ts
arsalansufi: Oh nice, very reasonable alternative! I like the current approach since it requires little mental overhead and feels like a very direct way of getting to what we need for tests. But am open to experimentation down the road

arsalansufi: And in terms of the new mocking approach, I found the error messages when calls didn't match up to be very helpful as I was debugging failing tests!

libs/types/src/auth/dipped_smart_card_auth.ts
```diff
@@ -0,0 +1,69 @@
...
+    | PollWorkerUser['role'];
+}
+
+export type ProgramCard = (
+  input: ProgramCardInput
+) => Promise<Result<{ pin?: string }, Error>>;
+
+export type UnprogramCard = () => Promise<Result<void, Error>>;
+
+export const DefaultAuthStatus: Readonly<AuthStatus> = {
```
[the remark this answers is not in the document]
arsalansufi: Good catch re this constant looking like a type on first glance. Updated: https://github.com/votingworks/vxsuite/pull/2960/commits/d783f475c3e24dbb1bb36e71c7f77e2e26bca5f6

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  checkPin: (input: { pin: string }) => void;
+  logOut: () => void;
+
+  programCard: DippedSmartCardAuth.ProgramCard;
+  unprogramCard: DippedSmartCardAuth.UnprogramCard;
```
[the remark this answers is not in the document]
arsalansufi: Ah you know what, in an earlier iteration of this code, I needed the `ProgramCard` and `UnprogramCard` types on the frontend, so I moved them from `libs/auth` to `libs/types`. That's no longer the case, so they can be moved back to `libs/auth`: https://github.com/votingworks/vxsuite/pull/2960/commits/c2b0873d06cab2ba69911a8789be6dfb8940c633

libs/ui/src/unlock_machine_screen.tsx
```diff
@@ -39,43 +43,67 @@ const EnteredCode = styled.div`
...
-    if (currentPasscode.length === SECURITY_PIN_LENGTH) {
-      auth.checkPasscode(currentPasscode);
-      setCurrentPasscode('');
+    if (isFeatureFlagEnabled(BooleanEnvironmentVariableName.SKIP_PIN_ENTRY)) {
+      void bypassPinEntry();
     }
-  }, [currentPasscode, auth]);
+    // Run this hook once on mount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
arsalansufi: Yes that's the plan! In fact, I plan on removing PINs from the user objects exposed to the frontend, so skipping PIN entry will have to move to the backend

I just can't make these changes yet since `UnlockMachineScreen` is a shared component still in use by machines with frontend auth

services/admin/src/server.ts
```diff
@@ -26,99 +33,134 @@ type NoParams = never;
...
+function buildApi(auth: DippedSmartCardAuthApi) {
+  return grout.createApi({
+    getAuthStatus: () => auth.getAuthStatus(),
+    checkPin: (input: { pin: string }) => auth.checkPin(input),
+    logOut: () => auth.logOut(),
+    programCard: (input: {
+      userRole: 'system_administrator' | 'election_manager' | 'poll_worker';
+    }) => auth.programCard(input),
+    unprogramCard: () => auth.unprogramCard(),
+  });
```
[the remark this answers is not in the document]
arsalansufi: So I initially hadn't wrapped, but as you've noted, binding (or the lack of) was an issue.

I tried binding by using fat arrow syntax, but that interfered with a lint rule:

error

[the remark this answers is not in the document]
jonahkagan: Oh well

services/admin/src/server.ts
```diff
@@ -782,7 +826,14 @@ export async function start({
   logger = new Logger(LogSource.VxAdminService),
   port = PORT,
   workspace,
-}: Partial<StartOptions>): Promise<void> {
+}: Partial<StartOptions>): Promise<Server> {
+  const auth = new DippedSmartCardAuthWithMemoryCard({
+    card: new WebServiceCard({ baseUrl: 'http://localhost:3001' }),
```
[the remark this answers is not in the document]
arsalansufi: Good q! We're in a way already hardcoding if you look at the proxies for frontend communication with the smartcards service, e.g. https://github.com/votingworks/vxsuite/blob/5cf6e3799241734a67f63b87663cd081b416ad73/frontends/election-manager/prodserver/setupProxy.js#L18

So that makes me feel confident that doing so on the backend is safe as well. From a code cleanliness perspective, a config var probably still makes sense, but this code path will be shortlived, so doesn't feel worth the trouble

frontends/election-manager/src/app_root.tsx
arsalansufi: The main issue that remains without the logout on mount is the one that I mention in https://github.com/votingworks/vxsuite/pull/2960#discussion_r1092694177

Took some time to investigate further and one of the deeper problems is that, while election(s) are persisted in the admin service, the current election ID is still persisted in frontend local storage. This explains a lot of the funky behavior that I've seen when switching between incognito windows and election definitions

Would be open to more fully sharing my findings and discussing paths forward tomorrow!

jonahkagan: Ah interesting! Yes, let's discuss tomorrow. Though also if you want to move forward it doesn't necessarily need to block, since it's only a concern for dev

libs/ui/src/hooks/auth/use_dipped_smart_card_auth.ts
arsalansufi: I honestly hadn't been planning to since I didn't see much gain from using react-query for this data. But for consistency sake and for machines where we don't have an `AppContext` available, using react-query might make sense after all

Will try it out but would also be a good thing to discuss tomorrow!

jonahkagan: Yes, agreed that there's not much use for a cache but that consistency would be a benefit. Unified error handling could also be nice. I was also under the impression that we would try to get rid of `AppContext` by the end of this

frontends/election-manager/src/app_root.tsx
arsalansufi: I thought a bit more about this and decided to change course. All of the issues that remain have to do with multiple isolated browsers backed by the same backend service - a situation that only really comes up during Cypress tests. Rather than modifying source code to support that case, I decided to just update Cypress tests to explicitly log out before starting

With that, we now do get the handy benefit of not having to re-auth when the frontend restarts during development!

https://github.com/votingworks/vxsuite/pull/2960/commits/1dd74c8b0fbc628cb6b4b8d56528c230e9108668

libs/ui/src/hooks/auth/use_dipped_smart_card_auth.ts
arsalansufi: Further thinking on this below in https://github.com/votingworks/vxsuite/pull/2960#discussion_r1092700857

libs/ui/src/hooks/auth/use_dipped_smart_card_auth.ts
arsalansufi: Realizing I wasn't doing a great job of distinguishing the short term from the long term!

> I was also under the impression that we would try to get rid of `AppContext` by the end of this.

Definitely! I still see this as a long term goal. I was just thinking about the best way to get backend auth wired up and running in the short term. That said, I don't think it's that much work to use react-query.

I had been envisioning one possible alternative long term solution where we create a more targeted `AuthContext` for auth since it is conceptually different than the rest of the react-query data (e.g. isn't persisted in the backend, is polled so doesn't require explicit invalidation on mutation, etc.). But the point about unified error handling is a compelling one in favor of just using react-query.

I think one of the subconscious things driving my hesitancy is the complexity and indirection that react-query adds. (Worth it for most data we store in it, to be clear.)

tldr: Leaning toward using react-query but a sanity check is worthwhile as we broaden the scope of what we use react-query for

frontends/election-manager/src/app_root.tsx
jonahkagan: Awesome!

libs/ui/src/hooks/auth/use_dipped_smart_card_auth.ts
jonahkagan: Yeah, I've wondered about `AuthContext` as a potential direction as well. I agree with most of your thinking here. Let's discuss and get to the bottom of it

arsalansufi: Discussed in person and we're aligned on the benefits and pain points of react-query. And despite the pain points (the key one being the clunkiness of error-handling-friendly mutations), agree that we should use react-query consistently as a wrapper around the `apiClient` ✅

frontends/election-manager/src/index.tsx
```diff
@@ -11,42 +11,52 @@ import {
...
+      errorMessage={
+        <Prose textCenter>
+          <h1>Something went wrong</h1>
+          <Text>Please restart the machine.</Text>
+        </Prose>
+      }
+    >
       <ServicesContext.Provider value={{ backend, logger, storage }}>
-        <App />
+        <ApiClientContext.Provider value={apiClient}>
```
arsalansufi, line 46: I moved the `ApiClientContext.Provider` up a level to `index.tsx` because I realized that, when it was in `App`, it was getting rendered twice in tests:

https://github.com/votingworks/vxsuite/blob/a51a57ee4f8ec5818f54b2ea1778cd9a5cc2547c/frontends/election-manager/test/render_in_app_context.tsx#L94-L102

I couldn't just remove the `ApiClientContext.Provider` from the test util because the `renderRootElement` helper is also used to render elements other than `App` in tests.

I think we'd ideally move all the wrappings in `index.tsx` into `App` to better mirror VxMark and VxScan (and update tests and test helpers accordingly), but that was out of scope for this PR

Relatedly, with the `QueryClientProvider` also in this file, I had to render the `ErrorBoundary` here, making it hard to write an error boundary test (because of the `ReactDom.render` in this file). But I did test manually

integration-testing/election-manager/cypress/support/auth.ts
```diff
@@ -43,5 +45,5 @@ export function mockCardRemoval(): void {
 }
 
 export function logOut(): void {
-  cy.request('POST', 'http://localhost:3000/api/logOut', {});
+  cy.request('POST', methodUrl('logOut', 'http://localhost:3000/api'), {});
```
arsalansufi, line 48: I know this is super rudimentary, and we'd ideally use the `Api` type to guarantee that the specified method is actually present in the API, but I wanted to time box. Made an issue for an improved version: https://github.com/votingworks/vxsuite/issues/2969

arsalansufi:

libs/ui/src/react_query.ts
```diff
@@ -0,0 +1,34 @@
+// eslint-disable-next-line vx/gts-no-import-export-type
+import type { DefaultOptions } from '@tanstack/react-query';
+
+/**
+ * Recommended default options for react-query query clients
+ */
+export const QUERY_CLIENT_DEFAULT_OPTIONS: DefaultOptions = {
```
jonahkagan, line 7: 👏

frontends/election-manager/src/index.tsx
jonahkagan: Cool, that sounds fine for now. If you wanted you could leave a TODO comment, but not necessary

frontends/election-manager/src/api.ts
```diff
@@ -19,3 +25,64 @@ export function useApiClient(): ApiClient {
...
+  },
+} as const;
+
+export const checkPin = {
+  useMutation() {
+    const apiClient = useApiClient();
+    const queryClient = useQueryClient();
+    return useMutation(apiClient.checkPin, {
+      async onSuccess() {
+        // Not strictly needed right now
```
jonahkagan, line 49: Maybe elaborate a bit more on why this isn't needed? Our future selves might be puzzled

libs/ui/src/unlock_machine_screen.tsx
```diff
@@ -73,7 +73,7 @@ export function UnlockMachineScreen({
           auth.checkPasscode(passcode);
         } else {
           assert(checkPinFromProps !== undefined);
-          await checkPinFromProps(passcode);
+          checkPinFromProps(passcode);
```
jonahkagan, line 76: Just making sure we're ok with not `await`ing here before resetting the passcode to `''`

integration-testing/election-manager/cypress/support/auth.ts
jonahkagan: Sounds good!

jonahkagan (approved):
Looks great!

frontends/election-manager/src/api.ts
arsalansufi: Oh oops! I dropped this comment here during our sync earlier and forgot to come back and clean it up 🤦‍♂️

libs/ui/src/unlock_machine_screen.tsx
arsalansufi: Yes good q - I wasn't able to detect a noticeable difference. Even when we had the await, the passcode clearing was visible before the transition to the remove card screen (because auth status doesn't update until the next re-fetch)

Side note: This is another interesting downside to our current approach to mutations - losing the ability to `await` at code points outside of the mutation `onSuccess` callback

jonahkagan: Cool, makes sense.

Re side note: yes, agreed. There are workarounds: e.g. using a `try/catch` with `mutateAsync` or  `await new Promise(resolve => mutation.mutate(input, { onSuccess: resolve })` but it's not ideal. Sadly there's just no way to have unhandled promise rejections in React event handlers work nicely.

Issue #2744, opened 2022-10-25
Text size (display)


mattroe:
Research: Add text-size selector to VxScan and measure text. 

Research: Update text-size selector on VxMark and measure text. These are different screen sizes and thus they will render text and different sizes. We may need to update how we specify base text sizes to be based upon physical units.

Implement font-size hierarchy ratios for each supported screen size:
- For smaller base text sizes, the headers can be larger.
- For the larger base font sizes, the headings will need to be relatively smaller in order to fit on the screen.

Review every component in every screen orientation to ensure that all content fits

mattroe:
closing this issue as the main work is done it's all now just polish/cleanup

PR #2961, opened 2023-02-01
[libs/ui] Add typography lib and flesh out VVSG font size themes

kofi-q:
## Overview
Related to: https://github.com/votingworks/vxsuite/issues/2744

- Adding semantic typography components (`<P>`, `<H1>`, `<H2>`, etc) that vary styles based on the selected theme.
- Attempting to simplify things a bit and reduce the surface area of what we support to just the essentials, so this makes a break from the "wrap blocks of text in `<Prose>`" approach we currently have, since that encourages a bit too much styling variability that might not play well with the themes. Should be able to do most of what we need with just these few building blocks (and we can add more semantic components or styling props later, as needed).
- VVSG 2.0 has requirements around capital letter text size being physically `n millimeters` on any given screen, which requires us to know the pixel density of the screen and the ratio of capital letter height to pixel size to make sure out sizes fall within the right ranges. Will need to tweak some of these values after testing on actual devices, but this is a decent starting point.

### On ELO 15" Screen Dimensions & Estimated Pixel Density of 1.476:
https://user-images.githubusercontent.com/264902/216123155-7a059c87-7928-47d4-8e35-46936b84aa63.mov

### Formatting Props Demo:
https://user-images.githubusercontent.com/264902/216127744-cee8641a-4d45-4343-84b6-69330045c314.mov

## Testing Plan 
- Added unit tests verifying formatting props have the desired effect and respect theme selection - not testing any specific theme values, since those are just configuration.
- Added storybook entry for visual testing/verification

## Checklist


libs/ui/src/app_base.test.tsx
```diff
@@ -98,3 +98,19 @@ test('renders with selected themes', () => {
...
+  );
+
+  expect(container).toContainHTML('<div>foo</div>');
+
+  const htmlNode = document.body.parentElement;
+  assert(htmlNode);
+  const computedStyles = window.getComputedStyle(htmlNode);
+
+  expect(computedStyles.overflow).toEqual('auto');
+});
```
adghayes, line 116: Other than meeting coverage, what is the point of this test? This is the kind of test I'm uncertain about, in terms of value.

libs/ui/src/themes/make_theme.ts
```diff
@@ -34,14 +64,123 @@ const colorThemes: Record<ColorMode, ColorTheme> = {
...
+const PIXELS_PER_INCH = 72;
+
+function mmToPx(mm: number): number {
+  // TODO: This is just a stop-gap proxy - need to add device detection logic to
+  // determine what screen we're on (ELO/VSAP/laptop) and use pre-defined
+  // ratios, since we can measure physical display dimensions more accurately.
+  const { devicePixelRatio } = window;
+
+  return mm * INCHES_PER_MM * devicePixelRatio * PIXELS_PER_INCH;
+}
```
adghayes, line 77: So my understanding is that the `mmToPx` logic is consistent and correct across devices, just being math, and the only `TODO` is to determine the `devicePixelRatio` on each device rather than relying on the `window`'s value? Is the `window` value typically not correct?

libs/ui/src/themes/make_theme.ts
```diff
@@ -34,14 +64,123 @@ const colorThemes: Record<ColorMode, ColorTheme> = {
...
+    },
+    headingsRem: {
+      h1: 2.5,
+      h2: 1.75,
+      h3: 1.5,
+      h4: 1.25,
+      h5: 1.125,
+      h6: 1,
+    },
+    letterSpacingEm: 0.01,
```
adghayes, line 108: Why headings in `rem` and letter spacing in `em`?

libs/ui/src/typography.tsx
```diff
@@ -0,0 +1,234 @@
...
+/** Styled block caption for small sub-text copy. */
+export function Caption(props: FontProps): JSX.Element {
+  return <StyledCaption {...props} />;
+}
+
+/**
+ * Styled block text for pre-formatted copy.
+ * Sanitizes and renders HTML text as HTML elements and preserves whitespace and
+ * newline formatting.
+ */
```
adghayes, line 224: Is this inspired by certain use cases within the apps?

adghayes (approved):
I'm excited to see `Prose` removed in favor of something more transparent. This seems like a good base but it's hard to tell without seeing it applied to an app. Once you start applying them to the apps I suggest you put out a PSA to the engineers because it seems like a significant change in how we do things. 

If you have the VVSG guideline numbers on hand, it'd be helpful for me to put those in the PR description so I can reference. 

I'd like to sync to understand the font size strategy (with `rem` vs `em` in the themes) but not blocking.

libs/ui/src/app_base.test.tsx
adghayes: I suppose in this case, we're confirming that a property is being passed to to global styles. That actually seems worth it to me. So I would take it slightly out of the category of tests that just confirm that passing in "green" makes the styling "green".

kofi-q: TBH, was originally driven by meeting coverage, but agreed that this sort of falls in the functional category

libs/ui/src/themes/make_theme.ts
kofi-q: That's right - for the most part, the `window.devicePixelRatio` value seemed to be accurate on my browser (it's `2` on my macbook screen and `1` when I drag the window to my external non-retina monitor), but I wasn't seeing the same thing in kiosk-browser (stayed at `1` regardless of which screen I was using). So, that didn't give me too much confidence in using that value and I want to do a bit more testing on the ELO devices

libs/ui/src/themes/make_theme.ts
kofi-q: Making the headings use `rem` so that they have consistent sizing relative to the base font size, whereas using `em` for the letter spacing makes that scale proportionally to the size of the particular text/heading being rendered, rather than having one letter spacing value for all font sizes

libs/ui/src/typography.tsx
kofi-q: Yeah, noticed there's an existing `TextWithLineBreaks` component used when rendering some pre-formatted text in election definitions (I think it's text configured by election officials)

PR #2966, opened 2023-02-02
Artisan CVR CDF

adghayes:
Artisan CVR CDF.

Sampled Current Format Row Size (no write-in): ~250 characters
Sampled New Format Row Size (no write-in): ~2500 characters (1200 w/o spaces)

Let's say it's roughly a 10x increase in the size of a CVR file, not counting write-ins. On its own, this is not enough to change our size considerations. The CVR files we export from a single scanner will never be big enough to cause problems (without write-ins). Separating write-in files is something we should do, but not because we need to because of the larger CVR file. When we export aggregated CVRs, however, then it may be an issue.


libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,375 @@
...
+              "CandidateIds": ["triggerfish"]
+            },
+            {
+              "@type": "CVR.CandidateSelection",
+              "isWriteIn": true
+            },
+            {
+              "@type": "CVR.CandidateSelection",
+              "isWriteIn": true
+            }
```
adghayes, line 117: Our system infers write in positions with `allowWriteIns` but here we explicitly list them, because there's not a similar option in CDF. Same sort of thing in Ballot Definition CDF if I remember correctly.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,375 @@
...
+      "@id": "precinct-1",
+      "Name": "Precinct 1",
+      "Type": "precinct"
+    }
+  ],
+  "CVR": [
+    {
+      "@type": "CVR.CVR",
+      "BallotStyleId": "2F",
+      "BallotStyleUnitId": "precinct-1",
```
adghayes, line 161: precinctId ➡️  BallotStyleUnitId

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,375 @@
...
+      "PartyIds": ["1"],
```
adghayes, line 162: Included but not required by the CDF, but required by VVSG, so adding here.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,375 @@
...
+      "@type": "CVR.Party",
+      "@id": "1",
+      "Name": "Fish Party",
+      "Abbreviation": "F"
+    }
+  ],
+  "Election": [
+    {
+      "@type": "CVR.Election",
+      "@id": "some-id",
```
adghayes, line 29: I don't think this `id` matters, could be anything. We don't currently have an election ID on our CVRs.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,375 @@
...
+  ],
+  "CVR": [
+    {
+      "@type": "CVR.CVR",
+      "BallotStyleId": "2F",
+      "BallotStyleUnitId": "precinct-1",
+      "PartyIds": ["1"],
+      "CreatingDeviceId": "scanner-1",
+      "ElectionId": "some-id",
+      "UniqueId": "id-168",
```
adghayes, line 165: The CDF description of this field is:

> The sequence number for this CVR. This represents the ordinal number that this CVR was processed by the tabulating device

Our actual IDs (not in these fixtures) are not serial, however, they're UUIDs. Although we'd be going outside of the description, I think what we have is a "Unique ID" and we should just use this field.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,375 @@
...
+      "vxBallotType": "absentee",
+      "vxTestBallot": true,
```
adghayes, line 167: Prefix our own fields with `vx`?

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,375 @@
...
+      "CurrentSnapshotId": "id-168-original",
+      "CVRSnapshot": [
+        {
+          "@id": "id-168-original",
+          "@type": "CVR.CVRSnapshot",
+          "Type": "original",
```
adghayes, line 173: I think that the CVRs we export from our scanning devices would all be considered "original" rather than "modified." They report the selections for overvote, even if they won't count. 

Modified = "After contest rules applied."
Original = "As scanned, no contest rules applied."

It'll keep it simpler for us if we just stick to original.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: Will just use election hash!

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,400 @@
...
+            {
+              "@type": "CVR.CandidateSelection",
+              "@id": "write-in-0",
+              "isWriteIn": true
+            },
+            {
+              "@type": "CVR.CandidateSelection",
+              "@id": "write-in-1",
+              "isWriteIn": true
+            }
```
adghayes, line 120: Write-in positions in the contest descriptions are just listed, as opposed to an `allowsWriteIns` flag.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,400 @@
...
+              "@type": "CVR.CVRContest",
+              "ContestId": "fishing",
+              "CVRContestSelection": [
+                {
+                  "@type": "CVR.CVRContestSelection",
+                  "ContestSelectionId": "yes",
+                  "SelectionPosition": [
+                    {
+                      "HasIndication": "yes",
+                      "IsAllocable": "yes",
```
adghayes, line 208: I meant to take this out but, for demo purposes, there's this `IsAllocable` field which can be used to indicate whether a mark is supposed to be counted based on contest rules. We could include this on all marks, or just where it's "no", but it's not required so I'll just leave it out.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,400 @@
...
+      "Type": "precinct"
+    }
+  ],
+  "Batch": [
+    {
+      "@type": "VX.Batch",
+      "@id": "batch-2",
+      "BatchLabel": "Batch 2"
+    }
+  ],
```
adghayes, line 166: Should be `vxBatch`, this isn't a part of the CDF spec but I think we might as well take the opportunity the format presents to normalize the batch label data instead of repeating it.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,400 @@
...
+      "BallotStyleId": "2F",
+      "BallotStyleUnitId": "precinct-1",
+      "PartyIds": ["1"],
+      "CreatingDeviceId": "scanner-1",
+      "ElectionId": "some-id",
+      "BatchId": "batch-2",
+      "UniqueId": "id-172",
+      "vxBallotType": "absentee",
+      "vxTestBallot": true,
+      "BallotImage": [
```
adghayes, line 312: This one has a `BallotImage` because there is a write-in. You can also include a `WriteInImage` next to the particular write in but this seems like a more appropriate place to include it given we include the whole image.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,400 @@
...
+        {
+          "@type": "CVR.ImageData"
+        },
```
adghayes, line 315: I'm thinking that the length of this array will always be 2, and we just won't have a link to an image if we aren't including the image.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,400 @@
...
+        {
+          "@type": "CVR.ImageData",
+          "Location": "./ballot-12312321312-back.jpeg"
```
adghayes, line 318: We'll move images to a directory structure and reference them.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,400 @@
...
+                            "y": 0,
+                            "width": 800,
+                            "height": 200
+                          },
+                          "WriteInRect": {
+                            "x": 0,
+                            "y": 0,
+                            "width": 800,
+                            "height": 100
+                          }
```
adghayes, line 380: We currently include layout information in the CVR to be able to tell where to zoom in in our transcription UI. Currently we include the entire computed layout in the CVR. We have a few options:
- Continue including the entire computed layout and put it here
- Continue including the entire computed layout and put a file reference to it here
- Only include the layout relevant to the contest

The last option is what I lean towards because it's the most specific and compact, even though it's a bit of extra work.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: cc @jonahkagan @carolinemodic

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: After discussing with @mattroe, I'm sure "original" is correct as we're not applying contest rules to determine selections (without straight party voting0.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: cc @jonahkagan @carolinemodic @eventualbuddha

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
carolinemodic: prefixing with vx seems reasonable to me

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
carolinemodic: agree with the choice here

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
carolinemodic: I like just moving the logic of this to the scanners instead of VxAdmin, it feels crazy in VxAdmin to be trying to figure out the coordinates

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
jonahkagan: Sequence number is useful in RLAs. Are the UUIDs ordered based on tabulation order? If so, then that's fine. If not, then I might replace this with an incrementing counter of some sort.

jonahkagan: Oh wait, but I see there's `BatchSequenceId` for that... Very confusing

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
jonahkagan: Yeah, let's pick a convention and make it consistent across all our CDF implementations

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: Yes, I think there are some issues with the documentation around here, a couple descriptions that look more copy pasted than written.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
jonahkagan: What are "contest rules"?

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: I think on the precinct scanner we don't _want_ a sequence ID because that would compromise voter privacy cc @mattroe. On central scanner i'll probably have a follow-up task to make it work.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
jonahkagan: Also, are we later going to include adjudication information for adjudicated ballots when exporting final CVRs from VxAdmin?

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
jonahkagan: What would be the second item in the array in that case?

jonahkagan: Is the idea to use the array index to indicate front vs back? That seems reasonable

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
jonahkagan: It looks like the CDF spec intends for us to just include an image of the write-in here directly. Why customize to include layout info instead of just doing that? Is the goal to reduce the number of images and thus the size of the CVR?

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
mattroe: Yeah, my perspective was sequence ID on central scanned batches, but no ID on precinct scanned batches for that reason unless someone is thinking differently. VVSG says "if applicable" so I assume that is to allow to not include it where it would be a privacy violation.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: For our transcription UX, we use the image of the entire ballot. It's possible that the ballot has multiple write-ins, so I think we should include the image of the ballot at the top-level to and avoid duplication. 

We could include an image of the write-in, but we also would need to include an image of the contest because we use that in our UX as well. So, without changing our UX, we have to somehow add something custom.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: Yeah!

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
mattroe: actually very sensible

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
mattroe: @jonahkagan see the context around 1.1.5-F here: https://docs.google.com/document/d/15B-LjdB-i5JhcVBi6b_GxmBtke6NYiBiMhoLzisIwPA/edit

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
jonahkagan: Gotcha, that makes sense.

A couple more thoughts:
- The contest layout info will be duplicated for each write-in. Not a huge deal but could be normalized.
- Are there times that we would need layout info for anything other than adjudicating write-ins? Are there other issues that might require adjudication? If so, it might make sense to include layout info on the `CVRContest` object and on the `CVRContestSelection` object.

None of these are strong recommendations, just poking and exploring options

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
jonahkagan: I'm just not clear about how overvotes and undervotes factor in. I thought those could potentially fall under "contest rules". If so, I would interpret the CDF as follows:
- original: what bubbles the scanner believes the voter bubbled
- modified: the logical interpreted "vote" based on the bubbles
- adjudicated: any changes to that logical interpretation based on adjudication

So when I originally assessed this, I thought our current CVRs looked more like "modified" than "original." But my interpretation of all this could be wrong, I don't have much context

mattroe: My understanding on the categorization is the same as @jonahkagan's. Section 3.3 in the spec clarifies this pretty well too.

After reading the spec more carefully, I think the intention is that under/overvote logic is a contest/election rule after all. The CDF spec repeatedly mentions that this only applies to HMPB because DREs prevent marking outside of contest/election rules. Section 2.2 says that scanners apply interpretation regularly based on election rules, and cites overvote & undervote rules as examples of those cases.

When talking with @adghayes today, I was under the impression that we had an "original" representation in our CVR today, but I didn't double check that.

Ultimately, what I would want to see in the system is a CVR that is as auditable as possible: where the original ballot marks are in the record, any interpretation done by the scanner is specified, and the results of human adjudication are also eventually identifiable.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
eventualbuddha: We include the layout information per CVR because the location of these things in each image is not unique due to rotation, stretching, cropping, etc. from the scanning process.

We could choose to normalize images like with currently do with VX ballots. This would mean that we can use a layout per ballot style, but it has a downside. We either have to interpret using the normalized image (not sure this is a good idea), ship both the normalized and original images with the CVRs, or only use original images with per-ballot layout information.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: @jonahkagan I've been confused about the same thing.

If our CVRs didn't include the overvoted selections, then I would think it's modified. But our CVRs do contain all the overvoted selections, so I now think of it as the "original." Although we are doing the logical interpretation for tabulation on VxScan, the CVR does not reflect that.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: OK, proposed convention:
1. Added fields are prefixed with lowercase `vx` - `vxBallotType`, `vxTestBallot`, `vxBatch`
2. Added "classes" are `CVR.vxBatch` or `BallotDefinition.vxWhatever`.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: I think our CVRs are "original" because, in an overvote case, we do include all of those options. Even though we do the logical interpretation for tabulation for a precinct scanner report, the CVR is pre-logical interpretation. So I think we're good here, but have been confused as well.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
```diff
@@ -0,0 +1,400 @@
...
+            },
+            {
+              "@type": "CVR.CandidateSelection",
+              "@id": "triggerfish",
+              "CandidateIds": ["triggerfish"]
+            },
+            {
+              "@type": "CVR.CandidateSelection",
+              "@id": "write-in-0",
+              "isWriteIn": true
```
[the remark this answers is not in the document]
adghayes: I think we can decide, it's an optional field. I would opt to not have it since I presume undefined is false here.

libs/fixtures/data/electionMinimalExhaustiveSample/cvrFiles/cdf_cast_vote_record_report.json
adghayes: @eventualbuddha I completely understand the rationale.

But looking at the existing code, it looks to me like we're _not_ using the calculated layout and we're using the ballot style layout. Here's where we pass layouts to the cast vote record, and we're querying the ballot style layout metadata saved in the DB:
https://github.com/votingworks/vxsuite/blob/a487926b23ee5dd30ce6e9078927ec230151586e/apps/scan/backend/src/cvrs/export.ts#L122-131

If we _are_ using the calculated layouts, then could you explain what I'm missing. If we actually _aren't_, why is that and does it matter to start using them?

Issue #184, opened 2021-01-23
remove ballot locale from CVR


benadida:
as per VVSG 1.1 Section 3.2.3

It's a privacy risk if a particular language is used only rarely.

PR #2968, opened 2023-02-02
Remove "_locales" from CVR

adghayes:
## Overview
Closes https://github.com/votingworks/vxsuite/issues/184. Removes ballot locale from CVR as it's a privacy risk and as part of cleanup before CVR CDF.

Here we are _only_ removing `_locales` from the CVRs. The `locales` field is still on `BallotMetadata` and it being used to key layouts on the scan backends and when they are exported / created in VxAdmin. I am wondering if, in a follow-up PR, we want to remove `BallotLocales` entirely from the scan and VxAdmin side? We want to support multi-lingual ballots but my experience is that it's not the best abstraction and we'll want to start over.


eventualbuddha (approved):
Seems reasonable.

PR #2974, opened 2023-02-03
add bootstrap directive to Makefile

amcmanus:
Our toplevel bootstrap script (`script/bootstrap`) expects each `Makefile` to include a `bootstrap` directive. This adds that to the vx-mark app backend.

### DIFF
--- a/apps/vx-mark/backend/Makefile
+++ b/apps/vx-mark/backend/Makefile
@@ -4,5 +4,7 @@ FORCE:
 build: FORCE
 	pnpm install && pnpm build
 
+bootstrap: build
+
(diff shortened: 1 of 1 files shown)


carolinemodic (approved):
Oh this reminds me I never updated complete-system to work with this new structure...

amcmanus:
> Oh this reminds me I never updated complete-system to work with this new structure...

Unfortunately, we're not consistent across these repos. In complete-system, the expectation is `make install`. I can update this commit to include that as well if you'd like.

carolinemodic (approved):

Issue #2971, opened 2023-02-02
Address "Election definition must be configured first" warnings in VxAdmin frontend tests


arsalansufi:
These are happeing in `src/components/import_cvrfiles_modal.test.tsx` and `src/screens/manual_data_import_index_screen.test.tsx`: 

```
PASS  src/components/import_cvrfiles_modal.test.tsx
  ● Console

    console.error
      Error: Election definition must be configured first
          at ElectionManagerStoreMemoryBackend.getCvrFiles (/home/circleci/project/frontends/election-manager/src/lib/backends/memory_backend.ts:184:13)
          at Object.queryFn (/home/circleci/project/frontends/election-manager/src/hooks/use_cvr_files_query.ts:23:56)
          at Object.fetchFn [as fn] (/home/circleci/project/node_modules/.pnpm/@tanstack+react-query@4.2.3_w7o5yyljkiidx2s2nzb26ottzu/node_modules/@tanstack/query-core/build/esm/index.js:1037:27)
          at run (/home/circleci/project/node_modules/.pnpm/@tanstack+react-query@4.2.3_w7o5yyljkiidx2s2nzb26ottzu/node_modules/@tanstack/query-core/build/esm/index.js:607:31)
          at createRetryer (/home/circleci/project/node_modules/.pnpm/@tanstack+react-query@4.2.3_w7o5yyljkiidx2s2nzb26ottzu/node_modules/@tanstack/query-core/build/esm/index.js:653:5)
          at Query.fetch (/home/circleci/project/node_modules/.pnpm/@tanstack+react-query@4.2.3_w7o5yyljkiidx2s2nzb26ottzu/node_modules/@tanstack/query-core/build/esm/index.js:1092:20)
          at QueryObserver.executeFetch (/home/circleci/project/node_modules/.pnpm/@tanstack+react-query@4.2.3_w7o5yyljkiidx2s2nzb26ottzu/node_modules/@tanstack/query-core/build/esm/index.js:2305:37)
          at QueryObserver.onSubscribe (/home/circleci/project/node_modules/.pnpm/@tanstack+react-query@4.2.3_w7o5yyljkiidx2s2nzb26ottzu/node_modules/@tanstack/query-core/build/esm/index.js:2169:14)
[...]

Issue #2965, opened 2023-02-01
Miscellaneous cleanup


arsalansufi:
A catch-all issue for miscellaneous cleanup items as I work on Auth v2 implementation. Will move items from here to standalone issues as needed.

- [x] Handle potential event handler errors (https://github.com/votingworks/vxsuite/pull/2960#discussion_r1092702301)
- [x] Use the same query client settings in the real VxAdmin app and VxAdmin tests (https://github.com/votingworks/vxsuite/pull/2975#discussion_r1096317253)
- [ ] Handle failures to save scanner report data to card (https://github.com/votingworks/vxsuite/pull/2986#discussion_r1099317079)
- [ ] Reset the backend between Cypress tests (https://github.com/votingworks/vxsuite/pull/3001#discussion_r1102105332)
- [ ] Change "smartcard" to "smart card" in copy
- [ ] Revisit auth error handling holistically and introduce a consistent pattern for auth checks via an auth decorator (https://github.com/votingworks/vxsuite/pull/3011#discussion_r1108885410)

arsalansufi:
And done:

- https://github.com/votingworks/vxsuite/issues/3678
- https://github.com/votingworks/vxsuite/issues/3679

Closing this issue out accordingly

PR #2975, opened 2023-02-03
Move VxCentralScan auth to the backend

arsalansufi:
_Recommend reviewing by commit_

## Overview

Issue links: https://github.com/votingworks/vxsuite/issues/2901

This PR moves VxCentralScan auth to the backend, using the backend dipped smart card auth logic set up in https://github.com/votingworks/vxsuite/pull/2960.

## Testing

- [x] Updated test suites
- [x] Manually tested VxCentralScan auth

## Checklist

### DIFF
--- a/frontends/election-manager/test/render_in_app_context.tsx
+++ b/frontends/election-manager/test/render_in_app_context.tsx
@@ -71,8 +71,8 @@ interface RenderInAppContextParams {
   hasPrinterAttached?: boolean;
   logger?: Logger;
   backend?: ElectionManagerStoreBackend;
-  queryClient?: QueryClient;
   apiClient?: ApiClient;
+  queryClient?: QueryClient;
 }
 
 export function renderRootElement(
@@ -81,14 +81,16 @@ export function renderRootElement(
     backend = new ElectionManagerStoreMemoryBackend(),
     logger = fakeLogger(),
     storage = new MemoryStorage(),
-    queryClient = new QueryClient(),
     apiClient = createMockApiClient(),
+    // TODO: Determine why tests fail when using createQueryClient and, by extension,
+    // QUERY_CLIENT_DEFAULT_OPTIONS
+    queryClient = new QueryClient(),
   }: {
     backend?: ElectionManagerStoreBackend;
     logger?: Logger;
     storage?: Storage;
-    queryClient?: QueryClient;
     apiClient?: ApiClient;
+    queryClient?: QueryClient;
   } = {}
 ): RenderResult {
   return testRender(
@@ -132,8 +134,8 @@ export function renderInAppContext(
     hasPrinterAttached = true,
     logger = new Logger(LogSource.VxAdminFrontend),
     backend,
-    queryClient,
     apiClient,
+    queryClient,
   }: RenderInAppContextParams = {}
 ): RenderResult {
   return renderRootElement(
--- a/services/admin/test/utils.ts
+++ b/services/admin/test/utils.ts
@@ -1,10 +1,6 @@
 import { Admin } from '@votingworks/api';
 import { CandidateContest } from '@votingworks/types';
-import {
-  DippedSmartCardAuthApi,
-  DippedSmartCardAuthWithMemoryCard,
-} from '@votingworks/auth';
-import { MemoryCard as MockCard } from '@votingworks/utils';
+import { DippedSmartCardAuthApi } from '@votingworks/auth';
 
 /**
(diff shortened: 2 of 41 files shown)


services/admin/test/utils.ts
```diff
@@ -26,18 +22,16 @@ export function buildOfficialCandidatesWriteInAdjudicationOptionGroup(
...
+export function buildMockAuth(): DippedSmartCardAuthApi {
+  return {
+    getAuthStatus: jest.fn(),
+    checkPin: jest.fn(),
+    logOut: jest.fn(),
+    programCard: jest.fn(),
+    unprogramCard: jest.fn(),
+    setElectionDefinition: jest.fn(),
+    clearElectionDefinition: jest.fn(),
+  };
```
arsalansufi, line 36: Realized that I could just set up a fully mocked auth instance instead of a real auth instance with a mocked card reader. No need to test the internals of the auth lib here (and in every other Vx machine) when it's going to test itself. And might as well make it easier for us to mock auth states, especially as we start checking auth on the backend. We still have integration tests to test the system holistically.

What initially got me to revisit this was a warning in the test output:

```
 worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.
```

This was due to the polling within the auth lib. I was initially ideating ways to turn off the polling in these tests when I realized that I didn't need the internals of the auth lib at all here.

frontends/election-manager/test/render_in_app_context.tsx
```diff
@@ -81,14 +81,16 @@ export function renderRootElement(
     backend = new ElectionManagerStoreMemoryBackend(),
     logger = fakeLogger(),
     storage = new MemoryStorage(),
-    queryClient = new QueryClient(),
     apiClient = createMockApiClient(),
+    // TODO: Determine why tests fail when using createQueryClient and, by extension,
+    // QUERY_CLIENT_DEFAULT_OPTIONS
```
arsalansufi, line 86: One of the issues seems to be the error boundary creating a hard failure out of this otherwise silent (test-specific, as far as I can tell) error: https://github.com/votingworks/vxsuite/issues/2971. But there's definitely more going on because disabling the error boundary setting didn't address all the failures. I did verify that these issues predate the move to backend auth

services/admin/test/utils.ts
jonahkagan: Nice, that sounds good. I almost suggested this change in your first PR but I didn't have a compelling reason

frontends/election-manager/test/render_in_app_context.tsx
jonahkagan: Hmm, interesting. I would definitely be wary of removing the default options completely in test, because they strongly influence the behavior of react-query and it's caching, which could potentially change the logic of the app. Maybe I'm being paranoid and you've thought it through though

services/scan/tsconfig.build.json
```diff
@@ -10,6 +10,7 @@
   },
   "references": [
     { "path": "../../libs/api/tsconfig.build.json" },
+    { "path": "../../libs/auth/tsconfig.build.json" },
```
jonahkagan, line 13: Add `declarationMap: true`?

jonahkagan (approved):

frontends/election-manager/test/render_in_app_context.tsx
arsalansufi: Yes totally fair! One thing worth noting is that this discrepancy between the query client in the real app vs. in tests has been around since before Auth v2 work. It was just less explicit / maybe even accidental, so I wanted to call it out now.

_index.tsx before Auth v2_

https://github.com/votingworks/vxsuite/blob/6460a1e8480fb5500a73024324c9c4b05f63e5b8/frontends/election-manager/src/index.tsx#L18-L29

_render_in_app_context.tsx before Auth v2_

https://github.com/votingworks/vxsuite/blob/43d00b66766b56e63ac39c3b9c4066b35a0efeef/frontends/election-manager/test/render_in_app_context.tsx#L81

What has changed since Auth v2 is the addition of settings, like the error boundary setting, which does increase risk. I've added this to my miscellaneous cleanup ticket so I don't lose track of it: https://github.com/votingworks/vxsuite/issues/2965

jonahkagan: Ah gotcha. Thanks for tracking!

PR #2978, opened 2023-02-06
Fix env loading for vxmark

carolinemodic:
Realized I forgot to update the env variable loading for the root vxsuite file to account for the extra nested folder


arsalansufi (approved):
Good catch!

PR #2982, opened 2023-02-06
fix subtle issue with setup dev script

carolinemodic:
When the setup-dev script runs even if node doesn't need to be updated we should still run corepack enable in case things like pnpm need updates. VxDev relies on this script for keeping things up to date.

### DIFF
--- a/script/setup-dev
+++ b/script/setup-dev
@@ -22,37 +22,36 @@ NODE_VERSION=$(cat "${DIR}/../.node-version")
 
 if command node -v 2>/dev/null | grep -q "${NODE_VERSION}"; then
   echo "Node version ${NODE_VERSION} already installed"
-  sudo corepack enable
-  exit
+else 
+
+  echo "Installing Node version ${NODE_VERSION}"
+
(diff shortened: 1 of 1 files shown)


script/setup-dev
```diff
@@ -22,6 +22,7 @@ NODE_VERSION=$(cat "${DIR}/../.node-version")
 
 if command node -v 2>/dev/null | grep -q "${NODE_VERSION}"; then
   echo "Node version ${NODE_VERSION} already installed"
+  sudo corepack enable
```
eventualbuddha, line 25: Rather than duplicating this line maybe do the install stuff in an `else` and have the existing `sudo corepack enable` after the `fi`?

amcmanus: I like this suggestion. It makes it easier to understand we always want `corepack` to run, even if Node is already installed.

carolinemodic:
updated

amcmanus (approved):

eventualbuddha (approved):

PR #2985, opened 2023-02-07
feat(codemods): remove babel-based codemods

eventualbuddha:
## Overview

The babel-based codemods were too hard to keep the types working for and we don't actually expect to use them again–they were only there to serve as examples for future codemods. To that end, `ts-morph` is a better utility for making wide ranging changes to TypeScript projects. Thus, this commit keeps one example codemod using `ts-morph` and deletes everything else. It also improves the ergonomics of running it by using `esbuild-runner`.

## Demo Video or Screenshot
n/a

## Testing Plan 
Automated.

## Checklist


kofi-q (approved):
Nice! I remember struggling with this quite a bit 😅

eventualbuddha:
Okay, turns out I was wrong and the types were still broken. They were working locally until I ran `jest --clearCache` 😭. I decided it's not worth the trouble of keeping them typed and working, so I just deleted them all except one that uses `ts-morph` instead of the babel-based stuff.

kofi-q (approved):
🚀

PR #2986, opened 2023-02-07
Move VxScan auth to the backend

arsalansufi:

## Overview

Issue links: https://github.com/votingworks/vxsuite/issues/2906, https://github.com/votingworks/vxsuite/issues/2903

And another machine's auth moved to the backend. This PR sets up inserted smart card auth on the backend and updates VxScan to use it instead of frontend auth.

## Testing

- [x] Updated test suites
- [x] Manually tested VxScan auth and writing reports to cards

## Checklist

### DIFF
--- a/apps/vx-scan/backend/src/app.ts
+++ b/apps/vx-scan/backend/src/app.ts
@@ -311,7 +311,7 @@ function buildApi(
       const authStatus = auth.getAuthStatus();
 
       // Though we only initiate this action when a poll worker is logged in, a poll worker could
-      // remove their card right after, so we treat these as handled errors rather than unexpected
+      // remove their card right after, so we treat these as user errors rather than unexpected
       // errors
       if (authStatus.status !== 'logged_in') {
         return err(new Error('User is not logged in'));
--- a/apps/vx-scan/backend/src/app.ts
+++ b/apps/vx-scan/backend/src/app.ts
@@ -5,10 +5,12 @@ import {
   MarkThresholds,
   PollsState,
   PrecinctSelection,
+  SinglePrecinctSelection,
 } from '@votingworks/types';
 import {
   BALLOT_PACKAGE_FOLDER,
   readBallotPackageFromBuffer,
+  ScannerReportDataSchema,
   singlePrecinctSelectionFor,
 } from '@votingworks/utils';
 import express, { Application } from 'express';
@@ -17,6 +19,8 @@ import { ExportDataError } from '@votingworks/data';
 import path from 'path';
 import { existsSync } from 'fs';
 import { assert, err, ok, Result } from '@votingworks/basics';
+import { InsertedSmartCardAuthApi } from '@votingworks/auth';
+import { z } from 'zod';
 import { backupToUsbDrive } from './backup';
 import {
   exportCastVoteRecords,
@@ -33,7 +37,14 @@ import {
 } from './types';
 import { getMachineConfig } from './machine_config';
 
+export type CardDataSchemaName = 'ScannerReportDataSchema';
+
+const CARD_DATA_SCHEMA_MAPPING: Record<CardDataSchemaName, z.ZodSchema> = {
+  ScannerReportDataSchema,
+};
+
 function buildApi(
+  auth: InsertedSmartCardAuthApi,
   machine: PrecinctScannerStateMachine,
   interpreter: PrecinctScannerInterpreter,
   workspace: Workspace,
@@ -44,6 +55,19 @@ function buildApi(
 
   return grout.createApi({
     getMachineConfig,
+
+    getAuthStatus: () => auth.getAuthStatus(),
+    checkPin: (input: { pin: string }) => auth.checkPin(input),
+    readCardData: <T>({ schema }: { schema: CardDataSchemaName }) =>
+      auth.readCardData<T>({ schema: CARD_DATA_SCHEMA_MAPPING[schema] }),
+    writeCardData: <T>({
+      data,
+      schema,
+    }: {
+      data: T;
+      schema: CardDataSchemaName;
+    }) =>
+      auth.writeCardData<T>({ data, schema: CARD_DATA_SCHEMA_MAPPING[schema] }),
 
     async configureFromBallotPackageOnUsbDrive(): Promise<
       Result<void, ConfigurationError>
@@ -88,21 +112,28 @@ function buildApi(
         await fs.readFile(mostRecentBallotPackageFile.filePath)
       );
 
+      const { electionDefinition, ballots } = ballotPackage;
+
+      // If the election has only one precinct, set it automatically
+      let precinctSelection: SinglePrecinctSelection | undefined;
+      if (electionDefinition.election.precincts.length === 1) {
+        precinctSelection = singlePrecinctSelectionFor(
+          electionDefinition.election.precincts[0].id
+        );
+      }
+
       await store.withTransaction(async () => {
-        const { electionDefinition, ballots } = ballotPackage;
         store.setElection(electionDefinition.electionData);
-
-        // If the election has only one precinct, set it automatically
-        if (electionDefinition.election.precincts.length === 1) {
-          store.setPrecinctSelection(
-            singlePrecinctSelectionFor(
-              electionDefinition.election.precincts[0].id
-            )
-          );
+        if (precinctSelection) {
+          store.setPrecinctSelection(precinctSelection);
         }
-
         await store.setHmpbTemplates(ballots);
       });
+
+      auth.setElectionDefinition(electionDefinition);
+      if (precinctSelection) {
+        auth.setPrecinctSelection(precinctSelection);
+      }
 
       return ok();
     },
@@ -127,6 +158,8 @@ function buildApi(
       );
       interpreter.unconfigure();
       workspace.reset();
+      auth.clearElectionDefinition();
+      auth.clearPrecinctSelection();
     },
 
     setPrecinctSelection(input: {
@@ -138,6 +171,7 @@ function buildApi(
       );
       store.setPrecinctSelection(input.precinctSelection);
       workspace.resetElectionSession();
+      auth.setPrecinctSelection(input.precinctSelection);
     },
 
     setMarkThresholdOverrides(input: {
@@ -290,6 +324,7 @@ function buildApi(
 export type Api = ReturnType<typeof buildApi>;
 
 export function buildApp(
+  auth: InsertedSmartCardAuthApi,
   machine: PrecinctScannerStateMachine,
   interpreter: PrecinctScannerInterpreter,
   workspace: Workspace,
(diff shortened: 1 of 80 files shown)


libs/auth/src/inserted_smart_card_auth_with_memory_card.ts
```diff
@@ -0,0 +1,294 @@
...
+      await this.card.writeLongObject(data);
+    } catch (error) {
+      return wrapException(error);
+    }
+
+    // Verify that the write was in fact successful by reading the data
+    const readResult = await this.readCardData({ schema });
+    if (readResult.isErr()) {
+      return err(new Error('Verification of write by reading data failed'));
+    }
```
arsalansufi, line 123: We used to perform this validation client-side, e.g.

https://github.com/votingworks/vxsuite/blob/46b0c7ec6da8b39f46eec25b8133e48272cec097/apps/vx-scan/frontend/src/screens/poll_worker_screen.tsx#L79-L90

Especially with the shift to react-query and hooks on the frontend, I found it cleaner to encapsulate this validation on the backend.

I do recognize that I'm not validating the `timeSaved` field as we used to on the frontend. Still getting a sense of how important that is before adding generalized support for field validation on the backend. My hunch is that this will all be less relevant with the shift to Java Cards, which provide more storage space than our current memory cards. The main silent error today seems to be data truncation, hence this logic on the frontend:

https://github.com/votingworks/vxsuite/blob/46b0c7ec6da8b39f46eec25b8133e48272cec097/apps/vx-scan/frontend/src/screens/poll_worker_screen.tsx#L300-L311

apps/vx-scan/backend/src/app.ts
```diff
@@ -45,6 +56,19 @@ function buildApi(
...
+    readCardData: <T>({ schema }: { schema: CardDataSchemaName }) =>
+      auth.readCardData<T>({ schema: CARD_DATA_SCHEMA_MAPPING[schema] }),
+    writeCardData: <T>({
+      data,
+      schema,
+    }: {
+      data: T;
+      schema: CardDataSchemaName;
+    }) =>
+      auth.writeCardData<T>({ data, schema: CARD_DATA_SCHEMA_MAPPING[schema] }),
```
arsalansufi, line 70: Serialization strikes again - this time for Zod schemas. While I could add serialization support for Zod schemas to Grout, e.g. through a combo of https://github.com/StefanTerdell/zod-to-json-schema and https://github.com/StefanTerdell/json-schema-to-zod, I didn't want to increase the scope of this PR too much.

For now, I figured the frontend could just specify the schema name and the backend could look up the Zod schema given that. We do lose some type safety because of this since the frontend can no longer guarantee that the data and schema match, but again, we can always revisit.

apps/vx-scan/backend/src/app_config.test.ts
```diff
@@ -313,3 +340,46 @@ test('ballot batching', async () => {
...
+  expect(mockAuth.checkPin).toHaveBeenNthCalledWith(1, { pin: '123456' });
+  expect(mockAuth.readCardData).toHaveBeenCalledTimes(1);
+  expect(mockAuth.readCardData).toHaveBeenNthCalledWith(1, {
+    schema: ScannerReportDataSchema,
+  });
+  expect(mockAuth.writeCardData).toHaveBeenCalledTimes(1);
+  expect(mockAuth.writeCardData).toHaveBeenNthCalledWith(1, {
+    data: {},
+    schema: ScannerReportDataSchema,
+  });
```
arsalansufi, line 384: As simple as these auth tests are, these are what caught the Zod schema serialization error!

apps/vx-scan/frontend/src/api.ts
```diff
@@ -33,6 +52,67 @@ export const getMachineConfig = {
...
+    return ['readCardData'];
+  },
+  useQuery<T>(schema: CardDataSchemaName) {
+    const apiClient = useApiClient();
+    return useQuery(
+      this.queryKey(),
+      () =>
+        // In the conversion from the Api type to the ApiClient type, generics are lost, so we cast
+        // back to the original type here
+        (apiClient.readCardData as Api['readCardData'])<T>({ schema }),
```
arsalansufi, line 92: So in addition to the Zod schema serialization challenge affecting type safety, there's a separate issue of the conversion from the `Api` type to the `ApiClient` type causing generics to be lost.

You can see that here when I remove the cast:

error

Haven't gotten to the bottom of it yet, but something about this utility is having that effect:

https://github.com/votingworks/vxsuite/blob/46b0c7ec6da8b39f46eec25b8133e48272cec097/libs/grout/src/client.ts#L8-L13

The following:

```ts
grout.AsyncRpcMethod<(<T>({ schema }: {
    schema: "ScannerReportDataSchema";
}) => Promise<Result<Optional<T>, Error | SyntaxError | ZodError<any>>>)
```

Becomes:

```ts
(args_0: {
    schema: "ScannerReportDataSchema";
}) => Promise<Result<unknown, Error | SyntaxError | ZodError<any>>>
```

apps/vx-scan/frontend/test/helpers/mock_api_client.tsx
```diff
@@ -40,16 +46,83 @@ export const statusNoPaper: PrecinctScannerStatus = {
...
+  MockClient<Api>,
+  'getAuthStatus' | 'writeCardData'
+> & {
+  // Because this is polled so frequently, we opt for a standard jest mock instead of a
+  // libs/test-utils mock since the latter requires every call to be explicitly mocked
+  getAuthStatus: jest.Mock;
+  // Because the values passed to this are so complex, we opt for a standard jest mock instead of a
+  // libs/test-utils mock since the latter requires exact input matching and doesn't support
+  // matchers like expect.objectContaining
+  writeCardData: jest.Mock;
```
arsalansufi, line 59: I contemplated just providing the exact values, but they were hard to generate and also included timestamps. I know the timestamps can be worked around by mocking time, but this didn't feel worth the trouble

apps/vx-scan/frontend/src/app.test.tsx
```diff
@@ -93,7 +79,7 @@ beforeEach(() => {
   );
   window.kiosk = kiosk;
 
-  apiMock.mockApiClient.reset();
+  apiMock = createApiMock();
```
arsalansufi, line 82: The `reset` command didn't quite suit my needs since I've overridden some of the default mocks with jest mocks. Instead of reasoning through how to cleanly reset everything in a `reset` wrapper, I found it easier to just create a fresh mock for every test

To avoid gotchas down the road in other VxScan tests, I may proactively update all VxScan tests to take this approach

apps/vx-scan/frontend/src/app_tally_report_paths.test.tsx
```diff
@@ -348,16 +341,15 @@ test('printing: polls closed, primary election, all precincts + quickresults on'
     electionMinimalExhaustiveSampleWithReportingUrlDefinition;
   const { election } = electionDefinition;
   apiMock.expectGetConfig({ electionDefinition, pollsState: 'polls_open' });
-  apiMock.expectGetScannerStatus({ ...statusNoPaper, ballotsCounted: 3 }, 3);
-  const { card } = renderApp({ connectPrinter: true });
+  apiMock.expectGetScannerStatus({ ...statusNoPaper, ballotsCounted: 3 }, 4);
```
arsalansufi, line 344: One notable pain point while updating these tests was having to tweak get scanner status counts, through a process of guess and check. Why this was necessary I'm still not exactly sure but so many things can affect test duration and by extension these counts.

This, in conjunction with my manual mocking of auth status, is making think that a built-in solution for mocking polled methods is worthwhile

apps/vx-scan/frontend/src/app.test.tsx
jonahkagan: Go for it. I saw that you used this pattern in the other apps, so I think making it consistent makes sense. Feel free to remove the `reset` method from the mock API client completely and make this the standard approach.

apps/vx-scan/backend/src/app.ts
arsalansufi: Mm really good point. In line with what we discussed in the backend auth spec about encapsulating as much as possible in the backend

For this case, I like the idea of a specific RPC method for saving the report data to the card

apps/vx-scan/frontend/src/api.ts
arsalansufi: 👍 Irrelevant for now given what I landed on

arsalansufi:

apps/vx-scan/backend/src/app.ts
```diff
@@ -310,6 +310,9 @@ function buildApi(
     }): Promise<Result<void, Error>> {
       const authStatus = auth.getAuthStatus();
 
+      // Though we only initiate this action when a poll worker is logged in, a poll worker could
+      // remove their card right after, so we treat these as handled errors rather than unexpected
+      // errors
```
jonahkagan, line 315: What should happen overall in the case that a poll worker removes their card right after? Currently, we're just swallowing the error on the frontend. Will that leave things in a consistent state? The user won't necessarily know something went wrong.

This makes me think we should either use assertions here and treat them as unhandled errors (which will at least fail loudly) or think a little more about how to make this safe.

libs/grout/test-utils/src/mock_client.ts
```diff
@@ -73,12 +72,6 @@ export function createMockClient<Api extends AnyApi>(options?: {
         mockMethod.assertComplete();
       }
     },
-
-    reset(): void {
-      for (const mockMethod of Object.values(mockMethods)) {
-        mockMethod.reset();
-      }
-    },
```
jonahkagan, line 81: Can you update the grout README with the new pattern?

jonahkagan (approved):

apps/vx-scan/backend/src/app.ts
arsalansufi: Totally agree that we should have clear feedback for the user in this case!

In practice, if you remove your card right after this is action is triggered, the failure won't happen here but in the actual card writing logic, which also wraps errors in a `Result`. So modifying these to be assertions won't make much of practical difference. My manual testing verified this.

On that note, this is a pre-existing problem with pre-existing TODOs, e.g.

https://github.com/votingworks/vxsuite/blob/449583919a1d2f4d3b10e0c05bac7192d38b4b3b/apps/vx-scan/frontend/src/screens/poll_worker_screen.tsx#L305-L310


Since this isn't a regression, I'm going to move forward with this PR and have added this to my [miscellaneous cleanup issue](https://github.com/votingworks/vxsuite/issues/2965)!

arsalansufi: Also, for completeness, I tested what actually happens when you remove a card preemptively both before and after this PR, and in both cases, polls successfully transition, but the report written to the card is corrupted/unreadable

jonahkagan: Thanks for looking deeper into this! Given what you found out about the report data getting corrupted, it makes me think that ideally the polls transition + report data writing should be somehow transactional. Hard to get right, so maybe not important to prioritize. But I agree with you that showing an error to the user seems like a bare minimum

PR #3001, opened 2023-02-09
Take election def as input in auth lib methods

arsalansufi:

## Overview

This PR adjusts how the auth lib accesses the election definition. I've thus far been having backends configure the auth lib with the election definition on startup and subsequent election definition changes, meaning that the auth lib is essentially caching the election definition. This is error prone since we have to remember to update the auth lib's copy of the election definition any time it changes in the containing backend.

This PR looks at an alternative: having the backend provide the election definition on every auth lib call. Even though the frontend polls auth status, we're not worried about this from a performance perspective since 1) SQLite can comfortably handle this read pattern and 2) we're in a fully local context.

## Testing

- [x] Updated/added automated tests
- [x] Tested VxAdmin auth manually
- [x] Tested VxCentralScan auth manually
- [x] Tested VxScan auth manually

## Checklist

### DIFF
--- a/apps/admin/backend/src/server.ts
+++ b/apps/admin/backend/src/server.ts
@@ -38,11 +38,22 @@ function constructDippedSmartCardAuthMachineState(
   workspace: Workspace
 ): DippedSmartCardAuthMachineState {
   // TODO: Once we actually support multiple elections, configure the auth instance with the
-  // currently selected election rather than the first. In fact, do so as soon as the currently
-  // selected election is persisted on the backend instead of the frontend since, even today, in
-  // dev, we can end up with multiple election definitions under the hood via incognito windows
+  // currently selected election rather than the most recently created. In fact, do so as soon as
+  // the currently selected election is persisted on the backend instead of the frontend since,
+  // even today, in dev, we can end up with multiple election definitions under the hood via
+  // incognito windows
   const elections = workspace.store.getElections();
-  return { electionDefinition: elections[0]?.electionDefinition };
+  const mostRecentlyCreatedElection =
+    elections.length > 0
+      ? elections.reduce((e1, e2) =>
+          new Date(e1.createdAt) > new Date(e2.createdAt)
+            ? /* istanbul ignore next */ e1
+            : e2
+        )
+      : undefined;
+  return {
+    electionDefinition: mostRecentlyCreatedElection?.electionDefinition,
+  };
 }
 
 function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
--- a/apps/admin/backend/src/server.ts
+++ b/apps/admin/backend/src/server.ts
@@ -4,16 +4,18 @@ import {
   BallotPageLayout,
   CandidateContest,
   CastVoteRecord,
+  ElectionDefinition,
   getBallotStyle,
   getContests,
   Id,
   InlineBallotImage,
+  Optional,
   Rect,
   safeParse,
   safeParseJson,
   safeParseNumber,
 } from '@votingworks/types';
-import { zip } from '@votingworks/basics';
+import { assert, zip } from '@votingworks/basics';
 import express, { Application } from 'express';
 import compression from 'compression';
 import * as fs from 'fs/promises';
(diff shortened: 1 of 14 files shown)


apps/admin/backend/src/server.ts
```diff
@@ -38,11 +38,22 @@ function constructDippedSmartCardAuthMachineState(
   workspace: Workspace
 ): DippedSmartCardAuthMachineState {
   // TODO: Once we actually support multiple elections, configure the auth instance with the
-  // currently selected election rather than the first. In fact, do so as soon as the currently
-  // selected election is persisted on the backend instead of the frontend since, even today, in
-  // dev, we can end up with multiple election definitions under the hood via incognito windows
+  // currently selected election rather than the most recently created. In fact, do so as soon as
+  // the currently selected election is persisted on the backend instead of the frontend since,
+  // even today, in dev, we can end up with multiple election definitions under the hood via
+  // incognito windows
```
arsalansufi, line 44: This was an interesting one. Since Cypress tests do trigger this "multiple election definitions under the hood" state, but we no longer set the auth lib election definition when a user loads an election definition (every integration test starts with the load of an election definition), I had to update the selection logic here to select the most recently created election definition rather than the first

Again, all inconsequential for prod

libs/auth/src/inserted_smart_card_auth_with_memory_card.ts
```diff
@@ -49,70 +45,56 @@ export class InsertedSmartCardAuthWithMemoryCard
...
-  readCardData<T>({
-    schema,
-  }: {
-    schema: z.ZodSchema<T>;
-  }): Promise<Result<Optional<T>, SyntaxError | z.ZodError | Error>> {
-    return this.card.readLongObject(schema);
+  readCardData<T>(
+    machineState: InsertedSmartCardAuthMachineState,
+    input: { schema: z.ZodSchema<T> }
+  ): Promise<Result<Optional<T>, SyntaxError | z.ZodError | Error>> {
```
arsalansufi, line 80: Though I'm not doing this yet since this implementation will be short lived, I could call `this.checkCardReaderAndUpdateAuthStatus(machineState)` and then verify that the user is logged in as a sys admin and that there's a card in the card reader

apps/admin/backend/src/server.ts
jonahkagan: Is there any way to make the Cypress tests unconfigure when they are done so that we don't have this problem?

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -12,27 +12,40 @@ import {
...
 export interface DippedSmartCardAuthConfig {
   allowElectionManagersToAccessUnconfiguredMachines?: boolean;
+}
+
+/**
+ * Machine state that the consumer is responsible for providing
+ */
+export interface DippedSmartCardAuthMachineState {
   electionDefinition?: ElectionDefinition;
 }
```
jonahkagan, line 51: Since we're really just checking the election hash from the machine against the hash from the card, should we just pass in the election hash instead of this abstracted machine state? Might be a little easier to grok. Or is there a reason we might want the full election def or want to add other fields to the machine state?

I know card programming is one place where we use the `electionData` string as well, but you could argue that it's more clear to have that be its own argument passed into `programCard` to indicate that we're programming the card with that data (vs all the other methods that are just doing a hash comparison to validate the current card)

libs/auth/src/dipped_smart_card_auth_with_memory_card.ts
```diff
@@ -161,24 +151,31 @@ export class DippedSmartCardAuthWithMemoryCard
...
-  clearElectionDefinition(): void {
-    delete this.config.electionDefinition;
+  private async checkCardReaderAndUpdateAuthStatus(
+    machineState: DippedSmartCardAuthMachineState
+  ) {
+    const cardSummary = await this.card.readSummary();
+    this.updateAuthStatus(machineState, {
+      type: 'check_card_reader',
+      cardSummary,
+    });
```
jonahkagan, line 161: This is quite an interesting piece of code after this change. It's an auth state manager that only changes state when you check it. Quite Schroedinger's Cat-like. I don't see any problem with it, but I don't know if I've seen something quite like this before, so it makes me slightly wary.

jonahkagan (approved):

libs/auth/src/dipped_smart_card_auth_with_memory_card.ts
arsalansufi: The Schroedinger's Cat analogy really works 😆

Yeah, I felt this too. I couldn't conceive of a reasonable way to compute the auth status before it was requested without returning to some form of election def caching in the auth lib.

The main risk that I see with this new approach is forgetting to refresh auth status internally before performing another auth status changing operation, like PIN checking. But at least that risk is contained to the auth lib, unlike the risk of correctly caching the election def, which would affect all consumers.

So given the smaller risk surface area of this new approach, I do think I prefer it. Will think of ways to mitigate the risk of forgetting to refresh auth status internally before performing another auth status changing operation when I create the Java Card implementations!

apps/admin/backend/src/server.ts
arsalansufi: Went ahead and added this to my [miscellaneous cleanup issue](https://github.com/votingworks/vxsuite/issues/2965). I think a generalized solution for resetting the backend between Cypress tests would be great

libs/auth/src/dipped_smart_card_auth_with_memory_card.ts
arsalansufi: I'll also be curious to see whether the need to always provide election hash to auth methods becomes cumbersome, e.g. would this complicate creating an auth decorator? (I don't think so but just noting)

jonahkagan: The reasoning about reducing the risk surface area makes sense to me!

PR #3004, opened 2023-02-13
Ansible and setup-dev

amcmanus:
## Overview
This PR begins the transition from using shell-scripts and Makefile configs to configure operating system level concerns, e.g. users, filesystems, packages, etc...

### Ansible
Ansible and any dependencies are installed via the new `script/install-ansible.sh` script. This is kept as minimal as possible so all future system changes are done via Ansible instead of more scripts/Makefiles. 

### setup-dev
I have kept `script/setup-dev` to minimize changes to documentation and developer processes, but the script now calls the appropriate Ansible playbook to manage Node and any dependencies. 

### Misc
This also begins our effort to bring consistency across the `vxsuite` and `vxsuite-complete-system` repos. We will be able to use/re-use Ansible playbooks within this repo elsewhere, helping ensure consistency in our build processes, regardless of repo.


ansible/playbooks/setup-node.yaml
```diff
@@ -0,0 +1,42 @@
+---
+- name: Install Node and other dependencies
+  hosts: 127.0.0.1
+  connection: local
+  become: true
+
+  vars:
+    node_version: "16.*"
```
carolinemodic, line 8: I think there was some very subtle behavior difference people were getting running unit tests between minor versions of node. Is it possible in ansible to specify a more specific version or do we have to just live with 16.* (same question for pnpm version but I don't think we've had any issue there)

script/setup-dev
```diff
@@ -1,57 +1,17 @@
...
-### Note: To update the version of NodeJS, update the `.node-version` file at
-### the root of the repository and add `hashes/node-v*.sha256` files for the
-### new version.
+# This used to install node and other dependencies.
+# That functionality has been migrated to Ansible.
+# This script now simply calls that functionality.
 
 set -euo pipefail
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
carolinemodic, line 9: is there any harm in calling setup_ansible in here as well? I guess I'm wondering for VxDev where this script is automatically called, but they don't have ansible installed yet (we could also just make the update_code script in VxDev call install_ansible)

carolinemodic (commented):
Two questions

script/setup-dev
amcmanus: No harm at all, great idea. I'll incorporate it with additional changes related to node/pnpm versioning.

ansible/playbooks/setup-node.yaml
benadida: +1 if we can specify a specific version, I think we win to rule out small changes in behavior across versions – since our systems are offline, the incremental win we get from automatic minor version updates are not as high as the risk of behavior changes.

benadida (approved):

ansible/playbooks/setup-node.yaml
amcmanus: @benadida, @carolinemodic, totally agree that pinning versions is ideal. It does not appear to be the way Node expects installation to go though: https://github.com/nodesource/distributions#using-debian-as-root-2

We can certainly pin to a major.minor.patch version, but we currently do not do that in `vxsuite-complete-system`, even though we do in `vxsuite`.  I have pushed up an example pinning it to `16.19.0` (which is what a new build in `vxsuite-complete-system` would use), but since we do a very manual install in `vxsuite` to `16.4.2`, I'd like to get more feedback and tests.

PR #3005, opened 2023-02-13
Rename `libs/data` to `libs/backend`

adghayes:
To make it clearer that `libs/data` is a library intended for use across the backends.


arsalansufi (approved):

adghayes:
Separating these renaming chances into separate PRs because they're actually being quite annoying and want to get them through.

@carolinemodic, branch protection `test-libs-data` => `test-libs-shared-backend`.

Issue #2927, opened 2023-01-25
Convert all VXF election definitions to CDF


jonahkagan:
Write a script

Will need to add a custom field to CDF to support `party.shortName`

jonahkagan:
Decided it wasn't necessary to convert all the fixtures during this project

PR #3006, opened 2023-02-13
feat(types): Ballot Definition CDF

jonahkagan:
## Overview
Task: #2927 

- Generates the ballot definition CDF types
- Implements functions to convert between VXF and CDF
- Extends CDF with a `Party.vxBallotLabel` field

Next, I'll be using these conversion functions to start converting our fixtures, so they may need to get tweaked, but I wanted to land an initial PR first, since this is already a sizeable chunk of code.

_Review by commit_

## Demo Video or Screenshot
N/A

## Testing Plan 
- Some small unit tests for now

## Checklist


libs/types/src/cdf/README.md
```diff
@@ -0,0 +1,18 @@
...
+[CDF Schema Builder](../../../cdf-schema-builder/).
+
+However, in certain cases we have needed to add extensions and/or constraints to
+the CDF specs based on the needs of our system. In each subdirectory, you will
+find a `vx-schema.json` file that contains the VotingWorks version of the CDF
+schema. The original NIST schemas can be found in the
+[data directory](../../data/). Note that due to some weirdness with how NIST
+provides their schemas, we use both the XSD (XML) format of the schema, which
+includes documentation, and the JSON format, which includes the information we
+need to generate the types.
```
jonahkagan, line 18: This isn't quite true yet, but I'm assuming that if we like this approach, @adghayes or I will update the other CDF subdirectories accordingly

libs/basics/src/unique.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Returns an array with duplicate values removed.
+ */
+export function unique<T>(array: readonly T[]): T[] {
+  return Array.from(new Set(array));
+}
```
adghayes, line 6: Hell yes, this is a utility function if there ever was one.

libs/cdf-schema-builder/README.md
```diff
@@ -23,10 +23,10 @@ With the CLI:
 
 ```sh
 # in a package.json script:
-cdf-schema-builder schema.xml > src/schema.ts
+cdf-schema-builder schema.xsd schema.json > schema.ts
```
adghayes, line 26: Thank you for updating this.

libs/types/src/cdf/ballot-definition/index.ts
```diff
@@ -0,0 +1,3712 @@
...
+   */
+  readonly PartyScopeGpUnitIds?: readonly string[];
+
+  /**
+   * The slogan or motto used by a political party.
+   */
+  readonly Slogan?: InternationalizedText;
+
+  /**
+   * The label to describe candidates from this party on a ballot.
```
adghayes, line 2586: Glad this documentation works from the JSON Schema.

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -0,0 +1,412 @@
...
+  return {
+    '@type': 'BallotDefinition.BallotDefinition',
+
+    Election: [
+      {
+        '@type': 'BallotDefinition.Election',
+        ElectionScopeId: stateId,
+        StartDate: dateString(vxfElection.date),
+        EndDate: dateString(vxfElection.date),
+        Type: Cdf.ElectionType.General,
```
adghayes, line 51: Could this also be `Primary`?

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -0,0 +1,412 @@
...
+        // TODO Eventually, we might want to list all of the contests included
+        // in the ballot style. Currently, our system infers it from the
+        // associated precincts' districts.
+        BallotStyle: vxfElection.ballotStyles.map(
+          (ballotStyle): Cdf.BallotStyle => ({
+            '@type': 'BallotDefinition.BallotStyle',
+            GpUnitIds: ballotStyle.precincts,
+            PartyIds: ballotStyle.partyId ? [ballotStyle.partyId] : undefined,
+            // In CDF, ballot styles don't have an id field. I think this might be
+            // because each ballot style can be uniquely identified by its
```
adghayes, line 138: This is fascinating, because the CVR CDF _does_ identify CVRs by a ballot style id.

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -0,0 +1,412 @@
...
+  const gpUnits = assertDefined(cdfBallotDefinition.GpUnit);
+
+  const state = find(
+    gpUnits,
+    (gpUnit) => gpUnit.Type === Cdf.ReportingUnitType.State
+  );
+  const county = find(
+    gpUnits,
+    (gpUnit) => gpUnit.Type === Cdf.ReportingUnitType.County
+  );
```
adghayes, line 237: Are these guaranteed to exist? Is that one of the constraints you'd like to enforce?

adghayes (approved):
LGTM, pending some system for enforcing our constraints!

libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan: Ooh good call. How can I tell from VXF? Maybe looking for contests and/or ballot styles with `partyId`?

libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan: 🤔 that is very odd

libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan: Hmm, good question. I think I put this logic in when I was only thinking about round trip conversion from VXF<->CDF. I wonder how many places in the product actual use `state` and `county`. Assuming there are some, I would imagine it makes sense to enforce this as a constraint to start.

libs/types/src/cdf/ballot-definition/convert.ts
adghayes: Ugh, I've been unsure about this. And we've employed two ways looking in the VXF, looking at contests with `partyId` vs. looking at ballot styles with `partyId`. But, because there is the possibility of non-partisan ballots during a primary or an open primary (not supported yet), I think looking at contests is the best bet.

jonahkagan: Ok, that's helpful. Since we're just using this logic for converting existing fixtures, I'm not too worried about figuring out a long-term solution. In fact, I think this problem will go away once we've moved to CDF.

f37a21beb

jonahkagan:
Gonna merge this and follow up with validation/constraint enforcement as a separate PR!

PR #3010, opened 2023-02-14
[libs/ui] Add theme-aware Card component

kofi-q:
## Demo Video or Screenshot
https://user-images.githubusercontent.com/264902/218809240-56dae693-3421-4016-a5c2-a18dc0220917.mov

### Intended in-app usage:
Screenshot 2023-02-14 at 11 12 47

### Existing cards in VxMark, for reference:
Screenshot 2023-02-14 at 11 36 04

## Testing Plan 
- Visual testing within storybook.js
- Added unit tests to verify content and optional footer are rendered as expected.

## Checklist

### DIFF
--- a/libs/types/src/ui_theme.ts
+++ b/libs/types/src/ui_theme.ts
@@ -15,7 +15,7 @@ export enum Color {
   DANGER_LOW_CONTRAST = '#ff3d3d',
   DANGER_MEDIUM_CONTRAST = '#820b0b',
   GRAY_DARK = '#222222',
-  GRAY_LIGHT = '#757575',
+  GRAY_LIGHT = '#8a8a8a',
   GRAY_MEDIUM = '#5c5c5c',
   OFF_BLACK = '#080808',
   OFF_WHITE = '#fafafa',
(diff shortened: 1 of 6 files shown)


libs/types/src/ui_theme.ts
```diff
@@ -15,7 +15,7 @@ export enum Color {
   DANGER_LOW_CONTRAST = '#ff3d3d',
   DANGER_MEDIUM_CONTRAST = '#820b0b',
   GRAY_DARK = '#222222',
-  GRAY_LIGHT = '#757575',
+  GRAY_LIGHT = '#8a8a8a',
```
kofi-q, line 18: Missed this in the last PR, but saw an accessibility alert in storybook about this not passing the 4.5:1 contrast ratio requirement, so bumping up the brightness a bit (used as foreground in low-contrast theme)

adghayes: storybook gives you contrast ratio warnings? If so, very neat.

adghayes (approved):
LGTM.

libs/types/src/ui_theme.ts
kofi-q: Yeah! We currently have to remember to check the accessibility tab they provide, but I remember reading about some teams rolling storybook into their automated tests, so that they get that reporting on each commit.

That could also be a useful way for us to cut down on the duplication from writing tests, if the stories are just run as tests - hoping to look into that a little more after this project 🤞

PR #3011, opened 2023-02-14
Move VxMark auth to the backend

arsalansufi:

## Overview

Issue links: https://github.com/votingworks/vxsuite/issues/2906, https://github.com/votingworks/vxsuite/issues/2902

Last machine's auth moved to the backend! This PR adds support for cardless voter sessions to libs/auth and then updates VxMark to use backend auth instead of frontend auth.

I'll open a separate PR once this is merged to clean up no longer used frontend auth code.

## Testing

- [x] Updated test suites
- [x] Manually tested VxMark auth, election def configuration from card, and reading/clearing poll worker card reports

## Checklist


libs/auth/src/inserted_smart_card_auth_with_memory_card.ts
```diff
@@ -74,11 +104,51 @@ export class InsertedSmartCardAuthWithMemoryCard
...
+  ): Promise<void> {
+    assert(this.config.allowedUserRoles.includes('cardless_voter'));
+
+    await this.checkCardReaderAndUpdateAuthStatus(machineState);
+    if (
+      this.authStatus.status !== 'logged_in' ||
+      this.authStatus.user.role !== 'poll_worker'
+    ) {
+      return;
+    }
```
arsalansufi, line 119: I'm landing on the following mental model for auth lib errors:

_For methods that transition auth status, e.g. `checkPin`, `logOut`, `startCardlessVoterSession`, `endCardlessVoterSession`_

- Use `assert` for states incompatible with the static config (`this.config`)
- For other incompatible states (e.g. trying to check PIN when you're not in the PIN-checking state or trying to start a cardless voter session when you're not logged in as a poll worker), just ignore the request and leave the auth status unchanged

_For card reading and writing methods, e.g. `programCard`, `unprogramCard`, `readCardData`, `writeCardData`_

- Use `assert` for states incompatible with the static config (`this.config`)
- Otherwise always use an error `Result` instead of throwing
- Methods that can perform access checks themselves should (e.g. `programCard` should check if you're logged in as a system administrator and return an error `Result` if need be). Methods that can't because the access checks are dependent on the context in which they're used (e.g. `readCardData`) should leave those checks to backend consumers. Those backend consumers should also use an error `Result` instead of throwing.

You'll see changes in this and the next few commits to align with the above!

apps/mark/backend/src/app.ts
```diff
@@ -1,18 +1,106 @@
...
+
 import { getMachineConfig } from './machine_config';
 
-function buildApi() {
+function buildApi(auth: InsertedSmartCardAuthApi) {
   return grout.createApi({
     getMachineConfig,
+
+    // TODO: Once election definition has been moved to the backend, no longer require the frontend
+    // to provide election hash to this and other methods that use the auth lib
```
arsalansufi, line 16: I contemplated moving election definition to the backend but again wanted to limit the scope of this PR

apps/mark/backend/tsconfig.build.json
```diff
@@ -10,10 +10,8 @@
     "declarationMap": true
   },
   "references": [
-    { "path": "../../../libs/api/tsconfig.build.json" },
+    { "path": "../../../libs/auth/tsconfig.build.json" },
     { "path": "../../../libs/basics/tsconfig.build.json" },
-    { "path": "../../../libs/backend/tsconfig.build.json" },
-    { "path": "../../../libs/db/tsconfig.build.json" },
```
arsalansufi, line 16: Looks like a few libs were specified here that aren't actually used, a la copy paste

libs/ui/src/hooks/auth/auth_helpers.ts
```diff
@@ -348,6 +348,15 @@ export function isPollWorkerAuth(
...
+): auth is InsertedSmartcardAuth.CardlessVoterLoggedIn;
+export function isCardlessVoterAuth(
+  auth: InsertedSmartCardAuth.AuthStatus
+): auth is InsertedSmartCardAuth.CardlessVoterLoggedIn;
+export function isCardlessVoterAuth(
+  auth: InsertedSmartcardAuth.Auth | InsertedSmartCardAuth.AuthStatus
+): auth is InsertedSmartCardAuth.CardlessVoterLoggedIn;
+export function isCardlessVoterAuth(
+  auth: InsertedSmartcardAuth.Auth | InsertedSmartCardAuth.AuthStatus
+): boolean {
```
arsalansufi, line 360: After this PR is merged, I'll clean up all the old, now unused auth code, including the old types!

apps/mark/frontend/src/app_root.tsx
```diff
@@ -364,9 +364,12 @@ export function AppRoot({
 
   const hidePostVotingInstructions = useCallback(() => {
     clearTimeout(PostVotingInstructionsTimeout.current);
-    if (isCardlessVoterAuth(auth)) auth.logOut();
```
arsalansufi, line 367: Got rid of the old `isCardlessVoterAuth` check because `endCardlessVoterSession` is always safe to run, kind of like `logOut`. And removing the check made it a bit easier to maintain 100% code coverage

apps/mark/frontend/src/app_root.tsx
```diff
@@ -507,34 +517,51 @@ export function AppRoot({
...
+      ) {
         resetBallot();
       }
     }
     resetBallotOnLogout();
-  }, [auth, resetBallot, initializedFromStorage]);
+  }, [authStatus, resetBallot, initializedFromStorage]);
 
-  const endVoterSession = useCallback(() => {
-    assert(isCardlessVoterAuth(auth));
```
arsalansufi, line 534: Similar reason here as above for removing the `isCardlessVoterAuth` assertion

apps/mark/frontend/src/pages/idle_page.tsx
```diff
@@ -28,7 +27,6 @@ export function IdlePage(): JSX.Element {
   useEffect(() => {
     async function reset() {
       setIsLoading(true);
-      await sleep(1000); // Allow "Clearing ballot" text to display
```
arsalansufi, line 31: This sleep is no longer necessary because `endVoterSession` is now truly async again

apps/mark/frontend/src/pages/poll_worker_screen.tsx
```diff
@@ -495,18 +509,18 @@ export function PollWorkerScreen({
   const [isDiagnosticsScreenOpen, setIsDiagnosticsScreenOpen] = useState(false);
   const [scannerReportData, setScannerReportData] =
     useState<ScannerReportData>();
+
   useEffect(() => {
-    if (scannerReportData) return;
-    if (!pollworkerAuth.card.hasStoredData) return;
```
arsalansufi, line 500: Got rid of the `hasStoredData` field in new models since, in Java Card land, the best way to populate such a boolean would just be to try to read data. And as you'll see below anyway, I got rid of the need to even check `hasStoredData` here

apps/mark/frontend/src/pages/poll_worker_screen.tsx
```diff
@@ -495,18 +509,18 @@ export function PollWorkerScreen({
...
+    readScannerReportDataFromCardMutation.mutate(undefined, {
+      onSuccess(result) {
+        if (result.isOk() && result.ok() !== undefined) {
+          setScannerReportData(result.ok());
+        }
+      },
+    });
+    // Try to read scanner report data from the card once on mount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
arsalansufi, line 523: Only performing this action on mount has conceptually always been what we wanted. When I actually populated the dependency array fully, `readScannerReportDataFromCardMutation` caused the dreaded `useEffect` infinite loop because it's redefined on every render. I didn't want to use a `useCallback` as a bandaid to work around that and instead wanted to be more explicit.

I spent some time thinking about how to move this out of a `useEffect` altogether and into an event handler, but what makes this challenging is that this read isn't triggered by a button click - it's triggered by card insertion.

Overall, I'm okay with where this landed!

apps/mark/frontend/src/pages/replace_election_screen.tsx
```diff
@@ -51,7 +51,9 @@ export function ReplaceElectionScreen({
         await makeCancelable(getElectionDefinitionFromCard())
       );
     })();
-  }, [getElectionDefinitionFromCard, makeCancelable]);
+    // Try to read scanner report data from the card once on mount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
arsalansufi, line 56: A case basically identical to the one above: A read that needs to happen just once on component mount that's also triggered by card insertion and that also infinite loops when all dependencies are listed

apps/admin/backend/src/server.ts
```diff
@@ -68,20 +68,28 @@ function constructDippedSmartCardAuthMachineState(
 
 function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
   return grout.createApi({
-    getAuthStatus: () =>
-      auth.getAuthStatus(constructDippedSmartCardAuthMachineState(workspace)),
+    getAuthStatus() {
+      return auth.getAuthStatus(
+        constructDippedSmartCardAuthMachineState(workspace)
+      );
+    },
```
arsalansufi, line 75: Noticed that all Grout methods before I started adding mine for auth use function syntax instead of fat arrow syntax so decided to be consistent

apps/scan/backend/src/app_scan.test.ts
```diff
@@ -1200,3 +1212,58 @@ test('stops completely if plustekctl freezes and cant be killed', async () => {
     error: 'paper_status_timed_out',
   });
 });
+
+test('write scanner report data to card', async () => {
```
arsalansufi, line 1216: Moved this test from `app_config.test.ts` to `app_scan.test.ts` since it fits better under the latter

libs/auth/src/inserted_smart_card_auth_with_memory_card.ts
jonahkagan: Thanks for outlining this mental model. A couple of questions:
- What's the motivation for "just ignore the request and leave the auth status unchanged" vs failing more loudly?
- Why do access checks in methods like `programCard` return an error `Result`? To me, that would imply that the consumer isn't responsible for checking the auth before calling the method and needs to explicitly think about an "unauthorized" use case (e.g. if we allowed any user type to press a "program card" button ). Vs using an `assert` would imply that the system should only allow users to call `programCard` if they're logged in as a sys admin and if that doesn't happen, the programmer made a mistake.

apps/mark/backend/src/app.ts
jonahkagan: Loving that grout makes this a low cost decision!

apps/mark/backend/src/app.ts
```diff
@@ -1,18 +1,106 @@
...
+    }: {
+      electionHash?: string;
+    }): Promise<Result<void, Error>> {
+      const authStatus = await auth.getAuthStatus({ electionHash });
+      if (authStatus.status !== 'logged_in') {
+        return err(new Error('User is not logged in'));
+      }
+      if (authStatus.user.role !== 'poll_worker') {
+        return err(new Error('User is not a poll worker'));
+      }
```
jonahkagan, line 92: I like this model of checking role-based permissions in the API. That meshes with general patterns I've seen for authentication vs authorization.

I wonder if the `logged_in` check should be part of the auth lib or stay here in the API. I don't have a strong opinion.

apps/mark/backend/tsconfig.build.json
jonahkagan: nice catch!

apps/mark/frontend/src/pages/poll_worker_screen.tsx
jonahkagan: How about using `useQueryChangeListener` to detect when  `authStatusQuery` changes to a logged in pollworker? (It would have to be moved up to a parent component, but that might not be the worst thing in the world...)

apps/mark/frontend/src/pages/poll_worker_screen.tsx
jonahkagan: That makes sense

apps/mark/frontend/src/app_root.tsx
```diff
@@ -293,29 +298,24 @@ export function AppRoot({
...
-    },
-    logger,
-  });
-
-  const precinctId = isCardlessVoterAuth(auth)
-    ? auth.user.precinctId
+  const authStatusQuery = getAuthStatus.useQuery(electionHash);
+  const authStatus = authStatusQuery.isSuccess
+    ? authStatusQuery.data
+    : InsertedSmartCardAuth.DEFAULT_AUTH_STATUS;
```
jonahkagan, line 304: Any particular reason to use a default status vs just rendering `null` while loading?

apps/admin/backend/src/server.ts
jonahkagan: Oh nice catch

apps/mark/backend/src/app.ts
arsalansufi: Yes very true! :tada:

libs/auth/src/inserted_smart_card_auth_with_memory_card.ts
arsalansufi: Yes good questions!

The motivation for both of the above is actually similar. I've been thinking about the case that someone triggers an action in a valid state but then does something like removing their card right after, switching to an invalid state. I wouldn't want to trigger the error boundary in that case.

I know this may not be likely in practice since the timing would have to be just right, but the fact that all these methods depend not only on code but the status of the card in the card reader makes me lean toward treating errors as user errors instead of code errors that trigger the error boundary.

For what it's worth, I have been able to trigger the error boundary by updating the `checkPin` method to assert that `this.authStatus.status === 'checking_passcode'` and removing my card right after entering the last digit of my PIN:

https://user-images.githubusercontent.com/12616928/218866133-7549f67f-8e68-4ffc-94bf-a5f792ca02b3.mov

apps/mark/backend/src/app.ts
arsalansufi: Hmm interesting question. I do like having these checks exist together as opposed to split across the codebase. Also, once we start gating the backend more consistently, we'll have API methods that don't read or write card data but that also need to check if the user is logged in. For those, we'll have to put the logged-in check here in the API layer, so might as well do so consistently

apps/mark/frontend/src/app_root.tsx
arsalansufi: So we actually do render `null` while loading auth status:

https://github.com/votingworks/vxsuite/blob/2b546fb4fe48345b3301970b8d0205b7dec6a264/apps/mark/frontend/src/app_root.tsx#L644-L646

There are just a lot of pieces of data and hooks over the next few lines that use `authStatus`, and I didn't want to have to check if `authStatus` exists in all of them. And I can't move the `return null` up before the hooks because the number of hooks has to be consistent across renders.

libs/auth/src/inserted_smart_card_auth_with_memory_card.ts
jonahkagan: Ok cool that makes sense. Maybe leave some documentation of this approach somewhere so we preserve it over time? Seems like something that could be easily forgotten

apps/mark/backend/src/app.ts
jonahkagan: Sounds good!

apps/mark/frontend/src/app_root.tsx
jonahkagan: Gotcha, I didn't read closely, whoops. That makes sense. Hopefully the hooks will go away in time

apps/mark/frontend/src/pages/replace_election_screen.tsx
```diff
@@ -35,33 +30,20 @@ export function ReplaceElectionScreen({
...
 }: Props): JSX.Element {
   const { election, electionHash } = electionDefinition;
-  const makeCancelable = useCancelablePromise();
-  const [cardElectionDefinition, setCardElectionDefinition] =
-    useState<ElectionDefinition>();
-
-  useEffect(() => {
-    void (async () => {
-      setCardElectionDefinition(
-        await makeCancelable(getElectionDefinitionFromCard())
```
arsalansufi, line 51: I no longer need to wrap the call to `getElectionDefinitionFromCard` in a `makeCancelable` because the `AppRoot` component is always mounted, unlike the `ReplaceElectionScreen` component, which could unmount in the middle of this request if the election manager card was removed

apps/mark/frontend/src/pages/poll_worker_screen.tsx
arsalansufi: Ah nice great suggestion! Updated ✅

libs/auth/src/inserted_smart_card_auth_with_memory_card.ts
arsalansufi: Added a README and documented there: https://github.com/votingworks/vxsuite/pull/3011/commits/b7f9a9aa72033030590c7f7566225d3c9a925750

arsalansufi:

apps/mark/frontend/src/app_root.tsx
```diff
@@ -619,6 +630,32 @@ export function AppRoot({
...
+            if (result.isOk() && result.ok() !== undefined) {
+              setScannerReportDataToBePrinted(result.ok());
+            }
+          },
+        });
+      } else if (!isPollWorkerAuth(newAuthStatus)) {
+        setScannerReportDataToBePrinted(undefined);
+      }
+    }
+  );
```
jonahkagan, line 677: Now that I'm seeing this again, I'm wondering why reading the report data is a mutation rather than a query. I should have asked before but it didn't occur to me until now.

arsalansufi: Mm that is a good catch - a query (with no caching so that we don't have to worry about clearing when a card is removed) probably makes more sense. Let me update

libs/auth/README.md
```diff
@@ -0,0 +1,38 @@
...
+  because the checks are dependent on the context in which they're used (e.g.
+  `readCardData`) should leave those checks to backend consumers. Those backend
+  consumers should also return an error `Result` instead of throwing.
+
+The motivation for not throwing errors and so often returning an error `Result`,
+which we typically reserve for expected errors / user errors, is that all these
+methods depend not only on code but the status of the card in the card reader.
+Someone could trigger an action in a valid state but then remove their card
+right after, switching to an invalid state. We wouldn't want this to trigger the
+frontend error boundary, which throwing errors on the backend typically does.
```
jonahkagan, line 38: Maybe add a pointer to this documentation from the relevant lines of code so that people know to read this before changing it?

arsalansufi: Since there are so many potential lines to reference, I added high-level comments to the auth classes and made myself a code owner for libs/auth: https://github.com/votingworks/vxsuite/pull/3011/commits/df99d6d42253ec653e7a33a002cbc116095e8319

apps/mark/frontend/src/api.ts
```diff
@@ -63,6 +71,34 @@ export const getAuthStatus = {
...
+      this.queryKey(),
+      async () => {
+        const result = await apiClient.readElectionDefinitionFromCard({
+          electionHash,
+        });
+        const electionData = result.ok();
+        const electionDefinition = electionData
+          ? safeParseElectionDefinition(electionData).ok()
+          : undefined;
+        return ok<Optional<ElectionDefinition>, Error>(electionDefinition);
```
arsalansufi, line 93: I moved this election definition parsing logic into the API client to avoid having to duplicate it, though I suppose I could move it even further upstream to the backend itself, in the spirit of avoiding substantial logic in the API client. Can go ahead and do that

apps/mark/frontend/src/pages/admin_screen.tsx
```diff
@@ -54,11 +56,22 @@ export function AdminScreen({
...
+    getElectionDefinitionFromCard.useQuery(electionHash, {
+      // Disable automatic fetching and only allow manual fetching through .refetch()
+      enabled: false,
+    });
+
+  async function loadElection() {
+    const { data } = await electionDefinitionFromCardQuery.refetch();
+    assert(data !== undefined);
+    const electionDefinitionFromCard = data.ok();
+    // TODO: Handle case that electionDefinitionFromCard is undefined, e.g. because it's malformed
```
arsalansufi, line 71: This isn't a regression - we already weren't handling this case. I'm just taking note

apps/mark/frontend/src/app_root.tsx
arsalansufi: Required some thinking / exploration of the react-query API but got things working!

apps/mark/frontend/src/api.ts
arsalansufi: And done: https://github.com/votingworks/vxsuite/pull/3011/commits/fd8ed5b2cfd79c0139f7181302fc04da7da198b7

arsalansufi:
@jonahkagan This is ready again!

apps/mark/frontend/src/api.ts
```diff
@@ -63,6 +63,22 @@ export const getAuthStatus = {
...
+    return ['getScannerReportDataFromCard'];
+  },
+  useQuery(electionHash?: string) {
+    const apiClient = useApiClient();
+    return useQuery(
+      this.queryKey(),
+      () => apiClient.readScannerReportDataFromCard({ electionHash }),
+      // Don't cache this since caching would require invalidation in response to external
+      // circumstances, like card removal
+      { cacheTime: 0, staleTime: 0 }
```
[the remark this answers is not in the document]
jonahkagan: That makes sense.

With this configuration, does the query refetch on rerenders? Or does it give the "fetch once on mount" behavior that we were going for?

apps/mark/frontend/src/pages/replace_election_screen.tsx
```diff
@@ -30,31 +30,48 @@ export function ReplaceElectionScreen({
...
+  const electionDefinitionFromCard = electionDefinitionFromCardQuery.data?.ok();
 
   useEffect(() => {
     const muted = screenReader.isMuted();
     screenReader.mute();
     return () => screenReader.toggleMuted(muted);
   }, [screenReader]);
 
-  if (!electionDefinitionFromCard) {
+  if (!electionDefinitionFromCardQuery.data) {
```
jonahkagan, line 48: Generally I like to use `!query.isSuccess` to be more precise about what I'm checking, since the `data` might well be falsy for some queries.

apps/mark/frontend/src/pages/replace_election_screen.tsx
```diff
@@ -30,31 +30,48 @@ export function ReplaceElectionScreen({
...
-  electionDefinitionFromCard,
   machineConfig,
   screenReader,
   unconfigure,
-}: Props): JSX.Element {
+}: ReplaceElectionScreenProps): JSX.Element {
   const { election, electionHash } = electionDefinition;
+  const electionDefinitionFromCardQuery =
+    getElectionDefinitionFromCard.useQuery(electionHash);
+  const electionDefinitionFromCard = electionDefinitionFromCardQuery.data?.ok();
```
jonahkagan, line 40: maybe move this down to L59 to avoid the `data?` check?

apps/mark/backend/src/app.ts
```diff
@@ -54,7 +60,18 @@ function buildApi(auth: InsertedSmartCardAuthApi) {
...
+      const electionDefinition = electionData
+        ? safeParseElectionDefinition(electionData).ok()
+        : undefined;
+
+      if (!electionDefinition) {
+        // While we could provide more specific error messages for different error cases, the
+        // frontend doesn't need that much detail
+        return err(new Error('Unable to read election definition from card'));
+      }
+      return ok(electionDefinition);
```
jonahkagan, line 74: Are we using the error case at all? If we're just checking on the frontend that either `electionDefinition` is ok or not rather than using the error message, we could just return `ElectionDefinition | undefined`

apps/mark/frontend/src/api.ts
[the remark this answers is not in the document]
arsalansufi: Verified through manual testing that we aren't refetching on re-renders! Just mounts (and re-mounts)

apps/mark/backend/src/app.ts
arsalansufi: This comment got me thinking about error handling holistically again since I do ultimately want to introduce a consistent pattern for auth checks via an auth decorator. And thus far, I've been returning error `Result` objects, e.g.

```ts
if (authStatus.status !== 'logged_in') {
  return err(new Error('User is not logged in'));
}
```

This comment is making me realize that requiring every API endpoint to return a `Result` object could become cumbersome for the UI.

An alternative would be to throw a special authorization error on the server that's converted to a 401, which the react-query query client could then globally special case and not trigger the error boundary for.

I've added an item to my miscellaneous cleanup issue to revisit this and am going to leave this specific case as is for now, still using a `Result`.

Issue #3013, opened 2023-02-15
Remove YesNoContest.partyId from VXF election def


jonahkagan:
Yes/no contests (aka ballot measures) are not actually associated with a party. Party associations only apply to candidate contests and are used for primary elections. VXF supports associating a partyId with ballot measures by accident, but CDF does not support this.

adghayes:
Closed by #3014.

PR #3014, opened 2023-02-15
Remove party associations for ballot measure (yes/no) contests

jonahkagan:
## Overview
Task: https://github.com/votingworks/vxsuite/issues/3013

Removes `YesNoContest.partyId` and deals with the consequences:
- Updating election fixtures to remove `partyId` from yes/no contests
- Updating logic that checked/filtered on `contest.partyId`

_Review by commits_

## Demo Video or Screenshot
N/A

## Testing Plan 
- Existing automated tests

## Checklist


apps/admin/frontend/src/app_write_in_flows.test.tsx
```diff
@@ -172,7 +172,7 @@ test('manual write-in data end-to-end test', async () => {
     .closest('tr')!;
   expect(
     within(precinct1SummaryRow).getByTestId('numBallots').textContent
-  ).toEqual('18');
+  ).toEqual('11');
```
jonahkagan, line 175: I couldn't make sense of this test well enough to understand if these changes are reasonable or indicate a bug

jonahkagan:
Here ya go: https://github.com/votingworks/vxsuite/issues/3015

apps/admin/frontend/src/app_write_in_flows.test.tsx
adghayes: I think this points to an underlying bug where our method of determining the total number of ballots in a precinct based on the total number of ballots in each contest is flawed. 

Basically, there should be 18 ballots in this precinct because we have 11 ballots from the mammal party and 7 ballots from the fish party. With the way we're calculating the total number, however, it sees the shared race (the now non-partisan races) and thinks they're the same ballot style, so it takes the maximum, 11. 

@jonahkagan this could take some thinking to get right so I'm not sure how you want to handle. Feel free to create a ticket and tag @adghayes and @carolinemodic, who may have more context. 

https://github.com/votingworks/vxsuite/blob/d43d42816b6a4d82d8491b9aa60448d5b70c93fc/apps/admin/frontend/src/utils/external_tallies.ts#L68-L122

jonahkagan: Ok, since this is a known bug now, I'm gonna not block this PR on it

PR #3021, opened 2023-02-16
Remove unused frontend auth code

arsalansufi:

## Overview

Now that we've moved auth to the backend on all machines, we can remove unused frontend auth code!

## Testing

- [x] Automated tests
- [x] Manual testing

## Checklist


libs/ui/src/export_logs_modal.test.tsx
```diff
@@ -30,6 +32,18 @@ const fileSystemEntry: KioskBrowser.FileSystemEntry = {
   mtime: new Date(),
 };
 
+const systemAdministratorAuthStatus: DippedSmartCardAuth.SystemAdministratorLoggedIn =
+  {
+    status: 'logged_in',
+    user: fakeSystemAdministratorUser(),
+    programmableCard: { status: 'no_card' },
+  };
```
arsalansufi, line 40: I ultimately didn't replace the old `Dipped.fakeSystemAdministratorAuth()`, etc. utils with analogs for the new auth types. I think the old auth types really benefited from the helpers because of all of the functions on the types, e.g. `systemAdministratorAuth.card.programUser`. The new auth types are simple enough to construct as needed. Given some of our new API mocking patterns, there are fewer points where we need to construct mock auth statuses anyway (usually just in the API mocks themselves)

apps/mark/frontend/test/helpers/timers.ts
```diff
@@ -0,0 +1,14 @@
+import { waitFor } from '@testing-library/react';
+import { advanceTimers as advanceTimersBase } from '@votingworks/test-utils';
+import { AUTH_STATUS_POLLING_INTERVAL_MS } from '@votingworks/ui';
+
+export function advanceTimers(seconds = 0): void {
```
arsalansufi, line 5: Moved these few pieces out of `test/helpers/smartcards.ts` and removed the rest of `test/helpers/smartcards.ts`, which was all for voter cards

apps/mark/integration-testing/cypress/support/e2e.js
```diff
@@ -17,17 +17,19 @@ import '@testing-library/cypress/add-commands';
...
-  '000000'
-);
-const POLL_WORKER_CARD_DATA = makePollWorkerCard(
-  electionSampleDefinition.electionHash
-);
+const ELECTION_MANAGER_CARD_DATA = {
+  t: 'election_manager',
+  h: electionSampleDefinition.electionHash,
+  p: '000000',
+};
```
arsalansufi, line 27: Will likely restore some form of abstraction here once I've finished Java Card implementation

libs/ui/src/export_logs_modal.test.tsx
jonahkagan: Sounds good to me!

libs/utils/src/auth.ts
```diff
@@ -0,0 +1,45 @@
+import { DippedSmartCardAuth, InsertedSmartCardAuth } from '@votingworks/types';
```
jonahkagan, line 1: How did you choose `libs/utils` vs `libs/types` for these? Maybe doesn't super matter since often both libs are used together, but seems like type guards could live with the associated types

jonahkagan (approved):

libs/utils/src/auth.ts
arsalansufi: Yeah good q! I was torn between these two options. I went with `libs/utils` following the simple rule that `libs/types` should only contain types (though I know that's not entirely the case even today)

I do like the idea of colocating logic though, so I think I'll move the type guards to `libs/types` after all

arsalansufi: Ah! Actually ran into a hiccup here. The tests for the type guards use `@votingworks/test-utils`. Moving the type guards and their corresponding tests to `@votingworks/types` would create a circular dependency between `@votingworks/test-utils` and `@votingworks/types`. I could of course just not use the test utils, but I think the pros of keeping the type guards in `@votingworks/utils` now outweigh the pros of moving them to `@votingworks/types`

jonahkagan: Classic dependency cycle hiccup. Sounds good

PR #3023, opened 2023-02-21
Modify react-query query client default options

arsalansufi:
## Overview

With the shift to react-query and our specific default options (notably `retry: false` and `useErrorBoundary: true`), we often see the frontend error boundary get triggered when the backend is spinning up or restarting due to changes. We've seen this in dev, but I suspect it could affect prod, too.

<table><tr><td>

</table></tr></td>

This PR modifies our react-query query client default options with a custom retry function in place of `retry: false` to prevent this.

## Testing

- [x] Added unit tests for custom retry function
- [x] Tested initial backend startup, backend restart, and true backend failure manually

## Checklist

### DIFF
--- a/libs/ui/src/react_query.ts
+++ b/libs/ui/src/react_query.ts
@@ -2,6 +2,25 @@
 import type { DefaultOptions } from '@tanstack/react-query';
 
 /**
+ * A custom retry function
+ *
+ * A 504 in VxSuite usually just indicates that the backend is starting up (or restarting), in
+ * which case we should retry, but still with a bound in case the backend is completely failing to
+ * start up.
+ *
+ * 1 original attempt + 4 retry attempts
+ * = 0 + 1 + 2 + 4 + 8 seconds (because of exponential backoff)
+ * = 15 seconds of total wait time
+ *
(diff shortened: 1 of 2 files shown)


libs/ui/src/react_query.ts
```diff
@@ -1,6 +1,25 @@
...
+ *
+ * 1 original attempt + 4 retry attempts
+ * = 0 + 1 + 2 + 4 + 8 seconds (because of exponential backoff)
+ * = 15 seconds of total wait time
+ *
+ * Otherwise, don't retry. Because everything is local, we don't expect intermittent errors.
+ */
+export function shouldRetry(failedRetryCount: number, error: unknown): boolean {
+  const isBackendLikelyStartingUp =
+    error instanceof Error && error.message.includes('504');
```
arsalansufi, line 19: I wanted to make this condition a little bit more specific, e.g. by checking for `"ECONNREFUSED"` in the error message, but that piece of the backend error message isn't propagated through the frontend proxy

This feels like a reasonable starting point, though

jonahkagan (approved):
Safe to try!

PR #3024, opened 2023-02-21
fix(grout): Fix latent bug in serializing undefined

jonahkagan:
## Overview
The helper function to determine if a value is a JSON built-in value was mistakenly identifying `undefined` as a built-in, which it's not (it only exists in JS value space, not JSON). This didn't actually impact the end logic that used this function, but that was just by chance. Fixing this to prevent any future bugs that might occur from this being wrong.

## Demo Video or Screenshot
N/A

## Testing Plan
Existing tests still pass

## Checklist

- [ ] I have added


arsalansufi (approved):
Nice catch!

PR #3029, opened 2023-02-22
Remove PINs from user models and consistently use "PIN" instead of "passcode"

arsalansufi:

## Overview

This PR removes PINs from user models such that current PINs never leave the backend, except when a new card has been programmed or a PIN has been reset and we need to display the new PIN to the user. (Back when auth was managed by the frontend, we always needed PINs there.)

This PR also updates code and user-facing copy to consistently use the term "PIN" instead of "passcode" / "code" / "security code".

## Testing

- [x] Automated tests
- [x] Manual testing of valid PIN entry
- [x] Manual testing of invalid PIN entry
- [x] Manual testing of `SKIP_PIN_ENTRY` feature flag
- [x] Manual testing of card programming

## Checklist


libs/ui/src/unlock_machine_screen.tsx
```diff
@@ -76,17 +72,6 @@ export function UnlockMachineScreen({
...
-  useEffect(() => {
-    function bypassPinEntry() {
-      checkPin(auth.user.passcode);
-    }
-    if (isFeatureFlagEnabled(BooleanEnvironmentVariableName.SKIP_PIN_ENTRY)) {
-      bypassPinEntry();
-    }
-    // Run this hook once on mount
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
```
arsalansufi, line 88: I add back support for skipping PIN entry in a subsequent commit

libs/auth/src/inserted_smart_card_auth_with_memory_card.ts
```diff
@@ -232,11 +236,19 @@ export class InsertedSmartCardAuthWithMemoryCard
...
+                    )
+                      ? { status: 'logged_in', user }
+                      : { status: 'checking_passcode', user };
+                  }
+                  if (user.role === 'election_manager') {
+                    return isFeatureFlagEnabled(
+                      BooleanEnvironmentVariableName.SKIP_PIN_ENTRY
+                    )
+                      ? { status: 'logged_in', user }
+                      : { status: 'checking_passcode', user };
```
arsalansufi: Even though the logic is identical for sys admins and election managers, when I group everything under one case, e.g.

```ts
if (user.role === 'system_administrator' || user.role === 'election_manager') {
  return isFeatureFlagEnabled(
    BooleanEnvironmentVariableName.SKIP_PIN_ENTRY
  )
    ? { status: 'logged_in', user }
    : { status: 'checking_passcode', user };
}
```

I get the following type error:

type-error

Hence the separate cases when it seems like I could just use one

jonahkagan (approved):

apps/admin/frontend/src/components/smartcard_modal/status_message.tsx
```diff
@@ -15,6 +15,7 @@ export type SmartcardAction = 'Program' | 'PinReset' | 'Unprogram';
 
 interface SmartcardActionComplete {
   action: SmartcardAction;
+  newPin?: string;
   role: UserRole;
   status: 'Success' | 'Error';
```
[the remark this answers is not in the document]
arsalansufi: True! Though I think I'll leave as is for consistency with the `SmartcardActionInProgress` type below:

https://github.com/votingworks/vxsuite/blob/ea07e4264607270396018c090d502f102f1d877b/apps/admin/frontend/src/components/smartcard_modal/status_message.tsx#L26

PR #3030, opened 2023-02-22
Fix for automatically computing total ballots from a contest tally

jonahkagan:
Creating a PR so that the branch will stick around, even though we aren't going to use this code


jonahkagan:
Closing in favor of https://github.com/votingworks/vxsuite/issues/3031

PR #3032, opened 2023-02-22
Separate Images in CDF Exports

adghayes:
## Overview

This is a follow-up to #3017 and part of #2360. The main change (first commit) is moving to exporting our cast vote record report as a directory containing:
- singular JSON cast vote record report in CDF format
- subdirectory for linked ballot images
- subdirectory for ballot layouts. 

@jonahkagan this PR is also some related cleanup work:
- moving to the altered JSON schema model and
- renaming the `validate` module to `canonicalize` because I agree that makes more sense

Note that all of these chances are still behind an environment variable to export, so it doesn't affect the default flow.

Example export, zipped (we are not zipping them in the product): [TEST__machine_000__4_ballots__2023-02-22_15-25-47.zip](https://github.com/votingworks/vxsuite/files/10817040/TEST__machine_000__4_ballots__2023-02-22_15-25-47.zip)


libs/backend/src/scan/cast_vote_records/build_report_metadata.ts
```diff
@@ -146,7 +146,7 @@ interface BuildCastVoteRecordReportParams {
  * necessary to make sense of those CVRs. We generate this separately because
  * we want to write the CVRs to file separately, one-by-one.
  */
-export function buildCastVoteRecordReport({
+export function buildCastVoteRecordReportWithoutCastVoteRecords({
```
jonahkagan, line 149: Maybe call this `buildCastVoteRecordReportMetadata` to match with the file name? It takes a bit of thinking to figure out what it means to build a cast vote record report without cast vote records.

libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -118,181 +205,201 @@ async function* getCastVoteRecordGenerator({
...
+      isTestMode,
+      batchInfo,
+    });
 
-  // yield last element without trailing comma
-  if (previous) {
-    yield previous;
-  }
+  return ok(
+    streamifyCastVoteRecordReport({
```
jonahkagan, line 240: Is there a reason we're still building the report metadata and the cast vote records and combining them in a stream if we're just doing it all in memory? Seems like it might be simpler to just build the full report object and then stringify.

libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -142,7 +142,7 @@ export async function buildCastVoteRecordReport({
     frontNormalizedFilename: sideOneFilename,
     backNormalizedFilename: sideTwoFilename,
   } of resultSheetGenerator) {
-    const validationResult = validateSheetInterpretation([sideOne, sideTwo]);
+    const validationResult = canonicalizeSheet([sideOne, sideTwo]);
```
jonahkagan, line 145: rename `validationResult` to `canonicalizationResult`?

libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -151,14 +151,14 @@ export async function buildCastVoteRecordReport({
       });
     }
 
-    const validatedSheet = validationResult.ok();
-    const [frontFilename, backFilename] = validatedSheet.wasReversed
+    const canonicalizedSheet = validationResult.ok();
+    const [frontFilename, backFilename] = canonicalizedSheet.wasReversed
       ? [sideTwoFilename, sideOneFilename]
       : [sideOneFilename, sideTwoFilename];
```
jonahkagan, line 157: Could consider passing in the file names with the page interpretations to `canonicalizeSheet` in order to avoid having to return the `wasReversed` flag

apps/scan/backend/src/backup.ts
```diff
@@ -107,24 +105,27 @@ export class Backup {
         exportCastVoteRecordsAsNdJson({ store: this.store })
       );
     } else {
+      // eslint-disable-next-line @typescript-eslint/await-thenable
```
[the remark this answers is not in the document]
adghayes: I was experiencing a weird linting thing in my dev environment, I think this is removed in a later commit but if not I'll remove it.

libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -118,181 +205,201 @@ async function* getCastVoteRecordGenerator({
...
+    isTestMode,
+    resultSheetGenerator,
+    ballotPageLayoutsLookup,
+    definiteMarkThreshold,
+    batchInfo,
+    imageOptions: {
+      which: whichImages,
+      directory: CVR_BALLOT_IMAGES_SUBDIRECTORY,
+    },
+    sheetExportCallback: async ({ validatedSheet, imageFilenames }) => {
```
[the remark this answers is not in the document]
adghayes: Sounds good to me. I was over-prioritizing only making one pass, but it would be easier if we don't.

libs/backend/src/scan/cast_vote_records/build_report_metadata.ts
adghayes: Sounds good to me.

libs/backend/src/scan/cast_vote_records/export.ts
adghayes: Our max string size is like 0.5GB. With inline ballot images, we're limited to like 1,000 ballots if all ballots have images (scaled down). Without inline ballot images, that goes up to way higher, like 100,000. So I thought about just not worrying about it. But I think it's conceivable that we change our CVRs to be even more information dense in the future and I didn't want to bank on it all fitting in the one string.

libs/backend/src/scan/cast_vote_records/export.ts
adghayes: dang, thanks

libs/backend/src/scan/cast_vote_records/export.ts
adghayes: will consider

libs/backend/src/scan/cast_vote_records/export.ts
jonahkagan: Ah ok, so there's a max string size that's diff from the process memory limit? That's helpful context. I think it's worth documenting some of this thinking in explanatory comments, since it's kind of an unconventional halfway between full streamed CVRs and in-memory CVRs

Or just going full streaming would be another option.

adghayes: Exactly, max string size is much more of a limit than what we can potentially have in an array or an object. Full streaming was hard because I was doing the sheet callback, but now that I'm making two passes, I'll see if it's easy to do full streaming.

adghayes:
@jonahkagan I have incorporated all of your feedback. I did think it made sense to go all on in generation / streaming, but that introduced a little complexity into the error handling. You can use a generator for a `Result` but it gets a little clunky, so I chose to just `throw` on validation errors.

libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -400,5 +343,66 @@ export async function exportCastVoteRecordReportToUsbDrive({
...
+    backNormalizedFilename: sideTwoFilename,
+  } of resultSheetGenerator()) {
+    const canonicalizationResult = canonicalizeSheet([sideOne, sideTwo]);
+
+    if (canonicalizationResult.isErr()) {
+      return err({
+        type: 'invalid-sheet-found',
+        message: describeSheetValidationError(canonicalizationResult.err()),
+      });
+    }
```
jonahkagan, line 426: Should be impossible to fail canonicalization on the second pass, right?

jonahkagan (approved):

libs/backend/src/scan/cast_vote_records/export.ts
adghayes: Should be! I think it's OK to have this redundancy though, in case the order of operations gets moved around.

PR #3033, opened 2023-02-23
VxAdmin: Remove feature to import external tally from SEMS

jonahkagan:
## Overview

This feature is no longer needed, so we don't want to keep maintaining it.

_Review by commits_

## Demo Video or Screenshot
N/A

## Testing Plan
- Existing automated tests

## Checklist

- [ ] I have added

- [x] I have added the "user_facing_change" label to this PR to automate an


libs/types/src/tallies.ts
```diff
@@ -59,8 +59,10 @@ export interface ExternalTally {
   readonly numberOfBallotsCounted: number;
 }
 
+// Why have a source type with only one value? We used to support importing
+// external tallies from state election management systems and had more source
+// types.
 export enum ExternalTallySourceType {
-  SEMS = 'sems',
   Manual = 'manual-data',
 }
```
jonahkagan, line 68: Currently, all of the code for external tallies is generalized (since we had different external tally source types). Now, the only supported external tallies are manual. We could go change all of that code to be specific to manual tallies (e.g. changing the language used, removing the checking of the source type), which would probably make things a bit clearer.

I'm not sure if it's worthwhile for two reasons:
- I'm not sure how much clearer it would be
- I'm not sure if we'll want to support importing other types of external tallies anytime soon

adghayes: I think this is something we should do when we move that data to the backend.

adghayes (approved):
LGTM

PR #3035, opened 2023-02-23
Rethink card abstraction

arsalansufi:

## Overview

I'd initially planned to create two implementations for each auth API, one for memory cards (as a stopgap) and one for Java Cards (the future):

- `DippedSmartCardAuthWithMemoryCard` implements `DippedSmartCardAuthApi`
- `DippedSmartCardAuthWithJavaCard` implements `DippedSmartCardAuthApi`
- `InsertedSmartCardAuthWithMemoryCard` implements `InsertedSmartCardAuthApi`
- `InsertedSmartCardAuthWithJavaCard` implements `InsertedSmartCardAuthApi`

I'm now taking a different approach with only one implementation for each auth API:

- `DippedSmartCardAuth` implements `DippedSmartCardAuthApi`
- `InsertedSmartCardAuth` implements `InsertedSmartCardAuthApi`

And I'm moving card variations a level deeper via a new card API that's more compatible with Java Cards but also still compatible with memory cards. The current card API is a bit too low level and specific to our memory card implementation, for example, with its methods for short and long values.

With the new card API, we can do as follows:

- `MemoryCard` implements `Card`
- `JavaCard` implements `Card`

Both `DippedSmartCardAuth` and `InsertedSmartCardAuth` will take a `Card` in their constructor.

*Note: `Card` is referring to the new API and not the existing one with the same name. Similarly, `MemoryCard` is not to be confused with the existing implementation of the old card API that we previously used in tests.

To minimize confusion, I've moved all the old card types and interfaces into a `legacy/` directory in `libs/auth`.

## Testing

- [x] Automated tests
- [x] Manually tested every machines' auth flows
- [x] Manually tested card programming on VxAdmin
- [x] Manually tested report writing, reading, and clearing on VxScan and VxMark

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Still on my radar for later


libs/utils/src/Card/memory_card.ts
```diff
@@ -1,160 +0,0 @@
...
-  ShortAndLongValues,
-} from '@votingworks/types';
-import { z } from 'zod';
-
-/* eslint-disable @typescript-eslint/require-await */
-
-/**
- * Implements the `Card` API with an in-memory implementation.
- */
-export class MemoryCard implements Card {
```
arsalansufi, line 16: I got rid of this `MemoryCard` implementation of the old card API because:

1. We no longer use it anywhere (it was previously used in tests, but application-level tests now mock auth).
2. Its name can be quite confusing. It's not referring to an actual memory card. It's referring to a mock of a card that uses computer memory.

libs/auth/src/inserted_smart_card_auth.ts
```diff
@@ -184,7 +179,7 @@ export class InsertedSmartCardAuth implements InsertedSmartCardAuthApi {
 
   async clearCardData(): Promise<Result<void, Error>> {
     try {
-      await this.card.writeLongUint8Array(Uint8Array.of());
+      await this.card.writeData(Buffer.from([]));
```
jonahkagan, line 182: Should the Card API have a `clearData` method? I'm thinking that since it has a `clearUserAndData` method, the concept of clearing data is already defined behind the Card abstraction, so it's possible that this way of clearing data that the consumer is doing here (writing an empty buffer) might not align with the way that `clearUserAndData` is doing it.

In other words, who gets to decide how clearing data is implemented - the Card API or its consumer?

Related: there's a bit of a conflation of an empty buffer and "no data". Are we good with that? I don't think we have any reason we'd want to support explicitly writing an empty buffer as valid data, but just thought I'd check.

apps/admin/frontend/src/components/smartcard_modal/smartcard_modal.tsx
```diff
@@ -45,7 +46,13 @@ export function SmartcardModal(): JSX.Element | null {
           smartcardType: 'system-administrator',
         });
       let contents: JSX.Element;
-      if (auth.programmableCard.programmedUser) {
+      if (
+        auth.programmableCard.programmedUser &&
+        // Writing user data may finish before writing additional data (e.g. the election
+        // definition). Don't transition from the card programming view to the card details view
+        // until all data has been written
+        !isSmartcardProgrammingInProgress(actionStatus)
```
jonahkagan, line 54: I was wondering about this when I saw that it was two separate writes. Is there a way that we could hide this in within the auth lib somehow so that the card programming action is atomic from the consumer's point of view?

jonahkagan (commented):
I like this layering - very clear!

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -51,10 +51,8 @@ type AuthAction =
  *
  * See the libs/auth README for notes on error handling
  */
-export class DippedSmartCardAuthWithMemoryCard
-  implements DippedSmartCardAuthApi
-{
-  private authStatus: DippedSmartCardAuth.AuthStatus;
+export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
```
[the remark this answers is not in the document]
arsalansufi: Yes good q! There is, mainly for mocking purposes in application-level tests. If we were to have app backends use `DippedSmartCardAuth` directly instead of `DippedSmartCardAuthApi`, we'd have to mock not only the public methods but also all the internal fields and methods of `DippedSmartCardAuth`:

mocking

Less importantly, there is also something nice about being able to see the public interface at a glance

libs/auth/src/card.ts
```diff
@@ -0,0 +1,67 @@
...
+  getCardStatus(): Promise<CardStatus>;
+
+  checkPin(pin: string): Promise<CheckPinResponse>;
+
+  writeUser(
+    input:
+      | { user: SystemAdministratorUser; pin: string }
+      | { user: ElectionManagerUser; pin: string }
+      | { user: PollWorkerUser }
+  ): Promise<void>;
```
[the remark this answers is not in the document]
arsalansufi: By unspecified user data blob, do you mean something like `user: User` or even more freeform, like `user: { [key: string]: string }`?

Regardless, I can share the reasoning behind what I've landed on!

1. I wanted parity between what `getCardStatus` returns and what `writeUser` expects as input. Having both work with `User` objects makes for a consistent interface.
2. With Java Cards, we'll be packaging user data into an x509 certificate. I want the transformation from `User` to certificate to be opaque to consumers.
3. If you were just referring to `user: User` as an alternative, I wanted to explicitly exclude `CardlessVoterUser` and also enforce which user roles require PINs.

libs/auth/src/dipped_smart_card_auth.ts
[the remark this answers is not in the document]
jonahkagan: Gotcha, thanks!

libs/auth/src/card.ts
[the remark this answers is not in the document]
jonahkagan: Cool, that all makes sense!

libs/auth/src/inserted_smart_card_auth.ts
arsalansufi: Ooh good call about ensuring consistency between standalone clearing of data and clearing of both the user and data! Will add a `clearData` method to the card API.

To the second point, as far as I can tell, under the hood, an empty buffer and "no data" are equivalent. `Buffer.from([])` basically means no bytes of data.

apps/admin/frontend/src/components/smartcard_modal/smartcard_modal.tsx
arsalansufi: Yeah this piece is interesting 🤔

While we can make this seem atomic from the auth lib interface perspective, e.g. `writeUserAndData` or `programUser`, under the hood is a different story. With our current memory cards, we do ultimately store data on the card in one continuous concatenated blob. But with Java Cards, user data and other data will live in separate data slots and require separate sequential commands to the card.

We can still make things seem atomic in the Java Card world though, by tweaking how `getCardStatus` operates. Instead of having it actually read from the card on every call, it could manage a cache of the current card data and pull from that instead of the underlying card. Of course, all the challenges with caching hold, but I've been considering this anyway since, unlike on our current memory cards, retrieving user info from Java Cards will be more involved (we'll be retrieving an x509 certificate and parsing it with openssl).

Bringing this back to the current PR, I do like the idea of a single card API method for card programming so will at least make that change

jonahkagan: What if `getCardStatus` returned some sort of "processing" status while writes were in progress? Where would that lead?

arsalansufi: That seems like a very reasonable option! If it's okay, I'm going to hold off on that for now since I'd like to explore that change in the context of the Java Card implementation

I've at least removed the logic above from the smart card modal since the joint write of user and election data with memory cards is atomic again (as of https://github.com/votingworks/vxsuite/pull/3035/commits/a17960594cc076da4221ead12e9e2a27084f84c1)

arsalansufi:
@jonahkagan Ready for another look!

apps/admin/frontend/src/components/smartcard_modal/smartcard_modal.tsx
jonahkagan: Sounds good!

libs/auth/test/utils.ts
```diff
@@ -7,7 +7,7 @@ export function buildMockCard(): Card {
   return {
     getCardStatus: jest.fn(() => Promise.resolve({ status: 'no_card' })),
     checkPin: jest.fn(),
-    writeUser: jest.fn(),
+    program: jest.fn(),
```
jonahkagan, line 10: Maybe now `clearUserAndData` should be `unprogram` to match?

jonahkagan (approved):

libs/auth/test/utils.ts
arsalansufi: 👍 I went back and forth between `unprogram` and `clearUserAndData`, with the latter feeling more descriptive, but being consistent with `program` makes sense

PR #3036, opened 2023-02-24
feat(admin): Import CDF ballot definition

jonahkagan:
## Overview

Task: #2930 

Adds support for configuring VxAdmin with a CDF ballot definition file.
 
To implement this feature, we modify `safeParseElection` to try parsing a given election both as VXF and CDF, returning whichever one works.

_Review by commits_

## Demo Video or Screenshot

https://user-images.githubusercontent.com/530106/221712824-a1588ea5-67ea-41c1-8871-c9d463269ca0.mov

## Testing Plan
- Manual test
- Added some new frontend automated tests

## Checklist

- [ ] I have added

- [x] I have added the "user_facing_change" label to this PR to automate an


libs/types/src/election_utils.ts
```diff
@@ -600,17 +601,18 @@ function preprocessElection(value: unknown): unknown {
...
+  value: unknown
+): Result<Election, z.ZodError> {
+  return safeParse(ElectionSchema, preprocessElection(value));
+}
+
 /**
- * Parses `value` as an `Election` object. Equivalent to
- * `safeParse(Election, value)`.
+ * Parses `value` as an `Election` object. Supports both VXF and CDF. If given a
+ * string, will attempt to parse it as JSON first.
```
jonahkagan, line 614: So I'm pretty sure that after this change, not only does VxAdmin now support importing CDF, but the resulting `ElectionDefinition` will have the CDF raw JSON as its `electionData` (which will also be the data that is hashed). I think this means that the CDF will be saved in the db and propagated throughout the rest of the system (e.g. via smartcards/ballot packages). So maybe all of VxSuite now supports CDF ballot definitions? Am I missing something?

adghayes: - CDF will be saved to the DB

adghayes: My understanding is the same, after taking a look at the relevant code:

- CDF will be saved to the DB
- CDF will be saved to card. With the current cards there is the space concern but with @arsalansufi's new cards that is changing?
- CDF will be in ballot package
- CDF election hash will be encoded into the ballots

I don't know of anything we're missing!

adghayes (approved):
LGTM!

PR #3041, opened 2023-02-27
Move current election state to backend

adghayes:
In preparation for importing CDF CVRs, I've started moving some state, especially w.r.t CVR files, to the admin backend. In this PR, I move the state of the "Current Election" to the backend. 

@jonahkagan @eventualbuddha I learned that this will be challenging to do in a piecemeal fashion. It's easy to move the pieces over but difficult to test. The frontend currently has the `backend` abstraction which has a lot of its own state on its own for testing. We're moving away from mocking our backend fully, with `fetch-mock`, to only mocking responses and requests with the new testing framework. There are some places where it will be strange and confusing in the transition.

I recommend we do this in one fell swoop and, instead of merging this PR, I start working on moving all of the admin endpoints over to `grout`.

### DIFF
--- a/apps/admin/frontend/src/lib/backends/backends.test.ts
+++ b/apps/admin/frontend/src/lib/backends/backends.test.ts
@@ -8,12 +8,13 @@ import { fakeLogger } from '@votingworks/logging';
 import {
   ExternalTallySourceType,
   Id,
+  Optional,
   safeParse,
   safeParseJson,
   VotingMethod,
 } from '@votingworks/types';
 import { MemoryStorage } from '@votingworks/utils';
-import { typedAs } from '@votingworks/basics';
+import { assert, assertDefined, typedAs } from '@votingworks/basics';
 import fetchMock from 'fetch-mock';
 import { convertTalliesByPrecinctToFullExternalTally } from '../../utils/external_tallies';
 import { ElectionManagerStoreAdminBackend } from './admin_backend';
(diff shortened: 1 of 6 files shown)


apps/admin/frontend/src/app.test.tsx
```diff
@@ -586,8 +589,14 @@ test('tabulating CVRs', async () => {
...
   getByText('Mark Unofficial Tally Results as Official Tally Results?');
   const modal = await screen.findByRole('alertdialog');
+  mockApiClient.markResultsOfficial.expectCallWith().resolves();
   fireEvent.click(within(modal).getByText('Mark Tally Results as Official'));
 
+  // since isOfficial is fetched from the memory backend, not the mock api
+  // client, have to set its results as official and also invalidate the query
+  await backend.markResultsOfficial();
+  await queryClient.invalidateQueries();
+
```
adghayes, line 599: Maybe there's a way to avoid `invalidateQueries`, but in a piecemeal transition we will have these points where we have to update the `backend` manually.

apps/admin/frontend/src/lib/backends/backends.test.ts
```diff
@@ -32,113 +33,67 @@ function makeAdminBackend(): ElectionManagerStoreAdminBackend {
...
-      ).unsafeUnwrap();
-
-      if (typeof body.isOfficialResults === 'boolean') {
-        dbEntry.electionRecord = {
-          ...dbEntry.electionRecord,
-          isOfficialResults: body.isOfficialResults,
-        };
-      }
-
-      return { body: typedAs<Admin.PatchElectionResponse>({ status: 'ok' }) };
```
adghayes, line 98: Will have to do a lot of work to maintain this testing setup during a transition.

PR #3042, opened 2023-02-27
refactor(backend): use `jsonStream` to simplify CVR streams

eventualbuddha:
## Overview
This is some follow-up to #3032, which I was doing a light review of while it got merged. I found the `JSON.stringify(…).replace(…)` we had there to be pretty janky and thought we should do better. This is my attempt to do that.

## Checklist

- [ ] I have added

### DIFF
--- a/libs/backend/src/scan/cast_vote_records/export.ts
+++ b/libs/backend/src/scan/cast_vote_records/export.ts
@@ -11,7 +11,7 @@ import {
   SheetOf,
   unsafeParse,
 } from '@votingworks/types';
-import { err, ok, Result } from '@votingworks/basics';
+import { err, ok, Result, typedAs } from '@votingworks/basics';
 import { Readable } from 'stream';
 import {
   generateCastVoteRecordReportDirectoryName,
@@ -31,6 +31,14 @@ import { getUsbDrives } from '../../get_usb_drives';
 import { SCAN_ALLOWED_EXPORT_PATTERNS, VX_MACHINE_ID } from '../globals';
 import { BallotPageLayoutsLookup, getBallotPageLayout } from './page_layouts';
 import { buildCastVoteRecordReportMetadata } from './build_report_metadata';
+
+type WithArraysAsIterables<T> = {
+  [P in keyof T]: T[P] extends ReadonlyArray<infer U>
+    ? Iterable<U>
+    : T[P] extends ReadonlyArray<infer U> | undefined
+    ? Optional<Iterable<U>>
+    : T[P];
+};
 
 /**
  * Properties of each sheet that are needed to generate a cast vote record
@@ -235,10 +243,12 @@ export function getCastVoteRecordReportStream({
   });
 
   return Readable.from(
-    jsonStream({
-      ...castVoteRecordReportMetadata,
-      CVR: castVoteRecordGenerator,
-    })
+    jsonStream(
+      typedAs<WithArraysAsIterables<CVR.CastVoteRecordReport>>({
+        ...castVoteRecordReportMetadata,
+        CVR: castVoteRecordGenerator,
+      })
+    )
   );
 }
 
--- a/libs/backend/src/scan/cast_vote_records/export.ts
+++ b/libs/backend/src/scan/cast_vote_records/export.ts
@@ -192,27 +192,6 @@ function* getCastVoteRecordGenerator({
   }
 }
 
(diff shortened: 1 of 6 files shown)


libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -206,31 +207,10 @@ function* getCastVoteRecordReportGenerator({
...
-    yield `\n    ${JSON.stringify(previousCastVoteRecord)}\n  `;
-  }
-
-  // closing brackets to make JSON valid
-  yield ']\n}';
+  yield* jsonStream({
+    ...castVoteRecordReportMetadata,
+    CVR: castVoteRecordGenerator,
+  });
 }
```
jonahkagan, line 214: We can probably remove this function now and just inline this expression in its caller

libs/utils/src/json_stream.ts
```diff
@@ -0,0 +1,78 @@
...
+ * }
+ *
+ * const stream = jsonStream({ foo: [1, 2, 3], bar: generateLargeArray() });
+ * const writer = createWriteStream('output.json');
+ * for (const chunk of stream) {
+ *  writer.write(chunk);
+ * }
+ * ```
+ */
+export function* jsonStream(input: unknown, indent = 0): Generator<string> {
```
jonahkagan, line 28: What do you think about making a type for json streamable input values? I suggested one here: https://github.com/votingworks/vxsuite/pull/3017#discussion_r1113453460

libs/utils/src/json_stream.ts
```diff
@@ -0,0 +1,78 @@
...
+    const nextIndentString = ' '.repeat(nextIndent * 2);
+    yield '[';
+    let hasEntries = false;
+    for (const value of input) {
+      if (hasEntries) {
+        yield ',';
+      } else {
+        hasEntries = true;
+      }
+      yield `\n${nextIndentString}`;
```
jonahkagan, line 49: If `indent` is 0, should we skip line breaks?

jonahkagan (commented):
💯

libs/utils/src/json_stream.ts
eventualbuddha: Tried that, didn't seem to work for reasons that weren't totally clear:

image

```ts
type JsonableWithIterable =
  | null
  | boolean
  | string
  | JsonableWithIterable[]
  | Iterable<JsonableWithIterable>
  | Generator<JsonableWithIterable>
  | { [key: string]: JsonableWithIterable };
```

jonahkagan: Weird

libs/utils/src/json_stream.ts
jonahkagan: I guess I was just thinking if the caller specifies no indent, maybe they would want a compressed version with no line breaks. Similar to:

```js
> JSON.stringify([1, 2], null, 2)
'[\n  1,\n  2\n]'
> JSON.stringify([1, 2], null, 0)
'[1,2]'
```

libs/utils/src/json_stream.ts
eventualbuddha: I opted to enforce some type safety at the `jsonStream` call site in [this commit](https://github.com/votingworks/vxsuite/pull/3042/commits/6e4a341416ce80b7c8111844dd696cafd81628e6):

```ts
    jsonStream(
      typedAs<WithArraysAsIterables<CVR.CastVoteRecordReport>>({
        ...castVoteRecordReportMetadata,
        CVR: castVoteRecordGenerator,
      })
    )
```

eventualbuddha: And then I changed it to have `jsonStream` take a type parameter:

```ts
    jsonStream<CVR.CastVoteRecordReport>({
      ...castVoteRecordReportMetadata,
      CVR: castVoteRecordGenerator,
    })
```

jonahkagan: Nice!

jonahkagan (approved):
This looks great from my perspective, but maybe @adghayes wants to weigh in since he has the most context

libs/utils/src/json_stream.ts
eventualbuddha: Ah, I see. Yeah that makes sense. I added a `compact` option for that and it's `true` by default.

PR #3045, opened 2023-02-28
Complete Java Card implementation of data storage methods

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2908

This PR completes the Java Card implementation of the card API's data storage methods (`readData`, `writeData`, and `clearData`).

## Testing

- [x] Manually tested reading, writing, and clearing data of various sizes

Automated tests will be coming in follow-up PRs!

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Coming in follow-up PRs

### DIFF
--- a/libs/auth/src/apdu.ts
+++ b/libs/auth/src/apdu.ts
@@ -156,7 +156,7 @@ export class CardCommand {
           p2: this.p2,
           data: this.data.subarray(
             i * MAX_COMMAND_APDU_DATA_LENGTH,
-            // Okay if this is larger than this.data.length
+            // Okay if this is larger than this.data.length as .subarray() automatically caps
             i * MAX_COMMAND_APDU_DATA_LENGTH + MAX_COMMAND_APDU_DATA_LENGTH
           ),
         })
--- a/libs/auth/src/apdu.ts
+++ b/libs/auth/src/apdu.ts
@@ -31,10 +31,15 @@ const CLA = {
  * - 0x61 0xXX is also equivalent to an HTTP 200 but indicates that XX more bytes of response data
  *   have yet to be retrieved via a GET RESPONSE command. Like command APDUs, response APDUs have a
  *   max length.
+ *
+ * See https://www.eftlab.com/knowledge-base/complete-list-of-apdu-responses for a list of all
+ * known status words.
  */
 export const STATUS_WORD = {
   SUCCESS: { SW1: 0x90, SW2: 0x00 },
   SUCCESS_MORE_DATA_AVAILABLE: { SW1: 0x61 },
+  FILE_NOT_FOUND: { SW1: 0x6a, SW2: 0x82 },
+  VERIFY_FAIL: { SW1: 0x63 },
 } as const;
 
 /**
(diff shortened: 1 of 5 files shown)


libs/auth/src/apdu.ts
```diff
@@ -150,6 +204,32 @@ export function constructTlv(
...
+ */
+export function parseTlv(
+  tagAsByteOrBuffer: Byte | Buffer,
+  tlv: Buffer
+): [Buffer, Buffer, Buffer] {
+  const tagLength = Buffer.isBuffer(tagAsByteOrBuffer)
+    ? tagAsByteOrBuffer.length
+    : 1;
+
+  let tlvLengthLength = 1; // Derpiest variable name, I know
```
jonahkagan, line 218: 🥸

libs/auth/src/java_card.ts
```diff
@@ -81,6 +98,29 @@ const VX_ADMIN_CERT_AUTHORITY_CERT = {
   OBJECT_ID: pivDataObjectId(0xf2),
 } as const;
 
+/**
+ * A rough upper bound on the size of a single data object
```
jonahkagan, line 102: Maybe add some context around how you came to this upper bound just in case we need to reason about it in the future?

libs/auth/src/java_card.ts
```diff
@@ -236,20 +284,66 @@ export class JavaCard implements Card {
...
+            data: constructTlv(GET_DATA.TAG_LIST_TAG, objectId),
+          })
+        );
+      } catch (error) {
+        if (
+          error instanceof ResponseApduError &&
+          isFileNotFoundStatusWord(error.statusWord())
+        ) {
+          chunks.push(Buffer.from([]));
+          continue;
```
jonahkagan, line 310: When would this case occur? Why is it skippable?

libs/auth/src/java_card.ts
```diff
@@ -236,20 +284,66 @@ export class JavaCard implements Card {
...
-  clearData(): Promise<void> {
-    return Promise.resolve();
+  async writeData(data: Buffer): Promise<void> {
+    if (data.length > GENERIC_STORAGE_SPACE_CAPACITY) {
+      throw new Error('Not enough space on card');
+    }
+
+    for (const [i, objectId] of GENERIC_STORAGE_SPACE.OBJECT_IDS.entries()) {
+      const chunk = data.subarray(
+        // Okay if these are larger than data.length
```
jonahkagan, line 327: Could you elaborate a bit on what happens when the chunks exceed the size of `data`? Inferring from the context, it seems like maybe this would result in "clearing" the data stored in the various slots.

libs/auth/src/java_card.ts
benadida: also, isn't the max object size 2^5 Kb?

benadida (approved):

libs/auth/src/java_card.ts
arsalansufi: Yes good call!

@ben It's very specifically `2^15 - 5` bytes. I went with 30K to leave breathing room, but on second thought, it doesn't make sense to add the breathing room at this level. It makes sense to add the breathing room a level up in the total capacity variable.

Made tweaks accordingly and added lots of context in: https://github.com/votingworks/vxsuite/pull/3045/commits/8f0f81f8a079225b17a521ee73eb3e5d7e4f7f62

libs/auth/src/java_card.ts
arsalansufi: Explained in https://github.com/votingworks/vxsuite/pull/3045/commits/7d86f4f78d9e3f8c98acc57d9e27eb9f5fa15092

libs/auth/src/java_card.ts
arsalansufi: Elaborated on in https://github.com/votingworks/vxsuite/pull/3045/commits/7d86f4f78d9e3f8c98acc57d9e27eb9f5fa15092

Issue #3048, opened 2023-02-28
Load election definition files directly from backend


adghayes:
Currently, we pass the file data from the frontend to the backend. File reads should be moved to the backend in the interest of deprecating most of `kiosk-browser`'s APIs.

adghayes:
We now read election packages on the backend, and don't support directly loading election definition files.

Issue #3047, opened 2023-02-28
Move VxAdmin External Tally to Backend


adghayes:
We may want to do some refactoring along the way since all "external tallies" are now our manual tally.

PR #3052, opened 2023-03-01
Admin Grout

adghayes:
## Requirements
- Convert all routes to `grout`
- Update testing framework on frontend and backend
- Move from `mutateAsync` to `mutate` wherever possible
- Remove `backend` abstraction

## Non-Requirements
- #3047
- #3048
- #3076 
- #3077 
- Reorganizing state on frontend (e.g. `Context`)
- Moving away from `useEffect`
- Optimize UI behavior while loading, although we don't want to make it worse

## Testing
Rewrote most of higher-level testing on the frontend and all the server testing on the backend. Currently working through a manual pass, although I did that throughout the changes.

I think the new testing in this change should be a first move toward a new pattern. On the backend, we should test flows rather than previously, mostly endpoints. On the frontend, we should _not_ test flows because it is not very useful. Our frontend testing should aim to test that:

1. Data from the backend correctly informs the user interface (enabled or disabled options, data shown to user, etc.)
2. Actions on the frontend correctly push mutations to the backend
3. The overall app user interface is properly navigable

1 & 2 are more efficiently and thoroughly accomplished using lower-level component or screen tests. 3 still requires some sort of high-level test but those should be limited to navigating the app in different auth modes or states. 

We can't move toward this fully quite yet because some data, the manual tally data, is still on the frontend. Also, because tallying happens at various places in the frontend - at the top-level for VotingWorks CVRs and then incorporating manual data and write-ins at the component level - it it still seems safer to have the tests that currently test results at a higher-level.


apps/admin/backend/src/server.ts
```diff
@@ -109,528 +163,279 @@ function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
...
-        electionId,
+    getWriteIns({
+      contestId,
+      status,
+      limit,
+    }: {
+      contestId?: ContestId;
+      status?: Admin.WriteInAdjudicationStatus;
+      limit?: number;
+    } = {}) {
```
jonahkagan, line 346: Two conventions I thought might be nice when I was writing the first grout API:
- Always name the single parameter `input` (makes it a bit easier to read the signature, and also helps with type tooltips). Destructure later if needed
- Always write an explicit return type (not strictly needed, but since we have a lot of type inference happening inside grout, I thought it might help with clarity to readers)

apps/admin/frontend/src/api.ts
```diff
@@ -102,3 +104,243 @@ export const unprogramCard = {
     });
   },
 } as const;
+
+// Queries
+
+type QueryProps<T extends keyof Api> = Parameters<Api[T]>[0];
```
jonahkagan, line 110: While this works and is very cool, it's breaking the type abstraction of grout a bit (`Api` should be treated as an opaque type by consumers). The client types, however, are fair game to use, so how about something like this:

```ts
type QueryProps<Method extends keyof ApiClient> = Parameters<ApiClient[Method]>[0]
```

apps/admin/frontend/src/api.ts
```diff
@@ -102,3 +104,243 @@ export const unprogramCard = {
...
+    return useQuery(this.queryKey(props), () =>
+      apiClient.getWriteInImage(props)
+    );
+  },
+} as const;
+
+// Grouped Invalidations
+
+function invalidateCastVoteRecordQueries(queryClient: QueryClient) {
+  void queryClient.invalidateQueries(getCastVoteRecordFileMode.queryKey());
```
jonahkagan, line 212: I think @kofi-q and I dug into this and decided that we should always `await` the call to `invalidateQueries`. That way the mutation doesn't complete until the invalidated data is refetched, so the user will see whatever UI we have for "mutation in progress" until the effects of the mutation are ready to be rendered

apps/admin/frontend/src/api.ts
```diff
@@ -102,3 +104,243 @@ export const unprogramCard = {
...
+  },
+} as const;
+
+export const unconfigure = {
+  useMutation() {
+    const apiClient = useApiClient();
+    const queryClient = useQueryClient();
+    return useMutation(apiClient.unconfigure, {
+      onSuccess() {
+        // invalidate all queries but auth
```
jonahkagan, line 245: I wonder if we should actually put in some wildcard key to actually invalidate _all_ queries in the cache. It might be hard to keep this up to date as we add more queries.

jonahkagan (commented):
Just reviewed server.ts and api.ts

apps/admin/frontend/src/api.ts
adghayes: if you just call `invalidateQueries` with no arguments it will do just that. I had that initially, but I was reticent to invalidate the auth status query. Could be fine, or maybe there's a way to invalidate all but.

apps/admin/frontend/src/api.ts
adghayes: If it works, it's good to me.

apps/admin/frontend/src/api.ts
jonahkagan: I can't imagine invalidating the auth status query would have much impact. It's refetched constantly anyway.

adghayes: Right now it'll return `null` for the whole app while loading, but that doesn't really matter / could be changed.

jonahkagan: After invalidating the auth status query, the stale auth status will still be shown while the new status is fetching. That's the foundational contract of react-query (it's a caching strategy called "stale while revalidate")

apps/admin/frontend/src/app.test.tsx
```diff
@@ -510,84 +435,50 @@ test('printing ballots and printed ballots report', async () => {
...
     expect.anything()
   );
 
-  const markOfficialButton = getByText('Mark Tally Results as Official');
-  await waitFor(() => expect(markOfficialButton).toBeEnabled());
-  fireEvent.click(markOfficialButton);
-  getByText('Mark Unofficial Tally Results as Official Tally Results?');
-  const modal = await screen.findByRole('alertdialog');
-  fireEvent.click(within(modal).getByText('Mark Tally Results as Official'));
-
```
adghayes, line 590: Moved testing to `tally_report_screen`

apps/admin/frontend/src/app.test.tsx
```diff
@@ -698,72 +596,35 @@ test('tabulating CVRs', async () => {
...
-  // Confirm that L&A Materials are unavailable after live CVRs have been loaded
-  fireEvent.click(getByText('L&A'));
-  getByText(
-    'L&A testing documents are not available after official election CVRs have been loaded.',
-    { exact: false }
-  );
-  expect(queryByText('List Precinct L&A Packages')).not.toBeInTheDocument();
-  expect(
-    queryByText('Print Full Test Deck Tally Report')
-  ).not.toBeInTheDocument();
```
adghayes, line 710: Moved testing to L&A screen

apps/admin/frontend/src/app.test.tsx
```diff
@@ -698,72 +596,35 @@ test('tabulating CVRs', async () => {
...
-  });
-  fireEvent.click(
-    await screen.findByText('Unofficial Full Election Tally Report')
-  );
-
-  const reportPreview2 = getByTestId('report-preview');
-  expect(within(reportPreview2).getAllByText('0').length).toEqual(40);
-  // useQuery's in AppRoot are refetching data, and there's no change to wait on.
-  // TODO: Remove after upgrade to React 18, which does not warn in this case.
-  await advanceTimersAndPromises();
```
adghayes, line 740: Covered in "clearing results" test.

apps/admin/frontend/src/app_root.tsx
```diff
@@ -142,16 +142,28 @@ export function AppRoot({
...
+      });
     },
-    [store]
+    [configureMutation, store]
   );
 
-  const resetElection: ResetElection = useCallback(async () => {
-    await store.reset();
-  }, [store]);
-
```
adghayes, line 154: Was able to move this out of the top-level `Context`.

apps/admin/frontend/src/screens/unconfigured_screen.tsx
```diff
@@ -79,35 +78,30 @@ export function UnconfiguredScreen(): JSX.Element {
...
     const file = input.files && input.files[0];
 
     if (file) {
       setVxElectionFileIsInvalid(false);
-      try {
-        const fileContent = await readFileAsync(file);
-        await validateAndSaveElection(fileContent);
-      } catch (error) {
+      // TODO: read file content from backend
+      const fileContent = await readFileAsync(file);
```
adghayes, line 94: #3048

apps/admin/backend/src/server.ts
```diff
@@ -192,11 +206,131 @@ function buildApi({
...
+        });
+      }
+
+      return addFileResult;
+    },
+
+    clearCastVoteRecordFiles() {
+      const electionId = loadCurrentElectionIdOrThrow(workspace);
+      store.deleteCastVoteRecordFiles(electionId);
+      store.setElectionResultsOfficial(electionId, false);
```
adghayes, line 331: Unrelated to the choices or changes in this PR, but currently when we clear the cast vote record files, we chance an election from "marked official" to "not marked official." Haven't thought through this too much - is it a potential security loophole? cc @mattroe @benadida? We could instead do it only in "Test Ballot Mode."

apps/admin/backend/src/server.cvr_file_mode.test.ts
```diff
@@ -1,53 +0,0 @@
-import { Admin } from '@votingworks/api';
```
adghayes, line 1: testing the file mode endpoint in a more integrated way in `server.test.ts`

apps/admin/backend/src/server.ts
```diff
@@ -109,528 +161,264 @@ function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
...
-                message: `expected file field to be named "${CVR_FILE_ATTACHMENT_NAME}"`,
-              },
-            ],
-          });
-          return;
+    async configure(input: { electionData: string }): Promise<ConfigureResult> {
+      const parseResult = safeParseElectionDefinition(input.electionData);
+      if (parseResult.isErr()) {
+        return err({ type: 'parsing', message: parseResult.err().message });
+      }
```
adghayes, line 169: Previously parsing the election definition on the frontend, now on the backend.

apps/admin/backend/src/server.ts
```diff
@@ -109,528 +161,264 @@ function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
...
-          status: 'error',
-          errors: [
-            {
-              type: 'ValidationError',
-              message: parseQueryResult.err().message,
-            },
-          ],
+    async addCastVoteRecordFile(input: {
+      path: string;
+    }): Promise<AddCastVoteRecordFileResult> {
```
adghayes, line 253: Previously passing a `File` object to the backend, now just a path.

apps/admin/backend/src/server.ts
```diff
@@ -761,117 +519,53 @@ export function buildApp({
...
+}: {
+  auth: DippedSmartCardAuthApi;
+  workspace: Workspace;
+  logger: Logger;
+}): Application {
+  const app: Application = express();
+  const api = buildApi({ auth, workspace, logger });
+  app.use('/api', grout.buildRouter(api, express));
   return app;
 }
```
adghayes, line 570: Could move `buildApi`, or have a separate `app` file cc @jonahkagan.

apps/admin/frontend/src/app_root.tsx
```diff
@@ -142,16 +147,21 @@ export function AppRoot({
...
+          },
+          {
+            onSuccess: resolve,
+          }
+        );
+      });
     },
-    [store]
+    [configureMutation]
   );
```
adghayes, line 163: Would like to move this out of top-level `Context`, but the definition screen is using it also so I left it here. Could move it down anyway.

@mattroe can we axe the definition editing screens on VxAdmin? It is work to maintain them.

apps/admin/frontend/src/components/confirm_removing_file_modal.tsx
```diff
@@ -19,33 +18,25 @@ export function ConfirmRemovingFileModal({
...
-  const cvrFilesQuery = useCvrFilesQuery();
+  const castVoteRecordFilesQuery = getCastVoteRecordFiles.useQuery();
 
-  const isLoading =
-    (fileType === ResultsFileType.CastVoteRecord ||
-      fileType === ResultsFileType.All) &&
-    cvrFilesQuery.isLoading;
+  if (!castVoteRecordFilesQuery.isSuccess) {
+    return null;
+  }
```
adghayes, line 32: On modals I feel completely fine just returning null until we're ready.

apps/admin/frontend/src/components/export_batch_tally_results_button.tsx
```diff
@@ -6,15 +6,22 @@ import { assert } from '@votingworks/basics';
...
   const { fullElectionTally, electionDefinition } = useContext(AppContext);
   assert(electionDefinition);
   const { election } = electionDefinition;
 
-  const isTestMode = useCvrFileModeQuery().data === Admin.CvrFileMode.Test;
+  const castVoteRecordFileModeQuery = getCastVoteRecordFileMode.useQuery();
+
+  if (!castVoteRecordFileModeQuery.isSuccess) {
+    return null;
+  }
```
adghayes, line 21: Returning `null` on this button, I'm not as sure about.

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -79,82 +81,61 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
...
-      assert(error instanceof Error);
-      setCurrentState({ state: 'error', error, filename });
-      await logger.log(LogEventId.CvrLoaded, userRole, {
-        message: `Failed to load CVR file: ${error.message}`,
-        disposition: 'failure',
-        filename,
-        error: error.message,
-        result: 'File not loaded, error shown to user.',
-      });
-    }
```
adghayes, line 125: This logging has been moved to the backend.

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -314,7 +303,7 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
         onOverlayClick={onClose}
         actions={
           <React.Fragment>
-            {(!window.kiosk || process.env.NODE_ENV === 'development') && (
+            {process.env.NODE_ENV === 'development' && (
```
adghayes, line 306: This is what's breaking the VxAdmin integration tests until #3076.

apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -328,7 +330,9 @@ export function ManualDataImportPrecinctScreen(): JSX.Element {
     existingPrecinctTally
   );
 
-  const writeInSummaryQuery = useWriteInSummaryQuery({ status: 'adjudicated' });
+  const writeInSummaryQuery = getWriteInSummary.useQuery({
+    status: 'adjudicated',
+  }) as UseQueryResult<Admin.WriteInSummaryEntryAdjudicated[]>;
```
adghayes, line 335: Previously there was an overload on the query that provided this. It's harder to overload a `grout` API so I'm just typecasting. We could just make a query specifically for `status: 'adjudicated`` however, as the frontend only uses the query with that status anyway.

apps/admin/frontend/src/screens/tally_screen.tsx
```diff
@@ -54,16 +52,30 @@ export function TallyScreen(): JSX.Element {
...
+    return (
+      <NavigationScreen>
+        <Prose maxWidth={false}>
+          <h1>Cast Vote Record (CVR) Management</h1>
+        </Prose>
+      </NavigationScreen>
+    );
+  }
+
+  const castVoteRecordFileList = castVoteRecordFilesQuery.data;
```
adghayes, line 71: If queries aren't ready, at least render the header of the page.

apps/admin/frontend/src/api.ts
```diff
@@ -223,16 +223,20 @@ export const getPrintedBallots = {
...
-async function invalidateCastVoteRecordQueries(queryClient: QueryClient) {
-  await queryClient.invalidateQueries(getCastVoteRecordFileMode.queryKey());
-  await queryClient.invalidateQueries(getCastVoteRecordFiles.queryKey());
-  await queryClient.invalidateQueries(getCastVoteRecords.queryKey());
+function invalidateCastVoteRecordQueries(queryClient: QueryClient) {
+  return Promise.all([
+    queryClient.invalidateQueries(getCastVoteRecordFileMode.queryKey()),
+    queryClient.invalidateQueries(getCastVoteRecordFiles.queryKey()),
+    queryClient.invalidateQueries(getCastVoteRecords.queryKey()),
+  ]);
```
adghayes, line 231: I changed this when I was struggling with the queries in tests. Can revert it if it's confusing but doesn't matter one way or another functionally.

apps/admin/frontend/src/app.test.tsx
```diff
@@ -204,20 +182,18 @@ test('configuring with a demo election definition', async () => {
   // You can not save as CDF when there is no election.
   expect(screen.queryAllByText('No Log File Present')).toHaveLength(0);
 
-  fireEvent.click(screen.getByText('Definition'));
+  userEvent.click(screen.getByText('Definition'));
   await screen.findByText('Load Demo Election Definition');
 });
 
 test('configuring with a VXF election definition from a file', async () => {
```
adghayes: The tests you recently added @jonahkagan get moved to the backend in a later commit, since the backend is now where parsing takes place.

apps/admin/frontend/src/app.test.tsx
```diff
@@ -482,68 +427,33 @@ test('L&A (logic and accuracy) flow', async () => {
   );
 });
 
-test('L&A features are available after test results are loaded', async () => {
```
adghayes, line 485: Moved to l&a specific test file.

apps/admin/frontend/src/app.test.tsx
```diff
@@ -482,68 +427,33 @@ test('L&A (logic and accuracy) flow', async () => {
...
-  screen.getByText('Print Full Test Deck Tally Report');
-});
-
-test('printing ballots and printed ballots report', async () => {
-  const backend = new ElectionManagerStoreMemoryBackend({
-    electionDefinition: eitherNeitherElectionDefinition,
-  });
-
-  const printer = fakePrinter();
+test('printing ballots', async () => {
```
adghayes, line 371: printed ballots report moved to a different test below focusing on rending information in the reports screen.

apps/admin/frontend/src/app_write_in_flows.test.tsx
```diff
@@ -262,97 +252,4 @@ test('manual write-in data end-to-end test', async () => {
   within(zooCouncilMammal2).getByText(hasTextAcrossElements('Chimera10'));
   within(zooCouncilMammal2).getByText(hasTextAcrossElements('Rapidash20'));
   within(zooCouncilFish2).getByText(hasTextAcrossElements('Relicanth14'));
-
-  userEvent.click(screen.getByText('Reports'));
-});
-
-test('availability of write-in tally report', async () => {
```
adghayes, line 269: moved this testing to more thorough testing in `tally_writein_report_screen.test.tsx`.

apps/admin/frontend/src/components/import_cvrfiles_modal.test.tsx
```diff
@@ -355,115 +299,24 @@ describe('Screens display properly when USB is mounted', () => {
...
-
-    fireEvent.click(domGetByText(tableRows[0], 'Load'));
-    await screen.findByText('Loading');
-    await waitFor(() => {
-      expect(backend.addCastVoteRecordFile).toHaveBeenCalledTimes(1);
-      getByText('0 new CVRs Loaded');
-    });
-  });
-
-  test('Shows previously loaded files when all files have already been loaded', async () => {
```
adghayes, line 415: lots of duplicate coverage of this file across tests, tried to consolidate into more targeted tests.

adghayes (commented):
Alright @jonahkagan, it's ready for you! :) 

I did end up having to rebase because there was too much churn in my feature-by-feature commits. Now it's organized by:
- rewrite backend
- rewrite frontend (no tests)
- rewrite frontend tests
- rebase testing fixes
- disabled integration testing

adghayes:
@jonahkagan just did a manual pass and found one issue. On the `manual_tally_precinct_screen` we were waiting for data to be refetched after mount, which the new shared react query settings do not always do. Fixed it in a small tweak.

apps/admin/frontend/test/render_in_app_context.tsx
```diff
@@ -89,7 +87,13 @@ export function renderRootElement(
...
     // TODO: Determine why tests fail when using createQueryClient and, by extension,
     // QUERY_CLIENT_DEFAULT_OPTIONS
-    queryClient = new QueryClient(),
+    queryClient = new QueryClient({
+      defaultOptions: {
+        queries: {
+          staleTime: Infinity,
+        },
+      },
+    }),
```
arsalansufi, line 96: 🙈 I discovered this discrepancy between the query client in tests and the query client in source code a while back, hence the TODO

To minimize the differences between the test query client and the real query client without having to get to the bottom of the test failures when using the default options, might be nice to still use the default options but with explicit overrides, e.g.

```ts
queryClient = new QueryClient({
  defaultOptions: {
    ...QUERY_CLIENT_DEFAULT_OPTIONS,
    queries: {
      ...(QUERY_CLIENT_DEFAULT_OPTIONS.queries ?? {}),
      useErrorBoundary: false,
    },
  },
}),
```

Doing a quick manual test, I think it's just `queries.useErrorBoundary: true` that's now resulting in test failures

apps/admin/frontend/test/helpers/api_mock.ts
```diff
@@ -0,0 +1,219 @@
...
+export function setAuthStatus(
+  mockApiClient: MockApiClient,
+  authStatus: DippedSmartCardAuth.AuthStatus
+): void {
+  mockApiClient.getAuthStatus.mockImplementation(() =>
+    Promise.resolve(authStatus)
+  );
+}
+
+export function expectGetCurrentElectionMetadata(
```
jonahkagan, line 41: why is this defined once here and once on the wrapper?

apps/admin/frontend/test/helpers/api_mock.ts
```diff
@@ -0,0 +1,219 @@
...
+    },
+
+    expectGetCurrentElectionMetadata(
+      metadata: {
+        electionDefinition: ElectionDefinition;
+        isOfficialResults?: boolean;
+        id?: string;
+        createdAt?: string;
+      } | null,
+      times = 1
```
jonahkagan, line 141: do we need the `times` param?

apps/admin/frontend/test/helpers/api_mock.ts
```diff
@@ -0,0 +1,219 @@
...
+                createdAt: new Date().toISOString(),
+                isOfficialResults: false,
+                ...metadata,
+              }
+            : null
+        );
+      }
+    },
+
+    expectGetCastVoteRecords(castVoteRecords: CastVoteRecord[], times = 1) {
```
jonahkagan, line 157: What about here - do we need `times`?

apps/admin/frontend/package.json
```diff
@@ -62,6 +62,7 @@
     "@types/array-unique": "^0.3.0",
     "@types/dashify": "^1.0.0",
     "@types/dompurify": "^2.0.3",
+    "@types/electron": "^1.6.10",
```
jonahkagan, line 65: What is this used for?

jonahkagan (commented):
Thanks for breaking it up into chunks - that definitely helped!

apps/admin/frontend/src/api.ts
```diff
@@ -102,3 +104,265 @@ export const unprogramCard = {
...
+    return useQuery(this.queryKey(), () =>
+      apiClient.getCastVoteRecordFileMode()
+    );
+  },
+} as const;
+
+type GetWriteInsProps = QueryProps<'getWriteIns'>;
+export const getWriteIns = {
+  queryKey(props?: GetWriteInsProps): QueryKey {
+    return props ? ['getWriteIns', props] : ['getWriteIns'];
```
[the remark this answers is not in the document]
adghayes: The `['getWriteIns']` variant is important for when you're using the `queryKey` to invalidate queries. I don't think `['getWriteIns', undefined]` would clear the query. But, we could special case the query invalidation differently by either making a specific method for it, or a specific property with the query root `'getWriteIns'`.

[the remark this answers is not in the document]
jonahkagan: I think it would work with [`getWriteIns`, undefined]` since the query key matching uses prefixes: https://tanstack.com/query/v4/docs/react/guides/query-invalidation#query-matching-with-invalidatequeries. But I could be wrong

apps/admin/frontend/src/app_root.tsx
```diff
@@ -69,16 +71,19 @@ export function AppRoot({
   });
```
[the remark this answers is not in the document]
adghayes: Yeah, that's definitely a follow-up task, but will punt here.

apps/admin/frontend/src/screens/tally_report_screen.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+    electionDefinition: electionMinimalExhaustiveSampleDefinition,
+    isOfficialResults: true,
+    logger,
+    apiClient,
+    route: routerPaths.tallyFullReport,
+  });
+
+  await waitFor(() => {
+    apiMock.apiClient.assertComplete();
+  });
```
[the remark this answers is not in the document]
adghayes: In this case, we have a button that is disabled and is only enabled once the information from the backend is loaded. If we check that it's disabled right out the gate, it will always be true, and not actually do the test properly.

apps/admin/frontend/src/screens/tally_report_screen.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+      screen.getByRole('button', { name: 'Mark Tally Results as Official' })
+    ).not.toBeDisabled();
+  });
+  userEvent.click(screen.getByText('Mark Tally Results as Official'));
+  screen.getByText('Mark Unofficial Tally Results as Official Tally Results?');
+  userEvent.click(
+    within(screen.getByRole('alertdialog')).getByText(
+      'Mark Tally Results as Official'
+    )
+  );
```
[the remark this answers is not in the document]
adghayes: The visual feedback is things on the page flipping from "Unofficial" to "Official," but that happens as a result of the app state (from the server) changing as a result of the mutation. So I don't think we want to make assertions on it.

apps/admin/frontend/test/helpers/api_mock.ts
adghayes: relic

apps/admin/frontend/test/helpers/api_mock.ts
adghayes: nope, relic

apps/admin/frontend/test/helpers/api_mock.ts
adghayes: nope, relic

apps/admin/frontend/package.json
adghayes: relic, will remove

apps/admin/frontend/src/components/smartcard_modal/smartcard_modal.test.tsx
```diff
@@ -61,11 +51,11 @@ afterEach(() => {
 });
 
 test('Smartcard modal displays card details', async () => {
-  renderRootElement(<App hardware={mockHardware} />, {
-    apiClient: mockApiClient,
-    backend: mockBackend,
-  });
-  await authenticateAsSystemAdministrator(mockApiClient);
+  const { apiMock, renderApp } = buildApp(mockApiClient);
```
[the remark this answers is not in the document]
adghayes: Some of them could, but I think the special thing about the `smartcard_modal` is that it can pop up anywhere in the app and is nested at a high level. I don't intend to touch them.

apps/admin/backend/src/store.ts
```diff
@@ -84,7 +84,10 @@ type AddCvrFileResult = Result<Admin.CvrFileImportInfo, AddCastVoteRecordError>;
  */
 export class Store {
   private constructor(private readonly client: DbClient) {
-    this.client.run('insert into settings default values');
+    const settings = Boolean(this.client.one('select id from settings'));
+    if (!settings) {
+      this.client.run('insert into settings default values');
+    }
```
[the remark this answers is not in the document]
adghayes: good idea, will do one or the other. I like the idea of the schema file.

apps/admin/backend/src/server.ts
adghayes: Motivation for changing would just be to be the same across app backends. I didn't know VxScan was actually doing it the same way, except that it has a separate `server.ts` and `app.ts` file. Will rearrange to be the same.

apps/admin/frontend/src/app_root.tsx
adghayes: I probably won't axe them in this PR because it has other implications, so I'll just move it down.

apps/admin/frontend/src/components/export_batch_tally_results_button.tsx
adghayes: OK, will look into options here.

apps/admin/backend/src/server.ts
```diff
@@ -109,528 +161,264 @@ function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
...
+          message: `Failed to access cast vote record file for import.`,
+          disposition: 'failure',
+          filename,
+          error: (error as Error).message,
+          result: 'File not loaded, error shown to user.',
+        });
+        return err({
+          type: 'invalid-file',
+          userFriendlyMessage:
+            'the selected file could not be opened by the system',
```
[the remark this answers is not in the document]
adghayes: In this case, although this is buried in the parsing CVR file code, we're already providing a `userFriendlyMessage` on parsing issues and I wanted it to be the same here.

apps/admin/backend/src/server.ts
```diff
@@ -109,528 +161,264 @@ function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
...
-            errors: [
-              {
-                type: 'invalid-request-body',
-                message: parseBodyResult.err().message,
-              },
-            ],
-          });
-          return;
-        }
+      return null;
```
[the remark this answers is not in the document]
adghayes: I remember changing it to `null` because I thought `grout` was complaining about `undefined`. Maybe that was unrelated, I'll see if I can change it back.

apps/admin/frontend/src/components/print_all_ballots_button.test.tsx
```diff
@@ -153,15 +155,15 @@ test('print sequence proceeds as expected', async () => {
...
-    );
+    // const printedBallots = await backend.loadPrintedBallots();
+    // expect(printedBallots).toHaveLength(i + 1);
+    // expect(printedBallots[printedBallots.length - 1]).toEqual(
+    //   expect.objectContaining(
+    //     typedAs<Partial<Admin.PrintedBallotRecord>>({
+    //       numCopies: 1,
+    //     })
+    //   )
+    // );
```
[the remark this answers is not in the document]
adghayes: something that should just be removed

apps/admin/backend/src/store.ts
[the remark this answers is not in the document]
adghayes: moved to schema file

apps/admin/backend/src/server.ts
benadida: I don't see a problem.

apps/admin/frontend/src/components/export_batch_tally_results_button.tsx
adghayes: this is below the fold so it practically doesn't matter, but that could change in future. for now I switched it to just have the button disabled until the query is ready, which seems to me like a fine pattern

apps/admin/backend/src/server.ts
```diff
@@ -109,528 +161,264 @@ function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
...
-          errors: [
-            {
-              type: 'ValidationError',
-              message: parseResult.err().message,
-            },
-          ],
-        });
-        return;
-      }
+    // TODO: Not being used by frontend. Do we need it?
```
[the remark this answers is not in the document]
adghayes: It's used by a lot of tests, is querying what seems to me like salient data from the backend, and isn't introducing much "cognitive overhead" so I'd actually rather just leave it in until our write-in work hardens.

apps/admin/backend/src/server.ts
```diff
@@ -109,528 +161,264 @@ function buildApi(auth: DippedSmartCardAuthApi, workspace: Workspace) {
...
-  deprecatedApiRouter.get<
-    Admin.GetWriteInAdjudicationTableUrlParams,
-    Admin.GetWriteInAdjudicationTableResponse,
-    Admin.GetWriteInAdjudicationTableRequest
-  >(
-    '/admin/elections/:electionId/contests/:contestId/write-in-adjudication-table',
-    (request, response) => {
-      const { electionId, contestId } = request.params;
-      const electionRecord = store.getElection(electionId);
+    // TODO: Frontend only uses with status "adjudicated". Change to improve typing?
```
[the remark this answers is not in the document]
adghayes: I'd also like to keep this for now, but can remove it if you insist. Low cost other than the typecasting. Updated comment.

apps/admin/frontend/src/api.ts
```diff
@@ -102,3 +104,265 @@ export const unprogramCard = {
     });
   },
 } as const;
+
+// Queries
+
+type QueryProps<Method extends keyof ApiClient> = Parameters<
```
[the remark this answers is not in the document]
adghayes: nit accepted

apps/admin/frontend/src/api.ts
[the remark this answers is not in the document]
adghayes: OK, I tested and it did not work. My understanding of that reference page is that `['getWriteIns']` would operate as a prefix for `['getWriteIns', anything]` but that `['getWriteIns, undefined]` is not a prefix for `['getWriteIns', 'key']`.

apps/admin/frontend/src/api.ts
```diff
@@ -102,3 +104,265 @@ export const unprogramCard = {
...
+  },
+  useQuery(props?: GetPrintedBallotsProps) {
+    const apiClient = useApiClient();
+    return useQuery(this.queryKey(props), () =>
+      apiClient.getPrintedBallots(props)
+    );
+  },
+} as const;
+
+// Grouped Invalidations
```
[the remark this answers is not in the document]
adghayes: Yeah, that's a cool idea I hadn't thought about. At this point, I don't know if I feel like they are already invalidated together. Right now I think there are one or two cases within the write-ins group where the invalidation is unnecessary but I'm just doing it anyway. I lean toward leaving it as-is for now.

apps/admin/backend/src/server.ts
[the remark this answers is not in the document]
adghayes: ok yeah it can just be `undefined`, not sure what I hit before. switched.

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -79,82 +81,61 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element {
...
-      setCurrentState({ state: 'error', error, filename });
-      await logger.log(LogEventId.CvrLoaded, userRole, {
-        message: `Failed to load CVR file: ${error.message}`,
-        disposition: 'failure',
-        filename,
-        error: error.message,
-        result: 'File not loaded, error shown to user.',
-      });
-    }
+    return new Promise<void>((resolve) => {
```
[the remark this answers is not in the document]
adghayes: sounds good, switched

apps/admin/frontend/src/components/remove_election_modal.tsx
```diff
@@ -1,20 +1,24 @@
...
   const history = useHistory();
-  const { resetElection } = useContext(AppContext);
+  const unconfigureMutation = unconfigure.useMutation();
+  const store = useElectionManagerStore();
 
   async function unconfigureElection() {
-    await resetElection();
+    // TODO: remove line once external tallies are in the backend
+    await store.clearFullElectionExternalTallies();
+    unconfigureMutation.mutate();
```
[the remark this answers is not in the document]
adghayes: I didn't notice anything funny in testing, but I think it would be better to wait. Changed.

apps/admin/frontend/src/components/save_results_button.tsx
```diff
@@ -6,23 +6,34 @@ import { Admin } from '@votingworks/api';
...
+}): JSX.Element | null {
   const [isSaveModalOpen, setIsSaveModalOpen] = useState(false);
   const { fullElectionTally, electionDefinition } = useContext(AppContext);
   assert(electionDefinition);
   const { election } = electionDefinition;
 
-  const isTestMode = useCvrFileModeQuery().data === Admin.CvrFileMode.Test;
+  const castVoteRecordFileModeQuery = getCastVoteRecordFileMode.useQuery();
+  const isTestMode =
+    castVoteRecordFileModeQuery.data === Admin.CvrFileMode.Test;
```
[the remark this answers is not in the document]
adghayes: I think disabled until ready is fine here, lmk if you disagree.

apps/admin/backend/src/server.ts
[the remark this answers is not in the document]
adghayes: Now I'm reminded why I switched it, I get this error:

```Query data cannot be undefined. Please make sure to return a value other than undefined from your query function. Affected query key: ["getCurrentElectionMetadata"]```

Seems like a React Query requirement, presumable because the data being `undefined` can indicate something else (no data).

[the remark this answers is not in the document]
Comment (adghayes) [reply apps/admin/backend/src/server.ts]: I see there are tests in `grout.test.ts` for endpoints that return `Person | undefined` but maybe that's not actually supported?

File: apps/admin/backend/src/server.ts (hunk 43 above)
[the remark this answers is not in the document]
Comment (jonahkagan) [reply apps/admin/backend/src/server.ts]: No strong opinion

File: apps/admin/frontend/src/api.ts (hunk 33 above)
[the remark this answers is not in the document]
Comment (jonahkagan) [reply apps/admin/frontend/src/api.ts:161]: Yeah that makes sense. Thanks for testing it out!

File: apps/admin/frontend/src/api.ts (hunk 45 above)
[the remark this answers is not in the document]
Comment (jonahkagan) [reply apps/admin/frontend/src/api.ts:224]: Cool, sounds fine to leave it for now.

File: apps/admin/backend/src/server.ts (hunk 40 above)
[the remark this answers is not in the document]
Comment (jonahkagan) [reply apps/admin/backend/src/server.ts:173]: Oh yeah I do recall learning that limitation of react-query at some point. It's definitely a react-query limitation, not a grout limitation - grout supports `undefined`. Maybe just add a comment explaining why?

File: apps/admin/frontend/src/components/save_results_button.tsx (hunk 48 above)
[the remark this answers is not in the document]
Comment (jonahkagan) [reply apps/admin/frontend/src/components/save_results_button.tsx:23]: Yeah, now that you changed the other button to have the same pattern, I feel good about it. Consistency was my goal

File: apps/admin/frontend/src/components/smartcard_modal/smartcard_modal.test.tsx (hunk 37 above)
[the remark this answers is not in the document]
Comment (jonahkagan) [reply apps/admin/frontend/src/components/smartcard_modal/smartcard_modal.test.tsx]: That's fair, though the "popping up anywhere" could be tested once at the app level, and then all the detailed tests could be just for the modal. But fine to kick that down the road for later.

File: apps/admin/frontend/src/screens/tally_report_screen.test.tsx (hunk 35 above)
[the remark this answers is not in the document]
Comment (jonahkagan) [reply apps/admin/frontend/src/screens/tally_report_screen.test.tsx:74]: Oh, I see. That makes sense. If there's no other visual indicator on the screen, then I can't think of any other way. Maybe add an explanatory comment?

I do worry a bit that this test is kind of brittle. What's the guarantee that the UI will "settle" after `advanceTimersAndPromises()`?

File: apps/admin/backend/src/server.ts (hunk 42 above)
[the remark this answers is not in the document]
Comment (jonahkagan) [reply apps/admin/backend/src/server.ts]: Cool, fine with me!

File: apps/admin/frontend/src/screens/unconfigured_screen.tsx [hunk 49]
```diff
@@ -61,8 +64,25 @@ const demoElection =
...
+            electionData,
+          },
+          {
+            onSuccess: resolve,
+          }
+        );
+      });
+    },
+    [configureMutation]
+  );
```
jonahkagan, line 83: This is basically the same as `configureMutation.mutateAsync` except it doesn't throw errors. I wonder if it would be clearer to just use `mutateAsync` here and catch the errors. Reading this code, it's a little unclear why we're using `mutate`

apps/admin/frontend/src/api.ts
```diff
@@ -208,7 +208,7 @@ export const getWriteInImage = {
   },
 } as const;
 
-type GetPrintedBallotsProps = QueryProps<'getPrintedBallots'>;
+type GetPrintedBallotsProps = QueryInput<'getPrintedBallots'>;
```
jonahkagan, line 211: Can you also change `GetPrintedBallotsProps` to `GetPrintedBallotsInput` and then change `queryKey(props)` to `queryKey(input)`? Sorry if I wasn't clear in my original suggestion

jonahkagan (approved):

apps/admin/frontend/src/api.ts
adghayes: You were clear, I didn't think it through.

apps/admin/frontend/src/screens/unconfigured_screen.tsx
adghayes: I misunderstood your previous comments and I thought I had to always use `mutate` or else the errors would escape off into the abyss. I'm going to need an explainer on this.

apps/admin/frontend/src/api.ts
adghayes: Fixed.

apps/admin/frontend/src/screens/unconfigured_screen.tsx
adghayes: So, _normally_ the unhandled promise would crash the app but you've added code _across apps_ so that's not the case?

In `jest`, wouldn't the rejected promise only happen if you mocked the rejection? So you'd know where it is?

If the confusion and crashing is only in test environments, and I don't expect this mutation to throw errors, then I don't see why I wouldn't just use `await configure.mutateAsync()`?

jonahkagan: > So, normally the unhandled promise would crash the app but you've added code across apps so that's not the case?

Yep!

> In jest, wouldn't the rejected promise only happen if you mocked the rejection? So you'd know where it is?

The main way it happens is if the API mock calls are incorrect and there's a missing mock call, which immediately throws an error. Even if you don't throw an error and just log an error (which is an option I built in the API mock), you still need to return some dummy value, which may trigger an error since the consuming code might be expecting a different shape. All of these cases lead to difficult to debug jest crashes.

PR #3053, opened 2023-03-01
Complete Java Card implementation of card programming methods

arsalansufi:

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2909

This PR completes the Java Card implementation of the card API's card programming methods (`program` and `unprogram`). By extension, this completes the Java Card implementation of the card API!

It also cleans up existing Java Card code, better organizing across files:

- `openssl.ts` - Generic openssl commands (not VotingWorks-specific)
- `piv.ts` - PIV standards (not VotingWorks-specific)
- `certs.ts` - VotingWorks-specific cert logic
- `java_card.ts` - Top-level logic

## Testing

- [x] Manually tested programming and unprogramming

Automated tests will be coming in follow-up PRs!

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Coming in follow-up PRs

### DIFF
--- a/libs/auth/src/card_reader.ts
+++ b/libs/auth/src/card_reader.ts
@@ -86,10 +86,6 @@ export class CardReader {
     });
   }
 
-  getReaderStatus(): Reader['status'] {
-    return this.reader.status;
-  }
-
   /**
    * Transmits command APDUs to a smart card. On success, returns response data. On error, throws.
    * Specifically throws a ResponseApduError when a response APDU with a non-success status word is
--- a/libs/auth/src/certs.ts
+++ b/libs/auth/src/certs.ts
@@ -1,0 +1,111 @@
+import { Buffer } from 'buffer';
+import { z } from 'zod';
+import { assert, throwIllegalValue } from '@votingworks/basics';
+import { User } from '@votingworks/types';
+
+import { openssl } from './openssl';
+
+/**
+ * VotingWorks's IANA-assigned enterprise OID
+ */
+const VX_IANA_ENTERPRISE_OID = '1.3.6.1.4.1.59817';
+
+/**
+ * Instead of overloading existing X.509 cert fields, we're using our own custom cert fields.
+ */
+const VX_CUSTOM_CERT_FIELD = {
+  /** One of: card, admin, central-scan, mark, scan (the latter four referring to machines) */
+  COMPONENT: `${VX_IANA_ENTERPRISE_OID}.1`,
+  /** Format: {state-2-letter-abbreviation}.{county-or-town} (e.g. MS.Warren) */
+  JURISDICTION: `${VX_IANA_ENTERPRISE_OID}.2`,
+  /** One of: sa, em, pw (system administrator, election manager, or poll worker) */
+  CARD_TYPE: `${VX_IANA_ENTERPRISE_OID}.3`,
+  /** The SHA-256 hash of the election definition  */
+  ELECTION_HASH: `${VX_IANA_ENTERPRISE_OID}.4`,
+} as const;
+
+/**
+ * Parsed custom cert fields
+ */
+interface CustomCertFields {
+  component: 'card' | 'admin' | 'central-scan' | 'mark' | 'scan';
+  jurisdiction: string;
+  cardType?: 'sa' | 'em' | 'pw';
+  electionHash?: string;
+}
+
+/**
+ * A schema to facilitate parsing custom cert fields
+ */
+const CustomCertFieldsSchema: z.ZodSchema<CustomCertFields> = z.object({
+  component: z.enum(['card', 'admin', 'central-scan', 'mark', 'scan']),
+  jurisdiction: z.string(),
+  cardType: z.optional(z.enum(['sa', 'em', 'pw'])),
+  electionHash: z.optional(z.string()),
+});
+
+/**
+ * Parses the provided cert and returns the custom cert fields. Throws an error if the cert doesn't
+ * follow VotingWorks's cert format.
+ */
+export async function parseCert(cert: Buffer): Promise<CustomCertFields> {
+  const response = await openssl(['x509', '-noout', '-subject', '-in', cert]);
+
+  const responseString = response.toString();
+  assert(responseString.startsWith('subject='));
+  const certSubject = responseString.replace('subject=', '').trimEnd();
+
+  const certFieldsList = certSubject
+    .split(',')
+    .map((field) => field.trimStart());
+  const certFields: { [fieldName: string]: string } = {};
+  for (const certField of certFieldsList) {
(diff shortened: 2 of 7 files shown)


libs/auth/src/card_reader.ts
```diff
@@ -86,10 +86,6 @@ export class CardReader {
     });
   }
 
-  getReaderStatus(): Reader['status'] {
-    return this.reader.status;
-  }
```
arsalansufi, line 91: Realized that this method wasn't getting used anywhere so I removed it (the `onReaderStatusChange` listener covers all needs for now)

libs/auth/src/certs.ts
```diff
@@ -0,0 +1,111 @@
+import { Buffer } from 'buffer';
```
arsalansufi, line 1: This first commit isn't adding any new code. It's just moving code around and reorganizing helpers. As noted in the PR description and commit message, the breakdown of files is:

- `openssl.ts `- Generic openssl commands (not VotingWorks-specific)
- `piv.ts` - PIV standards (not VotingWorks-specific)
- `certs.ts` - VotingWorks-specific cert logic
- `java_card.ts` - Top-level logic

libs/auth/src/java_card.ts
```diff
@@ -255,6 +156,8 @@ export class JavaCard implements Card {
   }
 
   async checkPin(pin: string): Promise<CheckPinResponse> {
+    await this.selectApplet();
```
arsalansufi, line 196: It's not totally clear to me how frequently applet selection has to happen - I _think_ it's once per card insertion. Removing a card and re-inserting it does require the applet to be selected again.

So while `checkPin` doesn't explicitly require it (the applet will already have been selected in order to read the user), `programCard` definitely does. To play it safe, I'm just starting every public command with applet selection.

libs/auth/src/openssl.ts
```diff
@@ -171,3 +179,52 @@ export async function verifySignature({
     challenge,
   ]);
 }
+
+/**
+ * Creates a cert by signing a public key with a signing cert authority cert and private key. All
+ * key and cert inputs should be in PEM format. The outputted cert will also be in PEM format.
+ */
+export async function createCertBySigningPublicKey({
```
benadida, line 187: nit: a cert is basically always signing a public key, so you could simply call this `createCert`

libs/auth/src/java_card.ts
```diff
@@ -46,6 +64,12 @@ const OPEN_FIPS_201_AID = Buffer.from([
  */
 const DEFAULT_PIN = '000000';
 
+/**
+ * The PIN unblocking key or PUK. Typically a sensitive value but not for us given our modification
+ * to OpenFIPS201 to invalidate the card on PIN reset.
```
benadida, line 69: nit: "invalidate the card" is a bit ambiguous, maybe more "erase the private keys"

libs/auth/src/java_card.ts
```diff
@@ -103,6 +127,16 @@ const GENERIC_STORAGE_SPACE = {
...
 } as const;
 
+/**
+ * Additional constructor inputs that only VxAdmin needs to provide, for card programming
+ */
+interface VxAdminParams {
+  vxAdminCertAuthorityCertPath: string;
+  vxAdminOpensslConfigPath: string;
+  vxAdminPrivateKeyPassword: string;
+  vxAdminPrivateKeyPath: string;
```
benadida, line 137: eventually this is going to be based on the TPM-held vxadmin private key, right? That won't have a password, and we have to look into how OpenSSL can use the TPM.

But, that's a next PR I imagine.

benadida (approved):

libs/auth/src/java_card.ts
arsalansufi: Really helpful reminder! This seems promising: https://github.com/tpm2-software/tpm2-tss-engine

This is going to be hard for me to test without real hardware, but lucky for us, I have a Lenovo ThinkPad on the way 😄

For VM development on Macs, we'll probably still want to support an old-school private key file and password. But I'll think of a way to abstract that away (in follow-up PRs)

libs/auth/src/java_card.ts
arsalansufi: Settled on a combo, to be precise while also making the impact clear to someone with less context on the effect of clearing PIN-gated keys: https://github.com/votingworks/vxsuite/pull/3053/commits/f9fff7927d0faf88e5e845ea69a2879cfd3556e2

There's a more detailed comment later in the code as well: https://github.com/votingworks/vxsuite/blob/f9fff7927d0faf88e5e845ea69a2879cfd3556e2/libs/auth/src/java_card.ts#L567-L576

libs/auth/src/java_card.ts
benadida: worth a chat about how we architect this maybe? I was thinking the TPM-based signature, since we have to do it in a number of different cases, might be provided as a library that calls out into a configurable system executable (e.g. `/vx/code/functions/sign.sh`).

arsalansufi: Ah good point re the need to perform TPM-based signatures in other contexts. Threw some time on the calendar for tomorrow morning - feel free to move around!

libs/auth/src/card_reader.ts
jonahkagan: Nice, I was wondering earlier if both were necessary

libs/auth/src/java_card.ts
```diff
@@ -112,14 +146,17 @@ export class JavaCard implements Card {
   private readonly cardReader: CardReader;
   private cardStatus: CardStatus;
   private readonly jurisdiction: string;
+  private readonly vxAdminParams?: VxAdminParams;
   private readonly vxCertAuthorityCertPath: string;
 
   constructor(input: {
     jurisdiction: string;
+    vxAdminParams?: VxAdminParams;
```
jonahkagan, line 154: If these are only used in `program`, could they be passed in directly there? Also maybe name them something related to programming rather than VxAdmin?

jonahkagan (approved):

libs/auth/src/java_card.ts
arsalansufi: Yeah I went back and forth on whether to name this unit in terms of function or machine but agree that function makes more sense!

In terms of passing params to `program` (and `unprogram`), that would require me to change the general `Card` interface, leaking aspects of and making it more specific to the Java Card + X.509 cert implementation. While I know we'll be getting rid of the memory card implementation of the `Card` interface soon, leaving only one implementation, I'd still like to keep the `Card` interface general. The params are also pretty low-level and not something that I'd want engineers to have to think about when calling `program`.

Issue #2928, opened 2023-01-25
Serialization and validation for CDF election definitions


jonahkagan:
- Edit our JSON schema to only allow our constrained version of the format
- Ensure that constraints not expressible in the schema are enforced via a validation function
- Check some overall CDF constraints (e.g. that `@id` fields are globally unique)

PR #3058, opened 2023-03-01
Ballot Definition CDF constraints

jonahkagan:
## Overview

Task: #2928 

To ensure we only accept CDF ballot definitions that our system can support, we modify the VX schema to accept a constrained version of CDF.

Note that I didn't remove any of the physical ballot layout stuff, since I expect we will be using at least some of it shortly.

In order to do this safely, we add some tests that ensure the schema is valid and accepts a subset of values accepted by the original NIST schema.

_Review by commits_

## Demo Video or Screenshot
N/A

## Testing Plan
Automated tests

## Checklist

- [ ] I have added


libs/types/src/cdf/ballot-definition/index.test.ts
```diff
@@ -1,23 +1,44 @@
...
-  );
-  const vxJson = readFileSync(join(__dirname, './vx-schema.json'), 'utf-8');
   const generatedTypes = readFileSync(join(__dirname, './index.ts'), 'utf-8');
   const out = mockWritable();
   buildSchema(nistXsd, vxJson, out).unsafeUnwrap();
   const expectedTypes = out.toString();
   expect(generatedTypes).toEqual(expectedTypes);
 });
+
+test('VX schema is a valid JSON schema', () => {
```
adghayes, line 38: Looks like test is really "VX schema and NIST schema are valid JSON schemas", although we should hope the NIST schema is valid.

adghayes (approved):
LGTM! Thank for for the comments in `cdf_schema_utils.test.ts`, it made it easy to understand the intent. Also cool to automatically check for unused definitions. I don't have as much deep a context as you on all the things being removed so mostly skimmed over that.

What do you think the MO will be when we eventually want to add some things back in? I wonder if we'll find ourselves looking back at this PR.

jonahkagan:
> What do you think the MO will be when we eventually want to add some things back in? I wonder if we'll find ourselves looking back at this PR.

Yeah, that's why I tried to chunk up the removals in logical commits so we could easily look back and undo one of them.

libs/types/src/cdf/ballot-definition/index.test.ts
jonahkagan: Haha yeah I would hope! Updated the test title to make it less confusing: 2cff1610f

jonahkagan:
@adghayes I added two more commits that enforce uniqueness of `@id` fields (and deal with the consequences). Could you take a look at those? Forgot to do them in the first go-round

adghayes (approved):
Could you add a comment that specifies that every ID in the document must be unique, as opposed to just unique within a list of items or a class of objects? Or maybe instead a test to that affect, since the current test doesn't represent that possibility? Some comment in `safeParseCdfBallotDefinition` or over `findDuplicateIds` would do.

Now CDF imports are required to have unique IDs, and VX definitions converted from CDF definitions will have unique IDs. But things loaded from VX definitions do not have that guarantee. So when we go to export option IDs in the cast vote record export, for example, we'd have to have branching logic based on whether we expect unique IDs. In which case I wonder if we should add a `definitionSourceType` field to the `ElectionDefinition` type. Another option is adding the requirement for VX Definitions too, which would require changing all the fixtures and giving the yes/no contests unique IDs.

Functionality LGTM, appreciate the utilities you've been adding to `basics`.

jonahkagan:
Great call, added a comment and made the test more explicit: 56fe5826d

Re: VX definitions: I think it would make the most sense to just enforce the same requirement unless it's prohibitively difficult to change the existing definitions. Maybe we just give those a pass somehow?

PR #3064, opened 2023-03-02
handle single sided blank ballots properly

carolinemodic:
In a different part of the code this is the logic for figuring out if a sheet requires adjudication, both pages must be "blank" but any pages with 0 possible marks are automatically considered blank, so I'm updating this part of the code that tells the state machine what to tell the frontend about the adjudication reason to do the same thing

### DIFF
--- a/services/scan/src/precinct_scanner_interpreter.ts
+++ b/services/scan/src/precinct_scanner_interpreter.ts
@@ -89,12 +89,14 @@ function combinePageInterpretationsForSheet(
     let reasons: AdjudicationReasonInfo[];
     // If both sides are blank, the ballot is blank
     if (
-      frontReasons.some(
+      (frontReasons.some(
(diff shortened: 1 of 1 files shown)


kshen0 (approved):
LGTM. Out of curiosity, why does a ballot with 0 marks not already evaluate true for `reason.type === AdjudicationReason.BlankBallot`? Is `0 marks` not exactly the same as `BlankBallot`, and therefore we need to keep them separate?

benadida (approved):

carolinemodic:
> not exactly the same as `BlankBallot`

0 marks here is if there are 0 optionals for marks like no contests on the page at all, versus BlankBallot which is when there are contests but none of them are marked. I guess we could just evaluate to BlankBallot when there are no contests in the first place I don't really know where that code lives or why we don't do that

PR #3066, opened 2023-03-02
Fix blank ballot adjudication on precinct scanner

carolinemodic:
## Overview
Fixes a bug where if a ballot is blank, the BlankBallot adjudication reason will not work properly if there are only contests on one side of the ballot in precinct-scanner. It will be marked as requiring adjudication as the code in vx_interpret does the right thing, but this part of the code does not (why is this logic duplicated?). Testing this would be complicated as I don't think we have an election definition that has this case so I'm just fixing the bug right now. 

## Checklist

- [ ] I have added

### DIFF
--- a/apps/scan/backend/src/interpret.ts
+++ b/apps/scan/backend/src/interpret.ts
@@ -90,12 +90,14 @@ function combinePageInterpretationsForSheet(
     let reasons: AdjudicationReasonInfo[];
     // If both sides are blank, the ballot is blank
     if (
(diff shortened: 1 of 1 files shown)


adghayes (approved):
LGTM as a quick fix. I wonder if we should be processing the "reason" or interpretation of pages with no contests as something specifically typed higher upstream, so we don't have to rely on these checks. I didn't dig into the logic further upstream so am not sure.

PR #3075, opened 2023-03-06
Add libs/auth tests, part 1

arsalansufi:

## Overview

I'm finally going back and adding all the `libs/auth` tests that I've been punting on 😅

This PR covers some of the lower-level files: `apdu.ts`, `card_reader.ts`, `certs.ts`, and `piv.ts`. I'll cover the rest of the files and reach 100% code coverage in a next PR (or PRs).

## Testing

- [x] How meta

## Checklist

### DIFF
--- a/libs/auth/src/certs.test.ts
+++ b/libs/auth/src/certs.test.ts
@@ -18,35 +18,35 @@ import { openssl } from './openssl';
 
 jest.mock('./openssl');
 
-const mockCert = Buffer.from([]);
-const mockElectionHash =
+const cert = Buffer.from([]);
+const electionHash =
   '43939f8d6b94dd85827c1d151d0b75f4617e934979d53b6d5ce2abf4535a93d4';
-const mockJurisdiction = 'ST.Jurisdiction';
+const jurisdiction = 'ST.Jurisdiction';
 
 test.each<{ subject: string; expectedCustomCertFields: CustomCertFields }>([
   {
     subject:
       'subject=C = US, ST = CA, O = VotingWorks, ' +
       '1.3.6.1.4.1.59817.1 = card, ' +
-      `1.3.6.1.4.1.59817.2 = ${mockJurisdiction}, ` +
+      `1.3.6.1.4.1.59817.2 = ${jurisdiction}, ` +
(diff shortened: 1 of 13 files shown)


libs/auth/src/certs.test.ts
```diff
@@ -0,0 +1,237 @@
...
+  {
+    user: {
+      role: 'poll_worker',
+      electionHash: mockElectionHash,
+    },
+    expectedSubject:
+      '/C=US/ST=CA/O=VotingWorks' +
+      '/1.3.6.1.4.1.59817.1=card' +
+      `/1.3.6.1.4.1.59817.2=${mockJurisdiction}` +
+      '/1.3.6.1.4.1.59817.3=pw' +
```
jonahkagan, line 222: Is it useful to have the card type be a two-character abbreviation? Seems harder to read, so I was wondering if there's an intentional tradeoff

jonahkagan (approved):

libs/auth/src/certs.test.ts
arsalansufi: Good q! I'd been following the plan laid out by Ben in [the certificates section of this doc](https://docs.google.com/document/d/1-6K-ayWUAIQAO3n50PZAAODVaXsBUkZhGoXcFNja6aI/edit#heading=h.j01awri45014) and hadn't questioned this. My initial read was that the two-character abbreviation was a space-saving measure, but that's truly inconsequential in the context of Java Cards, which give us so much more space than our old memory cards.

Ultimately, the decision doesn't matter too much since the we're not going to be working with the underlying card certificates other than in this file, but I'm very open to removing the abbreviations in favor of complete strings.

cc @benadida in case there are any other reasons for the user role abbreviations

arsalansufi: Going to merge this PR but will revisit this comment in a follow-up PR if need be!

benadida: My goal was space-saving in cert, which makes a bigger difference than it might seem because cert format is highly compressed. That said, it's still not a huge difference, as @arsalansufi points out, so I'm game to change these.

PR #3085, opened 2023-03-07
Adam/custom circleci images

amcmanus:
## Overview

1. Updates our CircleCI config to use our custom images, to more closely reflect our production release.
2. Uses direct downloads of Node to ensure our explicit version remains accessible, even after the apt repo removes it. (We should still make every effort to regularly update, but that is outside the scope of this PR.)
3. Updates a few tests that broke due to date formatting differences between Node 16.14.2 --> 16.19.1

## Testing Plan

Run CircleCI tests and confirm Success.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,13 +6,13 @@ orbs:
 executors:
   nodejs-browsers:
     docker:
-      - image: adammcmanus/cimg-debian11-browsers:1.0.0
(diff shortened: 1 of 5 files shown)


carolinemodic (approved):
exciting!

eventualbuddha (commented):
A little late to the party but LGTM 🎉

Issue #2746, opened 2022-10-25
Touch area size


mattroe:
Confirm we meet the VVSG touch target area size requirements: https://docs.google.com/document/d/1A5UX8Z0v12Vvf_jVOUmrOEvUNah2Xxmtltll6ZJW6Ps/edit#heading=h.wl5r11o23kzl

Issue #1713, opened 2022-04-19
Make it possible to retrieve buttons via `getByRole('button', { name: '...' })` in tests


arsalansufi:
The current inability to do so for many buttons may be indicative of an accessibility issue. Much more context can be found in this GitHub comment thread: https://github.com/votingworks/vxsuite/pull/1708#discussion_r853293129

PR #3086, opened 2023-03-07
Update Button component to satisfy VVSG design requirements

kofi-q:
## Overview
Related to: https://github.com/votingworks/vxsuite/issues/2743, https://github.com/votingworks/vxsuite/issues/2744
Resolves https://github.com/votingworks/vxsuite/issues/2746, resolves https://github.com/votingworks/vxsuite/issues/1713

Sorry for the PR size - tried to keep it down to essential updates, but a lot of it is snapshot/test updates and repetitive find-and-replace updates of the variant props 😅 

- Updated the shared `<Button>` component to use the new VVSG-compatible themes
  - The `forwardRef` wrapper function we were using before was getting in the way of rendering within storybook and didn't allow for adding type safety to the `value` and `onPress(value)` props, so just updated from functional component to class component instead.
  - Moved away from the boolean `primary`, `warning` etc props to a single enum `variant` prop to make configuration easier and avoid confusing cases where we were rendering `<Button primary danger>`, for example.
  - Updating the medium contrast `'warning'` colour to dark gray per Figma feedback, since it looks more brown than yellow/amber.
  - Simplified the event handling a bit so that `onPress` is called with whatever's provided in the `value` prop (so clients don't have to assign arbitrary `data-*` attributes and pull them out of an event object - this enables/simplifies some upcoming work around the ballot choice buttons and the segmented button.
- Removed duplicate `<Button>` and `<LinkButton>` components from `central-scan`, to avoid confusion.
- Removed the `onPress` prop from `<LinkButton>`, since it's only ever used when we don't use the `to` prop (and the `to` prop is ignored anyway), in which case, it's just a `<Button>` - made updating the LinkButton interface a little easier.
- Had to switch some old tests to use `getByRole('button', ...)` instead of `getByText`, since the button now wraps text content in a child `span` to allow for adding an icon on coloured buttons. Went ahead and added convenience `[get|find]Button()` and `[getAll|findAll]Buttons()` methods to the shared render function's result, [as suggested by Drew](https://votingworks.slack.com/archives/CEL6D3GAD/p1675981808029889?thread_ts=1675966572.953689&cid=CEL6D3GAD), to make our lives a little easier.
- Noticed a case where the `expectPrint` utilities weren't properly tearing down the print DOM when a test exception occurs, which got in the way of debugging, so just updated those while I was at it.
- Also ran into an edge case where react-modal doesn't remove the `aria-hidden` attribute from the root element in the DOM after unmounting the modal when running in jest/jsdom, so had to add a hack to remove it manually, since it prevents us from using `getByRole()` (anything within an `aria-hidden` element is hidden from the accessibility tree.

### Interaction:
https://user-images.githubusercontent.com/264902/223770402-a1c16bbb-d204-4b55-a524-4c2720c6e971.mov

## Styling remains unchanged in 'legacy' mode:
Screenshot 2023-03-07 at 16 41 41

## Testing Plan
- Updated tests in cases where we need to verify state on the button element (no need to update the ones that just fire click events for now, since firing a click event on a child of a `<button>` triggers a button click event.
- Added tests to verify button behaviour and API method and tests to verify that styling changes with different themes without testing specific values

## Checklist

- [n/a] I have added

- [n/a] I have added the "user_facing_change" label to this PR to automate an


adghayes:
@kofi-q just digging into the code review now. From a design perspective, two questions from looking in storybook:

- Currently if a button is disabled, it's size will change. I don't know if we have many cases of buttons enabling or disabling on screen in real-time, but it could lead to screens moving around a bit if we're not careful.
- The `touch-action: manipulation;` is cool for bordered buttons but for solid buttons it looks a bit strange to me. I'm not picky about it but I wonder if there are other options. In both the border and non-border case, it would be nice if the expanded border / edge was rounded like the button itself. 

Otherwise, just digging into this, loving the look and reorganization so far.

libs/ui/src/button.tsx
```diff
@@ -1,122 +1,353 @@
...
+  'previous',
+  'previousPrimary',
+  'primary',
+  'regular',
+  'secondary',
+  'settings',
+  'warning',
+] as const;
+
+export type ButtonVariant = typeof ALL_BUTTON_VARIANTS[number];
```
adghayes, line 25: TIL a new Typescript pattern.

libs/ui/src/button.tsx
```diff
@@ -1,122 +1,353 @@
...
+    color: 'accentPrimary',
+    icon: Icons.Previous,
+    isSolidColor: true,
+  },
+  primary: { color: 'accentPrimary', isSolidColor: true },
+  regular: { color: 'foreground' },
+  secondary: { color: 'accentSecondary', isSolidColor: true },
+  settings: { color: 'foreground', icon: Icons.Settings },
+  warning: { color: 'accentWarning', icon: Icons.Warning, isSolidColor: true },
+};
```
adghayes, line 98: I appreciate this organization! I was wondering how I was going to keep track of all this.

libs/ui/src/button.test.tsx
```diff
@@ -111,22 +259,3 @@ describe('renders Button', () => {
...
-
-  // Click is disabled
-  userEvent.click(button);
-  expect(onPress).toHaveBeenCalledTimes(0);
-
-  // Touch is disabled
-  fireEvent.touchStart(button, createTouchStartEventProperties(100, 100));
-  fireEvent.touchEnd(button, createTouchEndEventProperties(100, 100));
-  expect(onPress).toHaveBeenCalledTimes(0);
-});
```
adghayes, line 132: Rationale for removing this test?

adghayes: Oh, it looks like it moved up into the other `disabled` test? Seems good. I remember when we got pinged on this bug so just want to make sure we have coverage.

libs/ui/src/themes/render_with_themes.tsx
```diff
@@ -1,16 +1,48 @@
...
+  findAllButtons(
+    buttonText: string | RegExp,
+    opts?: ButtonQueryOptions
+  ): ReturnType<queries.FindAllByRole>;
+  findButton(
+    buttonText: string | RegExp,
+    opts?: ButtonQueryOptions
+  ): ReturnType<queries.FindByRole>;
+};
+
```
adghayes, line 45: This is great! Thanks for making this happen!

I would prefer, however, that we extend the `screen` option to include these queries rather than destructuring the render result in tests. Using `screen` is [recommended practice](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen) and I think just easier. 

It doesn't matter much to me if you change the tests you've updated to use the `screen` object - I think it's OK that we update the tests as we go, and I'm sure you've already done a lot of fishing in the all the frontends. But I like the idea of putting these queries on the `screen` in this PR or a follow-up PR. 

Thoughts?

libs/test-utils/src/expect_print.ts
```diff
@@ -117,8 +119,12 @@ function inspectPrintedElement(inspect: InspectPrintFunction): void {
...
   });
-  inspect(renderResult, lastPrintOptions);
-  renderResult.unmount();
+
+  try {
+    inspect(renderResult, lastPrintOptions);
+  } finally {
+    renderResult.unmount();
+  }
 }
```
adghayes, line 128: Ugh, sorry you had to get tripped up by this. Thanks for making it better.

libs/ui/src/button.tsx
```diff
@@ -1,122 +1,353 @@
...
+} & ( // Require a matching `value` if the provided click handler expects a value.
+    | {
+        onPress: ClickHandler;
+        value?: never;
+      }
+    | {
+        onPress: TypedClickHandler<T>;
+        value: T;
+      }
+  );
```
adghayes, line 68: Great.

apps/scan/frontend/src/app.test.tsx
```diff
@@ -58,15 +58,31 @@ function renderApp(props: Partial<AppProps> = {}) {
...
+ * We're removing the attribute here to make sure our getByRole queries work
+ * properly.
+ */
+async function hackActuallyCleanUpReactModal() {
+  await waitFor(() => {
+    expect(screen.queryByRole('alertdialog')).not.toBeInTheDocument();
+  });
+
+  window.document.body.firstElementChild?.removeAttribute('aria-hidden');
 }
```
adghayes, line 86: There's definitely some history on this in Slack. Hack looks great.

apps/mark/frontend/src/pages/review_page.tsx
```diff
@@ -550,17 +529,13 @@ export function ReviewPage(): JSX.Element {
...
         </ButtonFooter>
       ) : (
         <Sidebar
           footer={
             <React.Fragment>
-              <ButtonFooterLandscape>
-                {languageSettingsButton}
-                {settingsButton}
-              </ButtonFooterLandscape>
+              <ButtonFooterLandscape>{settingsButton}</ButtonFooterLandscape>
```
adghayes, line 538: Remind me the story with the `languageSettingsButton`?

apps/mark/frontend/src/components/candidate_contest.tsx
```diff
@@ -539,9 +531,7 @@ export function CandidateContest({
           actions={
             <React.Fragment>
               <Button
-                primary={
-                  normalizeCandidateName(writeInCandidateName).length > 0
-                }
+                variant="primary"
```
adghayes, line 534: Does this need to be conditionally primary? Style change?

kofi-q:
> - I see that if a button is disabled, its size will change. I don't know if we have many cases of buttons enabling or disabling on screen in real-time, but it could lead to screens moving around a bit if we're not careful.

Ugh yeah, the disabled button's been haunting me for a while now 😅 - not completely happy with it in its current state, but will look into at least trying reduce the size change.
Wouldn't give it too much priority at this stage - want to iterate on it later, post NH demo, since Jonah also mentioned it doesn't really look disabled to him. At the moment, it's not used in any voter-facing flows, so will try to address in a follow-up

> - The touch-action: manipulation; is cool for bordered buttons but for solid buttons it looks a bit strange to me. I'm not picky about it but I wonder if there are other options. In both the border and non-border case, it would be nice if the expanded border / edge was rounded like the button itself.

Yeah, the outer border is a normal border and the "grow" effect is via an inset shadow, so that interaction is a little weird. I haven't iterated since moving from all the buttons being in the outlined style to solid buttons, so definitely room for improvement there. My initial approach was the usual shadow that shrinks on press, but that would only really be viable in the low-contrast mode. Can just disable the interaction styles for now until we have implement something cleaner

adghayes (commented):
@kofi-q this looks great and I agree with all the little decisions you had to make along the way. I appreciate how this improves the type safety and easy-of-use of the `Button` component. And with the themes it just looks better.

I posted a couple of appearance questions in a separate comment and have a few questions as part of this code review. The biggest one, probably, in terms of work, is my feedback on the `getButton` extension. Having it from `render` is an improvement that outweighs the testing disadvantage of not having it on `screen`, so it should not slow down merging this. But I do think it would be great to have it on `screen` and also accessible from `within`, although I'm not sure how to do the `within` part :).

adghayes:
@kofi-q up to you what you want to do on the `touch-action` and the disabled cases, I just wanted to point them out. Sounds like you're on top of it!

apps/mark/frontend/src/pages/review_page.tsx
kofi-q: It's something we'll have to implement eventually, but I noticed it's currently just a placeholder and not rendered, so removed in the meantime to avoid the maintenance cost and just revisit when we're ready for a full implementation

libs/ui/src/button.tsx
kofi-q: Same 😄

adghayes (approved):
LGTM!

kofi-q:
Side note: one unfortunate thing I'm noticing is that `getByRole()` calls take a lot longer than `getByText()` and the test cases I had to update are significantly slower now :/

Found this thread discussing the issue and it seems like it's a necessary evil, but potentially something we can mitigate by paring down styles for test renders and maybe limiting usage to smaller render trees(?): https://github.com/testing-library/dom-testing-library/issues/552

jonahkagan:
@kofi-q If `getByRole` is super slow, maybe we can use `getByText` under the hood of `getButton` and then check that the element it gets is actually a button afterwards.

kofi-q:
> If getByRole is super slow, maybe we can use getByText under the hood of getButton and then check that the element it gets is actually a button afterwards.

Yeah, that was my first thought as well - that does side-step some of the a11y checks that `getByRole` is doing though, which go beyond checking that it's a button element (you could theoretically have a non-`<button>` element with a button role). I wouldn't want us to go down the path of rewriting some of that logic. So, agreed that a viable option is to try writing a lighter weight `getByRole` for buttons

PR #3087, opened 2023-03-07
Add Batch Info to CVR

adghayes:
Per [internal discussion](https://votingworks.slack.com/archives/CEL6D3GAD/p1678208279298859) about making our cast vote records more audit friendly, this change adds the following batch information to the CDF CVR (still feature-flagged):

- SequenceId (or batchNumber)
- Batch Start Time
- Batch End Time (if exists)
- Batch Size

One wrinkle on precinct scanner may be that we may export cast vote records before the polls are officially closed, so the last (and likely only) batch will not have an end time. I don't think solving that problem is within the scope of this PR if it is important.

### DIFF
--- a/libs/types/src/cdf/cast-vote-records/vx-schema.json
+++ b/libs/types/src/cdf/cast-vote-records/vx-schema.json
@@ -1203,7 +1203,15 @@
     },
     "CVR.vxBatch": {
       "description": "Entity containing metadata about a scanned batch. Cast vote records link to batches via CVR::BatchId.",
-      "required": ["@id", "@type", "BatchLabel"],
+      "required": [
+        "@id",
+        "@type",
+        "BatchLabel",
+        "SequenceId",
+        "StartTime",
+        "Size",
+        "CreatingDeviceId"
+      ],
       "additionalProperties": false,
       "properties": {
         "@id": {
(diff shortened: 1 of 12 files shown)


adghayes:
cc @ginvdr

libs/types/src/cdf/cast-vote-records/vx-schema.json
```diff
@@ -1214,7 +1222,31 @@
...
+        },
+        "Size": {
+          "description": "The number of sheets included in a batch.",
+          "type": "integer"
+        },
+        "CreatingDeviceId": {
+          "description": "The tabulator that created the batch.",
+          "type": "string",
+          "refTypes": ["CVR.ReportingDevice"]
         }
```
adghayes, line 1250: Open to naming changes about any of these fields, was just going with what made sense to me and fit within style of the NIST specification cc @jonahkagan @ginvdr.

libs/types/src/cdf/cast-vote-records/vx-schema.json
```diff
@@ -1214,7 +1222,31 @@
...
+          "description": "The start time of the batch. On a precinct scanner, the start time is when the polls are opened or voting is resumed. On a central scanner, the start time is when the user initiates scanning a batch.",
+          "type": "string",
+          "format": "date-time"
+        },
+        "EndTime": {
+          "description": "The end time of the batch. On a precinct scanner, the end time is when the polls are closed or voting is paused. On a central scanner, the end time is when a batch scan is complete",
+          "type": "string",
+          "format": "date-time"
+        },
+        "Size": {
```
jonahkagan, line 1242: maybe `NumberSheets` would be more CDF-y (thinking of `NumberVotes`)

jonahkagan (approved):

libs/types/src/cdf/cast-vote-records/vx-schema.json
adghayes: Great!

PR #3089, opened 2023-03-08
fix issue with double disconnect for custom

carolinemodic:
Fixes https://votingworks.slack.com/archives/CEL6D3GAD/p1678304275934649 
This triggers the error states and crazy things were happening for reasons I don't totally understand when we were disconnecting twice. This now fixes the issue, and simplifies the state machine.


eventualbuddha:
@carolinemodic rebase?

Issue #1338, opened 2022-01-11
CVR: sequence of the ballot in the batch


jonahkagan:
sequence of the corresponding voted ballot in the batch, when applicable

adghayes:
We have this marked as completed in the matrix but I do not believe it is completed.

adghayes:
Closed by #3283.

PR #3101, opened 2023-03-09
Add Batch Sequence ID to CDF CVR

adghayes:
This addresses #1338, meeting VVSG 2.0 1.1.5-G.7, but does not close it since CDF CVRs will not be the default until further along in this project. _For the central scanner only_, adds a the ordinal position of the ballot in its batch under CDF field `BatchSequenceId`.


apps/central-scan/backend/src/store.ts
```diff
@@ -982,10 +980,12 @@ export class Store {
       backInterpretationJson: string;
       frontNormalizedFilename: string;
       backNormalizedFilename: string;
+      batchSequenceId: number;
     }>) {
       yield {
         id: row.id,
         batchId: row.batchId,
+        batchSequenceId: row.batchSequenceId,
```
jonahkagan, line 988: Can we call this something more clear like `indexInBatch` in the code (even if we have to use `BatchSequenceId` in the CDF CVR)?

libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -39,6 +39,7 @@ import { buildCastVoteRecordReportMetadata } from './build_report_metadata';
 export interface ResultSheet {
   readonly id: Id;
   readonly batchId: Id;
+  readonly batchSequenceId?: number; // central scanner only
```
jonahkagan, line 42: Why central scanner only? Can you add some more context to this comment?

jonahkagan (approved):

libs/backend/src/scan/cast_vote_records/export.ts
adghayes: It would compromise voter privacy in a precinct scanner, can add more of a comment.

apps/central-scan/backend/src/store.ts
adghayes: Sure!

PR #3105, opened 2023-03-09
Add Total Write-In Count

adghayes:
We've discussed whether this is necessary, but might as well just do it as long as it's easy per discussions with Jessica. See discussion in #1240. We already have a count of write-ins per contest if any exist. This adds a cumulative count of write-ins at the at the CVr snapshot level, if it exists.

### DIFF
--- a/libs/backend/src/scan/cast_vote_records/build_cast_vote_record.ts
+++ b/libs/backend/src/scan/cast_vote_records/build_cast_vote_record.ts
@@ -59,8 +59,8 @@ function buildCVRBallotMeasureContest({
   return {
     '@type': 'CVR.CVRContest',
     ContestId: contest.id,
-    Overvotes: overvoted ? 1 : undefined,
-    Undervotes: undervoted ? 1 : undefined,
+    Overvotes: Math.max(vote.length - 1, 0),
+    Undervotes: Math.max(1 - vote.length, 0),
     Status: overvoted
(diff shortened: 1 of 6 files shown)


libs/backend/src/scan/cast_vote_records/build_cast_vote_record.ts
```diff
@@ -489,6 +512,7 @@ export function buildCastVoteRecord({
               ballotMarkingMode: 'machine',
             },
           }),
+          vxWriteIns: writeInCount > 0 ? writeInCount : undefined,
```
jonahkagan, line 515: Why does it need to be greater than 0 to be included? 0 seems like a valid count

adghayes: I'm sticking with the pattern I established in the other parts of this file of not including keys with undefined or empty values. Just want to be consistent. Can switch but would want to switch them all.

jonahkagan: Ok interesting I didn't catch that before. I think that general approach sounds fine. I would only recommend tweaking it for counts so we can communicate a message like: "yes, we did count this, but the count was 0" rather than "we don't know what the count was". I think in the other cases using `undefined` is clear.

adghayes: That's a good point, I've updated it in this PR.

jonahkagan (approved):

PR #3107, opened 2023-03-09
Add libs/auth tests, part 2

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3057

Another day 🌅, another `libs/auth` tests PR. This PR specifically adds tests for openssl helpers.

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ N/A
- [ ] ~I have added JSDoc comments to any newly introduced exports~ N/A

PR #3115, opened 2023-03-10
feat(test-utils): Add mockFunction.expectRepeatedCallsWith

jonahkagan:
## Overview

Support expecting repeated calls while still maintaining the requirements that every actual call match up with an expected call.

Also updates some app frontend tests to use this new ability!

_Review by commits_

## Demo Video or Screenshot
N/A

## Testing Plan
- Added automated tests

## Checklist

- [ ] I have added

### DIFF
--- a/apps/admin/frontend/test/render_in_app_context.tsx
+++ b/apps/admin/frontend/test/render_in_app_context.tsx
@@ -41,7 +41,7 @@ import {
 } from '../src/lib/backends';
 import { mockUsbDrive } from './helpers/mock_usb_drive';
 import { ApiClient, ApiClientContext } from '../src/api';
-import { ApiMock, createMockApiClient } from './helpers/api_mock';
+import { ApiMock } from './helpers/api_mock';
 
 export const eitherNeitherElectionDefinition =
   electionWithMsEitherNeitherDefinition;
@@ -82,7 +82,12 @@ export function renderRootElement(
     backend = new ElectionManagerStoreMemoryBackend(),
     logger = fakeLogger(),
     storage = new MemoryStorage(),
-    apiClient = createMockApiClient(),
+    // If there's no apiClient given, we don't want to create one by default,
+    // since the apiClient needs to have assertComplete called by the test. If
+    // the test doesn't need to make API calls, then it should not pass in an
+    // apiClient here, which will cause an error if the test tries to make an
+    // API call.
+    apiClient,
     // TODO: Determine why tests fail when using useErrorBoundary = true
     queryClient = new QueryClient({
       defaultOptions: {
--- a/apps/admin/frontend/test/render_in_app_context.tsx
+++ b/apps/admin/frontend/test/render_in_app_context.tsx
@@ -22,7 +22,7 @@ import {
 import { fakeLogger, Logger, LogSource } from '@votingworks/logging';
 
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { QUERY_CLIENT_DEFAULT_OPTIONS, UsbDrive } from '@votingworks/ui';
+import { UsbDrive } from '@votingworks/ui';
 import {
   fakeElectionManagerUser,
   fakeSystemAdministratorUser,
@@ -40,7 +40,7 @@ import {
   ElectionManagerStoreMemoryBackend,
 } from '../src/lib/backends';
 import { mockUsbDrive } from './helpers/mock_usb_drive';
-import { ApiClient, ApiClientContext } from '../src/api';
+import { ApiClient, ApiClientContext, createQueryClient } from '../src/api';
 import { ApiMock } from './helpers/api_mock';
 
 export const eitherNeitherElectionDefinition =
@@ -88,16 +88,7 @@ export function renderRootElement(
     // apiClient here, which will cause an error if the test tries to make an
     // API call.
     apiClient,
-    // TODO: Determine why tests fail when using useErrorBoundary = true
-    queryClient = new QueryClient({
-      defaultOptions: {
-        ...QUERY_CLIENT_DEFAULT_OPTIONS,
-        queries: {
-          ...(QUERY_CLIENT_DEFAULT_OPTIONS.queries ?? {}),
-          useErrorBoundary: false,
-        },
-      },
-    }),
+    queryClient = createQueryClient(),
   }: {
     backend?: ElectionManagerStoreBackend;
     logger?: Logger;
--- a/apps/scan/frontend/test/helpers/mock_api_client.tsx
+++ b/apps/scan/frontend/test/helpers/mock_api_client.tsx
@@ -46,13 +46,7 @@ export const statusNoPaper: PrecinctScannerStatus = {
   ballotsCounted: 0,
 };
 
-type MockApiClient = Omit<
-  MockClient<Api>,
-  'getAuthStatus' | 'saveScannerReportDataToCard'
-> & {
-  // Because this is polled so frequently, we opt for a standard jest mock instead of a
-  // libs/test-utils mock since the latter requires every call to be explicitly mocked
-  getAuthStatus: jest.Mock;
+type MockApiClient = Omit<MockClient<Api>, 'saveScannerReportDataToCard'> & {
   // Because the values passed to this are so complex, we opt for a standard jest mock instead of a
   // libs/test-utils mock since the latter requires exact input matching and doesn't support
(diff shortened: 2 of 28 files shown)


apps/scan/frontend/test/helpers/mock_api_client.tsx
```diff
@@ -81,9 +72,7 @@ export function createApiMock() {
   const mockApiClient = createMockApiClient();
 
   function setAuthStatus(authStatus: InsertedSmartCardAuth.AuthStatus): void {
-    mockApiClient.getAuthStatus.mockImplementation(() =>
-      Promise.resolve(authStatus)
-    );
+    mockApiClient.getAuthStatus.expectRepeatedCallsWith().resolves(authStatus);
```
jonahkagan, line 75: If we like this approach, I will update the rest of the apps accordingly. Just wanted to get feedback first.

libs/test-utils/src/mock_function.ts
```diff
@@ -167,22 +249,54 @@ export function mockFunction<Func extends AnyFunc>(
...
+      ...actualCallsWithTheirExpectedCalls,
+      ...unusedExpectedCalls.map((expectedCall) => ({
+        actualCall: undefined,
+        expectedCall,
+      })),
+    ];
+
+    // We can only accurately report a correspondence of expected and actual
+    // calls up to the first mismatched call. After that, we don't know which
+    // expected calls were used for which actual calls.
```
jonahkagan, line 286: This is the only major change to existing functionality. We now no longer show the calls that occurred after a mismatched call. Before, since we expected a 1-1 mapping of actual calls to expected calls, it was easy to just zip the two lists together. Now that there are repeated calls, it's basically impossible to show what was "supposed" to happen after a mismatch.

arsalansufi: Makes sense! Seeing the calls after the mismatch doesn't provide that much value anyway in my opinion

libs/test-utils/src/mock_function.ts
```diff
@@ -114,22 +128,69 @@ export function mockFunction<Func extends AnyFunc>(
...
+    const previousActualCall = state.actualCalls[actualCallIndex - 1];
+    if (!previousActualCall) {
+      return state.expectedCalls[0];
+    }
+
+    // If the previous actual call didn't match any expected call, then we don't
+    // know what the next expected call should be.
+    if (previousActualCall.matchingExpectedCallIndex === undefined) {
+      throw new MockFunctionError(
+        `Previous call to mock function did not match expected calls`
```
arsalansufi, line 144: nit: Can use standard quotes here instead of backticks

libs/test-utils/src/mock_function.ts
```diff
@@ -114,22 +128,69 @@ export function mockFunction<Func extends AnyFunc>(
...
+      return state.expectedCalls[nextExpectedCallIndex];
+    }
+
+    // Otherwise, maybe the previous expected call was a repeated expected call,
+    // so we can use that
+    const previousExpectedCall =
+      state.expectedCalls[previousActualCall.matchingExpectedCallIndex];
+    if (assertDefined(previousExpectedCall).repeated) {
+      return previousExpectedCall;
+    }
```
arsalansufi, line 161: I'm having a little trouble gleaning what's supported looking at this logic, but is a repeated call allowed to be interleaved with non-repeated calls, e.g.

Given the following expectations:
- Expect repeated calls for auth status
- Expect call to endpoint 1
- Expect call to endpoint 2

Is this sequence valid?
- Call for auth status
- Call to endpoint 1
- Call for auth status
- Call to endpoint 2

We're probably in agreement that this should be valid. I'm just having a hard time seeing whether this is supported looking at this logic

apps/scan/frontend/test/helpers/mock_api_client.tsx
```diff
@@ -61,11 +55,8 @@ type MockApiClient = Omit<
 
 function createMockApiClient(): MockApiClient {
   const mockApiClient = createMockClient<Api>();
-  // For some reason, using an object spread to override the getAuthStatus method breaks the rest
-  // of the mockApiClient, so we override like this instead
-  (mockApiClient.getAuthStatus as unknown as jest.Mock) = jest.fn(() =>
-    Promise.resolve({ status: 'logged_out', reason: 'no_card' })
-  );
+  // Because mockApiClient uses a Proxy under the hood, we add an explicit field
+  // to the object to override the Proxy implementation.
```
arsalansufi, line 59: Thanks for updating my "For some reason" comment 😅

apps/scan/frontend/test/helpers/mock_api_client.tsx
arsalansufi: Overall, I like the interface!

My main concern is just that the implementation of the interface is fairly hard to reason about. We shouldn't have to update the underlying implementation often, so it's not a big deal. But I am once again wondering whether it's the right call to reinvent the (Jest) wheel or not. I do still see the merits of strictly enforcing the call pattern, as brought up in this thread: https://github.com/votingworks/vxsuite/pull/3107#discussion_r1131736705. So I'm happy to continue down this line for now

jonahkagan: Yeah I agree it's hard to reason about! I puzzled a bit over ways to make it clearer.

I think the core difference between the Jest mock function and this mock is that Jest mocks differentiate the mock implementation from assertions about what calls were expected. Whereas with this mock, they are one and the same - the expected calls provide the "implementation" (return values). I think it's a pretty major difference that makes me feel like it's a new wheel, not necessarily a reinvention.

That being said, I'm still very open to this not being worthwhile. It's definitely a bit confusing having two different types of mock functions out there. (Though we did have that previously with fetch-mock anyway.)

I feel similar to you - down to keep experimenting with this while holding open the possibility that's it's not worth it.

arsalansufi: > Whereas with this mock, they are one and the same - the expected calls provide the "implementation" (return values). I think it's a pretty major difference that makes me feel like it's a new wheel, not necessarily a reinvention.

Thanks for that framing! Agree that that's more substantial than my earlier framing of just strictly enforcing the call pattern

> I feel similar to you - down to keep experimenting with this while holding open the possibility that's it's not worth it.

:100:

libs/test-utils/src/mock_function.ts
arsalansufi: Ah yes of course - thanks for clarifying that! Lost sight of that as I was in the weeds thinking about the current code block

libs/test-utils/src/mock_function.ts
jonahkagan: Thanks, fixed

jonahkagan:
@arsalansufi I updated apps/admin and apps/central-scan to use `expectRepeatedCallsWith` for `getAuthStatus`. In the process, I figured out why the error boundary setting was causing test failures (see the commit log for full explanation).

I also realized that we can't update apps/mark to use this approach yet because of the intersection of a few factors:
- We have to pass the election hash to `getAuthStatus` because the election is still stored in the frontend, which means the mock function would need to specify that input (as opposed to the jest mock, which doesn't check input). 
- We poll for auth status very quickly, so once the election hash changes (e.g. during configuration), the mock for `getAuthStatus` needs to also change its expected input very quickly. I tried this out but it was hard to make it deterministic.
A solution might be increasing the polling interval in test, but given that we are planning move the election to the backend, it didn't seem worthwhile for the moment.

jonahkagan:
cc @adghayes you might be interested to see what was going on with the apps/admin tests

apps/admin/frontend/test/render_in_app_context.tsx
```diff
@@ -88,16 +88,7 @@ export function renderRootElement(
...
-    queryClient = new QueryClient({
-      defaultOptions: {
-        ...QUERY_CLIENT_DEFAULT_OPTIONS,
-        queries: {
-          ...(QUERY_CLIENT_DEFAULT_OPTIONS.queries ?? {}),
-          useErrorBoundary: false,
-        },
-      },
-    }),
+    queryClient = createQueryClient(),
```
arsalansufi, line 91: Thanks for getting to the bottom of this! Figured that was probably the case - that a real bug in a test was being masked

arsalansufi (approved):
This looks great! I agree with the call not to update VxMark yet

PR #3120, opened 2023-03-10
Fix Broken Either Neither

adghayes:
## Overview

Currently, the text of "Yes" and "No" options in our ballot measures are not allowed to wrap in the ballot. With the recent change removing the special casing of either-neither in the hand-marked paper ballot (#2972), that means that our either-neither ballots are currently broken. This means they won't scan, their layouts won't be generated correctly, and we can't infer a lot of information about them.

I know we're going to change our ballot generation and layout soon but even if it's only in the interim I'd like to merge this and not have some of or main fixtures broken.

### DIFF
--- a/apps/admin/frontend/src/components/hand_marked_paper_ballot.tsx
+++ b/apps/admin/frontend/src/components/hand_marked_paper_ballot.tsx
@@ -1114,7 +1114,7 @@ export function HandMarkedPaperBallot({
                           }}
                         />
                       )}
-                      <Text bold noWrap>
+                      <Text bold>
                         <BubbleMark
                           position={election.ballotLayout?.targetMarkPosition}
                           checked={hasVote(votes?.[contest.id], 'yes')}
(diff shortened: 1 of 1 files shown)


adghayes:
After: 
Screen Shot 2023-03-10 at 11 21 05 AM

Before:
Screen Shot 2023-03-10 at 11 21 37 AM

kshen0 (approved):
I don't have a ton of background on this ballot type but removing `noWrap` seems like it would fix the issue of not wrapping! ~Do you have a screenshot to confirm?~

Thanks for uploading, visual changes LGTM. Typically for ballot layout changes (even fixes) are scanning changes needed to? IIUC in this case it may not be worth it if all layouts and scanning behavior is changing soon.

adghayes:
yup @kshen0, was just getting it up

adghayes:
In our timing-mark free world, generally not. Not sure if that will always be the case. In the case of our own ballots (not NH ballots), we generate layouts after from the images and then use those layouts as a comparison for the scanning. This will change generated layouts but scanning should proceed correctly from those layouts.

In fact, what motivated me to move on this was that because the ballot images looked broken, the generated layouts were also broken and I was unable to do some post-processing on layouts that is necessary in formatting the cast vote records.

PR #3122, opened 2023-03-13
CDF CVR Schema Restrictions & Testing

adghayes:
_Review commit-by-commit_

## Overview

Add a few restrictions on the CDF CVR. Primarily:
- Requiring some VVSG 2.0 necessary metadata on the CVR
- Requiring batch information in the report
- Removing party & retention contests

## Testing Plan

Also added testing for the CDF CVR schema in this PR similar to the ballot definition schema. A few adjustments had to be made:

- The NIST schema has unused definitions, so only testing that on the VX schema
- The NIST schema (and forked VX schema) are a different version of the JSON schema spec and thus require a variant library of `ajv`
- In the VX schema for cast vote records we've added a definition, so I needed to add a carveout to the subset validation that allowed custom definitions prefixed with `vx`.

### DIFF
--- a/libs/types/package.json
+++ b/libs/types/package.json
@@ -54,6 +54,7 @@
     "@typescript-eslint/parser": "^5.37.0",
     "@votingworks/cdf-schema-builder": "workspace:*",
     "ajv": "^8.12.0",
+    "ajv-draft-04": "^1.0.0",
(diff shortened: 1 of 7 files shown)


jonahkagan (approved):
Nice work!

PR #3125, opened 2023-03-13
VxAdmin: List Available CVR Files on Backend

### LINKED ISSUE #3076: VxAdmin: Read CVR File Entries on Backend
Follow-up to #3052. We're reading the file contents from the backend after that PR, but not listing the files available to be read, which is still being done in `kiosk-browser`. 

Must re-enable the integration-tests that were disabled as part of #3052.

adghayes:
## Overview

We list available CVR files on a USB drive for the user when they are importing CVR files. This PR moves the listing (reading of the directory on the USB drive) from the frontend to the backend. We'll have to update this listing and loading code when we switch to CVR CDF so it seemed appropriate to first move it to the backend.

Closes #3076, but does not achieve one of its main goals - re-enabling the integration testing disabled in #3052. See below for discussion.

## Testing Plan
- Testing in `libs/backend`, `admin/backend` and `admin/frontend`
- Manual Testing

## Checklist

- [ ] I have added

### DIFF
--- a/apps/admin/backend/src/app.test.ts
+++ b/apps/admin/backend/src/app.test.ts
@@ -378,13 +378,9 @@ test('addCastVoteRecordFile - handle parsing error', async () => {
 test('auth', async () => {
   const { electionDefinition } = electionFamousNames2021Fixtures;
   const { electionData, electionHash } = electionDefinition;
-  const { apiClient, auth, workspace } = buildTestEnvironment();
-  workspace.store.addElection(
-    electionMinimalExhaustiveSampleFixtures.electionDefinition.electionData
-  );
-  workspace.store.addElection(
-    electionFamousNames2021Fixtures.electionDefinition.electionData
-  );
+  const { apiClient, auth } = buildTestEnvironment();
+  await configureMachine(apiClient, auth, electionDefinition);
+  expect(auth.getAuthStatus).toHaveBeenCalledTimes(1);
 
   await apiClient.getAuthStatus();
   await apiClient.checkPin({ pin: '123456' });
@@ -393,8 +389,8 @@ test('auth', async () => {
   void (await apiClient.programCard({ userRole: 'election_manager' }));
   void (await apiClient.unprogramCard());
 
-  expect(auth.getAuthStatus).toHaveBeenCalledTimes(1);
-  expect(auth.getAuthStatus).toHaveBeenNthCalledWith(1, { electionHash });
+  expect(auth.getAuthStatus).toHaveBeenCalledTimes(2);
+  expect(auth.getAuthStatus).toHaveBeenNthCalledWith(2, { electionHash });
   expect(auth.checkPin).toHaveBeenCalledTimes(1);
   expect(auth.checkPin).toHaveBeenNthCalledWith(
     1,
--- a/apps/admin/backend/src/app.test.ts
+++ b/apps/admin/backend/src/app.test.ts
@@ -6,24 +6,23 @@ import {
   electionSampleCdfDefinition,
 } from '@votingworks/fixtures';
 import { LogEventId } from '@votingworks/logging';
-import { mockOf } from '@votingworks/test-utils';
+
 import { promises as fs } from 'fs';
 import { sha256 } from 'js-sha256';
 import { tmpdir } from 'os';
-import { join, sep } from 'path';
+import { basename, join, sep } from 'path';
+import tmp from 'tmp';
 import {
   buildTestEnvironment,
   configureMachine,
+  mockCastVoteRecordFileTree,
   mockElectionManagerAuth,
   mockSystemAdministratorAuth,
 } from '../test/app';
-import { listCastVoteRecordFilesOnUsb } from './cvr_files';
 
 beforeEach(() => {
   jest.restoreAllMocks();
 });
-
-jest.mock('./cvr_files');
 
 test('managing the current election', async () => {
   const { apiClient, auth, logger } = buildTestEnvironment();
@@ -136,7 +135,7 @@ test('configuring with a CDF election', async () => {
 });
 
 test('cast vote records - happy path election flow', async () => {
-  const { apiClient, auth } = buildTestEnvironment();
+  const { apiClient, auth, mockUsb, logger } = buildTestEnvironment();
 
   const { electionDefinition, standardCvrFile, standardLiveCvrFile } =
     electionMinimalExhaustiveSampleFixtures;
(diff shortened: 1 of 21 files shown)


adghayes:
@jonahkagan I realized I was wrong to think that this change would allow me to flip on the VxAdmin integration tests that I disabled. This change affects loading cast vote record files from a default location, but what broke in the integration tests is loading from manual input since we are now relying on Electron to saturate the `File` object with the `path`. So I still don't have a clear path to enabling integration testing in VxAdmin.

Right now I'm looking into running our integration tests in `Electron`. But we have a general inability to mock USB drive actions and I think it's severely limiting our ability to do integration tests since our software is so USB drive reliant.

apps/admin/backend/src/cvr_files.ts
```diff
@@ -0,0 +1,108 @@
...
+        break;
+      case 'no-usb-drive':
+      case 'usb-drive-not-mounted':
+        // we're just polling without a USB drive in these cases, no issue
+        break;
+      default:
+        throwIllegalValue(errorType);
+    }
+
+    return [];
```
jonahkagan, line 59: If we're returning `[]` in all of these cases, how does the app know to differentiate between the corrupted USB case (where it should prob show an error to the user) and the other cases where they can manually select a file?

apps/admin/backend/src/cvr_files.ts
```diff
@@ -0,0 +1,108 @@
...
+        // only interested in valid CVR files so ignore it.
+        continue;
+      }
+    }
+  }
+
+  await logger.log(LogEventId.CvrFilesReadFromUsb, 'system', {
+    message: `Found ${castVoteRecordFileMetadataList.length} CVR files on USB drive, user shown option to load.`,
+    disposition: 'success',
+  });
```
jonahkagan, line 92: I'm searching for a pattern where we don't need to interleave log statements in our business logic, because it makes it harder to read and reason about the logic.

Any ideas? Maybe we could return a Result and then log success/failure one level up?

apps/admin/backend/src/app.ts
```diff
@@ -30,33 +30,23 @@ import { AddCastVoteRecordFileResult, ConfigureResult } from './types';
...
 }
 
 function constructDippedSmartCardAuthMachineState(
   workspace: Workspace
 ): DippedSmartCardAuthMachineState {
-  // TODO: Once we actually support multiple elections, configure the auth instance with the
-  // currently selected election rather than the most recently created. In fact, do so as soon as
-  // the currently selected election is persisted on the backend instead of the frontend since,
-  // even today, in dev, we can end up with multiple election definitions under the hood via
-  // incognito windows
```
jonahkagan, line 54: 🎉

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -140,62 +132,15 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element | null {
...
-
-  useEffect(() => {
+  // TODO: Rather than explicitly refetching, which is outside of the standard
+  // React Query pattern, we should invalidate the query on USB drive status
+  // change. Currently React Query is not managing USB drive status and polling
+  // is being performed elsewhere.
+  useExternalStateChangeListener(usbDrive.status, () => {
     if (usbDrive.status === 'mounted') {
-      void fetchFilenames();
+      void cvrFilesOnUsbQuery.refetch();
```
jonahkagan, line 141: Any reason you chose `query.refetch()` over `queryClient.invalidateQueries()` here? I'm not sure which would be more appropriate, but I would have defaulted to the latter

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -140,62 +132,15 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element | null {
...
-      }
-    }
-  }
-
-  useEffect(() => {
+  // TODO: Rather than explicitly refetching, which is outside of the standard
+  // React Query pattern, we should invalidate the query on USB drive status
+  // change. Currently React Query is not managing USB drive status and polling
+  // is being performed elsewhere.
+  useExternalStateChangeListener(usbDrive.status, () => {
```
jonahkagan, line 139: Nice, this is exactly the kind of situation I was imagining we could use `useExternalStateChangeListener` for - a safe stopgap

adghayes:
OK, no easy win in switching to testing in `electron`. 

Another alternative is to use some kind of environment variable or constant prefix to determine the path from the filename in the case that there is not one from Electron. Not sure if that's too hacky...

But more generally, it seems to me that if we really want to leverage integration tests we will need to invest in better USB mocking.

apps/admin/backend/src/app.test.ts
```diff
@@ -146,6 +150,10 @@ test('cast vote records - happy path election flow', async () => {
     Admin.CvrFileMode.Unlocked
   );
 
+  // frontend can poll for available cast vote record files
+  mockOf(listCastVoteRecordFilesOnUsb).mockResolvedValueOnce([]);
+  expect(await apiClient.listCastVoteRecordFilesOnUsb()).toEqual([]);
```
jonahkagan, line 155: This approach to testing is what I would call a "tower of mocks." Each layer in the call stack gets unit tested with the layer immediately below mocked. I personally try to avoid these kinds of towers because I think they are brittle to changes. If any code in the stack gets refactored, many of the tests and mocks in the tower may have to change. In particular, it can be dangerous if I change a function's behavior and then forget to update a mock of that function in another test somewhere.

Instead, I try to stretch out my tests as much as possible so they are testing from the highest interface in the call stack that makes sense (e.g. in this case, the API) and mock the lowest interface in the call stack as possible (in this case, the USB drive itself). That gives me the most freedom to refactor the code in between with the tests as my safeguard.

I don't know that everybody agrees with me on this take, and it's certainly got some tradeoffs, but figured I'd share it and see what you thought.

jonahkagan (commented):
Let's talk through the integration tests issue. I'm not sure I have a full grasp on it yet

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
adghayes: Mostly because I feel confident that `query.refetch()` will trigger a refetch but I wasn't sure if invalidating a query while it was already mounted would trigger a refetch. But I'm realizing that we already do that in some places, so it would definitely work. I don't really have an opinion here, seems like they both work.

jonahkagan: Found some discussion here: https://github.com/TanStack/query/discussions/2468.

Doesn't seem like it matters!

apps/admin/backend/src/app.test.ts
adghayes: I don't prefer this style of testing either. When I worked on the VxAdmin `grout` transition, I tried to remove any dependence on `store` mocking and just test with the whole `store` running.

I mostly mocked in this way because it was easier, and the lower layers are a little harder to mock. Since `getUsbDrives` is internal to `libs/backend`, I can't mock it on its own. I have to mock the touchpoint with that module, `listDirectoryOnUsbDrive`. I could mock that here in `app.test.ts` as well, but for whatever reason I also deemed that to be more work.

To really get rid of the tower here we'd have to change `listDirectory` to accept `getUsbDrives` as an argument, similar to how the `Exporter` that @eventualbuddha created works, and mock that. That would be the more "testable" code pattern.

I agree with your calling this out but on its own it doesn't seem quite important enough to block and spend time on, but I don't feel strongly and can mock this differently if you do think it's important.

apps/admin/backend/src/cvr_files.ts
adghayes: No ideas, although your `Result` one is the best (maybe only) one I've heard. The current logging is pretty hard to read. I can imagine that more of our methods return `Result` objects and then we have a `logging.ts` that just has logging functions for each type of result. I am not wholly convinced that the current pattern is unreadable enough to make that worth it - the colocation also seems valuable.

apps/admin/backend/src/app.test.ts
jonahkagan: Maybe there's a way we can set up the mocking for `getUsbDrives` once and reuse it? I already made an attempt at this here: https://github.com/votingworks/vxsuite/blob/1f3cbf7c4d8c480f768c2ac1f06065442316425a/apps/scan/backend/test/helpers/app_helpers.ts#L73. In that case I used dependency injection like you mentioned.

Talking this through, I think a good path forward is to have as thin as possible of an abstraction layer over the actual filesystem commands that are talking to the USB drive and then mock that.

I don't feel strongly about whether you move towards that in this PR.

apps/admin/backend/src/cvr_files.ts
adghayes: Formerly, if one of these errors happen an error would just be thrown to the error boundary. 

Now, I'm just ignoring it. Despite the logging, the user is still allowed to manually select a file in this case. Although they will likely hit another problem with that USB drive, it's possible they do not. 

I mostly didn't care about this case because I did not find it worth building a user error message for. In order to get either the permission denied or the not directory error, the user would have had to manipulate the file system manually at another computer rather than export from one of our machines. We don't want them doing that anyway.

apps/admin/backend/src/app.test.ts
adghayes: I understand this `createMockUsb` and the dependency injection - that seems like a good path forward here. But I'm not sure what you mean by the following: 

> Talking this through, I think a good path forward is to have as thin as possible of an abstraction layer over the actual filesystem commands that are talking to the USB drive and then mock that.

What do you mean by that? Something different than `createMockUsb` is doing?

jonahkagan: I think by that last part I just was reiterating the same thing that `Usb` and `createMockUsb` is already doing. Sorry for being confusing lol

adghayes:
@jonahkagan CircleCI is out but code is ready for re-review. I decided to add the `getUsbDrives` mocking in VxAdmin because it seems like a better testing pattern, but at this time not breaking it out into a `libs/usb` or `libs/backend/test-utils`.

jonahkagan (approved):

PR #3131, opened 2023-03-14
(Re-)implement card mocking for local development and integration tests

arsalansufi:
_Easiest to review by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2915

This PR (re-)implements card mocking for local development and integration tests, since the old mocking approach is tied to memory cards and `services/smartcards`, which will both soon be removed.

The new approach takes advantage of the new generic card API (the abstraction continues to work well!), implementing the API using a file under the hood.

## Demo

_Card mocking script_

```
$ ./scripts/mock-card 
Options:
  --card-type            The type of card to mock
   [string] [choices: "system-administrator", "election-manager", "poll-worker",
                                                      "unprogrammed", "no-card"]
  --election-definition  The election definition to use for an election manager
                         or poll worker card                            [string]

Examples:
  $ ./scripts/mock-card --help
  $ ./scripts/mock-card --card-type system-administrator
  $ ./scripts/mock-card --card-type election-manager --election-definition \
  ../fixtures/data/electionFamousNames2021/election.json
  $ ./scripts/mock-card --card-type poll-worker --election-definition \
  ../fixtures/data/electionSample.json
  $ ./scripts/mock-card --card-type unprogrammed
  $ ./scripts/mock-card --card-type no-card
```

```
$ ./scripts/mock-card --card-type spades
❌ Invalid values:
  Argument: card-type, Given: "spades", Choices: "system-administrator", "election-manager", "poll-worker", "unprogrammed", "no-card"

Options:
... [ Rest of usage/help message ]
```

```
$ ./scripts/mock-card --election-definition ../fixtures/data/electionFamousNames2021/election.json
❌ Must specify card type

Options:
... [ Rest of usage/help message ]
```

```
$ ./scripts/mock-card --card-type election-manager
❌ Must specify election definition for election manager and poll worker cards

Options:
... [ Rest of usage/help message ]
```

```
$ ./scripts/mock-card --card-type election-manager --election-definition ../fixtures/data/non-existent-file.json
❌ ENOENT: no such file or directory, open '../fixtures/data/non-existent-file.json'
```

```
$ ./scripts/mock-card --card-type election-manager --election-definition ../fixtures/data/sample-ballot-images/blank-page.png 
❌ ../fixtures/data/sample-ballot-images/blank-page.png isn't a valid election definition
```

```
$ ./scripts/mock-card --card-type system-administrator
$ ./scripts/mock-card --card-type election-manager --election-definition ../fixtures/data/electionFamousNames2021/election.json
$ ./scripts/mock-card --card-type poll-worker --election-definition ../fixtures/data/electionSample.json
$ ./scripts/mock-card --card-type unprogrammed
$ ./scripts/mock-card --card-type no-card
```

## Testing

- [x] Added automated tests for `MockFileCard`
- [x] Manually tested mocking during local development
- [x] Verified that mocking is working in integration tests

## Checklist


libs/auth/scripts/mock_card.ts
```diff
@@ -0,0 +1,186 @@
...
+      '$ ./scripts/mock-card --card-type election-manager --election-definition electionFamousNames2021',
+      ''
+    )
+    .example(
+      '$ ./scripts/mock-card --card-type poll-worker --election-definition electionSample',
+      ''
+    )
+    .example('$ ./scripts/mock-card --card-type unprogrammed', '')
+    .example('$ ./scripts/mock-card --card-type no-card', '')
+    .strict();
```
arsalansufi, line 78: See the PR description for the usage/help message as well as error messages!

apps/admin/integration-testing/Makefile
```diff
@@ -27,10 +27,9 @@ build-smartcards:
 build: build-frontend build-backend build-converter-ms-sems build-smartcards
 
 run:
-	export NODE_ENV=production
```
arsalansufi, line 30: This was a fun find - every line of a Makefile is run in a separate shell context. So integration tests haven't actually been running with `NODE_ENV` set to `production`. I fix that here

apps/admin/integration-testing/cypress/support/auth.ts
```diff
@@ -1,14 +1,28 @@
 import { methodUrl } from '@votingworks/grout';
 
+// Importing all of @votingworks/auth causes Cypress tests to fail since Cypress doesn't seem to
+// interact well with PCSC Lite card reader code
+// eslint-disable-next-line vx/no-import-workspace-subfolders
```
arsalansufi, line 5: Even though I'm not using the PCSC Lite card reader code, just importing it in seemed to cause issues for Cypress

apps/admin/integration-testing/cypress/support/auth.ts
```diff
@@ -1,14 +1,28 @@
...
+import {
+  mockCard,
+  MockFileContents,
+} from '@votingworks/auth/src/mock_file_card';
+
 const PIN = '000000';
 
+function mockCardCypress(mockFileContents: MockFileContents): void {
+  mockCard(mockFileContents, cy.writeFile);
+}
```
arsalansufi, line 15: I was running into this error when trying to use the base `mockCard` here: https://stackoverflow.com/questions/63797683/how-to-remove-fs-readfilesync-is-not-a-function-error-from-cypress-test

Using `cy.writeFile` instead of `fs.writeFileSync` did the trick

libs/auth/scripts/generate_dev_keys_and_certs.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 import { Buffer } from 'buffer';
-import * as fs from 'fs/promises';
+import { promises as fs } from 'fs';
```
arsalansufi, line 3: Another change to avoid Cypress errors, this time `Can't resolve 'fs/promises'`, which actually seems to be a Webpack error (Cypress uses Webpack as its default preprocessor).

This import change may no longer be necessary since I'm importing `@votingworks/auth/src/mock_file_card` instead of all of `@votingworks/auth` (and `@votingworks/auth/src/mock_file_card` uses standard `fs` instead of `fs/promises`), but I kept it to avoid headaches down the road.

I'm sure we could (and probably should) update Cypress to use Vite as its preprocessor, but that was out of scope for this PR.

libs/auth/scripts/mock_card.ts
```diff
@@ -0,0 +1,186 @@
...
+        description: 'The type of card to mock',
+        type: 'string',
+        choices: CARD_TYPES,
+      },
+      'election-definition': {
+        description:
+          'The election definition to use for an election manager or poll worker card',
+        type: 'string',
+        choices: Object.keys(ELECTION_DEFINITIONS).sort(),
+      },
```
jonahkagan, line 62: I wonder if it would be easier to just allow passing in a path to an election definition like the previous script. Advantages I can think of:
- flexibly supports creating new/modified election definitions (which can happen when testing new features)
- leverages bash autocomplete for typing in a fixture name

I imagine you may have had a reason for doing it this way though, so I'm curious what that might be

apps/admin/integration-testing/Makefile
jonahkagan: Lol after all that discussion! Nice find

apps/admin/integration-testing/cypress/support/auth.ts
jonahkagan: Ah, I was wondering why you had a parameter for the `writeFileFn`, that explains it!

libs/auth/src/openssl.test.ts
```diff
@@ -12,8 +11,14 @@ import {
...
-jest.mock('fs');
-jest.mock('fs/promises');
+jest.mock('fs', () => ({
+  existsSync: jest.fn(),
+  promises: {
+    mkdir: jest.fn(),
+    unlink: jest.fn(),
+    writeFile: jest.fn(),
+  },
+}));
```
jonahkagan, line 21: Why this change? Just curious

jonahkagan (approved):

apps/admin/integration-testing/cypress/support/auth.ts
arsalansufi: Ah yeah, I should've left a note on the earlier commit! But glad it's all clear now

libs/auth/src/openssl.test.ts
arsalansufi: This is just a side effect of having to change how we import `fs/promises`. `jest.mock(...)` only mocks one level deep in a library.

So while this mocks all the `fs` promise methods:

```ts
import * as fs from 'fs/promises';

jest.mock('fs/promises');
```

This does not:

```ts
import { promises as fs } from 'fs';

jest.mock('fs');
```

libs/auth/scripts/mock_card.ts
arsalansufi: Yeah I went back and forth on this. I decided to map from strings to fixtures to avoid having to manually read files and compute election hashes, since it felt easy to get pieces of that wrong.

But thinking about it some more and prototyping some code, it's not that bad. For the benefits you mention, I'll switch over!

PR #3135, opened 2023-03-14
Miscellaneous auth cleanup

arsalansufi:
_Easiest reviewed by commit_

## Overview

A miscellaneous auth cleanup PR. I was making these changes as I was working on auth tests and decided to extract them into their own PR.

The changes include:

- More comprehensively and holistically handling PIN check errors
- Cleaning up auth logging (specifically card programming logging) and including more details where possible

## Demo Screenshot

As part of the updates to PIN check error handling, I introduced this new screen:

error

Should be seen very rarely if ever (e.g. if a card malfunctions during PIN checking) but is now there for completeness!

## Testing

- [x] Tested PIN checking and logging manually

Automated tests are on their way in my next PR 🤖

## Checklist

### DIFF
--- a/libs/auth/src/dipped_smart_card_auth.ts
+++ b/libs/auth/src/dipped_smart_card_auth.ts
@@ -384,97 +384,93 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
 
     switch (action.type) {
       case 'check_card_reader': {
-        const newAuthStatus = ((): DippedSmartCardAuthTypes.AuthStatus => {
-          switch (currentAuthStatus.status) {
-            case 'logged_out': {
-              switch (action.cardStatus.status) {
-                // TODO: Consider an alternative screen on the frontend for unknown errors
-                case 'no_card':
-                case 'unknown_error': {
-                  return { status: 'logged_out', reason: 'machine_locked' };
-                }
-
-                case 'card_error': {
-                  return { status: 'logged_out', reason: 'card_error' };
-                }
-
-                case 'ready': {
-                  const { user } = action.cardStatus;
-                  const validationResult = this.validateCardUser(
-                    machineState,
-                    user
-                  );
-                  if (validationResult.isOk()) {
-                    assert(
-                      user &&
-                        (user.role === 'system_administrator' ||
-                          user.role === 'election_manager')
-                    );
-                    return isFeatureFlagEnabled(
-                      BooleanEnvironmentVariableName.SKIP_PIN_ENTRY
-                    )
-                      ? { status: 'remove_card', user }
-                      : { status: 'checking_pin', user };
-                  }
-                  return {
-                    status: 'logged_out',
-                    reason: validationResult.err(),
-                    cardUserRole: user?.role,
-                  };
-                }
-
-                /* istanbul ignore next: Compile-time check for completeness */
-                default: {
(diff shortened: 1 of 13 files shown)


libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -180,7 +183,19 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
     input: { pin: string }
   ): Promise<void> {
     await this.checkCardReaderAndUpdateAuthStatus(machineState);
-    const checkPinResponse = await this.card.checkPin(input.pin);
+    let checkPinResponse: CheckPinResponseExtended;
+    try {
+      checkPinResponse = await this.card.checkPin(input.pin);
```
arsalansufi, line 187: Rather than having card API implementations remember to convert errors to `CheckPinResponseError` objects, I decided it was safer to just have the auth classes take care of this

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -384,97 +384,93 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
 
     switch (action.type) {
       case 'check_card_reader': {
-        const newAuthStatus = ((): DippedSmartCardAuthTypes.AuthStatus => {
```
arsalansufi, line 360: This commit is easiest to review using [?w=1](https://github.com/votingworks/vxsuite/pull/3135/commits/676ce525fb7f50a9a1b7b67e84cfe375b26262df?w=1)

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -384,97 +384,93 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
...
+              };
             }
+            return currentAuthStatus;
           }
-        })();
 
-        return newAuthStatus;
+          /* istanbul ignore next: Compile-time check for completeness */
+          default: {
+            return throwIllegalValue(currentAuthStatus, 'status');
```
eventualbuddha, line 471: Slight preference for leaving this without the `return` and putting a `break` in the outer `switch` `case`.

libs/auth/src/inserted_smart_card_auth.ts
```diff
@@ -328,63 +328,56 @@ export class InsertedSmartCardAuth implements InsertedSmartCardAuthApi {
...
+              reason: validationResult.err(),
+              cardUserRole: user?.role,
+            };
           }
-        })();
 
-        return newAuthStatus;
+          /* istanbul ignore next: Compile-time check for completeness */
+          default: {
+            return throwIllegalValue(action.cardStatus, 'status');
```
eventualbuddha, line 378: Ditto.

eventualbuddha (approved):

jonahkagan (approved):

libs/auth/src/dipped_smart_card_auth.ts
arsalansufi: I also prefer a `break`, but using a `break` here results in TS detecting unreachable code:

unreachable-code

And for full visibility, just removing the `return` isn't an option either since that interferes with eslint's no-fallthrough rule:

eslint

So a bit of a catch-22 haha

PR #3139, opened 2023-03-15
CVR Directory Fixture Friendliness

adghayes:
_review commit-by-commit_

## Overview

Adds some infrastructure to enable importing the new CDF CVR directory structure into VxAdmin. Order of work / commits: 
1. Update `res-to-ts` to handle directories specially
2. Add initial CVR fixture, update `pnpm:build-resources` to special case directories under a `cdf-cvr-files` directory
3. Introduce some new dependencies and build tool to stream in large cast vote record reports. It is meant to be the inverse of `jsonStream`. Also introduce a validation tool for the directory structure.
4. Use new import tool to re-enable test in `cvr-fixture-generator`. 
5. Regenerate fixture


libs/backend/src/cast_vote_record_report_import.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  CVR_BALLOT_IMAGES_SUBDIRECTORY,
+  CVR_BALLOT_LAYOUTS_SUBDIRECTORY,
+} from './scan';
+
+const cdfCvrFile = electionMinimalExhaustiveSampleFixtures.standardCdfCvrFile;
+
+describe('getCastVoteRecordReportImport', () => {
+  test('imports a valid cast vote record report', async () => {
+    const castVoteRecordImportResult = await getCastVoteRecordReportImport(
+      join(cdfCvrFile.asDirectoryPath(), CAST_VOTE_RECORD_REPORT_FILENAME)
```
eventualbuddha, line 21: I suspect all these temporary files will pile up fairly quickly. Maybe wrap this in something that cleans up after itself?

```ts
const cleanupPaths: string[] = [];

afterAll(() => {
  for (const path of cleanupPaths) {
    rmSync(path, { recursive: true });
  }
});

function makeCdfCvrFixture(): string {
  const path = cdfCvrFile.asDirectoryPath();
  cleanupPaths.push(path);
  return path;
}

File: libs/backend/src/cast_vote_record_report_import.ts [hunk 2]
```diff
@@ -0,0 +1,221 @@
...
+  );
+  if (listBallotImageContentsResult.isErr()) {
+    return invalidDirectoryError(listBallotImageContentsResult.err());
+  }
+  const ballotImageFiles = listBallotImageContentsResult
+    .ok()
+    .filter(
+      (fileEntry) =>
+        fileEntry.type !== FileSystemEntryType.Directory &&
+        fileEntry.name.endsWith('.jpg')
```
eventualbuddha, line 176: I just know someone in the future is going to spend an inordinate amount of time figuring out why it's not working and they'll eventually realize it's because the images are `.jpeg` instead of `.jpg`.

libs/backend/src/list_directory.ts
```diff
@@ -110,6 +110,35 @@ export async function listDirectory(
   }
 }
 
+/**
+ * Get entries for a directory recursively, includes stat information for each entry.
+ * Requires that the path be absolute. Includes directories in result.
+ */
+export async function listDirectoryRecursive(
```
eventualbuddha, line 117: This could end up loading a lot of data in memory at once. Maybe consider something like this, and for now you can just `Array.from(listDirectoryRecursive(…))` and bail if you find an `Err`?

```ts
/**
 * Get entries for a directory recursively, includes stat information for each entry.
 * Requires that the path be absolute. Includes directories in result.
 */
export async function* listDirectoryRecursive(
  path: string
): AsyncGenerator<Result<FileSystemEntry[], ListDirectoryError>> {
  const listRootResult = await listDirectory(path);
  if (listRootResult.isErr()) {
    yield listRootResult;
  } else {
    for (const fileEntry of listRootResult.ok()) {
      if (fileEntry.type === FileSystemEntryType.Directory) {
        yield* listDirectoryRecursive(fileEntry.path);
      }
    }
  }
}
```

libs/backend/src/list_directory.ts
```diff
@@ -110,6 +110,35 @@ export async function listDirectory(
...
+  path: string
+): Promise<Result<FileSystemEntry[], ListDirectoryError>> {
+  const listRootResult = await listDirectory(path);
+  if (listRootResult.isErr()) {
+    return err(listRootResult.err());
+  }
```
eventualbuddha, line 123: As in my generator-based version, you can just do this:

```suggestion
  if (listRootResult.isErr()) {
    return listRootResult;
  }
```

Wishing for this:

```rust
let file_entries = list_directory(path)?;
```

✨

eventualbuddha (approved):

libs/backend/src/cast_vote_record_report_import.test.ts
adghayes: @eventualbuddha we could wrap it in this test file, but the problem will happen in other test files too. Perhaps if we use `tmp` in `res-to-ts` and `fixtures` then we can force cleanup? My understanding is that `tmp` has some ability to clean up after itself when a process ends that using the `fs.mktdirsync` does not.

adghayes: https://github.com/raszi/node-tmp#graceful-cleanup

libs/backend/src/list_directory.ts
adghayes: In the current use case I'm not concerned about memory over-usage but this utility could be used in that way. Just counted the total number of files on my dev machine and it was over a million :). I'll switch to the generator version.

jonahkagan (approved):

libs/fixtures/package.json
```diff
@@ -14,7 +14,7 @@
   "scripts": {
     "type-check": "tsc --build",
     "build": "tsc --build tsconfig.build.json",
-    "build:resources": "res-to-ts --rootDir data --outDir src/data 'data/**/*.{csv,jsonl,json,txt,jpeg,jpg,png,pdf,xml,zip}'",
+    "build:resources": "res-to-ts --rootDir data --outDir src/data 'data/**/*.{csv,jsonl,json,txt,jpeg,jpg,png,pdf,xml,zip}' '!data/**/cdf-cvr-files/**/*.{csv,jsonl,json,txt,jpeg,jpg,png,pdf,xml,zip}' 'data/**/cdf-cvr-files/*'",
```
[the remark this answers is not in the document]
adghayes: Making that change would remove the third pattern in this sequence, but the second would still be necessary and I don't see a way around that unless we want to establish some sort of naming convention.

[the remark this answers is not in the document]
jonahkagan: Ok, that's fine. Maybe we could have a res-to-ts config file or a wrapper script for this command ? It just gets a little hard to work with in a package.json command. Not a blocker though

libs/backend/src/list_directory.ts
adghayes: OK, I just pushed a change to use the generator version. It seems clunky but perhaps that is the cost of memory efficiency. Take a look?

libs/backend/src/cast_vote_record_report_import.test.ts
adghayes: @eventualbuddha can you provide some feedback on my most recent commit as well, where I introduce some easy cleanup functionality? By putting copied fixtures into a subdirectory of `tmp`, we have an easy way to clean them all up at once. It's a bit clunky because it requires adding a `jest` teardown script and having a constant semi-shared between `res-to-ts` and `fixtures` (which could be one library anyway), but I like it because it offers a very simple way to cleanup fixtures across a test suite.

libs/backend/src/list_directory.ts
adghayes: I intend to use the the utils from #3148 once that is merged.

libs/backend/src/cast_vote_record_report_import.ts
```diff
@@ -162,59 +162,67 @@ export async function validateCastVoteRecordReportDirectoryStructure(
...
-    join(directoryAbsolutePath, CVR_BALLOT_LAYOUTS_SUBDIRECTORY)
+  const listBallotLayoutContentsResults = await arrayFromAsyncIterator(
+    listDirectoryRecursive(
+      join(directoryAbsolutePath, CVR_BALLOT_LAYOUTS_SUBDIRECTORY)
+    )
   );
-  if (listBallotLayoutContentsResult.isErr()) {
-    return invalidDirectoryError(listBallotLayoutContentsResult.err());
+  const relativeLayoutPaths: string[] = [];
+  for (const result of listBallotLayoutContentsResults) {
```
eventualbuddha:
```suggestion
  const listBallotImageContentsResults = listDirectoryRecursive(
    join(directoryAbsolutePath, CVR_BALLOT_IMAGES_SUBDIRECTORY)
  );
  const relativeImagePaths: string[] = [];
  for await (const result of listBallotImageContentsResults) {
```

eventualbuddha (approved):
Approach seems sane 👍

libs/fixtures/package.json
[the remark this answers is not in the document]
adghayes: made a wrapper script in `bash`. not sure if that's the ideal approach but it works and is more readable

libs/fixtures/src/data/electionMinimalExhaustiveSample/index.ts
```diff
@@ -7,6 +7,7 @@ export * as standardCvrFile from './cvrFiles/standard.jsonl';
 export * as standardLiveCvrFile from './cvrFiles/standard.live.jsonl';
 export * as partial1CvrFile from './cvrFiles/partial1.jsonl';
 export * as partial2CvrFile from './cvrFiles/partial2.jsonl';
+export * as standardCdfCvrFile from './cdf-cvr-files/standard';
```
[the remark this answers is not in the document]
adghayes: used `standardCdfCvrReport`

Issue #2527, opened 2022-09-02
Implement account lockout after failed PIN entry


mattroe:
## Product Requirements
- A system administrator must be able to configure the number of PIN entry attempts (5-10) before preventing additional login attempts.
- A system administrator must be able to configure the duration of lockout (duration option TBD) after the configured login attempt number is met.
- After a user attempts the configured number of PIN entry attempts, they must wait the configured lockout duration before attempting to authenticate again.

## VVSG Requirements
- 11.5-C: The voting system must lockout roles or individuals after an administrator-specified number of consecutive failed authentications attempts. This requirement prevents certain classes of password guessing attacks. This requirement can be implemented using a technique such as exponential backoff. NIST SP800-63B recommends allowing 5-10 attempts before starting exponential backoff. Exponential backoff requires that after each unsuccessful authentication attempt, the time period before another authentication attempt can be made grows exponentially. For instance: The wait after 1 unsuccessful authentication attempt is 0 seconds; The wait after 2 unsuccessful attempts is 2 seconds; The wait after 3 unsuccessful attempts is 4 seconds, and so on.
- 11.5-D: The voting system must allow only an administrator to define the lockout duration. This requirement extends [VVSG2005] I.7.2.1.2 by allowing the administrator flexibility in configuring the account lockout policy. The lockout policy should not lockout voters.

PR #3145, opened 2023-03-16
Add libs/auth tests, part 3

arsalansufi:
_Easiest reviewed by commit_

## Overview

More `libs/auth` tests, this time for the top-level auth API implementations

## Checklist

### DIFF
--- a/libs/auth/src/inserted_smart_card_auth.test.ts
+++ b/libs/auth/src/inserted_smart_card_auth.test.ts
@@ -1,20 +1,928 @@
-import { fakeLogger } from '@votingworks/logging';
+import { Buffer } from 'buffer';
+import { z } from 'zod';
+import { err, ok } from '@votingworks/basics';
+import {
+  electionSample2Definition,
+  electionSampleDefinition,
+} from '@votingworks/fixtures';
+import {
+  fakeLogger,
+  LogDispositionStandardTypes,
+  LogEventId,
+  Logger,
+} from '@votingworks/logging';
+import {
+  fakeCardlessVoterUser,
+  fakeElectionManagerUser,
+  fakePollWorkerUser,
+  fakeSystemAdministratorUser,
+  mockOf,
+} from '@votingworks/test-utils';
+import {
+  ElectionSchema,
+  InsertedSmartCardAuth as InsertedSmartCardAuthTypes,
+  UserWithCard,
+} from '@votingworks/types';
+import {
+  BooleanEnvironmentVariableName,
+  generatePin,
+  isFeatureFlagEnabled,
+} from '@votingworks/utils';
 
-import { buildMockCard } from '../test/utils';
+import { buildMockCard, MockCard, mockCardAssertComplete } from '../test/utils';
+import { CardStatus } from './card';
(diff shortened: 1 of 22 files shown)


libs/types/src/auth/auth.ts
```diff
@@ -28,6 +28,8 @@ export type User =
   | PollWorkerUser
   | CardlessVoterUser;
 
+export type UserWithCard = Exclude<User, CardlessVoterUser>;
```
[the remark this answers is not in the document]
jonahkagan: Great idea

libs/auth/test/utils.ts
```diff
@@ -10,17 +12,47 @@ export function numericArray(input: {
...
+export function mockCardAssertComplete(mockCard: MockCard): void {
+  for (const method of [
+    'getCardStatus',
+    'checkPin',
+    'program',
+    'readData',
+    'writeData',
+    'clearData',
+    'unprogram',
+  ] as const) {
```
jonahkagan, line 55: Any way to make sure this doesn't get out of date if we add a method?

libs/auth/src/inserted_smart_card_auth.test.ts
```diff
@@ -1,20 +1,928 @@
...
+
+  mockCardStatus({ status: 'ready', user });
+  expect(await auth.getAuthStatus(machineState)).toEqual({
+    status: 'checking_pin',
+    user,
+  });
+
+  mockCard.checkPin
+    .expectCallWith(wrongPin)
+    .resolves({ response: 'incorrect', numRemainingAttempts: Infinity });
```
jonahkagan, line 225: Do we need any testing for what happens when there are no remaining pin attempts and the user tries to enter their pin? I can't remember what happens in that case and what code is responsible for it

jonahkagan (approved):
Exhaustive! Exhausting! 😅

libs/auth/src/inserted_smart_card_auth.test.ts
arsalansufi: Yeah good q! I haven't actually implemented that functionality yet. Will be doing so as part of https://github.com/votingworks/vxsuite/issues/2527

libs/auth/test/utils.ts
arsalansufi: Good call! Came up with this: https://github.com/votingworks/vxsuite/pull/3145/commits/20cc3e4648e4404eb05344c3f31344986e9d6e3e

arsalansufi: ^ (TS will complain if we're not exhaustive)

example

jonahkagan: Nice, that's clever.

I looked to see how we do it in the Grout client and [it's a bit simpler](https://github.com/votingworks/vxsuite/blob/7ca8e5abd1953b51f8daf71e04be4e7e8b8385ed/libs/grout/test-utils/src/mock_client.ts#L70-L74). Wondering if the same approach would work here

arsalansufi: Ah nice, yeah much simpler: https://github.com/votingworks/vxsuite/pull/3145/commits/94ff53443e3197b6fb30700c0beff85b82e1adc0

Issue #3080, opened 2023-03-07
[admin] Sysadmin "Definition" page loads system settings as part of zip


kshen0:

kshen0:
Updated title to load zip file per Slack discussion

PR #3147, opened 2023-03-16
Kevin/vxadmin system settings demo election

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3079
https://github.com/votingworks/vxsuite/issues/3080
https://github.com/votingworks/vxsuite/issues/3081

* Refactors some file reading utils into common utils
* Adds db table, types, and schema for system settings
* Introduces `Initial Setup Package` concept, a zip file with `election.json` and `systemSettings.json` used to configure VxAdmin only
* Adds API and db logic to save and load system settings to/from db
* Adds system settings export to ballot package for configuration of other machines

## Demo Video or Screenshot
https://user-images.githubusercontent.com/1060688/225710696-9c958938-4570-40a3-94f8-541b9af8ddda.mov

## Testing Plan
* Added tests for `setup package` util
* Added UI tests

## Checklist


apps/admin/backend/schema.sql
```diff
@@ -80,11 +80,19 @@ create table printed_ballots (
     on delete cascade
 );
 
+create table system_settings (
```
kshen0, line 83: Note we already have a settings table. I considered adding system settings to settings but didn't to avoid locking us in to making system settings a singleton. There's a case to be made that `system settings` should be renamed something else to avoid confusion, which I would gladly do if there's a clearly better option.

apps/admin/backend/src/app.ts
```diff
@@ -120,6 +126,54 @@ function buildApi({
...
+        const typedError = error as Error;
+        await logger.log(
+          LogEventId.SystemSettingsSaved,
+          assertDefined(await getUserRole()),
+          { disposition: 'failure', error: typedError.message }
+        );
+        return err({
+          type: 'database',
+          message: `Failed to save system settings to db. Error: ${typedError.message}`,
+        });
```
arsalansufi, line 164: We typically use [`Result`](https://github.com/votingworks/vxsuite/blob/25857c511eb778be3647db3c4a9a843bc1efc051/libs/basics/src/result.ts) error objects (`err(...)`) for expected/user error cases. So for an invalid file, this makes sense. But for a DB error, less so.

For that, I'd probably recommend letting the thrown error bubble up (as other handlers in this file do). (For context, these errors trigger the frontend error boundary.)

There are definitely some fuzzy areas when it comes to classifying errors. For example, for a lot of smart card errors, I've been using `err(...)`s because those are often triggered by user actions like preemptively removing a card from the card reader.

apps/admin/backend/src/app.ts
```diff
@@ -120,6 +126,54 @@ function buildApi({
...
+      }
+
+      await logger.log(
+        LogEventId.SystemSettingsSaved,
+        assertDefined(await getUserRole()),
+        { disposition: 'success', system_settings_id: id }
```
arsalansufi, line 170: I believe most of our additional/custom log fields use camelCase, e.g. https://github.com/votingworks/vxsuite/blob/87da7bf630353891a84a81c39ef5ed8a9af7c538/libs/auth/src/dipped_smart_card_auth.ts#L241

apps/admin/backend/src/app.ts
```diff
@@ -164,6 +218,19 @@ function buildApi({
       return null;
     },
 
+    getCurrentSystemSettings(): SystemSettings | null {
+      const currentSystemSettingsId = store.getCurrentSystemSettingsId();
```
arsalansufi, line 222: Do we plan to support multiple system settings objects in VxAdmin (like we plan to one day support multiple election objects)?

My hunch is that, unlike elections, we'll only ever store one system settings object at a time, since these settings are global / applicable to all elections. With that in mind, we could simplify and get rid of this "current system settings ID" persistence, knowing that there'll only ever be one system settings object. The only reason the analogous "current election ID" persistence exists is to one day support multiple elections.

Worth sanity checking that product assumption with Roe!

apps/admin/frontend/src/api.ts
```diff
@@ -233,6 +233,18 @@ export const getPrintedBallots = {
   },
 } as const;
 
+export const getCurrentSystemSettings = {
```
arsalansufi, line 236: Nit that might be related to my above comment re not needing to persist multiple system settings objects: Could shorten this to `getSystemSettings`

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.test.tsx
```diff
@@ -94,8 +112,15 @@ test('Button renders properly when not clicked', () => {
 
 test('Modal renders insert usb screen appropriately', async () => {
   const usbStatuses: UsbDriveStatus[] = ['absent', 'ejected', 'bad_format'];
+  // Need 1 ApiMock per test, so we recreate the global-scoped mock to clobber
+  // the expectations we created in `beforeAll()`.
+  // TODO add an explicit 'clear expectations' function
```
arsalansufi, line 117: We actually used to provide a reset function, but it didn't always suit our needs so we intentionally shifted to a "recreate the mock for each test" approach: https://github.com/votingworks/vxsuite/pull/2986#discussion_r1098921442

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.test.tsx
```diff
@@ -94,8 +112,15 @@ test('Button renders properly when not clicked', () => {
...
+  apiMock = createApiMock();
```
arsalansufi, line 118: Curious why we need this when we I _think_ we could just do

```
apiMock = createApiMock();
```

(without the `const`) inside the `usbStatus` loop below

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.tsx
```diff
@@ -254,8 +268,13 @@ export function ExportElectionBallotPackageModalButton(): JSX.Element {
         case 'mounted': {
           actions = (
             <React.Fragment>
-              <Button variant="primary" onPress={() => saveFileCallback(false)}>
-                Save
+              <Button
+                disabled={!loaded}
+                variant="primary"
+                // onPress={() => console.log('on click')}
```
arsalansufi, line 274: Classic lingering log line 🪵

libs/types/src/system_settings.ts
```diff
@@ -0,0 +1,24 @@
+import { z } from 'zod';
+import { Id, Iso8601Timestamp } from './generic';
+
+/**
+ * System settings as used by frontends. Several database fields are hidden from the app
+ * because they are omitted in this model; see schema.sql.
+ */
+export interface SystemSettings {
+  pollWorkerCardPinsEnabled: boolean;
```
arsalansufi, line 9: Totally feel free to ignore this, especially since I put this variable name in the original ticket 😅, but I know we often prefix boolean variable names with `is` or `are` for consistency/readability, so this could be `arePollWorkerCardPinsEnabled`

arsalansufi (commented):
Still reviewing some odds and ends but wanted to get this batch to you sooner rather than later!

apps/admin/backend/src/app.test.ts
```diff
@@ -496,3 +497,170 @@ test('auth before election definition has been configured', async () => {
...
+  } as const;
+
+  const result = await apiClient.setSystemSettings({
+    systemSettings: JSON.stringify(malformedInput),
+  });
+  expect(result.isOk()).toEqual(false);
+  expect(result.isErr()).toEqual(true);
+
+  const err = result.err();
+  assert(err);
```
eventualbuddha, line 619: This could be like this:

```suggestion
  assert(result.isErr());
  const err = result.err();
```

The `assert` will help TS narrow the type of `result` from `Result` to `Err`.

libs/utils/src/file_reading.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Buffer } from 'buffer';
+
+export function readFileAsyncAsString(file: File): Promise<string> {
+  return new Promise<string>((resolve, reject) => {
+    const reader = new FileReader();
+    reader.onload = () => {
+      const { result } = reader;
+      resolve(typeof result === 'string' ? result : '');
+    };
+    reader.onerror = reject;
```
arsalansufi, line 11: I know this is just an existing function that was moved but should this `onerror` look like the one for `readFile` below?

```ts
reader.onerror = () => {
  reject(reader.error);
};
```

libs/utils/src/file_reading.ts
```diff
@@ -0,0 +1,73 @@
...
+  return new Promise((resolve, reject) => {
+    const reader = new FileReader();
+
+    /* istanbul ignore next */
+    reader.onerror = () => {
+      reject(reader.error);
+    };
+
+    reader.onload = () => {
+      resolve(Buffer.from(reader.result as ArrayBuffer));
```
arsalansufi, line 26: I know the TS types say that `reader.result` can be any of `string | ArrayBuffer | null`. I imagine `string` isn't actually possible here since we're calling `reader.readAsArrayBuffer`, but `null` might be?

`Buffer.from(null)` will throw an error so might want to explicitly do `Buffer.from([])` (an empty Buffer) when the `reader.result` is `null`

apps/admin/frontend/package.json
```diff
@@ -19,8 +19,8 @@
     "stylelint:run": "stylelint 'src/**/*.{js,jsx,ts,tsx}' && stylelint 'src/**/*.css' --config .stylelintrc-css.js",
     "stylelint:run:fix": "stylelint 'src/**/*.{js,jsx,ts,tsx}' --fix && stylelint 'src/**/*.css' --config .stylelintrc-css.js --fix",
     "test": "is-ci test:ci test:watch",
-    "test:ci": "TZ=UTC node scripts/test.js --coverage --watchAll=false --reporters=default --reporters=jest-junit --maxWorkers=7",
-    "test:coverage": "TZ=UTC node scripts/test.js --coverage --watchAll=false",
+    "test:ci": "TZ=UTC node scripts/test.js --coverage --watchAll=false --reporters=default --reporters=jest-junit --maxWorkers=7 --env=jest-environment-jsdom-sixteen",
+    "test:coverage": "TZ=UTC node scripts/test.js --coverage --watchAll=false --env=jest-environment-jsdom-sixteen",
```
arsalansufi, line 23: Thanks for digging into this and making it consistent!

arsalansufi (commented):
Just a few last thoughts - let me know if you want to discuss any of the bigger pieces in the first pass of comments above!

apps/admin/backend/src/app.ts
kshen0: Ah sorry, wires got crossed in my brain since I think I was working on the snake-case schema before this.

apps/admin/backend/src/app.ts
kshen0: I think in general I tend towards preserving optionality (eg. maybe we want to support a test mode settings profile and live mode one, where the test mode one has no poll worker card pins because a single official is simulating the role of multiple poll workers). But I'll check with Roe, and if we don't have a strong use case we should definitely simplify

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.test.tsx
kshen0: Probably should update this file to just move the mock creation out of `beforeAll` and accept some repeated boilerplate

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.test.tsx
arsalansufi: Got it!

One of the pros to the current pattern (of the global mock variable but a fresh mock for each test):

```ts
let apiMock: ApiMock;

beforeEach(() => {
  apiMock = createApiMock();
});

afterEach(() => {
  apiMock.assertComplete();
});
```

is that we guarantee that `apiMock.assertComplete()` is called - something that's easy to forget when you're doing, say:

```ts
test(() => {
  const apiMock = createApiMock();
  ...
  // Have to remember to call this
  apiMock.assertComplete();
});
```

-----

Taking a closer look at this current case, a better solution that eliminates the need for any overrides is to just run each of these USB status tests as a separate test using a `test.each` (here's an [example](https://github.com/votingworks/vxsuite/blob/87da7bf630353891a84a81c39ef5ed8a9af7c538/libs/auth/src/dipped_smart_card_auth.test.ts#L231-L260))

apps/admin/backend/src/app.test.ts
```diff
@@ -496,3 +497,170 @@ test('auth before election definition has been configured', async () => {
...
+  await configureMachine(apiClient, auth, electionDefinition);
+
+  mockSystemAdministratorAuth(auth);
+
+  // configure with well-formed system settings
+  const setResult = await apiClient.setSystemSettings({
+    systemSettings: systemSettings.asText(),
+  });
+  assert(setResult.isOk());
+  setResult.ok();
```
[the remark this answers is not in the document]
kshen0: The intent was to assert both `isOk()` and `ok()` work on the `Result` but I realize this is a bad half-state because we don't assert on the result of `ok()`. We only implicitly test that it doesn't throw. I'll remove this since it's mostly (or entirely?) redundant with `isOk()`.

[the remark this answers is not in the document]
kshen0: Also somewhat messy test practice to test `Result` like this at this higher level

apps/admin/backend/src/store.test.ts
```diff
@@ -1247,3 +1254,45 @@ test('current election id', () => {
...
   expect(store.getCurrentElectionId()).toBeUndefined();
 });
+
+/**
+ * System settings tests
+ */
+function makeSystemSettings() {
+  const systemSettingsString =
+    electionMinimalExhaustiveSampleFixtures.systemSettings.asText();
+  return unsafeParse(SystemSettingsSchema, JSON.parse(systemSettingsString));
```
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/3147/commits/0257bdde25d5b7929cc3b3746d1735c9674b60c9#diff-cbcfdf781631a75a8d41b38426a10e689a12ddc43ea6895110d0344eb9b27315R1264

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.test.tsx
kshen0: see discussion below; https://github.com/votingworks/vxsuite/pull/3147/commits/8cbf449e251c2e618419cd0b359f3bcc10b12c2f

apps/admin/backend/src/app.ts
kshen0: Removed support for multiple system settings and simplified here: https://github.com/votingworks/vxsuite/pull/3147/commits/fa25e9d6c2ad4f9e002ad5d9409a7fd5fdd5a09f

apps/admin/backend/src/app.test.ts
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/3147/commits/e0d7d876fde6dcf19d0c5a7d7c0a77e202eda086#diff-4a69b2fa57321af9c7b222c3cff0b954bf33b35e6293291bdee466c7bb262842L642

apps/admin/backend/src/app.test.ts
kshen0: https://github.com/votingworks/vxsuite/pull/3147/commits/6a7c3e8b412178f24a94699d9c141ea1ce4e7ac0

libs/types/src/system_settings.ts
kshen0: https://github.com/votingworks/vxsuite/pull/3147/commits/1e30a9e3822ec9c80008fbe7a57366d0ed5da3a6

kshen0:
Updates are made, feel free to take another look when you have time @arsalansufi

apps/admin/backend/src/app.test.ts
```diff
@@ -496,3 +496,124 @@ test('auth before election definition has been configured', async () => {
...
+  mockSystemAdministratorAuth(auth);
+
+  // configure with well-formed system settings
+  const setResult = await apiClient.setSystemSettings({
+    systemSettings: systemSettings.asText(),
+  });
+  assert(setResult.isOk());
+
+  const systemSettingsResult = await apiClient.getSystemSettings();
+  expect(systemSettingsResult).not.toBeUndefined();
```
arsalansufi, line 605: If you replace this with an `assert(systemSettingsResult !== undefined)`, on the next line, you can get by without the optional chaining `?`:

```ts
expect(systemSettingsResult.arePollWorkerCardPinsEnabled).toEqual(true);
```

apps/admin/backend/src/store.ts
```diff
@@ -235,6 +237,39 @@ export class Store {
     return settings?.currentElectionId ?? undefined;
   }
 
+  /**
+   * Creates a system settings record and returns its ID.
+   * Note `system_settings` are logical settings that span other machines eg. VxScan.
+   * `settings` are local to VxAdmin
+   */
+  addSystemSettings(systemSettings: SystemSettings): void {
```
arsalansufi, line 245: What do you think of renaming this `setSystemSettings` or `saveSystemSettings`, since `addSystemSettings` loosely implies that there can be multiple system settings objects?

apps/admin/frontend/src/api.ts
```diff
@@ -382,3 +392,17 @@ export const addPrintedBallots = {
...
 } as const;
+
+export const setSystemSettings = {
+  useMutation() {
+    const apiClient = useApiClient();
+    const queryClient = useQueryClient();
+    return useMutation(apiClient.setSystemSettings, {
+      async onSuccess() {
+        await queryClient.invalidateQueries(
+          getCurrentElectionMetadata.queryKey()
```
arsalansufi, line 403: I think `getCurrentElectionMetadata` should be `getSystemSettings` since that's the query we want to invalidate (probably just a copy-paste error)

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.test.tsx
```diff
@@ -80,31 +95,36 @@ beforeEach(() => {
...
-    } = renderInAppContext(<ExportElectionBallotPackageModalButton />, {
-      usbDrive: mockUsbDrive(usbStatus),
-    });
+test.each<{
+  usbStatus: UsbDriveStatus;
+}>([
+  { usbStatus: 'absent' },
+  { usbStatus: 'ejected' },
+  { usbStatus: 'bad_format' },
+])(
```
arsalansufi, line 120: :tada:

arsalansufi (approved):
A few last things then LGTM!

apps/admin/backend/src/store.ts
kshen0: Agree, changed in https://github.com/votingworks/vxsuite/pull/3147/commits/d6b5efd7f572c85c2a475a53edd0db0c76de2cc6

apps/admin/frontend/src/api.ts
kshen0: Thanks for the catch, indeed a copy and paste error https://github.com/votingworks/vxsuite/pull/3147/commits/d6b5efd7f572c85c2a475a53edd0db0c76de2cc6#diff-a21e7f57fc6e4d06d99aa89c6921cbc36789c5bd69e48f5c617f3805eb0c9943R424

PR #3148, opened 2023-03-16
refactor(basics): introduce `IteratorPlus` and async variant

eventualbuddha:
## Overview
Replaces the free functions like `map` and `take` that worked with iterators and replaces them with an `iter` function that returns either an `IteratorPlus` or `AsyncIteratorPlus`, which are just `Iterable` and `AsyncIterable` with a bunch of useful methods tacked on. The main difference from before is that instead of doing everything inside out you can chain things to make them more readable (and discoverable).

I'm hoping to make this feature more ergonomic and useful to facilitate working with streams of data rather than loading everything at once.

## Demo Video or Screenshot
```ts
// before
Array.from(map(
  take(5, naturals()),
  (n) => n * n
));

// after
naturals()
  .take(5)
  .map((n) => n * n)
  .toArray()
```

## Testing Plan
Automated tests, both existing and new.

## Checklist


libs/basics/src/iterators/iterator_plus.test.ts
```diff
@@ -0,0 +1,301 @@
...
+  expect(iter([]).take(-1).toArray()).toEqual([]);
+});
+
+test('drop', () => {
+  expect(iter(['a', 'b']).skip(1).take(1).toArray()).toEqual(['b']);
+  expect(iter(integers()).skip(2).take(2).toArray()).toEqual([2, 3]);
+  expect(iter(integers()).skip(0).take(2).toArray()).toEqual([0, 1]);
+  expect(iter(integers()).skip(-2).take(2).toArray()).toEqual([0, 1]);
+  expect(iter([1]).skip(2).take(2).toArray()).toEqual([]);
+});
```
adghayes, line 167: `skip`?

libs/basics/src/iterators/async_iterator_plus.test.ts
```diff
@@ -0,0 +1,435 @@
...
+  expect(await iter([]).async().take(1).toArray()).toEqual([]);
+  expect(await iter(['a', 'b']).async().take(1).toArray()).toEqual(['a']);
+  expect(await iter(integers()).async().take(1).toArray()).toEqual([0]);
+  expect(await iter(integers()).async().take(5).toArray()).toEqual([
+    0, 1, 2, 3, 4,
+  ]);
+  expect(await iter([]).async().take(-1).toArray()).toEqual([]);
+});
+
+test('drop', async () => {
```
adghayes, line 218: `skip`

libs/cvr-fixture-generator/src/generate_cvrs.ts
```diff
@@ -115,164 +120,172 @@ interface GenerateCvrsParams {
...
-              ballot.ballotConfig.ballotStyleId === ballotStyleId &&
-              ballot.ballotConfig.precinctId === precinctId &&
-              ballot.ballotConfig.isLiveMode === !testMode
-          ).layout;
-          if (ballotPageLayouts.length > 2) {
-            throw new Error('only single-sheet ballots are supported');
-          }
+}: GenerateCvrsParams): IteratorPlus<CVR.CVR> {
+  return iter(
+    (function* gen() {
```
adghayes, line 130: The syntax of defining generators in TS is already a bit confusing and in this case I can see how it's a bit weirder. No complaints, I would do it the same way, but I know I would find this confusing if I stumbled upon it and I wish there were an easier syntax.

adghayes (approved):
I'm impressed how quickly you put this together. LGTM!

libs/cvr-fixture-generator/src/generate_cvrs.ts
eventualbuddha: Yeah I don't love this either. Certainly open to suggestions on how to improve this. @jonahkagan?

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -99,13 +98,15 @@ export function convertVxfElectionToCdfBallotDefinition(
...
                   ...(contest.allowWriteIns
-                    ? take(contest.seats, naturals()).map(
-                        (writeInIndex): Cdf.CandidateOption => ({
-                          '@type': 'BallotDefinition.CandidateOption',
-                          '@id': `${contest.id}-option-write-in-${writeInIndex}`,
-                          IsWriteIn: true,
-                        })
-                      )
+                    ? naturals()
+                        .take(contest.seats)
```
jonahkagan, line 102: When I originally wrote this code, I wanted to just write `range(0, contest.seats)` and get an array. But I just worked with what we had.

Sometimes it takes a little more mental gymnastics to work in infinite sequence land when you're trying to do a finite thing. I wonder if we could have some finite versions of all these awesome infinite sequence tools.

eventualbuddha: It could also be `integers({ from: 1, through: contest.seats }).toArray()`, which would yield the same results (`naturals()` is just `integers({ from: 1 })`). It sounds like you wanted this to be 0-indexed, though. That'd be `integers({ through: contest.seats - 1 }).toArray()`.

libs/cvr-fixture-generator/src/generate_cvrs.ts
eventualbuddha: I wasn't sure which approach to use. It's nice to be able to just use these methods on the result, but you're right that this is pretty 👎. I'll just return `Generator<T>`.

PR #3149, opened 2023-03-16
VxAdmin - Move Machine Config to Backend

adghayes:
## Overview

Moves the `MachineConfig` state fetching to `frontend -> backend` rather than `frontend -> frontend server proxy`. Now more closely resembles VxScan.

## Testing Plan

Updated automated testing in the frontend, added a couple in the backend.


apps/admin/frontend/src/api.ts
```diff
@@ -273,7 +286,16 @@ export const unconfigure = {
...
+        // invalidate all queries except a select few
+        await queryClient.invalidateQueries({
+          predicate: (query: Query) => {
+            const queryKeyPrefix = query.queryKey[0];
+            return (
+              getMachineConfig.queryKeyPrefix !== queryKeyPrefix &&
+              getAuthStatus.queryKeyPrefix !== queryKeyPrefix
+            );
+          },
+        });
```
adghayes, line 298: I wanted to not clear on `unconfigure` mostly because it makes testing easier but also because we don't need to ever refetch the machine config.

I can see a couple of other ways of handling this. We could have a query key that prefixes all things that should be cleared on `unconfigure`, and then clear with that prefix here. I chose to avoid that because I was concerned that someone could forget to add this prefix to something that needed it, and then it wouldn't be cleared properly. We could also treat the `getMachineConfig` endpoint like the `getAuthStatus` endpoint and just mock a value. I'm would be happy with that if that feels less convoluted.

jonahkagan: This approach makes sense to me. I think it's a little overcomplicated by checking the prefix of the query key vs just checking the whole query key, since the two query keys we care about don't have anything other than the prefix.

jonahkagan (approved):
Glad to see this change, thanks for doing it!

Issue #2911, opened 2023-01-20
Update VxAdmin configuration flow


arsalansufi:

arsalansufi:
This ticket has become a little vague and catch-all, and I'm tracking remaining work through other issues now so am going to close this out

PR #3150, opened 2023-03-17
Move jurisdiction validation up from card layer to auth layer to better support non-VxAdmin configuration flows

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue links: Loosely https://github.com/votingworks/vxsuite/issues/2911 and https://github.com/votingworks/vxsuite/issues/2912

An important subtlety came up while discussing machine configuration flows: We _won't_ be locking non-VxAdmin machines to a specific jurisdiction at the Bakery (as we will be for VxAdmins). Just as these machines go from being election-agnostic to election-specific after configuration with an election manager card, they'll similarly go from being jurisdiction-agnostic to jurisdiction-specific after that configuration. After unconfiguration, they'll return to being jurisdiction-agnostic.

This strikes a balance between security (unlocking an active machine with a system administrator card from another jurisdiction will not be possible) and configuration simplicity (our Bakery configuration process for non-VxAdmin machines won't require any jurisdiction-specific actions - we can keep a general inventory and ship as needed). This also opens up the possibility of a jurisdiction lending another jurisdiction a machine - an unlikely possibility any time soon but could come up down the road.

To allow for this, this PR moves jurisdiction validation code up from the card layer to the auth layer and adds jurisdiction to the `DippedSmartCardAuthMachineState` and `InsertedSmartCardAuthMachineState` objects, where it can be provided by the machine alongside election hash since it's no longer static.

I could've alternatively kept jurisdiction validation in the card layer, but this would've required passing through the auth layer anyway. It also would've required rewiring the card layer to mirror the Schrodinger vibes of the auth layer (courtesy @jonahkagan for the analogy: https://github.com/votingworks/vxsuite/pull/3001#discussion_r1102151926). I also actually like the idea of keeping the card layer agnostic of machine state that can change - nice separation of concerns between the card layer and auth layer.

## Testing

- [x] Updated existing tests
- [x] Tested login and card programming manually

Will be accounting for these changes in my upcoming Java Card implementation tests, too

## Checklist


libs/auth/src/certs.ts
```diff
@@ -16,9 +16,9 @@ const VX_IANA_ENTERPRISE_OID = '1.3.6.1.4.1.59817';
 const VX_CUSTOM_CERT_FIELD = {
   /** One of: card, admin, central-scan, mark, scan (the latter four referring to machines) */
   COMPONENT: `${VX_IANA_ENTERPRISE_OID}.1`,
-  /** Format: {state-2-letter-abbreviation}.{county-or-town} (e.g. MS.Warren) */
+  /** Format: {state-2-letter-abbreviation}.{county-or-town} (e.g. ms.warren or ca.los-angeles) */
```
arsalansufi, line 19: Switching this to lower-case kebab-case for consistency with the other custom cert fields

libs/auth/src/certs.ts
```diff
@@ -39,7 +39,7 @@ export const STANDARD_CERT_FIELDS = [
 export interface CustomCertFields {
   component: 'card' | 'admin' | 'central-scan' | 'mark' | 'scan';
   jurisdiction: string;
-  cardType?: 'sa' | 'em' | 'pw';
+  cardType?: 'system-administrator' | 'election-manager' | 'poll-worker';
```
arsalansufi, line 42: Following up on an earlier piece of feedback to just use the full user role names since we're not strapped for space in the same way anymore

libs/auth/certs/dev/vx-admin-cert-authority-cert.pem
```diff
@@ -1,13 +1,13 @@
 -----BEGIN CERTIFICATE-----
```
arsalansufi, line 1: Just re-ran my handy `./scripts/generate-dev-keys-and-certs` script to update these! This will require anyone who has set up cards to re-run the `./scripts/configure-dev-java-card` script, but I think that's just me 😝

libs/auth/src/java_card.ts
```diff
@@ -378,10 +387,8 @@ export class JavaCard implements Card {
     // VotingWorks cert
     await this.verifyCardPrivateKey(CARD_VX_CERT.PRIVATE_KEY_ID, cardVxCert);
 
-    const user = await parseUserDataFromCert(
-      cardVxAdminCert,
-      this.jurisdiction
-    );
+    const { user, jurisdiction } = await parseUserDataFromCert(cardVxAdminCert);
+    assert(jurisdiction === vxAdminCertAuthorityCertDetails.jurisdiction);
```
arsalansufi, line 391: Though I'm no longer checking the jurisdiction against the machine jurisdiction in this layer, I'm still checking that the jurisdiction is consistent across certs stored on the card

apps/admin/backend/src/app.ts
```diff
@@ -49,12 +54,18 @@ function getCurrentElectionDefinition(
...
 }
 
-function constructDippedSmartCardAuthMachineState(
+function constructAuthMachineState(
   workspace: Workspace
 ): DippedSmartCardAuthMachineState {
   const currentElectionDefinition = getCurrentElectionDefinition(workspace);
   return {
     electionHash: currentElectionDefinition?.electionHash,
+    // TODO: Pull jurisdiction from VxAdmin cert authority cert
```
arsalansufi, line 63: I'll tackle this and the other TODOs in this commit as I complete https://github.com/votingworks/vxsuite/issues/2911 and https://github.com/votingworks/vxsuite/issues/2912

apps/admin/backend/src/app.ts
```diff
@@ -49,12 +54,18 @@ function getCurrentElectionDefinition(
...
+    jurisdiction: isFeatureFlagEnabled(
+      BooleanEnvironmentVariableName.ENABLE_JAVA_CARDS
+    )
+      ? /* istanbul ignore next */ DEV_JURISDICTION
+      : undefined,
```
arsalansufi, line 68: Falling back to `undefined` when `ENABLE_JAVA_CARDS` is `false` just makes sure that development with memory cards is unaffected by this check that doesn't apply to them

arsalansufi:
@benadida I don't think you need to explicitly review this but have added you to keep you in the loop, especially since this does touch validation logic

libs/auth/src/certs.ts
arsalansufi, line 19: Switching this to lower-case kebab-case for consistency with the other custom cert fields

libs/auth/src/certs.ts
arsalansufi, line 42: Following up on an earlier piece of feedback to just use the full user role names since we're not strapped for space in the same way anymore

libs/auth/certs/dev/vx-admin-cert-authority-cert.pem
arsalansufi, line 1: Just re-ran my handy `./scripts/generate-dev-keys-and-certs` script to update these! This will require anyone who has set up dev Java Cards to re-run the `./scripts/configure-dev-java-card` script, but I think that's just me 😝

apps/admin/backend/src/app.ts
arsalansufi, line 63: I'll tackle this and the other TODOs in this commit as I complete https://github.com/votingworks/vxsuite/issues/2911 and https://github.com/votingworks/vxsuite/issues/2912

apps/admin/backend/src/app.ts
arsalansufi, line 68: Falling back to `undefined` when `ENABLE_JAVA_CARDS` is `false` just makes sure that interim development with memory cards is unaffected (as a reminder, we don't validate jurisdiction when `machineState.jurisdiction` is `undefined`)

libs/auth/src/java_card_config.ts
```diff
@@ -37,18 +31,14 @@ function constructDevCardProgrammingConfig(
...
-}: {
+export function constructDevJavaCardConfig(input: {
   includeCardProgrammingConfig?: boolean;
   pathToAuthLibRoot: string;
 }): JavaCardConfig {
   return {
-    cardProgrammingConfig: includeCardProgrammingConfig
-      ? constructDevCardProgrammingConfig(pathToAuthLibRoot)
+    cardProgrammingConfig: input.includeCardProgrammingConfig
+      ? constructDevCardProgrammingConfig(input.pathToAuthLibRoot)
```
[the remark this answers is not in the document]
arsalansufi: No good reason here - just being loosely consistent with other points in the auth code where we don't destructure, e.g. https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/libs/auth/src/dipped_smart_card_auth.ts#L203-L209

But those other examples feel a bit different in that the grouping facilitates understanding, e.g. distinguishing what comes from `machineState` vs. `input`.

With this example that you've pointed out, I don't see a good reason to group under `input`, so I've switched back to destructuring: https://github.com/votingworks/vxsuite/pull/3150/commits/ad02399915d5987d243718472e12778a1c661daa

libs/auth/src/java_card.ts
```diff
@@ -323,24 +323,25 @@ export class JavaCard implements Card {
...
    *    reader
+   * This wrapper should never throw errors.
    */
-  private async safeReadUser(): Promise<Optional<UserWithCard>> {
+  private async safeReadCardDetails(): Promise<Optional<CardDetails>> {
     try {
-      return await this.readUser();
+      return await this.readCardDetails();
     } catch {
       return undefined;
```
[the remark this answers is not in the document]
arsalansufi: This actually gets logged upstream here: https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/libs/auth/src/dipped_smart_card_auth.ts#L73-L81

With the reason being `invalid_user_on_card`: https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/libs/auth/src/dipped_smart_card_auth.ts#L511-L513

I've definitely contemplated adding a lower-level log here to get more visibility into specific situations but ultimately decided that the value in that was minimal. The most common error case will almost certainly be inserting an unprogrammed card.

libs/auth/src/java_card.ts
```diff
@@ -323,24 +323,25 @@ export class JavaCard implements Card {
   }
 
   /**
-   * We wrap this.readUser() because it:
+   * We wrap readCardDetails to create safeReadCardDetails because readCardDetails:
```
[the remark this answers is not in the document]
arsalansufi: Yeah great questions!

> Are there places where we want readCardDetails to throw?

We ultimately never want the top-level card reading operation to throw. A bad card shouldn't be able to make a machine crash. It should instead surface a screen like this when the card is inserted to login:

image

This screen is surfaced by the auth layer when the card layer indicates that a card is in the card reader but a user could not be extracted from the card: https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/libs/auth/src/dipped_smart_card_auth.ts#L511-L513

Helpers are of course a different story and can throw all they need to communicate up to the top-level operation that the card is somehow invalid. Which leads to your second question:

> Can we simplify by getting rid of the wrapper and making readCardDetails itself safe?

We _could_ make the lower-level `readCardDetails` safe, but that would almost certainly complicate things rather than simplify. The simplest way to make it safe would be to wrap everything in a try-catch (basically what we do today but without the wrapper function). The more nuanced way to make it safe would be to have both it and all downstream helpers return `Result` objects (which as I'm sure you've now seen allow packaging both success results and error results without throwing). The main challenge here would be a constant need to check `Result` objects, through layers and layers of helpers. I touch on this in this old thread: https://github.com/votingworks/vxsuite/pull/3037#discussion_r1119011184

arsalansufi:
@kshen0 Responded to your comments/questions - let me know if you'd like to discuss any of them in person!

apps/mark/backend/src/app.ts
```diff
@@ -10,57 +14,74 @@ import {
...
+  isFeatureFlagEnabled,
+  ScannerReportData,
+  ScannerReportDataSchema,
+} from '@votingworks/utils';
 
 import { getMachineConfig } from './machine_config';
 
+function constructAuthMachineState(input: {
+  electionHash?: string;
+}): InsertedSmartCardAuthMachineState {
```
kshen0, line 28: Perhaps this is another place where `input` is unnecessary, since currently there's not much else in this function to distinguish from. But with outstanding TODOs that might change, so I'll leave it up to you

apps/scan/backend/src/app.ts
```diff
@@ -46,11 +49,19 @@ import { getMachineConfig } from './machine_config';
 import { CVR_EXPORT_FORMAT } from './globals';
 import { DefaultMarkThresholds } from './store';
 
-function constructInsertedSmartCardAuthMachineState(
+function constructAuthMachineState(
   workspace: Workspace
 ): InsertedSmartCardAuthMachineState {
   const electionDefinition = workspace.store.getElectionDefinition();
```
kshen0, line 55: I see here we don't use `input`; consider standardizing all `constructAuthMachineState` functions to use the same convention.

kshen0 (approved):
Just one more nit about standardizing `constructAuthMachineState` args (no strong opinion on if you use `input` or not). Then LGTM!

apps/mark/backend/src/app.ts
arsalansufi: Agreed - went ahead and changed! https://github.com/votingworks/vxsuite/pull/3150/commits/cc7f767279b4cb457fe2f4365ee0ff4cca84f0a5

apps/scan/backend/src/app.ts
arsalansufi: Once nuance here is that VxMark's machine state construction currently looks different than the other machines because it's the one machine where the election definition doesn't yet live on the backend and needs to be provided by the frontend. Once we make that change, VxMark too will just take the `workspace` as its only input

PR #3151, opened 2023-03-20
Add tests for Java Card implementation (and bring libs/auth code coverage to 100%)

arsalansufi:
_Easiest reviewed by commit_

## Overview

Finally done with `libs/auth` tests! :tada: :relieved:

## Checklist


libs/auth/src/java_card.test.ts
```diff
@@ -0,0 +1,938 @@
...
+} from './piv';
+
+jest.mock('./card_reader');
+jest.mock('./openssl', (): typeof import('./openssl') => ({
+  // We use real openssl commands in these tests to ensure end-to-end correctness, the one
+  // exception being openssl commands for cert creation since two cert creation commands with the
+  // exact same inputs won't necessarily generate the same outputs, making assertions difficult
+  ...jest.requireActual('./openssl'),
+  createCert: jest.fn(),
+}));
```
arsalansufi, line 63: I'd initially planned to completely mock the `openssl` lib but then realized that at no point would I be testing the openssl commands underlying the Java Card implementation. With the new approach of using realistic test files and real openssl commands, I'm much more confident in our ability to prevent regressions

libs/auth/scripts/generate_dev_keys_and_certs.ts
```diff
@@ -100,6 +117,78 @@ async function generateDevVxAndVxAdminKeysAndCerts(
...
+      { role: 'poll_worker', electionHash },
+    ];
+    for (const user of users) {
+      const prefix = user.role.replace('_', '-');
+      const cardVxPrivateKeyPath = `${outputDir}/${prefix}-card-vx-private-key.pem`;
+      const cardVxPublicKeyPath = `${outputDir}/${prefix}-card-vx-public-key.der`;
+      const cardVxCertPath = `${outputDir}/${prefix}-card-vx-cert.der`;
+      const cardVxAdminPrivateKeyPath = `${outputDir}/${prefix}-card-vx-admin-private-key.pem`;
+      const cardVxAdminPublicKeyPath = `${outputDir}/${prefix}-card-vx-admin-public-key.der`;
+      const cardVxAdminCertPath = `${outputDir}/${prefix}-card-vx-admin-cert.der`;
```
adghayes, line 141: At this point in my review, I am a little confused why we are generating all the `cardVxAdmin` keys as fixtures rather than generating them as part of the tests, but I assume my review will reveal this.

libs/auth/test/utils.ts
```diff
@@ -12,6 +18,25 @@ export function numericArray(input: {
...
+/**
+ * A mock card reader
+ */
+export class MockCardReader implements Pick<CardReader, 'transmit'> {
+  private readonly onReaderStatusChange: OnReaderStatusChange;
+
+  constructor(...params: ConstructorParameters<typeof CardReader>) {
+    this.onReaderStatusChange = params[0].onReaderStatusChange;
+  }
+
```
adghayes, line 30: TIL `ConstructorParameters`. I would prefer this because it makes it a tiny bit easier to read and consistent with your `CardReader` constructor signature, but this is a nitty nit nit.

```suggestion
  constructor(input: ConstructorParameters<typeof CardReader>[0]) {
    this.onReaderStatusChange = input.onReaderStatusChange;
  }
```

adghayes (approved):
LGTM! Exciting to have this fully covered. Some nit feedback.

I don't have a nitty-gritty understanding of all the mechanisms that are being mocked in the main test file, but I don't feel that I need to in order to approve this. The test cases do a good job of laying out intent.

libs/auth/test/utils.ts
```diff
@@ -50,3 +75,44 @@ export function mockCardAssertComplete(mockCard: MockCard): void {
...
+  fileType:
+    | 'vx-admin-cert-authority-cert.der'
+    | 'vx-admin-cert-authority-cert.pem'
+    | 'vx-admin-private-key.pem'
+    | 'vx-cert-authority-cert.pem'
+    | 'vx-private-key.pem';
+  setId?: TestFileSetId;
+}
+
+interface UserSpecificTestFile {
```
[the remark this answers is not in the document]
adghayes: Oh, yes I ended up commenting on this in my review. I think that renaming make sense.

libs/auth/src/java_card.ts
```diff
@@ -318,7 +328,6 @@ export class JavaCard implements Card {
   }
 
   async clearData(): Promise<void> {
-    await this.selectApplet();
```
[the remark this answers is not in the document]
arsalansufi: Ah yes good q - should've left a note clarifying!

Because `this.clearData` just calls `this.writeData`, and `this.writeData` calls `this.selectApplet` internally, there's no need to call it here. I was initially worried about someone forgetting to add a `this.selectApplet` back in if they ever updated `this.clearData` to no longer call `this.writeData`, but we know have tests to prevent a regression like this.

All that said, a comment does feel in order since this is the only top-level method that doesn't call `this.selectApplet`

libs/auth/test/utils.ts
arsalansufi: Agree that what you have is more intuitive! It does always take me a second to process `...params`

libs/auth/scripts/generate_dev_keys_and_certs.ts
```diff
@@ -108,10 +197,24 @@ async function generateDevVxAndVxAdminKeysAndCerts(
...
+ * - A card VxAdmin cert by signing the second key pair's public key with the VxAdmin cert
+ *   authority cert
  */
 export async function main(): Promise<void> {
   try {
-    await generateDevVxAndVxAdminKeysAndCerts('./certs/dev');
+    await generateDevVxAndVxAdminKeysAndCerts({
+      forTests: asBoolean(process.env['FOR_TESTS']),
+      outputDir: process.env['OUTPUT_DIR'] ?? './certs/dev',
+    });
```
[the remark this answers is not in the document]
arsalansufi: Good point re env var conflicts! Of course not too likely here but was easy enough to switch to command-line args, given that I use them in another `libs/auth` script: https://github.com/votingworks/vxsuite/pull/3151/commits/b5af512f411835e2c673b57d54c8d7c734aa9ce4

PR #3159, opened 2023-03-21
chore(scan): remove `UninterpretedHmpbPage`

eventualbuddha:
## Overview
We don't ever use the fact that this `PageInterpretation` has `metadata`, so it doesn't distinguish itself meaningfully from `UnreadablePage`. So let's just use `UnreadablePage`.

This is part of an effort to simplify the scanning types.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist


apps/central-scan/backend/src/cvrs/export.ts
```diff
@@ -25,16 +25,13 @@ import {
 
 function isHmpbPage(
   interpretation: PageInterpretation
-): interpretation is HmpbPageInterpretation {
-  return (
-    interpretation.type === 'InterpretedHmpbPage' ||
-    interpretation.type === 'UninterpretedHmpbPage'
-  );
+): interpretation is InterpretedHmpbPage {
+  return interpretation.type === 'InterpretedHmpbPage';
```
carolinemodic, line 29: how are these helpers used? Is this going to change whether or not these pages are exported if this now returns false instead of true?

carolinemodic (approved):
Isn't there frontend logic on these page types too does anything need to change there?

eventualbuddha:
> Isn't there frontend logic on these page types too does anything need to change there?

I believe [`BallotEjectScreen`](https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/apps/central-scan/frontend/src/screens/ballot_eject_screen.tsx#L68) was the only place that referenced it. And that means that for a `UninterpretedHmpbPage`-sheet it would:
1. set [`isUnreadableSheet = true`](https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/apps/central-scan/frontend/src/screens/ballot_eject_screen.tsx#L295)
2. cause [`allowBallotDuplication` to be `false`](https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/apps/central-scan/frontend/src/screens/ballot_eject_screen.tsx#L299-L302)
3. make it so [`Tabulate As Is` will never appear](https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/apps/central-scan/frontend/src/screens/ballot_eject_screen.tsx#L355-L376)
4. which means we never call [`continueScanning` with `forceAccept: true`](https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/apps/central-scan/frontend/src/screens/ballot_eject_screen.tsx#L475-L484)
5. so [`continueImport` will only ever be called with `forceAccept: false`](https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/apps/central-scan/backend/src/central_scanner_app.ts#L392)
6. which means [the sheet will be deleted](https://github.com/votingworks/vxsuite/blob/b5328da61d490fd54899882c8d1fd22f3337e5cb/apps/central-scan/backend/src/importer.ts#L501)
7. which means it'll never get exported in the first place, at least from `central-scan`.

PR #3129, opened 2023-03-14
perf(scan-nh): introduce a faster interpreter

eventualbuddha:
## Overview

Adds a new AccuVote-compatible interpreter that uses 🦀 Rust to be much faster. It lives in `ballot-interpreter-nh-next` and is intended to replace `ballot-interpreter-nh` eventually.

## TODO
- [ ] make debug logging from Rust work properly

## Demo Video or Screenshot
Benchmarking `ballot-interpreter-nh` vs `benchmark-interpreter-nh-next`:
| Interpreter | Average (ms) | Ops/sec | Multiple slower than fastest |
|-|-|-|-|
| ballot-interpreter-nh | 1331 | 0.75 | 19x |
| ballot-interpreter-nh-next | 71 | 14 | fastest |

<details>
  <summary>Benchmark source code</summary>

```js
const { Bench } = require('tinybench');
const ballotInterpreterNh = require('./libs/ballot-interpreter-nh');
const ballotInterpreterNhNext = require('./libs/ballot-interpreter-nh-next');
const fixtures = require('./libs/fixtures');
const amherst = fixtures.electionGridLayoutNewHampshireAmherstFixtures;

async function main() {
  /** @type {import('./libs/types').SheetOf<string>} */
  const sheet = [
    amherst.scanMarkedFront.asFilePath(),
    amherst.scanMarkedBack.asFilePath(),
  ];
  const bench = new Bench();

  bench
    .add('ballot-interpreter-nh', async () => {
      await ballotInterpreterNh.interpret(amherst.electionDefinition, sheet, {
        isTestMode: true,
      });
    })
    .add('ballot-interpreter-nh-next', () => {
      ballotInterpreterNhNext.interpret(amherst.electionDefinition, sheet);
    });

  await bench.run();

  const fastest = bench.tasks.reduce((fastest, task) =>
    !fastest.result
      ? task
      : task.result && task.result.mean < fastest.result.mean
      ? task
      : fastest
  );

  console.table(
    bench.tasks.map(({ name, result }) => ({
      'Task Name': name,
      'Average Time (ms)': result?.mean,
      'Ops Per Second': result?.hz,
      Multiplier:
        !result || !fastest.result
          ? NaN
          : `${Math.round(result.mean / fastest.result.mean)}x`,
    }))
  );
}

if (require.main === module) {
  main().catch((error) => {
    console.error(error);
    process.exit(1);
  });
}
```

</details>

## Testing Plan
[...]

PR #3173, opened 2023-03-21
refactor(scan): centralize various ballot mark utilities

eventualbuddha:
## Overview

_(review by commit)_

I plan to use this in #3129 to reduce the duplication of converting marks to votes that is introduced for the third time in that PR.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist


libs/utils/src/votes.ts
```diff
@@ -653,3 +660,73 @@ export function castVoteRecordHasWriteIns(cvr: CastVoteRecord): boolean {
...
+    assert(contest, `Contest not found: ${mark.contestId}`);
+    const existingVotes = votesDict[mark.contestId] ?? [];
+    const newVotes =
+      contest.type === 'candidate'
+        ? markToCandidateVotes(contest, markThresholds, mark)
+        : contest.type === 'yesno'
+        ? markToYesNoVotes(markThresholds, mark)
+        : /* istanbul ignore next */ throwIllegalValue(contest, 'type');
+
+    if (newVotes.length > 0) {
```
jonahkagan, line 745: This check doesn't seem necessary

libs/ballot-interpreter-vx/src/get_votes_from_marks.ts
```diff
@@ -22,63 +11,12 @@ export function getVotesFromMarks(
...
-        throwIllegalValue(mark, 'type');
-    }
-  }
-
-  return votes;
+  return convertMarksToVotesDict(
+    election.contests,
+    {
+      marginal: markScoreVoteThreshold,
+      definite: markScoreVoteThreshold,
```
jonahkagan, line 18: Seems like this is taking advantage of insider knowledge that `convertMarksToVotesDict` isn't going to use `markThresholds.marginal` at all. Maybe change `convertMarksToVotesDict` to take `markThresholds: Pick<MarkThresholds, 'definite'>` so that you don't have to pass in `marginal`?

libs/backend/src/scan/cast_vote_records/build_cast_vote_record.ts
```diff
@@ -372,7 +374,10 @@ function buildOriginalSnapshot({
                 (Math.floor(mark.score * 100) / 100).toString(),
               ],
               HasIndication:
-                mark.score >= definiteMarkThreshold
+                getMarkStatus(mark.score, {
+                  marginal: definiteMarkThreshold,
```
jonahkagan, line 378: This is a bit tricky. Maybe omitting `marginal` would be a clearer way to specify that you want to ignore marginal marks?

Alternatively could change this to check `!== MarkStatus.Unmarked`

jonahkagan (approved):

libs/backend/src/scan/cast_vote_records/build_cast_vote_record.ts
eventualbuddha: You can now omit `marginal` as you suggested.

PR #3175, opened 2023-03-21
Import CDF CVRs into VxAdmin Pt. 1

adghayes:
## Overview

Adds the ability to import CDF CVRs into VxAdmin, behind a feature flag. The feature is not yet complete - a subsequent "Pt. 2" of this PR will grab the layouts and ballot images as needed. I wanted to separate them out to avoid the PR getting too grande. 

## Testing Plan
New test file specifically for CDF CVR files, which will be the CVR import test file after the conversion. I preferred to test at the `app` level rather than write tests at the `store` or helper file level.

## Checklist


apps/admin/backend/src/cvr_files.ts
```diff
@@ -96,3 +115,282 @@ export async function listCastVoteRecordFilesOnUsb(
...
+  }
+}
+
+/**
+ * Converts the vote data in the CDF cast vote record into the simple
+ * dictionary of contest ids to contest selection ids that VxAdmin uses
+ * internally as a basis for tallying votes.
+ *
+ * This method assumes that if a contest selection is present, then it an
+ * indication. It ignores the possibility that there are multiple selection
```
jonahkagan, line 353: "This method assumes that if a contest selection is present, then it an indication" is missing a verb between "it" and "an"

apps/admin/backend/src/cvr_files.ts
```diff
@@ -96,3 +115,282 @@ export async function listCastVoteRecordFilesOnUsb(
...
+ * positions within the contest selection (only applicable to ranked choice and
+ * other forms of voting) or that the selection position has no indication.
```
jonahkagan, line 355: Would it make sense to assert these assumptions in the code to help us remember to change it later if we ever need to?

apps/admin/backend/src/cvr_files.ts
```diff
@@ -96,3 +115,282 @@ export async function listCastVoteRecordFilesOnUsb(
...
+ */
+function convertCastVoteRecordVotesToLegacyVotes(
+  cvrSnapshot: CVR.CVRSnapshot
+): Record<string, string[]> {
```
jonahkagan, line 359: Is this different from the `VotesDict` type?

apps/admin/backend/src/store.ts
```diff
@@ -105,6 +105,22 @@ export class Store {
...
+   *
+   * Returns the result of the function.
+   */
+  withTransaction<T>(
+    fn: () => Promise<T>,
+    shouldCommit?: (result: T) => boolean
+  ): Promise<T>;
+  withTransaction<T>(fn: () => T, shouldCommit?: (result: T) => boolean): T {
+    return this.client.transaction(() => fn(), shouldCommit);
+  }
```
jonahkagan, line 122: I think it's safe to assume that throwing an exception or returning an `err` should roll back. Accepting a `shouldCommit` parameter feels a bit too flexible for our use cases

apps/admin/backend/src/cvr_files.ts
```diff
@@ -394,3 +406,242 @@ export function convertCastVoteRecordToLegacyFormat({
...
+  exportedTimestamp,
+}: {
+  store: Store;
+  reportDirectoryPath: string;
+  exportedTimestamp: Iso8601Timestamp;
+}): Promise<AddCastVoteRecordReportResult> {
+  const electionId = store.getCurrentElectionId();
+  if (!electionId) {
+    throw new Error('no current election id');
+  }
```
jonahkagan, line 464: Nit: I prefer `assert` in these cases to express that we're expecting this to never happen

apps/admin/backend/src/cvr_files.ts
```diff
@@ -394,3 +406,242 @@ export function convertCastVoteRecordToLegacyFormat({
...
+    });
+  }
+  const relativeBallotImagePaths = directoryValidationResult.ok();
+  const reportPath = join(
+    reportDirectoryPath,
+    CAST_VOTE_RECORD_REPORT_FILENAME
+  );
+
+  // We are hashing only the JSON report here - perhaps we should hash some
+  // structure of the report directory as well or even the images.
```
jonahkagan, line 487: What's the purpose of the hash?

jonahkagan (approved):
Commits were pretty clearly separated - thanks! It's helpful if you leave a note telling me to review by commits so I know if that's your intention or not

Issue #2524, opened 2022-09-02
Implement poll worker card PINs


mattroe:
VVSG requires that all authentication implements MFA, which is met through smart card PINs today. We decided to not implement this for poll workers initially as we determined it to be unnecessarily arduous for our customers. The language of VVSG states the system "must be capable of", which suggests that this can be an optional configuration for administrators.

VVSG 11.3.1-B.4: At a minimum, the voting system must be capable of using multi-factor authentication to verify a user has authorized access to perform critical operations, including: At a minimum, the voting system must be capable of using multi-factor authentication to verify a user has authorized access to perform critical operations, including: *changing device states, including opening and closing the polls*

Before creating smart cards, a system administrator should have the option of enabling poll worker PINs in VxAdmin. Cards programmed with this option enabled should be programmed with PINs in the same manner as Election Manager cards.

kofi-q:
@mattroe - is this a setting we should make consistent across all cards for a given election? Or would we allow an admin to enable/disable poll worker PINs at any point between programming the cards?

mattroe:
@kofi-q As we discussed synchronously:
- a user will not want to change this within an election (and should probably not be allowed to)
- administrators will most likely keep the same setting across all elections
- this could be a VxAdmin-device-level setting with some of the additional device-level auth settings we'll need to implement

PR #3179, opened 2023-03-22
Add support for poll worker cards with PINs (disabled by default)

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2524

This PR adds support for poll worker cards with PINs, assuming the relevant system setting is enabled.

We're disabling the setting by default because we believe that poll worker card PINs are more likely to cause election-day disruption than add security, especially given that they'll often be written on the backs of cards. (We still believe that sys admin and election manager card PINs are a must.) The real motivation for this is VVSG2 compliance, which requires support for poll worker card PINs. So one step closer to compliance - woohoo!

Also of note, we're handing inconsistencies as follows:

- If the system settings says that poll worker card PINs are enabled, and a previously programmed poll worker card without a PIN (but valid in all other respects) is inserted, it'll be treated as unprogrammed.
- If the system settings says that poll worker card PINs are not enabled, and a previously programmed poll worker card with a PIN is inserted, it'll also be treated as unprogrammed.

Lastly, I haven't finished integrating with system settings in VxCentralScan, VxMark, and VxScan because that thread is in flight - will take care of that once I can!

## Screenshots

_Poll worker card programming when poll worker card PINs are enabled_

pw-card-w-pin-programmed

pw-card-w-pin

_Poll worker card programming when poll worker card PINs are not enabled_

pw-card-wout-pin-programmed

pw-card-wout-pin

## Testing

- [x] Updated `libs/auth` tests
- [x] Updated VxAdmin frontend tests
- [x] Manually tested programming poll worker cards with and without PINs
- [x] Manually tested inconsistency detection

## Checklist


libs/auth/src/card.ts
```diff
@@ -3,17 +3,63 @@ import {
...
+}
+
+interface PollWorkerCardDetails {
+  jurisdiction: string;
+  user: PollWorkerUser;
+  /**
+   * Unlike system administrator and election manager cards, which always have PINs, poll worker
+   * cards by default don't have PINs but can if the relevant system setting is enabled.
+   */
+  hasPin: boolean;
```
arsalansufi, line 25: I consciously decided not to add hasPin to the card details for the other user roles that always have PINs. Didn't want to create the false sense that having a PIN or not was an option for them

libs/auth/src/certs.ts
```diff
@@ -18,7 +18,7 @@ const VX_CUSTOM_CERT_FIELD = {
   COMPONENT: `${VX_IANA_ENTERPRISE_OID}.1`,
   /** Format: {state-2-letter-abbreviation}.{county-or-town} (e.g. ms.warren or ca.los-angeles) */
   JURISDICTION: `${VX_IANA_ENTERPRISE_OID}.2`,
-  /** One of: system-administrator, election-manager, poll-worker */
+  /** One of: system-administrator, election-manager, poll-worker, poll-worker-with-pin */
   CARD_TYPE: `${VX_IANA_ENTERPRISE_OID}.3`,
```
arsalansufi, line 22: I decided to use the existing cert card type field as opposed to a new field, once again to avoid creating the false sense that having a PIN or not was an option for sys admins and election managers

libs/auth/src/java_card.ts
```diff
@@ -393,15 +420,17 @@ export class JavaCard implements Card {
...
      *
-     * If the user does have a PIN:
+     * If the card does have a PIN:
      * Perform this verification later in checkPin because operations with this private key are
      * PIN-gated
      */
-    if (cardDetails.user.role === 'poll_worker') {
+    const cardDoesNotHavePin =
+      arePollWorkerCardDetails(cardDetails) && !cardDetails.hasPin;
+    if (cardDoesNotHavePin) {
```
arsalansufi, line 436: @benadida I imagine that this is a key point you'll be looking at in your security review - making sure that we can't somehow trick the system into skipping this card VxAdmin cert validation.

What makes this safe is the inconsistency handling that I describe in the PR description:

> Also of note, we're handing inconsistencies as follows:
> - If the system settings says that poll worker card PINs are enabled, and a previously programmed poll worker card without a PIN (but valid in all other respects) is inserted, it'll be treated as unprogrammed.
> - If the system settings says that poll worker card PINs are not enabled, and a previously programmed poll worker card with a PIN is inserted, it'll also be treated as unprogrammed.

So for example, if poll worker card PINs are not enabled, and we accordingly never call `checkPin` for them, could someone use a poll worker card with a PIN to completely skip validation of the card VxAdmin cert? The answer is no, because the card will be treated as unprogrammed here: https://github.com/votingworks/vxsuite/blob/b01a1687edd10e9ee2a13faa628c0a05597e8396/libs/auth/src/inserted_smart_card_auth.ts#L469-L478

libs/auth/src/memory_card.ts
```diff
@@ -141,11 +145,43 @@ export class MemoryCard implements Card {
...
+          jurisdiction: DEV_JURISDICTION,
+          user,
+        };
+        break;
+      }
+      case 'poll_worker': {
+        cardDetails = {
+          jurisdiction: DEV_JURISDICTION,
+          user,
+          hasPin: pin !== undefined,
```
arsalansufi, line 175: Was nearly going to punt on updating this deprecated memory card code path but had to do more than half the work just to get the TS compiler happy. So went ahead and completed things

libs/auth/src/inserted_smart_card_auth.ts
```diff
@@ -342,21 +348,33 @@ export class InsertedSmartCardAuth implements InsertedSmartCardAuthApi {
...
+                  case 'system_administrator': {
+                    return skipPinEntry
+                      ? { status: 'logged_in', user }
+                      : { status: 'checking_pin', user };
+                  }
+                  case 'election_manager': {
+                    return skipPinEntry
+                      ? { status: 'logged_in', user }
+                      : { status: 'checking_pin', user };
+                  }
```
arsalansufi, line 364: I know it looks like some of these cases could be merged, but TS requires me to separate them out

arsalansufi:
@kshen0 I'll have you review this one since it involves the `arePollWorkerCardPinsEnabled` settings. (@adghayes you got auto-assigned, but I removed you - planning to assign my next PR to you)

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -336,8 +348,9 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
         assert(electionHash !== undefined);
         await this.card.program({
           user: { role: 'poll_worker', electionHash },
+          pin: arePollWorkerCardPinsEnabled ? pin : undefined,
         });
-        return undefined;
+        return arePollWorkerCardPinsEnabled ? pin : undefined;
```
kshen0, line 353: nit: IMO it's a little safer to put all the logic in a single if block, like
```
    if (arePollWorkerCardPinsEnabled) {
      await this.card.program({
        user: { role: 'poll_worker', electionHash },
        pin,
      });
      return pin;
    }

    await this.card.program({
      user: { role: 'poll_worker', electionHash },
    });
```

but just personal preference

apps/admin/frontend/src/components/smartcard_modal/card_details_view.tsx
```diff
@@ -104,11 +111,11 @@ export function CardDetailsView({
...
-    (role === 'election_manager' &&
-      // We can support PIN resets on cards from other elections once we update PIN resetting to
-      // change only PINs and leave other card data, like election definitions, intact. As of
-      // 8/4/22, PIN resetting reprograms cards entirely
-      doesCardElectionHashMatchMachineElectionHash)
+    // Because PIN resetting completely reprograms the card under the hood, we also need the
+    // relevant election definition to be loaded for election cards
+    (doesCardElectionHashMatchMachineElectionHash &&
+      (role === 'election_manager' ||
+        (arePollWorkerCardPinsEnabled && role === 'poll_worker')))
```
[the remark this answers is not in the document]
benadida: That comment should say "election *manager* cards". The election definition is stored on the election manager card currently, as a way to load it onto machines. But not on the poll worker card.

libs/auth/src/certs.ts
benadida: I like it, fewer ways to make mistakes.

libs/auth/src/certs.ts
```diff
@@ -39,7 +39,8 @@ export const STANDARD_CERT_FIELDS = [
 export type CardType =
   | 'system-administrator'
   | 'election-manager'
-  | 'poll-worker';
+  | 'poll-worker'
+  | 'poll-worker-with-pin';
```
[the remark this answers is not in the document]
benadida: agreed @kshen0 with your analysis, and I think we can safely say we don't want other cards without PINs.

benadida (approved):
looks good to me on the security front.

apps/admin/frontend/src/components/smartcard_modal/card_details_view.tsx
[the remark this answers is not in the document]
arsalansufi: Ah yeah a few points of clarification here:

- Poll worker cards are election cards. Election cards = election manager cards and poll worker cards
- While we only write the full election definition to election manager cards (as Ben has pointed out), poll worker cards do also need the election definition for the election hash. Both election manager cards and poll worker cards have an election hash on them.
- Returning to the condition here, the `doesCardElectionHashMatchMachineElectionHash` is actually applying to both election manager cards and poll worker cards if you look at the (many 😅) parentheses.

I'll polish the comment and maybe the booleans, too, for clarity but hope that clears things up!

libs/auth/src/certs.ts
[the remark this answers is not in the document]
arsalansufi: Yup, exactly right!

> But if we don't anticipate more role-specific options

And as Ben has noted, we definitely don't envision making PINs an option for other cards. And there are no other role--specific options that I can anticipate given either product direction or VVSG2 compliance needs.

libs/auth/src/dipped_smart_card_auth.ts
arsalansufi: Nice, I agree - will update!

kshen0 (approved):
Thanks for updating the comment!

PR #3180, opened 2023-03-22
Require contest ids in layouts

adghayes:
_Review commit-by-commit_

## Overview

We use our layouts to determine where a contest is on a ballot, but on the VotingWorks-style ballots we don't have contest IDs associated with the contest layouts. This means that to find the layout of a given contest, we have to iterate through all the contests on all its layouts until we find the correct index. We do this when exporting CVRs and when finding boundaries for write-in images, and possibly other places. 

When exporting CDF CVRs, I made the decision to only export images and layouts that have write-ins (not front _and_ back, only the side with write-ins). However, I was unable to find the right contest in VxAdmin because I wasn't able to loop through all the layouts. Even if we reverted the CDF CVR decision, I'm not sure how the current style will play with multi-page ballots?

In this PR, we make contest id's required and add it to the vx interpreter.

## Testing Plan
- Light automated testing on new `ballot-interpreter-vx` export
- Light manual testing with the new code paths

## Checklist


apps/central-scan/backend/src/store.ts
```diff
@@ -1103,15 +1103,18 @@ export class Store {
...
         loadedLayouts.push(
           await interpretTemplate({
             electionDefinition,
             imageData: page,
             metadata: ballotPageLayout.metadata,
+            contestOffset,
           })
         );
+        contestOffset += ballotPageLayout.contests.length;
       }
```
adghayes, line 1118: Why are we using `interpretTemplate` in the scanner stores? We already have both the images and the layouts, why re-interpret??

apps/central-scan/integration-testing/cypress/e2e/configuration.cy.ts
```diff
@@ -103,7 +103,10 @@ describe('BSD and services/Scan', () => {
       .click();
     cy.contains('No ballots have been scanned', { timeout: 30000 });
     cy.contains('Scan New Batch').click();
-    cy.contains('A total of 1 ballot has been scanned in 1 batch.');
+
+    // The ballot package is currently out of sync with the image fixtures
+    // TODO: fix this, probably once our ballot package format changes
+    cy.contains('Wrong Election');
```
adghayes, line 109: This is lazy, but it still satisfies the integration test's goal of seeing if we can scan a ballot. Don't want to invest time in re-crafting images.

libs/ballot-interpreter-vx/src/layout.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    ballotPdfData,
+    metadata,
+  });
+
+  expect(layoutsWithImages).toHaveLength(2);
+  expect(
+    layoutsWithImages[0].ballotPageLayout.contests.map(
+      (contest) => contest.contestId
+    )
+  ).toMatchObject([
```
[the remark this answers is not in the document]
adghayes: Yup, good point.

adghayes, line 31:
```suggestion
  ).toEqual([
```

libs/ballot-interpreter-vx/src/layout.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    '775020892',
+    '775020890',
+    '775021420',
+    '775021421',
+  ]);
+  expect(
+    layoutsWithImages[1].ballotPageLayout.contests.map(
+      (contest) => contest.contestId
+    )
+  ).toMatchObject(['750000261', '750000262', '750000263']);
```
adghayes, line 43:
```suggestion
  ).toEqual(['750000261', '750000262', '750000263']);
```

libs/ballot-interpreter-vx/src/layout.test.ts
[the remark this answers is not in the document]
adghayes: Changed.

libs/ballot-interpreter-vx/src/interpreter/index.ts
```diff
@@ -176,6 +176,10 @@ export class Interpreter {
       imageData,
       metadata,
       imdebug,
+      // the below is incorrect and may mean that contest ids are not accurate
+      // for this method. currently this method is only being used in tests and
+      // the CLI
+      contestOffset: 0,
```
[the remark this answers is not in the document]
eventualbuddha: I think we should get rid of it or deprecate it, yeah. I took a stab at removing its uses in https://github.com/votingworks/vxsuite/pull/3183, which you can use if you like.

apps/central-scan/backend/src/store.ts
eventualbuddha: Yeah seems not necessary.

eventualbuddha:
My high-level viewpoint is that this is something we should do and this seems like a good start in that direction.

libs/ballot-interpreter-vx/src/interpreter/index.ts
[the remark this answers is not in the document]
eventualbuddha: I finished this in 61581bd4bbe6d9685e588644d01440bd56bb1556.

apps/central-scan/integration-testing/cypress/e2e/configuration.cy.ts
```diff
@@ -104,10 +104,6 @@ describe('BSD and services/Scan', () => {
     cy.contains('No ballots have been scanned', { timeout: 30000 });
     cy.contains('Scan New Batch').click();
 
-    // The ballot package is currently out of sync with the image fixtures
-    // TODO: fix this, probably once our ballot package format changes
-    cy.contains('Wrong Election');
-
```
adghayes: @eventualbuddha are you removing this because you updated the image fixtures? In which case we could check for the correct ballot interpretation.

If not, I do think it's valuable to check this because it tests that the interpreter actually tried to do something.

adghayes:
@eventualbuddha does this just need approval and merging?

apps/central-scan/integration-testing/cypress/e2e/configuration.cy.ts
eventualbuddha: This was actually failing, I think because I updated the fixtures.

eventualbuddha:
> @eventualbuddha does this just need approval and merging?

Yes.

eventualbuddha (approved):
Approving, but @adghayes you should look over the code I added.

adghayes:
Yup @eventualbuddha, looked it over and the changes look great to me. I'm glad you were able to just remove the CLI. After your updated the ballot package, the "Wrong Election" hack was no longer necessary so I added back in the original test assertion which is better.

PR #3181, opened 2023-03-22
feat(basics): add `windows` method for `iter`

eventualbuddha:
## Overview
As long as we're aping Rust's `Iterator`, let's do it properly! https://doc.rust-lang.org/1.68.0/std/iter/trait.Iterator.html

I took [this comment](https://github.com/votingworks/vxsuite/pull/3129#discussion_r1142593332) as tacit support for this.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist

### DIFF
--- a/libs/basics/src/iterators/async_iterator_plus.ts
+++ b/libs/basics/src/iterators/async_iterator_plus.ts
@@ -307,6 +307,33 @@ export class AsyncIteratorPlusImpl<T> implements AsyncIteratorPlus<T> {
     return set;
   }
 
+  windows(groupSize: 0): never;
+  windows(groupSize: 1): AsyncIteratorPlus<[T]>;
+  windows(groupSize: 2): AsyncIteratorPlus<[T, T]>;
+  windows(groupSize: 3): AsyncIteratorPlus<[T, T, T]>;
+  windows(groupSize: 4): AsyncIteratorPlus<[T, T, T, T]>;
+  windows(groupSize: 5): AsyncIteratorPlus<[T, T, T, T, T]>;
+  windows(groupSize: number): AsyncIteratorPlus<T[]>;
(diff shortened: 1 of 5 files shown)


libs/basics/src/iterators/async_iterator_plus.ts
```diff
@@ -294,6 +294,33 @@ export class AsyncIteratorPlusImpl<T> implements AsyncIteratorPlus<T> {
...
   }
 
+  windows(groupSize: 0): never;
+  windows(groupSize: 1): AsyncIteratorPlus<[T]>;
+  windows(groupSize: 2): AsyncIteratorPlus<[T, T]>;
+  windows(groupSize: 3): AsyncIteratorPlus<[T, T, T]>;
+  windows(groupSize: 4): AsyncIteratorPlus<[T, T, T, T]>;
+  windows(groupSize: 5): AsyncIteratorPlus<[T, T, T, T, T]>;
+  windows(groupSize: number): AsyncIteratorPlus<T[]>;
+  windows(groupSize: number): AsyncIteratorPlus<T[]> {
```
adghayes, line 304: Wouldn't it be nice if there was a better way to do this? I would guess other languages have this right? But it would be relying on a runtime-variable to set the compile-time length of an array, so I'm not sure if it is possible...

adghayes (approved):
LGTM

jonahkagan (approved):

libs/basics/src/iterators/async_iterator_plus.ts
eventualbuddha: Yeah, I tried to make this happen using type magic but TS doesn't like it because it has a hard time knowing how deep to go:
image

PR #3188, opened 2023-03-23
use updated image with rust installed

amcmanus:
## Overview

This PR updates our CircleCI config to use an updated set of docker images with Rust already installed to a pinned version. It also sets up the necessary ENV vars.

## Testing Plan

Run CircleCI with it. All tests should pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,13 +6,13 @@ orbs:
 executors:
   nodejs-browsers:
     docker:
-      - image: votingworks/cimg-debian11-browsers:2.0.0
+      - image: votingworks/cimg-debian11-browsers:2.0.2
         auth:
           username: $VX_DOCKER_USERNAME
           password: $VX_DOCKER_PASSWORD
   nodejs:
     docker:
-      - image: votingworks/cimg-debian11:2.0.0
+      - image: votingworks/cimg-debian11:2.0.2
         auth:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -986,6 +986,7 @@ commands:
           paths:
             - /root/.local/share/pnpm/store/v3
             - /root/.cache/Cypress
+            - /root/.cargo
```
eventualbuddha, line 989: This is where we've configured cargo to put its downloaded crates, but it doesn't do anything to cache build results. Is that your understanding?

.circleci/config.yml
```diff
@@ -6,13 +6,13 @@ orbs:
 executors:
   nodejs-browsers:
     docker:
-      - image: votingworks/cimg-debian11-browsers:2.0.0
+      - image: votingworks/cimg-debian11-browsers:2.0.2
```
eventualbuddha, line 9: Is there a commit/PR associated with this version bump?

eventualbuddha (approved):

.circleci/config.yml
amcmanus: Yes. We'll have to use something else for that. I've started looking.

.circleci/config.yml
amcmanus: https://github.com/votingworks/vxsuite-build-system/pull/19

PR #3190, opened 2023-03-23
VxAdmin CDF CVR Import Pt. 2

adghayes:
_review commit-by-commit_

## Overview

Completes #3175 by importing ballot images, layouts, and write-in records with CDF imports. Depends on #3180 and refactors some of the write-in image fetching on the backend, but otherwise does not normalize or change how we store images and layouts, simply converts at the boundary. Later, when CDF is the only format, we can normalize the ballot image storage.

## Testing Plan

Updated some automated testing and did a full pass of manual testing with the CDF export on VxScan to the CDF import in VxAdmin, with both VotingWorks ballots (HMPB & BMD) and NH ballots, comparing it with the current format. Everything checked out, other than the (mostly known) deficiencies of VotingWorks ballots with write-in adjudication. Specifically, those deficiencies are:
- On VotingWorks HMPB ballots, attempting to load a write-in image crashes the app. The app searches for a definition for the contest option within the layout and, finding none, throws an error.
- Assuming the above was fixed, I believe the layout on screen of the ballot, contest, and option would be messed up but I may be wrong here.
- On VotingWorks BMD ballots, no image is rendered because there is none. I assume we will want to treat BMD write-ins as "transcribed values" which would change the import process, but we have no handling for it now and it is a larger product discussion.

## Checklist


libs/types/src/cast_vote_record.ts
```diff
@@ -18,12 +18,12 @@ export interface CastVoteRecord
     | number
     | readonly number[]
     | BallotLocale
-    | SheetOf<InlineBallotImage>
-    | SheetOf<BallotPageLayout>
+    | SheetOf<InlineBallotImage | null>
+    | SheetOf<BallotPageLayout | null>
```
adghayes, line 22: CDF CVRs don't necessarily have images for both sheets.

adghayes: `undefined` is seralized to `null` in an array anyway, so it seemed less confusing to just type it as `null`.

apps/admin/backend/src/app.ts
```diff
@@ -548,92 +539,61 @@ function buildApi({
...
+      // Identify the page layout, contest layout, and associated image.
+      let pageLayout: Optional<BallotPageLayout>;
+      let contestLayout: Optional<BallotPageContestLayout>;
+      let ballotImage: InlineBallotImage | null | undefined;
+      for (const [index, layout] of cvr._layouts.entries()) {
+        if (layout === null) continue;
+
+        const foundContestLayout = layout.contests.find(
+          (contest) => contest.contestId === contestId
         );
```
adghayes, line 556: We're now using the changes in #3180 to find a layout without necessarily having all the layouts, and iterating through them.

apps/admin/backend/src/cvr_files.ts
```diff
@@ -592,6 +606,59 @@ export async function addCastVoteRecordReport({
         ).BatchLabel,
       });
 
+      // Add referenced images and layouts to the cast vote record
+      if (cvr.BallotImage) {
+        // Convention is that we always have two entries in the BallotImage
+        // array, allowing us to indicate front and back via array index.
+        assert(cvr.BallotImage.length === 2);
```
adghayes, line 613: As long as the write-in entries nested in the CDF CVR reference the correct image which is in the CVR's `BallotImage` field, I'm not sure if this convention is necessary in the long run. But for now, it is a useful simpllification.

adghayes:
@eventualbuddha I'm back on top of this one, it's ready for your review.

adghayes:
Per Slack thread, @jonahkagan is on point to review rather than @eventualbuddha.

apps/admin/backend/src/app.ts
```diff
@@ -558,92 +549,61 @@ function buildApi({
...
+      if (!ballotImage) {
+        throw new Error('no ballot image associated with the ballot layout');
       }
-      const contestLayout =
-        layouts[currentLayoutOptionIdx]?.contests[contestIdx];
 
-      // Options are laid out from the bottom up, so we reverse write-ins to get the correct bounds
-      const writeInOptions = contestLayout?.options
+      // Identify the write-in option layout. Options are laid out from the
+      // bottom up, so we reverse write-ins to get the correct bounds.
```
jonahkagan, line 582: Is this true? Are the write-ins in the layout in reverse order for some reason? That doesn't match with my intuition for the write-in option indices

libs/backend/src/load_ballot_image.ts
```diff
@@ -0,0 +1,25 @@
+import { loadImage, toDataUrl, toImageData } from '@votingworks/image-utils';
+
+/**
+ * Default factor by which to scale ballot images.
+ */
+export const CVR_BALLOT_IMAGE_SCALE = 0.5;
```
jonahkagan, line 6: If we shrink the images on export, does that mean we're also factoring this in to the layouts we use when cropping them in the adjudication flow? I'm assuming that must be what's going on but just want to check for my understanding

jonahkagan (approved):

apps/admin/backend/src/app.ts
adghayes: Sounds like you're more on top of this question than I am. I had no idea.

libs/backend/src/load_ballot_image.ts
adghayes: I do not know, I just kept the image scale the same so it wouldn't mess anything up. Eventually we won't scale them at all, once we revisit how we're storing this images in VxAdmin.

libs/backend/src/load_ballot_image.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Loads a ballot image from file into a data URL. We are currently shrinking
+ * the ballot images for transfer, so this will shrink the image as well. Uses
+ * a default scale factor of {@link CVR_BALLOT_IMAGE_SCALE}.
+ */
+export async function loadBallotImageBase64(
+  path: string,
+  factor: number = CVR_BALLOT_IMAGE_SCALE
```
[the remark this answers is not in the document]
adghayes: Good point, could strip it, but I'm hoping this function dies sometime soon so I'm just going to leave it in.

apps/admin/backend/src/app.cdf_cvr_files.test.ts
```diff
@@ -7,15 +7,20 @@ import { basename, join } from 'path';
...
   buildTestEnvironment,
   configureMachine,
   mockCastVoteRecordFileTree,
   mockElectionManagerAuth,
 } from '../test/app';
 import { modifyCastVoteRecordReport } from '../test/utils';
+import { getWriteInsFromCastVoteRecord } from './util/cvrs';
 
-jest.setTimeout(30_000);
+jest.setTimeout(60_000);
```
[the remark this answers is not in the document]
adghayes: This was a flaky case - it would sometimes hit above 30 seconds and sometimes not. It is usually not going close to 60 seconds. Still super slow though!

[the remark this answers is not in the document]
jonahkagan: Gotcha, that makes sense. Do we need to have 3000 records in the fixture? Or is there a more minimal one we could use?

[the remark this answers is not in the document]
adghayes: We don't need a fixture with 3000 records in _this_ test, but I think we do need a big fixture once we start doing tallying on the backend. That way we can have a report with the full mix of context options (whatever we decide that full mix to be) and then some. For now, I lean toward keeping it as is since it will be compatible with tally testing.

I just ran the tests locally and the times don't look too bad. It's only the big integration test that takes a really long time, and I think that's OK.

```  
  ✓ happy path - mock election flow (29026 ms)
  ✓ adding a duplicate file returns OK to client but logs an error (10497 ms)
  ✓ handles file with previously added entries by adding only the new entries (2105 ms)
  ✓ error if path to report is not valid (15 ms)
  ✓ error if report has invalid directory structure (26 ms)
  ✓ error if report metadata is not parseable (7869 ms)
  ✓ error if adding test report while in official mode (2035 ms)
  ✓ error if adding official report while in test mode (10138 ms)
  ✓ error if a cast vote record not parseable (1033 ms)
  ✓ error if a cast vote record is somehow invalid (1007 ms)
  ✓ error if cast vote records from different files share same ballot id but have different data (11039 ms)
  ✓ error if a layout is invalid (1086 ms)
```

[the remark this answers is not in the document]
jonahkagan: I actually had noticed the slow happy path test when I was working on something that touched it. That's more why I called it out now. I do think 30s is on the long side for a test case, but it's not a blocking concern

PR #3191, opened 2023-03-24
Unconfigure Screen Re-Render Cycle Fix

adghayes:
## Overview

I noticed when working in VxAdmin that the `UnconfiguredScreen` was rendering around 1000 times / second. After some debugging, I found there was re-rendering loop:

1. `UnconfiguredScreen` renders
2. `configureMutation` changes - @jonahkagan, every time we use `useMutation` it's a distinct entity because the `useMutation` function creates and returns a new function. Is this desirable behavior? Could it be memoized?
3. `configureMutateAsync` callback changes
4. `getOutputFile` callback changes
5. `updateStatus` callback changes
6. `useEffect` runs
7. `updateStatus` called
8. `setInputConversionFiles` called
9. `inputConversionFiles` state changes
10. `UnconfiguredScreen` renders
11. ...

I must have introduced this in #3052.

I'm not sure why, but this re-rendering loop was preventing me from accessing the NH File Conversion menu. I would click `Convert from NH File` and the app would freeze, not crash. 

I have a small fix in this PR to limit the cycle and only call `setInputConversionFiles` when necessary. It's not an ideal solution but it is a fix.

The bigger issue seems to be how we're trying to keep the UI state in sync with the converter service state, and doing so outside of `grout` so they don't play nicely. Ideally all the conversion stuff would be moved to the backend but, with much of it intended to leave the product, that seems not worth it at the moment.

## Demo Video or Screenshot

console logs before:
Screen Shot 2023-03-24 at 1 09 01 PM

console logs after:
Screen Shot 2023-03-24 at 1 08 27 PM

### DIFF
--- a/apps/admin/frontend/src/screens/unconfigured_screen.tsx
+++ b/apps/admin/frontend/src/screens/unconfigured_screen.tsx
@@ -344,7 +344,7 @@ export function UnconfiguredScreen(): JSX.Element {
           <HorizontalRule />
           <p>
             <Button small onPress={resetUploadFilesAndGoBack}>
-              back
+              Back
             </Button>
           </p>
         </Prose>
--- a/apps/admin/frontend/src/screens/unconfigured_screen.tsx
+++ b/apps/admin/frontend/src/screens/unconfigured_screen.tsx
@@ -242,11 +242,25 @@ export function UnconfiguredScreen(): JSX.Element {
         return;
       }
 
-      setInputConversionFiles(files.inputFiles);
+      if (
+        inputConversionFiles.length !== files.inputFiles.length ||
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
Thanks for detailing the problem out! Still thinking through what a long-term solution might be, but this of course looks good in the interim

jonahkagan:
Nice catch!

I found a discussion about this in the react-query issues: https://github.com/TanStack/query/issues/1858. Sounds like `mutateAsync` itself is stable, so we could pull that out from the `useMutation` object to fix the issue here.

Another solution would be to put `updateStatus` into a `useQuery` hook and kill the `useEffect`.

PR #3192, opened 2023-03-24
Implement card lockout after failed PIN entry

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2527

This PR implements card lockout after failed PIN entry. Lockout will kick in after N incorrect PIN attempts, with a starting lockout duration of D seconds that will double after every subsequent failure. I've started with defaults of N = 5 and D = 15 though we can of course adjust. These settings will be configurable by sys admins through the system settings file that Kevin is working on. Though we think that few election officials will ever modify these, VVSG2 requires them to be configurable. Note that before a machine has been configured, the defaults will be used.

Of note from an implementation perspective is the fact that the number of incorrect PIN attempts is stored in a secure location on the card, rather than the machine. This means that someone can't bypass a lockout by trying a card on a second machine. The second machine will display the same lockout time as the first. Relatedly, because we're storing no lockout data on the machine, if someone removes a card mid-lockout and re-inserts it, the lockout timer will start over. All good things from a security and architecture simplicity perspective. For non-malicious users, our hope is that the default 5 attempts without lockout will leave plenty of wiggle room for mistakes.

There's still a hard cap of 15 incorrect PIN attempts on a Java Card after which a card will need to be reprogrammed. (The 15 comes from the fact that 15 is the largest number you can represent with 4 bits, which is how the relevant count is stored on the card - it's all bits at the end of the day!) I haven't implemented any special handling for that case since the lockout time will be so high by then anyway, I have a hard time imagining anyone waiting for it to complete (>2 hours with the current defaults). I'll open an issue to document the edge case, though.

## Video

_N = 3 incorrect PIN attempts without lockout, D = 5 seconds_

https://user-images.githubusercontent.com/12616928/227623040-3f94d4ed-8ad2-4e70-9aee-cbd792267833.mov

## Screenshots

_1-minute+ lockout_

lockout-1

_1-hour+ lockout_

lockout-2

_Absurd lockout_

lockout-3

## Testing

- [x] Updated various layers of tests, maintaining 100% code coverage in `libs/auth` and `libs/ui`
- [x] Tested lockout manually on a machine that uses dipped smart card auth
- [x] Tested lockout manually on a machine that uses inserted smart card auth

## Checklist


libs/auth/src/java_card.ts
```diff
@@ -440,7 +459,10 @@ export class JavaCard implements Card {
       );
     }
 
-    return cardDetails;
+    const numIncorrectPinAttempts = cardDoesNotHavePin
+      ? undefined
+      : (await this.getNumIncorrectPinAttempts()) || undefined;
```
arsalansufi, line 465: Rather than using a mix of `0` and `undefined` for the zero-case, I'm consistently using `undefined`.

I initially tried reserving `undefined` for cards without PINs and always defining a number, whether 0, for cards with PINs, but that added complexity to the types/code for little gain.

libs/auth/src/lockout.ts
```diff
@@ -0,0 +1,61 @@
+import { Optional } from '@votingworks/basics';
+
+const DEFAULT_NUM_INCORRECT_PIN_ATTEMPTS_ALLOWED_BEFORE_CARD_LOCKOUT = 5;
+const DEFAULT_STARTING_CARD_LOCKOUT_DURATION_SECONDS = 15;
```
arsalansufi, line 4: @benadida Any thoughts on these defaults? I know we've already discussed the number of incorrect PIN attempts with lockout, but we hadn't yet discussed the starting lockout duration (that will of course double with every subsequent incorrect PIN attempt)

libs/types/src/auth/dipped_smart_card_auth.ts
```diff
@@ -21,7 +21,10 @@ export interface CheckingPin {
   readonly status: 'checking_pin';
   readonly user: SystemAdministratorUser | ElectionManagerUser;
   readonly error?: true;
-  readonly wrongPinEnteredAt?: Date;
+  /** A Unix timestamp in milliseconds for easy serialization */
+  readonly lockedOutUntil?: number;
```
arsalansufi, line 25: I encountered some interesting serialization bugs when I was using Dates. `JSON.stringify` automatically converts Dates to strings. So when I was using a Date type here, the type was inaccurate on the frontend, which was actually working with a string.

I looked into updating Grout's serialization to handle Dates, but Grout's serialization special-casing works best for types that can't be serialized at all. Dates actually are serialized by `JSON.stringify`, so I'd have to do some guess work to gauge whether a value was originally a Date before it became a string but then run the risk of accidentally converting some values that were meant to be strings into Dates. Ultimately decided that finding a general solution was out of scope, so I switched to Unix timestamps in milliseconds instead. Will open an issue for this! cc @jonahkagan

libs/auth/src/lockout.ts
benadida: lgtm

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -434,11 +439,16 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
                     user.role === 'system_administrator' ||
                       user.role === 'election_manager'
                   );
-                  return isFeatureFlagEnabled(
+                  const skipPinEntry = isFeatureFlagEnabled(
                     BooleanEnvironmentVariableName.SKIP_PIN_ENTRY
```
adghayes, line 443: my feature flag baby is still alive, what great news!

libs/auth/src/inserted_smart_card_auth.ts
```diff
@@ -409,7 +418,11 @@ export class InsertedSmartCardAuth implements InsertedSmartCardAuthApi {
             return {
               ...currentAuthStatus,
               error: undefined,
-              wrongPinEnteredAt: new Date(),
+              lockedOutUntil: computeCardLockoutEndTime(
+                machineState,
+                action.checkPinResponse.numIncorrectPinAttempts
+              )?.getTime(),
+              wrongPinEnteredAt: new Date().getTime(),
```
adghayes, line 425: maybe this will become apparent as I get to later commits, but noting that I'm not sure why we need `wrongPinEnteredAt` at all

adghayes (approved):
looks really good!

libs/auth/src/inserted_smart_card_auth.ts
adghayes: it looks like you're essentially just using this as a boolean flag in the frontend to tell whether the PIN was _just_ incorrectly entered or if it's a lockout attempt from possibly another machine. that works

libs/auth/src/dipped_smart_card_auth.ts
arsalansufi: It's made it through many a refactor haha

libs/auth/src/inserted_smart_card_auth.ts
arsalansufi: That's right - `wrongPinEnteredAt` actually predates this PR and is more fundamentally used to indicate whether an incorrect PIN has been entered to the frontend.

Code from before this PR: https://github.com/votingworks/vxsuite/blob/fa6a1566ca2d50c4eebe08f70fa46344565042e2/libs/ui/src/unlock_machine_screen.tsx#L81-L83

In addition to the reason you lay out, we also couldn't just rely on `lockedOutUntil` since it's `undefined` for the first few incorrect PIN attempts. We need something else to capture whether an incorrect PIN was entered.

libs/types/src/auth/dipped_smart_card_auth.ts
jonahkagan: Did a little digging on this because I was surprised. It turns out that the `JSON.stringify` replacer function (which is what Grout uses for tagging values during serialization) is called with a transformed value, not the original value. However, you can still access the original value using `this[key]`. Wild!

```ts
> JSON.stringify(new Date(), function (k, v) { console.log(typeof v, typeof this[k]); return this[k] instanceof Date ? 1 : v})
string object
'1'
```

I think this should allow us to catch and handle dates.

Alternatively we could just write our own recursive tagger and then call `JSON.stringify` on the resulting transformed value. I considered this originally but decided to try out the built-in replacer functionality to see how it worked.

I do think it's worth fixing because this is a potential footgun for anyone who tries to use a Date.

jonahkagan: https://github.com/votingworks/vxsuite/pull/3239

arsalansufi: Thanks for digging into this!

PR #3204, opened 2023-04-04
[VxScan] load system settings from ballot package

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3082

Recommend reviewing commit-by-commit.

* VxScan now reads system settings from the ballot package, writes them to db, and returns them in `getConfig()`
* Backwards compatibility: if the ballot package contains no `systemSettings.json`, we load the default system settings defined in code. We can probably remove this once all machines read system settings
* Some, but not all, fixtures have been updated to include `systemSettings.json` (just enough to get tests passing)

## Demo Video or Screenshot
No user facing changes

## Testing Plan
- added automated tests for writing, reading, and backwards compatibility

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (will do this when ballot package logic is separated out)


libs/utils/tsconfig.build.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["src", "src/data/*.json"],
+  "include": ["src", "src/data/*.json", "src/data/*.zip"],
```
kshen0, line 3: These changes were needed to get libs/utils/src/data/ballot-package-state-of-hamilton.zip to build. Tests were failing because that zip file had no `systemSettings.json` file. Since I later introduced a fallback to load `DEFAULT_SYSTEM_SETTINGS`, the update to `ballot-package-state-of-hamilton.zip` isn't strictly necessary, but is good hygiene to keep.

arsalansufi (approved):
LGTM!

PR #3209, opened 2023-04-04
feat(basics): add `#flatMap`

eventualbuddha:
## Overview
Also improves support for the async iterable methods to take either async or sync iterables.

## Demo Video or Screenshot
```ts
/// BEFORE
const loadedLayouts: BallotPageLayoutWithImage[] = [];

for (const [pdf, layouts] of templates) {
  for await (const { page, pageNumber } of pdfToImages(pdf, { scale: 2 })) {
    const ballotPageLayout = layouts[pageNumber - 1];
    loadedLayouts.push({
      ballotPageLayout,
      imageData: page,
    });
  }
}

this.cachedLayouts = loadedLayouts;
return loadedLayouts;

/// AFTER
this.cachedLayouts = await iter(templates)
  .async()
  .flatMap(([pdf, layouts]) =>
    iter(pdfToImages(pdf, { scale: 2 }))
      .zip(layouts)
      .map(([{ page: imageData }, ballotPageLayout]) => ({
        ballotPageLayout,
        imageData,
      }))
  )
  .toArray();
return this.cachedLayouts;
```

## Checklist

### DIFF
--- a/libs/basics/src/iterators/async_iterator_plus.test.ts
+++ b/libs/basics/src/iterators/async_iterator_plus.test.ts
@@ -81,12 +81,12 @@ test('chain', async () => {
 });
 
 test('zip', async () => {
-  expect(await iter([]).async().zip(iter([]).async()).toArray()).toEqual([]);
+  expect(await iter([]).async().zip(iter([])).toArray()).toEqual([]);
   expect(await iter([1]).async().zip().toArray()).toEqual([[1]]);
   expect(
     await iter([1])
       .async()
-      .zip(iter([2]).async())
+      .zip(iter([2]))
       .toArray()
   ).toEqual([[1, 2]]);
   expect(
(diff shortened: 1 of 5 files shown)


libs/basics/src/iterators/async_iterator_plus.test.ts
```diff
@@ -308,6 +308,30 @@ test('first', async () => {
...
+  expect(
+    await iter([])
+      .async()
+      .flatMap(() => [])
+      .toArray()
+  ).toEqual([]);
+  expect(
+    await iter([1, 2, 3])
+      .async()
+      .flatMap((a) => [a, a])
```
arsalansufi, line 321: nit: Maybe a slight preference for `item` or `value` instead of `a`

arsalansufi (approved):
LGTM!

I think I actually find the before code in the PR description easier to grok than the after code, but also have no issue with the after. Does the after code improve parallelizability, e.g. is it more like a `Promise.all`?

eventualbuddha:
> I think I actually find the before code in the PR description easier to grok than the after code, but also have no issue with the after. Does the after code improve parallelizability, e.g. is it more like a `Promise.all`?

This is an interesting point. I agree that the "after" code is not necessarily easier to read. I think with more familiarity with these APIs it will seem less weird, but I'm sympathetic to that view. As written it won't parallelize operations, though that's something it could in theory do a la [`rayon`'s `par_iter`](https://rust-lang-nursery.github.io/rust-cookbook/concurrency/parallel.html).

There is actually a functional difference between the "before" and "after" code because of the `zip` call, which not only prevents the need to index into a separate array but also ensures that the lengths are the same. This is a benefit, but there's also a cost in terms of being less JS-idiomatic and a bit more difficult to debug when things go wrong. If anyone has suggestions for making this easier to work with and debug I'd be happy to hear them.

arsalansufi:
> because of the zip call, which not only prevents the need to index into a separate array but also ensures that the lengths are the same

Thanks for that context! That does seem like an important benefit to the after code

Issue #3059, opened 2023-03-01
VxScan: show ballot count on all screens


mattroe:
We currently show the ballot count on some voter and poll-worker screens, but not all. The lack of a ballot count shown in some scenarios leads to scenarios where it's not clear if a ballot was counted or not. We should show the ballot count at all times for clarification as this is the most important piece of data on the scanner.

PR #3216, opened 2023-04-05
Vxscan vvsg ballot counter

kofi-q:
## Overview

Related to: https://github.com/votingworks/vxsuite/issues/3185
Closes: https://github.com/votingworks/vxsuite/issues/3059

Relevant changes in first commit (4b182f0455dce82d1575d9076784011e9bfd0175)
Second commit has test snapshot updates for VxMark (can't figure out why those changed).

- Added `BigMetric` component in libs/ui (builds on `LabelledText`) to use for the `ScannedBallotCount` component in VxScan
- Updated the base VxScan layout to persist the ballot count across all screens, per https://github.com/votingworks/vxsuite/issues/3059

## Demo Video or Screenshot

https://user-images.githubusercontent.com/264902/230153613-86eeeb00-668d-4369-832d-5deea89e2f1c.mov

### In context in VxScan:
Screenshot 2023-04-04 at 17 22 10

## Testing Plan
- Added unit test for `BigMetric`
- Updated tests that indirectly render the base layout component to mock out the API call for scanner status

## Checklist

### DIFF
--- a/libs/ui/src/big_metric.test.tsx
+++ b/libs/ui/src/big_metric.test.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+
+import { render, screen } from '../test/react_testing_library';
+import { BigMetric } from './big_metric';
+
+test('renders label and formatted value', () => {
+  render(<BigMetric label="Number of Machines" value={2} />);
+  render(<BigMetric label="Ballots Scanned" value={4096} />);
+
+  // Verify labels are rendered:
+  screen.getByText('Number of Machines');
+  screen.getByText('Ballots Scanned');
+
+  // Verify counts are rendered as formatted numbers with accessible labels:
+  expect(screen.getByText('2')).toHaveAccessibleName('Number of Machines: 2');
+  expect(screen.getByText('4,096')).toHaveAccessibleName(
+    'Ballots Scanned: 4,096'
+  );
+});
+
+// To pass satisfy coverage requirements:
+test('renders in both "legacy" and VVSG size modes', () => {
+  render(<BigMetric label="Number of Machines" value={2} />, {
+    vxTheme: { sizeMode: 'legacy' },
+  });
+
+  render(<BigMetric label="Number of Machines" value={2} />, {
+    vxTheme: { sizeMode: 'm' },
+  });
+});
--- a/libs/ui/src/big_metric.test.tsx
+++ b/libs/ui/src/big_metric.test.tsx
@@ -18,7 +18,7 @@ test('renders label and formatted value', () => {
   );
 });
 
-// To pass satisfy coverage requirements:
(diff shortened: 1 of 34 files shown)


libs/ui/src/big_metric.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const formattedValue = format.count(value);
+  const formattedLabel = `${label}: ${formattedValue}`;
+
+  return (
+    <StyledContainer>
+      <LabelledText aria-hidden label={label}>
+        <StyledValue
+          aria-label={formattedLabel}
+          data-testid={valueElementTestId}
```
eventualbuddha, line 50: Is there a way to make `<BigMetric data-testid="…"` do what we want? Or is this the better way to do this?

libs/ui/src/big_metric.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  screen.getByText('Ballots Scanned');
+
+  // Verify counts are rendered as formatted numbers with accessible labels:
+  expect(screen.getByText('2')).toHaveAccessibleName('Number of Machines: 2');
+  expect(screen.getByText('4,096')).toHaveAccessibleName(
+    'Ballots Scanned: 4,096'
+  );
+});
+
+// To pass satisfy coverage requirements:
```
eventualbuddha, line 21:
```suggestion
// To satisfy coverage requirements:
```

libs/ui/src/big_metric.tsx
kofi-q: Yeah, I believe that would work - I was avoiding having to update the test assertions to include the label text as well, but I'm down to do that folks feel strongly

libs/ui/src/big_metric.test.tsx
kofi-q: thanks!

libs/ui/src/big_metric.tsx
```diff
@@ -0,0 +1,57 @@
...
+ */
+export function BigMetric(props: BigMetricProps): JSX.Element {
+  const { label, value, valueElementTestId } = props;
+
+  const formattedValue = format.count(value);
+  const formattedLabel = `${label}: ${formattedValue}`;
+
+  return (
+    <StyledContainer>
+      <LabelledText aria-hidden label={label}>
```
kshen0, line 47: Why do we want this to be `aria-hidden`? Based on the `aria-label` on the child element, it seems like we want this tree to be rendered on accessibility devices.

kofi-q: Oh, good catch! I meant to put that on the label, not the whole component - will fix that up

kshen0 (approved):
LGTM!

Issue #2526, opened 2022-09-02
Implement user session time limits


mattroe:
## Product Requirements
* A System Administrator must be able to set a configurable value for the maximum time limit for a user's session & inactivity
* This limit can be the same for all users and roles.
* We should implement the NIST recommended max values as default and allow configuration on a 1-12(?) hours for session time and 5-30(?) minutes for inactivity.
* Once the time limit is set, the application must log the user out and require re-authentication via smart card.

## VVSG Requirements
11.5-A: The voting system must enable an administrator the ability to do the following: set the maximum time limit for a user’s session, set the maximum time limit for user inactivity. NIST SP 800-63B [NIST17d] recommends a max session time of 12 hours regardless of inactivity and a max inactivity time of 30 minutes. Elections consist of temporary employees and user access may only be required during an election. A user’s access may expire and terminate automatically at the end of an election.
11.5-B: The voting system must require reauthentication of an authorized user after the administrator-specified time limit for the user’s session or for user inactivity. After authentication, a user’s access to a voting system will time-out after a specified period of time.  This will avoid unauthorized access to the voting system by unauthorized users. Once a user’s access has timed-out, the user will have to re-authenticate to continue using the voting system. For voters, session times are specified under requirement 7.2-O – Inactivity alerts. For more information, see NIST SP 800-63B [NIST17d].

PR #3217, opened 2023-04-05
Add inactive and overall session time limits

arsalansufi:
_Easiest reviewied by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2526

This PR implements the session time limits required by VVSG2, specifically an inactive/idle session time limit and an overall session time limit. Both will eventually be configurable using the system settings file that Kevin is working on, but the defaults are 30 minutes for the inactive session time limit and 12 hours for the overall session time limit.

For the inactive session time limit, after the limit is hit, the user is given 1 minute to wrap up what they're working on. For the overall session time limit, the user is given a warning 15 minutes before the limit hits. There's no option to extend your session; reauthentication is required, per VVSG2.

Note that these time limits only apply to user-authenticated screens. Screens like the VxScan insert-your-ballot screen are unaffected and can remain inactive/open.

## Demo Videos

_Inactive session time limit_

https://user-images.githubusercontent.com/12616928/230183545-1551f6c8-79ff-4193-b77d-cccfddfd5419.mov

-----

_Overall session time limit_

https://user-images.githubusercontent.com/12616928/230183550-18cb2e6c-7ac2-438e-af71-705c5f902f0b.mov

-----

_Timer running out_

https://user-images.githubusercontent.com/12616928/230183555-c723a668-1c0b-46ca-b1a0-57c5134e179f.mov

-----

_Edge case: Session time limit hit on remove-card screen_

https://user-images.githubusercontent.com/12616928/230183552-0fa75106-b6cb-442c-a634-4cfeda5bb2ae.mov

## Demo Screenshots

_Inactive session time limit_

inactive-session-time-limit

_Overall session time limit_

overall-session-time-limit

_VxAdmin nav timer_

vx-admin-nav-timer

_VxCentralScan nav timer_

vx-central-scan-nav-time

I didn't bother adding these nav timers to VxMark and VxScan since they don't have the same top nav as VxAdmin and VxCentralScan. Those machines are also much less likely to hit these time limits.

## Testing

- [x] Updated libs/auth tests, maintaining 100% code coverage
- [x] Updated existing app tests
- [x] Tested VxAdmin manually
- [x] Tested VxCentralScan manually
- [ ] Tested VxMark manually
- [ ] Tested VxScan manually

Punted on tests for the `SessionTimeLimitTracker` component since this PR is large enough and will add those in a follow-up.

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ I could update the logout log to indicate whether the cause of the logout was a session time limit, but this PR is large enough so I'm pushing that to a follow-up.


libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -226,6 +234,17 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
     await this.updateAuthStatus(machineState, { type: 'log_out' });
   }
 
+  async updateSessionExpiry(
```
arsalansufi, line 237: This method exists for two cases:

1. Bringing the session expiry up when the inactive session time limit is hit. I know I could just have the frontend hit the logout endpoint instead of modifying the auth status expiry, but updating the expiry makes accessing the new logout time easier across the UI, which further makes implementing the nav bar timer much easier.
2. Handling the case that the system clock is updated and the expiry needs to be adjusted accordingly. I haven't handled that in this PR but will in a follow-up. (Thanks to @eventualbuddha for reminding me about system clock updating.)

After I've incorporated 2, I may add some restrictions to how the session expiry can be updated for enhanced security such that, even if someone compromises the frontend, they can't extend a session forever.

libs/auth/src/inserted_smart_card_auth.ts
```diff
@@ -219,6 +227,21 @@ export class InsertedSmartCardAuth implements InsertedSmartCardAuthApi {
     });
   }
 
+  async logOut(machineState: InsertedSmartCardAuthMachineState): Promise<void> {
+    await this.updateAuthStatus(machineState, { type: 'log_out' });
+  }
```
arsalansufi, line 232: Why we need a log-out method for inserted smart card auth, where logout is typically accomplished by removing your card, is described in a comment later in this commit

libs/ui/src/session_time_limit_tracker.tsx
```diff
@@ -0,0 +1,206 @@
+/* istanbul ignore file */
```
arsalansufi, line 1: As I mentioned in the PR description, this PR is large enough as is so I'm going to add tests for this component in a follow-up

apps/mark/frontend/src/api.ts
```diff
@@ -122,6 +123,40 @@ export const checkPin = {
   },
 } as const;
 
+/* istanbul ignore next */
```
arsalansufi, line 126: More tests I'm punting to a follow-up. Using `istanbul ignore` comments in this commit so that we can still keep VxMark at 100% code coverage and avoid coverage regressions elsewhere

apps/admin/backend/src/app.test.ts
```diff
@@ -469,22 +469,33 @@ test('auth', async () => {
...
-  await apiClient.checkPin({ pin: '123456' });
-  await apiClient.logOut();
-  void (await apiClient.programCard({ userRole: 'system_administrator' }));
-  void (await apiClient.programCard({ userRole: 'election_manager' }));
-  void (await apiClient.unprogramCard());
-
   expect(auth.getAuthStatus).toHaveBeenCalledTimes(2);
   expect(auth.getAuthStatus).toHaveBeenNthCalledWith(2, { electionHash });
+
+  await apiClient.checkPin({ pin: '123456' });
```
arsalansufi, line 475: Reordered these tests to pair calls and assertions together

apps/admin/frontend/src/app.test.tsx
```diff
@@ -106,7 +108,6 @@ beforeEach(() => {
 
 afterEach(() => {
   delete window.kiosk;
-  MockDate.reset();
```
arsalansufi, line 109: This `reset` was calling tests to fail because it was triggering session time limits. I removed it because it isn't actually needed - we're already resetting the date in this file's `beforeEach`

apps/admin/frontend/src/app.test.tsx
```diff
@@ -292,7 +295,7 @@ test('L&A (logic and accuracy) flow', async () => {
       expect.anything()
     )
   );
-  jest.advanceTimersByTime(5000);
+  advanceTimers(5);
```
arsalansufi, line 298: Needed to switch from plain `jest.advanceTimersByTime` to our `advanceTimers` helper to prevent warnings about missing `act` wrappings:

```
Warning: An update to SessionTimeLimitTimer inside a test was not wrapped in act(...).
```

apps/mark/frontend/src/app_root.tsx
```diff
@@ -649,17 +650,23 @@ export function AppRoot({
...
-        unconfigureMachine={unconfigure}
-        isMachineConfigured={Boolean(optionalElectionDefinition)}
-        resetPollsToPaused={
-          pollsState === 'polls_closed_final'
-            ? makeAsync(resetPollsToPaused)
-            : undefined
-        }
-        usbDriveStatus={usbDrive.status}
-      />
+      <React.Fragment>
```
arsalansufi, line 653: This file is easier to review using https://github.com/votingworks/vxsuite/pull/3217/commits/8b215a6322844b05523f248dbfaca6442ce88bad?w=1

apps/central-scan/backend/src/central_scanner_app.test.ts
```diff
@@ -809,17 +809,28 @@ test('auth', async () => {
...
     { pin: '123456' }
   );
+
+  await apiClient.logOut();
   expect(auth.logOut).toHaveBeenCalledTimes(1);
   expect(auth.logOut).toHaveBeenNthCalledWith(1, { electionHash });
+
+  await apiClient.updateSessionExpiry({
+    sessionExpiresAt: new Date().getTime(),
+  });
```
kshen0, line 829: Is it expected that real user flows will lead to `updateSessionExpiry` being called by the app after `logOut`? Maybe I missed this in the code earlier but it feels unintuitive; after logging out how can there be a session expiry time (as there is no session)?

libs/ui/src/session_time_limit_tracker.tsx
```diff
@@ -0,0 +1,206 @@
...
+  authStatus,
+  logOut,
+  updateSessionExpiry,
+}: SessionTimeLimitTrackerHelperProps): JSX.Element | null {
+  // TODO: Pull values from system settings
+  const inactiveSessionTimeLimitMinutes =
+    DEFAULT_INACTIVE_SESSION_TIME_LIMIT_MINUTES;
+  const overallSessionTimeLimitHours = DEFAULT_OVERALL_SESSION_TIME_LIMIT_HOURS;
+
+  const now = useNow().toJSDate();
```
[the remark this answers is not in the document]
arsalansufi: Yes good question! I'm using `useNow` here because it returns an updated date every second, triggering a re-render every second and making sure we constantly check whether it's time to surface the prompt or not:

https://github.com/votingworks/vxsuite/blob/04631463a3c68e8796a24f4c1bda114e234042e2/libs/ui/src/hooks/use_now.ts#L11-L13

With a static `new Date()`, we'd potentially never surface the prompt (unless a re-render was triggered by some other external factor). I'd have to instead set up a `setInterval` or `setTimeout` myself to make that happen. With `useNow`, I can get that for free.

In the `useIdleTimer` `onIdle` function, I don't need something like `useNow` because it's an event handler. So whenever it's called, `new Date()` will be up to date.

apps/central-scan/backend/src/central_scanner_app.test.ts
arsalansufi: Ah good catch - I wasn't thinking too much about making this test realistic since the auth instance is mocked out and I'm just verifying that the API call results in the correct auth lib call with the correct input. (I'm relying on `libs/auth` tests for more complete tests of user flows.)

But let me adjust here and elsewhere to make this a little more representative!

libs/ui/src/session_time_limit_tracker.tsx
```diff
@@ -0,0 +1,206 @@
...
+                {pluralize('hours', overallSessionTimeLimitHours, true)}.
+                <br />
+                The machine will automatically lock in{' '}
+                <Timer countDownTo={new Date(authStatus.sessionExpiresAt)} />.
+              </p>
+            )}
+            <p>
+              Lock the machine now and reauthenticate
+              <br />
+              with your smart card to continue working.
```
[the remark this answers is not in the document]
arsalansufi: I was on the fence here and thought the text looked a little funny with the natural wrapping as applied by the modal width but also do like the idea of letting things wrap naturally. If it's alright with you, I'm going to leave this as is since I got an LGTM from Roe on the screenshots with this particular wrapping

apps/mark/backend/src/app.ts
```diff
@@ -55,6 +56,19 @@ function buildApi(auth: InsertedSmartCardAuthApi) {
       });
     },
 
+    logOut(input: { electionHash?: string }) {
+      return auth.logOut(constructAuthMachineState(input));
+    },
+
+    updateSessionExpiry(input: {
+      electionHash?: string;
```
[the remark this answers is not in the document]
arsalansufi: Ah, so it actually is used downstream.

The call to `constructAuthMachineState` uses it:

https://github.com/votingworks/vxsuite/blob/8b215a6322844b05523f248dbfaca6442ce88bad/apps/mark/backend/src/app.ts#L67

https://github.com/votingworks/vxsuite/blob/8b215a6322844b05523f248dbfaca6442ce88bad/apps/mark/backend/src/app.ts#L26-L41

And inside the auth lib, the first thing `updateSessionExpiry` does is refresh the auth status, which requires the machine state, election hash included:

https://github.com/votingworks/vxsuite/blob/8b215a6322844b05523f248dbfaca6442ce88bad/libs/auth/src/inserted_smart_card_auth.ts#L238

-----

VxMark is definitely a bit wonky in that it still stores election definition on the frontend, so election hash has to be provided by the frontend for auth calls. But as you set up a VxMark store, I can build off of that and remove this passing of election hash from frontend to backend.

apps/central-scan/backend/src/central_scanner_app.test.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/3217/commits/a5534082a3362fb73fb070e9a275941d2a103639

libs/ui/src/session_time_limit_tracker.tsx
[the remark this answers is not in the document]
kshen0: Thanks!

apps/mark/backend/src/app.ts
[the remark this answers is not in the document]
kshen0: Ah sorry didn't see that it was passed down through `input`!

apps/central-scan/backend/src/central_scanner_app.test.ts
kshen0: Hmm not sure of our testing philosophy - I tend to lean towards making tests as atomic as possible, but that may be the old Mocha way of thinking when you could more easily group like tests under a single `describe` block. Under that philosophy it might make sense to break this larger `auth` test out into smaller atomic tests, but it's definitely low priority since it's an existing pattern to have a single larger test. No strong opinion.

kshen0 (approved):

apps/central-scan/backend/src/central_scanner_app.test.ts
arsalansufi: That is a great point/suggestion! Updated the app-level auth tests to be more atomic for clarity: https://github.com/votingworks/vxsuite/pull/3217/commits/8f38166cc60cbfce3632edcee34392d9389b5365

PR #3218, opened 2023-04-05
add autoscan functionality to custom scanner demo program

carolinemodic:
## Overview
Adds the ability to go into autoscan mode for load testing in the custom scanner demo program, this will continually watch for paper scan it and eject it without interpretation and keep track of the number of ballots counted. 

cc @benadida this should unblock load testing for you if you run the bin/demo program from libs/custom-scanner here


libs/custom-scanner/src/utils/status_watcher.ts
```diff
@@ -0,0 +1,139 @@
...
+ *   if (statusResult.isErr()) {
+ *     console.error(statusResult.err());
+ *   } else {
+ *     const status = statusResult.ok();
+ *     console.log(status);
+ *   }
+ * }
+ * ```
+ */
+export function watchStatus(
```
carolinemodic, line 59: @eventualbuddha I copied this from a previous version of the code you had.

Issue #3110, opened 2023-03-09
Implement card cert expiries


arsalansufi:
Plan:

1. 100 years for card VotingWorks certs
2. 5 years for sys admin card VxAdmin certs
3. 6 months for election manager and poll worker card VxAdmin certs

https://github.com/votingworks/vxsuite/issues/3109 includes a potential optimization to 2.

arsalansufi:
Reminder to self to incorporate: https://github.com/votingworks/vxsuite/pull/3106#discussion_r1131653901

PR #3219, opened 2023-04-05
Implement card cert expiries

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3110

This PR implements card cert expiries, specifically:

- 100 years for VotingWorks-issued card certs 🧓
- 5 years for VxAdmin-issued sys admin card certs 🧒
- 6 months for VxAdmin-issued election card certs 👶

Certs generated by the `generate-dev-keys-and-certs` and `generate-test-keys-and-certs` scripts will continue to last for 100 years.

The PR also cleans up the openssl command that we're using for cert generation for added clarity.

## Testing

- [x] Updated unit tests
- [x] Inspected certs manually

## Checklist

### DIFF
--- a/libs/auth/src/openssl.ts
+++ b/libs/auth/src/openssl.ts
@@ -192,7 +192,7 @@ export async function verifySignature({
 export async function createCert({
   certSubject,
   certType = 'standard',
-  expiryInDays = 365,
+  expiryInDays,
   opensslConfig,
   publicKeyToSign,
   signingCertAuthorityCert,
@@ -201,7 +201,7 @@ export async function createCert({
 }: {
   certSubject: string;
   certType?: 'standard' | 'certAuthorityCert';
-  expiryInDays?: number;
+  expiryInDays: number;
   opensslConfig: FilePathOrBuffer;
   publicKeyToSign: FilePathOrBuffer;
   signingCertAuthorityCert: FilePathOrBuffer;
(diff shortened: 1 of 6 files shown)


libs/auth/src/openssl.ts
```diff
@@ -208,15 +208,23 @@ export async function createCert({
...
   const certSigningRequest = await openssl([
     'req',
     '-new',
     '-config',
     opensslConfig,
     '-key',
-    signingPrivateKey,
-    '-passin',
-    `pass:${signingPrivateKeyPassword}`,
+    throwawayPrivateKey,
```
arsalansufi, line 227: While the output of this function is still the same (given the `-force_pubkey` below), the previous code was quite misleading. We'd ideally be using the private key of the public key to sign to generate the CSR, but until we figure out how to do that with Java Cards, we're at least now being more explicit about the fact that we're using a throwaway private key to generate the CSR.

arsalansufi, line 227: While the output of this function is still the same (given the -force_pubkey below), the previous code was quite misleading. We'd ideally be using the private key of the public key to sign to generate the CSR, but until we figure out how to do that with Java Cards, we're at least now being more explicit about the fact that we're using a throwaway private key to generate the CSR.

benadida (approved):

PR #3225, opened 2023-04-06
refactor(central-scan): extract "Delete Batch" modal

eventualbuddha:
## Overview
I wanted to try out transitioning an API endpoint to use Grout and this seemed like a good one to illustrate both the actual Grout usage and how using `react-query` with a mutation can simplify our client code.

## Demo Video or Screenshot
n/a

## Testing Plan
New automated tests, light manual testing.

## Checklist

### DIFF
--- a/apps/central-scan/frontend/src/api.ts
+++ b/apps/central-scan/frontend/src/api.ts
@@ -78,8 +78,15 @@ export const logOut = {
 export const updateSessionExpiry = {
   useMutation() {
     const apiClient = useApiClient();
+    return useMutation(apiClient.updateSessionExpiry);
+  },
+} as const;
+
+export const deleteBatch = {
+  useMutation() {
+    const apiClient = useApiClient();
     const queryClient = useQueryClient();
-    return useMutation(apiClient.updateSessionExpiry, {
+    return useMutation(apiClient.deleteBatch, {
       async onSuccess() {
         // Because we poll auth status with high frequency, this invalidation isn't strictly
         // necessary
--- a/apps/central-scan/frontend/src/components/delete_batch_modal.test.tsx
+++ b/apps/central-scan/frontend/src/components/delete_batch_modal.test.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import userEvent from '@testing-library/user-event';
-import { suppressingConsoleOutput } from '@votingworks/test-utils';
 import { MockApiClient, createMockApiClient, provideApi } from '../../test/api';
 import { render, screen, waitFor } from '../../test/react_testing_library';
 import { DeleteBatchModal } from './delete_batch_modal';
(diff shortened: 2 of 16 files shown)


kofi-q (approved):

apps/central-scan/frontend/src/api.ts
```diff
@@ -74,3 +74,17 @@ export const logOut = {
...
+
+export const deleteBatch = {
+  useMutation() {
+    const apiClient = useApiClient();
+    const queryClient = useQueryClient();
+    return useMutation(apiClient.deleteBatch, {
+      async onSuccess() {
+        // Because we poll auth status with high frequency, this invalidation isn't strictly
+        // necessary
+        await queryClient.invalidateQueries(getAuthStatus.queryKey());
```
jonahkagan, line 86: Do we need to invalidate auth status after deleting a batch, or was this just accidentally left during copy/paste?

apps/central-scan/frontend/src/components/delete_batch_modal.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+afterEach(() => {
+  mockApiClient.assertComplete();
+});
+
+test('allows canceling', async () => {
+  const onClose = jest.fn();
+
+  render(
+    <QueryClientProvider client={new QueryClient()}>
```
jonahkagan, line 25: It's generally good to use the same query client options in tests as in the app, since they can influence the logic significantly.

apps/central-scan/frontend/src/components/delete_batch_modal.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <ApiClientContext.Provider value={mockApiClient}>
+        <DeleteBatchModal batchId="a" batchLabel="Batch 1" onClose={onClose} />
+      </ApiClientContext.Provider>
+    </QueryClientProvider>
```
jonahkagan, line 29: Here's an example of how I tried to abstract some of the boilerplate here: https://github.com/votingworks/vxsuite/blob/7a2879e5d28ec6ab43e1750362091c545b58ce78/apps/mark/frontend/test/helpers/mock_api_client.tsx#L150.

apps/central-scan/frontend/src/components/delete_batch_modal.tsx
```diff
@@ -0,0 +1,62 @@
...
+  onClose,
+}: Props): JSX.Element {
+  const deleteBatchMutation = api.deleteBatch.useMutation();
+
+  async function doDeleteBatch() {
+    try {
+      // await the mutation to ensure the modal is closed only after the mutation
+      // is complete and the batch is deleted.
+      await deleteBatchMutation.mutateAsync({ batchId });
+      onClose();
```
jonahkagan, line 26: The pattern I've been recommending is to do `mutate` with an `onSuccess` handler for this. Any reason you would prefer `mutateAsync` with an empty `catch`? I know it's a bit more readable to be able to use `await`, but I'm interested in having a consistent approach as much as possible to reduce cognitive overhead.

apps/central-scan/frontend/src/api.ts
eventualbuddha: I don't know. Probably not, given the action being performed. I'll assume not unless @arsalansufi tells me otherwise.

arsalansufi: Nope, no need to invalidate auth status here!

apps/central-scan/frontend/src/components/delete_batch_modal.tsx
eventualbuddha: I had played around with a couple different options and never landed on that for some reason, but it does seem better. Going with the callback.

Issue #3169, opened 2023-03-21
Prepare for Java Card use during NH pilot


mattroe:

arsalansufi:
Relevant spec: https://docs.google.com/document/d/169VWj8W9AX2ir6T1fb9Tr3aEouJO9Ptt2MC8dzvN7FQ/edit?usp=sharing

arsalansufi:
Relevant state of the world for NH: [NH Pilot Build Instructions](https://docs.google.com/document/d/1K_5XY94qi-zBljIJQhSFOIsO-hpmJIzWzGiNrB032sI/edit?usp=sharing)

PR #3226, opened 2023-04-06
Make various updates to Java Card configuration in prep for production flows

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3169

A very first step as I work on production machine/card configuration flows, this PR makes various updates to Java Card configuration. It:

- Removes the openssl config file as a config param, since it's the same across all envs
- Adds dev vs. prod branching
- Uses `__dirname` to avoid having to pass `pathToAuthLibRoot`

## Testing

- [x] Updated unit tests
- [x] Manually tested that dev still works
- [x] Manually tested that prod will hypothetically work (by simulating aspects of prod in my dev environment)

## Checklist


libs/auth/src/java_card_config.ts
```diff
@@ -15,36 +18,62 @@ export interface JavaCardConfig {
...
+      __dirname,
+      '../certs/dev/vx-admin-cert-authority-cert.pem'
+    ),
     vxAdminPrivateKeyPassword: DEV_PRIVATE_KEY_PASSWORD,
-    vxAdminPrivateKeyPath: `${pathToAuthLibRoot}/certs/dev/vx-admin-private-key.pem`,
+    vxAdminPrivateKeyPath: path.join(
+      __dirname,
+      '../certs/dev/vx-admin-private-key.pem'
+    ),
   };
```
arsalansufi, line 60: As you can sort of see here, my plan for prod is to have `vxsuite-complete-system` machine configuration scripts create and store the VxAdmin cert authority cert at `/vx/config/vx-admin-cert-authority-cert.pem`. Until we're using the TPM for signing, the scripts can also store the VxAdmin private key at `/vx/config/vx-admin-private-key.pem` and provide the private key password through the same env var mechanism that we use to provide [machine ID](https://github.com/votingworks/vxsuite-complete-system/blob/7226af99dfc1a552036f35e2e769fd95786abb93/config/read-vx-machine-config.sh#L36-L39).

Dev (including VxDev) will continue to use the dev keys and certs committed to the `certs/dev` directory.

libs/auth/certs/prod/vx-cert-authority-cert.pem
```diff
@@ -0,0 +1,15 @@
+# A placeholder (and copy of certs/dev/vx-cert-authority-cert.pem)
+# until we have an official cert
```
arsalansufi, line 2: TIL re PEM file comments - openssl knows to ignore anything outside the `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----`

libs/utils/src/environment_variable.ts
```diff
@@ -124,7 +124,7 @@ export function getBooleanEnvVarConfig(
     case BooleanEnvironmentVariableName.ENABLE_JAVA_CARDS:
       return {
         name,
-        allowInProduction: false,
+        allowInProduction: true,
```
arsalansufi, line 127: NH pilot, here we come!

libs/auth/src/java_card_config.ts
```diff
@@ -15,36 +18,62 @@ export interface JavaCardConfig {
...
 /**
  * The password for all dev private keys
  */
 export const DEV_PRIVATE_KEY_PASSWORD = '1234';
 
-function constructDevCardProgrammingConfig(
-  pathToAuthLibRoot: string
-): JavaCardConfig['cardProgrammingConfig'] {
+function shouldUseProdCerts(): boolean {
+  return process.env.NODE_ENV === 'production' && !isVxDev();
```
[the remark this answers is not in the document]
arsalansufi: Come next week, I'll either add a feature flag to enable usage of dev certs in prod or will have enough of the production flow ready that this won't be necessary

libs/auth/src/java_card_config.ts
```diff
@@ -15,36 +18,62 @@ export interface JavaCardConfig {
   vxCertAuthorityCertPath: string;
 }
 
+/**
+ * Should be kept in sync with vxsuite-complete-system scripts
+ */
+const PROD_CONFIG_DIRECTORY = '/vx/config';
```
benadida, line 24: I'd love for this to be an environment variable for configuration. Not required in this PR, but just food for thought.

benadida (approved):

libs/auth/src/java_card_config.ts
arsalansufi: Ah nice, I like that - will make it such that we don't have to keep vxsuite in sync with vxsuite-complete-system. It'll just be vxsuite-complete-system's responsibility to be internally consistent

arsalansufi: https://github.com/votingworks/vxsuite/pull/3226/commits/55aaca6ebb915c34f5944cb9fd8c28b0a6e48d6a

PR #3232, opened 2023-04-07
build(rust): cargo workspaces

eventualbuddha:
## Overview

Cargo workspaces are similar to pnpm workspaces in that you can have multiple packages (crates) that depend on each other. We'll likely want to split out the election types so they can be reused across multiple crates, and this is the first step toward that.

https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist


arsalansufi (approved):

libs/ballot-interpreter-nh-next/src/interpret.rs
```diff
@@ -349,20 +349,15 @@ pub fn interpret_ballot_card(side_a_path: &Path, side_b_path: &Path, options: &O
...
 
     #[test]
     fn test_interpret_ballot_card() {
-        let fixture_path = Path::new(file!())
-            .parent()
-            .unwrap()
-            .parent()
-            .unwrap()
-            .join("test/fixtures/ashland");
+        let fixture_path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("test/fixtures/ashland");
```
eventualbuddha, line 360: This change was because the `file!()` macro gives a path relative to the workspace root when using workspaces: https://stackoverflow.com/questions/30003921/how-can-i-locate-resources-for-testing-with-cargo.

PR #3237, opened 2023-04-07
ignore hidden files with "." prefix in ballot package dir

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3236
When loading ballot packages from a USB, the ballot package IO library will load the newest file in the `ballot-packages` directory. If the newest file is a hidden `._foo.zip` file [left by OS X](https://apple.stackexchange.com/questions/14980/why-are-dot-underscore-files-created-and-how-can-i-avoid-them) the ballot package library will try and fail to parse it as a zip.

This PR makes it so we ignore `._`-prefixed files when reading `ballot-packages`. There's an argument to be made to ignore all hidden files prefixed by `.`, but since we also check for the `.zip` suffix, I don't know of any other cases where an OS would accidentally make `.foo.zip` files.

## Demo Video or Screenshot
Old

https://user-images.githubusercontent.com/1060688/230693060-59ed6677-4b04-439e-8e78-eb0dd1b39808.mov

New

https://user-images.githubusercontent.com/1060688/230693109-5796db99-b2ac-458e-be8d-40b1e68db6fc.mov

## Testing Plan
- added tests for `._` edge case

## Checklist

### DIFF
--- a/libs/backend/src/ballot_package/ballot_package_io.test.ts
+++ b/libs/backend/src/ballot_package/ballot_package_io.test.ts
@@ -16,7 +16,7 @@ import {
 import { assert } from '@votingworks/basics';
 import { safeParseSystemSettings } from '@votingworks/utils';
 import { join } from 'path';
-import * as fsSync from 'fs';
+import * as fs from 'fs';
 import { Buffer } from 'buffer';
 import { createBallotPackageWithoutTemplates } from './test_utils';
 import { readBallotPackageFromUsb } from './ballot_package_io';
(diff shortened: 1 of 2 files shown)


libs/backend/src/ballot_package/ballot_package_io.ts
```diff
@@ -51,7 +51,9 @@ export async function readBallotPackageFromUsb(
 
   const files = await fs.readdir(directoryPath, { withFileTypes: true });
   const ballotPackageFiles = files.filter(
-    (file) => file.isFile() && file.name.endsWith('.zip')
+    // OsX hidden files prefixed with `._` are not valid zip files and will break zip parsing
```
eventualbuddha, line 54: macOS is the current name.

libs/backend/src/ballot_package/ballot_package_io.ts
```diff
@@ -51,7 +51,9 @@ export async function readBallotPackageFromUsb(
...
+    (file) =>
+      file.isFile() && !file.name.startsWith('._') && file.name.endsWith('.zip')
```
eventualbuddha, line 56: I think it'd be reasonable to exclude anything starting with `.` since they're generally considered "hidden".

libs/backend/src/ballot_package/ballot_package_io.test.ts
```diff
@@ -15,9 +15,37 @@ import {
...
+  const files = fsSync.readdirSync(dirPath);
+  const filesWithStats = files.map((file) => ({
+    file,
+    ...fsSync.statSync(join(dirPath, file)),
+  }));
+  assert(filesWithStats[0] !== undefined && filesWithStats[1] !== undefined);
+  expect(filesWithStats[0].file).toContain(newerFileName);
+  expect(filesWithStats[1].file).toContain(olderFileName);
+  // expect(filesWithStats[0].file).toContain('newer-ballot-package.zip');
+  // expect(filesWithStats[1].file).toContain('older-ballot-package.zip');
```
eventualbuddha, line 44: Commented out?

libs/backend/src/ballot_package/ballot_package_io.test.ts
```diff
@@ -15,9 +15,37 @@ import {
 } from '@votingworks/fixtures';
 import { assert } from '@votingworks/basics';
 import { safeParseSystemSettings } from '@votingworks/utils';
+import { join } from 'path';
+import * as fsSync from 'fs';
```
eventualbuddha, line 19:
```suggestion
import * as fs from 'fs';
```

eventualbuddha (approved):

PR #3238, opened 2023-04-10
For NH pilot: Support using dev certs in prod for Java Card auth

arsalansufi:
This PR updates VxSuite to support using dev certs in prod for Java Card auth, through a `USE_DEV_CERTS_IN_PROD` feature flag.

We very likely won't need this for the NH pilot as I should be able to finish more of the production flows this week before the NH pilot code freeze, but I'm opening this just in case. Will also come in handy for QA ahead of the freeze. Note that `USE_DEV_CERTS_IN_PROD` is only relevant if the `ENABLE_JAVA_CARDS` feature flag is also on.

### DIFF
--- a/libs/auth/src/env.d.ts
+++ b/libs/auth/src/env.d.ts
@@ -1,7 +1,7 @@
 declare namespace NodeJS {
   export interface ProcessEnv {
-    CONFIG_DIRECTORY?: string;
     NODE_ENV: 'development' | 'production' | 'test';
     VX_ADMIN_PRIVATE_KEY_PASSWORD?: string;
+    VX_CONFIG_ROOT?: string;
   }
 }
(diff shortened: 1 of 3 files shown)


arsalansufi:
@carolinemodic Assigned to you for visibility since I know you'll be creating images for QA this week and will want to set this flag!

carolinemodic:
@arsalansufi So with this flag enabled will we still have to go the cert exchange you are adding in the machine config script but its with dev certs / I can do it on my own without a VxCertifier?

arsalansufi:
Ah yes good q @carolinemodic - with this flag enabled, the existing machine config script will actually work as is without any changes at all. No cert signing required (VxCertifier or on your own). We're still using a representative VxAdmin private key and cert. They've just been pre-created and committed to the code base in https://github.com/votingworks/vxsuite/tree/main/libs/auth/certs/dev

Let me know if I can clarify any further!

kofi-q (approved):

PR #3244, opened 2023-04-10
Default to CDF CVR Exports in Scanners, Remove Exporting in Other Format

adghayes:
_review commit-by-commit_
_commits 1 - 9: VxScan_
_commits 10 - 15: VxCentralScan_
_commits 16+: cleanup_

## Overview

Removes support for exporting CVRs in the legacy format on the precinct and central scanner - all exports are in CDF format. This should not and will be merged until a PR to _import_ into VxAdmin only in CDF format is ready. 

- In VxCentralScan, removes ability to export VXF CVRs. Was able to remove most of the old CVR-generating code entirely.
- In VxCentralScan, introduced the `MockUsb` testing model we're using in VxAdmin and VxScan for consistency.
- In VxScan, we still have to maintain some of the old CVR code because we're sending CVRs to the frontend for tallying, but I was able to strip down and simplify a lot of it (we don't need images, for example, which was a lot of the logic)
- Backups also default to CDF now

## Testing Plan
Most added code in this PR was updating the automated testing to check for the new format. I have not done any manual testing, but would like to. Once the VxAdmin Import Default PR is ready, I'll test the manual flows for both scanners and the VxAdmin.

## Checklist


apps/scan/backend/src/scanners/custom/app_config.test.ts
```diff
@@ -252,44 +257,54 @@ test('export CVRs to USB in deprecated VotingWorks format', async () => {
...
+        .assertOk('test')
+        .CVR.toArray();
+
+      const cvr = cvrs[0];
+      expect(cvrs.length).toEqual(1);
+      expect(
+        convertCastVoteRecordVotesToLegacyVotes(
+          unsafeParse(CVR.CVRSchema, cvr).CVRSnapshot[0]
+        )
+      ).toMatchObject({
```
jonahkagan, line 288: Maybe `toMatchInlineSnapshot` here could make life easier

jonahkagan (approved):
Thanks for the step-by-step tour!

adghayes:
Merged in #3283.

PR #3246, opened 2023-04-10
Move jurisdiction into user model in prep for non-VxAdmin jurisdiction setting

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: A first step toward of https://github.com/votingworks/vxsuite/issues/2912

I'll soon be updating VxCentralScan, VxMark, and VxScan to store the current jurisdiction, set when the machine is configured with an election definition and unset when the machine is unconfigured. (This jurisdiction will then be used by the auth lib for jurisdiction validation - logic that's already been set up.)

We'll be getting the jurisdiction during configuration from the election manager card used to unlock the machine, which means that the auth lib needs to somehow expose the jurisdiction on the card. The simplest way to accomplish this was to move jurisdiction into the user model (as opposed to its current location outside the user model at the top-level of the card details model). In retrospect, the user model was always a good fit for this data.

## Testing

- [x] Updated existing tests
- [x] Tested auth manually

## Checklist

### DIFF
--- a/apps/mark/backend/src/app.test.ts
+++ b/apps/mark/backend/src/app.test.ts
@@ -10,7 +10,7 @@ import {
   mockOf,
 } from '@votingworks/test-utils';
 import { ElectionDefinition } from '@votingworks/types';
-import { InsertedSmartCardAuthApi } from '@votingworks/auth';
+import { DEV_JURISDICTION, InsertedSmartCardAuthApi } from '@votingworks/auth';
 import {
   ALL_PRECINCTS_SELECTION,
   ReportSourceMachineType,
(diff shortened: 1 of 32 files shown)


kshen0 (approved):

apps/mark/backend/src/app.test.ts
```diff
@@ -68,57 +70,6 @@ test('uses default machine config if not set', async () => {
   });
 });
 
-test('auth', async () => {
```
arsalansufi, line 71: I actually previously moved these tests to [app.auth.test.ts](https://github.com/votingworks/vxsuite/blob/main/apps/mark/backend/src/app.auth.test.ts) and forgot to remove the old ones still in this file

kshen0 (approved):
Nice catch, thanks for cleaning up

Issue #3243, opened 2023-04-10
moultonborough seal


mattroe:
https://votingworks.slack.com/archives/C04U11L9SCR/p1680891565441179?thread_ts=1680824231.396769&cid=C04U11L9SCR

PR #3247, opened 2023-04-10
Add seals for moultonborough

carolinemodic:
https://github.com/votingworks/vxsuite/issues/3243


kofi-q (approved):
I've been really wanting to fix this duplicated-seal situation 😅

carolinemodic:
@kofi-q yes! We should just make it a file outside of the codebase though really

carolinemodic:
but not duplicated would certainly be an improvement

Issue #3233, opened 2023-04-07
Grout: Update serialization logic to better support Dates


arsalansufi:
Copying from https://github.com/votingworks/vxsuite/pull/3192#discussion_r1148012901:

> I encountered some interesting serialization bugs when I was using Dates. JSON.stringify automatically converts Dates to strings. So when I was using a Date type here, the type was inaccurate on the frontend, which was actually working with a string.

> I looked into updating Grout's serialization to handle Dates, but Grout's serialization special-casing works best for types that can't be serialized at all. Dates actually are serialized by JSON.stringify, so I'd have to do some guess work to gauge whether a value was originally a Date before it became a string but then run the risk of accidentally converting some values that were meant to be strings into Dates. I ultimately decided that finding a general solution was out of scope, so I switched to Unix timestamps in milliseconds instead. Will open an issue for this!

eventualbuddha:
I think there’s a case to be made that we should use ISO-8601 strings or add support for `DateTime` from luxon which has a better API.

PR #3249, opened 2023-04-10
Use Dates again instead of Unix timestamps in milliseconds in auth lib code

arsalansufi:
## Overview

Since @jonahkagan identified a fix for https://github.com/votingworks/vxsuite/issues/3233 :tada:, we can use Dates again instead of Unix timestamps in milliseconds in auth lib code. (The former are just a bit more convenient than the latter.) This PR switches back to Dates accordingly.

## Testing

- [x] Updated automated tests
- [x] Tested session time limits manually

## Checklist

### DIFF
--- a/apps/mark/backend/src/app.test.ts
+++ b/apps/mark/backend/src/app.test.ts
@@ -230,7 +230,7 @@ test('configureBallotPackageFromUsb reads to and writes from store', async () =>
     Promise.resolve({
       status: 'logged_in',
       user: fakeElectionManagerUser(electionDefinition),
-      sessionExpiresAt: new Date().getTime() + 60 * 1000,
+      sessionExpiresAt: fakeSessionExpiresAt(),
     })
   );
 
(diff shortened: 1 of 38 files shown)


apps/mark/backend/src/app.test.ts
```diff
@@ -68,57 +68,6 @@ test('uses default machine config if not set', async () => {
   });
 });
 
-test('auth', async () => {
```
arsalansufi, line 71: I actually previously moved these tests to [app.auth.test.ts](https://github.com/votingworks/vxsuite/blob/main/apps/mark/backend/src/app.auth.test.ts) and forgot to remove the old ones still in this file

apps/scan/backend/src/scanners/custom/app_config.test.ts
```diff
@@ -295,7 +295,7 @@ test('export CVRs to USB in deprecated VotingWorks format', async () => {
   );
 });
 
-test('setPrecinctSelection will reset polls to closed and update auth instance', async () => {
```
arsalansufi, line 298: A relic of a much earlier iteration of auth code

eventualbuddha (approved):

apps/scan/backend/src/scanners/custom/app_config.test.ts
eventualbuddha: Untitled

eventualbuddha (approved):
`DateTime` changes look great. Thanks!

PR #3267, opened 2023-04-12
feat(nh-next): Return normalized images and computed ballot layouts

jonahkagan:
## Overview

Task: #3143

In order to get write-in adjudication working with the nh-next interpreter, we modify the interpreter to return normalized ballot images and computed ballot layout information.

_Review by commit_

## Demo Video or Screenshot
N/A

## Testing Plan
- Manual tests
- Added some basic automated tests (mainly testing plumbing, plus a few invariants on the computed layouts)
- @eventualbuddha did performance testing

## Checklist

### DIFF
--- a/apps/central-scan/backend/src/cvrs/export.ts
+++ b/apps/central-scan/backend/src/cvrs/export.ts
@@ -94,11 +94,12 @@ export async function* exportCastVoteRecords({
       includeImages && isHmpbSheet(interpretation)
         ? mapSheet(
             interpretation,
-            ({ metadata }) =>
-              store.getBallotPageLayoutForMetadata(
+            ({ metadata, layout }) =>
+              layout ??
+              (store.getBallotPageLayoutForMetadata(
                 metadata,
                 electionDefinition
-              ) as BallotPageLayout
+              ) as BallotPageLayout)
           )
         : undefined
     );
--- a/libs/ballot-interpreter-nh-next/src/interpret.rs
+++ b/libs/ballot-interpreter-nh-next/src/interpret.rs
@@ -1,5 +1,6 @@
 use std::path::Path;
 
+use image::DynamicImage;
 use image::GenericImage;
 use image::GrayImage;
 use imageproc::contrast::otsu_level;
@@ -48,10 +49,18 @@ pub struct LoadedBallotCard {
 }
 
 #[derive(Debug, Serialize)]
+pub struct NormalizedImageBuffer {
+    width: u32,
+    height: u32,
+    data: Vec<u8>,
+}
+
+#[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct InterpretedBallotPage {
     grid: TimingMarkGrid,
     marks: ScoredOvalMarks,
+    normalized_image: NormalizedImageBuffer,
 }
 #[derive(Debug, Serialize)]
 pub struct InterpretedBallotCard {
@@ -218,6 +227,17 @@ pub fn load_and_prepare_ballot_page_image(
 }
 
 #[time]
+fn make_normalized_image_buffer(image: GrayImage) -> NormalizedImageBuffer {
+    let (width, height) = image.dimensions();
+    let rgba_image = DynamicImage::ImageLuma8(image).into_rgba8();
+    NormalizedImageBuffer {
+        width,
+        height,
+        data: rgba_image.into_raw(),
+    }
+}
+
+#[time]
 pub fn interpret_ballot_card(side_a_path: &Path, side_b_path: &Path, options: &Options) -> Result {
     let LoadedBallotCard {
         side_a,
@@ -339,10 +359,12 @@ pub fn interpret_ballot_card(side_a_path: &Path, side_b_path: &Path, options: &O
         front: InterpretedBallotPage {
             grid: front_grid,
             marks: front_scored_oval_marks,
+            normalized_image: make_normalized_image_buffer(front_image),
         },
         back: InterpretedBallotPage {
             grid: back_grid,
             marks: back_scored_oval_marks,
+            normalized_image: make_normalized_image_buffer(back_image),
         },
     })
 }
--- a/libs/ballot-interpreter-nh-next/src/interpret.rs
+++ b/libs/ballot-interpreter-nh-next/src/interpret.rs
@@ -19,6 +19,8 @@ use crate::geometry::Size;
 use crate::image_utils::find_scanned_document_inset;
 use crate::image_utils::maybe_resize_image_to_fit;
 use crate::image_utils::Inset;
+use crate::layout::build_interpreted_page_layout;
+use crate::layout::InterpretedContestLayout;
 use crate::metadata::BallotPageMetadata;
 use crate::metadata::BallotPageMetadataError;
 use crate::scoring::score_oval_marks_from_grid_layout;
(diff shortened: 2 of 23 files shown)


apps/central-scan/backend/src/cvrs/export.ts
jonahkagan, line 1: @adghayes is there a similar change needed for the codepath for CDF CVRs?

adghayes: Yes. Would you suggest creating a separate PR for this or tacking a commit onto this one?

jonahkagan: Tacking on to this one is fine! I'm inferring that you're offering to make the change?

adghayes: yeah! if you can wait an hour or two

adghayes: added, sans testing

adghayes:
the test that's currently failing on CircleCI is a `custom` one in VxScan that is proving to be **super** flaky

libs/ballot-interpreter-nh-next/src/interpret.rs
```diff
@@ -339,10 +359,12 @@ pub fn interpret_ballot_card(side_a_path: &Path, side_b_path: &Path, options: &O
         front: InterpretedBallotPage {
             grid: front_grid,
             marks: front_scored_oval_marks,
+            normalized_image: make_normalized_image_buffer(front_image),
         },
         back: InterpretedBallotPage {
             grid: back_grid,
             marks: back_scored_oval_marks,
+            normalized_image: make_normalized_image_buffer(back_image),
```
eventualbuddha, line 387: Serializing a relatively large image as JSON seems likely to take a _lot_ of memory and maybe be slow. Napkin math: 1700 pixels wide × 2200 pixels high × 4 bytes per pixel × 2 images is 29.9MB, which is our baseline that we can't really do better than without compression. Then we serialize it such that, on average, each byte becomes four bytes e.g. `0xff` (one byte) → `"255,"` (four bytes). That brings our total up to ~120MB. I'll do some testing, but I wanted to comment on this because it jumped out at me.

eventualbuddha: I added some basic logging around the interpretation, serialization, and parsing and got this with `ts/src/interpret.test.ts`:

```
Interpretation took 79.192995ms
Serialization took 576.007626ms (size: 386871478)
  console.time
    parse JSON: 421 ms
```

It adds about 1s on my M1 MAX MBP and 386,871,478 bytes (387MB) worth of Rust string. I'm guessing this is because we're doing a pretty-print so there's lots of extra spaces per byte. Also, I think v8 is smart enough to keep the data encoded as UTF-8, but JS engines used to be(/ still sometimes are?) UTF-16 which would bring this up over 700MB!

eventualbuddha: And yes, not doing pretty-printing sped it up and reduced the size significantly. That seems a worthwhile change even though in practice it'll be with a much smaller amount of data:

```
Interpretation took 78.782609ms
Serialization took 179.387241ms (size: 117576475)
  console.time
    parse JSON: 396 ms
```

eventualbuddha: And if we don't serialize the normalized images as JSON:

```
Interpretation took 76.642562ms
Serialization took 107.414µs (size: 68016)
  console.time
    parse JSON: 0 ms
```

apps/central-scan/backend/src/cvrs/export.ts
jonahkagan: Thanks!

libs/ballot-interpreter-nh-next/src/interpret.rs
eventualbuddha: This should be resolved by a8b2e60cb50ccc2b87bbc96815154c313f0a27e4.

libs/ballot-interpreter-nh-next/src/layout.rs
```diff
@@ -0,0 +1,112 @@
...
+    );
+
+    let top_left_subpixel_point = grid.point_for_location(
+        top_left_grid_point.x as GridUnit,
+        top_left_grid_point.y as GridUnit,
+    )?;
+    let bottom_right_subpixel_point = grid.point_for_location(
+        bottom_right_grid_point.x as GridUnit,
+        bottom_right_grid_point.y as GridUnit,
+    )?;
```
eventualbuddha, line 62: Unnecessary casts:

```suggestion
    let top_left_subpixel_point =
        grid.point_for_location(top_left_grid_point.x, top_left_grid_point.y)?;
    let bottom_right_subpixel_point =
        grid.point_for_location(bottom_right_grid_point.x, bottom_right_grid_point.y)?;
```

libs/ballot-interpreter-nh-next/src/layout.rs
```diff
@@ -0,0 +1,112 @@
...
+                .reduce(|a, b| a.union(&b))
+                .expect("Contest must have options");
+
+            Some(InterpretedContestLayout {
+                contest_id: (*contest_id).clone(),
+                bounds,
+                options,
+            })
+        })
+        .collect::<Option<Vec<_>>>()
```
eventualbuddha, line 111: The type parameter can be inferred.

```suggestion
        .collect()
```

libs/ballot-interpreter-nh-next/src/layout.rs
```diff
@@ -0,0 +1,112 @@
...
+        bottom_right_grid_point.y as GridUnit,
+    )?;
+    let top_left_pixel_point = Point::new(
+        top_left_subpixel_point.x as PixelPosition,
+        top_left_subpixel_point.y as PixelPosition,
+    );
+    let bottom_right_pixel_point = Point::new(
+        bottom_right_subpixel_point.x as PixelPosition,
+        bottom_right_subpixel_point.y as PixelPosition,
+    );
```
eventualbuddha, line 70: Makes me wish for a `Point::round`. Then you could do this:

```suggestion
    let top_left_pixel_point = top_left_subpixel_point.round();
    let bottom_right_pixel_point = bottom_right_subpixel_point.round();
```

Actually, that's not hard to do:

```rust
impl Point<SubPixelUnit> {
    pub fn round(self) -> Point<PixelPosition> {
        Point::new(
            self.x.round() as PixelPosition,
            self.y.round() as PixelPosition,
        )
    }
}
```

libs/ballot-interpreter-nh-next/src/layout.rs
```diff
@@ -0,0 +1,112 @@
...
+
+            // Use the union of the option bounds as an approximation of the contest bounds
+            let bounds = options
+                .iter()
+                .map(|option| option.bounds)
+                .reduce(|a, b| a.union(&b))
+                .expect("Contest must have options");
+
+            Some(InterpretedContestLayout {
+                contest_id: (*contest_id).clone(),
```
eventualbuddha, line 106: Unnecessary dereference:

```suggestion
                contest_id: contest_id.clone(),
```

eventualbuddha (commented):
Looking good so far. For testing, I'd add some basic tests to `ts/src/interpret.test.ts` in `ballot-interpreter-nh-next` to ensure we're getting the layout information (maybe not bother checking the exact values? or use a snapshot?) and the image data (checking width/height/data length is probably sufficient).

libs/ballot-interpreter-nh-next/src/layout.rs
jonahkagan: Yes, was thinking of something like this! Thanks!

jonahkagan:
Ok addressed feedback and added tests!

eventualbuddha (approved):

PR #3268, opened 2023-04-12
build(deps): ensure specific packages have one version

eventualbuddha:
## Overview

Pins versions of certain packages. Includes a change to `validate-monorepo` that ensures we don't regress. This should fix issues with VS Code + ESLint loading incompatible versions of ESLint which caused it to stop working in the editor.

It's important that these packages are pinned to a specific version.  Otherwise, we can end up with multiple versions of the same package in the same monorepo, which can cause issues. For example, if we have two versions of React, then we can end up with two copies of the React context, which can cause issues. Or if we have two versions of TypeScript, we can have code flagged as an error by one version of TypeScript, but not the other.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist


script/src/validate-monorepo/validation/index.ts
```diff
@@ -18,7 +18,27 @@ export async function* validateMonorepo(): AsyncGenerator<ValidationIssue> {
   const libs = await readdir(join(root, 'libs'));
   const packages = [...services, ...libs, ...appPackages];
 
-  yield* pkgs.checkConfig({ packages: [root, ...packages] });
+  yield* pkgs.checkConfig({
+    packages: [root, ...packages],
+    // It's important that these packages are pinned to a specific version.
```
jonahkagan, line 23: Provocation: Is there any package that this invariant _shouldn't_ apply to?

jonahkagan (approved):
This is gonna save us so much pain

script/src/validate-monorepo/validation/index.ts
eventualbuddha: It's a good question. One example is Jest. We don't have much in the way of shared code that uses the Jest API, and it's likely we will want to upgrade Jest one package at a time to make it tractable. I'd be open to making this opt out instead of opt in in the future, but for now I think we should go with this.

jonahkagan: Sounds good!

PR #3273, opened 2023-04-12
Make various tweaks for vxsuite-complete-system scripts

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue links: A combo of https://github.com/votingworks/vxsuite/issues/2911, https://github.com/votingworks/vxsuite/issues/3265, https://github.com/votingworks/vxsuite/issues/3272

This PR makes various tweaks for vxsuite-complete-system scripts, specifically the basic configuration wizard. The two main tweaks are:

1. Updating vxsuite to use the latest env vars prepared by vxsuite-complete-system
2. Updating vxsuite's sys admin card programming script so that it can be used by vxsuite-complete-system for production sys admin card bootstrapping

I know 2 feels a little bit circular, but the alternative would be to recreate card programming logic in vxsuite-complete-system, which I don't think makes sense. Java Card programming is a complex operation, and the script in vxsuite is only simple because it can hook into source code.

## Testing

- [x] Tested the updated sys admin card programming scripts manually

## Checklist


libs/auth/src/env.d.ts
```diff
@@ -1,7 +1,7 @@
 declare namespace NodeJS {
   export interface ProcessEnv {
     NODE_ENV: 'development' | 'production' | 'test';
-    VX_ADMIN_PRIVATE_KEY_PASSWORD?: string;
     VX_CONFIG_ROOT?: string;
+    VX_MACHINE_PRIVATE_KEY_PASSWORD?: string;
```
arsalansufi, line 6: [Relevant line in vxsuite-complete-system](https://github.com/votingworks/vxsuite-complete-system/blob/89e898ca2dcf5d12bd64bd3b9efdb25c6104a252/config/read-vx-machine-config.sh#L53)

apps/admin/backend/src/app.ts
```diff
@@ -59,8 +59,7 @@ function constructAuthMachineState(
   return {
     arePollWorkerCardPinsEnabled: systemSettings?.arePollWorkerCardPinsEnabled,
     electionHash: currentElectionDefinition?.electionHash,
-    // TODO: Pull jurisdiction from VxAdmin cert authority cert
-    jurisdiction: DEV_JURISDICTION,
+    jurisdiction: process.env.VX_MACHINE_JURISDICTION ?? DEV_JURISDICTION,
```
arsalansufi, line 62: [Now set by vxsuite-complete-system here](https://github.com/votingworks/vxsuite-complete-system/blob/89e898ca2dcf5d12bd64bd3b9efdb25c6104a252/config/read-vx-machine-config.sh#L49)

benadida (approved):

PR #3278, opened 2023-04-12
Kevin/vxmark usb and system settings frontend

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3084
https://github.com/votingworks/vxsuite/issues/3207

* VxMark frontend loads election definition from the ballot package on USB instead of card. This reuses `UnconfiguredElectionScreen` factored out from VxScan
* Loading election definition from card has been removed, but we still check against the election hash on the card to enforce that the correct ballot package is being loaded
* VxMark loads system settings from the ballot package

## Demo Video or Screenshot
**Before**

Loading election definition from card:

https://user-images.githubusercontent.com/1060688/231608632-87585327-b544-4a6c-be53-869eea59d119.mov

---

Trying to unlock a configured machine with a card that has been configured with a different election:

https://user-images.githubusercontent.com/1060688/231608719-f6ec5614-9eaa-405d-9332-8e2cf2b336ff.mov

---

**After**

Loading election definition from USB

https://user-images.githubusercontent.com/1060688/231609097-e361738a-460d-4e71-b360-29af6b48a4a1.mov

_Note_ you can see after configuration, React warns there was a state update after component unmount. I think (but am not sure) that comes from the call to `updateElectionDefinition` in `UnconfiguredElectionScreenWrapper` since the warning doesn't fire when configuration fails. If we're removing `electionDefinition` from the frontend store soon it doesn't seem worth it to chase this down, but let me know what you think.

---

Trying to unlock a configured machine with a card that has been configured with a different election:

https://user-images.githubusercontent.com/1060688/231609271-1a05886a-eef3-4194-89c8-1cc5582430a0.mov

---

Trying to configure a fresh machine but card and USB ballot package election hashes don't match:

https://user-images.githubusercontent.com/1060688/231609348-65846ddc-21b5-4190-aa4f-24d90ea9e808.mov

## Testing Plan
* Updated existing tests, including e2e tests, to work with the new USB/React Query flow
* Added an endpoint (but no API client method) for Cypress to hit to configure the machine. Cypress previously configured machines by interacting directly with a file `input` that was removed as part of this change. Cypress has no support for mock USBs so this was the fastest and cleanest way to fix Cypress tests.

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (Logging is built in to the ballot package lib)

### DIFF
--- a/apps/mark/backend/src/app.ts
+++ b/apps/mark/backend/src/app.ts
@@ -13,7 +13,6 @@ import {
   ElectionDefinition,
   PrecinctId,
   SystemSettings,
-  safeParseElectionDefinition,
 } from '@votingworks/types';
 import { ScannerReportData, ScannerReportDataSchema } from '@votingworks/utils';
 
@@ -84,13 +83,22 @@ function buildApi(
       return auth.endCardlessVoterSession(constructAuthMachineState(input));
     },
 
-    getSystemSettings(): SystemSettings | undefined {
-      return workspace.store.getSystemSettings();
+    getElectionDefinition(): ElectionDefinition | null {
+      return workspace.store.getElectionDefinition() ?? null;
+    },
+
+    getSystemSettings(): SystemSettings | null {
+      return workspace.store.getSystemSettings() ?? null;
+    },
+
+    unconfigureMachine() {
+      workspace.store.setElection(undefined);
+      workspace.store.deleteSystemSettings();
     },
 
     async configureBallotPackageFromUsb(input: {
-      electionHash: string;
-    }): Promise<Result<void, BallotPackageConfigurationError>> {
+      electionHash?: string;
+    }): Promise<Result<ElectionDefinition, BallotPackageConfigurationError>> {
       const machineState = constructAuthMachineState(input);
       const authStatus = await auth.getAuthStatus(machineState);
 
@@ -111,32 +119,6 @@ function buildApi(
       assert(systemSettings);
       workspace.store.setElection(electionDefinition.electionData);
       workspace.store.setSystemSettings(systemSettings);
-      return ok();
-    },
-
-    async readElectionDefinitionFromCard(input: {
-      electionHash?: string;
-    }): Promise<Result<ElectionDefinition, Error>> {
-      const machineState = constructAuthMachineState(input);
-      const authStatus = await auth.getAuthStatus(machineState);
-      if (authStatus.status !== 'logged_in') {
-        return err(new Error('User is not logged in'));
-      }
-      if (authStatus.user.role !== 'election_manager') {
-        return err(new Error('User is not an election manager'));
-      }
-
-      const result = await auth.readCardDataAsString(machineState);
-      const electionData = result.ok();
-      const electionDefinition = electionData
-        ? safeParseElectionDefinition(electionData).ok()
-        : undefined;
-
-      if (!electionDefinition) {
-        // While we could provide more specific error messages for different error cases, the
-        // frontend doesn't need that much detail
-        return err(new Error('Unable to read election definition from card'));
-      }
       return ok(electionDefinition);
     },
 
--- a/apps/mark/backend/src/app.ts
+++ b/apps/mark/backend/src/app.ts
@@ -13,11 +13,13 @@ import {
   ElectionDefinition,
   PrecinctId,
   SystemSettings,
+  DEFAULT_SYSTEM_SETTINGS,
 } from '@votingworks/types';
 import { ScannerReportData, ScannerReportDataSchema } from '@votingworks/utils';
 
 import { Usb, readBallotPackageFromUsb } from '@votingworks/backend';
 import { Logger } from '@votingworks/logging';
+import { electionSampleDefinition } from '@votingworks/fixtures';
 import { getMachineConfig } from './machine_config';
 import { Workspace } from './util/workspace';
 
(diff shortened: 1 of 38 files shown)


apps/mark/frontend/jest.config.js
```diff
@@ -12,6 +12,14 @@ module.exports = {
...
     '!src/contexts/ballot_context.ts',
   ],
+  coverageThreshold: {
+    global: {
+      statements: 99.7,
+      branches: 99.7,
+      functions: 99.4,
+      lines: 99.8,
+    },
+  },
```
kshen0, line 22: Lowered primarily because `getSystemSettings.useQuery` is unused and untested.

apps/mark/frontend/src/app_root.tsx
```diff
@@ -331,6 +336,27 @@ export function AppRoot({
...
 
+  const updateElectionDefinition = useCallback(
+    (electionDefinition: ElectionDefinition) => {
+      dispatchAppState({
+        type: 'updateElectionDefinition',
+        electionDefinition,
+      });
+    },
+    []
+  );
```
kshen0, line 347: I went back and forth on the best approach here: should we remove electionDefinition from the frontend store completely? It's cleaner to do so in a single PR but ultimately to keep scope and diff size down I went with this approach of updating the frontend store every time the backend store changes. This way, places that access the frontend store for `electionDefinition` (which I'm guessing is most screens) are not impacted in this PR.

eventualbuddha: Can you mark whatever local functions etc that will be removed as `@deprecated`?

apps/mark/backend/src/app.ts
```diff
@@ -84,13 +85,33 @@ function buildApi(
...
+      const systemSettings = DEFAULT_SYSTEM_SETTINGS;
+      workspace.store.setElection(electionDefinition.electionData);
+      workspace.store.setSystemSettings(systemSettings);
+      return ok(electionDefinition);
     },
 
     async configureBallotPackageFromUsb(input: {
-      electionHash: string;
-    }): Promise<Result<void, BallotPackageConfigurationError>> {
+      electionHash?: string;
```
arsalansufi, line 113: Now that we have election def on the backend, it's no longer necessary to have the frontend provide election hash to the backend for auth checks. So we could just have the backend pull election hash directly.

That said, I know there's a separate comment thread on this PR about tackling things piecemeal, so that shift could def come in its own PR (I know I'd even said that I might tackle that). Just thought I'd note in case you wanted to start using the new approach here.

arsalansufi (commented):
Looks great!

apps/mark/frontend/src/api.ts
```diff
@@ -209,3 +204,34 @@ export const clearScannerReportDataFromCard = {
     );
   },
 } as const;
+
+/* istanbul ignore next */
```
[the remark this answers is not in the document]
kshen0: You are correct, I removed these

apps/mark/backend/src/app.ts
kshen0: Summarizing our chat, `electionHash` will never be defined on the auth object at this point (this endpoint is called to load election definition from USB so the machine can't yet have an election hash) so we can just construct a machine state with `{electionHash: undefined}` and remove the electionHash param.

apps/mark/frontend/src/app_root.tsx
kshen0: The most important thing to mark deprecated is reading `electionDefinition` from the frontend store. But as far as I can tell there's no function for reading from the frontend store; it's just dereferenced from `appState` and passed from `app_root` to its children via props.

However, based on your feedback I did mark `updateElectionDefinition` as deprecated in https://github.com/votingworks/vxsuite/pull/3278/commits/88bc90d1fcf867bf7e19bf4577317b43c27aa1f0#diff-1cc7f1707c76fa44e5e42e6fd40172387d8b03bfc55bd5fa63b7c2f334c78e6eR339

apps/mark/backend/src/app.ts
```diff
@@ -84,13 +85,33 @@ function buildApi(
       return auth.endCardlessVoterSession(constructAuthMachineState(input));
     },
 
-    getSystemSettings(): SystemSettings | undefined {
-      return workspace.store.getSystemSettings();
+    getElectionDefinition(): ElectionDefinition | null {
+      return workspace.store.getElectionDefinition() ?? null;
```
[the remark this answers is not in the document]
arsalansufi: Mm got it! I have no issue with using `null` - was just curious what the reasoning was. And I see now that it was react-query not allowing `undefined`

Using a `Result` is also reasonable! Whatever we pick, I would just say consistency across the machines would be nice. Looking at other machines, I'm now seeing that they also use `null` 🙈

arsalansufi (approved):
LGTM!

apps/mark/backend/src/app.ts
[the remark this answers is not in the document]
kshen0: I'm still fuzzy on use cases for `Result` vs returning the value directly, but handling Result in this case can feel clunky eg https://github.com/votingworks/vxsuite/commit/a8e13623d0b320b72e94cbc33844efea65899cba#diff-1cc7f1707c76fa44e5e42e6fd40172387d8b03bfc55bd5fa63b7c2f334c78e6eR354

So to maintain consistency and better ergonomics I'll revert

PR #3279, opened 2023-04-13
Enable double sheet detection with the option to disable

carolinemodic:
## Overview
Makes double sheet detection enabled by default but with the option to disable. When there is a double sheet jam shows a specific UI relevant to multiple sheets. 

https://github.com/votingworks/vxsuite/issues/3235

## Demo Video or Screenshot

https://user-images.githubusercontent.com/14897017/231652221-0d5aeb23-07cc-4c7f-8d65-fffe78b3f3b3.mov

All scans in above video are two sheets layered on top of one another

## Testing Plan

Manually tested various moments of double sheet feeding, manually tested that killing and restarting app keeps the disable state 

## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/custom/state_machine.ts
+++ b/apps/scan/backend/src/scanners/custom/state_machine.ts
@@ -95,6 +95,7 @@ type ScannerStatusEvent =
   | { type: 'SCANNER_BOTH_SIDES_HAVE_PAPER' }
   | { type: 'SCANNER_JAM' }
   | { type: 'SCANNER_JAM_CLEARED' }
+  | { type: 'SCANNER_JAM_DOUBLE_SHEET' }
   | { type: 'SCANNER_DISCONNECTED' };
 
 type CommandEvent = { type: 'SCAN' } | { type: 'ACCEPT' } | { type: 'RETURN' };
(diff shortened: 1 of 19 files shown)


apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -270,13 +275,16 @@ async function reset({ client }: Context): Promise<void> {
...
   debug('Scanning');
+  const isUltrasonicDisabled = workspace.store.getIsUltrasonicDisabled();
   const scanResult = await client.scan({
     wantedScanSide: ScanSide.A_AND_B,
     resolution: ImageResolution.RESOLUTION_200_DPI,
     imageColorDepth: ImageColorDepthType.Grey8bpp,
     formStandingAfterScan: FormStanding.HOLD_TICKET,
     // The A4 scanner does not support double sheet detection.
-    doubleSheetDetection: DoubleSheetDetectOpt.DetectOff,
+    doubleSheetDetection: isUltrasonicDisabled
```
jonahkagan, line 285: Given that the custom lib calls this "double sheet detection," it seems a bit weird that we then start calling it "ultrasonic" at higher levels in the stack. For consistency, I think it would be clearer to call it "double sheet detection" throughout the system. I recognize it's kind of an annoying change to make but maybe is worth it if you agree?

jonahkagan (approved):
Looks great overall!

apps/scan/backend/src/scanners/custom/state_machine.ts
carolinemodic: yeah good point. Going to merge this as is but I made a note to make a follow up PR for this

Issue #3015, opened 2023-02-15
fixtures: Remove electionPrimaryNonpartisanContests


jonahkagan:
It's now an exact duplicate of electionMinimalExhaustiveSample

Issue #1341, opened 2022-01-11
VxAdmin: import CDF CVRs


jonahkagan:

adghayes:
Closed by #3283.

Issue #1340, opened 2022-01-11
VxScan & VxCentralScan: export CDF CVRs


jonahkagan:

adghayes:
Closed by #3283.

PR #3283, opened 2023-04-13
VxAdmin, VxScan, VxCentralScan: Export & Import CVRs in VVSG CDF Format

### LINKED ISSUE #1240: CVR: Add count of write-ins on ballot
This satisfies VVSG 2.9 requirement 1.1.5-D.4:
The voting system must record write-in information in the CVR that includes the total number of write-ins in the CVR.

Presumably this means the number of write-ins on each ballot/CVR in the CVR file, not the total file.

### LINKED ISSUE #2360: Reevaluate how to transfer write in images from scanning to VxAdmin

adghayes:
_review commit-by-commit_
_commits 1-3: already reviewed in #3244, no new code (besides rebase resolutions)_
_commits 4+: new_

## Overview

This PR is a continuation of #3244. It will finally close #1240, #1241, #1333, #1334, #1335, #1336, #1337, #1338, #1340, and #1341.

The previous PR #3244 makes CDF the default export from the scanners and this PR completes the CDF work by making it the default import format in VxAdmin. There are still some loose ends, but for the most part this ends the project.

Loose Ends:
- Legacy CVR Files are still in the codebase and used as fixtures in `libs/utils` and `admin/frontend`. The tests using it are not using it to test file imports generally, just using those files as a source of vote data to test tallying and such. @carolinemodic and I decided it was OK to leave those in and gradually remove them as the tallying code gets revamped, but open to disagreement.
- I'd like to do #3015 as a follow-up to just have one fewer of the legacy files
- I'm unsure whether to close #2360. We are now transferring files differently but I think we need to normalize them on import on the VxAdmin side. This was not possible when we were briefly supporting both the formats, but is now possible. I started a discussion in Slack. This will roll nicely into the VxAdmin tally project. 
- Not sure what the thinking is behind #833. @carolinemodic?


apps/admin/backend/src/app.ts
```diff
@@ -292,25 +297,30 @@ function buildApi({
 
     async addCastVoteRecordFile(input: {
       path: string;
-    }): Promise<AddCastVoteRecordFileResult> {
+    }): Promise<
+      Result<
+        Admin.CvrFileImportInfo,
+        AddCastVoteRecordReportError & { message: string }
```
adghayes, line 306: @jonahkagan I know that we discussed moving all error message creating and handling to the frontend, but since I wanted to use the same message for logging (which happens on the backend) I just kept the message creation on the backend.

apps/admin/backend/src/store.test.ts
```diff
@@ -56,700 +52,25 @@ test('assert election exists', () => {
   );
 });
 
-test('add a CVR file', async () => {
```
adghayes, line 59: Previously, we were testing a lot of CVR actions in `store.test.ts` and then testing them again at a higher level in `app.test.ts`. Now we're mostly only testing it in `app.cvr_files.test.ts` (or `app.cdf_cvr_files` at this point in the PR), so there's less testing overall but the coverage is pretty much the same. I'm for testing at the higher level (which uses the store) and not belaboring the endpoints here.

apps/admin/backend/src/cvr_files.ts
```diff
@@ -436,13 +438,14 @@ export function getAddCastVoteRecordReportErrorMessage(
...
       if (error.currentFileMode === Admin.CvrFileMode.Official) {
-        return `You are currently tabulating official results but the selected cast vote record report report contains test results.`;
+        return `You are currently tabulating official results but the selected cast vote record report contains test results.`;
       }
       return `You are currently tabulating test results but the selected cast vote record report contains official results.`;
     case 'invalid-cast-vote-record': {
       const messageBase = `Found an invalid cast vote record at index ${error.index} in the current report.`;
       const messageDetail = (() => {
         const subErrorType = error.error;
+        /* istanbul ignore next  - write testing when error handling requirements and implementation harden */
```
adghayes, line 432: Happy to go back and write testing for this but it seemed like a lot of branches that would require a bunch of copy-pasted tests that don't actually verify much. I wasn't sure about this error handling approach in general so punted.

apps/admin/backend/src/cvr_files.ts
```diff
@@ -665,7 +665,7 @@ export async function addCastVoteRecordReport({
               CVR_BALLOT_IMAGES_SUBDIRECTORY,
               CVR_BALLOT_LAYOUTS_SUBDIRECTORY
             )
-            .replace(/jpg|jpeg/, 'layout.json');
+            .replace(/(jpg|jpeg)$/, 'layout.json');
```
adghayes, line 649: I think the normalized filenames with the rust interpreter are different than they were before, requiring me to be more precise here. The new filenames are like: `324123-12312-uuid-1234123.jpg-12312-3129-1232-uuid-12321-normalized.jpg`

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -376,6 +376,7 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element | null {
             <FileInputButton
               data-testid="manual-input"
               onChange={processCastVoteRecordFileFromFilePicker}
+              accept=".json"
```
adghayes, line 379: There are ways to allow the user to select directories, but I think they're not intuitive to the user (and require some annoying React HTML workarounds), so let's just let them select the `.json` report and handle the path on the backend.

libs/utils/src/filenames.ts
```diff
@@ -171,7 +153,10 @@ export function parseCastVoteRecordReportDirectoryName(
   // extract number of ballots
   assert(typeof postTestPrefixSegments[1] !== 'undefined');
   const ballotSegments = postTestPrefixSegments[1].split(SUBSECTION_SEPARATOR);
-  if (ballotSegments.length !== 2 || ballotSegments[1] !== 'ballots') {
+  if (
+    ballotSegments.length !== 2 ||
+    (ballotSegments[1] !== 'ballots' && ballotSegments[1] !== 'ballot')
```
adghayes, line 158: Bug fix.

libs/types/src/cdf/cast-vote-records/vx-schema.json
```diff
@@ -293,7 +293,7 @@
       "type": "object"
     },
     "CVR.CVRSnapshot": {
-      "required": ["@id", "@type", "Type"],
+      "required": ["@id", "@type", "Type", "CVRContest"],
```
adghayes, line 296: Rationale: every snapshot should have contests. If it doesn't, the CVR just an entirely blank sheet. In VVSG, undervotes should be represented explicitly by contest, so there will be a list of contests even if a voter does cast an entirely blank sheet.

libs/types/src/cdf/cast-vote-records/vx-schema.json
```diff
@@ -244,7 +244,7 @@
       "type": "object"
     },
     "CVR.CVRContestSelection": {
-      "required": ["@type", "SelectionPosition"],
+      "required": ["@type", "SelectionPosition", "ContestSelectionId"],
```
adghayes, line 247: Rationale: we have no use for a contest selection if we cannot tie it to an ID in the election definition.

libs/types/src/cdf/cast-vote-records/vx-schema.json
```diff
@@ -196,7 +196,7 @@
       "type": "object"
     },
     "CVR.CVRContest": {
-      "required": ["@type", "ContestId"],
+      "required": ["@type", "ContestId", "CVRContestSelection"],
```
adghayes, line 199: Rationale: this just made the typing easier. Use an empty array if there are no selections. Can revert this one if it seems excessive.

apps/admin/backend/src/app.writeins.test.ts
```diff
@@ -190,11 +197,15 @@ test('write-in adjudication lifecycle', async () => {
...
-    expect(writeInImageEntries).toHaveLength(0); // the fixtures do not have ballot images
+    // Skipping this check on the write-in images. The fixture does now have
+    // write in images but attempting to retrieve them will throw an error
+    // because the endpoint is not compatible with VotingWorks format
+    // ballot layouts.
+
+    // const writeInImageEntries = await apiClient.getWriteInImage({
+    //   writeInId: writeInRecord.id,
+    // });
+    // expect(writeInImageEntries).toHaveLength(0); // the fixtures do not have ballot images
```
jonahkagan, line 208: Unless you have an eventual path to resolution, maybe just remove this whole chunk

apps/admin/backend/src/store.test.ts
jonahkagan: I'm into it

apps/admin/backend/src/cvr_files.ts
jonahkagan: Kind of a weird filename format, but hardening the regex seems like a net positive anyway

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
jonahkagan: Cute workaround

libs/utils/src/filenames.ts
jonahkagan: Regression test?

libs/types/src/cdf/cast-vote-records/vx-schema.json
jonahkagan: I'm into this. I think in general empty array communicates "there were no selections" better than undefined, which sounds more like "we don't know if there were selections"

jonahkagan (approved):
🎉 Looks great!

apps/admin/backend/src/app.ts
```diff
@@ -292,25 +297,30 @@ function buildApi({
...
+      >
+    > {
       const { path } = input;
       const userRole = assertDefined(await getUserRole());
       const filename = basename(path);
       let fileStat: Stats;
       try {
         fileStat = await fs.stat(path);
       } catch (error) {
+        const message = `Failed to access cast vote record report for import.`;
```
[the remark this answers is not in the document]
adghayes: I copied it just because it to generate the message with `getAddCastVoteRecordReportErrorMessage` I would have had to sort of manufacture an error to pass into the method. But I can do that instead. Changed in e191dddc28eefaf40b3e17a77615d65c8c74a297

libs/types/src/cdf/cast-vote-records/vx-schema.json
adghayes: OK, I wasn't sure if making changes to our spec for typing purposes was overkill, because we technically are restricting the interoperability of the format. But for the CVR project we've never been terribly concerned about robust interoperability, so I thought it better to make adjustments for ourselves.

apps/admin/backend/src/app.cdf_cvr_files.test.ts
```diff
@@ -269,15 +267,13 @@ test('adding a duplicate file returns OK to client but logs an error', async ()
   const reportDirectoryPath = standardCdfCvrReport.asDirectoryPath();
 
   // add file once
-  await apiClient.addCastVoteRecordFile({
+  void (await apiClient.addCastVoteRecordFile({
```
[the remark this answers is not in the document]
adghayes: We have a lint rule that requires `Result`s to be handled somehow and it was either this or make assertions on `isOk` everytime.

libs/types/src/cdf/cast-vote-records/vx-schema.json
jonahkagan: I think it's useful to be as constrained as possible until we need more flexibility

apps/admin/backend/src/app.cdf_cvr_files.test.ts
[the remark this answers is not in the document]
jonahkagan: Will the test silently pass if for some reason the result is not `ok` here? If so, I think an assertion would make it easier to debug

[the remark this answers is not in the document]
adghayes: no, all these tests should fail if the result is not `ok` here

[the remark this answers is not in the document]
adghayes: changed anyway in a61f411710da5c024c74c8e0f96611bbac22b589 in case I'm wrong

apps/admin/backend/src/app.writeins.test.ts
adghayes: removed in a61f411710da5c024c74c8e0f96611bbac22b589

libs/utils/src/filenames.ts
adghayes: added in 045ff193fd7d6fd6db767acf33180f182970fcfe

PR #3285, opened 2023-04-13
[VxScan] Reduce crowding on ElectionManagerScreen

kofi-q:
## Overview

At the default VVSG size setting, the Election Manager Screen is getting a little crowded and in some states, content overflows and gets rendered below the fold.

This page could us a re-design, but in the meantime, taming a few elements on the screen so things fit at least at the default "medium" text size:
- Render an `H2`-sized page title instead of `H1` (with `as="h1"` to preserve a11y heading hierarchy)
- Stop rendering the `ChangePrecinctButton` as `<Select large>` so we make a bit of room - it's a little too large now with the VVSG styling.
- Change the disabled button explanation text at the bottom from `<P>` to `<Caption>`

## Testing Plan
- Visual verification

## Checklist

### DIFF
--- a/apps/scan/frontend/src/screens/election_manager_screen.tsx
+++ b/apps/scan/frontend/src/screens/election_manager_screen.tsx
@@ -9,8 +9,9 @@ import {
   SetClockButton,
   UsbDrive,
   ChangePrecinctButton,
-  H1,
   P,
+  H2,
+  Caption,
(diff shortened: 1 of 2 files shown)


kshen0 (approved):
Didn't know about the `as` prop, thanks for the explanation!

PR #3300, opened 2023-04-17
refactor(auth): use `tmp` directly

eventualbuddha:
## Overview
Rather than implementing our own `tmp`-like functionality, let's just use the `tmp` package. Advantages:
- we don't have to hardcode a particular place to put files
- we don't have to implement logic to avoid writing to an existing file
- we can use the existing `tmp` cleanup facility

All of the above simplifies the tests somewhat as well.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist


arsalansufi (approved):
Nice! Thanks for making this change - agree that we shouldn't reinvent the wheel

PR #3304, opened 2023-04-18
Kevin/vxcentralscan usb and system settings

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3083
https://github.com/votingworks/vxsuite/issues/3208

In VxCentralScan, moves ballot package loading from frontend to backend. This means
* The USB programming flow is standardized with VxMark and VxScan and allows us to reuse loading and error states in `UnconfiguredElectionScreen`
* We get `system settings` loading for ~free

Unfortunately we lose the detailed reporting of ballot template loading status since the frontend is now unaware of how many templates have been loaded until all templates have been loaded. 

## Demo Video or Screenshot

Path 1:
* Auth
* Insert USB to start automatic configuration
* Eject USB when prompted
* Unconfigure machine (does not prompt for USB ejection again)
* 
https://user-images.githubusercontent.com/1060688/232956090-b6b1b5c3-1501-4686-b71a-05bdd8a045d4.mov

---

Path 2:
* Auth
* Insert USB to start automatic configuration
* Decline to eject USB
* USB is ejected right before unconfigure is confirmed

https://user-images.githubusercontent.com/1060688/232956084-0ecdc181-f004-4920-ae7b-6555b27af20d.mov

---

Edge case: USB is mounted before auth

https://user-images.githubusercontent.com/1060688/232956061-38b02fc1-9da7-43cf-b334-6c45bd0931ea.mov

## Testing Plan
- updated existing tests to check for `UnconfiguredElectionScreen` flows

## Checklist


apps/scan/backend/src/store.ts
```diff
@@ -657,16 +657,6 @@ export class Store {
     return scannerBackedUpAt >= DateTime.max(...cvrsLastUpdatedDates);
   }
 
-  addBallotCard(batchId: string): string {
```
kshen0, line 660: This method was unused and `ballot_cards` doesn't exist in the schema. I deleted it to because test coverage was under threshold by 0.01% even though I didn't make any changes to /scan. Deleting these lines puts this PR back over threshold

apps/central-scan/backend/src/central_scanner_app.ts
```diff
@@ -119,6 +128,64 @@ function buildApi({
         throw error;
       }
     },
+
+    async configureWithSampleBallotPackageForIntegrationTest(): Promise<void> {
```
kshen0, line 132: This is a workaround because we don't have mock USB support for cypress yet

apps/central-scan/integration-testing/cypress/e2e/configuration.cy.ts
```diff
@@ -68,35 +68,20 @@ describe('BSD and services/Scan', () => {
     cy.visit('/');
     enterPin();
     removeCard();
-    cy.contains('Load Election Configuration', { timeout: 10000 });
-  });
 
-  it('BSD can be configured with services/scan with an election JSON file', () => {
```
kshen0, line 74: This option was removed from the UI

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -460,10 +473,41 @@ export function AppRoot({
...
+
+  // The `configure` useEffect hook above depends on `authStatus`. Since hooks can't
+  // be rendered conditionally, we need to short circuit on authStatusQuery after the
+  // hook is defined.
   if (!authStatusQuery.isSuccess) {
     return null;
   }
-  const authStatus = authStatusQuery.data;
+  // assert authStatus so TypeScript knows it's not undefined
+  assert(authStatus);
```
arsalansufi, line 510: Super minor but I think you can drop this assertion if you move the defining of `authStatus` back here:

```ts
const authStatus = authStatusQuery.data;
```

And just use `authStatusQuery.data` in the `useEffect` above, which might be preferable since the `useEffect` will eventually go away when we fully move the election definition to react-query.

But also feel free to leave as is!

arsalansufi (approved):
LGTM!

apps/central-scan/frontend/src/app_root.tsx
kshen0: Good call - [5057e2d](https://github.com/votingworks/vxsuite/pull/3304/commits/5057e2dfb8907c5edb394aad4c0b2ff0373e332c)

PR #3305, opened 2023-04-18
feat(scan): remove worker pool

eventualbuddha:
## Overview

Cleaning up some old VxCentralScan code. Though it does regress the performance of VxCentralScan, I assert this doesn't matter because:
- we don't use this in VxScan and it's acceptable there.
- the parallelism provided by the worker pool was used by `ballot-interpreter-vx` and `ballot-interpreter-nh`, which are both deprecated.

I left the "worker" terminology in place, but it barely makes any sense to call them that. I'm open to alternative naming suggestions.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist


apps/central-scan/backend/src/importer.ts
```diff
@@ -49,58 +45,38 @@ export class Importer {
...
-  private async getWorkerPool(): Promise<
-    WorkerPool<workers.Input, workers.Output>
-  > {
-    if (!this.workerPool) {
-      this.workerPool = this.workerPoolProvider();
-      this.workerPool.start();
-      await this.workerPool.callAll({
-        action: 'configure',
-        dbPath: this.workspace.store.getDbPath(),
+  private async interpreterReady(): Promise<void> {
```
jonahkagan, line 61: `configureInterpreter` might be a bit clearer

apps/central-scan/backend/src/importer.ts
```diff
@@ -22,23 +22,19 @@ import { join } from 'path';
...
 import { Castability, checkSheetCastability } from './util/castability';
-import { HmpbInterpretationError } from './util/hmpb_interpretation_error';
 import { Workspace } from './util/workspace';
 import {
   describeValidationError,
   validateSheetInterpretation,
 } from './validation';
-import * as workers from './workers/combined';
 import * as interpretNhWorker from './workers/interpret_nh';
 import * as interpretVxWorker from './workers/interpret_vx';
```
jonahkagan, line 31: Maybe call these "interpreters" rather than "workers"? That's how we colloquially talk about them

apps/central-scan/backend/src/importer.ts
```diff
@@ -49,58 +45,37 @@ export class Importer {
...
-  private interpreterReady = true;
-
-  constructor({
-    workspace,
-    scanner,
-    workerPoolProvider = (): WorkerPool<workers.Input, workers.Output> =>
-      inlinePool<workers.Input, workers.Output>(workers.call),
-  }: Options) {
+  private interpreterState: 'init' | 'configuring' | 'ready' = 'init';
+  private interpreterReadyPromise?: Promise<void>;
```
jonahkagan, line 49: Do we need to store this promise? Seems like we could equivalently just check `interpreterState`

jonahkagan (approved):
🧹

apps/central-scan/backend/src/importer.ts
eventualbuddha: We can't just check `interpreterState` because if configuration is in progress and `configureInterpreter` is called again it will just resolve immediately. Holding on to the promise ensures that all callers are resolved at the same time and only _after_ the interpreter is ready.

jonahkagan: Gotcha, thanks!

PR #3306, opened 2023-04-18
SKIP_ELECTION_HASH_CHECK Env. Variable

adghayes:
## Checklist


apps/scan/backend/src/vx_interpreter.test.ts
```diff
@@ -36,15 +38,20 @@ import { Interpreter, sheetRequiresAdjudication } from './vx_interpreter';
...
+jest.mock('@votingworks/utils', () => {
   return {
-    ...jest.requireActual('@votingworks/ballot-encoder'),
-    // to allow changing election definitions without changing the image fixtures
-    // TODO: generate image fixtures from election definitions more easily
-    // this election hash is for the MS demo fixture images
-    sliceElectionHash: () => 'b0260b4e9d492dab3813',
+    ...jest.requireActual('@votingworks/utils'),
+    isFeatureFlagEnabled: (flag: BooleanEnvironmentVariableName) =>
+      featureFlagMock.isEnabled(flag),
```
jonahkagan, line 47: This boilerplate is probably my least favorite part of this approach. Would it be possible to write it like this:

```ts
jest.mock('@votingworks/utils', () => {
  return {
    ...featureFlagMock.exports,
  }
})
```

Or something like that. Just thinking of ways to reduce this

apps/scan/backend/src/vx_interpreter.test.ts
```diff
@@ -36,15 +38,20 @@ import { Interpreter, sheetRequiresAdjudication } from './vx_interpreter';
...
   };
 });
+beforeEach(() => {
+  featureFlagMock.enableFeatureFlag(
+    BooleanEnvironmentVariableName.SKIP_ELECTION_HASH_CHECK
+  );
+});
```
jonahkagan, line 54: Do we need to do any resetting after each test?

jonahkagan (commented):
Good compromise. I think this is a solid overall direction. No blocking feedback, just ideas

apps/scan/backend/src/vx_interpreter.test.ts
adghayes: Yeah, good call, I realized that later and have it added in an in-progress commit.

apps/scan/backend/src/vx_interpreter.test.ts
adghayes: I would love to be able to reduce this. At first I just had this: 
```ts
  return {
    isFeatureFlagEnabled: featureFlagMock.isEnabled,
  }
```
I like that it's slightly less boilerplate, but it won't work because the `jest.mock` is hoisted and `featureFlagMock` is not defined at the time that its property would be accessed.

I'm trying to wrap my head around your `.exports` idea. On of the reasons to actually instantiate something (to have a `getFeatureFlagMock` rather than a singleton mock) is because I was concerned about `jest` files being run in parallel, and feature flag mocks interfering with one another. But if they're being run in parallel, can they interfere with one another if they share a dependency? Now that I think about it, I think that `jest` re-imports dependencies for every test suite so it would actually be safe to have a singleton pattern, then you could spread in like this. Thoughts?

jonahkagan: Ugh I forgot about the hoisting. That's annoying. I think the less shared state across modules, the less likely it is for something confusing to happen, so I'm inclined to just keep it as you have it here.

We could also explore avoiding the hoisting. Looks like `jest.doMock` is the way to do that: https://jestjs.io/docs/jest-object#jestdomockmodulename-factory-options

adghayes: I've tried to get `jest.doMock` to work in our codebase several times and never gotten it to work. There are some conditions around it I don't fully understand. If you ever figure it out, that would be excellent!

libs/utils/src/environment_variable.ts
```diff
@@ -26,6 +26,9 @@ export enum BooleanEnvironmentVariableName {
   // For NH pilot: Use dev certs in prod for Java Card auth. Only relevant if ENABLE_JAVA_CARDS is
   // also true.
   USE_DEV_CERTS_IN_PROD = 'REACT_APP_VX_USE_DEV_CERTS_IN_PROD',
+  // Skips election hash checks in various apps to allow using old fixtures in
+  // development even as their respective election definitions change
+  SKIP_ELECTION_HASH_CHECK = 'REACT_APP_VX_SKIP_ELECTION_HASH_CHECK',
```
arsalansufi, line 31: I might suggest a slightly more specific name, e.g. `SKIP_CVR_ELECTION_HASH_CHECK`. When I first saw this variable name, I thought we might be skipping election hash checks for smart cards

adghayes: OK, good call. It is meant to be a more generic purpose flag - it's skipping the election hash checks in scanning interpretation as well, not just for CVRs. Could maybe be `SKIP_IMPORT_ELECTION_HASH_CHECK` but that's also vague.

arsalansufi: `SKIP_ELECTION_HASH_CHECK_ON_IMPORT` sounds alright to me! I'm also okay with more verbose if need be, e.g.
`SKIP_SCAN_AND_CVR_IMPORT_ELECTION_HASH_CHECKS`

jonahkagan: Could also make separate env variables for each place we want to skip the hash check if we want to be more targeted.

libs/utils/src/features_mock.ts
```diff
@@ -0,0 +1,37 @@
...
+    disableFeatureFlag: (flag: BooleanEnvironmentVariableName) => {
+      mockedFlags[flag] = false;
+    },
+    resetFeatureFlag: (flag?: BooleanEnvironmentVariableName) => {
+      if (flag) {
+        delete mockedFlags[flag];
+      } else {
+        mockedFlags = {};
+      }
+    },
```
jonahkagan, line 35: Given that this is currently only being used to reset all flags, maybe just change to `resetFeatureFlags` and don't take any `flag` parameter

jonahkagan (approved):

libs/utils/src/environment_variable.ts
adghayes: Cool, I'll split the variables.

libs/utils/src/features_mock.ts
adghayes: Done.

PR #3308, opened 2023-04-18
feat(scan): parameterize write-in area

eventualbuddha:
## Overview
Requires that `gridLayout` elections specify an area around the target mark to crop the containing contest option properly. This is used for cropping the write-in area in the write-in adjudication flow.

Note that any previous `gridLayout` elections will no longer be valid.

## Demo Video or Screenshot
n/a

## Testing Plan
- [ ] Automated
- [x] Manually scanned a ballot with write-ins and imported the CVRs into the VxAdmin adjudication flow

## Checklist


libs/types/src/election.ts
```diff
@@ -459,13 +459,20 @@ export interface GridLayout {
   readonly ballotStyleId: BallotStyleId;
   readonly columns: number;
   readonly rows: number;
+  /**
+   * Area in timing mark units around a target mark (i.e. bubble) to consider
+   * part of the option for that target mark. This is used to crop the image
+   * to show the write-in area for a given grid position.
+   */
+  readonly optionBoundsFromTargetMark: Rect;
```
jonahkagan, line 467: How do you think about `gridLayouts` vs `ballotLayout`? Since `ballotLayout.targetMarkPosition` is related to the option bounds, I could see a case for colocating them

libs/types/src/election.ts
```diff
@@ -459,13 +459,20 @@ export interface GridLayout {
   readonly ballotStyleId: BallotStyleId;
   readonly columns: number;
   readonly rows: number;
+  /**
+   * Area in timing mark units around a target mark (i.e. bubble) to consider
+   * part of the option for that target mark. This is used to crop the image
```
jonahkagan, line 464: maybe s/image/ballot image/

libs/ballot-interpreter-nh-next/src/election.rs
```diff
@@ -16,6 +16,14 @@ idtype!(PrecinctId);
 //
 // IF YOU CHANGE ANYTHING HERE, YOU MUST ALSO CHANGE IT THERE.
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Rect {
+    pub x: i32,
+    pub y: i32,
+    pub width: u32,
+    pub height: u32,
+}
```
jonahkagan, line 25: It might be nice to unify `geometry::Rect` with `election::Rect`

jonahkagan (approved):

libs/types/src/election.ts
eventualbuddha: One potential advantage of putting this in `ballotLayout` is reduced duplication. However, it also precludes having e.g. an election with both state and local ballots that have different values. Maybe that wouldn't ever happen, though. The biggest reason I put it here is because if I put it in `ballotLayout` it has to be optional, but I wanted it to be required if `gridLayouts` is present. We could enforce that with a Zod `refine` call, but I preferred to let the type system handle it.

jonahkagan: Ah, yeah I haven't thought about elections that might have multiple entries in `gridLayouts` before...

What is `ballotLayout` used for? Should it potentially be moved inside `gridLayouts`?

Maybe all of this is moot if we switch to CDF for this anyway. We don't need to over-rotate on it now

eventualbuddha: I could maybe see us merging them together, but I agree we don't need to fix this now.

libs/ballot-interpreter-nh-next/src/election.rs
eventualbuddha: Ended up creating an `Outset` struct instead.

jonahkagan: Nice, that's more descriptive

PR #3318, opened 2023-04-19
[libs/ui] Add a theme management API context to AppBase

kofi-q:
## Overview

Adding a theme management API (`setColorMode`, `setSizeMode`, `resetThemes`) which will initially be used by the voter-facing display settings component.

- Updating  the `colorMode`, `sizeMode` props on `AppBase` to a more appropriately named `default*Mode` - these will now represent the initial/default state of the UI that we return to when `resetThemes` is used.
- Also updating the storybook bootstrap code to use the theme manager API when switching between modes.

## Testing Plan
- Added test case for the `AppBase` implementation of the `ThemeManagerContext`

## Checklist

### DIFF
--- a/apps/admin/frontend/src/app.tsx
+++ b/apps/admin/frontend/src/app.tsx
@@ -29,7 +29,7 @@ export function App({
   return (
     <BrowserRouter>
       <AppBase
-        colorMode={colorMode}
+        defaultColorMode={colorMode}
         legacyBaseFontSizePx={baseFontSizePx}
         legacyPrintFontSizePx={printFontSizePx}
       >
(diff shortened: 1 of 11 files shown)


adghayes (approved):
LGTM. Looking at this makes me realize I don't understand the following... how have the settings menus been affecting font size and contrast until now, without a theme management API?

kofi-q:
> how have the settings menus been affecting font size and contrast until now, without a theme management API?

Ah yeah, they were previously changing the `color|sizeMode` props on the `AppBase` component and that would trigger a re-render with the new theme, but wanted a plug-and-play pattern for the upcoming display settings component where it could trigger updates without needing additional logic in the host app

kofi-q:
(A piece I left out is that the controls in the storybook menu bar also update global state within storybook, which trigger a re-render)

PR #3325, opened 2023-04-20
refactor(admin): store ballot images differently, not as base 64

adghayes:
## Overview

We have been storing ballot images in VxAdmin as base64 encoded strings alongside our CVRs. We want to store in a different format because, as @eventualbuddha, pointed out, the base64 requires conversion whenever we want to export and requires more storage. I've also modified the data model to normalize each page image and its data (tying layout to image).

## Demo Video or Screenshot
Nothing has changed in the flow, but the performance of a couple aspects has changed. The CVR loading is significantly faster. Loading the write-in images on the transcription screen is a bit slower, though. As a result, I've added some prefetching logic which makes it _faster_ than before, with the exception of the first image in a contest. I also tried an alternate approach with serving images out of a static folder, but did not see sufficient advantage to that (although it is certainly an option if we go the file route).

previously:

https://user-images.githubusercontent.com/37960853/233423442-f1bc1dad-1f2d-4860-9c4d-89f5c02b6a03.mov

after refactor, no prefetching:

https://user-images.githubusercontent.com/37960853/233423736-2d273357-bd8a-402f-a10e-14e353e9bd88.mov

after refactor, with prefetching:

https://user-images.githubusercontent.com/37960853/233473556-c42a3da4-696b-4a37-8699-80e738a7a349.mov

alternate approach (serving images as static assets rather than base64):

https://user-images.githubusercontent.com/37960853/233441440-6ab6ee2b-4715-4533-8c5c-419b83393c15.mov

## Testing Plan
- [x] manual testing on `electionGridLayoutsNewHampshireAmherst` fixtures
- [x] update automated testing

## Checklist

- [ ] ~~I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates~~


libs/backend/src/cast_vote_record_helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+  for (const cvrContest of currentSnapshot.CVRContest) {
+    for (const cvrContestSelection of cvrContest.CVRContestSelection) {
+      for (const selectionPosition of cvrContestSelection.SelectionPosition) {
+        const cvrWriteIn = selectionPosition.CVRWriteIn;
+        if (
+          selectionPosition.HasIndication === CVR.IndicationStatus.Yes &&
+          cvrWriteIn
+        ) {
+          // Check for a BMD write-in
+          if (cvrWriteIn.Text) {
```
arsalansufi, line 50: Just curious: How does `cvrWriteIn.Text` indicate that the write-in is a BMD write-in?

Also, very minor, but rather than adding a comment to indicate what this check is doing, we could wrap it in a function to make that clear, e.g.

```ts
function isBmdWriteIn(cvrWriteIn: ...): boolean {
  // Might still want some comment explaining why this indicates that the write-in is a BMD
  // write-in
  return Boolean(cvrWriteIn.Text);
}
```

libs/backend/src/cast_vote_record_helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+                cvrWriteIn.WriteInImage?.Location &&
+                cvrImageData.Location === cvrWriteIn.WriteInImage.Location
+            );
+            castVoteRecordWriteIns.push({
+              contestId: cvrContest.ContestId,
+              optionId: cvrContestSelection.ContestSelectionId,
+              side:
+                pageIndex === undefined || pageIndex === -1
+                  ? undefined
+                  : pageIndex === 0
```
arsalansufi, line 70: Sanity check: Does this logic work okay for multi-sheet ballots? I can't remember whether we store a separate CVR per sheet or not

apps/admin/backend/schema.sql
```diff
@@ -72,7 +74,8 @@ create table ballot_images (
   image blob not null,
   layout text not null,
   foreign key (cvr_id) references cvrs(id)
-    on delete cascade
+    on delete cascade,
+  unique(cvr_id, is_back)
```
arsalansufi, line 78: Enums might be tricky in SQLite but curious whether you considered something like `front_or_back` (where the possible values are `"front"` and `"back"`) instead of a boolean `is_back`.

Just thinking about cases where we might have to check that a sheet is a front sheet with a slightly confusing `!is_back` condition.

Regardless, I'm okay with `is_back`!

apps/admin/frontend/src/screens/write_ins_transcription_screen.tsx
```diff
@@ -305,6 +306,23 @@ export function WriteInsTranscriptionScreen({
   });
   const imageData = imageDataQuery.data ? imageDataQuery.data[0] : undefined;
 
+  // prefetch the next write-in image
+  const queryClient = useQueryClient();
+  const apiClient = useApiClient();
+  useEffect(() => {
+    const nextAdjudication = adjudications[offset + 1];
+    if (nextAdjudication) {
+      void queryClient.prefetchQuery({
```
arsalansufi, line 315: Sanity check: What happens if the prefetch fails? Does react-query's default error handling take care of this (even with the void)?

libs/api/src/services/admin/types.ts
```diff
@@ -485,25 +484,13 @@ export const WriteInSummaryEntryNonPendingSchema: z.ZodSchema<WriteInSummaryEntr
 /**
  * Write-in image information.
  */
-export interface WriteInImageEntry {
+export interface WriteInImageView {
```
arsalansufi, line 487: I like `View` more than `Entry`! Maybe worth updating the comment for this interface to make it even clearer what it represents?

arsalansufi (approved):
This looks great! Thanks for testing the speeds of the various permutations!

apps/admin/backend/src/cvr_files.ts
```diff
@@ -689,7 +606,79 @@ export async function addCastVoteRecordReport({
...
+            );
+            const imageData = await fs.readFile(imagePath);
+
+            // Load and verify layout data
+            const layoutPath = imagePath
+              .replace(
+                CVR_BALLOT_IMAGES_SUBDIRECTORY,
+                CVR_BALLOT_LAYOUTS_SUBDIRECTORY
+              )
+              .replace(/(jpg|jpeg)$/, 'layout.json');
```
eventualbuddha, line 639: All of this is a bit fragile and may not do the right thing depending on exactly where on the file system this is. For example, if the USB drive gets mounted at a directory with `ballot-images` in the name this will produce an incorrect path. Similarly, if we ever use non-`jpeg` or `jpg` extensions this will not work correctly. There's an argument to be made that we just append `.layout.json` to whatever filename the image has. However, this function does the same thing yours does but works around these issues and also ensures that we only read files within the `reportDirectoryPath`:

```ts
async function loadImageAndAssociatedLayout(
  cvrImageDataLocation: string,
  reportDirectoryPath: string
): Promise<LoadedCvrImageAndLayout> {
  assert(cvrImageDataLocation.startsWith('file:'));
  const relativeImagePath = normalize(
    cvrImageDataLocation.slice('file:'.length)
  );
  assert(relativeImagePath.startsWith(`${CVR_BALLOT_IMAGES_SUBDIRECTORY}/`));
  const parsedBallotImagePath = parse(
    relativeImagePath.slice(`${CVR_BALLOT_IMAGES_SUBDIRECTORY}/`.length)
  );
  const imagePath = join(reportDirectoryPath, relativeImagePath);
  const imageData = await fs.readFile(imagePath);

  // Load layout data
  const layoutPath = join(
    reportDirectoryPath,
    CVR_BALLOT_LAYOUTS_SUBDIRECTORY,
    parsedBallotImagePath.dir,
    `${parsedBallotImagePath.name}.layout.json`
  );
  const layoutJson = await fs.readFile(layoutPath, 'utf8');

  return { imagePath, layoutPath, imageData, layoutJson };
}
```

eventualbuddha (commented):
Looks pretty good. There's a security issue that needs addressing, though.

apps/admin/backend/src/cvr_files.ts
```diff
@@ -689,7 +606,79 @@ export async function addCastVoteRecordReport({
...
+              continue;
+            }
+
+            // Load image data
+            assert(cvrImageData.Location.startsWith('file:'));
+            const imagePath = join(
+              reportDirectoryPath,
+              cvrImageData.Location.slice('file:'.length)
+            );
+            const imageData = await fs.readFile(imagePath);
```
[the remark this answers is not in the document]
arsalansufi: Ooh really good catch 🔐

apps/admin/backend/src/app.ts
```diff
@@ -550,9 +530,11 @@ function buildApi({
 
       return [
         {
-          image: ballotImage.normalized,
+          image: toDataUrl(await loadImageData(image), 'image/jpeg').slice(
+            'data:image/jpeg;base64,'.length
+          ),
```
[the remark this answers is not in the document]
adghayes: OK, two thoughts here I think:
- You're suggesting this could just be the full data URL. I think that's a good idea. I thought about changing it but ended up just leaving it.
- At a later point we could return a URL that points back to the server.  I don't think you're suggesting doing that here but yeah, definitely, we could. I experimented with that and it works, I just didn't see the advantage at the moment but if someone made a compelling case for file storage and served images, I'm in.

apps/admin/backend/src/cvr_files.ts
```diff
@@ -622,32 +620,44 @@ export async function addCastVoteRecordReport({
...
+        fileId,
+        cvr.UniqueId as BallotId,
+        cvrData
+      );
+      if (addCastVoteRecordResult.isErr()) {
+        return err({
+          type: 'ballot-id-already-exists-with-different-data',
+          index: castVoteRecordIndex,
+        });
+      }
```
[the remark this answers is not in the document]
adghayes: Well, it's my existing code that has been moved so it's a fair point. I wasn't worried about it since it's the only type of error that method can return, but I've added a `throwIllegalValue`check in e8d537c77 to keep it safer.

libs/backend/src/cast_vote_record_helpers.ts
adghayes: That's just how we're using the CDF specification (and how it's intended to be used). If you vote on paper, we don't have a piece of text we only have an image (unless we were using OCR). If you vote on a BMD, you don't have an image you have a piece of text, basically a write-in at the "transcribed" stage.

The function suggestion is great, incorporated in 540047482afa3cf9ca602d4c0609e522463943fc.

apps/admin/backend/schema.sql
adghayes: I did think about it and I had a reason but now I can't remember it... Will circle back and change this if I don't think of it.

apps/admin/backend/schema.sql
```diff
@@ -65,6 +67,17 @@ create table cvr_file_entries (
...
 );
 
+create table ballot_images (
+  id varchar(36) primary key,
+  cvr_id varchar(36) not null,
+  is_back boolean not null,
+  image blob not null,
+  layout text not null,
+  foreign key (cvr_id) references cvrs(id)
+    on delete cascade,
```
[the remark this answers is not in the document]
adghayes: What this does is delete the `ballot_images` record if the `cvrs` parent record is deleted, not the reverse. We have the same relationship for write-ins to CVRs and from CVRs to the election. From the SQLite docs:

> CASCADE: A "CASCADE" action propagates the delete or update operation on the parent key to each dependent child key. For an "ON DELETE CASCADE" action, this means that each row in the child table that was associated with the deleted parent row is also deleted. For an "ON UPDATE CASCADE" action, it means that the values stored in each dependent child key are modified to match the new parent key values.

Just tested locally to confirm:
```
sqlite> select count(*) from ballot_images;
50
sqlite> select count(*) from cvrs;
25
sqlite> delete from ballot_images;
sqlite> select count(*) from ballot_images;
0
sqlite> select count(*) from cvrs;
25
```

apps/admin/frontend/src/api.ts
```diff
@@ -239,6 +239,9 @@ export const getWriteInImage = {
   queryKey(input?: GetWriteInImageInput): QueryKey {
     return input ? ['getWriteInImage', input] : ['getWriteInImage'];
   },
+  queryFn(input: GetWriteInImageInput, apiClient: ApiClient) {
+    return apiClient.getWriteInImage(input);
+  },
```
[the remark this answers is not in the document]
adghayes: My thinking was that the bit of code was small enough that it wasn't worth consolidating. I could use `queryFn` in the `useQuery` hook but then that `useQuery` hook would have a distinct pattern from all the rest in the file, and the consistency seemed preferable.

libs/backend/src/cast_vote_record_helper.test.ts
```diff
@@ -0,0 +1,297 @@
+/* eslint-disable vx/gts-identifiers */
```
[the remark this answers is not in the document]
adghayes: Good suggestion, I didn't know that was possible. Updated in 7bed487d0.

apps/admin/backend/src/util/cvrs.ts
```diff
@@ -10,9 +10,9 @@ export function getWriteInVotes(
 }
 
 /**
- * Gets all the write-in options from a CVR.
+ * @deprecated Gets all the write-in options from a legacy CVR.
```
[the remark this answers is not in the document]
adghayes: Function added in `libs/backend/cast_vote_record_helpers.ts` that replaces it, going from the CDF format to a list of write-ins. This is only being used in a test and, once CVRs in legacy format are not retrievable from the backend at all, that test will have to be changed and this function will have no callers. Happy to undeprecate but I see it being removed soon-ish.

benadida:
My main question is: do we need a separate table for the ballot images? Or can they just live in the write-in table? Normalization into a separate table is only really useful if we intend to use that table for other purposes than write-ins.

adghayes:
@benadida my rationale for that normalization is that multiple write-ins can refer to the same ballot image and layout. An alternate approach would be to have the frontImage, frontLayout, backImage, and backLayout, but I leaned away from that for a possibly non-existent reason. We're going to be searching through the `cvrs` table a lot for tallies and I thought that keeping `blob`s out of it might make scanning through database pages scale better, but maybe that's not how SQLite works.

apps/admin/backend/src/cvr_files.ts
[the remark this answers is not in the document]
adghayes: Thanks for catching this!

benadida:
> @benadida my rationale for that normalization is that multiple write-ins can refer to the same ballot image and layout.

Ahah of course. Yes good.

> An alternate approach would be to have the frontImage, frontLayout, backImage, and backLayout on the `cvrs` table, but I leaned away from that for a possibly non-existent reason. We're going to be searching through the `cvrs` table a lot for tallies and I thought that keeping `blob`s out of it might make scanning through database pages scale better, but maybe that's not how SQLite works.

I don't think it would make a big performance difference inherently, except if we have a lot of select * that we'd have to change, but I doubt we have that much. 

That said, i can live with a separate table, because if you did put it in CVRs you'd need to double up all the field names.

I'm left with one suggestion: a better data model, IMO, would have ballot_images primary-keyed on cvr_id, side, where side is constrained to be 'front' or 'back'. Then, the reference from write-in would be cvr_id,side. It already has cvr_id, and encoding the info of which side it's on is a good idea, too, and then boom you've got your reference to the right image. Plus you have an appropriate constraint that a CVR can't have more than one image for each of its sides.

apps/admin/frontend/src/screens/write_ins_transcription_screen.tsx
adghayes: I just tried, experimentally, by returning a rejected promise from the `queryFn` just as `grout` would. I found that if the prefetch fails then I see the error in the console but it doesn't crash anything. I'm not sure what that means w.r.t. React Query's default error handling. Stack trace:
```
  | onError | @ | query.ts:438
  | reject | @ | retryer.ts:112
  | (anonymous) | @ | retryer.ts:172
  | Promise.catch (async) |   |  
  | run | @ | retryer.ts:152
  | createRetryer | @ | retryer.ts:202
  | fetch | @ | query.ts:450
  | fetchQuery | @ | queryClient.ts:453
  | prefetchQuery | @ | queryClient.ts:496
  | (anonymous) | @ | write_ins_transcription_screen.tsx:317
  | invokePassiveEffectCreate | @ | react-dom.development.js:23487
  | callCallback2 | @ | react-dom.development.js:3945
  | invokeGuardedCallbackDev | @ | react-dom.development.js:3994
  | invokeGuardedCallback | @ | react-dom.development.js:4056
  | flushPassiveEffectsImpl | @ | react-dom.development.js:23574
  | unstable_runWithPriority | @ | scheduler.development.js:468
  | runWithPriority$1 | @ | react-dom.development.js:11276
  | flushPassiveEffects | @ | react-dom.development.js:23447
  | (anonymous) | @ | react-dom.development.js:23324
  | workLoop | @ | scheduler.development.js:417
  | flushWork | @ | scheduler.development.js:390
  | performWorkUntilDeadline | @ | scheduler.development.js:157
```

I think this is a good outcome, but not sure.

adghayes:
@benadida that makes sense to me, I'll make a change to that data model. There's a unique constraint already on `cvr_id` and `is_back` but it's better to just make it the primary key and pass fewer ID's around. I'll also take @arsalansufi's suggestion and change `is_back` to `side`, a constrained text field which matches our `Side` type.

apps/admin/backend/src/cvr_files.ts
adghayes: @eventualbuddha a version of this has been added, which addresses your other security comment.

adghayes: @eventualbuddha I actually think this change is not strictly necessary due to the validation logic upstream of this code. The import code finds all the images _actually_ contained in the report (at the correct locations within the report) and during validation, if the an image reference is not in the list of those _actually_ contained in the report, it will fail. I think this is still a great improvement, however, so I'll keep it in.

PR #3327, opened 2023-04-20
fix(mark): Ensure workspace dir exists

jonahkagan:
## Overview

The workspace directory needs to be created if it doesn't exist. Copied this fix from apps/admin.

## Demo Video or Screenshot
N/A
## Testing Plan
Couldn't run the app before, now I can!
## Checklist

### DIFF
--- a/libs/db/src/client.ts
+++ b/libs/db/src/client.ts
@@ -7,7 +7,7 @@ import Database = require('better-sqlite3');
 
 type Database = Database.Database;
 
-const debug = makeDebug('scan:db-client');
+const debug = makeDebug('db-client');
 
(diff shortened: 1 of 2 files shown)


libs/db/src/client.ts
```diff
@@ -7,7 +7,7 @@ import Database = require('better-sqlite3');
 
 type Database = Database.Database;
 
-const debug = makeDebug('scan:db-client');
+const debug = makeDebug('db-client');
```
jonahkagan, line 10: Just something I found while debugging

kshen0 (approved):
Thanks for the fix!

PR #3331, opened 2023-04-20
Enable Java Cards by default and remove *all* the old memory card code

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3307

It's time! We're running a real election with Java Cards, so let's also enable Java Cards by default in dev. And while we're at it, let's remove *all* the old memory card code, from code in the auth lib to the entirety of `services/smartcards` - wowzas. Spring cleaning am I right? 🧹

### Follow-Ups

- [ ] Remove `services/smartcards` references from vxsuite-complete-system (holding off until we're fully ready for NH)
- [x] Re-remind the eng team how to configure Java Cards for local development
- [x] Re-remind the eng team how to use the new card-mocking script in case people have still been using the old Python script (though hopefully the need for this will be shortlived with the upcoming "dock" tool!)

## Testing

- [x] Automated tests
- [x] Manual testing of basic auth flows

## Checklist


apps/admin/backend/src/server.ts
```diff
@@ -70,24 +51,38 @@ export async function start({
...
-    buildApp({
+  if (!resolvedApp) {
+    const auth = new DippedSmartCardAuth({
+      card:
+        isFeatureFlagEnabled(BooleanEnvironmentVariableName.USE_MOCK_CARDS) ||
+        isIntegrationTest()
+          ? new MockFileCard()
+          : new JavaCard(
+              constructJavaCardConfig({ includeCardProgrammingConfig: true })
+            ),
```
arsalansufi, line 70: The core change in this commit is the removal of `new MemoryCard(...)` from all the app backends.

But I also had to shuffle code around such that we don't initialize a real auth instance in tests. Initializing a real auth instance, even if it goes unused, causes tests to fail since we can't actually communicate with a card reader in the test environment.

Had to make similar changes to all the app backends.

apps/central-scan/backend/src/server.ts
```diff
@@ -74,34 +58,45 @@ export async function start({
     resolvedWorkspace = await createWorkspace(workspacePath);
   }
 
-  // clear any cached data
+  // Clear any cached data
   resolvedWorkspace.clearUploads();
+  resolvedWorkspace.store.cleanupIncompleteBatches();
```
arsalansufi, line 63: Not actually new, just moved up from later in this file

apps/scan/frontend/vite.config.ts
```diff
@@ -6,16 +6,15 @@ import setupProxy from './prodserver/setupProxy';
 
 export default defineConfig(async (env) => {
   const workspaceRootPath = join(__dirname, '../../..');
-  const workspacePackages = await getWorkspacePackageInfo(
-    workspaceRootPath
-  );
+  const workspacePackages = await getWorkspacePackageInfo(workspaceRootPath);
```
arsalansufi, line 9: Just some Prettier auto-formatting

.circleci/config.yml
```diff
@@ -857,30 +857,6 @@ jobs:
...
-            make -C services/smartcards coverage
-      - run:
-          name: Typecheck
-          command: |
-            make -C services/smartcards typecheck
-      - run:
-          name: Check Format
-          command: |
-            make -C services/smartcards check-format
-
```
eventualbuddha, line 883: 🎉

apps/admin/backend/src/server.ts
```diff
@@ -70,24 +51,38 @@ export async function start({
...
   /* istanbul ignore next */
-  const resolvedApp =
-    app ??
-    buildApp({
+  if (!resolvedApp) {
+    const auth = new DippedSmartCardAuth({
+      card:
+        isFeatureFlagEnabled(BooleanEnvironmentVariableName.USE_MOCK_CARDS) ||
+        isIntegrationTest()
+          ? new MockFileCard()
```
eventualbuddha, line 67: I see that this isn't new, but I find myself wondering: what happens if someone runs `rm -rf /tmp/*`? It appears that when anything tries to read from the mock card that it'll just throw an exception.

eventualbuddha (approved):

eventualbuddha:
FYI I removed the branch protection rule that required `services/smartcards` to pass CI.

apps/admin/backend/src/server.ts
```diff
@@ -70,24 +51,38 @@ export async function start({
...
+          : new JavaCard(
+              constructJavaCardConfig({ includeCardProgrammingConfig: true })
+            ),
+      config: {
+        allowElectionManagersToAccessUnconfiguredMachines: false,
```
kshen0, line 72: Will this always be `false`? I don't remember running into this setting during the USB programming flow work but it's de facto always enforced. Or put another way, I don't think we allow `allowElectionManagersToAccessUnconfiguredMachines: true` because the frontend is unaware of the setting. Do we need to support that?

kshen0 (approved):

apps/admin/backend/src/server.ts
arsalansufi: Yup you're right - the `getAuthStatus` call will fail and the UI will crash if someone runs `rm -rf /tmp/*` while using mock cards. Let me open a ticket to fix this. Should be as simple as creating the file if it doesn't exist!

arsalansufi:
> FYI I removed the branch protection rule that required services/smartcards to pass CI

Thank you @eventualbuddha!

apps/admin/backend/src/server.ts
arsalansufi: So the frontend doesn't have to be aware of this setting directly - the backend handles it internally, here:

https://github.com/votingworks/vxsuite/blob/c63f0dedef27622c62c4b650225c890f3dbcff2a/libs/auth/src/dipped_smart_card_auth.ts#L617-L622

Because the setting is false in VxAdmin, using an election manager card on an unconfigured VxAdmin will surface this error screen:

https://github.com/votingworks/vxsuite/blob/c63f0dedef27622c62c4b650225c890f3dbcff2a/apps/admin/frontend/src/components/election_manager.tsx#L75-L82

Because the setting is true in VxCentralScan, on the other hand, using an election manager card on an unconfigured VxCentralScan is allowed (and actually required to configure the machine).

Also relevant is the fact that this setting is only exposed for `DippedSmartCardAuth` and not `InsertedSmartCardAuth`, which VxMark and VxScan use. On those two machines, it's understood that election managers can always access unconfigured machines.

Let me know if I can clarify any of this further!

PR #3333, opened 2023-04-21
feat(grout): add `Buffer` support

eventualbuddha:
## Overview

In general we won't use this because we do most data processing on the backend. However, we still have some use cases for sending binary data back and forth. This makes doing so easier.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist

### DIFF
--- a/libs/grout/src/serialization.ts
+++ b/libs/grout/src/serialization.ts
@@ -1,3 +1,4 @@
+import { Buffer } from 'buffer';
 import { err, isResult, ok, Result } from '@votingworks/basics';
 import {
   isArray,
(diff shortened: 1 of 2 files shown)


libs/grout/src/serialization.ts
```diff
@@ -131,6 +140,9 @@ function unserializableError(value: unknown) {
 }
 
 function throwIfUnserializable(value: unknown): void {
+  if (taggers.some((tagger) => tagger.shouldTag(value))) {
+    return;
+  }
   if (
     isObject(value) &&
     !(value instanceof Date) &&
```
jonahkagan, line 148: I think you can remove this line now.

jonahkagan (approved):

Issue #3078, opened 2023-03-07
[admin] Sysadmin "Load Demo Election Definition" button loads separate system settings file


kshen0:
Start with `pollWorkerCardPinsEnabled: boolean`

PR #3345, opened 2023-04-21
save default system settings/polish

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3078
https://github.com/votingworks/vxsuite/issues/3302

Minor changes in VxAdmin related to system settings config.

* Polish: put an "Or" horizontal rule between the first 2 buttons of the unconfigured screen
* Load `DEFAULT_SYSTEM_SETTINGS` when loading an `election.json` only or when loading the default election
  * Currently this is a no-op because VxAdmin doesn't do anything with stored system settings except export it in the ballot package, and we also have logic in the export to fallback to `DEFAULT_SYSTEM_SETTINGS`. But in the future, if we allow configuration, display, or usage of system settings from VxAdmin we'll need to load and store system settings. I'd rather load these explicitly now and be redundant than leave this as a trapdoor for later.

## Demo Video or Screenshot

Screenshot 2023-04-21 at 1 39 52 PM

## Testing Plan
- [x] manually confirm rendering
- [x] add tests for new system settings write
## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/unconfigured_screen.tsx
+++ b/apps/admin/frontend/src/screens/unconfigured_screen.tsx
@@ -15,6 +15,7 @@ import { assert } from '@votingworks/basics';
 // eslint-disable-next-line vx/gts-no-import-export-type
 import type { ConfigureResult } from '@votingworks/admin-backend';
 import { readFileAsyncAsString } from '@votingworks/utils';
+import { DEFAULT_SYSTEM_SETTINGS } from '@votingworks/types';
 import { readInitialAdminSetupPackageFromFile } from '../utils/initial_setup_package';
 import {
   getElectionDefinitionConverterClient,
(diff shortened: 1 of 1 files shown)


apps/admin/frontend/src/screens/unconfigured_screen.tsx
```diff
@@ -149,6 +151,7 @@ export function UnconfiguredScreen(): JSX.Element {
       // TODO: read file content from backend
       const fileContent = await readFileAsyncAsString(file);
       await saveElectionToBackend(fileContent);
+      saveSystemSettingsToBackend(JSON.stringify(DEFAULT_SYSTEM_SETTINGS));
```
arsalansufi, line 154: Not directly related to this PR but as I was passing through all the machines' backend stores, I noticed that we delete old settings before loading new ones in VxCentralScan and VxMark, but not VxAdmin and VxScan.

_VxCentralScan and VxMark_

https://github.com/votingworks/vxsuite/blob/aa21fe78e1b1470613a434f6c5aca1d1d704173c/apps/central-scan/backend/src/store.ts#L184-L190

https://github.com/votingworks/vxsuite/blob/aa21fe78e1b1470613a434f6c5aca1d1d704173c/apps/mark/backend/src/store.ts#L107-L113

_VxAdmin and VxScan_

https://github.com/votingworks/vxsuite/blob/aa21fe78e1b1470613a434f6c5aca1d1d704173c/apps/admin/backend/src/store.ts#L230-L235

https://github.com/votingworks/vxsuite/blob/aa21fe78e1b1470613a434f6c5aca1d1d704173c/apps/scan/backend/src/store.ts#L1241-L1246

Maybe this doesn't matter much since, in order to upload new settings, you have to unconfigure a machine first, which presumably deletes the old settings. But we might want to be consistent.

arsalansufi (approved):
This PR LGTM! Left a comment about something not directly related that I noticed

apps/admin/frontend/src/screens/unconfigured_screen.tsx
kshen0: I agree we should be consistent: added in [b0e56a9](https://github.com/votingworks/vxsuite/pull/3345/commits/b0e56a9b4ab0c3a8bb38cf40f858a78f6ee8d763)

PR #3348, opened 2023-04-24
Add jurisdiction state management to VxCentralScan

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2912

This PR replaces the hardcoded dev jurisdiction in VxCentralScan with proper jurisdiction state management. When the machine is configured, the jurisdiction will be set to the jurisdiction on the election manager card used to auth onto the machine. And when the machine is unconfigured, the jurisdiction will be cleared.

When a jurisdiction is set, the machine will only accept cards from that jurisdiction. And when no jurisdiction is set, the machine will accept cards from all jurisdictions. This logic already exists and lives in the auth lib.

## Testing

- [x] Updated automated tests
- [x] Manually tested that jurisdiction is set on configure
- [x] Manually tested that jurisdiction is cleared on unconfigure

I could definitely add more tests for configuration / unconfiguration, e.g. there currently isn't a test for the `configureFromBallotPackageOnUsbDrive` function, but that's an existing gap. I'm going to prioritize updating the other machines first and will then come back to add more tests if I have time later this week.

## Checklist


apps/central-scan/backend/schema.sql
```diff
@@ -69,3 +69,9 @@ create table system_settings (
   id integer primary key check (id = 1),
   are_poll_worker_card_pins_enabled boolean not null default false
 );
+
+create table jurisdiction (
```
jonahkagan, line 73: Since this changes with the election configuration, could it just live in the same table?

apps/central-scan/backend/src/central_scanner_app.ts
```diff
@@ -168,11 +169,12 @@ function buildApi({
       if (ballotPackageResult.isErr()) {
         return ballotPackageResult;
       }
-
+      assert(isElectionManagerAuth(authStatus));
       const ballotPackage = ballotPackageResult.ok();
       const { electionDefinition, systemSettings, ballots } = ballotPackage;
       assert(systemSettings);
-      store.setElection(electionDefinition.electionData);
```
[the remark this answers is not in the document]
jonahkagan: Ah nice catch!

apps/central-scan/backend/schema.sql
arsalansufi: Updated! https://github.com/votingworks/vxsuite/pull/3348/commits/44c3e6fc8da181184fd8d4a8558bb49ad0419e96

This was one of those changes that ballooned a bit. In adding the jurisdiction to the election table and making it non-nullable, I had to update the store interface for setting the election to support simultaneously setting the jurisdiction. This required updating all consumers of `store.setElection` to use the new `store.setElectionAndJurisdiction`.

This updating led me to some unused handlers (unused as of https://github.com/votingworks/vxsuite/pull/3304), which instead of updating, I completely removed. I timeboxed this and left a TODO on two last handlers I didn't remove because removing them would require updates to trickier end-to-end tests.

jonahkagan: Thanks for doing that cleanup!

apps/central-scan/backend/src/central_scanner_app.ts
```diff
@@ -612,6 +519,8 @@ export async function buildCentralScannerApp({
     }
   );
 
+  // TODO: Update end_to_end_hmpb.test.ts to use configureFromBallotPackageOnUsbDrive and then
+  // remove this, as it's no longer used by the frontend
```
kshen0, line 523: Sorry for missing these in my PR! It seems fragile that the backend e2e tests need to reconstruct the sequence of requests the frontend makes, and I'm wondering if we should deprecate these e2e tests for that reason. But the only alternative I can think of for this type of coverage is an integration test and I know we're not convinced that's the best place to invest. Ticketed in https://github.com/orgs/votingworks/projects/23/views/23?filterQuery=label%3A%22proj%3Asys-config%22&pane=issue&itemId=26607650 and I can pick this up in the next 1-2 weeks if you don't think this is urgent.

kshen0 (approved):

apps/central-scan/backend/src/central_scanner_app.ts
arsalansufi: No worries at all! It also wasn't immediately apparent to me that they were no longer used

> It seems fragile that the backend e2e tests need to reconstruct the sequence of requests the frontend makes

Yeah I agree that we should avoid these tests of longer sequences. They're fragile and harder to reason about than more atomic tests. I do, however, see value in tests of smaller sequences, like the common "POST followed by a GET" test or a "configure followed by a scan" test, where the underlying store isn't mocked and we're actually testing store logic, through the top-level API

Thank you for opening a ticket! I def don't think it's urgent

PR #3354, opened 2023-04-24
CVR Fixture Generation Tweaks

adghayes:
A couple quick additions to CVR fixture generation to more easily make CVR fixtures for performance testing:
1. **Ballot Prefix ID** - useful when wanting to create multiple CVRs for the same election without ID collision
2. **BMD Ballots** - useful when wanting to test without images. We can generate CVRs without images by default, if we don't use the `--includeBallotImages` flag, but the CVRs still have write-ins so they fail VxAdmin validation. Rather than tweaking validation, I wanted to be able to produce BMD ballots. They aren't procedurally valid - they contain overvotes which are not possible on a BMD - but they are formally valid.


kofi-q (approved):
🚀

PR #3355, opened 2023-04-24
refactor(admin): normalize CVR data

adghayes:
_review commit-by commit_

## Overview

We want to de-blobify our data storage when possible. Currently, CVRs are stored as just a blob in the DB. In this PR, we leave the `votes` component of the CVR as a blob but make typed columns for the metadata. Eventually, this will allow us to create tallies faster because we don't have to parse all of the CVRs in order to begin filtering them by metadata.

I did some investigation into how this impacts performance [here](https://docs.google.com/document/d/1kWQbt9kSyylpGkSrSTS-6bmOQqodPEl8rI8pCTiXqXY/edit#heading=h.w879v8uqcdk4). Read Note 1 to get context on Note 2. 

## Testing Plan
- this is mostly a refactor so current tests should cover it
- added a little automated testing for the new `scanner_batches` table
- "manual" performance testing


apps/admin/backend/src/store.ts
```diff
@@ -353,45 +360,80 @@ export class Store {
...
+          ballotStyleId: string;
+          ballotType: string;
+          batchId: string;
+          batchLabel: string;
+          precinctId: string;
+          scannerId: string;
+          sheetNumber: number | null;
+          votes: string;
+        }
+      | undefined;
```
adghayes: Here we're just retrieving all the data about a record in order to compare it to the new record. I wonder if it would be better to just hash all records and compare to a hash? Slightly higher space requirements, maybe slightly lower time requirements (this dupe checking doesn't take very long in perf testing), but this code becomes more readable. Based on those tradeoffs, decided not to pursue the hashing now, but maybe there's a case for it.

apps/admin/backend/src/store.ts
```diff
@@ -651,17 +714,44 @@ export class Store {
...
+        id: entry.id,
+        ballotId: entry.ballotId,
+        electionId,
+        data: JSON.stringify({
+          ...castVoteRecordLegacyMetadata,
+          ...JSON.parse(entry.votes),
+        }),
+        createdAt: convertSqliteTimestampToIso8601(entry.createdAt),
+      };
+    });
```
adghayes: You'll see in the performance testing document that this conversion is slowing us down quite a bit, but it's temporary.

apps/admin/backend/schema.sql
```diff
@@ -42,13 +42,23 @@ create table cvrs (
...
   scanner_id text not null,
   sheet_number integer check (sheet_number is null or sheet_number > 0),
   votes text not null,
   created_at timestamp not null default current_timestamp,
+  foreign key (election_id) references elections(id)
+    on delete cascade,
+  foreign key (batch_id, scanner_id) references scanner_batches(batch_id, scanner_id)
+);
+
+create table scanner_batches (
```
adghayes, line 56: I can definitely see an argument for tossing away this commit and just leaving the batch data not normalized, but i wanted to make it cleaner and it's possible we will want additional batch metadata and don't want to shove it into the CVR table.

apps/admin/backend/src/store.ts
```diff
@@ -790,6 +842,7 @@ export class Store {
         `,
         electionId
       );
+      this.deleteEmptyScannerBatches(electionId);
```
kofi-q, line 845: More out of curiosity: is there a case where we couldn't just wipe all the batches for `electionId` after deleting all CVRs? (i.e. when is a batch non-empty for an election after deleting the CVRs?)

apps/admin/backend/schema.sql
```diff
@@ -42,13 +42,23 @@ create table cvrs (
...
   party_id text,
   precinct_id text not null,
   scanner_id text not null,
   sheet_number integer check (sheet_number is null or sheet_number > 0),
   votes text not null,
   created_at timestamp not null default current_timestamp,
+  foreign key (election_id) references elections(id)
+    on delete cascade,
+  foreign key (batch_id, scanner_id) references scanner_batches(batch_id, scanner_id)
+);
```
kofi-q, line 53: nit: The index creation statements for `cvrs` aren't co-located with the table anymore

apps/admin/backend/schema.sql
```diff
@@ -42,13 +42,23 @@ create table cvrs (
...
+
+create table scanner_batches (
+  batch_id text not null,
+  label text not null,
+  scanner_id text not null,
```
kofi-q, line 58: If all CVRs in a batch have the same scannerId, then maybe we can remove it from the `cvrs` table and join `scanner_batches` whenever we need it?

kofi-q: (And maybe the same applies to `precinct_id` if that makes sense to move into `scanner_batches`?)

apps/admin/backend/schema.sql
```diff
@@ -42,13 +42,23 @@ create table cvrs (
...
+  election_id varchar(36) not null,
+  primary key (batch_id, scanner_id),
```
kofi-q, line 60: For my education: is this because batch IDs are only unique across batches from a single scanner?

(If that's the case, I guess that cancels out my comment about moving `scanner_id` out of `cvrs`, since we'd need it to join `scanner_batches` anyway)

kofi-q (approved):
Looks good! Couple nits, couple questions, nothing blocking

apps/admin/backend/src/store.ts
```diff
@@ -651,17 +714,44 @@ export class Store {
...
 
-    return entries.map((entry) => ({
-      id: entry.id,
-      ballotId: entry.ballotId,
-      electionId,
-      data: entry.data,
-      createdAt: convertSqliteTimestampToIso8601(entry.createdAt),
-    }));
+    const fileMode = this.getCurrentCvrFileModeForElection(electionId);
+    if (fileMode === Admin.CvrFileMode.Unlocked) return [];
```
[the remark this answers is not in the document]
adghayes: good call, 02047043d.

apps/admin/backend/src/store.ts
adghayes: In this case you could do that. When I wrote the helper I was also thinking of the case of removing only a single CVR file which we plan to support: https://votingworks.slack.com/archives/CEL6D3GAD/p1682555456417949.

On that topic, it's a simple change so I'm just going to leave it undone until we build the frontend for it.

apps/admin/backend/schema.sql
adghayes: good nit, d517c0085

apps/admin/backend/schema.sql
adghayes: In the real world, batch IDs should be unique. They are UUID's. I'm never quite sure about this point - do I assume UUIDs are, indeed unique? That the probability of collision is so low I shouldn't build around it? Because if that's the case, then we could have have simple batch_id here.

I also had our fixtures in mind, which do *not* have unique batch ID's, but I can fix that.

apps/admin/backend/schema.sql
adghayes: Good idea on the `scanner_id`. On the `precinct_id`, I think that breaks down with the "All Precincts" use case. We have CVRs from multiple precincts in the same batch and need to keep track of each's precinct individually.

kofi-q: Ah, that makes sense, re: `precinct_id` - thanks for the context!

apps/admin/backend/schema.sql
kofi-q: Oh yeah, if we're using UUIDs, I believe we'll be fine even when combining data across multiple machines

apps/admin/backend/src/store.ts
kofi-q: Ah good point, I remember browsing that thread - makes sense to keep as is

PR #3357, opened 2023-04-25
Update VxMark backend to pull election hash from backend store

arsalansufi:
_Easiest reviewed by commit_

## Overview

Now that VxMark persists the election definition on the backend (thanks Kevin!), we can pull election hash from the backend store for auth calls, instead of having the frontend provide it. This simplifies the code quite a bit and makes VxMark better match the other machines.

Tackling this as a precursor to VxMark jurisdiction state management.

## Testing

- [x] Updated automated tests
- [x] Tested manually

## Checklist

### DIFF
--- a/apps/mark/backend/src/app.ts
+++ b/apps/mark/backend/src/app.ts
@@ -43,8 +43,6 @@ function buildApi(
   return grout.createApi({
     getMachineConfig,
 
-    // TODO: Once election definition has been moved to the backend, no longer require the frontend
-    // to provide election hash to this and other methods that use the auth lib
     getAuthStatus() {
       return auth.getAuthStatus(constructAuthMachineState(workspace));
     },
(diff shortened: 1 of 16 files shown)


apps/mark/backend/src/app.ts
```diff
@@ -47,42 +45,41 @@ function buildApi(
 
     // TODO: Once election definition has been moved to the backend, no longer require the frontend
     // to provide election hash to this and other methods that use the auth lib
```
kshen0, line 47: Delete this comment?

apps/mark/backend/test/app_helpers.ts
```diff
@@ -42,3 +48,31 @@ export function createApp(): MockAppContents {
     server,
   };
 }
+
+export async function configureApp(
```
kshen0, line 52: Thanks for adding this!

kshen0 (approved):
LGTM, minor nit that I think we can delete an old comment

apps/mark/backend/src/app.ts
arsalansufi: Good catch! https://github.com/votingworks/vxsuite/pull/3357/commits/74d6b15c20b323502aa988053ad2b9e115b8f22f

PR #3364, opened 2023-04-25
admin: add tools for performance testing & add index to improve performance of cvr file loading

adghayes:
_review commit by commit_

## Overview

Adds some tools to `admin-backend` to make it easier to test performance. Including:
- ability to take backups of the workspace from tests, and load that workspace in other tests
- temporary directory cleanup after tests
- performance timer

Adds current version of performance tests. These are tests which are meant to be run locally and manually, so I wasn't sure if I should commit them? But it's also been frustrating switching between branches to use the various pieces. @kofi-q let me know what you think of including it all.

Finally, adds two indices to the `cvrs` table which dramatically increases the speed of adding CVR files: 
Adding CVR Files (1)
This does increase the size of the database but by a manageable 10%. See [slack discussion.](https://votingworks.slack.com/archives/CEL6D3GAD/p1682360680789959)

## Testing Plan

Most of this is not tested as it is already testing code.


apps/admin/backend/src/perf.test.ts
```diff
@@ -0,0 +1,70 @@
...
+} from '../test/app';
+import { getBackupPath, takeBackup } from '../test/backups';
+import { getPerformanceTimer } from '../test/timer';
+
+// depending on the test conditions, may need to increase this
+jest.setTimeout(600000);
+
+const electionDefinition = electionMinimalExhaustiveSampleDefinition;
+
+test.skip('load CVR files', async () => {
```
kofi-q, line 19: nit: Is it worth moving this file elsewhere, like `admin/backend/perf` or `admin/backend/scripts` and having a manual way to run it (maybe a separate `pnpm test:perf` command)?
Might not be clear at first glance why these are being skipped (but +1 to committing the perf test code for reuse)

apps/admin/backend/src/util/workspace.ts
```diff
@@ -22,16 +16,11 @@ export interface Workspace {
 export function createWorkspace(root: string): Workspace {
   const resolvedRoot = resolve(root);
   const dbPath = join(resolvedRoot, 'data.db');
-  const uploadsPath = join(resolvedRoot, 'uploads');
```
kofi-q, line 25: Is this just old, unused code?

apps/admin/backend/src/perf.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  takeBackup({
+    workspace,
+    backupName: 'performance',
+  });
+});
+
+test.skip('tally performance', () => {
+  const timer = getPerformanceTimer();
+  const { workspace, auth } = buildTestEnvironment(
+    getBackupPath('performance')
```
kofi-q, line 50: Would it work instead to do the common test env setup in a `beforeAll()` block? Or is this meant to facilitate persisting across multiple runs of the test file?

apps/admin/backend/test/app.ts
```diff
@@ -163,10 +162,17 @@ export async function configureMachine(
 }
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-export function buildTestEnvironment() {
+export function buildTestEnvironment(workspaceRoot?: string) {
   const logger = fakeLogger();
   const auth = buildMockDippedSmartCardAuth();
-  const workspace = createWorkspace(tmp.dirSync().name);
+  const resolvedWorkspaceRoot =
+    workspaceRoot ??
```
kofi-q, line 169: Should this be `||` instead of `??`? I think this is allowing the workspace root to be set to an empty string, but not sure if that's valid value

kofi-q (commented):
Generally looks good! One open question about the need for test env backups

apps/admin/backend/test/app.ts
adghayes: That's a good point! I didn't realize those two operators had different behavior like that.

apps/admin/backend/src/util/workspace.ts
adghayes: AFAICT. @eventualbuddha what was the `uploads` subdirectory in VxAdmin backend supposed to be used for?

apps/admin/backend/src/perf.test.ts
adghayes: yeah, I think that would be better. I looked into it briefly but wasn't sure how to set up `jest` to run that way. it must be possible though, will make it work.

apps/admin/backend/src/perf.test.ts
adghayes: The main motivation for the backup is that loading up the DB can take a long time. If I want 1,000,000 votes in the DB, that can take almost 20 minutes. I want to be able to do that once and then run the same test file multiple times, either to have a sample size larger than 1, to try variations in the code, or to repeatedly test destructive actions.

kofi-q: Ah, this feature makes sense, then - thanks for explaining

apps/admin/backend/test/app.ts
adghayes: 9e2bbcefb96857da244e9ef3e86d1b1c786c3649

apps/admin/backend/src/perf.test.ts
kofi-q: Sounds good - can definitely be a followup later down the line, though. I think this is fine for now, to get things moving

kofi-q (approved):
🚀

apps/admin/backend/src/perf.test.ts
adghayes: I figured out how to do the `jest` regex's the right way, and it's a big improvement! Thanks for the prod, 6b703d5.

kofi-q: Oh nice! 🙌

apps/admin/backend/src/util/workspace.ts
eventualbuddha: Probably where we temporarily stored things uploaded via express+multer. Old code.

PR #3367, opened 2023-04-26
feat(basics): add early-return for `Result` functions

eventualbuddha:
_(I recommend reviewing with ["Hide whitespace" enabled](https://github.com/votingworks/vxsuite/pull/3367/files?diff=split&w=1))_

## Overview
Rust has the [`?` operator](https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator) to return early from functions if the `Result` operand is `Err`. This is convenient, and the lack of it with our `Result` type means a lot of extra error checks with early returns. This is annoying boilerplate, but it also makes complete code coverage difficult to achieve since you need to write tests that cover what is essentially just propagating an error.

A note about the design of this API: it uses JavaScript exceptions. This is not ideal but is unavoidable since it's the only way to return early from a function without `return`. Additionally, you might ask why `resultBlock` has a function to be passed to `or` rather than simply having `or` (or some other `Result` method) throw the exception. The reason for that is to ensure the errors propagated have the right type. Otherwise a thrown `Err` might not be compatible with the type expected by the containing function.

## Demo Video or Screenshot
```ts
// before
function doSomething(): Result<string, Error> {
  const valueResult = doSomethingThatMightFail();
  if (valueResult.isErr()) {
    return valueResult;
  }
  const value = valueResult.ok();
  const value2Result = doSomethingElseThatMightFail(value);
  if (value2Result.isErr()) {
    return value2Result;
  }
  const value2 = value2Result.ok();
  return anotherThingThatMightFail(value2);
}

// after
function doSomething(): Result<string, Error> {
  return resultBlock((fail) => {
    const value = doSomethingThatMightFail().okOrElse(fail);
    const value2 = doSomethingElseThatMightFail(value).okOrElse(fail);
    return anotherThingThatMightFail(value2);
  });
}
```

## Testing Plan
Automated.

## Checklist

### DIFF
--- a/libs/basics/src/result.ts
+++ b/libs/basics/src/result.ts
@@ -50,11 +50,10 @@ class Ok<T> {
   }
 
   /**
-   * Returns the contained value. For use with `resultBlock` or
-   * `asyncResultBlock`.
+   * Returns the contained value.
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  or<E>(ret: (error: E) => void): T {
+  okOrElse<E>(fn: (error: E) => void): T {
     return this.value;
   }
 
@@ -140,11 +139,12 @@ class Err<E> {
   }
 
   /**
-   * Calls `ret` with the contained error. For use with `resultBlock` or
-   * `asyncResultBlock`.
+   * Calls the given callback with the contained error.
    */
-  or(ret: (error: E) => never): never {
-    ret(this.error);
+  okOrElse(fn: (error: E) => never): never;
+  okOrElse<T>(fn: (error: E) => T): T;
+  okOrElse<T>(fn: (error: E) => T): T {
+    return fn(this.error);
   }
 
   /**
--- a/libs/basics/src/result.ts
+++ b/libs/basics/src/result.ts
@@ -50,6 +50,15 @@ class Ok<T> {
   }
 
   /**
+   * Returns the contained value. For use with `resultBlock` or
+   * `asyncResultBlock`.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  or<E>(ret: (error: E) => void): T {
+    return this.value;
+  }
+
+  /**
    * Returns the contained value.
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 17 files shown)


libs/basics/src/result.ts
```diff
@@ -215,3 +232,87 @@ export type Result<T, E> = Ok<T> | Err<E>;
...
+ * @example
+ *
+ * ```ts
+ * function doSomething(): Result<string, Error> {
+ *   return resultBlock((ret) => {
+ *     const value = doSomethingThatMightFail().or(ret);
+ *     const value2 = doSomethingElseThatMightFail(value).or(ret);
+ *     return anotherThingThatMightFail(value2);
+ *   });
+ * }
```
jonahkagan, line 261: I think it might be worth thinking about making `or` work outside of a `resultBlock` so that we can have a better mental model of what it does. An idea: rename to `okOrElse` or `unwrapOrElse` and let the user provide a default value or function that returns a default value.

I would also suggest maybe changing the naming convention for `ret` to something a bit more descriptive (maybe `bail` or `fail`).

Putting those together:

```ts
return resultBlock((fail) => {
  const value = doSomethingThatMightFail().okOrElse(fail);
  const value2 = doSomethingElseThatMightFail(value).okOrElse(fail);
  return anotherThingThatMightFail(value2);
});
```

libs/basics/src/result.ts
```diff
@@ -215,3 +232,87 @@ export type Result<T, E> = Ok<T> | Err<E>;
...
+ * function doSomething(): Promise<Result<string, Error>> {
+ *   return asyncResultBlock(async (ret) => {
+ *     const value = (await doSomethingThatMightFail()).or(ret);
+ *     const value2 = (await doSomethingElseThatMightFail(value)).or(ret);
+ *     return anotherThingThatMightFail(value2);
+ *   });
+ * }
+ * ```
+ */
+export async function asyncResultBlock<T, E>(
```
jonahkagan, line 302: Is it possible to write one version of `resultBlock` that supports sync and async based on the type of the callback given?

jonahkagan (approved):
Cool idea! I think it may turn out a bit easier to think about than `mapResult`. Let's try it and see.

libs/basics/src/result.ts
eventualbuddha: Yeah, I was trying to keep it short, but I think you're right.

eventualbuddha: Okay, this is now changed.

libs/basics/src/result.ts
eventualbuddha: I tried this and could not get it to work. It's straightforward enough to build something that runs correctly, but making it so TypeScript can properly type check it seems hard. No matter what I tried it would always end up inferring the type parameters as `unknown`.

PR #3368, opened 2023-04-26
[VxScan] Use the new shared DisplaySettings component

kofi-q:
## Overview

Integrating the `DisplaySettings` component into VxScan to allow voters to change text size and color contrast.

- Added `DisplaySettingsButton` which navigates to  `/display-settings`
- Extracted a `ScreenHeader` component which renders the ballot counter and the `DisplaySettingsButton`
- Added a react-router route to a new  `DisplaySettingsScreen`, which is just a thin wrapper around `DisplaySettings` that navigates back to the previous URL when `onClose` is fired. Initially wanted to explicitly navigate to `/`, but that's incompatible with the preview dashboard, which is rendered at `/preview`.

## Demo Video or Screenshot

https://user-images.githubusercontent.com/264902/235526675-8945a584-1956-4e77-a272-a549343647d2.mov

## Testing Plan
- Unit tests to verify routing for the new button and screen
- Visual testing via the preview dashboard

## Checklist


apps/scan/frontend/src/components/display_settings_button.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Paths } from '../constants';
+
+const LabelContainer = styled.span`
+  align-items: center;
+  display: flex;
+  flex-wrap: nowrap;
+  font-weight: ${(p) => p.theme.sizes.fontWeight.semiBold};
+  gap: 0.5rem;
+  text-align: left;
+  width: min-content;
```
jonahkagan, line 15: Curious what this would look like with the button text all on one line. Maybe it would seem less intrusive on the screen?

jonahkagan (approved):

mattroe:
@kofi-q placement works well for the current VxScan prototype with the screen below paper infeed, but we may want to adjust this if we revisit overall scanner hardware.

Re: the button itself, the only problem is that this doesn't exactly convey what is available to change. I don't think we should expect the user to know what display settings are available, so I'd like to communicate that text size & contrast/color are specifically available for adjusting somehow. 

One possible solution is just keep the same display icon and changing the copy to be `Color & Size`, `Font Size/Color`, or other similar more specific copy.

apps/scan/frontend/src/components/display_settings_button.tsx
kofi-q: Yeah, I had it that way previously and it looked odd at the larger sizes (mostly weird wrapping at the largest size) - but thought about it a bit over the weekend and want to try again with some workarounds (esp. considering people aren't going to be flipping back and forth between sizes)
Will add a screen grab of the the single line option for comparison and try out Roe's copy suggestions as well

kofi-q:
> Re: the button copy/iconography, the only problem is that this doesn't exactly convey what is available to change. I don't think we should expect the user to know what display settings are available, so I'd like to communicate that text size & contrast/color are specifically available for adjusting somehow.

> One possible solution is just keep the same display icon and changing the copy to be Color & Size, Font Size/Color, or other similar more specific copy.

Agreed - I like the idea of trying to convey color/size options. Will try out some options and grab screenshots for review

kofi-q:
Oh, good call on "zoom"! And +1 to more action-oriented copy

Issue #3296, opened 2023-04-14
USB Encapsulation & Mocking Capabilities


adghayes:
We have repeatedly hit roadblocks when trying to develop with USB drives, in particular when we need to mock them. We interface with USB drives in two places: `kiosk-browser` and in the app backends. In both those places, we are querying system files directly (like `/dev/disk/by-id`) and so have trouble mocking. We are completely unable to mock USB drives in integration tests and our approach in development (in `libs/usb-mocking`) is limited, clunky, and requires special permissions. Finally, we'd like to move away from any of this happening via `kiosk-browser`.

In order to overcome these problems, we need an encapsulated `libs/usb` similar to `libs/auth` that would allow us to interact with the USB, maintain contextual state (e.g. "ejected" status), and inject mocking. The project would include moving apps away from the `kiosk-browser` USB APIs and instead accessing that information through the backend, much like `Auth`. 

It's difficult to estimate the work here but I think it is 3 weeks plus-minus of a single developer's time. They would have to:
- move USB scripts from `vxsuite-complete-system` to `vxsuite`
- build an abstraction layer for the USB
- build a real USB implementation using the USB scripts
- build a mock USB implementation which would probably mount something in `/tmp`
- wire up the USB implementation to the backend, rewrite testing
- scrap the `useUsbDrive` react hook, get USB information on the frontends from the backend, and rewrite frontend testing accordingly
- update permissions in `vxsuite-complete-system` build scripts such that the USB scripts that are now in `vxsuite` can be run in a locked down version of the software

adghayes:
Next: #3447

adghayes:
Completed

PR #3378, opened 2023-04-27
Create libs/usb-drive, integrate into VxScan

jonahkagan:
## Overview

_Review by commit_

First step towards https://github.com/votingworks/vxsuite/issues/3296

Creates a new library libs/usb-drive that interfaces with a single USB drive (following the logic from kiosk-browser). Then uses that library to replace existing USB code in apps/scan. Also adds a mock for use in backend tests.

## Demo Video or Screenshot

N/A

## Testing Plan
- Updated existing tests
- New tests for library
- Some basic manual testing

## Checklist

### DIFF
--- a/apps/scan/backend/src/app_usb_drive.test.ts
+++ b/apps/scan/backend/src/app_usb_drive.test.ts
@@ -1,0 +1,23 @@
+import { withApp } from '../test/helpers/custom_helpers';
+
+test('getUsbDriveStatus', async () => {
+  await withApp({}, async ({ apiClient, mockUsbDrive }) => {
+    mockUsbDrive.removeUsbDrive();
+    await expect(apiClient.getUsbDriveStatus()).resolves.toEqual({
+      status: 'no_drive',
+    });
+    mockUsbDrive.insertUsbDrive({});
+    await expect(apiClient.getUsbDriveStatus()).resolves.toEqual({
+      status: 'mounted',
+      mountPoint: expect.any(String),
+      deviceName: 'mock-usb-drive',
+    });
+  });
+});
+
+test('ejectUsbDrive', async () => {
+  await withApp({}, async ({ apiClient, mockUsbDrive }) => {
+    mockUsbDrive.usbDrive.eject.expectCallWith().resolves();
+    await expect(apiClient.ejectUsbDrive()).resolves.toBeUndefined();
+  });
+});
--- a/apps/scan/backend/test/helpers/shared_helpers.ts
+++ b/apps/scan/backend/test/helpers/shared_helpers.ts
@@ -1,6 +1,6 @@
 import { InsertedSmartCardAuthApi } from '@votingworks/auth';
 import { ok } from '@votingworks/basics';
-import { MockUsb, createBallotPackageZipArchive } from '@votingworks/backend';
+import { createBallotPackageZipArchive } from '@votingworks/backend';
 import { electionFamousNames2021Fixtures } from '@votingworks/fixtures';
 import * as grout from '@votingworks/grout';
 import {
@@ -14,6 +14,7 @@ import {
   singlePrecinctSelectionFor,
 } from '@votingworks/utils';
 import waitForExpect from 'wait-for-expect';
+import { MockUsbDrive } from '@votingworks/usb-drive';
 import { Api } from '../../src/app';
 import { PrecinctScannerInterpreter } from '../../src/interpret';
 import {
@@ -54,7 +55,7 @@ export async function waitForStatus(
 /**
  * configureApp is a testing convenience function that handles some common configuration of the VxScan app.
  * @param apiClient - a VxScan API client
- * @param mockUsb - a mock USB
+ * @param mockUsbDrive - a mock USB drive
  * @param options - an object containing optional arguments
  * @param options.mockAuth - a mock InsertedSmartCardAuthApi. Passing this will automatically
  *                           create a mock that auths the user as an election manager of the same
@@ -62,7 +63,7 @@ export async function waitForStatus(
  */
 export async function configureApp(
   apiClient: grout.Client<Api>,
-  mockUsb: MockUsb,
+  mockUsbDrive: MockUsbDrive,
   {
     ballotPackage = electionFamousNames2021Fixtures.electionJson.toBallotPackage(),
     precinctId,
@@ -85,7 +86,7 @@ export async function configureApp(
     );
   }
 
-  mockUsb.insertUsbDrive({
+  mockUsbDrive.insertUsbDrive({
     'ballot-packages': {
       'test-ballot-package.zip': await createBallotPackageZipArchive(
         ballotPackage
--- a/apps/scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
+++ b/apps/scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
@@ -1,14 +1,16 @@
 import React from 'react';
 import {
   UnconfiguredElectionScreen,
-  UsbDriveStatus,
-  useExternalStateChangeListener,
+  useQueryChangeListener,
 } from '@votingworks/ui';
 import { ScreenMainCenterChild } from '../components/layout';
-import { configureFromBallotPackageOnUsbDrive } from '../api';
+import {
+  configureFromBallotPackageOnUsbDrive,
+  getUsbDriveStatus,
+  legacyUsbDriveStatus,
+} from '../api';
 
 interface Props {
-  usbDriveStatus: UsbDriveStatus;
   isElectionManagerAuth: boolean;
 }
 
@@ -16,21 +18,27 @@ interface Props {
  * UnconfiguredElectionScreenWrapper wraps the shared UnconfiguredElectionScreen component
  * with VxScan-specific logic (primarily calls to the VxScan API)
  */
-export function UnconfiguredElectionScreenWrapper(props: Props): JSX.Element {
-  const { usbDriveStatus, isElectionManagerAuth } = props;
+export function UnconfiguredElectionScreenWrapper(
+  props: Props
+): JSX.Element | null {
+  const { isElectionManagerAuth } = props;
 
+  const usbDriveStatusQuery = getUsbDriveStatus.useQuery();
   const configureMutation = configureFromBallotPackageOnUsbDrive.useMutation();
-  useExternalStateChangeListener(usbDriveStatus, (newUsbDriveStatus) => {
-    if (newUsbDriveStatus === 'mounted') {
+  // TODO move watching for USB drive to configure to the backend
+  useQueryChangeListener(usbDriveStatusQuery, (newUsbDriveStatus) => {
+    if (newUsbDriveStatus.status === 'mounted') {
       configureMutation.mutate();
     }
   });
   const error = configureMutation.data?.err();
 
+  if (!usbDriveStatusQuery.isSuccess) return null;
+
   return (
     <ScreenMainCenterChild>
       <UnconfiguredElectionScreen
-        usbDriveStatus={usbDriveStatus}
+        usbDriveStatus={legacyUsbDriveStatus(usbDriveStatusQuery.data)}
         isElectionManagerAuth={isElectionManagerAuth}
         backendConfigError={error}
         machineName="VxScan"
@@ -41,10 +49,5 @@ export function UnconfiguredElectionScreenWrapper(props: Props): JSX.Element {
 
 /* istanbul ignore next */
 export function DefaultPreview(): JSX.Element {
-  return (
-    <UnconfiguredElectionScreenWrapper
-      usbDriveStatus="absent"
-      isElectionManagerAuth
-    />
-  );
+  return <UnconfiguredElectionScreenWrapper isElectionManagerAuth />;
 }
--- a/libs/usb-drive/package.json
+++ b/libs/usb-drive/package.json
@@ -1,0 +1,64 @@
+{
+  "name": "@votingworks/usb-drive",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Library for interacting with a USB drive",
+  "license": "GPL-3.0",
+  "author": "VotingWorks Eng <eng@voting.works>",
+  "main": "build/index.js",
+  "types": "build/index.d.js",
+  "scripts": {
+    "build": "tsc --build tsconfig.build.json",
+    "clean": "rm -rf build *.tsbuildinfo",
+    "lint": "pnpm type-check && eslint .",
+    "lint:fix": "pnpm type-check && eslint . --fix",
+    "pre-commit": "lint-staged",
+    "test": "is-ci test:ci test:watch",
+    "test:ci": "TZ=UTC pnpm build && pnpm test:coverage --reporters=default --reporters=jest-junit --maxWorkers=7",
+    "test:coverage": "TZ=UTC jest --coverage",
+    "test:watch": "TZ=UTC jest --watch",
+    "type-check": "tsc --build"
+  },
+  "lint-staged": {
+    "*.+(css|graphql|json|less|md|mdx|sass|scss|yaml|yml)": [
+      "prettier --write"
+    ],
+    "*.+(js|jsx|ts|tsx)": [
+      "eslint --quiet --fix"
+    ],
+    "package.json": [
+      "sort-package-json"
+    ]
+  },
+  "dependencies": {
+    "@votingworks/basics": "workspace:*",
+    "debug": "^4.3.2"
+  },
+  "devDependencies": {
+    "@types/debug": "^4.1.7",
+    "@types/jest": "^27.0.3",
+    "@types/node": "^18.11.18",
+    "@types/styled-components": "^5.1.9",
+    "@typescript-eslint/eslint-plugin": "^5.37.0",
+    "@typescript-eslint/parser": "^5.37.0",
+    "esbuild": "^0.17.11",
+    "esbuild-runner": "^2.2.2",
+    "eslint": "8.23.1",
+    "eslint-config-prettier": "^8.5.0",
+    "eslint-import-resolver-node": "^0.3.6",
+    "eslint-plugin-import": "^2.26.0",
+    "eslint-plugin-jest": "^26.1.5",
+    "eslint-plugin-prettier": "^4.2.1",
+    "eslint-plugin-vx": "workspace:*",
+    "is-ci-cli": "^2.2.0",
+    "jest": "^27.3.1",
+    "jest-junit": "^14.0.1",
+    "jest-watch-typeahead": "^0.6.4",
+    "lint-staged": "^11.0.0",
+    "prettier": "^2.6.2",
+    "sort-package-json": "^1.50.0",
+    "ts-jest": "^27.0.7",
+    "typescript": "4.6.3"
+  },
+  "packageManager": "pnpm@7.24.3"
+}
--- a/libs/usb-drive/package.json
+++ b/libs/usb-drive/package.json
@@ -39,11 +39,11 @@
   "devDependencies": {
     "@types/debug": "^4.1.7",
     "@types/jest": "^27.0.3",
-    "@types/node": "^18.11.18",
+    "@types/node": "16.18.23",
     "@types/styled-components": "^5.1.9",
     "@types/tmp": "^0.2.3",
-    "@typescript-eslint/eslint-plugin": "^5.37.0",
-    "@typescript-eslint/parser": "^5.37.0",
+    "@typescript-eslint/eslint-plugin": "5.37.0",
+    "@typescript-eslint/parser": "5.37.0",
     "esbuild": "^0.17.11",
     "esbuild-runner": "^2.2.2",
     "eslint": "8.23.1",
@@ -58,10 +58,10 @@
     "jest-junit": "^14.0.1",
     "jest-watch-typeahead": "^0.6.4",
     "lint-staged": "^11.0.0",
-    "prettier": "^2.6.2",
+    "prettier": "2.6.2",
     "sort-package-json": "^1.50.0",
     "ts-jest": "^27.0.7",
     "typescript": "4.6.3"
   },
-  "packageManager": "pnpm@7.24.3"
+  "packageManager": "pnpm@8.1.0"
 }
--- a/libs/usb-drive/package.json
+++ b/libs/usb-drive/package.json
@@ -32,13 +32,16 @@
   },
   "dependencies": {
     "@votingworks/basics": "workspace:*",
-    "debug": "^4.3.2"
+    "@votingworks/test-utils": "workspace:*",
+    "debug": "^4.3.2",
+    "tmp": "^0.2.1"
   },
   "devDependencies": {
     "@types/debug": "^4.1.7",
     "@types/jest": "^27.0.3",
     "@types/node": "^18.11.18",
     "@types/styled-components": "^5.1.9",
+    "@types/tmp": "^0.2.3",
     "@typescript-eslint/eslint-plugin": "^5.37.0",
     "@typescript-eslint/parser": "^5.37.0",
     "esbuild": "^0.17.11",
--- a/libs/usb-drive/src/mock_usb_drive.ts
+++ b/libs/usb-drive/src/mock_usb_drive.ts
@@ -2,7 +2,7 @@
 import { MockFunction, mockFunction } from '@votingworks/test-utils';
 import { execSync } from 'child_process';
 import { join } from 'path';
-import { mkdirSync, writeFileSync } from 'fs';
+import { mkdirSync, readdirSync, rmSync, writeFileSync } from 'fs';
 import tmp from 'tmp';
 import { Buffer } from 'buffer';
 import { UsbDrive } from './usb_drive';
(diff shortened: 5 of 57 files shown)


libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+export function detectUsbDrive(): UsbDrive {
+  let didEject = false;
+
+  return {
+    async status(): Promise<UsbDriveStatus> {
+      let deviceInfo = await getUsbDriveStatus();
+      if (!deviceInfo) return { status: 'no_drive' };
+      // Automatically mount the drive if it's not already mounted
+      // TODO if the drive is mounted at a different mount point than our
+      // default VX mount point, do we want to unmount and remount it?
```
jonahkagan, line 131: Open question

libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+      await unmountUsbDrive(deviceInfo.mountpoint);
+      didEject = true;
+    },
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    async format(): Promise<void> {
+      throw new Error('Not implemented');
+    },
+  };
+}
```
jonahkagan, line 160: I omitted the logic about unmounting phantom USB drives: https://github.com/votingworks/kiosk-browser/blob/main/src/ipc/get-usb-drive-info.ts#L83. Is that ok?

libs/usb-drive/src/usb_drive.ts
```diff
@@ -103,6 +107,7 @@ async function unmountUsbDrive(mountPoint: string): Promise<void> {
   await exec('umount', [mountPoint]);
 }
 
+// TODO check format?
```
jonahkagan: stay tuned for a later commit

apps/scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
```diff
@@ -1,36 +1,44 @@
...
+export function UnconfiguredElectionScreenWrapper(
+  props: Props
+): JSX.Element | null {
+  const { isElectionManagerAuth } = props;
 
+  const usbDriveStatusQuery = getUsbDriveStatus.useQuery();
   const configureMutation = configureFromBallotPackageOnUsbDrive.useMutation();
-  useExternalStateChangeListener(usbDriveStatus, (newUsbDriveStatus) => {
-    if (newUsbDriveStatus === 'mounted') {
+  // TODO move watching for USB drive to configure to the backend
```
jonahkagan, line 28: Probably gonna leave this for later to keep changes minimal

libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+async function mountUsbDrive(devicePath: string): Promise<void> {
+  debug(`Mounting USB drive ${devicePath} at ${VX_MOUNT_POINT}`);
+  await fs.mkdir(VX_MOUNT_POINT, { recursive: true });
+  await exec('mount', [
+    '-w',
+    '-o',
+    'umask=000,nosuid,nodev,noexec',
+    devicePath,
+    VX_MOUNT_POINT,
+  ]);
```
jonahkagan, line 101: Ok to call these commands directly instead of using the `mount-usb.sh` app scripts from vxsuite-complete-system?

libs/usb-drive/src/usb_drive.ts
adghayes: When I first moved the mounting logic from `kiosk-browser` to a script called by `kiosk-browser`, I omitted this step and then found it to be problematic. The case comes up quite a lot and, as an OS problem, I think the way we were handling is the best we can do.

Just diving into a review now.

libs/usb-drive/src/usb_drive.ts
adghayes: We have a preference for using scripts, see @benadida's comment [here](https://github.com/votingworks/kiosk-browser/pull/142#issuecomment-1382585164). I think this case is analogous but not confident.

To make the above work, would you not not need to call `mount` with `sudo`? And to call `mount` with `sudo`, you'd have to give the running user complete, no-password access to `sudo mount` via `sudoers.etc` (there are ways to restrict what arguments a command can be called with in the `sudoers.etc` file, but not a way to allow a dynamic argument like the device path we'll be passing in). We'd prefer to only give the running user access to a script so, if compromised, they could not mount whatever they like.

libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+  await exec('umount', [mountPoint]);
+}
+
+export function detectUsbDrive(): UsbDrive {
+  let didEject = false;
+
+  return {
+    async status(): Promise<UsbDriveStatus> {
+      let deviceInfo = await getUsbDriveStatus();
+      if (!deviceInfo) return { status: 'no_drive' };
```
adghayes: I'm guessing we'd `didEject = false` here

libs/usb-drive/src/mock_usb_drive.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    assertComplete() {
+      for (const method of Object.values(usbDrive)) {
+        method.assertComplete();
+      }
+    },
+
+    insertUsbDrive(contents: MockFileTree) {
+      mockUsbTmpDir?.removeCallback();
+      mockUsbTmpDir = tmp.dirSync({ unsafeCleanup: true });
```
adghayes, line 72: It's worth checking if `unsafeCleanup` actually works. I think I checked once and found it did not, because of the way `jest` exits (or fails to exit).

libs/usb-drive/src/usb_drive.ts
jonahkagan: Thanks for the context, I figured there was a reason.

Should these scripts continue to live in vxsuite-complete-system? If so, how do we support accessing them in dev? Maybe we have a dev version of the scripts in vxsuite?

adghayes: I'm not sure. 

The way `kiosk-browser` supports it in dev is just assuming there's a `vxsuite-complete-system` repo adjacent to a `kiosk-browser` repo, which could be finicky but we've had no problems. We'd still have to figure out and test pathing so that the scripts are accessible when locked down.

I think it would probably be better to move them into `vxsuite` to make pathing to it intuitive and have all the code colocated. There are a couple consumers in `vxsuite-complete-system` but they could use a path to `vxsuite` or copies of the scripts. I'd need to refresh my memory on what happens to `vxsuite` code when we lock down, and how to path to it (from `sudoers.etc`).

jonahkagan: Colocating the USB mount/unmount scripts in vxsuite would definitely make development easier. If that's workable for prod, I'd support that.

@amcmanus Do you have any thoughts on this?

libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+
+const exec = promisify(execFile);
+
+const debug = makeDebug('usb-drive');
+
+export type UsbDriveStatus =
+  | { status: 'no_drive' }
+  | { status: 'mounted'; mountPoint: string }
+  | { status: 'ejected' }
+  | { status: 'error'; reason: 'bad_format' };
```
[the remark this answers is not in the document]
jonahkagan: I don't think so and here's my reasoning:
- I don't think we have a need to tell users when mounting is in progress. Pretty sure we don't really do this today. Though if you have a diff impression I'm totally open to hearing it. I could be convinced it's worth telling users about if it's particularly slow - like "hey we saw you inserted a USB drive and we're checking it out"
- We can show an "ejecting in progress" state to users by waiting for the promise to complete

libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+  return {
+    async status(): Promise<UsbDriveStatus> {
+      let deviceInfo = await getUsbDriveStatus();
+      if (!deviceInfo) return { status: 'no_drive' };
+      // Automatically mount the drive if it's not already mounted
+      // TODO if the drive is mounted at a different mount point than our
+      // default VX mount point, do we want to unmount and remount it?
+      if (!deviceInfo.mountpoint && !didEject) {
+        await mountUsbDrive(deviceInfo.path);
+        deviceInfo = assertDefined(await getUsbDriveStatus());
```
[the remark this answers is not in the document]
jonahkagan: Good question. I don't know the answer. Do you have any prior evidence that this will be a problem?

libs/usb-drive/src/usb_drive.ts
jonahkagan: I don't totally understand what you're getting at here. Can you elaborate?

libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+    },
+
+    async eject(): Promise<void> {
+      const deviceInfo = await getUsbDriveStatus();
+      if (!deviceInfo) {
+        throw new Error('No USB drive detected');
+      }
+      if (!deviceInfo.mountpoint) {
+        throw new Error('USB drive is not mounted');
+      }
```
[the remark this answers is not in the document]
jonahkagan: Hmm, yes I agree there's a potential race condition. I can't think of a reason these errors would be useful to surface, so I'm down to change this to a no-op if there's no USB drive mounted

libs/usb-drive/src/usb_drive.ts
amcmanus: I think putting the scripts in vxsuite makes sense. If there's system level config that needs to happen in support of script functionality, that'll be handled by complete-system / build-system.

libs/usb-drive/src/mock_usb_drive.ts
jonahkagan: Good callout. Do you have ideas for how to make this cleanup well otherwise though? Manual cleanup is tough to enforce. I'm kind of wondering if letting it leak is really that bad, since the worst case is just using up a lot of space on a dev machine.

libs/usb-drive/src/usb_drive.ts
```diff
@@ -131,6 +144,7 @@ export function detectUsbDrive(): UsbDrive {
       return { status: 'ejected' };
     },
 
+    // TODO do we need to run a `sync` command before unmounting?
```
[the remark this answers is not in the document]
jonahkagan: @adghayes would like your insight here

libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+    '-o',
+    'umask=000,nosuid,nodev,noexec',
+    devicePath,
+    VX_MOUNT_POINT,
+  ]);
+}
+
+async function unmountUsbDrive(mountPoint: string): Promise<void> {
+  debug(`Unmounting USB drive at ${mountPoint}`);
+  await exec('umount', [mountPoint]);
```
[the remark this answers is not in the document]
jonahkagan: Yup, sorry for the confusion

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
jonahkagan: Updated a1ff83cd4

libs/usb-drive/src/usb_drive.ts
arsalansufi: Just thought I'd note that, while working on auth changes, I also found it really useful to be able to put scripts in vxsuite (instead of vxsuite-complete-system). And I did in fact end up having a vxsuite-complete-system script consume a vxsuite script ([link](https://github.com/votingworks/vxsuite-complete-system/blob/c77994a22449046ddcb1d0e18dcafe7112850215/config/admin-functions/program-system-administrator-cards.sh#L12-L20))

libs/usb-drive/src/mock_usb_drive.ts
adghayes: It's not necessary, for sure. our new CVR files are bigger and I've hit low disk space once before, and had to manually cleanup. if something like this were easy, it would be a nice to have, but i'd have to think about how this is being run within tests:
https://github.com/votingworks/vxsuite/blob/6fe6a1e3efb0b3cac757f692596e62f1bc91111b/libs/fixtures/src/data/electionMinimalExhaustiveSample/cvr-files/standard.ts#L11-L17

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
adghayes: This is what we do currently and I I think we should. Much like the eject button your OS gives you, the user should reasonably be able to expect that it's not going to mess up data. And I don't see a reason to expose an unsafe API. In `usbstick.ts` I used the language "Eject" to represent "Sync + Unmount".

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
adghayes: re: mounting, we do show this in a couple places. If you are looking at the USB button in the top right corner of VxAdmin or VxCentralScan, it will show a mounting state when mounting. I don't think this is terribly important. But we also show it in modals that require USBs. If you open the modal without a USB drive, it will show the insert USB drive icon and when you do insert, will give feedback that it's mounting. I've always felt that the earlier feedback is nice-to-have and avoids the user having a momentary "it's not working" moment. but I can't remember how long the mounting takes exactly - maybe it wouldn't be noticeable to the user

re: ejecting, that seems fine. I wonder if it will be annoying to pass or convert promises into props for UI components that need ejecting feedback, but if you find it's not a problem then I think better to have one less status.

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
adghayes: I actually think this is fine. I read it as asserting that the `mountPoint` exists, which I can see the OS having a delay in reporting (like it delays scanning the filesystem on a mountable drive in [insert.sh](https://github.com/votingworks/vxsuite/blob/7ff5fc08ebf286f716545e24364394b019794c7d/libs/usb-mocking/insert.sh#L27). But if we're just asserting the `blockDevice` exists, that's fine - it just may not have a `mountPoint` yet.

libs/usb-drive/src/mock_usb_drive.ts
```diff
@@ -0,0 +1,88 @@
...
+ * filesystem directory. If using this interface, the mock functions will
+ * automatically be updated to return the correct status.
+ *
+ * Warning: if you use both interfaces, they may get out of sync.
+ */
+export function createMockUsbDrive(): MockUsbDrive {
+  let mockUsbTmpDir: tmp.DirResult | undefined;
+
+  const usbDrive: MockedUsbDrive = {
+    status: mockFunction('status'),
```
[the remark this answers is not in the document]
adghayes: Does that require specifying the number of repeated calls?

libs/usb-drive/src/usb_drive.ts
adghayes: I don't see anywhere where we clear a `didEject = true` status once the USB drive is no longer inserted - am I missing it or is it missing?

adghayes:
@jonahkagan how do you think this would be mocked in an `integration-testing` environment?

libs/usb-drive/src/mock_usb_drive.ts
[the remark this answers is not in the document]
jonahkagan: Nope!

libs/usb-drive/src/usb_drive.ts
jonahkagan: Great, I agree with you

libs/usb-drive/src/usb_drive.ts
jonahkagan: Updated to use scripts here: a39278fea

libs/usb-drive/src/usb_drive.ts
jonahkagan: 🤦 great catch. Fixed and added regression test: 6d174455b

jonahkagan:
> how do you think this would be mocked in an integration-testing environment?

I haven't worked with the integration tests much. Do you think using the provided `MockUsbDrive` would work?

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
jonahkagan: This makes me wonder if we even need `status()` to block on mounting. Maybe we just kick off a mount async and return `no_drive`. Then `status` can catch up whenever. We would need some locking but could be nice...

libs/usb-drive/src/usb_drive.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  return devicePath;
+}
+
+async function getUsbDriveStatus(): Promise<BlockDeviceInfo | undefined> {
+  const devicePath = await findUsbDriveDevice();
+  if (!devicePath) return undefined;
+  debug(`Found USB drive at ${devicePath}`);
+  return await getBlockDeviceInfo(devicePath);
+}
```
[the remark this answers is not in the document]
jonahkagan: Good point: 69b8e1376

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
jonahkagan: Great, thanks! Added a sync here: 1ed1bc2d87ba47820057c69f0319f2ff807b541c

libs/usb-drive/src/usb_drive.ts
jonahkagan: Discussed and decided to leave this bit out for now until we can repro the problem and understand better the underlying issue.

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
jonahkagan: Ok that's helpful context about mounting. I'm game to remove it and see if it feels laggy, or just support it. I could go either way. Mounting seems pretty much instantaneous in my VM, which is why I didn't originally think to support it.

Re: ejecting, I wouldn't use the promise directly, I would use the react-query mutation `isLoading` flag.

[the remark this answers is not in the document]
adghayes: I think this is a good question for @mattroe who may have locked down machines on hardware to see how long it takes in production conditions.

@mattroe do you have locked down VxAdmin, or VxDev VxAdmin? If you go to a modal that requires a USB and then insert it, how long does it show the "Loading..." message for? Could we go without it?

libs/usb-drive/src/usb_drive.ts
adghayes: > > how do you think this would be mocked in an integration-testing environment?
> 
> I haven't worked with the integration tests much. Do you think using the provided `MockUsbDrive` would work?

I don't think so because, although we could have a flag to use `MockUsbDrive`, we wouldn't be able to access its API.   We need an exposed API. We could do a [`MockFileUsbDrive`](https://github.com/votingworks/vxsuite/blob/main/libs/auth/src/mock_file_card.ts) similar to what @arsalansufi made for `libs/auth`.

adghayes:
> > how do you think this would be mocked in an integration-testing environment?

> I haven't worked with the integration tests much. Do you think using the provided MockUsbDrive would work?

I don't think so because, although we could have a flag to use MockUsbDrive, we wouldn't be able to access its API. We need an exposed API. We could do a MockFileUsbDrive similar to [what @arsalansufi made for libs/auth](https://github.com/votingworks/vxsuite/blob/main/libs/auth/src/mock_file_card.ts).

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
jonahkagan: I realized that even though the code I had before would likely work, it was still possible for an overlapping call to `status` to cause duplicate mount commands, so we needed some degree of protection anyway. I coupled this with the idea of mounting in the background to arrive at what I think is a pretty straightforward solution: e6af305ff

jonahkagan:
@adghayes yeah that seems like a good approach. Could also replace libs/usb-mocking with that.

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
mattroe: @adghayes I don't see any Loading screen when testing on a locked down VxAdmin. The screen just changes state within what seems like less than a second.

libs/usb-drive/src/mock_usb_drive.ts
jonahkagan: Added cleanup here: 1178426b2. Thanks for the suggestion!

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
adghayes: Huh, that's not what I remember, but it seems like enough of an indication that it's OK to remove.

jonahkagan:
@adghayes I think I've addressed all the feedback (great feedback, btw!). Ready for a second look

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
jonahkagan: Thanks! We can keep an eye on it

libs/usb-drive/src/usb_drive.ts
jonahkagan: @amcmanus can you point me to what I might need to modify in complete-system/build-system to support these new scripts I'm adding?

libs/usb-drive/src/mount.sh
```diff
@@ -0,0 +1,29 @@
...
+}
+
+if ! [[ $# -eq 1 ]]; then
+    usage
+fi
+
+USB_DRIVE_DEVICE_REGEX=^/dev/sd[a-z][0-9]$
+LOOP_DEVICE_REGEX=^/dev/loop[0-9]p[0-9]$
+
+if ! [[ $1 =~ $USB_DRIVE_DEVICE_REGEX || $1 =~ $LOOP_DEVICE_REGEX ]]; then
```
adghayes, line 17: this loop device regex is specific to the way that `libs/usb-mocking` is attaching a mock volume. if you don't see this library being used with `libs/usb-mocking`, you can remove. if you do see them being used together in the interim, makes sense to keep.

libs/usb-drive/src/usb_drive.test.ts
```diff
@@ -0,0 +1,252 @@
...
+
+jest.mock('fs', () => ({
+  promises: {
+    readdir: jest.fn().mockRejectedValue(new Error('Not mocked')),
+    readlink: jest.fn().mockRejectedValue(new Error('Not mocked')),
+  },
+}));
+jest.mock('./exec', () => ({
+  exec: jest.fn().mockRejectedValue(new Error('Not mocked')),
+}));
```
adghayes, line 13: i like this pattern

libs/usb-drive/package.json
```diff
@@ -0,0 +1,67 @@
...
+    ],
+    "package.json": [
+      "sort-package-json"
+    ]
+  },
+  "dependencies": {
+    "@votingworks/basics": "workspace:*",
+    "@votingworks/test-utils": "workspace:*",
+    "debug": "^4.3.2",
+    "tmp": "^0.2.1"
```
adghayes, line 37: thank you for keeping things that should be in dev dependencies in dev dependencies, this has been bothering me in other libs

apps/scan/backend/src/app_usb_drive.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      mountPoint: expect.any(String),
+      deviceName: 'mock-usb-drive',
+    });
+  });
+});
+
+test('ejectUsbDrive', async () => {
+  await withApp({}, async ({ apiClient, mockUsbDrive }) => {
+    mockUsbDrive.usbDrive.eject.expectCallWith().resolves();
+    await expect(apiClient.ejectUsbDrive()).resolves.toBeUndefined();
```
adghayes, line 21: do you need to check status after ejecting?

apps/scan/backend/test/helpers/shared_helpers.ts
adghayes, line 1: maybe i'm imagining things but I thought there were more USB test helpers in the scan/backend that could be removed with this change

adghayes (approved):
LGTM. Some random comments. Excited to see this merged.

Reviewing made me realize that I might have introduced a regression when I moved exporting CVRs to the backend. Previously we would await `sync` (but not `unmount`) after the CVR export started, to ensure that we didn't move through the poll worker flow (or the export results modal) prematurely on VxScan. I don't see any `sync` calls anymore. 

I think that `libs/usb-drive` will need an API for simply calling `sync`, but this is not a regression (if it is one) introduced by this PR. Just made me notice.

apps/scan/backend/src/app_usb_drive.test.ts
jonahkagan: I don't think so. Since the usb drive library is mocked, I don't think checking status after ejecting would actually test anything here.

libs/usb-drive/src/mount.sh
jonahkagan: I think we will probably use them together for now, but that's good to know!

apps/scan/backend/test/helpers/shared_helpers.ts
jonahkagan: Lol, I couldn't find anything else. There are definitely others scattered around the monorepo that we can eventually clean up once we convert everything to use this lib

PR #3360, opened 2023-04-25
Add jurisdiction state management to VxMark

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2912

The VxMark equivalent of https://github.com/votingworks/vxsuite/pull/3348 for VxCentralScan. Largely copying from that PR's description:

This PR replaces the hardcoded dev jurisdiction in VxMark with proper jurisdiction state management. When the machine is configured, the jurisdiction will be set to the jurisdiction on the election manager card used to auth onto the machine. And when the machine is unconfigured, the jurisdiction will be cleared.

When a jurisdiction is set, the machine will only accept cards from that jurisdiction. And when no jurisdiction is set, the machine will accept cards from all jurisdictions. This logic already exists and lives in the auth lib.

## Testing

- [x] Updated automated tests
- [x] Manually tested that jurisdiction is set on configure
- [x] Manually tested that jurisdiction is cleared on unconfigure

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ N/A

PR #3384, opened 2023-05-01
Add jurisdiction state management to VxScan

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/2912

The VxScan equivalent of https://github.com/votingworks/vxsuite/pull/3348 for VxCentralScan and https://github.com/votingworks/vxsuite/pull/3360 for VxMark. Largely copying from those PRs' descriptions:

This PR replaces the hardcoded dev jurisdiction in VxScan with proper jurisdiction state management. When the machine is configured, the jurisdiction will be set to the jurisdiction on the election manager card used to auth onto the machine. And when the machine is unconfigured, the jurisdiction will be cleared.

When a jurisdiction is set, the machine will only accept cards from that jurisdiction. And when no jurisdiction is set, the machine will accept cards from all jurisdictions. This logic already exists and lives in the auth lib.

While I was at it, I also tweaked how we define jurisdiction in tests.

## Testing

- [x] Updated automated tests
- [x] Manually tested that jurisdiction is set on configure
- [x] Manually tested that jurisdiction is cleared on unconfigure

## Checklist

### DIFF
--- a/apps/scan/backend/src/app.ts
+++ b/apps/scan/backend/src/app.ts
@@ -12,6 +12,7 @@ import {
 import {
   ScannerReportData,
   ScannerReportDataSchema,
+  isElectionManagerAuth,
   singlePrecinctSelectionFor,
 } from '@votingworks/utils';
 import express, { Application } from 'express';
@@ -23,7 +24,6 @@ import {
 } from '@votingworks/backend';
 import { assert, err, iter, ok, Result } from '@votingworks/basics';
 import {
-  DEV_JURISDICTION,
   InsertedSmartCardAuthApi,
   InsertedSmartCardAuthMachineState,
 } from '@votingworks/auth';
@@ -44,10 +44,10 @@ function constructAuthMachineState(
   workspace: Workspace
 ): InsertedSmartCardAuthMachineState {
   const electionDefinition = workspace.store.getElectionDefinition();
+  const jurisdiction = workspace.store.getJurisdiction();
   return {
     electionHash: electionDefinition?.electionHash,
-    // TODO: Persist jurisdiction in store and pull from there
-    jurisdiction: DEV_JURISDICTION,
+    jurisdiction,
   };
 }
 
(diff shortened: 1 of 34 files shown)


libs/test-utils/src/auth.ts
```diff
@@ -5,14 +5,15 @@ import {
...
+  TEST_JURISDICTION,
 } from '@votingworks/types';
 
 export function fakeSystemAdministratorUser(
   props: Partial<SystemAdministratorUser> = {}
 ): SystemAdministratorUser {
   return {
     role: 'system_administrator',
-    jurisdiction: 'jurisdiction',
+    jurisdiction: TEST_JURISDICTION,
```
arsalansufi, line 16: I realized that the jurisdiction on fake user objects created by `@votingworks/test-utils` wasn't matching the jurisdiction we use in app tests (`DEV_JURISDICTION`). Not a big deal since we're mocking auth in app tests and the jurisdiction mismatch won't actually be detected but did feel a bit misleading. Also made it hard to know which jurisdiction to expect on user objects when writing test expectations.

Through this commit and the next one, I introduce a `TEST_JURISDICTION`, separate from the `DEV_JURISDICTION`, and consistently use it across app tests.

pnpm-lock.yaml
```diff
@@ -1461,6 +1461,9 @@ importers:
       '@votingworks/grout':
         specifier: workspace:*
         version: link:../../../libs/grout
+      '@votingworks/types':
+        specifier: workspace:*
+        version: link:../../../libs/types
```
arsalansufi, line 1466: This large `pnpm-lock.yaml` diff seems to be undoing what I _think_ were accidental `pnpm-lock.yaml` changes introduced by https://github.com/votingworks/vxsuite/pull/3355

Will post in #vxsuite-prodeng later to make sure we're all on the same version of pnpm etc.

apps/scan/backend/src/app.ts
```diff
@@ -86,13 +86,12 @@ function buildApi(
...
-      assert(!store.getElectionDefinition(), 'Already configured');
-      const [usbDrive] = await usb.getUsbDrives();
-      assert(usbDrive?.mountPoint !== undefined, 'No USB drive mounted');
-
       const authStatus = await auth.getAuthStatus(
         constructAuthMachineState(workspace)
       );
+      assert(!store.getElectionDefinition(), 'Already configured');
+      const [usbDrive] = await usb.getUsbDrives();
+      assert(usbDrive?.mountPoint !== undefined, 'No USB drive mounted');
```
kshen0, line 94: Can you remind me why this has to move under the auth check?

arsalansufi: Ah yeah this move isn't strictly necessary!

I've just been getting in the habit of making auth status retrievals the first step in backend handlers since I eventually want to move auth status retrievals into auth check middlewares/decorators. So having auth status retrievals come first is more representative of the planned long-term state.

kshen0 (approved):

PR #3393, opened 2023-05-02
Turn off gts-no-import-export-type lint rule

jonahkagan:
## Overview

We often disable this rule to allow importing types from backend packages into frontend packages. Here, we change the recommended setting to disable this rule by default.

_Review by commits_

## Demo Video or Screenshot
N/A

## Testing Plan
CI should cover it

## Checklist

### DIFF
--- a/libs/eslint-plugin-vx/src/configs/recommended.ts
+++ b/libs/eslint-plugin-vx/src/configs/recommended.ts
@@ -49,7 +49,10 @@ export = {
     'vx/gts-no-default-exports': 'error',
     'vx/gts-no-foreach': 'error',
     'vx/gts-no-for-in-loop': 'error',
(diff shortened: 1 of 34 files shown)


libs/eslint-plugin-vx/src/configs/recommended.ts
```diff
@@ -49,7 +49,10 @@ export = {
     'vx/gts-no-default-exports': 'error',
     'vx/gts-no-foreach': 'error',
     'vx/gts-no-for-in-loop': 'error',
-    'vx/gts-no-import-export-type': ['error', { allowReexport: true }],
+    // Importing types allows a package to list another package as a dev
+    // dependencies if only using the package's types. This makes it possible
```
adghayes, line 53: `dependency`

adghayes (approved):
there's a small typo in the disabling comment , but otherwise LGTM

PR #3405, opened 2023-05-03
m16 replace ballot bag any time poll worker wants to

benadida:
This is a minimal change to allow for replacing the ballot bag at any time from the poll worker screen.

This is a little bit more convoluted than I would have liked, in that the existing replace-ballot-bag-screen was fairly tightly bound, in its behavior, to a situation where the ballot bag *must* be replaced. I made two changes:

- update the ballot bag counter the moment the confirm button is hit, independent of auth status.
- force the replace ballot bag screen to continue to be shown until the poll worker card is removed.

I *think* this logic is better, but this is a slightly bigger change than I would have liked.

### DIFF
--- a/frontends/precinct-scanner/src/app_root.tsx
+++ b/frontends/precinct-scanner/src/app_root.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useReducer } from 'react';
+import React, { useCallback, useEffect, useReducer, useState } from 'react';
 import 'normalize.css';
 import makeDebug from 'debug';
 
@@ -238,6 +238,8 @@ export function AppRoot({
   machineConfig: machineConfigProvider,
   logger,
 }: Props): JSX.Element | null {
+  const [showReplaceBallotBagScreen, setShowReplaceBallotBagScreen] =
+    useState(false);
   const [appState, dispatchAppState] = useReducer(appReducer, initialState);
   const {
     electionDefinition,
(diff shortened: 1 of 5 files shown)


adghayes:
Will take a look at this in an hour or so!

frontends/precinct-scanner/src/components/replace_ballot_bag_screen.tsx
```diff
@@ -17,12 +17,14 @@ export function ReplaceBallotBagScreen({
   onComplete,
 }: Props): JSX.Element {
   const [confirmed, setConfirmed] = useState(false);
+  const [onCompleteCalled, setOnCompleteCalled] = useState(false);
 
   useEffect(() => {
-    if (confirmed && !pollWorkerAuthenticated) {
+    if (confirmed && !onCompleteCalled) {
       onComplete();
```
adghayes, line 24: note to self: now we call `onComplete` from the `ReplaceBallotBagScreen` as soon as it's confirmed, not waiting until the PW card is removed.

frontends/precinct-scanner/src/screens/poll_worker_screen.tsx
```diff
@@ -523,6 +527,23 @@ export function PollWorkerScreen({
...
+  // because we depend on unmounting for that to happen.
+  if (isReplacingBallotBag) {
+    return (
+      <ReplaceBallotBagScreen
+        scannedBallotCount={scannedBallotCount}
+        pollWorkerAuthenticated
+        onComplete={() => {
+          if (onReplaceBallotBag) {
+            onReplaceBallotBag();
+          }
```
adghayes, line 541: note to self: `onComplete` via the PW screen (not forced ballot bag change, voluntary one) simply updates the app state. it is poll worker removing card which will unmount this PW screen and reset this.

frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -544,17 +546,29 @@ export function AppRoot({
...
     return (
       <ReplaceBallotBagScreen
         scannedBallotCount={scannerStatus.ballotsCounted}
         pollWorkerAuthenticated={isPollWorkerAuth(auth)}
-        onComplete={() =>
+        onComplete={() => {
+          // pin the state to be in "replacing ballot bag" mode
+          // so that, when we update the state, we don't immediately
+          // unmount this component.
+          setShowReplaceBallotBagScreen(true);
```
adghayes, line 566: note to self: this flag is only applicable to the original, forced ballot bag replacement path

adghayes: maybe this could cause a flicker, not sure

adghayes (approved):
Looks like a fair tweak to me! It is indeed convoluted and took me too long to wrap my head around, but I'd rather a small convoluted patch than a large one. I am assuming you've manually tested both paths (the forced and voluntary bag replacement).

benadida:
> Looks like a fair tweak to me! It is indeed convoluted and took me too long to wrap my head around, but I'd rather a small convoluted patch than a large one. I am assuming you've manually tested both paths (the forced and voluntary bag replacement).

Yes, I have tested both paths indeed.

frontends/precinct-scanner/src/app_root.tsx
benadida: did not see a flicker. I think this shouldn't, since this just flips the second of the OR'ed conditions for showing the replace-bag-screen.

PR #3396, opened 2023-05-02
chore(scan): remove VX HMPB interpretation and metadata encoding

eventualbuddha:
## Overview

We're no longer including rendered VX HMPBs in ballot packages and no longer trying to interpret them. This PR removes the actual `Interpreter` class that handled interpreting VX HMPBs, as well as the encoder/decoder for the QR code data. Adds a basic version of an interpreter for VX BMD ballots. I'll use that in `scan`/`central-scan` in a future PR.

Refs #3374, follow-up to #3392.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

Issue #3374, opened 2023-04-27
Remove VX-style hand-marked paper ballots


eventualbuddha:
To be replaced by a 2.0 version.

PR #3407, opened 2023-05-03
consolidate and restructure scanning interpretation

eventualbuddha:
## Overview

Refs #3374, follow-up to #3396.

The split of `services/scan` into `apps/central-scan/backend` and `apps/scan/backend` created a lot of duplicate code. This duplicate code has drifted such that the copies are, in many cases, no longer identical. A lot of these changes are not actually relevant to the specific app the copies reside in, and so could be consolidated.

Specifically, both backends did something like this before this PR:
1. get a pair of scanned images from the scanner
2. figure out which underlying interpreter to use (right now it's just VX BMD or NH HMPB)
3. delegate to the interpreter, but then haphazardly apply validation for things like the selected precinct/test mode/election hash check
4. save the ballot images in the persistent storage location

They each did these things in somewhat different ways, possibly incompletely. This PR consolidates the shared functionality mostly into a function called `interpretSheetAndSaveImages` within `libs/backend`. There's still more to clean up, but this gets things to a decent place and should make additional changes easier.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist


apps/central-scan/backend/src/interpret_sheet.test.ts
```diff
@@ -0,0 +1,261 @@
...
+});
+
+test('properly detects test ballot in live mode', async () => {
+  const ballotImagePath =
+    electionFamousNames2021Fixtures.machineMarkedBallotPage1.asFilePath();
+  const interpretationResult = await interpretSheet(
+    {
+      electionDefinition: electionFamousNames2021Fixtures.electionDefinition,
+      precinctSelection: ALL_PRECINCTS_SELECTION,
+      testMode: false, // this is the test mode
```
arsalansufi: I know this is just copied from a prior file, but I think this comment is a bit confusing - can probably remove

arsalansufi (approved):
I don't have too much context on these code paths, but the high-level intent makes sense to me - LGTM!

PR #3381, opened 2023-04-28
chore(fixtures): remove extraneous ballot package files

eventualbuddha:
## Overview
Now that we're no longer producing or importing ballot PDFs and layouts in ballot packages, this commit removes the ballot package fixtures.

Refs #3374, follow-up to #3380.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #3408, opened 2023-05-03
Generate CVR Fixtures from JSON & Replace Existing CVR Fixtures w/ Saner, Smaller Ones

adghayes:
## Overview

#3381 removed a ballot package I wanted to use to generate a fixture, but then I realized - why use ballot packages at all anymore? We're not getting images from them. And we're not getting layouts from them. In commit 1, I change `cvr-fixture-generator` to create fixtures based off of election definitions. We can switch it back if we ever need to, but the great advantage of using election definitions is that we could start to autogenerate or validate fixture CVR files. 

In commit 2, I regenerate the minimal exhaustive cvr file, see commit description.

In commit 3, I generate an Amherst cvr file, because we need a fixture with "images."

In commit 4, I fix testing in VxAdmin.

In commit 5, I fix testing in `libs/backend`.


adghayes:
Another advantage of downsizing the default CVR files is that tests run faster!

eventualbuddha (approved):

PR #3412, opened 2023-05-04
[VxMark] Replace text size management code with newer DisplaySettings component

kofi-q:
## Overview

Removing the current text size management code (size selector button, associated modal, and context data and API) and replacing it with the `DisplaySettings` component from libs/ui, which pairs with the `AppBase` wrapper at the root of the app tree.

Was planning on doing this as a last step, since the new themes need a bit of cleaning up within VxMark, but this will be short-lived anyway.

TODO (future PRs):
- Add a display settings reset at the end of every voter session
- Figure out interplay between voter display settings and election-official-screens (e.g. do we reset back to default when an election manager/poll worker card is inserted after a voter has changed themes?)
- Try to break the "see more" button's dependency on text size

## Demo Video or Screenshot
https://user-images.githubusercontent.com/264902/236270657-6c2309de-3167-4ee6-a42d-bad2350d1080.mov

## Testing Plan
- Updated tests that were expecting the older settings button/modal
- Removed a bunch of tests that felt redundant with this change - replaced with smaller unit tests verifying that the added `DisplaySettingsButton` and `DisplaySettingsPage` are rendered as expected.

## Checklist


apps/mark/frontend/src/app_root.tsx
```diff
@@ -632,28 +598,16 @@ export function AppRoot({
   const machineConfig = machineConfigQuery.data;
 
   if (!cardReader) {
-    return (
-      <SetupCardReaderPage
-        useEffectToggleLargeDisplay={useEffectToggleLargeDisplay}
```
adghayes, line 637: @kofi-q could you remind me what the `useEffectToggleLargeDisplay` props were originally intended to do, and why they're no longer necessary?

apps/mark/frontend/src/pages/review_page.test.tsx
```diff
@@ -20,7 +23,7 @@ it('Renders ContestPage', () => {
   screen.getByText(/color & size/i);
 });
 
-it('Renders ContestPage in Landscape orientation', () => {
+it('Renders ReviewPage in Landscape orientation', () => {
```
adghayes, line 26: Where are we w.r.t. supporting or not supporting landscape? Is this test necessary anymore?

adghayes (approved):
@kofi-q looks good to me, thank you for the tight organization of commits. Two questions from me in the comments.

apps/mark/frontend/src/app_root.tsx
kofi-q: It was adjusting the font size to the largest on some of the simple full-page screens that just have a large icon and a prompt, to make the text stand out larger. Not needed anymore, with the current heading size hierarchy we have (and in the rare case we ever need to override the size mode, we can now wrap the relevant UI in a `ThemeProvider` instead)

apps/mark/frontend/src/pages/review_page.test.tsx
kofi-q: I think we'll be able to move support eventually, but Caroline mentioned we'll support it in the short term (though I'm already blanking out on the specifics of the conversation). Will follow up again with her and Roe on what  the timeline for that looks like - would simplify a lot of code if we could remove it altogether

PR #3413, opened 2023-05-04
feat(election-manager): add USB formatting feature.

benadida:
## Overview

Simple USB formatting in VxAdmin.

Because we have a more complete version of this on `main`, doing minimal testing / logging here.


libs/auth/scripts/gp.jar
eventualbuddha, line 1: What is this?

benadida: zoinks, aggresive git add.

adghayes (approved):
This fit in so nicely! LGTM. Two low-pri thoughts:
- If it were for my own purposes, I want to add a truncated UUID to the USB label. I get really confused when my USBs have the same label, but if they're exclusively using them with our software (and not plugging more than one into any of their external computers), shouldn't matter.
- My read of the behavior here is the USB drive is ejected. It is then formatted which does not change the state, so it will remain "ejected" and then you just remove it. Is that right? That makes perfect sense to me as a user flow.

benadida:
> * If it were for my own purposes, I want to add a truncated UUID to the USB label. I get really confused when my USBs have the same label, but if they're exclusively using them with our software (and not plugging more than one into any of their external computers), shouldn't matter.

Yes, good call on the naming, I'll think about a simple tweak there.

> * My read of the behavior here is the USB drive is ejected. It is then formatted which does not change the state, so it will remain "ejected" and then you just remove it. Is that right? That makes perfect sense to me as a user flow.

You got it, that was my thinking.

PR #3418, opened 2023-05-05
Kevin/paper handler driver

kshen0:
## Overview
Still WIP and not ready for review

Since this is just copy+pasted from another repo then modified, the diff shows old code as new changes. If helpful, I can restructure so the copy+paste is a separate branch and open this PR against that branch, so the diff will be only net new changes.

## Checklist


libs/custom-paper-handler/src/driver/deprecated_driver.ts
kshen0, line 1: Unmodified copy of `driver.ts` prototype; kept for reference but will probably remove before review

libs/custom-paper-handler/src/driver/deprecated_driver.test.ts
kshen0, line 1: Since tests don't exist for the old driver, this test file serves as the "before" tests ie. passing tests on main. TODO this file should be an exact copy of `driver.test.ts` except it should import `deprecated_driver.ts`. If both test files pass then we have confidence the migration didn't break anything.

libs/custom-paper-handler/src/driver/driver.test.ts
kshen0, line 1: New driver file that uses message-coder

libs/custom-paper-handler/src/driver/driver.ts
kshen0, line 1: New driver file that uses message-coder

kshen0:
Closing for cleaner diff in https://github.com/votingworks/vxsuite/pull/3443

PR #3420, opened 2023-05-08
[VxMark] VVSG Design Fit and Finish - Part 1

kofi-q:
## Overview

Tweaking styling on each screen to fix anything that looks broken after enabling the new VVSG-compatible themes and making sure everything looks reasonable at the default text size on the ELO 15".

## Testing Plan
- Visual testing
- Updating any affected tests as I go

## Checklist

### DIFF
--- a/libs/ui/src/button_list.tsx
+++ b/libs/ui/src/button_list.tsx
@@ -10,6 +10,7 @@ export const ButtonList = styled.p`
   }
   & > button {
     margin-bottom: 0.5rem;
+    width: 100%;
     overflow: hidden;
   }
 `;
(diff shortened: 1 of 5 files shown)


libs/ui/src/button_list.tsx
```diff
@@ -10,6 +10,7 @@ export const ButtonList = styled.p`
   }
   & > button {
     margin-bottom: 0.5rem;
+    width: 100%;
```
adghayes, line 13: Do you have good reason to suspect this won't have unintended impacts in other places?

adghayes (approved):
lgtm, see single question

libs/ui/src/button_list.tsx
kofi-q: Yeah, good question - I noticed all the usage instances of `ButtonList` are relying on explicitly setting the (now-deprecated) `fullWidth` prop of the `Button` component (which just applies `width: 100%` in legacy mode). `fullWidth` is ignored with the new themes, so  decided to move that logic up into `ButtonList`, since it semantically fits better there.

PR #3426, opened 2023-05-09
feat(election-manager): number the CVR files.

benadida:
## Overview

When jurisdictions import CVRs from multiple precinct scanners, having a quick count of how many CVR files have been loaded makes it very easy to know how many you've got left to load. We've seen this in the field. This just adds numbering to the list of CVR files in election-manager.

I'll also PR this into `main` separately.

## Demo Video or Screenshot

Before:
Screenshot 2023-05-09 at 12 08 12 PM

After:
Screenshot 2023-05-09 at 12 07 05 PM

## Testing Plan

Manually tested.

## Checklist

- [ X ] I have added a screenshot and/or video to this PR to demo the change
- [ X ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

### DIFF
--- a/frontends/election-manager/src/screens/tally_screen.tsx
+++ b/frontends/election-manager/src/screens/tally_screen.tsx
@@ -163,6 +163,9 @@ export function TallyScreen(): JSX.Element {
               {hasAnyFiles ? (
                 <React.Fragment>
                   <tr>
+                    <TD as="th" narrow nowrap textAlign="right">
+                      #
+                    </TD>
                     <TD as="th" narrow nowrap>
                       Created At
                     </TD>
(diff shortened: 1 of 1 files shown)


frontends/election-manager/src/screens/tally_screen.tsx
jonahkagan, line 1: The total CVR count footer row is now misaligned with the header rows.

frontends/election-manager/src/screens/tally_screen.tsx
```diff
@@ -163,6 +163,9 @@ export function TallyScreen(): JSX.Element {
               {hasAnyFiles ? (
                 <React.Fragment>
                   <tr>
+                    <TD as="th" narrow nowrap textAlign="right">
+                      #
+                    </TD>
```
jonahkagan, line 168: This column seems unnecessarily wide. I'm not sure if there's any way to shrink it to fit the content, but that might look better.

jonahkagan (changes requested):

frontends/election-manager/src/screens/tally_screen.tsx
benadida: oh duh, thank you.

frontends/election-manager/src/screens/tally_screen.tsx
benadida: probably because of misalignment of bottom row, can't believe i missed that.

benadida:
@jonahkagan much better. Fixing the bottom row is all that was needed:
Screenshot 2023-05-09 at 1 24 56 PM

PR #3429, opened 2023-05-09
Improve interpreter timing mark detection

jonahkagan:
## Overview

When exploring designing our own ballots from scratch, I ran into some small snags with the interpreter:
- A confusing error message when the corner timing marks couldn't be found
- An expectation that the horizontal spacing between timing marks is equivalent to the vertical spacing

I made attempts to fix both of these in a way that I think is a net improvement, but would like to hear if there's a good reason things were set up that way before.

## Demo Video or Screenshot
N/A

## Testing Plan
Manually tested on the ballots I've been designing, relying on existing tests to prevent regressions.

(Side note: now that we're investing in this interpreter, it might be good to think about beefing up the testing so that it's easy to add new tests or change tests for situations like this)

## Checklist

### DIFF
--- a/libs/ballot-interpreter-nh/src/metadata.rs
+++ b/libs/ballot-interpreter-nh/src/metadata.rs
@@ -195,19 +195,22 @@ pub fn compute_bits_from_bottom_timing_marks(
 
     // The first and last timing marks must be present.
     match (bottom_timing_marks.first(), bottom_timing_marks.last()) {
-        (Some(Some(_)), Some(Some(_))) => {}
-        _ => {
+        (Some(Some(_)), Some(Some(_))) => {
+            // both timing marks are present
+        }
+        (Some(None), Some(None)) => {
             return Err(BallotPageMetadataError::InvalidTimingMarkCount {
                 expected: 2,
-                actual: vec![
-                    bottom_timing_marks.first().unwrap(),
-                    bottom_timing_marks.last().unwrap(),
-                ]
-                .iter()
-                .filter(|&mark| mark.is_some())
-                .count(),
+                actual: 0,
             });
         }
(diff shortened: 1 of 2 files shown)


libs/ballot-interpreter-nh/src/metadata.rs
```diff
@@ -199,10 +199,13 @@ pub fn compute_bits_from_bottom_timing_marks(
...
-                    .iter()
-                    .filter(|&mark| mark.is_some())
-                    .count(),
+                actual: vec![
+                    bottom_timing_marks.first().unwrap(),
+                    bottom_timing_marks.last().unwrap(),
+                ]
+                .iter()
+                .filter(|&mark| mark.is_some())
+                .count(),
```
eventualbuddha, line 208: Maybe this instead? It's more verbose, but pretty clear and doesn't `unwrap`.

```rust
    match (bottom_timing_marks.first(), bottom_timing_marks.last()) {
        (Some(Some(_)), Some(Some(_))) => {
            // both timing marks are present
        }
        (Some(None), Some(None)) => {
            return Err(BallotPageMetadataError::InvalidTimingMarkCount {
                expected: 2,
                actual: 0,
            });
        }
        (Some(None), _) | (_, Some(None)) => {
            return Err(BallotPageMetadataError::InvalidTimingMarkCount {
                expected: 2,
                actual: 1,
            });
        }
        _ => unreachable!(),
    }
```

eventualbuddha (approved):

libs/ballot-interpreter-nh/src/metadata.rs
jonahkagan: Sure!

PR #3440, opened 2023-05-11
change how we handle write-in candidates in manual tally entry

adghayes:
## Overview

In order to move manual tally data to the backend, I needed to make this transition PR changing how we treat write-in candidates in manual data.

I added the ability to add write-in candidates to manual tallies in #2649. I recommend watching the video in that PR description for context. The way I implemented it had two main problems:

1. Feature Limitation: Write-in candidates from on-screen adjudication would pre-populate in the manual tally entry screens, but write-in candidates from manual tally entry would not pre-populate in on-screen adjudication. We assumed that the flow would move in that direction, but we know we can't always guarantee it
2. Technical Issue: The way that manual data added write-in candidates were tracked was by having a key value pair in the manual tally. But, there is a manual tally for each precinct, so it required some overcomplicated logic to update _all_ precinct tallies when you save the current one. When moving to the backend, this makes it more complicated than just updating one precinct at a time.

Both of these issues can be addressed now that write-in candidates are normalized after #3417, which is the purpose of this PR. A good deal of the logic you see added here is very temporary and will be cleared in the next PR, which I'll call out in comments.

This is a refactor mostly but there were a couple product choices I made:
- Previously we allowed removing a manual data added write-in candidate. You would delete them from a precinct screen and they would be removed from all the tallies. Now that I understand more about manual tally data, I think this is an unsafe action. It allows you to possibly (probably) invalidate a bunch of other data that has been validated on entry. I removed the modal entirely
- I allow you to remove a newly added candidate (from the manual interface) if you haven't saved it yet, but if you have saved it, there's no way to delete them! Ever! (without clearing the election). In a subsequent PR moving manual tallies to the backend, I will implement logic to remove such candidates if they are no longer being referenced, similar to how candidates via on screen adjudication are removed automatically.

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
+++ b/apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
@@ -37,6 +37,7 @@ import {
   getWriteInCandidates,
   addWriteInCandidate as addWriteInCandidateApi,
 } from '../api';
+import { normalizeWriteInName } from '../utils/write_ins';
 
 const TallyInput = styled(TextInput)`
   width: 4em;
@@ -85,10 +86,6 @@ function ContestDataRow({
       </TD>
     </tr>
   );
-}
-
-function normalizeWriteInName(name: string) {
-  return name.toLowerCase().trim();
 }
 
 function AddWriteInRow({
--- a/apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
+++ b/apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
@@ -1,4 +1,4 @@
-import { assert, assertDefined } from '@votingworks/basics';
+import { Optional, assert, assertDefined, find } from '@votingworks/basics';
 import React, { useCallback, useContext, useState } from 'react';
 import { useHistory, useParams } from 'react-router-dom';
 import styled from 'styled-components';
@@ -33,7 +33,6 @@ import {
   getEmptyManualTalliesByPrecinct,
   getTotalNumberOfBallots,
 } from '../utils/manual_tallies';
-import { isManuallyAdjudicatedWriteInCandidate } from '../utils/write_ins';
 import {
   getWriteInCandidates,
   addWriteInCandidate as addWriteInCandidateApi,
@@ -88,6 +87,10 @@ function ContestDataRow({
   );
 }
 
+function normalizeWriteInName(name: string) {
+  return name.toLowerCase().trim();
+}
+
 function AddWriteInRow({
   addWriteInCandidate,
   contestId,
@@ -115,7 +118,9 @@ function AddWriteInRow({
             onPress={onAdd}
             disabled={
               writeInName.length === 0 ||
-              disallowedCandidateNames.includes(writeInName)
+              disallowedCandidateNames.includes(
+                normalizeWriteInName(writeInName)
+              )
             }
           >
             Add
@@ -173,6 +178,12 @@ interface TempContestTally {
 interface TempManualTally {
   readonly contestTallies: Dictionary<TempContestTally>;
   readonly numberOfBallotsCounted: number;
+}
+
+interface TempWriteInCandidate {
+  readonly id: string;
+  readonly name: string;
+  readonly contestId: string;
 }
 
 function getNumericalValueForTally(tally: number | EmptyValue): number {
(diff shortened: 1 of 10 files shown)


apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -175,6 +180,12 @@ interface TempManualTally {
   readonly numberOfBallotsCounted: number;
 }
 
+interface TempWriteInCandidate {
+  readonly id: string;
+  readonly name: string;
+  readonly contestId: string;
+}
+
```
adghayes, line 188: The idea here is that you don't actually create new write-in candidates until you save the tally.

apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -308,25 +302,81 @@ export function ManualDataImportPrecinctScreen(): JSX.Element {
...
+    try {
+      // add temporary write-in candidates to backend, get ids
+      const writeInCandidateRecords = await Promise.all(
+        nonZeroTempWriteInCandidates.map((candidate) =>
+          addWriteInCandidateMutation.mutateAsync({
+            contestId: candidate.contestId,
+            name: candidate.name,
+          })
+        )
+      );
```
adghayes, line 347: It's a bit convoluted to do a bunch of requests here. It would be better to do it as one mutation, but this works and it is temporary and will be removed soon.

When manual tally data is moved to the backend, you'll just submit the whole "form" including the temporary write-in candidates and the backend will handle adding or removing candidates as necessary.

apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -401,32 +472,49 @@ export function ManualDataImportPrecinctScreen(): JSX.Element {
         break;
       default: {
         const tally = contestTally.tallies[dataKey];
-        assert(tally);
```
adghayes, line 463: The contract of how manual tallies work has changed with this PR.

Previously, we'd expect the contestTallies in a manual tally to include an option tally for every candidate. So when we wanted to add a write-in candidate, we'd add it there. But it's fragile to use that list as the source of truth when there are so many copies of the list (for each precinct).

Now, we expect manual tallies to include an option tally for each official candidate but, for write-in candidates, we can create them on the fly and just assume `undefined` is 0.

apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -478,51 +568,106 @@ export function ManualDataImportPrecinctScreen(): JSX.Element {
...
             <ContestData key={contest.id}>
               <Text small>{getContestDistrictName(election, contest)}</Text>
               <h3>{contestTitle}</h3>
               <Table borderTop condensed>
                 <tbody>
-                  {contest.type === 'candidate' &&
-                    getCandidatesFromContestTally(contestTally).map(
-                      (candidate) => (
+                  {contest.type === 'candidate' && (
+                    <React.Fragment>
```
adghayes, line 593: This form has a lot of repeated buts but when I tried to consolidate it felt harder to think through.

apps/admin/frontend/src/utils/manual_tallies.ts
```diff
@@ -246,8 +244,7 @@ export function convertTalliesByPrecinctToFullManualTally(
 }
 
 export function getEmptyContestTallies(
-  election: Election,
-  allAdjudicatedValues?: Map<ContestId, string[]>
```
adghayes, line 250: Because of the change in contract (we don't need to include write-ins in contestTallies if they're 0, and we have a different source of truth for who the write-in candidates are), we don't need to include these names from adjudicated values.

kofi-q (approved):
Minor write-in de-dupe logic comment, but LGTM overall!

apps/admin/frontend/src/screens/manual_data_import_precinct_screen.tsx
```diff
@@ -88,6 +87,10 @@ function ContestDataRow({
   );
 }
 
+function normalizeWriteInName(name: string) {
```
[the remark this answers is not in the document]
adghayes: sgtm, I wasn't sure what other things to do here bug the spacing makes sense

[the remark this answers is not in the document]
adghayes: addressed in [3f35cab](https://github.com/votingworks/vxsuite/pull/3440/commits/3f35cabc5ce83275ded07562697dd4515db550fd)

PR #3443, opened 2023-05-11
Kevin/paper handler driver 2

kshen0:
## Overview
To try to manage diff size of merging paper-handler changes into vxsuite, there will be 3 PRs
1. Import paper-handler basics from `vxsuite-print-scan` ~as is (with minor updates to imports, build system, etc since monorepo structure has changed since then)
2. Big-endian support in message-coder
3. Changes to tests and implementation in paper-handler driver itself

This PR is `#3` and converts most of the driver to message-coder, with a few exceptions for response decoding
* `scan`
* `getScannerCapability`

and one exception for request encoding:
* `bufferChunk` (used by `print`)

This means the bulk of the diff to pay attention to is in `driver.ts` and `driver.test.ts`. See `Testing Plan` for more details on diff you can skim over.

## Demo Video or Screenshot
N/A

## Testing Plan
**The purpose of `deprecated_driver.ts` and `deprecated_driver.test.ts`** is to give confidence that the new driver behaves like the old driver. Both files are ~1300 lines of diff and can be skimmed over in this review if you agree with the reasoning below.

* Since there were no automated tests, `deprecated_driver.test.ts` can be thought of as "previously successful tests on main"
* `deprecated_driver` is a copy+paste of the old driver before changes in this PR
* `deprecated_driver.test.ts` is a copy+paste of the current test file with 2 differences
  * It imports and tests `deprecated_driver` instead of `driver`
  * It expects `Uint8Array`s, not `Buffer`s, to be passed to `WebUSBDevice.transferOut`

The expectation differs because `WebUSBDevice.transferOut` expects a `TypedArray`, which both `Buffer` and `Uint8Array` are. The new driver implementation passes a `Buffer` while the old passes `Uint8Array`. Both work (I have confirmed this manually on the hardware) but I wanted to be as explicit as possible when defining test expectations.

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (not done; likely will add logging at a higher level of abstraction for user actions)


kshen0:
See https://github.com/votingworks/vxsuite/pull/3499

PR #3464, opened 2023-05-15
[VxMark] VVSG Design Updates for "see more" scroll buttons

kofi-q:
## Overview

_Recommend reviewing commit 2 with `?w=1` to ignore indentation changes - a little messy otherwise: https://github.com/votingworks/vxsuite/pull/3464/commits/5ac8b8a400ff919b86d0955034ad02893a73c7c4?w=1_

Updating the "see more" scroll buttons used on the contest and review pages.

Code was duplicated across a few components, so extracted to a shared `WithScrollButtons` wrapper component  in libs/ui. Kept functionality mostly equivalent, except for:
- No longer fading out the scroll buttons when removing them after reaching the min/max scroll, so removing the logic around disabling them for a moment before fading them out. Simplifying by removing them from the DOM altogether.
- Rendering scroll buttons as regular (vs primary) buttons - they looked a little distracting as solid coloured buttons.

### Candidate Contest:
https://github.com/votingworks/vxsuite/assets/264902/dcf12ed5-f374-4f35-92f0-79efb35eea2f

### Yes/No Contest:
https://github.com/votingworks/vxsuite/assets/264902/138b862c-cf16-4d69-8027-39482cf9a091

### MS Either/Neither Contest:
https://github.com/votingworks/vxsuite/assets/264902/28e8fcd1-c7fd-46a1-ad9d-e4b95c27e9ec

### Gamepad Navigation Regression Check:
https://github.com/votingworks/vxsuite/assets/264902/20950df9-5e3f-4fda-81ff-acb381199fdd

## Testing Plan
- Updated the scroll button cypress test - excluding the new component from coverage in libs/ui otherwise, since it's not to possible to test element size-based logic in jsdom tests without some hacky DOM API hackery.

## Checklist


apps/mark/frontend/src/components/ms_either_neither_contest.tsx
```diff
@@ -327,11 +201,7 @@ export function MsEitherNeitherContest({
             gridArea: 'pick-one-label',
           }}
         >
-          <Prose>
-            <P weight="bold">
```
kofi-q, line 331: Noticed these were adding unnecessary spacing to the UI, eating up precious real estate in `XL` mode.

apps/mark/frontend/src/components/contest_screen_layout.tsx
```diff
@@ -128,5 +23,4 @@ export const ChoicesGrid = styled.div.attrs({
   display: grid;
   grid-auto-rows: minmax(auto, 1fr);
   grid-gap: 0.5rem;
-  margin-bottom: 1rem;
```
kofi-q, line 131: No longer needed - was used as a spacer to account for the scroll container shadow - handled within `WithScrollButtons` now

libs/ui/src/with_scroll_buttons.tsx
```diff
@@ -0,0 +1,203 @@
...
+const SCROLL_DISTANCE_TO_CONTENT_HEIGHT_RATIO = 0.75;
+
+const SCROLL_BUTTON_SIZE_MODE_OVERRIDES: Readonly<Record<SizeMode, SizeMode>> =
+  {
+    s: 's',
+    m: 'm',
+    l: 'm',
+    xl: 'm',
+    legacy: 'legacy',
+  };
```
adghayes, line 25: What's the VVSG rationale here? I agree it seems totally necessary, but want to understand if we have an explicit VVSG carveout or not.

adghayes (approved):
LGTM. There is always a case to make the "See More" bigger, as putting candidates below the fold already makes people unhappy. But I think this is the right tradeoff. Looks a lot better.

libs/ui/src/with_scroll_buttons.tsx
kofi-q: Yeah, nothing specifically around navigation/scroll controls, so this is more to keep the button sizes tame in our specific UI. WRT text size, VVSG does allow for some text elements to be smaller than the currently active text size, as long as crucial elements like candidate names are rendered at the selected size.

I felt we could've gotten by with capping it at `L` instead of `M`, but it falls apart in the ms-either-neither case (not sure if that contest is something we'll support long-term in its current format)

PR #3466, opened 2023-05-16
Move Manual Tally Data to Backend

### LINKED ISSUE #3448: Move Manual Tally Data to Backend
Part 1: #3440 
Part 2: In Progress

adghayes:
## Overview

Closes #3448. As a result, we longer have any frontend persistent (across restarts) state, per our architecture plan :tada::tada::tada: The data structure that was used to store it in the frontend, however, is deeply embedded in the app. So instead of really breaking up all the requests for manual tally data into individual server endpoints, we just have one main one that is called at the top level of the app. Over the course of this project that will be broken up, and the old data structure will be discarded. But keeping it now is the only path for gradual change. In summary...

**What does this PR do?**
- Adds DB storage for manual tallies, compromising between explicitly typed columns (for total number of ballots) and blobs (for entered contest results)
- Adds endpoints on the backend for post, delete, and get
- Creates and deletes write-in candidates on the backend as needed in response to entered manual tally data
- Deletes the "`backend`" abstraction on the backend (which is not to be confused with `admin/backend`), which was only being used with local storage. No more local storage!
- Moves the 'delete' and 'set' requests down in the React component hierarchy, as is our pattern
- Replaces any use of the local storage `backend` with use of the real backend.
- Temporarily removes ability to toggle between "Absentee" and "Precinct" manual data. We want to be able to enter both, and easier to just remove for now.

**What does this PR not do that needs to be done later?**
- Break up uses of manual tally data or summary data on the frontend into distinct server requests, moving them down the React component hierarchy.
- Simplify our existing `ManualTally` type which is unwieldy and has way more data than it needs. We're basically storing a contest definition in the DB for every precinct / ballot style
- Enhance manual tally data UX entry in any way

## Testing Plan
- Rewrote a lot of automated tests on frontend
- Added test coverage to new functions on backend
- Manual testing

## Checklist


apps/admin/backend/schema.sql
```diff
@@ -106,6 +106,28 @@ create table ballot_images (
...
+create table manual_tally_write_in_candidate_references (
+  manual_tally_id integer not null,
+  write_in_candidate_id varchar(36) not null,
+  primary key (manual_tally_id, write_in_candidate_id),
+  foreign key (manual_tally_id) references manual_tallies(id)
+    on delete cascade,
+  foreign key (write_in_candidate_id) references write_in_candidates(id)
+    on delete cascade
+);
+
```
adghayes, line 130: I want to maintain a system, like we have for on-screen adjudication currently, where if you create a write-in candidate it will exist as long as it is referenced. If it's no longer referenced (no more adjudications or no more manual tallies that reference it), it's removed. The alternative is having these records pile up. 

In the future we will have an "official write-in candidates" interface where you can specify official candidates and possibly do any deleting there, but @mattroe is working on the importance / timeline for that feature. Until then, we'll just auto-delete.

The auto-deleting entails some non-trivial overhead (see `store.ts` but it's confined to the store). @kofi-q please let me know if there's any way to make it cleaner or clearer.

apps/admin/backend/src/store.ts
```diff
@@ -951,24 +956,70 @@ export class Store {
...
-      `,
+          delete from write_in_candidates
+          where id = ?
+        `,
         id
       );
     }
   }
 
+  private deleteAllChildlessWriteInCandidates(): void {
```
adghayes, line 984: could probably do all this in one query with nested select statements but I don't trust `sqlite3` to optimize the queries properly

apps/admin/backend/src/store.ts
```diff
@@ -1274,6 +1325,149 @@ export class Store {
...
+          ballot_count,
+          contest_tallies
+        ) values 
+          (?, ?, ?, ?)
+        on conflict
+          (election_id, precinct_id)
+        do update set
+          ballot_count = excluded.ballot_count,
+          contest_tallies = excluded.contest_tallies
+        returning (id)
```
adghayes, line 1365: i've never seen us use `on conflict` or `returning` in our codebase but it allows some efficiencies - fewer database requests - and i think it's fair game

apps/admin/backend/schema.sql
adghayes: And @kofi-q another possible idea - instead of managing this reference table, normalize the manual tally data so the links are direct. I thought about this but I don't think that manual tally data normalizes well (we'd have to list metadata and option totals in separate rows with some nullable fields) and it doesn't have other advantages based on how we will query, but I could be convinced.

apps/admin/backend/src/util/manual_tallies.ts
```diff
@@ -0,0 +1,339 @@
...
+  }
+  resultsByCategory.set(TallyCategory.Party, contestTalliesByParty);
+
+  return {
+    overallTally,
+    resultsByCategory,
+    votingMethod,
+    timestampCreated,
+  };
+}
```
adghayes, line 239: This is all just copied from the frontend. It will someday be removed from the frontend. I could have pulled it out into `libs/utils` but was easiest just to copy, and it will be deleted from the frontend later on in this project.

apps/admin/backend/src/types.ts
```diff
@@ -392,3 +395,15 @@ export const DatabaseSerializedCastVoteRecordVotesSchema: z.ZodSchema<DatabaseSe
...
+ * @deprecated The {@link FullElectionManualTally} object is not serializable
+ * by `grout` so we have an alternative on the server side, and convert on the
+ * frontend.
+ */
+export type ServerFullElectionManualTally = Omit<
+  FullElectionManualTally,
+  'resultsByCategory'
+> & {
+  resultsByCategory: Dictionary<Dictionary<ManualTally>>;
+};
```
adghayes, line 409: We actually had the same problem on the frontend when serializing for local storage. Now we just have the problem somewhere else. This is a temporary measure while we're serving the whole full election tally to the frontend.

apps/admin/frontend/src/app_root.tsx
```diff
@@ -53,27 +48,28 @@ export function AppRoot({
   const { cardReader, printer: printerInfo } = useDevices({ hardware, logger });
 
   const [isTabulationRunning, setIsTabulationRunning] = useState(false);
-  const [manualTallyVotingMethod, setManualTallyVotingMethod] = useState(
-    VotingMethod.Precinct
-  );
```
adghayes, line 58: We agreed to just remove the toggling voting method feature temporarily as it's not relevant now: https://votingworks.slack.com/archives/CEL6D3GAD/p1683914471084739.

apps/admin/frontend/src/contexts/services_context.ts
```diff
@@ -1,25 +1,16 @@
...
-} from '../lib/backends';
 
 /**
  * Provides access to various services.
  */
 export interface ServicesContextInterface {
-  readonly backend: ElectionManagerStoreBackend;
   readonly logger: Logger;
-  readonly storage: Storage;
 }
```
adghayes, line 9: Now that this is just the logger I will, in a future commit, remove this additional context and place the logger in the AppContext. But wanted to keep any avoidable churn outside this PR.

apps/admin/frontend/jest.config.js
```diff
@@ -25,9 +25,9 @@ module.exports = {
   ],
   coverageThreshold: {
     global: {
-      statements: 83,
+      statements: 82,
       branches: 70,
-      functions: 78,
+      functions: 77,
```
adghayes, line 30: I don't care to keep or increase coverage while so much of this testing will be further rewritten soon anyway.

adghayes:
@kofi-q I'm a bit confused why the `function` test coverage is failing in `admin/backend` - I would assume the line that it indicates would be a function definition but it's not. Any suggestion? I can just reduce coverage if it's something unimportant but don't yet understand what it's complaining about in `store.ts`.

apps/admin/backend/schema.sql
kofi-q: Yeah, in the spirit of "optimise as needed", seems fine to go this route for now. I assume the downside of letting them pile up is that they show up in the UI and make it increasingly congested?

kofi-q (commented):
Looks good! Nothing blocking here - most of my comments feel like nice-to-have optimisations.

WRT the failing coverage check, it looks like there are no tests verifying that write-in candidate records aren't deleted if there are still CVR write-ins referencing them (so that anonymous function in the map call is untested, since no CVR write-ins exist)

apps/admin/backend/schema.sql
adghayes: Yeah, it's definitely a UI congestion issue rather than a data / performance congestion issue.

apps/admin/backend/src/store.ts
```diff
@@ -1274,6 +1325,149 @@ export class Store {
...
+      this.client.run(
+        `
+          insert into manual_tally_write_in_candidate_references (
+            manual_tally_id,
+            write_in_candidate_id
+          ) values ${questionMarks.join(', ')}
+        `,
+        ...params
+      );
+    }
```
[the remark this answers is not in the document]
adghayes: OK my answer here is related to my answer for https://github.com/votingworks/vxsuite/pull/3466#discussion_r1196784797, so combining response here.

When new manual tally data comes in from the frontend, we handle any temporary write-in candidates and add them to the database. That felt like _application_ logic to me, and @jonahkagan's been advocating for minimizing the logic in `store` to data management and pulling out helpers for application logic. 

However, the application logic is always paired with actually inserting the tallies into the database, which happens at the `store` level. Originally I had the store method wrapped in a transaction and then the store method + temp write-in handling wrapped in a transaction, but you can't start a transaction within a transaction and that didn't work. So I pulled the transaction out to the app level.

Not sure what the best approach is here!

apps/admin/backend/src/app.ts
```diff
@@ -393,10 +400,18 @@ function buildApi({
       return addCastVoteRecordReportResult;
     },
 
-    clearCastVoteRecordFiles(): void {
+    async clearCastVoteRecordFiles(): Promise<void> {
       const electionId = loadCurrentElectionIdOrThrow(workspace);
       store.deleteCastVoteRecordFiles(electionId);
       store.setElectionResultsOfficial(electionId, false);
+      await logger.log(
```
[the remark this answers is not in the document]
adghayes: I really don't expect logging to fail. It makes sense to me to do fire-and-forget, but I don't anticipate it would make a UX difference either way. @carolinemodic might have thoughts.

kofi-q (approved):
Simplified delete query looks great!

:shipit:

Issue #2866, opened 2023-01-04
Sign and verify all files in and out of VotingWorks equipment


benadida:
## Summary

Every file produced by VotingWorks equipment and loaded into VotingWorks equipment should be digitally signed so they can be authenticated.

## Tasks

- [ ] design of where signature and certificate information will be stored
- [ ] design of how to check that files are properly signed at load time, e.g. CVRs
- [ ] sign CVRs files out of VxScan and VxCentralScan
- [ ] sign election definition (and ballot package?) out of VxAdmin
- [ ] verify signed CVRs when loading into VxAdmin
- [ ] verify signed election definition when loading into VxMark, VxScan, VxCentralScan

PR #3476, opened 2023-05-17
Add methods for signing and verifying machine-exported artifacts

arsalansufi:
## Overview

A first step toward meeting VVSG2 data protection requirements, this PR adds methods for signing and verifying machine-exported artifacts 🔏, in the form of an `ArtifactAuthenticator` class in the auth lib.

I've stuck to the high-level plan laid out in [VxSuite Security Architecture 2022+](https://docs.google.com/document/d/1-6K-ayWUAIQAO3n50PZAAODVaXsBUkZhGoXcFNja6aI/edit?usp=sharing), specifically the doc's [Application-Level Signing](https://docs.google.com/document/d/1-6K-ayWUAIQAO3n50PZAAODVaXsBUkZhGoXcFNja6aI/edit#heading=h.teuvwrhwjtfk) and [Authenticating Artifacts](https://docs.google.com/document/d/1-6K-ayWUAIQAO3n50PZAAODVaXsBUkZhGoXcFNja6aI/edit#heading=h.t5l7q1em6muv) sections.

## Testing

- [x] Added automated tests
- [x] Manually verified that Java Card auth still works after config changes

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Intentionally punting on this for now but will open an issue to track


libs/auth/src/env_vars.ts
```diff
@@ -3,8 +3,11 @@ import { assert } from '@votingworks/basics';
 /**
  * Gets an env var and throws an error if it's not set
  */
-export function getRequiredEnvVar(envVarName: string): string {
+export function getRequiredEnvVar<T extends string>(
+  envVarName: T
+): NonNullable<NodeJS.ProcessEnv[T]> {
```
arsalansufi, line 8: This change ensures that the typings in [libs/auth/src/env.d.ts](https://github.com/votingworks/vxsuite/blob/main/libs/auth/src/env.d.ts) still apply when using this helper

script/src/run-dev/index.ts
```diff
@@ -169,7 +169,7 @@ export async function main(
             npmBinCommand({
               name: `${name}:run`,
               command:
-                'while [ ! -f build/index.js ]; do echo "Waiting for build…"; sleep 1; done; nodemon --watch build --delay 1 --exitcrash --exec NODE_ENV=development node build/index.js',
+                `while [ ! -f build/index.js ]; do echo "Waiting for build…"; sleep 1; done; nodemon --watch build --delay 1 --exitcrash --exec NODE_ENV=development VX_MACHINE_TYPE=${frontend} node build/index.js`,
```
arsalansufi, line 172: `VX_MACHINE_TYPE` is specified in prod through [this line](https://github.com/votingworks/vxsuite-complete-system/blob/a6f272bc74684807c87d80be06926550fe61ab4a/config/read-vx-machine-config.sh#L38)

libs/auth/src/java_card.ts
```diff
@@ -154,7 +154,11 @@ export class JavaCard implements Card {
   private readonly customChallengeGenerator?: () => string;
   private readonly vxCertAuthorityCertPath: string;
 
-  constructor(input: JavaCardConfig) {
+  constructor(
+    // Support specifying a custom config for tests
```
arsalansufi, line 158: Why we can't get rid of the constructor input in its entirety

libs/auth/src/openssl.ts
```diff
@@ -200,16 +200,16 @@ export function verifyFirstCertWasSignedBySecondCert(
 }
 
 /**
- * Verifies a challenge signature against an original challenge using a public key (in PEM format).
+ * Verifies a message signature against an original message using a public key (in PEM format).
```
arsalansufi, line 203: A challenge is one kind of message - shifted to the more general term in prep for new upcoming uses of this function

libs/auth/src/openssl.ts
```diff
@@ -218,11 +218,24 @@ export async function verifySignature({
...
-    challenge,
+    messageSignature,
+    message,
   ]);
 }
 
+/**
+ * Signs a message using a private key file (in PEM format)
+ */
+export async function signMessageUsingPrivateKeyFile({
```
arsalansufi, line 229: This is just for dev. In prod, we'll be using the TPM and a variant of [this script](https://github.com/votingworks/vxsuite-complete-system/blob/a6f272bc74684807c87d80be06926550fe61ab4a/app-scripts/sign.sh)

libs/auth/src/artifact_authenticator.ts
```diff
@@ -0,0 +1,201 @@
...
+    switch (this.signingMachinePrivateKey.source) {
+      case 'file': {
+        return await signMessageUsingPrivateKeyFile({
+          message,
+          privateKey: this.signingMachinePrivateKey.path,
+        });
+      }
+      /* istanbul ignore next */
+      case 'tpm': {
+        // TODO: Move TPM signing script from vxsuite-complete-system to vxsuite and call here
```
arsalansufi, line 188: Decided to defer this to a follow-up PR since this PR was getting large enough

libs/auth/src/config.ts
```diff
@@ -4,6 +4,46 @@ import { isVxDev } from '@votingworks/utils';
...
+
+function getMachineCertPathAndPrivateKey(): {
+  certPath: string;
+  privateKey: FileKey | TpmKey;
+} {
+  const machineType = getRequiredEnvVar('VX_MACHINE_TYPE');
+  const machineCertFileName =
+    machineType === 'admin'
+      ? 'vx-admin-cert-authority-cert.pem'
+      : `vx-${machineType}-cert.pem`;
```
arsalansufi, line 26: For reference, the non-VxAdmin machine certs are not cert authority (CA) certs (certs capable of issuing other certs), hence their different naming convention. This convention is consistent with the [create-machine-cert.sh script in vxsuite-complete-system](https://github.com/votingworks/vxsuite-complete-system/blob/a6f272bc74684807c87d80be06926550fe61ab4a/config/admin-functions/create-machine-cert.sh#L11-L15)

libs/auth/src/config.ts
```diff
@@ -4,6 +4,46 @@ import { isVxDev } from '@votingworks/utils';
...
+    certPath: path.join(__dirname, '../certs/dev', machineCertFileName),
+    privateKey: {
+      source: 'file',
+      path: path.join(
+        __dirname,
+        '../certs/dev',
+        `vx-${machineType}-private-key.pem`
+      ),
+    },
+  };
```
arsalansufi, line 44: In subsequent commits, I generate all the new non-VxAdmin machine keys and certs that we need for dev

libs/auth/src/config.test.ts
```diff
@@ -1,29 +1,51 @@
...
   ...jest.requireActual('@votingworks/utils'),
   isVxDev: jest.fn(),
 }));
 
 beforeEach(() => {
-  process.env.NODE_ENV = 'test';
-  process.env.VX_CONFIG_ROOT = '/vx/config';
-
+  (process.env.NODE_ENV as string) = 'test';
+  (process.env.VX_CONFIG_ROOT as string) = '/vx/config';
```
arsalansufi, line 18: Need these casts now because I've made `process.env` values read-only

libs/auth/src/artifact_authenticator.ts
```diff
@@ -0,0 +1,201 @@
...
+      publicKey: signingMachinePublicKey,
+    });
+  }
+
+  private serializeSignatureFileContents(
+    signatureFileContents: SignatureFileContents
+  ): Buffer {
+    const { signature, signingMachineCert } = signatureFileContents;
+    return Buffer.concat([
+      Buffer.from([signature.length]),
```
arsalansufi, line 148: ECC signature length can vary ever so slightly, hence the inclusion of this length value. See [this Slack thread](https://votingworks.slack.com/archives/CEL6D3GAD/p1684342298051899) for more context

libs/auth/src/artifact_authenticator.ts
```diff
@@ -0,0 +1,201 @@
...
+      signature,
+      signingMachineCert,
+    ]);
```
arsalansufi, line 151: Settled on simple buffer concatenation instead of, say, a JSON file, because the latter made it too easy to accidentally add new unverified/unsigned fields.

I actually tried JSON out and myself made this mistake, accidentally moving the version number (which you'll see below) out of the signed message and into a top-level JSON field, only to remember that the new location would allow someone malicious to manually modify the version number and thereby influence parsing.

libs/auth/src/artifact_authenticator.ts
```diff
@@ -0,0 +1,201 @@
+import { Buffer } from 'buffer';
```
arsalansufi, line 1: @benadida This is the most relevant file for you to security review

libs/auth/src/artifact_authenticator.ts
```diff
@@ -0,0 +1,202 @@
...
+      signature,
+      signingMachineCert,
+    ]);
+  }
+
+  private deserializeSignatureFileContents(
+    file: Buffer
+  ): SignatureFileContents {
+    const signatureLength = file[0];
+    assert(signatureLength !== undefined);
```
benadida, line 159: let's also assert that the length is within some tight bounds we expect.

and while we're at it, that the total length of the file is at least long enough that we have a normal looking signingMachineCert. Paranoia paranoia here about potential parsing attacks :)

libs/auth/src/artifact_authenticator.ts
```diff
@@ -0,0 +1,202 @@
...
+    const signature = file.subarray(1, signatureLength + 1);
+    const signingMachineCert = file.subarray(signatureLength + 1);
+    return { signature, signingMachineCert };
+  }
+
+  private async constructMessage(artifact: Artifact): Promise<Buffer> {
+    const messageFormatVersion = Buffer.from('1', 'utf-8');
```
benadida, line 166: utf-8 or ascii? Does utf-8 make it take up two bytes, I forget?

libs/auth/src/artifact_authenticator.ts
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Verifies the authenticity of the provided artifact using its signature file, which is expected
+   * to be found at the same file path as the artifact, but with a .vxsig extension, e.g.
+   * /path/to/artifact.txt.vxsig.
+   */
+  async authenticateArtifactUsingSignatureFile(
+    artifact: Artifact
+  ): Promise<Result<void, Error>> {
+    try {
+      await this.authenticateArtifactUsingSignatureFileHelper(artifact);
```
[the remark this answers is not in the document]
benadida: Actually, I do see a reason for the helper, but I'd use it to separate the file paths from the cryptography, cause intermingling those two feels unnecessary.

How about `authenticateArtifactUsingSignatureFile` goes and find the `.vxsig` file, loads it, and then calls the helper with just `Buffer`s?

benadida (approved):
on security, looks good, a couple of paranoid tweaks suggested, and a helper-function tweak suggested.

libs/auth/src/artifact_authenticator.ts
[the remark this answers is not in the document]
arsalansufi: Okay, this ended up being a really helpful nudge! Given that the contents of the signature "file" may in some cases not be stored in a file at all and instead in, say, a QR code on a printed report, I've better separated cryptography and persistence logic: https://github.com/votingworks/vxsuite/pull/3476/commits/aebe9063a6c4e366486f75777a06fe17bec46627

libs/auth/src/artifact_authenticator.ts
arsalansufi: For my understanding @benadida, what exactly would validation of the length of the `signingMachineCert` catch that wouldn't be caught during subsequent parsing of the cert? Is the thinking just that it's safer to short circuit early?

benadida: To be clear this is just layers of defense and extra paranoia. I worry about to-be-discovered buffer bugs in OpenSSL that could be triggered by malicious inputs, and so any ways we can constrain what we feed OpenSSL is a good thing.

arsalansufi: Got it! That makes sense. Added both length checks accordingly: https://github.com/votingworks/vxsuite/pull/3476/commits/e32eb684370a543e9d1bf59c39de3fd27c502f84 https://github.com/votingworks/vxsuite/pull/3476/commits/5d4e48b111e21beb86fa541a94341c228dbaa6cc

libs/auth/src/artifact_authenticator.ts
```diff
@@ -0,0 +1,201 @@
...
+    } catch {
+      // TODO: Log raw error
```
kshen0, line 88: Is this TODO addressed in a later commit?

kshen0 (approved):

libs/auth/src/openssl.ts
arsalansufi: Good q! It's already enforced in the sense that we have no private key files in prod, just the private key in the TPM, which can't be directly accessed.

I considered a `NODE_ENV` check, but there's some nuance there since we have some prod-like environments (VxDev, integration tests) where this function is okay to use.

All this said, I do think a comment makes sense! And maybe some code tweaks - let me think some more about this

arsalansufi: Okay, thought some more about this and restructured this helper to support both the dev and prod (i.e. file key and TPM key) use cases: https://github.com/votingworks/vxsuite/pull/3476/commits/8344fdc081601a8c5da15049ddc37db713e78e09

I'd previously not done this because I wasn't planning on using OpenSSL for signing messages with the TPM key (I was planning on using [this script](https://github.com/votingworks/vxsuite-complete-system/blob/a6f272bc74684807c87d80be06926550fe61ab4a/app-scripts/sign.sh) instead). But I just successfully used OpenSSL to sign with the TPM key on my ThinkPad and am now leaning toward that approach for consistency with cert creation functions and consistency in general.

I haven't actually added the code for signing messages with the TPM key via OpenSSL to this PR since I think that merits a separate review.

libs/auth/src/artifact_authenticator.ts
arsalansufi: Not yet! I have an issue open for a few logs that I want to add: https://github.com/votingworks/vxsuite/issues/3484

PR #3477, opened 2023-05-17
support big-endian in uint16 and uint32 coders

kshen0:
## Overview
Supports big-endian `Uint16` and `Uint32` in `message-coder`.

## Testing Plan
* Updated automated tests for big endian case

## Checklist

- [ x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

### DIFF
--- a/libs/message-coder/src/uint16_coder.ts
+++ b/libs/message-coder/src/uint16_coder.ts
@@ -11,11 +11,25 @@ import {
 } from './types';
 import { UintCoder } from './uint_coder';
 
+interface Uint16CoderOptions {
+  littleEndian: boolean;
+}
+
 /**
  * Coder for a uint16, aka a 16-bit unsigned integer. Uses little-endian byte
  * order.
  */
 export class Uint16Coder extends UintCoder {
+  private readonly littleEndian: boolean;
+
+  constructor(
+    enumeration?: unknown,
+    options: Uint16CoderOptions = { littleEndian: true }
+  ) {
+    super(enumeration);
+    this.littleEndian = options?.littleEndian;
+  }
+
   bitLength(): BitLength {
     return 16;
   }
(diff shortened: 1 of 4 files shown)


libs/message-coder/src/uint32_coder.ts
```diff
@@ -11,11 +11,25 @@ import {
...
 export class Uint32Coder extends UintCoder {
+  private readonly littleEndian: boolean;
+
+  constructor(
+    enumeration?: unknown,
+    options: Uint32CoderOptions = { littleEndian: true }
+  ) {
+    super(enumeration);
+    this.littleEndian = options?.littleEndian;
+  }
```
kshen0, line 31: Wasn't sure if we should put this in `UintCoder`, but since we only need big endian support in Uint16 and Uint32 right now, it felt like overkill to add it at the next level up.

libs/message-coder/src/uint16_coder.ts
```diff
@@ -11,11 +11,25 @@ import {
...
 /**
  * Coder for a uint16, aka a 16-bit unsigned integer. Uses little-endian byte
  * order.
  */
 export class Uint16Coder extends UintCoder {
+  private readonly littleEndian: boolean;
+
+  constructor(
+    enumeration?: unknown,
+    options: Uint16CoderOptions = { littleEndian: true }
```
eventualbuddha, line 27: This is equivalent for now, but if we add additional properties it allows specifying defaults for one at a time rather than all or nothing.
```suggestion
    { littleEndian = true }: Partial<Uint16CoderOptions> = {}
```

libs/message-coder/src/uint32_coder.ts
```diff
@@ -11,11 +11,25 @@ import {
...
 /**
  * Coder for a uint32, aka a 32-bit unsigned integer. Uses little-endian byte
  * order.
  */
 export class Uint32Coder extends UintCoder {
+  private readonly littleEndian: boolean;
+
+  constructor(
+    enumeration?: unknown,
+    options: Uint32CoderOptions = { littleEndian: true }
```
eventualbuddha, line 27: Ditto.
```suggestion
    { littleEndian = true }: Partial<Uint32CoderOptions> = {}
```

eventualbuddha (approved):

PR #3480, opened 2023-05-18
[VxAdmin] Reduce unnecessary rapid re-renders of the full React tree

kofi-q:
## Overview

Noticed VxAdmin was running slowly and stuttering sometimes, CSS transitions weren't working, and my computer generally seemed to be working a little harder than it should.

Dug a bit and realised our auth status polling is done at the root component level and triggers a re-render of everything (every 100ms - just short enough to kill my beautiful transitions).

Started off by using the `isDataEqual()` react-query param, but got console warnings about it being deprecated, so using the unfortunately named `structuralSharing()` param instead. This does mean we're doing an object compare every 100ms, but it's a negligible cost compared to recomputing the render tree. Open to change suggestions if there are better workarounds here.

@adghayes - not 100% sure, but this could be what was slowing your machine to crawl whenever you were presenting.

### Before:
https://github.com/votingworks/vxsuite/assets/264902/5b059336-66ba-44d5-99fc-1fe1b7c36c3f

### After:
https://github.com/votingworks/vxsuite/assets/264902/54358dcb-673a-4b5a-af16-2a668844397d

## Testing Plan
- Listened to my computer fan 😜 

## Checklist

### DIFF
--- a/apps/admin/frontend/src/api.ts
+++ b/apps/admin/frontend/src/api.ts
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import React from 'react';
 import type { Api } from '@votingworks/admin-backend';
 import {
@@ -59,6 +60,15 @@ export const getAuthStatus = {
     const apiClient = useApiClient();
     return useQuery(this.queryKey(), () => apiClient.getAuthStatus(), {
       refetchInterval: AUTH_STATUS_POLLING_INTERVAL_MS,
+      structuralSharing(oldData, newData) {
+        if (!oldData) {
+          return newData;
+        }
+
+        // Prevent infinite re-renders of the app tree:
+        const isUnchanged = _.isEqual(oldData, newData);
+        return isUnchanged ? oldData : newData;
+      },
     });
   },
 } as const;


eventualbuddha (approved):

adghayes:
Nice optimization!

arsalansufi:
+1 thanks for finding this!

I am a bit curious why this behavior isn't applying by default 🤔. Since I believe structural sharing is an optimization that's enabled by default on the version of react-query that we're on: https://tkdodo.eu/blog/react-query-render-optimizations#structural-sharing

arsalansufi:
Conducting a quick experiment of my own on the latest main 🧑‍🔬, I'm interestingly not seeing rapid re-renders by default. I only see rapid re-renders when I explicitly set `structuralSharing` to false 🤔

https://github.com/votingworks/vxsuite/assets/12616928/293db458-531c-49b7-bb1a-6eef230d2dea

kofi-q:
> I am a bit curious why this behavior isn't applying by default

I was confused by that too - was trying to figure out it didn't have the intended effect here, but that I think about it, it could be because we're passing the full object down the tree and that's what triggers the re-renders. So even when react-query persists the internal references, maybe the outer object reference changes and triggers re-renders? If that's the case, the longer-term solution would be to update any child consumers to use the query directly, maybe?

kofi-q:
> Conducting a quick experiment of my own on the latest main 🧑‍🔬, I'm interestingly not seeing rapid re-renders by default. I only see rapid re-renders when I explicitly set structuralSharing to false 🤔

Oh, just saw this! That's interesting - maybe it's something env related? Curious to try again in a new clone of the repo - don't know what could be causing it to behave differently here

arsalansufi:
I'll try again tomorrow in a fresh env as well, but also feel free to merge this in the interim if it unblocks you! We can always undo this once we get to the bottom of it

kofi-q:
Sounds good - thanks! Might just leave this open till we learn more, since I can iterate locally with this workaround in place for now

adghayes:
Any developments here? I'd prefer to merge it to make development a little easier.

kofi-q:
Oh thanks, forgot about this - doesn't hurt to commit, so will merge for now and try the fresh-env test tomorrow to see if the issue will repro there (and revert if it doesn't)

PR #3482, opened 2023-05-19
remove the shortcut open/close polls screens.

benadida:
## Overview

Removed the shortcut screens "do you want to open/close the polls?" because it became confusing to change the ballot bag before the hard limit is reached, you'd have insert poll worker card, tap "no" explicitly to move to the next screen, and that's just unintuitive.

## Testing Plan

Manually tested and updated automated tests.

## Checklist

~- [ ] I have added a screenshot and/or video to this PR to demo the change~
- [ X ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

### DIFF
--- a/frontends/precinct-scanner/jest.config.js
+++ b/frontends/precinct-scanner/jest.config.js
@@ -13,9 +13,9 @@ module.exports = {
   ],
   coverageThreshold: {
     global: {
-      statements: 94.98,
+      statements: 94.96,
       branches: 87,
-      functions: 88.4,
+      functions: 88.34,
       lines: 95,
     },
   },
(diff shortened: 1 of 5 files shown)


carolinemodic (approved):
We will need to update docs to reflect this, people are trained on the previous screens already so I'm not really sure this is worth doing even though I do like this better overall but if you and Roe agreed to remove it then thats fine.

benadida:
> We will need to update docs to reflect this, people are trained on the previous screens already so I'm not really sure this is worth doing even though I do like this better overall but if you and Roe agreed to remove it then thats fine.

Checked with Ginny, and discussed with Roe last week, everyone is good with this, so merging.

PR #3489, opened 2023-05-20
pre-warm cache of tts

benadida:
## Overview

Make use of the new kiosk-browser API that caches spoken utterances, to warm the text-to-speech cache.

## Testing Plan

Tested manually

## Checklist

### DIFF
--- a/frontends/bmd/src/app_root.tsx
+++ b/frontends/bmd/src/app_root.tsx
@@ -16,9 +16,11 @@ import {
   PrecinctId,
   BallotStyleId,
   PrecinctSelection,
+  getCandidatePartiesDescription,
 } from '@votingworks/types';
 import { decodeBallot, encodeBallot } from '@votingworks/ballot-encoder';
 import 'normalize.css';
+import pluralize from 'pluralize';
 
 import Gamepad from 'react-gamepad';
 import { useHistory } from 'react-router-dom';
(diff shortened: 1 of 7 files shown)


frontends/bmd/src/pages/admin_screen.tsx
```diff
@@ -71,6 +75,100 @@ export function AdminScreen({
     fetchElection();
   }
 
+  useEffect(() => {
```
benadida, line 78: this brute-forcing of strings to pre-warm the cache is... not my favorite approach. But, it has the virtue of being simple, not super likely to fail, and if it fails, the worst outcome is some strings aren't cached and will be when they're later spoken.

frontends/bmd/src/app_root.tsx
```diff
@@ -535,8 +538,98 @@ export function AppRoot({
     /* istanbul ignore else */
     if (newElectionDefinition) {
       updateElectionDefinition(newElectionDefinition);
+
+      void (async () => {
```
benadida, line 542: this isn't the most resilient approach, as it assumes a lot about what the subsequent strings are. But, for a first pass, it has the benefit of being unlikely to fail in a way that breaks the app. If it fails to cache, worst case is that the utterances are delayed on first use.

benadida: putting this here in the app_root also means that loading the election is not in any way delayed. The utterances are generated in the background, even if the election manager card is removed.

carolinemodic: So if we miss a string here the failure mode is it just takes a while to be said?

carolinemodic (approved):

PR #3493, opened 2023-05-22
apps/scan/backend: Ensure scanner mocks simulate paper status after scan

jonahkagan:
## Overview

There's a race condition in the interaction between scanner mocks and the state machine. The state machine [tries to check the paper status](https://github.com/votingworks/vxsuite/blob/80511f6597388683b51c23126f7f31eaa971d78b/apps/scan/backend/src/scanners/custom/state_machine.ts#L828-L835) after completing the scan command. If it still says READY_TO_SCAN, then the scanner transitions to retry scanning, then checks the paper status again. If it then says READY_TO_EJECT, the state machine [moves to the rejecting state](https://github.com/votingworks/vxsuite/blob/80511f6597388683b51c23126f7f31eaa971d78b/apps/scan/backend/src/scanners/custom/state_machine.ts#L879).

Since we weren't updating the mocks to switch to a READY_TO_EJECT state soon enough, it was possible that it would switch in between these two checks. To mitigate, I built a little helper function that switches the mock scanner status return value immediately when the scan command completes. It's a small step back towards building a stateful scanner mock, which I don't love, but I couldn't think of a better way to fix the issue.

One alternative idea is to try to pause time and stop the state machine from advancing until we're able to reconfigure the mocks, but I felt like trying to mess around with fake timers would potentially create even more problems and felt like a bigger project to consider if we keep running into issues.

## Testing Plan
Ran tests on CI ~10 times and confirmed no failures.

## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/custom/app_scan.test.ts
+++ b/apps/scan/backend/src/scanners/custom/app_scan.test.ts
@@ -32,7 +32,11 @@ import {
   waitForStatus,
 } from '../../../test/helpers/shared_helpers';
 import { SheetInterpretation } from '../../types';
-import { ballotImages, withApp } from '../../../test/helpers/custom_helpers';
+import {
+  ballotImages,
+  simulateScan,
+  withApp,
+} from '../../../test/helpers/custom_helpers';
 
 jest.setTimeout(20_000);
 /**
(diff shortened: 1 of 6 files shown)


apps/scan/backend/src/scanners/custom/app_scan.test.ts
```diff
@@ -88,10 +92,8 @@ test('configure and scan hmpb', async () => {
         type: 'ValidSheet',
       };
 
-      mockScanner.scan.mockResolvedValue(ok(await ballotImages.completeHmpb()));
+      simulateScan(mockScanner, await ballotImages.completeHmpb());
       await apiClient.scanBallot();
-      await expectStatus(apiClient, { state: 'scanning' });
```
jonahkagan, line 93: With this new approach, we can't guarantee that the state machine will stay in the `scanning` state long enough to check it reliably, so I had to just remove that check, which is not ideal. The only reason it was reliably in the `scanning` state before is that it would be stuck in a retry loop.

jonahkagan:
Also noting that the logic to check paper status after scan has a note that says it's a holdover from the Plustek scanner. We could potentially remove that logic, which would remove the root cause of this flakiness. I don't have enough experience with the Custom scanner to know how feasible that is.

PR #3495, opened 2023-05-22
Paper handler driver test

kshen0:
## Overview
Adds tests for most commands in the existing paper handler driver (ie. tests for legacy implementation). After this merges and #3443 is rebased on main, the test file diff should be easier to read.

## Demo Video or Screenshot
N/A

## Testing Plan
Automated

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (No user actions or storage writes introduced. Error handling is not done yet so logging would be premature.)

PR #3499, opened 2023-05-22
Kevin/message coder paper handler driver

kshen0:
## Overview
Converts most of the driver to message-coder, with a few exceptions
* The device's scan command response doesn't return a prefix byte indicating length of scan data; `MessageCoder`'s `byteArray` requires a length. We could compute the length inside the scan handling loop and redefine the coder each time. However, this feels inconsistent with other `MessageCoder` usage. I think the better solution is to look into supporting `unboundedByteArray` that requires no length and must be the last property in the coder (similar to `unboundedString`). I'll ticket this if we agree that approach is best.
* `bufferChunk()` suffers from the same problem, but in this case I do redefine the coder for each call since the tradeoffs are different - if the arbitrary-length print data is not included in the coder, we can't use our existing `transferOutGeneric` helper and would have to special case transfers out for this use case
* `getScannerCapability` is unused by the state machine at the moment so I didn't convert it; I'll convert if we need it

Also adds a basic CLI for interacting with the driver.

## Demo Video or Screenshot
N/A

## Testing Plan
* Updated tests added in https://github.com/votingworks/vxsuite/pull/3495
* Added tests for `scan` and `bufferChunk`

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (not done; likely will add logging at a higher level of abstraction for user actions)


libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -244,11 +233,14 @@ export class PaperHandlerDriver {
   /**
    * Transfers data out on the real time bulk out endpoint.
    */
-  transferOutRealTime(command: Uint8[]): Promise<USBOutTransferResult> {
-    return this.webDevice.transferOut(
-      REAL_TIME_ENDPOINT_OUT,
-      new Uint8Array(command)
-    );
+  transferOutRealTime(requestId: Uint8): Promise<USBOutTransferResult> {
+    // TODO error handling
```
kshen0, line 237: Leaving this as a TODO because we'll have a better idea of how to best handle errors when we get more into product development

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -275,9 +267,59 @@ export class PaperHandlerDriver {
...
+    }
+    const data = encodeResult.ok();
+    assert(data);
+    return this.webDevice.transferOut(GENERIC_ENDPOINT_OUT, data);
+  }
+
+  /**
+   * Send commands or data on the generic bulk out endpoint.
+   */
+  transferOutGenericDeprecated(
```
kshen0, line 296: This is still used by `print()`

kshen0:
Needed to rebase and it was easier to spin off into a new PR: https://github.com/votingworks/vxsuite/pull/3545

PR #3500, opened 2023-05-23
feat(message-coder): add `byteArray` & related coders

eventualbuddha:
## Overview
These coder handle byte arrays, especially the common pattern of an encoded length of bytes followed by the number of bytes specified by that length.

## Demo Video or Screenshot
```ts
export const RealTimeExchangeResponse = message({
  startOfPacket: literal(0x82),
  requestId: uint8(),
  token: literal(TOKEN),
  returnCode: uint8(),
  optionalBytes: byteArrayWithLengthPrefix(uint8()),
});
```

## Testing Plan
Automated.

## Checklist


kshen0:
> I don't love the name dynamicBytes. Got a better suggestion?
`variableLengthByteSequence`? `arbitraryBytes`?

`unboundedBytes` would be consistent with `unboundedString` but I guess in this case it is actually bounded

libs/message-coder/src/dynamic_bytes.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  expect(coder.default()).toEqual(Uint8Array.of());
+});
+
+test('bitLength', () => {
+  const coder = dynamicBytes(uint8());
+  expect(coder.bitLength(Uint8Array.of(1, 2, 3))).toEqual(ok(8 + 24));
+});
+
+test('encode', () => {
+  const coder = dynamicBytes(uint8());
```
kshen0, line 24: It's overkill to test every coder, but can we test `dynamicBytes(literal(...))`? `literal` and `uint8` are the 2 cases I've seen in the paper-handler manual.

libs/message-coder/src/dynamic_bytes.ts
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @example
+ *
+ * ```ts
+ * import { dynamicBytes, uint8 } from '@votingworks/message-coder';
+ *
+ * const coder = dynamicBytes(uint8());
+ * coder.encode(Uint8Array.from([1, 2, 3]));
+ * // => Uint8Array.of(3, 1, 2, 3)
+ * //           length ┘  └──┴──┴ data
```
kshen0, line 98: Very useful example

kshen0 (approved):

libs/message-coder/src/dynamic_bytes.test.ts
eventualbuddha: `literal` doesn't actually work here, but I added a couple variants to achieve what you want:

```ts
byteArray(64); // encodes/decodes 64 bytes as-is
byteArrayWithLengthPrefix(uint8()); // encodes/decodes 0-255 bytes by encoding the length followed by the bytes
byteArrayWithLengthPrefix(64); // encodes/decodes 64 bytes by encoding the (fixed) length followed by the bytes
```

kshen0 (approved):
Name change LGTM. Thanks for adding the fixed length decoder.

PR #3505, opened 2023-05-23
fix(tts): stop utterance repeating, it's unreliable in its effects. 

benadida:
Also turn off tts for poll worker screen, it's more confusing than helpful.

### DIFF
--- a/frontends/bmd/src/pages/poll_worker_screen.tsx
+++ b/frontends/bmd/src/pages/poll_worker_screen.tsx
@@ -377,6 +377,13 @@ export function PollWorkerScreen({
     appPrecinct
   );
 
+  // Disable the audiotrack when on poll-worker screen
+  useEffect(() => {
+    const initialMuted = screenReader.isMuted();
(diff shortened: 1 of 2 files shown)


frontends/bmd/src/pages/poll_worker_screen.tsx
```diff
@@ -377,6 +377,13 @@ export function PollWorkerScreen({
     appPrecinct
   );
 
+  // Disable the audiotrack when on poll-worker screen
```
carolinemodic, line 380: did you mean to include this?

carolinemodic (approved):

frontends/bmd/src/pages/poll_worker_screen.tsx
benadida: yes I did. I don't think it's useful in Mississippi at all and it was a contributing factor to things sounding bad. I don't want to give the TTS system a workout if unnecessary.

PR #3518, opened 2023-05-26
Support message signing with the TPM as well as signing of larger messages

arsalansufi:
## Overview

This PR handles two messaging signing needs for VVSG2 data protection:

- Signing with the TPM private key in prod
- Signing of larger messages by using streams

While I know these two needs seem unrelated, their implementations ended up related. More on this in GitHub comments.

## Testing

- [x] Updated automated tests

## Checklist


libs/auth/src/shell.ts
```diff
@@ -1,10 +1,13 @@
 import { Buffer } from 'buffer';
 import { spawn } from 'child_process';
+import { Stream } from 'stream';
 import { assert } from '@votingworks/basics';
 
 interface RunCommandOptions {
   /** A function to run on child process close, e.g. for cleanup */
   onClose?: () => Promise<void>;
+  /** Data to pipe to the child process's standard input */
+  stdin?: Stream;
```
arsalansufi, line 10: In case you're wondering why I've used the `Stream` type here and not the more specific `NodeJS.ReadableStream`, you'll see a few commits from now that I need to support using a `CombinedStream` (a la [combined-stream](https://github.com/felixge/node-combined-stream)), which extends `Stream` and not specifically `NodeJS.ReadableStream`.

Another option here would be `NodeJS.ReadableStream | CombinedStream`. Will think some more and potentially change.

libs/auth/src/openssl.ts
```diff
@@ -434,3 +414,109 @@ export async function createCert(input: CreateCertInput): Promise<Buffer> {
...
+ * Digitally signs a message using the specified private key.
+ *
+ * Uses a slightly roundabout control flow for permissions purposes:
+ *
+ * signMessage() --> ../src/intermediate-scripts/sign-message --> signMessageHelper()
+ *
+ * When using TPM keys in production, we need sudo access. Creating the intermediate sign-message
+ * script allows us to grant password-less sudo access to the relevant system users for just that
+ * operation, instead of having to grant password-less sudo access more globally.
+ */
```
arsalansufi, line 507: I'm using the same pattern that I landed on for `createCert`:

https://github.com/votingworks/vxsuite/blob/417f9242cfba32b266911265e0ef5faf7e6dad12/libs/auth/src/openssl.ts#L452-L464

Just as I did for `createCert`, I'll be opening a PR in vxsuite-complete-system for the permissions update:

https://github.com/votingworks/vxsuite-complete-system/blob/a6f272bc74684807c87d80be06926550fe61ab4a/config/sudoers#L33

libs/auth/src/openssl.ts
```diff
@@ -434,3 +414,109 @@ export async function createCert(input: CreateCertInput): Promise<Buffer> {
...
+   * We use streaming APIs to process the message since the message could be large and infeasible
+   * to store in memory (i.e. in a Buffer), e.g. the message might include the contents of a
+   * multi-GB zip file.
+   *
+   * A seeming alternative would be to require a file path since the openssl command streams under
+   * the hood, but many of our messages will include the contents of a file preceded by a custom
+   * prefix, so there is no file path to provide (unless a temporary file is created, which too
+   * would have to be done with streaming APIs).
+   */
+  message: Stream;
```
arsalansufi, line 436: I did some manual testing with a dummy 2GB file. Creating its VVSG2 data protection signature file took on average between 3 to 5 seconds on my machine with the streaming approach that I've landed on. Earlier iterations without streaming caused my JS heap to run out of memory. Good validation!

arsalansufi:
@benadida I don't think there's much here for you to security review, but if you did want to sanity check any commits, it'd probably be https://github.com/votingworks/vxsuite/pull/3518/commits/62263250fbd428944514a0faa2da71e8161bf7d6 and https://github.com/votingworks/vxsuite/pull/3518/commits/d72e2e265a6af65dd675fe8ffdd7c2122f43cf94

libs/auth/src/openssl.test.ts
```diff
@@ -493,37 +501,147 @@ test.each<{
...
+      OPENSSL_TPM_ENGINE_NAME,
+      '-passin',
+      `pass:${TPM_KEY_PASSWORD}`,
+    ],
+  },
+])(
+  'signMessageHelper - $description',
+  async ({ signingPrivateKey, expectedOpensslSignatureRequestParams }) => {
+    setTimeout(() => {
+      // process.stdin.destroy();
```
jonahkagan, line 583: remove?

jonahkagan (approved):

libs/auth/src/openssl.test.ts
arsalansufi: 🙈 Thank you for catching! https://github.com/votingworks/vxsuite/pull/3518/commits/f0b5c7a6c083e1c13f5d82a3d5e080e41b99607f

PR #3522, opened 2023-05-27
fix(election-manager): fixed table column alignment for external tallies

benadida:
Another majestic PR that took serious PhD thinking to fix my serious PhD bug.

Screenshot 2023-05-27 at 10 25 14 AM


carolinemodic (approved):
monumental change

PR #3523, opened 2023-05-27
fix(precinct-scanner): properly handle all the ways plustek crashes

benadida:
Rework the error handling for a disconnected plustek to always prompt for unplug / replug.

Wait 3s before deciding what to show so that we don't show an unnecessary and confusing error if the cord is unplugged.

### DIFF
--- a/frontends/precinct-scanner/src/app_root.tsx
+++ b/frontends/precinct-scanner/src/app_root.tsx
@@ -675,12 +675,9 @@ export function AppRoot({
         return <ScanProcessingScreen />;
       case 'unrecoverable_error':
         return (
-          <ScanErrorScreen
-            error={scannerStatus.error}
-            isTestMode={isTestMode}
-            scannedBallotCount={scannerStatus.ballotsCounted}
-            restartRequired
-            powerConnected={computer.batteryIsCharging}
+          <SetupScannerScreen
+            batteryIsCharging={computer.batteryIsCharging}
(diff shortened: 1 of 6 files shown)


frontends/precinct-scanner/src/app_root.tsx
```diff
@@ -675,12 +675,9 @@ export function AppRoot({
         return <ScanProcessingScreen />;
       case 'unrecoverable_error':
         return (
-          <ScanErrorScreen
-            error={scannerStatus.error}
-            isTestMode={isTestMode}
-            scannedBallotCount={scannerStatus.ballotsCounted}
-            restartRequired
-            powerConnected={computer.batteryIsCharging}
+          <SetupScannerScreen
```
carolinemodic, line 678: don't you also need to update the app_root state that you changed in the last PR to now point here instead of internal connection problem? Or is that the same as this screen?

carolinemodic (approved):

frontends/precinct-scanner/src/app_root.tsx
benadida: I put the unplug/replug flow inside `<SetupScannerScreen>`, so both this `unrecoverable_error` path and the `disconnected` path now point to this same flow.

`<ScanErrorScreen>` now only handles ballot-not-counted issues.

One could argue that there's a worthwhile renaming refactor to be done here... I chose not to do that at this time to keep things simpler.

PR #3534, opened 2023-06-01
Apply VVSG2 data protection to (i.e. sign and verify) CVR files

arsalansufi:
_Easiest reviewed by commit_

## Overview

This PR applies VVSG2 data protection to CVR files. On export, scanners will now sign CVRs, and on import, VxAdmin will now verify those signatures to ensure authenticity. 

Below is a screen capture of what could make for a great security demo! I first load some CVRs into VxAdmin. I then edit those same CVRs, try reloading them, and am prevented from doing so.

https://github.com/votingworks/vxsuite/assets/12616928/79766941-8589-41af-9aa6-cf942cc261b8

## Testing

- [x] Updated automated tests
- [x] Tested end-to-end, from scanner to VxAdmin, manually

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced ➡️ Relying on existing logging paths

### DIFF
--- a/apps/admin/backend/src/cvr_files.ts
+++ b/apps/admin/backend/src/cvr_files.ts
@@ -350,7 +350,7 @@ export function getAddCastVoteRecordReportErrorMessage(
     case 'report-access-failure':
       return 'Unable to access cast vote record report for import.';
     case 'cast-vote-records-authentication-error':
-      return 'Unable to authenticate cast vote records. Try re-exporting from the scanner.';
+      return 'Unable to authenticate cast vote records. Try exporting them from the scanner again.';
     case 'invalid-report-structure':
       return 'Cast vote record report has invalid file structure.';
     case 'malformed-report-metadata':
--- a/apps/admin/backend/src/app.cvr_files.test.ts
+++ b/apps/admin/backend/src/app.cvr_files.test.ts
@@ -460,7 +460,7 @@ test('cast vote records authentication error', async () => {
     err({
       type: 'cast-vote-records-authentication-error',
       message:
-        'Unable to authenticate cast vote records. Try re-exporting from the scanner.',
+        'Unable to authenticate cast vote records. Try exporting them from the scanner again.',
     })
   );
   expect(logger.log).toHaveBeenLastCalledWith(
@@ -469,7 +469,7 @@ test('cast vote records authentication error', async () => {
     expect.objectContaining({
       disposition: 'failure',
       message:
-        'Unable to authenticate cast vote records. Try re-exporting from the scanner.',
+        'Unable to authenticate cast vote records. Try exporting them from the scanner again.',
     })
   );
   expect(await apiClient.getCastVoteRecordFiles()).toHaveLength(0);
--- a/libs/ui/src/unconfigured_election_screen.test.tsx
+++ b/libs/ui/src/unconfigured_election_screen.test.tsx
@@ -45,7 +45,7 @@ test.each([
     description: 'ballot package authentication errs',
     errorString: 'ballot_package_authentication_error',
     expectedErrorMessage:
-      'Error authenticating ballot package. Try re-exporting from VxAdmin.',
+      'Error authenticating ballot package. Try exporting it from VxAdmin again.',
   },
   {
     description:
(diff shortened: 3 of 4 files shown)


apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -148,10 +148,9 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element | null {
           <Prose>
             <h1>Error</h1>
             <p>
-              There was an error reading the content of the file{' '}
+              There was an error reading the contents of{' '}
               <strong>{currentState.filename}</strong>:{' '}
-              {currentState.errorMessage}. Please ensure this file only contains
-              valid CVR data for this election.
+              {currentState.errorMessage}
```
arsalansufi, line 153: I noticed some wonkiness with the current CVR import error messaging, e.g. we mention reading a file but we now read a directory. We also had some double periods `..` because nearly all of the server-returned user-friendly error messages include a period already. I also decided to remove the hardcoded "Please ensure this file only contains valid CVR data for this election" help message because 1) we export a directory not a file, and 2) all the server-returned error messages include more specific guidance that this second generic message distracts from.

Note that I haven't yet replaced all instances of the variable name `filename` with, say, `directoryName` because the long-term state of CVR exports/imports is still in a bit of flux.

benadida (commented):
I have a concern, that maybe doesn't need to be addressed right away, but writing it down here for discussion.

I think, if I'm reading the code correctly, that the signature process involves reading the CVR files from the USB drive, where the files were just written. There's an assumption here that the USB data that's about to be signed is the true data.

Imagine an evil USB that silently ignores writes except for the signature file. So we write the CVRs, but those writes are silently ignored. The , we read whatever the USB is telling us we just wrote (but the USB is lying), and we generate a signature on the wrong content.

Ideally, we'd be generating the signature not on what is on the USB, bit rather based on what is in our database/on our internal drive.

apps/admin/backend/src/cvr_files.ts
```diff
@@ -344,7 +348,9 @@ export function getAddCastVoteRecordReportErrorMessage(
   const errorType = error.type;
   switch (errorType) {
     case 'report-access-failure':
-      return 'Failed to access cast vote record report for import.';
+      return 'Unable to access cast vote record report for import.';
+    case 'cast-vote-records-authentication-error':
+      return 'Unable to authenticate cast vote records. Try re-exporting from the scanner.';
```
jonahkagan, line 353: Suggestion for slightly plainer language in the copy:

"Try exporting them from the scanner again."

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
jonahkagan: Nice catch. I think your cleanup makes sense.

This is one example of why I think it makes more sense for the backend to return error identifiers rather than full user-facing error copy. (I see you've been using that approach in this PR.)

jonahkagan (approved):
Aside from @benadida's comment, this looks good.

apps/admin/backend/src/cvr_files.ts
arsalansufi: Good call - adjusted copy in a few places accordingly: https://github.com/votingworks/vxsuite/pull/3534/commits/6e8e5f57aa5879c0dc0dcee962966fe1262c14ca

arsalansufi:
@benadida Thanks for flagging! I think it makes a lot of sense to generate files on the machine, sign them there, and then copy to the USB.

This issue affects not only CVRs but also ballot packages. Because of that, I'm going to open up a follow-up PR to update both of those code paths to write files to the machine's temp directory first and then copy them.

benadida:
> @benadida Thanks for flagging! I think it makes a lot of sense to generate files on the machine, sign them there, and then copy to the USB.
> 
> This issue affects not only CVRs but also ballot packages. Because of that, I'm going to open up a follow-up PR to update both of those code paths to write files to the machine's temp directory first and then copy them.
> 
> Main downside I can think of is a slightly slower export process. Currently inconsequential for ballot packages. And could become less consequential for CVR files if we do switch to continuous CVR exporting.

Another possibility here is to hash the data in memory as you're writing it to disk and then sign the hash and write that to disk. But if it's easier to write to local disk, then sign, then copy, that's fine too.

arsalansufi:
^ Nice, I'll timebox an exploration of alternatives like the above!

PR #3535, opened 2023-06-01
VxAdmin Tally Foundations 2.1 - Manual Data Types Refactor

### LINKED ISSUE #3473: Manual Tallies: Type Cleanup
We want to streamline the way we store manual tallies for more type safety and easier data manipulation. I believe this will be easier after we move all tallying to the backend, so punting this to the end of the project. Result of #3466 .

adghayes:
_Review commit-by-commit_

## Overview

Closes #3473. Creates a new `ManualElectionResults` type to replace the `ManualTally`. It's more compact for serialization but mostly the change is so that, on the backend, the `ManualElectionResults` and the `ElectionResults` are roughly the same data structure for easier reasoning and combining / manipulating.

## Checklist


libs/basics/src/map_object.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Takes a dictionary and returns a dictionary with the same keys but values
+ * remapped by a provided function.
+ */
+export function mapObject<T, U>(
```
kofi-q, line 5: Could use `_.mapValues()` instead: https://lodash.com/docs/4.17.15#mapValues

(That is, unless we have some guiding principles on the team around not using lodash - I did notice we don't have too many usage instances of it in the codebase)

kofi-q (approved):
🚀

libs/basics/src/map_object.ts
adghayes: It's ambiguous, but I think we have an unspecified preference for not using `lodash` or other easily replicable dependencies 🤷

PR #3536, opened 2023-06-01
refactor(mark): extract vote marking flow UI

eventualbuddha:
## Overview

These components will be reused in the RAVE voter terminal.

## Demo Video or Screenshot
n/a

## Testing Plan
- [x] New automated tests for `libs/mark-flow-ui` (95% coverage threshold)
- [x] Manual testing of VxMark voter flow

## Checklist


libs/mark-flow-ui/src/lib/screen_orientation.ts
```diff
@@ -0,0 +1,15 @@
...
+
+interface ScreenOrientationReturnType {
+  isLandscape: boolean;
+  isPortrait: boolean;
+}
+
+export function screenOrientation(
+  machineConfig: MachineConfig
+): ScreenOrientationReturnType {
+  const isLandscape = machineConfig.screenOrientation === 'landscape';
```
kofi-q, line 11: This makes me wonder: will we support a landscape mode for RAVE as well?
I might be doing some cleanup in VxMark later to remove landscape support, so want to make sure I don't end up removing anything that will be useful later

libs/mark-flow-ui/src/pages/cast_ballot_page.tsx
```diff
@@ -0,0 +1,85 @@
...
+  position: absolute;
+  right: 1rem;
+  bottom: 1rem;
+`;
+
+interface Props {
+  hidePostVotingInstructions: () => void;
+}
+
+export function CastBallotPage({
```
kofi-q, line 49: Was this meant to be a file move? (Didn't see a corresponding delete) Or will the original copy continue to be used in VxMark?

libs/mark-flow-ui/src/themes/make_theme.ts
```diff
@@ -0,0 +1,318 @@
+import {
```
kofi-q, line 1: Did you mean to fork the `themes` folder as well? Can update the reference the existing ones in libs/ui otherwise

kofi-q (approved):
LGTM - couple minor things

Excited for storybooked mark-ui components - wishing I had that during the design refresh 😅

libs/mark-flow-ui/src/lib/screen_orientation.ts
eventualbuddha: I think the idea is we'll _only_ have landscape. Tagging @mattroe for confirmation.

libs/mark-flow-ui/src/themes/make_theme.ts
eventualbuddha: I initially created the `mark-flow-ui` package by copying everything from `libs/ui`, then deleting everything that wasn't needed. `makeTheme` doesn't appear to be exported from `libs/ui`. Should I just export it, or is there a better way to achieve this?

eventualbuddha: Nevermind, I see `renderWithTheme` is already exported.

libs/mark-flow-ui/src/pages/cast_ballot_page.tsx
eventualbuddha: This one should not have been moved or copied as the corresponding RAVE one will be different.

libs/mark-flow-ui/src/lib/screen_orientation.ts
kofi-q: Ah, good to know! I guess when the time comes, I can just focus on the pieces specific to the machine UI that are no longer needed. Can leave all orientation logic within libs/mark-flow-ui intact

PR #3541, opened 2023-06-02
VxAdmin Tally Foundations Pt. 2.2 - Integrating Manual Results

adghayes:
_review commit-by-commit_

## Overview

Adds manual result tabulation to the tally backend.

## Testing Plan
I think I stumbled on a good testing approach for the manual tally tabulation which I'll apply to the other tabulation methods in #3543. Testing coverage will be the focus of that next PR.


apps/admin/backend/src/store.ts
```diff
@@ -1682,28 +1683,30 @@ export class Store {
 
   getManualResults({
     electionId,
-    precinctId,
-    ballotStyleId,
-    ballotType,
+    precinctIds,
+    ballotStyleIds,
+    votingMethods,
```
adghayes, line 1659: Change: Enable filtering on lists of ids rather than only on specific ids.

apps/admin/backend/src/tabulation/manual_results.ts
```diff
@@ -0,0 +1,130 @@
...
+ */
+export function queryManualResults({
+  store,
+  filter = {},
+  groupBy = {},
+}: {
+  store: Store;
+  filter?: Tabulation.Filter;
+  groupBy?: Tabulation.GroupBy;
+}): QueryManualResultsResult {
```
adghayes: In a later commit, I rename this `tabulateManualResults` for consistency and I make the return type a `Result` rather than this custom type. The rebase was tricky so just left it as-is 🤷

kofi-q (approved):

PR #3546, opened 2023-06-05
[VxScan] Hide election info bar in large size modes

kofi-q:
## Overview

Since we're moving from the 15" ELO screen to the smaller 13" (and switching to a landscape orientation), the election info bar no longer really works at larger size modes.

Briefly discussed with @mattroe the possibility of hiding the bar on some screens like we used to, but we fortunately don't need to at the default size. Makes sense to hide it when a voter explicitly decides to bump up the text size.

## Testing Plan
- Manual

## Checklist

### DIFF
--- a/apps/scan/frontend/src/components/layout.tsx
+++ b/apps/scan/frontend/src/components/layout.tsx
@@ -16,7 +16,10 @@ import { ScreenHeader } from './screen_header';
  * At larger text sizes, the election info bar takes up too much valuable screen
  * space, so we're hiding it in cases where a voter increases text size.
  */
-const ELECTION_BAR_HIDDEN_SIZE_MODES: ReadonlySet<SizeMode> = new Set(['l', 'xl']);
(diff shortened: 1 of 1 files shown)


apps/scan/frontend/src/components/layout.tsx
```diff
@@ -7,9 +7,17 @@ import {
...
+import { ThemeContext } from 'styled-components';
+import { SizeMode } from '@votingworks/types';
 import { getConfig, getMachineConfig, getScannerStatus } from '../api';
 import { ScreenHeader } from './screen_header';
 
+/**
+ * At larger text sizes, the election info bar takes up too much valuable screen
+ * space, so we're hiding it in cases where a voter increases text size.
+ */
+const ELECTION_BAR_HIDDEN_SIZE_MODES: Set<SizeMode> = new Set(['l', 'xl']);
```
eventualbuddha, line 19:
```suggestion
const ELECTION_BAR_HIDDEN_SIZE_MODES: ReadonlySet<SizeMode> = new Set(['l', 'xl']);
```

eventualbuddha (approved):

PR #3549, opened 2023-06-06
Fix the design app to work with complete system build script

carolinemodic:
The design app needs the script folder to be properly built by vxsuite-complete-system. It doesn't actually make sense to be built there though so I think maybe we should keep "apps" as production apps and move this somewhere else to fix the problem in a better way, but I want to unblock Ginny using VxDev.

### DIFF
--- a/apps/design/frontend/script/build-stubs
+++ b/apps/design/frontend/script/build-stubs
@@ -1,0 +1,1 @@
+../../../../script/build-stubs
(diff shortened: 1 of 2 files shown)


carolinemodic:
this is actually unnecessary after I took a more brutal hammer in https://github.com/votingworks/vxsuite-complete-system/pull/310

PR #3554, opened 2023-06-06
Sign ballot packages using data on machine rather than USB (and remove custom save location option)

arsalansufi:
_Easiest reviewed by commit_

## Overview

@benadida raised an important concern in https://github.com/votingworks/vxsuite/pull/3534#pullrequestreview-1456378640. If we write data to USB and then sign that USB data, our security narrative relies on USBs not being compromised/faulty, as a compromised/faulty USB could claim to have written the data that we asked it to but actually have written something else. So even if the artifact and its signature match, the underlying artifact might be invalid. While it's highly unlikely that this situation would arise in practice, let's harden our security narrative by signing data on machine rather than USB. This PR applies this change to ballot package signing.

In the process, this PR also 1) moves ballot package saving logic to the backend (in line with our long-term vision for VxSuite architecture) and 2) removes the ability to save ballot packages to a custom location. (2 was only useful for development, and that need is now largely covered by dev dock USB mocking.) 2 made 1 easier, which in turn made the core goal of this PR easier.

_No more custom save location option_

updated

## Testing

- [x] Updated automated tests
- [x] Tested ballot package export and import manually

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced ➡️ Maintained existing logs as I moved logic from the frontend to the backend


apps/admin/backend/src/app.test.ts
```diff
@@ -266,3 +265,22 @@ test('getSystemSettings returns default system settings when no system settings
...
 });
+
+test('saveBallotPackageToUsb', async () => {
+  const { apiClient, auth, mockUsb } = buildTestEnvironment();
+  const { electionDefinition } = electionMinimalExhaustiveSampleFixtures;
+  await configureMachine(apiClient, auth, electionDefinition);
+
+  mockUsb.insertUsbDrive({});
+  const response = await apiClient.saveBallotPackageToUsb();
+  expect(response).toEqual(ok());
```
arsalansufi, line 276: I may come back and add more detailed assertions here but for now am just prioritizing moving quickly so that we can create images for testing this week

apps/admin/backend/src/app.ts
```diff
@@ -166,13 +171,81 @@ function buildApi({
       return auth.unprogramCard(constructAuthMachineState(workspace));
     },
 
-    async writeBallotPackageSignatureFile(input: {
-      ballotPackagePath: string;
-    }): Promise<void> {
-      await artifactAuthenticator.writeSignatureFile({
-        type: 'ballot_package',
-        path: input.ballotPackagePath,
+    async saveBallotPackageToUsb(): Promise<Result<void, 'no_usb_drive'>> {
```
kshen0, line 174: I'm wondering if we should give errors an explicit type here, even though there's only 1 possible error value at the moment. This feels a bit like a magic string - is this common practice for the `Result` abstraction when there's only 1 possible error value?

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.tsx
```diff
@@ -1,142 +1,64 @@
...
 
-import * as workflow from '../workflows/export_election_ballot_package_workflow';
-
 const UsbImage = styled.img`
   margin-right: auto;
   margin-left: auto;
   height: 200px;
 `;
 
+type SaveError = 'no_usb_drive';
```
kshen0, line 20: re: previous comment, we could define this on the backend and import here

apps/admin/frontend/test/helpers/api_mock.ts
```diff
@@ -326,10 +326,8 @@ export function createApiMock(
       );
     },
 
-    expectWriteBallotPackageSignatureFile(ballotPackagePath: string) {
-      apiClient.writeBallotPackageSignatureFile
-        .expectCallWith({ ballotPackagePath })
-        .resolves();
+    expectSaveBallotPackageToUsb(result: Result<void, 'no_usb_drive'> = ok()) {
```
kshen0, line 329: Maybe also worth it to give the ballot package save result a defined type for use here and in the backend?

apps/admin/backend/src/app.ts
```diff
@@ -166,13 +171,81 @@ function buildApi({
...
+        await logger.log(
+          LogEventId.SaveBallotPackageComplete,
+          'election_manager',
+          {
+            disposition: 'failure',
+            message: 'Error saving ballot package: no USB drive',
+            result: 'Ballot package not saved, error shown to user.',
+          }
+        );
+        return err('no_usb_drive');
```
eventualbuddha, line 213: In this circumstance the temporary directory will never be cleaned up. Put the cleanup in a `finally` block?

apps/admin/backend/src/app.ts
```diff
@@ -166,13 +171,81 @@ function buildApi({
...
       });
+      ballotPackageZipStream.finalize();
+
+      const tempDirectory = dirSync().name;
+      const ballotPackageFileName = generateFilenameForBallotExportPackage(
+        electionDefinition,
+        new Date()
+      );
+      await fs.writeFile(
+        join(tempDirectory, ballotPackageFileName),
```
eventualbuddha, line 198: We're computing this value multiple times. Make a variable?

apps/admin/backend/src/app.test.ts
```diff
@@ -231,8 +231,7 @@ test('setSystemSettings returns an error for malformed input', async () => {
     systemSettings: JSON.stringify(malformedInput),
   });
   assert(result.isErr());
-  const err = result.err();
-  expect(err.type).toEqual('parsing');
+  expect(result.err().type).toEqual('parsing');
```
eventualbuddha, line 234:
```suggestion
  expect(result).toEqual(err('parsing'));
```

apps/admin/backend/src/app.ts
arsalansufi: Good call! A shared type feels much clearer: https://github.com/votingworks/vxsuite/pull/3554/commits/2ec12c1f5a79001e58c9a9b923d85048614b97c1

apps/admin/frontend/test/helpers/api_mock.ts
arsalansufi: Agreed https://github.com/votingworks/vxsuite/pull/3554/commits/18f42ecab9d61ced650358f0f80bb74ff382ac26

apps/admin/backend/src/app.ts
arsalansufi: Nice catch! https://github.com/votingworks/vxsuite/pull/3554/commits/4c6ff323bb93c4e5131127670406aec797bb7e3f?w=1

apps/admin/backend/src/app.ts
arsalansufi: I'd actually done this intentionally because I thought it made the code clearer, e.g.

```ts
await fs.copyFile(
  join(tempDirectory, ballotPackageFileName),
  join(usbBallotPackageDirectory, ballotPackageFileName)
);
```

reads a bit clearer to me than

```ts
await fs.copyFile(
  tempDirectoryBallotPackageFilePath,
  usbBallotPackageFilePath
);
```

But on second thought, they're probably comparable in readability. Can update

arsalansufi: ✅ https://github.com/votingworks/vxsuite/pull/3554/commits/7a4bb362091acadcaa701f1bdcfabf8e13347455

kshen0 (approved):

apps/admin/backend/src/app.ts
eventualbuddha: Thanks. My thinking was that someone updating it might forget to `join` one of them.

PR #3557, opened 2023-06-07
Kevin/print scan app scaffold

kshen0:
## Overview
Forks VxMark to a new app. The only changes are
* renames `VxMark`, `mark-backend` etc. strings to `VxMarkScan` and `mark-scan-backend`
* Updates external references to machine name and logger sources

Recommend skimming the first commit, which is a fork of `apps/mark` with a codemod to rename.

## Demo Video or Screenshot
Nothing new to demo but I verified this builds and runs:
Screenshot 2023-06-06 at 4 56 11 PM

## Testing Plan
Ports VxMark tests

## Checklist


.circleci/config.yml
```diff
@@ -266,6 +266,80 @@ jobs:
       - store_artifacts:
           path: apps/mark/integration-testing/cypress/videos/
 
+  # @votingworks/mark-scan-backend
```
kshen0, line 269: I don't love this atm because we're re-running the same tests twice, but I feel strongly that mark-scan should be its own app. VxMarkScan will run on a different device and UX will diverge from VxMark; benefits of keeping the apps united are outweighed by how troublesome it would be to maintain 2 separate flows for 2 different devices. As mark-scan diverges running these tests will become worth the time and compute budget to run them. If there's a way to disable them until then while still meeting the requirement that `.circleci/config.yml` is up to date, happy to update.

eventualbuddha: I don't think this is a problem, but if others disagree we can disable the tests by removing (or renaming) the `test` script in `package.json`.

kofi-q (approved):
🚀 

Makes a lot of sense as a separate app, but WRT to a full fork, I worry a bit about getting into a situation where certain feature updates require modifying similar/identical code in two different places, so it might mean being aggressive about consolidating common code if and when we run into those cases.

kshen0:
Thanks for raising - definitely a real concern that changes to `VxMark.*` will require extra work to duplicate across 2 apps. Summarizing Slack conversation, @kofi-q points out that forking now will force us to be more intentional about code sharing, which should produce a cleaner result than preemptively extracting common code paths.

PR #3558, opened 2023-06-07
[libs/ui] Add theme configuration for the 'lenovo' screen type

kofi-q:
## Overview

Adding a `'lenovo'` screen type, for UI theme generation, with dimensions and pixel density matching the Lenovo machines used for VxAdmin builds. Ensures that sizes are scaled more appropriately/consistently than the `'builtIn'` screen type allows for.

## Testing Plan
- Dimensions from Lenovo tech specs, rendering tested on Thinkpad

## Checklist

### DIFF
--- a/libs/types/src/ui_theme.ts
+++ b/libs/types/src/ui_theme.ts
@@ -1,5 +1,5 @@
 /** Supported screen types for VxSuite apps. */
-export type ScreenType = 'builtIn' | 'elo13' | 'elo15';
+export type ScreenType = 'builtIn' | 'elo13' | 'elo15' | 'lenovo';
 
 /** Options for supported UI color themes. */
 export type ColorMode =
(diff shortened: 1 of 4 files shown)


libs/types/src/ui_theme.ts
```diff
@@ -1,5 +1,5 @@
 /** Supported screen types for VxSuite apps. */
-export type ScreenType = 'builtIn' | 'elo13' | 'elo15';
+export type ScreenType = 'builtIn' | 'elo13' | 'elo15' | 'lenovo';
```
eventualbuddha, line 2: Might be worth making this `lenovo-thinkpad` or `lenovo-thinkpad-15` or even something more specific, but I can live with this too.

eventualbuddha (approved):

libs/types/src/ui_theme.ts
kofi-q: Yeah, that's a good call - will go with `lenovo-thinkpad-15` for clarity

PR #3560, opened 2023-06-07
Sign CVRs using data on machine rather than USB

arsalansufi:
## Overview

The equivalent of https://github.com/votingworks/vxsuite/pull/3554 but for CVRs. Copying from that PR's description:

>  If we write data to USB and then sign that USB data, our security narrative relies on USBs not being compromised/faulty, as a compromised/faulty USB could claim to have written the data that we asked it to but actually have written something else. So even if the artifact and its signature match, the underlying artifact might be invalid. While it's highly unlikely that this situation would arise in practice, let's harden our security narrative by signing data on machine rather than USB.

This PR applies this change to CVR signing.

## Testing

- [x] Updated automated tests

## Checklist

### DIFF
--- a/libs/backend/src/scan/cast_vote_records/export.ts
+++ b/libs/backend/src/scan/cast_vote_records/export.ts
@@ -21,6 +21,8 @@ import {
 import { basename, join, parse } from 'path';
 import fs from 'fs';
 import { ArtifactAuthenticatorApi } from '@votingworks/auth';
+import { dirSync } from 'tmp';
+import { rm } from 'fs/promises';
 import {
   describeSheetValidationError,
   canonicalizeSheet,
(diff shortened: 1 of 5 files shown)


libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -349,7 +357,8 @@ export async function exportCastVoteRecordReportToUsbDrive(
     const exportReportResult = await exporter.exportDataToUsbDrive(
       reportDirectory,
       CAST_VOTE_RECORD_REPORT_FILENAME,
-      castVoteRecordReportStream
+      castVoteRecordReportStream,
+      { machineDirectoryToWriteToFirst }
```
arsalansufi, line 361: Definitely a bit clunky, but I think it's clear, and it also makes for the most minimal changes elsewhere in CVR export logic, which is very likely going to change after SLI anyway, to better accommodate continuous backup.

Other alternatives included:

- Explicitly calling both `exporter.exportData` and `exporter.exportDataToUsbDrive` everywhere that we need to write to both places. This would require lots of repeated code.
- Writing all files to disk and then copying the whole directory to USB. This would require either using the experimental `fs.cp` or writing our own code to recursively copy a directory.

kshen0 (approved):

PR #3586, opened 2023-06-15
refactor(admin): Migrate Ballot Count Displays

### LINKED ISSUE #3455: VxAdmin: move tally summary data odds and ends to backend
Part of #3449.

After moving all the tally reporting to the backend, there will still be a lot of odds and ends in the frontend that rely on tally metadata (overall ballot counts in various places, etc.) that will need to be cleaned up.

adghayes:
## Overview

Closes #3455.

We have the "Ballot Counts Tables" on the reports screen that give a ballot count for each precinct, each party (if primary election), each voting method, each scanner and each batch (if selected). This is not a design that we will certify on. It does not scale, and is inflexible. In the future, we might show ballot counts for voting method and party, maybe on the reports page and maybe elsewhere. Given that, I wanted to create the backend methods for getting those ballot counts in a way that will continue to be useful when we move to another format. Instead of giving the frontend a blob with all the different category breakdowns, each one is a different query. For the short term, this leads a less than ideal scenario where the reports page is making 5 - 7 requests for ballot counts, each one involving table scans. I think that's fine for the short term, and I'd much rather the endpoints be reusable.

## Testing Plan
- Automated
- Manual

### DIFF
--- a/apps/admin/backend/src/store.ts
+++ b/apps/admin/backend/src/store.ts
@@ -991,31 +991,31 @@ export class Store {
     election: Election;
     groupBy?: Tabulation.GroupBy;
   }): Generator<Tabulation.GroupOf<CardTally>> {
-    const cvrSelectParts: string[] = [];
+    const selectParts: string[] = [];
     const groupByParts: string[] = [];
 
     if (groupBy.groupByBallotStyle || groupBy.groupByParty) {
-      cvrSelectParts.push('cvrs.ballot_style_id as ballotStyleId');
+      selectParts.push('cvrs.ballot_style_id as ballotStyleId');
       groupByParts.push('cvrs.ballot_style_id');
     }
 
     if (groupBy.groupByBatch) {
-      cvrSelectParts.push('cvrs.batch_id as batchId');
+      selectParts.push('cvrs.batch_id as batchId');
       groupByParts.push('cvrs.batch_id');
     }
 
     if (groupBy.groupByPrecinct) {
-      cvrSelectParts.push('cvrs.precinct_id as precinctId');
+      selectParts.push('cvrs.precinct_id as precinctId');
       groupByParts.push('cvrs.precinct_id');
     }
 
     if (groupBy.groupByScanner) {
-      cvrSelectParts.push('scanner_batches.scanner_id as scannerId');
+      selectParts.push('scanner_batches.scanner_id as scannerId');
       groupByParts.push('scanner_batches.scanner_id');
     }
 
     if (groupBy.groupByVotingMethod) {
-      cvrSelectParts.push('cvrs.ballot_type as votingMethod');
+      selectParts.push('cvrs.ballot_type as votingMethod');
       groupByParts.push('cvrs.ballot_type');
     }
 
@@ -1024,7 +1024,7 @@ export class Store {
     for (const row of this.client.each(
       `
           select
-            ${cvrSelectParts.map((line) => `${line},`).join('\n')}
+            ${selectParts.map((line) => `${line},`).join('\n')}
             cvrs.sheet_number as sheetNumber,
             count(cvrs.id) as tally
           from cvrs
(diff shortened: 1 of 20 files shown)


apps/admin/frontend/src/components/ballot_counts_table.tsx
```diff
@@ -255,15 +266,14 @@ export function BallotCountsTable({
               <TD as="th">Ballot Count</TD>
               <TD as="th">Report</TD>
             </tr>
-            {Object.values(VotingMethod).map((votingMethod) => {
```
adghayes, line 288: Phasing out this `enum`.

apps/admin/frontend/src/components/ballot_counts_table.tsx
```diff
@@ -185,9 +198,6 @@ export function BallotCountsTable({
     }
     case TallyCategory.Party: {
       const partiesForPrimaries = getPartiesWithPrimaryElections(election);
-      if (partiesForPrimaries.length === 0) {
-        return null;
-      }
```
adghayes, line 211: Consider it the responsibility of the caller to not render this if it's not a primary. That's already how we use it / don't use it. Removed relevant test.

apps/admin/frontend/src/app.test.tsx
```diff
@@ -408,9 +418,21 @@ test('tabulating CVRs', async () => {
...
+  apiMock.expectGetManualResultsMetadata([]);
+  apiMock.expectGetScannerBatches([
+    // need to mock scanner batches to get link button to batch report
+    {
+      electionId: 'election-id',
+      batchId: '2',
+      label: 'Batch 2',
+      scannerId: 'scanner-id',
+    },
+  ]);
```
adghayes, line 434: Some of these tests at the app level are pretty clunky. When we transition tally reports (next piece of work) they will have no purpose and will be removed.

apps/admin/frontend/src/app.test.tsx
```diff
@@ -530,36 +566,6 @@ test('manual tally data appears in reporting', async () => {
...
-  userEvent.click(screen.getByText('Reports'));
-
-  // total ballot count combination of manual and cast vote record data
-  await waitFor(() =>
-    expect(screen.getAllByTestId('total-ballot-count')[0].textContent).toEqual(
-      '3,100'
-    )
-  );
-});
-
```
adghayes, line 562: Tested in `reports_screen.test.tsx`

apps/admin/frontend/test/helpers/api_expect_helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+): Tabulation.GroupBy {
+  return {
+    groupByBatch: false,
+    groupByParty: false,
+    groupByPrecinct: false,
+    groupByScanner: false,
+    groupByVotingMethod: false,
+    ...groupBy,
+  };
+}
```
adghayes, line 20: This is necessary because `grout`'s `deepEqual` naturally does not consider `false` and `undefined` as the same.

apps/admin/backend/src/store.ts
```diff
@@ -966,14 +973,98 @@ export class Store {
...
+      groupByParts.push('cvrs.batch_id');
+    }
+
+    if (groupBy.groupByPrecinct) {
+      cvrSelectParts.push('cvrs.precinct_id as precinctId');
+      groupByParts.push('cvrs.precinct_id');
+    }
+
+    if (groupBy.groupByScanner) {
+      cvrSelectParts.push('scanner_batches.scanner_id as scannerId');
```
kofi-q, line 1013: nit: should this just be `selectParts` if it includes fields from both `cvrs` and `scanner_batches`?

apps/admin/backend/src/tabulation/card_counts.ts
```diff
@@ -0,0 +1,148 @@
...
+  isGroupByEmpty,
+} from '@votingworks/utils';
+import { CardTally } from '../types';
+import { Store } from '../store';
+
+import { tabulateManualBallotCounts } from './manual_results';
+
+/**
+ * Adds a card tally to a card counts object. Mutates the card counts object
+ * in place, and assumes overwrites previous value for the same card.
```
kofi-q, line 17: nit: typo on `and assumes overwrites`?

kofi-q (approved):

apps/admin/backend/src/tabulation/card_counts.ts
adghayes: thanks! a174c0d23

apps/admin/backend/src/store.ts
adghayes: I don't have an meaningful opinion one way or another (in other places, I'm treating `scannerId` as a "CVR attribute" but in the SQL context I think it's a fair clarification), so switched it - cb4f2153a.

PR #3587, opened 2023-06-15
refactor(grout): Grout Serialization Plain Object Prototype Requirement Tweak

adghayes:
## Overview

I've come across an issue in the VxAdmin backend where, if I send records pulled directly from the store through `grout`, serialization fails. It says the records are not plain objects because their prototypes do not equal `Object.prototype`. Actually, both of the prototypes are "null prototypes", but two different null prototypes don't evaluate as equal. Why `better-sqlite3` returns objects with a different null prototype, I do not know.

### DIFF
--- a/libs/grout/src/util.ts
+++ b/libs/grout/src/util.ts
@@ -25,8 +25,7 @@ export function isPlainObject(
 ): value is Record<string, unknown> {
   return (
     isObject(value) &&
-    Object.getPrototypeOf(value) === Object.prototype &&
-    value.constructor.name === Object.name
+    Object.getPrototypeOf(Object.getPrototypeOf(value)) === null
   );
 }
 
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -679,13 +679,9 @@ function buildApi({
     },
 
     getManualResultsMetadata(): ManualResultsMetadataRecord[] {
-      return store
-        .getManualResultsMetadata({
-          electionId: loadCurrentElectionIdOrThrow(workspace),
-        })
-        .map((record) => ({ ...record }));
-      // TODO: grout is having trouble serializing records from the store
-      // without destructuring them. need to investigate
+      return store.getManualResultsMetadata({
+        electionId: loadCurrentElectionIdOrThrow(workspace),
+      });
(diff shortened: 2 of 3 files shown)


libs/grout/src/util.ts
```diff
@@ -25,8 +25,7 @@ export function isPlainObject(
 ): value is Record<string, unknown> {
   return (
     isObject(value) &&
-    Object.getPrototypeOf(value) === Object.prototype &&
-    value.constructor.name === Object.name
+    Object.getPrototypeOf(Object.getPrototypeOf(value)) === null
```
adghayes, line 28: When comparing to `Object.prototype`, we're checking if an object has the null prototype. But it turns out there can be multiple, logically identical null prototypes that are not `===`. So, check for null prototype in a different way.

I also thought to remove the `constructor.name` check because I also thing it's in theory OK if objects have unique constructors as long as their prototype checks out. It's not about where ya came from, it's about who ya are.

jonahkagan: Should we just be checking `Object.getPrototypeOf(value) === null`? Vs checking the prototype of the prototype of the value like you're doing here?

I'm ok with removing the `constructor.name` check, though I would bet there was a reason for it based on the sources I used when originally researching how to write this function. Maybe it's in the git history? That being said, if there's no comment or test for it, I can't really give a good reason not to drop it.

adghayes: I think we have to check the double prototype, because `Object.getPrototypeOf(value)` evaluates to `[Object: null prototype] {}`, not `null`. Plain objects have the "null prototype" but their prototype is not `null`.

adghayes: Here's the context: https://github.com/votingworks/vxsuite/pull/2851#discussion_r1043978582

adghayes: Re-reading that thread, it sounds like we hit the issue that @eventualbuddha brought up. I think that the above change might fix it in a different way.

jonahkagan: Ok cool, thanks for looking into that. Happy to try this and see how it goes!

jonahkagan (approved):

Issue #3581, opened 2023-06-13
Automatic polls-close CVR exports can't be imported in VxAdmin


arsalansufi:
Automatic poll-close CVR exports can't be imported by VxAdmin. Exports via the EM "Save CVRs" function, however, can be imported just fine by VxAdmin. The `.vxsig` file seems to be missing for the former. We noticed this in test mode, but I suspect that it'll also affect live mode.

![Image](https://github.com/votingworks/vxsuite/assets/12616928/ff0039bb-8ede-42b0-a10d-573699efe213)

arsalansufi:
Fix merged into main! https://github.com/votingworks/vxsuite/commit/e9cc7ce71c4041dfe76908fc7ba99128a3d59cad

PR #3589, opened 2023-06-16
Ensure CVR export signature files are flushed to USB drives

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3581

Both @kofi-q and I have encountered sporadic auth errors when importing CVRs exported by VxScan in prod for m17a QA.

In these error cases, the CVRs are missing a corresponding `.vxsig` signature file. Turns out, the signature file isn't being flushed to the USB drive before the USB drive is removed. We make use of `sync -f` throughout the codebase to ensure data is written to USBs before their removal. We were just missing a `sync -f` in this specific code path. As we continue to hash out `libs/usb-drive`, I'm sure we can find a way to make this even more foolproof.

## Testing

- [x] Tested manually in prod
- [x] Maintained 100% code coverage in `libs/auth`

## Checklist

### DIFF
--- a/libs/auth/src/java_card.ts
+++ b/libs/auth/src/java_card.ts
@@ -141,6 +141,11 @@ export const GENERIC_STORAGE_SPACE = {
   ],
 } as const;
 
+/* istanbul ignore next */
+function generateChallenge(): string {
+  return `VotingWorks/${new Date().toISOString()}/${uuid()}`;
+}
+
 /**
  * An implementation of the card API that uses a Java Card running our fork of the OpenFIPS201
  * applet (https://github.com/votingworks/openfips201) and X.509 certs. The implementation takes
@@ -151,7 +156,7 @@ export class JavaCard implements Card {
   private readonly cardReader: CardReader;
   // See TestJavaCard in test/utils.ts to understand why this is protected instead of private
   protected cardStatus: CardStatus;
-  private readonly customChallengeGenerator?: () => string;
+  private readonly generateChallenge: () => string;
   private readonly vxCertAuthorityCertPath: string;
 
   constructor(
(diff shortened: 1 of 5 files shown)


libs/auth/src/java_card.ts
```diff
@@ -161,7 +166,9 @@ export class JavaCard implements Card {
   ) {
     this.cardProgrammingConfig = input.cardProgrammingConfig;
     this.cardStatus = { status: 'no_card' };
-    this.customChallengeGenerator = input.customChallengeGenerator;
+    this.generateChallenge =
+      input.customChallengeGenerator ??
+      /* istanbul ignore next */ generateChallenge;
```
arsalansufi, line 171: Unrelated to the fix - just making code for test-specific overrides consistent

kshen0 (commented):
> // Mock writeFile since we don't want this test to actually write to /media

Why is this? From this constraint I see why we want to rely on dependency injection in the test. But then we don't actually test that `sync -f` is called.

arsalansufi:
Ignore my last (since deleted) comment - I think I misunderstood your question haha

I don't want to write to `/media` because that's where real removable devices like USBs live, and I wouldn't want a test to accidentally overwrite data on a connected USB. All the other tests in this file write to `/tmp`, which is fine to actually write to. Let me know if I can clarify any further!

arsalansufi:
I see what you're saying about not testing the use of `sync -f`, though. Maybe I could mock the means of determining whether the device is removable instead of the means of data flushing - let me try that!

kshen0 (approved):
Ah yes I had conflated the two. Thanks for making these changes to test for `sync -f` regardless.

Issue #3452, opened 2023-05-12
VxAdmin: move main tally reports (all reports but write-in) to backend


adghayes:
Part of #3449.

Subtasks:

- [ ] Generate tallies from CVRs on backend
- [ ] Generate tallies from manual data on backend
- [ ] Modify CVR tallies with write-in data
- [ ] Wire backend tallies up to frontend
- [ ] Rewrite a lot of testing

PR #3594, opened 2023-06-20
Tally Report Part 1 - Backend Methods

adghayes:
_highly recommend commit-by-commit review_

## Overview

Part of #3452. Adds methods to get data for tally report from the frontend. Specifically, that is: scanned results with write-in adjudication information, but write-ins grouped and, separately, manual results. 

The first two commits are refactors of the tabulation infrastructure that I've been building. The first adds a helper to help merge tabulation groups. The second is a bigger change - tabulation groups are now keyed with keys that can be converted into their group specifiers, so it's not required that the groups always maintain group specifiers. This was beginning to look like a footgun - e.g. you merge some tabulation groups and forget to include the `precinctId` labels, making the groups impossible to distinguish at the end of the tabulation pipeline.

### DIFF
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -35,6 +35,7 @@ import {
   BALLOT_PACKAGE_FOLDER,
   CAST_VOTE_RECORD_REPORT_FILENAME,
   generateFilenameForBallotExportPackage,
+  groupMapToGroupList,
   isIntegrationTest,
   parseCastVoteRecordReportDirectoryName,
 } from '@votingworks/utils';
@@ -692,7 +693,7 @@ function buildApi({
       groupBy: Tabulation.GroupBy;
     }): Array<Tabulation.GroupOf<Tabulation.CardCounts>> {
       const electionId = loadCurrentElectionIdOrThrow(workspace);
-      return Object.values(
+      return groupMapToGroupList(
         tabulateFullCardCounts({
           electionId,
           store,
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -55,6 +55,7 @@ import {
   SemsExportableTallies,
   ServerFullElectionManualTally,
   SetSystemSettingsResult,
+  TallyReportResults,
   WriteInAdjudicationAction,
   WriteInAdjudicationStatus,
   WriteInCandidateRecord,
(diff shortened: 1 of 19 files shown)


apps/admin/backend/src/app.ts
```diff
@@ -692,7 +693,7 @@ function buildApi({
       groupBy: Tabulation.GroupBy;
     }): Array<Tabulation.GroupOf<Tabulation.CardCounts>> {
       const electionId = loadCurrentElectionIdOrThrow(workspace);
-      return Object.values(
+      return groupMapToGroupList(
```
adghayes, line 700: Now, when converting at the boundaries of the tabulation code, you have to use `groupMapToGroupList` rather than `Object.values`. The typing should help encourage this, because `GroupList` items have attributes like `precinctId`, `partyId`, etc. while the `GroupMap` items do not, because they assume that data is encoded in the key.

libs/utils/src/tabulation.ts
```diff
@@ -310,6 +381,25 @@ export function extractGroupSpecifier(
...
+ * preferable for most consumers.
+ */
+export function groupMapToGroupList<T>(
+  groupMap: Tabulation.GroupMap<T>
+): Tabulation.GroupList<T> {
+  const list: Tabulation.GroupList<T> = [];
+  for (const [groupKey, group] of Object.entries(groupMap)) {
+    list.push({
+      ...getGroupSpecifierFromGroupKey(groupKey),
+      // eslint-disable-next-line vx/gts-spread-like-types
```
adghayes, line 396: So, this is a fair callout by the linter. It's possible to form a `GroupMap` with with values that are not objects and therefore not spreadable. Looking on the internet, I don't see a good means to define a type which includes any plain object type. Sounds simple, right? If you can find it, I appreciate it. You can't do `T extends {}` because apparently that matches any unknown value. And you can't do `T extends { [key: string]: value }` because that only matches objects with an explicit index signature.

libs/utils/src/tabulation.ts
```diff
@@ -825,7 +825,8 @@ export function mergeManualWriteInTallies(
   }
 
   return {
-    ...manualResults,
+    // eslint-disable-next-line vx/gts-spread-like-types
+    ...anyResults,
```
adghayes, line 829: This seems like an unfair callout by the linter, I'm happy to disable this. `anyResults` extends an `Object` type so it should be spreadable.

libs/utils/src/tabulation.ts
kofi-q: Hm, my type-fu isn't up to the challenge, but on the other hand, I've been feeling like the `Tabulation.Group*` types are maybe a little too generic anyway. We only expect to use them with a handful of types related to tallies right? Is there anything stopping us from doing something explicit like this?:

```ts
type TallyData = ElectionResults | CardCounts | CardTally | etc;

type GroupMap<TallyData> = Record<GroupKey, TallyData>;

type GroupOf<TallyData> = TallyData & GroupSpecifier;

// etc...
```

Not 100% sure that solves the eslint warning issue without trying it out, but maybe?

kofi-q (approved):
:shipit:

libs/utils/src/tabulation.ts
adghayes: The only thing that stops us is that not all the included types are defined at the level of `libs/types` where the `Group*` types are defined.

Issue #3570, opened 2023-06-08
Apply new styles to central system frontend


mattroe:

mattroe:
admin essentially complete, next up is updating centralscan. adjusting this to the next milestone now.

PR #3597, opened 2023-06-20
[VxAdmin] VVSG Design Updates: UnconfiguredScreen

kofi-q:
## Overview

Applying the new VVSG design updates to the `UnconfiguredScreen` in VxAdmin.

## Testing Plan
- N/A - no functional changes

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/unconfigured_screen.tsx
+++ b/apps/admin/frontend/src/screens/unconfigured_screen.tsx
@@ -8,14 +8,7 @@ import React, {
 import { useHistory } from 'react-router-dom';
 
 import { electionFamousNames2021Fixtures } from '@votingworks/fixtures';
-import {
-  Button,
-  Font,
-  Icons,
-  P,
-  Prose,
-  useMountedState,
-} from '@votingworks/ui';
+import { Button, Font, Icons, P, useMountedState } from '@votingworks/ui';
 import { assert } from '@votingworks/basics';
 
 import type { ConfigureResult } from '@votingworks/admin-backend';
@@ -304,7 +297,7 @@ export function UnconfiguredScreen(): JSX.Element {
         centerChild
         title={`Convert from ${client?.getDisplayName()} files`}
       >
-        <Prose textCenter>
+        <Font align="center">
           <P> Select the following files from a USB drive, etc.</P>
           {inputConversionFiles.map((file) =>
             file.path ? (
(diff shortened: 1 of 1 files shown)


apps/admin/frontend/src/screens/unconfigured_screen.tsx
```diff
@@ -314,12 +305,17 @@ export function UnconfiguredScreen(): JSX.Element {
         title={`Convert from ${client?.getDisplayName()} files`}
       >
         <Prose textCenter>
-          <p> Select the following files from a USB drive, etc.</p>
+          <P> Select the following files from a USB drive, etc.</P>
```
kshen0, line 308: I know this predates your change, but just want to confirm we still support configuration of these files from non-USB drive storage (files on machine storage? SD card?)

ie. configuration of downstream machines requires a USB drive with signed files but since this UI is for import of SEMS files, requiring signing wouldn't make sense.

kshen0 (approved):

apps/admin/frontend/src/screens/unconfigured_screen.tsx
```diff
@@ -359,47 +354,52 @@ export function UnconfiguredScreen(): JSX.Element {
...
         )}
-        <HorizontalRule>or</HorizontalRule>
-        <p>
+        <P>
           <Button onPress={loadDemoElection}>
             Load Demo Election Definition
           </Button>
-        </p>
+        </P>
       </Prose>
```
adghayes, line 403: @kofi-q is there a reason you're not removing `Prose` even though you marked it as deprecated?

apps/admin/frontend/src/screens/unconfigured_screen.tsx
adghayes: AFAIK, we do. All of these configuration flows import a file from a path given by the system file picker, which does not restrict paths to the USB drive. I think importing from elsewhere is relevant for development, though, right? As in locked-down production, we wouldn't want someone importing from the machine itself and we wouldn't want to enable SD cards.

apps/admin/frontend/src/screens/unconfigured_screen.tsx
kofi-q: Ah, thanks for catching that - been meaning to remove those as I go

apps/admin/frontend/src/screens/unconfigured_screen.tsx
kofi-q: Yeah, don't have too much context on this flow, but I believe it's supported via USB without requiring the SEMS files to be signed

apps/admin/frontend/src/screens/unconfigured_screen.tsx
kofi-q: ✅  https://github.com/votingworks/vxsuite/pull/3597/commits/6982fdc1f88ceba8b43bf3fe5c5fb467478804cd

Issue #3602, opened 2023-06-21
Unable to change definition on election manager card to all bubble ballot defs


mattroe:
https://votingworks.slack.com/archives/C05BNBX5C3G/p1687366250696509

PR #3603, opened 2023-06-21
No longer write full election definition to election manager cards 

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3602

This PR updates card programming code to no longer write the full election definition to election manager cards. Now that VxMark is also configured via USB, nowhere in VxSuite do we now read the full election definition from election manager cards.

This cleanup has become more urgent because the all-bubble-ballot election definitions that SLI will be using are too large to fit on a Java Card and result in card programming errors with the current code.

## Testing

- [x] Updated unit tests
- [x] Tested manually

### DIFF
--- a/libs/auth/src/mock_file_card.test.ts
+++ b/libs/auth/src/mock_file_card.test.ts
@@ -108,7 +108,6 @@ test('MockFileCard basic mocking', async () => {
         user: electionManagerUser,
       },
     },
-    data: undefined,
     pin,
   });
   expect(await card.getCardStatus()).toEqual({
--- a/libs/auth/src/mock_file_card.test.ts
+++ b/libs/auth/src/mock_file_card.test.ts
@@ -16,7 +16,7 @@ import {
   serializeMockFileContents,
 } from './mock_file_card';
 
-const { electionData, electionHash } = electionSampleDefinition;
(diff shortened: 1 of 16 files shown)


libs/auth/src/mock_file_card.test.ts
```diff
@@ -108,7 +108,7 @@ test('MockFileCard basic mocking', async () => {
         user: electionManagerUser,
       },
     },
-    data: Buffer.from(electionData, 'utf-8'),
+    data: undefined,
```
kshen0, line 111: Do we need the `data` prop at all now?

kshen0 (approved):

libs/auth/src/mock_file_card.test.ts
arsalansufi: Yeah good q! While we don't write election definitions to cards any more, we do still read/write scanner report data:

https://github.com/votingworks/vxsuite/blob/7058c9ab6e5eddaf417d05537a0c35ef249a5363/apps/mark/backend/src/app.ts#L162-L164

So the card interface still needs methods for reading and writing card data, and by extension, the mock card implementation of the card interface still needs to support that, too

PR #3596, opened 2023-06-20
Tally Reports Part 2 - UI components

adghayes:
__review commit-by-commit, read description below, and feel free to play around with component in storybook__

## Overview

Part of #3452. 

**Commit 1** - File reorganization

**Commit 2** - Update snapshots

**Commit 3** - Creates a `ContestResultsTable` component to eventually fully replace the `ContestTally` component. The component differs in that it relies on the new types, it removes an "dimming" functionality for irrelevant contests. It represents only one contest, and it parent components are responsible for listing and choosing the contests, unlike in `ContestTally`. Styles are just copied, and there was no iteration on styling.

**Commit 4** - Creates an `AdminTallyReport` component. I've always been frustrated that our tally report existed not as a UI component but as one of several return results from the `ElectionManagerTallyReport` component in VxAdmin. I'm excited to use `storybook` to make viewing and iterating on these reports easier. Component mostly copied from `ElectionManagerTallyReport`, styling mostly untouched

**Commit 5** - Add card counts component to show card counts on reports instead of voting method breakdown.

## Demo Video or Screenshot

Screen Shot 2023-06-20 at 9 44 37 AM
Screen Shot 2023-06-20 at 9 44 30 AM

PR #3604, opened 2023-06-21
[VxAdmin] Enable new VVSG theme

kofi-q:
## Overview

Enabling the new VVSG2-compatible theme in VxAdmin.

Not all components have been updated to their theme-aware versions yet, but this is close enough to the final state that it unblocks code freeze and documentation pass for the m17a release.

### Before/After:
Screenshot 2023-06-21 at 13 52 50
Screenshot 2023-06-21 at 13 52 11

## Testing Plan
- Did a click-through of the app to make sure nothing's too severely wonky (see [slack thread](https://votingworks.slack.com/archives/C05BNBX5C3G/p1687367406860259) for video/discussion).

## Checklist

### DIFF
--- a/apps/admin/frontend/src/app.tsx
+++ b/apps/admin/frontend/src/app.tsx
@@ -1,4 +1,3 @@
-import { ColorMode } from '@votingworks/types';
 import { AppBase } from '@votingworks/ui';
 import {
   getHardware,
@@ -12,6 +11,9 @@ import { LogSource, Logger } from '@votingworks/logging';
 import { AppRoot, Props as AppRootProps } from './app_root';
 import { SessionTimeLimitTracker } from './components/session_time_limit_tracker';
 
+/* Copied from old App.css */
+const PRINT_FONT_SIZE_PX = 14;
+
 export type Props = Partial<AppRootProps>;
 
 export function App({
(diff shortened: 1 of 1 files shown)


kofi-q:
cc @arsalansufi - this will be the only critical PR to cherrypick for the m17a release, but will let you know if any lat-minute fixes come in that should be included as well

adghayes (commented):
@kofi-q have you sanity checked the tally reports in addition to the app screens? Because I think this will have have a substantial impact on styling. In #3596 you can play with the two reports in story book and toggle theme, but that component is not yet what VxAdmin is pulling from. IIRC, the new theme makes it look better.

kofi-q:
> have you sanity checked the tally reports in addition to the app screens? Because I think this will have have a substantial impact on report styling. In https://github.com/votingworks/vxsuite/pull/3596 you can play with the two reports in story book and toggle theme, but that component is not yet what VxAdmin is pulling from. IIRC, the new theme makes it look better.

Ah yeah, I was actually taking a look at that earlier and started down the path of modifying the `Prose` component to keep its overrides for print media even when we switch to the new theme, but that was causing some styling issues.

With this PR, the print preview changes quite a bit, but not so much the actual printout (since we do lock the print font size for actual prints) - that inconsistency's definitely an issue that needs to be sorted out, but here's a print-to-pdf comparison of the before and after for reference:

https://github.com/votingworks/vxsuite/assets/264902/f65b45eb-1f40-4275-bdf8-567fe188ea47

Don't have too much context around the report printing flows, so not sure if there other reports I should spot-check or if maybe it's worth testing actual physical prints to be sure. Let me know what you think

adghayes:
Ah, so you're saying that of the two versions above, one will show in the preview and the other will actually print? I think both look fine. Maybe we could just flip the styling on the printed reports too? Maybe that would help you get rid of `Prose`? 

We must have run into this with VxScan, do you remember how we handled there?

kofi-q:
My bad - wasn't clear in that description. The video shows the PDF print before and after the change.

I initially thought the preview that's shown within the app looked different from what gets printed to paper/PDF, since we set different base fonts for screen vs print, but I just double-checked and it actually matches the printout - that's mainly because I haven't yet made any changes to printable components and they're still using the old `Text` components, which retain the legacy font sizes (which is also why there weren't any major changes to VxScan printouts).

And +1 to (eventually) enabling the VVSG styles on printed reports as well - on my wishlist, but there's just a bit more work to do there to iron out the kinks first. The ideal final state would be to have all code using the same typography components consistently, but with appropriate size overrides for print, regardless of what size mode we're in.

adghayes (approved):
Sweet. As long as everything looks sane, which it does, seems good to merge. Consistent, even better!

Issue #3324, opened 2023-04-20
Make sure clock updates play well with session time limits


arsalansufi:
You can currently bypass session time limits by changing the clock to a time in the past. Similarly, if you change the clock to a time far enough in the future, you'll be automatically logged out.

We could make use of the auth lib's [`updateSessionExpiry` method](https://github.com/votingworks/vxsuite/blob/12edde0d8b4bc5abfabe6ada6040027e45903275/libs/auth/src/dipped_smart_card_auth_api.ts#L24-L26) to intelligently update the session expiry time such that the remaining session time is maintained after a clock update.

Open to other solutions as well!

PR #3608, opened 2023-06-22
Prevent bypassing session time limits by changing the clock

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3324

image

Updating the clock can have funky consequences! As noted in the linked issue:

> You can currently bypass session time limits by changing the clock to a time in the past. Similarly, if you change the clock to a time far enough in the future, you'll be automatically logged out.

This PR addresses both of the above by updating the session expiry whenever we update the clock. We update the session expiry such that the time remaining on the session is maintained, e.g. if 1 hour remained before the clock update, 1 hour will remain after the clock update.

## Testing

- [x] Added/updated unit tests

Working on testing manually ahead of m17a rc2 image creation but may ultimately end up testing on real hardware through those images.

### DIFF
--- a/apps/admin/frontend/src/screens/settings_screen.tsx
+++ b/apps/admin/frontend/src/screens/settings_screen.tsx
@@ -1,4 +1,5 @@
 import React, { useContext } from 'react';
+import { assert } from '@votingworks/basics';
 import {
   CurrentDateAndTime,
   Prose,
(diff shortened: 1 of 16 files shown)


apps/admin/frontend/src/screens/settings_screen.tsx
```diff
@@ -1,25 +1,42 @@
...
+import { isSystemAdministratorAuth } from '@votingworks/utils';
 
 import { AppContext } from '../contexts/app_context';
 import { NavigationScreen } from '../components/navigation_screen';
 import { FormatUsbButton } from '../components/format_usb_modal';
+import { updateSessionExpiry } from '../api';
 
 export function SettingsScreen(): JSX.Element {
-  const { logger, usbDrive } = useContext(AppContext);
+  const { auth, logger, usbDrive } = useContext(AppContext);
```
arsalansufi, line 18: I know I should probably use the `getAuthStatus` query instead of context to get the auth status. Was just fearing having to update tests 🙈

Will change if it ends up being easy and doesn't delay m17a rc2 image creation

apps/admin/frontend/src/screens/settings_screen.tsx
```diff
@@ -1,25 +1,42 @@
...
+          <SetClockButton
+            sessionExpiresAt={auth.sessionExpiresAt}
+            updateSessionExpiry={async (sessionExpiresAt: Date) => {
+              try {
+                await updateSessionExpiryMutation.mutateAsync({
+                  sessionExpiresAt,
+                });
+              } catch {
+                // Handled by default query client error handling
+              }
```
arsalansufi, line 37: I know this pattern is clunky, but it's currently the best option we have for passing promisified react-query mutations to reusable components while also preventing unhandled promise rejections.

Touched on in [this presentation slide](https://docs.google.com/presentation/d/1qcohhkeavCVjTZAAhGbbB7sbnvDJLlNov3OTLijfcvs/edit#slide=id.g1f8414d696e_0_361) from a while back

eventualbuddha:
Broadly speaking, it seems like the frontend is still doing too much. I assume this is because we're still using `kiosk-browser` APIs to set the clock. Instead, we should be setting the clock on the backend and updating the session expiration at the same time. The frontend shouldn't even have the ability to update the session expiration, IMO.

Time constraints may mean we have to go with this for now, but I think we should come back to it to address these issues.

arsalansufi:
After testing more fully, I'm deciding not to proceed with this PR. This fix 1) introduces other bugs, like flashing modals or the inactive session timeout being triggered, all stemming from the fact that changing the clock and updating the session expiry aren't an atomic operation, and 2) pushes us farther away from a desired long-term state where clock changes happen on the backend.

Came up with https://github.com/votingworks/vxsuite/pull/3614 as an alternative for SLI and CISA testing!

PR #3614, opened 2023-06-22
Log out after clock changes

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3324

This PR updates VxSuite apps to log out after clock changes as a heavy-handed means of preventing people from bypassing the session time limit by changing the clock.

As noted in the linked issue:

> You can currently bypass session time limits by changing the clock to a time in the past. Similarly, if you change the clock to a time far enough in the future, you'll be automatically logged out.

My first attempt at a fix (https://github.com/votingworks/vxsuite/pull/3608) 1) introduced other bugs, like flashing modals or the inactive session timeout being triggered, all stemming from the fact that changing the clock and updating the session expiry weren't an atomic operation, and 2) pushed us farther away from a desired long-term state where clock changes happen on the backend.

This new fix, though heavy-handed, is easy to reason about and requires minimal code. It'll do the trick for upcoming SLI and CISA testing, and we can always revisit down the road.

# Demo Screenshot and Video

modal

https://github.com/votingworks/vxsuite/assets/12616928/230e1f9e-f494-4ad0-92e3-f02ed5ef3519

## Testing

- [x] Updated unit tests
- [x] Tested manually

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/settings_screen.tsx
+++ b/apps/admin/frontend/src/screens/settings_screen.tsx
@@ -10,16 +10,26 @@ import {
 import { AppContext } from '../contexts/app_context';
 import { NavigationScreen } from '../components/navigation_screen';
 import { FormatUsbButton } from '../components/format_usb_modal';
+import { logOut } from '../api';
 
 export function SettingsScreen(): JSX.Element {
   const { logger, usbDrive } = useContext(AppContext);
+  const logOutMutation = logOut.useMutation();
 
   return (
     <NavigationScreen title="Settings">
       <Prose maxWidth={false}>
         <h2>Current Date and Time</h2>
         <p>
-          <SetClockButton>
+          <SetClockButton
+            logOut={async () => {
+              try {
+                await logOutMutation.mutateAsync();
+              } catch {
+                // Handled by default query client error handling
+              }
+            }}
+          >
             <CurrentDateAndTime />
           </SetClockButton>
         </p>
--- a/apps/admin/frontend/src/screens/settings_screen.tsx
+++ b/apps/admin/frontend/src/screens/settings_screen.tsx
@@ -21,15 +21,7 @@ export function SettingsScreen(): JSX.Element {
       <Prose maxWidth={false}>
         <h2>Current Date and Time</h2>
         <p>
-          <SetClockButton
-            logOut={async () => {
-              try {
-                await logOutMutation.mutateAsync();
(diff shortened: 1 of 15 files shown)


apps/admin/frontend/src/screens/settings_screen.tsx
```diff
@@ -10,16 +10,26 @@ import {
...
         <h2>Current Date and Time</h2>
         <p>
-          <SetClockButton>
+          <SetClockButton
+            logOut={async () => {
+              try {
+                await logOutMutation.mutateAsync();
+              } catch {
+                // Handled by default query client error handling
+              }
```
arsalansufi, line 30: I know this pattern is clunky, but it's currently the best option we have for passing promisified react-query mutations to reusable components while also preventing unhandled promise rejections.

Touched on in [this presentation slide](https://docs.google.com/presentation/d/1qcohhkeavCVjTZAAhGbbB7sbnvDJLlNov3OTLijfcvs/edit#slide=id.g1f8414d696e_0_361) from a while back

carolinemodic (approved):

apps/admin/frontend/src/screens/settings_screen.tsx
kofi-q: FWIW, I believe we're leaning toward `.mutate()` over `.mutateAsync()` - and if we need to vary UI based on mutation status, we can get that from `logOutMutation.[status|data|error]`. Similar discussion here: https://github.com/votingworks/vxsuite/pull/3573#discussion_r1229774807

That being said, I think @jonahkagan  might have the most up-to-date guidance here.

(Also, this is not important & non-blocking for this PR - just made me remember previous discussions)

arsalansufi: Ah yes you're totally right! Copied and pasted from my previous PR (https://github.com/votingworks/vxsuite/pull/3608), which benefitted from the ability to `await` the mutation that I was passing to the `SetClockButton`

Here, I can def use plain `.mutate` since I don't need to `await` it in `SetClockButton`

arsalansufi: https://github.com/votingworks/vxsuite/pull/3614/commits/7322f0a3badd6b74880556a7cf06cb4760da48eb

kshen0 (approved):

Issue #3584, opened 2023-06-13
VxCentralScan (and likely VxScan) cannot import ballot package if multiple are present and most recent is for a different election


arsalansufi:
VxCentralScan (and likely VxScan) cannot import ballot package if multiple are present and the most recent is for a different election.

PR #3616, opened 2023-06-23
wait for USB eject before attempting configuration again

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3584

Fixes a bug where failed configuration of VxCentralScan from USB would result in a loop of attempting and failing to configure.

## Demo Video or Screenshot

**Before**
* Auth with a `primary election` election manager card
* Insert a mock USB with a `general election` ballot package

https://github.com/votingworks/vxsuite/assets/1060688/bef38369-de79-443a-ac45-00d1605470bb

**After**
1. Auth with a `general election` election manager card
2. Insert a mock USB with both `general` and `primary` ballot packages, but `primary` is newer
  a. Screenshot 2023-06-22 at 6 58 53 PM
3. Configuration fails and error message successfully displays
4. Delete `primary` ballot package, leaving only correct `general` ballot package
5. Eject USB drive
6. Remount USB drive
7. Configuration succeeds

https://github.com/votingworks/vxsuite/assets/1060688/27d8c595-fe85-4664-a460-9f57162ac019

## Testing Plan
- [x] Add automated tests for unhappy path
- [x] Also manually verified fix with 2 physical USB drives, one with correct ballot package and one without

## Checklist

### DIFF
--- a/apps/central-scan/frontend/src/app_root.tsx
+++ b/apps/central-scan/frontend/src/app_root.tsx
@@ -476,10 +476,24 @@ export function AppRoot({
     }
   }, [electionJustLoaded, usbDrive.status]);
 
+  // We check configureMutation's status and stop configuration attempts after a failed attempt.
+  // This prevents the app from looping with continued failed configurations.
+  // If the user ejects the USB drive, we reset configureMutation's status so configuration
+  // may be attempted when a new or modified USB drive is inserted.
+  useEffect(() => {
+    if (
+      (configureMutation.isError || configureMutation?.data?.err) &&
+      usbDrive.status === 'ejected'
+    ) {
+      configureMutation.reset();
+    }
+  }, [usbDrive.status, configureMutation]);
+
   useEffect(() => {
     async function configure() {
       if (
         !configureMutation.isLoading &&
+        !configureMutation?.data?.err() &&
         !electionDefinition &&
         authStatusQuery.data &&
         isElectionManagerAuth(authStatusQuery.data) &&
@@ -723,6 +737,10 @@ export function AppRoot({
           <Button small onPress={() => logOutMutation.mutate()}>
             Lock Machine
           </Button>
+          <UsbControllerButton
+            usbDriveStatus={usbDrive.status}
+            usbDriveEject={() => usbDrive.eject(userRole)}
+          />
         </MainNav>
         <Main centerChild>
           <UnconfiguredElectionScreen
--- a/apps/central-scan/frontend/src/app_root.tsx
+++ b/apps/central-scan/frontend/src/app_root.tsx
@@ -482,7 +482,7 @@ export function AppRoot({
   // may be attempted when a new or modified USB drive is inserted.
   useEffect(() => {
     if (
-      (configureMutation.isError || configureMutation?.data?.err) &&
+      (configureMutation.isError || configureMutation?.data?.err()) &&
       usbDrive.status === 'ejected'
     ) {
       configureMutation.reset();
--- a/apps/central-scan/frontend/src/app_root.tsx
+++ b/apps/central-scan/frontend/src/app_root.tsx
@@ -501,8 +501,10 @@ export function AppRoot({
       ) {
         const result = await configureMutation.mutateAsync();
         const electionDefinitionFromResult = result.ok();
-        assert(electionDefinitionFromResult !== undefined);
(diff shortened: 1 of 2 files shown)


apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -476,19 +476,35 @@ export function AppRoot({
     }
   }, [electionJustLoaded, usbDrive.status]);
 
+  // We check configureMutation's status and stop configuration attempts after a failed attempt.
+  // This prevents the app from looping with continued failed configurations.
+  // If the user ejects the USB drive, we reset configureMutation's status so configuration
+  // may be attempted when a new or modified USB drive is inserted.
+  useEffect(() => {
+    if (
+      (configureMutation.isError || configureMutation?.data?.err) &&
```
eventualbuddha, line 485: You're checking whether a function is truthy. Did you mean `configureMutation?.data?.err()`?

kshen0: Good catch, sorry about that. Updated

kshen0: [1871198](https://github.com/votingworks/vxsuite/pull/3616/commits/1871198de2506e418f2d90209f5e350af97132d4)

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -476,19 +476,35 @@ export function AppRoot({
...
+  // We check configureMutation's status and stop configuration attempts after a failed attempt.
+  // This prevents the app from looping with continued failed configurations.
+  // If the user ejects the USB drive, we reset configureMutation's status so configuration
+  // may be attempted when a new or modified USB drive is inserted.
+  useEffect(() => {
+    if (
+      (configureMutation.isError || configureMutation?.data?.err()) &&
+      usbDrive.status === 'ejected'
+    ) {
+      configureMutation.reset();
```
eventualbuddha, line 488: This getting called twice should be okay, right? In React 18 `useEffect` may be called multiple times in development.

eventualbuddha (approved):

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -476,19 +476,35 @@ export function AppRoot({
...
+    }
+  }, [usbDrive.status, configureMutation]);
+
   useEffect(() => {
     async function configure() {
       if (
         !configureMutation.isLoading &&
+        !configureMutation?.data?.err() &&
```
arsalansufi, line 496: `configureMutation?.data?.isOk()` might be more readable than a double negative

arsalansufi (approved):
LGTM!

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -476,19 +476,35 @@ export function AppRoot({
...
+  // may be attempted when a new or modified USB drive is inserted.
+  useEffect(() => {
+    if (
+      (configureMutation.isError || configureMutation?.data?.err()) &&
+      usbDrive.status === 'ejected'
+    ) {
+      configureMutation.reset();
+    }
+  }, [usbDrive.status, configureMutation]);
+
```
jonahkagan, line 491: I'm not sure if this is applicable here, but it might be worth looking at how we approach this in VxScan: https://github.com/votingworks/vxsuite/blob/b5cc9385dfb37791f9b3d0142915f6d93e65f584/apps/scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx#L26-L33.

We have a listener on the unconfigured screen to check if the USB status changes to `mounted` and only configure in that case. I'm pretty sure it has the correct behavior.

apps/central-scan/frontend/src/app_root.tsx
kshen0: I believe so, since we check for `ejected` status this shouldn't interrupt a configuration in progress

apps/central-scan/frontend/src/app_root.tsx
kshen0: Discussed on Slack: the best long term fix is to move `central-scan` over to the newer `@votingworks/usb-drive` lib and get USB drive status from the backend. This will allow us to match the pattern in `scan`.

If we want to make a shorter fix to similar effect, we can use `useExternalStateChangeListener` as a stopgap.

For now we'll merge this to the m17 branch so at least some fix is in.

PR #3620, opened 2023-06-23
Stop Silly Parsing

adghayes:
## Overview

CVR endpoint on backend was really inefficient, weren't optimizing it since it's going to be removed soon anyway.


arsalansufi (approved):
LGTM! A+ name PR name lol

eventualbuddha:
It's not obvious to me why this newer version is faster. Large memory required to hold the original and the mapped arrays that creates GC thrash? What's your hypothesis?

PR #3623, opened 2023-06-23
make vote conversion operation faster for legacy tally method

adghayes:
## Overview

Addresses extreme slowness in frontend rendering found during m17a QA.

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #3621, opened 2023-06-23
build lookup for contest option indices so we don't have log(n^2) 

adghayes:
## Overview

Addresses a time complexity issue that arises when contests have many candidates, as in the bubble ballot. I'm hoping this will speed up the export.

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #3624, opened 2023-06-23
Speed up VxAdmin CVR imports

arsalansufi:
## Overview

A tweak similar in spirit to https://github.com/votingworks/vxsuite/pull/3621 and https://github.com/votingworks/vxsuite/pull/3623, this PR speeds up VxAdmin CVR imports, specifically for ballots with large numbers of candidates (as is the case for the all-bubble ballot SLI will be using).

## Testing

On my Mac VM, I'm seeing an import of 715 filled all-bubble ballot CVRs go from 12:40 minutes to 5:40 minutes!

### DIFF
--- a/apps/admin/backend/src/cvr_files.ts
+++ b/apps/admin/backend/src/cvr_files.ts
@@ -163,12 +163,11 @@ function snapshotHasValidContestReferences(
     );
     if (!electionContest) return err('invalid-contest');
 
+    const validContestOptions = new Set(
+      getValidContestOptions(electionContest)
+    );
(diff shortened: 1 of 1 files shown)


adghayes:
@arsalansufi do you think you could get significantly more speedup here by pushing the contest option generation further up the chain so it's not called for every contest on every sheet?

adghayes:
I'm not sure if you could, but maybe worth a try.

arsalansufi:
Yeah good q @adghayes! I considered that, but contest operation generation is so fast (less than 1 millisecond) so moving it would have little effect as far as I can tell

adghayes (approved):

PR #3629, opened 2023-06-26
Fix overflowing tally reports

jonahkagan:
## Overview

Tally reports with longer content were overflowing the letter size page when printed. My guess is that the printer interpreted it as a legal size page instead of letter (I'm not sure exactly why). I couldn't find anywhere in the stack that the page size was specified, so it must be using some default logic.

I noticed that the CSS for BMD ballots specified a letter page size, so I tried adding that to the tally report CSS and that fixed the problem!

## Demo Video or Screenshot
Before (left) and after (right)
IMG_7712

## Testing Plan
Manually tested with Polls Opened and Polls Closed reports in VxScan

## Checklist

### DIFF
--- a/libs/ui/src/reports/tally_report.tsx
+++ b/libs/ui/src/reports/tally_report.tsx
@@ -20,6 +20,9 @@ export const TallyReport = styled.div`
   @media print {
     font-size: 12px;
   }
+  @page {
+    size: letter portrait;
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
Amazing! Glad this ended up being a simple fix

PR #3631, opened 2023-06-26
[VxAdmin] VVSG Design Updates: ImportCvrFilesModal

kofi-q:
## Overview

Updating the `ImportCvrFilesModal` in VxAdmin to use the new theme-aware typography components and removing instances of the now-deprecated `Prose` component.

Also noticed the modal contents aren't scrollable with a lot of CVR files on the USB drive, so locking the max height of the shared `Modal` component and adding scroll controls to the CVR files table, so that the modal actions are always visible.

Out of scope for now:
- Table styling (borders, spacing, contents)

### Before/After:
Screenshot 2023-06-26 at 14 16 06

https://github.com/votingworks/vxsuite/assets/264902/97c18607-5e46-4a76-adc4-4f9fa12d6dbf

## Testing Plan
- Updated changed snapshots

## Checklist

### DIFF
--- a/apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
+++ b/apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
@@ -8,10 +8,12 @@ import {
   ModalWidth,
   Table,
   TD,
-  Prose,
   Button,
   ElectronFile,
   useExternalStateChangeListener,
+  WithScrollButtons,
+  P,
+  Font,
 } from '@votingworks/ui';
 import {
   isElectionManagerAuth,
@@ -52,8 +54,11 @@ const LabelText = styled.span`
   font-weight: 500;
 `;
 
-const TestMode = styled.span`
-  color: #ff8c00;
+const Content = styled.div`
+  display: flex;
+  flex-direction: column;
+  max-height: 100%;
+  overflow: hidden;
 `;
 
 type ModalState =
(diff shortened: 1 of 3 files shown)


apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -285,7 +282,11 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element | null {
           {!fileModeLocked && (
             <td>
               <LabelText>
-                {isTestModeResults ? <TestMode>Test</TestMode> : 'Official'}
+                {isTestModeResults ? (
+                  <Font color="warning">Test</Font>
```
jonahkagan, line 286: In the preview video, this text looked black. Is it meant to be orange?

jonahkagan (approved):

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
kofi-q: The "warning" color falls back to the default text color in medium and high-contrast modes, since we were unable to find any appropriate amber-adjacent colors that meet the >= 10:1 contrast ratio for medium contrast. (We don't yet provide a way to switch to low-contrast/dark mode in VxAdmin, but if we did, this would then show up as amber)

PR #3633, opened 2023-06-26
Kevin/vxcentralscan wait for usb eject to reattempt configure

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3584

Fixes a bug where failed configuration of VxCentralScan from USB would result in a loop of attempting and failing to configure.

## Demo Video or Screenshot

**Before**
* Auth with a `primary election` election manager card
* Insert a mock USB with a `general election` ballot package

https://github.com/votingworks/vxsuite/assets/1060688/bef38369-de79-443a-ac45-00d1605470bb

**After**
1. Auth with a `general election` election manager card
2. Insert a mock USB with both `general` and `primary` ballot packages, but `primary` is newer
  a. Screenshot 2023-06-22 at 6 58 53 PM
3. Configuration fails and error message successfully displays
4. Delete `primary` ballot package, leaving only correct `general` ballot package
5. Eject USB drive
6. Remount USB drive
7. Configuration succeeds

https://github.com/votingworks/vxsuite/assets/1060688/27d8c595-fe85-4664-a460-9f57162ac019

## Testing Plan
- [x] Add automated tests for unhappy path
- [x] Also manually verified fix with 2 physical USB drives, one with correct ballot package and one without

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [x] I have added a screenshot and/or video to this PR to demo the change
- [x] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #3638, opened 2023-06-27
Revert "Kevin/vxcentralscan wait for usb eject to reattempt configure"

kshen0:
Reverts votingworks/vxsuite#3633

@mattroe reported a bug where forcing the error caused every subsequent configure attempt to fail with the same error. It's likely the mutation is not being reattempted.


adghayes:
@kshen0 I'm working on a fix in #3644 that hopefully will address both problems.

PR #3639, opened 2023-06-27
Ensure VxDesign ballots can be interpreted

jonahkagan:
## Overview

In #3599, I introduced an end-to-end prototype of VxDesign. There was a catch: the ballots it produced couldn't actually be interpreted by VxSuite.

This PR makes a few tweaks to ensure the ballots produced can be interpreted.

I also fixed an issue where the fonts we use weren't embedded into the PDF, which meant that the ballot PDF only worked if you already had those fonts on your computer.

_Review by commits_

## Testing Plan
Manually tested scanning ballot images on VxCentralScan.

## Checklist


apps/design/shared/src/layout.ts
```diff
@@ -485,14 +481,17 @@ function Contest({
 }): [Rectangle, GridPosition[]] {
   assert(contest.type === 'candidate');
 
-  const titleLines = textWrap(
-    contest.title,
-    gridWidth(CONTEST_COLUMN_WIDTH - 0.5)
-  );
+  // Temp hack until we can change the timing mark grid dimensions: expand the
+  // last column to fill the page
```
arsalansufi, line 487: Took me a second to register that "last column" here means the last overall ballot column, not the last grid column. A quick addition to the comment could help!

arsalansufi (approved):
LGTM!

apps/design/shared/src/layout.ts
jonahkagan: Thanks, will update!

PR #3640, opened 2023-06-27
[VxAdmin] VVSG Design Updates: Adjudication Screen

kofi-q:
## Overview

Updating the `WriteInsAdjudicationScreen` in VxAdmin to use the new theme-aware typography components and removing instances of the now-deprecated `Prose` and `Text` components.

Also re-organising the UI to better utilise available screen space and fit more candidate options above the fold:
- Moved contest info, ballot info, and navigation controls to a full-width horizontal header

Additional nit update to the action button labels to be phrase/sentence case instead of title case.

### Before:
https://github.com/votingworks/vxsuite/assets/264902/f01ff88a-1f68-414a-a4c1-8b05e55a9c74

### After:
https://github.com/votingworks/vxsuite/assets/264902/68445b2b-dd0d-4c74-86b5-dc4fc4a9e3a3

## Testing Plan
- Updated affected existing tests.

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
+++ b/apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
@@ -71,6 +71,8 @@ const ContestTitleContainer = styled.div`
 const ContestTitle = styled(H2)`
   display: flex;
   flex-direction: column;
+  font-size: 1.2rem;
+  font-weight: ${(p) => p.theme.sizes.fontWeight.regular};
 
   /*
    * Override heading styling.
@@ -89,7 +91,7 @@ const AdjudicationNav = styled.div`
 const BallotViews = styled.div`
   background: ${(p) => p.theme.colors.foreground};
   padding: 0 0.5rem;
-  width: 62.5vw;
+  width: 75vw;
 `;
 
 const AdjudicationControls = styled.div`
@@ -105,9 +107,8 @@ const AdjudicationForm = styled.div`
 
 const TranscribedButtons = styled.div`
   display: grid;
-  grid-auto-rows: 1fr;
   grid-gap: max(${(p) => p.theme.sizes.minTouchAreaSeparationPx}px, 0.25rem);
-  grid-template-columns: 1fr 1fr;
+  grid-template-columns: 1fr;
 
   &:not(:last-child) {
     margin-bottom: 0.5rem;
--- a/apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
+++ b/apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
@@ -1,3 +1,4 @@
+/* stylelint-disable order/properties-order */
 import React, { useEffect, useRef, useState } from 'react';
 import styled from 'styled-components';
 
@@ -13,13 +14,14 @@ import {
   Button,
   Main,
   Screen,
-  Text,
-  Prose,
-  HorizontalRule,
   Icons,
   Modal,
   P,
   Font,
+  Caption,
+  LabelledText,
+  H2,
+  H4,
 } from '@votingworks/ui';
 import { format } from '@votingworks/utils';
 import { assert, throwIllegalValue } from '@votingworks/basics';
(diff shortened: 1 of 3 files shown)


apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -692,9 +710,9 @@ export function WriteInsAdjudicationScreen({
                 variant={currentWriteInMarkedInvalid ? 'secondary' : 'regular'}
                 disabled={!isWriteInDetailViewFresh}
               >
-                Mark Write-In Invalid
+                <Icons.DangerX /> Mark write-in invalid
               </Button>
-            </Prose>
+            </div>
```
jonahkagan, line 715: I wonder if it might be confusing for these buttons to start offscreen when the candidate list is long? Maybe they should be in a fixed footer instead of inside the scrollable content

That would probably work best if they were side-by-side - maybe changing the text to "Add candidate" and "Mark invalid" would do it

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -38,66 +40,81 @@ import {
...
+
+const ContestTitle = styled(H2)`
+  display: flex;
+  flex-direction: column;
+
+  /*
+   * Override heading styling.
+   * TODO: Update shared heading components to omit margins when heading is the
+   * last/only child in its container.
+   */
```
jonahkagan, line 79: Nice idea

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -38,66 +40,81 @@ import {
...
-  flex: 3;
-  background: #455a64;
-  padding: 0 0.5rem;
+const AdjudicationScreen = styled(Screen)`
+  /* Matches the focus style applied in libs/ui/global_styles.tsx, which are
+   * disabled by default in VxAdmin and enabled on the touch-only VxSuite
+   * machines.
+   * TODO: We should probably figure out a more consistent approach to
+   * enabling/disabling focus outlines across a single app, instead of
+   * conditionally enabling on certain pages.
```
jonahkagan, line 49: I think there was an intention when this screen was designed to make it super easy to navigate by keyboard so that you could adjudicate a lot of write-ins quickly, hence the need for focus outlines. I don't know whether or not users actually use it. My hunch would be no, so we could probably just drop it.

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
jonahkagan, line 1: Good idea to reorganize to show more candidate options at once.

Taking a step back and looking at the resulting screen, it's pretty visually busy (especially with the dual header bar and two columns of buttons). I think a big challenge here is that so many elements are the same size and color, so all of the elements have the same visual weight. There's not enough hierarchy for my eye to organize around.

Maybe that's just what we have to deal with, but I'm not sure I totally understand why we're implementing the high contrast themes for these fairly data-intensive screens. Are they required?

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
kofi-q: Had a similar thought and initially started heading down the path of having them at the top of the section, but I'd already overshot my time box for this task.

I like the idea idea of a sticky footer even more - will check in with @adghayes when he's back on how common a case it is to have a long list of write-ins to see if we should prioritise a task to address

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
kofi-q: That's good feedback - not a fan of where things are at with the page either, but didn't want to rabbit-hole too much. Don't want to end up in a worse state than before either, so open to suggestions on quick wins. Maybe we just move back to the single column view for now and bring the contest title and nav buttons back in?

WRT the themes - there's no VVSG requirement for election-official-facing screens, so this part of the project was motivated more by product and code consistency. We're defaulting here to the "medium contrast" option recommended by VVSG, which is still fairly high at 10:1, but not as high as the 20:1 high contrast themes. We do have the option of a creating another theme configuration somewhere between the low-contrast and medium-contrast options if you think that would help.

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
jonahkagan: Sounds good! I've noticed we spend a lot of effort maintaining it, so I'm onboard with dropping that requirement

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
jonahkagan: I think if we went back to single column but shrunk the contest title/ballot metadata to a smaller font size it might help. I also think your change to make the candidate buttons all the same width is a huge improvement, so we could keep that.

Thanks for the context on themes. I do think having some ability to deemphasize elements could really help. E.g. in this case we could create visual contrast between the two header bars by making one of them gray or something. But I do like the idea of having code consistency. Maybe we can make a theme that extends the VVSG theme with a few additional tints/shades so we can create slightly more complex UI with clearer hierarchy for the central system machines.

kofi-q: Was just playing around with a couple directions to try to unclutter - I think the first screenshot falls along the lines of what you suggested, so I can work on tweaking that a little more:

Screenshot 2023-06-27 at 16 35 11

Screenshot 2023-06-27 at 16 44 34

jonahkagan: Nice. Nit; The padding of the ballot metadata looks slightly different than the padding of the candidate column.

kofi-q: Ah, thanks - that's for the second screenshot, right?

jonahkagan: Yes

kofi-q:
Demo with latest commit:

https://github.com/votingworks/vxsuite/assets/264902/02ddff2c-aa68-4490-a7ee-54f49ebf6c72

Let me know what you think/what else could be tweaked

jonahkagan:
Nice! Maybe expand the next button so the previous button left-aligns with the candidate buttons? And potentially do the same with the "back to all write ins" button (less sure about that idea)

jonahkagan (approved):

kofi-q:
Ah, I agree with the visual intent there - usually not a big fan of trying to keep sizes across in sync across separate elements though. Just started tweaking and it was getting a little hacky, so will punt on that for now - makes me think that maybe what we want is a right sidebar that extends through the full screen height, so that all the controls can live within that visually

PR #3642, opened 2023-06-27
VxDesign ballot interpretation tests

jonahkagan:
## Overview

Add some basic tests to make sure we maintain the interpretability of ballots created by VxDesign.

This is a landmark moment (IMHO) because it these are the first tests that use dynamically generated ballots which can never get outdated (rather than fixtures).

I thought about putting these tests in libs/ballot-interpreter with the rest of our interpretation tests, but decided to keep them here for now since I'll be doing the majority of my development in apps/design for the moment. Open to feedback on this

## Demo Video or Screenshot
A dynamically generated and marked ballot
image

## Checklist

### DIFF
--- a/apps/design/backend/package.json
+++ b/apps/design/backend/package.json
@@ -40,7 +40,7 @@
   "dependencies": {
     "@votingworks/basics": "workspace:*",
     "@votingworks/db": "workspace:*",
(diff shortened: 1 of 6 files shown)


kofi-q (approved):
Exciting times!

PR #3643, opened 2023-06-28
[VxMarkScan] add paper feed and ballot print flow

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3562

* Adds a screen before voting session start for the poll worker to insert paper
* Prints ballot to Custom paper handler instead of hp printer
* Disables backend test coverage (I thought this was disabled already, but that might have been frontend only). Since UX is still in flux I don't want to spend lots of time hitting 100% coverage only to delete and move things around later
* Removes the blocking check for HP printer connection since the printer must be disconnected while ballot box is attached

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/211b6902-1bef-449b-bfd4-a98560cdf1e2

## Testing Plan
- Updated existing tests to use `expectPrintToPdf` and expect `api.printBallot` calls
- Added test for `no_hardware` state when paper handler is not connected
- Manually tested both sides of hot reloading env var toggle

## Checklist


apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -0,0 +1,276 @@
+import makeDebug from 'debug';
```
kshen0, line 1: Adapted from prototype state machine; expect this to change significantly

libs/test-utils/src/expect_print.ts
```diff
@@ -16,6 +16,9 @@ let lastPrintedElementWithCallback: Optional<ElementWithCallback>;
 let lastPrintOptions: Optional<PrintOptions>;
 let errorOnNextPrint: Optional<Error>;
 
+let lastPdfPrintedElement: Optional<JSX.Element>;
+let lastPdfPrintOptions: Optional<PrintOptions>;
```
kshen0, line 20: Open to feedback on the best way to separate this. My thought was that printing to PDF and printing to paper are 2 distinct user actions so we should not have them share the same expectations ie. `expectPrint` should not succeed if the user prints to PDF.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -0,0 +1,266 @@
...
+    time = Date.now();
+
+    await enablePrintPromise;
+    let dotsSkipped = 0;
+    debug(`begin printing ${customChunkedBitmaps.length} chunks`);
+    let i = 0;
+    for (const customChunkedBitmap of customChunkedBitmaps) {
+      debug(`printing chunk ${i}`);
+      if (customChunkedBitmap.empty) {
+        dotsSkipped += 24;
```
eventualbuddha, line 241: Magic number.

apps/mark-scan/frontend/src/index.tsx
```diff
@@ -23,6 +23,15 @@ const screenReader = new AriaScreenReader(
     : new SpeechSynthesisTextToSpeech(memoize(getUsEnglishVoice))
 );
 
+// @ts-expect-error `hot` prop does not exist on ImportMeta, but we explicity check for it
```
eventualbuddha, line 26: You can probably add this to `ImportMeta` via a `.d.ts` file so you don't have to flag it as an error. Also, these comments don't explain _why_ this is disabled.

apps/mark-scan/frontend/src/pages/print_page.tsx
```diff
@@ -18,6 +23,17 @@ export function PrintPage(): JSX.Element {
...
+  const printBallotMutation = printBallot.useMutation();
+
+  async function printElementToCustomPaperHandler(
+    element: JSX.Element,
+    options: PrintOptions
+  ): Promise<void> {
+    debug(`Ignoring print options with keys: ${Object.keys(options)}`);
+    const pdfData = await printElementToPdf(element);
+    debug(`got pdf data length ${pdfData.byteLength}`);
+    printBallotMutation.mutate({ pdfData: Array.from(pdfData) });
```
eventualbuddha, line 35: Consider using `Buffer`.

apps/mark-scan/frontend/vite.config.ts
```diff
@@ -84,6 +84,10 @@ export default defineConfig((env) => {
       },
     ],
 
+    server: {
+      hmr: false,
```
eventualbuddha, line 88: Why? If this is here does that other code that seems to disable it in JS also need to exist?

apps/mark-scan/integration-testing/cypress/support/e2e.js
```diff
@@ -89,7 +89,18 @@ function configureWithSampleDefinitionAndSystemSettings() {
   );
 }
 
+function fakeLoadPaper() {
+  cy.intercept('POST', 'http://localhost:3000/api/getPaperHandlerState', {
```
eventualbuddha, line 93: Don't need the full URL.

```suggestion
  cy.intercept('POST', '/api/getPaperHandlerState', {
```

libs/test-utils/src/expect_print.ts
eventualbuddha: Seems reasonable. I know you're duplicating the existing code structure, but this doesn't feel like we need two separate bindings here. Could just be `lastPdfPrintCommand`.

libs/test-utils/src/expect_print.ts
```diff
@@ -37,9 +45,24 @@ export function simulateErrorOnNextPrint(error: Error = new Error()): void {
...
+    if (lastPrintedElementWithCallback) {
+      culprits.push('lastPrintedElementWithCallback');
+    }
+    if (lastPdfPrintedElement) {
+      culprits.push('lastPdfPrintedElement');
+    }
     resetExpectPrint();
-    throw new ExpectPrintError(message);
+    resetExpectPrintToPdf();
+    throw new ExpectPrintError(`${message}. Culrpit: ${culprits}`);
```
eventualbuddha, line 65:
```suggestion
    throw new ExpectPrintError(`${message}. Culprit: ${culprits.join(', ')}`);
```

apps/mark-scan/frontend/package.json
```diff
@@ -87,6 +87,7 @@
     "babel-plugin-named-asset-import": "^0.3.7",
     "babel-preset-react-app": "^10.0.0",
     "buffer": "^6.0.3",
+    "canvas": "2.9.1",
```
eventualbuddha, line 90: Used?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -0,0 +1,266 @@
...
+  );
+}
+
+function isPaperAnywhere(paperHandlerStatus: PaperHandlerStatus): boolean {
+  return (
+    isPaperInInput(paperHandlerStatus) || isPaperInScanner(paperHandlerStatus)
+  );
+}
+
+export class PaperHandlerStateMachine {
```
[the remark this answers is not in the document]
kshen0: That's a good point. We lock at the web device level in the driver to prevent simultaneous data transfer but nothing prevents callers from simultaneously invoking 2 commands that use different channels, or even 2 commands that use the same channel but happen to release the lock before completion. I'll think about how to do this.

apps/mark-scan/frontend/vite.config.ts
kshen0: I tried this first but couldn't get it working, so I added the code to disable in JS and forgot to remove this. But upon further review, this config on its own disables HMR, but the code change on its own doesn't. It's possible I forgot a server restart after adding this config, and only restarted once I added the code change. Regardless, I'm updating to keep this config and delete the lines that (unsuccessfully) disable it from code.

apps/mark-scan/frontend/src/index.tsx
kshen0: Documenting that adding the following to `env.d.ts` works

```
/// <reference types="vite/client" />
```

From https://vitejs.dev/guide/features.html#client-types. But we no longer need this block since we can disable from vite config.

apps/mark-scan/frontend/src/pages/print_page.tsx
kshen0: [f13440b](https://github.com/votingworks/vxsuite/pull/3643/commits/f13440bf2c7805693d14ff5efc1e1edd2d1364f1)

apps/mark-scan/backend/src/app.ts
```diff
@@ -176,6 +195,38 @@ function buildApi(
...
+    // prints and presents a completed ballot
+    async printBallot(input: {
+      // Uint8Array is not serializable as JSON, so we serialize to number[] and deserialize below
+      pdfData: number[];
+    }): Promise<SimpleServerStatus> {
+      if (!stateMachine) {
+        return 'no_hardware';
+      }
+
+      const deserialized: Uint8Array = new Uint8Array(input.pdfData);
```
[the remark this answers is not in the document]
kshen0: [f13440b](https://github.com/votingworks/vxsuite/pull/3643/commits/f13440bf2c7805693d14ff5efc1e1edd2d1364f1)

apps/mark-scan/backend/src/app.ts
```diff
@@ -122,8 +130,19 @@ function buildApi(
       const authStatus = await auth.getAuthStatus(
         constructAuthMachineState(workspace)
       );
-      const [usbDrive] = await usb.getUsbDrives();
-      assert(usbDrive?.mountPoint !== undefined, 'No USB drive mounted');
+      const usbDrives = await usb.getUsbDrives();
+      // mark-scan hardware boots off a USB drive so we need to find the media drive
+      const mountPoints: string[] = [];
+      const usbDrive = usbDrives.find((drive) => {
+        mountPoints.push(drive.mountPoint || '<none>');
```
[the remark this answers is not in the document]
kshen0: [d149e93](https://github.com/votingworks/vxsuite/pull/3643/commits/d149e934e72441647b2b4a170696da7c437ca98d)

apps/mark-scan/frontend/package.json
kshen0: [eb0d046](https://github.com/votingworks/vxsuite/pull/3643/commits/eb0d04666c575c3935a34166d8161980012bf778)

apps/mark-scan/frontend/vite.config.ts
kshen0: [5ec62b4](https://github.com/votingworks/vxsuite/pull/3643/commits/5ec62b4b73a6161ae54e76aa014c57e588291cf3)

apps/mark-scan/frontend/src/index.tsx
kshen0: [5ec62b4](https://github.com/votingworks/vxsuite/pull/3643/commits/5ec62b4b73a6161ae54e76aa014c57e588291cf3)

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -0,0 +1,266 @@
...
+  imageDataToBinaryBitmap,
+  Uint8,
+} from '@votingworks/custom-paper-handler';
+import { SimpleServerStatus, SimpleStatus } from './types';
+
+const serverDebug = makeDebug('custom-paper-handler:machine');
+
+function debug(message: string) {
+  serverDebug(message);
+  console.log(message);
```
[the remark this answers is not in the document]
kshen0: Cleaned up debugging https://github.com/votingworks/vxsuite/pull/3643/commits/f13440bf2c7805693d14ff5efc1e1edd2d1364f1#diff-ea4f2566cf950ca374305ad00616d5eda7f645377dcf5f5cf6167e1fd0add195R17

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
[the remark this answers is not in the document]
kshen0: Recalling our previous conversation about updating the driver channel locks, this is a good opportunity to
* factor out `libs/custom-scanner/mutex`
* update the paper handler driver to use `Mutex` for its channels
* update the mark-scan state machine to use `Mutex` for a single global lock that controls access to any driver function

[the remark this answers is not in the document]
kshen0: @eventualbuddha actually, since I'm in the process of rewriting the state machine with `xstate` it might make sense to do the above in a separate PR. Thoughts?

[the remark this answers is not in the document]
eventualbuddha: I'm okay with either as long as it doesn't get lost in the shuffle.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: I think this is the line height but I'll confirm this and update

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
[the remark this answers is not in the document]
kshen0: tracking in https://github.com/votingworks/vxsuite/issues/3703

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -0,0 +1,266 @@
...
+  /**
+   * Moves print head to UP position, does not move paper
+   */
+  async disablePrint(): Promise<void> {
+    await this.driver.disablePrint();
+  }
+
+  async scanAndSave(pathOut: string): Promise<void> {
+    await this.driver.setScanDirection('backward');
+    const grayscaleResult = await this.driver.scan();
```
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/3643/commits/1fbf0dcc525b044d864930dcb553f4f8510d4467

libs/test-utils/src/expect_print.ts
kshen0: [73ea37b](https://github.com/votingworks/vxsuite/pull/3643/commits/73ea37b288f6ebd94d0d7438d026d4b6f0d7f95d)

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: 24 is the number of dots in the vertical direction when we buffer chunks to print: https://github.com/votingworks/vxsuite/pull/3643/commits/bba9d146154c248c91476d669a65d940c99090e8

apps/mark-scan/frontend/src/index.tsx
```diff
@@ -23,15 +23,6 @@ const screenReader = new AriaScreenReader(
     : new SpeechSynthesisTextToSpeech(memoize(getUsEnglishVoice))
 );
 
-// Disable hot reloading because VxMarkScan hardware lacks resources to support it
```
eventualbuddha: Would be nice to retain this comment in the `vite.config.ts` file. Also, maybe only disable if you can tell you're actually using the VxMarkScan hardware?

libs/test-utils/src/expect_print.ts
```diff
@@ -183,6 +237,35 @@ export async function expectPrint(
   }
 }
 
+/**
+ * Asserts that a print to PDF took place. Curently only printElementToPdf is supported;
```
eventualbuddha, line 241:
```suggestion
 * Asserts that a print to PDF took place. Currently only printElementToPdf is supported;
```

eventualbuddha (approved):
With one issue that must be addressed before merging (`.buffer`), LGTM.

apps/mark-scan/backend/src/app.ts
```diff
@@ -122,8 +131,20 @@ function buildApi(
...
+      // mark-scan hardware boots off a USB drive so we need to find the media drive
+      const usbDrive = usbDrives.find((drive) =>
+        drive.mountPoint?.startsWith(dataUsbMountPrefix)
+      );
+      const mountPoints = usbDrives
+        .map((drive) => drive.mountPoint || '<none>')
+        .join(', ');
+      assert(
+        usbDrive !== undefined,
+        `No USB drive mounted to ${dataUsbMountPrefix}. Got mount points: ${JSON.stringify(
```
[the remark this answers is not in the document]
kshen0: 🤦‍♂️

apps/mark-scan/frontend/src/pages/print_page.tsx
```diff
@@ -18,6 +24,17 @@ export function PrintPage(): JSX.Element {
...
+  const printBallotMutation = printBallot.useMutation();
+
+  async function printElementToCustomPaperHandler(
+    element: JSX.Element,
+    options: PrintOptions
+  ): Promise<void> {
+    debug(`Ignoring print options with keys: ${Object.keys(options)}`);
+    const pdfData = await printElementToPdf(element);
+    debug(`got pdf data length ${pdfData.byteLength}`);
+    printBallotMutation.mutate({ pdfData: Buffer.from(pdfData.buffer) });
```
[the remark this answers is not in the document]
kshen0: Interesting, thank you for clarifying 🙏

apps/mark-scan/frontend/src/index.tsx
kshen0: Comment and environment variable to toggle hot reloading added in https://github.com/votingworks/vxsuite/pull/3643/commits/eb995df1a945bb6e1459b9b05f773ec8f4468f12

I'm not currently in front of the hardware so I wasn't able to play around with detecting the hardware. Bikeshedding on variable naming: I went back and forth on whether this should be an `ENABLE` or `DISABLE` variable, but since hot reloading is the default for all apps, it felt right to have a variable for disabling one specific app. Furthermore I don't love how single-app config exists in the root `.env`, but AFAIK we don't have project-level env files.

apps/mark-scan/frontend/test/helpers/mock_api_client.tsx
```diff
@@ -161,6 +174,18 @@ export function createApiMock() {
       mockApiClient.unconfigureMachine.expectCallWith().resolves();
     },
 
+    expectParkPaper(): void {
+      mockApiClient.parkPaper.expectCallWith().resolves('paper_parked');
+    },
+
+    expectPrintBallot(pdfData: Buffer = Buffer.from([])): void {
```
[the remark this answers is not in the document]
kshen0: Updated in https://github.com/votingworks/vxsuite/pull/3643/commits/eb995df1a945bb6e1459b9b05f773ec8f4468f12

apps/mark-scan/frontend/vite.config.ts
```diff
@@ -72,8 +72,13 @@ export default defineConfig((env) => {
...
     },
 
+    // When env var is set, disable hot reloading. We support this option because VxMarkScan
+    // hardware has limited resources and hot reloading can cause poor performance during development
     server: {
-      hmr: false
+      hmr:
+        // Values are not parsed to booleans in this file
+        rootDotenvValues['REACT_APP_VX_DISABLE_MARKSCAN_HOT_RELOAD'] === 'TRUE'
+        ? false : undefined
```
kshen0, line 81: This is clunky but felt necessary since I don't think the feature flag lib is usable this early in the build process

eventualbuddha: At this point you can just access `process.env.DISABLE_MARKSCAN_HOT_RELOAD` directly–it's just NodeJS.

libs/utils/src/environment_variable.ts
```diff
@@ -35,6 +35,8 @@ export enum BooleanEnvironmentVariableName {
   SKIP_CAST_VOTE_RECORDS_AUTHENTICATION = 'REACT_APP_VX_SKIP_CAST_VOTE_RECORDS_AUTHENTICATION',
   // Disables exporting original snapshots with CVRs
   DISABLE_CVR_ORIGINAL_SNAPSHOTS = 'REACT_APP_VX_DISABLE_CVR_ORIGINAL_SNAPSHOTS',
+  // Disables hot reloading for VxMarkScan. Helps reduce performance problems when developing on VxMarkScan's low-resource hardware.
+  DISABLE_MARKSCAN_HOT_RELOAD = 'REACT_APP_VX_DISABLE_MARKSCAN_HOT_RELOAD',
```
eventualbuddha, line 39: You can do it this way, but since we only need this flag in the vite config and not in the frontend at all this whole file can be reverted and just use `process.env. DISABLE_MARKSCAN_HOT_RELOAD === 'true'` in `vite.config.ts`.

eventualbuddha (approved):
Consider simplifying the environment variable for disabling HMR.

PR #3644, opened 2023-06-28
VxCentralScan - move configuration endpoints to `grout`

adghayes:
## Overview

Moves a handful of endpoints to `grout`, particularly those around configuration since they were logically bunched together. Moved calls down the component hierarchy where possible. Along they way found a couple bugs and inconsistencies in this part of the codebase that needs some TLC, will call out in the comments.

## Testing Plan
- new automated testing
- manual testing sanity check

## Checklist

- [ ] ~~I have added a screenshot and/or video to this PR to demo the change~~
- [ ] ~~I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates~~


apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -418,97 +321,69 @@ export function AppRoot({
...
-      }
+    if (
+      electionDefinitionQuery.isSuccess &&
+      !electionDefinitionQuery.data &&
+      authStatusQuery.data &&
+      isElectionManagerAuth(authStatusQuery.data) &&
+      usbDrive.status === 'mounted' &&
+      !configureMutation.isLoading
+    ) {
+      configureMutation.mutate();
```
adghayes: I simplify the logic here, but it still has an infinite retry sort of bug identified in m17a QA which led to #3633. I address this in the final commit after manual testing.

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -569,7 +448,7 @@ export function AppRoot({
               </React.Fragment>
             }
             unconfigureMachine={() =>
-              unconfigureServer({ ignoreBackupRequirement: true })
+              Promise.resolve(systemAdministratorUnconfigure())
```
adghayes, line 444: I manually confirmed that wrapping this method in a promise still means that errors get caught in the error boundary.

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -593,7 +472,11 @@ export function AppRoot({
   }
 
   if (electionDefinition) {
-    if (electionJustLoaded) {
+    if (
+      wasJustConfigured &&
+      configureMutation.data.isOk() &&
+      !configureSuccessScreenClosed
+    ) {
```
adghayes: See final commit for changes.

apps/central-scan/frontend/src/components/toggle_test_mode_button.tsx
```diff
@@ -47,7 +47,7 @@ export function ToggleTestModeButton({
     <React.Fragment>
       <Button
         onPress={() => setFlowState('confirmation')}
-        disabled={!canUnconfigure && !isTestMode}
+        disabled={!canUnconfigure}
```
adghayes: our `getCanUnconfigure` method on the `store` already contains all this logic, so these extra checks are redundant and may allow us to create behavior in tests with non-existent combinations of parameters

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -368,6 +354,13 @@ export function AppRoot({
     electionDefinitionQuery.data,
   ]);
 
+  // reset the configuration mutation whenever USB is removed
+  useEffect(() => {
+    if (usbDrive.status !== 'mounted' && !configureMutation.isIdle) {
+      configureMutation.reset();
+    }
+  }, [configureMutation, usbDrive.status]);
+
```
adghayes, line 363: The idea of this commit is that we need to manage the configuration mutation more closely (to get our auto-configure and then hold off until you re-insert USB approach correct) _and_ use the mutation's state as a source of truth for component state.

adghayes: see: https://votingworks.slack.com/archives/C05BNBX5C3G/p1687800639587809
see: https://votingworks.slack.com/archives/C05BNBX5C3G/p1687884921336539

adghayes: @jonahkagan I was looking at the VxScan solution - does that work because the mutation resets when the unconfigured screen is unmounted?

apps/central-scan/frontend/src/screens/admin_actions_screen.test.tsx
```diff
@@ -95,108 +90,72 @@ test('clicking "Save Backup" shows progress', async () => {
...
-  resolve();
-  await waitFor(() => !screen.getByText('Deleting election data'));
+  // we are redirected to the dashboard
+  expect(history.location.pathname).toEqual('/');
 });
 
-test('clicking "Delete Ballot Data" shows progress', async () => {
-  const zeroData = jest.fn();
-  renderScreen({ zeroData, hasBatches: true, canUnconfigure: true });
+test('clicking "Delete Ballot Data" calls progress', async () => {
```
jonahkagan, line 135: is this meant to say "calls backend" not "calls progress"?

apps/central-scan/frontend/src/app_root.tsx
jonahkagan: I'm not totally sure what you're asking. Maybe let's discuss?

My quick take (without having dug into this code), is that having a `useEffect` with `configureMutation` in the dependency array makes me nervous.

I think we want the same behavior for scan and central-scan, right? So I would guess that using the same logic from scan should work...

jonahkagan (commented):
Wow, this is awesome work! Thanks for converting all of these endpoints and for the clear PR tour.

I'd like to resolve the unconfiguring issue from the last commit before approving.

apps/central-scan/backend/src/central_scanner_app.ts
```diff
@@ -180,12 +194,92 @@ function buildApi({
...
+      ) {
+        const message =
+          'Unable to unconfigure the machine while it has data that has not been backed up.';
+        await logger.log(LogEventId.ElectionUnconfigured, userRole, {
+          disposition: 'failure',
+          message,
+          result: 'Machine remains configured and data unchanged.',
+        });
+        return err(new Error(message));
+      }
```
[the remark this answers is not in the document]
adghayes: The guards in the frontend, in most of these cases, are buttons being disabled. Which I think we should definitely keep. So I'll go the `assert` path.

apps/central-scan/frontend/src/app_root.tsx
adghayes: The only difference between the two is that in VxCentralScan we show a sort of "success screen" afterwards whereas in VxScan I think we just jump to the next component. We could maintain the success screen with callbacks but I think it exists to remind the user to eject the USB, but moving forward VxCentralScan also needs continuous backup... so I'm in favor of just removing it.

apps/central-scan/frontend/src/api.ts
```diff
@@ -57,6 +59,40 @@ export const getTestMode = {
...
+    const apiClient = useApiClient();
+    return useQuery(this.queryKey(), () =>
+      apiClient.getMarkThresholdOverrides()
+    );
+  },
+} as const;
+
+// Mutations
+
 export const setTestMode = {
```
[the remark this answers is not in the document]
adghayes: Since all the ballot data is not going through `grout` currently, it actually does not need to invalidate anything else.

apps/central-scan/backend/src/central_scanner_app.ts
```diff
@@ -180,12 +194,92 @@ function buildApi({
...
+      return store.getMarkThresholdOverrides() ?? null;
+    },
+
+    setMarkThresholdOverrides(input: {
+      markThresholdOverrides?: MarkThresholds;
+    }) {
+      importer.setMarkThresholdOverrides(input.markThresholdOverrides);
+    },
+
+    async zeroScanningData(): Promise<Result<void, Error>> {
```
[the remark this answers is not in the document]
adghayes: yeah, the historical name here has been zero (that's what it is at the `importer` and `store`) level and I was trying to maintain some connection, but I agree that `zero` is generally confusing. will move away from it here.

apps/central-scan/frontend/src/screens/admin_actions_screen.tsx
```diff
@@ -94,28 +110,10 @@ export function AdminActionsScreen({
...
-        if (isMounted) {
-          setIsFactoryResetting(false);
-        }
-      })();
-      return () => {
-        isMounted = false;
-      };
-    }
-  }, [isFactoryResetting, unconfigureServer, usbDriveEject, userRole]);
+  const markThresholdOverrides = markThresholdOverridesQuery.data ?? undefined;
```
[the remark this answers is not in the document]
adghayes: Originally I had "Loading Admin Actions..." show but I didn't like the whole UI jumpiness. In 99% of cases there will not be override data and the default will be accurate.

The `?? undefined` is because it might be `null` coming from the backend and I just like to convert at the boundaries for consistency.

apps/central-scan/backend/src/central_scanner_app.ts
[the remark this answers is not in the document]
adghayes: 38aedde4d

apps/central-scan/backend/src/central_scanner_app.ts
```diff
@@ -95,8 +96,21 @@ function buildApi({
       return store.getTestMode();
     },
 
-    setTestMode(input: { testMode: boolean }) {
-      return store.setTestMode(input.testMode);
+    async setTestMode(input: { testMode: boolean }) {
+      const userRole = await getUserRole();
+      const { testMode } = input;
```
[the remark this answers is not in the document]
adghayes: 03e610037

apps/central-scan/frontend/src/components/set_mark_thresholds_modal.tsx
```diff
@@ -90,16 +94,10 @@ export function SetMarkThresholdsModal({
...
-    try {
-      await setMarkThresholdOverrides(undefined);
-      onClose();
-    } catch (error) {
-      assert(error instanceof Error);
-      setCurrentState(ModalState.ERROR);
-      setErrorMessage(`Error setting thresholds: ${error.message}`);
-    }
+    setMarkThresholdOverridesMutation.mutate({});
+    onClose();
```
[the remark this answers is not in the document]
adghayes: 152fc78f1

apps/central-scan/backend/src/central_scanner_app.ts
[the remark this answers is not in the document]
adghayes: a004a818a

apps/central-scan/frontend/src/screens/admin_actions_screen.tsx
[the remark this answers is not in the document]
jonahkagan: Ok, thanks for explaining!

apps/central-scan/frontend/src/app_root.tsx
jonahkagan: Sounds good to me, maybe worth talking through with @mattroe

apps/central-scan/backend/test/helpers/setup_app.ts
```diff
@@ -0,0 +1,84 @@
...
+      importer,
+      app,
+      logger,
+      apiClient,
+      server,
+    });
+  } finally {
+    const { promise, resolve, reject } = deferred<void>();
+    server.close((error) => (error ? reject(error) : resolve()));
+    await promise;
```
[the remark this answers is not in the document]
adghayes: d3ebbadc1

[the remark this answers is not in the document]
adghayes: i think `promisify` didn't work here because it expects an `(error, value) =>` callback whereas `server.close` is just an `(error) =>` callback.  Didn't dig much.

Applied the change to VxScan as well.

apps/central-scan/frontend/src/screens/admin_actions_screen.test.tsx
adghayes: 4d79f44f2

apps/central-scan/frontend/src/app_root.tsx
adghayes: Sounds good. This thread is the biggest piece of feedback, I'll wait till I'm actually at my desk tomorrow to check in with folks and resolve. 

@kofi-q if you get to VxCentralScan UI work before this is merged, you may want to branch off of this PR.

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -472,11 +465,7 @@ export function AppRoot({
   }
 
   if (electionDefinition) {
-    if (
-      wasJustConfigured &&
-      configureMutation.data.isOk() &&
-      !configureSuccessScreenClosed
-    ) {
+    if (configureMutation.isSuccess && configureMutation.data.isOk()) {
```
kshen0, line 468: Is this check redundant? Should it be impossible to have an `electionDefinition` without a successful configure?

adghayes: Not redundant, because you could have configured the machine last time you powered on or before we reset the mutation. Will be reworking this to be more similar to VxScan anyway, though, so this logic will likely be removed.

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -334,33 +322,6 @@ export function AppRoot({
...
-    electionDefinitionQuery.isSuccess,
-    electionDefinitionQuery.data,
-  ]);
-
-  // reset the configuration mutation whenever USB is removed
-  useEffect(() => {
-    if (usbDrive.status !== 'mounted' && !configureMutation.isIdle) {
-      configureMutation.reset();
-    }
-  }, [configureMutation, usbDrive.status]);
```
jonahkagan: ✂️ 

Love to see this

jonahkagan (approved):

PR #3648, opened 2023-06-28
[VxAdmin] VVSG Design Updates: WriteInsScreen

kofi-q:
## Overview

Applying VVSG design updates to the `WriteInsScreen` in VxAdmin:
- Replacing typography elements with theme-aware versions
- Removing deprecated `Prose` and `Text` instances
- Replaced `<Text muted>` with `<Font weight="light">` where appropriate - the former was applying a faded gray, which we don't currently have in the default color theme, so using font weight to convey which rows require action instead.
- Changed the table header font size to the document default, since it looked noticeably small compared to the table body.

Additional cleanup to render the adjudication flow screen as a full-page screen instead of a `<Modal fullscreen>`

### Before:
Screenshot 2023-06-28 at 11 55 08

### After:
Screenshot 2023-06-28 at 11 50 46

Screenshot 2023-06-28 at 11 50 18

Screenshot 2023-06-28 at 11 51 37

## Testing Plan
- Manual - no functional changes

## Checklist

### DIFF
--- a/libs/ui/src/table.tsx
+++ b/libs/ui/src/table.tsx
@@ -27,10 +27,10 @@ export const Table = styled.table<TableProps>`
   }
   & th {
     border-top: 1px solid ${tableBorderColor};
-    font-size: 0.75rem;
   }
   @media print {
     border-top: 1px solid ${tableBorderColor};
+    font-size: 0.75rem;
   }
 `;
 
--- a/libs/ui/src/table.tsx
+++ b/libs/ui/src/table.tsx
@@ -27,10 +27,13 @@ export const Table = styled.table<TableProps>`
   }
   & th {
     border-top: 1px solid ${tableBorderColor};
+
+    @media print {
+      font-size: 0.75rem;
+    }
   }
   @media print {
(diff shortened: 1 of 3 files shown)


libs/ui/src/table.tsx
```diff
@@ -27,10 +27,10 @@ export const Table = styled.table<TableProps>`
   }
   & th {
     border-top: 1px solid ${tableBorderColor};
-    font-size: 0.75rem;
```
jonahkagan, line 30: I like this change! Have you spot checked other screen/modals with tables to make sure it looks good in other contexts as well? I know we use them in a lot of places so it might be worth a glance

jonahkagan (approved):

libs/ui/src/table.tsx
kofi-q: Yeah, good nudge! I'd spot-checked a few screens in VxAdmin, but will also take a look at one or two other apps, just to be safe.

Also just noticed I should've placed the print media override within this `th` block to properly preserve previous behaviour for prints, so I'll fix that up

PR #3658, opened 2023-06-29
style(jsx): remove unnecessary `React` from scope

eventualbuddha:
## Overview

Per the [docs on `react/react-in-jsx-scope`][1], the rule is only needed when not using the `react-jsx` runtime, which we are using. So we disable it by extending the `jsx-runtime` config as described in the note at the bottom.

[1]: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

## Checklist

### DIFF
--- a/apps/admin/frontend/babel.config.js
+++ b/apps/admin/frontend/babel.config.js
@@ -2,5 +2,5 @@
  * @type {import('@babel/core').TransformOptions}
  */
 module.exports = {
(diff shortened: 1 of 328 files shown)


jonahkagan (approved):
Nice find!

PR #3662, opened 2023-06-30
Migrate WIA Report

### LINKED ISSUE #3453: VxAdmin: move write-in scatter tally creation to backend
Part of #3449.

adghayes:
## Overview

Closes #3453. 

## Checklist


adghayes:
@kofi-q I wasn't sure how best to do the loading behavior on this new report screen in the frontend. the way I did it for the other reports ultimately doesn't look too good.

here's the page once it loads:

Screen Shot 2023-07-03 at 12 38 12 PM

the data for the report might take several seconds to load. originally, I had, in the place of the report, a "Loading..." component where the preview is. but it looks a bit messy (like a lot of the intact `Loading` components, IMO). do we want to establish a pattern of using the "Progress Ellipses"? Or do we want to move toward using the spinner icon you've added? that's what I have in this PR, the spinner. wasn't sure to put it on the left or right. left is what I have, as it seems more natural, but it means the header jumps to the side when done.

Screen Shot 2023-07-03 at 12 36 07 PM

Screen Shot 2023-07-03 at 12 36 42 PM

wondering about your opinion in this case and also whether we should have a general pattern. I actually like the "Progress Ellipsis" but that component needs some rework if we're going to keep using it, has a lot of styling baggage. also it can make modals change in width as the ellipses move, or some weirder behavior (see video):


https://github.com/votingworks/vxsuite/assets/37960853/3ffdb7a5-7009-48f6-ace6-c60f096c226f

kofi-q:
Yeah, good question about the loading content pattern - I don't believe we'll be able to have a single pattern for every situation, but it's definitely worth settling on consistent patterns for various categories (single-line item loading vs. page section loading vs whole page loading). Not something I've spent much time thinking about, but can put on the wishlist if we're running into more delayed data scenarios as our apps get more complex.

In this case, it feels like a good fit might be something along the lines of a block placeholder, like you'd see on sites where outlines of the content fade in and out - for our case though, I think it's fine to just keep the `Report Preview` heading unchanged and add the `<Icons.Spinner /> Generating report` text under the heading instead of having the heading change.

WRT specific loading component, I think the spinner icon works well for smaller single-line elements (or stuff like button labels), so in this case, up to you on which component feels more appropriate. Maybe worth wrapping the spinner icon in a `<FullScreenIconWrapper>` and seeing if it doesn't look too ridiculous at that size 😅

kofi-q:
This didn't occur to me in the slack conversation, but it might be worth using bullet points for the "Contests With No Write-Ins" section and dropping the italics so that section's a bit more legible. It might also just be the indentation and italics that are throwing me off, so another option could be to remove the indentation and italics for a start.

libs/ui/src/reports/write_in_adjudication_report.tsx
```diff
@@ -0,0 +1,122 @@
...
+          }
+          const sectionKey = partyId || 'none';
+
+          return (
+            <ReportSection
+              key={sectionKey}
+              data-testid={`write-in-tally-report-${sectionKey}`}
+            >
+              <LogoMark />
+              <Prose maxWidth={false}>
```
kofi-q, line 86: nit: This is probably just copied from the pre-existing reports I believe in the current state, `<Prose maxWidth={false}>` is equivalent to a regular div, so worth trying to drop if it doesn't affect the formatting of the page

kofi-q (approved):
LGTM

Only thing worth considering is the "no write-ins" section - IMO, fine-tuning the loading state stuff isn't a big deal to figure out in this single case, unless it's bugging you 😄

libs/ui/src/reports/write_in_adjudication_report.tsx
adghayes: it does change the formatting, but actually brings it more in line with the other report (makes header slightly more pronounced), so I think it looks better. dropped it in 52353c2e6

adghayes:
@kofi-q took out the italics and minimized the indentation, think that looks good. also added border lines between the zero contests heading and beneath headings of contests with write-ins but non adjudicated, looks better to me. 69273b321

Screen Shot 2023-07-03 at 3 40 55 PM

PR #3663, opened 2023-06-30
Prune Old Tally Code

adghayes:
## Overview

Removes the stuff that can be removed now that we're at the end of the VxAdmin tally infrastructure project. A lot of former dependencies cannot be removed until VxScan tally code is changed.


kofi-q (approved):
✂️ ✂️

PR #3667, opened 2023-07-03
[CentralScan] VVSG Design Updates: AdminActionsScreen

kofi-q:
## Overview

Apply VVSG design updates to the `AdminActionsScreen` in CentralScan:
- Replacing typography components with theme-aware versions and removing deprecated `<Prose>` 
- Moving modal titles into `title` prop
- Removing `danger` button variant styling from the non-destructive buttons that open confirmation modals.

### Before:
Screenshot 2023-07-03 at 14 43 12

### After:
https://github.com/votingworks/vxsuite/assets/264902/6de1c25e-cca0-41d6-8670-d6af736edf05

Screenshot 2023-07-03 at 15 55 27

## Testing Plan
- Visual

## Checklist

### DIFF
--- a/apps/central-scan/frontend/src/screens/admin_actions_screen.tsx
+++ b/apps/central-scan/frontend/src/screens/admin_actions_screen.tsx
@@ -182,16 +182,13 @@ export function AdminActionsScreen({
               >
                 <Icons.Delete /> Delete Election Data from VxCentralScan
               </Button>{' '}
-              {!canUnconfigure && !isTestMode && (
-                <React.Fragment>
-                  <br />
-                  <Font color="warning">
-                    <Icons.Warning /> You must &quot;Save Backup&quot; before
-                    you may delete election data.
-                  </Font>
-                </React.Fragment>
-              )}
             </ButtonRow>
+            {!canUnconfigure && !isTestMode && (
+              <Font color="warning">
+                <Icons.Warning /> You must &quot;Save Backup&quot; before you
+                may delete election data.
+              </Font>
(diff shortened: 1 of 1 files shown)


apps/central-scan/frontend/src/screens/admin_actions_screen.tsx
```diff
@@ -249,13 +254,8 @@ export function AdminActionsScreen({
...
         <Modal
-          centerContent
-          content={
-            <Prose textCenter>
-              <h1>Are you sure?</h1>
-              <p>This can not be undone.</p>
-            </Prose>
-          }
+          title="Are you sure?"
+          content={<P>This can not be undone.</P>}
```
adghayes, line 258: i know this is copied over but i think grammar police would dictate that this be changed to "This **cannot** be undone."

adghayes (approved):
aesthetic nit - i think the warning text at the bottom (about needing to back up the scanner) could use a little more spacing because the text against the dotted disabled button is a bit messy looking. but if you're pointedly trying to avoid these sorts of one-off style adjustments, that sounds good

apps/central-scan/frontend/src/screens/admin_actions_screen.tsx
kofi-q: Ah, good catch - agreed, definitely worth rolling into this change

kofi-q (commented):
I like the idea of adding that bit of separation for the disabled button warning text - will tweak a bit

PR #3668, opened 2023-07-03
refactor(mark): more generic print page callback

eventualbuddha:
## Overview
Rather than assuming that all users of this component has `updateTally` and `resetBallot` steps, just call back once the print job has started and let the parent component handle it.

I plan to use this in `rave-mark`.

## Demo Video or Screenshot
n/a

## Testing Plan
n/a

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/pages/print_page.tsx
+++ b/apps/mark-scan/frontend/src/pages/print_page.tsx
@@ -1,7 +1,8 @@
-import { useContext } from 'react';
+import { useContext, useEffect, useRef } from 'react';
 import { PrintPage as MarkFlowPrintPage } from '@votingworks/mark-flow-ui';
 import { assert } from '@votingworks/basics';
 import { BallotContext } from '../contexts/ballot_context';
+import { BALLOT_PRINTING_TIMEOUT_SECONDS } from '../config/globals';
 
 export function PrintPage(): JSX.Element {
   const {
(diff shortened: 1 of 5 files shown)


apps/mark/frontend/src/pages/print_page.tsx
```diff
@@ -18,6 +19,22 @@ export function PrintPage(): JSX.Element {
   assert(typeof ballotStyleId === 'string', 'ballotStyleId is not defined');
   assert(typeof precinctId === 'string', 'precinctId is not defined');
 
+  const printerTimer = useRef(0);
+
+  function onPrintStarted() {
```
adghayes, line 24: because it's being passed as a prop below, should this be an `onCallback` in principle?

apps/mark-scan/frontend/src/pages/print_page.tsx
```diff
@@ -18,6 +19,22 @@ export function PrintPage(): JSX.Element {
   assert(typeof ballotStyleId === 'string', 'ballotStyleId is not defined');
   assert(typeof precinctId === 'string', 'precinctId is not defined');
 
+  const printerTimer = useRef(0);
```
adghayes, line 22: same question

adghayes (approved):

apps/mark/frontend/src/pages/print_page.tsx
eventualbuddha: _(I'm interpreting your question as whether this should use `useCallback`–if you meant something else, please correct me)_

I've gone back and forth on this. I started out just using functions normally, then I used to try to use `useCallback` on everything that could potentially cause a re-render if it changed, but after reading things like [this](https://kentcdodds.com/blog/usememo-and-usecallback) I've decided to mostly avoid `useCallback` unless there's a specific reason to use it.

apps/mark-scan/frontend/src/pages/print_page.tsx
eventualbuddha: What are you expecting this should be? I don't quite understand.

adghayes: oh, this was just bad line selection, I just meant same question as before, apologies

apps/mark/frontend/src/pages/print_page.tsx
adghayes: correct interpretation of my question. had to ask it as I think using `useCallback` was a piece of feedback you gave on an early PR of mine 😉 . makes sense to not use it.

PR #3670, opened 2023-07-05
Update Live Check in preparation for CISA testing

arsalansufi:
_Easiest reviewed by commit_

## Overview

This PR makes a number of updates to Live Check as we prepare for CISA testing. It specifically:

- Updates Live Check to use new signing functions (the same ones that we use for artifact authentication)
- Updates Live Check to include the machine cert in QR codes
- Adds the Live Check button to the VxScan election manager screen
- Adds the Live Check button to VxAdmin sys admin settings screen

I'm additionally hoping to add the Live Check button to the VxScan sys admin UI and the VxAdmin election manager UI. Will add those in a follow-up PR - wanted to get this main PR up for review as soon as possible.

## Demo Videos

_VxScan poll worker screen_

https://github.com/votingworks/vxsuite/assets/12616928/de9599e6-37a0-49f2-a047-ab8c25a02cd5

_VxScan election manager screen_

https://github.com/votingworks/vxsuite/assets/12616928/344ce1b1-548b-447c-bcaf-1b87464fb955

_VxAdmin sys admin settings screen_

https://github.com/votingworks/vxsuite/assets/12616928/d300e5e1-9e4e-44b1-8940-71045c00bfcf

## Testing

- [x] Added automated tests
- [x] Verified that my phone could scan the QR code

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced ➡️ I haven't yet given time constraints but will make an issue for this


libs/ui/src/live_check_modal.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+    <LiveCheckModal
+      onClose={jest.fn()}
+      qrCodeValue="qrCodeValue"
+      signatureInputs={{
+        machineId: 'machineId',
+        date: new Date(),
+        electionHashPrefix: 'electionHashPrefix',
+      }}
+    />
+  );
```
arsalansufi, line 15: Punting on more complete test assertions for now

apps/scan/backend/src/app.ts
```diff
@@ -90,6 +91,15 @@ function buildApi(
       );
     },
 
+    generateLiveCheckQrCodeValue() {
+      const { machineId } = getMachineConfig();
+      const electionDefinition = workspace.store.getElectionDefinition();
+      return new LiveCheck().generateQrCodeValue({
```
arsalansufi, line 97: Will eventually dependency inject live check like I dependency inject the auth instance and artifact authenticator so that I can write proper app-level tests. Punted on this for now to keep this PR as minimal as possible

benadida (approved):
looks good.

only comment is whether the UI components can be shared rather than duped, but I'm fine if it has to be duped for now.

arsalansufi:
> only comment is whether the UI components can be shared rather than duped, but I'm fine if it has to be duped for now.

Yeah the tricky thing here is that the duplicated component needs to interact with react-query. To fully share it, it'd need to take a react-query query as a prop, which would make testing it in libs/ui difficult. So I opted to share just the visual logic in libs/ui, letting frontends wire the component up with react-query as needed. A solvable problem for sure! Just didn't want to tackle it right now

PR #3683, opened 2023-07-06
[VxMark] Update StartPage to fit legacy landscape screen

kofi-q:
## Overview

We will continue to support and upgrade the old 14" landscape machines out in the field, so VxMark needs to be updated so the new styling also works in landscape mode. 

- Removing the old `Sidebar` component in favour of varying the layouts based on CSS `orientation` media queries
- Removing branching logic in `ElectionInfo` as well and using media queries.

### Portrait:
https://github.com/votingworks/vxsuite/assets/264902/81520e60-b28f-4e86-a994-fdeb7136524b

### Landscape:
https://github.com/votingworks/vxsuite/assets/264902/349d85f8-1a52-49ca-ab63-1dff43e6f108

## Testing Plan
- Visual + existing tests as regression guards

## Checklist


apps/mark/frontend/src/pages/start_page.tsx
```diff
@@ -1,46 +1,47 @@
 /* stylelint-disable order/properties-order */
 import { singlePrecinctSelectionFor } from '@votingworks/utils';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
adghayes, line 3: why do you need this? can't you just remove `React` after #3658?

adghayes (approved):
as long as we're keeping landscape around, I appreciate that you're keeping it simple. LGTM

PR #3704, opened 2023-07-11
feat(scan-nh): optionally count unmarked write-ins

eventualbuddha:
## Overview

Refs WIA: ability to adjudicate name-only marks (no oval) #2458

If `markThresholds.writeInTextArea` is set, the NH interpreter will now compute a score for the dark pixels in each write-in area. If the score crosses the threshold, the fill score for that write-in will be set to 100% (1). This is an interim solution for detecting unmarked write-ins that we may decide to keep as one of multiple different configurations, e.g. alongside including such ballots in the write-in adjudication flow.

To enable this feature, simply populate `markThresholds.writeInTextArea` within an NH election definition with a value. I recommend 5-7% based on my testing.

The interpreter does not automatically score write-ins and instead makes that an option that is currently enabled when the threshold is set. I structured it this way primarily because the interpreter does not use the knowledge of the mark thresholds and I'd like to keep it that way. It also doesn't modify the score as described earlier, but returns the write-in scores separately. Setting the score to 100% happens at a higher level of the interpretation stack.

## Demo Video or Screenshot
| Front | Back |
|-|-|
| scan-marked-front_debug_scored_write_in_areas | scan-marked-back_debug_scored_write_in_areas |

## Testing Plan
- [x] Automated
- [ ] TBD manual with m17 election


libs/ballot-interpreter-nh/src/scoring.rs
```diff
@@ -238,3 +239,103 @@ pub fn score_bubble_mark(
...
+    let write_in_layout = contest_layout.options.iter().find(|option| {
+        option.grid_location.side == side
+            && option.grid_location.column == column
+            && option.grid_location.row == row
+    })?;
+    let write_in_layout_width = write_in_layout.bounds.width();
+
+    // 75% of the width of the write-in crop area and 2x the height of the
+    // bubble turned out to work pretty well with the NH state ballot layout.
+    // For now we're not trying to be more general than that.
```
jonahkagan, line 316: Seems fine for now. Do you have any ideas how we might solve this in the more general case?

libs/ballot-interpreter-nh/ts/src/cli.ts
```diff
@@ -272,10 +276,10 @@ function tryReadElectionFromElectionTable(
       db
         .prepare('SELECT election_data as electionData FROM election LIMIT 1')
         .get() as Optional<{ electionData: string }>
-    )?.electionData;
+    )?.electionData?.toString();
 
     return electionData
-      ? safeParseElectionDefinition(electionData).ok()
+      ? { electionData, election: JSON.parse(electionData), electionHash: '' }
```
jonahkagan, line 282: Why this change?

libs/ballot-interpreter/src/legacy_adapter.ts
```diff
@@ -77,9 +81,23 @@ function convertNewHampshireNextMarkToSharedMark(
...
+        ? scoredMark.fillScore
+        : // or, if there's enough text in the write-in area, score that at 100%
+        scoredWriteInArea &&
+          typeof markThresholds.writeInTextArea === 'number' &&
+          scoredWriteInArea.score >= markThresholds.writeInTextArea
+        ? 1
+        : // otherwise, just leave the mark fill score alone
+          scoredMark.fillScore;
+    if (scoredWriteInArea) {
+      console.log({ scoredMark, scoredWriteInArea, score });
```
jonahkagan, line 96: Is this meant to be left in? Maybe convert to `debug`?

jonahkagan (approved):

libs/ballot-interpreter/src/legacy_adapter.ts
eventualbuddha: Nope, will remove.

libs/ballot-interpreter-nh/ts/src/cli.ts
eventualbuddha: This was part of an unrelated change that I had mostly backed out. I forgot about this piece, but I'll revert. The reason for it was because I was trying to work with a backup that had an incompatible version of the `Election` struct.

libs/ballot-interpreter-nh/src/scoring.rs
eventualbuddha: I think we'll need to specify per-contest or maybe even per-option what the corresponding offset/area from the bubble is that we should inspect. See [this Slack thread for the most recent context](https://votingworks.slack.com/archives/CEL6D3GAD/p1689021283637969).

PR #3706, opened 2023-07-11
[VxScan] Reset voter display settings after each session

kofi-q:
## Overview

From VVSG 2.0:
> 7.1-A – Reset to default settings
> If the adjustable settings of the voter interface have been changed by the voter or election worker during the voting session, the system must automatically reset to the default setting when the voter finishes voting, verifying, and casting.

- Adds a utility component that monitors the scanner status and resets the display settings to the default when we detect the end of a scanning session (currently, when we move from `'accepted'` to `'no_paper'`).
- Additionally, resetting to the default settings whenever an election official card is inserted, since non-voter-facing screens are optimised for the default display settings.
  - When the card is removed, we restore the display settings to the previous voter state before the card was inserted.

### Limitations:
- This currently doesn't cover the potential case where a voter is unable to complete the scanning process - will try to address that in a follow-up, but open to ideas on covering that edge case if you think of anything off the top.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/264902/a7290710-733c-45e2-b5fe-220a7f4ac9b3

## Testing Plan
- Added unit tests verifying the logic in the utility component and verifying that it's rendered at the app root.

## Checklist

### DIFF
--- a/apps/scan/frontend/src/components/display_settings_manager.tsx
+++ b/apps/scan/frontend/src/components/display_settings_manager.tsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import { ThemeManagerContext, useQueryChangeListener } from '@votingworks/ui';
+import { DefaultTheme, ThemeContext } from 'styled-components';
+import { getAuthStatus, getScannerStatus } from '../api';
+
+/**
+ * Side-effect component for monitoring for auth and voter session changes and
+ * resetting/restoring voter display settings as needed.
+ */
+export function DisplaySettingsManager(): JSX.Element | null {
+  const themeManager = React.useContext(ThemeManagerContext);
+  const currentTheme = React.useContext(ThemeContext);
+
+  const authStatusQuery = getAuthStatus.useQuery();
+  const scannerStatusQuery = getScannerStatus.useQuery();
+
+  const [voterSessionTheme, setVoterSessionTheme] =
+    React.useState<DefaultTheme | null>(null);
+
+  useQueryChangeListener(authStatusQuery, (newStatus, previousStatus) => {
+    // Reset to default theme when election official logs in:
+    if (
+      previousStatus?.status === 'logged_out' &&
+      newStatus.status !== 'logged_out'
+    ) {
+      setVoterSessionTheme(currentTheme);
+      themeManager.resetThemes();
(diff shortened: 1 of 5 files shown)


apps/scan/frontend/src/components/display_settings_manager.tsx
```diff
@@ -0,0 +1,53 @@
...
+      newStatus.status === 'logged_out' &&
+      voterSessionTheme
+    ) {
+      themeManager.setColorMode(voterSessionTheme.colorMode);
+      themeManager.setSizeMode(voterSessionTheme.sizeMode);
+      setVoterSessionTheme(null);
+    }
+  });
+
+  // [VVSG 2.0 7.1-A] Reset to default theme when voter is done scanning:
```
adghayes, line 42: putting VVSG assertions in the source code for the win 💯

adghayes (approved):
The `DisplaySettingsManager` component looks great. I was concerned that something so side effect centric would be unwieldy, but with `useQueryChangeListener` it ends up being quite readable and reasonable.

UX looks good, although now the transition from the "Accepted" screen to the "Insert Your Ballot" is more jarring. It makes me want to extend the accepted timeout, but considering that this feature will rarely be used, it seems not worth touching.

On the question of unsuccessful scans, I don't know what else we could do off the top of my head except add a timeout to reset to default. If the ballot is rejected for whatever reason, the voter presumably wants the machine in the same settings they previously set to read the error or issues with the ballot. We could just reset when they remove their ballot (basically whenever `newState` transitions to `no_paper` from not `no_paper`), but there are cases where a ballot will be removed and re-inserted so I don't think that's right. If anything, I would just suggest a timeout.

kofi-q:
Yeah, good point on extending the timeout between "your ballot was counted" and "insert your ballot" - I'll add to the list of cleanup/followup tasks.

And I like the idea of a theme/session timeout if there's been no activity for a while. I didn't want to simply reset every time we return to "insert your ballot" for the reasons you pointed out, so I think that's a good compromise if we just assume that no scanner status change for `x` minutes probably means no one's around. 

Crazy idea: we add a motion/proximity sensor to the machine and use that to determine voter presence 😜

PR #3708, opened 2023-07-12
Brother Thermal Printer Application Tweaks

adghayes:
## Overview

We still need to make the Brother Thermal Printer work _at all_ in prod images, but this should add functionality in the app sufficient for **m17**. 

Commit 1 - adds page count display to print prompt
Commit 2 - frontend responds to PJ-823 in the same way as PJ-822

## Demo Video or Screenshot
Screen Shot 2023-07-12 at 9 31 35 AM

## Testing Plan
- manual testing in dev
- not checked in prod, but that's a separate issue
- no automated testing: automated testing with `pdf-dist` in the browser has some hiccups which, although surmountable, were not worth working through to get this into the product. should be revisited later if we proceed with the Brother thermal printers


apps/scan/frontend/src/utils/get_page_count.ts
```diff
@@ -0,0 +1,10 @@
+import { printElementToPdf } from '@votingworks/ui';
+import pdfjs from 'pdfjs-dist';
+import { Buffer } from 'buffer';
+import 'pdfjs-dist/build/pdf.worker.entry';
+
+export async function getPageCount(element: JSX.Element): Promise<number> {
+  const pdfBuffer = Buffer.from(await printElementToPdf(element));
+  const pdf = await pdfjs.getDocument(pdfBuffer).promise;
+  return pdf.numPages;
```
eventualbuddha, line 9: This is a pretty heavy-weight way to get the page count. Is there no way to tell from the element? Also, I'd prefer to keep pdfjs out of the frontend as I remember it complicating the build and testing a bit. Though maybe not since you didn't change those things?

eventualbuddha (approved):

mattroe (commented):
frontend LGTM

apps/scan/frontend/src/utils/get_page_count.ts
adghayes: It does complicate testing considerably, but I'm simply not testing it 😄 .

Totally agree on the heavy-weight perspective. If all sub-reports were single paged, we could tell from the element. But because a single sub-report can have multiple pages, we can't actually tell how many pages will be printed until we render to PDF. 

It would be better to get the page count in a way consistent with the `kiosk-browser` API, because right now we're essentially rendering to PDF twice. For example, return the page count after printing. But that would be a bigger refactor, touching multiple repos, that I'd rather not explore for this m17 demonstration.

If this stays in the codebase, I'd support doing this in a smoother way.

adghayes: As for "complicating the build," I'm not sure what you mean, but I haven't seen any problems.

PR #3716, opened 2023-07-13
VxDesign: Bubble position option (left/right)

jonahkagan:
## Overview

Adds a new toggle in VxDesign to lay out bubbles on the left or the right. Updates the ballot layout code accordingly. Note that moving bubbles to the right also means right-aligning all of the text, which required some new SVG rendering functionality.

## Demo Video or Screenshot

Screen Shot 2023-07-13 at 10 46 05 AM

https://github.com/votingworks/vxsuite/assets/530106/1e2e656e-b1b3-4d9e-9978-6e7326030abc

## Testing Plan
- Visual inspection of ballots
- Automated tests for interpretation

## Checklist

### DIFF
--- a/apps/design/frontend/src/ballots_screen.tsx
+++ b/apps/design/frontend/src/ballots_screen.tsx
@@ -1,12 +1,96 @@
-import { H1, Table, TH, TD, LinkButton, P } from '@votingworks/ui';
+import {
+  H1,
+  Table,
+  TH,
+  TD,
+  LinkButton,
+  P,
+  H2,
+  Icons,
+  Button,
+} from '@votingworks/ui';
 import { Route, Switch, useParams } from 'react-router-dom';
 import { find } from '@votingworks/basics';
-import { getElection } from './api';
-import { NestedTr } from './layout';
+import {
+  BallotLayout,
+  BallotPaperSize,
+  BallotTargetMarkPosition,
+  Election,
+} from '@votingworks/types';
+import { useState } from 'react';
+import { getElection, updateElection } from './api';
+import { Form, FormActionsRow, FormField, NestedTr } from './layout';
 import { ElectionNavScreen } from './nav_screen';
 import { ElectionIdParams, electionParamRoutes, routes } from './routes';
 import { hasSplits } from './geography_screen';
 import { BallotScreen } from './ballot_screen';
+import { SegmentedControl } from './segmented_control';
+
+const defaultBallotLayout: Required<BallotLayout> = {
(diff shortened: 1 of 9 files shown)


apps/design/frontend/src/ballots_screen.tsx
```diff
@@ -98,6 +182,8 @@ function BallotsListScreen(): JSX.Element | null {
           })}
         </tbody>
       </Table>
+      <H2>Ballot Layout</H2>
+      <BallotDesignForm electionId={electionId} savedElection={election} />
```
jonahkagan, line 186: Maybe this moves into a tab soon when there are more settings to toggle

apps/design/frontend/src/segmented_control.tsx
jonahkagan, line 1: Decided to make my own segmented control because I needed it to show the selected state when disabled. The existing SegmentedButton just grayed out both buttons (since they were both being passed the `disabled` prop). I didn't want to go messing with both the SegmentedButton and Button components to change this behavior.

cc @kofi-q something to think about - should SegmentedButton show which option is selected when disabled?

apps/design/shared/src/document_svg.tsx
```diff
@@ -49,8 +50,9 @@ export function SvgTextBox({
         <text
           // eslint-disable-next-line react/no-array-index-key
           key={textLine + index}
-          x={0}
+          x={align === 'right' ? width : 0}
           y={(index + 1) * lineHeight}
+          textAnchor={align === 'left' ? 'start' : 'end'}
```
jonahkagan, line 55: [textAnchor](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-anchor#example) is a crazy property. It moves the text to a diff position in relation to the `x` coordinate (rather than just moving the text within the content box like we're used to with HTML). So if you switch `textAnchor` to `'end'`, you need to move the `x` coordinate to the other end of the content box to get the box in the same position as it was before. I encapsulated this adjustment within the rendering so that the `TextBox` component behaves more like what we're used to.

apps/design/frontend/src/segmented_control.tsx
kofi-q: Yeah, that makes sense to me - I think I just didn't come across a disabled case in the wild, so didn't end up handling that specifically, but I'd expect it to show what's selected, disabled or not.

Tangentially, I've wanted to remove `SegmentedButton` altogether and replace instances with either toggle switches or radio buttons, since that's really what we're using them for in the cases I've seen...but that's a battle for another day 😅

adghayes (approved):
LGTM

PR #3718, opened 2023-07-13
feat(scan): expect printers when printing reports

adghayes:
## Overview

Currently VxScan outputs it's report opportunistically - if there's a printer available, it prints. If not, it writes to smartcard. With this change, we instead specify where to output report for a given build - either `smartcard`, `laser-printer`, and `thermal-sheet-printer` - with the `REACT_APP_VX_PRECINCT_REPORT_DESTINATION` env. variable.

If the setting is `smartcard`, the report is always exported to the smartcard regardless of printers. We anticipate removing smartcard report transfer eventually anyway.

If the setting is `laser-printer` or `thermal-sheet-printer`, we prevent transitioning polls until a printer is attached. The language should be workshopped but, as a stopgap to prevent an accident, this will do the job:

| Setting | No Printer | Printer |
| ------------- | ------------- | ------------- |
| either `printer` setting, before printing | Screen Shot 2023-07-13 at 2 18 16 PM | Screen Shot 2023-07-13 at 2 18 53 PM |
| `laser-printer` | Screen Shot 2023-07-13 at 2 20 34 PM  | Screen Shot 2023-07-13 at 2 20 47 PM  |
| `thermal-sheet-printer` | Screen Shot 2023-07-13 at 2 23 16 PM  | Screen Shot 2023-07-13 at 2 23 07 PM |

Note: I'm using the ENV variable to fix the problem of telling which kind of printer is attached, rather than trying to parse the printer information. This means that if you attach a laser printer in `thermal-sheet-printer` mode, the printing will work find but you'll still get the "Insert X Sheets..." prompt.

I've left the options to open, close, and pause polls in the other poll worker menu untouched to serve as an escape hatch.

## Testing Plan
- Manual Testing
- Added light automated testing covering the m17 use case

## Checklist


carolinemodic (approved):
thank you drew!

PR #3723, opened 2023-07-14
env variable config for bakeoff

carolinemodic:
Set up the env file for images for m17b. @benadida what are we doing with livecheck?


benadida:
> Set up the env file for images for m17b. @benadida what are we doing with livecheck?

I'd like to have it as an option, even if I don't know yet if I'm going to demo it.

adghayes:
LGTM. Sidebar: I think we should change the dev dock and zero smartcard pin on `main` because I think it's a risk to have them be defaults.

carolinemodic:
update with livecheck as true

mattroe:
oh nice we added the env variable for printer vs smartcard!

.env
```diff
@@ -1,17 +1,16 @@
...
 REACT_APP_VX_SKIP_CVR_ELECTION_HASH_CHECK=FALSE
 REACT_APP_VX_SKIP_SCAN_ELECTION_HASH_CHECK=FALSE
 REACT_APP_VX_SKIP_BALLOT_PACKAGE_AUTHENTICATION=FALSE
 REACT_APP_VX_SKIP_CAST_VOTE_RECORDS_AUTHENTICATION=FALSE
-REACT_APP_VX_DISABLE_CVR_ORIGINAL_SNAPSHOTS=FALSE
-REACT_APP_VX_CONVERTER=ms-sems
+REACT_APP_VX_DISABLE_CVR_ORIGINAL_SNAPSHOTS=TRUE
 REACT_APP_VX_SCREEN_ORIENTATION=portrait
-REACT_APP_VX_PRECINCT_REPORT_DESTINATION=smartcard
+REACT_APP_VX_PRECINCT_REPORT_DESTINATION=laser-sheet-printer
```
benadida, line 16: laser? not thermal?

mattroe:
@adghayes did that so fast I didn't even notice yesterday!

.env
carolinemodic: oh @adghayes what is the right string? you said laser-sheet-printer in the doc

mattroe: thermal per https://github.com/votingworks/vxsuite/pull/3718

carolinemodic: updated to thermal

adghayes: oops, thanks for catching my mistake

benadida (approved):

PR #3726, opened 2023-07-17
admin(perf, refactor): quicker information about write-in queue sizes

adghayes:
With 500,000 CVRs loaded up and around 600,000 write-ins, the write-in adjudication experience is super slow. #3725 made a very significant improvement, but it's still slow. The main reason is that after _every_ adjudication, the adjudication flow refetches the _entire_ list of write-ins for that contest. We need to avoid doing that! 

My idea for doing that is to:

a) fetch only the sequence of write-in IDs, with no status, so it does not have to be updated after every adjudication
b) fetch counts of completed vs. left to adjudicate separately

We _could_ do (a) even more cleanly by implementing a pagination API, but I think that will be somewhat complex given other requirements, and fetching a list of 500,000 or even 5,000,000 IDs should be OK. Leaving pagination for another day.

This PR sets up (b). Alters our over-complex endpoint for getting write-in summaries (which has been effectively replaced by backend write-in tabulation) with an endpoint that only gets the counts of adjudicated vs. pending. In addition, to avoid joining on the `cvrs` table for every `write-ins` query, denormalizes a smidgen of data. 

I'll follow this up with a PR focusing on (a).

### DIFF
--- a/apps/admin/backend/schema.sql
+++ b/apps/admin/backend/schema.sql
@@ -20,6 +20,7 @@ create table write_in_candidates (
 create table write_ins (
   id varchar(36) primary key,
   cvr_id varchar(36) not null,
+  election_id varchar(36) not null,
   side text not null check (side = 'front' or side = 'back'),
   contest_id text not null,
   option_id text not null,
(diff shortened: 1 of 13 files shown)


kofi-q (approved):
LGTM!

PR #3731, opened 2023-07-17
Write-In Adjudication Performance Changes + Flow Tweaks

adghayes:
@kofi-q review commit-by-commit recommended

## Overview

When loading up VxAdmin with lots of CVRs and write-ins, the write-in adjudication flow became very sluggish. We were fetching a ton of data after every adjudication. The main point of this PR is to make the data fetching more efficient so the WIA UI is more responsive.

The PR also introduces two flow tweaks to make WIA easier:
- maintains a write-in sequence, based on when a write-in was imported, such that whenever write-ins are added they are added to the end of the queue
- when you click to adjudicate a contest, will automatically drop you on the _first pending_ write-in. so you don't have to scroll forward to adjudicate a new batch, and if you accidentally miss one you can find it easily.

## Demo Video or Screenshot

(has audio)

https://github.com/votingworks/vxsuite/assets/37960853/60547e3a-3245-4671-b52d-a815f8156a28

## Testing Plan
- Manual Testing
- New & Updated Automated Testing

## Checklist

### DIFF
--- a/apps/admin/backend/schema.sql
+++ b/apps/admin/backend/schema.sql
@@ -18,7 +18,8 @@ create table write_in_candidates (
 );
 
 create table write_ins (
-  id varchar(36) primary key,
+  sequence_id integer primary key autoincrement,
+  id varchar(36) not null unique,
   cvr_id varchar(36) not null,
   election_id varchar(36) not null,
   side text not null check (side = 'front' or side = 'back'),
--- a/apps/admin/frontend/src/api.ts
+++ b/apps/admin/frontend/src/api.ts
@@ -14,7 +14,6 @@ import {
   useQueryClient,
 } from '@tanstack/react-query';
 import * as grout from '@votingworks/grout';
-import { Id } from '@votingworks/types';
 
 export type ApiClient = grout.Client<Api>;
 
@@ -216,14 +215,19 @@ export const getCastVoteRecordFileMode = {
   },
 } as const;
 
-type GetWriteInsInput = QueryInput<'getWriteIns'>;
-export const getWriteIns = {
-  queryKey(input?: GetWriteInsInput): QueryKey {
-    return input ? ['getWriteIns', input] : ['getWriteIns'];
+type GetWriteInAdjudicationQueueInput =
+  QueryInput<'getWriteInAdjudicationQueue'>;
+export const getWriteInAdjudicationQueue = {
+  queryKey(input?: GetWriteInAdjudicationQueueInput): QueryKey {
+    return input
+      ? ['getWriteInAdjudicationQueue', input]
+      : ['getWriteInAdjudicationQueue'];
   },
-  useQuery(input?: GetWriteInsInput) {
+  useQuery(input: GetWriteInAdjudicationQueueInput) {
     const apiClient = useApiClient();
-    return useQuery(this.queryKey(input), () => apiClient.getWriteIns(input));
+    return useQuery(this.queryKey(input), () =>
+      apiClient.getWriteInAdjudicationQueue(input)
+    );
   },
 } as const;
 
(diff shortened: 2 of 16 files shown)


apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -428,6 +431,43 @@ export function WriteInsAdjudicationScreen(): JSX.Element {
...
+  const writeInAdjudicatedWriteInCandidateIds = writeInAdjudicationContext
+    ? writeInAdjudicationContext.relatedWriteIns
+        .filter(
+          (w): w is WriteInRecordAdjudicatedOfficialCandidate =>
+            w.status === 'adjudicated' &&
+            w.adjudicationType === 'write-in-candidate'
+        )
+        .map((w) => w.candidateId)
+    : [];
+
```
adghayes, line 507: This logic of identifying off-limits adjudication options was originally in the backend, but it made more sense to just give the frontend all the data and figure out the limits on the frontend.

apps/admin/frontend/src/api.ts
```diff
@@ -256,42 +260,37 @@ export const getWriteInCandidates = {
...
-    return queryClient.invalidateQueries({
-      predicate(query) {
-        return (
-          query.queryKey[0] === 'getWriteInDetailView' &&
-          query.queryKey[1] === castVoteRecordId &&
-          query.queryKey[2] === contestId &&
-          query.queryKey[3] !== writeInId
-        );
-      },
-    });
```
adghayes, line 288: Previously there was some extra logic to strategically invalidate the `DetailView` cache, because adjudicating a write-in can change the options for a write-in in the same ballot, same contest, but not other ballots or contests. But I think there was a lot of possible footguns, and the split between `Image` and `AdjudicationContext` allows us to prefetch and not invalidate the big data requests - the images - and always invalidate the `AdjudicationContext`.

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -622,16 +638,16 @@ export function WriteInsAdjudicationScreen(): JSX.Element {
...
-              imageUrl={writeInDetailView.imageUrl}
-              ballotBounds={writeInDetailView.ballotCoordinates}
-              writeInBounds={writeInDetailView.writeInCoordinates}
+              key={currentWriteInId} // Reset zoom state for each write-in
+              imageUrl={writeInImageView.imageUrl}
+              ballotBounds={writeInImageView.ballotCoordinates}
+              writeInBounds={writeInImageView.writeInCoordinates}
             />
-          )}
+          ) : null}
```
adghayes, line 687: If image isn't ready, just show dark background of ballot viewer. I felt this looks better than "Loading..."

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -648,12 +665,14 @@ export function WriteInsAdjudicationScreen(): JSX.Element {
...
-                          await adjudicateAsOfficialCandidate(candidate);
+                      onPress={() => {
+                        if (
+                          isWriteInAdjudicationContextFresh &&
+                          !isCurrentAdjudication
+                        ) {
+                          adjudicateAsOfficialCandidate(candidate);
                         }
                       }}
-                      disabled={!isWriteInDetailViewFresh}
```
adghayes, line 656: So, we want to prevent adjudication while the `AdjudicationContext` is not fresh. Previously we would disable the button, but I think it looked a little jarring to have the buttons flash disabled (dotted outline) and back. So I've just made the buttons no-op for that split second. I think it would be hard to hit the button at that time.

apps/admin/frontend/src/api.ts
```diff
@@ -231,6 +231,25 @@ export const getWriteInAdjudicationQueue = {
...
+      ? ['getFirstPendingWriteInId', input]
+      : ['getFirstPendingWriteInId'];
+  },
+  useQuery(input: GetFirstPendingWriteInIdInput) {
+    const apiClient = useApiClient();
+    return useQuery(
+      this.queryKey(input),
+      () => apiClient.getFirstPendingWriteInId(input),
+      {
+        cacheTime: 0,
```
adghayes, line 247: I want the query to re-run anytime its mounted and never rely on cached queries.

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -373,32 +382,60 @@ export function WriteInsAdjudicationScreen(): JSX.Element {
...
+      if (firstPendingWriteInId) {
+        setOffset(writeInQueue.indexOf(firstPendingWriteInId));
+      }
+      setInitialOffsetSet(true);
+    }
+  }, [firstPendingWriteInIdQuery, writeInQueueQuery, initialOffsetSet]);
 
-  // prefetch the next write-in image
+  // prefetch the next and previous write-in image
   useEffect(() => {
```
adghayes, line 402: This part - these two dependent effects and the dependent queries - is probably the most fragile part of the whole adjudication page. I'm not quite sure how to make less side-effect-y, but it seems reliable in manual and automated testing.

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -338,25 +344,34 @@ export function WriteInsAdjudicationScreen(): JSX.Element {
 
   const queryClient = useQueryClient();
   const apiClient = useApiClient();
-  const writeInsQuery = getWriteIns.useQuery({ contestId: contest.id });
+  const writeInQueueMetadataQuery =
+    getWriteInAdjudicationQueueMetadata.useQuery({ contestId });
```
adghayes: We're fetching this only to get the `pending` count - we could get the `total` count from the queue. It's a bit overkill to do a full table scan after every adjudication, but I realized I'm overthinking this.

apps/admin/backend/schema.sql
```diff
@@ -18,7 +18,8 @@ create table write_in_candidates (
 );
 
 create table write_ins (
-  id varchar(36) primary key,
+  sequence_id integer primary key autoincrement,
```
kofi-q, line 21: I'm probably lacking some context here, but does this provides more reliable ordering than the existing `created_at` timestamp?

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -338,55 +346,97 @@ export function WriteInsAdjudicationScreen(): JSX.Element {
...
+      !!currentWriteInId
+    );
 
-  const [doubleVoteAlert, setDoubleVoteAlert] = useState<DoubleVoteAlert>();
-  const [showNewWriteInCandidateForm, setShowNewWriteInCandidateForm] =
-    useState(false);
-  const newWriteInCandidateInput = useRef<HTMLInputElement>(null);
-  const nextButton = useRef<Button>(null);
-  const firstAdjudicationButton = useRef<Button>(null);
+  // sets the user's position in the adjudication clue to the first pending write-in
```
kofi-q, line 385: sp: `s/clue/queue/`?

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
kofi-q, line 1: No action needed in this PR, but high-level nudge as you continue to refine this flow: this file has become way too huge/hard to read and continues to grow, so wherever you see opportunities, might be worth breaking stuff out (and at least, adding new code in helper utils that get used here vs adding directly here)

kofi-q (approved):
LGTM - glad to see these perf improvements

apps/admin/backend/schema.sql
adghayes: Yeah, I looked at the `created_at` timestamp and found that it didn't have the granularity to distinguish between write-ins added in sequence. I think that it uses unix "seconds since" date storage, unless that was purely how it was formatting data in the console. I think SQLite has other ways of storing dates but it seemed cleanest just to have a different field.

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
adghayes: Sounds good. I just went and separated out the UI components that were easily separable.

I think in the future we'll want a screen that renders a particular write-in and then a component that contains it that manages queuing, prefetching, etc. But until we have a clearer vision for how we want to manage the queueing it's easier to leave as-is.

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
adghayes: fixed

PR #3737, opened 2023-07-18
interpreter: Fix timing mark detection for partially grainy marks

jonahkagan:
## Overview

Sometimes based on lighting conditions, there can be a strip of grainy pixels on one side of a ballot image. This causes any timing marks on a full-bleed ballot to be partially grainy. The contour detection algorithm sometimes finds contours in the grainy area, which can lead to rejected timing marks.

To fix this, we remove the logic to filter out timing marks contours that contain other contours inside of them. We think this filter was actually unnecessarily strict.

## Demo Video or Screenshot
image

## Testing Plan
Manually tested with various scan backups, diffed the results and found negligible differences in scored marks.

## Checklist

### DIFF
--- a/libs/ballot-interpreter-nh/src/timing_marks.rs
+++ b/libs/ballot-interpreter-nh/src/timing_marks.rs
@@ -403,7 +403,7 @@ pub fn find_timing_mark_shapes(
     let candidate_timing_marks = contours
         .iter()
         .enumerate()
-        .filter_map(|(i, contour)| {
+        .filter_map(|(_i, contour)| {
             if contour.border_type == BorderType::Hole {
                 let contour_bounds = get_contour_bounding_rect(contour).offset(
                     -PixelPosition::from(BORDER_SIZE),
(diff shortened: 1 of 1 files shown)


benadida (approved):

libs/ballot-interpreter-nh/src/timing_marks.rs
```diff
@@ -403,15 +403,14 @@ pub fn find_timing_mark_shapes(
     let candidate_timing_marks = contours
         .iter()
         .enumerate()
-        .filter_map(|(i, contour)| {
+        .filter_map(|(_i, contour)| {
```
eventualbuddha, line 406: Just `_` is the idiomatic way to do it if you want to ignore a parameter.

eventualbuddha (approved):

libs/ballot-interpreter-nh/src/timing_marks.rs
jonahkagan: 👍 good to know. I was just following the compiler suggestion 🤷

eventualbuddha: Also works. In practice I most see just underscore.

PR #3740, opened 2023-07-18
Fix CircleCI Tests Temporarily

adghayes:
There should be a proper fix for this later today. However, we should consider pinning our Chrome version to that which we're actually using via `electron`?


adghayes:
Need to clean up commit history, reverting to draft.

carolinemodic (approved):

adghayes:
It won't actually just let me change the `config.yml` because we are auto-generating it. @eventualbuddha any suggested workaround?

PR #3743, opened 2023-07-19
refactor(message-coder): fix flaky test

adghayes:
This flake conforms to @eventualbuddha's `fast-check` hypothesis, that sometimes `fast-check` generates values that lead to 100% coverage and sometimes they don't. 

In this case I actually refactor the code so that `bytesRequired` is never 0, because we always require at least one byte for a number. A semantic change, really, which helps avoid the flaky switch case.

### DIFF
--- a/libs/message-coder/src/byte_array_with_length_prefix.ts
+++ b/libs/message-coder/src/byte_array_with_length_prefix.ts
@@ -85,7 +85,6 @@ class FixedLengthByteArrayCoder extends BaseCoder<Uint8Array> {
     // we only support uint16 and smaller for now,
     // because otherwise the buffer sizes get too large
     switch (bytesRequired) {
-      case 0:
       case 1:
         this.lengthCoder = uint8();
         break;
--- a/libs/custom-paper-handler/src/driver/coders.ts
+++ b/libs/custom-paper-handler/src/driver/coders.ts
@@ -1,6 +1,5 @@
 import {
   CoderType,
-  byteArrayWithLengthPrefix,
   literal,
   message,
   padding,
@@ -115,7 +114,7 @@ export const RealTimeExchangeResponseWithoutData = message({
   requestId: uint8(),
   token: literal(TOKEN),
   returnCode: uint8(),
-  optionalBytes: byteArrayWithLengthPrefix(0),
+  optionalBytes: padding(1),
 });
 export type RealTimeExchangeResponseWithoutData = CoderType<
   typeof RealTimeExchangeResponseWithoutData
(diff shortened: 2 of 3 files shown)


libs/message-coder/src/byte_array_with_length_prefix.ts
```diff
@@ -80,12 +80,14 @@ class FixedLengthByteArrayCoder extends BaseCoder<Uint8Array> {
   constructor(private readonly length: number) {
     super();
 
-    const bytesRequired = Math.ceil(Math.log2(length + 1) / BITS_PER_BYTE);
+    const bytesRequired = Math.max(
+      Math.ceil(Math.log2(length + 1) / BITS_PER_BYTE),
+      1
```
eventualbuddha, line 85: Rather than clapping to 1, just assert that `length > 0`?

adghayes: When `length = 1`, `bytesRequired = 0` based on the math, so we still need to do something here.

We could further make the assertion that `length !== 0`, because why would you be encoding a fixed length array of length 0, but that's not addressing any flakiness.

kshen0: Doesn't `length=1` result in `bytesRequired=1` even without `Math.max`? Evaluating in browser console:

```
> const length = 1;
< undefined
> const BITS_PER_BYTE = 8;
< undefined
> Math.ceil(Math.log2(length + 1) / BITS_PER_BYTE)
< 1
```

Separately, I also agree we can assert `length !== 0` since encoding fixed length arrays of length 0 has ~undefined behavior.

adghayes: Ah, you're right, I was mistaken. I'll switch to making that assertion.

adghayes: Fixed it. I don't think the `assert` is even necessary as it is captured by the `default` case of the `switch` block, with accurate logging.

kshen0: Ah true, without `Math.max`, `bytesRequired = 0` when `length = 0`

kshen0:
Looking at failed tests, it seems I did the nonsensical thing of creating a byte array with `length = 0`. I think that coder `RealTimeExchangeResponseWithoutData` can be updated to `optionalBytes: padding(1)` or `optionalBytes: literal(0)`. In the manual this is semantically a response that supports a byte array, but the byte array is empty so we define it as a length 0. But fine with me to just use one of the above workarounds.

adghayes:
thanks @kshen0, I replaced it with `padding(1)`. is there a reason we couldn't we just drop it?

kshen0:
> thanks @kshen0, I replaced it with `padding(1)`. is there a reason we couldn't we just drop it?

I can't remember exactly but I _think_ all coders will error if there are unexpected bytes in the message ie. trailing bytes are not ignored

kshen0 (approved):
Thanks for fixing the flakiness and previous test with 0-length array. I had totally forgotten about that use case but I think it's cleaner for the coder library to disallow 0-length byte arrays than to allow this carveout for the paper-handler's niche edge case.

adghayes:
Thanks for helping figure this one out @kshen0!

PR #3744, opened 2023-07-19
Counting Blank Ballots Backend

adghayes:
## Overview

To meet VVSG 1.1.9-B.5, we must be able to offer a count of blank ballots. It was quick to add this to the backend, so just did it now while `admin-backend` is fresh for me. We can build the UI when we build the other reporting UI. Although I wonder if these counts will be more useful in an expanded ballot adjudication flow. Either way.

### DIFF
--- a/apps/admin/backend/src/tabulation/utils.ts
+++ b/apps/admin/backend/src/tabulation/utils.ts
@@ -30,3 +30,10 @@ export function replacePartyIdFilter(
     batchIds: filter.batchIds,
   };
 }
+
+/**
+ * Tests whether CVR votes are empty.
+ */
+export function isBlankVotes(votes: Tabulation.Votes): boolean {
+  return Object.values(votes).every((selections) => selections.length === 0);
+}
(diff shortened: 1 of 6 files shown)


apps/admin/backend/src/tabulation/utils.ts
```diff
@@ -30,3 +30,10 @@ export function replacePartyIdFilter(
     batchIds: filter.batchIds,
   };
 }
+
+/**
+ * Tests whether CVR votes are empty.
+ */
+export function isBlankVotes(votes: Tabulation.Votes): boolean {
```
kofi-q, line 37: opt: maybe something along the lines of `isBlankBallotCard` or `isBlankCard` instead?

kofi-q (approved):

apps/admin/backend/src/tabulation/utils.ts
adghayes: that sounds better, i sincerely appreciate the naming suggestions 🤖

PR #3745, opened 2023-07-19
test(custom-scanner): use seed for flaky fast-check test

adghayes:
Another `fast-check` case of flakiness, I addressed this one by fixing the random seed.

### DIFF
--- a/libs/custom-scanner/src/protocol.test.ts
+++ b/libs/custom-scanner/src/protocol.test.ts
@@ -801,7 +801,10 @@ test('getImageData', async () => {
         );
         expect(onRead).toHaveBeenCalledTimes(1);
       }
-    )
+    ),
+    {
+      seed: 0, // with a random seed, some branches are sporadically missed
+    }
   );
 });
 


libs/custom-scanner/src/protocol.test.ts
```diff
@@ -801,7 +801,10 @@ test('getImageData', async () => {
         );
         expect(onRead).toHaveBeenCalledTimes(1);
       }
-    )
+    ),
+    {
+      seed: 0, // with a random seed, some branches are sporadically missed
```
kofi-q, line 806: Don't know too much about fast-check, but does this defeat the purpose? (i.e. does this mean every test run will use the same set of random values now?)

kofi-q (approved):

libs/custom-scanner/src/protocol.test.ts
adghayes: Every test will run with the same set of random values now, yes. From my perspective, `fast-check` is useful in generating an array of test values and running through all of them. I think that the property of that array being random is less desirable for us, as it leads to test coverage issues. I will wait to merge this until @eventualbuddha is back though so he can chime in.

You could explicitly write out the case that it's missing, but figuring that out was not worth the time.

jonahkagan: We could strike a compromise by only using a fixed seed in CI

eventualbuddha: I like @jonahkagan's suggestion, but I don't feel strongly about it.

PR #3748, opened 2023-07-19
feat(scan): remove smartcard reports

adghayes:
## Overview

Removes support for saving tally reports to smartcards in VxScan.

## Checklist


.env
```diff
@@ -14,4 +14,4 @@ REACT_APP_VX_SKIP_CAST_VOTE_RECORDS_AUTHENTICATION=FALSE
 REACT_APP_VX_DISABLE_CVR_ORIGINAL_SNAPSHOTS=FALSE
 REACT_APP_VX_CONVERTER=ms-sems
 REACT_APP_VX_SCREEN_ORIENTATION=portrait
-REACT_APP_VX_PRECINCT_REPORT_DESTINATION=smartcard
+REACT_APP_VX_PRECINCT_REPORT_DESTINATION=thermal-sheet-printer
```
adghayes, line 17: We could just remove this now, but that would require differentiating on the frontend between thermal and non-thermal printers (or at least the ones we support). That's doable but it does require quite a bit of work, so easier just to punt until we settle on a single kind of printer.

PR #3750, opened 2023-07-20
[VxScan] Always reset display settings when paper leaves the scanner

kofi-q:
## Overview

Followup from https://github.com/votingworks/vxsuite/pull/3706, addressing VVSG 7.1-A (Reset to display default settings).

Got guidance from @mattroe about how to handle rejected ballots:
> I think we should reset once a ballot leaves the document scanner -- whether it's casted in the box or a voter takes it from the machine.

Updating the reset logic to always reset when the scanner moves into the `no_paper` state (vs when moving from `accepted` to `no_paper` as before).

## Demo Video or Screenshot
https://github.com/votingworks/vxsuite/assets/264902/378a12d9-794b-4192-a612-cb2746714b63

## Testing Plan
- Updating the existing unit test accordingly

## Checklist

- [x] I have added a screenshot and/or video to this PR to demo the change~

### DIFF
--- a/apps/scan/backend/src/types.ts
+++ b/apps/scan/backend/src/types.ts
@@ -7,6 +7,7 @@ import {
   PrecinctSelection,
   SystemSettings,
 } from '@votingworks/types';
+import { PrecinctScannerState } from '@votingworks/types/src/precinct_scanner';
 
 export interface MachineConfig {
   machineId: string;
@@ -19,31 +20,6 @@ export interface PageInterpretationWithAdjudication<
   interpretation: T;
   contestIds?: readonly string[];
 }
-
-export const PRECINCT_SCANNER_STATES = [
-  'connecting',
-  'disconnected',
-  'no_paper',
-  'ready_to_scan',
-  'scanning',
-  'returning_to_rescan',
-  'ready_to_accept',
-  'accepting',
-  'accepted',
-  'needs_review',
-  'accepting_after_review',
-  'returning',
-  'returned',
-  'rejecting',
-  'rejected',
-  'jammed',
-  'both_sides_have_paper',
-  'recovering_from_error',
-  'double_sheet_jammed',
-  'unrecoverable_error',
-] as const;
-
-export type PrecinctScannerState = typeof PRECINCT_SCANNER_STATES[number];
 
 export type InvalidInterpretationReason =
   | 'invalid_test_mode'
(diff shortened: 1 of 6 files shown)


adghayes (approved):

adghayes (commented):
@kofi-q I noticed while looking into `jest` upgrades that this file imports a _constant_ from `@votingworks/backend`, rather than just types. It means that the test inherits some dependencies from the backend which can break tests. Normally we just `import type` from the backend.

Can you rescope the test to a smaller number of hardcoded states so it doesn't introduce this dependency issue?

adghayes:
Alternately we could upgrade or resolve the backend package (`uuid`) that's causing trouble. I'm not sure if our `import type` rule applies to test files as well as actual files. @jonahkagan?

jonahkagan:
I'm not sure I understand what test is broken by importing that constant here. Can you point me to it?

adghayes:
> I'm not sure I understand what test is broken by importing that constant here. Can you point me to it?

The issue doesn't actually emerge until upgrade to Jest 28 - https://circleci.com/gh/votingworks/vxsuite/346867.

jonahkagan:
Oh weird. I would probably just punt and hardcode a few key precinct scanner states for now, since I don't think exhaustiveness is core to this particular test. Seems like the quickest way through this particular hurdle.

kofi-q:
Thanks for surfacing! I feel like this exposes a bit of a code organisation hole maybe? e.g. I'd expect to be able to share certain constants/types/pieces of logic between both frontend and backend packages, which would mean moving those out to some shared location - so, maybe another fix here is to move these states out to a shared library, right?

I would personally just move them out to libs/types for simplicity, but I expect others might feel strongly about this not being broad enough to belong there?

adghayes:
@kofi-q I 100% agree, I feel like this exposes two holes.

First, there was a purpose to `libs/api` of simply having shared frontend/backend _code_ and _data_ rather than simply _types_. I don't think we should re-invest in the `libs/api` model - I think that when the use case is big enough we can do the `shared` model as @jonahkagan has done for `design` or move it into whatever other library - in this case I think that a file in `libs/types` that's specific to the app is totally fine. I'm not picky about that "other library" personally. 

Second, I think we should have a lint rule preventing non-`type` imports from app backends.

kofi-q:
Done! Very minimal usage for that type, so moving it was straightforward. Should be easy to adjust if needed in the future.
https://github.com/votingworks/vxsuite/pull/3750/commits/391ab3fc6004b2883b35b98aa2a14f2d029f4877

PR #3756, opened 2023-07-25
rewrite state machine in xstate, scan and interpret

kshen0:
Recommend reviewing commit-by-commit.

## Overview
https://github.com/votingworks/vxsuite/issues/3686

Primarily
* Rewrites the mark-scan state machine to use `xstate`, borrowing from `scan`'s implementation
* Adds `scanning` and `interpreting` states between `print` and `present`

**Notes**
* The interpreting state prints the ballot interpretation to `debug`. In a future PR we'll display it on screen for the voter to confirm
* Some types are factored out from `apps/scan` and into `libs/ballot-interpreter`. Following discussion in [Slack](https://votingworks.slack.com/archives/CEL6D3GAD/p1690330715086859), `mark-scan` integrates directly with `libs/ballot-interpreter`. To avoid expanding scope of this PR, the implementation duplicates a small amount of logic for combining page interpretations into sheet interpretations. By doing this we avoid adding another dependency to the `apps/scan` interpreter, which we want to deprecate.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/32fbc22d-71b5-4e5d-a61b-5e024acf2389

## Testing Plan
* Added basic state machine test to validate it can transition between states
* Updated existing tests to work with new state machine
* More extensive testing on hold until state ordering is finalized

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (On hold until flow is finalized)


apps/scan/backend/src/interpret.test.ts
```diff
@@ -34,6 +42,36 @@ afterEach(async () => {
   await fs.rm(ballotImagesPath, { recursive: true });
 });
 
+test('configure get/set', () => {
```
kshen0, line 45: I added these tests because I had previously factored out `scan/backend/src/interpret.ts` to `libs/ballot-interpreter`. That dropped coverage in the latter package so I brought coverage back up with these tests. I reverted my changes that factored out the package, but thought I'd leave the tests in for coverage anyway.

libs/custom-paper-handler/src/driver/driver_interface.ts
```diff
@@ -0,0 +1,112 @@
+import { Coder, CoderError, Uint16, Uint8 } from '@votingworks/message-coder';
```
kshen0, line 1: I added this interface since we'll need one for an eventual hardware mock for dev environments.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
+    return (
+      timer(0, pollingIntervalMs)
+        // `pipe` forwards the value from the previous function to the next unary function ie. switchMap.
+        // In this case there is no value. The combination of timer(...).pipe() is so we can execute the
+        // function supplied to `switchMap` on the specified interval.
+        .pipe(
+          // `switchMap` returns an Observable that emits events.
+          // Why do we use `switchMap`? Is it because `switchMap` only emits items from the most recent
+          // inner Observable ie. it will only emit the latest PaperHandlerStatusEvent?
+          switchMap(async () => {
```
kshen0, line 162: @jonahkagan I think I have a decent idea of how this implementation of the status poller works except for our use of `swichMap` here. Do you remember why we use it?

jonahkagan: I don't remember, but maybe it has to do with making the timeout work?

kshen0:
Assigning @eventualbuddha for scan/interpret review and @jonahkagan for state machine review

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
```diff
@@ -0,0 +1,225 @@
...
+  debug('Completed printing %d chunks total', customChunkedBitmaps.length);
+}
+
+export async function scanAndSave(
+  driver: PaperHandlerDriver
+): Promise<SheetOf<string>> {
+  debug('+scanAndSave');
+  const dir = '/tmp';
+  const dateString = new Date().toISOString();
+  const pathOutFront = join(dir, `ballot-statemachine-${dateString}-A.jpg`);
```
eventualbuddha, line 155: I don't think there's any reason to build our own temporary file path generator. This one has some downsides:
- it may not be that `/tmp` is always where we want to write temporary files
- this file may already exist (though it shouldn't)
- if this function were somehow called multiple times in the same instant you'd get the same path for both

```suggestion
  const pathOutFront = tmpNameSync();
```

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
```diff
@@ -0,0 +1,225 @@
...
+  // We can only print to one side from the thermal printer, but the interpret flow expects
+  // a SheetOf 2 pages. Use an image of a blank sheet for the 2nd page.
+  const blankSheetFixturePath = join(__dirname, 'fixtures', 'blank-sheet.jpg');
```
eventualbuddha, line 158: It's unfortunate that we're still passing paths around instead of dealing with image data directly. If we were using image data directly we could just do `createImageData(width, height)` to make a blank page.

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
```diff
@@ -0,0 +1,225 @@
...
+  );
+
+  const combinedInterpretation: SheetInterpretationWithPages = {
+    ...combinePageInterpretationsForSheet(pageInterpretations),
+    pages: pageInterpretations,
+  };
+
+  // Use JSON.stringify instead of string interpolation because the latter
+  // only prints one level deep
+  debug(`interpretation: ${JSON.stringify(combinedInterpretation, null, 2)}`);
```
eventualbuddha, line 223: Does this do what you want?
```suggestion
  debug('interpretation: %O', combinedInterpretation);
```

apps/mark-scan/backend/src/custom-paper-handler/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const PAPER_HANDLER_STATUS_POLLING_INTERVAL_MS = 100;
+// Slower interval for limited hardware
+export const DEV_PAPER_HANDLER_STATUS_POLLING_INTERVAL_MS = 200;
+export const PAPER_HANDLER_STATUS_POLLING_TIMEOUT_MS = 10_000;
+
+export const BALLOT_IMAGES_PATH = '/tmp';
```
eventualbuddha, line 6: Placeholder, I assume? Maybe add a comment explaining what it should be?

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
```diff
@@ -0,0 +1,225 @@
...
+  let time = Date.now();
+  const pages: ImageData[] = [];
+  for await (const { page, pageCount } of pdfToImages(Buffer.from(pdfData), {
+    scale: 200 / 72,
+  })) {
+    assert(pageCount === 1, `Unexpected page count ${pageCount}`);
+    pages.push(page);
+  }
+  const page = pages[0];
+  assert(page, 'Unexpected undefined page');
```
eventualbuddha, line 118: Using a loop and pushing to an array make it a bit less clear that you're looking for the first/only page. Consider something like this instead?
```suggestion
  const pageInfo = await iter(
    pdfToImages(Buffer.from(pdfData), { scale: 200 / 72 })
  ).first();
  assert(
    pageInfo?.pageCount === 1,
    `Unexpected page count ${pageInfo?.pageCount ?? 0}`
  );
  const { page } = pageInfo;
```

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
```diff
@@ -0,0 +1,225 @@
...
+  debug(`pdf to image took ${Date.now() - time} ms`);
```
eventualbuddha, line 119: Consider using `time` from `@votingworks/utils`.

libs/custom-paper-handler/src/bits.ts
```diff
@@ -348,6 +353,12 @@ export function Uint8ToBinaryArray(value: Uint8): BinaryArray {
   return bitArray;
 }
 
+export function arrayBufferToHexString(buffer: ArrayBuffer): string {
```
eventualbuddha, line 356: Probably useful for debugging, but not used anywhere in this PR. Consider removing.

libs/custom-paper-handler/src/cli/driver_cli.ts
```diff
@@ -51,11 +63,17 @@ async function outputSampleBallotPdfWidth() {
   console.log(`First page of sample ballot is ${page.width} dots`);
 }
 
+async function scan(driver: PaperHandlerDriverInterface): Promise<void> {
+  const dateString = new Date().toISOString();
+  const pathOut = join('/tmp', `ballot-driver-cli-${dateString}.jpg`);
```
eventualbuddha, line 68: Similar note as above. Consider `tmpNameSync()`, possibly with options to add `.jpeg` extension.

libs/custom-paper-handler/src/driver/coders.ts
```diff
@@ -8,7 +8,21 @@ import {
...
+  data: Uint8Array;
+  width: number;
+}
+
+export const TransferOutRealTimeRequest = message({
+  stx: literal(START_OF_PACKET),
+  requestId: uint8(),
+  token: literal(TOKEN),
+  // Treat "optional data length" byte as padding when no additional data is supplied
+  padding: padding(8),
```
eventualbuddha, line 23:
```suggestion
  unusedOptionalDataLength: padding(8),
```

libs/custom-paper-handler/src/driver/coders.ts
```diff
@@ -128,13 +142,28 @@ export const ScanResponse = message({
   returnCode: uint8(),
   cis: uint8(0x00 | 0x01 | 0x02),
   scan: uint8(),
-  sizeX: uint16(),
-  sizeY: uint16(),
+  sizeX: uint16(undefined, { littleEndian: false }),
+  sizeY: uint16(undefined, { littleEndian: false }),
   status: uint16(),
```
eventualbuddha, line 147: The `sizeX` and `sizeY` are big endian but `status` isn't? 👻

libs/custom-paper-handler/src/driver/coders.ts
```diff
@@ -128,13 +142,28 @@ export const ScanResponse = message({
...
   dummy: padding(32),
 });
 export type ScanResponse = CoderType<typeof ScanResponse>;
 
+export function scanResponseToHexString(r: ScanResponse): string {
```
eventualbuddha, line 152: Unused in this PR. Consider removing.

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -477,53 +443,95 @@ export class PaperHandlerDriver {
...
-
       const header = responseBuffer.slice(0, SCAN_HEADER_LENGTH_BYTES);
       const response: ScanResponse = ScanResponse.decode(
         Buffer.from(header)
       ).unsafeUnwrap();
-
       scanStatus = response.returnCode;
       const { sizeX, sizeY } = response;
-      const pixelsPerByte = response.scan <= 5 ? 1 : 8;
+      // The last sizeX returned by the scan command is 0, so store the first one for use later
```
eventualbuddha, line 464: I don't think I understand this comment. Where is the check for 0?

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -477,53 +443,95 @@ export class PaperHandlerDriver {
...
     const imageData: Uint8Array[] = [];
 
+    // Assumptions:
+    // 1. There's at least 1 data block
+    // 2. Each data block can fit within a single `transferInGeneric` call
     while (scanStatus === OK_CONTINUE) {
       const rawResponse = await this.transferInGeneric();
       assert(rawResponse?.data);
 
       const responseBuffer = rawResponse.data.buffer;
```
eventualbuddha, line 457: I think we're just getting lucky here that the returned `ArrayBuffer` is _only_ the data we care about and not some larger shared buffer managed by `usb`. Accessing the underlying `ArrayBuffer` of a `DataView` should _always_ take into account the `byteOffset` and `byteLength` values:

```suggestion
      const responseBuffer = new Uint8Array(
        rawResponse.data.buffer,
        rawResponse.data.byteOffset,
        rawResponse.data.byteLength
      );
```

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -477,53 +443,95 @@ export class PaperHandlerDriver {
...
       ).unsafeUnwrap();
-
       scanStatus = response.returnCode;
       const { sizeX, sizeY } = response;
-      const pixelsPerByte = response.scan <= 5 ? 1 : 8;
+      // The last sizeX returned by the scan command is 0, so store the first one for use later
+      if (width === -1) {
+        width = sizeX;
+      }
+      const pixelsPerByte = response.scan <= 0x05 ? 1 : 8;
```
eventualbuddha, line 468: `0x05` feels like a magic number. Are we using a straight numeric `uint8` for something that should be an 8-bit enum?

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -477,53 +443,95 @@ export class PaperHandlerDriver {
...
+          .toUpperCase()}`
+      );
+    }
+    const imageBuf = Buffer.concat(imageData);
     return createImageData(
-      Uint8ClampedArray.from(Buffer.concat(imageData)),
-      1728
+      Uint8ClampedArray.from(imageBuf),
+      // Start scan ticket (0x1c 0x53 0x50 0x53) returns width of data block [Size X]; hardcode this value for now but
+      // we should read it from the response header
```
eventualbuddha, line 493: What's the hardcoded part?

eventualbuddha (commented):
I only really reviewed the non-state machine parts. Looks generally okay but there are a few things that need to change.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
+    pdfData: Uint8Array,
+    options: Partial<ImageConversionOptions> = {}
+  ): Promise<void> {
+    this.machineService.send({
+      type: 'VOTER_INITIATED_PRINT',
+    });
 
-  async logStatus(): Promise<void> {
-    debug('%O', await this.driver.getPaperHandlerStatus());
+    return driverPrintBallot(this.driver, pdfData, options);
```
jonahkagan, line 117: Should this printing method get called from within the state machine?

The approach we used in apps/scan was to have the state machine manage all of the business logic, including triggering side effects, so that it could all be understood in one place.

Could that approach be applied here?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
+  constructor(
+    private readonly driver: PaperHandlerDriver,
+    private readonly machineService: Interpreter<
+      Context,
+      any,
+      PaperHandlerStatusEvent,
+      any,
+      any
+    >
+  ) {}
```
jonahkagan, line 77: I know this class was here before you started, but I wonder if it would be better to match the pattern of apps/scan (constructing a plain object with methods) in order to have less cognitive overhead when looking at the two files.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
+    if (isPaperInOutput(paperHandlerStatus)) {
+      event = { type: 'BALLOT_PRINTED' };
+    } else {
+      event = { type: 'PAPER_INSIDE' };
+    }
+  } else if (isPaperReadyToLoad(paperHandlerStatus)) {
+    event = { type: 'PAPER_READY_TO_LOAD' };
   }
+  debug(`Emitting event ${event.type}`);
+  return event;
```
jonahkagan, line 136: I would recommend having the scanner status events all be descriptive of what the scanner sees rather than interpreting them. E.g. if paper is in the output position, it doesn't necessarily mean there was a ballot printed - that's contextual information that you can only know based on when you receive the status. So I might change that to `PAPER_IN_OUTPUT` or something. That way if you get a scanner status that says `PAPER_IN_OUTPUT` in a time other than when the ballot was printed, you know something went wrong.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
+    },
+    states: {
+      no_paper: NoPaperState,
+      loading_paper: LoadingPaperState,
+      waiting_for_ballot_data: WaitingForBallotDataState,
+      printing_ballot: PrintingBallotState,
+      scanning: ScanningState,
+      interpreting: InterpretingState,
+      presenting_ballot: PresentingBallotState,
+      final_state: FinalState,
```
jonahkagan, line 366: What do you think about inlining the states? I think it's a little easier to read when you can easily find the state name referenced by another state in the same object.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
+  BaseActionObject
+> = {
+  invoke: pollPaperStatus(),
+  entry: (context) => {
+    // Paper can trigger sensors before it's actually able to be loaded,
+    // so we wait to decrease chance of failed load
+    setTimeout(async () => {
+      await context.driver.loadPaper();
+      await context.driver.parkPaper();
+    }, 300);
```
jonahkagan, line 214: Consider using the xstate delay functionality

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
-  }
+const NoPaperState: StateNodeConfig<
+  Context,
+  any,
+  PaperHandlerStatusEvent,
+  BaseActionObject
+> = {
+  invoke: pollPaperStatus(),
+  on: {
+    PAPER_INSIDE: {
```
jonahkagan, line 191: How does it happen that the status goes straight from NO_PAPER to PAPER_INSIDE without going through PAPER_READY_TO_LOAD first?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
-    }
-    await writeImageData(pathOut, colorResult);
-  }
+const WaitingForBallotDataState: StateNodeConfig<
+  Context,
+  any,
+  PaperHandlerStatusEvent,
+  BaseActionObject
+> = {
+  invoke: pollPaperStatus(),
```
jonahkagan, line 229: Do we need to poll for paper status if the only transition out of this state is for the voter to initiate a print?

Also, I'm not sure I understand what it means to wait for ballot data here. Could you give me a bit of context to help me understand?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
-    debug(`image to binary took ${Date.now() - time} ms`);
-    time = Date.now();
+const InterpretingState: StateNodeConfig<
+  Context,
+  any,
+  PaperHandlerStatusEvent,
+  BaseActionObject
+> = {
+  invoke: [
+    { ...pollPaperStatus() },
```
jonahkagan, line 276: Do we need to poll paper status here? I don't see any transitions out of this state based on paper status.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
+        const precinct = precincts[precincts.length - 1];
+        const precinctSelection = singlePrecinctSelectionFor(precinct.id);
+        const testMode = true;
+        return interpretScannedBallots(
+          electionDefinition,
+          precinctSelection,
+          testMode,
+          scannedImagePaths,
+          sheetId
+        );
```
jonahkagan, line 300: Nit: maybe factor this out into a helper function to keep the state machine more concise

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
-  Optional<PaperHandlerStateMachine>
+const PresentingBallotState: StateNodeConfig<
+  Context,
+  any,
+  PaperHandlerStatusEvent,
+  BaseActionObject
+> = {
+  entry: async (context) => {
+    await context.driver.presentPaper();
+  },
```
jonahkagan, line 317: How do you get out of this state?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1,267 +1,394 @@
...
+const FinalState: StateNodeConfig<
+  Context,
+  any,
+  PaperHandlerStatusEvent,
+  BaseActionObject
+> = {
+  type: 'final',
+  entry: () => {
+    debug('Entered ending state');
+  },
```
jonahkagan, line 329: How do you get into this state?

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
kshen0: Fixed path was useful when debugging why the scan wasn't working, but agree this is a better approach now that scanning is working reliably.

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
kshen0: To clarify, this would be blocked by a future refactor of the interpret pipeline ie. no action needed in this PR?

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
kshen0: It prints the first level but level 2 evaluates to `[object Object]` :(

libs/custom-paper-handler/src/cli/driver_cli.ts
kshen0: Since the CLI is used only for debugging on the hardware, I think writing to `/tmp` with a human-readable name is the desired behavior here. What do you think?

libs/custom-paper-handler/src/driver/driver.ts
kshen0: Hmm sorry I know you flagged this in a previous PR but looks like I missed this incorrect usage.

libs/custom-paper-handler/src/driver/driver.ts
kshen0: We want to use `sizeX` returned by the scan command later in this function. Each packet sent by the scanner includes a header that specifies sizeX and sizeY. So for packets 0..n-1 we get a packet like this (pseudocode):

```
scanStatus: OK_CONTINUE,
sizeX: 1738,
sizeY: ...
dataBlock: // Scan image data
```

the last packet _n_ is empty so we get

```
scanStatus: OK_NO_MORE_DATA,
sizeX: 0,
sizeY: 0,
dataBlock: // no bytes
```

Ideally there would be no empty packet, and the last packet would be:

```
scanStatus: OK_NO_MORE_DATA,
sizeX: 1738,
sizeY: ...
dataBlock: // Scan image data
```

In that case we could just store the last `sizeX` value. But since we _do_ get an empty packet with `sizeX=0` we store `sizeX` from the first packet instead.

I wrote this comment because I felt the empty packet at the end was nonintuitive and it wasn't clear why we store the first `sizeX` value. I'll try to rewrite this comment to be more clear.

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
kshen0: https://github.com/votingworks/vxsuite/pull/3756/commits/d3dd266a08d0e239ffae8085bfb715f31235d930#diff-7222935cb6da120f6165330ebde04dd946fdd1cdfa28d1a417fd79e7c6d7cfb3R153

apps/mark-scan/backend/src/custom-paper-handler/constants.ts
kshen0: This was used when calling `interpretSheetAndSaveImages`, but in https://github.com/votingworks/vxsuite/pull/3756/commits/4669a7b630e4cc2e724a7685f0a8930449beefa9 I removed it.

That commit simplifies the state machine to just call `interpretSheet` and remove `combinePageInterpretationsForSheet`. I think this is the right approach because
* mark-scan doesn't ballot images as an output. It saves after scan because image filepaths are needed as an input for interpretation, but we don't need the images to be saved after interpretation, which is what `interpretSheetAndSaveImages` does
* `combinePageInterpretationsForSheet` is a pared down version of logic from VxScan that determines whether the combination of sheet interpretations is valid, since VxScan may be fed a HMPB or BMD ballot (and those ballots may have invalid sheet configurations due to human error). But VxMarkScan (a) scans BMD ballots only and (b) prints and scans before the paper is moved outside the machine, greatly reducing risk of invalid sheet configurations.

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
kshen0: Great point, thanks for the suggestion.
https://github.com/votingworks/vxsuite/pull/3756/commits/736f6f48826c785778fe1d8096878def667c56a4
nb. `200 / 72` is a magic number from the print prototype. If I had to guess we scale like this because the pdf is at 200 dpi and we want to print at 72 dpi. I'll try to move these numbers into constants if possible.

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
kshen0: Removed timing info since we don't do anything with that info. Can always re add it https://github.com/votingworks/vxsuite/pull/3756/commits/736f6f48826c785778fe1d8096878def667c56a4

libs/custom-paper-handler/src/bits.ts
kshen0: https://github.com/votingworks/vxsuite/pull/3756/commits/eb2ca2aa84cb657a35727fcbf2498525601d48fc

libs/custom-paper-handler/src/driver/driver.ts
kshen0: https://github.com/votingworks/vxsuite/pull/3756/commits/bb9464c00ad71cfccdf46d47bd4cab15e96a6d7e

libs/custom-paper-handler/src/driver/driver.ts
kshen0: Turned into an enum in https://github.com/votingworks/vxsuite/pull/3756/commits/2a8fe012d65eb2ae0083392855055749add9cad4

libs/custom-paper-handler/src/driver/driver.ts
kshen0: Outdated comment, sorry for leaving that in. We are actually reading width from the response header.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: We don't need to poll paper status at the moment, and we're unlikely to in the next round of exception handling, so I'll remove this.

Waiting for ballot data means the state machine is waiting to receive the ballot image as a `Buffer` from the frontend. In terms of user flows, this means the blank paper has been loaded successfully and the voter is still making selections. The state machine is waiting for the user to finish their selections and press "Print", which sends the data and kicks off the print command.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: `paperStatus` is a bit of a misnomer; it should be something like `pollHardwareStatus`. I've been leaving this in every step because in the future we'll need to handle changes to paper and hardware status. But like you note in `WaitingForBallotDataState` it's not needed right now and this step is at lower risk of abnormal hardware status, so I'll remove as well.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: This is the end of the flow implemented in this PR. In the future, you get out of this state by confirming your selections are correct. That will transition to a state that ingests the printed ballot and ejects out the rear to the ballot box.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: The test is the only place that transitions to this state. I added this because transitioning away from a state that invokes `pollPaperStatus` is necessary to gc the machine, and I wasn't sure which states would invoke `pollPaperStatus` (right now most of them do). But since we explicitly `stop()` the machine I think we can remove this.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: I saw `scan` does that and was split on approach for `mark-scan`. On one hand I like being able to see all the high level states at a glance, something that's hard to do with `scan`'s inlined states, but on the other hand

* it's annoying and error-prone to keep the separate state declarations in order
* I agree it's easier to do an in-depth read of the code when they're inlined

I think the pros of inlining outweigh the cons so I'll update.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Do you mean just the `driverPrintBallot` call? We could definitely move that call to a state; I agree it's cleaner that way.

Not sure if the way I support outside interaction with the state machine here is the best way. The goal is to expose a method to allow the API to cause state transitions (in this case from WaitingForBallotData to Printing). Just wrapping `send` seemed like the most straightforward way but if there's a better way let me know.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Good call, this is left over from before I separated out `LoadingPaperState` from `NoPaperState`. This event handler should have been deleted before requesting review, sorry!

libs/custom-paper-handler/src/cli/driver_cli.ts
eventualbuddha: I could buy that. Consider using `os.tmpdir()`?

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
eventualbuddha: I think it's actually the reverse: the PDF is 72 dpi and we want to print at 200 dpi. But yes, that's the rationale.

libs/custom-paper-handler/src/driver/constants.ts
```diff
@@ -17,6 +17,31 @@ export const START_OF_PACKET: Uint8 = 0x02;
...
+// Exhaustive list of scan types whose data blocks will be 8 bpp.
+// Manual reference: "Start scan ticket" 0x1C 0x53 0x50 0x53
+export enum ScanTypes8BitsPerPixel {
+  NA = 0x00,
+  RED = 0x01,
+  GREEN = 0x02,
+  BLUE = 0x03,
+  ULTRAVIOLET = 0x04,
+  GRAY = 0x05,
+}
```
eventualbuddha, line 44: I find this name a little misleading because the actual return value from "Start scan ticket" does _not_ need to represent an 8bpp value. You're only encoding those values that _do_ represent them and then using presence in the enum to imply 8bpp. I think this might be clearer, if more verbose:

```suggestion
// Exhaustive list of scan types.
// Manual reference: "Start scan ticket" 0x1C 0x53 0x50 0x53
export enum ScanTypes {
  NA = 0x00,
  /**
   * Red light scan (8bpp)
   */
  RED = 0x01,
  /**
   * Green light scan (8bpp)
   */
  GREEN = 0x02,
  /**
   * Blue light scan (8bpp)
   */
  BLUE = 0x03,
  /**
   * Ultraviolet light scan (8bpp)
   */
  ULTRAVIOLET = 0x04,
  /**
   * White light scan (8bpp)
   */
  GRAY = 0x05,
  /**
   * first line RED, second line GREEN, third line BLUE (24bpp)
   */
  RGB_RAW = 0x06,
  /**
   * first line RED, second line GREEN, third line BLUE, fourth line UV (32bpp)
   */
  RGBU_RAW = 0x07,
  /**
   * Black and white with red light (1bpp)
   */
  BW_RED = 0x08,
  /**
   * Black and white with green light (1bpp)
   */
  BW_GREEN = 0x09,
  /**
   * Black and white with blue light (1bpp)
   */
  BW_BLUE = 0x0a,
  /**
   * Black and white with ultraviolet light (1bpp)
   */
  BW_ULTRAVIOLET = 0x0b,
  /**
   * Black and white with white light (1bpp)
   */
  BW = 0x0c,
}

export function getBitsPerPixelForScanType(scanType: ScanTypes): number {
  switch (scanType) {
    case ScanTypes.NA:
      return 0;
    case ScanTypes.RED:
    case ScanTypes.GREEN:
    case ScanTypes.BLUE:
    case ScanTypes.ULTRAVIOLET:
    case ScanTypes.GRAY:
    case ScanTypes.BW_RED:
    case ScanTypes.BW_GREEN:
    case ScanTypes.BW_BLUE:
    case ScanTypes.BW_ULTRAVIOLET:
    case ScanTypes.BW:
      return 8;
    case ScanTypes.RGB_RAW:
      return 24;
    case ScanTypes.RGBU_RAW:
      return 32;
    default:
      throwIllegalValue(scanType);
  }
}
```

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: I think that's a valuable change to make. Would you be opposed if we keep this thin wrapper class in this PR and refactor after we hit the demo milestone? I don't think the cost of moving to the`apps/scan` pattern will increase much if we wait, and I think we should prioritize exception handling first.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
jonahkagan: Yes, I meant the `driverPrintBallot` call.

Wrapping `send` seems like the right approach for managing outside interaction. Pretty sure that's how I did it in `apps/scan`.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
jonahkagan: Yeah, fine to punt til later. I didn't realize this PR was just aiming for a demo. That helps explain why the state machine is so much simpler than `apps/scan`!

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
jonahkagan: Thanks, that's helpful!

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
jonahkagan: Gotcha

libs/custom-paper-handler/src/driver/constants.ts
kshen0: I agree with your take, thanks for the suggestion. I'll merge this in.

apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
kshen0: [3b78ff6](https://github.com/votingworks/vxsuite/pull/3756/commits/3b78ff654295ec18c6f65777425c50dc280ee0b5)

libs/custom-paper-handler/src/cli/driver_cli.ts
kshen0: [d00818c](https://github.com/votingworks/vxsuite/pull/3756/commits/d00818cc97ffa81cfbe7ea6ae569d2be1195ee4c)

libs/custom-paper-handler/src/driver/constants.ts
kshen0: https://github.com/votingworks/vxsuite/pull/3756/commits/51a0803a462c3a7eb61208d3925e6a07f4989f8f

PR #3767, opened 2023-07-27
Kevin/vxmarkscan move precinct selection to backend

kshen0:
## Overview
Moves `precinctSelection` CRUD to the backend.
* Uses the same schema as VxMark
* Adds grout endpoints to get and set precinct selection
* Only allows unsetting of precinct selection when entire machine is unconfigured via `unconfigureMachine` endpoint. AFAICT this is consistent with the UI, which does not allow unsetting precinct selection once it's set.

## Demo Video or Screenshot
No new behavior. Manual tests of existing behavior:

**Set precinct**

https://github.com/votingworks/vxsuite/assets/1060688/c0bc6646-c660-4eec-be96-996e7fb51ffd

**Precinct selection is cleared when entire machine is cleared**

https://github.com/votingworks/vxsuite/assets/1060688/7f56481f-9b3a-44ca-b65c-16561fafd37f

## Testing Plan
* Manually tested existing set/unset flow
* Added automated tests for automatic precinct selection when there is only 1 precinct in the election
* Updated existing tests to expect new react query calls

## Checklist

- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates (no new behavior)


apps/mark-scan/backend/src/app.ts
```diff
@@ -101,6 +106,16 @@ function buildApi(
...
 
+    setPrecinctSelection(input: {
+      precinctSelection: PrecinctSelection | AllPrecinctsSelection;
+    }): void {
+      workspace.store.setPrecinctSelection(input.precinctSelection);
+    },
+
+    getPrecinctSelection(): Result<Optional<PrecinctSelection>, Error> {
+      return ok(workspace.store.getPrecinctSelection());
+    },
```
kshen0, line 117: Open to feedback on when to use `Result`. In this case, I use `Result` for `get` because precinct selection can be undefined but grout requires a defined response, so we meet that requirement by wrapping `undefined` in a `Result`. In `set` there's nothing to return so it felt unnecessary to return an empty `Result` 100% of the time.

apps/mark-scan/backend/src/app.ts
```diff
@@ -158,6 +173,13 @@ function buildApi(
       });
       workspace.store.setSystemSettings(systemSettings);
 
+      const { precincts } = electionDefinition.election;
+      if (precincts.length === 1) {
+        workspace.store.setPrecinctSelection(
+          singlePrecinctSelectionFor(precincts[0].id)
+        );
+      }
```
kshen0, line 181: This is existing behavior in the frontend that I ported to the backend. See: Deleted test `apps/mark-scan/frontend/src/app_single_precinct_election.test.tsx`

apps/mark-scan/frontend/src/pages/insert_card_screen.tsx
```diff
@@ -34,7 +38,10 @@ export function InsertCardScreen({
   showNoAccessibleControllerWarning,
 }: Props): JSX.Element {
   const getPrecinctSelectionQuery = getPrecinctSelection.useQuery();
-  const precinctSelection = getPrecinctSelectionQuery.data?.ok();
+  let precinctSelection: Optional<PrecinctSelection>;
+  if (getPrecinctSelectionQuery.data) {
+    precinctSelection = getPrecinctSelectionQuery.data.ok();
+  }
```
kshen0, line 44: I'm not sure why the optional chaining operator `?.` isn't getting picked up by the test coverage tool as covered, but breaking this out like so solves the coverage problem.

apps/mark-scan/backend/src/app.ts
arsalansufi: This reasoning makes sense to me!

Re the get method, the main alternative that I've seen is to return `null` when the store returns `undefined`, e.g.:

https://github.com/votingworks/vxsuite/blob/b26ecaf006712bf8caeec943e0bef68913e4c752/apps/central-scan/backend/src/central_scanner_app.ts#L209-L211

That might be preferable if we never expect to return an error `Result` (`err(...)`). But using a `Result` also seems totally fine.

apps/mark-scan/backend/src/store.ts
```diff
@@ -91,6 +95,48 @@ export class Store {
...
+
+    this.client.run(
+      'update election set precinct_selection = ?',
+      precinctSelection ? JSON.stringify(precinctSelection) : null
+    );
+  }
+
+  /**
+   * Gets the current precinct `scan` is accepting ballots for. If set to
+   * `undefined`, ballots from all precincts will be accepted.
```
arsalansufi, line 114: I was under the impression that the precinct selection had to be the special `AllPrecincts` value for ballots from all precincts to be accepted. I thought that `undefined` just indicated that a precinct hadn't been selected yet. I might be missing something!

apps/mark-scan/frontend/src/pages/insert_card_screen.tsx
arsalansufi: Yeah I think this might be because the coverage tool wants you to cover both the cases that the value before the `?.` is defined and not defined. With `if` statements on the other hand, I believe the coverage tool only cares that you meet the condition.

apps/mark-scan/frontend/src/pages/admin_screen.tsx
```diff
@@ -62,6 +58,18 @@ export function AdminScreen({
 }: AdminScreenProps): JSX.Element {
   const { election } = electionDefinition;
   const logOutMutation = logOut.useMutation();
+  const getPrecinctSelectionQuery = getPrecinctSelection.useQuery();
+  const precinctSelection = getPrecinctSelectionQuery.data?.ok();
```
arsalansufi, line 62: Looking at this code, `precinctSelection` can end up `undefined` for one of two reasons: either 1) data hasn't finished loading yet, or 2) a precinct selection has actually not yet been made.

If it's not important to distinguish between these two cases, then this code is fine! But if it is important, you'll often see code like this:

```ts
if (!getPrecinctSelectionQuery.isSuccess) {
  return null; // Basically a data-loading UI state
}
const precinctSelection =  getPrecinctSelectionQuery.data.ok(); // No `?.` necessary
```

https://github.com/votingworks/vxsuite/blob/47369cc56898fe3d6cff2971e853141b92071b58/apps/admin/frontend/src/components/live_check_button.tsx#L13-L16

apps/mark-scan/frontend/src/pages/insert_card_screen.tsx
kshen0: The coverage tool is ok with other files where we use `const precinctSelection = getPrecinctSelectionQuery.data?.ok()` but in this particular file it's not 🤷‍♂️

apps/mark-scan/backend/src/store.ts
kshen0: No I think you're right. I copied this part of the schema from apps/scan, including comment, but an undefined precinct selection might mean something different in `scan`. In `mark-scan` it should mean the same thing as in `mark`, which is how you describe it.

apps/mark-scan/frontend/src/pages/admin_screen.tsx
kshen0: This is a great point, we should return a loading state. I'll update callers of this query. This should also solve the coverage problem above.

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -295,6 +278,8 @@ export function AppRoot({
   const stateMachineState = getStateMachineStateQuery.isSuccess
     ? getStateMachineStateQuery.data
     : 'no_hardware';
+  const getPrecinctSelectionQuery = getPrecinctSelection.useQuery();
+  const precinctSelection = getPrecinctSelectionQuery.data?.ok();
```
kshen0, line 282: We don't want to return `null` in `app_root` because it'll cause the whole app to flicker

apps/mark-scan/backend/src/store.ts
kshen0: Deleted comment on getter: [bef2b8c](https://github.com/votingworks/vxsuite/pull/3767/commits/bef2b8c2333ba9ff8ab7f8e2571f12d74c8e7d79)

apps/mark-scan/frontend/src/pages/admin_screen.tsx
arsalansufi: Thanks for leaving the explanations for cases that couldn't be updated!

arsalansufi (approved):
LGTM!

apps/mark-scan/backend/src/app.ts
jonahkagan: Grout actually doesn't require a defined response, it supports `undefined`. ReactQuery is the culprit here.

Given that, I personally think it makes the most sense to substitute in the `null` value in the `useQuery` hook to make it clearer why it's happening.

I think we're generally trying to use `Result` in API endpoints to signify errors that need to be shown to the user.

apps/mark-scan/frontend/src/pages/poll_worker_screen.tsx
```diff
@@ -181,9 +178,14 @@ export function PollWorkerScreen({
   const getStateMachineStateQuery = getStateMachineState.useQuery();
   const stateMachineState = getStateMachineStateQuery.data;
 
+  const getPrecinctSelectionQuery = getPrecinctSelection.useQuery();
+  const precinctSelection = getPrecinctSelectionQuery.data?.ok();
```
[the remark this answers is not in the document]
jonahkagan: I think this is actually masking a subtle bug. If `getPrecinctSelectionQuery` is being loaded for the first time, on the initial render, its `data` will be `undefined`. That will cause the default value for `selectedCardlessVoterPrecinctId` to be `undefined`. Once the query loads and rerenders, `selectedCardlessVoterPrecinctId` won't be updated, since the default value for the `useState` hook only applies on the first render. This may not happen in practice if the query data is already cached before this component is rendered, but it's still a latent bug waiting to happen.

I recommend always checking `query.isSuccess` before accessing `data` because it's easy for things to behave in weird ways otherwise. In this case, I would either pass in the precinct selection from a parent component or make a wrapper parent component to load the necessary data, putting the `useState` in the child component. It's a bit verbose but I think it's important.

apps/mark-scan/frontend/src/pages/admin_screen.tsx
```diff
@@ -62,6 +58,18 @@ export function AdminScreen({
...
+
+  const setPrecinctSelectionMutation = setPrecinctSelection.useMutation();
+  const updatePrecinctSelection = useCallback(
+    (newPrecinctSelection: PrecinctSelection) => {
+      setPrecinctSelectionMutation.mutate({
+        precinctSelection: newPrecinctSelection,
+      });
+    },
+    [setPrecinctSelectionMutation]
```
jonahkagan, line 69: Mutation objects aren't memoized, so I'm pretty sure this `useCallback` is unnecessary because it will recreate the function on every render anyway.

apps/mark-scan/frontend/src/app_root.tsx
jonahkagan: I'm surprised that checking for `isSuccess` here would cause flickering. My expectation would be that it would show the loading state on the first load, and then cache the precinct selection for all future renders.

Also, I think there might be an issue with the check involving `precinctSelection` on L674:

```ts
 if (optionalElectionDefinition && precinctSelection) {
```
Before `getPrecinctSelectionQuery` is loaded, `precinctSelection` will be `undefined`, so this entire case will be skipped and we'll show the unconfigured screen. Then, once it loads, if there was a saved precinct selection, we'll switch to this case, which would change to a different screen.

kshen0: My assumption was returning `null` from `app_root` would cause a blank screen to render, so better to prevent the render of a single subcomponent. But actually I see we already return `null` if there are other queries that aren't yet successful. Is there a loading page higher up the component tree?

jonahkagan: My approach has been to use `null` for loading states since the backend is local and the data we're loading are mostly small. If we noticed a case where there was a significantly long blank screen shown to the user, then I think we could consider adding an explicit loading screen. I've found that showing a very quick loading screen feels like a flicker.

apps/mark-scan/frontend/src/pages/poll_worker_screen.tsx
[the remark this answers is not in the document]
kshen0: Thanks for flagging this bug. https://github.com/votingworks/vxsuite/pull/3786 returns `PollWorkerScreen` to its original contract of requiring `precinctSelection` to be passed as a non-optional prop.

apps/mark-scan/frontend/src/pages/admin_screen.tsx
kshen0: Updated in https://github.com/votingworks/vxsuite/pull/3786/files#diff-eba06bf39f2aae2ca6cf5b42615541312f20619de8c93b109f2eaa45a134fff5L63

PR #3771, opened 2023-07-28
chore(test): Upgrade Jest to v27 Minimum

adghayes:
## Overview

Our `jest` versions are all over the place which isn't a problem in itself, but increases overall dependencies and can create confusion. In this PR I upgrade `jest` to v27 anywhere it was lower than that along with closely related packages. 

The most significant change is to how timers work. `jest` has replaced the old timers implementation with a "modern" implementation. It works almost exactly the same, except now `MockDate` doesn't work because the functionality is baked into `jest` itself. In this PR I also strip out `mockdate` and `jest-date-mock` which should not be necessary moving forward. 

I then do some miscellaneous cleanup in the last 3 commits.

At some point in a future PR, we can upgrade Jest to v28 minimum and then v29 minimum. To v28 _should_ require almost no changes, and v29 should require few changes outside of snapshot formatting changes. Once things are standardized, we could lock down `jest` versions at the root of the project for consistency. 

We _also_ need to move `jest` and it's related packages to `devDependencies` because right now some are in dependencies and it's increasing production / trusted build package size.


adghayes:
I just let reviewers automatically be tagged based on file ownership. If you do intend to review this PR, please drop a comment so others know they don't have to.

apps/mark-scan/frontend/src/setupTests.tsx
```diff
@@ -72,3 +73,5 @@ afterEach(() => {
 
 globalThis.TextDecoder = TextDecoder as typeof globalThis.TextDecoder;
 globalThis.TextEncoder = TextEncoder as typeof globalThis.TextEncoder;
+
+global.setImmediate = jest.useRealTimers as unknown as typeof setImmediate;
```
adghayes: whoops, this is wrong. removed in a later commit

apps/admin/frontend/src/__snapshots__/app.test.tsx.snap
```diff
@@ -215,7 +215,7 @@ exports[`L&A (logic and accuracy) flow 1`] = `
         <span
           class="c3"
         >
-          This report was created on Tuesday, November 3, 2020 at 10:22:00 PM UTC.
+          This report was created on Tuesday, November 3, 2020 at 10:22:01 PM UTC.
```
adghayes, line 218: Previously `mockdate` and `jest` where operating out of sync, so the time _should_ have been the new time (based on initial time and the `sleep` times, but it wasn't). Now they operate together.

apps/admin/frontend/package.json
```diff
@@ -19,9 +19,9 @@
     "stylelint:run": "stylelint 'src/**/*.{js,jsx,ts,tsx}' && stylelint 'src/**/*.css' --config .stylelintrc-css.js",
     "stylelint:run:fix": "stylelint 'src/**/*.{js,jsx,ts,tsx}' --fix && stylelint 'src/**/*.css' --config .stylelintrc-css.js --fix",
     "test": "is-ci test:ci test:watch",
-    "test:ci": "TZ=UTC node scripts/test.js --coverage --watchAll=false --reporters=default --reporters=jest-junit --maxWorkers=6 --env=jest-environment-jsdom",
-    "test:coverage": "TZ=UTC node scripts/test.js --coverage --watchAll=false --env=jest-environment-jsdom",
-    "test:watch": "TZ=UTC node scripts/test.js --env=jest-environment-jsdom",
```
adghayes, line 24: this is set in the `jest` config file, doesn't need to be here

jonahkagan (approved):
Nice!

I reviewed the whole thing, not just my codeowner parts

PR #3772, opened 2023-07-29
chore(test): Upgrade Jest to v28 Minimum

adghayes:
Upgrades `jest` and close dependencies to v28 everywhere. Relatively few changes. 

Something mysteriously broke with importing `@votingworks/test-utils` into `admin/integration-testing` due to the `jest-diff` upgrade (I think), but it only broke for tests which we are skipping anyway. I just made sure to skip the entire file, because those tests need to be rewritten anyway at some point. Currently it's always the longest running test suite, largely because it's loading a bunch of files for 3 end-to-end tests it never actually runs.

One test failure still, discussed [here](https://github.com/votingworks/vxsuite/pull/3750#pullrequestreview-1553442188), that will have to be fixed before merging, but I don't think it blocks review.


adghayes:
ignore the "ready to review" status, still need to rebase but away from computer. sorry!

adghayes:
OK, now it's rebased and actually ready to review @jonahkagan @arsalansufi @carolinemodic @eventualbuddha @kshen0, the automatically tagged folks. No urgency on this ofc but again, if you start an overall review, please drop a note so others know they do not have to.

jonahkagan:
Reviewing

jonahkagan (approved):

PR #3774, opened 2023-08-01
chore(test): Upgrade Jest to v29

adghayes:
## Overview

Upgrades all `jest` packages to v29. I don't do it in this PR, but in a subsequent PR, I'd like to move the universal `jest` dependencies to the root `package.json` to avoid unintentionally out-of-sync versions in the future.


arsalansufi (approved):
LGTM!

PR #3779, opened 2023-08-01
ERR Hand-Crafted Sample

adghayes:
## Overview

I made a very miniature version of an ERR file for the minimal exhaustive animal election, only writing out one contest. This is just to give myself and others a sense of the format. 

The format doesn't contain any big surprises. I mainly wanted to see how the format allows specifying vote totals across splits. In a custom CSV we export, we can create a column for every split (precinct, ballot, etc.) that we support. In ERR, there are three splits to differentiate ballot counts and vote counts:
- `CountItemType`: this is equivalent to our "voting method" and differentiates between absentee, election-day, early voting, and more
- `DeviceClass`: this allows you to specify the kind of device that tabulated the result e.g. precinct vs. central scanner. we don't have an equivalent filter in our system nowadays, but we may someday
- `GpUnit`: but beware, this `GpUnit` looks a little different than in the other specs, read below

Our type definition makes a little clearer what `GpUnit` actually means in this spec:
```
readonly GpUnit?: ReadonlyArray<ReportingDevice | ReportingUnit>;
```
It's actually much more flexible than a "geopolitical unit" - you can use `ReportingDevice` to specify a scanner and use the `ReportingUnit` type to specify a batch, split precinct, ballot style other, and specific categories of geopolitical units. You could also have units of type `other` type and do arbitrary splits. 

So, as far as supporting ERR splits, we have a few options:
1. For MCVP, only export ERR in one supported format, not part of the query editor
2. Allow some splits in ERR - only those easily and clearly in the spec. This would mean we allow splitting on voting method + 1 additional field.
3. Allow arbitrary splits in ERR, using the "Other" type to describe any combined split
4. Allow arbitrary splits in ERR, updating our version of ERR to allow other specifiers

I think 3 & 4 are bad ideas. 2 is probably where we want to be, but for MCVP I suggest we just do 1.

And for interfacing with VxDesign, I would prefer just a `.csv` format we control.


libs/types/src/cdf/election-results-reporting/sample.ts
```diff
@@ -0,0 +1,305 @@
...
+      '@id': 'precinct-1',
+      '@type': 'ElectionResults.ReportingUnit',
+      Type: ReportingUnitType.Precinct,
+      ComposingGpUnitIds: ['precinct-1-1M', 'precinct-1-2F'],
+    },
+    {
+      '@id': 'precinct-1-1M',
+      '@type': 'ElectionResults.ReportingUnit',
+      Type: ReportingUnitType.SplitPrecinct,
+    },
```
adghayes, line 37: I defined these, and we could do vote totals at this granularity, but did not do so below.

PR #3781, opened 2023-08-02
chore(lint): public-hoist all eslint packages

adghayes:
I'd like to make our dependencies easier to understand. Currently, all of our package.json files reference a bunch of eslint configs and plugins that they don't explicitly use in their eslint configs. Why? Because they are referenced in eslint-plugin-vx and eslint is not able to resolve the packages if they're not in a flat node module structure (like npm). To replicate this in pnpm, we "publicly hoist" all eslint packages. This is the default config that pnpm now ships with. All publicly hoisted packages exist in a flat node modules directory at the root of the repo.

I hope this helps with two things:
- make the package.json more informative by only including explicitly referenced packages
- avoid drifting versions of eslint. there were places in the repo where we had upgraded plugin dependencies of eslint-plugin-vx but hadn't upgraded those plugins in other packages, meaning linting was not getting bug fixes / improvements in those packages as intende

## Testing Plan

I tested with `pnpm lint` and VSCode linting that rules were still being applied as expected.


.npmrc
```diff
@@ -4,9 +4,12 @@
 # it might get the wrong version of.
 hoist = false
 
-# We _do_ try to hoist `@types` packages so that TS will not end up with type
-# errors from comparing types from different versions of the same package.
-public-hoist-pattern=['@types/*']
```
adghayes, line 9: I think due to incorrect format, this `public-hoist-pattern` was not doing anything.

apps/admin/backend/package.json
```diff
@@ -99,8 +91,7 @@
     "nodemon": "^2.0.20",
     "sort-package-json": "^1.50.0",
     "supertest": "^6.0.1",
-    "ts-jest": "^29.1.1",
-    "typescript": "4.6.3"
```
adghayes, line 103: I also moved `typescript` to the workspace root where possible, as it's universal to the repo.

apps/admin/frontend/package.json
```diff
@@ -153,25 +152,11 @@
...
-    "eslint-config-airbnb": "^19.0.4",
-    "eslint-config-prettier": "^8.5.0",
-    "eslint-config-react-app": "^7.0.1",
-    "eslint-import-resolver-node": "^0.3.4",
-    "eslint-plugin-cypress": "^2.12.1",
-    "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-jest": "^27.2.3",
-    "eslint-plugin-jsx-a11y": "^6.6.1",
-    "eslint-plugin-react": "^7.31.8",
-    "eslint-plugin-react-hooks": "^4.6.0",
```
adghayes, line 173: `react-hook`s was no longer necessary because the rules are bundled with `airbnb`

apps/central-scan/frontend/package.json
```diff
@@ -47,9 +47,6 @@
       "last 1 safari version"
     ]
   },
-  "eslintConfig": {
-    "extends": "react-app"
-  },
```
adghayes, line 52: this setting in `package.json` is lower in priority than `.eslintrc` so it, and the `eslint-config-react-app` package, were not doing anything and could be removed.

eventualbuddha (approved):
👏

PR #3784, opened 2023-08-02
[libs/ui] Fix BmdPaperBallot printout styling

kofi-q:
## Overview

With the new UI themes enabled, most of the `Prose` component's text-specific styling gets stripped - this breaks the styling of the `BmdPaperBallot` a bit, so updating the layout and styling, using the newer typography components, to more closely match the pre-themes styling.

Also adding a `VxThemeProvider` wrapper around the `styled-components/ThemeProvider` to simplify theme installation, since `BmdPaperBallot` now relies on themed components.

### New Styling:
Screenshot 2023-08-02 at 12 26 58

### Pre-Themes Styling:
Screenshot 2023-08-02 at 12 26 29

### Existing Post-Themes Styling:
Screenshot 2023-08-02 at 12 39 40

## Testing Plan
- Manual testing in VxMark (ballot printing) and VxAdmin (L&A packages) to verify styling
- Updated affected tests/snapshots

## Checklist

### DIFF
--- a/apps/admin/frontend/src/__snapshots__/app.test.tsx.snap
+++ b/apps/admin/frontend/src/__snapshots__/app.test.tsx.snap
@@ -1080,28 +1080,97 @@ House Bill 1796 - Flag Referendum
 `;
 
 exports[`L&A (logic and accuracy) flow 2`] = `
-.c11 {
-  word-break: break-word;
-}
-
-.c12 {
+.c15 {
+  font-size: 1em;
   font-weight: 600;
+  line-height: 1.3;
+  margin: 0;
 }
 
-.c13 {
-  font-weight: 600;
-  word-break: break-word;
-}
-
-.c6 {
-  white-space: nowrap;
+.c5 {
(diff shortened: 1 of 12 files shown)


adghayes (approved):
code-wise LGTM

I think the updated ballot looks great, but I want to check in with @benadida @carolinemodic @mattroe to confirm there aren't any size concerns. We've specifically aimed for compactness in the past because we don't support a second page on a BMD ballot. I don't think we have a specific metric or election definition we try to fit, though, so I don't think we can worry about it.

mattroe:
Thanks for flagging @adghayes. The difference in compactness between `New Styling` and `Pre-Themes Styling` doesn't look that significant to me. `Existing Post-Themes Styling` is definitely unnecessarily spaced.

We're eventually going to have to solve the single-page BMD problem, but I think we can keep that out of scope for now and run with `New Styling.

kofi-q:
Sounds good - thanks!

PR #3786, opened 2023-08-03
Safer react query handling

kshen0:
## Overview
* Always checks for `query.isSuccess` when querying precinct selection
* Returns null instead of an empty `Result`

## Demo Video or Screenshot
No behavior change

## Testing Plan
Updated existing tests

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/pages/admin_screen.tsx
+++ b/apps/mark-scan/frontend/src/pages/admin_screen.tsx
@@ -28,7 +28,6 @@ import {
 import { makeAsync } from '@votingworks/utils';
 import { Logger } from '@votingworks/logging';
 import type { MachineConfig } from '@votingworks/mark-scan-backend';
-import { Optional } from '@votingworks/basics';
 import { ScreenReader } from '../config/types';
 import { getPrecinctSelection, logOut, setPrecinctSelection } from '../api';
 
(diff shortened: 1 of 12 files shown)


apps/mark-scan/frontend/src/pages/admin_screen.tsx
```diff
@@ -79,7 +77,9 @@ export function AdminScreen({
   if (!getPrecinctSelectionQuery.isSuccess) {
     return null;
   }
-  const precinctSelection = getPrecinctSelectionQuery.data.ok();
+
+  const precinctSelection: Optional<PrecinctSelection> =
+    getPrecinctSelectionQuery.data || undefined;
```
kshen0, line 82: This gets passed to `ElectionInfoBar` during render. That component's props allow `undefined` `PrecinctSelection` but not `null`. It's probably safe to update that component to accept `precinctSelection?: PrecinctSelection | null` but it seemed lower impact + very explicit to do it this way.

jonahkagan: Agreed.

I would even consider isolating the conversion between `null` and `undefined` to the query handler to keep it as localized as possible, since it's really a react-query quirk.

E.g. change the backend method to return `undefined`, then:
```ts
  useQuery() {
    return useQuery(
      this.queryKey(),
      // Since query functions are not allowed to return undefined, convert to null
      async () => (await apiClient.getPrecinctSelection()) ?? null,
      // Convert back to undefined when reading the query results
      { select: (precinctSelection) => precinctSelection ?? undefined }
    );
  },
```
I don't know if this will actually work.

jonahkagan (approved):

PR #3787, opened 2023-08-03
ci(chore): only setup necessary dependencies for each test

adghayes:
__"review" as a whole, not commit-by-commit, needs to be rebased, etc.__

## Overview

When doing package upgrades lately, I've been waiting on a lot of tests. I noticed we're installing dependencies for the whole repo (including `ballot-interpreter-nh`) for every single test suite, even if it's very lowly on the dependency tree. I experimented with installing only necessary dependencies.

It does seem to improve test running. In terms of CircleCI credits:

| Dependency Approach | Lockfile Change? | Credit Cost (sample size = 1) |
| ------------- | ------------- | ---------- |
| All Dependencies | Yes | 10,233  |
| Selected Dependencies  | Yes  |  5,941 |
| All Dependencies | No | 5,371  |
| Selected Dependencies  | No  |  4,425 |

I'm too lazy to do an analysis of the test durations (and CircleCi has limited export formats) but a quick glance suggests that it parallels the credit differences, with greater proportional differences for lowlier test suites.

This could make tests harder to debug for weird dependency issues, but maybe also will help us be better about our dependency trees. What do you think?


apps/admin/backend/Makefile
```diff
@@ -8,7 +8,7 @@ install:
 	sudo apt -y install libpcsclite1 libpcsclite-dev
 
 build: FORCE
-	pnpm install && pnpm build
+	pnpm install --filter {.}... --filter {.}... --filter {.}... --filter {.}... --filter {.}... --filter {.}... --filter {.}... && pnpm build
```
adghayes, line 11: typo

apps/scan/backend/tsconfig.build.json
```diff
@@ -10,7 +10,6 @@
     "declarationMap": true
   },
   "references": [
-    { "path": "../../../libs/api/tsconfig.build.json" },
```
adghayes, line 13: places where we reference build dependencies that are no longer _actual_ dependencies break during typescript build, because it can't find the relevant node module

libs/monorepo-utils/package.json
```diff
@@ -37,15 +37,7 @@
...
     "@types/node": "16.18.23",
-    "@typescript-eslint/eslint-plugin": "5.37.0",
-    "@typescript-eslint/parser": "5.37.0",
     "esbuild-runner": "^2.2.2",
-    "eslint": "8.23.1",
-    "eslint-config-prettier": "^8.5.0",
-    "eslint-import-resolver-node": "^0.3.6",
-    "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-jest": "^27.2.3",
-    "eslint-plugin-prettier": "^4.2.1",
```
adghayes, line 48: just cleanup

libs/monorepo-utils/src/circleci.ts
```diff
@@ -155,6 +157,10 @@ ${jobIds.map((jobId) => `      - ${jobId}`).join('\n')}
 commands:
   checkout-and-install:
     description: Get the code and install dependencies.
+    parameters:
+      path:
+        type: string
+        default: '.'
```
adghayes, line 163: We could edit the parameterization to default to installing all packages for the repo instead of just root packages.

PR #3788, opened 2023-08-03
Flexible Results CSV Backend

adghayes:
## Overview

We already support CSV exports, but of a specific format that splits results by precinct and voting method. In preparation for more flexible reporting on the frontend, we need to create CSV's based on various splits and filters.

I [wrote up](https://docs.google.com/document/d/1mgcHpOrGQarxNPuOyTGf6LLHRaeCMLA7Od-12MR7HGc/edit#heading=h.yfj5jq5kfm5l) some thoughts on combining splits and filters. It's unfortunately complicated to do. There are a few things to figure out:

1. We calculate grouped totals opportunistically, meaning we add new groups as we encounter them. For splits, we need to answer the general question - what splits do we expect? For example if you split on precinct and ballot style, you'd only expect splits for precinct x ballot style combinations that are actually possible. This PR adds utilities for determining what splits to expect.

2. Different filters / splits affect what contests can appear in results. If we're filtering on "Mammal Party," it doesn't make sense to include "Fish Party" contests in those results. This PR adds utilities for determining what contests to include. We had something previously, "getRelevantContests," but I wanted something more generalized.

I tried to be exhaustive for the splits Voting Method, Party, Ballot Style, and Precinct. I punt on determining correct splits for scanner or batch splits. 

## Testing Plan
- Lots of new automated testing
- Manually tested the existing results export


libs/utils/src/tabulation_helpers.ts
adghayes, line 1: I can understand some apprehension about this caching pattern, but I think that it's safe. We identify the cache based on election hashes, which would change if the election changed. And the caches are scoped to the higher-order functions, and cannot be accessed elsewhere.

Generally speaking, I want data structures that contain only IDs of objects rather than entire objects. Having entire objects (like how we used to have the entire `Contest` definitions in contest results) makes mocking difficult, serialization potentially long, and forces us to construct data that might never be consumed. I want to look up needed information at the point of consuming (like exporting a file) and do it in a way that's reasonably ergonomic and O(1).

libs/utils/src/tabulation/contest_filtering.ts
```diff
@@ -0,0 +1,260 @@
...
+ * are no ballot styles in a report, it's not a valid report.
+ */
+export function getContestIdsForFilter(
+  electionDefinition: ElectionDefinition,
+  filter?: Tabulation.Filter
+): Set<ContestId> {
+  return getContestIdsForBallotStyleIds(electionDefinition, [
+    ...getBallotStyleIdsForFilter(electionDefinition, filter),
+  ]);
+}
```
adghayes, line 196: The idea here is that a consumer, handling a report that is both filtered and split, could `getContestIdsForFilter` once and then `getContestIdsForSplit` for each split and intersect the two sets to get the actual contest IDs.

libs/utils/test/fixtures.ts
```diff
@@ -0,0 +1,139 @@
+import { electionMinimalExhaustiveSampleDefinition } from '@votingworks/fixtures';
+import {
+  AnyContest,
+  DistrictId,
+  ElectionDefinition,
+  PartyId,
+} from '@votingworks/types';
+
+function contestFactory(
```
kofi-q, line 9: nit: usually find `doSomething()` method names more readable - maybe `generateContest()` or `buildContest()` here?

kofi-q (approved):
Couple optional nits, but logic looks good and well-tested! A little dense, but I didn't have any obvious lightbulbs go off on how to simplify, so possibly not avoidable

adghayes:
@kofi-q my main concern with this PR is the density, but I think that iterating on this over the next couple months and adding more requirements will force it to smooth out. Already thinking of ways to determine which splits to expect in a more readable way.

PR #3789, opened 2023-08-03
Require Election.ballotLayout

jonahkagan:
## Overview

We're relying on this field more and more in VxDesign and the interpreter, so we should make it a required part of the election definition rather than setting independent defaults in various locations where we try to use it.

_Review by commit_

## Demo Video or Screenshot
N/A

## Testing Plan
Updated all test fixtures.

## Checklist

### DIFF
--- a/libs/types/src/election_utils.ts
+++ b/libs/types/src/election_utils.ts
@@ -6,6 +6,7 @@ import { safeParseCdfBallotDefinition } from './cdf/ballot-definition/convert';
 import {
   AdjudicationReason,
   AnyContest,
+  BallotPaperSize,
   BallotStyle,
   BallotStyleId,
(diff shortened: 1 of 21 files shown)


libs/types/src/election_utils.ts
```diff
@@ -522,6 +523,13 @@ function preprocessElection(value: unknown): unknown {
     }
   }
 
+  if (!('ballotLayout' in election)) {
+    election = {
+      ...election,
+      ballotLayout: { paperSize: BallotPaperSize.Letter },
+    };
+  }
```
jonahkagan, line 531: This supports backwards compatibility with old election defs

adghayes (approved):

PR #3790, opened 2023-08-04
fix(lint): reverse stylelint media query notation rule & add comments to config

adghayes:
All modals in `vxsuite` are currently appearing fullscreen. This is because the new `stylelint` config applied a new syntax for `@media` queries that isn't supported until Chromium 104, but our version of `electron` is many versions behind that.

### DIFF
--- a/stylelint.shared.js
+++ b/stylelint.shared.js
@@ -1,8 +1,14 @@
+// Configuration for our CSS linting with stylelint
 module.exports = {
   extends: ['stylelint-config-standard'],
   customSyntax: 'postcss-styled-syntax',
   rules: {
+    // We use color functions (e.g. rgb) in nested JS that, being JS, need comma-separated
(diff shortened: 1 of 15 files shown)


stylelint.shared.js
```diff
@@ -1,14 +1,26 @@
+// Configuration for our CSS linting with stylelint
 module.exports = {
   extends: ['stylelint-config-standard'],
   customSyntax: 'postcss-styled-syntax',
   rules: {
+    // We use color functions (e.g. rgb) in nested JS that, being JS, need comma-separated
+    // parameters. "Modern" notation in CSS is space-separated parameters. "Legacy"
+    // notation enforces comma-separated parameters.
```
arsalansufi, line 8: Thanks for adding these comments!

arsalansufi (approved):

PR #3793, opened 2023-08-06
Upgrade React 18

### LINKED ISSUE #1660: Upgrade to React 18

### LINKED ISSUE #2798: Testing Warnings in Election Manager due to React Query
Since we've started using `ReactQuery` in `election-manager`, we get all sorts of console warnings during tests because `ReactQuery` is trying to do state updates after unmounting. It looks like others have had this issue: https://github.com/TanStack/query/issues/432. It would make the tests a lot more readable if we could get this sorted out.

adghayes:
## Overview

Closes #1660. Closes #2798. 

## Checklist


apps/admin/frontend/package.json
```diff
@@ -125,10 +125,9 @@
   },
   "devDependencies": {
     "@jest/types": "^29.6.1",
-    "@testing-library/jest-dom": "^5.16.4",
-    "@testing-library/react": "^12.1.5",
-    "@testing-library/react-hooks": "^8.0.1",
```
adghayes, line 130: `@testing-library/react-hooks` has been absorbed by `@testing-library/react`, and you'll see it being replaced in tests in subsequent commits.

apps/admin/frontend/package.json
```diff
@@ -125,10 +125,9 @@
...
-    "@testing-library/user-event": "^12.6.0",
+    "@testing-library/jest-dom": "^5.17.0",
+    "@testing-library/react": "^14.0.0",
+    "@testing-library/user-event": "^13.5.0",
```
adghayes, line 130: We have to upgrade `@testing-library/react` two major versions to v14 for React 18 compatibility, and it's not backward compatible so it had to happen at the same time.

`@testing-library/user-event` only needs to go to v13, though, which is good because v14 introduces a ton of breaking changes.

package.json
```diff
@@ -53,5 +53,11 @@
     "postcss-styled-syntax": "^0.4.0",
     "stylelint": "^15.10.2",
     "stylelint-config-standard": "^34.0.0"
+  },
+  "pnpm": {
+    "overrides": {
+      "@types/eslint": "8.4.1",
+      "typescript": "4.6.3"
+    }
```
adghayes, line 61: The package upgrades created some linting chaos. Specifically, some dependency started pulling a different conflicting version of `@types/eslint` and another dependency started pulling typescript 5, both of which created type collisions that made linting fail. One way to deal with this is with pnpm package overrides like this, which force all versions of a package in the dependency array to have the same version.

libs/ui/src/export_logs_modal.test.tsx
```diff
@@ -65,7 +67,7 @@ test('renders loading screen when usb drive is mounting or ejecting in export mo
     );
     userEvent.click(screen.getByText('Save Log File'));
     await screen.findByText('Loading');
-    userEvent.click(screen.getByText('Cancel'));
+    userEvent.click(await screen.findByText('Cancel'));
```
adghayes, line 70: Since React 18 is about concurrency and automatic batching of state updates, it means that some things are faster or slower in tests. The largest category of test tweaks is around properly waiting for things on the page _or_ finding workarounds to test things that are now too fast to catch on screen by default.

libs/ui/src/export_logs_modal.test.tsx
```diff
@@ -168,7 +169,9 @@ test('successful save raw log flow', async () => {
   userEvent.click(screen.getByText('Save'));
   await screen.findByText(/Saving Logs/);
   expect(mockKiosk.readFile).toHaveBeenCalled();
-  jest.advanceTimersByTime(2001);
+  act(() => {
+    jest.advanceTimersByTime(2001);
+  });
```
adghayes, line 174: In some places, the upgrade forces us to have proper `act` warnings. In others, previously necessary `act` wraps are no longer necessary. In general, I wouldn't say it's become more or less important, but React & RTL seem to be more consistent and logical about where they're required.

libs/ui/src/hooks/use_usb_drive.test.tsx
```diff
@@ -53,16 +55,19 @@ test('full lifecycle with USBControllerButton', async () => {
 
   const mockKiosk = fakeKiosk();
   mockKiosk.getUsbDriveInfo.mockResolvedValue([]);
+  const { promise: mountPromise, resolve: mountResolve } = deferred<void>();
+  mockKiosk.mountUsbDrive.mockReturnValueOnce(mountPromise);
```
adghayes, line 59: Now that state updates within `useEffect`s are batched, some loading states never actually appear on screen during tests because, without actual things to wait on, components just go from `setState('loading')` to `setState('loaded')`. When we actually want to capture these loading states it may be necessary to mock unresolved promises like so.

libs/ui/src/index.ts
```diff
@@ -25,7 +25,6 @@ export * from './hooks/use_mounted_state';
 export * from './hooks/use_now';
 export * from './hooks/use_previous';
 export * from './hooks/use_screen_info';
-export * from './hooks/use_stored_state';
```
adghayes, line 28: Not used anywhere and we don't want to use stored state in the frontend anyway.

libs/ui/src/print_element.test.tsx
```diff
@@ -29,110 +30,110 @@ const fakeOptions: PrintOptions = { sides: 'one-sided' };
...
-    await waitFor(async () => {
-      const printContainer = await screen.findByTestId('print-root');
-      within(printContainer).getByText('Print me!');
+    await suppressingConsoleOutput(async () => {
+      await printElement(simpleElement, fakeOptions);
+      expect(printer.print).toHaveBeenCalledTimes(1);
+      expect(printer.print).toHaveBeenLastCalledWith(
+        expect.objectContaining(fakeOptions)
+      );
     });
```
adghayes, line 39: RTL really doesn't like how we're testing this `printElement` component so I had to wrap all tests in this file with `suppressingConsoleOutput`. I don't like it and I tried to find lots of other ways to do this, but the `act` warnings are brutal and if wrap in asynchronous `act` calls the tests just don't complete.

libs/ui/src/print_element.tsx
```diff
@@ -58,9 +59,8 @@ async function printElementWhenReadyHelper<PrintResult>(
       await printAndTeardown();
     }
 
-    ReactDom.render(
-      <PrintOnly>{elementWithOnReadyCallback(onElementReady)}</PrintOnly>,
-      printRoot
+    reactPrintRoot.render(
+      <PrintOnly>{elementWithOnReadyCallback(onElementReady)}</PrintOnly>
```
adghayes, line 63: The main change to non-test files is this change in how `react` is rendered at roots.

apps/scan/frontend/src/screens/poll_worker_screen.test.tsx
```diff
@@ -233,7 +232,6 @@ describe('transitions from polls paused', () => {
     apiMock.expectSetPollsState('polls_open');
     apiMock.expectGetConfig({ pollsState: 'polls_open' });
     userEvent.click(screen.getByText('Yes, Resume Voting'));
-    await screen.findByText('Resuming Voting…');
```
adghayes, line 236: This is among a few cases where loading states are no longer visible in tests without setting up some sort of `deferred` promise, and it just didn't seem worth the trouble.

apps/admin/frontend/src/screens/print_test_deck_screen.tsx
```diff
@@ -525,6 +523,18 @@ export function PrintTestDeckScreen(): JSX.Element {
         return handMarkedPaperBallotsWithCallback;
       });
 
+      const allTallyReportResults: Record<
```
adghayes, line 526: This method was previously returning, an asynchronous React component? I'm not sure, but linting didn't like it.

apps/mark/frontend/src/pages/accessible_controller_diagnostic_screen.test.tsx
```diff
@@ -29,12 +29,14 @@ function renderScreen(
...
+jest.useFakeTimers();
+
 describe('Accessible Controller Diagnostic Screen', () => {
   beforeEach(() => {
-    jest.useFakeTimers().setSystemTime(new Date(now.toISO()));
+    jest.setSystemTime(new Date(now.toISO()));
   });
 
-  it('yields a success result when all steps are completed', async () => {
+  it.skip('yields a success result when all steps are completed', async () => {
```
adghayes, line 39: The two tests for the successful diagnostics flow are failing. In the diagnostics flow, you press a button to play the sound, and then advance once the sound is played. You cannot confirm the test passed until the sound is played. So `hasSoundPlayed` is state of that component. [See the useEffect here.](https://github.com/votingworks/vxsuite/blob/3ee24946c1b0271928b6787dabf327a45fc1f237/apps/mark/frontend/src/pages/accessible_controller_diagnostic_screen.tsx#L183-202:)

In testing, however, the state change `setHasAudioPlayed` never triggers the `useEffect` to run again so, when the `{Enter}` button is pressed, the test doesn't advance to passed. I tested in the app itself, and this works, so it seems specific to the testing environment. If you add some element of the render that is dependent on `hasPlayedAudio`, that forces a re-render, which does run the `useEffect`. But state changes should still trigger hooks to re-run even if there's no change on the page? I am confused by this - not sure if it's a `react` issue, an RTL issue, or expected behavior.

In any case, I can fix this in one of two ways:
1. Changing the screen to display something differently after audio has played. Maybe the "Sound is Not Working" button should be disabled until the sound has played? Or maybe we should have a check?
2. Explicitly calling `rerender` in the tests that this affects after the sound plays.

apps/mark/frontend/src/app_quit_on_idle.test.tsx
```diff
@@ -59,20 +59,19 @@ test('Insert Card screen idle timeout to quit app', async () => {
...
-  await advanceTimersAndPromises();
   await advanceTimersAndPromises(QUIT_KIOSK_IDLE_SECONDS);
-  expect(window.kiosk?.quit).toHaveBeenCalledTimes(1);
+  await waitFor(() => {
+    expect(window.kiosk?.quit).toHaveBeenCalledTimes(1);
+  });
 });
 
-test('Voter idle timeout', async () => {
+test.skip('Voter idle timeout', async () => {
```
adghayes: This is the thorniest test failure of the them all. It appears that once the idle timeout passes, this `useEffect` in the shared `IdlePage` component runs many, many times: https://github.com/votingworks/vxsuite/blob/3ee24946c1b0271928b6787dabf327a45fc1f237/libs/mark-flow-ui/src/pages/idle_page.tsx#L37-L42
Setting `isLoading` to `true` _should_ prevent the effect from re-running, but for some reason that state change doesn't propagate fast enough.

I tried this out in the app, in `kiosk-browser`, and found that it worked as expected. But I also noticed that the tree was re-rendering a ton, and traced it to this callback: https://github.com/votingworks/vxsuite/blob/3ee24946c1b0271928b6787dabf327a45fc1f237/apps/mark/frontend/src/app_root.tsx#L520-L526
This is similar to we discussed here: https://github.com/votingworks/vxsuite/pull/3191#issuecomment-1502308643. Mutations returned from `useMutation` actually change on every render. And for some reason the app is re-rendering a lot. What I think is happening (due to React 18 concurrency and batching) is that the mutation within the `IdlePage` `useEffect` is triggering a circular issue where the mutation changes, the dependencies change, and the `useEffect` runs again before batched state changes take effect, so `isLoading` is still false.

One solution may be to have a ref for `isLoading`. In a subsequent PR, I fix by removing the mutation as a dependency of the callback, instead using the `mutate` function of the mutation.

Thoughts?

apps/mark/frontend/src/pages/idle_page.tsx
```diff
@@ -1,14 +1,19 @@
...
 
-  async function reset() {
-    await endVoterSession();
-    resetBallot();
-  }
+  const reset = useCallback(() => {
+    endCardlessVoterSessionMutate(undefined, {
+      onSuccess: () => resetBallot(),
+    });
+  }, [endCardlessVoterSessionMutate, resetBallot]);
```
adghayes, line 16: The `mutate` method has a consistent identity so doesn't cause runaway `useEffect`s like were apparently botching that test. 

I think we're still using mutations as dependencies in a lot of places. Annoyingly, the lint rule doesn't let you specify properties in the dependency array, some discussion here: https://github.com/facebook/react/issues/16265. That lint rule is otherwise very helpful, so we don't want to disable it often. Which is a shame, because I think passing `someMutation.mutate` to the dependency array is more idiomatic than what I do above.

Thoughts?

package.json
jonahkagan: Nice, glad this was easy to fix

libs/ui/src/export_logs_modal.test.tsx
jonahkagan: I think this particular example was only working by chance before.

libs/ui/src/hooks/use_devices.test.ts
```diff
@@ -120,13 +120,13 @@ test('can connect printer as expected', async () => {
   );
 
   // Prevent `act` warning.
```
jonahkagan, line 122: Is there still an act warning to prevent here?

libs/ui/src/hooks/use_devices.test.ts
```diff
@@ -120,13 +120,13 @@ test('can connect printer as expected', async () => {
...
-  await waitForNextUpdate();
```
jonahkagan, line 123: Is there still an act warning to prevent here?

libs/ui/src/hooks/use_devices.test.ts
```diff
@@ -365,59 +365,63 @@ test('can handle logs for a random device as expected', async () => {
...
   jest.useFakeTimers();
   const hardware = new MemoryHardware();
   const logger = fakeLogger();
-  const { result, waitForNextUpdate } = renderHook(() =>
+  const { result, unmount } = renderHook(() =>
     useDevices({ hardware, logger })
   );
   expect(result.current).toEqual(emptyDevices);
 
-  await waitForNextUpdate();
```
jonahkagan, line 380: Does `waitForNextUpdate` not work anymore?

libs/ui/src/print_element.test.tsx
jonahkagan: Happy to investigate this more with you or just leave as is

apps/central-scan/frontend/src/components/delete_batch_modal.tsx
```diff
@@ -29,9 +29,9 @@ export function DeleteBatchModal({
       content={
         <React.Fragment>
           <P>This action cannot be undone.</P>
-          {deleteBatchMutation.error && (
-            <P color="danger">{`${deleteBatchMutation.error}`}</P>
-          )}
+          {deleteBatchMutation.error ? (
+            <P color="danger">{deleteBatchMutation.error as string}</P>
+          ) : null}
```
jonahkagan, line 34: Why this change?

apps/scan/frontend/src/app.test.tsx
```diff
@@ -153,11 +160,14 @@ test('app can load and configure from a usb stick', async () => {
 
   // Insert a USB with a ballot package
   apiMock.expectGetUsbDriveStatus('mounted');
+  const { promise: configurePromise, resolve: configureResolve } =
+    deferred<Result<void, BallotPackageConfigurationError>>();
   apiMock.mockApiClient.configureFromBallotPackageOnUsbDrive
     .expectCallWith()
-    .resolves(ok());
+    .returns(configurePromise);
```
jonahkagan, line 167: If this ends up being a common pattern we might build it into the API mock. E.g. `.resolves(value, { deferred: true })` could return a `deferred`

apps/scan/frontend/src/app.test.tsx
```diff
@@ -786,7 +788,10 @@ test('poll worker can open, pause, unpause, and close poll without scanning any
   apiMock.expectSetPollsState('polls_paused');
   apiMock.expectGetConfig({ pollsState: 'polls_paused' });
   userEvent.click(await screen.findByText('Pause Voting'));
-  await screen.findByText('Pausing Voting…');
+  // avoid `act` warning due state changes after async report printing runs
```
jonahkagan, line 791: I thought this warning got removed in React 18?

apps/scan/frontend/src/screens/poll_worker_screen.test.tsx
jonahkagan: Do we have other coverage of this? I think one of the main responsibilities of the frontend tests should be covering the various states we expect to see.

I get that it feels tedious when it takes a lot of work. Maybe that just means we need to make it easier to set up

apps/mark/frontend/src/app_end_to_end.test.tsx
jonahkagan, line 1: Heroic update to this file! Love seeing all the `advanceTimersAndPromises` go away. Would love to kill that function someday

apps/mark/frontend/src/app_quit_on_idle.test.tsx
jonahkagan: We could move the logic of that effect into the `useInterval`.

I also think we need to remove the mutation object from the dependency array like you mentioned. We should never be doing that, it's a footgun.

apps/mark/frontend/src/app_setup_errors.test.tsx
```diff
@@ -200,52 +192,43 @@ describe('Displays setup warning messages and errors screens', () => {
...
     // Start on Insert Card screen
-    screen.getByText(insertCardScreenText);
+    await screen.findByText(insertCardScreenText);
 
     // Remove charger and reduce battery level slightly
     act(() => {
       hardware.setBatteryDischarging(true);
       hardware.setBatteryLevel(0.6);
     });
-    await advanceTimersAndPromises(BATTERY_POLLING_INTERVAL / 1000);
```
jonahkagan, line 216: Does removing these significantly slow down the test overall? I think it's good to use fake timers and advance them when we have any sort of significant polling delay.

apps/mark/frontend/src/pages/accessible_controller_diagnostic_screen.test.tsx
jonahkagan: Have you confirmed that there is no re-render happening? Or is it just happening after we try to press `{Enter}`? I'm happy to look at this one closer with you.

apps/mark/frontend/src/pages/idle_page.tsx
jonahkagan: How would you access the `.mutate` method inside the `useCallback` without also depending on the mutation object as well? Wouldn't the callback function close over the mutation object?

jonahkagan (approved):
Awesome work! Excited to get this in.

libs/ui/src/print_element.test.tsx
adghayes: I spent a while with this one and originally wrote the tests. I think it's best to just leave as-is, but feel free if you're interested.

apps/scan/frontend/src/screens/poll_worker_screen.test.tsx
adghayes: Actually I was wrong, I think I was just avoiding an `act` warning here. Added back the assertions, wrapped. a0b04d0b9

apps/mark/frontend/src/pages/accessible_controller_diagnostic_screen.test.tsx
adghayes: Yeah, I have confirmed no re-render is happening, before or after `{Enter}`. Below is a summarized console output from adding `console.log` in a bunch of places. The render function never re-runs.   

I would appreciate it if you looked into this one with me. My read of it, that state changes aren't propagating because they don't affect the DOM, is troubling.

Let me know if you want to play with it or want me to try something else specific.

```
console.log
    run render AccessibleControllerSoundDiagnostic
    run useEffect to set/reset listener
    trigger {ArrowRight} in test, to play sound
    keydown listener runs {ArrowRight}
    hasPlayedAudio is false
    listener sets hasPlayedAudio to true
    trigger {Enter} in test
    keydown listener receives {Enter}
    hasPlayedAudio is false
    test passed callback not triggered
```

apps/mark/frontend/src/app_quit_on_idle.test.tsx
adghayes: Excellent suggestion! That works. Fixed and pushed.

I created a ticket for the fix / lint rule for `Mutation` dependencies: https://github.com/votingworks/vxsuite/issues/3794

apps/mark/frontend/src/pages/idle_page.tsx
adghayes: The callback function would close over the mutation object but, since the identify of `.mutate` is consistent, even if the mutation object we're accessing is old the `.mutate` reference should still be correct? 

But we can't allow such a think within the react-hooks lint rule, unless there's a way to make programmatic carve-outs to rules (like post-processing to ignore certain cases). I assume that such a thing is possible within `eslint`, but maybe complex.

libs/ui/src/hooks/use_devices.test.ts
adghayes: Yup :/

libs/ui/src/hooks/use_devices.test.ts
adghayes: It doesn't exist on the `renderHook` API added into mainline RTL. In the unmount case, I think we're supposed to use `unmount`. In the waiting on a status case, I think we're just supposed to use `waitFor` as is the pattern elsewhere.

apps/scan/frontend/src/app.test.tsx
adghayes: The warning that got removed is "a state update was made to an unmounted component", the "act" updates are alive and well. You can actually turn them off I think if you just don't care, it's just a testing environment thing.

apps/mark/frontend/src/app_end_to_end.test.tsx
adghayes: 100%

apps/scan/frontend/src/app.test.tsx
adghayes: That could be neat! It's a bit unintuitive to switch from `resolves(value)` to `returns(promise)`, so a `resolves` API would be easier to use.

apps/central-scan/frontend/src/components/delete_batch_modal.tsx
adghayes: The original, with `@types` upgrades, no throws lint error: 
```
src/components/delete_batch_modal.tsx:30:10 - error TS2746: This JSX tag's 'children' prop expects a single child of type 'ReactNode', but multiple children were provided.

30         <React.Fragment>
            ~~~~~~~~~~~~~~

```

That doesn't make sense to me, because I thought the whole point of a fragment was that it could have multiple children. But if you change it to:

```
          {deleteBatchMutation.error ? (
            <P color="danger">{`${deleteBatchMutation.error}`}</P>
          ) : null}
```
Then it passes, and is also valid and effectively the same. The `as string` part can be removed. f0ac108d1.

apps/mark/frontend/src/app_setup_errors.test.tsx
adghayes: Nope, it has no impact on the speed of the test. Just confirmed. This test file uses fake timers and the `waitFor` and `findBy...` APIs are also advancing fake timers between each of their checks.

apps/mark/frontend/src/pages/idle_page.tsx
jonahkagan: Ah yes, I see how that would work. I don't mind the bit of extra boilerplate to pull it out into a separate variable. A little annoying, but that's ok.

libs/ui/src/hooks/use_devices.test.ts
jonahkagan: Weird that they got rid of that. Oh well

apps/central-scan/frontend/src/components/delete_batch_modal.tsx
jonahkagan: I think the problem here is that `deleteBatchMutation.error` has type `unknown`. A few ideas for working around this...

Add a type assertion for that value:

```ts
{deleteBatchMutation.error as Error | undefined && <P color="danger">{`${deleteBatchMutation.error}`}</P>}
```

Plug in the Error type in the mutation:

```ts
    return useMutation<void, Error, { batchId: string }>(apiClient.deleteBatch);
```
The problem with this is that TS doesn't support partial inference of type parameters, so you have to explicitly list all of them.

apps/mark/frontend/src/pages/accessible_controller_diagnostic_screen.test.tsx
jonahkagan: My take after debugging this a bit is that the test has no way to wait for the re-render to happen since there's nothing changed in the DOM.

I think it's reasonable to add something to the DOM to help the test out (maybe something hidden if we can't think of a reasonable change to make on screen).

My other idea is that we could `setHasPlayedAudio(true)` right before actually playing the audio, which would result in a fairly similar UX.

```diff
       if (event.key === 'ArrowRight') {
         const wasMuted = screenReader.isMuted();
         screenReader.unmute();
+        setHasPlayedAudio(true);
         await screenReader.speak(
           'Press the select button to confirm sound is working.'
         );
         screenReader.toggleMuted(wasMuted);
-        setHasPlayedAudio(true);
       }
```

PR #3795, opened 2023-08-08
Check election hash for VxDesign ballots

jonahkagan:
## Overview

Modifies libs/ballot-interpreter to check the election hash (there was already logic for checking precinct/test mode).

In order to have the correct election hash encoded in the ballot metadata QR code, we have to first lay out all of the ballots to compute gridLayouts. To support this change, we also undo passing an ElectionDefinition around in VxDesign, since we only need the Election data until we're finally ready to layout the ballots and compute the hash.

Also changes VxDesign to show a placeholder QR code in ballot previews to help distinguish them from real ballots that have an encoded election hash.

## Demo Video or Screenshot

In browser preview with placeholder QR code
Screen Shot 2023-08-08 at 11 56 44 AM

## Testing Plan
Updated existing automated tests

## Checklist

### DIFF
--- a/apps/design/shared/src/layout.ts
+++ b/apps/design/shared/src/layout.ts
@@ -18,9 +18,11 @@ import {
   ElectionDefinition,
   getCandidatePartiesDescription,
   getContests,
+  getPrecinctById,
   GridLayout,
   GridPosition,
   Precinct,
+  safeParseElectionDefinition,
 } from '@votingworks/types';
 import makeDebug from 'debug';
 import {
@@ -30,7 +32,11 @@ import {
   Rectangle,
   TextBox,
 } from './document_types';
-import { encodeMetadataInQrCode, QrCodeData } from './encode_metadata';
+import {
+  encodeInQrCode,
+  encodeMetadataInQrCode,
+  QrCodeData,
+} from './encode_metadata';
 import { range } from './util';
 
 const debug = makeDebug('layout');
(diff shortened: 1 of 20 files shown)


apps/design/shared/src/layout.ts
```diff
@@ -1641,3 +1706,74 @@ export function layOutBallot(
...
+}
+
+/**
+ * Given an election without gridLayouts, lays out all of the ballots for the
+ * election (for every ballot style/precinct combo). Returns the laid out
+ * ballots as well as the election definition with gridLayouts.
+ *
+ * We have to lay out all of the ballots at once in order to make sure the
+ * correct election hash is encoded in the ballot metadata. The election hash
+ * needs to be from the completed election with gridLayouts defined.
```
arsalansufi, line 1741: > We have to lay out all of the ballots at once in order to make sure the correct election hash is encoded in the ballot metadata.

I initially read this as meaning that we're performing a check on election hash in the first pass, which isn't the case. We're just preparing the data that we need to compute a proper election hash. I'm parsing the sentence correctly now, after reading the third paragraph in this comment.

I think you could even drop this paragraph altogether. The third paragraph was the one that gave me the most clarity!

arsalansufi (approved):
LGTM!

apps/design/shared/src/layout.ts
jonahkagan: Great, will change

PR #3827, opened 2023-08-12
WIP Tally Infrastructure Refactor

adghayes:
## Checklist

### DIFF
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -42,10 +42,13 @@ import { basename, dirname, join } from 'path';
 import {
   BALLOT_PACKAGE_FOLDER,
   CAST_VOTE_RECORD_REPORT_FILENAME,
+  combineCardCounts,
+  combineElectionResults,
   generateFilenameForBallotExportPackage,
-  groupMapToGroupList,
   isIntegrationTest,
   parseCastVoteRecordReportDirectoryName,
+  resolveFundamentalGroupMap,
+  resolveGroupByToFundamentalGroupBy,
 } from '@votingworks/utils';
 import { dirSync } from 'tmp';
 import ZipStream from 'zip-stream';
@@ -87,10 +90,8 @@ import { handleEnteredWriteInCandidateData } from './util/manual_results';
 import { addFileToZipStream } from './util/zip';
 import { exportFile } from './util/export_file';
 import { generateBatchResultsFile } from './exports/batch_results';
-import {
-  tabulateElectionResults,
-  tabulateTallyReportResults,
-} from './tabulation/full_results';
+import { tabulateElectionResults } from './tabulation/full_results';
+import { tabulateTallyReportResults } from './tabulation/tally_report_results';
 import { getSemsExportableTallies } from './exports/sems_tallies';
 import { generateResultsCsv } from './exports/csv_results';
 import { tabulateFullCardCounts } from './tabulation/card_counts';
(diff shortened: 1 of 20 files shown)


adghayes:
Hey @kofi-q do you mind taking a cursory look (not a review look) at this PR while I'm out next week, and let me know what you think? I haven't rewritten any testing yet, but have confirmed that at a high-level (app level) the functionality is working the same. In this PR, I make things more complicated, in some ways, in the hopes of making things simpler and less error-prone in the long run. 

### Problem 
I felt that as the code was it was a) prone to a certain kind of footguns and b) not extensible for new filters and groupings.

 I noticed a bug [here](https://github.com/votingworks/vxsuite/blob/fc0d2861fe6940d426088e2b4aeb744ce273a565/apps/admin/backend/src/store.ts#L954) where, if grouping by `party`, the `store` query would group by ballot style but then give the ballot style grouping the `party` group identifier, as if it included ballots for the whole party rather than just one ballot style of a party. I probably missed this in testing because our primary fixtures have one ballot style per party.

The problem is rooted in the fact that our data model only really cares about 4 core attributes - `ballotStyleId`, `batchId`, `votingMethod`, and `precinctId`. Any other attributes we've discussed for MCVP can be derived from those 4, like `partyId`, `scannerId`, `scannerType`, geographical groupings, etc. We've had discussions in the past about how `partyId` is in a weird place and maybe we should just add it to the `cvrs` table. But I think there could be more and more examples like this in the future. E.g. in a more complex geographical data modal, there would be all sorts of geographical filters / groupings - ward, county, congressional district, etc. And for voting method, there might be higher level categories too - e.g. 'early in-person', 'in-person election day', and 'absentee' might resolve to just an 'in-person' and 'absentee' category.

### Solution
Differentiate between operations / filters / groups on the core data model and the higher-level ones we define that resolve down to the lower level ones. In this PR I use the word "Fundamental" as in `FundamentalFilter` vs. `Filter`. 

Applying this to filtering is easy - the store understands `FundamentalFilter` and [we use our type system](https://github.com/votingworks/vxsuite/blob/94cca23f93c83e0aabaf413367eb386893fcd7cf/libs/types/src/tabulation.ts#L63) to force us to resolve every `Filter` to `FundamentalFilter` in the right places.

Applying this to grouping is, as usual, harder. We want to do all grouped calculations in terms of the `FundamentalGroupBy` and, if the higher-level query is using a higher level grouping field like `partyId` or `scannerId` then we have to merge the results of each group appropriately. The logic to do this can be tricky, but it's the same logic in my last PR (#3788). By organizing it into a single conversion step between "Groups" and "FundamentalGroups", we avoid some conversion footguns. But there are two main drawbacks I see:
- Higher-level consumers (any endpoint serving data to the frontend, for example) has to convert inputs to "Fundamental", do operations, convert back. The type system will enforce this, but it's not always simple steps.
- Performance hit due to computing results for finer groups. This is mainly a problem for `partyId` groups, which I think should be rare (after I do another refactor hehe)


## TL;DR

I think this is the right way to go but I know it's dense, looking for thoughts on what's most confusing about it and how to make less dense.

adghayes:
Or maybe for now I should just extend the VxAdmin data model to include partyId normalizing some election data) and worry about the aggregate use cases at a different time. I can reuse the "fundamental" concept when doing the splits work, and build aggregation of results in later when needed. I've got a week to ponder 😅.

adghayes:
Yeah... now I'm thinking I should just do that because the fundamental groups -> groups idea is too convoluted. Although eventually I think I'll have to make a distinction between filters in the UI vs. filters in the DB, I don't have to yet! I'm thinking I'll scope this down to a smaller refactor. Leaving it open to get your 10 cents.

PR #3836, opened 2023-08-17
Implement core continuous export logic

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3835

This PR lays the foundation 🪨 for continuous export. I've deferred some functionality (notably hashing and signing of CVRs) to subsequent PRs since this PR is large enough as is. This PR also doesn't touch the import side of things at all. Note that I'm using a feature flag to to allow for incremental updates without breaking existing import/export functionality.

## Demo Video

Notice the new directories created as we scan!

https://github.com/votingworks/vxsuite/assets/12616928/5c1df19e-b1aa-4441-ab48-39c61cd7fd20

## Testing Plan

- [x] Tested manually
- [x] Began adding tests - will add more in subsequent PRs

## Checklist

- [x] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Will add in subsequent PRs


libs/utils/src/filenames.ts
```diff
@@ -118,6 +120,30 @@ export function generateCastVoteRecordReportDirectoryName(
...
+ */
+export function generateCastVoteRecordExportDirectoryName({
+  inTestMode,
+  machineId,
+  time = new Date(),
+}: {
+  inTestMode: boolean;
+  machineId: string;
+  time?: Date;
+}): string {
```
arsalansufi, line 134: Main difference between this function and the deprecated one is that this one doesn't include ballot count in the generated name. Including a ballot count in the directory name doesn't make much sense if we're continuously exporting, as the count's always changing.

I know this intermediate state with two similarly named functions is a bit confusing, but I'm going to prune a lot of old code once this project is done (in just a couple of weeks).

libs/backend/src/scan/cast_vote_records/cast_vote_record_exporter.ts
```diff
@@ -0,0 +1,415 @@
+/* istanbul ignore file */
```
arsalansufi, line 1: Going to add tests in subsequent PRs

Issue to track: https://github.com/votingworks/vxsuite/issues/3837

libs/backend/src/scan/cast_vote_records/cast_vote_record_exporter.ts
```diff
@@ -0,0 +1,415 @@
...
+      getUsbDrives: async () => {
+        const drive = await this.usbDrive.status();
+        return drive.status === 'mounted' ? [drive] : [];
+      },
+    });
+  }
+
+  async exportCastVoteRecordsToUsbDrive(
+    resultSheets: ResultSheet[] | Generator<ResultSheet>
+  ): Promise<Result<void, ExportCastVoteRecordToUsbDriveError>> {
```
arsalansufi, line 129: This function mirrors the existing [exportCastVoteRecordReportToUsbDrive](https://github.com/votingworks/vxsuite/blob/325446e15f477c4587a644b5a238df6902e5d1d6/libs/backend/src/scan/cast_vote_records/export.ts#L481-L484) but is designed for continuous export. The structure of the export directory now looks something like this:

```
cast-vote-records
└── test-ballot_general-election_5fb429f84c
    ├── TEST__machine_000__2023-08-17_10-47-02
    │   ├── 0c976e40-9ac5-468b-b167-fe7fa37fab38
    │   │   ├── 0c976e40-9ac5-468b-b167-fe7fa37fab38-back.jpg
    │   │   ├── 0c976e40-9ac5-468b-b167-fe7fa37fab38-back.layout.json
    │   │   ├── 0c976e40-9ac5-468b-b167-fe7fa37fab38-front.jpg
    │   │   ├── 0c976e40-9ac5-468b-b167-fe7fa37fab38-front.layout.json
    │   │   └── cast-vote-record-report.json
    │   ├── 2d902f55-4a02-4a9a-922b-d01f8213010e
    │   │   ├── 2d902f55-4a02-4a9a-922b-d01f8213010e-back.jpg
    │   │   ├── 2d902f55-4a02-4a9a-922b-d01f8213010e-back.layout.json
    │   │   ├── 2d902f55-4a02-4a9a-922b-d01f8213010e-front.jpg
    │   │   ├── 2d902f55-4a02-4a9a-922b-d01f8213010e-front.layout.json
    │   │   └── cast-vote-record-report.json
    │   ├── cf5a6f3b-a7c1-4f1b-8af5-b969dbe62f41
    │   │   ├── cast-vote-record-report.json
    │   │   ├── cf5a6f3b-a7c1-4f1b-8af5-b969dbe62f41-back.jpg
    │   │   ├── cf5a6f3b-a7c1-4f1b-8af5-b969dbe62f41-back.layout.json
    │   │   ├── cf5a6f3b-a7c1-4f1b-8af5-b969dbe62f41-front.jpg
    │   │   └── cf5a6f3b-a7c1-4f1b-8af5-b969dbe62f41-front.layout.json
    │   └── metadata.json
    └── TEST__machine_000__2023-08-17_10-47-02.vxsig
```

In VxScan, the common case will be to pass a result sheets array of length 1 right after scanning a ballot (as you'll see in a subsequent commit). But we still want to support passing a generator capable of iterating over all result sheets for full exports, e.g. on VxCentralScan or on VxScan when using the EM button to export all CVRs.

libs/backend/src/scan/cast_vote_records/cast_vote_record_exporter.ts
```diff
@@ -0,0 +1,415 @@
...
+    };
+  }
+
+  private getExportDirectoryPathRelativeToUsbMountPoint(
+    scannerState: ScannerStateUnchangedByExport
+  ): string {
+    const { electionDefinition, inTestMode } = scannerState;
+    const { election, electionHash } = electionDefinition;
+
+    let exportDirectoryName = this.scannerStore.getExportDirectoryName();
```
arsalansufi, line 177: One of the new pieces of state that we need to manage for continuous export is the name of the export directory. We could try to infer it, e.g. find all directories on the USB with the correct name format and then pick the one with the most recent date. But that quickly gets complicated by edge cases.

libs/backend/src/scan/cast_vote_records/cast_vote_record_exporter.ts
```diff
@@ -0,0 +1,415 @@
...
+    const exportResult = await this.exporter.exportDataToUsbDrive(
+      exportDirectoryPathRelativeToUsbMountPoint,
+      'metadata.json',
+      JSON.stringify(metadata)
+    );
+    if (exportResult.isErr()) {
+      return exportResult;
+    }
+
+    // TODO: Sign metadata file
```
arsalansufi, line 319: Issue to track: https://github.com/votingworks/vxsuite/issues/3838

apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -418,6 +431,17 @@ function recordAcceptedSheet(store: Store, { interpretation }: Context) {
...
+    isFeatureFlagEnabled(
+      BooleanEnvironmentVariableName.ENABLE_CONTINUOUS_EXPORT
+    )
+  ) {
+    const exportResult =
+      await castVoteRecordExporter.exportCastVoteRecordsToUsbDrive([
+        assertDefined(store.getResultSheet(sheetId)),
+      ]);
+    exportResult.unsafeUnwrap();
+  }
```
arsalansufi, line 444: In my testing, I've found that exporting CVRs does not block the display of the success check, which is great! The exporting happens in the background after the success check is displayed.

I do intend to think more deeply about and revisit error handling. For now, if the export fails, you'll see the catch-all "something went wrong, please restart" error screen.

Issue to track: https://github.com/votingworks/vxsuite/issues/3839

libs/backend/src/scan/cast_vote_records/cast_vote_record_exporter.ts
```diff
@@ -0,0 +1,415 @@
...
+export interface CastVoteRecordExporterApi {
+  exportCastVoteRecordsToUsbDrive(
+    resultSheets: ResultSheet[] | Generator<ResultSheet>
+  ): Promise<Result<void, ExportCastVoteRecordToUsbDriveError>>;
+}
+
+/**
+ * The implementation of the cast vote record exporter API
+ */
+export class CastVoteRecordExporter implements CastVoteRecordExporterApi {
```
jonahkagan, line 103: It looks like this class doesn't actually hold any state in memory. Does it need to be its own object, or could it just be a function `exportCastVoteRecordsToUsbDrive` that takes some of these stateful objects as arguments?

The main motivation I could think of for making it an object would be to inject it at the top level of the app and then be able to mock it, but I've generally been pushing for as little intermediate mocking as possible in our backends (to try to make the tests more realistic).

Curious to hear your thoughts!

apps/scan/backend/src/store.ts
```diff
@@ -993,4 +1010,48 @@ export class Store {
...
+      'insert into export_directory_name (export_directory_name) values (?)',
+      exportDirectoryName
+    );
+  }
+}
+
+/**
+ * Builds a ScannerStore object to be passed through to a CastVoteRecordExporter
+ */
+export function buildScannerStoreForCastVoteRecordExporter(
```
jonahkagan, line 1041: What's the motivation for building this new store object vs just having the store implement the `ScannerStore` interface?

I could see it being a little confusing to have an extra wrapper layer that adds slight modifications to the logic.

arsalansufi: Yeah great callout - I was just trying to minimize modifications to the existing VxScan `Store` while allowing myself to use names that felt a bit clearer in new code, e.g. `.getBatches` instead of `.batchStatus`.

But upon closer inspection, the amount of work required to make the VxScan `Store` implement the `ScannerStore` interface is minimal, including renaming existing methods in the VxScan `Store`. Will modify and get rid of the wrapper, which I agree is a little confusing

libs/backend/src/scan/cast_vote_records/cast_vote_record_exporter.ts
arsalansufi: Really good question!

I did notice that the class doesn't hold any state in memory, so you're right that it doesn't need to be a class and could just be a function with several inputs.

But I've been finding functions with so many inputs to be a little clunky, given how much you have to pass to them when you want to call them. With a class, you can at least pre-wire certain required inputs when you initialize the class, say in an `index.ts` file, and not have to worry about those from then on.

This benefit then extends to internal helpers, too, in that you don't have to continue passing around the same inputs over and over. Though I know this can be mitigated by grouping inputs in an interface.

That said, I definitely agree with the point about minimizing intermediate mocking in our backend tests and testing more completely. This made me realize that I was unnecessarily mocking the cast vote record exporter in VxScan backend tests when I could just initialize the real deal. Updated accordingly in https://github.com/votingworks/vxsuite/pull/3836/commits/d09c99d2a9a69bccad7047c08729bdde30aac0b9

Overall though, I'm definitely open to switching to plain functions, with well encapsulated inputs

jonahkagan: Yeah, I hear you on the annoyance of passing lots of inputs around through many functions.

In this case, would you need to pass anything except the `UsbDrive`? I think the `Store` is already available and the `ArtifactAuthenticator` could be constructed on demand.

One thing I notice about this pattern is that it encourages creating stateless singletons and then feeling like they need to be passed around instead of building them on demand. So the pattern then proliferates. When really these objects are just a bundle of code with preconfigured arguments.

I guess I'm not opposed to that approach, but it does make it a bit hard to know what's stateful and what's not. But this is mostly a style/readability concern, and my opinions are weakly held.

arsalansufi: > One thing I notice about this pattern is that it encourages creating stateless singletons and then feeling like they need to be passed around instead of building them on demand

Really good point!

I know I'm guilty of this with the `ArtifactAuthenticator`, though there was another reason at play there. I initially planned for it to be constructed as needed, but it needs certain env vars to configure itself, which aren't available in test environments. So I switched to a dependency injection approach to make it easier to mock in tests. In retrospect, I could've just set up the right env vars in tests. That's what I ended up doing for integration tests.

> the `ArtifactAuthenticator` could be constructed on demand

See the above haha. That said, in my next PR, I'll be revisiting the `ArtifactAuthenticator`, and I'm thinking of addressing the above while I'm at it, so that it can in fact be constructed on demand, or just turned into plain functions.

I'm convinced by this thread and am thinking of switching to plain functions in this PR, too. The "too many arguments" problem can be solved by grouping arguments meaningfully. I think the `ScannerStore` abstraction largely accomplishes that.

jonahkagan: Oh interesting puzzle with the env vars. I hadn't thought of that. Maybe having parameters that default to env vars could be a good approach? That way they can be overridden in test, but the consuming app code doesn't need to worry about managing the env vars.

arsalansufi:
Made the edits discussed with @jonahkagan but opened a new PR to make reviewing easier for @adghayes when he's back! https://github.com/votingworks/vxsuite/pull/3842

libs/backend/src/scan/cast_vote_records/cast_vote_record_exporter.ts
arsalansufi: ^ So that actually is how the `ArtifactAuthenticator` works ([relevant link](https://github.com/votingworks/vxsuite/blob/dffb657cad7f104fc8f2f345186eeea236dd33b2/libs/auth/src/artifact_authenticator.ts#L68-L72)) and how I test it ([relevant link](https://github.com/votingworks/vxsuite/blob/dffb657cad7f104fc8f2f345186eeea236dd33b2/libs/auth/src/artifact_authenticator.test.ts#L99-L101)).

The challenge in app-level tests, assuming that we were to construct the `ArtifactAuthenticator` wherever it was needed, is that it would exist embedded within other code, where we couldn't easily modify how it's constructed. Ways around this of course, which I'll explore!

PR #3852, opened 2023-08-21
Move mark thresholds from election definition to system settings

jonahkagan:
## Overview

In order to fully support the election definition CDF, we need to align it exactly with our existing election definition fields. Til now, we've included the mark thresholds in the election definition. However, there's no corresponding field for these in CDF. So, here we move the mark thresholds to system settings instead.

## Demo Video or Screenshot
N/A

## Testing Plan
- Updated existing tests

## Checklist


libs/types/src/system_settings.ts
```diff
@@ -15,20 +15,24 @@ import {
...
  * (and therefore not needing to reprint ballots, for example).
  */
 export interface SystemSettings {
-  arePollWorkerCardPinsEnabled: boolean;
-  inactiveSessionTimeLimitMinutes: InactiveSessionTimeLimitMinutes;
-  numIncorrectPinAttemptsAllowedBeforeCardLockout: NumIncorrectPinAttemptsAllowedBeforeCardLockout;
-  overallSessionTimeLimitHours: OverallSessionTimeLimitHours;
-  startingCardLockoutDurationSeconds: StartingCardLockoutDurationSeconds;
+  readonly auth: AuthSettings;
 }
```
arsalansufi: So there's one more place where we need to account for this updated structure, namely all the `constructAuthMachineState` functions in the app backends:

https://github.com/votingworks/vxsuite/blob/500be04f064cda0511843c578b5c278b35addf08/apps/admin/backend/src/app.ts#L113-L125

TS isn't complaining because all the machine state fields are optional:

https://github.com/votingworks/vxsuite/blob/500be04f064cda0511843c578b5c278b35addf08/libs/auth/src/dipped_smart_card_auth_api.ts#L50-L60

Retroactively realizing the risk that can come with optional fields 😅 Might make sense to just mark the relevant fields as required and then make it the app's responsibility to fall back to the defaults when there are no system settings (i.e. before a machine has been configured)

arsalansufi (commented):
Still reviewing but wanted to get my one more substantial comment in front of you proactively!

libs/ballot-interpreter/src/hmpb-ts/cli.ts
```diff
@@ -516,14 +552,41 @@ export async function main(args: string[]): Promise<number> {
...
+      );
+      usage(stderr);
+      return 1;
+    }
+
+    const electionDefinition = parseElectionDefinitionResult.ok();
+
+    const systemSettings = safeParseSystemSettings(
+      await fs.readFile(systemSettingsPath, 'utf8')
+    ).ok();
```
arsalansufi, line 584: Should we apply the same error handling that we do for election definition parsing if we fail to parse the system settings file? Looks like we're currently silently ignoring failures to parse the system settings file

libs/auth/src/java_card.test.ts
```diff
@@ -639,6 +639,8 @@ test.each<{
 
     mockCardReader.setReaderStatus('ready');
     await waitForExpect(async () => {
+      // This might fail if the test keys and certs are outdated (e.g. if the election hash changes).
+      // Run ./scripts/generate-test-keys-and-certs to update them.
```
arsalansufi, line 643: Thanks for adding this note!

apps/central-scan/backend/src/end_to_end_hmpb.test.ts
```diff
@@ -146,7 +152,15 @@ test('going through the whole process works', async () => {
...
     'ballot-packages': {
       'ballot-package.zip': await createBallotPackageZipArchive(
-        electionGridLayoutNewHampshireAmherstFixtures.electionJson.toBallotPackage()
+        electionGridLayoutNewHampshireAmherstFixtures.electionJson.toBallotPackage(
+          {
+            ...DEFAULT_SYSTEM_SETTINGS,
+            markThresholds: {
+              definite: 0.08,
+              marginal: 0.05,
+            },
```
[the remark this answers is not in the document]
jonahkagan: Yeah, those are the mark thresholds that were in the election definition for this election.

libs/types/src/system_settings.ts
jonahkagan: Ooh nice catch!

I think your suggestion to fix it makes sense. The other idea that comes to mind is having a test in each app that ensures auth is correctly integrated.

arsalansufi: ^ That makes sense! I'm fine with either

libs/ballot-interpreter/src/hmpb-ts/cli.ts
jonahkagan: Nice catch, thanks. Added a warning and explanatory comment: 1619cfbe8

libs/types/src/system_settings.ts
jonahkagan: Implemented required fields here: https://github.com/votingworks/vxsuite/pull/3860. It turns out there were tests, but the tests never set the system settings during machine configuration, so they weren't catching this.

jonahkagan:
@arsalansufi This is ready for re-review

arsalansufi (approved):
LGTM!

PR #3853, opened 2023-08-21
[VxScan] Ensure voter has the ability to review all overvote/undervote contests

kofi-q:
## Overview

Addressing VVSG2.0 7.3-H and 7.3-I:

> 7.3-H – Overvotes
> 2. A scanner or other device that a voter uses to cast a paper ballot must be capable of
providing feedback that identifies specific contests that have been overvoted in visual
format, and with either audio format or sound cues.

> 7.3-I – Undervotes
The voting system must notify voters in both visual and audio formats of the specific contest
in which they select fewer than the allowable number of options (that is, for undervotes).

This updates the `ScanWarningScreen` in VxScan to list out all contests with undervotes/overvotes - we list them all out on the main screen if the list is short enough to fit; otherwise, we render a summary along with a button to see the full list.

This also combines the previously separate overvote and undervote warnings, since in the previous case, it was possible to dismiss the overvote warning and have your ballot cast without the undervotes being surfaced.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/264902/085e8cbd-0544-431c-9a90-3530ba9514a0

## Testing Plan
- Added unit tests for the new components and updated existing functional tests for `ScanWarningScreen`

## Checklist

### DIFF
--- a/apps/scan/frontend/src/components/misvote_warnings/test_utils.test.ts
+++ b/apps/scan/frontend/src/components/misvote_warnings/test_utils.test.ts
@@ -1,0 +1,23 @@
+import { electionSample } from '@votingworks/fixtures';
+import { AnyContest } from '@votingworks/types';
+
+const CONTEST_TEMPLATE: AnyContest = electionSample.contests[0];
+
+export function generateContests(count: number): AnyContest[] {
+  const contests: AnyContest[] = [];
+
+  for (let i = 0; i < count; i += 1) {
+    contests.push({
+      ...CONTEST_TEMPLATE,
+      id: `${i}`,
+      title: `[TEST ${i}] ${CONTEST_TEMPLATE.title}`,
+    });
+  }
+
+  return contests;
+}
+
+test('generates contests', () => {
+  const contests = generateContests(2);
+  expect(contests[0]).not.toEqual(contests[1]);
+});
--- a/apps/scan/frontend/src/components/misvote_warnings/types.ts
+++ b/apps/scan/frontend/src/components/misvote_warnings/types.ts
@@ -1,0 +1,19 @@
+import { AnyContest } from '@votingworks/types';
+
+export interface MisvoteWarningsProps {
+  blankContests: readonly AnyContest[];
+  overvoteContests: readonly AnyContest[];
+  partiallyVotedContests: readonly AnyContest[];
+}
+
+export interface MisvoteWarningsConfig {
+  maxCardsPerRow: number;
+  maxColumnsPerCard: number;
+  maxPreviewContestRows: number;
+}
+
+export interface Layout {
+  maxColumnsPerCard: number;
+  numCardsPerRow: number;
+  showSummaryInPreview?: boolean;
+}
--- a/apps/scan/frontend/src/components/misvote_warnings/use_layout_config_hook.test.tsx
+++ b/apps/scan/frontend/src/components/misvote_warnings/use_layout_config_hook.test.tsx
@@ -1,0 +1,168 @@
+import { SizeMode } from '@votingworks/types';
+import { render } from '../../../test/react_testing_library';
+import { Layout, MisvoteWarningsProps } from './types';
+import { useLayoutConfig } from './use_layout_config_hook';
+import { CONFIG } from './constants';
+import { generateContests } from './test_utils.test';
+
+let hookResult: Layout | null = null;
+
+function TestComponent(props: MisvoteWarningsProps) {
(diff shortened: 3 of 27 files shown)


apps/scan/frontend/src/components/misvote_warnings/test_utils.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      title: `[TEST ${i}] ${CONTEST_TEMPLATE.title}`,
+    });
+  }
+
+  return contests;
+}
+
+test('generates contests', () => {
+  const contests = generateContests(2);
+  expect(contests[0]).not.toEqual(contests[1]);
```
adghayes, line 22: I think `toMatchObject` would probably be more complete here, to ensure they're not different objects with the same contents.

apps/scan/frontend/src/components/misvote_warnings/use_layout_config_hook.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+      overvoteContests={[]}
+      partiallyVotedContests={[]}
+    />
+  );
+  expect(hookResult).toEqual(
+    expect.objectContaining<Partial<Layout>>({
+      numCardsPerRow: 1,
+    })
+  );
+});
```
adghayes, line 66: It looks like maybe you separated the layout config as a hook for test-ability? That seems fine.

Maybe I'm misremembering, but I think the only data actually used is the `showPreview` flag so I don't know if you need to assert against these other things / have them in there.

adghayes (approved):
LGTM, some minor questions about the new hook.

libs/ui/src/list_item.tsx
```diff
@@ -0,0 +1,39 @@
...
+`;
+
+export function ListItem(props: ListItemProps): JSX.Element {
+  const { children, className } = props;
+
+  return (
+    <Container className={className}>
+      <IconContainer>
+        <Icons.CircleSolid />
+      </IconContainer>
```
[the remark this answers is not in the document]
kofi-q: yeah, i was having trouble getting it to play nice with the grid layout, but it's worth another shot, since it would simplify this code a bit - just had another idea for getting it to work that I'd like to try out

apps/scan/frontend/src/components/misvote_warnings/use_layout_config_hook.ts
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { useTheme } from 'styled-components';
+
+import { CONFIG } from './constants';
+import { Layout, MisvoteWarningsProps } from './types';
+
+export function useLayoutConfig(props: MisvoteWarningsProps): Layout {
```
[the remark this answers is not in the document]
kofi-q: ah yeah, it's used in both `misvote_warnings` and `warning_details` and it depends on the `useTheme` hook (so, the calculations and result of the hook change when the size theme is changed) - decided to make it a hook as well instead of having clients pass in the theme themselves

apps/scan/frontend/src/components/misvote_warnings/use_layout_config_hook.test.tsx
kofi-q: yeah, the other fields are referenced in `warning_details`, which was the initial motivation for extracting to a common util, but it definitely made testing easier as well!

apps/scan/frontend/src/components/misvote_warnings/test_utils.test.ts
kofi-q: ah, IIUC `toMatchObject()` only matches object subsets, so `{a: "one", b: "two"}` would match `{a: "one"}`, whereas `toEqual()` matches all fields (and `toBe()` matches the actual object references)

apps/scan/frontend/src/components/misvote_warnings/types.ts
```diff
@@ -0,0 +1,19 @@
+import { AnyContest } from '@votingworks/types';
+
+export interface MisvoteWarningsProps {
+  blankContests: readonly AnyContest[];
+  overvoteContests: readonly AnyContest[];
+  partiallyVotedContests: readonly AnyContest[];
```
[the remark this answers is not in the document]
kofi-q: yeah, was on the fence about this - initially just copied what we previously had in the `ScanWarningScreen` component but also couldn't think of anything better to call it. IIUC, a blank contest is technically an undervoted contest, so this naming keeps the two separate for display purposes

PR #3857, opened 2023-08-22
refactor(admin): use DB logic to determine splits/groups instead of our own utils

adghayes:
_review by commit_

## Overview

I don't really like the way that I wrote for determining splits in #3788. I don't trust that code to manage all the relationships between the different pieces of election data (ballot style, etc.). And I realized that all that code is essentially mimicking the behavior of a SQL database when doing joins or filtering. E.g. the function `getCartesianProductSplits` was just doing an `inner join` without an `on` clause.

I think this is much better and more trustworthy in the DB. It does introduce some annoying SQL manipulation code, but that's an acceptable tradeoff.

## Testing Plan

Rewrote testing for utils as testing for the admin store.


libs/fixtures/data/electionComplexGeoSample/README.md
```diff
@@ -0,0 +1,8 @@
+### Fixture Purpose
```
adghayes, line 1: Not sure where else to provide this context, which I think is relevant to a reader of the code.

apps/admin/backend/src/store.ts
```diff
@@ -329,6 +375,159 @@ export class Store {
...
+            voting_method
+          ) 
+          values 
+            ${Tabulation.SUPPORTED_VOTING_METHODS.map(() => '(?, ?)').join(
+              ',\n'
+            )}
+        `,
+      ...params
+    );
+  }
```
adghayes, line 428: These records exist simply to be able to get the voting method "cartesian product" with other splits through the DB rather than adding logic outside of DB.

eventualbuddha: Could they be added after the `create table` in `schema.sql`?

apps/admin/backend/schema.sql
```diff
@@ -6,6 +6,16 @@ create table elections (
   deleted_at timestamp
 );
 
+create table precincts(
```
kofi-q, line 9: glad we're breaking some of these pieces out of the json blob - always wondered if there was anything stopping us from doing so

apps/admin/backend/src/store.ts
adghayes: Definitely. My thinking here, though, was that eventually this may be a jurisdiction by jurisdiction setting so I didn't want it at the `schema` level.

kofi-q (approved):

apps/admin/backend/schema.sql
adghayes: I thought you'd like it!

apps/admin/backend/schema.sql
```diff
@@ -6,6 +6,16 @@ create table elections (
...
+  election_id integer not null,
+  id text not null,
+  name text not null,
+  sort_index integer not null,
```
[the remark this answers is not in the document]
adghayes: In the MS case, SEMS exports precincts in a non-alphabetical order that I think customers are used to seeing in reports.

But... I just looked at some samples are realized that non-alphabetical is actually alphabetical on `id`, but not `name`. So I think I'll just do that for simplicity.

PR #3858, opened 2023-08-22
refactor(admin): strict SQL tables

adghayes:
## Overview

Use strict mode in `admin/backend` SQL tables. The only practical type safety this brings for us, AFAICT, is preventing inserting text values into integer columns. For the `schema.sql` file, it forces us to stop using made up types that SQLite ignores anyway and only use the types that they are actually interpreted as (e.g. don't use `varchar(36)`, use `text`, because SQLite ignores `varchar` limits anyways).


apps/admin/backend/schema.sql
```diff
@@ -1,44 +1,74 @@
 create table elections (
-  id serial primary key,
+  id text primary key,
```
eventualbuddha, line 2: I find myself wishing to PostgreSQL's `uuid` type.

apps/admin/backend/schema.sql
```diff
@@ -1,44 +1,74 @@
...
   data text not null,
-  is_official_results boolean not null default false,
-  created_at timestamp not null default current_timestamp,
-  deleted_at timestamp
-);
+  is_official_results integer not null default false,
```
eventualbuddha, line 4: Worth adding a check here?
```suggestion
  is_official_results integer not null default false
    check (is_official_results = true or is_official_results = false),
```

eventualbuddha (commented):
Are the other changes related to the schema changes?

adghayes:
Closing this to revive when everyone's on Debian 12.

PR #3859, opened 2023-08-22
Use functions instead of a class for artifact authentication

arsalansufi:
_Easiest reviewed by commit_

## Overview

Following [this discussion](https://github.com/votingworks/vxsuite/pull/3836#discussion_r1297531983) with @jonahkagan, I've decided to update artifact authentication to use functions instead of a class. The logic involves no state, so a class isn't necessary. And using a class encourages dependency injection of a singleton, making the code more involved than it needs to be.

While dependency injection of a singleton does make mocking easier in tests, avoiding this kind of intermediate mocking allows us to test as close to the real deal as possible. I'm still mocking and/or skipping checks using feature flags in many cases, so this PR doesn't finish following through on that idea. But I have ideas as for how to do so that I'll tackle in follow-up PRs, e.g. adding signature files to fixtures or just creating them on the fly, etc. Didn't want to balloon this PR too much.

## Testing

- [x] Updated automated tests
- [x] Tested artifact authentication manually

## Checklist


libs/auth/src/artifact_authentication.ts
```diff
@@ -0,0 +1,265 @@
+import { Buffer } from 'buffer';
```
arsalansufi, line 1: No need to review this commit in detail. I've just taken the methods on the `ArtifactAuthenticator` class and made them functions, reordering to avoid "this function is used before it's defined" errors.

I haven't changed the bodies of the functions at all, minus accessing config values from a `config` param instead of `this.`

libs/backend/src/ballot_package/ballot_package_io.test.ts
```diff
@@ -35,17 +33,19 @@ import { UsbDrive } from '../get_usb_drives';
...
+
 jest.mock('@votingworks/utils', (): typeof import('@votingworks/utils') => ({
   ...jest.requireActual('@votingworks/utils'),
   isFeatureFlagEnabled: (flag) => mockFeatureFlagger.isEnabled(flag),
 }));
 
-let mockArtifactAuthenticator: jest.Mocked<ArtifactAuthenticatorApi>;
-
 beforeEach(() => {
+  mockOf(authenticateArtifactUsingSignatureFile).mockResolvedValue(ok());
```
arsalansufi, line 47: As noted in the PR description, I'm still mocking artifact authentication in many places. I have plans to stop mocking in subsequent PRs, where I'll either add signature files to fixtures or just create them on the fly, etc.

libs/backend/src/scan/cast_vote_records/legacy_export.test.ts
```diff
@@ -32,6 +29,11 @@ import {
   ResultSheet,
 } from './legacy_export';
 
+jest.mock('@votingworks/auth', (): typeof import('@votingworks/auth') => ({
+  ...jest.requireActual('@votingworks/auth'),
+  writeSignatureFile: jest.fn(),
```
arsalansufi, line 34: I've removed mocking of the writing of signature files in most places. I just couldn't in this test file since all data writing is itself mocked, so `writeSignatureFile` can't read in the data that it needs to sign. In other test files, where we actually write data to the file system, `writeSignatureFile` can access the data that it needs to sign and works just fine.

Didn't feel worth updating since I'll soon be replacing this test file anyway.

apps/admin/backend/test/set_env_vars.ts
```diff
@@ -0,0 +1,3 @@
+process.env['VX_MACHINE_TYPE'] = 'admin';
```
arsalansufi, line 1: Artifact authentication relies on this variable, which is guaranteed to be set on all real machines

jonahkagan: Maybe put this comment in this file for posterity?

jonahkagan (approved):
💯

apps/admin/backend/test/set_env_vars.ts
arsalansufi: Good call: https://github.com/votingworks/vxsuite/pull/3859/commits/81f1edb8d55ec0a507d6af9f5dc8140663846999

PR #3863, opened 2023-08-22
CDF election definition

jonahkagan:
## Overview

Changes VxDesign to export election definitions in CDF. The bulk of the work to implement this was to extend the code that converts between VXF and CDF to support the physical ballot definition - the information that the scanners need to be able to locate bubbles on the ballots.

## Testing Plan
- Updated automated tests

## Checklist


libs/types/src/cdf/ballot-definition/index.ts
jonahkagan, line 1: Reminder that this file is auto-generated based on vx-schema.json

libs/types/src/election_parsing.ts
jonahkagan, line 1: The functions in this file were moved from election_utils to avoid a cyclic dependency. No code changes

libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan, line 1: This file has the meat of the change!

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -45,6 +58,174 @@ export function convertVxfElectionToCdfBallotDefinition(
   const stateId = vxfElection.state.toLowerCase().replaceAll(' ', '-');
   const electionDate = dateString(new Date(vxfElection.date));
 
+  const precinctSplits: Record<
```
jonahkagan, line 61: I think at some point it would be nice to integrate our split precinct support into the VXF format. Currently the flow looks like this:
- User enters explicit split precinct data which in VxDesign
- That gets converted to VXF (encoding splits in ballot styles)
- The VXF is then converted to CDF for export (back to explicit split precincts)
- The CDF is converted back to VXF on import to VxAdmin (back to encoding splits in ballot styles)

It's a bit convoluted, but I think it works for now.

jonahkagan:
Whoops, found a small bug: the target mark position (left or right) needs to be included in CDF. The rest of the PR should be fine to review and I'll add that on top

jonahkagan:
Ok I fixed the left/right bubble issue in #3877, so this is now ready for review for real!

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -45,6 +58,174 @@ export function convertVxfElectionToCdfBallotDefinition(
...
+      }>
+    | undefined
+  > = Object.fromEntries(
+    vxfElection.precincts.map((precinct) => {
+      const precinctBallotStyles = vxfElection.ballotStyles.filter(
+        (ballotStyle) => ballotStyle.precincts.includes(precinct.id)
+      );
+      // There may be multiple ballot styles with the same districts but
+      // different partyIds, but we only want to split precincts that are part
+      // of different districts.
```
arsalansufi, line 75: So this comment is basically noting that there's a second reason why a single precinct might have multiple ballot styles. The first reason is a split precinct. The second reason is ballot styles for different party affiliations.

arsalansufi (commented):
Still reviewing but just getting some questions in front of you (mostly just understanding questions)

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -433,7 +433,11 @@ export function convertVxfElectionToCdfBallotDefinition(
       },
     ],
 
-    GeneratedDate: dateTimeString(new Date()),
+    // Since we don't have a generated date in VXF, we use the election date. If
+    // we were to use the current date, it would cause changes every time we
+    // hash the object. We want hashes to be based on the content of the
+    // election, not the date generated.
+    GeneratedDate: dateTimeString(new Date(vxfElection.date)),
```
arsalansufi, line 440: 👍

libs/monorepo-utils/src/dependencies.test.ts
```diff
@@ -81,8 +81,9 @@ test('findAllMonorepoDependencies yields all dependencies', () => {
   const pkgs = getWorkspacePackageInfo(join(__dirname, '../../..'));
   const basicsPkg = pkgs.get('@votingworks/basics')!;
 
-  // single dependency
+  // simple dependencies
```
arsalansufi, line 84: What exactly do you mean by "simple dependencies"? (To be clear, I don't think I understand "single dependency" either 😅)

arsalansufi (approved):
Looks good! :rocket:

libs/types/src/cdf/ballot-definition/vx-schema.json
```diff
@@ -26,7 +27,8 @@
           "items": {
             "$ref": "#/definitions/BallotDefinition.BallotFormat"
           },
-          "minItems": 0,
+          "minItems": 1,
+          "maxItems": 1,
```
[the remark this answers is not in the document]
jonahkagan: Yeah exactly. In general I've taken the approach of allowing our schema (vx-schema.json) to be stricter than the base schema (nist-schema.json), but not the reverse (except for adding custom fields). That way we can be sure that we're compatible with a subset of CDF. There's an automated test that enforces this.

libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan: Yup! This is used for primary elections. VxDesign doesn't actually generate the separate ballot styles for party affiliations yet (it's next on the roadmap), but we have fixtures that already have party-affiliated ballot styles.

libs/monorepo-utils/src/dependencies.test.ts
jonahkagan: Yeah I was confused as well. This function finds all of the deps of the given package that are workspace packages (i.e. directories in the monorepo) rather than npm packages. This particular assertion was trying to check a simple case (a package with only one workspace dep), but my changes added another dep, so I updated the comment to try to avoid it being misleading. I don't know that I increased the clarity at all haha

libs/types/src/cdf/ballot-definition/vx-schema.json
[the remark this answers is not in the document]
arsalansufi: Perfect, makes sense! Just now becoming aware of the existence of `nist-schema.json` in the repo. The `vx-schema.json` file name for a CDF schema is also now clearer to me

PR #3864, opened 2023-08-22
Require explicit options for YesNoContests

jonahkagan:
## Overview

Previously, `YesNoContest.yesOption/noOption` were _optional_ 😉.  CDF requires explicit options with globally unique IDs. The conversion code to make this work out was getting really confusing, especially since these option IDs need to correspond with the option IDs in GridLayouts. Furthermore, there were a bunch of places in our code that didn't account for the possibility of an explicit option id and just used `'yes'` and `'no'` strings.

To clean this all up, I made `yesOption` and `noOption` required. I also changed the `YesNoContestOption` type to reference these ids rather than use `'yes'` and `'no'`. However, I left the `YesNoVote` type the same to try to keep changes minimal.

This turned out to be a bit of an adventure, as you'll see in the commits.

## Testing Plan
Updated existing tests

## Checklist


libs/mark-flow-ui/src/components/yes_no_contest.tsx
```diff
@@ -134,15 +133,16 @@ export function YesNoContest({
...
-                    }
+                    {overvoteSelection === 'yes'
+                      ? contest.yesOption.label
+                      : contest.noOption.label}
+                  </strong>
+                  ? To change your vote, first unselect your vote for{' '}
+                  <strong>
+                    {overvoteSelection === 'yes'
+                      ? contest.noOption.label
+                      : contest.yesOption.label}
```
arsalansufi, line 145: Wow! Good find 🙈

arsalansufi (approved):
LGTM!

PR #3877, opened 2023-08-23
Remove BallotLayout.targetMarkPosition and BallotLayout.layoutDensity from Election

jonahkagan:
## Overview

VxSuite doesn't use these fields for interpretation, but VxDesign uses them for ballot layout. To ensure that the election exported is as minimal as possible, I'm removing these fields from `Election`. (Also partially motivated by the fact that they don't exist in CDF, so it's an easy place to accidentally make mistakes when converting.)

Instead, these two settings are now stored separately in VxDesign.

One large piece of code that depended on these fields was the legacy HMPB component in VxAdmin, which I removed (since it's already been replaced by VxDesign).

_Review by commits_

## Demo Video or Screenshot
N/A

## Testing Plan
Automated tests

## Checklist


libs/types/src/election.ts
```diff
@@ -351,30 +351,12 @@ export enum BallotPaperSize {
...
-/**
- * Schema for {@link BallotTargetMarkPosition}.
- */
-export const BallotTargetMarkPositionSchema: z.ZodSchema<BallotTargetMarkPosition> =
-  z.nativeEnum(BallotTargetMarkPosition);
-
 export interface BallotLayout {
   paperSize: BallotPaperSize;
-  layoutDensity?: number;
-  targetMarkPosition?: BallotTargetMarkPosition;
```
arsalansufi, line 371: Reasoning through the implications of this: Changing the bubble position or layout density will no longer change the election hash. Which means that, even after changing bubble position or layout density, ballots with the previous bubble position or layout density will still be scannable, assuming all else is unchanged. Is that correct? (Seems reasonable to me, just wanted to check my understanding)

arsalansufi (approved):
LGTM!

libs/types/src/election.ts
jonahkagan: Changing these layout settings will change the actual ballot layout, which will change the gridLayouts in the election definition (which is what actually tells the scanner where the bubbles are), which will change the election hash.

arsalansufi: Ah perfect - hadn't considered the grid layouts tie-in. Makes sense 👍

PR #3878, opened 2023-08-23
admin(refactor): tally report data refactor

adghayes:
_review by commit_

## Overview

As I look towards making a custom report builder, I wanted to simplify the results operations that the frontend is responsible for. There are two hurdles that make reports just a little more complex than I'd like them to be.

First, all reports for primaries are split by party. As a result, those reports require ballot counts by party. But the tabulation code in `utils` only calculates the overall ballot counts for each group. As a result, we tabulate results by party and then merge the results, retaining the ballot counts. Previously that merging happened on the frontend, now it happens on the backend. We could build ballot counts by party into the initial tabulation, but at the moment I felt like that would be over-rotating on this one requirement.

Second, for each report we need to exclude contests that don't make sense to appear on that report. #3861 added the ability to do this with the VxAdmin store. In this PR, I shift to calculating the contest list on the backend and passing it to the frontend. Doing this allowed me to remove some of the code I couldn't remove in #3861 that I didn't particularly trust. I think there's also a case for moving contest filtering up into the initial tabulation, similar to doing the same for splits in #3865, but not going to act on that yet as it would be a bigger / riskier refactor.

The goal is that all the serious computation is done on the backend and the frontend just renders it, per our preferred architecture.

In the last commit I just generally reorganize the `tabulation` directory in `libs/utils` after all the recent changes.

## Testing Plan

- added quite a bit more automated testing than I removed, for 100% coverage on new code
- manual tested this on a primary election fixture


apps/admin/backend/src/app.results.test.ts
```diff
@@ -104,265 +102,6 @@ test('card counts', async () => {
   );
 });
 
-test('tally report data - overall report & write-ins', async () => {
```
adghayes, line 107: testing removed from this file is moved to `app.tally_reports.test.ts`

apps/admin/backend/src/tabulation/full_results.ts
```diff
@@ -162,66 +160,3 @@ export async function tabulateElectionResults({
   debug('done tabulating election results');
   return groupedElectionResults;
 }
-
-/**
- * Tabulates grouped tally reports for an election. This includes scanned results
- * adjusted with write-in adjudication data (but combining all unofficial write-ins)
- * and manual results separately.
- */
-export async function tabulateTallyReportResults({
```
adghayes, line 171: tally report specific tabulation is moved to `tally_reports.ts`

apps/admin/frontend/src/components/admin_tally_report_by_party.tsx
```diff
@@ -97,33 +83,24 @@ export function AdminTallyReportByParty({
...
-      allElectionResults: partyCompleteTallyResults.map(
-        (r) => r.scannedResults
-      ),
-    });
-    const allManualResults = partyCompleteTallyResults
-      .map((r) => r.manualResults)
-      .filter((r): r is Tabulation.ManualElectionResults => !!r);
-    const combinedManualResults = allManualResults.length
-      ? combineManualElectionResults({ election, allManualResults })
-      : undefined;
```
adghayes, line 126: This is the party results merging - now happening on the backend.

apps/admin/frontend/src/screens/tally_report_screen.tsx
```diff
@@ -363,12 +350,10 @@ export function FullElectionTallyReportScreen(): JSX.Element {
 export function AllPrecinctsTallyReportScreen(): JSX.Element {
   const { electionDefinition, isOfficialResults } = useContext(AppContext);
   assert(electionDefinition);
-  const { election } = electionDefinition;
 
   const reportResultsQuery = getResultsForTallyReports.useQuery({
     groupBy: {
       groupByPrecinct: true,
-      groupByParty: electionHasPrimaryContest(election),
```
adghayes, line 371: Previously it was up to the frontend to tell the backend whether it wanted things grouped by party - now we default to doing that on the backend if it's a primary. Moving it as a default to the backend avoids the footgun of forgetting to do this for new pages that use this query.

kofi-q (approved):
Thanks for the helpful comments on what's moving where/what's changing!

PR #3879, opened 2023-08-24
VxAdmin: Store system settings with election

jonahkagan:
## Overview

System settings are election-specific, since they are imported at the same time (in the election package). Since the backend of VxAdmin supports storing multiple elections, it makes sense to store the system settings with each election.

In making this change, I also thought it made sense for the API to configure the election to reflect this change as well.

## Demo Video or Screenshot
N/A

## Testing Plan
- Updated automated tests
- Manually tested configuring/unconfiguring

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/unconfigured_screen.tsx
+++ b/apps/admin/frontend/src/screens/unconfigured_screen.tsx
@@ -3,11 +3,11 @@ import { useHistory } from 'react-router-dom';
 
 import { electionFamousNames2021Fixtures } from '@votingworks/fixtures';
 import { Button, Font, Icons, P, useMountedState } from '@votingworks/ui';
-import { assert } from '@votingworks/basics';
+import { assert, throwIllegalValue } from '@votingworks/basics';
 
-import type { ConfigureResult } from '@votingworks/admin-backend';
 import { readFileAsyncAsString } from '@votingworks/utils';
 import { DEFAULT_SYSTEM_SETTINGS } from '@votingworks/types';
+import type { ConfigureError } from '@votingworks/admin-backend';
 import { readInitialAdminSetupPackageFromFile } from '../utils/initial_setup_package';
 import {
   getElectionDefinitionConverterClient,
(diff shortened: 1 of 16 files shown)


apps/admin/frontend/src/screens/unconfigured_screen.tsx
```diff
@@ -71,64 +54,56 @@ export function UnconfiguredScreen(): JSX.Element {
...
+          electionData,
+          systemSettingsData,
+        });
+        if (result?.isErr()) {
+          setConfigureError(result.err());
+          // eslint-disable-next-line no-console
+          console.error(
+            'configure failed in saveElectionToBackend',
+            result.err().message
+          );
```
arsalansufi, line 83: Curious what your thinking is around `console.error` logs on the frontend instead of, say, using a frontend `Logger` or backend logging

apps/admin/frontend/src/screens/unconfigured_screen.tsx
```diff
@@ -71,64 +54,56 @@ export function UnconfiguredScreen(): JSX.Element {
...
         }
-      );
+        return result;
+      } catch {
+        // Do nothing
```
arsalansufi, line 87: Since the default React query error handling will kick in, the "Handled by default query client error handling" comment that we typically use might be more informative. While it's true that we're doing nothing in the catch, the current comment might be misleading as the failures here won't actually be silent

arsalansufi (approved):
LGTM!

apps/admin/frontend/src/screens/unconfigured_screen.tsx
jonahkagan: I just maintained what was already there 🤷

apps/admin/frontend/src/screens/unconfigured_screen.tsx
jonahkagan: Thanks, will update!

apps/admin/frontend/src/screens/unconfigured_screen.tsx
arsalansufi: Ah missed that!

PR #3881, opened 2023-08-24
fix(auth): follow the protocol more closely

eventualbuddha:
## Overview
This changes `CardReader::transmit` to record the number of bytes still available to be read and to then request exactly that number of bytes while reading the response.

Some Java cards, like CAC cards, are pretty strict about this. This commit backports some of @benadida's changes made to support CAC cards.

## Demo Video or Screenshot
n/a

## Testing Plan
- [x] verified that installing the applet on a Java card and setting it up as a VxSuite sysadmin card still works

### DIFF
--- a/libs/auth/src/apdu.test.ts
+++ b/libs/auth/src/apdu.test.ts
@@ -25,6 +25,7 @@ test.each<{
     ins: 0x01,
     p1: 0x02,
     p2: 0x03,
+    data: Buffer.of(),
   });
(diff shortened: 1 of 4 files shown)


arsalansufi (approved):
Good to know that some cards/applets are more strict about this - LGTM!

benadida (approved):
assuming this was tested with our Javacards, lgtm!

PR #3884, opened 2023-08-25
lint: add rule to prevent using mutations as dependencies in react hook dep arrays

### LINKED ISSUE #3794: chore(lint): create rule preventing passing `Mutation`s into the dependency array of React hooks
## Summary

`Mutation`s change on every render and this can lead to repeated renders or even cyclical renders. Instead, pass `mutate` or `mutateAsync` to dependency arrays because they have consistent identities.

## Requirements / Tasks
1. Create rule
2. Fix examples in codebase

adghayes:
## Overview

Closes #3794. We've seen a bug or two related to re-render cycles due to mutations changing on every render. The resulting pattern is not cute, but it avoids bugs!

### DIFF
--- a/apps/central-scan/frontend/src/app_root.tsx
+++ b/apps/central-scan/frontend/src/app_root.tsx
@@ -95,6 +95,7 @@ export function AppRoot({
 
   const electionDefinitionQuery = getElectionDefinition.useQuery();
   const unconfigureMutation = unconfigure.useMutation();
+  const unconfigureMutate = unconfigureMutation.mutate;
 
   const [isExportingCvrs, setIsExportingCvrs] = useState(false);
 
@@ -197,8 +198,8 @@ export function AppRoot({
   }, [setStatus]);
 
   const systemAdministratorUnconfigure = useCallback(() => {
-    unconfigureMutation.mutate({ ignoreBackupRequirement: true });
-  }, [unconfigureMutation]);
+    unconfigureMutate({ ignoreBackupRequirement: true });
+  }, [unconfigureMutate]);
 
   const scanBatch = useCallback(async () => {
     setIsScanning(true);
--- a/libs/eslint-plugin-vx/src/rules/no_react_hook_mutation_dependency.ts
+++ b/libs/eslint-plugin-vx/src/rules/no_react_hook_mutation_dependency.ts
@@ -1,0 +1,71 @@
+import { AST_NODE_TYPES, TSESLint, TSESTree } from '@typescript-eslint/utils';
+import { createRule } from '../util';
+
+const HOOKS_WITH_DEPS = [
+  'useEffect',
+  'useLayoutEffect',
+  'useMemo',
+  'useCallback',
+  'useImperativeHandle',
+];
+
+const MUTATION_TYPE_NAME = 'UseMutationResult';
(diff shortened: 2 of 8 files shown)


libs/eslint-plugin-vx/src/rules/no_react_hook_mutation_dependency.ts
```diff
@@ -0,0 +1,71 @@
...
+          }
+
+          const { parserServices } = context.getSourceCode();
+          const tsNodeMap = parserServices.esTreeNodeToTSNodeMap;
+          const typeChecker = parserServices.program.getTypeChecker();
+
+          const mutation = deps.elements.find((element) => {
+            const type = typeChecker.getTypeAtLocation(tsNodeMap.get(element));
+            // istanbul ignore next - unsure how to reproduce aliasSymbol in tests
+            const typeName = type.symbol?.name ?? type.aliasSymbol?.name;
```
adghayes, line 56: I do not understand the TS tree. For whatever reason, the type shows up as `symbol` in the test and as `aliasSymbol` in our actual linting.

apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -197,8 +198,8 @@ export function AppRoot({
   }, [setStatus]);
 
   const systemAdministratorUnconfigure = useCallback(() => {
-    unconfigureMutation.mutate({ ignoreBackupRequirement: true });
-  }, [unconfigureMutation]);
+    unconfigureMutate({ ignoreBackupRequirement: true });
+  }, [unconfigureMutate]);
```
eventualbuddha, line 202: Couldn't this just be `unconfigure.mutate`?

adghayes: You would think, right? The `exhaustive-deps` rule does not allow it in this case.

adghayes: I think the reasoning is that, as a method, `mutate` may refer to `this` which needs to be captured in the dependency array. Which is why I think that sort of property-specific dependency is allowed for non-functions but not functions.

adghayes: https://github.com/facebook/react/issues/16265#issuecomment-517518539

jonahkagan (approved):
Can we also make one that warns against using `useEffect`/`useCallback` at all?

adghayes:
@jonahkagan what's the reason to not use `useCallback`? 

For `useEffect`, it'd be easy to make a rule. I know we've agreed that `useEffect`s are not a good choice in most cases but would you extend that to all cases?

jonahkagan:
What @eventualbuddha said.

For both `useEffect` and `useCallback`, I already avoid them at all costs and only use them in exceptional circumstances. Though as you called out, it's for different reasons:
- `useEffect` is rarely needed since we have safer tools for synchronizing with external systems (e.g. react-query)
- `useCallback` is rarely needed since it's purely for improving performance, which isn't often a blocker.

A lint rule would help externalize those "best practices" into something a bit more tangible. I'm imagining it would be nice to have to explicitly write a comment explaining why I'm bypassing the lint rule to use one of those.

PR #3886, opened 2023-08-25
refactor/test(admin): coverage to 100%

adghayes:
## Overview

We had a conversation a while back about testing practices in which we generally agreed that a 100% coverage approach in which you have to explicitly ignore any uncovered lines is better than a partial coverage approach. I wanted to apply that approach to `admin/backend`.


apps/admin/backend/src/app.ts
```diff
@@ -286,20 +296,21 @@ function buildApi({
...
       try {
-        const settings = store.getSystemSettings();
-        await logger.log(
-          LogEventId.SystemSettingsRetrieved,
-          (await getUserRole()) ?? 'unknown',
-          { disposition: 'success' }
-        );
-        return settings ?? DEFAULT_SYSTEM_SETTINGS;
+        const settings = store.getSystemSettings(electionId);
+        await logger.log(LogEventId.SystemSettingsRetrieved, 'system', {
```
adghayes, line 306: I took the liberty of switching `userRole ?? 'unknown'` to `'system'` for the logging user role. The way that this is used in the app is as a system call, not something triggered by a particular user.

adghayes:
I think @kofi-q makes the most sense as a reviewer because he's our testing lead and has been reviewing my VxAdmin PRs. But since I touch the way we handle system settings, would appreciate it if @jonahkagan or @kshen0 looked at just [003b033](https://github.com/votingworks/vxsuite/pull/3886/commits/003b033900ac411ed3b12506bd5602e7055b9427). I'm inferring how to handle system settings from how the data model is now.

apps/admin/backend/src/app.test.ts
```diff
@@ -186,18 +187,38 @@ test('getSystemSettings happy path', async () => {
...
 
+test('getSystemSettings failure path', async () => {
+  const { apiClient, workspace, logger } = buildTestEnvironment();
+  const { store } = workspace;
+  const {
+    electionDefinition: { electionData },
+  } = electionMinimalExhaustiveSampleFixtures;
+
+  // configure with malformed system settings data, which is the only way for
+  // this path to fail
```
kshen0, line 205: I'm surprised we don't do validation when storing system settings data

apps/admin/backend/src/app.test.ts
```diff
@@ -186,18 +187,38 @@ test('getSystemSettings happy path', async () => {
...
+  const electionId = store.addElection({
+    electionData,
+    systemSettingsData: '{ bad json }',
+  });
+  store.setCurrentElectionId(electionId);
```
kshen0, line 210: Can this fail if we give a nonexistent election ID?

kshen0 (commented):
I'm less familiar with how we get/set system settings after the conversion to JSON blob but from my read of the schema and store methods, this looks correct

apps/admin/backend/src/app.test.ts
adghayes: We do validation at the `app` layer not the `store` layer, which is why I had to inject the bad data at the `store` level in this test.

apps/admin/backend/src/app.test.ts
adghayes: It would fail if we gave it a nonexistent election ID, but that would break most of the other endpoints too. I like to make the assumption that the current election ID always exists, otherwise we end up coding for a bunch of unrealistic failure cases.

adghayes:
Thanks for taking a look @kshen0!

jonahkagan (approved):

jonahkagan:
@adghayes I just reviewed the one commit you asked me to, FYI

apps/admin/backend/src/app.ts
jonahkagan: I actually meant all the logging in this endpoint

adghayes: I caught that, I'm just particularly annoyed with the failure logging. Planning to remove both.

apps/admin/backend/src/tabulation/card_counts.test.ts
```diff
@@ -7,7 +7,10 @@ import {
   MockCastVoteRecordFile,
   addMockCvrFileToStore,
 } from '../../test/mock_cvr_file';
-import { tabulateScannedCardCounts } from './card_counts';
+import {
+  tabulateFullCardCounts,
+  tabulateScannedCardCounts,
+} from './card_counts';
 
 test('tabulateScannedCardCounts - grouping', () => {
```
kofi-q, line 15: Might need to update test case description since we're testing through a different method now

kofi-q (approved):
:shipit:

PR #3887, opened 2023-08-25
Fix L&A Primary Tally Reports

carolinemodic:
Fixes tally reports for L&A Test decks, by pulling in https://github.com/votingworks/vxsuite/pull/2818/commits which fixes the overall tally report in the first commit, and the second commit has a hack that fixes the tally reports by precinct and party. That one is much hackier as those reports are NOT precomputed and generated on the fly in general and it changes it to only "precompute" for test decks but not non-test decks.

### DIFF
--- a/frontends/election-manager/src/components/ballot_counts_table.tsx
+++ b/frontends/election-manager/src/components/ballot_counts_table.tsx
@@ -423,6 +423,8 @@ export function BallotCountsTable({
         </Table>
       );
     }
+    case TallyCategory.PartyPrecinct:
+      throw new Error('unsupported');
     default:
       throwIllegalValue(breakdownCategory);
   }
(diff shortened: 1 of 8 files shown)


adghayes (commented):
I thought that after the first commit, the precinct reports worked as well, so I don't understand why the second commit is necessary.

For the precinct test deck tally reports, every precinct has its own `FullElectionTally`, so in theory the `Party` category should be sufficient to distinguish the different reports.

adghayes (commented):
I think my original PR is hacky, in that it tries to precompute tally categories outside of the tally infrastructure. The additional commit is hacky too, and it makes me nervous that it touches the non-test deck tally code (even though I think that the conditions you place in `filterTallies` make it correct). 

Instead of either commit, can you take a look at #3888? I didn't meant to spend an hour on this but I think the strategy in #3888 is safer and less hacky - just converting test deck ballots to `CastVoteRecords` and then allowing our `precompute` method to handle things.

I checked `electionMinimalExhaustiveSample` and `electionFamousNames` and they looked correct.

carolinemodic:
@adghayes Thank you for looking into that @adghayes !! I assumed that way of fixing it would be much more involved, I just tested your branch locally and the tally reports match the ones I'm getting here and I agree thats the much better approach. I'm going to close this can you open a PR for your changes?

PR #3889, opened 2023-08-28
Add signing to new CVR export function

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3838

This PR adds signing to the new CVR export function for continuous export. Alongside each CVR export directory will now be a `.vxsig` file that contains a signature of the export's top-level metadata file, where that metadata file further contains a hash of all the contents of the CVR export directory. (The hashing logic is coming in its own PR.) To accomplish this, I made a number of changes to artifact auth functions in the auth lib, highlighted in PR comments.

## Testing Plan

- [x] Updated automated tests
- [x] Tested election package export and import manually 

## Checklist

### DIFF
--- a/libs/auth/src/artifact_authentication.ts
+++ b/libs/auth/src/artifact_authentication.ts
@@ -286,7 +286,7 @@ function constructSignatureFilePath(artifact: ArtifactToImport): string {
 }
 
 // eslint-disable-next-line @typescript-eslint/require-await
-async function performArtifactSpecificValidation(
+async function performArtifactSpecificAuthenticationChecks(
   artifact: ArtifactToImport
 ): Promise<void> {
   switch (artifact.type) {
@@ -345,7 +345,7 @@ export async function authenticateArtifactUsingSignatureFile(
       artifact,
       artifactSignatureBundle
     );
-    await performArtifactSpecificValidation(artifact);
+    await performArtifactSpecificAuthenticationChecks(artifact);
   } catch {
     // TODO: Log raw error
     return err(
--- a/libs/auth/src/artifact_authentication.ts
+++ b/libs/auth/src/artifact_authentication.ts
@@ -4,7 +4,7 @@ import fs from 'fs/promises';
 import { sha256 } from 'js-sha256';
 import path from 'path';
 import recursiveReadDir from 'recursive-readdir';
-import { Stream } from 'stream';
+import { Readable, Stream } from 'stream';
 import {
   assert,
   err,
@@ -24,17 +24,56 @@ import {
   verifyFirstCertWasSignedBySecondCert,
   verifySignature,
 } from './openssl';
-import { runCommand } from './shell';
 import { constructPrefixedMessage } from './signatures';
+
+interface CastVoteRecordsToExport {
+  type: 'cast_vote_records';
+  context: 'export';
+  directoryName: string;
+  metadataFileContents: string;
+}
+
+interface CastVoteRecordsToImport {
+  type: 'cast_vote_records';
+  context: 'import';
+  directoryPath: string;
+}
+
+type CastVoteRecords = CastVoteRecordsToExport | CastVoteRecordsToImport;
+
+interface ElectionPackage {
+  type: 'election_package';
+  filePath: string;
+}
+
+interface LegacyCastVoteRecords {
+  type: 'legacy_cast_vote_records';
+  directoryPath: string;
+}
+
+/**
+ * An export-time representation of an {@link Artifact}
+ */
+export type ArtifactToExport =
+  | CastVoteRecordsToExport
+  | ElectionPackage
+  | LegacyCastVoteRecords;
+
+/**
+ * An import-time representation of an {@link Artifact}
+ */
+export type ArtifactToImport =
+  | CastVoteRecordsToImport
+  | ElectionPackage
+  | LegacyCastVoteRecords;
 
 /**
  * A machine-exported artifact whose authenticity we want to be able to verify
  */
-export interface Artifact {
-  type: 'ballot_package' | 'cast_vote_records';
-  /** A file path for ballot packages and a directory path for cast vote records */
-  path: string;
-}
+export type Artifact =
+  | CastVoteRecords
+  | ElectionPackage
+  | LegacyCastVoteRecords;
 
 interface ArtifactSignatureBundle {
   signature: Buffer;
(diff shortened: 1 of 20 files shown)


libs/auth/src/artifact_authentication.ts
```diff
@@ -54,12 +93,13 @@ interface ArtifactSignatureBundle {
  *
  * File paths are sorted alphabetically to ensure a consistent output.
  */
-async function hashDirectoryContents(directory: string): Promise<Buffer> {
-  const filePaths = (await recursiveReadDir(directory)).sort();
+/* istanbul ignore next */
```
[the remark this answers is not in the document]
adghayes: so `hashDirectoryContents` was only for legacy CVRs - for new CVRs we don't have to hash directory contents because we are only hashing the tree's top-level file...

libs/auth/src/artifact_authentication.ts
```diff
@@ -255,10 +345,13 @@ export async function authenticateArtifactUsingSignatureFile(
       artifact,
       artifactSignatureBundle
     );
+    await performArtifactSpecificValidation(artifact);
```
adghayes, line 348: nit: At first glance, it would make more sense to me for "artifact specific validation" to occur outside the main `artifact_authentication` paths, rather than making this code manage the consumers' exceptions. Is there a particular reason to put it here?

libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -321,21 +318,54 @@ async function exportMetadataFileToUsbDrive(
...
+
+  const exportResult = await exporter.exportDataToUsbDrive(
+    // The signature file should live adjacent to the export directory, not within it.
+    path.parse(exportDirectoryPathRelativeToUsbMountPoint).dir,
+    signatureFile.fileName,
+    signatureFile.fileContents
+  );
+  if (exportResult.isErr()) {
+    return exportResult;
+  }
```
adghayes, line 360: I don't have a sense of how this would have been organized when the `auth` lib was writing the file, but it reads very nicely now.

libs/types/src/cast_vote_records.ts
```diff
@@ -20,3 +21,14 @@ export interface CastVoteRecord
...
+
+/**
+ * Metadata stored in the top-level metadata file for a cast vote record export
+ */
+export interface CastVoteRecordExportMetadata {
+  arePollsClosed?: boolean;
+  /** Global data relevant to all cast vote records in an export, e.g. election info */
+  castVoteRecordReportMetadata: CastVoteRecordReport;
+  /** A hash of all cast vote record files in an export */
+  castVoteRecordRootHash: string;
```
adghayes, line 33: I would guess you address this later or somewhere else, but right now not sure why the root hash exists alongside the report metadata rather than entirely separately, because I would think the root hash should include the report metadata.

adghayes (approved):
LGTM. Most of my comments are just me commenting as I read to make sure I'm understanding it. I asked about one organizational nit and one detail of the root hashing but otherwise it looks good, I like the refactor choices you made.

libs/auth/src/artifact_authentication.ts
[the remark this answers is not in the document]
arsalansufi: Yeah that's essentially right! Though just to be extra precise, we're hashing and signing the tree's top-level file, but that top-level file further contains a hash of everything else in the directory. So we're still indirectly incorporating everything into the signature file (which is a must)

libs/auth/src/artifact_authentication.ts
arsalansufi: Ah yeah so the one thing that this function will currently do is recompute the cast vote record root hash to verify that it matches the one in the metadata file. That operation isn't consumer-specific and is in fact a core part of CVR authentication. Without that check, we'd just be verifying that the metadata file hasn't been tampered with. We'd miss any tampering with the many other files in the export.

This comment is making me realizing that this function could use a better name, maybe `performArtifactSpecificAuthenticationChecks`

libs/types/src/cast_vote_records.ts
arsalansufi: Good question! This hash doesn't need to contain the metadata file itself because the metadata file is what's we're signing with OpenSSL. So everything in it is inherently incorporated into the signature.

Side note: When you ask OpenSSL to sign something, it actually hashes whatever you pass to it and then signs that hash. That's why we don't have to pre-hash the metadata file, itself.

That said, I do see why this might cause confusion. I'll think some more about whether it's worth changing in subsequent PRs. One nice property of the current setup is that we can trust the metadata file after a very speedy signature check. If we were to incorporate the metadata file alongside all the other files into the root hash and just sign that hash, we'd have to recompute the root hash (a slower operation) before we could trust the metadata file. There might be some convenience in being able to trust the metadata file more quickly, ahead of the rest of the files in the export.

Getting a bit in the weeds here but we can chat further about this during our sync tomorrow!

adghayes: That explains it. Not everything needs to be hashed - everything just needs to be signed.

libs/auth/src/artifact_authentication.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/3889/commits/39623324179c75d8d1187049e1f328aa88ef9ee0

adghayes: I think the name is fine, I understood the purpose from it. I just was unsure if this was the right place for the hash-checking, both because it applies to only one type of artifact and because right now the import vs. export logic seems asymmetrical. The authenticator is responsible for both creating and checking the signature. And it's responsible for checking the hash, but _not_ creating it, as it needs the hash already created to create the signature file.

I don't see it as problem at all, just observing the patterns as they emerge.

libs/types/src/cast_vote_records.ts
arsalansufi: > Not everything needs to be hashed

^ Exactly! More specifically, not everything needs to be hashed by our code, pre-OpenSSL

Ultimately, even the metadata file is hashed because that's what OpenSSL always does before signing (following the general best practices around signatures)

libs/auth/src/artifact_authentication.ts
arsalansufi: > the import vs. export logic seems asymmetrical

This is a really good observation! I found a way to place the hash creation logic in my next PR in `libs/auth`, even though it ultimately has to be the app (VxScan/VxCentralScan) backend that creates the hashes table. So while the asymmetry will still exist, the majority of the hashing logic for export and import will be colocated.

In terms of my hesitance to move the checking logic out to consumers (which I know you're not suggesting), I just want to make it as hard as possible to forget that check. I want the mental model to be "I've called authenticateArtifact, so I'm good to go".

PR #3893, opened 2023-08-28
Select & Multi-Select Component

adghayes:
## Overview

I'm working toward building a report builder, like roughly seen in this video:

[report-builder.webm](https://github.com/votingworks/vxsuite/assets/37960853/4458944f-89b2-4e40-aeca-2755ad1dad3b)

Due to our interpretation of VVSG requirements around reporting capabilities (e.g. 1.1.9-D, "The voting system must be capable of generating reports that consolidate vote data from selected precincts."), we think that we need to enable reports that might combine multiple, but not all, precincts or ballot styles. For a user to select multiple, but not all, of items in a list, I thought a multi-select was the best route. 

Using `react-select`.

[react-select.webm](https://github.com/votingworks/vxsuite/assets/37960853/589d8458-e168-47cc-a71e-7b5be7af92c8)

I added a storybook story if you'd like to experiment.

## Testing Plan

- visual / manual testing in `storybook` and in context
- automated testing of different configurations

### DIFF
--- a/libs/ui/src/search_select.tsx
+++ b/libs/ui/src/search_select.tsx
@@ -1,0 +1,171 @@
+import { typedAs } from '@votingworks/basics';
+import Select, {
+  components,
+  DropdownIndicatorProps,
+  MultiValueRemoveProps,
+  StylesConfig,
+} from 'react-select';
+import { useTheme } from 'styled-components';
+import { Icons } from './icons';
+
+function DropdownIndicator(
+  props: DropdownIndicatorProps<unknown, true>
+): JSX.Element {
+  return (
+    <components.DropdownIndicator {...props}>
+      <Icons.CaretDown />
+    </components.DropdownIndicator>
+  );
+}
+
+function MultiValueRemove(
+  props: MultiValueRemoveProps<unknown, true>
+): JSX.Element {
+  return (
+    <components.MultiValueRemove {...props}>
+      <Icons.X />
+    </components.MultiValueRemove>
+  );
+}
+
+export interface SelectOption<T extends string = string> {
+  value: T;
+  label: string;
+}
+
+interface SearchSelectBaseProps<T extends string = string> {
+  isMulti: boolean;
+  isSearchable: boolean;
+  options: Array<SelectOption<T>>;
+  ariaLabel?: string;
+}
+
+export interface SearchSelectMultiProps<T extends string = string>
+  extends SearchSelectBaseProps<T> {
+  isMulti: true;
+  value: T[];
+  onChange: (values: T[]) => void;
+}
+
+export interface SearchSelectSingleProps<T extends string = string>
+  extends SearchSelectBaseProps<T> {
+  isMulti: false;
+  value?: T;
+  onChange: (value?: T) => void;
+}
+
+export type SearchSelectProps<T extends string = string> =
+  | SearchSelectSingleProps<T>
+  | SearchSelectMultiProps<T>;
+
+export function SearchSelect<T extends string = string>({
+  isMulti,
+  isSearchable,
+  options,
+  value,
+  onChange,
+  ariaLabel,
+}: SearchSelectSingleProps<T> | SearchSelectMultiProps<T>): JSX.Element {
+  const theme = useTheme();
+
+  return (
+    <Select
+      isMulti={isMulti}
+      isSearchable={isSearchable}
+      isClearable={false}
+      options={options}
+      defaultValue={value}
+      onChange={onChange}
+      placeholder={null}
+      aria-label={ariaLabel}
+      unstyled
+      components={{ DropdownIndicator, MultiValueRemove }}
+      styles={typedAs<StylesConfig>({
+        container: (baseStyles) => ({
+          ...baseStyles,
+          lineHeight: theme.sizes.lineHeight,
+        }),
+        control: (baseStyles) => ({
+          ...baseStyles,
+          border: `solid ${theme.colors.foreground} ${theme.sizes.bordersRem.thin}rem`,
+          borderRadius: '0.25rem',
+          backgroundColor: theme.colors.background,
+          minHeight: isMulti ? '2.2rem' : undefined,
+        }),
+        valueContainer: (baseStyles) => ({
+          ...baseStyles,
+          margin: `0.25rem`,
+          gap: '0.2rem',
+        }),
+        singleValue: (baseStyles) => ({
+          ...baseStyles,
+          fontWeight: theme.sizes.fontWeight.semiBold,
+        }),
+        multiValue: (baseStyles) => ({
+          ...baseStyles,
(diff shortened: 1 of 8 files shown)


libs/ui/src/search_select.tsx
```diff
@@ -0,0 +1,342 @@
...
+}
+
+const ComponentContainer = styled.div`
+  width: 100%;
+  position: relative;
+`;
+
+const SelectBar = styled.div`
+  width: 100%;
+  min-height: 2.2rem;
```
adghayes, line 22: I just manually adjusted this so that, in multi-select mode, the addition of options wouldn't cause the box's size to noticeably change. It can fit a single row of options.

libs/ui/src/search_select.tsx
```diff
@@ -0,0 +1,342 @@
...
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 0.1rem;
+  margin: 0.2rem 0.3rem;
+  border-radius: 0.25rem;
+  overflow: hidden;
+`;
+
+const INPUT_CONTAINER_GRID_AREA = '1 / 1 / 2 / 3';
+const INPUT_GRID_AREA = '1 / 2 / auto / auto';
```
adghayes, line 46: The problem of a multi-select is that you need an `input` component that's going to resize as you type, going to the next line if necessary (forgot to demo this in video but it's neat!). But `input` components normally don't resize.

Here's how `react-select` handles this while not totally running afoul of semantic HMTL. Basically, we create a grid and snap multiple elements to the _same_ grid-area, which allows the `input`s to resize according to another element, even though `input`s normally do not resize.

`INPUT_CONTAINER_GRID_AREA` is for, in the single select case, colocating the current value and the input.

`INPUT_GRID_AREA` is for, in the multi-select case`, resizing the `input` according to a computed `::after` element of the `InputContainer` that is based on the current search string's length.

libs/ui/src/search_select.tsx
```diff
@@ -0,0 +1,342 @@
...
+const SelectedMultiValueText = styled.div`
+  padding: ${(p) => {
+    const thickRem = p.theme.sizes.bordersRem.thick;
+    return `${thickRem}rem ${thickRem}rem ${thickRem}rem ${thickRem * 2}rem`;
+  }};
+  white-space: nowrap;
+  font-weight: ${(p) => p.theme.sizes.fontWeight.semiBold};
+`;
+
+const SelectedMultiValueRemove = styled.div`
```
kofi-q, line 71: we should probably use `styled.button` here, since it's playing the role of a button (or at least assign it a `role="button"` attribute and give it a `tabindex="0"` so it's properly accessible)

libs/ui/src/search_select.tsx
```diff
@@ -0,0 +1,342 @@
...
+        <ValuesContainer isMulti={isMulti}>
+          {isMulti &&
+            currentValues.map((value) => {
+              const option = optionLookup[value];
+              return (
+                <SelectedMultiValue key={option.value}>
+                  <SelectedMultiValueText>
+                    {option.label}
+                  </SelectedMultiValueText>
+                  <SelectedMultiValueRemove
```
kofi-q, line 283: Since this is an icon-only button, worth assigning an `aria-label={Remove ${option label}}` - also makes it easier to write a more semantic test with `getButton("Remove foo")`, when combined with the note above about making this a `<button>` or `<div role="button">`.

Trying not to make assumptions about shared components only being used in a kiosk-browser context (where native web semantics don't matter as much), since we might have stuff like VxDesign hosted in a browser-accessible env

adghayes:
@kofi-q thanks for the feedback! I don't have a cutoff for bundle size, `react-select` just seemed like it had its own language esp. in terms of styling that I didn't want to get into. Also - when I tried to use `react-select` the typing just straight up doesn't work. I don't know if I'm doing something wrong or not, but the type definitions don't populate and the props show up as `any`. Supposedly v5 is written in native Typescript so the props should all be there - either I'm importing it incorrectly or it's actually badly packaged.

I'm giving it another go now, but still having issues with the typing. If you have any idea what my problem might me, lmk!

kofi-q:
Hm, yeah if it's not a straightforward integration out the box, I agree we should just go with our own component for now. Been bitten enough times by `react-select` that I'd err on the side of not using it 😅 
We can backfill the a11y pieces here if and when we need them in a web context

adghayes:
I think it's actually pretty straightforward if we allow a little gap in our type system (the props of the `Select` component exported by `react-select` are `any`, but we can type specific arguments with types they export for ease). I'm playing with it now and if I can get it spun up in a couple hours I think it'll be worth using the package. Sound OK?

adghayes:
Hey @kofi-q, this is ready for your re-review. Now it uses `react-select`. It gets us a lot of stuff for free, and I was able to insert the existing styles pretty easily. The problem is that the `Select` component exported by `react-select` has its `props` as type `any` for some reason. For individual props, type bindings are exported by the library so I was able to use them for the most complicated type, `styles`. Otherwise the types are pretty simple, and on the wrapper component that this PR creates accesses the `react-select` component in this type-lax way.

adghayes:
@kofi-q I did make one styling change - instead of using an icon to indicate focus in the menu options, I use a dotted border. I think this works well except on the first and last child, where it's a little less clear, but I like how it fits our contrast ratios and doesn't look too out of place.

libs/ui/src/search_select.tsx
```diff
@@ -0,0 +1,342 @@
...
+      {isShowingMenu && (
+        <Menu>
+          {displayedOptions.length === 0 ? (
+            <MenuNoOption>No options</MenuNoOption>
+          ) : (
+            displayedOptions.map((option) => {
+              return (
+                <MenuOption
+                  key={option.value}
+                  tabIndex={-1}
```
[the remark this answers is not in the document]
adghayes: handled by `react-select`

libs/ui/src/search_select.tsx
adghayes: handled by `react-select`, it's now a button and `aria-label` is generated by default

libs/ui/src/search_select.tsx
adghayes: handled by `react-select`

libs/ui/src/search_select.tsx
```diff
@@ -0,0 +1,342 @@
...
+}
+
+export function SearchSelect<T extends string = string>({
+  isMulti,
+  isSearchable,
+  options,
+  defaultValue,
+  onSelectionChange,
+}: SearchSelectProps<T>): JSX.Element {
+  const [currentValues, setCurrentValues] = useState<T[]>(
```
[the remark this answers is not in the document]
adghayes: now it is a controlled component

kofi-q (approved):
Glad that worked out - thanks for the extra work getting it hooked up! Everything LGTM

PR #3902, opened 2023-08-30
Use yes/no option ids for votes

jonahkagan:
## Overview

Follow up to https://github.com/votingworks/vxsuite/pull/3864

It turns out I missed a few spots where there were still `'yes'/'no'` contest option ids being used, specifically in places where we were interfacing with the `YesNoVote` type, which used these same string literals. To prevent the risk of this kind of mistake being made further, this PR changes the `YesNoVote` type to also use an explicit option ID. This ensures that the same option ID is being used throughout the system consistently.

_Review by commits_

## Testing Plan
- Updated automated tests
- Grepped for 'yes'/'no' strings

## Checklist


adghayes (approved):
LGTM, thank you for doing this.

PR #3906, opened 2023-08-30
VxDesign: Ballot types and ballot modes

jonahkagan:
## Overview

Adds support for ballot types (precinct/absentee) and ballot modes (official/test/sample) to VxDesign. You can try out different type/mode combos on the ballot preview screen. Exporting all ballots now generates one ballot of each ballot type/mode combo (which makes it 6x slower, but that's a later problem).

_Review by commit_

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/b9716c9a-844c-4368-86df-728816c070c0

## Testing Plan
- Updated automated tests
- Manual testing of the ballot viewer
- Visual confirmation of ballot layout (including different density settings)

## Checklist


apps/design/frontend/src/contests_screen.tsx
jonahkagan, line 1: Recommend hiding whitespace for this commit

apps/design/frontend/src/ballot_viewer.tsx
```diff
@@ -305,21 +311,51 @@ export function BallotViewer({
...
+                value={ballotType}
+                onChange={setBallotType}
+              />
+            </FormField>
+
+            <FormField label="Tabulation Mode">
+              <RadioGroup
+                options={[
+                  { value: 'official', label: 'Official Ballot' },
+                  { value: 'test', label: 'L&A Test Ballot' },
```
arsalansufi, line 346: Nice! I like the inclusion of "L&A" in this label. Makes the distinction between test and sample ballots extra clear (I'm sure this distinction is already be clear to election officials but still helpful for us internally)

apps/design/frontend/src/ballot_viewer.tsx
```diff
@@ -305,21 +311,51 @@ export function BallotViewer({
...
+            <FormField label="Timing Mark Grid">
+              {grid.columns} columns x {grid.rows} rows
+              <div style={{ marginTop: '0.25rem' }}>
+                <label style={{ cursor: 'pointer' }}>
+                  <input
+                    type="checkbox"
+                    checked={showGridLines}
+                    onChange={() => setShowGridLines(!showGridLines)}
+                  />{' '}
+                  Show grid lines
```
arsalansufi, line 363: I see that we've switched to a checkbox here, which makes a lot sense!

spacing

I do think that the styling feels a bit incongruous with the radio buttons above - I've typically seen design systems style radio buttons and checkboxes similarly, at least in terms of spacing.

Not blocking to tweak the checkbox styling, just wanted to share an opinion!

apps/design/backend/src/app.ts
```diff
@@ -139,43 +144,63 @@ function buildApi({ store }: { store: Store }) {
...
-          ' ',
-          '_'
-        )}-${ballotStyleId}.pdf`;
-        zip.file(fileName, pdf);
-        pdf.end();
+      const ballotTypes: Array<[BallotType, string]> = [
+        [BallotType.Standard, 'precinct'],
+        [BallotType.Absentee, 'absentee'],
+      ];
+      const ballotModes: BallotMode[] = ['official', 'test', 'sample'];
```
arsalansufi, line 154: One technique that I've seen used to guarantee that an array like this is exhaustive is as follows:

```ts
const BALLOT_MODES = ['official', 'test', 'sample'] as const;
type BallotMode = typeof BALLOT_MODES[number];
```

Basically define the exhaustive array first and then define the type from that

arsalansufi (approved):
Looks great!

apps/design/frontend/src/ballot_viewer.tsx
jonahkagan: Yeah agreed, this is a bit of laziness on my part because I don't think "show grid lines" will stick around very long. Once we move to letting bubbles be placed anywhere instead of only on grid intersections, it won't be an important feature anymore. It's really only been needed to help me when building the layout algorithm.

apps/design/frontend/src/ballot_viewer.tsx
jonahkagan: Yes, I was confused about it for a while 😆

apps/design/backend/src/app.ts
jonahkagan: Good suggestion, thanks! d1054c3fd

apps/design/frontend/src/ballot_viewer.tsx
arsalansufi: Got it, thanks for that context!

PR #3907, opened 2023-08-31
Change BallotType enum to string, fix related bugs

jonahkagan:
## Overview

While working on #3906, I noticed that the interpreter wasn't actually parsing the ballot type correctly due to a misordering of the enum variants. So I decided to change the enum to use string values instead. It turned out there were a couple of other bugs hidden in there as well:
- The interpreter was returning string ballot type values instead of numbers
- I forgot to encode the correct ballot type in the QR code and was still using a previous hardcoded value

This PR fixes the bugs, changes the enum to use string values for future clarity, and adds more comprehensive testing to the metadata parsing. Lastly, it renames `BallotType.Standard` to `BallotType.Precinct` to be more semantic and consistent with CDF.

_Review by commits_

## Demo Video or Screenshot
N/A

## Testing Plan
Automated tests

## Checklist


apps/admin/backend/test/mock_cvr_file.ts
jonahkagan, line 1: This commit just used find and replace with a little human discernment sprinkled in

adghayes (approved):
LGTM. Appreciate the renaming to "Precinct", that's always been confusing.

PR #3911, opened 2023-08-31
perf(auth): read file data as needed

eventualbuddha:
## Overview
Some perf improvements to #3898.

Rather than reading all the file data for a given directory upfront, read it at the point where we actually need it to compute the hash. Additionally, we don't need to read all the file data for a given file all at once. It can be streamed while we compute the hash incrementally.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests.


arsalansufi (approved):
This is really great, thank you Brian! Had it in the back of my mind that we should eventually stream but had been punting because there's no longer a giant CVR file that _has_ to be streamed. That said, even with smaller files, I'm sure streaming will improve perf

PR #3913, opened 2023-09-01
factor out mutex pkg

kshen0:
## Overview
Factors out `Mutex` from `libs/custom-scanner` for use in other packages.

## Demo Video or Screenshot
N/A

## Testing Plan
Moved mutex tests to mutex package. No new tests added.

## Checklist


libs/mutex/jest.config.js
```diff
@@ -0,0 +1,16 @@
...
+ */
+module.exports = {
+  ...shared,
+  collectCoverageFrom: [
+    ...shared.collectCoverageFrom,
+    '!src/**/index.ts',
+    '!src/cli/**/*.ts',
+    '!src/mocks/**/*.ts',
+    '!src/types/**/*.ts',
+  ],
```
adghayes, line 14: could remove / simplify this

libs/mutex/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@votingworks/mutex",
+  "version": "1.0.0",
+  "description": "Generic mutex package for controlling access to a shared resource.",
+  "keywords": [],
+  "license": "AGPL-3.0",
+  "author": "",
+  "main": "build/index.js",
+  "scripts": {
+    "analyze-packets": "cd tools/analyze-packets && vite",
```
adghayes, line 10: could probably remove this script

adghayes (commented):
Is there a compelling reason that this needs to be its own package rather than exist in `basics` or `utils`? With what I know, I think this would be better there.

Because of the way our CircleCI works, this will spin up an entire other instance on every run just to run the couple tests in `mutex.test`. And it makes our dependency tree just a little more complicated.

kshen0:
Fair point, I'll move it to utils and that should also address the config bloat you commented on.

adghayes (approved):
Thanks for making the adjustment!

libs/custom-scanner/src/custom_a4_scanner.ts
```diff
@@ -569,7 +568,7 @@ export class CustomA4Scanner implements CustomScanner {
   /**
    * Resets the hardware.
    */
-  async resetHardware(): Promise<Result<void, ErrorCode>> {
+  resetHardware(): Promise<Result<void, ErrorCode>> {
```
kshen0, line 571: I'm not sure why these functions were declared as `async` when they don't have any `await` expressions. Our linter prohibits this so I went ahead and removed the `async` declarations.

PR #3915, opened 2023-09-05
Have VxScan use new export format for full exports

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue links: https://github.com/votingworks/vxsuite/issues/3914, https://github.com/votingworks/vxsuite/issues/3916

This PR has VxScan use the new export format for full exports, when the continuous export feature flag is enabled. It also has VxScan update the export metadata on polls close (to indicate that polls have been closed).

## Testing

- [x] Added/updated automated tests
- [x] Tested manually

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Will revisit logging in later PRs

### DIFF
--- a/libs/backend/src/scan/cast_vote_records/export.ts
+++ b/libs/backend/src/scan/cast_vote_records/export.ts
@@ -76,11 +76,17 @@ interface ScannerStateUnchangedByExport {
   pollsState?: PollsState;
 }
 
+interface ExportOptions {
+  arePollsClosing?: boolean;
+  isFullExport?: boolean;
+}
+
 /**
  * A grouping of inputs needed by helpers throughout this file
  */
 interface ExportContext {
   exporter: Exporter;
+  exportOptions: ExportOptions;
   scannerState: ScannerStateUnchangedByExport;
   scannerStore: ScannerStore;
 }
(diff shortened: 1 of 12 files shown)


libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -404,14 +429,20 @@ export async function exportCastVoteRecordsToUsbDrive(
     }
     const { castVoteRecordId, castVoteRecordHash } =
       exportCastVoteRecordFilesResult.ok();
+    castVoteRecordHashes[castVoteRecordId] = castVoteRecordHash;
+  }
+
+  for (const [castVoteRecordId, castVoteRecordHash] of Object.entries(
+    castVoteRecordHashes
+  )) {
```
arsalansufi, line 437: An interesting bug here. I had to move the updating of cast vote record hashes out of the result-sheet iterating loop to avoid SQLite "DB client is busy" errors. There was no issue before when I'd only used this function to export one result sheet at a time for continuous export, as the result-sheet iterating query was always finished by the time we ran hash-updating queries.

I have tests later in this PR that will make sure that we don't accidentally regress here.

adghayes: 🐛

adghayes (approved):
LGTM.

Separate from the changes in this PR, I'm unsure if our sequence of events at polls closed is the right one. When we close polls, we wait for the CVRs to export and the reports to print before changing the app state to "closed." I think we did it this way to make sure those events happened. But the results export can be done manually after the fact, as a backup path. We need to be able to print the report after the fact in the same way. But then I think we should just close polls and _then_ do actions. 

I'm thinking specifically of a nit I have with the process, where the batch metadata exported at polls closed actually has the current batch as _not finished_ even though it should be.

arsalansufi:
> the batch metadata exported at polls closed actually has the current batch as not finished even though it should be

@adghayes Ooh, thanks for noting that! Just now seeing CDF's `EndTime` field for batches, and how the field isn't populated by our current automatic export flow. I agree with your take that we should change order of operations. Opened an issue to track: https://github.com/votingworks/vxsuite/issues/3929

PR #3919, opened 2023-09-05
Have VxCentralScan use new export format

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3918

While we're aren't bringing continuous export to VxCentralScan, we do still want to use the new export format on VxCentralScan (multiple CVR JSON files instead of just one) so that VxAdmin only has to worry about one format on import. This PR has VxCentralScan use the new export format, when the continuous export feature flag is enabled.

We've talked about using a slightly modified version of the format on VxCentralScan, where each directory and CVR JSON file corresponds to a batch instead of a single CVR. But I'm not going to make that change until I do some performance testing on the current format. If there's a significant performance benefit to exporting by batch, I'll make changes. Otherwise, the fewer formats, the better.

We've also talked about VxCentralScan only exporting what VxAdmin needs, i.e. skipping images for CVRs without write-ins, with the option to export everything if desired. I plan on introducing that functionality in follow-up PRs. Just focusing on the basics here.

## Testing

- [x] Added/updated automated tests
- [x] Tested manually

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced ➡️ Left existing logging intact

### DIFF
--- a/libs/backend/src/scan/cast_vote_records/export.ts
+++ b/libs/backend/src/scan/cast_vote_records/export.ts
@@ -49,7 +49,7 @@ import {
 import { ResultSheet } from './legacy_export';
 import { buildElectionOptionPositionMap } from './option_map';
 
-type ExportCastVoteRecordToUsbDriveError =
+type ExportCastVoteRecordsToUsbDriveError =
   | { type: 'invalid-sheet-found'; message: string }
   | ExportDataError;
 
@@ -258,7 +258,7 @@ async function exportCastVoteRecordFilesToUsbDrive(
 ): Promise<
   Result<
     { castVoteRecordId: string; castVoteRecordHash: string },
-    ExportCastVoteRecordToUsbDriveError
+    ExportCastVoteRecordsToUsbDriveError
   >
 > {
   const { exporter } = exportContext;
@@ -350,7 +350,7 @@ async function exportMetadataFileToUsbDrive(
   castVoteRecordRootHash: string,
   exportDirectoryPathRelativeToUsbMountPoint: string
 ): Promise<
-  Result<{ metadataFileContents: string }, ExportCastVoteRecordToUsbDriveError>
+  Result<{ metadataFileContents: string }, ExportCastVoteRecordsToUsbDriveError>
 > {
   const { exporter, exportOptions, scannerState } = exportContext;
   const { pollsState } = scannerState;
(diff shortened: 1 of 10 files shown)


libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -76,11 +85,18 @@ interface ScannerStateUnchangedByExport {
...
+  scannerType: 'central';
+}
+
+interface PrecinctScannerOptions {
+  scannerType: 'precinct';
   arePollsClosing?: boolean;
   isFullExport?: boolean;
 }
 
+type ExportOptions = CentralScannerOptions | PrecinctScannerOptions;
```
arsalansufi, line 98: There are ways to get by without a `scannerType` field in the options, but I ultimately found that the `scannerType` field made the code clearer

libs/backend/src/scan/cast_vote_records/export.ts
```diff
@@ -383,33 +434,36 @@ async function exportSignatureFileToUsbDrive(
...
-        const drive = await usbDrive.status();
-        return drive.status === 'mounted' ? [drive] : [];
-      },
+      getUsbDrives:
+        'getUsbDrives' in usbDrive
+          ? usbDrive.getUsbDrives
+          : async () => {
+              const drive = await usbDrive.status();
+              return drive.status === 'mounted' ? [drive] : [];
+            },
```
arsalansufi, line 450: Planning to consolidate on the new `@votingworks/usb-drive` `UsbDrive` interface later (as I promised to during the recent re-architecture check-in)

adghayes (approved):
LGTM.

I'm don't understand the note in the description about potentially not requiring images in VxCentralScan exports in the future. Don't we need them for write-in adjudication? Unless we implement proper write-in adjudication for VxCentralScan. Which I thought was _not_ our plan, but may be mistaken here.

arsalansufi:
> I'm don't understand the note in the description about potentially not requiring images in VxCentralScan exports in the future. Don't we need them for write-in adjudication?

@adghayes My bad, I had a typo in my description!

> We've also talked about VxCentralScan only exporting what VxAdmin needs, i.e. skipping images for CVRs ~with~ _without_ write-ins, with the option to export everything if desired.

The "with" should've read "without" - just fixed! Hopefully that clears things up

PR #3932, opened 2023-09-05
Kevin/vxmarkscan debt paydown

kshen0:
## Overview
* Removes legacy paper handler class
* Uses `utils/Mutex` to prevent concurrent access to the driver

Seeking feedback on a few design decisions:

**Public API lock**
The public API has its own lock so only one driver method may be executing at any time. This follows `libs/custom-scanner`'s implementation. It felt appropriate to copy this behavior because concurrent driver commands have undefined behavior.

**Locking the whole WebUsbDevice vs locking generic and real-time channels individually**
`libs/custom-scanner` has a channel abstraction and locks per-channel.

The prototype version of the paper handler driver also locked per-channel. The current `libs/custom-paper-handler` implementation locks the entire WebUsbDevice. This means both pairs of channels (pair 1: real-time in/out; pair 2: generic in/out) are locked even though most callers should need only 1 pair. In theory this might lose time in a situation like:
* Caller 1 needs generic endpoints only and locks the WebUsbDevice
* Caller 2 needs real-time endpoints only and must wait for the overly-general WebUsbDevice lock

but in practice I haven't found a situation that matches. The manual implies commands use _either_ the generic endpoints or the real-time endpoints, but no commands use both [0]. This assumption along with the public API lock means we should never need to simultaneously access the real-time and generic channels. So by locking the entire WebUsbDevice we're trading off some potential for performance optimization for simpler code and logic.

[0]
> The scanner is configured as vendor specific device, with 4 bulk endpoints. The first pair of endpoints is used to communicate with generic commands. The second pair of endpoints is used to communicate with real time commands. The 2nd pair of endpoints in this manual is also called the second interface.

## Checklist


libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -140,83 +140,110 @@ export async function getPaperHandlerWebDevice(): Promise<
...
   }
 
   /**
-   * Should be private, but exposed for development.
+   * Send commands or data on the generic bulk out endpoint.
+   * Expects the caller to acquire the lock for and pass in the WebUsbDevice.
    */
-  getWebDevice(): MinimalWebUsbDevice {
-    return this.webDevice;
+  transferOutGeneric<T>(
```
kshen0, line 177: `transferOutGeneric` and `transferInGeneric` are used in a few different ways by callers, so we leave the locking up to the caller and require the WebUsbDevice to be passed into these functions. The `Mutex` package encourages good locking practices in callers because `.withLock` is called to get a reference to `webDevice`, but that usage is not enforced, so it's possible to regress if a direct reference to `webDevice` is added somewhere.

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -225,37 +252,20 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
...
   ): Promise<Result<T, CoderError>> {
-    await this.realTimeLock.acquire();
-
-    const transferOutResult = await this.transferOutRealTime(requestId);
-    assert(transferOutResult.status === 'ok'); // TODO: handling
-
-    const transferInResult = await this.transferInRealTime();
-    this.realTimeLock.release();
-    assert(transferInResult.status === 'ok'); // TODO: handling
+    return await this.webDeviceMutex.withLock(async (webDevice) => {
```
kshen0, line 255: All real-time commands call `handleRealTimeExchange` ie. none call `transferOutRealTime` or `transferInRealTime` directly. Therefore it's cleanest to handle locking in this helper once rather than with boilerplate in each real-time command.

kshen0:
Closing - blocked by some deadlock issues with my mutex usage and not working on actively. Will reopen when I revisit.

PR #3934, opened 2023-09-05
feat(admin): Tally Report Builder - Export Report PDF

adghayes:
## Overview

Follow up to #3927, adding ability to export reports as PDFs to the report builder. Functionality already exists for other tally pages so it should look familiar.

For the file names, I took the same approach as with the report titling - if there are only two levels of simple filters, generate a name. Otherwise, just call it "custom." We hit "custom" in fewer cases here because the filename represents all reports in a list rather than just one, so the filter is only the `filter` and the `groupBy` is separately encoded in the filename. Whereas for report titles, we try to combine the `filter` and the identity of the current group.

For example, say you have "Absentee" filter and "By Precinct" grouping. The report names might be:

1. Precinct 1 Absentee Ballots Tally Report
2. Precinct 2 Absentee Ballots Tally Report

Whereas the file name _containing_ both those reports will be `absentee-tally-reports-by-precinct.pdf`.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/37960853/b672f1c2-717c-4865-9858-976e43c6fcd5

## Testing Plan
- automated testing added
- re-used the `SaveFrontendFileModal` component, so a lot of cases are already covered
- manual testing

## Checklist

### DIFF
--- a/apps/admin/frontend/src/components/reporting/export_report_pdf_button.test.tsx
+++ b/apps/admin/frontend/src/components/reporting/export_report_pdf_button.test.tsx
@@ -1,0 +1,38 @@
+import { electionMinimalExhaustiveSampleDefinition } from '@votingworks/fixtures';
+import { fakeKiosk } from '@votingworks/test-utils';
+import { renderInAppContext } from '../../../test/render_in_app_context';
+import { ExportReportPdfButton } from './export_report_pdf_button';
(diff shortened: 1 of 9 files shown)


mattroe (commented):
frontend / filenames LGTM

kofi-q (approved):

PR #3936, opened 2023-09-06
support a large top margin on bmd_paper_ballot

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3872

* Adds an optional 1.75in margin to the BMD ballot component so the top of the ballot isn't cut off when presenting the ballot for validation
* Changes the implementation of the print state
  * The old implementation triggered the scan state too early. When printing longer ballots, the paper would advance such that `paperInOutput` would return true in the middle of a print, causing the paper to be pulled into the scanner before printing was done.

## Demo Video or Screenshot

**Before**
IMG_4236

**After**
IMG_4238
IMG_4237

## Testing Plan
* Added test to BMD ballot component

## Checklist

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -372,20 +372,24 @@ export function buildMachine(
         },
       },
       printing_ballot: {
-        invoke: pollPaperStatus(),
-        entry: (context, event) => {
-          // Need to hint to Typescript that we want the 'VOTER_INITIATED_PRINT' event in our union type of events
-          if ('pdfData' in event) {
-            void driverPrintBallot(context.driver, event.pdfData, {});
-          } else {
-            throw new Error(
-              `printing_ballot entry called by unsupported event type: ${event.type}`
-            );
-          }
-        },
-        on: {
-          PAPER_IN_OUTPUT: 'scanning',
-        },
+        invoke: [
+          {
+            id: 'printBallot',
+            src: (context, event) => {
+              // Need to hint to Typescript that we want the 'VOTER_INITIATED_PRINT' event in our union type of events
+              if ('pdfData' in event) {
+                return driverPrintBallot(context.driver, event.pdfData, {});
+              }
+              throw new Error(
+                `printing_ballot entry called by unsupported event type: ${event.type}`
+              );
+            },
+            onDone: {
+              target: 'scanning',
+            },
+          },
+          pollPaperStatus(),
+        ],
       },
       scanning: {
         invoke: [
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -376,13 +376,8 @@ export function buildMachine(
           {
             id: 'printBallot',
             src: (context, event) => {
-              // Need to hint to Typescript that we want the 'VOTER_INITIATED_PRINT' event in our union type of events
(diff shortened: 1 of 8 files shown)


libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -275,7 +281,10 @@ export function BmdPaperBallot({
 
   return withPrintTheme(
     <Ballot aria-hidden>
-      <Header>
+      <Header
+        className={largeTopMargin ? 'large-top-margin' : undefined}
```
jonahkagan, line 285: Any reason to use a class here vs just passing a prop to the `Header` styled component? (that's the pattern I usually see in our code for parameterized styling)

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -372,20 +372,29 @@ export function buildMachine(
...
-        },
-        on: {
-          PAPER_IN_OUTPUT: 'scanning',
-        },
+        invoke: [
+          {
+            id: 'printBallot',
+            src: (context, event) => {
+              // Need to hint to Typescript that we want the 'VOTER_INITIATED_PRINT' event in our union type of events
+              if ('pdfData' in event) {
```
jonahkagan, line 380: You could do `assert(event.type === 'VOTER_INITIATED_PRINT')` to make this a bit more concise. Our `assert` function will hint the type checker

jonahkagan (approved):

libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -51,11 +51,15 @@ const Ballot = styled.div`
...
-const Header = styled.div`
+interface StyledHeaderProps {
+  largeTopMargin?: boolean;
+}
+const Header = styled.div<StyledHeaderProps>`
   display: flex;
   flex-direction: row;
   align-items: center;
   border-bottom: 0.2em solid #000;
+  margin-top: ${(p) => (p.largeTopMargin ? '1.75in' : undefined)};
```
kshen0, line 62: Not sure if this is the right way to fallback, but I explicitly don't want to overwrite margin with `0` in the case where `largeTopMargin` is falsy.

jonahkagan: If it works, then seems fine to me!

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -372,20 +372,19 @@ export function buildMachine(
...
+            id: 'printBallot',
+            src: (context, event) => {
+              assert(event.type === 'VOTER_INITIATED_PRINT');
+              return driverPrintBallot(context.driver, event.pdfData, {});
+            },
+            onDone: {
+              target: 'scanning',
+            },
+          },
+          pollPaperStatus(),
```
kshen0, line 386: nb. pollPaperStatus() is still needed here to inform the top-level handler for `jammed`.

libs/ui/src/bmd_paper_ballot.tsx
kshen0: Didn't know about conditional styling with the `styled-components` lib!

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: [723c854](https://github.com/votingworks/vxsuite/pull/3936/commits/723c854a1a00a702f440201a0feb2151f90fc49b)

PR #3938, opened 2023-09-06
Self-Documenting `.csv` Exports

adghayes:
_probably review by commit_

## Overview

Please read [this Slack conversation](https://votingworks.slack.com/archives/CEL6D3GAD/p1693945053137789) for full context. In short, we want all filters that are applied to a report to be apparent in the `.csv` export. There's an alternate approach I could have chosen (see slack convo) and I'm absolutely open to pushback if you think the alternate is better, but this way means less change for the final exports.

## Testing Plan

Alongside the change to `.csv` files, I basically rewrote all the testing for results `.csv` exports. For the better! More coverage.

### DIFF
--- a/apps/admin/backend/src/app.results_csv_export.test.ts
+++ b/apps/admin/backend/src/app.results_csv_export.test.ts
@@ -10,16 +10,15 @@ import {
 import { tmpNameSync } from 'tmp';
 import { readFileSync } from 'fs';
 import { LogEventId } from '@votingworks/logging';
-import { Tabulation, safeParse, CVR as CVRType } from '@votingworks/types';
+import { Tabulation } from '@votingworks/types';
 import { Client } from '@votingworks/grout';
-import { vxBallotType } from '@votingworks/types/src/cdf/cast-vote-records';
+import { parseCsv } from '../test/csv';
 import {
   buildTestEnvironment,
   configureMachine,
   mockElectionManagerAuth,
 } from '../test/app';
 import { Api } from './app';
-import { modifyCastVoteRecordReport } from '../test/utils';
 
 jest.setTimeout(60_000);
 
@@ -47,28 +46,26 @@ afterEach(() => {
   featureFlagMock.resetFeatureFlags();
 });
 
-async function exportResultsSnippet({
+async function getParsedExport({
   apiClient,
   groupBy,
   filter,
-  numLines,
 }: {
   apiClient: Client<Api>;
   groupBy?: Tabulation.GroupBy;
   filter?: Tabulation.Filter;
-  numLines?: number;
-}): Promise<string[]> {
+}): Promise<ReturnType<typeof parseCsv>> {
   const path = tmpNameSync();
-  const precinctGroupedExportResult = await apiClient.exportResultsCsv({
+  const exportResult = await apiClient.exportResultsCsv({
     path,
     groupBy,
     filter,
   });
-  expect(precinctGroupedExportResult.isOk()).toEqual(true);
-  return readFileSync(path, 'utf-8').toString().split('\n').slice(0, numLines);
+  expect(exportResult.isOk()).toEqual(true);
+  return parseCsv(readFileSync(path, 'utf-8').toString());
 }
 
-it('exports expected results, without splits or filter', async () => {
+it('exports expected results for full election', async () => {
   const { electionDefinition, castVoteRecordReport } =
     electionMinimalExhaustiveSampleFixtures;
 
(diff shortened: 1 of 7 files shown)


apps/admin/backend/src/app.results_csv_export.test.ts
```diff
@@ -194,114 +160,6 @@ it('logs failure if export fails for some reason', async () => {
   );
 });
 
-it('uses appropriate file headers based on group by', async () => {
```
adghayes, line 197: All the testing removed from this file was reproduced in `./exports/csv_results.test.ts`.

apps/admin/backend/test/csv.ts
```diff
@@ -0,0 +1,21 @@
+// eslint-disable-next-line import/no-unresolved
+import { parse } from 'csv-parse/sync';
```
adghayes, line 2: What's nice about using this library is that it automatically validates some things we weren't even checking before, like that all rows in the `.csv` have the same number of columns.

apps/admin/backend/test/csv.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export function streamToString(stream: NodeJS.ReadableStream): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const chunks: string[] = [];
+    stream.on('data', (chunk: string) => chunks.push(chunk));
+    stream.on('end', () => resolve(chunks.join('')));
+    stream.on('error', reject);
+  });
+}
```
adghayes, line 21: This is a 100% Github Copilot written method and I take no responsibility for it.

apps/admin/backend/src/exports/csv_results.ts
```diff
@@ -21,135 +22,295 @@ import { ScannerBatch } from '../types';
...
+ * the columns will appear in the CSV.
+ */
+const METADATA_ATTRIBUTES = [
+  'precinct',
+  'party',
+  'ballotStyle',
+  'votingMethod',
+  'scanner',
+  'batch',
+] as const;
```
adghayes, line 36: You may be thinking, or have thought before - we're reproducing and converting this list of attributes all over the place, maybe there's a refactor that we could do to reduce some of the boilerplate and even avoid bugs? 

And I think that's definitely true.

But, right now all of the filters or IDs are just strings, which would make generic typing easy (except that `votingMethod` is a restricted string). But that might not be the case for very long. We might need to describe geographical filters or groupings with additional level identifiers (e.g. "precinct" vs. "ward" vs. "county"). Or we might need to group voting method in multiple overlapping ways (e.g. in-person vs. mail or early voting in-person vs. election day in-person).

So, I'm punting on refactoring and am currently leaving a lot of this annoying boilerplate because I worry something more streamlined will restrict us in the near future.

apps/admin/backend/src/exports/csv_results.ts
```diff
@@ -21,135 +22,295 @@ import { ScannerBatch } from '../types';
...
 }: {
+  filter: Tabulation.Filter;
   groupBy: Tabulation.GroupBy;
-  isPrimaryElection: boolean;
+}): MetadataStructure {
+  const filterStructure: MetadataStructure = {
+    precinct: filter.precinctIds
+      ? filter.precinctIds.length > 1
+        ? 'multi'
+        : 'single'
```
adghayes, line 74: The case where `filter.precinctIds.length === 0` is irrelevant because in that case there would be no report, because no records match that filter.

apps/admin/backend/test/csv.ts
kofi-q: big fan of this change 🎉

apps/admin/backend/src/exports/csv_results.ts
kofi-q: helpful context - thanks!

apps/admin/backend/src/exports/csv_results.ts
```diff
@@ -21,135 +22,292 @@ import { ScannerBatch } from '../types';
...
   }
 
+  // Multi Attributes
+  // -------------------
+
+  if (metadataStructure.precinct === 'multi') {
+    values.push(
+      assertDefined(filter.precinctIds)
+        .map((id) => getPrecinctById(electionDefinition, id).name)
+        .join(',')
```
kofi-q, line 276: nit: since these are meant to be human-readable, should we add a space after the commas for readability? might wrap more naturally in CSV viewers/editors too

kofi-q (approved):
what an interesting little problem! solution sounds and looks good to me :shipit:

apps/admin/backend/src/exports/csv_results.ts
adghayes: Good point! Will do.

adghayes: [ac44edc](https://github.com/votingworks/vxsuite/pull/3938/commits/ac44edc3c943e3da96876aadfee5b89fa7df196c)

PR #3941, opened 2023-09-07
admin: Ballot Count Report Builder Backend

adghayes:
## Overview

We already have code to generate ballot counts, but the original use case was only to make the precinct or voting method breakdowns that appear on the reports page. It supported grouping, but not filtering.

We'll be creating a "Ballot Count Report Builder" similar to the "Tally Report Builder," which will require filtering of results. This PR adds that capability to the VxAdmin backend, leaving the frontend untouched for now.

### DIFF
--- a/apps/admin/backend/src/tabulation/card_counts.ts
+++ b/apps/admin/backend/src/tabulation/card_counts.ts
@@ -1,12 +1,13 @@
 import { Id, Tabulation } from '@votingworks/types';
-import { assertDefined } from '@votingworks/basics';
 import {
   GROUP_KEY_ROOT,
   getEmptyCardCounts,
   getGroupKey,
+  groupBySupportsZeroSplits,
   isGroupByEmpty,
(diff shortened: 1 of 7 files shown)


apps/admin/backend/src/tabulation/card_counts.ts
```diff
@@ -74,8 +80,22 @@ export function tabulateScannedCardCounts({
...
+      electionId,
+      groupBy,
+      filter,
+    });
+    for (const expectedGroup of expectedGroups) {
+      cardCountsGroupMap[getGroupKey(expectedGroup, groupBy)] =
+        getEmptyCardCounts();
+    }
+  }
+
```
adghayes, line 96: determining the expected groupings is necessary so we that expected 0 counts show up on reports.

adghayes:
Sorry @kshen0, forgot to set reviewer to @kofi-q!

kofi-q (approved):
LGTM!

PR #3942, opened 2023-09-07
VxDesign: Primary elections

jonahkagan:
## Overview

Adds basic support for primary elections to VxDesign, including:
- A field to select either a "general" or "primary" election type
- Only showing the field to add party associations to contests for primary elections
- Generating party-specific ballot styles
- Displaying the party associated with each ballot style
- Displaying the party name on primary election ballots

_Review by commits_

## Demo Video or Screenshot
(sound on!)

https://github.com/votingworks/vxsuite/assets/530106/278adfa0-60ad-4eaa-a431-519e2b9f31cb

## Testing Plan
- Updated automated tests
- Visual confirmation of ballot design

(Note: since I don't have basic test coverage for a lot of VxDesign from the prototyping phase, I'm trying to add it as I improve various features. I did that for most of this PR, but ran out of time to cover the contests screen)

## Checklist

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -22,7 +22,7 @@ import JsZip from 'jszip';
 import { renderDocumentToPdf } from '@votingworks/hmpb-render-backend';
 import { ElectionRecord, Precinct, Store } from './store';
 
-function createBlankElection(): Election {
+export function createBlankElection(): Election {
   return {
     type: 'general',
     title: '',
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -11,7 +11,7 @@ import {
   BallotType,
 } from '@votingworks/types';
 import express, { Application } from 'express';
-import { assertDefined, find, ok, Result } from '@votingworks/basics';
+import { assertDefined, find, groupBy, ok, Result } from '@votingworks/basics';
 import {
   BallotMode,
   BALLOT_MODES,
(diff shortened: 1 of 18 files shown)


adghayes:
Exciting! 

When creating a new contest, the party affiliation should not default to anything but require you to make a selection even if it's "No Party Affiliation." This may already be the case but I didn't see contest creation from scratch in the video so I'm not sure.

I really like how the party name is presented simply and clearly at the top of the ballot.

apps/design/backend/src/app.ts
```diff
@@ -49,9 +49,19 @@ function createBlankElection(): Election {
 // precincts to have splits based on the ballot styles.
 export function convertVxfPrecincts(election: Election): Precinct[] {
   return election.precincts.map((precinct) => {
-    const ballotStyles = election.ballotStyles.filter((ballotStyle) =>
+    const precinctBallotStyles = election.ballotStyles.filter((ballotStyle) =>
       ballotStyle.precincts.includes(precinct.id)
     );
+    // Since there may be multiple ballot styles for a precinct for different parties, we
+    // dedupe them based on the district IDs when creating splits.
```
arsalansufi, line 56: Having a little bit of trouble grok-ing this comment, mind clarifying a bit? I think the "for different parties" is throwing me off. How does deduping by district relate to parties? (I don't think it does, I'm probably just parsing this sentence wrong.)

arsalansufi (approved):
The UX looks great!

apps/design/backend/src/app.ts
jonahkagan: Let's say you have a primary election with multiple parties. In a given precinct split, each of those parties will get its own ballot style. Those ballot styles will each have the same list of districts. When creating the ballots, we'll look through those districts' contests and filter down to the appropriate ones for each party.

So to work backwards from ballot styles to precinct splits, we need to ignore the parties and look at the unique groups of districts.

arsalansufi: Okay that does help! I was forgetting that this function is working backwards from ballot styles to precincts. With that context in mind, things make more sense

PR #3943, opened 2023-09-07
Remove candidate party labels on primary election HMP ballots

jonahkagan:
## Overview

Since primary ballots already have a big party label at the top, and all of the candidates should be from the same party by definition, having a party label for each candidate is redundant.

_Review by commit_

## Demo Video or Screenshot
Before
Screen Shot 2023-09-07 at 11 48 43 AM

After
Screen Shot 2023-09-07 at 11 49 12 AM

## Testing Plan
Added new generated fixtures for primary election ballots, then made the change and regenerated the fixtures.

## Checklist


libs/hmpb/render-backend/src/generate_fixtures.ts
```diff
@@ -89,10 +91,33 @@ function generateSampleElectionFixtures(): void {
...
+  }
+}
+
 export function main(): void {
   fs.rmSync(fixturesDir, { recursive: true });
 
   generateAllBubbleBallotFixtures();
   generateFamousNamesFixtures();
   generateSampleElectionFixtures();
+  generatePrimaryElectionFixtures();
```
arsalansufi, line 122: I take it `generateSampleElectionFixtures()` is generating general election fixtures? Might want to rename it `generateGeneralElectionFixtures()` for clarity, since all of these are "sample" elections

libs/hmpb/layout/src/layout.ts
```diff
@@ -882,7 +882,10 @@ function CandidateContest({
   const options: Rectangle[] = [];
   let rowHeightUsed = headingRowHeight;
   for (const candidate of contest.candidates) {
-    const partyText = getCandidatePartiesDescription(election, candidate);
+    const partyText =
+      election.type === 'primary'
+        ? undefined
+        : getCandidatePartiesDescription(election, candidate);
```
arsalansufi, line 888: I'm noticing that the spacing between the last option and the box border when there are no parties is getting to be a little small:

spacing

Could modify the styles to add some space back for this case

arsalansufi (approved):
Makes a lot of sense!

libs/hmpb/layout/src/layout.ts
jonahkagan: Yeah, good catch. I've looked into this before and you have to add a whole timing mark row space to make the bubble alignment for the subsequent contest work out, which also looks weird. So I just left it as is for now until we remove the bubble-on-grid-intersection constraint.

libs/hmpb/render-backend/src/generate_fixtures.ts
jonahkagan: Yes, I totally agree. If I do that renaming, I'd like to rename the base election fixtures as well.

Though I realize that here I did deviate in the naming of this function from the base election fixture name, which is a bit inconsistent.

libs/hmpb/layout/src/layout.ts
arsalansufi: Ah right the timing-mark constraint makes sense - leaving as is is fine then!

libs/hmpb/render-backend/src/generate_fixtures.ts
jonahkagan: Will tackle some renaming in a separate PR

PR #3948, opened 2023-09-09
refactor(admin): use `libs/usb-drive`

### LINKED ISSUE #3446: limit sudo privileges
## Description

In https://github.com/votingworks/vxsuite-complete-system/blob/main/config/sudoers

we've granted sudo permissions to `mkfs.fat` and other filesystem utils to `vx-ui`. That is likely a lot more than we strictly need to do. I'd prefer if we had a wrapper shell script that creates filesystems only on `/media` partitions, for example, and it would be much better if this could be on the server side so only `vx-services` has those limited permissions.

### LINKED ISSUE #3447: Migrate VxAdmin to `libs/usb-drive`
In #3378, @jonahkagan added `libs/usb-drive` and moved VxScan to use it. We should do this for VxAdmin (and the other apps, represented in separate tickets). Note: should close #3446.

adghayes:
_review by commit_

## Overview

Closes #3447, and enables us to close #3446 after a follow-up. Uses `libs/usb-drive` in VxAdmin.

I realized that `libs/usb-drive` doesn't have any logging like the older code. We should add that, I made #3949.

Now that we have `libs/usb-drive`, we should build out the stub for `MockFileUsbDrive` and re-enable, possibly expand, VxAdmin integration tests (#3950).

## Testing Plan
 - Updated tests along the way
 - Manually tested the format USB flow in a lot of different circumstances. Also tested exporting to file from the frontend and backend.

## Checklist


libs/usb-drive/src/format_fat32.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# partition the device with a single FAT32 (type=c) partition
+echo 'type=c' | sfdisk --wipe always --wipe-partitions always $DEVICE
+
+# there may be a delay before the partition can be recognized by the OS, particularly
+# when there is simultaneous polling (e.g. via lsblk). this pause allows for the delay
+sleep 0.01
+
+# format the partition with a FAT32 filesystem, which must be done as root
+sudo mkfs.fat -F 32 -n $LABEL ${DEVICE}1
```
adghayes, line 33: I'm not sure why running the entire script with `sudo` doesn't do the trick, but I had to also preface specific command with `sudo`.

libs/usb-drive/src/mock_usb_drive.ts
```diff
@@ -2,20 +2,23 @@
...
 interface MockDirectory {
   [name: string]: MockFileTree;
 }
+type MockFileTree = MockFile | MockDirectory | ActualDirectory;
 
 function writeMockFileTree(destinationPath: string, tree: MockFileTree): void {
   if (Buffer.isBuffer(tree)) {
     writeFileSync(destinationPath, tree);
+  } else if (typeof tree === 'string') {
+    cpSync(tree, destinationPath, { recursive: true });
```
adghayes, line 21: This is copied from `apps/admin/backend`, and allows copying in entire existing directories more easily (specifically important for cast vote record file fixtures).

libs/usb-drive/src/usb_drive.ts
```diff
@@ -113,8 +161,16 @@ export function detectUsbDrive(): UsbDrive {
...
+  // The block device info blips during formatting, so we need to track the
+  // state and ignore the OS-provided device info during formatting.
+  let isFormatting = false;
+
   return {
     async status(): Promise<UsbDriveStatus> {
+      if (isFormatting) {
+        return { status: 'ejected' };
+      }
+
```
adghayes, line 173: Due to the aforementioned blips, I think that it would be thrashing consumers to actually base status on the `lsblk` received info. So I added this escape hatch, as existed for `useUsbDrive`. Re-using `ejected` rather than a new status is preferable because it involves fewer statuses that consumers have to account for, and `ejected` represents the case pretty well.

libs/usb-drive/src/usb_drive.ts
```diff
@@ -161,5 +217,32 @@ export function detectUsbDrive(): UsbDrive {
...
+        const label = generateVxUsbLabel(deviceInfo.label ?? undefined);
+        await formatUsbDrive(getRootDeviceName(deviceInfo.path), label);
+        debug('USB drive formatted successfully');
+      } catch (error) {
+        debug(`USB drive formatting failed`);
+        throw error;
+      } finally {
+        isFormatting = false;
+        didEject = true; // prevent remount
```
adghayes, line 244: We may want use cases in the future where we want to remount immediately, but none exist now. The only use case is the formatting in the VxAdmin frontend.

libs/usb-drive/src/mock_file_usb_drive.ts
```diff
@@ -0,0 +1,18 @@
+import { UsbDrive, UsbDriveStatus } from './usb_drive';
+
+/**
+ * TODO: Implement properly to enable USB drive operations in integration tests.
+ */
+export class MockFileUsbDrive implements UsbDrive {
```
adghayes, line 6: I'm excited about this as it will allow us to revive and expand `admin` integration tests, but there's already a lot in this PR so punting on implementation anything but `no_drive`.

apps/admin/backend/src/app.cvr_files.test.ts
```diff
@@ -176,6 +177,10 @@ test('happy path - mock election flow', async () => {
 
   expect(await apiClient.getCastVoteRecordFileMode()).toEqual('unlocked');
 
+  const availableCastVoteRecordFiles3 =
+    await apiClient.listCastVoteRecordFilesOnUsb();
+  expect(availableCastVoteRecordFiles3).toMatchObject([]);
+
```
adghayes, line 183: @jonahkagan I needed to add this assertion. Above, there's an `removeUsbDrive` call. Below, there's an `insertUsbDrive` call. Each of those calls set up `expectRepeatedCalls`. But it seems like `expectRepeatedCalls` expects _one or more_ calls, when in this case I think expecting _zero or more_ calls would be more useful. Do you think this would be useful to add to `mockFunction`? It would also cover some of the other cases in these tests where I have to add ` usbDrive.status.expectRepeatedCallsWith().resolves({ status: 'no_drive' });` at the beginning. The `mockUsbDrive` could initialize expecting zero or more calls resolving to `no_drive`.

apps/admin/backend/src/app.ts
```diff
@@ -198,12 +198,26 @@ function buildApi({
...
+      return usbDrive.eject();
+    },
+
+    async formatUsbDrive(): Promise<Result<void, Error>> {
+      try {
+        await usbDrive.format();
+        return ok();
+      } catch (error) {
+        return err(error as Error);
+      }
```
adghayes, line 215: I made this into a `Result` endpoint because, even wrapped in a `try/catch` block on the frontend, errors from `mutateAsync` would go to the error boundary but skip the `catch` block. I didn't realize it worked like that. And I believe the recommendation is to use the `Result` type here anyway.

I could move the `Result` return value down a layer into `libs/usb-drive`, but since this is currently the only consumer it doesn't really matter.

apps/admin/frontend/src/components/export_election_ballot_package_modal_button.tsx
```diff
@@ -77,15 +80,6 @@ export function ExportElectionBallotPackageModalButton(): JSX.Element {
...
           break;
-        case 'ejecting':
-        case 'mounting':
-          mainContent = <Loading />;
-          actions = (
-            <Button onPress={closeModal} disabled>
-              Cancel
-            </Button>
-          );
-          break;
```
adghayes, line 88: Transition states are gettin' scrapped! Less to maintain, this is great.

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -141,8 +141,8 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element | null {
   // React Query pattern, we should invalidate the query on USB drive status
   // change. Currently React Query is not managing USB drive status and polling
   // is being performed elsewhere.
-  useExternalStateChangeListener(usbDrive.status, () => {
-    if (usbDrive.status === 'mounted') {
+  useExternalStateChangeListener(usbDriveStatus.status, () => {
+    if (usbDriveStatus.status === 'mounted') {
       void cvrFilesOnUsbQuery.refetch();
     }
   });
```
adghayes, line 148: @jonahkagan I thought I might have been able to remove this, but not quite. We're still in a position where we're trying to invalidate a query based off of a change in data of another query. I'm not sure what a better way to do this. I thought about `onSuccess`, but that's actually being deprecated for queries and the React Query creator recommends using `useEffect` instead if its unavoidable.

You've previously suggesting the possibility of polling in this case, while mounted.

Open to whatever suggestions here, or just changing the comment. This would become easier if we move away from polling, I am guessing.

apps/admin/frontend/src/components/reporting/export_csv_button.tsx
```diff
@@ -60,7 +60,7 @@ export function ExportCsvResultsButton({
 
   return (
     <React.Fragment>
-      <Button disabled={disabled || !window.kiosk} onPress={openModal}>
+      <Button disabled={disabled} onPress={openModal}>
```
adghayes, line 63: Now the USB drive mocking via the dev dock with even work in the browser.

apps/admin/frontend/src/components/save_backend_file_modal.tsx
```diff
@@ -74,29 +75,19 @@ export function SaveBackendFileModal({
...
 
-  const [usbDrivePath, setUsbDrivePath] = useState<string>();
-
-  useExternalStateChangeListener(usbDrive.status, async (newStatus) => {
-    if (newStatus === 'mounted') {
-      const usbDriveInfo = await usbstick.getInfo();
-      assert(usbDriveInfo);
-      setUsbDrivePath(usbDriveInfo.mountPoint);
-    }
-  });
```
adghayes, line 88: :)

apps/admin/frontend/src/components/save_backend_file_modal.tsx
```diff
@@ -167,10 +149,11 @@ export function SaveBackendFileModal({
                   options={{
                     // Provide a file name and default to the USB drive's root directory.
                     defaultPath: join(
-                      usbDrivePath ?? '',
+                      usbDriveStatus.mountPoint,
                       basename(defaultRelativePath)
                     ),
                   }}
+                  disabled={!window.kiosk}
```
adghayes, line 156: Although the happy path of saving to the USB drive via this modal now _does not_ need `kiosk-browser`, doing "Save As..." still does, as it uses `kiosk-browser` to open the save dialog.

apps/admin/frontend/src/components/save_frontend_file_modal.tsx
```diff
@@ -233,31 +235,26 @@ export function SaveFrontendFileModal({
...
       return (
         <Modal
           title="No USB Drive Detected"
           content={
             <P>
-              <UsbImage
-                src="/assets/usb-drive.svg"
-                alt="Insert USB Image"
-                // hidden feature to save with file dialog by double-clicking
-                onDoubleClick={() => exportResults(true)}
```
adghayes, line 251: My opinion is that we shouldn't have "hidden features" in our open source code base. Even though they are, in that way, not hidden, it may be bad optics like calling mock functions "fake" functions. Also - the same capability is on this same modal anyway...

apps/admin/frontend/test/helpers/mock_usb_drive.ts
```diff
@@ -0,0 +1,26 @@
+import { throwIllegalValue } from '@votingworks/basics';
+
+import type { UsbDriveStatus } from '@votingworks/usb-drive';
+
+export function mockUsbDriveStatus(
```
adghayes, line 5: Copied from VxScan's `fakeUsbDriveStatus`, which I rename in a later commit. Could move this to `test-utils`.

libs/usb-drive/src/format_fat32.sh
adghayes, line 1: @benadida just want to call your attention to this script. Instead of granting `vx-ui` access to `mkfs.fat` and `sfdisk`, it will just get access to this script. Per your issue #3446.

benadida: LOVE IT!

apps/admin/frontend/src/components/save_backend_file_modal.tsx
eventualbuddha: Do we want "Save As…" still?

libs/usb-drive/src/usb_drive.ts
```diff
@@ -161,5 +217,32 @@ export function detectUsbDrive(): UsbDrive {
...
+        debug('No USB drive detected, skipping format');
+        return;
+      }
+
+      isFormatting = true;
+      try {
+        if (deviceInfo.mountpoint) {
+          debug('USB drive is mounted, unmounting before formatting');
+          await unmountUsbDrive(deviceInfo.mountpoint);
+          didEject = true;
```
jonahkagan, line 233: Is this line necessary?

apps/admin/backend/src/util/exporter.ts
```diff
@@ -0,0 +1,33 @@
...
+import { UsbDrive } from '@votingworks/usb-drive';
+import { ADMIN_ALLOWED_EXPORT_PATTERNS } from '../globals';
+
+/**
+ * TODO: Convert Exporter to use libs/usb-drive types.
+ *
+ * Wraps the libs/usb-drive interface in a `getUsbDrives` function
+ * which is the pattern currently expected the libs/backend.
+ */
+export function getGetUsbDrives(
```
eventualbuddha, line 14: I might name this something like `buildGetUsbDrivesFunction` to make it clearer that it's not typo'd from `getUsbDrives`.

apps/admin/backend/src/app.cvr_files.test.ts
jonahkagan: If not sure I understand why there could be 0 or more calls in this case. Is it non-deterministic whether `usbDrive.status` is called?

apps/admin/frontend/src/api.ts
```diff
@@ -141,10 +145,62 @@ export const unprogramCard = {
...
+// USB
+
+export const getUsbDriveStatus = {
+  queryKey(): QueryKey {
+    return ['getUsbDriveStatus'];
+  },
+  useQuery() {
+    const apiClient = useApiClient();
+    return useQuery(this.queryKey(), () => apiClient.getUsbDriveStatus(), {
+      refetchInterval: USB_DRIVE_STATUS_POLLING_INTERVAL_MS,
```
eventualbuddha, line 157: I continue to lament all the polling, but I don't think it makes sense to do differently in this PR.

apps/admin/backend/src/app.ts
jonahkagan: If we're handling these errors explicitly to show messages to users on the frontend, then I agree `Result` is the right move!

jonahkagan: Also if it's true that exceptions thrown from `mutateAsync` don't bubble up in the original call stack but only go to the error boundary, that would be great news for us, because we could stop using `mutate`. The original reason I recommended using `mutate` over `mutateAsync` was to prevent uncaught exceptions from crashing tests with no useful error output, which was nearly impossible to debug and would happen whenever a mock function assertion failed.

Would you be down to explore this a bit more?

libs/usb-drive/src/format_fat32.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+set -euo pipefail
+
+usage () {
+    echo 'Usage: format.sh <device> <label>'
```
eventualbuddha, line 6:
```suggestion
    echo 'Usage: format_fat32.sh <device> <label>'
```

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
jonahkagan: I don't understand the exact requirements for when we want to fetch the CVR files query, but it may be addressable by a dependent query pattern: https://tanstack.com/query/v4/docs/react/guides/dependent-queries

libs/usb-drive/src/usb_drive.ts
```diff
@@ -100,6 +101,53 @@ async function unmountUsbDrive(mountPoint: string): Promise<void> {
...
 }
 
+async function formatUsbDrive(
+  devicePath: string,
+  label: string
+): Promise<void> {
+  debug(`Formatting disk ${devicePath} as FAT32 volume with label ${label}`);
+  await exec('sudo', [
+    '-n',
+    join(MOUNT_SCRIPT_PATH, 'format_fat32.sh'),
```
eventualbuddha, line 111: I realize this isn't a change that this PR introduced, but `MOUNT_SCRIPT_PATH` being in `../src` assumes that this library is running in place and not packaged up using e.g. `pnpm pack`. Maybe that's fine for now, but I do think it's at least worth adding `format_fat32.sh` to the `files` array in `package.json` and possibly moving it to the package root or a different subdirectory.

jonahkagan (approved):
Very exciting!! Thanks for doing this

libs/usb-drive/src/format_fat32.sh
eventualbuddha, line 33: This enables an attacker who can control the contents of `$LABEL` or `$DEVICE` to alter the flags passed to `mkfs.fat`. For example, if I call `format_32.sh /dev/sdb 'a --help'` then I cause `mkfs.fat` to not perform the formatting action but to still exit with status 0. Someone with more time and creative thinking may be able to do worse. To address this, we should always quote such expansions:

```suggestion
sudo mkfs.fat -F 32 -n "${LABEL}" "${DEVICE}1"
```

libs/usb-drive/src/format_fat32.sh
```diff
@@ -0,0 +1,33 @@
...
+    exit 1
+fi
+
+if [[ ${#LABEL} -gt 11 ]]; then
+    echo "error: \"${LABEL}\" has more than the allowed 11 characters for a FAT32 volume label" 
+    exit 1
+fi
+
+# partition the device with a single FAT32 (type=c) partition
+echo 'type=c' | sfdisk --wipe always --wipe-partitions always $DEVICE
```
eventualbuddha, line 26:
```suggestion
echo 'type=c' | sfdisk --wipe always --wipe-partitions always "${DEVICE}"
```

eventualbuddha (commented):
Yay! However, there's an issue that needs to be addressed in the shell script.

apps/admin/backend/src/app.cvr_files.test.ts
adghayes: Because `usbDrive.status` is only called when some `app.getUsbDriveStatus` is called or when it's called from the exporter, so sometimes it's not called if we don't explicitly poll `app.getUsbDriveStatus`.

This issue is non-existent in the frontends because polling is set up, and anytime we run a test the polling will surely hit the calls at least once. But we're not polling from the backend, so if we don't check status explicitly, it never gets called.

adghayes: In other words, it's deterministic but not automatic in the backend case.

apps/admin/backend/src/app.ts
adghayes: OK yeah, I'll give this a try later today or tomorrow and post on Slack what I find.

libs/usb-drive/src/format_fat32.sh
```diff
@@ -0,0 +1,33 @@
...
+
+usage () {
+    echo 'Usage: format.sh <device> <label>'
+    exit 1
+}
+
+ROOT_DEVICE_REGEX=^/dev/sd[a-z]$
+
+DEVICE=$1
+LABEL=$2
```
eventualbuddha, line 13: Also, for a minor quality of life improvement:

```suggestion
DEVICE=${1:-}
LABEL=${2:-}

if [[ -z $DEVICE || -z $LABEL ]]; then
    usage
fi
```

apps/admin/frontend/src/components/save_backend_file_modal.tsx
adghayes: I still use it in dev 🤷 but I wouldn't want to maintain it if it will cost us a lot in the future. It may.

The only reason `Save As...` can work is because in Electron, the returned `File` object from the system dialog has a path. In a regular browser, this doesn't work, because security. So if we are aiming to drop `kiosk-browser`, the replacement would either need to support this or we'd need to remove this.

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
adghayes: I've used dependent queries elsewhere and it satisfies the use case of "wait to run this query until this other query finishes," usually necessary because the inputs of the second query (e.g. a write-in ID) come from the first. 

But in this case we want to invalidate a query when another query is invalidated, which in my reading is not handled by dependent queries.

jonahkagan: I was thinking we'd just `enable` the CVR files query when the usb drive is mounted. Would that give us the behavior we want here?

apps/admin/backend/src/app.cvr_files.test.ts
jonahkagan: Are we able to know if a backend test is going to call `app.getUsbDriveStatus`? Can we just only add the mock expectation if needed?

libs/usb-drive/src/usb_drive.test.ts
```diff
@@ -224,29 +229,112 @@ describe('eject', () => {
...
+    const { promise: formatScriptPromise, resolve: formatScriptResolve } =
+      deferred<{
+        stdout: string;
+      }>();
+    execMock.mockReturnValueOnce(formatScriptPromise); // format
+
+    const formatPromise = usbDrive.format();
+    await new Promise((resolve) => {
+      setTimeout(resolve, 0);
+    }); // wait for format state
```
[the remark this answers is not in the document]
adghayes: I'm trying to remember, but still the case that when you remove this, the test fails.

I think that `format` will yield to the event loop on `getUsbDriveDeviceInfo()` so, if you run `status()` immediately, `format` has not yet set `isFormatting` to `true` so the escape hatch for that status isn't there yet. And that's what we want to test.

libs/usb-drive/src/usb_drive.ts
```diff
@@ -100,6 +101,53 @@ async function unmountUsbDrive(mountPoint: string): Promise<void> {
   await exec('sudo', ['-n', join(MOUNT_SCRIPT_PATH, 'unmount.sh'), mountPoint]);
 }
 
+async function formatUsbDrive(
+  devicePath: string,
+  label: string
+): Promise<void> {
+  debug(`Formatting disk ${devicePath} as FAT32 volume with label ${label}`);
+  await exec('sudo', [
```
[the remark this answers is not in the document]
adghayes: We do, in order to run `sfdisk` and to make the `sudo` within the script pass without giving the user full `sudo` access to either `sfdisk` or `mkfs.fat`, just this script.

libs/usb-drive/src/usb_drive.ts
adghayes: nope!

libs/usb-drive/src/usb_drive.test.ts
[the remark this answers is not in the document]
jonahkagan: Ok, gotcha. Seems like a good case for `waitFor`, though I guess we don't have that in backend tests...

libs/usb-drive/src/usb_drive.ts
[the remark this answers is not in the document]
adghayes: I'm not sure why, but `mkfs.fat` fails if it's not preceded by `sudo`. Even if you run the script with `sudo` but within the script it doesn't hav `sudo` before it, `mkfs.fat` fails. I haven't looked into why.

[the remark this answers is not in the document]
adghayes: Never mind, I'm just wrong. Must have been failing for some other reason when I thought that was the case, because now I can pull out the sudo.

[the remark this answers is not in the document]
adghayes: re: bailing - the script will fail with a "Permission Denied" error at `sfdisk` if it's not run as root, so that should be fine

libs/usb-drive/src/usb_drive.test.ts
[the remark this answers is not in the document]
adghayes: OK, I've added a `backendWaitFor` in 6d77a2090.

libs/usb-drive/src/usb_drive.ts
eventualbuddha: I think putting it under `scripts` is fine for now. `npm pack` is what would be used if we were, say, publishing this library to the npm registry using `npm publish`.

PR #3956, opened 2023-09-12
Update VxAdmin to support importing new CVR export format

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3926

This PR completes the end-to-end wiring of continuous export! It specifically updates VxAdmin to support importing the new CVR export format. The user experience is unchanged by this PR. Just a lot of under-the-hood tweaks.

All of the new logic is feature flag gated. Once I merge this PR, I'll work on pruning the feature flag and old CVR export/import logic.

## Testing

- [x] Updated some unit tests

I've mostly punted on automated tests, which I'll add in follow-up PRs, since this PR is large enough as is.

## Checklist

### DIFF
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -68,7 +68,7 @@ import {
   addCastVoteRecordReport,
   getAddCastVoteRecordReportErrorMessage,
   listCastVoteRecordFilesOnUsb,
-} from './cvr_files';
+} from './legacy_cast_vote_records';
 import { getMachineConfig } from './machine_config';
 import {
   getWriteInAdjudicationContext,
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -32,9 +32,11 @@ import { createReadStream, createWriteStream, promises as fs, Stats } from 'fs';
 import { basename, dirname, join } from 'path';
 import {
   BALLOT_PACKAGE_FOLDER,
+  BooleanEnvironmentVariableName,
   CAST_VOTE_RECORD_REPORT_FILENAME,
   generateFilenameForBallotExportPackage,
   groupMapToGroupList,
+  isFeatureFlagEnabled,
   isIntegrationTest,
   parseCastVoteRecordReportDirectoryName,
 } from '@votingworks/utils';
@@ -48,6 +50,7 @@ import {
   CvrFileMode,
   ElectionRecord,
   ExportDataResult,
+  ImportCastVoteRecordsError,
   ManualResultsIdentifier,
   ManualResultsMetadataRecord,
   ManualResultsRecord,
@@ -86,6 +89,7 @@ import { getOverallElectionWriteInSummary } from './tabulation/write_ins';
 import { rootDebug } from './util/debug';
 import { tabulateTallyReportResults } from './tabulation/tally_reports';
 import { buildExporter } from './util/exporter';
+import { importCastVoteRecords } from './cast_vote_records';
 
 const debug = rootDebug.extend('app');
 
@@ -404,9 +408,39 @@ function buildApi({
     }): Promise<
       Result<
         CvrFileImportInfo,
-        AddCastVoteRecordReportError & { message: string }
+        | (AddCastVoteRecordReportError & { message: string })
+        | ImportCastVoteRecordsError
       >
     > {
+      /* c8 ignore start */
+      if (
+        isFeatureFlagEnabled(
+          BooleanEnvironmentVariableName.ENABLE_CONTINUOUS_EXPORT
+        )
+      ) {
+        const userRole = assertDefined(await getUserRole());
+        const importResult = await importCastVoteRecords(store, input.path);
+        if (importResult.isErr()) {
+          await logger.log(LogEventId.CvrLoaded, userRole, {
+            disposition: 'failure',
+            errorDetails: JSON.stringify(importResult.err()),
+            errorType: importResult.err().type,
+            exportDirectoryPath: input.path,
+            result: 'Cast vote records not imported, error shown to user.',
+          });
+        } else {
+          await logger.log(LogEventId.CvrLoaded, userRole, {
+            disposition: 'success',
+            exportDirectoryPath: input.path,
+            numberOfBallotsImported: importResult.ok().newlyAdded,
+            numberOfDuplicateBallotsIgnored: importResult.ok().alreadyPresent,
+            result: 'Cast vote records imported.',
+          });
+        }
+        return importResult;
+      }
+      /* c8 ignore stop */
+
       const userRole = assertDefined(await getUserRole());
       const { path: inputPath } = input;
       // the path passed to the backend may be for the report directory or the
--- a/apps/admin/backend/src/cast_vote_records.ts
+++ b/apps/admin/backend/src/cast_vote_records.ts
@@ -1,0 +1,283 @@
+/* c8 ignore start */
+import * as fs from 'fs/promises';
+import { sha256 } from 'js-sha256';
+import path from 'path';
+import { v4 as uuid } from 'uuid';
+import { isTestReport, readCastVoteRecordExport } from '@votingworks/backend';
+import { assert, assertDefined, err, ok, Result } from '@votingworks/basics';
+import {
+  BallotId,
+  BallotPageLayoutSchema,
+  CVR,
+  ElectionDefinition,
+  getBallotStyle,
+  getContests,
+  getPrecinctById,
+  safeParseJson,
+} from '@votingworks/types';
+import {
+  BooleanEnvironmentVariableName,
+  castVoteRecordHasValidContestReferences,
+  convertCastVoteRecordVotesToTabulationVotes,
+  isFeatureFlagEnabled,
+} from '@votingworks/utils';
+
+import { Store } from './store';
+import {
+  CastVoteRecordElectionDefinitionValidationError,
+  CvrFileImportInfo,
+  CvrFileMode,
+  ImportCastVoteRecordsError,
+} from './types';
+
+/**
+ * Validates that the fields in a cast vote record and the election definition correspond
+ */
+function validateCastVoteRecordAgainstElectionDefinition(
+  castVoteRecord: CVR.CVR,
+  electionDefinition: ElectionDefinition
+): Result<void, CastVoteRecordElectionDefinitionValidationError> {
+  function wrapError(
+    error: Omit<CastVoteRecordElectionDefinitionValidationError, 'type'>
+  ): Result<void, CastVoteRecordElectionDefinitionValidationError> {
+    return err({ ...error, type: 'invalid-cast-vote-record' });
+  }
+
+  const { election, electionHash } = electionDefinition;
+
+  if (
+    castVoteRecord.ElectionId !== electionHash &&
+    !isFeatureFlagEnabled(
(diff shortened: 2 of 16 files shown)


libs/utils/src/filenames.test.ts
```diff
@@ -600,3 +569,110 @@ test('generateLogFilename', () => {
...
+    expectedDirectoryName: 'TEST__machine_3unicorn3__2023-08-16_17-02-24',
+  },
+])(
+  'generateCastVoteRecordExportDirectoryName',
+  ({ input, expectedDirectoryName }) => {
+    expect(generateCastVoteRecordExportDirectoryName(input)).toEqual(
+      expectedDirectoryName
+    );
+  }
+);
```
arsalansufi, line 608: This `test.each` is just moved down from above. The one below is new.

libs/utils/src/filenames.ts
```diff
@@ -326,3 +310,62 @@ export function generateLogFilename(
...
+    'machine',
+    maybeParse(MachineId, machineId) ?? sanitizeStringForFilename(machineId),
+  ].join(SUBSECTION_SEPARATOR);
+  const timeString = moment(time).format(TIME_FORMAT_STRING);
+  const directoryNameComponents = [machineString, timeString];
+  if (inTestMode) {
+    directoryNameComponents.unshift('TEST');
+  }
+  return directoryNameComponents.join(SECTION_SEPARATOR);
+}
```
arsalansufi, line 334: Similarly, this function is just moved down from above. The one below is new.

libs/auth/src/artifact_authentication.ts
```diff
@@ -296,18 +295,26 @@ async function performArtifactSpecificAuthenticationChecks(
...
-        await fs.readFile(path.join(artifact.directoryPath, 'metadata.json'))
-      ).toString('utf-8');
-      const metadata: CastVoteRecordExportMetadata = unsafeParse(
-        CastVoteRecordExportMetadataSchema,
-        metadataFileContents
+      const metadataFileContents = await fs.readFile(
+        path.join(artifact.directoryPath, 'metadata.json'),
+        'utf-8'
       );
+      const parseResult = safeParseJson(
```
arsalansufi, line 302: As noted in the commit description, the mistake here was that `unsafeParse` expects an object, not a JSON string. `safeParseJson`, on the other hand, expects a JSON string, which is what we're providing.

Tests missed this because of how I'd incorrectly mocked `unsafeParse` 🙈

I recognize that I should just not mock, but here's why I mocked for context:

```
// Avoid having to prepare a complete CVR.CastVoteRecordReport object for
// CastVoteRecordExportMetadata
```

libs/backend/src/cast_vote_records/import.ts
```diff
@@ -0,0 +1,268 @@
...
+
+/**
+ * Reads and parses a cast vote record export, authenticating the export in the process. The
+ * export's metadata file is parsed upfront whereas the cast vote records are parsed lazily.
+ *
+ * Basic validation is performed on the cast vote records. Referenced image files and layout files
+ * are not read/parsed, but their existence is validated such that consumers can safely access
+ * them.
+ */
+export async function readCastVoteRecordExport(
```
arsalansufi, line 228: My mental model for the separation of responsibilities between `libs/backend` and `apps/admin/backend` re import is that:

- `libs/backend` reads/parses data and performs validation internal to an export (i.e. it doesn't perform contextual validation like whether the fields in an export and the election definition correspond)
- `apps/admin/backend` actually persists data and performs contextual validation

This separation will allow us to reuse `libs/backend` import code in VxScan, when we build support for bootstrapping a VxScan with another VxScan's CVRs. The persistence logic and contextual validation for VxScan bootstrapping differ from the persistence logic and contextual validation for VxAdmin tabulation, so it's nice that those pieces are separated out.

apps/admin/backend/src/legacy_cast_vote_records.ts
```diff
@@ -113,6 +115,40 @@ export async function listCastVoteRecordFilesOnUsb(
...
+          exportTimestamp: new Date(
+            metadata.castVoteRecordReportMetadata.GeneratedDate
+          ),
+          isTestModeResults: directoryNameComponents.inTestMode,
+          name: entry.name,
+          path: entry.path,
+          scannerIds: [directoryNameComponents.machineId],
+        });
+        continue;
+      }
```
arsalansufi, line 149: Whereas previously, we could extract everything that we needed for the "list of exports on the USB" UI from the directory name, now, we have to pull some info from the metadata file.

Note that I think it's okay for us to extract this information without authenticating the export (which can be time consuming since we have to re-hash), as we aren't actually importing anything yet. If someone tries to trick the system by changing a directory name or the metadata file, the worst they'll be able to do is get the "list of exports on the USB" UI to display something misleading. Importing will still fail.

We've applied similar principles to Java Card auth (e.g. [here](https://docs.google.com/document/d/17qLbrHYkjlEtXx18Oh1TovZz1D1dQd3Ze7azknqd3jA/edit#bookmark=id.gf0zmlln19uj)).

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -62,6 +65,89 @@ const Content = styled.div`
   overflow: hidden;
 `;
 
+/* c8 ignore start */
+function userReadableMessageFromError(
```
arsalansufi, line 69: Note that this function lives on the frontend. The old import code has an equivalent function live on the backend and sends a user-readable message down to the frontend for it to use. Ultimately, this message-preparation is a UI/frontend concern and not a backend concern, hence my moving it to the frontend.

This will become even more necessary once we start working on multi-language support (as noted [here](https://docs.google.com/document/d/1a2GRj1g4aOawEssPdTCgbv-qs28d-M_dyMEfzm4I4nM/edit#heading=h.gno6ffrtd5t)).

apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -62,6 +65,89 @@ const Content = styled.div`
...
+    }
+    case 'metadata-file-parse-error': {
+      return 'Unable to parse metadata file.';
+    }
+    default: {
+      throwIllegalValue(error, 'type');
+    }
+  }
+}
+/* c8 ignore stop */
```
arsalansufi, line 149: I know I've added a lot of code coverage ignores. I've opened issues to track all the tests that I'm deferring to later PRs.

apps/admin/backend/src/cast_vote_records.ts
```diff
@@ -0,0 +1,283 @@
...
+  // imported records
+  const mode: CvrFileMode = isTestReport(castVoteRecordReportMetadata)
+    ? 'test'
+    : 'official';
+  const currentMode = store.getCurrentCvrFileModeForElection(electionId);
+  if (currentMode !== 'unlocked' && mode !== currentMode) {
+    return err({ type: 'invalid-mode', currentMode });
+  }
+
+  const existingImportId = store.getCastVoteRecordFileByHash(
```
arsalansufi, line 117: Throughout the code and UI, we talk about a singular CVR file per export, which is no longer the case. I'm going to update terminology in a follow-up PR

apps/admin/backend/src/cast_vote_records.ts
```diff
@@ -0,0 +1,283 @@
...
+
+      castVoteRecordIndex += 1;
+    }
+
+    // TODO: Calculate the precinct list before iterating through cast vote records, once there is
+    // only one geopolitical unit per batch
+    store.updateCastVoteRecordFileRecord({
+      id: importId,
+      precinctIds,
+    });
```
adghayes, line 270: Now that we have a metadata file, I'm hoping we can eventually get the precinct list in the file up front to avoid this.

apps/admin/backend/src/app.ts
```diff
@@ -404,9 +408,39 @@ function buildApi({
...
       >
     > {
+      /* c8 ignore start */
+      if (
+        isFeatureFlagEnabled(
+          BooleanEnvironmentVariableName.ENABLE_CONTINUOUS_EXPORT
+        )
+      ) {
+        const userRole = assertDefined(await getUserRole());
+        const importResult = await importCastVoteRecords(store, input.path);
```
adghayes, line 422: How are you handling the manual file selection use case? As in line 446 below.

adghayes (approved):

libs/backend/src/cast_vote_records/import.ts
```diff
@@ -0,0 +1,268 @@
...
+      ) as [string, string];
+
+      if (!imageFilePaths.every((filePath) => existsSync(filePath))) {
+        yield wrapError({ subType: 'image-file-not-found' });
+        return;
+      }
+      if (!layoutFilePaths.every((filePath) => existsSync(filePath))) {
+        yield wrapError({ subType: 'layout-file-not-found' });
+        return;
+      }
```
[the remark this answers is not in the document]
arsalansufi: That's correct!

apps/admin/backend/src/app.ts
arsalansufi: Ah good catch, I'm not handling that case.

That said, it turns out that manual file selection is already broken because it only allows you to select an individual file, but for quite some time now, we've been exporting a directory and not a file.

Gonna revisit this holistically as its own issue, potentially just removing the manual option altogether, after checking in with others: https://github.com/votingworks/vxsuite/issues/3967

apps/admin/backend/src/cast_vote_records.ts
```diff
@@ -0,0 +1,283 @@
...
+      BooleanEnvironmentVariableName.SKIP_CVR_ELECTION_HASH_CHECK
+    )
+  ) {
+    return wrapError({ subType: 'election-mismatch' });
+  }
+
+  const precinct = getPrecinctById({
+    election,
+    precinctId: castVoteRecord.BallotStyleUnitId,
+  });
```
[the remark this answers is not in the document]
arsalansufi: https://github.com/votingworks/vxsuite/pull/3956/commits/84480ad4a0a086f1ad359ccb8d9709bbcbb34165

well-arent-you-fancy

apps/admin/backend/src/cast_vote_records.ts
arsalansufi: 👍 Opened an issue for myself so that I don't forget to revisit this

apps/admin/backend/src/app.ts
adghayes: It's a bit of a hidden feature that I failed to properly publicize, but manual file selection is not broken. When I switched cast vote records to directories, I added the indicated logic so that you could select the `cast-vote-record-report.json` at the root of the directory, which would import the directory.

I use this escape hatch _all the time_ when doing manual testing in VxAdmin. It's far easier than having to always get a cast vote record report on a real or mocked USB drive, so I'd vote strongly to keep the option or something like it.

apps/admin/backend/src/cast_vote_records.ts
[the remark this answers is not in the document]
adghayes: Haters always gonna hate on the caching but we've got lots of data to process no time for haters.

apps/admin/backend/src/app.ts
arsalansufi: Ah really great to know! In that case, will plan on maintaining

PR #3959, opened 2023-09-12
Drew/libs usb logging

### LINKED ISSUE #3949: logging: add to `libs/usb-drive`
Most logging can be replicated from the deprecated hook in `libs/ui`.

adghayes:
## Overview

Closes #3949. Adds logging to `libs/usb-drive`. Copies logging mostly verbatim from `use_usb_drive.ts`. For mounting, ejecting, and formatting we log an "init" event and then a success or failure "complete" event. 

## Testing Plan
Added logging testing.

## Checklist

### DIFF
--- a/apps/scan/backend/src/index.ts
+++ b/apps/scan/backend/src/index.ts
@@ -56,7 +56,7 @@ async function resolveWorkspace(): Promise<Workspace> {
 
 async function main(): Promise<number> {
   const workspace = await resolveWorkspace();
-  const usbDrive = detectUsbDrive();
+  const usbDrive = detectUsbDrive(logger);
 
   const precinctScannerStateMachine =
     customStateMachine.createPrecinctScannerStateMachine({
@@ -70,6 +70,7 @@ async function main(): Promise<number> {
     precinctScannerStateMachine,
     workspace,
     usbDrive,
+    logger,
   });
 
   return 0;
--- a/libs/usb-drive/src/usb_drive.ts
+++ b/libs/usb-drive/src/usb_drive.ts
@@ -2,6 +2,7 @@ import { promises as fs } from 'fs';
 import { join } from 'path';
 import { assert, assertDefined } from '@votingworks/basics';
 import makeDebug from 'debug';
+import { LogEventId, Logger, LoggingUserRole } from '@votingworks/logging';
 import { exec } from './exec';
 
 const debug = makeDebug('usb-drive');
@@ -19,8 +20,8 @@ export type UsbDriveStatus =
 
 export interface UsbDrive {
   status(): Promise<UsbDriveStatus>;
-  eject(): Promise<void>;
-  format(): Promise<void>;
+  eject(loggingUserRole: LoggingUserRole): Promise<void>;
+  format(loggingUserRole: LoggingUserRole): Promise<void>;
 }
 
 export interface BlockDeviceInfo {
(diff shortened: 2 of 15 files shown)


libs/usb-drive/src/usb_drive.ts
```diff
@@ -152,7 +153,100 @@ function isFat32(deviceInfo: BlockDeviceInfo): boolean {
   return deviceInfo.fstype === 'vfat' && deviceInfo.fsver === 'FAT32';
 }
 
-export function detectUsbDrive(): UsbDrive {
+async function logMountInit(logger: Logger): Promise<void> {
```
adghayes, line 156: Separating out the logging as methods to improve readability of the main functions.

libs/usb-drive/src/usb_drive.ts
```diff
@@ -152,7 +153,100 @@ function isFat32(deviceInfo: BlockDeviceInfo): boolean {
...
+
+async function mount(
+  deviceInfo: BlockDeviceInfo,
+  logger: Logger
+): Promise<void> {
+  await logMountInit(logger);
+  try {
+    await mountUsbDrive(deviceInfo.path);
+    await logMountSuccess(logger);
+    debug('USB drive mounted successfully');
```
adghayes, line 242: @jonahkagan what was the reason you were using `console.log` before?

libs/usb-drive/src/usb_drive.ts
```diff
@@ -152,7 +153,100 @@ function isFat32(deviceInfo: BlockDeviceInfo): boolean {
...
+  } catch (error) {
+    await logMountFailure(logger, error as Error);
+    debug(`USB drive mounting failed: ${error}`);
+  }
```
adghayes, line 246: Note that mounting errors are, like before, not propagated.

libs/usb-drive/src/usb_drive.ts
```diff
@@ -208,24 +295,34 @@ export function detectUsbDrive(): UsbDrive {
...
+      await logEjectInit(logger, loggingUserRole);
+      try {
+        await unmountUsbDrive(deviceInfo.mountpoint);
+        didEject = true;
+        await logEjectSuccess(logger, loggingUserRole);
+        debug('USB drive ejected successfully');
+      } catch (error) {
+        await logEjectFailure(logger, loggingUserRole, error as Error);
+        debug(`USB drive ejection failed: ${error}`);
+        throw error;
```
adghayes, line 313: Ejecting errors are, like before, propagated.

apps/scan/backend/src/index.ts
```diff
@@ -70,6 +70,7 @@ async function main(): Promise<number> {
     precinctScannerStateMachine,
     workspace,
     usbDrive,
+    logger,
```
adghayes, line 73: I wasn't quite sure why logger wasn't being passed here, which the effect would be that there are two different logger instances when the app is running. I'm not sure that's a problem, but thought it made sense to pass here.

libs/usb-drive/src/usb_drive.ts
jonahkagan: Where was I using console.log? I don't remember any reason.

libs/usb-drive/src/usb_drive.ts
```diff
@@ -19,8 +20,8 @@ export type UsbDriveStatus =
 
 export interface UsbDrive {
   status(): Promise<UsbDriveStatus>;
-  eject(): Promise<void>;
-  format(): Promise<void>;
+  eject(loggingUserRole: LoggingUserRole): Promise<void>;
+  format(loggingUserRole: LoggingUserRole): Promise<void>;
```
jonahkagan, line 24: This API is kind of weird now. It makes me think that we're not separating concerns well when it comes to logging.

I wonder if we should initialize the logger with a way to query the current user role instead of passing it in when calling `log`...

E.g.:
```ts
const logger = new Logger({ auth });
logger.userLog(LogEventId.UserDidSomething); // queries auth for current user type
logger.systemLog(LogEventId.SystemDidSomething); // uses "system" user type
```

libs/usb-drive/src/usb_drive.ts
jonahkagan: We could take this one step further by decorating functions with logging, since the pattern is always the same.

A sketch of how this would look:
```ts
function withLogging(actionLabel, func) {
  try {
    logger.log(actionLabel, 'init');
    await func();
    logger.log(actionLabel, 'success');
  } catch (error) {
    logger.log(actionLabel, 'failure', error);
    throw error;
  }
}

const mount = withLogging(UsbDriveMount, () => {... mount implementation ...});
```

jonahkagan (approved):

libs/usb-drive/src/usb_drive.ts
adghayes: In the `catch` clause the `void mount()` promise. I wondered if there was a need to use `console.log` when it's out of the normal execution like that. You even had a lint rule disabled for it 🤷

jonahkagan: Probably was just cutting corners for initial implementation. My guilty secret is that I always delay adding formal logging because I don't like how it clutters the code (as you can see from my other comments)

libs/usb-drive/src/usb_drive.ts
adghayes: It is weird. I think this is a good idea. There are lots of places where we're passing user roles around awkwardly to log with the role. It was a problem with the over-computation of tallies in the frontend. That was happening in an effect (bad start) that depended on the user role to log properly. Very silly.

libs/usb-drive/src/usb_drive.ts
adghayes: I feel the same. I too often make decisions based off of code appearance.

Issue #3909, opened 2023-08-31
Return to review screen after spoiling ballot


mattroe:
Instead of restarting the entire voting session?

PR #3962, opened 2023-09-12
Kevin/vxmarkscan alert pollworker after invalidate

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3883

On ballot invalidation, displays a screen telling the voter to alert a pollworker for assistance. The pollworker must authenticate before continuing.

In the long term we'll want this state to allow the voter to go back to edit their selections, tracked in https://github.com/votingworks/vxsuite/issues/3909

## Demo Video or Screenshot
n.b. there's some rendering lag in the middle of the video. You may want to skip ahead past it.

https://github.com/votingworks/vxsuite/assets/1060688/09d74659-6c9c-46eb-b878-d9e31cf29e6f

## Testing Plan
* Added tests for different auth statuses in invalidation page.
* Added top-level state tests to keep coverage at 100

## Checklist

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -76,7 +76,7 @@ type PaperHandlerStatusEvent =
   | { type: 'VOTER_INITIATED_PRINT'; pdfData: Buffer }
   | { type: 'PAPER_IN_OUTPUT' }
   | { type: 'PAPER_IN_INPUT' }
-  | { type: 'CONFIRM_INVALIDATE_BALLOT' }
+  | { type: 'VOTER_CONFIRMED_INVALIDATED_BALLOT' }
   | { type: 'SCANNING' }
   | { type: 'VOTER_VALIDATED_BALLOT' }
   | { type: 'VOTER_INVALIDATED_BALLOT' };
@@ -124,8 +124,8 @@ export class PaperHandlerStateMachine {
         return 'scanning';
       case state.matches('interpreting'):
         return 'interpreting';
-      case state.matches('invalidating_ballot'):
-        return 'invalidating_ballot';
+      case state.matches('waiting_for_invalidated_ballot_confirmation'):
+        return 'waiting_for_invalidated_ballot_confirmation';
       case state.matches('presenting_ballot'):
         return 'presenting_ballot';
       case state.matches('eject_to_front'):
@@ -184,7 +184,7 @@ export class PaperHandlerStateMachine {
 
   confirmInvalidateBallot(): void {
     this.machineService.send({
-      type: 'CONFIRM_INVALIDATE_BALLOT',
+      type: 'VOTER_CONFIRMED_INVALIDATED_BALLOT',
     });
   }
 }
@@ -434,14 +434,15 @@ export function buildMachine(
         },
         on: {
           VOTER_VALIDATED_BALLOT: 'eject_to_rear',
-          VOTER_INVALIDATED_BALLOT: 'invalidating_ballot',
+          VOTER_INVALIDATED_BALLOT:
+            'waiting_for_invalidated_ballot_confirmation',
           NO_PAPER_ANYWHERE: 'resetting_state_machine_after_success',
         },
       },
       // Ballot invalidation is a 2-stage process so the frontend can prompt the voter to get a pollworker
-      invalidating_ballot: {
+      waiting_for_invalidated_ballot_confirmation: {
         on: {
-          CONFIRM_INVALIDATE_BALLOT: 'eject_to_front',
+          VOTER_CONFIRMED_INVALIDATED_BALLOT: 'eject_to_front',
           // Even if ballot is removed from front, we still want the frontend to require pollworker auth before continuing
           NO_PAPER_ANYWHERE: undefined,
         },
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -442,7 +442,8 @@ export function buildMachine(
       invalidating_ballot: {
         on: {
           CONFIRM_INVALIDATE_BALLOT: 'eject_to_front',
-          NO_PAPER_ANYWHERE: 'resetting_state_machine_after_success',
+          // Even if ballot is removed from front, we still want the frontend to require pollworker auth before continuing
+          NO_PAPER_ANYWHERE: undefined,
         },
       },
       // Eject-to-rear jam handling is a little clunky. It
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -76,6 +76,7 @@ type PaperHandlerStatusEvent =
   | { type: 'VOTER_INITIATED_PRINT'; pdfData: Buffer }
   | { type: 'PAPER_IN_OUTPUT' }
   | { type: 'PAPER_IN_INPUT' }
+  | { type: 'CONFIRM_INVALIDATE_BALLOT' }
   | { type: 'SCANNING' }
   | { type: 'VOTER_VALIDATED_BALLOT' }
   | { type: 'VOTER_INVALIDATED_BALLOT' };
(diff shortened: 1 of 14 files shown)


apps/mark-scan/frontend/src/pages/ballot_invalidated_page.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { resetBallot, endVoterSession } = useContext(BallotContext);
+
+  const confirmInvalidateBallotMutation = confirmInvalidateBallot.useMutation();
+
+  function onPressContinue() {
+    confirmInvalidateBallotMutation.mutate(undefined, {
+      async onSuccess() {
+        resetBallot();
+        await endVoterSession();
+      },
```
jonahkagan, line 21: I noticed in the video that the screen briefly flashes to the "start voting" screen before going back to the pollworker screen, which seemed unintentional. Might have to do with the order of actions here, though I'm not sure.

apps/mark-scan/frontend/src/pages/ballot_invalidated_page.tsx
```diff
@@ -0,0 +1,44 @@
...
+    });
+  }
+
+  return (
+    <Screen>
+      <Main padded centerChild>
+        <Text center>
+          <H1>
+            <span aria-label="Ballot Invalidated.">Ballot Invalidated</span>
```
jonahkagan, line 30: Since most people generally don't read the fine print, I might consider changing this headline to something like "Ask a Poll Worker for Help" so that it's clear what action is important.

apps/mark-scan/frontend/src/pages/ballot_invalidated_page.tsx
```diff
@@ -0,0 +1,44 @@
...
+          </H1>
+          <p>
+            You have indicated your ballot needs changes. Please alert a poll
+            worker to spoil your incorrect ballot and restart your voting
+            session.
+          </p>
+        </Text>
+      </Main>
+      <ButtonFooter>
+        <Button onPress={onPressContinue}>I Have Alerted a Poll Worker</Button>
```
jonahkagan, line 40: I wonder if we should just wait on this screen until a poll worker inserts their card and confirms the ballot was spoiled?

jonahkagan (commented):
First pass review just on UX in case that ends up changing implementation

kshen0:
Thanks for the feedback. I made the recommended changes and will reassign after I can QA this tomorrow.

kshen0:
@jonahkagan I made the following changes
* To fix flicker, reset auth and ballot state **before** changing state machine state. When in the `invalidating_ballot` state, both pollworker auth and cardless voter auth will render `BallotInvalidatedPage`, so there shouldn't be a race condition on auth status any more
* Changed headline to your suggestion
*  `BallotInvalidatedPage` requires pollworker auth to continue

Note that we expect the voter or poll worker may remove the ballot at any stage **after** pressing "My Ballot is Incorrect". Since we can't enforce or predict when the ballot is removed, we don't change state based on paper presence. We're relying somewhat heavily on the pollworker to handle the ballot correctly after restarting the voting session. While it would be nice to tell the poll worker they need to spoil the ballot first, this would make the interaction a little awkward:

* voter indicates their ballot needs changes and calls over a poll worker
* poll worker authenticates
* screen tells poll worker to spoil the ballot
* poll worker leaves to spoil the ballot, taking their card with them
* voter is left waiting with an unactivated machine
* poll worker returns to start a new session

To avoid making the voter wait I think it's better to create a new voter session immediately and trust the pollworker to spoil the ballot. What do you think?

jonahkagan:
Nice! This new flow seems great to me. I agree with your instinct to trust the poll worker to spoil the ballot. We could potentially change the headline for that page to "Remove and Spoil Ballot." Also can get feedback on this when you demo to the rest of the team. I'll review the implementation now!

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -76,6 +76,7 @@ type PaperHandlerStatusEvent =
   | { type: 'VOTER_INITIATED_PRINT'; pdfData: Buffer }
   | { type: 'PAPER_IN_OUTPUT' }
   | { type: 'PAPER_IN_INPUT' }
+  | { type: 'CONFIRM_INVALIDATE_BALLOT' }
```
jonahkagan, line 79: Maybe rename this `VOTER_CONFIRMED_INVALIDATED_BALLOT` to be consistent with the other voter actions?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -425,7 +434,16 @@ export function buildMachine(
         },
         on: {
           VOTER_VALIDATED_BALLOT: 'eject_to_rear',
-          VOTER_INVALIDATED_BALLOT: 'eject_to_front',
+          VOTER_INVALIDATED_BALLOT: 'invalidating_ballot',
+          NO_PAPER_ANYWHERE: 'resetting_state_machine_after_success',
+        },
+      },
+      // Ballot invalidation is a 2-stage process so the frontend can prompt the voter to get a pollworker
+      invalidating_ballot: {
```
jonahkagan, line 442: Nit: When I first read this state name, I thought maybe the backend was doing something to invalidate the ballot. Maybe would name this `waiting_for_invalidated_ballot_confirmation` or something.

apps/mark-scan/frontend/src/pages/ask_poll_worker_page.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </H1>
+        {cardlessVoterContent.body}
+      </React.Fragment>
+    );
+    button = (
+      // onPress is a required prop but we want the button to no op
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      <Button disabled onPress={() => {}}>
+        Insert a Poll Worker Card to Continue
+      </Button>
```
jonahkagan, line 58: I wonder if this should just be a plain text message instead of a disabled button, since it's not really asking for the user to press the button, but rather do something off screen

apps/mark-scan/frontend/src/pages/ballot_invalidated_page.tsx
```diff
@@ -0,0 +1,53 @@
...
+  async function onPressContinue() {
+    // Reset session and ballot before changing the state machine state. If done
+    // in the reverse order the screen will flicker
+    await endVoterSession();
+    resetBallot();
+    confirmInvalidateBallotMutation.mutate(undefined);
+  }
+
+  return (
+    <AskPollWorkerPage
```
jonahkagan, line 29: Is there a reason that this flow is abstracted out into a separate component? It seems like it's only used here in `BallotInvalidatedScreen`

jonahkagan (approved):

apps/mark-scan/frontend/src/pages/ballot_invalidated_page.tsx
kshen0: Great question, I use it in this branch that follows the same pattern but for blank sheet interpretations:
https://github.com/votingworks/vxsuite/pull/3981/files#diff-1db666e555eff795a25c5b430171548b3cbea77dec73e069948df2552e78376e

apps/mark-scan/frontend/src/pages/ask_poll_worker_page.tsx
kshen0: Good point. I added this as a button because I thought it would make it clearer that a poll worker card is required to continue. While I still think it does that more effectively than a text message (that people might skim over), it's somewhat of a dark pattern because the button will never actually work. It also might (?) be less accessible to screen readers. Removed in [b2aba7c](https://github.com/votingworks/vxsuite/pull/3962/commits/b2aba7c9a11c378009cdda4f0298e03c970fce88)

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Thanks! [ac47292](https://github.com/votingworks/vxsuite/pull/3962/commits/ac472921d2a0ae8222d0827d0ab717ae927c667a)

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: [ac47292](https://github.com/votingworks/vxsuite/pull/3962/commits/ac472921d2a0ae8222d0827d0ab717ae927c667a)

Issue #3923, opened 2023-09-05
Randomly "shuffle" CVR export contents during continuous export to maintain voter privacy


arsalansufi:
Also use a placeholder `GeneratedDate` timestamp in individual cast vote record files

PR #3972, opened 2023-09-14
Maintain voter privacy by obfuscating the order in which ballots are cast

arsalansufi:
~_Easiest reviewed by commit_~

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3923

This PR maintains voter privacy by obfuscating the order in which ballots are cast - something that requires extra care when exporting cast vote records continuously.

Copying a code comment:

> File creation timestamps could reveal the order in which ballots were cast. To maintain
> voter privacy, every time a ballot is cast, we randomly select 1 or 2 previously added cast vote
> records and update their creation timestamps to the present. This approach is equivalent to
> moving random ballots to the top of a stack every time a new ballot is added to the stack, a
> kind of shuffling as we go.
>
> This shuffling is irrelevant for full exports, where we already iterate over cast vote records
> in an order independent of creation timestamp.

The PR also removes time from the metadata for individual cast vote records.

## Demo Video

A boring or exciting video depending on your perspective 😆 that shows the contents of a cast vote record export, ordered by creation date. You can see how random existing directories hop to the top every time a new directory is added.

https://github.com/votingworks/vxsuite/assets/12616928/1217fbc4-3eeb-47b8-9d0e-a6660faa8ddf

## Testing

Tested manually and will add automated tests in subsequent PRs

## Checklist


libs/backend/src/cast_vote_records/export.ts
```diff
@@ -423,6 +425,76 @@ async function exportSignatureFileToUsbDrive(
...
+    .filter((entry) => entry.isFile())
+    .map((entry) => entry.name);
+  for (const fileName of fileNames) {
+    await fs.copyFile(
+      path.join(directoryPath, fileName),
+      path.join(tempDirectoryPath, fileName)
+    );
+  }
+  await fs.rm(directoryPath, { recursive: true });
+  await fs.rename(tempDirectoryPath, directoryPath);
```
adghayes, line 456: In the comment's description of this function, you explain what I would expect - copy and rename directory. But in the function itself you're iterating and copying each file individually rather than copying the directory recursively. Is there a reason for this?

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -423,6 +425,76 @@ async function exportSignatureFileToUsbDrive(
...
+ * and update their creation timestamps to the present. This approach is equivalent to taking
+ * random ballots from a stack and placing them on top every time a new ballot is added to the
+ * stack, a kind of shuffling as we go.
+ *
+ * This shuffling as we go is irrelevant for full exports, where we already iterate over cast vote
+ * records in an order independent of creation timestamp.
+ */
+async function maintainVoterPrivacy(
+  exportContext: ExportContext
+): Promise<void> {
```
adghayes, line 471: who needs election guard when you can have a javascript function called `maintainVoterPrivacy`

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -423,6 +425,76 @@ async function exportSignatureFileToUsbDrive(
...
+    await fs.readdir(exportDirectoryPath, { withFileTypes: true })
+  )
+    .filter((entry) => entry.isDirectory())
+    .map((entry) => entry.name);
+
+  if (castVoteRecordIds.length < 2) {
+    return;
+  }
+
+  const oneOrTwo = Math.random() < 0.5 ? 1 : 2;
```
adghayes, line 488: Is there a security reason for randomly picking a number here?

My first thought is that it would have been based on the size of the pile - the more CVRs there are, the more should be reshuffled.

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -174,6 +175,7 @@ function buildCastVoteRecordReportMetadata(
     batchInfo: batches,
     election,
     electionId,
+    generatedDate: options.hideTime ? new Date(election.date) : undefined,
```
adghayes, line 178: Why is this important for voter privacy?

adghayes: And why is including it at the end of the day _not_ a problem for voter privacy?

adghayes (approved):

adghayes:
@arsalansufi do you want @benadida's review on this too?

libs/backend/src/cast_vote_records/export.ts
eventualbuddha: Related: [`Math.random()` is not cryptographically secure](https://kemilbeltre.medium.com/why-do-not-use-math-random-a6f8b0ad38dd). We should probably add a lint rule against its use. You could do this instead:
```suggestion
  const oneOrTwo = getRandomValues(new Uint8Array(1))[0] % 2 === 1 ? 1 : 2;
```

libs/backend/src/cast_vote_records/export.ts
arsalansufi: Ah yes good q, meant to leave a note about this. `fs.cp`, which supports a `recursive` param, is [experimental](https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options), so I didn't want to use it

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -423,6 +429,76 @@ async function exportSignatureFileToUsbDrive(
...
+/**
+ * Updates the creation timestamp of a directory and its children files (assuming no
+ * sub-directories) using the one method guaranteed to work:
+ * ```
+ * cp -r <directory-path> <directory-path>-temp
+ * rm <directory-path>
+ * mv <directory-path>-temp <directory-path>
+ * ```
+ */
+async function updateCreationTimestampOfDirectoryAndChildrenFiles(
```
eventualbuddha, line 441: I think there should be tests for this to ensure that the timestamps do actually change.

libs/backend/src/cast_vote_records/export.ts
arsalansufi: So we use `buildCastVoteRecordReportMetadata` to generate the metadata in two places: 1) in the top-level `metadata.json` file but also 2) in individual `cast-vote-record-report.json` files, which we now create per CVR.

If we had real date times in 2, our shuffling would become irrelevant since you could just look at the `GeneratedDate` field in the `cast-vote-record-report.json` files.

There's no issue with a real date time in 1 since it's top-level and not tied to any individual CVR.

Let me know if I can clarify further!

libs/backend/src/cast_vote_records/export.ts
arsalansufi: 👍 I was planning to add tests in subsequent PRs, but I can go ahead and add this one now since it's important

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -423,6 +429,76 @@ async function exportSignatureFileToUsbDrive(
...
+  if (castVoteRecordIds.length < 2) {
+    return;
+  }
+
+  const oneOrTwo = Math.random() < 0.5 ? 1 : 2;
+  for (let i = 0; i < oneOrTwo; i += 1) {
+    const randomCastVoteRecordDirectoryPath = path.join(
+      exportDirectoryPath,
+      assertDefined(
+        castVoteRecordIds[Math.floor(Math.random() * castVoteRecordIds.length)]
```
eventualbuddha, line 493: Another `Math.random()`. Maybe add a function like this to use in both places?

```ts
function getRandomNumber({ min, max }: { min: number; max: number }): number {
  if (min >= max) {
    throw new Error('min must be less than max');
  }

  const range = max - min;
  if (range <= 0) {
    throw new Error('Range must be greater than 0');
  }

  const byteArray = randomBytes(4); // 4 bytes for a 32-bit integer
  const randomInt = byteArray.readUInt32LE(0); // Convert the bytes to an integer
  return min + (randomInt % range);
}
```

Then you can do this:
```suggestion
        castVoteRecordIds[getRandomNumber({ min: 0, max: castVoteRecordIds.length - 1 })]
```

arsalansufi:
Yeah good call @adghayes re adding @benadida! I'll add him once I've addressed the existing comments (plus some other thoughts I've had since opening this PR)

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -423,6 +429,76 @@ async function exportSignatureFileToUsbDrive(
...
+ * mv <directory-path>-temp <directory-path>
+ * ```
+ */
+async function updateCreationTimestampOfDirectoryAndChildrenFiles(
+  directoryPath: string
+): Promise<void> {
+  const tempDirectoryPath = `${directoryPath}-temp`;
+  await fs.mkdir(tempDirectoryPath);
+  const fileNames = (await fs.readdir(directoryPath, { withFileTypes: true }))
+    .filter((entry) => entry.isFile())
```
eventualbuddha, line 447: This line means the function silently ignores directories. If we're confident there won't be any directories, should we `throw` if we encounter one? Otherwise a change to the CVR directory structure will cause this function to slowly corrupt the CVRs.

libs/backend/src/cast_vote_records/export.ts
arsalansufi: Good q @adghayes! You can basically make more inferences if you're always moving the same number of CVRs. For example, if you're always moving one CVR (and assuming we already have lots of CVRs), there's a high likelihood that every other CVR among the latest CVRs is one of the last ones cast. Even if you're always moving two CVRs, the same situation can arise, but with every third ballot instead of every other. Adding this additional source of randomness makes the whole system less orderly.

Somewhat relatedly, since opening this PR, I've been thinking about protecting the privacy of the last voter to cast their ballot before polls close. And I'm thinking of adding one more source of randomness to accomplish that, since there currently are situations where you could infer which CVR belongs to that voter.

arsalansufi: Thank you for the reminder @eventualbuddha! I recall you updating PIN generation code some time a while back for this same reason. I think that the fact that `Math.random` is not cryptographically secure will really stick this time, but I also agree that a lint rule is a good idea

arsalansufi: https://github.com/votingworks/vxsuite/pull/3972/commits/5785f26535d5c1d5158144da7dfb7d74b6134dcc

libs/backend/src/cast_vote_records/export.ts
arsalansufi: Good call - added a helper to libs/basics: https://github.com/votingworks/vxsuite/pull/3972/commits/1c59eb2dcd23e8db25b5c0fbd11d89eec642004e

libs/backend/src/cast_vote_records/export.ts
arsalansufi: Added a clarifying comment in https://github.com/votingworks/vxsuite/pull/3972/commits/bc83adca281107c4fd81c5408af0802da159ba38

libs/backend/src/cast_vote_records/export.ts
arsalansufi: Updated the comment to make this clearer in https://github.com/votingworks/vxsuite/pull/3972/commits/bc83adca281107c4fd81c5408af0802da159ba38

libs/backend/src/cast_vote_records/export.ts
arsalansufi: Okay, did decide to throw here and in one other place relevant to artifact authentication https://github.com/votingworks/vxsuite/pull/3972/commits/28f0f9002b2321df5ee839eef496754444393862

I've been worried about hidden directories (and files) created by file system tools causing unexpected problems, but ultimately, I think the rigidity makes sense here

libs/backend/src/cast_vote_records/export.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/3972/commits/d4ef8523f534eb0d41ffc4e2f83f4b14291f3c3c

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -423,6 +432,113 @@ async function exportSignatureFileToUsbDrive(
...
+      });
+      castVoteRecordsToUpdate.push(
+        assertDefined(previouslyAddedCastVoteRecordIds[randomIndex])
+      );
+    }
+    // 25% of the time, move the most recently added cast vote record back to the top of the stack.
+    // This is meant to protect the privacy of the last voter, whose cast vote record would
+    // otherwise always have the earliest timestamp among the cluster of cast vote records with the
+    // latest timestamps.
+    if (getRandomInteger({ min: 1, max: 4 }) === 1) {
```
arsalansufi, line 526: I picked a 25% probability for this piece somewhat arbitrarily. Just trying to strike a balance between not having to copy too much but adding enough randomness that you can't know for sure which cast vote record belongs to the last voter.

There's an alternative way to go about this: probabilistically deciding whether we call `maintainVoterPrivacy` before or after creating the new cast vote record. But that makes `maintainVoterPrivacy` less self-contained and the overall code a bit harder to reason about in my opinion.

arsalansufi: Had a change of heart here 🙈 and decided that the alternative was in fact still easy enough to understand, while also being more performant since we never have to copy the just created cast vote record directory

https://github.com/votingworks/vxsuite/pull/3972/commits/7fc2c4f58daa618df81e24af3b08731c47aa226f

arsalansufi:
@benadida This one is now in a good state for you to security review! (No rush of course, I know you have plenty on your plate today with NH)

benadida (approved):
one comment that can be used for future improvement, not blocking this.

One question: is there recovery code for what happens if you power dies and you reboot and you have to scan the USB stick and finish whatever directory moves were in progress?

cjerdonek:
> File creation timestamps could reveal the order in which ballots were cast. To maintain voter privacy, every time a ballot is cast, we randomly select 1 or 2 previously added cast vote records and update their creation timestamps to the present. This approach is equivalent to moving random ballots to the top of a stack every time a new ballot is added to the stack, a kind of shuffling as we go.

From a security perspective, wouldn't one want the position of the most recently cast ballot to be anywhere in the order with equal likelihood? With the approach above, it seems like the most recently cast ballot will always be in one of the top three slots (so not a uniform distribution). Is there a reason there can't be an approach that would enable this?

cjerdonek:
I'm also wondering if information can be leaked with the approach above based on the timestamps that are "left" at the end of the day. For example, say a single ballot was cast at 7:15am. If at the end of the day there's a ballot with a timestamp of 7:15am (because it was never randomly selected), wouldn't this voter be revealed?

libs/basics/src/random.ts
```diff
@@ -0,0 +1,24 @@
...
+}: {
+  min: number;
+  max: number;
+}): number {
+  assert(Number.isInteger(min), 'min should be an integer');
+  assert(Number.isInteger(max), 'max should be an integer');
+  assert(min <= max, 'min should be less than or equal to max');
+
+  const range = max - min;
+  const byteArray = crypto.randomBytes(4); // 4 bytes for a 32-bit integer
```
cjerdonek, line 21: Is there a reason not to use [`crypto.randomInt()`](https://nodejs.org/docs/latest-v16.x/api/crypto.html#cryptorandomintmin-max-callback)?

benadida: this is a great comment, not sure how I missed this!

benadida:
@cjerdonek thanks for these comments. You are correct, of course, that the current plan doesn't have perfect privacy. The implementation we've chosen is one that at least defends against all the ballots being in cast order, while also being efficient enough to work according to VVSG2 requirements, which is that the USB stick needs to be always ready with all CVRs written. So we're doing just a little bit of shuffling.

We need to review more research into the literature on history-independent data structures, and we're hobbled by the filesystem constraints, which don't allow us to change creation time on USB sticks. If you've got better ideas, we welcome them!

cjerdonek:
Thanks for being open to my feedback, @benadida! A couple thoughts had occurred to me:

First, instead of selecting 1 or 2 of the previously added cast vote records at random and then updating their creation timestamps to the present, how about choosing the first N CVR's in the stack's _current_ order (where N is some small number chosen before the election, say 5), and then re-adding them, together with the newest CVR, in a random order at the top. That would both (1) ensure that the most recently cast ballot is in one of N spots with equal probability (N could be made as high as you want), as well as (2) address the second point I made and not risk revealing any voters' votes. (It would be interesting for someone to do an analysis of what the overall distribution would look like over time using this process.)

Second, that's too bad about the USB stick's limitations, because that was another possibility. However, what about the idea of pre-creating all of the files at the very beginning, and then writing the bytes to those pre-existing files (choosing the file to write to at random each time)? That way the creation times won't be giving anything away. (I'm not sure if there is a similar limitation on the last-modified times.)

libs/basics/src/random.ts
arsalansufi: Ah yes good catch! https://github.com/votingworks/vxsuite/pull/3972/commits/05f279b03b9032ccf863981a7d6ab9dfa554f28b

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -423,6 +443,96 @@ async function exportSignatureFileToUsbDrive(
...
+    );
+    return entry.name;
+  });
+  for (const fileName of fileNames) {
+    await fs.copyFile(
+      path.join(directoryPath, fileName),
+      path.join(tempDirectoryPath, fileName)
+    );
+  }
+  await fs.rm(directoryPath, { recursive: true });
```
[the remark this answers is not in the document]
arsalansufi: Makes sense! I'll be thinking through recovery / resync logic in subsequent PRs, so this is good to have in place

https://github.com/votingworks/vxsuite/pull/3972/commits/a2a4390d6c989f2fc196491eb66cb55efcb96a08

arsalansufi:
+1 - thanks @cjerdonek for the comments!

@benadida summarized the constraints well. VVSG2 essentially requires continuous export to USB, so while it's true that we'd

> want the position of the most recently cast ballot to be anywhere in the order with equal likelihood

it'd be hard to accomplish this without either 1) re-exporting all CVRs every time a ballot is cast, which would be prohibitively expensive/slow, or 2) finding a more efficient way to reliably change creation timestamps in the file system, which we haven't yet been able to do.

-----

> I'm also wondering if information can be leaked with the approach above based on the timestamps that are "left" at the end of the day. For example, say a single ballot was cast at 7:15am (and the rest after 8am, say). If at the end of the day there's a ballot with a timestamp of 7:15am (because it was never randomly selected), wouldn't this voter be revealed?

So I believe we still protect against this with the current implementation. Every time a ballot A is cast, at least one other ballot B is updated to have (essentially) the same timestamp. Even if one of ballot A or B is later randomly selected to have some other timestamp, leaving a lone ballot at the original timestamp, an attacker would have no way to know whether the lone ballot is ballot A or B. So in the case that you've laid out, an attacker would have no way to know whether the 7:15AM ballot is actually the ballot that was cast at that time or a ballot that was updated to have that time.

-----

That said, even if we aren't at risk of revealing any voters with the current implementation, I still like your following suggestion:

> First, instead of selecting 1 or 2 of the previously added cast vote records at random and then updating their creation timestamps to the present, how about choosing the first N CVR's in the stack's current order (where N is some small number chosen before the election, say 5), and then re-adding them, together with the newest CVR, in a random order at the top.

I could even imagine a hybrid of the current approach and this approach, where we move up a combo of ballots from a certain range in the stack plus some ballots from anywhere in the stack (since the latter, too, has its benefits).

-----

> what about the idea of pre-creating all of the files at the very beginning, and then writing the bytes to those pre-existing files (choosing the file to write to at random each time)?

Very interesting! There'd be some trickiness in getting the implementation right, since we'd need to pre-create both directories and files on the USB. We'd also need to make sure our digital signing operations properly account for these placeholder directories and files. And finally, as you've noted, we'd still need to account for / obfuscate file modification times (which generally are easier to update than creation times, e.g. `touch file.txt`).

-----

I'm leaning toward merging the current implementation but opening an issue to track further improvements, which I'll revisit after I finish the rest of my work on continuous CVR export.

Thanks again @cjerdonek for the critical look and ideation!

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -601,7 +596,7 @@ export async function exportCastVoteRecordsToUsbDrive(
     // identifiable, as either the first or the last cast vote record among the cluster of cast
     // vote records with the latest creation timestamps.
     const whenToShuffleRelativeToCastVoteRecordCreation = assertDefined(
-      (['before', 'after'] as const)[getRandomInteger({ min: 0, max: 1 })]
+      (['before', 'after'] as const)[crypto.randomInt(0, 2)]
```
eventualbuddha, line 603: Good to know this is a thing.

PR #3973, opened 2023-09-14
[libs/types] Add translation/audio schemas to BallotPackage

kofi-q:
## Overview

Related tech spec [here](https://docs.google.com/document/d/1a2GRj1g4aOawEssPdTCgbv-qs28d-M_dyMEfzm4I4nM/edit?usp=sharing).

Adding new types/schemas for translation/audio-related data files we'll be adding to the election package zip file:
- `BallotPackageMetadata`: Will be loaded from `metadata.json` - currently just contains a `version` field which will allow us to verify the ballot package version and the machine build version match up. Wasn't sure if this might make sense to add to the existing `SystemSettings` instead - LMK if you think so.
- `UiStringsPackage`: Per-language string translations for all enabled languages - to be loaded from `appStrings.json` and merged with additional election strings translations from the election CDF.
- `UiStringAudioClips`: List of audio clips, with associated metadata. These will load from an `audioClips.jsonl` file in the zip package - went with that instead of a single JSON object just in case we need to stream them to the DB in the future, instead of loading them all into memory first.
- `UiStringAudioKeysPackage`: Per-language mappings from string key to a sequence of audio keys, to be loaded from `audioKeys.json`.

## Testing Plan
- Added a few JSON parsing unit tests to verify the schemas are set up properly

## Checklist

### DIFF
--- a/libs/types/src/ballot_package.ts
+++ b/libs/types/src/ballot_package.ts
@@ -5,11 +5,28 @@ import {
   PrecinctId,
 } from './election';
 import { SystemSettings } from './system_settings';
+import { BallotPackageMetadata } from './ballot_package_metadata';
+import { UiStringAudioClips } from './ui_string_audio_clips';
+import { UiStringAudioKeysPackage } from './ui_string_audio_keys';
+import { UiStringsPackage } from './ui_string_translations';
+
+export enum BallotPackageFileName {
+  APP_STRINGS = 'appStrings.json',
+  AUDIO_CLIPS = 'audioClips.jsonl',
+  AUDIO_KEYS = 'audioKeys.json',
+  ELECTION = 'election.json',
+  METADATA = 'metadata.json',
+  SYSTEM_SETTINGS = 'systemSettings.json',
+}
 
 export interface BallotPackage {
   electionDefinition: ElectionDefinition;
+  metadata?: BallotPackageMetadata; // TODO(kofi): Make required
   // TODO(kevin) once all machines support system settings, make systemSettings required
   systemSettings?: SystemSettings;
+  uiStringAudioClips?: UiStringAudioClips; // TODO(kofi): Make required
+  uiStringAudioKeys?: UiStringAudioKeysPackage; // TODO(kofi): Make required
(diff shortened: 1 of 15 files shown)


kofi-q:
(cc @jonahkagan for optional review, since this relates to VxDesign exports)

libs/types/src/ballot_package.ts
```diff
@@ -5,11 +5,28 @@ import {
...
+import { UiStringAudioClips } from './ui_string_audio_clips';
+import { UiStringAudioKeysPackage } from './ui_string_audio_keys';
+import { UiStringsPackage } from './ui_string_translations';
+
+export enum BallotPackageFileName {
+  APP_STRINGS = 'appStrings.json',
+  AUDIO_CLIPS = 'audioClips.jsonl',
+  AUDIO_KEYS = 'audioKeys.json',
+  ELECTION = 'election.json',
+  METADATA = 'metadata.json',
```
arsalansufi, line 18: Putting the ballot package version in a separate `metadata.json` file, as opposed to the `systemSettings.json` makes sense to me. The ballot package version feels higher-level than any one file within the ballot package (and could even dictate which files are present to begin with, somewhere down the road)

libs/types/src/ui_string_audio_clips.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface UiStringAudioClip {
+  data: string;
+  key: string;
+  lang: LanguageCode;
+}
+
+/**
+ * A single audio clip record in the audio clips JSONL file in a ballot package.
+ */
+export const UiStringAudioClipJsonSchema: z.ZodType<UiStringAudioClip> =
```
arsalansufi, line 17: Small consistency nit: I believe most of our types and schemas follow a naming convention where:

- Type name = `<TypeName>`
- Schema name = `<TypeName>Schema`

As opposed to `<TypeName>JsonSchema` as we have here. There might be a conscious reason for this divergence, though, that I'm missing!

arsalansufi (approved):
LGTM!

jonahkagan (approved):
Lgtm!

Issue #3945, opened 2023-09-08
"Printing ballot" screen shows during ballot cast


kshen0:
Flagged in https://github.com/votingworks/vxsuite/pull/3931#discussion_r1317951958

PR #3981, opened 2023-09-18
Kevin/vxmarkscan warn when interpretation fails

kshen0:
## Overview
Adds a flow for when the ballot interpretation is `BlankPageInterpretation`. Barring hardware failures, we expect this to happen only when the thermal paper is inserted upside down.

## Demo Video or Screenshot

In this video I load non-thermal paper first, then load thermal paper in the correct orientation second.

https://github.com/votingworks/vxsuite/assets/1060688/d948f999-e500-421a-beda-28fe4f67f957

Known issues in this video:
* "Printing Ballot" screen shows when ballot is being cast https://github.com/votingworks/vxsuite/issues/3945
* Validation screen says "You may still vote in this contest" for contest with no votes. We should hide that copy because this suggests the voter can still edit their selections by pressing a contest card, which doesn't work. They have to press the "My Ballot is Incorrect" button and enter the ballot spoil flow. (Github Issues is down; I'll file this ticket once it's back up).

## Testing Plan
* Added tests for new frontend state

## Checklist


apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -324,15 +369,87 @@ export function buildMachine(
...
+            },
+          },
+          load_paper: {
+            entry: async (context) => {
+              await context.driver.loadPaper();
+              await context.driver.parkPaper();
+            },
+            on: {
+              PAPER_PARKED: 'done',
+              NO_PAPER_ANYWHERE: 'accepting_paper',
```
kshen0, line 426: Handle case of user pulling paper out of the infeed before it's fully loaded. Jammed paper is handled by top-level `on: { PAPER_JAM: 'jammed' }` transition

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -324,15 +369,86 @@ export function buildMachine(
...
+                interpretation: undefined,
+                scannedImagePaths: undefined,
+              });
+            },
+          },
+        },
+        onDone: { target: 'paper_reloaded' },
+      },
+      // paper_reloaded is a distinct state from paper_parked only
+      // because the frontend handles them differently
```
kshen0: `paper_reloaded` is only used by the `blank_page_interpretation` state and could be a child state of `blank_page_interpretation`, but there are 2 frontend components.
1. `BlankPageInterpretationPage` for all child states currently in `blank_page_interpretation`
2. `PaperReloadedPage` for the `paper_reloaded` state

So our conditions on `stateMachineState` are very concise; see https://github.com/votingworks/vxsuite/pull/3981/files#diff-dc3b4ba8825793529470eb8c571a54a06cff23c83696464aebf5bea255070ec5R702

If we make `paper_reloaded` a child of `blank_page_interpretation` then we'd have to check states like:

```
if (['blank_page_interpretation.presenting_paper', 'blank_page_interpretation.accepting_paper', 'blank_page_interpretation.load_paper'].includes(stateMachineState)) {
  return <BlankPageInterpretationPage ... />
}
if (stateMachineState === 'blank_page_interpreation.paper_reloaded') {
  return <PaperReloadedPage />
}
```

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -187,6 +192,46 @@ function pollPaperStatus(): InvokeConfig<Context, PaperHandlerStatusEvent> {
...
+      }),
+      timeout({
+        each: PAPER_HANDLER_STATUS_POLLING_TIMEOUT_MS,
+        with: () => throwError(() => new Error('paper_status_timed_out')),
+      })
+    );
+  };
+}
+
+function pollAuthStatus(): InvokeConfig<Context, PaperHandlerStatusEvent> {
```
kshen0, line 228: I don't love adding more polling, but this seemed cleaner than expecting the frontend to "watch" auth changes and change the backend state machine state via `useEffect`

apps/mark-scan/frontend/src/pages/poll_worker_screen.tsx
```diff
@@ -230,7 +230,14 @@ export function PollWorkerScreen({
     setIsConfirmingEnableLiveMode(false);
   }
 
-  if (hasVotes && pollWorkerAuth.cardlessVoterUser) {
+  if (
+    hasVotes &&
+    pollWorkerAuth.cardlessVoterUser &&
+    // It's expected there are votes in app state if the state machine reports a blank page after printing.
+    // The paper was likely inserted upside down so the solution is to reload paper correctly and go back to
+    // the voting screen
```
kshen0, line 238: I don't like leaving super specific comments like this that might diverge from implementation, but I think it's important to explain why we're ok with breaking a previous constraint (that you can't enter a voting session with existing frontend ballot state)

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
jonahkagan: Agree that polling on the backend is cleaner

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -324,15 +369,87 @@ export function buildMachine(
...
+                ? context.interpretation[0].interpretation.type ===
+                  'InterpretedBmdPage'
+                : false,
+          },
+          {
+            target: 'blank_page_interpretation',
+            cond: (context) =>
+              context.interpretation
+                ? context.interpretation[0].interpretation.type === 'BlankPage'
+                : false,
```
jonahkagan, line 395: What happens if there is no interpretation? It seems like both conds would be false and we'd get stuck in this state. Seems like maybe we expect `context.interpretation` to always be defined, in which case I think an `assert` would be more clear

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -324,15 +369,87 @@ export function buildMachine(
...
+            entry: async (context) => {
+              await context.driver.loadPaper();
+              await context.driver.parkPaper();
+            },
+            on: {
+              PAPER_PARKED: 'done',
+              NO_PAPER_ANYWHERE: 'accepting_paper',
+            },
+          },
+          done: {
```
jonahkagan, line 429: Could probably remove this state by adding the `assign` action contained here to the transition from `load_paper`

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -324,15 +369,87 @@ export function buildMachine(
...
+        onDone: { target: 'paper_reloaded' },
+      },
+      // `paper_reloaded` could be a substate of `blank_page_interpretation` but by keeping
+      // them separate we can avoid exposing all the substates of `blank_page_interpretation`
+      // to the frontend.
+      paper_reloaded: {
+        invoke: pollAuthStatus(),
+        entry: () => {
+          debug('Placeholder paper_reloaded');
+        },
```
jonahkagan, line 448: Remove debug?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -562,6 +682,13 @@ export async function getPaperHandlerStateMachine(
           return 'resetting_state_machine_after_jam';
         case state.matches('resetting_state_machine_after_success'):
           return 'resetting_state_machine_after_success';
+        case state.matches('transition_interpretation'):
+          return 'transition_interpretation';
```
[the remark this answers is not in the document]
kshen0: The frontend doesn't need to know about it, but if we don't handle it then the switch statement will fall through to `no_hardware` and the frontend will show an error. The chances of that happening are pretty slim - it would have to query status at the exact moment the transition state is running.

I think the best way to handle is to just return `interpreting` here.

apps/mark-scan/frontend/src/pages/paper_reloaded_page.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useHistory } from 'react-router-dom';
+
+export function PaperReloadedPage(): JSX.Element {
+  // Messy, but this flow eventually ends up in BallotContext. By setting path now
+  // we ensure we render the correct page once poll worker auth is ended. If we don't
+  // set the path now, we'll incorrectly render PrintPage (the last ballot screen rendered).
+  const history = useHistory();
+  useEffect(() => {
+    history.push('/ready-to-review');
+  });
```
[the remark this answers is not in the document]
kshen0: I agree - if we can get rid of frontend routing it would be a big win. Not sure yet if doing that is blocked by migrating the remaining legacy frontend state to the backend. I think this change would be too big to fit into this PR so I've created https://github.com/votingworks/vxsuite/issues/3985 to track.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Sorry, I think I cleaned this up but accidentally overwrote with a `git reset` before committing.

kshen0:
All suggestions addressed in [7669f7e](https://github.com/votingworks/vxsuite/pull/3981/commits/7669f7e9cf002622ebb94dc03bc725bec9875e5f) and [fa1cb2b](https://github.com/votingworks/vxsuite/pull/3981/commits/fa1cb2b010d5270d1db1632826ff6d40e7aa636a)

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
jonahkagan: I was thinking with the assert that you could remove the conditional check. So something like this:

```ts
assert(context.interpretation)
return context.interpretation[0].interpretation.type === 'BlankPage`
```
or even
```ts
assertDefined(context.interpretation)[0].interpretation.type === 'BlankPage'
```
Sorry I wasn't clear enough!

jonahkagan (approved):

apps/mark-scan/frontend/src/pages/paper_reloaded_page.tsx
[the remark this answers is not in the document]
jonahkagan: Cool, I agree it's probably better to do as a separate refactor.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Oops sorry your original comment was clear, just rushed the fix on my part. [700b335](https://github.com/votingworks/vxsuite/pull/3981/commits/700b3351b465b1f80fe509abb721c978ab0165ba) uses `assertDefined` and removes the conditional check.

It also adds an assertion on interpretation type in `entry`. It's redundant with the check in the `cond` transitions but
* it's valuable to throw on unexpected interpretation types. Without this assertion we'd get stuck in `transition_interpretation`
* Typescript isn't aware that `entry` is guaranteed to precede `cond` transitions so we need to re-assert with `assertDefined`

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Agree the `assign` call can go anywhere, but the state machine breaks when I get rid of the `done` state and transitioning from `load_paper`: https://app.circleci.com/pipelines/github/votingworks/vxsuite/12078/workflows/b8f38ca0-e4cd-4e24-bc63-48b103b367a0/jobs/400580

It's expecting the target state to be a sibling of `load_paper` ie. a child of `blank_page_interpretation` _unless_ we transition from the `done` state. When transitioning from `done` it can successfully transition to top-level states.

kshen0:
QA'd these changes after [700b335](https://github.com/votingworks/vxsuite/pull/3981/commits/700b3351b465b1f80fe509abb721c978ab0165ba)

PR #3984, opened 2023-09-20
Detect when a newly inserted VxScan USB needs CVRs (re)synced

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3924

Before continuous export, if a USB was removed from a VxScan mid-day, and a different one was connected in its place, say because the original USB failed and was no longer able to mount, no additional action would need to be taken. With continuous export, if a USB is removed mid-day, and a different one is connected in its place, CVRs need to be rewritten to the new USB, i.e. the USB needs CVRs (re)synced. This PR adds code and a UX for this.

## Demo Video

In this video, you'll see that I remove a USB and insert another in its place, triggering the (re)sync UX.

https://github.com/votingworks/vxsuite/assets/12616928/a6efc717-e354-4ff7-b66f-9212622cbffe

## Testing

- [x] Added VxScan frontend and backend tests
- [x] Tested manually

`libs/backend` tests are coming in a follow-up PR

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Adding logging in a follow-up PR

### DIFF
--- a/apps/scan/frontend/src/components/cast_vote_record_sync_modal.tsx
+++ b/apps/scan/frontend/src/components/cast_vote_record_sync_modal.tsx
@@ -1,0 +1,91 @@
+import { useState } from 'react';
+import { throwIllegalValue } from '@votingworks/basics';
+import {
+  Button,
+  Loading,
+  Modal,
+  P,
+  useQueryChangeListener,
+} from '@votingworks/ui';
+
+import { exportCastVoteRecordsToUsbDrive, getUsbDriveStatus } from '../api';
+
+type ModalState = 'closed' | 'prompt' | 'syncing' | 'success' | 'error';
+
+export function CastVoteRecordSyncModal(): JSX.Element | null {
+  const usbDriveStatusQuery = getUsbDriveStatus.useQuery();
+  const exportCastVoteRecordsToUsbDriveMutation =
+    exportCastVoteRecordsToUsbDrive.useMutation();
+
+  const [modalState, setModalState] = useState<ModalState>('closed');
+
+  useQueryChangeListener(usbDriveStatusQuery, (newUsbDriveStatus) => {
+    if (newUsbDriveStatus.doesUsbDriveRequireCastVoteRecordSync) {
+      setModalState('prompt');
+    }
+  });
+
+  function syncCastVoteRecords() {
+    setModalState('syncing');
+    exportCastVoteRecordsToUsbDriveMutation.mutate(
+      { mode: 'full_export' },
+      {
+        onSuccess: (result) => {
+          if (result.isErr()) {
+            setModalState('error');
+            return;
+          }
+          setModalState('success');
+        },
+      }
+    );
+  }
+
+  function closeModal() {
+    setModalState('closed');
+  }
+
+  switch (modalState) {
+    case 'closed': {
+      return null;
+    }
+    case 'prompt': {
+      return (
+        <Modal
+          title="CVR Sync Required"
+          content={<P>CVRs need to be synced to the inserted USB drive.</P>}
+          actions={
+            <Button variant="primary" onPress={syncCastVoteRecords}>
+              Sync CVRs
+            </Button>
+          }
+        />
+      );
+    }
+    case 'syncing': {
+      return <Modal content={<Loading>Syncing CVRs</Loading>} />;
+    }
+    case 'success': {
+      return (
+        <Modal
+          title="CVR Sync Complete"
+          content={<P>Voters may continue casting ballots.</P>}
+          actions={<Button onPress={closeModal}>Close</Button>}
+          onOverlayClick={closeModal}
+        />
+      );
+    }
+    case 'error': {
+      return (
+        <Modal
+          title="CVR Sync Error"
+          content={<P>Try inserting a different USB drive.</P>}
+        />
+      );
+    }
+    /* istanbul ignore next: Compile-time check for completeness */
+    default: {
+      throwIllegalValue(modalState);
+    }
+  }
+}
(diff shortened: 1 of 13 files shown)


libs/backend/src/cast_vote_records/export.ts
```diff
@@ -165,12 +178,10 @@ async function getExportDirectoryPathRelativeToUsbMountPoint(
...
-  if (exportDirectoryNeedsCreation) {
-    await fs.mkdir(
-      path.join(usbMountPoint, exportDirectoryPathRelativeToUsbMountPoint),
-      { recursive: true }
-    );
-  }
+  await fs.mkdir(
+    path.join(usbMountPoint, exportDirectoryPathRelativeToUsbMountPoint),
+    { recursive: true }
+  );
```
arsalansufi, line 184: Minor change to existing logic. Rather than only calling `fs.mkdir` when we believe it to be necessary, I'm now always calling it. The operation is inexpensive and idempotent, and after the addition of resync logic, the previous code was actually missing cases where export directory creation was necessary.

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -668,5 +679,81 @@ export async function exportCastVoteRecordsToUsbDrive(
...
+    }
+    return false;
+  })();
+  doesUsbDriveRequireCastVoteRecordSyncCachedResult = {
+    previousUsbDriveStatus: usbDriveStatus,
+    previousValue: value,
+  };
+
+  return value;
+}
```
arsalansufi, line 759: Going to add more complete tests for this function when I add `libs/backend` tests in a follow-up PR. But I do add some basic tests in this PR after wiring everything together.

apps/scan/frontend/src/components/cast_vote_record_sync_modal.tsx
```diff
@@ -0,0 +1,91 @@
...
+  Modal,
+  P,
+  useQueryChangeListener,
+} from '@votingworks/ui';
+
+import { exportCastVoteRecordsToUsbDrive, getUsbDriveStatus } from '../api';
+
+type ModalState = 'closed' | 'prompt' | 'syncing' | 'success' | 'error';
+
+export function CastVoteRecordSyncModal(): JSX.Element | null {
```
arsalansufi, line 15: This same code path and UX could be used for recovery after unexpected failures partway through an export operation, which is cool! Planning to add in a follow-up PR

libs/backend/src/cast_vote_records/export.ts
arsalansufi, line 184: Minor change to existing logic. Rather than only calling `fs.mkdir` when we believe it to be necessary, I'm now always calling it. The operation is inexpensive and idempotent, and after the addition of resync logic, the previous code was actually missing cases where export directory creation was necessary.

libs/backend/src/cast_vote_records/export.ts
arsalansufi, line 759: Going to add more complete tests for this function when I add `libs/backend` tests in a follow-up PR. But I do add some basic tests in this PR after wiring everything together.

apps/scan/frontend/src/components/cast_vote_record_sync_modal.tsx
arsalansufi, line 15: This same code path and UX could be used for recovery after unexpected failures partway through an export operation, which is cool! Planning to add in a follow-up PR.

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -668,5 +679,81 @@ export async function exportCastVoteRecordsToUsbDrive(
     return exportSignatureFileResult;
   }
 
+  clearDoesUsbDriveRequireCastVoteRecordSyncCachedResult();
+
```
adghayes, line 683: At the end of the export, I assume that we wouldn't require any sync. Why not update the cached value rather than clearing it, as if we don't know?

adghayes (approved):
LGTM, but I feel that the prompt text should be more detailed. Even for an election administrator, "CVRs need to be synced to the inserted USB" seems opaque. For me, that sentence doesn't indicate the problem that needs to be solved. Maybe something like "The inserted USB drive does not contain an up-to-date record of the ballots scanned at this scanner. Cast vote records (CVRs) need to be synced to the USB drive." What do you think?

arsalansufi:
@adghayes Good nudge!

We started with the more descriptive "The CVRs on the inserted USB drive are out of sync with the CVRs on the machine.", but decided that that was too in the weeds. So we switched to "CVRs need to be synced to the inserted USB drive." for simplicity. But I agree that the new message is opaque - we've maybe swung too far in the other direction.

I like your proposed copy as a more user-friendly / less technical version of the original, more descriptive copy. I've made one further tweak and landed on "The inserted USB drive does not contain up-to-date records of the votes cast at this scanner. Cast vote records (CVRs) need to be synced to the USB drive." (consistently talking about votes cast instead of both ballots scanned and votes cast). Updated the video in the PR description accordingly.

libs/backend/src/cast_vote_records/export.ts
arsalansufi: I considered that but ultimately decided that it was safer to just let the `doesUsbDriveRequireCastVoteRecordSync` function recompute, the performance impact being negligible. If we were to update at the end of the export function, we'd have to take extra care to only update if the export was a full export (and not a single-CVR export). Made me realize that updating at the end of export is trickier than it seems and could lead to bugs down the road as the export function is modified.

Issue #3890, opened 2023-08-28
Remove old CVR code


arsalansufi:
Reminder that this includes the `legacy_cast_vote_records` artifact type in `libs/auth`

Issue #3905, opened 2023-08-30
fixtures: merge `libs/cvr-fixture-generator` into `libs/fixtures`


adghayes:
We'd like to be able to auto-generate our CVR fixtures, which is a lot less convoluted if these two libraries are the same.

adghayes:
we've gone a different direction, moving all fixture generation out of `libs/fixtures`

PR #3988, opened 2023-09-21
Update libs/cvr-fixture-generator to use new CVR export format (plus other small tweaks)

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3890

A very first step in removing old CVR code, this PR updates `libs/cvr-fixture-generator` to use the new CVR export format. In my next PR, I'll actually generate new fixtures and have app tests use them.

This PR also makes various small tweaks to `libs/backend` functions, all geared around prep for removal of old CVR code.

## Testing

- [x] Updated automated CVR fixture generator tests
- [x] Ran CVR fixture generator and manually validated the generated contents

## Checklist

### DIFF
--- a/libs/backend/src/ballot_package/ballot_package_io.ts
+++ b/libs/backend/src/ballot_package/ballot_package_io.ts
@@ -94,17 +94,15 @@ export async function readBallotPackageFromUsb(
     return filepathResult;
   }
 
-  const artifactAuthenticationResult =
-    await authenticateArtifactUsingSignatureFile({
-      type: 'election_package',
-      filePath: filepathResult.ok(),
-    });
-  if (
-    artifactAuthenticationResult.isErr() &&
-    !isFeatureFlagEnabled(
-      BooleanEnvironmentVariableName.SKIP_BALLOT_PACKAGE_AUTHENTICATION
-    )
-  ) {
+  const artifactAuthenticationResult = isFeatureFlagEnabled(
+    BooleanEnvironmentVariableName.SKIP_BALLOT_PACKAGE_AUTHENTICATION
+  )
+    ? ok()
+    : await authenticateArtifactUsingSignatureFile({
+        type: 'election_package',
+        filePath: filepathResult.ok(),
+      });
+  if (artifactAuthenticationResult.isErr()) {
     await logger.log(LogEventId.BallotPackageLoadedFromUsb, 'system', {
       disposition: 'failure',
       message: 'Ballot package authentication erred.',
--- a/libs/backend/src/cast_vote_records/export.ts
+++ b/libs/backend/src/cast_vote_records/export.ts
@@ -24,8 +24,11 @@ import {
   CastVoteRecordExportMetadata,
   CVR,
   ElectionDefinition,
+  Id,
   MarkThresholds,
+  PageInterpretation,
   PollsState,
+  SheetOf,
   unsafeParse,
 } from '@votingworks/types';
 import { UsbDrive, UsbDriveStatus } from '@votingworks/usb-drive';
@@ -48,12 +51,7 @@ import {
   describeSheetValidationError,
 } from './canonicalize';
 import { readCastVoteRecordExportMetadata } from './import';
-import { ResultSheet } from './legacy_export';
 import { buildElectionOptionPositionMap } from './option_map';
-
-type ExportCastVoteRecordsToUsbDriveError =
-  | { type: 'invalid-sheet-found'; message: string }
-  | ExportDataError;
 
 /**
  * The subset of scanner store methods relevant to exporting cast vote records
@@ -109,6 +107,32 @@ interface ExportContext {
   scannerStore: ScannerStore;
   usbMountPoint: string;
 }
+
+/**
+ * A sheet to be included in a CVR export
+ */
+export interface ResultSheet {
+  readonly id: Id;
+  readonly batchId: Id;
+  readonly interpretation: SheetOf<PageInterpretation>;
+  readonly frontImagePath: string;
+  readonly backImagePath: string;
+
+  /**
+   * Required per VVSG 2.0 1.1.5-G.7 but only relevant for central scanners. On precinct scanners,
+   * this would compromise voter privacy.
+   */
+  readonly indexInBatch?: number;
+
+  /**
+   * TODO: Determine whether this field is still used and, if not, remove
+   */
+  readonly batchLabel?: string;
+}
+
+type ExportCastVoteRecordsToUsbDriveError =
+  | { type: 'invalid-sheet-found'; message: string }
+  | ExportDataError;
 
 //
 // Helpers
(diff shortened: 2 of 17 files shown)


libs/backend/src/ballot_package/ballot_package_io.ts
```diff
@@ -94,17 +94,15 @@ export async function readBallotPackageFromUsb(
...
-  ) {
+  const artifactAuthenticationResult = isFeatureFlagEnabled(
+    BooleanEnvironmentVariableName.SKIP_BALLOT_PACKAGE_AUTHENTICATION
+  )
+    ? ok()
+    : await authenticateArtifactUsingSignatureFile({
+        type: 'election_package',
+        filePath: filepathResult.ok(),
+      });
+  if (artifactAuthenticationResult.isErr()) {
```
arsalansufi, line 105: Thought it made more sense to actually skip the auth check, instead of still performing the auth check and ignoring the result, when the relevant feature flag is set

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -110,6 +108,32 @@ interface ExportContext {
   usbMountPoint: string;
 }
 
+/**
+ * A sheet to be included in a CVR export
+ */
+export interface ResultSheet {
```
arsalansufi, line 114: Moved this from `legacy_export.ts` to `export.ts` so that `export.ts` doesn't have to import from `legacy_export.ts`. A precursor to fully deleting `legacy_export.ts` in an upcoming PR

libs/backend/src/cast_vote_records/import.ts
```diff
@@ -146,11 +148,12 @@ async function* castVoteRecordGenerator(
 
     // Only relevant for HMPBs
     let castVoteRecordBallotSheetId: number | undefined;
+    const isHandMarkedPaperBallot = Boolean(castVoteRecord.BallotSheetId);
     if (castVoteRecord.BallotSheetId) {
       const parseBallotSheetIdResult = safeParseNumber(
         castVoteRecord.BallotSheetId
       );
-      if (parseResult.isErr()) {
+      if (parseBallotSheetIdResult.isErr()) {
```
arsalansufi, line 156: 🙈 Was referencing the wrong result variable here

Will be adding tests soon to get all the new CVR export code back up to 100% code coverage

libs/backend/src/cast_vote_records/import.ts
```diff
@@ -188,18 +191,23 @@ async function* castVoteRecordGenerator(
...
-        )
-      ) as [string, string];
+      const layoutFilePaths = isHandMarkedPaperBallot
+        ? (ballotImageLocations.map((location) =>
+            path.join(
+              castVoteRecordDirectoryPath,
+              `${path.parse(location).name}.layout.json`
+            )
+          ) as [string, string])
+        : undefined;
```
arsalansufi, line 201: Realized that I wasn't accounting for the fact that layout files won't exist for BMD ballots

libs/backend/src/cast_vote_records/import.ts
```diff
@@ -256,12 +263,31 @@ export async function readCastVoteRecordExport(
...
     castVoteRecordExportMetadata,
-    castVoteRecords,
+    castVoteRecordIterator,
   });
 }
+
+/**
+ * Determines whether or not a cast vote record report was generated on a machine in test mode
+ */
+export function isTestReport(
```
arsalansufi, line 279: Another function that I needed to move to unblock removing an old file, in this case `legacy_import.ts`, in an upcoming PR

libs/cvr-fixture-generator/src/cli/generate/main.ts
```diff
@@ -58,6 +52,8 @@ interface IO {
 
 /**
  * Command line interface for generating a cast vote record file.
+ *
+ * TODO: Make more full use of export functions in libs/backend to avoid duplicating logic.
```
arsalansufi, line 56: I'm pretty excited about this idea! My thinking:

- Replace the CVR-generating function in this lib with a function that generates `ResultSheet` objects
- Prepare a mock USB drive and mock store and pass them and the above `ResultSheet` objects to `exportCastVoteRecordsToUsbDrive` from `libs/backend`, to have that function do all the heavy lifting of generating a CVR fixture

Then, we wouldn't have to worry about keeping `libs/backend` and `libs/cvr-fixture-generator` in sync.

I'd tackle this now, but I want to first finish removing old code, and this is a bit of a tangent.

libs/cvr-fixture-generator/src/generate_cvrs.ts
```diff
@@ -371,15 +367,11 @@ export function* generateCvrs({
...
-                            ? frontImageFileUri
-                            : undefined,
+                          Location: frontImageFileUri,
                         },
                         {
                           '@type': 'CVR.ImageData',
-                          Location: backHasWriteIns
-                            ? backImageFileUri
-                            : undefined,
+                          Location: backImageFileUri,
```
arsalansufi, line 374: A bit in the weeds but in new CVR code, I'm taking an all or nothing approach to ballot images. For every ballot, either both front and back images are included, or no images are included.

I think that this will ultimately make code easier to reason about. And there's even a product argument to be made. During one of the demos that I attended in NH, a request came up to display both sides of a ballot for write-in adjudication, even if only one side had a write-in, to provide the maximal possible context when adjudicating.

libs/cvr-fixture-generator/src/cli/generate/main.ts
adghayes: That would be wonderful. As such it's like we are maintaining a separate CVR format entirely.

libs/cvr-fixture-generator/src/generate_cvrs.ts
adghayes: I agree. When I started on the CVR directories, I wanted to do only the relevant pages for write-ins. In the context where everything had to be exported at the end of the day, there was some reason to do that. But with continuous export, I think it makes much more sense to include all images. There's also a way to interpret VVSG that it's necessary to do so. Regardless of whether we buy into that interpretation, Roe and I have previously discussed that including all images is a better story for audit-ability.

Maybe this would also allow us to just use CVRs for some (mostly internal) use cases where we use backups? Not sure.

libs/cvr-fixture-generator/src/cli/generate/main.ts
adghayes: My only concern is that _ideally_, we'd like to be able to auto-generate CVRs in `libs/fixtures`. A little bit ago I created #3905.

Introducing more dependencies could make that harder.

But perhaps the solution is to allow those dependencies and _not_ merge the two libraries, but rather use scripting to call one library from the other.

adghayes (approved):

libs/cvr-fixture-generator/src/cli/generate/main.ts
arsalansufi: Ah yes thanks for pointing that out! I like the idea of using scripts to call from one library to another, instead of a hard dependency

Issue #3983, opened 2023-09-19
Add logging for user and system actions in VxMarkScan


kshen0:

kshen0:
https://github.com/votingworks/vxsuite/pull/4569

PR #3997, opened 2023-09-22
Kevin/pat device integration

kshen0:
## Overview

VVSG 2.0 requires support for dual-switch inputs ie. inputs that can send exactly 2 different signals. A common implementation is a [sip & puff device](https://www.orin.com/access/sip_puff/), but we can't assume exactly which device will be used. This PR adds support for dual-switch PAT devices:

* Adds general support so the device can be used to navigate the DOM and select the currently highlighted element, much like the existing accessible controller
* Adds a calibration flow to help the voter identify which input maps to which action
* Adds an environment variable to support developing in a VM
  * The flag disables hardware checks for the paper handler and allows `app_root` to render even when no hardware is found
  * Skips the paper loading step of voter session setup
  * Makes `setAcceptingPaperState` endpoint a no-op since it's called by frontend in the beginning of the voting flow
  * Breaks if you try to print a ballot
    * Supporting VM development for all flows would require a full hardware mock, but for much less work this allows us to develop parts of VxMarkScan that don't rely on printer/scanner status. Up until now I've been commenting out various checks but it felt more stable to add this env var. 

Note that this PAT implementation relies on the PAT device registering as an [HID device](https://en.wikipedia.org/wiki/Human_interface_device). It's somewhat unlikely that the built-in 3.5mm PAT input on the actual hardware will register as an HID device, so this implementation may change.

## Demo Video or Screenshot

**Hardware check disabled**
Screenshot 2023-10-04 at 10 47 25 AM

**Step 0: Instructions screen**
Screenshot 2023-10-05 at 3 19 19 PM

**Step 1.1: Identify "Move" Input**
Screenshot 2023-10-05 at 3 24 11 PM

**Step 1.2: "wrong" input triggered**
Screenshot 2023-10-05 at 3 24 24 PM

**Step 1.3: "Move" Input successfully identified**
Screenshot 2023-10-05 at 3 24 17 PM

Step 2 is the same as Step 1, except for the "Select" input instead of "Navigate"

**Step 3: Continue with voting or repeat identification flow**
Screenshot 2023-10-05 at 12 45 18 PM

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (Logging hasn't yet been prioritized, tracked in https://github.com/votingworks/vxsuite/issues/3983)


apps/mark-scan/frontend/src/api.ts
```diff
@@ -112,7 +116,11 @@ export const getStateMachineState = {
   useQuery() {
     const apiClient = useApiClient();
     return useQuery(this.queryKey(), () => apiClient.getPaperHandlerState(), {
-      refetchInterval: STATE_MACHINE_POLLING_INTERVAL_MS,
+      refetchInterval: !isFeatureFlagEnabled(
+        BooleanEnvironmentVariableName.DISABLE_VXMARKSCAN_STATE_MACHINE_POLLING
```
kshen0, line 120: This double negative is confusing but fits our existing pattern of development flags being opt-in eg. DISABLE_CARD_READER_CHECK, SKIP_PIN_ENTRY etc

jonahkagan: You could switch the order of the ternary expression to remove the double negative, if that would help

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -760,6 +783,12 @@ export function AppRoot({
         // Pages that condition on state machine state aren't nested under Ballot because Ballot uses
         // frontend browser routing for flow control and is completely independent of the state machine.
         // We still want to nest pages that condition on the state machine under BallotContext so we render them here.
+        if (!isPatDeviceCalibrated && patHidInput) {
```
jonahkagan, line 786: the comment above seems like it doesn't quite apply to this case

apps/mark-scan/frontend/src/pages/pat_device_calibration_page.tsx
jonahkagan, line 1: When I watched the video of this flow, it wasn't clear to me that I would need to press all the buttons in order to get through the calibration step. I wonder if it might be clearer to instruct the user to press each button in sequence, like we do for the accessible controller diagnostic flow in VxMark. It also might be nice to make those flows consistent in styling and such, though maybe not crucial.

libs/ui/src/hooks/use_devices.ts
```diff
@@ -38,6 +39,7 @@ export interface Devices {
   precinctScanner?: KioskBrowser.Device;
   batchScanner?: KioskBrowser.Device;
   printer?: KioskBrowser.PrinterInfo;
+  patHidInput?: KioskBrowser.Device;
```
jonahkagan, line 42: This moves us a bit further away from removing our reliance on kiosk-browser. Is there an easy way to read the PAT device status from the backend?

kshen0: I spent some time experimenting with this. We can successfully read PAT device status and input from the backend using `node-hid` but I think the frontend still needs access to the device because its primary purpose outside of this calibration flow is as an input device during the voting flow. We could build the calibration flow in the backend without much trouble because there are only 4 states. But it feels like listening to PAT input events and transmitting them to the frontend to handle DOM navigation would be messy. eg. the frontend would need to subscribe to or poll the backend for every single PAT event and simulate a browser `Tab` keypress/click event in response.

As long as we need the frontend to be aware of the PAT device for general voting flow, the marginal cost of using the frontend-integrated PAT device for the calibration flow is small to none.

That said, I'm not up to date on what the plan is to get rid of kiosk-browser. Is there something you had in mind that would be a cleaner solution than the one I mentioned above?

apps/mark-scan/frontend/src/pages/pat_device_calibration_page.tsx
kshen0: That's a good point - in general the styling in VxMarkScan needs another pass to bring it closer to the other apps.

I'll try to polish the style so it's closer to the accessible controller diagnostic flow. The main difference is we don't know what device the voter is using so we can't have a diagram or even opinionated instructions about what to press or sip/puff. But the header styling and step numbers should help with the ambiguity.

kshen0: The main changes I made were
* Added header
* Got rid of modals
* Structured code more similarly to the accessible controller diagnostic screen

Without a vector graphic or "This input is broken" button the pages still look somewhat different. I'm not sure what the best action is if a PAT device input is broken - since voters supply their own devices we can't recommend asking a poll worker for a replacement. In the meantime I've at least added a button to skip calibration so the poll worker can bail out of the flow. They can also just unplug the device to exit.

apps/mark-scan/frontend/src/app_pat_input.test.tsx
```diff
@@ -0,0 +1,127 @@
+import { electionGeneralDefinition } from '@votingworks/fixtures';
```
kshen0, line 1: Ignore these - will update when we settle on UI

libs/ui/src/hooks/use_devices.ts
jonahkagan: It seems like there are two different aspects to interacting with the PAT device:
- Device status (is it connected?)
- Device input (select/navigate triggers)

I think we currently use custom kiosk-browser functionality for device status (via this `useDevices` hook), but use regular web browser standard interfaces for device input. Is that true?

Eventually I think we want to move device status to the backend so we can migrate away from kiosk-browser. I was suggesting that this might be a good time to start moving device status to the backend. I totally agree that it makes sense to keep using the web browser standard interface for input.

But, I also think that continuing to use the existing device status from kiosk browser doesn't make it that much harder to move the backend later, so it's also fine to leave it as is for now.

apps/mark-scan/frontend/src/pages/pat_device_calibration_page.tsx
jonahkagan: Nice, I think this seems great!

A few suggestions to play with:
- Centered paragraph text is hard to read because each new line starts at a different x-coordinate. Maybe try left-aligning either the paragraph text or all of the text?
- I wonder if there's an icon we could use in place of the device graphics - e.g. a question mark circle for the prompt, then a checkmark circle or X circle on success/failure.
- Are "Navigate" and "Select" the standard names for the two inputs? Just curious if users will already understand these terms or if we should make it a bit more explanatory. And/or consider using different terms.

kshen0: Good point. I did try both left-aligned options and thought it looked off-balance. I think adding the graphics in your second bullet will correct the balance.

I don't think there's a standard for naming:
* [This resource](https://accessibleweb.com/assistive-technologies/assistive-technology-focus-sip-and-puff-devices/) uses "navigate", "open links", "activate buttons"
* [This resource](https://www.boia.org/blog/understanding-assistive-technologies-what-are-sip-and-puff-systems) uses "scrolling" and "activate a hyperlink"
* WCAG 2.1 calls it ["navigating focus"](https://www.w3.org/TR/WCAG21/#focus-order) or ["moving focus"](https://www.w3.org/TR/WCAG21/#no-keyboard-trap)

From this we could be a little more verbose and call it "navigating focus" and "activating focus"

libs/ui/src/hooks/use_devices.ts
kshen0: I was thinking device management should be all-or-nothing in the frontend or backend, but that might apply less to input devices and more to devices like the card reader. I agree that we can break down the concerns cleanly between status and input. I'll update this branch to remove kiosk-browser status and move it to the backend.

apps/mark-scan/frontend/src/pages/pat_device_calibration_page.tsx
jonahkagan: Thanks for sharing those resources. If there's no clear standard, I actually think what you chose already is pretty clear.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -289,267 +362,306 @@ export function buildMachine(
...
-        invoke: [
-          pollPaperStatus(),
-          {
-            id: 'loadAndPark',
-            src: (context) => {
-              return loadAndParkPaper(context.driver);
+    },
+    invoke: [pollPatDeviceConnectionStatus()],
+    states: {
+      voting_flow: {
```
kshen0, line 387: Nesting the voting flow in its own state like this allows us to use a history node.

From docs:
> A history [state node](https://xstate.js.org/docs/guides/statenodes.html) is a special kind of state node that, when reached, tells the machine to go to the last state value of that region

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -289,267 +362,306 @@ export function buildMachine(
...
       },
-      accepting_paper: {
-        invoke: pollPaperStatus(),
-        on: {
-          PAPER_READY_TO_LOAD: 'loading_paper',
-        },
+      PAT_DEVICE_DISCONNECTED: {
+        actions: assign({
+          patDevice: () => undefined,
+        }),
```
kshen0, line 379: Setting `context.patDevice = undefined` ensures we re-enter the calibration state when another device is connected.

apps/mark-scan/backend/README.md
```diff
@@ -9,6 +9,21 @@ Follow the instructions in the [VxSuite README](../../../README.md)
...
+SUBSYSTEM=="input", GROUP="input", MODE="0666"
+SUBSYSTEM=="usb", ATTR{idVendor}=="0a95", ATTR{idProduct}=="0012", MODE="0666", GROUP="plugdev"
+KERNEL=="hidraw*", ATTRS{idVendor}=="0a95", ATTRS{idProduct}=="0012", MODE="0666", GROUP="plugdev"
+```
+
+then run
+
+```
+sudo udevadm control --reload-rules && sudo udevadm trigger
+```
```
jonahkagan, line 25: Should we add this to automated dev env setup? cc @amcmanus

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -290,267 +362,306 @@ export function buildMachine(
...
-        on: {
-          // Paper may be inside the machine from previous testing or machine failure. We should eject
-          // the paper (not to ballot bin) because we don't know whether the page has been printed.
-          PAPER_INSIDE_NO_JAM: 'eject_to_front',
-          PAPER_PARKED: 'eject_to_front',
-          BEGIN_ACCEPTING_PAPER: 'accepting_paper',
-        },
+      PAPER_JAM: 'voting_flow.jammed',
+      JAMMED_STATUS_NO_PAPER: 'voting_flow.jam_physically_cleared',
+      PAT_DEVICE_CONNECTED: {
```
jonahkagan, line 370: Interesting, I hadn't imagined this being part of the state machine. I had kind of assumed it would just be another separate backend API similar to how we interface with the card reader for auth. But I can see a case for it being integrated here if we want to interrupt the voting flow when the PAT device is connected but not calibrated. It's interesting to try to figure out these new patterns as we move more stuff to the backend.

That makes me think - do we need to interrupt the voting flow to calibrate the PAT device? How/when do we imagine this device getting set up and calibrated? With the accessible controller, it seems like we have a flow where it gets calibrated from a menu as needed, not in real time when connected.

apps/mark-scan/backend/README.md
kshen0: If it's trivial to add! If not trivial, we can wait since these details (for example, vendor and device ID) will change when we integrate with the built-in 3.5mm port.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Great questions! You're correct that the main reason I made this part of the state machine was to support the ability to interrupt at any time.

The [product brief](https://docs.google.com/document/d/1lYqziJK7cPovJlWDthYJZiIPQAVv4UTZ4wPoWcomy8w/edit#heading=h.ktowgi2il3p) specifies that we should interrupt the voting flow to calibrate the device as soon as it's plugged in, and "At any point in the voter session, a user can connect a dual-switch device into the PAT port".

It's worth questioning whether this requirement in the product brief is necessary. I think it's a nice-to-have, since a voter or poll worker might forget to connect the device until they hand off control to the voter, or the voter might decide halfway through that the physical dimensions of the screen and elements require them to use a PAT device.

There's also some benefit to waiting until the cardless voter session is active (as opposed to having the poll worker set it up, like the accessible controller). The accessible controller test seems to be purely to test that the hardware works, whereas the PAT device calibration is primarily to teach the voter how to navigate the DOM in this particular app. Since the voter must be involved in that flow, I don't think we want to start the calibration flow from a menu.

Let me know if that makes sense; happy to discuss more.

jonahkagan: Thanks for sharing that context! I hadn't read the product brief before. I agree with your interpretation that this is primarily about teaching the voter.

Given that, it feel like this flow should be associated with the user session. E.g. if two voters in a row use the PAT device without unplugging it, both should see the PAT device tutorial. Is that how it behaves now? My read is no - that it only shows once per connection of the device.

kshen0: I've been assuming each voter will supply their own device ie. they will disconnect their device and take it with them after casting the ballot. But now that you mention it, I think there's also the possibility that the precinct supplies the device, so we should delete `context.patDevice` on session reset as well.

apps/mark-scan/frontend/src/pages/pat_device_identification/pat_device_identification_page.tsx
```diff
@@ -0,0 +1,65 @@
...
+  useEffect(() => {
+    // Child components override the gamepad keyboard listener while we take
+    // the voter through device identification.
+    document.removeEventListener('keydown', handleGamepadKeyboardEvent);
+
+    // Reattach the gamepad listener on cleanup
+    return () => {
+      document.addEventListener('keydown', handleGamepadKeyboardEvent);
+    };
+  });
```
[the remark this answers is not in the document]
kshen0: I wanted to prevent sip/puff events during the calibration from actually navigating and selecting elements, but in retrospect think `e.preventDefault` also does this. I'll confirm and remove this.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: https://github.com/votingworks/vxsuite/pull/3997/commits/72fb195a1989f9c38a6841b43121ca0fde59ddbd#diff-ea4f2566cf950ca374305ad00616d5eda7f645377dcf5f5cf6167e1fd0add195R631 resets `context.patDevice` and transitively the calibration flow after every voter session

apps/mark-scan/frontend/src/lib/gamepad.ts
```diff
@@ -104,6 +104,12 @@ export function handleGamepadKeyboardEvent(event: KeyboardEvent): void {
     case ']':
       handleClick();
       break;
+    case '1':
+      handleArrowDown();
+      break;
+    case '2':
+      handleClick();
+      break;
```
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/3997/commits/72fb195a1989f9c38a6841b43121ca0fde59ddbd#diff-f5e645f6d53aa188936702c41417e60125aa07a2eca400e069a04b900a65a80cR107

apps/mark-scan/backend/src/custom-paper-handler/cli/state_machine_cli.ts
```diff
@@ -80,11 +83,12 @@ export async function main(): Promise<number> {
...
 
   const stateMachine = await getPaperHandlerStateMachine(
-    driver,
     workspace,
     auth,
     logger,
-    DEV_PAPER_HANDLER_STATUS_POLLING_INTERVAL_MS
+    driver,
+    DEV_DEVICE_STATUS_POLLING_INTERVAL_MS,
+    AUTH_STATUS_POLLING_INTERVAL_MS
```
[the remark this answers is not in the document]
kshen0: Good idea, https://github.com/votingworks/vxsuite/pull/3997/commits/72fb195a1989f9c38a6841b43121ca0fde59ddbd#diff-ea4f2566cf950ca374305ad00616d5eda7f645377dcf5f5cf6167e1fd0add195R739

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -620,14 +621,27 @@ function setUpLogging(
...
+  driver?: PaperHandlerDriverInterface
+): PaperHandlerDriverInterface {
+  if (driver) {
+    debug('Using real driver');
+    return driver;
+  }
+
+  debug('No driver found. Starting state machine with mock driver');
+  return new MockPaperHandlerDriver();
+}
```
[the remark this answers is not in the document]
kshen0: Sort of like this? https://github.com/votingworks/vxsuite/pull/3997/commits/72fb195a1989f9c38a6841b43121ca0fde59ddbd#diff-f0b52a9aee3e9ffc9049ab9e6a9fb4d4133c482230c1b398d5159a2021db6ec3R39

There are 3 cases we want to support
1. Device exists and a real driver is returned
2. Device does not exist and no real driver is returned (so we show an error)
3. Device does not exist and mock driver is returned (for development)

apps/mark-scan/frontend/src/pages/pat_device_identification/identify_input_step.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  return (
+    <StepInnerContainer svgSize="medium" padding="0 40px 0 0">
+      <div>
+        <Text>
+          <H1>{headerContent}</H1>
+          {bodyContent && <P>{bodyContent}</P>}
+        </Text>
+      </div>
+      {icon}
```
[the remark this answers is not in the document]
kshen0: I'll play around with this to see what feels balanced when the icon and text are vertically stacked.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
mattroe: The main difference between the accessible controller input and dual-switch input is that we know the exact controls on the former, but don't necessarily know the latter. There's the possibility that some jurisdictions will provide dual-switch inputs at the polling place (though this is rare) so we have to support a flow that allows any dual-switch input. Even in the case where dual-switches are provided, a voter may still want to bring an alternative form that they are more comfortable with into the polling place and I'd like to be able to accommodate that.

All this is to say: I think we need to have a mapping/calibration step for each voter session, which is how I understand it to be currently implemented?

apps/mark-scan/frontend/src/pages/accessible_controller_diagnostic_screen.tsx
```diff
@@ -1,15 +1,42 @@
...
 import { ScreenReader } from '../config/types';
 import {
   DiagnosticScreenHeader,
   StepContainer,
-  StepInnerContainer,
 } from './diagnostic_screen_components';
 
 type ButtonName = 'Up' | 'Down' | 'Left' | 'Right' | 'Select';
 
+const StepInnerContainer = styled.div`
```
kshen0, line 13: I had factored this out earlier for use in the PAT flow, but since the designs have diverged it no longer made sense to share styles. This just adds this styled element back to this file; no diff from `main`.

apps/mark-scan/frontend/src/pages/pat_device_identification/pat_device_identification_page.tsx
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/3997/commits/ec3ea46da5d208dc7606afa9f645014e54dfddf2#diff-9667732ed6e98f6d65fe7f4b3a79f59dcb33cc61b08060a44ff3679ce5fee507L26

apps/mark-scan/frontend/src/pages/pat_device_identification/pat_device_identification_page.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <Screen>
+      <Main centerChild>
+        <DiagnosticScreenHeader>
+          <P>
+            <Font weight="bold">PAT Device Identification</Font> &mdash; Step{' '}
+            {step + 1} of {steps.length}
+          </P>
+          <Button onPress={onExitCalibration}>Skip Identification</Button>
```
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/3997/commits/ec3ea46da5d208dc7606afa9f645014e54dfddf2#diff-9667732ed6e98f6d65fe7f4b3a79f59dcb33cc61b08060a44ff3679ce5fee507R51

apps/mark-scan/frontend/src/pages/pat_device_identification/pat_introduction_step.tsx
```diff
@@ -0,0 +1,55 @@
...
+            Activate Focus eg. activate the focused candidate&apos;s row to mark
+            a vote for them
+          </li>
+        </ol>
+      </P>
+      <P>
+        The device test will ask you to trigger both device inputs in turn to
+        identify their functions. After both inputs are identified you will be
+        returned to the previous screen.
+      </P>
```
jonahkagan, line 49: Some copy suggestions:
- Use "Navigate" and "Activate" without the word focus to be consistent with the other screens.
- I know we've bike-shedded the specific terms a lot, but I liked "Select" for being a simpler word than "Activate". Maybe "Move" and "Select" are the two simplest terms I can think of. But I don't feel strongly.
- People often skim instructions so maybe trim them down

So maybe something like:

**Test Your Device**
Your device's two inputs can be used to **Move** focus between items on the screen and **Select** an item.

**Trigger any input to begin the test.**

mattroe: Agree with @jonahkagan on copy simplification suggestions. Particularly use of the words "move" and "select".

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Correct, it's per voter session right now. I think the question during review was whether we needed to support

a. connection at _any_ time during the voter flow vs
b. device connection just at the beginning

The implementation changes based on that requirement, but it's a moot point now since we're all on board with A. Thanks for the clarifications!

apps/mark-scan/frontend/src/pages/pat_device_identification/identify_input_step.tsx
```diff
@@ -37,6 +37,8 @@ export function IdentifyInputStep({
       event.preventDefault();
 
       if (event.key === behaviorToKeypressMap[inputName]) {
+        // Exhaustive switch statement so we don't need a default case
+        // eslint-disable-next-line default-case
```
jonahkagan, line 41: Generally our pattern is to have a default case with `throwIllegalValue(variableFromSwitch)` so that the type-checker enforces exhaustiveness.

jonahkagan (approved):
Looks great! Love where this landed

apps/mark-scan/frontend/src/pages/pat_device_identification/identify_input_step.tsx
eventualbuddha: And to add a little more context: the reason this is different than simply suppressing the error is that using `throwIllegalValue` prevents forgetting to update `switch` statements when you add new possible values. Let's say you add a new possible value `'v'` to type `T` and you're not using `throwIllegalValue`. All of the `switch` statements that switch on a value of type `T` will still type check just fine. If you're using `throwIllegalValue` then its argument will have its type narrowed to `'v'` instead of `never`, resulting in a type error. This ensures you won't forget to update the relevant `switch` statements.

PR #4001, opened 2023-09-26
Trusted Build changes

amcmanus:
This PR includes all the changes necessary to support Trusted Builds of VotingWorks applications. Due to the unique requirements of that process, `vxsuite` no longer performs installation of system level tools/frameworks. 

**Makefiles and apt packages:**
- We no longer install packages via Makefiles. This functionality has been moved to `vxsuite-build-system`
- When necessary, Makefiles were updated to continue to support the bootstrap script for ease of use

**Rust:**
- Rust install and build commands (invoked via pnpm) are now explicitly separate. Install is only responsible for fetching cargo crates, and build is responsible for the actual compilation. This is necessary for Trusted Build
- pnpm configuration has been changed to only run Rust as needed (it ran by default previously)
- CircleCI has been updated to support this new build approach

**Python / converter-ms-sems:**
- In an effort to remove python/pip dependencies from `vxsuite`, converter-ms-sems has been disabled during any build processes. It will eventually be removed/rewritten, but that is outside the scope of this PR.

Since CircleCI has also been updated to support these changes, running the test suite serves as the initial test of these changes. Additionally, extensive testing of the build process has been performed as part of the `vxsuite-build-system` Trusted Build effort.


arsalansufi (approved):
LGTM!

PR #4002, opened 2023-09-26
Remove old CVR code from VxCentralScan and VxScan

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3890

Last old CVR code removal PR! Includes removal of the `ENABLE_CONTINUOUS_EXPORT` feature flag, which will now be enabled by default.

## Testing

- [x] Automated testing

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Will be revisiting/checking logging holistically soon


apps/central-scan/backend/src/backup.ts
```diff
@@ -84,21 +83,6 @@ export class Backup {
...
-        definiteMarkThreshold: this.store.getMarkThresholds().definite,
-        isTestMode: this.store.getTestMode(),
-        resultSheetGenerator: this.store.forEachResultSheet(),
-        batchInfo: this.store.getBatches(),
-        reportContext: 'backup',
-      })
-    );
-
-    debug('added CVRs to backup');
-
```
arsalansufi, line 101: As noted in the commit description, I'll be completely revisiting scanner backups in a follow-up PR. This is just a temporary patch, necessary because I've removed the old CVR export function used here (`getCastVoteRecordReportStream`)

apps/scan/backend/src/store.ts
```diff
@@ -565,14 +561,6 @@ export class Store {
...
     }
 
-    if (
-      isFeatureFlagEnabled(
-        BooleanEnvironmentVariableName.ENABLE_CONTINUOUS_EXPORT
-      )
-    ) {
-      return true;
-    }
-
```
arsalansufi, line 575: This is also related to scanner backups. tldr: You have to export a backup before you can unconfigure a scanner. Continuous export should render this moot (because you're essentially always backed up), but I want to think through things fully when I work on overall backup functionality cleanup.

For now, I'm just removing the reference to the `ENABLE_CONTINUOUS_EXPORT` feature flag, which is itself removed in a subsequent commit.

apps/scan/backend/test/helpers/custom_helpers.ts
```diff
@@ -209,6 +209,10 @@ export function simulateScan(
   });
 }
 
+export async function waitForContinuousExportToUsbDrive(): Promise<void> {
```
arsalansufi, line 212: This operation happens in the background after a ballot is accepted, and this is the simplest way to ensure that it's able to finish in relevant tests. Ending a test before continuous export finishes can result in errors due to test directories getting cleaned up while they're still being written to.

apps/scan/backend/src/app_results.test.ts
```diff
@@ -23,49 +19,6 @@ jest.mock('@votingworks/utils', (): typeof import('@votingworks/utils') => {
   };
 });
 
-async function scanBallot(
```
arsalansufi, line 26: A lingering one-off copy of the `scanBallot` helper

apps/scan/backend/src/scanners/custom/app_config.test.ts
```diff
@@ -283,25 +263,17 @@ test('export CVRs to USB', async () => {
...
-  // Just test that the app has been wired properly. Rely on libs/backend tests for more detailed
-  // coverage of export logic.
+test('continuous CVR export, including polls closing, followed by a full export', async () => {
   await withApp(
     {},
     async ({ apiClient, mockAuth, mockScanner, mockUsbDrive }) => {
       await configureApp(apiClient, mockAuth, mockUsbDrive, { testMode: true });
       await scanBallot(mockScanner, apiClient, 0);
       await scanBallot(mockScanner, apiClient, 1);
-      await sleep(1000); // Let background continuous export to USB drive finish
```
arsalansufi, line 304: Build this sleep into `scanBallot`, as you saw in the last commit

apps/scan/backend/src/scanners/custom/app_scan_double_sheets.test.ts
```diff
@@ -90,10 +93,9 @@ test('insert second ballot while first ballot is accepting', async () => {
         DELAY_ACCEPTED_READY_FOR_NEXT_BALLOT: 1000,
         DELAY_ACCEPTED_RESET_TO_NO_PAPER: 2000,
       },
-      interpret: mockInterpret(interpretation),
```
arsalansufi, line 93: The tldr here is that you can only use mocked interpretation (as opposed to real interpretation) if you don't need to export CVRs in your test. With continuous export no longer feature flag gated, the number of tests that export CVRs and can't use mock interpretation has increased.

How I figured this out:

https://github.com/votingworks/vxsuite/blob/1483679cbc86ed63b93da1103b291a323a4c4a78/apps/scan/backend/test/helpers/shared_helpers.ts#L103-L117

apps/scan/backend/src/scanners/custom/app_scan_double_sheets.test.ts
```diff
@@ -113,31 +115,52 @@ test('insert second ballot while first ballot is accepting', async () => {
...
       });
+
+      await waitForContinuousExportToUsbDrive();
     }
   );
 });
 
 test('insert second ballot while first ballot needs review', async () => {
-  const interpretation = needsReviewInterpretation;
+  // const interpretation = needsReviewInterpretation;
```
kshen0: Delete this?

libs/utils/src/environment_variable.ts
```diff
@@ -38,9 +38,6 @@ export enum BooleanEnvironmentVariableName {
   SKIP_CAST_VOTE_RECORDS_AUTHENTICATION = 'REACT_APP_VX_SKIP_CAST_VOTE_RECORDS_AUTHENTICATION',
   // Disables exporting original snapshots with CVRs
   DISABLE_CVR_ORIGINAL_SNAPSHOTS = 'REACT_APP_VX_DISABLE_CVR_ORIGINAL_SNAPSHOTS',
-  // Enables continuous export to USB on VxScan, usage of the corresponding export format on
-  // VxCentralScan, and handling of these exports on VxAdmin
-  ENABLE_CONTINUOUS_EXPORT = 'REACT_APP_VX_ENABLE_CONTINUOUS_EXPORT',
```
kshen0, line 43: Does this need to be removed from the root `.env` file too?

apps/scan/backend/src/scanners/custom/app_scan_double_sheets.test.ts
kshen0: Never mind, I see this is deleted later

libs/utils/src/environment_variable.ts
arsalansufi: Looks like I never added it to that file (oops), so we should be good to go:

```
$ cd vxsuite
$ git grep ENABLE_CONTINUOUS_EXPORT
$
```

apps/central-scan/backend/src/central_scanner_app.test.ts
```diff
@@ -15,17 +15,12 @@ import {
   TEST_JURISDICTION,
 } from '@votingworks/types';
 import { Scan } from '@votingworks/api';
-import {
-  BooleanEnvironmentVariableName,
-  CAST_VOTE_RECORD_REPORT_FILENAME,
-  getFeatureFlagMock,
-} from '@votingworks/utils';
+import { getFeatureFlagMock } from '@votingworks/utils';
```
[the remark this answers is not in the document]
arsalansufi: Good catch, removed: https://github.com/votingworks/vxsuite/pull/4002/commits/f0c478f8d3ee25c06e4742fd766680cb12259ab8

apps/scan/backend/test/helpers/custom_helpers.ts
arsalansufi: Ah yes, good call re a comment! https://github.com/votingworks/vxsuite/pull/4002/commits/5362fff7f48d2468193720be07d5828a67919928

kshen0 (approved):

PR #4003, opened 2023-09-26
[i18n/audio] Read UI strings from USB ballot package

kofi-q:
## Overview

🔍  _Easier to review by commit._

- Adding scaffolding and starter code for extracting election UI strings from the CDF election.
  - Required modifying our election parsing code a bit to expose the parsed CDF election (if in CDF format) in addition to the VXF election. Definitely room for future cleanup/optimization, but picked least disruptive path for now.
- Adding logic for extracting app UI strings and combining with election strings.

## Testing Plan
- Added unit tests for CDF extractor utilities and for UI string combination in the ballot package loader.

## Checklist


libs/utils/src/extract_cdf_ui_strings.test.ts
```diff
@@ -0,0 +1,180 @@
...
+  },
+
+  [ElectionStringKey.STATE_NAME]() {
+    // TODO(kofi): Implement
+  },
+};
+
+for (const [electionStringKey, testFn] of Object.entries(tests)) {
+  test(electionStringKey, testFn);
+}
```
arsalansufi, line 180: I've been a fan of [`test.each` for repetitive tests](https://jestjs.io/docs/api#testeachtablename-fn-timeout) like these. This explicit loop is also totally fine! But just thought I'd note

libs/utils/src/extract_cdf_ui_strings.ts
```diff
@@ -0,0 +1,138 @@
...
+const SUPPORTED_LANGUAGES = new Set<string>(Object.values(LanguageCode));
+
+/**
+ * Sets the appropriate language strings in supported languages for the given
+ * internationalized CDF ballot content text.
+ */
+function setInternationalizedUiStrings(params: {
+  uiStrings: UiStringsPackage;
+  stringKey: ElectionStringKey;
+  entityId?: string;
```
arsalansufi, line 23: For my understanding, could you explain when `entityId` will be specified vs. not?

arsalansufi (approved):
LGTM!

libs/utils/src/extract_cdf_ui_strings.test.ts
kofi-q: Oh, love that! TIL - thanks!

libs/utils/src/extract_cdf_ui_strings.ts
kofi-q: Yeah, glad you asked - wasn't sure of the naming here. It's used as a sub-key for strings with multiple "instances" like `contestTitle` or `candidateName`, where the entity ID would be contest ID or candidate ID, respectively (vs `electionTitle` for example, where there's only one instance).

LMK if that makes any lightbulbs go off for you WRT naming. Was waffling between `stringSubKey` and `entityId`, but neither's too obvious. Kind of leaning `stringSubKey` now though, since it's a bit more explicit WRT what it's used for in the translation mapping.

kofi-q: K, I think https://github.com/votingworks/vxsuite/pull/4003/commits/ceaf8caf918eca9bcf937745ad25a01182116383 improves readability a bit - now defining `stringKey` as either a single `ElectionStringKey` or a pair of `[key: ElectionStringKey, subKey: string]`

arsalansufi: Ah perfect, this clarifies a lot! And I accordingly like the changes introduced in https://github.com/votingworks/vxsuite/commit/ceaf8caf918eca9bcf937745ad25a01182116383

PR #4005, opened 2023-09-27
[Mark/Scan/Mark-Scan] Load UI Strings from BallotPackage into DB

kofi-q:
## Overview

- Added shared app logic and store queries for writing/reading UI Strings to and from backend stores.
- Running that logic within the "configure machine" endpoints in `Mark`, `Scan`, and `Mark-Scan` when loading the ballot package.
  - Updated `Mark` and `Mark-Scan` to wrap configuration store queries in a transaction to match `Scan`'s behaviour.

## Testing Plan
- Using a shared test runner in each app to test against it's API and store 

## Checklist


libs/backend/src/ui_strings/ui_strings_machine_configuration.ts
```diff
@@ -0,0 +1,41 @@
+/* istanbul ignore file - tested via VxSuite apps. */
+
+import { BallotPackage, LanguageCode } from '@votingworks/types';
+import { Logger } from '@votingworks/logging';
+import { UiStringsStore } from './ui_strings_store';
+
+/** Input for {@link processBallotPackage}. */
```
arsalansufi, line 7: I can't seem to find a function with the name `processBallotPackage`, so I'm wondering if this `{@link processBallotPackage}` is a stale reference - maybe you're planning to add this function in the future?

apps/mark-scan/backend/src/app.ts
```diff
@@ -156,21 +157,30 @@ export function buildApi(
...
+      const ballotPackage = ballotPackageResult.ok();
+      const { electionDefinition, systemSettings } = ballotPackage;
       assert(systemSettings);
 
-      workspace.store.setElectionAndJurisdiction({
-        electionData: electionDefinition.electionData,
-        jurisdiction: authStatus.user.jurisdiction,
-      });
-      workspace.store.setSystemSettings(systemSettings);
+      workspace.store.withTransaction(() => {
```
arsalansufi, line 164: Good catch re the prior lack of a transaction!

apps/mark-scan/backend/src/app.ui_strings.test.ts
```diff
@@ -1,26 +1,89 @@
...
-    buildMockInsertedSmartCardAuth(),
-    createMockUsb().mock,
-    fakeLogger(),
-    workspace
-  ),
+  api: buildApi(mockAuth, createMockUsb().mock, fakeLogger(), workspace),
   store: store.getUiStringsStore(),
 });
+
+describe('configureMachine', () => {
```
arsalansufi, line 54: The most minor nit: This describe string `'configureMachine'` makes me think that there's a function/method named `configureMachine` that we're testing. But the method doesn't necessarily have that name, e.g. here it's `configureBallotPackageFromUsb`. `'machine configuration'` could work as a more general-purpose alternative

arsalansufi (approved):
LGTM!

libs/backend/src/ui_strings/ui_strings_machine_configuration.ts
kofi-q: Thanks! You're right about the stale reference - will update

apps/mark-scan/backend/src/app.ui_strings.test.ts
kofi-q: Ah, good nit - updating to match the endpoint methods

PR #4009, opened 2023-09-29
Decrease VxScan backend test flakiness and introduce proper recovery from failures midway through a CVR export operation

arsalansufi:
_Easiest reviewed by commit_

## Overview

VxScan backend tests have gotten quite flaky since my recent merge to enable continuous export by default, without a feature flag. Unsurprisingly, my [rudimentary approach](https://github.com/votingworks/vxsuite/blob/2a1441689e87312d56041930dd8fd8c19ac7ceda/apps/scan/backend/test/helpers/shared_helpers.ts#L137-L146) for waiting for continuous export to finish in tests isn't quite cutting it 😅.

This PR introduces a more robust waiting mechanism. Whenever a CVR export operation starts, we write a hidden `.export-in-progress` file to the USB drive. And when the operation completes, we delete that file. Tests can check for the lack of that file.

I've also combined this mechanism with https://github.com/votingworks/vxsuite/pull/3984 to introduce proper recovery from failures midway through a CVR export operation, which previously could've gone uncaught. Now, if an export operation fails midway, and the "something went wrong" screen prompts a poll worker to restart (existing behavior), the scanner will recognize that a resync is required on reboot :tada:. Lots of room for further optimization here (e.g. being smart about what we have to resync, etc.).

## Testing

- [x] Updated automated tests
- [x] Tested recovery from failures midway through an export manually

## Checklist

### DIFF
--- a/apps/scan/backend/test/helpers/custom_helpers.ts
+++ b/apps/scan/backend/test/helpers/custom_helpers.ts
@@ -21,7 +21,7 @@ import {
 import * as grout from '@votingworks/grout';
 import { getImageChannelCount } from '@votingworks/image-utils';
 import { Logger, fakeLogger } from '@votingworks/logging';
-import { SheetInterpretation, SheetOf, mapSheet } from '@votingworks/types';
+import { SheetOf, mapSheet } from '@votingworks/types';
 import { Application } from 'express';
 import { Server } from 'http';
 import { AddressInfo } from 'net';
@@ -228,29 +228,18 @@ export async function scanBallot(
       initialBallotsCounted === 0 || (await apiClient.getConfig()).isTestMode,
   });
 
-  const interpretation: SheetInterpretation = {
-    type: 'ValidSheet',
-  };
-
   mockScanner.scan.mockResolvedValueOnce(ok(await ballotImages.completeBmd()));
   await apiClient.scanBallot();
   mockScanner.getStatus.mockResolvedValue(ok(mocks.MOCK_READY_TO_EJECT));
   await waitForStatus(apiClient, {
     state: 'ready_to_accept',
-    interpretation,
     ballotsCounted: initialBallotsCounted,
     canUnconfigure: true,
-  });
-  await apiClient.acceptBallot();
-  mockScanner.getStatus.mockResolvedValue(ok(mocks.MOCK_NO_PAPER));
-  await waitForStatus(apiClient, {
-    ballotsCounted: initialBallotsCounted + 1,
-    state: 'accepted',
-    interpretation,
-    canUnconfigure: true,
+    interpretation: { type: 'ValidSheet' },
   });
 
-  // Wait for transition back to no paper
+  await apiClient.acceptBallot();
+  mockScanner.getStatus.mockResolvedValue(ok(mocks.MOCK_NO_PAPER));
   await waitForStatus(apiClient, {
     state: 'no_paper',
     ballotsCounted: initialBallotsCounted + 1,
--- a/apps/scan/backend/test/helpers/custom_helpers.ts
+++ b/apps/scan/backend/test/helpers/custom_helpers.ts
@@ -123,6 +123,7 @@ export async function withApp(
     });
     mockUsbDrive.assertComplete();
   } finally {
+    await waitForContinuousExportToUsbDrive(mockUsbDrive);
     await new Promise<void>((resolve, reject) => {
       server.close((error) => (error ? reject(error) : resolve()));
     });
@@ -215,9 +216,9 @@ export function simulateScan(
 
 export async function scanBallot(
   mockScanner: jest.Mocked<CustomScanner>,
+  mockUsbDrive: MockUsbDrive,
   apiClient: grout.Client<Api>,
-  initialBallotsCounted: number,
-  options: { skipWaitForContinuousExportToUsbDrive?: boolean } = {}
+  initialBallotsCounted: number
 ): Promise<void> {
   mockScanner.getStatus.mockResolvedValue(ok(mocks.MOCK_READY_TO_SCAN));
   await waitForStatus(apiClient, {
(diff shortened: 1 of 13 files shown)


apps/scan/backend/test/helpers/custom_helpers.ts
```diff
@@ -123,6 +123,7 @@ export async function withApp(
     });
     mockUsbDrive.assertComplete();
   } finally {
+    await waitForContinuousExportToUsbDrive(mockUsbDrive);
```
arsalansufi, line 126: Because this wait will now take no time if irrelevant, we can use it more liberally. Here, I run it at the end of the `withApp` helper that we use across most VxScan backend tests

apps/scan/backend/test/helpers/custom_helpers.ts
```diff
@@ -256,7 +257,5 @@ export async function scanBallot(
     canUnconfigure: true,
   });
 
-  if (!options.skipWaitForContinuousExportToUsbDrive) {
-    await waitForContinuousExportToUsbDrive();
-  }
+  await waitForContinuousExportToUsbDrive(mockUsbDrive);
```
arsalansufi, line 249: Same idea here as above - we can perform this wait more liberally now

Note that I'm still including this wait here in the `scanBallot` helper, in spite of the wait at the end of `withApp`, because some tests explicitly interact with the exported contents on the USB drive. So a wait is needed right away as opposed to at the end of the test.

apps/scan/backend/test/helpers/custom_helpers.ts
```diff
@@ -228,29 +228,18 @@ export async function scanBallot(
...
   });
+
   await apiClient.acceptBallot();
   mockScanner.getStatus.mockResolvedValue(ok(mocks.MOCK_NO_PAPER));
-  await waitForStatus(apiClient, {
-    ballotsCounted: initialBallotsCounted + 1,
-    state: 'accepted',
-    interpretation,
-    canUnconfigure: true,
-  });
```
arsalansufi, line 250: This is an existing (pre-continuous-export) source of test flakiness. The scanner state machine sometimes transitions so quickly from the `accepted` state to the `no_paper` state in tests that we never detect the `accepted` state. Checking for the `accepted` state is unimportant in the context of this `scanBallot` helper. We can still be sure that the ballot was accepted because of the `ballotsCounted: initialBallotsCounted + 1` in the `no_paper` state.

kshen0: Does this affect the user experience? If we transition too quickly from the `accepted` state is there a screen that doesn't get rendered?

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -755,6 +798,11 @@ export async function doesUsbDriveRequireCastVoteRecordSync(
     }
     const castVoteRecordRootHash = scannerStore.getCastVoteRecordRootHash();
 
+    // A previous export may have failed midway
+    if (areOrWereCastVoteRecordsBeingExportedToUsbDrive(usbDriveStatus)) {
+      return true;
+    }
```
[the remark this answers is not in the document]
kshen0: I'm trying to understand the rules we set for ourselves around the marker file. Can a resync still happen when the marker file is present? Other than actions that come after CVR writing, is there anything that's prevented if the marker file is present?

apps/scan/backend/test/helpers/custom_helpers.ts
arsalansufi: No effect here on actual UX since this is just a test helper. In a number of tests, we intentionally reconfigure state machine delays to trigger certain states and/or to speed tests up, making it possible to quickly pass the accepted state (in a way that we wouldn't outside of tests)

libs/backend/src/cast_vote_records/export.ts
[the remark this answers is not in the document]
arsalansufi: > Can a resync still happen when the marker file is present?

Yup, a resync can still happen! The marker doesn't prevent any CVR operations.

> Other than actions that come after CVR writing, is there anything that's prevented if the marker file is present?

The marker doesn't prevent anything directly. It just factors into the `doesUsbDriveRequireCastVoteRecordSync` function:

https://github.com/votingworks/vxsuite/blob/b0de8444b4462abbf7d16bd434d7cd36c4f204f6/libs/backend/src/cast_vote_records/export.ts#L713-L723

Which is used to determine when to display the CVR (re)sync modal:

modal

`doesUsbDriveRequireCastVoteRecordSync` further only recomputes when 1) the USB drive status changes (e.g. on insertion and removal) or 2) its cache is explicitly cleared (e.g. after cast vote record export).

When the CVR (re)sync modal is up, it only prevents casting of additional ballots. Sys admins, election managers, and poll workers can still insert their smart cards to access their screens.

Hope the clears things up a bit!

apps/scan/backend/test/helpers/custom_helpers.ts
kshen0: Missed the filepath, thanks!

libs/backend/src/cast_vote_records/export.ts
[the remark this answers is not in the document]
kshen0: Got it. I was thinking through whether the machine could deadlock while waiting for the marker to be removed but it sounds like the only place we block on the marker also forces CVR resync, which removes the marker.

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -569,8 +570,31 @@ async function randomlyUpdateCreationTimestamps(
   }
 }
 
+function getCastVoteRecordExportInProgressMarkerFilePath(
+  usbMountPoint: string
+): string {
+  return path.join(usbMountPoint, '.export-in-progress');
```
kshen0, line 576: nit: since we are writing to the filesystem and can't assume what other files are on the usb drive (or can we?), should we prefix this like `.vx-export-in-progress`?

kshen0 (approved):

libs/backend/src/cast_vote_records/export.ts
arsalansufi: We can and can't haha - VxAdmin does provide a USB formatting tool that we encourage election officials to use, but you never know what might end up on a USB. A `vx-` prefix makes sense to me: https://github.com/votingworks/vxsuite/pull/4009/commits/5d77a902e2e83ded4ede18c1f73f295b0297196d

Issue #3922, opened 2023-09-05
Introduce the ability to export all images, including images for ballots that weren't accepted, so that we can deprecate existing backup functionality


arsalansufi:

adghayes:
Does any of our backup tooling rely on the exported `db` file? Because we wouldn't be able to export that in the CVR export due to privacy concerns.

arsalansufi:
@adghayes Good q! In this [Slack thread](https://votingworks.slack.com/archives/CEL6D3GAD/p1691523484960869), we reached the conclusion that the DB file isn't critical for any analysis. And given the privacy concerns that you mention, we have all the more reason to never export it. Another reason why we want to deprecate the existing backup functionality and consolidate on just one kind of export.

Issue #3996, opened 2023-09-22
Miscellaneous continuous export cleanup


arsalansufi:
Will open standalone issues for relevant items when I wrap up my current stretch work on continuous export.

- [ ] ~Remove [this mocking](https://github.com/votingworks/vxsuite/blob/3ba81d4d418df406aa5304fac2dbbc6e9cf159ae/libs/auth/src/artifact_authentication.test.ts#L47-L49)~ https://github.com/votingworks/vxsuite/issues/4100
- [x] Have `BallotImage` field consistently include both a front and back image, when populated
- [x] Revisit [this TODO](https://github.com/votingworks/vxsuite/blob/3ba81d4d418df406aa5304fac2dbbc6e9cf159ae/libs/backend/src/cast_vote_records/build_cast_vote_record.ts#L163-L164)
- [x] Use `*File` vs. `*FileContents` more consistently in variable names
- [x] Fix flaky tests
- [ ] ~Only flush to USB once instead of after every write when performing a full export~ https://github.com/votingworks/vxsuite/issues/4093
- [x] Remove `cvrs_backed_up_at` tracking, after verifying that it's no longer used (check both VxScan and VxCentralScan)
- [x] Make `waitForContinuousExportToUsbDrive` test helper in VxScan less arbitrary
- [x] Recovery from mid-CVR-write failures
- [x] Don't allow polls close if a mismatch is detected
- [x] Don't allow clicks outside progress modals to dismiss them
- [ ] ~Add an indication of whether CVR exports are pre- or post-polls-close~ https://github.com/votingworks/vxsuite/issues/4094
- [ ] ~Determine if `batchLabel` field is still needed and remove if possible~ https://github.com/votingworks/vxsuite/issues/4095
- [ ] ~Have recovery logic export only what needs to be exported instead of everything~ https://github.com/votingworks/vxsuite/issues/4096
- [ ] ~Update code names and variables to no longer refer to CVRs as single files~ https://github.com/votingworks/vxsuite/issues/4097
- [ ] ~Update CVR fixture generator to use actual CVR export functions~ https://github.com/votingworks/vxsuite/issues/4098
- [ ] ~Re-enable CVR auth in CVR tests~ https://github.com/votingworks/vxsuite/issues/4101
- [ ] ~Export uninterpretable rejected ballots~ https://github.com/votingworks/vxsuite/issues/4099

arsalansufi:
Closing this issue out now that I've created standalone issues for the outstanding checklist items

PR #4011, opened 2023-10-02
Deprecate VxScan's existing backup functionality in favor of "extended" CVR exports

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3922

This PR deprecates VxScan's existing backup functionality in favor of "extended" CVR exports that contain images not only for accepted ballots but also rejected* ballots. (This is currently the key delta between CVR exports and backups - backups include rejected ballot images. There are other deltas, like backups' inclusion of the DB file, but we want to drop that for privacy reasons.)

In my next PR, I'll focus on VxCentralScan, where I'll also deprecate existing backup functionality in favor of extended CVR exports, but will also by default have VxCentralScan export only the images that are absolutely necessary for tabulation (images for accepted ballots with write-ins).

Summarizing, when all is said and done:

1. VxScan continuous export will include images for all ballots, accepted and rejected
2. VxScan "Save CVRs" button will include images for all ballots, accepted and rejected (same as 1)
3. VxScan will no longer have a "Save Backup" button
4. VxCentralScan "Save CVRs" button will include images for accepted ballots with write-ins
5. VxCentralScan "Save Backup" button will include images for all ballots, accepted and rejected (same as 1 and 2)

This PR handles 1, 2, and 3. My next PR will handle 4 and 5.

### Caveat

*By "rejected" ballots, I mean interpretable rejected ballots. Tracking and exporting _all_ rejected ballots, including ones that we couldn't interpret at all, will require some extra work. Today's backups also don't include these images, so we're not losing anything with this PR ([Slack thread with some context](https://votingworks.slack.com/archives/CEL6D3GAD/p1695844813451049)).

## Demo Screenshot and Video

_No more "Save Backup" button, just "Save CVRs"_

em-settings

_CVR export contents, including "rejected-" sub-directories_

https://github.com/votingworks/vxsuite/assets/12616928/1a8fe881-83f5-4798-b7eb-3d072b8a0f3c

## Testing

- [x] Updated automated tests
- [x] Tested manually

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ Revisiting logging holistically in a follow-up PR


libs/backend/src/cast_vote_records/build_cast_vote_record.ts
```diff
@@ -569,26 +578,19 @@ export function buildCastVoteRecord({
...
-            : {
-                // empty object to represent a page with no included image
-                '@type': 'CVR.ImageData',
-              }
-        )
+    BallotImage: imageFileUris
+      ? imageFileUris.map((imageFileUri) => ({
+          '@type': 'CVR.ImageData',
+          Location: imageFileUri,
+        }))
```
arsalansufi, line 593: Following up on our discussion in https://github.com/votingworks/vxsuite/pull/3988#discussion_r1333360383 re including either no images or both front and back images, but never just one image

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -526,47 +559,45 @@ export async function updateCreationTimestampOfDirectoryAndChildrenFiles(
...
- * This shuffling is irrelevant for full exports, where we already iterate over cast vote records
- * in an order independent of creation timestamp.
+ * This shuffling is irrelevant for full exports, where we already iterate over ballots in an order
+ * independent of creation timestamp.
  */
 async function randomlyUpdateCreationTimestamps(
   exportContext: ExportContext,
   exportDirectoryPathRelativeToUsbMountPoint: string,
-  options: { castVoteRecordIdToIgnore?: string } = {}
+  options: { subDirectoryNameToIgnore?: string } = {}
```
arsalansufi, line 573: Updated the variable names in this function because we're now shuffling not only cast vote record sub-directories but also sub-directories for rejected sheets, i.e. all export sub-directories

apps/scan/backend/schema.sql
```diff
@@ -9,8 +9,6 @@ create table election (
   ballot_count_when_ballot_bag_last_replaced integer not null default 0,
   is_sound_muted boolean not null default false,
   is_ultrasonic_disabled boolean not null default false,
-  cvrs_backed_up_at datetime,
-  scanner_backed_up_at datetime,
```
arsalansufi, line 13: Not only can we delete logic for the old-style of backup; we can also delete logic for the check that prevents unconfiguring a machine until you've backed. Continuous export makes it such that you're essentially always backed up

apps/scan/backend/src/store.ts
```diff
@@ -96,6 +96,28 @@ function resultSheetRowToResultSheet(row: ResultSheetRow): ResultSheet {
...
+  };
+}
+
+function sheetRowToSheet(row: SheetRow): Sheet {
+  assert(
+    row.requiresAdjudication === 0 ||
+      row.finishedAdjudicationAt !== null ||
+      row.deletedAt !== null,
+    'Every sheet requiring review should have been either accepted or rejected'
+  );
```
arsalansufi, line 115: This condition is guaranteed to hold given transactions in a subsequent commit

apps/scan/backend/test/helpers/shared_helpers.ts
```diff
@@ -103,40 +98,6 @@ export async function configureApp(
...
- *   should already be another test that covers the same interpretation case.
- * - The test doesn't check the CVR export at the end. The interpreter stores
- *   the ballot images which are used in the CVR, and mocking will forgo that
- *   logic.
- * - The test doesn't depend on the actual page interpretations. This function
- *   adds fake page interpretations that don't actually match the passed in
- *   ballot interpretation (because the state machine doesn't actually use those
- *   page interpretations, they are just stored for the CVR).
- */
-export function mockInterpret(
```
arsalansufi, line 120: We can't use this function any more because the second condition in the above list re CVR exports and ballot images is no longer met. We're now always exporting ballot images after scans (aside from the caveat mentioned in the PR description).

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -133,6 +133,21 @@ export interface ResultSheet {
   readonly batchLabel?: string;
```
adghayes, line 133: sidebar: is this still needed?

apps/scan/backend/src/store.ts
```diff
@@ -545,64 +505,6 @@ export class Store {
     return row?.ballotsCounted ?? 0;
   }
 
-  /**
-   * Returns whether the appropriate backups have been completed and it is safe
-   * to unconfigure a machine / reset the election session. Always returns
-   * true in test mode.
-   */
-  getCanUnconfigure(): boolean {
```
adghayes, line 553: OK, so now that we have continuous backup you can always reconfigure because it's always backed up. Yes?

When the USB is out of sync and needs to be updated it shows a warning and you click a button to sync. But IIRC it shows the prompt via the voter screen, and the election manager screens are still accessible normally with authentication. So does this mean that you could have the "out-of-sync" warning, insert an Election Manager card, and unconfigure anyway? I haven't confirmed this is actually possible so may just be missing something.

adghayes (approved):

libs/backend/src/cast_vote_records/export.ts
arsalansufi: I'm not entirely sure - left a note to revisit this in my temporary catch-all issue: https://github.com/votingworks/vxsuite/issues/3996

apps/scan/backend/src/store.ts
arsalansufi: This line of reasoning is all correct - it _is_ possible to access the PW, EM, and SA screens when the sync prompt is up on the voter screen.

I'd been thinking about how worth it it was to prevent unconfiguring while in this fairly rare state. Things are complicated by the fact that the USB likely won't be in the machine come time to unconfigure, since it'll have been taken to VxAdmin for tabulation. And we don't display the sync prompt when there's no USB in the machine at all. Which brings us back to the need for some separate piece of store state if we want to fully track this.

I'm leaning toward disabling the PW polls-close button and the EM delete-election-data button when in the sync-required state, as the simplest possible safeguard.

Yes, an EM could still remove the USB and then delete election data (since the EM screen is still accessible when there's no USB). But a PW couldn't remove the USB and then close polls (since the PW screen is not accessible when there's no USB). So if a machine has had its polls closed, it's guaranteed to be backed up.

I think that this covers our bases well but can double check with Roe, as well!

adghayes:
@arsalansufi one more thought about replacing backups... backups do / did include the log file, now our replacement doesn't include the log file. So that may be lost, and people will probably forget to export it. Not sure if that matters, but may be worth a Slack thread if there hasn't been one already.

arsalansufi:
> one more thought about replacing backups... backups do / did include the log file, now our replacement doesn't include the log file. So that may be lost, and people will probably forget to export it. Not sure if that matters, but may be worth a Slack thread if there hasn't been one already.

Yeah another good point. This has come up in past discussion, and I believe that we decided that it was okay for log export to remain separate. But I'll re-raise this in Slack, too

apps/scan/backend/src/store.ts
arsalansufi: And got confirmation that we're good to go with the edits that I've proposed in this thread ✅

arsalansufi: https://github.com/votingworks/vxsuite/pull/4011/commits/9615ad30db98ac71f5561d4d4e9f20afe1b984e0 + https://github.com/votingworks/vxsuite/pull/4011/commits/4e2090f50673a9f8d7f99dee143e30f5d63999ac

This thread made me realize that there are quite a few operations that we need to block when a CVR sync is detected as necessary:

- Switching from official mode to test mode (assuming ballots have been counted)
- Deleting election data
- Closing polls (including from the polls paused state)

This video demonstrates all of the above:

https://github.com/votingworks/vxsuite/assets/12616928/7af274d9-0248-4784-9595-ee415a7635fd

While I was at it, I also had to tweak the flow for switching from official mode to test mode when ballots have been counted, which used to really explicitly bring the need to backup. I've now just swapped that modal out with a confirmation modal:

https://github.com/votingworks/vxsuite/assets/12616928/139d8bb4-bfc9-4a9e-9bcc-dc125337de4c

PR #4016, opened 2023-10-02
[libs/ui] Shared UI Strings API react-query client

kofi-q:
## Overview

- Adding shared `react-query` frontend client for the shared UI Strings backend API.
  - Only implementing `getAvailableLanguages()` for now and will flesh out over time
- Exporting an instance of the shared API in each frontend for use in components.
  - Also triggering the included `onMachineConfigurationChange` whenever a configure or unconfigure mutation happens.

## Testing Plan
- Added libs/ui unit tests for the shared API logic.
- Added app frontend unit tests to verify `onMachineConfigurationChange` is being triggered appropriately.

## Checklist

### DIFF
--- a/libs/ui/src/hooks/ui_strings_api.ts
+++ b/libs/ui/src/hooks/ui_strings_api.ts
@@ -1,0 +1,46 @@
+import { QueryClient, QueryKey, useQuery } from '@tanstack/react-query';
+import { UiStringsApi } from '@votingworks/types';
+import * as grout from '@votingworks/grout';
+
+// Unable to use `grout.Client<UseStringsApi>` directly due to some mismatched
+// type inference with `grout.AnyMethods`, so copying the `grout.Client`
+// definition here for now:
+export type UiStringsApiClient = {
+  [Method in keyof UiStringsApi]: grout.AsyncRpcMethod<UiStringsApi[Method]>;
+};
+
+function createReactQueryApi(getApiClient: () => UiStringsApiClient) {
+  return {
+    getAvailableLanguages: {
+      getQueryKey(): QueryKey {
+        return ['getAvailableLanguages'];
+      },
+
+      useQuery() {
+        const apiClient = getApiClient();
+
+        return useQuery(this.getQueryKey(), () =>
+          apiClient.getAvailableLanguages()
+        );
+      },
+    },
+
+    async onMachineConfigurationChange(
+      queryClient: QueryClient
+    ): Promise<void> {
+      await queryClient.invalidateQueries(
+        this.getAvailableLanguages.getQueryKey()
+      );
+    },
+
+    // TODO(kofi): Fill out.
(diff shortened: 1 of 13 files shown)


libs/ui/src/hooks/ui_strings_api.ts
```diff
@@ -0,0 +1,46 @@
+import { QueryClient, QueryKey, useQuery } from '@tanstack/react-query';
+import { UiStringsApi } from '@votingworks/types';
+import * as grout from '@votingworks/grout';
+
+// Unable to use `grout.Client<UseStringsApi>` directly due to some mismatched
+// type inference with `grout.AnyMethods`, so copying the `grout.Client`
+// definition here for now:
```
arsalansufi, line 7: Seems like a reasonable workaround!

I think `grout.Client<UseStringsApi>` in the comment is maybe supposed to read `grout.Client<UiStringsApi>`

libs/ui/src/hooks/ui_strings_api.ts
```diff
@@ -0,0 +1,46 @@
...
+        const apiClient = getApiClient();
+
+        return useQuery(this.getQueryKey(), () =>
+          apiClient.getAvailableLanguages()
+        );
+      },
+    },
+
+    async onMachineConfigurationChange(
+      queryClient: QueryClient
```
arsalansufi, line 29: For consistency, would it make sense to pass the `queryClient` (or `getQueryClient`) as a top-level input to `createReactQueryApi` / `createUiStringsApi`?

```ts
createReactQueryApi(
  getApiClient: () => UiStringsApiClient,
  getQueryClient: () => QueryClient
)
```

So that all methods (`getAvailableLanguages`, `onMachineConfigurationChange`) always have access to both without them needing to be passed in.

I'm okay with the current approach, too - maybe there's a reason to differentiate the `apiClient` from the `queryClient` that I'm missing

arsalansufi (approved):
LGTM!

libs/ui/src/hooks/ui_strings_api.ts
kofi-q: oops - thanks!

libs/ui/src/hooks/ui_strings_api.ts
kofi-q: Ah yeah, +1 to consistency!

kofi-q: Bah, looks like we're not allowed to invoke `useQueryClient` inside of the `onSuccess` handler of mutations (since it's outside of the component context, so that forces us to pass in the query client to the change handler at execution time. Will drop the latest commit for now

arsalansufi: Ohh I see, makes sense - leaving as is is fine then!

PR #4017, opened 2023-10-02
[i18n/Audio] Clear UI Strings store on unconfigure

kofi-q:
## Overview

Updating the "unconfigure machine" API endpoints in `Mark` and `Mark-Scan` to `reset()` the store completely, matching what we're doing in `Scan`, instead of explicitly deleting data from each table.

We would need to revert back to explicitly clearing individual tables if we ever add a table that needs to persist across multiple configure/unconfigure events, but that seems unlikely (at least, in the short/medium term).

__TODO for later:__ noticed the "unconfigure" endpoints are unguarded WRT user type, so might be worth adding an  "is EM or SysAdmin" check, at least for consistency with the "configure" endpoints.

## Testing Plan
- Added a shared test runner for the `UiStringsStore` to verify all UI String data is cleared when unconfiguring machines.

## Checklist

### DIFF
--- a/apps/mark-scan/backend/src/app.ts
+++ b/apps/mark-scan/backend/src/app.ts
@@ -111,8 +111,7 @@ export function buildApi(
     },
 
     unconfigureMachine() {
-      workspace.store.setElectionAndJurisdiction(undefined);
-      workspace.store.deleteSystemSettings();
+      workspace.store.reset();
     },
 
     configureWithSampleBallotPackageForIntegrationTest(): Result<
--- a/apps/mark-scan/backend/src/app.ui_strings.test.ts
+++ b/apps/mark-scan/backend/src/app.ui_strings.test.ts
@@ -6,6 +6,7 @@ import {
   createMockUsb,
   runUiStringApiTests,
   runUiStringMachineConfigurationTests,
+  runUiStringMachineDeconfigurationTests,
 } from '@votingworks/backend';
 import { fakeLogger } from '@votingworks/logging';
 import { buildMockInsertedSmartCardAuth } from '@votingworks/auth';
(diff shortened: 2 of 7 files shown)


apps/mark-scan/backend/src/app.ts
```diff
@@ -111,8 +111,7 @@ export function buildApi(
     },
 
     unconfigureMachine() {
-      workspace.store.setElectionAndJurisdiction(undefined);
-      workspace.store.deleteSystemSettings();
+      workspace.store.reset();
```
arsalansufi, line 114: :100:

apps/mark-scan/backend/src/app.ts
```diff
@@ -111,8 +111,7 @@ export function buildApi(
...
     },
```
arsalansufi, line 115: > TODO for later: noticed the "unconfigure" endpoints are unguarded WRT user type, so might be worth adding an "is EM or SysAdmin" check, at least for consistency with the "configure" endpoints.

Makes sense! I've been wanting to add some kind of auth decorators, so that we can easily add access checks across app endpoints: https://github.com/votingworks/vxsuite/issues/3679

eventualbuddha:
This makes me wish for a bigger hammer for unconfigure. Essentially a "factory reset". It'd be great if we had a Docker or macOS-like file system layering feature that would allow us to simply remove the top layer and reboot the machine to go back to a clean state. cc @benadida

apps/mark-scan/backend/src/app.ts
kofi-q: Ahh, nice - having some easy/consistent way of annotating endpoints would be great!

kofi-q:
> This makes me wish for a bigger hammer for unconfigure. Essentially a "factory reset". 

Big +1!

PR #4022, opened 2023-10-03
Update primary font to Roboto across all apps

kofi-q:
## Overview

_Easier by commit - last 2 commits are just snapshot updates and dead code deletion._

Slack context [here](https://votingworks.slack.com/archives/C026YV9LPGB/p1693223190383199).

Replacing `Helvetica Neue` with `Roboto` as the primary font in apps, for consistency across the website, marketing materials and apps.

Fonts downloaded from [Google Fonts](https://fonts.google.com/specimen/Roboto) and converted to Base64 encoding with the [Font Squirrel](https://www.fontsquirrel.com) tool, to allow us to embed the fonts in a single shared source file.

- Loading the fonts in the `<AppBase>` component used across our apps, but disabling for unit tests, since it seemed to slow them down quite a bit.
- Deleting deprecated font files and references from frontend HTML files.
- No change yet to letter size calculations (for VVSG compatibility), since the body to total height proportions of `Roboto` are fairly similar to `Helvetica Neue`.
- Starting trying to remove `Noto Emoji` fonts, but we seem to be using them in a few places - we can replace those with `Icon`s later and then deprecate `Noto Emoji`.

## Demo Video or Screenshot

| 🤖 Roboto | Helvetica Neue |
| Screenshot 2023-10-03 at 14 48 13 | Screenshot 2023-10-02 at 17 15 44 |

## Testing Plan
- Added integration tests via the appa to verify that the new font is loaded correctly by `<AppBase>`

## Checklist


kofi-q:
Leaving Drew as primary reviewer, but probably also worth a sanity check from @jonahkagan

mattroe:
I don't see Roboto de-emphasized in the preview

kofi-q:
> I don't see Roboto de-emphasized in the preview

Oh, good catch! Will take a look and fix that font weight - thanks

libs/ui/src/app_base.tsx
```diff
@@ -50,6 +53,13 @@ export function AppBase(props: AppBaseProps): JSX.Element {
   const [colorMode, setColorMode] = React.useState<ColorMode>(defaultColorMode);
   const [sizeMode, setSizeMode] = React.useState<SizeMode>(defaultSizeMode);
 
+  React.useEffect(() => {
+    /* istanbul ignore next - tested via integration tests. */
+    if (!disableFontsForTests) {
+      loadFonts();
+    }
+  }, [disableFontsForTests]);
```
jonahkagan, line 61: Does this mean there will be an initial render without fonts loaded? Does that cause any visual weirdness?

jonahkagan (approved):

kofi-q:
> I don't see Roboto de-emphasized in the preview

Ok, turned out to be an old screenshot from when I had a bug in one of the font definitions - fixed the screenshot now

libs/ui/src/app_base.tsx
kofi-q: Yeah, good question - I didn't see any noticeable flicker on initial render when I was testing, but I'm going to try it out with caching disabled to see what kind of impact it might have

kofi-q: Couldn't spot it while watching the refresh, but managed to catch a frame on a screen recording where the text is in sans-serif just before switching to Roboto. Just to be safe, I'm going to delay the rendering until `loadFonts` has run

libs/ui/src/app_base.tsx
```diff
@@ -49,12 +52,29 @@ export function AppBase(props: AppBaseProps): JSX.Element {
 
   const [colorMode, setColorMode] = React.useState<ColorMode>(defaultColorMode);
   const [sizeMode, setSizeMode] = React.useState<SizeMode>(defaultSizeMode);
+  const [fontsLoaded, setFontsLoaded] = React.useState(false);
+
+  React.useEffect(() => {
+    /* istanbul ignore next - tested via integration tests. */
+    if (!disableFontsForTests) {
+      loadFonts();
```
eventualbuddha, line 60: React runs `useEffect` multiple times in development, so we end up with multiple `<style>` elements being added. While this may not cause any visual issues, it's maybe not great for resource usage and keeping things close to production if we can.

image

libs/ui/src/app_base.tsx
jonahkagan: Nice, thanks for looking into that

libs/ui/src/app_base.tsx
kofi-q: Oh, just seeing this! Thanks for the callout - didn't know about the multiple `useEffect` invocations in dev. Will add a cleanup then, to address the dev weirdness

PR #4029, opened 2023-10-04
feat(admin): Add "No Results" Warning to Report Builders

adghayes:
## Overview

Currently, if you create a combination of filters and group by clauses in a report builder such that there are no results at all, the report viewing area will just be blank. Although this should be rare, I wanted to add in some sort of warning text in this situation so a user does not think the app is broken. Examples where this might show up:

1) User tries to "group by batch" but no CVRs are loaded. There are no batches, so nothing to group by.
2) User creates an impossible filter (e.g. filtering for ballots associated with the democratic party but with republican ballot styles) and applies a grouping

This does not apply in some cases. For example, imagine a case (3) in which you set up an impossible filter as in example (2) but _do not_ have a grouping - you'll just get a zero or blank report. 

There may be other ways to resolve these. For example, we could analyze the filter for an impossible filter (probably hitting the backend would be easiest) and give a more specific warning to the filter. This would better catch cases (2) and (3). And for (1), maybe we disable those groupings until CVRs are added.

I wanted to do the least lift solution here because these cases are all rare, and I just don't want the screen to be confusingly empty in these rare cases.

The change prevents printing and exporting reports once an empty report is previewed, but if you print or export the report before the previewing, it will still print an empty page, export an empty PDF, or export an empty CSV. It's a lot more complicated to do otherwise, so I left this corner case alone.

## Demo Video or Screenshot

Screen Shot 2023-10-04 at 4 56 17 PM

## Testing Plan
added automated tests, checked manually

## Checklist

### DIFF
--- a/apps/admin/frontend/src/components/reporting/ballot_count_report_viewer.test.tsx
+++ b/apps/admin/frontend/src/components/reporting/ballot_count_report_viewer.test.tsx
@@ -107,6 +107,40 @@ test('autoPreview = false does not load preview automatically', async () => {
   await screen.findButton('Load Preview');
 });
 
+test('shows no results warning when no results', async () => {
+  const { electionDefinition } = electionTwoPartyPrimaryFixtures;
+  apiMock.expectGetCardCounts(
+    {
+      filter: {},
+      groupBy: { groupByBatch: true },
+    },
+    []
+  );
+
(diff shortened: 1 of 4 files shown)


kofi-q (approved):
LGTM!

Issue #3995, opened 2023-09-22
Add feature flag for not including redundant metadata in every individual cast-vote-record-report.json


arsalansufi:
I'm suggesting a feature flag because this change would make the individual `cast-vote-record-report.json` files not true CDF CVR reports. With a feature flag, we can be either CDF compliant or space efficient.

PR #4036, opened 2023-10-04
Add feature flag for excluding redundant metadata in individual CVR reports

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3995

In order to meet the VVSG2 requirements around continuous export, while also staying true to CDF, our export format has ended up a little strange.

We export a top-level `metadata.json` file for the whole export, plus a `cast-vote-record-report.json` file for every individual CVR. But we also repeat the metadata in the `metadata.json` file in every `cast-vote-record-report.json` file, so that the latter remain CDF-compliant. This PR introduces a feature flag to exclude that redundant metadata in `cast-vote-record-report.json` files, significantly decreasing their size. This feature flag is similar in spirit to the existing feature flag for excluding original snapshots in CVRs.

With a feature flag, we can toggle between efficiency and true CDF-compliance, as needed. Hopefully, something will come out of https://github.com/usnistgov/CastVoteRecords/issues/45 such that this feature flag becomes unnecessary 🤞.

I'm enabling both CVR optimization feature flags by default, since we'll want them enabled for real elections for the foreseeable future as far as I can tell. Faster performance on election day feels more important than hypothetical cross-system compatibility, for now.

## Demo Screenshot

An individual `cast-vote-record-report.json` file, before on the left, after on the right. All the extra data that's on the left but not on the right (i.e. the redundant metadata) is never used on import.

before-after

## Testing

- [x] Manually tested export and import with and without the feature flag

Automated tests are coming in their own PR

## Checklist


.env
```diff
@@ -11,8 +11,8 @@ REACT_APP_VX_SKIP_CVR_ELECTION_HASH_CHECK=FALSE
 REACT_APP_VX_SKIP_SCAN_ELECTION_HASH_CHECK=FALSE
 REACT_APP_VX_SKIP_BALLOT_PACKAGE_AUTHENTICATION=FALSE
 REACT_APP_VX_SKIP_CAST_VOTE_RECORDS_AUTHENTICATION=FALSE
-REACT_APP_VX_DISABLE_CVR_ORIGINAL_SNAPSHOTS=FALSE
+REACT_APP_VX_CAST_VOTE_RECORD_OPTIMIZATION_EXCLUDE_ORIGINAL_SNAPSHOTS=TRUE
+REACT_APP_VX_CAST_VOTE_RECORD_OPTIMIZATION_EXCLUDE_REDUNDANT_METADATA=TRUE
```
arsalansufi, line 15: As noted in the PR description:

> I'm enabling both CVR optimization feature flags by default, since we'll want them enabled for real elections for the foreseeable future as far as I can tell. Faster performance on election day feels more important than hypothetical cross-system compatibility, for now

adghayes (commented):
lgtm, thanks for the consistency changes around env. vars

adghayes (approved):

PR #4037, opened 2023-10-04
[i18n/audio] Pipe data through to getUiStrings react-query client

kofi-q:
## Overview

Wiring up a `getUiStrings` react-query client API to the backend store for future use with our string translation library.

Also updating the nullable return types in the API from `Optional<T>` to `T | null`, since react-query doesn't like `undefined` query data values.

## Testing Plan
- Updated backend and frontend API tests

## Checklist


arsalansufi (approved):
LGTM!

PR #4039, opened 2023-10-05
Integration Test w/ USB Drives

### LINKED ISSUE #3950: expand `MockFileUsbDrive` and re-enable VxAdmin integration tests
Possible after #3948 lands.

3 of 4 VxAdmin integration tests were disabled because we couldn't mock USB drives. Now we can! We need to extend the stubbed mocking capabilities and re-enable those integration tests, which as also probably wildly out of date.

adghayes:
## Overview

Closes #3950. Fleshes out the `MockFileUsbDrive` implementation of `UsbDrive`, modeled after @arsalansufi's `MockFileCard` implementation. Applies it in an integration test for reporting in VxAdmin, which loads CVRs from file and exports a CSV file, checking the exports contents.

The main challenge I encountered was CVR authentication. The import failed with OpenSSL error "Verification Failed." I'm not sure if the instance would need to be set up with certain certificates - @arsalansufi would that be an easy fix? I didn't know how to do that, so I extended our `allowInProduction` carveouts to allow using env. variables in integration tests, and loading them on test-startup.


libs/usb-drive/jest.config.js
```diff
@@ -5,5 +5,5 @@ const shared = require('../../jest.config.shared');
  */
 module.exports = {
   ...shared,
-  coveragePathIgnorePatterns: ['src/cli.ts'],
+  coveragePathIgnorePatterns: ['src/cli.ts', 'src/mocks'],
```
adghayes, line 8: The `memory_usb_drive` file was already ignoring coverage, but I'd rather have it here than in a file comment.

libs/usb-drive/src/mocks/memory_usb_drive.ts
adghayes, line 1: Content removed here moved to shared file `helpers.ts`.

apps/admin/integration-testing/e2e/support/auth.ts
```diff
@@ -50,10 +50,11 @@ export function mockElectionManagerCardInsertion({
  * Enters the PIN into the PIN pad.
  */
 export async function enterPin(page: Page): Promise<void> {
-  await page.waitForSelector('text=Enter the card PIN to unlock.');
+  await page.getByText('Enter the card PIN to unlock.').waitFor();
```
adghayes, line 53: To be more consistent with our RTL testing, I want to move away from the `text=something` style selectors. Instead, get the "locater" and make assertions against it.

In this case, Playwright docs recommend against using `waitForSelector` and instead recommend using `waitFor` with a locator.

apps/admin/integration-testing/playwright.config.ts
```diff
@@ -22,6 +25,7 @@ export default defineConfig({
     ['html', { open: 'never' }],
     ['junit', { outputFile: join(outputDir, 'results.xml') }],
   ],
+  workers: 1,
```
adghayes, line 28: Test suites still run in parallel with current settings, we have to only use one worker to give each test exclusive access to app state.

We could instead spin up different app instances for different tests, but that seems like trouble (and resource-intensive).

apps/admin/integration-testing/.env
```diff
@@ -0,0 +1 @@
+REACT_APP_VX_SKIP_CAST_VOTE_RECORDS_AUTHENTICATION=TRUE
```
arsalansufi, line 1: > The main challenge I encountered was CVR authentication. The import failed with OpenSSL error "Verification Failed." I'm not sure if the instance would need to be set up with certain certificates - @arsalansufi would that be an easy fix?

Hmm interesting - integration tests _should_ be using our dev certs, i.e. the certs committed to [libs/auth/certs/dev](https://github.com/votingworks/vxsuite/tree/main/libs/auth/certs/dev) and pointed to via these functions:

https://github.com/votingworks/vxsuite/blob/c36e82ba5d531624b53e7bb781c276313c9a3391/libs/auth/src/config.ts#L24-L28

https://github.com/votingworks/vxsuite/blob/c36e82ba5d531624b53e7bb781c276313c9a3391/libs/auth/src/config.ts#L30-L63

And I've signed our CVR fixtures using those same dev keys and certs:

https://github.com/votingworks/vxsuite/blob/c36e82ba5d531624b53e7bb781c276313c9a3391/libs/cvr-fixture-generator/src/cli/generate/main.ts#L281-L282

Seems like something just isn't matching up. Skipping CVR auth is a totally fine workaround for now, and if you want to open a follow-up issue, I'll take a look when I get the chance!

jonahkagan (approved):
Very exciting 🎉

libs/usb-drive/src/mocks/file_usb_drive.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import {
+  MOCK_USB_FILE_PATH,
+  MockFileUsbDrive,
+  getMockFileUsbDriveHandler,
+  initializeMockFile,
+} from './file_usb_drive';
+
+beforeEach(() => {
+  // clear mock file from other tests or development
+  initializeMockFile();
```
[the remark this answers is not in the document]
adghayes: Yeah, there are a lot of disadvantages to the fact that, in this first pass, all file mocks in all use cases use the same temporary file. It would be better to introduce the idea of "different instances." And it's not currently set up to replace our dev USB drive mocks, although we will get there.

The single file approach solves the problem of multiple processes (e.g. app and integration test runner) being able to access / edit the same data. This problem is a little different in the integration test case, the normal test case, and development. 

After doing this PR, I think that the integration test case could be solved by environment variables? As in, the test suite generates a UUID on boot up and places it in an environment variable, which can then be used to initialize the `usbDrive` from the app, so they can access the same temporary file. The other two cases - regular tests and dev - are easier. In tests we could mock the environment variable and in dev (`NODE_ENV=development`) we could actually have a `git` ignored file and mock USB root directory under `libs/usb-drive/dev-workspace`. Then we could kill `libs/usb-mocking`. What do you think of these ideas?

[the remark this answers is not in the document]
jonahkagan: > there are a lot of disadvantages to the fact that, in this first pass, all file mocks in all use cases use the same temporary file

I think we should get clear on what the disadvantages are before we try to solve this. I'm thinking about the auth mocks, which I believe are backed by a single file instance, and it seems like that hasn't caused too many issues so far.

I do like the idea of killing `libs/usb-mocking`, but I would imagine that's orthogonal to this discussion, since we only need a single mock instance for that.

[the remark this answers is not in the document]
adghayes: I agree. I started to solve this in the most general way and found it to be more trouble than it's worth.

In a more limited way, however, it's necessary - distinguishing the dev USB from the test USB. We're more concerned about data persistence in this case than in the auth case. I wouldn't want my integration tests clearing a dev mock USB drive I was separately using. This will be addressed in another PR not up for review yet.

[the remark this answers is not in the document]
jonahkagan: That makes sense!

PR #4052, opened 2023-10-07
`libs/usb-drive`: VxCentralScan

adghayes:
Converts VxCentralScan over to the new `libs/usb-drive` framework. 

Review by commit.


apps/central-scan/frontend/src/app.test.tsx
```diff
@@ -27,20 +25,27 @@ import {
...
+} from '../test/api';
+import { mockUsbDriveStatus } from '../test/helpers/mock_usb_drive';
 
 let mockApiClient: MockApiClient;
 
 beforeEach(() => {
   jest.restoreAllMocks();
 
-  window.kiosk = undefined;
-
```
adghayes, line 38: AFAICT, `central-scan`'s only use of `window.kiosk` are now the reboot methods and setting the date and time. These could easily be moved to `libs/backend`, and `central-scan` could be an initial test case for a `kiosk-browser` alternative.

adghayes: It's the only app that does not print.

apps/admin/backend/src/util/export_file.ts
```diff
@@ -16,9 +16,17 @@ export function exportFile({
...
+    /* c8 ignore start */
+    usbDrive: {
+      status: () =>
+        Promise.resolve({
+          status: 'no_drive',
+        }),
+      eject: () => Promise.resolve(),
+      format: () => Promise.resolve(),
+    },
+    /* c8 ignore stop */
```
adghayes: I think that this workaround is evidence that `Exporter` would be better off as a couple of functions rather than a class with state and multiple methods. I didn't want to get involved in that refactor here, though.

apps/admin/backend/src/util/export_file.ts
```diff
@@ -16,9 +16,17 @@ export function exportFile({
...
```
adghayes, line 29: I think that this workaround is evidence that Exporter would be better off as a couple of functions rather than a class with state and multiple methods. I didn't want to get involved in that refactor here, though.

apps/central-scan/backend/src/index.ts
```diff
@@ -2,25 +2,27 @@ import { Logger, LogSource, LogEventId } from '@votingworks/logging';
...
   `${dotenvPath}.${NODE_ENV}.local`,
   // Don't include `.env.local` for `test` environment
   // since normally you expect tests to produce the same
   // results for everyone
-  NODE_ENV !== 'test' ? `${dotenvPath}.local` : '',
+  !isTestEnvironment ? `${dotenvPath}.local` : '',
   `${dotenvPath}.${NODE_ENV}`,
   dotenvPath,
-  NODE_ENV !== 'test' ? `../../../${dotenvPath}.local` : '',
+  !isTestEnvironment ? `../../../${dotenvPath}.local` : '',
```
adghayes, line 25: Prevents `.env.local` files polluting integration tests run locally.

apps/central-scan/backend/src/index.ts
```diff
@@ -35,7 +37,7 @@ for (const dotenvFile of dotenvFiles) {
   }
 }
 
-const logger = new Logger(LogSource.VxScanService);
+const logger = new Logger(LogSource.VxCentralScanService);
```
adghayes, line 40: Noticed this bug.

apps/central-scan/integration-testing/.env
```diff
@@ -0,0 +1,2 @@
+REACT_APP_VX_SKIP_SCAN_ELECTION_HASH_CHECK=TRUE
```
adghayes, line 1: Allows test to pass without updating images if definition changes only slightly.

apps/central-scan/integration-testing/.env
```diff
@@ -0,0 +1,2 @@
...
+REACT_APP_VX_SKIP_BALLOT_PACKAGE_AUTHENTICATION=TRUE
```
adghayes, line 2: We're generating the ballot package on the fly (which is better for maintainability than having a fixture we need to regenerate) so we don't have a signature.

adghayes:
Flagging @jonahkagan for review (assuming you're reviewer for `libs/usb-drive` things unless you say otherwise), and flagging @arsalansufi just so you know.

apps/central-scan/frontend/src/app.test.tsx
jonahkagan: That's awesome!

apps/central-scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
```diff
@@ -3,26 +3,34 @@ import {
...
+  // USB drive status is guaranteed to exist because app root will not render
+  // this component until the USB drive query succeeds.
+  assert(usbDriveStatusQuery.isSuccess);
+
   const logOutMutation = logOut.useMutation();
   const configureMutation = configureFromBallotPackageOnUsbDrive.useMutation();
 
-  useExternalStateChangeListener(usbDriveStatus, (newUsbDriveStatus) => {
-    if (newUsbDriveStatus === 'mounted') {
+  useQueryChangeListener(usbDriveStatusQuery, (newUsbDriveStatus) => {
```
jonahkagan, line 32: Can this move to the backend now? Or do you want to save that for later?

apps/admin/backend/src/util/export_file.ts
jonahkagan: Agreed. For now, it might be worth making this "mock" usb drive throw an assertion error if any of its methods are called rather than fail silently

libs/backend/src/list_directory.ts
```diff
@@ -142,20 +145,22 @@ export type ListDirectoryOnUsbDriveError =
  * drive's filesystem. Looks at only the first found USB drive.
  */
 export async function listDirectoryOnUsbDrive(
-  relativePath: string,
-  getUsbDrives: () => Promise<UsbDrive[]> = defaultGetUsbDrives
+  usbDrive: UsbDrive,
+  relativePath: string
 ): Promise<Result<FileSystemEntry[], ListDirectoryOnUsbDriveError>> {
   // We currently do not support multiple USB drives
```
jonahkagan, line 151: This comment seems unnecessary now

apps/admin/backend/src/cast_vote_records.ts
```diff
@@ -47,7 +47,6 @@ import {
   CvrFileMode,
   ImportCastVoteRecordsError,
 } from './types';
-import { buildGetUsbDrivesFn } from './util/exporter';
```
jonahkagan, line 50: Can `buildGetUsbDrivesFn` be removed completely now?

jonahkagan (approved):

apps/admin/backend/src/cast_vote_records.ts
adghayes: I'm not sure if it can be removed in this PR, but it is removed in one of the PRs in this series.

apps/central-scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
adghayes: It didn't occur to me to move it to the backend. Are you suggesting that, instead of triggering the configuration from the frontend, it's triggered directly from the backend? Or are you suggesting something else?

If we were to do that, then we'd need a way to get the configuration status / error to the frontend. I am thinking we might augment the the `getUsbDriveStatus` API endpoint to have extra data about configuration status / attempt(s).

jonahkagan: Yes, that's what I'm suggesting. Why should the frontend be involved at all? The user doesn't need to interact with the screen during configuration.

I think it would make the most sense to put error/in-progress status in endpoint gets the current configuration (currently, `getElectionDefinition` is the closest thing)

arsalansufi: > I am thinking we might augment the the getUsbDriveStatus API endpoint to have extra data about configuration status / attempt(s).

I did something like this while working on the CVR resync modal!

https://github.com/votingworks/vxsuite/blob/021f5ea2247967b5d5ee3c2994537b0643d46039/apps/scan/backend/src/app.ts#L106-L118

PR #4054, opened 2023-10-08
`libs/usb-drive`: Cleanup

adghayes:
With all apps converted to the `libs/usb-drive` framework, this cleans up the old framework:
- removes `kiosk-browser` endpoints
- makes `libs/ui` use new types
- applies changes to frontends
- removes `libs/backend` usb utils


libs/ui/src/export_logs_modal.test.tsx
```diff
@@ -47,32 +47,6 @@ const electionManagerAuthStatus: DippedSmartCardAuth.ElectionManagerLoggedIn = {
 
 jest.useFakeTimers();
 
-test('renders loading screen when usb drive is mounting or ejecting in export modal', async () => {
```
adghayes, line 50: No longer have to test for this states because we no longer support them.

libs/ui/src/usbcontroller_button.tsx
```diff
@@ -31,7 +32,8 @@ export function UsbControllerButton({
 }: Props): JSX.Element | null {
   const variant: ButtonVariant = primary ? 'primary' : 'regular';
 
-  if (usbDriveStatus === 'mounted') {
+  const { status } = usbDriveStatus;
```
jonahkagan, line 35: We might want this button to disable immediately after pressing it to avoid double presses (and maybe show a spinner or something while the action is in progress). When I removed the "ejecting" state, I imagined we would do that by waiting on the eject mutation to succeed.

adghayes: Good point, will look into reviving the "ejecting" state via passing in some query state.

adghayes: I revived the "Ejecting..." text by using the `isLoading` flag on the mutation, which I think prevents the obvious issue of double ejecting.

jonahkagan: Will `usbDriveStatus` still be `'mounted'` while `isLoading` is true? If so, I don't think we'll actually hit the disabled case. Did you test this? Or am I just reading the code wrong?

adghayes: Oops 🤦 . Not reading the code wrong. I did test it briefly, but was focused on making sure the mutation status was resetting properly and didn't actually confirm I saw the "Ejecting..." status.

libs/usb-drive/src/usb_drive.ts
```diff
@@ -319,8 +319,8 @@ export function detectUsbDrive(logger: Logger): UsbDrive {
       }
       await logEjectInit(logger, loggingUserRole);
       try {
-        await unmountUsbDrive(deviceInfo.mountpoint);
         didEject = true;
+        await unmountUsbDrive(deviceInfo.mountpoint);
```
adghayes, line 323: @jonahkagan while testing the previous commit, I actually hit another issue I've never hit before. In a dev environment with a real USB drive, my USB drive would sometimes remount immediately after ejecting. In other words, it didn't seem like it was respecting the `ejected` state. Looking at the logging, it looked like the mount was getting triggered via `status` before `umountUsbDrive` exited. By moving the `didEject=true` status a bit earlier, I'm not longer able to reproduce. 

It means it's possible that `didEject` is set to `true` and the unmount actually fails. But if we're in a situation where the unmount fails, we probably have other problems, and there's not much we can do about. And removing the drive will reset the status.

jonahkagan: Sounds like there's a race condition between mounting and unmounting.

I think optimistically setting `didEject = true` is one way to handle it, but you're right it's a bit messy.

A safter approach might be to have a lock that the mounting/unmounting actions share to avoid contention. This could be as simple as an `isUnmounting` flag that we check before trying to mount.

jonahkagan: Now that I'm looking into this more closely, we have a few different "busy" states that we want to lock on (mounting, formatting, unmounting). It might be worth thinking about how to easily prevent actions if the drive is busy in any way. E.g. we could have a single `isBusy` flag. Or have multiple flags an a helper that checks if any of them is set.

jonahkagan (approved):

adghayes:
@jonahkagan thank you for the review on this PR and all the other `libs/usb-drive` PRs, it's a lot better for your review!

PR #4055, opened 2023-10-09
`libs/usb-drive`: mocking in development

adghayes:
## Overview

Moves the USB mocking (via the dev doc) to use `libs/usb-drive` instead of `libs/usb-mocking`. Now, the mocked USB drive uses `/libs/usb-drive/dev-workspace/mock-usb-data/` as a root. Previously the mock USB drive was in an `.img` file and fully mounted as a loop device.

Folks will have to enable `REACT_APP_VX_USE_MOCK_USB` to begin using the mocking capability. It's faster than before, is space-limited only by the VM rather than the mock `.img`, and the files are more easily inspectable / manipulable. The main downside is that the mock USB can no longer be used interchangeably with real USB drives - real USB drives do not work when `REACT_APP_VX_USE_MOCK_USB=true`. This is how auth mocking works, and separating real from mock may be cleaner anyway.

## Demo Video or Screenshot

I just copied the styling for the disabled smard card mocks to represent the disabled USB mocks. It doesn't fit the name of the env variable, but we could alternatively add a bar below all the controls to provide information about both environment variables.

Screen Shot 2023-10-09 at 6 59 29 PM

## Checklist


libs/dev-dock/frontend/src/dev_dock.tsx
```diff
@@ -336,17 +364,29 @@ function UsbDriveMockControls() {
...
         onClick={onUsbDriveClick}
         isInserted={isInserted}
+        disabled={disabled}
         aria-label="USB Drive"
       >
-        {status && <UsbDriveIcon isInserted={isInserted} />}
+        <UsbDriveIcon isInserted={isInserted} disabled={disabled} />
+        {!isFeatureEnabled && (
+          <UsbMocksDisabledMessage>
+            <p>USB mock disabled</p>
```
jonahkagan, line 385: If we wanted to add the info about the env var, I think an info icon with a tooltip would be a good solution. And we could use the same pattern for the smart card mocks disabled message.

Alternatively we could have an "Enable" button that pops up an alert with instructions (and eventually make an actual API for enabling/disabling mocks via the dev dock).

jonahkagan (approved):

libs/usb-drive/src/mocks/file_usb_drive.ts
```diff
@@ -1,39 +1,61 @@
...
+
+export const MOCK_USB_STATE_FILENAME = 'mock-usb-state.json';
+export const MOCK_USB_DATA_DIRNAME = 'mock-usb-data';
+export const DEFAULT_MOCK_USB_DIR = '/tmp/mock-usb';
+export const DEV_MOCK_USB_DIR = join(__dirname, '../../dev-workspace');
+export const DEV_MOCK_USB_GLOB_PATTERN = join(DEV_MOCK_USB_DIR, '**/*');
+
+function getMockUsbDirPath(): string {
+  if (process.env.NODE_ENV === 'development') {
+    return join(__dirname, '../../dev-workspace');
```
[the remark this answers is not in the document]
adghayes: Yes, thank you!

.env
```diff
@@ -7,6 +7,7 @@ REACT_APP_VX_ENABLE_REACT_QUERY_DEVTOOLS=FALSE
 REACT_APP_VX_ENABLE_DEV_DOCK=TRUE
 REACT_APP_VX_SKIP_PIN_ENTRY=FALSE
 REACT_APP_VX_USE_MOCK_CARDS=FALSE
+REACT_APP_VX_USE_MOCK_USB=FALSE
```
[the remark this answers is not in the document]
adghayes: good point, switched!

libs/dev-dock/frontend/src/dev_dock.tsx
adghayes: OK, going to punt on tweaking what I have now but like your ideas.

Is it possible to enable/disable mocks while the app is running? Or would the dev dock kill the app and restart (rebuild)?

jonahkagan: That's a good question and I don't know the answer

Issue #3951, opened 2023-09-11
Test at scale on real hardware


arsalansufi:

arsalansufi:
- Notes here: https://votingworks.slack.com/archives/CEL6D3GAD/p1696262663599539
- Worked on an optimization here: https://github.com/votingworks/vxsuite/pull/4056
- Main next step is testing with the all-bubble ballot

arsalansufi:
Opened a new issue to track all-bubble ballot testing: https://github.com/votingworks/vxsuite/issues/4106

PR #4056, opened 2023-10-09
Don't require reading in all images and layout files for CVR authentication

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3951

While testing CVR export and import, I identified one notable (and expected) source of slowness. VxScan CVR exports now always include all images and layout files, not just images and layout files for HMPBs with write-ins, even though VxAdmin only stores the latter on import. Despite VxAdmin only storing the latter, it currently reads in all images and layout files to compute the root hash for CVR authentication.

Putting some numbers to this slowness, I scanned 1000 NH test ballots, none with write-ins, on VxCentralScan and VxScan. Importing the VxCentralScan-exported CVRs on a ThinkPad VxAdmin took **2** seconds. Importing the VxScan-exported CVRs, on the other hand, took **28** seconds - a notable difference.

This PR tweaks CVR authentication, adding image and layout file hashes to `cast-vote-record-report.json` files and only incorporating `cast-vote-record-report.json` file hashes into the root hash. When VxAdmin actually encounters a CVR for which we need to store images and layout files (i.e. a CVR for an HMPB with a write-in), it'll rehash the images and layout files and compare the hashes against the known good hashes in the authenticated `cast-vote-record-report.json` file. I've written the code such that you can't forget to perform this secondary hash check - the check is built into the abstraction. This completely gets rid of the discrepancy between VxCentralScan-export import time and VxScan-export import time :tada:.

The diff in this PR is so large because I had to regenerate CVR fixtures to account for the format change.

## Testing

- [x] Updated automated tests and added some new ones
- [x] Tested export and import manually

## Checklist


libs/types/src/cdf/cast-vote-records/index.ts
```diff
@@ -1468,6 +1468,11 @@ export interface ImageData {
    * A pointer to the location of the image file.
    */
   readonly Location?: Uri;
+
+  /**
+   * The hash of the VotingWorks layout file corresponding to this image.
+   */
+  readonly vxLayoutFileHash?: string;
```
arsalansufi, line 1475: The CDF CVR schema includes fields for image hashes, which is great! For layout file hashes, I've opted for a custom field on CDF `CVR.ImageData` objects.

I could've alternatively created a separate file for those hashes, but that would've complicated upstream code.

libs/types/src/cast_vote_records.ts
```diff
@@ -39,6 +40,12 @@ export interface CastVoteRecord
   readonly _scannerId: string;
 }
 
+export enum CastVoteRecordExportFileName {
```
arsalansufi, line 43: Nothing new in this file - just moving existing types from `libs/backend` into it (and shuffling some existing types around)

libs/types/src/cast_vote_records.ts
```diff
@@ -101,3 +115,36 @@ export type SheetValidationError = {
...
+export type ReadCastVoteRecordExportError =
+  | ReadCastVoteRecordExportMetadataError
+  | { type: 'authentication-error' };
+
+type ReferencedFile = 'image' | 'layout-file';
+
+/**
+ * An error encountered while reading an individual cast vote record
+ */
+export type ReadCastVoteRecordError = { type: 'invalid-cast-vote-record' } & (
```
arsalansufi, line 139: This is the one type that I've actually tweaked a bit, adding some additional error `subType` values, which you'll see get used in subsequent commits

libs/auth/src/cast_vote_record_hashes.ts
```diff
@@ -78,29 +78,25 @@ export function readableFileFromDisk(filePath: string): ReadableFile {
...
-  for (const file of filesSorted) {
-    const filePath = path.join(directoryName, file.fileName);
-    // Be extra cautious and prevent spoofing the directory summary by using directory/file names
-    // with newlines
-    assert(!filePath.includes('\n'));
-    hasher.update(`${await file.computeSha256Hash()}  ${filePath}\n`);
-  }
-  return hasher.hex();
+  const cvrDirectorySummary = `${await cvrReportFile.computeSha256Hash()}  ${cvrReportPath}\n`;
+  return sha256(cvrDirectorySummary);
```
arsalansufi, line 99: You might be wondering why we have to hash this special string and can't just hash the `cast-vote-record-report.json`, alone, as the hash for a single CVR. Incorporating the name of the CVR directory into the hash (by way of `cvrReportPath`) is important because it prevents renaming a CVR directory.

Without this, CVR directory renames that keep the overall alphabetical order of the CVR directories the same could go undetected.

It was actually a `libs/auth` test that reminded me to cover this case! It failed on my first pass at this update, where I was directly using the hash of the `cast-vote-record-report.json` as the hash for the CVR.

https://github.com/votingworks/vxsuite/blob/93b214c8dc9ca8eca583e118650f919e8f662bc2/libs/auth/src/artifact_authentication.test.ts#L314-L315

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
```diff
@@ -55,6 +55,35 @@ export function toCdfBallotType(ballotType: BallotType): CVR.vxBallotType {
   }
 }
 
+/**
+ * The input to {@link buildCvrImageData}
+ */
+export interface CvrImageDataInput {
+  imageHash: string;
+  imageRelativePath: string;
+  layoutFileHash?: string;
```
arsalansufi, line 64: Optional because, even when exporting images, we don't export layout files for BMD ballots

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -359,6 +372,33 @@ async function exportCastVoteRecordFilesToUsbDrive(
       canonicalizedSheet.interpretation.some(({ votes }) => hasWriteIns(votes))
     : true;
 
+  const castVoteRecordFilesToExport: ReadableFile[] = [];
```
arsalansufi, line 375: Had to change the order of operations a bit. I now export images and layout files first (if applicable) and then the `cast-vote-record-report.json`, as opposed to the other way around, since the `cast-vote-record-report.json` needs the hashes of the images and layout files

libs/backend/src/cast_vote_records/referenced_files.ts
```diff
@@ -0,0 +1,123 @@
...
+        subType: 'image-read-error',
+      });
+    }
+
+    if (sha256(fileContents) !== this.expectedFileHash) {
+      return err({
+        type: 'invalid-cast-vote-record',
+        subType: 'incorrect-image-hash',
+      });
+    }
```
arsalansufi, line 69: As noted in the PR description:

> I've written the code such that you can't forget to perform this secondary hash check - the check is built into the abstraction.

This is where that happens! Calling `.read()` on a `ReferencedImageFile` or `ReferencedLayoutFile` automatically checks its hash.

libs/backend/src/cast_vote_records/import.ts
```diff
@@ -31,50 +36,11 @@ import {
...
 
 import { TEST_OTHER_REPORT_TYPE } from './build_report_metadata';
-
-/**
- * A cast vote record report without metadata
- */
-export type CastVoteRecordReportWithoutMetadata = Pick<
-  CVR.CastVoteRecordReport,
-  'CVR'
->;
```
arsalansufi, line 41: These are all the types that I moved to `libs/types` in an earlier commit

libs/cvr-fixture-generator/src/cli/generate/main.ts
```diff
@@ -261,9 +256,29 @@ export async function main(
...
+      ...reportMetadata,
+      CVR: [castVoteRecord],
+    };
+    fs.writeFileSync(
+      join(
+        castVoteRecordDirectory,
+        CastVoteRecordExportFileName.CAST_VOTE_RECORD_REPORT
+      ),
+      JSON.stringify(castVoteRecordReport)
+    );
```
arsalansufi, line 281: As in `libs/backend`, I had to change the order of operations here, exporting the `cast-vote-record-report.json` after the images and layout files (if applicable)

libs/cvr-fixture-generator/src/generate_cvrs.ts
```diff
@@ -383,3 +400,19 @@ export function* generateCvrs({
     }
   }
 }
+
+/**
+ * Retroactively populates the hashes on a cast vote record image data object
+ */
+export function populateImageAndLayoutFileHashes(
```
arsalansufi, line 407: I know this is a bit hacky 🙈

It was the easiest way to get the job done without significantly refactoring this code, which I didn't want to spend time on given my eventual plan to fully overhaul CVR fixture generation to use actual CVR export functions.

I am taking a quick shortcut in that I'm only populating hashes on the `CVR.ImageData` objects stored at `BallotImage`, not the ones stored at `WriteInImage`, since we don't actually use those and it would notably complicate this function.

libs/fixtures/.prettierignore
```diff
@@ -0,0 +1 @@
+/**/castVoteRecords/generated
```
arsalansufi, line 1: `libs/cvr-fixture-generator` automatically formats staged JSON files on commit, including generated CVR fixture data. This poses a problem because, post-formatting, the files' hashes will no longer match the hashes used to construct the generated signature file. I _think_ that this is why you ran into auth issues while working on https://github.com/votingworks/vxsuite/pull/4039 @adghayes!

arsalansufi:
FYI @benadida this is implementing the idea discussed [here in Slack](https://votingworks.slack.com/archives/CEL6D3GAD/p1696262837417009?thread_ts=1696262663.599539&cid=CEL6D3GAD)!

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
eventualbuddha: Thanks for the note. Might it be better as a code comment?

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -278,14 +283,22 @@ function buildCastVoteRecord(
...
+          layoutFileHash:
+            await referencedFiles.layoutFiles?.[0].computeSha256Hash(),
+        },
+        {
+          imageHash: await referencedFiles.imageFiles[1].computeSha256Hash(),
+          imageRelativePath: referencedFiles.imageFiles[1].fileName,
+          layoutFileHash:
+            await referencedFiles.layoutFiles?.[1].computeSha256Hash(),
+        },
+      ]
```
eventualbuddha, line 300: Seeing this sort of thing prompted me to create #4059.

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
arsalansufi: Good call: https://github.com/votingworks/vxsuite/pull/4056/commits/9072ae191f935d2a4a15886d0be6f419ea1e3fcf

libs/basics/src/errors.ts
```diff
@@ -4,3 +4,15 @@
 export function extractErrorMessage(error: unknown): string {
   return error instanceof Error ? error.message : String(error);
 }
+
+/**
+ * Determines whether an error is a Node fs module non-existent file or directory error
+ */
+export function isNonExistentFileOrDirectoryError(error: unknown): boolean {
```
eventualbuddha, line 11: Arguably this belongs in `@votingworks/backend` since the frontends use `/basics` as well. Since it happens to work in that environment I don't have a strong preference, but something to consider.

libs/backend/src/cast_vote_records/referenced_files.ts
```diff
@@ -0,0 +1,123 @@
...
+      return err({
+        type: 'invalid-cast-vote-record',
+        subType: 'layout-file-parse-error',
+      });
+    }
+    const layout = parseResult.ok();
+
+    return ok(layout);
+  }
+}
```
eventualbuddha, line 123: These two classes are really similar. Consider consolidating some of the shared logic in composible functions (as opposed to subclassing)?

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
```diff
@@ -585,11 +607,6 @@ export function buildCastVoteRecord({
...
           }),
         ],
-    BallotImage: imageFileUris
-      ? imageFileUris.map((imageFileUri) => ({
-          '@type': 'CVR.ImageData',
-          Location: imageFileUri,
-        }))
-      : undefined,
+    BallotImage: images ? images.map(buildCvrImageData) : undefined,
   };
```
eventualbuddha, line 611:
```suggestion
    BallotImage: images?.map(buildCvrImageData),
```

eventualbuddha (approved):
Thanks for all the clarification comments! I've reviewed this for code style and general functionality, but haven't actually tried it.

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
eventualbuddha: Leaving such comments in VS Code does not always translate well. I think you can see what I meant.

benadida:
> FYI @benadida this is implementing the idea discussed [here in Slack](https://votingworks.slack.com/archives/CEL6D3GAD/p1696262837417009?thread_ts=1696262663.599539&cid=CEL6D3GAD)!

Awesome, thank you! High level looks good, leaving the detailed code review to @eventualbuddha

libs/fixtures/.prettierignore
adghayes: Good find!

libs/basics/src/errors.ts
arsalansufi: Good point! I'm leaning toward keeping the helper in `libs/basics` because `libs/backend` depends on other modules that could potentially use this helper, e.g. `libs/auth`, `libs/db`, `libs/logging`, `libs/usb-drive`. If we moved the helper into `libs/backend`, we'd introduce a circular dependency by having those other modules pull from `libs/backend`

libs/backend/src/cast_vote_records/export.ts
arsalansufi: Ended up using the new and improved `mapSheet` at this very code point, too, haha: https://github.com/votingworks/vxsuite/pull/4056/commits/d2cf1cebee0df2a2be2039a69d95a001eed1a043

libs/backend/src/cast_vote_records/referenced_files.ts
arsalansufi: Updated the code to 1) better share logic and 2) use composable functions (a pattern that I haven't really used before but like!): https://github.com/votingworks/vxsuite/pull/4056/commits/89734d1adb6d1aa39c14280beb164019167c72ed

PR #4059, opened 2023-10-09
refactor: use `mapSheet` with split sheet data

eventualbuddha:
## Overview

We fairly often have data for a single sheet spread across multiple tuples rather than nested within a single tuple. To make this easier to work with, `mapSheet` now accepts multiple `SheetOf` values and spreads the pages from each in the call to the provided function.

## Demo Video or Screenshot
```ts
expect(mapSheet([1, 2], [3, 4], (x, y) => x + y)).toEqual([4, 6]);
```

## Testing Plan
Automated.


arsalansufi (approved):
Nice, I like this!

PR #4064, opened 2023-10-10
fix(auth): do more validation of TLV messages

eventualbuddha:
## Overview
We were not validating:
- that the encoded tag is the one we expected
- that the first length byte is ≤ 0x80 if it isn't one of the special 0x81 or 0x82 values
- that the TLV response didn't contain more data than the length said it should

See [Slack discussion](https://votingworks.slack.com/archives/CEL6D3GAD/p1696958808975429).

## Demo Video or Screenshot
n/a

## Testing Plan
Automated, manually tested that programming a sysadmin card and using it in VxScan still works.

### DIFF
--- a/libs/auth/src/apdu.ts
+++ b/libs/auth/src/apdu.ts
@@ -1,6 +1,7 @@
 /* eslint-disable max-classes-per-file */
 import { Buffer } from 'buffer';
-import { assert } from '@votingworks/basics';
+import { inspect } from 'util';
+import { assert, assertDefined } from '@votingworks/basics';
 import { asHexString, Byte, isByte } from '@votingworks/types';
 
 /**
@@ -179,7 +180,7 @@ export class CardCommand {
     this.ins = input.ins;
     this.p1 = input.p1;
     this.p2 = input.p2;
-    this.data = input.data ?? Buffer.from([]);
+    this.data = input.data ?? Buffer.of();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


libs/auth/src/apdu.ts
```diff
@@ -262,24 +267,54 @@ export function constructTlv(
...
+    tagBytes.equals(expectedTagBytes),
+    `TLV tag (${inspect(tagBytes)}) does not match expected tag (${inspect(
+      expectedTagBytes
+    )})`
+  );
 
-  let tlvLengthLength = 1; // Derpiest variable name, I know
-  const tlvLengthFirstByte = tlv.at(tagLength);
+  let tlvLengthLength: number;
+  let tlvLength: number;
```
arsalansufi, line 284: Maybe for consistency with the latest variable names across `constructTlv` and `parseTlv`, these should now be:

```ts
let lengthBytesLength: number;
let valueBytesLength: number;
```

libs/auth/src/apdu.ts
```diff
@@ -262,24 +267,54 @@ export function constructTlv(
...
+  const tlvLengthFirstByte = assertDefined(tlv.at(expectedTagLength));
+  const tlvLengthSecondByte = tlv.at(expectedTagLength + 1);
+  const tlvLengthThirdByte = tlv.at(expectedTagLength + 2);
```
arsalansufi, line 287: Slight preference for defining `tlvLengthSecondByte` and `tlvLengthThirdByte` within the `if` conditions for which they're relevant. Otherwise, what we store in `tlvLengthSecondByte` or `tlvLengthThirdByte` might actually be the start of the TLV value

libs/auth/src/apdu.ts
```diff
@@ -262,24 +267,54 @@ export function constructTlv(
...
+      assertDefined(tlvLengthThirdByte);
+  } else if (tlvLengthFirstByte <= 0x80) {
+    tlvLengthLength = 1;
+    tlvLength = assertDefined(tlvLengthFirstByte);
+  } else {
+    throw new Error(
+      `TLV length is invalid: received 0x${tlvLengthFirstByte.toString(
+        16
+      )}, but expected a value <= 0x80`
+    );
```
arsalansufi, line 305: Since `0x81` and `0x82` are valid first length bytes as well, a message like this might be preferable:

> TLV length is invalid: Received 0xXX but expected a value <= 0x82 for the first length byte

arsalansufi (commented):
Thank you for updating! A few nits

libs/auth/src/apdu.ts
```diff
@@ -225,35 +226,39 @@ export function constructTlv(
...
     ? tagAsByteOrBuffer
     : Buffer.of(tagAsByteOrBuffer);
 
   /**
    * The convention for TLV length is as follows:
    * - 0xXX           if value length < 128 bytes
-   * - 0x81 0xXX      if value length > 128 and < 256 bytes
-   * - 0x82 0xXX 0xXX if value length > 256 and < 65536 bytes
+   * - 0x81 0xXX      if value length ≥ 128 and < 256 bytes
+   * - 0x82 0xXX 0xXX if value length ≥ 256 and < 65536 bytes
```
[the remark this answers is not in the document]
arsalansufi: ^ FYI I had to edit a few times to get the above right haha, so you might need to refresh 😅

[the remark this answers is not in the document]
eventualbuddha: Thanks, inequality errors are a close cousin to off-by-1 errors 😆

libs/auth/src/apdu.ts
```diff
@@ -262,24 +267,54 @@ export function constructTlv(
...
+    `TLV tag (${inspect(tagBytes)}) does not match expected tag (${inspect(
+      expectedTagBytes
+    )})`
+  );
+
+  let lengthBytesLength: number;
+  let valueBytesLength: number;
+  const lengthBytesFirst = assertDefined(tlv.at(expectedTagLength));
+  if (lengthBytesFirst === 0x81) {
+    const tlvLengthSecondByte = tlv.at(expectedTagLength + 1);
```
arsalansufi, line 287: A missed rename here - I think we want `lengthBytesSecond`

arsalansufi (approved):
LGTM!

PR #4073, opened 2023-10-11
Manual Counts Policy - Ballot Count Reporting

adghayes:
## Overview

@mattroe and I discussed how our policy with manual results is that we _always_ show them separately on reports where they contribute to the overall ballot count or vote count. This is what we do for the tally reports - if there are any manual results, the reports flip to show manual counts.

When I built the ballot count reporting (PDF and CSV), I included options to show the ballot count breakdown differently. 'none' would never show manual counts, and 'manual' would always show manual counts. This is in violation of our decided policy.

In this PR, I simply remove the idea the ballot count breakdown as an editable feature - it is always the maximal option - BMD, HMPB, Total and, if applicable, Manual.

### DIFF
--- a/apps/admin/backend/src/app.ballot_count_report_csv_export.test.ts
+++ b/apps/admin/backend/src/app.ballot_count_report_csv_export.test.ts
@@ -62,7 +62,6 @@ it('logs success if export succeeds', async () => {
   const offLimitsPath = '/root/hidden';
   const failedExportResult = await apiClient.exportBallotCountReportCsv({
     path: offLimitsPath,
-    ballotCountBreakdown: 'none',
   });
   expect(failedExportResult.isErr()).toEqual(true);
   expect(logger.log).toHaveBeenLastCalledWith(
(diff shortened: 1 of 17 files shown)


mattroe:
@adghayes what do you think about implementing this as a checkbox under report by instead? Such as Report By "Ballot Type" instead of having a distinct option?

adghayes:
I think of it as a dropdown because I'm assuming we will eventually add sheet counts which will be in the same dropdown. That being said, we don't have sheet counts yet, so it could be a checkbox for now. 

re: collapsing the "Report By," here's what I wrote in the polish doc:

> Interesting observation. You can see on the chart that "Report By" breaks things onto new rows whereas the breakdown breaks things across rows. This is necessary to show row totals. From a technical standpoint, they are very different - what CVRs do we filter for vs. how do we display the data.

> I know it seems weird as a singular option but I think that has more to do with the fact that we only have one option so far rather than that it should be with "Report By."

> I would push to leave this as-is, if anything because the implementation makes more sense this way.

adghayes:
Yup, there's no reason we can't apply the same UX. And maybe you're right, that the "Report By" UX would be clearer for users as it was for you. Happy to make it that way if that's the consensus.

But I personally still don't like it, even more so for the sheet example than the previous example. Because the "Report By" entails some splitting of results into its constituent parts. Separating sheet counts is sort of that, sort of not... because the sum of sheet counts is not the ballot count. It could be a cute little checkbox.

It's not a hill I'll die on but that's my gut. It's possible I just can't see through the implementation to the UX.

kofi-q (approved):

PR #4074, opened 2023-10-11
Manual Tally Policy - CSV Tally Reports

### LINKED ISSUE #4069: VxAdmin Reports: Standardize Manual Counts Policy
We are not always showing manual counts in ballot count reports, and tally report CSV exports. Our policy is to always show manual counts separately if they exist.

adghayes:
_review by commit, there's a refactor and bug fix in the log_

## Overview

Closes #4069 alongside #4073. Updates the tally CSV exports so that they split manual and scanned results.

## Demo Video or Screenshot

N/A, example file attached:
[TEST-full-election-tally-report__2023-10-23_13-11-49.csv](https://github.com/votingworks/vxsuite/files/13078796/TEST-full-election-tally-report__2023-10-23_13-11-49.csv)

## Testing Plan

Automated testing.

## Checklist

### DIFF
--- a/apps/admin/backend/src/exports/csv_tally_report.ts
+++ b/apps/admin/backend/src/exports/csv_tally_report.ts
@@ -7,10 +7,12 @@ import {
   AnyContest,
   Election,
 } from '@votingworks/types';
-import { assert, assertDefined } from '@votingworks/basics';
+import { Optional, assert, assertDefined } from '@votingworks/basics';
 import {
   combineGroupSpecifierAndFilter,
+  getTallyReportCandidateRows,
   groupMapToGroupList,
+  mergeTabulationGroupMaps,
 } from '@votingworks/utils';
 import { Readable } from 'stream';
 import { Store } from '../store';
@@ -22,16 +24,28 @@ import {
   generateCsvMetadataHeaders,
   getCsvMetadataRowValues,
 } from './csv_shared';
+import { tabulateManualResults } from '../tabulation/manual_results';
+
+// eslint-disable-next-line vx/gts-no-return-type-only-generics
+function assertIsOptional<T>(_value?: unknown): asserts _value is Optional<T> {
+  // noop
+}
 
 function generateHeaders({
   election,
   metadataStructure,
+  hasManualResults,
 }: {
   election: Election;
   metadataStructure: CsvMetadataStructure;
+  hasManualResults: boolean;
 }): string[] {
   const headers = generateCsvMetadataHeaders({ election, metadataStructure });
-  headers.push('Contest', 'Contest ID', 'Selection', 'Selection ID', 'Votes');
+  headers.push('Contest', 'Contest ID', 'Selection', 'Selection ID');
+  if (hasManualResults) {
+    headers.push('Manual Votes', 'Scanned Votes');
+  }
+  headers.push('Total Votes');
   return headers;
 }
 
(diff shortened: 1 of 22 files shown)


libs/utils/src/tabulation/tabulation.ts
```diff
@@ -541,7 +541,9 @@ export function combineCandidateContestResults({
         combinedContestResults.tallies[candidateTally.id];
 
       if (!combinedCandidateTally) {
-        combinedContestResults.tallies[candidateTally.id] = candidateTally;
+        combinedContestResults.tallies[candidateTally.id] = {
+          ...candidateTally,
+        };
```
adghayes, line 545: 🙈

libs/utils/src/tabulation/tally_reports.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export function getTallyReportCandidateRows({
+  contest,
+  scannedContestResults,
+  manualContestResults,
+}: {
+  contest: CandidateContest;
+  scannedContestResults: Tabulation.CandidateContestResults;
+  manualContestResults?: Tabulation.CandidateContestResults;
+}): TallyReportCandidateRow[] {
```
adghayes, line 66: I think that for the use case in this PR, this is a little overbuilt. But. I'm breaking this logic out into a util because I will use the same logic for frontend tally reports, and add different algorithms for grouping / showing write-in candidates. Either "show all",  or "show significant." Future features, like the ability to have official write-in candidates, could add more algorithms here. Will build on this in next PR.

libs/types/src/election.ts
```diff
@@ -142,11 +142,6 @@ export const WriteInCandidateSchema: z.ZodSchema<WriteInCandidate> = z.object({
   partyIds: z.array(PartyIdSchema).optional(),
 });
 
-export const writeInCandidate: Candidate = {
-  id: 'write-in',
-  name: 'Write-In',
-  isWriteIn: true,
-};
```
adghayes, line 149: While not strictly necessary, I thought it was better to keep this under `tabulation` rather than `election` because it really is a tabulation tool. Also, use variable naming to indicate it represents a generic or pending write-in, rather than perhaps thinking it is an actual write-in.

libs/types/src/tabulation.ts
```diff
@@ -198,6 +198,32 @@ export interface ElectionWriteInSummary {
...
+ * Represents an unadjudicated write-in, which ultimately may not be a vote
+ * for a candidate.
+ */
+export const GENERIC_WRITE_IN_CANDIDATE: Candidate = {
+  id: GENERIC_WRITE_IN_ID,
+  name: GENERIC_WRITE_IN_NAME,
+  isWriteIn: true,
+};
+
+export const PENDING_WRITE_IN_ID = GENERIC_WRITE_IN_ID;
```
adghayes, line 215: I am leaving these two with the same IDs because they should never exist in a list together, and changing IDs risked breaking something. But I separated out `PENDING` from `GENERIC` to give them different default names, and to make code using them more readable.

apps/admin/backend/src/exports/csv_tally_report.ts
```diff
@@ -22,16 +24,28 @@ import {
...
   metadataStructure: CsvMetadataStructure;
+  hasManualResults: boolean;
 }): string[] {
   const headers = generateCsvMetadataHeaders({ election, metadataStructure });
-  headers.push('Contest', 'Contest ID', 'Selection', 'Selection ID', 'Votes');
+  headers.push('Contest', 'Contest ID', 'Selection', 'Selection ID');
+  if (hasManualResults) {
+    headers.push('Manual Votes', 'Scanned Votes');
+  }
+  headers.push('Total Votes');
```
adghayes, line 48: If there's no manual results, we could just keep the column name here as "Votes." But I assume for external systems consuming our .csv files, it will be less confusing if the column names do not change.

kofi-q (approved):

PR #4075, opened 2023-10-11
[i18n/audio] Replace a few strings in Mark[-Scan] with language-aware versions

kofi-q:
## Overview

_Easier by commit_

Adding `@votingworks/ui/appStrings` and `@votingworks/ui/electionStrings` shared string component renderers (building off `<UiString>`) and starting off with a few proof-of-concept strings to try out different usage patterns.

Replacing a few strings on the `CandidateContest` pages in `Mark` and `Mark-Scan`.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/264902/aaee2716-ddc2-4f62-b6c7-a80358c11958

## Testing Plan
- Manual: populated the DB with test translations and added the `UiStringContextProvider` to the app root of `Mark` to verify functionality. Using a stand-in language picker for testing until we have an actual voter-facing component.
- Added unit tests for the `renderCandidatePartyList` helper function.
- Will later add some sanity check tests for things like making sure the object keys match the `uiStringKey`s, since that feels easy to mess up or end up with copy/paste errors.

## Checklist

### DIFF
--- a/libs/ui/src/ui_strings/election_strings.tsx
+++ b/libs/ui/src/ui_strings/election_strings.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 
 import {
   Candidate,
-  Contest,
   Parties,
   Party,
   getCandidateParties,
@@ -14,9 +13,9 @@ import { UiString } from './ui_string';
 export const electionStrings = {
   // TODO(kofi): Fill out.
 
-  // NOTE: Omitting 'type' to support both the `Contest` and the
+  // NOTE: Using more lenient typing to support both the `Contest` and the
   // `MsEitherNeitherContest` types.
-  contestTitle: (contest: Omit<Contest, 'type'>) => (
+  contestTitle: (contest: { id: string; title: string }) => (
     <UiString uiStringKey="contestTitle" uiStringSubKey={contest.id}>
       {contest.title}
     </UiString>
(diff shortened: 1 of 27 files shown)


libs/ui/src/ui_strings/election_strings.tsx
```diff
@@ -0,0 +1,56 @@
...
+} from '@votingworks/types';
+
+import { UiString } from './ui_string';
+
+/* istanbul ignore next - mostly presentational, tested via apps where relevant */
+export const electionStrings = {
+  // TODO(kofi): Fill out.
+
+  // NOTE: Omitting 'type' to support both the `Contest` and the
+  // `MsEitherNeitherContest` types.
```
eventualbuddha, line 18: Would it be better to do either say which types you're expecting (`Contest | MsEitherNeitherContest`) or declare the shape of the object you need (`{ id: string; title: string }`)?

kofi-q: Yeah, good call. I initially wanted slightly stricter typing for the contest, since I thought `{ id, title }` might match some other election objects, but shouldn't be an issue, now that I think about it. (Couldn't go the `Contest | MsEitherNeitherContest` route without additional work, since `MsEitherNeitherContest` is declared in `mark-flow-ui`)

arsalansufi (approved):
LGTM - exciting to see things come together!

PR #4082, opened 2023-10-11
VSAP USB Drive Hotfix

adghayes:
VSAP has two "devices" which are causing problems:
- a USB partition acting as / formatted as an LVM (logical volume manager). we can disallow this by skipping drives which have a filesystem type of "LVM." There's no documentation I can find on a proper enumeration of all file types, so I'm simply excluding file systems which include "LVM" in the name
-  "logical volume" that is part of that LVM. this is easy to exclude, because we can limit our search to proper partitions

### DIFF
--- a/libs/usb-drive/src/usb_drive.test.ts
+++ b/libs/usb-drive/src/usb_drive.test.ts
@@ -40,6 +40,7 @@ function lsblkOutput(devices: Array<Partial<BlockDeviceInfo>> = []) {
         fstype: 'vfat',
         fsver: 'FAT32',
         label: 'VxUSB-00000',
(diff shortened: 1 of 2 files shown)


kshen0 (approved):
Confirmed this works on the hardware. Specifically
* `blockDeviceInfo.type === 'part'` catches `vx--vg-hashes`
* `!blockDeviceInfo.fstype?.includes('LVM')` catches the LVM2_member drive

adghayes:
Taking a second look at this, I think that the conditions are indeed the conditions we want to check. So, I'm merging and treating issue as closed, rather than a hotfix we need to revisit.

PR #4085, opened 2023-10-12
Remove legacy color and size themes

jonahkagan:
## Overview

Removes the `legacy` color and size themes, which we no longer use. This cleanup will help make it easier to implement the new desktop theme by removing old code that I would otherwise have to maintain.

Updates VxDesign to temporarily use the medium contrast color/small size theme, since it was previously using the legacy theme. It's useable with this theme, but everything is a bit too big. I'm planning to follow up this PR with a rudimentary desktop theme that VxDesign can use while I flesh out the eventual more complex desktop theme.

_Review by commit_

## Demo Video or Screenshot

VxDesign too big!
Screenshot 2023-10-12 at 9 37 14 AM

## Testing Plan
Automated tests

## Checklist


apps/design/frontend/src/app.tsx
```diff
@@ -74,7 +74,7 @@ const StyleOverrides = styled.div`
 
 export function App(): JSX.Element {
   return (
-    <AppBase>
+    <AppBase defaultColorMode="contrastMedium" defaultSizeMode="s">
```
kofi-q, line 77: could optionally set the `legacyBaseFontSizePx` prop to the more standard `16` or `14` to avoid having large UI elements in the interim (although, I think the button controls would still be pretty large because of the 'min touch area' setting on the `s` theme 😅 )

libs/ui/src/themes/make_theme.ts
```diff
@@ -82,18 +81,6 @@ const colorThemes: Record<ColorMode, ColorTheme> = {
     foreground: Color.GRAY_LIGHT,
     foregroundDisabled: Color.GRAY_LIGHT,
   },
-
-  legacy: {
```
kofi-q, line 86: could you also update the storybook preview setup code in `.storybook/preview.tsx`? makes me realise we need to add a storybook build step to the CI process to make sure it stays in sync

libs/ui/src/big_metric.tsx
```diff
@@ -24,9 +24,7 @@ const StyledContainer = styled.span`
 
 const StyledValue = styled(H1)`
   display: inline-block;
-
-  /* TODO: Remove once VxScan is updated to default to the new VVSG size theme: */
-  font-size: ${(p) => (p.theme.sizeMode === 'legacy' ? '72px' : undefined)};
+  font-size: 72px;
```
kofi-q, line 27: I think this needs to be removed for parity with the old behaviour

libs/ui/src/themes/make_theme.ts
adghayes: I tested `storybook` on this branch which built just fine, but breaks when you try to select the still extant Legacy color or size options. That being the case, I'm not sure how we could catch this in an automated way.

apps/design/frontend/src/app.tsx
jonahkagan: Yeah, I am just gonna follow up with a quick desktop size theme PR to deal with this.

libs/ui/src/themes/make_theme.ts
jonahkagan: I think the problem is just that the `.storybook` dir isn't being type-checked. I added it to the tsconfig and that caught the problem.

libs/ui/src/big_metric.tsx
jonahkagan: Oh whoops, that was a mistake. Thanks for catching!

libs/ui/src/themes/make_theme.ts
jonahkagan: 6bb3e7ea8

adghayes (approved):

PR #4086, opened 2023-10-12
[i18n/Audio] Adding renderers and helpers for remaining election strings

kofi-q:
## Overview

_Easier by commit_

Fleshing out `@votingworks/ui/electionStrings` to include all the translatable/spoken strings in the election definition so far.

Making the `VxMark` `StartPage` language-aware to test out the new election strings, which led to adding support for number and date i18n -- the audio piece of that is going to require a bit more planning and coordination with the VxDesign work, so I'll start writing up some thoughts in the spec.
- Also, to avoid audio interpolation, the previous `"Your ballot contains x contests."` string is now `"Number of contests on your ballot: x"`

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/264902/481d65a7-4c80-4d98-bcdd-3794d9613fca

## Testing Plan
- Unit tests for the new ui string utils
- Updated expectations in existing tests for changed strings/HTML structures
- Manually tested `StartPage` in different languages with mock translations.

## Checklist

- ~[x] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates~

### DIFF
--- a/libs/ui/src/ui_strings/election_strings.tsx
+++ b/libs/ui/src/ui_strings/election_strings.tsx
@@ -1,22 +1,104 @@
-import { Party } from '@votingworks/types';
+/* eslint-disable react/destructuring-assignment */
+
+import {
+  BallotStyleId,
+  Candidate,
+  County,
+  District,
+  Election,
+  ElectionStringKey as Key,
+  Party,
+  Precinct,
+  YesNoOption,
+} from '@votingworks/types';
 
 import { UiString } from './ui_string';
 
+// Using more lenient typing to support both the `Contest` and the
+// `MsEitherNeitherContest` types.
+interface ContestLike {
+  id: string;
+  title: string;
+}
+
+type ContestWithDescription = ContestLike & {
+  description: string;
+};
+
+/**
+ * Election-specific strings that need to be translated and/or spoken.
+ */
 /* istanbul ignore next - mostly presentational, tested via apps where relevant */
 export const electionStrings = {
   // TODO(kofi): Fill out.
 
-  // NOTE: Using more lenient typing to support both the `Contest` and the
-  // `MsEitherNeitherContest` types.
-  contestTitle: (contest: { id: string; title: string }) => (
-    <UiString uiStringKey="contestTitle" uiStringSubKey={contest.id}>
+  [Key.BALLOT_STYLE_ID]: (id: BallotStyleId) => (
+    <UiString uiStringKey={Key.BALLOT_STYLE_ID} uiStringSubKey={id}>
+      {id}
+    </UiString>
+  ),
+
+  [Key.CANDIDATE_NAME]: (candidate: Candidate) => (
+    <UiString uiStringKey={Key.CANDIDATE_NAME} uiStringSubKey={candidate.id}>
+      {candidate.name}
+    </UiString>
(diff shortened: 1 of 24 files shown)


libs/ui/src/ui_strings/number_string.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <H1>
+      <NumberString value={100000} />
+    </H1>
+  );
+
+  await screen.findByRole('heading', { name: '100,000' });
+
+  act(() => getLanguageContext()?.setLanguage(LanguageCode.SPANISH));
+
+  await screen.findByRole('heading', { name: '100.000' });
```
arsalansufi, line 27: I've always thought the smaller `.` for aesthetic separator and bigger `,` for decimal separator made more sense lol

libs/ui/src/ui_strings/election_strings.tsx
```diff
@@ -1,22 +1,104 @@
-import { Party } from '@votingworks/types';
+/* eslint-disable react/destructuring-assignment */
```
arsalansufi, line 1: Just curious, why is this necessary?

libs/ui/src/ui_strings/utils.tsx
```diff
@@ -29,3 +40,32 @@ export function renderCandidatePartyList(
     </React.Fragment>
   );
 }
+
+export function renderPrecinctSelectionName(
+  electionPrecincts: readonly Precinct[],
+  precinctSelection?: PrecinctSelection
+): React.ReactNode {
```
arsalansufi, line 47: Any particular reason for making these helpers functions that return JSX instead of components? E.g. this could alternatively be expressed as:

```jsx
<PrecinctSelectionName
  electionPrecincts={electionPrecincts}
  precinctSelection={precinctSelection}
/>
```

arsalansufi (approved):
LGTM!

libs/ui/src/ui_strings/election_strings.tsx
kofi-q: Not sure why this happened, but once I switched over to using the `ElectionStringKey` enum as the keys, the eslint rule started interpreting the method params as `props` for components, so it complained about not destructuring `candidate` before referencing `candidate.id`, for example.

libs/ui/src/ui_strings/number_string.test.tsx
kofi-q: same!

libs/ui/src/ui_strings/utils.tsx
kofi-q: oh, yeah! thanks for flagging - i initially had `candidatePartyList` as a method in the `electionStrings` mapping and didn't rethink the pattern when i broke it out into a separate function. these make more sense as components

libs/ui/src/ui_strings/election_strings.tsx
arsalansufi: Got it, thanks for explaining

libs/ui/src/ui_strings/utils.tsx
arsalansufi: Makes sense!

PR #4090, opened 2023-10-12
Get the new CVR export functions in libs/backend up to 100% code coverage

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4058

Finally adding the tests that I've punted on for so so long. This PR gets the new CVR export functions in `libs/backend` up to 💯% code coverage. I'll be opening a separate PR for the new CVR import functions.

### DIFF
--- a/apps/scan/backend/src/app_config.test.ts
+++ b/apps/scan/backend/src/app_config.test.ts
@@ -35,8 +35,8 @@ import {
 } from '@votingworks/backend';
 import { InsertedSmartCardAuthApi } from '@votingworks/auth';
 import { CVR, ElectionDefinition } from '@votingworks/types';
-import { configureApp } from '../../../test/helpers/shared_helpers';
-import { scanBallot, withApp } from '../../../test/helpers/custom_helpers';
+import { configureApp } from '../test/helpers/shared_helpers';
+import { scanBallot, withApp } from '../test/helpers/custom_helpers';
 
 jest.setTimeout(20_000);
 
--- a/libs/backend/src/cast_vote_records/export.ts
+++ b/libs/backend/src/cast_vote_records/export.ts
@@ -1,4 +1,3 @@
-/* istanbul ignore file */
 import crypto from 'crypto';
 import { existsSync } from 'fs';
 import fs from 'fs/promises';
@@ -111,7 +110,10 @@ interface PrecinctScannerOptions {
   isFullExport?: boolean;
 }
 
-type ExportOptions = CentralScannerOptions | PrecinctScannerOptions;
+/**
+ * Options for {@link exportCastVoteRecordsToUsbDrive}
+ */
+export type ExportOptions = CentralScannerOptions | PrecinctScannerOptions;
 
 /**
  * A grouping of inputs needed by helpers throughout this file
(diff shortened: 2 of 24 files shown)


apps/scan/backend/src/app_config.test.ts
```diff
@@ -35,8 +35,8 @@ import {
 } from '@votingworks/backend';
 import { InsertedSmartCardAuthApi } from '@votingworks/auth';
 import { CVR, ElectionDefinition } from '@votingworks/types';
-import { configureApp } from '../../../test/helpers/shared_helpers';
-import { scanBallot, withApp } from '../../../test/helpers/custom_helpers';
+import { configureApp } from '../test/helpers/shared_helpers';
+import { scanBallot, withApp } from '../test/helpers/custom_helpers';
```
arsalansufi, line 39: Completely unrelated. Just had a hunch that `app_config.test.ts` wasn't intentionally located inside `apps/scan/backed/src/scanners/custom/`. The other tests in `apps/scan/backed/src/scanners/custom/` are all about scanning and prefixed with `app_scan_*`. Moved it up alongside other more general tests in `apps/scan/backend/src/`.

libs/usb-drive/src/mocks/memory_usb_drive.ts
```diff
@@ -74,6 +75,7 @@ export function createMockUsbDrive(): MockUsbDrive {
     removeUsbDrive() {
       mockUsbTmpDir?.removeCallback();
       mockUsbTmpDir = undefined;
+      usbDrive.status.reset();
```
arsalansufi, line 78: Without these resets, inserting and removing the mock USB drive continues growing the list of expected repeated calls. This results in unexpected behavior like the mock USB drive indicating that it's no drive even after calling `mockUsbDrive.insertUsbDrive`

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -839,14 +850,17 @@ export async function doesUsbDriveRequireCastVoteRecordSync(
...
     // A previous export operation may have failed midway
     if (areOrWereCastVoteRecordsBeingExportedToUsbDrive(usbDriveStatus)) {
       return true;
     }
 
     const { election, electionHash } = electionDefinition;
+    const exportDirectoryName = scannerStore.getExportDirectoryName();
+    if (!exportDirectoryName) {
+      return true;
+    }
```
arsalansufi, line 863: In case you're wondering why this check has moved location and gone from checking `false` to `true`, this check, at its previous location, was almost equivalent to the `ballotsCounted === 0` check (since we create the export directory after the first ballot is counted).

Now, by moving it after the `ballotsCounted === 0` case, we're catching the case that we fail to create the export directory (and specifically save its path in the DB) after the first ballot is counted. Should be exceedingly rare but, regardless, the new check is more useful than the previous one.

If you want to get extra specific, technically the `areOrWereCastVoteRecordsBeingExportedToUsbDrive(usbDriveStatus)` would likely also catch this. But that check involves a marker file on the USB drive. If the USB drive were replaced with a blank USB drive, that check wouldn't catch anything (other checks after would).

Zooming out, we're talking about the edgiest of edge cases here

adghayes (approved):

PR #4092, opened 2023-10-19
fix(hmpb): improve QR code detection

eventualbuddha:
## Overview
Changes our QR code detection code in Rust to be closer to our code in TypeScript. We now use multiple QR code readers, specifically `rqrr` (slightly faster) with fallback to `zbar` (better detection, but not complete overlap with `rqrr`). Additionally, we now only look at the parts of the image that we know might contain a QR code: the top-right and bottom-left corners. This speeds up the detection a lot, dropping from 60-150ms on my M1 MAX to 2-12ms.

The crop factor of 20% width × 20% height was not chosen terribly scientifically. I just deemed it a reasonable crop with enough error margin based on looking at a few scans.

## Todo
- [x] we need to update the CI docker container to include `libzbar-dev`

## Demo Video or Screenshot
n/a

## Testing Plan
- [x] new automated testing

We'll want to do some more manual testing, but I don't think we need to do that before this is merged.


libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -625,6 +596,22 @@ pub fn interpret_ballot_card(
     })
 }
 
+pub fn par_map_pair<T, F, R>(left: T, right: T, mapper: F) -> (R, R)
```
arsalansufi, line 563: What does the `par_` prefix here indicate? Parallel? If so, would you be open to calling this `parallel_map_pair`, `parallelized_map_pair`, `map_pair_parallelized`, etc.? Unless `par` is an established Rust convention, of course

.circleci/config.yml
```diff
@@ -9,13 +9,13 @@ orbs:
 executors:
   nodejs-browsers:
     docker:
-      - image: votingworks/cimg-debian11-browsers:2.0.4
+      - image: votingworks/cimg-debian11-browsers:2.0.5
```
arsalansufi, line 12: Do we need to update vxsuite-build-system at all to include `libzbar-dev`, e.g. maybe in this file? https://github.com/votingworks/vxsuite-build-system/blob/main/playbooks/install-vxsuite-packages.yaml

arsalansufi (approved):
LGTM! Thanks for digging into this and fixing!

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
eventualbuddha: It's not so much a Rust idiom as the convention used by the [`rayon`](https://crates.io/crates/rayon) crate (e.g. `par_iter`). I don't feel strongly about it, but I am inclined to leave it for this reason.

.circleci/config.yml
eventualbuddha: Unsure. I created https://github.com/votingworks/vxsuite-build-system/pull/85 just in case.

PR #4104, opened 2023-10-23
Fix manually created CVR fixture

arsalansufi:
While testing write-in adjudication, @adghayes identified an inconsistency in our one manually created CVR fixture. All the `WriteInImage` fields reference the front image, but two of the contests with write-ins are actually on the back of the ballot and should reference the back image.

Our actual CVR export logic is doing the right thing. The reason this fixture has an issue is because I recently modified it by hand to use the new CVR export format and made a human error 🙈.


adghayes (approved):
Thanks!!!

Issue #3848, opened 2023-08-18
Scanners: support tabulation of long paper lengths


mattroe:
specific length TBD based on hardware testing; most likely 22"

align with vxdesign lengths

eventualbuddha:
Tested scanning a 8.5x24" blank sheet with `libs/custom-scanner/bin/demo`'s autoscan feature. The scanner acted as normal and I got these images from it:
| Side A | Side B|
|-|-|
|![Image](https://github.com/votingworks/vxsuite/assets/1938/92ac8230-4b24-4754-aa3a-c119645e4585)|![Image](https://github.com/votingworks/vxsuite/assets/1938/ea67db22-912d-4817-9386-f4d354adc8cc)|

I'd like to find a way to print an actual 24" or 22" ballot that can be scanned, but this result indicates that we likely can scan 22" ballots with our current setup.

eventualbuddha:
Also tested with the PDI PageScan v5 using their `scan_demo` program. I had to make a change to the program to set the `PAGE_SIZE` tag on connecting to the scanner:

```cpp
pdiscan_error = PdSetTagLong(scanning_handle, PDISCAN_TAG_PAGE_SIZE, PD_PAGE_SIZE_MAXIMUM_SIZE);
```

Once I did that I was able to scan the same 8.5x24" sheet I used with the Custom scanner. It wasn't very clear whether the whole thing was scanned properly because of the bitonal scanning instead of grayscale, so I drew on the front of the sheet as you'll see in the images:

| Side A | Side B |
|-|-|
|![Image](https://github.com/votingworks/vxsuite/assets/1938/cd8a3f8a-e86e-472a-88a8-6817c1c352ba)|![Image](https://github.com/votingworks/vxsuite/assets/1938/6ed416db-171f-428a-aae5-40ab847e9c10)|

The vertical line is apparently just my scanning unit. I've tried cleaning the sensor glass but the line doesn't go away. It's unrelated to the long sheet.

PR #4108, opened 2023-10-24
Interpret longer ballot paper sizes

jonahkagan:
## Overview

Task: #3848 

Adds support for interpreting all of the VxDesign custom longer ballot paper sizes. Also updates ballot fixtures and tests to match.

_Review by commit_

## Demo Video or Screenshot
N/A

## Testing Plan
Updated automated tests. At some point, we'll also want to test some physical ballots, but we'll need to work with a print shop for that.

## Checklist


libs/ballot-interpreter/src/hmpb-rust/ballot_card.rs
```diff
@@ -175,15 +226,11 @@ impl PaperInfo {
...
-            },
-            BallotPaperSize::Legal => Size {
-                width: 34,
-                height: 53,
-            },
+        let columns_per_inch = 4.0;
+        let rows_per_inch = 4.0;
+        let grid_size = Size {
+            width: (columns_per_inch * width) as GridUnit,
+            height: (rows_per_inch * height) as GridUnit - 3,
```
jonahkagan, line 233: Same formula used in VxDesign

libs/hmpb/render-backend/src/generate_fixtures.ts
```diff
@@ -15,21 +15,29 @@ import {
...
+    `Generating: ${join(fixtureDir.replace(fixturesDir, ''), label)}`
+  );
   fs.writeFileSync(
     join(fixtureDir, `${label}-document.json`),
     JSON.stringify(document, null, 2)
   );
   const pdf = renderDocumentToPdf(document);
-  pdf.pipe(fs.createWriteStream(join(fixtureDir, `${label}.pdf`)));
+  const fileStream = fs.createWriteStream(join(fixtureDir, `${label}.pdf`));
+  pdf.pipe(fileStream);
```
eventualbuddha, line 33: Can you use [`pipeline` from `streams/promises`](https://nodejs.org/api/stream.html#streampipelinestreams-options)?

libs/ballot-interpreter/src/hmpb-rust/ballot_card.rs
eventualbuddha: Maybe put a comment here pointing to the other location and vice versa?

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -382,12 +383,12 @@ pub fn find_partial_timing_marks_from_candidate_rects(
         .collect::<Vec<Rect>>();
     let left_half_rects = rects
         .iter()
-        .filter(|r| r.left() < half_height)
+        .filter(|r| r.left() < half_width)
```
eventualbuddha, line 386: Whoops!

eventualbuddha (approved):
I think this looks right at a high level. I'm approving it without having done any manual testing myself. If you'd like me to take a deeper look at e.g. debug images to try to discover potential problems with real longer scans I can do so, likely tomorrow.

libs/hmpb/render-backend/src/generate_fixtures.ts
jonahkagan: It's a little weird in this case because you have to call `pdf.end()`. But I found `finished` in `streams/promises` which is an improvement over what I have

jonahkagan:
>  If you'd like me to take a deeper look at e.g. debug images to try to discover potential problems with real longer scans I can do so, likely tomorrow.

Sure, that would be great! I'll hold off on merging for now.

PR #4109, opened 2023-10-24
[libs/ui] Add UI String catalog build/extract steps

kofi-q:
## Overview

Adding a build step for extracting a catalog of UI Strings for translation, using `i18next-parser`.
- Copied and modified the starter config from https://github.com/i18next/i18next-parser#options
- Currently only extracting app strings, since election strings/translations will be populated based on user-provided election config input.
- Had to modify a few `appStrings` definitions to use dot-path keys instead of relying on the `uiStringSubKey` prop, since there was no out-of-the-box way to configure i18next-parser to handle the latter pattern.
- Added a stub export step (for only `main`-branch builds) to the circleci config - will fill that in once we decide on a shared location for libs/ui & VxDesign.

### Current i18next-parser Output:
```json
{
  "buttonAddWriteIn": "add write-in candidate",
  "buttonBack": "Back",
  "buttonDisplaySettings": "Color/Size",
  "buttonNext": "Next",
  "buttonReview": "Review",
  "buttonStartVoting": "Start Voting",
  "contestNavigationInstructions": "To navigate through the contest choices, use the down button. To move to the next contest, use the right button.",
  "labelAllPrecinctsSelection": "All Precincts",
  "labelBallotStyle": "Ballot style:",
  "labelNumBallotContests": "Number of contests on your ballot:",
  "numSeatsInstructions": {
    "1": "Vote for 1.",
    "2": "Vote for 2.",
    "3": "Vote for 3.",
    "4": "Vote for 4."
  },
  "numVotesRemaining": {
    "1": "You may select 1 more.",
    "2": "You may select 2 more.",
    "3": "You may select 3 more.",
    "4": "You may select 4 more."
  },
  "numVotesSelected": {
    "1": "You have selected 1.",
    "2": "You have selected 2.",
    "3": "You have selected 3.",
    "4": "You have selected 4."
  }
}

```

## Testing Plan
- Manual at the moment, but thinking through integration/e2e test strategy for later, once we have all the pieces in place

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1083,6 +1083,13 @@ jobs:
             JEST_JUNIT_OUTPUT_DIR: ./reports/
       - store_test_results:
           path: libs/ui/reports/
+      - when:
+          condition:
+            equal: [ main, << pipeline.git.branch >> ]
+          steps:
+            - run:
+                name: Export App Strings Catalog
+                command: echo 0 # TODO(kofi): export to versioned folder accessible from VxDesign
 
   # @votingworks/usb-drive
   test-libs-usb-drive:
--- a/libs/ui/src/ui_strings/app_strings.tsx
+++ b/libs/ui/src/ui_strings/app_strings.tsx
@@ -57,26 +57,10 @@ export const appStrings = {
     // better voter experience.
     // This pattern only makes sense when there's a very limited value space.
     ({
-      1: (
-        <UiString uiStringKey="numSeatsInstructions" uiStringSubKey="1">
-          Vote for 1.
-        </UiString>
-      ),
-      2: (
-        <UiString uiStringKey="numSeatsInstructions" uiStringSubKey="2">
-          Vote for 2.
-        </UiString>
-      ),
-      3: (
-        <UiString uiStringKey="numSeatsInstructions" uiStringSubKey="3">
-          Vote for 3.
-        </UiString>
-      ),
-      4: (
-        <UiString uiStringKey="numSeatsInstructions" uiStringSubKey="4">
-          Vote for 4.
-        </UiString>
-      ),
+      1: <UiString uiStringKey="numSeatsInstructions.1">Vote for 1.</UiString>,
+      2: <UiString uiStringKey="numSeatsInstructions.2">Vote for 2.</UiString>,
+      3: <UiString uiStringKey="numSeatsInstructions.3">Vote for 3.</UiString>,
+      4: <UiString uiStringKey="numSeatsInstructions.4">Vote for 4.</UiString>,
       // TODO(kofi): Find out what a reasonable upper limit is for the number of
       // possible votes per contest.
     })[numSeats],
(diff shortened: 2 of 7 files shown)


.circleci/config.yml
```diff
@@ -1083,6 +1083,13 @@ jobs:
             JEST_JUNIT_OUTPUT_DIR: ./reports/
       - store_test_results:
           path: libs/ui/reports/
+      - when:
+          condition:
+            equal: [ main, << pipeline.git.branch >> ]
+          steps:
+            - run:
+                name: Export App Strings Catalog
+                command: echo 0 # TODO(kofi): export to versioned folder accessible from VxDesign
```
arsalansufi, line 1092: Feel free to hold off on this for just a bit! I'm hashing out a few ideas in the spec that I'm writing, and I'm thinking of publishing to S3 and owning that as part of the VxDesign / multi-language data pipeline work.

Might also be worth a quick sync since we're reaching the touch points between our threads of work

arsalansufi (approved):
LGTM!

.circleci/config.yml
kofi-q: Perfect! I figured this would be a good data "hand-off" point for your export/import pipeline, so will leave this open for you to update once you've settled on the plan 🤝

libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -57,49 +57,33 @@ export const appStrings = {
...
+      4: <UiString uiStringKey="numSeatsInstructions.4">Vote for 4.</UiString>,
       // TODO(kofi): Find out what a reasonable upper limit is for the number of
       // possible votes per contest.
     })[numSeats],
 
   numVotesSelected: (numVotes: number) =>
     ({
       1: (
-        <UiString uiStringKey="numVotesSelected" uiStringSubKey="1">
+        <UiString uiStringKey="numVotesSelected.1">
```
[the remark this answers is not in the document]
kofi-q: It's now only used in the `election_strings.tsx` module, but maybe, for consistency, it's worth removing completely and adding a `buildStringKey()` utility or something for the election strings - will give that some thought and follow up with a PR

.circleci/config.yml
arsalansufi: Awesome! And we can sync as needed once I've shared the spec out and you've had a chance to read it!

libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -57,49 +57,33 @@ export const appStrings = {
...
-      ),
-      4: (
-        <UiString uiStringKey="numSeatsInstructions" uiStringSubKey="4">
-          Vote for 4.
-        </UiString>
-      ),
+      1: <UiString uiStringKey="numSeatsInstructions.1">Vote for 1.</UiString>,
+      2: <UiString uiStringKey="numSeatsInstructions.2">Vote for 2.</UiString>,
+      3: <UiString uiStringKey="numSeatsInstructions.3">Vote for 3.</UiString>,
+      4: <UiString uiStringKey="numSeatsInstructions.4">Vote for 4.</UiString>,
```
eventualbuddha, line 63: My reading of this is that a caller will get `undefined` for anything that isn't 1, 2, 3 or 4, so the "Vote for N." label will just be missing. Is that right? If so, shouldn't we have a fallback that does interpolation?

kofi-q: Yeah, was considering that, but want to avoid interpolation altogether where possible -- was going to see if we can reliably assert an upper limit, otherwise fall back to a non-interpolated string like "Number of votes available: x"

eventualbuddha: I'm sure there is an upper limit, but it might be an absurd number like 50. I've seen as high as 13 in my own experience as a voter.

kofi-q: Oh thanks, that's a lot more than I thought! I think I'll just reword these strings, then, as long as they don't sound too clunky

PR #4115, opened 2023-10-24
[Mark] Replace remaining StartPage strings with <UiString>s

kofi-q:
## Overview

- Removing the aria label for the "start voting" button (which was intended as spoken instructions for advancing to the first contest) and combining it with the general ballot navigation instructions text.
- Added an `<AudioOnly>` component to make the `'screen-reader-only'` class name assignment a little less brittle.
- Moved things around so that the screen reader only reads the election info and navigation instructions and not the button labels (until the voter focuses on those).

## Testing Plan
- Manual, since these were audio-only strings

## Checklist


arsalansufi (approved):
LGTM!

PR #4120, opened 2023-10-25
Button icon via props

jonahkagan:
## Overview

This separates out button icons from the concept of button variants. A few motivations for this:
- There were times where we wanted to change just the color but not the icon, which was starting to lead to a variant explosion (e.g. `previousPrimary` for when we wanted a previous icon with primary coloring).
- There are a lot of cases where we don't use a preset variant but still want to add an icon. This approach standardizes these cases with the variant cases.
- The problem that this is trying to solve (standardizing which icons get used in certain semantic button roles in the UI) can be better solved by having a simplified set of icons with clear names.

To implement this change, this PR:
- Changes the `Button` component to take an `icon` prop to add an icon on the left of the label. Also adds a `rightIcon` prop for more rarely used icons on the right (e.g. for a directional "next" button).
- Removes all of the variants that specify an icon
- Cleans up our icons a bit to make it easier to be consistent

_Review by commits_

## Demo Video or Screenshot
Shouldn't change much visually. The notable change is that some places that were previously using the "danger" variant (which used the "Danger" icon by default) now use the "Delete" icon (X in a circle), which I felt was more appropriate for "delete"/"remove" actions. The "Danger" icon is still in use for other dangerous actions.

## Testing Plan
Automated testing, plus some visual spot checks

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/manual_data_entry_screen.tsx
+++ b/apps/admin/frontend/src/screens/manual_data_entry_screen.tsx
@@ -721,6 +721,7 @@ export function ManualDataEntryScreen(): JSX.Element {
         );
       })}
       <P weight="semiBold">
+        <Icons.Warning />{' '}
         {someContestHasInvalidResults
           ? 'At least one contest above has invalid results entered'
           : someContestHasNoResults
@@ -729,14 +730,7 @@ export function ManualDataEntryScreen(): JSX.Element {
       </P>
       <P>
         <LinkButton to={routerPaths.manualDataSummary}>Cancel</LinkButton>{' '}
-        <Button
-          variant={
-            someContestHasInvalidResults || someContestHasNoResults
-              ? 'warning'
-              : 'primary'
-          }
-          onPress={saveResults}
-        >
+        <Button variant="primary" onPress={saveResults}>
           Save Results
         </Button>{' '}
       </P>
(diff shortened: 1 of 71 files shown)


apps/admin/frontend/src/screens/manual_data_entry_screen.tsx
```diff
@@ -721,6 +721,7 @@ export function ManualDataEntryScreen(): JSX.Element {
         );
       })}
       <P weight="semiBold">
+        <Icons.Warning />{' '}
```
adghayes, line 724: It looks like the logic has changed here to always show the warning icon - what's the thinking?

libs/ui/src/icons.tsx
```diff
@@ -68,10 +67,6 @@ function FaIcon(props: InnerProps): JSX.Element {
  */
 export const Icons = {
   Add(): JSX.Element {
-    return <FaIcon type={faSquarePlus} />;
-  },
-
-  AddCircle(): JSX.Element {
```
adghayes, line 75: :) glad my the circle icon I added made it to the big leagues

adghayes (approved):
one q about a logic tweak but otherwise LGTM, good cleanup

apps/admin/frontend/src/screens/manual_data_entry_screen.tsx
jonahkagan: That was a mistake, thanks for catching. Fixed: f6fdd87e3

PR #4121, opened 2023-10-25
Desktop color theme (first draft)

jonahkagan:
## Overview

Adds the new desktop color theme without applying it anywhere yet. This is just the groundwork for updating the components in libs/ui.

_Review by commit_

## Demo Video or Screenshot
I suggest opening up Storybook to see it yourself, but here's a tour.

https://github.com/votingworks/vxsuite/assets/530106/35709bdb-2aea-420b-b0b4-6a18e2edfb11

## Testing Plan
- Existing automated tests
- Checked contrast via Storybook (eventually I would like to write automated tests for contrast)

Note that this contrast checking did identify two inaccessible color combos in the Medium contrast theme (which is our voter-facing default). I plan to fix those in a follow up.

## Checklist


adghayes (approved):
LGTM. 

Only thing I was a little confused about when I opened up storybook was that contrast ratios are labelled, for example, as "VVSG 10:1" if it is _at least_ VVSG 10:1. I had to check the code because I thought maybe they were all 10:1 exactly. And for the "inaccessible" contrast (white on primary blue) in the medium contrast theme, I was curious what the current contrast ratio was but it doesn't show. 

Not sure if that's worth changing any behavior or storybook language for... just the thing I noticed!

jonahkagan:
Thanks for that note. I added a tooltip with the actual contrast ratio so you can see it when you hover over the contrast tag.

PR #4124, opened 2023-10-26
VxScan: remove "Official" label from reports

adghayes:
## Overview

As we standardize report naming in VxAdmin (see #4122), we [agreed](https://votingworks.slack.com/archives/CEL6D3GAD/p1698179698723679) that we should be careful with the word "Official" and it shouldn't appear on VxScan reports.

#### Before

Screen Shot 2023-10-25 at 5 08 05 PM

Screen Shot 2023-10-25 at 5 08 21 PM

#### After

Screen Shot 2023-10-25 at 5 08 53 PM

Screen Shot 2023-10-26 at 10 41 11 AM


kofi-q (approved):
Good catch!

PR #4125, opened 2023-10-26
Regenerate HMPB Fixtures Correctly

eventualbuddha:
## Overview
[@arsalansufi discovered](https://votingworks.slack.com/archives/CEL6D3GAD/p1698335612883039) that our HMPB "all bubbles" fixtures were out of date. This was due to the pre-commit hook that had `prettier` reformat the `election.json` files:

https://github.com/votingworks/vxsuite/blob/2af7a0f3c4b17c95cb6185d968dde00b026d85f3/libs/hmpb/render-backend/package.json#L28-L30

The fix is to:
1. format the fixtures in a "pretty" JSON format initially,
2. tell `prettier` not to reformat them, and
3. regenerate the fixtures again.

This PR does those things split across two commits, one for the manual changes and one to regenerate the fixtures.

## Demo Video or Screenshot
See the [Slack thread](https://votingworks.slack.com/archives/CEL6D3GAD/p1698335612883039), and this verification that the base64-encoded QR code in the cycling all bubbles ballot has the right election hash.

```
❯ sha256sum fixtures/all-bubble-ballot/election.json 
e99996c904a142d11289d816b24f0b0c6bb36f82e2763e18a8ca6ee576d15884  fixtures/all-bubble-ballot/election.json
❯ node
> bytes = Buffer.from('VlAC6ZmWyQShQtESiQEBAgMA', 'base64')
<Buffer 56 50 02 e9 99 96 c9 04 a1 42 d1 12 89 01 01 02 03 00>
> be=require('@votingworks/ballot-encoder')
…
> be.decodeElectionHash(bytes)
'e99996c904a142d11289'
```

## Testing Plan
Manually verified that the election hash of the cycling all bubbles ballot matches. I did not print the PDF and scan it, but I'm confident that would work.


arsalansufi (approved):
Thanks for looking into this and putting up a fix!

PR #4129, opened 2023-10-26
[i18n/audio] Install UiStringContextProvider in voter-facing apps

kofi-q:
## Overview

Installing the shared `UiStringContextProvider` in `Vx[Scan|Mark|Mark-Scan]` to connect `<UiString>` instances to the backend API.

Only enabling the context provider for dev/prod app instances (and integration tests) and leaving it disabled for the functional jest tests, since it was going to be too much unnecessary overhead to add mock versions of the UiString API endpoints and set up expectations in all existing tests that render the root `<App>` component directly.

## Testing Plan
- Unit test coverage for the added conditional provider installation
- Planning on following up later integration test coverage to verify e2e functionality from configuration through changing languages, once all the pieces are ready.

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/app.tsx
+++ b/apps/mark-scan/frontend/src/app.tsx
@@ -9,7 +9,14 @@ import {
 } from '@votingworks/utils';
 import { Logger, LogSource } from '@votingworks/logging';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { AppBase, ErrorBoundary, H1, P, Text } from '@votingworks/ui';
+import {
+  AppBase,
+  ErrorBoundary,
+  H1,
+  P,
+  Text,
+  UiStringsContextProvider,
+} from '@votingworks/ui';
 import { ColorMode, ScreenType, SizeMode } from '@votingworks/types';
 import { memoize } from './utils/memoize';
 import {
@@ -27,6 +34,7 @@ import {
   ApiClientContext,
   createApiClient,
   createQueryClient,
+  uiStringsApi,
 } from './api';
 import { SessionTimeLimitTracker } from './components/session_time_limit_tracker';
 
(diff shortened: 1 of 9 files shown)


apps/mark-scan/frontend/src/app.tsx
```diff
@@ -149,14 +159,19 @@ export function App({
...
-                  hardware={hardware}
-                  storage={storage}
-                  screenReader={screenReader}
-                  reload={reload}
-                  logger={logger}
-                />
-                <SessionTimeLimitTracker />
+                <UiStringsContextProvider
+                  api={uiStringsApi}
+                  disabled={!enableStringTranslation}
```
kofi-q, line 164: The positive-to-negative mapping here feels admittedly awkward - initially had a `disableStringTranslationForTesting` prop on the `<App>` component, but that would've meant requiring every test using `<App>` to remember to explicitly disable, vs explicitly enabling once in `index.tsx`.

arsalansufi: Ah yes the art of enable/disable params haha 🎨

After taking a sec to reason through the various possible permutations of params on `App` and `UiStringsContextProvider`, I like what you've landed on

libs/ui/src/ui_strings/ui_strings_context.tsx
```diff
@@ -7,13 +7,18 @@ import { AudioContextProvider } from './audio_context';
...
 }
 
 export function UiStringsContextProvider(
   props: UiStringsContextProviderProps
 ): JSX.Element {
-  const { api, children, noAudio } = props;
+  const { api, children, disabled, noAudio } = props;
+
+  if (disabled) {
+    return <React.Fragment>{children}</React.Fragment>;
```
arsalansufi, line 20: Could this be simplified down to `return children;`?

arsalansufi (approved):
LGTM!

libs/ui/src/ui_strings/ui_strings_context.tsx
kofi-q: Oh yeah, good call!

PR #4131, opened 2023-10-26
Update loading ellipsis animation to account for font change

arsalansufi:
## Overview

I've been spending lots of time looking at CVR import/export modals, and I recently noticed that the dots in the ellipsis animation are getting clipped. It just occurred to me that our recent font change 🤖 is probably responsible.

_Before_

https://github.com/votingworks/vxsuite/assets/12616928/81d9c958-da53-4b8e-a347-5aa9f27c31c7

_After :tada:_

https://github.com/votingworks/vxsuite/assets/12616928/3f15f8a5-2885-4c31-a7e7-f988908b357c

### DIFF
--- a/libs/ui/src/__snapshots__/loading.test.tsx.snap
+++ b/libs/ui/src/__snapshots__/loading.test.tsx.snap
@@ -2,7 +2,7 @@
 
 exports[`Renders Loading with defaults 1`] = `
 .c1 {
-  margin-left: -1.4em;
+  margin-left: -1em;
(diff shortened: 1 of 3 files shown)


kofi-q (approved):
Oh, good catch! Reminds me that I might've seen one or two other places in the past where we do some text-size-specific maths - will take a look and see if anything else needs changing

arsalansufi:
^ Ah yes hadn't considered other text-size-specific computations - thanks!

PR #4142, opened 2023-10-30
change ballot package location on USB

### LINKED ISSUE #3935: Export files to USB in consistent directory structure
On creation of this issue, we export files to a bunch of different places on USB drives:

- VxAdmin: ballot packages are exported to `/ballot-packages`
- VxScan + VxCentralScan: cast vote records are exported to `/cast-vote-records/election-identifier`
- VxAdmin: many documents are just exported to `/`

We've [decided](https://votingworks.slack.com/archives/CEL6D3GAD/p1693940716686119) that, where possible, we want to stick to a consistent structure in the format: `/election-identifier/[file-type]/[filename]`. This allows easy copying and deletion of most files for an election. Most importantly, it provides a consistent UX.

The following places need to use the new format:

- [x] VxScan + VxCentralScan - cast vote record exports
- [x] ~~VxScan + VxCentralScan - backups?~~
- [ ] VxAdmin - ballot package export (assuming this continues to exist)
- [x] VxAdmin - `.csv` exports
- [x] VxAdmin - `.pdf` exports

adghayes:
## Overview

Closes #3935. Previously, ballot packages would be exported to a USB drive like so:

```./ballot-packages/sample-county_example-primary-election_06414087b1__2023-10-30_12-57-46.zip```

Now, we export them like this:

```./sample-county_example-primary-election_06414087b1/ballot-packages/ballot-package__2023-10-30_12-56-55.zip```

This means that all of our in-system files are being exported in the same way, to a top-level election directory, a file type subdirectory, then a filename. I kept `ballot-package` in the file name for the ballot package so that it wouldn't be _just_ a date if it got out of context. Could add file hash in too if desirable.

## Testing Plan
- updated automated testing, added a test
- manually tested exporting from VxAdmin and loading into VxMark

### DIFF
--- a/apps/scan/backend/src/app_config.test.ts
+++ b/apps/scan/backend/src/app_config.test.ts
@@ -13,7 +13,6 @@ import {
   singlePrecinctSelectionFor,
 } from '@votingworks/utils';
 import {
-  assert,
   assertDefined,
   err,
   find,
@@ -21,16 +20,14 @@ import {
   ok,
   unique,
 } from '@votingworks/basics';
-import fs from 'fs';
-import { join } from 'path';
 import {
   fakeElectionManagerUser,
   fakeSessionExpiresAt,
   mockOf,
 } from '@votingworks/test-utils';
 import {
-  createBallotPackageZipArchive,
   getCastVoteRecordExportDirectoryPaths,
+  mockBallotPackageFileTree,
   readCastVoteRecordExport,
 } from '@votingworks/backend';
 import { InsertedSmartCardAuthApi } from '@votingworks/auth';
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: apps/scan/backend/src/app_config.test.ts [hunk 1]
```diff
@@ -168,46 +161,6 @@ test("if there's only one precinct in the election, it's selected automatically
   });
 });
 
-test('configures using the most recently created ballot package on the usb drive', async () => {
```
adghayes, line 171: this is tested in `libs/backend` and doesn't need to be tested here too

adghayes:
@arsalansufi or @kshen0 tagging either of you for review as I've discussed this change with both of you on Slack.

libs/utils/src/filenames.ts
```diff
@@ -108,88 +59,6 @@ export function generateElectionBasedSubfolderName(
   )}`;
 }
 
-/**
- * Generate the directory name for the cast vote record report.
- *
- * @deprecated
```
arsalansufi, line 114: Thanks for removing these! Forgot about them when I was cleaning up old CVR code

arsalansufi (approved):
LGTM!

PR #4144, opened 2023-10-30
Button cleanup

jonahkagan:
## Overview

An assortment of small Button component cleanup chores that will make it easier to implement the upcoming desktop theme. Details in the commits!

## Demo Video or Screenshot
N/A 

## Testing Plan
Automated tests plus some manual testing of VxMark

## Checklist


libs/mark-flow-ui/src/components/review.tsx
```diff
@@ -210,7 +204,7 @@ export function Review({
             footerAlign={selectionsAreEditable ? 'right' : undefined}
             footer={
               selectionsAreEditable && (
-                <DecoyButton>
+                <Button onPress={() => {}}>
```
adghayes, line 207: Based on your commit description, should you be passing a `tabIndex` here?

libs/mark-flow-ui/src/pages/start_page.tsx
```diff
@@ -52,12 +50,6 @@ const Footer = styled.div`
   padding: 0.5rem;
 `;
 
-const LargeButtonText = styled.span`
-  font-size: 50px;
-  line-height: 2;
-  padding: 0 1rem;
-`;
-
```
adghayes, line 60: I am confused by the styling change here:

1. Was this button really a set `px` before, and thus unresponsive to size settings?
2. From my reading, it looks like you made it responsive but a lot smaller, from `50px` to `1.2rem`? Is that right?

libs/ui/src/button.tsx
```diff
@@ -28,15 +28,6 @@ export interface StyledButtonProps {
...
   rightIcon?: IconName | JSX.Element;
-
-  /** @deprecated Place the button within a flex or grid container instead. */
-  readonly fullWidth?: boolean;
-
-  /** @deprecated Incompatible with new VVSG size themes */
-  readonly large?: boolean;
-
-  /** @deprecated Incompatible with new VVSG size themes */
-  readonly small?: boolean;
```
adghayes, line 39: These were truly not doing anything at all? 

😮 

Scandalous.

libs/mark-flow-ui/src/components/review.tsx
jonahkagan: Great catch, I think it got removed accidentally in a rebase

libs/mark-flow-ui/src/pages/start_page.tsx
jonahkagan: 1. Yes, as far as I can tell it was a set px before.
2. The font size with the default medium size theme now works out to 48px, so it's pretty much in the same range.

libs/ui/src/button.tsx
jonahkagan: I think they were doing things long ago, but were deprecated and then unimplemented at some point along the way

adghayes (approved):

PR #4150, opened 2023-10-31
Voter Privacy Warming / Tally Report Viewer Refactor

adghayes:
## Checklist


adghayes:
@mattroe can you take a look at these two videos and give me some feedback?

The ballot report builder is getting kinda complex. I reduced a little header spacing above "Filters" and "Report By" to try to compact the "input" area of the page, vs. the "action" area of the page. I'm also making "Generate Report" artificially wide so that it seems separate from the other buttons.

This will get easier to style with the options introduced by @jonahkagan in his desktop theme. The "Generate Report" button can be primary, "Print Report" can be secondary, and the disabled styling will be less jarring.

https://github.com/votingworks/vxsuite/assets/37960853/9a7e686f-e687-473a-8394-8c58c9ffd6c6


https://github.com/votingworks/vxsuite/assets/37960853/d8ba7283-ad32-4f27-9277-f99d2a01a754

mattroe:
Overall experience LGTM. I do agree that we should consider this screen as @jonahkagan implements the new design system to simplify / clarify the button state in particular.

I know that I originally suggested we keep this, but now that this screen is getting cluttered, I think we should consider removing the pagination notice. It ultimately doesn't provide that much value. If we make that change, it could visually make more sense to put the warnings under all the print/export buttons?

kofi-q (approved):
LGTM - +1 to the "generate preview before print/export" pattern being a better UX

PR #4153, opened 2023-11-01
Switch status alerts from colored text to colored icons

jonahkagan:
## Overview

In the new desktop theme, our accent colors for status alerts (success/warning/danger) only meet the [3:1 contrast ratio minimum](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast), which is only permitted for icons/graphical elements, not text. I did it this way in order to have a bright yellow/orange color for warnings that isn't super dark and weird looking. Thus, we can't have text with these colors.

I think this is a fine constraint, because we rarely use colored text, and it's still effective to have a colored icon to grab attention accompanied by text with the default text color.

To implement this new approach, this PR:
- Adds a `color` prop to the `Icon` components
- Removes the `color` prop from the `Font` components
- Updates all the status message use cases accordingly

_Review by commit_

## Demo Video or Screenshot
A few sample changes from VxMark
Before
Screenshot 2023-11-01 at 10 12 17 AM
After
Screenshot 2023-11-01 at 10 21 55 AM

Before
Screenshot 2023-11-01 at 10 13 48 AM
After
Screenshot 2023-11-01 at 10 21 09 AM

## Testing Plan
- Automated tests
- Visual inspection of most of the changes to make sure they are still legible

## Checklist

### DIFF
--- a/apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
+++ b/apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
@@ -14,6 +14,7 @@ import {
   WithScrollButtons,
   P,
   Font,
+  Icons,
 } from '@votingworks/ui';
 import {
   format,
@@ -394,13 +395,13 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element | null {
           <td data-testid="cvr-count">{format.count(cvrCount)}</td>
           {!fileModeLocked && (
             <td>
-              <LabelText>
-                {isTestModeResults ? (
-                  <Font color="warning">Test</Font>
-                ) : (
-                  <Font color="success">Official</Font>
-                )}
-              </LabelText>
+              {isTestModeResults ? (
+                <LabelText>
+                  <Icons.Warning color="warning" /> Test
+                </LabelText>
+              ) : (
+                <LabelText>Official</LabelText>
+              )}
             </td>
           )}
           <TD textAlign="right">
--- a/apps/admin/frontend/src/components/smartcard_modal/status_message.tsx
+++ b/apps/admin/frontend/src/components/smartcard_modal/status_message.tsx
@@ -78,17 +78,15 @@ export function SuccessOrErrorStatusMessage({
     }
     return (
       <P>
-        <Font color="danger">
-          <Icons.Danger />
-        </Font>{' '}
-        {text} Please try again.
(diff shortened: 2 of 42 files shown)


apps/admin/frontend/src/components/import_cvrfiles_modal.tsx
```diff
@@ -394,13 +395,13 @@ export function ImportCvrFilesModal({ onClose }: Props): JSX.Element | null {
...
-                  <Font color="success">Official</Font>
-                )}
-              </LabelText>
+              {isTestModeResults ? (
+                <LabelText>
+                  <Icons.Warning color="warning" /> Test
+                </LabelText>
+              ) : (
+                <LabelText>Official</LabelText>
+              )}
```
jonahkagan, line 404: Screenshot 2023-11-01 at 10 28 58 AM

apps/admin/frontend/src/components/smartcard_modal/status_message.tsx
```diff
@@ -103,16 +101,18 @@ export function SuccessOrErrorStatusMessage({
   if (action === 'PinReset') {
     assert(newPin !== undefined);
     return (
-      <TextLarge color="success">
-        New card PIN is <Font weight="bold">{hyphenatePin(newPin)}</Font>.
+      <TextLarge>
+        <Icons.Done color="success" /> New card PIN is{' '}
+        <Font weight="bold">{hyphenatePin(newPin)}</Font>.
```
jonahkagan, line 106: Screenshot 2023-11-01 at 10 29 50 AM

adghayes (approved):
Looks better this way!

Is the "warning" red instead of orange just to draw attention / increase contrast? I'm very used to the warnings being orange, and now the `warning` vs. `danger` colors are exactly the same for icons.

jonahkagan:
> Is the "warning" red instead of orange just to draw attention / increase contrast? I'm very used to the warnings being orange, and now the `warning` vs. `danger` colors are exactly the same for icons.

Oh, I got confused. I thought it was already red on `main`, but it turns out I made that change in an earlier PR (https://github.com/votingworks/vxsuite/commit/cdde8fcd5cd681a9696ebdce209bf6ce4b47ac46#diff-01f30c555c79c11331b594d7ca37e59459b212f73b76186e19e83693d080776cR247). I'm going to switch it back to orange, which I agree is the correct choice.

PR #4160, opened 2023-11-02
Desktop UI components

jonahkagan:
## Overview

Converts the rest of the components in libs/ui to either work with the desktop theme or at least use the new color/size theme fields. Most of these changes involved only minor visual differences, so I grouped them all in one PR.

Also adds spacing guidelines and generally tries to implement those for spacing consistency across components.

_Review by commit_

## Demo Video or Screenshot
Included in relevant commits

## Testing Plan
Existing automated tests plus manual visual testing

## Checklist


libs/ui/src/themes/spacing.stories.tsx
jonahkagan, line 1: Screenshot 2023-11-02 at 10 51 29 AM
Screenshot 2023-11-02 at 10 51 42 AM

libs/ui/src/__snapshots__/button_bar.test.tsx.snap
jonahkagan, line 1: Screenshot 2023-11-02 at 10 52 57 AM

libs/ui/src/modal.tsx
jonahkagan, line 1: Screenshot 2023-11-02 at 10 55 17 AM

adghayes (approved):
LGTM!

We now have some components that are meant for both desktop and touch, and some that are only for touch. I don't know how much of a problem it would be in practice, but I wish there were some way we clearly distinguished which components should be used where.

jonahkagan:
> We now have some components that are meant for both desktop and touch, and some that are only for touch.

Yeah it's a good point. I'll think about this. Let me know if you have ideas

adghayes:
@jonahkagan I wonder if we could have a custom, required flag in our storybook types, so if we add something to storybook we have to indicate whether it's one or the other or both, and it could appear somewhere in the storybook interface.

jonahkagan:
@adghayes something like that sounds reasonable

PR #4162, opened 2023-11-02
Add scaffolding for VxDesign background task processing

arsalansufi:
_Easiest reviewed by commit_

## Overview

This PR adds scaffolding for VxDesign background task processing, something that we need for cloud translation and audio clip generation but that could also prove useful for ballot generation, which is likely to become more time consuming as we build additional VxDesign functionality.

Everything is running locally, but I'm designing things to make the eventual transition to a Heroku deployment as easy as possible. I'm accordingly 1) running the background worker in its own process separate from the backend server process and 2) mimicking the DB schema that Arlo uses for background task processing.

## Demo Video

Here, I'm having the election-creation endpoint kick off a no-op job:

https://github.com/votingworks/vxsuite/assets/12616928/7c09d907-b680-4c3d-bfef-65bd5a3d74dd

## Testing Plan

- [x] Added unit tests for store methods
- [x] Manually tested end-to-end

### DIFF
--- a/apps/design/backend/package.json
+++ b/apps/design/backend/package.json
@@ -17,7 +17,7 @@
     "lint": "pnpm type-check && eslint .",
     "lint:fix": "pnpm type-check && eslint . --fix",
     "pre-commit": "lint-staged",
-    "start": "node ./build/index.js",
+    "start": "node ./build/index.js & node ./build/worker/index.js",
     "test": "is-ci test:ci test:watch",
     "test:ci": "jest --coverage --reporters=default --reporters=jest-junit --maxWorkers=6",
     "test:coverage": "jest --coverage",
--- a/apps/design/backend/schema.sql
+++ b/apps/design/backend/schema.sql
@@ -6,3 +6,13 @@ create table elections (
   layout_options_data text not null,
   created_at timestamp not null default current_timestamp
 );
+
+create table background_tasks (
+  id text primary key,
+  task_name text not null,
+  payload text not null,
+  created_at timestamp not null default current_timestamp,
+  started_at timestamp,
+  completed_at timestamp,
+  error text
+);
--- a/apps/design/backend/src/store.ts
+++ b/apps/design/backend/src/store.ts
@@ -412,13 +412,13 @@ export class Store {
   // Background task processing
   //
 
-  getQueuedBackgroundTasks(): BackgroundTask[] {
+  getOldestQueuedBackgroundTask(): Optional<BackgroundTask> {
     const sql = `${getBackgroundTasksBaseQuery}
       where started_at is null
-      order by created_at asc
+      order by created_at asc limit 1
     `;
-    const rows = this.client.all(sql) as BackgroundTaskRow[];
-    return rows.map(backgroundTaskRowToBackgroundTask);
+    const row = this.client.one(sql) as Optional<BackgroundTaskRow>;
+    return row ? backgroundTaskRowToBackgroundTask(row) : undefined;
   }
 
   getBackgroundTask(taskId: Id): Optional<BackgroundTask> {
--- a/apps/design/backend/src/store.ts
+++ b/apps/design/backend/src/store.ts
@@ -1,4 +1,5 @@
 import {
+  Optional,
   assert,
   groupBy,
   throwIllegalValue,
@@ -23,6 +24,7 @@ import {
   PartyId,
 } from '@votingworks/types';
 import { join } from 'path';
+import { v4 as uuid } from 'uuid';
 
 export interface ElectionRecord {
   id: Id;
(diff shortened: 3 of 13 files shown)


apps/design/backend/schema.sql
```diff
@@ -5,4 +5,14 @@ create table elections (
...
+
+create table background_tasks (
+  id text primary key,
+  task_name text not null,
+  payload text not null,
+  created_at timestamp not null default current_timestamp,
+  started_at timestamp,
+  completed_at timestamp,
+  error text
+);
```
arsalansufi, line 18: As noted in the PR description, I decided to mimic the DB schema that Arlo uses for background task processing: https://github.com/votingworks/arlo/blob/7a2689cf938efcf8e2733e2d8732eadfc020be48/server/models.py#L977-L988

I didn't include fields for progress tracking yet. Can add those when the need arises.

I also considered foregoing `task_name` and incorporating it into the `payload`, but decided to keep the separate field since being able to query on specific `task_name` values feels useful.

apps/design/backend/src/store.ts
```diff
@@ -74,6 +76,52 @@ export interface BallotStyle {
   partyId?: PartyId;
 }
 
+export interface BackgroundTask {
+  id: Id;
+  taskName: string;
```
arsalansufi, line 81: Will likely make this a stricter union of string literals once the kinds of background tasks become clearer

apps/design/backend/src/store.ts
```diff
@@ -227,6 +275,13 @@ export class Store {
     return new Store(DbClient.fileClient(dbPath, SchemaPath));
   }
 
+  /**
+   * Builds and returns a new store whose data is kept in memory.
+   */
+  static memoryStore(): Store {
+    return new Store(DbClient.memoryClient(SchemaPath));
+  }
```
arsalansufi, line 283: Added this for tests (next commit)

apps/design/backend/package.json
```diff
@@ -17,7 +17,7 @@
     "lint": "pnpm type-check && eslint .",
     "lint:fix": "pnpm type-check && eslint . --fix",
     "pre-commit": "lint-staged",
-    "start": "node ./build/index.js",
+    "start": "node ./build/index.js & node ./build/worker/index.js",
```
arsalansufi, line 20: Did some testing to verify that ctrl-c closes both processes. Presumably handled by pnpm

apps/design/backend/src/worker/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { WORKSPACE } from '../globals';
+import { Store } from '../store';
+import * as worker from './worker';
+
+async function main(): Promise<void> {
+  const workspacePath = path.resolve(assertDefined(WORKSPACE));
+  ensureDirSync(workspacePath);
+  const dbPath = path.join(workspacePath, 'design-backend.db');
+  const store = Store.fileStore(dbPath);
```
arsalansufi, line 13: As far as I can tell, this is the first case of two processes accessing the same SQLite DB in VxSuite. I did verify that this is okay:

> Multiple processes can have the same database open at the same time. Multiple processes can be doing a SELECT at the same time. But only one process can be making changes to the database at any moment in time, however.
>
> Reference: https://www.sqlite.org/faq.html

We may run into some quirks with simultaneous writes by the server and the worker, but I wasn't able to trigger any such errors with my manual testing. I _believe_ `better-sqlite3` automatically retries when it finds that the DB is busy. Regardless, when we deploy using Heroku, we'll be using Postgres, rendering most of these concurrency concerns irrelevant.

script/src/run-dev/index.ts
```diff
@@ -169,7 +169,9 @@ export async function main(
             npmBinCommand({
               name: `${name}:run`,
               command:
-                `while [ ! -f build/index.js ]; do echo "Waiting for build…"; sleep 1; done; nodemon --watch build --delay 1 --exitcrash --exec NODE_ENV=development VX_MACHINE_TYPE=${frontend} node build/index.js`,
+                'while [ ! -f build/index.js ]; do echo "Waiting for build…"; sleep 1; done; ' +
+                'nodemon --watch build --delay 1 --exitcrash --exec ' +
+                'NODE_ENV=development pnpm start',
```
arsalansufi, line 174: I dropped the `VX_MACHINE_TYPE=${frontend}`  here in favor of the provided abstraction for env vars:

https://github.com/votingworks/vxsuite/blob/44baf7fcb5f97ffab5e3584eb977e57b73d369ac/script/src/run-dev/index.ts#L152

-----

I also switched from `node build/index.js` as the command for frontends' dependent services to `pnpm start` so that we have some more control. Every backend's `pnpm start` command is `node ./build/index.js`, with the exception of VxDesign, which uses `node ./build/index.js & node ./build/worker/index.js`

Zooming out, this just makes sure that the background worker is properly started when we run `pnpm start` inside `apps/design/frontend`

arsalansufi:
@kofi-q This can of course wait till next week since I know it's close to the end of your week!

apps/design/backend/src/store.ts
```diff
@@ -352,4 +407,73 @@ export class Store {
...
+    if (error) {
+      this.client.run(
+        `
+        update background_tasks
+        set completed_at = current_timestamp, error = ?
+        where id = ?
+        `,
+        error,
+        taskId
+      );
```
jonahkagan, line 467: What happens if you run this case when `error` is undefined? Wondering if it would just work

apps/design/backend/src/worker/worker.ts
```diff
@@ -0,0 +1,46 @@
...
+): Promise<void> {
+  // Simulate a long-running operation
+  // TODO(arsalan): Implement actual processing logic
+  await sleep(3000);
+}
+
+async function processBackgroundTasks(store: Store): Promise<void> {
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    const queuedTasks = store.getQueuedBackgroundTasks();
```
jonahkagan, line 17: Nit: Since we only ever want one task at a time, maybe we just query for a single task here?

jonahkagan (approved):
Exciting to see this start to come together!!

apps/design/backend/src/store.ts
arsalansufi: Ah good catch! Can indeed consolidate (with a conversion of `undefined` to `null`): https://github.com/votingworks/vxsuite/pull/4162/commits/d26a754961434311ec1e46058f53499b6a3b24be

apps/design/backend/src/worker/worker.ts
arsalansufi: 👍 I like that: https://github.com/votingworks/vxsuite/pull/4162/commits/89d6761bdf1f0ecefaabf9024db90512887facf3

PR #4165, opened 2023-11-03
UWI Plumbing: Separate UWIs in Legacy Adapter

adghayes:
## Overview

First of a series of PRs to expose unmarked write-ins, from the interpreter to VxAdmin, represented separately from marks. 

## Testing Plan

All automated.

### DIFF
--- a/libs/ballot-interpreter/src/interpret.ts
+++ b/libs/ballot-interpreter/src/interpret.ts
@@ -2,7 +2,6 @@ import { sliceElectionHash } from '@votingworks/ballot-encoder';
 import { Result, throwIllegalValue, typedAs } from '@votingworks/basics';
 import { loadImageData } from '@votingworks/image-utils';
 import {
-  AdjudicationReason,
   BallotMetadata,
   BallotType,
   ElectionDefinition,
@@ -10,7 +9,6 @@ import {
   InvalidPrecinctPage,
   InvalidTestModePage,
   mapSheet,
-  MarkThresholds,
   PageInterpretation,
   PageInterpretationWithFiles,
   PrecinctSelection,
@@ -22,6 +20,7 @@ import makeDebug from 'debug';
 import { interpret as interpretNhHmpbBallotSheet } from './hmpb-ts';
 import { interpret as interpretVxBmdBallotSheet } from './bmd';
 import { convertNhInterpretResultToLegacyResult } from './legacy_adapter';
+import { InterpreterOptions, getScoreWriteInsFlag } from './options';
 
 const debug = makeDebug('ballot-interpreter:scan:interpreter');
 
@@ -36,17 +35,6 @@ export type InterpretResult = Result<SheetOf<InterpretFileResult>, Error>;
 export interface InterpretFileResult {
   interpretation: PageInterpretation;
   normalizedImage: ImageData;
-}
-
-/**
- * Options for interpreting a sheet of ballot images.
- */
-export interface InterpreterOptions {
-  electionDefinition: ElectionDefinition;
-  precinctSelection: PrecinctSelection;
-  testMode: boolean;
-  markThresholds: MarkThresholds;
-  adjudicationReasons: readonly AdjudicationReason[];
 }
 
 /**
@@ -135,7 +123,7 @@ function interpretAndConvertNhHmpbResult(
   options: InterpreterOptions
 ): SheetOf<InterpretFileResult> {
   const result = interpretNhHmpbBallotSheet(electionDefinition, sheet, {
-    scoreWriteIns: options.markThresholds.writeInTextArea !== undefined,
+    scoreWriteIns: getScoreWriteInsFlag(options),
   });
 
   return validateInterpretResults(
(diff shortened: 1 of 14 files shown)


libs/ballot-interpreter/src/interpret.ts
```diff
@@ -38,17 +37,6 @@ export interface InterpretFileResult {
...
- * Options for interpreting a sheet of ballot images.
- */
-export interface InterpreterOptions {
-  electionDefinition: ElectionDefinition;
-  precinctSelection: PrecinctSelection;
-  testMode: boolean;
-  markThresholds: MarkThresholds;
-  adjudicationReasons: readonly AdjudicationReason[];
-}
-
```
adghayes, line 51: Had to pull this out to shared file in order to share a dependent helper.

libs/ballot-interpreter/src/options.ts
```diff
@@ -0,0 +1,23 @@
...
+  testMode: boolean;
+  markThresholds: MarkThresholds;
+  adjudicationReasons: readonly AdjudicationReason[];
+}
+
+export function getScoreWriteInsFlag(options: InterpreterOptions): boolean {
+  return options.adjudicationReasons.includes(
+    AdjudicationReason.UnmarkedWriteIn
+  );
+}
```
adghayes, line 23: Scoring write-ins is now based off of the adjudication flag.

libs/types/src/interpretation.ts
```diff
@@ -45,11 +49,24 @@ export const InterpretedBmdPageSchema: z.ZodSchema<InterpretedBmdPage> =
...
+  optionId: WriteInIdSchema,
+  score: z.number(),
+});
+
 export interface InterpretedHmpbPage {
   type: 'InterpretedHmpbPage';
   ballotId?: BallotId;
   metadata: HmpbBallotPageMetadata;
   markInfo: MarkInfo;
+  unmarkedWriteIns?: UnmarkedWriteIn[];
```
adghayes, line 69: I'm not quite sure what the best way to represent them here is.

The way we represent marks is currently confusing - we keep both the raw mark data and the interpreted mark data (votes) in the return data of the interpreter.

We could do the same for `unmarkedWriteIns` - an `unmarkedWriteInInfo` and a `unmarkedPossibleWriteInVotes`. I'm not sure which direction we're heading at this layer - exposing more detail or exposing less. I'm also not sure if changes to the interpreter re: parameter-izable write-in areas will require more information to be passed here, or if all that necessary information will be caught in the layout.

Anyway, welcoming feedback on this data representation.

libs/utils/src/hmpb/ballot_adjudication_reasons.ts
```diff
@@ -1,46 +1,50 @@
...
 import { throwIllegalValue } from '@votingworks/basics';
 import { allContestOptions } from './all_contest_options';
 
 export interface Options {
-  optionMarkStatus: (option: ContestOption) => MarkStatus;
+  optionStatus: (option: ContestOption) => {
+    markStatus: MarkStatus;
+    writeInAreaStatus: WriteInAreaStatus;
+  };
 }
```
adghayes, line 18: In this PR, I maintain the same layer of abstraction we currently have - this method returns _all_ possible adjudication reasons, and the consumer decides which ones are important. 

But I'm thinking like we probably will need to change how this works, because the "all possible adjudication reasons" is ostensibly different from state to state, and the combinations and interactions may be too complicated. 

For now, it works because if we're ignoring marked write-ins then the write-in area status is always `Ignored` and it doesn't factor in to overvote adjudication.

I want to commit this if it makes sense, because of time limits, but as adjudication rules get more complicated may want to rethink layers here.

libs/utils/src/hmpb/ballot_adjudication_reasons.ts
```diff
@@ -1,46 +1,50 @@
...
 
     for (const contest of contests) {
-      const selectedOptionsByContestId = new Map<
-        Contest['id'],
-        Array<{ id: ContestOption['id']; index: number }>
-      >();
+      const selectedContestOptions: Array<{
+        id: ContestOption['id'];
+        index: number;
+      }> = [];
```
adghayes, line 38: There was some useless code here, where we created a dictionary on every loop that only ever got one value in it. So I cleaned it up to be simpler.

libs/ballot-interpreter/src/interpret_nh_hmpb.test.ts
```diff
@@ -222,16 +234,20 @@ describe('HMPB - m17 backup', () => {
...
       );
 
       const interpretation = mapSheet(interpretationWithImages, (page) => {
         // Only snapshot stable fields - marks and layout coordinates may change slightly on CI
         if (page.interpretation.type === 'InterpretedHmpbPage') {
-          const { type, metadata, votes } = page.interpretation;
-          return { type, metadata, votes };
+          const { type, metadata, votes, unmarkedWriteIns } =
+            page.interpretation;
+          return { type, metadata, votes, unmarkedWriteIns };
```
adghayes, line 250: Arguably problematic that we're including unmarkedWriteIns unedited in the snapshot because it includes scores, but the scores are rounded to two decimal places so I think they should be consistent on CI.

adghayes:
Now just merging a PR that includes these changes instead.

Issue #4128, opened 2023-10-26
Ensure that VxCentralScan inactive session time limit doesn't interfere with CVR exports


arsalansufi:
While the performance of CVR exports and imports has improved after switching to the new multiple-file format (I was able to export 4000 cycled all-bubble ballots in 12 mins 55 secs on my Mac VM), things do take longer on production hardware. And there are consequences to the inactive session time limit (30 mins by default) logging you out while exporting CVRs. When you log back in, the CVR export continues, but in the background without any UI indication, bogging down the UI.

One option is to disable the inactive session time limit while exporting CVRs, backups, etc. Another is to introduce higher options for the time limit (configured in system settings) and use one of those higher settings for SLI testing. If we take the latter approach (which is definitely easier / less clunky than the former), we should double check that VVSG doesn't place hard restrictions on allowed options for the inactive session time limit.

PR #4167, opened 2023-11-06
Allow increasing inactive session time limit for SLI testing

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4128

A short-term precautionary measure to make sure that SLI doesn't hit the inactive session time limit when exporting large numbers of CVRs on VxCentralScan. We'll need to remember to set the higher time limit in the system settings file in the election package that we provide them, which I'll open an issue to track. Will also open an issue to track the longer-term fix of pausing the inactive session time limit timer during export operations.

### DIFF
--- a/libs/types/src/auth/auth.ts
+++ b/libs/types/src/auth/auth.ts
@@ -79,14 +79,17 @@ export const DEFAULT_STARTING_CARD_LOCKOUT_DURATION_SECONDS: StartingCardLockout
  * The inactive/idle session time limit, after which the user must reauthenticate - a VVSG2
  * requirement
  */
-export type InactiveSessionTimeLimitMinutes = 10 | 15 | 20 | 25 | 30;
+export type InactiveSessionTimeLimitMinutes = 10 | 15 | 20 | 30 | 360;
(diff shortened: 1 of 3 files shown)


libs/types/src/auth/auth.ts
```diff
@@ -79,14 +79,17 @@ export const DEFAULT_STARTING_CARD_LOCKOUT_DURATION_SECONDS: StartingCardLockout
...
  * requirement
  */
-export type InactiveSessionTimeLimitMinutes = 10 | 15 | 20 | 25 | 30;
+export type InactiveSessionTimeLimitMinutes = 10 | 15 | 20 | 30 | 360;
 export const InactiveSessionTimeLimitMinutesSchema: z.ZodSchema<InactiveSessionTimeLimitMinutes> =
   z.union([
     z.literal(10),
     z.literal(15),
     z.literal(20),
-    z.literal(25),
```
arsalansufi, line 88: `25` felt like a somewhat random option to me, so I just dropped it hah.

I was thinking of dropping the stricter validation altogether such that the data model supports any positive integer value. But the validation feels helpful for now, since we're editing system settings files manually and not through a UI (i.e. VxDesign).

jonahkagan (approved):

PR #4171, opened 2023-11-06
Add Google Cloud Translation and Text-to-Speech API integrations to VxDesign

arsalansufi:
_Easiest reviewed by commit_

## Overview

This PR adds Google Cloud Translation and Text-to-Speech API integrations to VxDesign, in preparation for language and audio file generation. I've gone ahead and added caching to minimize API calls and, by extension, costs.

I've also done some thinking around Google Cloud auth, for both deployed and local environments. I'll add some notes to the VxDesign README in a follow-up PR, but the tldr is that I'm planning for us to use a service account in deployed environments and service account impersonation in local environments (https://cloud.google.com/iam/docs/service-account-overview + https://cloud.google.com/iam/docs/service-account-impersonation), both of which I've successfully set up.
## Demo

Through this code:

```ts
const store = Store.memoryStore();
const translator = new GoogleCloudTranslator({ store });
const speechSynthesizer = new GoogleCloudSpeechSynthesizer({ store });

const text = 'VotingWorks is a non-profit making better election technology';
const [translatedText] = await translator.translateText(
  [text],
  LanguageCode.SPANISH
);
const audioClipBase64 = await speechSynthesizer.synthesizeSpeech(
  translatedText,
  LanguageCode.SPANISH
);
fs.writeFileSync('audioClip.mp3', Buffer.from(audioClipBase64, 'base64'));
```

I'm able to generate this translated audio clip! :tada:

https://github.com/votingworks/vxsuite/assets/12616928/c530844a-83a8-4b2a-850f-4a7297704548

I manually converted the MP3 to an MP4 so that I could attach it to this PR description (GitHub doesn't support MP3 files in PR descriptions).

## Testing Plan

- [x] Added unit tests
- [x] Tested manually (see above)

## Checklist

### DIFF
--- a/apps/design/backend/schema.sql
+++ b/apps/design/backend/schema.sql
@@ -16,3 +16,19 @@ create table background_tasks (
   completed_at timestamp,
   error text
 );
+
+create table translation_cache (
+  source_text text not null,
+  target_language_code text not null,
+  translated_text text not null
+);
+
+create unique index idx_translation_cache on translation_cache (
+  source_text,
+  target_language_code
+);
+
+create table speech_synthesis_cache (
+  source_text text primary key,
+  audio_clip_base64 text not null
+);
--- a/apps/design/backend/src/language_and_audio/translator.ts
+++ b/apps/design/backend/src/language_and_audio/translator.ts
@@ -1,0 +1,99 @@
+import { TranslationServiceClient as GoogleCloudTranslationClient } from '@google-cloud/translate';
+import { assertDefined } from '@votingworks/basics';
+import { LanguageCode } from '@votingworks/types';
+
+import { Store } from '../store';
+import { GOOGLE_CLOUD_PROJECT_ID } from './google_cloud_config';
+
+export interface Translator {
+  translateText(
+    textArray: string[],
+    targetLanguageCode: LanguageCode
+  ): Promise<string[]>;
+}
+
+/**
+ * The subset of {@link GoogleCloudTranslationClient} that we actually use
+ */
+export type MinimalGoogleCloudTranslationClient = Pick<
+  GoogleCloudTranslationClient,
+  'translateText'
+>;
+
+/**
+ * An implementation of {@link Translator} that uses the Google Cloud Translation API
+ */
+export class GoogleCloudTranslator implements Translator {
+  private readonly store: Store;
+  private readonly translationClient: MinimalGoogleCloudTranslationClient;
+
+  constructor(input: {
+    store: Store;
+    // Support providing a mock client for tests
+    translationClient?: MinimalGoogleCloudTranslationClient;
+  }) {
+    this.store = input.store;
+    this.translationClient =
+      input.translationClient ??
+      /* istanbul ignore next */ new GoogleCloudTranslationClient();
+  }
+
+  async translateText(
+    textArray: string[],
+    targetLanguageCode: LanguageCode
+  ): Promise<string[]> {
+    const translatedTextArray: string[] = Array.from<string>({
+      length: textArray.length,
+    }).fill('');
+
+    const cacheMisses: Array<{ index: number; text: string }> = [];
+    for (const [index, text] of textArray.entries()) {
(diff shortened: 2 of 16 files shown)


apps/design/backend/schema.sql
```diff
@@ -16,3 +16,19 @@ create table background_tasks (
...
+
+create unique index idx_translations_cache on translation_cache (
+  source_text,
+  target_language_code
+);
+
+create table speech_synthesis_cache (
+  source_text text primary key,
+  audio_clip_base64 text not null
+);
```
arsalansufi, line 34: I recognize that, while we're all running locally, these caches 1) won't be shared across Vx employees and 2) will frequently get cleared, e.g. every time we make a DB schema change or run `pnpm clean-all`, etc. I'm not too concerned about costs right now, but as a stopgap until we deploy VxDesign, I may later add some logic to "bootstrap" everyone's local cache.

libs/basics/src/assert.ts
```diff
@@ -15,7 +15,7 @@ export function assert(
  * Asserts that `condition` is true. This function exists to avoid a polyfill
  * for `assert` in the browser.
  */
-export function assertDefined<T>(entity?: T, message?: string): T {
+export function assertDefined<T>(entity?: T | null, message?: string): T {
```
arsalansufi, line 18: This change is relevant for subsequent commits, as the Google Cloud API types include a number of values that can sometimes be `null`. And before this change, calling `assertDefined` on these values wasn't properly communicating to TS that they weren't `null`.

apps/design/backend/schema.sql
```diff
@@ -16,3 +16,19 @@ create table background_tasks (
...
+
+create unique index idx_translation_cache on translation_cache (
+  source_text,
+  target_language_code
+);
+
+create table speech_synthesis_cache (
+  source_text text primary key,
+  audio_clip_base64 text not null
+);
```
arsalansufi, line 34: I recognize that, while we're all running locally, these caches 1) won't be shared across Vx employees and 2) will frequently get cleared, e.g. every time we make a DB schema change or run `pnpm clean-all`, etc. I'm not too concerned about costs right now, but as a stopgap until we deploy VxDesign, I may later add some logic to "bootstrap" everyone's local cache.

libs/basics/src/assert.ts
arsalansufi, line 18: This change is relevant for subsequent commits, as the Google Cloud API types include a number of values that can sometimes be `null`. And before this change, calling `assertDefined` on these values wasn't properly communicating to TS that they weren't `null`.

apps/design/backend/src/language_and_audio/translator.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface Translator {
+  translateText(
+    textArray: string[],
+    targetLanguageCode: LanguageCode
+  ): Promise<string[]>;
+}
+
+/**
+ * The subset of {@link GoogleCloudTranslationClient} that we actually use
```
kofi-q, line 16: 💯

libs/basics/src/assert.ts
kofi-q: ah, thanks - this has been bugging me 😅

kofi-q (approved):
Exciting times!

Doesn't have to be part of this PR, but one thing I forgot to mention I noticed while generating translations for testing at translate.google.com: the text translation seems to handle markup just fine, but TTS includes the markup in the generated audio, reading out each character (e.g. https://translate.google.com/?sl=en&tl=es&text=Do%20%3Cb%3Eyou%3C%2Fb%3E%20like%20apples%3F&op=translate) -- I think we'll need to sanitise the text and strip out those placeholder tags before synthesis to handle those cases.

arsalansufi:
> I think we'll need to sanitise the text and strip out those placeholder tags before synthesis to handle those cases

Great catch! I'll tackle that in a follow-up PR

PR #4173, opened 2023-11-07
UWI Plumbing: Get UWIs from VxScanners to VxAdmin

adghayes:
_review by commit_

## Overview

Pipes UWIs from VxScanners to VxAdmin:

1. Incorporates UWIs into CVRs
2. Imports UWIs as write-ins, but not votes, into VxAdmin
3. Applies filters such that UWI status is retrievable from the database

Does _not_ add the ability to adjudicate marks, which will be added in a subsequent PR. Since it's not possible to adjudicate the marks yet, this is a non-issue.

## Testing Plan

Automated.


libs/ballot-interpreter/src/options.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export interface InterpreterOptions {
+  electionDefinition: ElectionDefinition;
+  precinctSelection: PrecinctSelection;
+  testMode: boolean;
+  markThresholds: MarkThresholds;
+  adjudicationReasons: readonly AdjudicationReason[];
+}
+
+export function getScoreWriteInsFlag(options: InterpreterOptions): boolean {
```
jonahkagan, line 19: Nit: maybe name this `shouldScoreWriteIns`

Also, do we need to make sure that `markThresholds.writeInTextArea` is defined if unmarked write-in detection is turned on? Or maybe we should just make `markThresholds.writeInTextArea` required, since the adjudication reason is now the on/off switch.

libs/utils/src/hmpb/ballot_adjudication_reasons.ts
```diff
@@ -1,46 +1,50 @@
...
  * Enumerates all the reasons a series of contests might need adjudication.
- * Callers must provide a function that can get the mark status for any contest
- * option in the contests given.
+ * Callers must provide a function that can get the mark status and write-in
+ * area status for any contest option in the contests given.
  */
 export function* ballotAdjudicationReasons(
   contests: Contests,
-  { optionMarkStatus }: Options
+  { optionStatus }: Options
```
jonahkagan, line 27: It looks like the `optionStatus` parameter is only used in one place. Can we just move the logic of that function into this function? It's kind of confusing to go back and forth to figure out what is determining the option status, and it seems like we don't actually need the flexibility of having it as a parameter.

libs/ballot-interpreter/src/legacy_adapter.ts
```diff
@@ -99,47 +177,68 @@ function convertNewHampshireNextMarkToSharedMark(
...
+            );
           }
-
-          return mark.optionId === option.id;
-        });
-        assert(
-          contestMarks.length > 0,
-          `mark for option ${option.id} not found`
         );
+        assert(contestOptionScore, `mark for option ${option.id} not found`);
```
jonahkagan, line 238: Can use `find` from libs/basics which has a built in assertion

libs/utils/src/hmpb/ballot_adjudication_reasons.ts
```diff
@@ -61,44 +68,64 @@ export function* ballotAdjudicationReasons(
 
           /* c8 ignore next 2 */
           default:
-            throwIllegalValue(status);
+            throwIllegalValue(markStatus);
+        }
+
+        if (
+          markStatus !== MarkStatus.Marked &&
+          writeInAreaStatus === 'filled'
```
jonahkagan, line 76: Nit: Maybe use the enum `WriteInAreaStatus.Filled` here for consistency

libs/ballot-interpreter/src/hmpb-ts/types.ts
```diff
@@ -130,6 +133,19 @@ export interface ScoredPositionArea {
...
+ * contest option, the mark score, and the write-in area score if applicable.
+ */
+export interface ScoredContestOption {
+  option: ContestOption;
+  gridPosition: GridPosition;
+  scoredMark: ScoredBubbleMark;
+  markStatus: MarkStatus;
+  scoredWriteInArea?: ScoredPositionArea;
+  writeInAreaStatus: WriteInAreaStatus;
+}
```
jonahkagan, line 147: Should this type just live in `legacy_adapter.ts` since it's only used there? I think this file is generally meant for types that interface between TS and Rust

libs/ballot-interpreter/src/legacy_adapter.ts
```diff
@@ -38,23 +41,19 @@ import {
...
-  InterpretFileResult,
-  InterpretResult,
-  InterpreterOptions,
-} from './interpret';
+import { type InterpretFileResult, type InterpretResult } from './interpret';
+import { getScoreWriteInsFlag, InterpreterOptions } from './options';
 
 type OkType<T> = T extends Ok<infer U> ? U : never;
 
-function convertNewHampshireNextMarkToSharedMark(
```
jonahkagan, line 53: Glad to see the New Hampshire references removed! It's about time.

I wonder if we should consider renaming the legacy adapter since it's kind of what we've got...

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
```diff
@@ -304,7 +300,52 @@ function buildCVRCandidateContest({
...
+            },
+          },
+        ],
+      };
+    }) ?? [];
+
+  return {
+    '@type': 'CVR.CVRContest',
+    ContestId: contest.id,
+    Overvotes: vote.length > contest.seats ? contest.seats : 0, // VVSG 2.0 1.1.5-E.2
```
jonahkagan, line 342: How do overvotes and unmarked write-ins interact here? If a voter overvotes by adding an unmarked write-in, what will the CVR look like?

libs/ballot-interpreter/src/legacy_adapter.ts
```diff
@@ -99,47 +177,68 @@ function convertNewHampshireNextMarkToSharedMark(
...
+      convertScoredContestOptionToLegacyMark(scoredContestOption)
+    ),
+  };
+
+  return markInfo;
+}
+
+function getUnmarkedWriteInsFromScoredContestOptions(
+  scoredContestOptions: ScoredContestOption[]
+): InterpretedHmpbPage['unmarkedWriteIns'] {
```
jonahkagan, line 196: I noticed there are a number of places that you use `InterpretedHmpbPage['unmarkedWriteIns']` rather than using whatever type it is directly. Is there a reason for that? Curious because it's different from what I usually see, so when I see it I usually think there's some meaning to it

libs/utils/src/cast_vote_records.ts
```diff
@@ -124,6 +131,11 @@ export function getWriteInsFromCastVoteRecord(
                   : pageIndex === 0
                   ? 'front'
                   : 'back',
+              isUnmarked:
+                selectionPosition.OtherStatus ===
+                UNMARKED_WRITE_IN_SELECTION_POSITION_OTHER_STATUS
+                  ? true
+                  : undefined,
```
jonahkagan, line 138: Why `true` or `undefined` vs just a boolean?

apps/admin/backend/src/store.ts
```diff
@@ -1725,6 +1725,11 @@ export class Store {
           left join
             write_in_candidates on write_in_candidates.id = write_ins.write_in_candidate_id
           where ${whereParts.join(' and ')}
+            and (
+              write_ins.is_unmarked = 0 or
+              official_candidate_id is not null or
+              write_in_candidate_id is not null
+            )
```
jonahkagan, line 1732: I'm having some trouble understanding what this new clause does. Not sure if `write_ins.is_unmarked = 0` is filtering unmarked write-ins in or out (is 0 true or false?). Also not sure what it means when `official_candidate_id` or `write_in_candidate_id` are not null. Maybe add an explanatory comment?

jonahkagan (commented):
I don't think I have any major blockers but I don't fully understand everything yet so not approving for now.

libs/ballot-interpreter/src/options.ts
adghayes: That's a good point. I substitute in the default in the Typescript, but maybe the rust interpreter is asserting it exists if the flag is on?

This is one of the pieces that I was unsure about pending your work.

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
adghayes: In the NH case, we will reject the ballot at VxScan if it's an unmarked write-in overvote.

If there's a future were we don't reject those, and there's a possibility of adjudicating it as an overvote at VxAdmin, it would still appear as normal / undervote in the CVR. 

Because UWIs can only become votes through adjudication and the scanner CVRs are pre-adjudication, they should be treated as undervotes.

libs/ballot-interpreter/src/legacy_adapter.ts
adghayes: Not a strong reason, although sometimes I prefer it because it communicates more. It indicates that the UWI type is a component of the interpretation type, which tells me a little something about how to find it and its purpose.

apps/admin/backend/src/store.ts
adghayes: Yeah, the adjudication logic is difficult to parse. I will add a comment.

libs/utils/src/hmpb/ballot_adjudication_reasons.ts
adghayes: The way that this logic is split between the adapter and `utils` is confusingly specific, and I opted to keep it in case there was an equally specific justification it.

I think that it should just be moved from `utils` into `ballot-interpreter`.

libs/ballot-interpreter/src/options.ts
adghayes: OK. Yeah, the place where the threshold is required I just have a temporary default in place until our threshold strategy gets clearer.

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
jonahkagan: Can't the voter choose to cast the ballot despite the overvote at VxScan?

I agree it makes sense to treat the unmarked write-in as an undervote in the pre-adjudication CVR.

libs/utils/src/hmpb/ballot_adjudication_reasons.ts
jonahkagan: Agreed

libs/utils/src/cast_vote_records.ts
jonahkagan: Cool, I'd lean towards a required boolean since we always know one way or the other. When I see an optional field I have to think about if there's a diff between false and undefined or not

libs/backend/src/cast_vote_records/build_cast_vote_record.ts
adghayes: In the NH case they cannot choose to cast the ballot despite the overvote. @mattroe can correct me if I'm wrong.

apps/admin/backend/src/store.ts
```diff
@@ -1705,6 +1705,13 @@ export class Store {
     const officialCandidateNameLookup =
       getOfficialCandidateNameLookup(election);
 
+    // we ignore unmarked write-ins that are not adjudicated for a candidate
+    whereParts.push(`not (
+      write_ins.is_unmarked = 1 and
```
jonahkagan, line 1710: Thanks for adding the comment, it's more clear now.

I'm a bit concerned that the 0 changed to a 1 but I don't see a corresponding test update. Is this case covered?

libs/ballot-interpreter/src/adjudication_reasons.ts
```diff
@@ -0,0 +1,175 @@
...
+  WriteInAreaStatus,
+} from '@votingworks/types';
+import { find, throwIllegalValue } from '@votingworks/basics';
+import { allContestOptions } from '@votingworks/utils';
+import { type ScoredContestOption } from './legacy_adapter';
+
+/**
+ * Enumerates all the reasons a series of contests might need adjudication.
+ * Callers must provide a function that can get the mark status and write-in
+ * area status for any contest option in the contests given.
```
jonahkagan, line 17: I don't think this part of the comment applies anymore

jonahkagan (approved):

PR #4178, opened 2023-11-07
Update libs/auth scripts such that they can be run from anywhere

arsalansufi:
_Easiest reviewed by commit_

## Overview

Was working on some scripts for testing the Google Cloud Translation and Text-to-Speech API integrations and took a quick detour to make some low-hanging improvements to `libs/auth` scripts.

Main update is that `libs/auth` scripts can now be run from anywhere. Previously, many had to be run from the root of the `libs/auth` directory, as noted in the `libs/auth` README:

> Note that all scripts are meant to be run from `libs/auth/` and not `libs/auth/scripts/`.

I also made some tweaks to the recently added PIN-checking script, most notably no longer requiring specifying `VX_MACHINE_TYPE=admin` when checking a VxSuite card's PIN.

## Testing Plan

- [x] Manually tested all modified scripts

### DIFF
--- a/libs/auth/scripts/check_pin.ts
+++ b/libs/auth/scripts/check_pin.ts
@@ -11,7 +11,7 @@ const usageMessage = 'Usage: check-pin [--cac|--vxsuite (default)]';
 type CardType = 'cac' | 'vxsuite';
 
 function parseCommandLineArgs(args: readonly string[]): { cardType: CardType } {
-  if (![undefined, '--cac', '--vxsuite'].includes(args[0]) || args.length > 1) {
+  if (args.length > 1 || ![undefined, '--cac', '--vxsuite'].includes(args[0])) {
     console.log(usageMessage);
     process.exit(0);
   }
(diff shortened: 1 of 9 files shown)


libs/auth/scripts/check_pin.ts
```diff
@@ -1,58 +1,63 @@
...
-  console.timeEnd('waitForReadyCardStatus');
+  let card: PinProtectedCard &
+    StatefulCard<CardDetails | CommonAccessCardDetails>;
+  switch (cardType) {
+    case 'cac': {
+      card = new CommonAccessCard();
+      break;
+    }
+    case 'vxsuite': {
+      (process.env.VX_MACHINE_TYPE as string) = 'admin';
```
arsalansufi, line 37: The main update in this commit - previously, this had to be manually set. Rest of the updates are just for consistency with how other `libs/auth` scripts are structured

kofi-q (approved):

PR #4188, opened 2023-11-09
Commit app strings catalog to VxSuite codebase

arsalansufi:
## Overview

This PR commits the i18next-parser-generated app strings catalog to the VxSuite codebase so that VxDesign can access it for language and audio file generation.

I've tweaked pnpm commands such that:

1. The catalog is automatically updated when `app_strings.tsx` is modified, via a pre-commit hook.
2. CI will detect if the catalog is out of sync, e.g. because someone made a commit bypassing pre-commit hooks.

I'll implement a versioning scheme in a later PR. For now, I'm just storing a "latest" catalog.

## Demo Videos

_Catalog update via pre-commit hook_

https://github.com/votingworks/vxsuite/assets/12616928/e611d366-993b-4a01-8902-41c19d2647bf

Note that the catalog file does still have to be committed after it's updated. The pre-commit hook handles the update but not the commit.

_Detecting if the catalog is out of sync_

https://github.com/votingworks/vxsuite/assets/12616928/1f3256bf-def2-4108-abd4-7e76aa94cb1b

## Testing Plan

- [x] Tested pnpm commands manually

## Checklist


.circleci/config.yml
```diff
@@ -1083,13 +1083,6 @@ jobs:
             JEST_JUNIT_OUTPUT_DIR: ./reports/
       - store_test_results:
           path: libs/ui/reports/
-      - when:
-          condition:
-            equal: [ main, << pipeline.git.branch >> ]
-          steps:
-            - run:
-                name: Export App Strings Catalog
-                command: echo 0 # TODO(kofi): export to versioned folder accessible from VxDesign
```
arsalansufi, line 1092: The hypothetical versioned folder accessible from VxDesign mentioned here is now just `libs/ui` itself, specifically `libs/ui/src/ui_strings/app_strings_catalog/`

PR #4190, opened 2023-11-09
[fix/uistrings] Avoid numeric entity IDs getting treated as array indices

kofi-q:
## Overview

@adghayes ran into a bug with ui string extraction where older Vx election fixtures with numeric IDs (precinct ID, party ID, etc) would cause issues when converted to CDF elections and imported into voter-facing apps.

This change explicitly sets numeric IDs in the `UiStringTranslations` object.

Before:
```ts
{
  precinctName: [
    'Precinct A',
    'Precinct B',
  ],
}
```

After:
```ts
{
  precinctName: {
    '0': 'Precinct A',
    '1': 'Precinct B',
  },
}
```

## Testing Plan
- Added a unit test to repro the bug and then verify the fix

## Checklist


adghayes (approved):
Simple fix, confirmed it works. Unblocked. Thank you!

PR #4192, opened 2023-11-09
VxDesign blank ballot adjudication option

adghayes:
## Overview

Adds option on the Tabulation screen to adjudication unmarked write-ins. Also starts including the set system settings in the setup package - previously it was just exporting the default regardless of what was set in the app.

## Demo Video or Screenshot

Screen Shot 2023-11-09 at 11 16 36 AM

## Testing Plan

Edited automated tests accordingly + tested manually.

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -6,7 +6,6 @@ import {
   Id,
   safeParseElection,
   BallotPaperSize,
-  DEFAULT_SYSTEM_SETTINGS,
   SystemSettings,
   BallotType,
   BallotPackageFileName,
@@ -225,7 +224,9 @@ function buildApi({ store }: { store: Store }) {
     async exportSetupPackage(input: {
       electionId: Id;
     }): Promise<{ zipContents: Buffer; electionHash: string }> {
-      const { election, layoutOptions } = store.getElection(input.electionId);
+      const { election, layoutOptions, systemSettings } = store.getElection(
+        input.electionId
+      );
       const { electionDefinition } = layOutAllBallotStyles({
         election,
         // Ballot type and ballot mode shouldn't change the election definition, so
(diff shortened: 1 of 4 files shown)


apps/design/frontend/src/tabulation_screen.tsx
jonahkagan, line 1: We could also use a field here to set the write-in detection threshold. We also still need to resolve how these two things (adjudication reason and threshold) relate to each other. My thought is to just require that the write-in detection threshold is always set so we don't have to worry about it in the code. If we want, the UI here in VxDesign can be smarter and only show the field to set that threshold if the adjudication reason is turned on, but I don't know that we even need to do that given that we're already showing the marginal mark threshold regardless of whether its used.

apps/design/backend/src/app.ts
```diff
@@ -239,7 +240,7 @@ function buildApi({ store }: { store: Store }) {
       zip.file(BallotPackageFileName.ELECTION, electionDefinition.electionData);
       zip.file(
         BallotPackageFileName.SYSTEM_SETTINGS,
-        JSON.stringify(DEFAULT_SYSTEM_SETTINGS, null, 2)
+        JSON.stringify(systemSettings, null, 2)
```
jonahkagan, line 243: Oops, nice catch

jonahkagan (approved):

apps/design/frontend/src/tabulation_screen.tsx
adghayes: I think for now we could just add it like the marginal mark threshold, yeah, I will tack that on.

adghayes: I ended up making it so that it only shows up when relevant - I didn't want users getting confused about it affecting write-ins when the adjudication reason is not on:


https://github.com/votingworks/vxsuite/assets/37960853/114ff18f-5968-4e57-abc2-2669afa59be8

jonahkagan: I might label it `Unmarked Write-In Threshold` to make it even clearer that they are related.

What do you think about applying the same show/hide logic to the marginal mark threshold?

And what do you think about making the `writeInTextArea` field required in mark thresholds so we don't have to worry about it getting inconsistent with the adjudication reasons?

adghayes: I think that would be a better label, good suggestion.

I support applying the same logic to the marginal mark threshold!

I support making the field required for type safety / assumptions.

I also think we might consider renaming "Mark Thresholds" if we can think of a better name, because it's not exclusively marks anymore.

PR #4195, opened 2023-11-09
fix(scan): we were ignoring undervoted ballots on VxScan reports

adghayes:
In testing another feature, I noticed a bug where the VxScan precinct report has the wrong number of ballots in the ballot count above each contest. This is because the `votes` object returned from the interpreter doesn't include keys / empty arrays for undervoted contests, and the tabulation code relies on those keys to know the number of ballots. To fix the bug and to make the `votes` object more useful, I added in the empty contests. This puts `votes` more in alignment with how we represent undervotes in the CVR CDF, where they have to be explicitly listed rather than silently omitted.


libs/ballot-interpreter/src/__snapshots__/interpret_nh_hmpb.test.ts.snap
adghayes, line 1: I have checked that the only difference in snapshot is adding in keys with empty arrays.

libs/utils/src/votes.ts
```diff
@@ -156,9 +156,7 @@ export function convertMarksToVotesDict(
         : /* c8 ignore next */
           throwIllegalValue(contest, 'type');
 
-    if (newVotes.length > 0) {
-      votesDict[mark.contestId] = [...existingVotes, ...newVotes] as Vote;
-    }
+    votesDict[mark.contestId] = [...existingVotes, ...newVotes] as Vote;
```
adghayes, line 159: Everything other than this is just testing updates.

adghayes:
@eventualbuddha I realized I probably need to make a similar change for BMD ballots, reverting to drarft

libs/utils/src/votes.ts
adghayes: (within this commit)

adghayes:
The `libs/ui` test is failing due to a coverage miss that I can't reproduce locally - an `istanbul ignore next` covers it. I think I can just add a test to cover it but it's a weird case.

libs/utils/src/votes.ts
jonahkagan: Do we need a regression test for the tally code where you first noticed this issue? I.e. a test that would have failed before that passes after this change?

jonahkagan (approved):

libs/utils/src/votes.ts
adghayes: The tally code doesn't need a regression test, I don't think, because the tally code and report display code never changed. It was correctly handling the interpreter results. I don't think there was a good, clean way to address this issue at the tally code level.

I did want to make sure that the BMD and HMPB interpreters were outputting full `votes` dictionaries, and updating the existing tests covered that. I thought about adding am explicit test called "intepreter exports `votes` for all contests on page" but I think it's an extensive enough assertion now that the intent should be apparent.

PR #4197, opened 2023-11-12
unreadable pages in the NH legacy adapter should show up as unreadable pages, not as scanning errors

benadida:
In the NH legacy ballot adapter, we were treating uninterpretable ballots as scanning errors, which in central-scanner just interrupt the batch scanning and effectively cause one ballot go unaccounted for.

This PR fixes the error that gets reported in this case.

We should probably also fix the silent swallowed error.

~TODO: needs a test, probably a high-level test, to make sure the right error gets surfaced when scanning, say, a blank sheet of paper.~

### DIFF
--- a/libs/ballot-interpreter/src/legacy_adapter.ts
+++ b/libs/ballot-interpreter/src/legacy_adapter.ts
@@ -1,6 +1,5 @@
 import {
   assert,
-  err,
   find,
   iter,
   ok,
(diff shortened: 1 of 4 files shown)


libs/ballot-interpreter/src/legacy_adapter.ts
```diff
@@ -407,13 +408,29 @@ function convertNextInterpretedBallotPage(
...
+  sheet: SheetOf<string>
+): Promise<InterpretResult> {
   /* istanbul ignore next */
   if (nextResult.isErr()) {
-    return err(new Error(JSON.stringify(nextResult.err())));
+    return ok([
+      {
+        interpretation: {
+          type: 'UnreadablePage',
+          reason: JSON.stringify(nextResult.err().type),
```
eventualbuddha, line 422: Was there a particular reason to `JSON.stringify` this? The `type` is just a string, so we could just pass it through as-is.

benadida: nope, copy-pasted and then changed the content and forgot to remove the stringify. I'll update.

eventualbuddha (approved):

PR #4203, opened 2023-11-13
fix(ballot-interpreter): improve QR code detection

eventualbuddha:
## Overview

Changes the search area bounds to better handle skew. In comparing the search area with the actual QR code location, I found that the area was sometimes not wide enough and also often taller than needed. This also changes the height of the search area to be based on the ballot image width, since the QR code is anchored to the bottom of the image and searching more in a taller ballot will not help.

This has a negligible impact on performance. For a letter-sized ballot image (1700x2200 pixels), we previously were searching 149,600 pixels and we're now searching 180,625 pixels. This did not seem to have a noticeable impact on runtime.

([Slack thread](https://votingworks.slack.com/archives/C060FEVB2L8/p1699500954906219))

## Demo Video or Screenshot
| Before | After |
|-|-|
|image|image|

## Testing Plan
Tested manually, added automated regression test.


libs/ballot-interpreter/src/hmpb-rust/qr_code/detect.rs
```diff
@@ -235,4 +238,22 @@ mod test {
...
+    fn test_detect_qr_code_in_skewed_image() {
+        let fixture_path =
+            PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("test/fixtures");
+        let scan_side_a_path = fixture_path.join("scan-skewed-side-a.jpeg");
+        let scan_side_b_path = fixture_path.join("scan-skewed-side-b.jpeg");
+        detect(
+            &image::open(scan_side_a_path).unwrap().into_luma8(),
+            &ImageDebugWriter::disabled(),
+        )
+        .expect("side A QR code not detected");
```
jonahkagan, line 252: When I first read this, I really thought it was an assertion that the expected result should be "side A QR code not detected." I guess this is a pitfall of the `expect` nomenclature

jonahkagan (approved):

benadida (approved):

libs/ballot-interpreter/src/hmpb-rust/qr_code/detect.rs
eventualbuddha: I changed it to be clearer since it looks like we mostly use it the other way.

PR #4205, opened 2023-11-14
Improved desktop RadioGroup

jonahkagan:
## Overview

- Consolidated to a single file
- Disable entire group, not individual options
- Uses Button component directly, remove hidden input
- Uses icons instead of CSS for radio circle graphic
- Updates styles for desktop theme
- Ensures tabbing through options works

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/7ba89373-ac64-4fc5-85cf-9d91c50054e8

## Testing Plan
- Updated unit tests
- Manual testing in Storybook
- Manually tested tabbing and screen reader in VxMark

## Checklist


kofi-q:
Saw the note about tabbing in the description -- just a heads-up that recommended a11y patterns for radio is that `tab` key should focus in and out of the group as a whole and `up`/`down` and `left`/`right` move between values in the group: https://www.w3.org/WAI/ARIA/apg/patterns/radio/

jonahkagan:
Thanks @kofi-q for pointing that out! I had looked at that briefly but didn't realize it was working correctly because I didn't see the focus outline, so I assumed it was broken already. It turned out that while the outline wasn't showing, the focus behavior was working as specified. I managed to figured out a way to get the focus behavior to work (using the hidden input you previously were using) while also getting the outline to work!

ec76f1fce

https://github.com/votingworks/vxsuite/assets/530106/204d1efc-0f64-4ce4-82ce-f65c394d3d60

(Side note: I found a bug in the screen reader that I will fix in a separate PR)

PR #4212, opened 2023-11-15
VxDesign desktop theme

jonahkagan:
## Overview

Uses the desktop color theme for VxDesign. Along the way, switches VxDesign to use as many libs/ui components as possible, replacing the custom components I threw together during prototyping.

There are probably still design improvements we can make now that we're using the desktop theme, but I tried to keep this as simple as possible of a change to start.

_Review by commit_

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/a57b2afe-8f10-463b-a1d7-9de180915341

## Testing Plan
Updated existing tests, plus a bunch of manual testing and visual inspection

## Checklist


apps/design/frontend/src/layout.tsx
```diff
@@ -18,22 +18,17 @@ export const Form = styled.form`
...
-    background: transparent;
+  input[type='text'] {
+    min-width: 18rem;
+  }
+  .search-select {
+    min-width: 18rem;
   }
 `;
 
+export const Input = styled.input``;
```
jonahkagan: removed in a later commit

libs/ui/src/global_styles.tsx
```diff
@@ -152,10 +152,10 @@ export const GlobalStyles = createGlobalStyle<GlobalStylesProps>`
     --focus-outline: ${(p) =>
       p.isTouchscreen
         ? `${p.theme.colors.primary} dashed ${p.theme.sizes.bordersRem.medium}rem`
-        : 'none'};
+        : `${p.theme.colors.primary} solid ${p.theme.sizes.bordersRem.medium}rem`}
   }
 
-  :focus {
+  :focus-visible {
```
jonahkagan, line 158: This change helped make SearchSelect focus correctly. Details on why this is useful here: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#focus_vs_focus-visible

libs/ui/src/reports/contest_results_table.tsx
```diff
@@ -12,7 +12,8 @@ import { format, getTallyReportCandidateRows } from '@votingworks/utils';
 import { throwIllegalValue, assert, Optional } from '@votingworks/basics';
 
 import { Text, NoWrap } from '../text';
-import { tableBorderColor } from '../table';
+
+const tableBorderColor = 'rgb(194, 200, 203)';
```
jonahkagan, line 16: I didn't want to mess with the print styles, so I just inlined the old border color here for now.

adghayes: I am envisioning that, with eventually moving toward a different print infrastructure, we'll totally partition print and UI styles.

adghayes (approved):
Looks great to me! I played around with it locally and didn't notice anything I didn't see in your video.

Having purple "Save" buttons in contrast with red "Delete" buttons is a little confusing to me since I'm so used to the larger green vs. red contrast. But I like the purple throughout the app and the consistency, so I'm not suggesting re-introducing green unless we hear about it from a lot of other people.

Issue #4096, opened 2023-10-23
Improve continuous export failure recovery logic


arsalansufi:
Potential improvements include:

1. Handling the case that even [this first write to the USB drive](https://github.com/votingworks/vxsuite/blob/f8e32c36326b46d66f275b8cb7609966f5fd3371/libs/backend/src/cast_vote_records/export.ts#L700) fails and [this check](https://github.com/votingworks/vxsuite/blob/f8e32c36326b46d66f275b8cb7609966f5fd3371/libs/backend/src/cast_vote_records/export.ts#L883-L886) accordingly misses the need for recovery. Could be solved by relying on a piece of state on the machine rather than the USB drive.
2. Only re-syncing the one CVR that failed to be written to the USB drive, instead of all CVRs, when possible.

arsalansufi:
Going to close this issue since (1) is done and will open a separate issue for just (2)

PR #4213, opened 2023-11-15
Improve continuous export failure recovery by moving the export operation "progress marker" from USB drive to VxScan store

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4096

This PR improves continuous export failure recovery by moving the export operation "progress marker" from the USB drive to the VxScan store.

We currently write a `.vx-export-in-progress` file to the USB drive at the start of every continuous export operation. We then delete the file when the operation completes. If an export operation fails midway and a machine has to be restarted, the existence of that file on startup tells us that a resync is required.

The main downside to using a file on the USB drive is that the write of that file itself may fail. In that case, the machine won't know that a resync is required after restart. Switching to a piece of state on the machine is more robust, especially when paired with a transaction.

## Testing

- [x] Verified that existing tests still pass
- [x] Added unit tests
- [x] Tested failure recovery manually by removing the USB drive right after scanning a ballot

## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/custom/state_machine.ts
+++ b/apps/scan/backend/src/scanners/custom/state_machine.ts
@@ -434,11 +434,15 @@ async function recordAcceptedSheet(
   const { sheetId } = interpretation;
   store.withTransaction(() => {
     storeInterpretedSheet(store, sheetId, interpretation);
+
     // If we're storing an accepted sheet that needed review, that means that it was "adjudicated"
     // (i.e. the voter said to count it without changing anything).
     if (interpretation.type === 'NeedsReviewSheet') {
       store.adjudicateSheet(sheetId);
     }
+
+    // Gets reset to false within exportCastVoteRecordsToUsbDrive
+    store.setIsContinuousExportOperationInProgress(true);
   });
   (
     await exportCastVoteRecordsToUsbDrive(
@@ -460,10 +464,14 @@ async function recordRejectedSheet(
   const { sheetId } = interpretation;
   store.withTransaction(() => {
     storeInterpretedSheet(store, sheetId, interpretation);
+
     // We want to keep rejected ballots in the store, but not count them. We accomplish this by
     // "deleting" them, which just marks them as deleted and is how we indicate that an interpreted
     // ballot wasn't counted.
     store.deleteSheet(sheetId);
+
+    // Gets reset to false within exportCastVoteRecordsToUsbDrive
+    store.setIsContinuousExportOperationInProgress(true);
   });
   (
     await exportCastVoteRecordsToUsbDrive(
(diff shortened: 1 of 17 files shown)


apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -434,11 +434,15 @@ async function recordAcceptedSheet(
...
     storeInterpretedSheet(store, sheetId, interpretation);
+
     // If we're storing an accepted sheet that needed review, that means that it was "adjudicated"
     // (i.e. the voter said to count it without changing anything).
     if (interpretation.type === 'NeedsReviewSheet') {
       store.adjudicateSheet(sheetId);
     }
+
+    // Gets reset to false within exportCastVoteRecordsToUsbDrive
+    store.setIsContinuousExportOperationInProgress(true);
```
arsalansufi, line 445: I actually implement this store method in the next commit

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -72,12 +71,36 @@ export interface ScannerStore {
...
+  scannerType: 'precinct';
+
+  getBallotsCounted(): number;
+  getExportDirectoryName(): string | undefined;
+  getPollsState(): PollsState;
+  isContinuousExportOperationInProgress(): boolean;
+  setExportDirectoryName(exportDirectoryName: string): void;
+  setIsContinuousExportOperationInProgress(
+    isContinuousExportOperationInProgress: boolean
+  ): void;
```
arsalansufi, line 96: Rather than having all these methods be optional on the `ScannerStore` interface, I've now updated `ScannerStore` to be a union of two distinct interfaces, one for the central scanner and one for the precinct scanner (still sharing what I can though, via `ScannerStoreBase`)

apps/scan/backend/src/store.ts
```diff
@@ -823,15 +827,44 @@ export class Store {
...
 
+  isContinuousExportOperationInProgress(): boolean {
+    const row = this.client.one(
+      `
+      select
+        is_continuous_export_operation_in_progress as isContinuousExportOperationInProgress
+      from is_continuous_export_operation_in_progress
+      `
+    ) as { isContinuousExportOperationInProgress: number } | undefined;
+    return Boolean(row?.isContinuousExportOperationInProgress);
```
adghayes, line 838: I just added a type + basic utils to `admin/backend/src/store.ts` for being more explicit about these boolean conversions. What do you think about copying those or moving them to a shared library to use in this `store` file?

adghayes (approved):

apps/scan/backend/src/store.ts
arsalansufi: Oh nice! Can move to a shared location

arsalansufi: ✅ https://github.com/votingworks/vxsuite/pull/4213/commits/a9caa1408ae3d1b3c36cc007074922209d75f28c

PR #4215, opened 2023-11-15
Create UiStringsApi type from API object

jonahkagan:
## Overview

Grout is designed to infer the type from the API object literal, rather
than have a separate interface to define the type. This change removes
the need for a workaround to create the grout client type for the
UiStringsApi.

## Checklist


arsalansufi (approved):
Thanks for updating!

Issue #3615, opened 2023-06-22
Add a more specific error message for jurisdiction mismatches


arsalansufi:
Jurisdiction mismatches come up quite frequently in dev, so let's make it easier to notice when they're the cause of an error. Could be additionally helpful to surface the machine's current jurisdiction in the error message.

arsalansufi:
Could be additionally helpful to surface the machine's current jurisdiction in the error message

PR #4218, opened 2023-11-15
Add a more specific error message for jurisdiction mismatches between card and machine

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3615

TODO

## Demo Screenshots

TODO

## Testing Plan

- [x] Automated tests
- [x] Manually tested VxAdmin and VxScan

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced ➡️ Relying on existing logging pathways

### DIFF
--- a/apps/scan/frontend/src/screens/invalid_card_screen.tsx
+++ b/apps/scan/frontend/src/screens/invalid_card_screen.tsx
@@ -1,11 +1,17 @@
+import { InsertedSmartCardAuth } from '@votingworks/types';
 import { InvalidCardScreen as SharedInvalidCardScreen } from '@votingworks/ui';
+
 import { ScreenMainCenterChild } from '../components/layout';
 
-export function InvalidCardScreen(): JSX.Element {
+export function InvalidCardScreen({
+  authStatus,
+}: {
+  authStatus: InsertedSmartCardAuth.LoggedOut;
+}): JSX.Element {
   return (
     <ScreenMainCenterChild infoBarMode="pollworker">
       <SharedInvalidCardScreen
-        reason="invalid_user_on_card"
+        reasonAndContext={authStatus}
         recommendedAction="Remove the card to continue."
       />
     </ScreenMainCenterChild>
@@ -14,5 +20,9 @@ export function InvalidCardScreen(): JSX.Element {
 
 /* istanbul ignore next */
 export function DefaultPreview(): JSX.Element {
-  return <InvalidCardScreen />;
+  return (
+    <InvalidCardScreen
+      authStatus={{ status: 'logged_out', reason: 'invalid_user_on_card' }}
+    />
+  );
 }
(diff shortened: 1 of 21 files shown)


libs/ui/src/invalid_card_screen.tsx
```diff
@@ -1,47 +1,75 @@
...
-    heading = 'Card is Backwards';
-    recommendedAction = 'Remove the card, turn it around, and insert it again.';
-  }
-
-  if (reason === 'election_manager_wrong_election') {
-    errorDescription =
-      'The inserted Election Manager card is programmed for another election ' +
-      'and cannot be used to unlock this machine.';
-  }
+    recommendedActionOverride ?? 'Please insert a valid card.';
```
arsalansufi, line 31: Switched from "Please insert a valid Election Manager or System Administrator card" to "Please insert a valid card" as the default recommended action since this component is used not only in VxAdmin and VxCentralScan but also in VxScan, where poll worker cards are also valid. (When this component was first created, it was only used in VxAdmin and VxCentralScan.)

In most contexts, we do already provide a `recommendedAction` prop, so this is fairly inconsequential.

apps/scan/frontend/src/screens/invalid_card_screen.tsx
```diff
@@ -1,11 +1,17 @@
...
+export function InvalidCardScreen({
+  authStatus,
+}: {
+  authStatus: InsertedSmartCardAuth.LoggedOut;
+}): JSX.Element {
   return (
     <ScreenMainCenterChild infoBarMode="pollworker">
       <SharedInvalidCardScreen
-        reason="invalid_user_on_card"
+        reasonAndContext={authStatus}
```
arsalansufi, line 14: We'll actually now get better error messages on VxScan not just for jurisdiction mismatches but also other cases, e.g. election mismatches

libs/ui/src/invalid_card_screen.tsx
```diff
@@ -1,47 +1,75 @@
...
 
-  if (reason === 'machine_not_configured') {
-    errorDescription =
-      'This machine is unconfigured and cannot be unlocked with this card.';
+  switch (reason) {
+    case 'card_error': {
+      graphic = <RotateCardImage />;
+      // The only cause we currently know of for a card error
+      heading = 'Card is Backwards';
```
kshen0, line 37: I think I hit this case when I had a faulty card reader (or faulty connection to the card reader), where the card reader was detected on VxMarkScan but couldn't read any cards. Replacing the card reader solved the problem.

Still, if a backwards card is much more likely to happen than a faulty card reader, makes sense to keep this as is.

kshen0 (approved):

libs/ui/src/invalid_card_screen.tsx
arsalansufi: Ah right, thanks for reminding me of that. I've left the code as is but updated the comment: https://github.com/votingworks/vxsuite/pull/4218/commits/9f3acae111a4c0b8b0cb5caae62d8dba036b4aa6

PR #4221, opened 2023-11-16
feature(export-logs): export all historical logs at new location.

benadida:
A revamp of log export. Moved to the backend. This now includes all historical logs, everything in `/var/log/votingworks`. For the moment, CDF export is disabled.

### DIFF
--- a/apps/admin/frontend/src/screens/logs_screen.test.tsx
+++ b/apps/admin/frontend/src/screens/logs_screen.test.tsx
@@ -1,7 +1,8 @@
-import userEvent from '@testing-library/user-event';
 import { fakeKiosk } from '@votingworks/test-utils';
+import { ok } from '@votingworks/basics';
+import { mockUsbDriveStatus } from '@votingworks/ui';
+import userEvent from '@testing-library/user-event';
 import { screen } from '../../test/react_testing_library';
-
 import { LogsScreen } from './logs_screen';
 import { renderInAppContext } from '../../test/render_in_app_context';
 import { ApiMock, createApiMock } from '../../test/helpers/mock_api_client';
@@ -20,28 +21,15 @@ afterEach(() => {
 });
 
 test('Exporting logs', async () => {
-  renderInAppContext(<LogsScreen />, { apiMock });
+  renderInAppContext(<LogsScreen />, {
+    apiMock,
+    usbDriveStatus: mockUsbDriveStatus('mounted'),
+  });
+
+  apiMock.apiClient.exportLogsToUsb.expectCallWith().resolves(ok());
 
   // Log saving is tested fully in src/components/export_logs_modal.test.tsx
   userEvent.click(screen.getByText('Save Log File'));
-  await screen.findByText('No Log File Present');
-  userEvent.click(screen.getByText('Close'));
-
-  // Log saving is tested fully in src/components/export_logs_modal.test.tsx
-  userEvent.click(screen.getByText('Save CDF Log File'));
-  await screen.findByText('No Log File Present');
-  userEvent.click(screen.getByText('Close'));
-});
-
-test('Exporting logs when no election definition', async () => {
-  renderInAppContext(<LogsScreen />, { electionDefinition: 'NONE', apiMock });
-
-  // Log saving is tested fully in src/components/export_logs_modal.test.tsx
-  userEvent.click(screen.getByText('Save Log File'));
-  await screen.findByText('No Log File Present');
-  userEvent.click(screen.getByText('Close'));
-
-  expect(
-    screen.getByText('Save CDF Log File').closest('button')
-  ).toHaveAttribute('disabled');
+  await screen.findByText('Save logs on the inserted USB drive?');
+  userEvent.click(screen.getByText('Save'));
 });
--- a/apps/admin/frontend/src/screens/logs_screen.tsx
+++ b/apps/admin/frontend/src/screens/logs_screen.tsx
@@ -1,21 +1,32 @@
 import { useContext } from 'react';
 import { ExportLogsButtonRow } from '@votingworks/ui';
+import { err } from '@votingworks/basics';
+import { LogsResultType } from '@votingworks/backend';
+import { exportLogsToUsb } from '../api';
 
 import { AppContext } from '../contexts/app_context';
 import { NavigationScreen } from '../components/navigation_screen';
 
 export function LogsScreen(): JSX.Element {
-  const { electionDefinition, usbDriveStatus, auth, logger, machineConfig } =
-    useContext(AppContext);
+  const { usbDriveStatus, auth, logger } = useContext(AppContext);
+
+  const exportLogsToUsbMutation = exportLogsToUsb.useMutation();
+
+  async function doExportLogs(): Promise<LogsResultType> {
+    try {
+      return await exportLogsToUsbMutation.mutateAsync();
+    } catch (e) {
+      return err('copy-failed');
+    }
+  }
 
   return (
     <NavigationScreen title="Logs">
       <ExportLogsButtonRow
-        electionDefinition={electionDefinition}
         usbDriveStatus={usbDriveStatus}
         auth={auth}
         logger={logger}
-        machineConfig={machineConfig}
+        onExportLogs={doExportLogs}
       />
     </NavigationScreen>
   );
--- a/apps/scan/frontend/src/app_root.tsx
+++ b/apps/scan/frontend/src/app_root.tsx
@@ -17,7 +17,8 @@ import {
 } from '@votingworks/utils';
 import { Logger } from '@votingworks/logging';
 
-import { assert } from '@votingworks/basics';
+import { assert, err } from '@votingworks/basics';
+import { LogsResultType } from '@votingworks/backend';
 import { PrecinctReportDestination } from '@votingworks/types';
 import { LoadingConfigurationScreen } from './screens/loading_configuration_screen';
 import { ElectionManagerScreen } from './screens/election_manager_screen';
@@ -45,6 +46,7 @@ import {
   getUsbDriveStatus,
   transitionPolls,
   unconfigureElection,
+  exportLogsToUsb,
 } from './api';
 import { VoterScreen } from './screens/voter_screen';
 import { LoginPromptScreen } from './screens/login_prompt_screen';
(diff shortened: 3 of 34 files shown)


apps/admin/frontend/src/screens/logs_screen.tsx
```diff
@@ -8,6 +9,13 @@ export function LogsScreen(): JSX.Element {
   const { electionDefinition, usbDriveStatus, auth, logger, machineConfig } =
     useContext(AppContext);
 
+  const exportLogsToUsbMutationAsync =
+    exportLogsToUsb.useMutation().mutateAsync;
```
arsalansufi, line 13: Question for you @jonahkagan: Are we still recommending usage of `.mutate` instead of `.mutateAsync` (unless wrapped in a try-catch) for error handling purposes, as documented on [this slide](https://docs.google.com/presentation/d/1qcohhkeavCVjTZAAhGbbB7sbnvDJLlNov3OTLijfcvs/edit#slide=id.g1f8414d696e_0_361)?

libs/backend/src/logs/logs_api.ts
```diff
@@ -0,0 +1,48 @@
...
+/** Creates a shareable implementation of {@link LogsApi}. */
+export function createLogsApi({
+  usbDrive,
+  machineId,
+}: {
+  usbDrive: UsbDrive;
+  machineId: string;
+}): LogsApi {
+  return {
+    async exportLogsToUsb(): Promise<boolean> {
```
arsalansufi, line 17: Instead of returning a boolean to indicate success vs. failure, this might be a good use case for the [`Result` class](https://github.com/votingworks/vxsuite/blob/main/libs/basics/src/result.ts), e.g. `Result<void, 'no-log-directory' | 'no-usb-drive'>`.

On error, we'd return `err('no-log-directory')`, `err('no-usb-drive')`, etc. And on success, we'd return `ok()`.

Consumers could then condition on `result.isOk()`.

libs/ui/src/export_logs_modal.tsx
```diff
@@ -101,6 +103,19 @@ export function ExportLogsModal({
...
 
+    if (logFileType === LogFileType.Raw && onExportLogs && !openFileDialog) {
+      const result = await onExportLogs(logFileType);
+      await logger.log(LogEventId.FileSaved, userRole, {
+        disposition: result ? 'success' : 'failure',
+        message: result
+          ? 'Sucessfully saved logs on the usb drive.'
+          : 'Failed to save logs to usb drive.',
+        fileType: 'logs',
+      });
```
arsalansufi, line 114: I've been moving more and more event logging from the frontend to the backend, when possible. The `createLogsApi` function that you've added could take a logger as an input and then handle the logging internally.

My thinking is that every time we call the backend's log-exporting function, we want to emit this log, so might as well colocate the log instead of requiring consumers to remember to log. Also avoids cluttering the frontend.

But I don't feel too strongly about this and am okay with leaving as is!

arsalansufi (commented):
Looks great! I have one question for Jonah about our latest recommended react-query patterns and two other smaller thoughts

libs/backend/src/logs/logs_api.ts
benadida: yep, I definitely had on my list a "need to fix this" but wasn't sure yet how to do it, so your pointer is perfect.

libs/ui/src/export_logs_modal.tsx
arsalansufi: Ah got it! Def feel free to time-box. And a good reminder of another challenge that we need to figure out around backend code sharing patterns

libs/types/src/logs_api.ts
```diff
@@ -0,0 +1,10 @@
+import { Result } from '@votingworks/basics';
```
jonahkagan, line 1: Instead of defining these types in libs/types, the pattern recommended for grout is to colocate them with the code that creates the API.

Also, you shouldn't define an interface for the API, but rather let grout infer it from the API object itself.

See https://github.com/votingworks/vxsuite/pull/4215

apps/scan/frontend/src/screens/election_manager_screen.test.tsx
```diff
@@ -50,6 +52,11 @@ afterEach(() => {
 function renderScreen(
   props: Partial<ElectionManagerScreenProps> = {}
 ): RenderResult {
+  // eslint-disable-next-line @typescript-eslint/require-await
+  const defaultDoExportLogs: () => Promise<LogsResultType> = async () => {
+    return ok();
+  };
+  const doExportLogs = props.doExportLogs || defaultDoExportLogs;
```
jonahkagan, line 59: A simpler way to make a mock function is:

`jest.fn().mockResolvedValue(ok())`

apps/admin/frontend/src/screens/logs_screen.test.tsx
```diff
@@ -20,28 +20,15 @@ afterEach(() => {
...
-
-  // Log saving is tested fully in src/components/export_logs_modal.test.tsx
-  userEvent.click(screen.getByText('Save Log File'));
-  await screen.findByText('No Log File Present');
-  userEvent.click(screen.getByText('Close'));
-
-  expect(
-    screen.getByText('Save CDF Log File').closest('button')
-  ).toHaveAttribute('disabled');
+  fireEvent.click(screen.getByText('Save Log File'));
```
jonahkagan, line 31: Use `userEvent` - `fireEvent` is deprecated

apps/admin/frontend/src/screens/logs_screen.tsx
jonahkagan: Yes, that's still the recommendation. This seems like a case where the consuming code needs to await the mutation, so using `mutateAsync` makes sense. But it does still need to be wrapped in a try/catch to avoid unhandled promise crashes.

Also, I would say you don't need the `useCallback` here - I don't think there will be any performance issue without it.

apps/scan/frontend/src/screens/election_manager_screen.test.tsx
benadida: whoops, yeah, of course, not sure why I didn't use a jest mock here.

libs/ui/src/export_logs_modal.tsx
```diff
@@ -54,114 +42,34 @@ export function ExportLogsModal({
...
-      await logger.log(LogEventId.FileSaved, userRole, {
-        disposition: 'failure',
-        message: `Error saving log file: ${error.message}`,
-        result: 'File not saved, error message shown to user.',
-        fileType: 'logs',
-      });
-      setCurrentState(ModalState.Error);
+
+      if (result.isErr()) {
+        setErrorMessage(result.err());
```
jonahkagan, line 68: Do we want to expose the raw error message tags to the user (see L172 below)? I think it's ok but just wanted to check that it was intentional. Other options would be to turn it into a user-friendly message or just remove it from the UI completely.

apps/scan/frontend/src/app_root.tsx
```diff
@@ -71,6 +72,17 @@ export function AppRoot({
   const transitionPollsMutation = transitionPolls.useMutation();
   const unconfigureMutation = unconfigureElection.useMutation();
 
+  const exportLogsToUsbMutationAsync =
+    exportLogsToUsb.useMutation().mutateAsync;
```
jonahkagan, line 76: Nit: for stylistic consistency:

```ts
const exportLogsToUsbMutation = exportLogsToUsb.useMutation()`

...

// in doExportLogs
return await exportLogsToUsbMutation.mutateAsync()
```

apps/scan/frontend/src/screens/election_manager_screen.tsx
```diff
@@ -376,6 +379,10 @@ export function ElectionManagerScreen({
 /* istanbul ignore next */
 export function DefaultPreview(): JSX.Element {
   const { electionDefinition } = usePreviewContext();
+  // eslint-disable-next-line @typescript-eslint/require-await
+  const doExportLogs: () => Promise<LogsResultType> = async () => {
+    return ok();
+  };
```
jonahkagan, line 385: Another mock to turn into `jest.fn()`

jonahkagan (approved):

libs/ui/src/export_logs_modal.tsx
benadida: yeah that was intentional. I could make it nicer but it's going to be rare enough in production that it didn't feel necessary.

apps/scan/frontend/src/screens/election_manager_screen.tsx
benadida: this one's not a test, it's an actual default but maybe the issue here is we shouldn't have this be a reasonable default. I think there was a reason I did this but... let me see if I can undo it.

benadida: oh wait.. it's a defaultpreview.... maybe it is a test? what's the right way to think about this? Like a test?

jonahkagan: Oh whoops, I didn't read closely. I think it's fine how you have it

jonahkagan: This is for the "preview mode" where you can view screens of the app directly by going to `/preview`

adghayes:
@benadida @jonahkagan I know I'm late to this party, but just noticing that the frontends now have `@votingworks/backend` as a dependency. Shouldn't that be a `devDependency`, with `import type`?

benadida:
> @benadida @jonahkagan I know I'm late to this party, but just noticing that the frontends now have `@votingworks/backend` as a dependency. Shouldn't that be a `devDependency`, with `import type`?

maybe some typescript magic I don't understand there! Say more?

jonahkagan:
Yes @adghayes, nice catch. @benadida you can see a bit about the pattern in the grout docs: https://github.com/votingworks/vxsuite/tree/main/libs/grout#client. The basic idea is that since the types are only needed at compile time, you can use `import type` to avoid loading in the actual dependency values at runtime. It mostly matters for us in cases where we can't load the values (e.g. client-side code trying to load a module with Node deps). We've worked around that by stubbing server-side code as well, so in this case it still worked.

adghayes:
@benadida fixed it in https://github.com/votingworks/vxsuite/pull/4248

PR #4222, opened 2023-11-16
VxScan Report Reprinting

### LINKED ISSUE #2839: VxScan: persistent additional report printing function
I know we just decided not to do this with sus-polls but I still think we should provide a way to print extra report copies from the pollworker screen to help this situation (after polls open before any ballots are cast, or after polls closed) : https://votingworks.slack.com/archives/C02Q3M874SV/p1667912860296459?thread_ts=1667912600.790539&amp;cid=C02Q3M874SV

Issue automatically created from slack feedback reported by caroline 
https://votingworks.slack.com/archives/CJU9MSC6S/p1667931764800919?thread_ts=1667931764.800919&cid=CJU9MSC6S

adghayes:
## Overview

Closes #2839.

## Demo Video or Screenshot

Depicts reprinting after polls closed, but it could happen at any stage.

https://github.com/votingworks/vxsuite/assets/37960853/059c66c6-38be-41b0-be5f-03cdba751c67

## Testing Plan

New automated testing + manual testing.

## Checklist


apps/scan/backend/schema.sql
```diff
@@ -6,6 +6,9 @@ create table election (
   precinct_selection text,
   is_test_mode boolean not null default true,
   polls_state text not null default "polls_closed_initial",
+  last_polls_transition_type text,
+  last_polls_transition_time integer,
+  last_polls_transition_ballot_count integer,
```
adghayes, line 11: There are a few ways to do this. Could split it out to its own table. Could keep track of all transitions in another table and infer the current state from the last transition. This seemed simplest to start.

apps/scan/backend/src/types.ts
```diff
@@ -70,3 +70,18 @@ export interface PrecinctScannerStateMachine {
...
+}
+
+export type PrecinctScannerPollsInfo =
+  | {
+      pollsState: Extract<PollsState, 'polls_closed_initial'>;
+    }
+  | {
+      pollsState: Exclude<PollsState, 'polls_closed_initial'>;
+      lastPollsTransition: PollsTransition;
+    };
```
adghayes, line 87: This is why I renamed the type in the previous commit. Naming is hard.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -246,30 +178,42 @@ export function PollWorkerScreen({
         return;
       }
 
-      const timePollsTransitioned = Date.now();
-      setCurrentPollsTransition(pollsTransition);
+      const pollsTransitionTime = Date.now();
+      await transitionPollsMutation.mutateAsync({
+        type: pollsTransitionType,
+        time: pollsTransitionTime,
+      });
```
adghayes, line 185: I'm moving this mutation up in the order of events. Previoiusly it was:

1. Print Report
2. Export CVRs
3. Change Polls State

Part of the reasoning for this was, I believe, that printing the report is so essential that we didn't want to change state without guaranteeing we could do it. But with the ability to reprint reports and the existing ability to re-export CVRs, it's lower risk and I think this is a better order of operations. It makes more sense - for example, the last batch in the CVR export now has an end timestamp, as it should, where previously it may have appeared in the file like it had not ended. cc @arsalansufi I know I've brought up that last bit before.

apps/scan/frontend/src/utils/get_page_count.ts
```diff
@@ -4,6 +4,16 @@ import { Buffer } from 'buffer';
...
+  // mock for tests
+  if (process.env.NODE_ENV === 'test') {
+    return Promise.resolve(1);
+  }
+
+  // mock if doing development in the browser
+  if (!window.kiosk) {
+    return Promise.resolve(1);
+  }
+
```
adghayes, line 16: The way that we get the page count here is a little overdone - we reprint the report as a PDF through `kiosk-browser` and use `pdfjs` to get a page count. It doesn't work in tests because `kiosk.printToPDF` doesn't work. I've just added mocks so I can use this in the browser or tests without hanging on this.

apps/scan/frontend/src/utils/print_report.tsx
adghayes, line 1: This file is pretty much copied from `poll_worker_screen`.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -438,41 +382,42 @@ export function PollWorkerScreen({
...
+                  )} of paper`
+                : 'paper'}{' '}
+              into the printer to print the report.
             </P>
-          </Prose>
+          )}
+          <P>
+            Remove the poll worker card once you have printed all necessary
+            reports.
+          </P>
```
adghayes, line 408: Changed the copy on this line due to a @mattroe request.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -438,41 +382,42 @@ export function PollWorkerScreen({
...
-          throwIllegalValue(currentPollsTransition);
+          throwIllegalValue(pollsTransitionType);
       }
     })();
 
     return (
       <ScreenMainCenterChild infoBarMode="pollworker">
         <CenteredLargeProse>
-          <H1>{pollsTransitionCompleteText}</H1>
-          <Prose themeDeprecated={fontSizeTheme.medium}>
```
adghayes, line 449: Death to `Prose`!

adghayes: I didn't notice any changes after removing it.

apps/scan/frontend/test/helpers/mock_api_client.tsx
```diff
@@ -50,13 +52,70 @@ export const statusNoPaper: PrecinctScannerStatus = {
...
+function createMockApiClient(): MockApiClient {
+  const mockApiClient = createMockClient<Api>();
+  // For some reason, using an object spread to override methods breaks the rest
+  // of the mockApiClient, so we override like this instead
+  (mockApiClient.transitionPolls as unknown as jest.Mock) = jest.fn(() => {
+    throw new Error('transitionPolls not mocked');
+  });
+  return mockApiClient as unknown as MockApiClient;
+}
+
```
adghayes, line 111: I tried to avoid introducing something like this, but I couldn't correctly mock the timestamp that is required on transitioning polls. Even if you mock the system time via `jest`, it will mock advancing time throughout the test so the actual timestamp will have drifted from the initial mock.

apps/scan/frontend/src/utils/get_page_count.ts
adghayes: I'm OK leaving it for now - and leaving a number of other convoluted aspects of this PR - to be resolved by eventual printing from the backend. 

Part of the whole challenge here is setting up the frontend with all the information it needs to print the report, while the backend already has it.

mattroe:
UX / app flow LGTM!

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -489,8 +434,23 @@ export function PollWorkerScreen({
...
+  assertFalsy(pollWorkerFlowState);
+
   const commonActions = (
     <React.Fragment>
+      {pollsState !== 'polls_closed_initial' && (
+        <ReprintReportButton
+          printerInfo={printerInfo}
+          scannedBallotCount={scannedBallotCount}
+          lastPollsTransition={lastPollsTransition}
+          reportLabel="Previous"
```
jonahkagan, line 447: Could we put the actual name of the report in this label easily? E.g. in the case you showed in the video I think it would be clearer for this button to say "Print Polls Closed Report"

apps/scan/frontend/src/utils/get_page_count.ts
jonahkagan: Sounds good, thanks for the context

apps/scan/frontend/src/components/reprint_report_button.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { getConfig, getMachineConfig, getScannerResultsByParty } from '../api';
+import { printReport } from '../utils/print_report';
+
+interface ReprintReportButtonProps {
+  printerInfo?: KioskBrowser.PrinterInfo;
+  lastPollsTransition?: PollsTransition;
+  scannedBallotCount: number;
+  reportLabel: 'Additional' | 'Previous';
+  beforePrint: () => void;
+  afterPrint: (numPages: number) => void;
```
jonahkagan, line 13: I'm not sure I see the benefit of abstracting the data loading and logic for reprinting into this special button component, since the poll worker screen already has the data needed for printing polls transition reports in general.

It's a little confusing to track the control flow with the `beforePrint` and `afterPrint` callbacks. I think the previous approach of having a function defined in the poll worker screen component to do the printing logic was a bit more straightforward.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -154,88 +145,29 @@ export function PollWorkerScreen({
   const [pollWorkerFlowState, setPollWorkerFlowState] = useState<
     Optional<PollWorkerFlowState>
   >(initialPollWorkerFlowState());
-  const [currentPollsTransition, setCurrentPollsTransition] =
-    useState<PollsTransition>();
-  const [currentPollsTransitionTime, setCurrentPollsTransitionTime] =
-    useState<number>();
+
+  // optimistically set lastPollsTransition based on mutation input if it exists
```
jonahkagan, line 149: Why do we need the optimistic `lastPollsTransition` here?

apps/scan/backend/schema.sql
jonahkagan: Yeah this seems fine. One thing to note is that now the `polls_state` field is actually duplicating some of the information held in `last_polls_transition_type`. I think it's still fine to store it for clarity's sake, but thought I'd call it out.

apps/scan/frontend/test/helpers/mock_api_client.tsx
jonahkagan: One workaround might be to have a util function `getCurrentTimestamp` that wraps the `Date` API. Then you can mock that function in test and control it more precisely.

If you want to keep it like this, that's fine too, but maybe add a comment explaining why

jonahkagan (commented):
Left a few comments I'd like to at least hear your thoughts on before approving

apps/scan/frontend/src/api.ts
```diff
@@ -185,6 +195,8 @@ export const setPrecinctSelection = {
     return useMutation(apiClient.setPrecinctSelection, {
       async onSuccess() {
         await queryClient.invalidateQueries(getConfig.queryKey());
+        // changing the precinct selection after polls open will reset polls to closed
+        await queryClient.invalidateQueries(getPollsInfo.queryKey());
```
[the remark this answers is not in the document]
adghayes: Yes, great catch, thank you. I guess the advantage of having the `Config` bundled is that we just refreshed everything all the time.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: It used to say that, but I removed it just to keep it simple. Yeah, I can totally add it back it.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: Because after the printing page, we land on the `transition_complete` page which needs the current transition to show accurate text. The other query will almost certainly be complete, but I don't want to bet on it.

jonahkagan: Maybe there's a way to encode that constraint more explicitly? E.g. block leaving the printing page until the mutation completes and the new transition info has loaded?

(Or at the very least a comment explaining this in a bit more detail would be helpful)

apps/scan/frontend/src/components/reprint_report_button.tsx
jonahkagan: I like the idea of breaking this screen into smaller components. I wonder if there's a way to do it with screens rather than buttons though. Anyway, not a blocking issue

adghayes: Yeah, each part of the flow could be a different screen. That would probably be best. And it will become a heck of a lot easier once certain pieces (*cough* printing) are on the backend and we can just hit endpoints rather than have elaborate functions.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: I think that sort of block would require doing some sort of `await reactQuerySettledIfNotAlreadySetted(...)`, which is something I've wished for in the past but never found. And it's even a little more convoluted here because we're waiting on both a mutation and the subsequent query. I think you'd have to do some sort of `waitFor` which I wanted to avoid.

For now I'll just add a comment!

jonahkagan: Mutations can automatically wait for subsequent queries by awaiting the call to `queryClient.invalidateQueries` in the `onSuccess` handler, for future reference. So all you have to do is await the mutation to ensure that the invalidated queries are updated.

adghayes: But does that wait for those queries to be invalidated or wait until they've been invalidated and successfully refeched?

apps/scan/frontend/test/helpers/mock_api_client.tsx
adghayes: 4d26214f25eb45df7e0ed156a9c80af4454a232d, thanks for the suggestion!

apps/scan/frontend/src/api.ts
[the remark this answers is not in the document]
adghayes: dfe9f388c

apps/scan/frontend/src/screens/poll_worker_screen.tsx
jonahkagan: I'm pretty sure it waits for them to be refetched: https://tkdodo.eu/blog/mastering-mutations-in-react-query#awaited-promises. But I could be wrong

adghayes: I just did some testing, and I don't think they do refetch in time. I removed the optimistic updating and the tests failed because we assert that the last polls transition exists before we render the transition screens, but it did not exist. If the refetching actually completed before `await mutateAsync` resolved, then it should just work as-is without the optimistic updating.

adghayes: comment: 64525cc64

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: fa6773de1

https://github.com/votingworks/vxsuite/assets/37960853/61caa631-35ee-48a0-9c77-f6e81e65ee6d

adghayes: fa6773de1

https://github.com/votingworks/vxsuite/assets/37960853/61caa631-35ee-48a0-9c77-f6e81e65ee6d

apps/scan/frontend/src/screens/poll_worker_screen.tsx
jonahkagan: Dang, that's really surprising and unfortunate

jonahkagan (approved):

Issue #4025, opened 2023-10-03
WIA: visually clarify selected write-in


carolinemodic:
the write-in highlights are a little tall, so when there are two write-ins in a row, it could be hard to know for sure which one I'm transcribing. Two thoughts:
• we could make it less tall
• we could highlight in color the filled bubble that we identified, so there's clarity as to which one we're transcribing.

Issue automatically created from slack feedback reported by ben 
https://votingworks.slack.com/archives/CJU9MSC6S/p1682451531541429?thread_ts=1682451531.541429&cid=CJU9MSC6S

mattroe:
https://votingworks.slack.com/archives/CEL6D3GAD/p1699987754781639?thread_ts=1699914454.155639&cid=CEL6D3GAD

PR #4223, opened 2023-11-16
Improve WIA Cropping

adghayes:
## Overview

#4025 

## Demo Video or Screenshot

| Before  | After |
| Screen Shot 2023-11-13 at 4 42 03 PM  | Screen Shot 2023-11-16 at 2 03 45 PM  |
| Screen Shot 2023-11-13 at 4 40 33 PM |  Screen Shot 2023-11-16 at 1 37 11 PM  | 

## Testing Plan

Manual mostly, updated snapshots / fixtures.

## Checklist


libs/hmpb/layout/src/layout.ts
```diff
@@ -1687,7 +1687,7 @@ function layOutBallotHelper({
         bottom: 1,
         left: 1,
         right: 9,
-        top: 2,
+        top: 1,
```
adghayes, line 1690: We were cropping a little high on the VxBallots, see difference between highlight and desired area:

Screen Shot 2023-11-13 at 4 42 13 PM

The cropping too low is dealt with in the interpreter. The above is correct - it says `1` and we want `1`. That off-by-one issue is in the interpreter.

libs/types/src/cdf/ballot-definition/fixtures.ts
adghayes, line 1: This is not necessary but thought I'd make it consistent.

libs/ballot-interpreter/src/hmpb-rust/layout.rs
```diff
@@ -33,11 +33,9 @@ fn build_option_layout(
     let column_offset = -(grid_layout.option_bounds_from_target_mark.left as i32);
     let row_offset = -(grid_layout.option_bounds_from_target_mark.top as i32);
     let width: GridUnit = grid_layout.option_bounds_from_target_mark.left
-        + grid_layout.option_bounds_from_target_mark.right
-        + 1;
+        + grid_layout.option_bounds_from_target_mark.right;
     let height: GridUnit = grid_layout.option_bounds_from_target_mark.top
-        + grid_layout.option_bounds_from_target_mark.bottom
-        + 1;
+        + grid_layout.option_bounds_from_target_mark.bottom;
```
adghayes, line 38: Here's the off-by-one issue. We're adding in 1 to the width and the height, in addition to the `top + bottom` or `left + right` specified by the `gridLayouts`. I imagine the original thought here was to account for the size of the bubble itself? But we don't need to do that, the measurements are from the center of the bubble in all directions.

jonahkagan: Weird, I don't know why we would have done that! Nice catch

jonahkagan (approved):

PR #4224, opened 2023-11-16
fix(auth): correct card interface APIs

eventualbuddha:
## Overview

I noticed that the interfaces for `Card` and `CommonAccessCardCompatibleCard` didn't actually include all the methods that they should have. I added a `BaseCard` interface to capture the `disconnect` method, then fixed `CommonAccessCardCompatibleCard` to include `PinProtectedCard` to ensure it has `checkPin`.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated type checking.

### DIFF
--- a/libs/auth/src/card.ts
+++ b/libs/auth/src/card.ts
@@ -100,6 +100,13 @@ export type CheckPinResponse =
   | CheckPinResponseIncorrect;
 
 /**
+ * The base API any card should support
+ */
+export interface BaseCard {
+  disconnect(): Promise<void>;
+}
+
+/**
  * The API for a card that can provide its status
  */
 export interface StatefulCard<T = CardDetails> {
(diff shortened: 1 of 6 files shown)


libs/auth/src/card.ts
```diff
@@ -92,6 +92,13 @@ interface CheckPinResponseIncorrect {
   numIncorrectPinAttempts: number;
 }
 
+/**
+ * The base API any card should support
+ */
+export interface BaseCard {
+  disconnect(): Promise<void>;
+}
```
arsalansufi, line 100: nit: Could you move this below `CheckPinResponse` so that `CheckPinResponse` comes right after `CheckPinResponseCorrect` and `CheckPinResponseIncorrect`?

arsalansufi (approved):
Good call!

PR #4225, opened 2023-11-16
Add sheet-copying scripts to facilitate scale testing

arsalansufi:
## Overview

In preparation for upcoming QA, I'm committing a script that I've been using to facilitate my own scale testing. The script copies a scanner store's sheet records to quickly bump the sheet count up without having to physically scan ballots.

### Gotchas

Because the script is copying sheets rather than creating them from scratch, a first sheet does have to be physically scanned before the script can be used. The script reminds you of this, if relevant:

```
$ ./copy-sheets 10
❌ Scanner store should contain at least one sheet to copy
```

-----

For safety, I added a check to ensure that the script cannot be run in production:

```
$ NODE_ENV=production ./copy-sheets 
❌ Cannot be run in production
```

-----

And finally, the script does not cover continuous export to the USB drive, but it does surface the CVR re-sync modal on completion.

### Update 11/20

I've also added a script to facilitate scale testing on VxCentralScan. This script copies a specified batch and all its sheets. Here's the usage message for both scripts.

_VxScan script_

```
Usage: copy-sheets <target-sheet-count>

Arguments:
  <target-sheet-count>	The target total sheet count after copying
```

_VxCentralScan script_

```
Usage: copy-batch '<batch-name>' <num-copies>

Arguments:
  <batch-name>	The name of the batch to copy
  <num-copies>	The number of copies to create
```

## Demo Video

https://github.com/votingworks/vxsuite/assets/12616928/608a9175-2e2a-4e22-837a-b1bb38b713c4

## Testing Plan

- [x] Tested manually

### DIFF
--- a/apps/scan/backend/scripts/.eslintrc.json
+++ b/apps/scan/backend/scripts/.eslintrc.json
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "no-console": "off"
+  }
+}
--- a/apps/scan/backend/scripts/.eslintrc.json
+++ b/apps/scan/backend/scripts/.eslintrc.json
@@ -1,5 +1,0 @@
-{
-  "rules": {
-    "no-console": "off"
-  }
-}
--- a/apps/scan/backend/scripts/copy_sheets.ts
+++ b/apps/scan/backend/scripts/copy_sheets.ts
@@ -91,7 +91,6 @@ export function main(args: readonly string[]): void {
   try {
     checkEnvironment();
     copySheets(parseCommandLineArgs(args));
-    process.exit(0);
   } catch (error) {
     console.error(`❌ ${extractErrorMessage(error)}`);
     process.exit(1);
--- a/apps/scan/backend/scripts/copy_sheets.ts
+++ b/apps/scan/backend/scripts/copy_sheets.ts
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import path from 'path';
 import { v4 as uuid } from 'uuid';
 import { AcceptedSheet } from '@votingworks/backend';
 import {
@@ -40,8 +41,14 @@ function copySheet(store: Store, sheet: AcceptedSheet): void {
   const newSheet: AcceptedSheet = {
     ...sheet,
     id: newSheetId,
-    frontImagePath: sheet.frontImagePath.replace(sheet.id, newSheetId),
-    backImagePath: sheet.backImagePath.replace(sheet.id, newSheetId),
+    frontImagePath: path.join(
+      path.dirname(sheet.frontImagePath),
+      `${newSheetId}-front.jpg`
+    ),
+    backImagePath: path.join(
+      path.dirname(sheet.backImagePath),
+      `${newSheetId}-back.jpg`
+    ),
   };
 
   fs.copyFileSync(sheet.frontImagePath, newSheet.frontImagePath);
@@ -93,8 +100,9 @@ function copySheets({ targetSheetCount }: CopySheetsInput): void {
     copySheet(store, sheet);
   }
 
+  const sheetOrSheets = numSheetsToCreate === 1 ? 'sheet' : 'sheets';
   console.log(
-    `✅ Created ${numSheetsToCreate} new sheets by copying existing sheets, ` +
+    `✅ Created ${numSheetsToCreate} new ${sheetOrSheets} by copying existing sheets, ` +
       `bringing total sheet count to ${targetSheetCount}`
   );
 }
--- a/apps/scan/backend/scripts/copy_sheets.ts
+++ b/apps/scan/backend/scripts/copy_sheets.ts
@@ -1,5 +1,4 @@
 import fs from 'fs';
-import path from 'path';
 import { v4 as uuid } from 'uuid';
 import { AcceptedSheet } from '@votingworks/backend';
 import {
@@ -11,6 +10,7 @@ import { safeParseNumber } from '@votingworks/types';
 
 import { SCAN_WORKSPACE } from '../src/globals';
 import { Store } from '../src/store';
+import { createWorkspace } from '../src/util/workspace';
 
 const usageMessage = `Usage: copy-sheets <target-sheet-count>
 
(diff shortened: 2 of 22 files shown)


apps/scan/backend/scripts/copy_sheets.ts
```diff
@@ -0,0 +1,99 @@
...
+    {
+      imagePath: newSheet.backImagePath,
+      interpretation: newSheet.interpretation[1],
+    },
+  ]);
+}
+
+function copySheets({ targetSheetCount }: CopySheetsInput): void {
+  const store = Store.fileStore(
+    path.join(assertDefined(SCAN_WORKSPACE), 'ballots.db')
```
arsalansufi, line 63: I _think_ that this script may just work for VxCentralScan, too, if you provide a `SCAN_WORKSPACE` env var that points it to VxCentralScan's workspace directory. I don't have my Fujitsu on me right now to test, but will soon.

arsalansufi: Actually, on second thought, this won't just work for VxCentralScan, as we need to instantiate a VxCentralScan `Store` instance, not a VxScan `Store` instance, which is fine. That was wishful thinking hah

apps/scan/backend/scripts/.eslintrc.json
eventualbuddha, line 1: I don't have a strong preference, but consider putting this as an `overrides` in the package's main `.eslintrc.json`.

apps/scan/backend/scripts/copy_sheets.ts
```diff
@@ -0,0 +1,99 @@
...
+  return { targetSheetCount };
+}
+
+function copySheet(store: Store, sheet: AcceptedSheet): void {
+  const newSheetId = uuid();
+  const newSheet: AcceptedSheet = {
+    ...sheet,
+    id: newSheetId,
+    frontImagePath: sheet.frontImagePath.replace(sheet.id, newSheetId),
+    backImagePath: sheet.backImagePath.replace(sheet.id, newSheetId),
```
eventualbuddha, line 43: I'm not sure we want to assume that the image path has the sheet ID in it here, even though that is the case right now. Consider either calling the same logic that generates the path in the first place _or_ just generates a unique name in the same directory?

apps/scan/backend/scripts/copy_sheets.ts
```diff
@@ -0,0 +1,99 @@
...
+      imagePath: newSheet.backImagePath,
+      interpretation: newSheet.interpretation[1],
+    },
+  ]);
+}
+
+function copySheets({ targetSheetCount }: CopySheetsInput): void {
+  const store = Store.fileStore(
+    path.join(assertDefined(SCAN_WORKSPACE), 'ballots.db')
+  );
```
eventualbuddha, line 64:
```suggestion
  const workspace = createWorkspace(assertDefined(SCAN_WORKSPACE));
  const store = workspace.store;
```

apps/scan/backend/scripts/copy_sheets.ts
```diff
@@ -0,0 +1,99 @@
...
+  assert(
+    sheets.length > 0,
+    'Scanner store should contain at least one sheet to copy'
+  );
+  assert(
+    targetSheetCount > sheets.length,
+    `Target sheet count should be greater than current sheet count (${sheets.length})`
+  );
+
+  const numSheetsToCreate = Math.max(targetSheetCount - sheets.length, 0);
```
eventualbuddha, line 76: You've asserted that `targetSheetCount > sheets.length`, so no need for the `Math.max`.
```suggestion
  const numSheetsToCreate = targetSheetCount - sheets.length;
```

apps/scan/backend/scripts/copy_sheets.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * A script for copying a scanner store's sheet records to facilitate scale testing
+ */
+export function main(args: readonly string[]): void {
+  try {
+    checkEnvironment();
+    copySheets(parseCommandLineArgs(args));
+    process.exit(0);
```
eventualbuddha, line 94: Unnecessary `process.exit`? In general it's best to let the process exit naturally in case there are any e.g. buffers to flush etc.

apps/scan/backend/scripts/copy_sheets.ts
```diff
@@ -0,0 +1,99 @@
...
+  assert(process.env.NODE_ENV !== 'production', 'Cannot be run in production');
+}
+
+function parseCommandLineArgs(args: readonly string[]): CopySheetsInput {
+  if (args.length !== 1 || !safeParseNumber(args[0]).isOk()) {
+    console.log(usageMessage);
+    process.exit(0);
+  }
+  const targetSheetCount = safeParseNumber(args[0]).unsafeUnwrap();
+  return { targetSheetCount };
```
eventualbuddha, line 34: Consider:
- using `safeParseInt` instead of `safeParseNumber`
- only parsing once
- using type narrowing instead of `unsafeUnwrap`
- printing the usage message to stderr instead of stdout
- exiting with a non-zero exit status so e.g. a scripting-based caller knows it failed
```suggestion
  const parseTargetSheetCountResult = safeParseInt(args[0]);
  if (parseTargetSheetCountResult.isErr()) {
    process.stderr.write(`${usageMessage}\n`);
    process.exit(1);
  }
  return { targetSheetCount: parseTargetSheetCountResult.ok() };
```

apps/scan/backend/scripts/copy_sheets.ts
```diff
@@ -0,0 +1,99 @@
...
+      interpretation: newSheet.interpretation[1],
+    },
+  ]);
+}
+
+function copySheets({ targetSheetCount }: CopySheetsInput): void {
+  const store = Store.fileStore(
+    path.join(assertDefined(SCAN_WORKSPACE), 'ballots.db')
+  );
+  const sheets = Array.from(store.forEachAcceptedSheet());
```
eventualbuddha, line 65: This could load quite a lot of data. Consider getting the length first, determining the number of sheets to create, then loading only as many as might be needed but up to some limit, like 500.

eventualbuddha (approved):
I'm happy to see tooling to make larger volume testing of software features easier! I have a few suggestions for improvements, but this mostly LGTM.

apps/scan/backend/scripts/.eslintrc.json
arsalansufi: 👍 Updated `apps/scan/backend` as well as other packages where I've used the same pattern: https://github.com/votingworks/vxsuite/pull/4225/commits/5f078110543495fa2777980eed689bb3d4244ba7

apps/scan/backend/scripts/copy_sheets.ts
arsalansufi: Ah good catch! Forgot to remove the `Math.max` after I added the assertion: https://github.com/votingworks/vxsuite/pull/4225/commits/ce8c0a3368544c2ca5f2a19c34551aeadc8156d5

apps/scan/backend/scripts/copy_sheets.ts
arsalansufi: Ah makes sense re flushing buffers, etc. - updated this script and all others that I've written where I've used an explicit `process.exit(0)`: https://github.com/votingworks/vxsuite/pull/4225/commits/a15b9fa677d75598c95fbcb460bc5f2bb4f0cbcd

apps/scan/backend/scripts/copy_sheets.ts
arsalansufi: Good call: https://github.com/votingworks/vxsuite/pull/4225/commits/abb2dbdd663c5bad282cdef21c54d088f7d6737f

apps/scan/backend/scripts/copy_sheets.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/4225/commits/abb2dbdd663c5bad282cdef21c54d088f7d6737f

apps/scan/backend/scripts/copy_sheets.ts
arsalansufi: 👍 https://github.com/votingworks/vxsuite/pull/4225/commits/3a94d52cd7682d9c2b56425f0a71eddf789d5043

apps/scan/backend/scripts/copy_sheets.ts
arsalansufi: Good call: https://github.com/votingworks/vxsuite/pull/4225/commits/c083493de8335a3e153306abbc43fa484aded67f

PR #4226, opened 2023-11-16
fix(auth): handle mock CAC card bad PIN responses

eventualbuddha:
## Overview

Our mock CAC cards act the same as our VxSuite cards when receiving an invalid PIN. Instead of responding with a "security condition not satisfied" code, they respond with "counter has reached the value X".

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests. @benadida to put up a parallel PR that will enable me to manually test this (or he can).


arsalansufi (approved):
LGTM!

PR #4227, opened 2023-11-16
add scripts and capabilities for building a simulated CAC card out of a JavaCard

benadida:
Allow us to program a JavaCard to mimick a CAC card as best as possible. Important changes:
- use the same key slots as the CAC card
- use RSA2048 instead of ECC

all the changes are downstream from those two requirements.

### DIFF
--- a/libs/auth/scripts/configure_simulated_cac_card.ts
+++ b/libs/auth/scripts/configure_simulated_cac_card.ts
@@ -1,0 +1,292 @@
+import { Buffer } from 'buffer';
+import { existsSync } from 'fs';
+import path from 'path';
+import { extractErrorMessage } from '@votingworks/basics';
+import { Byte } from '@votingworks/types';
+
+import { CommandApdu, constructTlv } from '../src/apdu';
+import { getRequiredEnvVar } from '../src/env_vars';
+import {
+  MAX_NUM_INCORRECT_PIN_ATTEMPTS,
+  OPEN_FIPS_201_AID,
+  PUK,
+} from '../src/java_card';
+import {
+  construct8BytePinBuffer,
+  CRYPTOGRAPHIC_ALGORITHM_IDENTIFIER,
+} from '../src/piv';
+import { runCommand } from '../src/shell';
+import { waitForReadyCardStatus } from './utils';
+
+import { CARD_DOD_CERT, CommonAccessCard, DEFAULT_PIN } from '../src/cac';
+
+const APPLET_PATH = path.join(
+  __dirname,
+  '../applets/OpenFIPS201-v1.10.2-with-vx-mods.cap'
+);
+const GLOBAL_PLATFORM_JAR_FILE_PATH = path.join(__dirname, '../scripts/gp.jar');
+
+/**
+ * CHANGE REFERENCE DATA ADMIN is an OpenFIPS201-specific extension of the PIV-standard CHANGE
+ * REFERENCE DATA command. It allows setting PINs and PUKs without the current value. APDUs must be
+ * sent over a GlobalPlatform Secure Channel.
+ */
+const CHANGE_REFERENCE_DATA_ADMIN = {
+  INS: 0x24,
+  P1: 0xff,
+  P2_PIN: 0x80,
+  P2_PUK: 0x81,
+} as const;
+
+/**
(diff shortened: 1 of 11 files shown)


libs/auth/src/cac/common_access_card.ts
```diff
@@ -108,12 +117,18 @@ export class CommonAccessCard implements CommonAccessCardCompatibleCard {
   private readonly cardReader: CardReader;
   private cardStatus: CardStatus<CommonAccessCardDetails>;
   private readonly customChallengeGenerator?: () => string;
+  private readonly certPath?: string;
```
eventualbuddha, line 120: This is only here for `createAndStoreCert`. Remove this and make it a parameter to that method?

libs/auth/scripts/configure_simulated_cac_card.ts
```diff
@@ -0,0 +1,293 @@
...
+    {
+      source: 'file',
+      path: vxPrivateKeyPath,
+    },
+    commonName
+  );
+}
+
+/**
+ * An initial Java Card configuration script to be run at the Bakery
```
eventualbuddha, line 278:
```suggestion
 * Create a mock Common Access Card for use with RAVE.
```

libs/auth/src/cac/common_access_card_certs.ts
```diff
@@ -107,14 +127,19 @@ export async function parseCert(
...
+      certDetails;
+    return {
+      commonAccessCardId,
+      givenName,
+      middleName,
+      familyName,
+    };
+  } catch {
+    return undefined;
+  }
```
eventualbuddha, line 144: Is this "bail on error" required? I'm guessing it's for when a card isn't configured. Maybe add a debug log in the `catch` block?

```ts
debug('error parsing CAC details from certificate: %s', error);
```

libs/auth/src/cac/common_access_card.ts
```diff
@@ -294,16 +348,134 @@ export class CommonAccessCard implements CommonAccessCardCompatibleCard {
...
+    }
+  }
+
+  /**
+   * Generates an asymmetric key pair on the card. The public key is exported, and the private key
+   * never leaves the card. The returned public key will be in PEM format.
+   *
+   * A PIN must be provided if the specified private key is PIN-gated.
+   */
+  private async generateAsymmetricKeyPair(
```
eventualbuddha, line 373: I think it'd be useful to pull the tag values out as constants, e.g.

```ts
const GENERATE_RSA_PUBLIC_KEY_RESPONSE_TAG = Buffer.of(0x7f, 0x49);
const RSA_PUBLIC_KEY_MODULUS_TAG = 0x81;
const RSA_PUBLIC_KEY_EXPONENT_TAG = 0x82;

const SEQUENCE_TAG = 0x30;
```

etc.

libs/auth/src/cac/common_access_card.ts
```diff
@@ -294,16 +348,134 @@ export class CommonAccessCard implements CommonAccessCardCompatibleCard {
...
+                constructTlv(0x02, exponentValue),
+              ])
+            ),
+          ])
+        ),
+      ])
+    );
+
+    const publicKeyInPemFormat = await publicKeyDerToPem(publicKeyInDerFormat);
+    return publicKeyInPemFormat;
```
eventualbuddha, line 442:
```suggestion
    return await publicKeyDerToPem(publicKeyInDerFormat);
```

eventualbuddha (approved):
I tested this manually and got it to work. Requested changes are minor, though I do think we should make them.

libs/auth/src/cac/common_access_card.ts
benadida: I see your point but... this approach mirrors the approach taken on `JavaCard`, so I think best to leave it for consistency?

libs/auth/src/cac/common_access_card_certs.ts
benadida: yes, it's for the case where the card isn't configured yet and then on insert we read the data and boom.

libs/auth/src/cac/common_access_card.ts
eventualbuddha: Sure, that's fine.

PR #4231, opened 2023-11-17
drew/refactor/mark/remove-frontend-state

adghayes:
## Checklist


apps/mark/frontend/src/app_cardless_voting.test.tsx
adghayes, line 1: I chopped a lot of this file because it makes less sense in a world where state comes from the backend, and it's already covered by `end_to_end` anyway. Focused the tests on the voting parts only.

apps/mark/frontend/src/app_contest_candidate_no_party.test.tsx
adghayes, line 1: Could these `contest` files be tested in `mark-flow-ui`?

adghayes: In a separate PR...

apps/mark/frontend/src/app_data.test.tsx
adghayes, line 1: This seemed like it's covered in other places.

apps/mark/frontend/test/helpers/mock_api_client.tsx
```diff
@@ -193,6 +204,36 @@ export function createApiMock() {
...
+      electionStateRef.current = electionState
+        ? {
+            ...electionStateRef.current,
+            ...electionState,
+          }
+        : initialElectionState;
+      mockApiClient.getElectionState
+        .expectCallWith()
+        .resolves(electionStateRef.current);
+    },
```
adghayes, line 218: To make testing easier on myself, I gave `apiMock` some state so you only have to expect the diffs.

apps/mark/frontend/src/app_single_precinct_election.test.tsx
adghayes, line 1: covered on backend and in `admin_screen` now

libs/ui/src/change_precinct_button.tsx
```diff
@@ -81,6 +90,11 @@ export function ChangePrecinctButton({
     const { value } = event.currentTarget;
     if (!value) return; // in case of blur on placeholder option
 
+    if (value === dropdownCurrentValue) {
+      // blur on current value should be no-op
+      return;
+    }
+
```
adghayes, line 97: It was firing twice in some (but not all) tests, once on click and again on blur.

apps/mark/backend/src/app.ts
```diff
@@ -159,6 +182,64 @@ export function buildApi(
...
+      store.setPollsState(newPollsState);
+
+      assert(newPollsState !== 'polls_closed_initial');
+      const logEvent = (() => {
+        switch (newPollsState) {
+          case 'polls_closed_final':
+            return LogEventId.PollsClosed;
+          case 'polls_paused':
+            if (oldPollsState === 'polls_closed_final') {
+              // logging case handled by ResetPollsToPausedButton
```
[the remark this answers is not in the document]
adghayes: Would be ideal, but since it's in `libs/ui` it's difficult to do typescript-wise. Kofi did something like this for `uiStrings`, but that took a while to set up so I don't want to go there in this PR.

apps/mark/backend/src/app.ts
```diff
@@ -159,6 +182,64 @@ export function buildApi(
...
+      store.setPrecinctSelection(input.precinctSelection);
+      store.setBallotsPrintedCount(0);
+    },
+
+    getElectionState(): ElectionState {
+      return {
+        precinctSelection: store.getPrecinctSelection(),
+        ballotsPrintedCount: store.getBallotsPrintedCount(),
+        isTestMode: store.getTestMode(),
+        pollsState: store.getPollsState(),
```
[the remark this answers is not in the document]
adghayes: I lean toward just leaving the logic separate for now. It's what we do in `scan`, and since it's a read from a table with one row I'm not very concerned about the performance.

apps/mark/backend/src/store.ts
```diff
@@ -163,4 +170,147 @@ export class Store {
...
+      return undefined;
+    }
+
+    const precinctSelectionParseResult = safeParseJson(
+      rawPrecinctSelection,
+      PrecinctSelectionSchema
+    );
+
+    // istanbul ignore next
+    if (precinctSelectionParseResult.isErr()) {
```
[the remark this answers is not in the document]
adghayes: The error that a malformed `PrecinctSelection` is written to the DB.

apps/mark/backend/src/store.ts
```diff
@@ -163,4 +170,147 @@ export class Store {
...
+  /**
+   * Gets the current test mode setting value.
+   */
+  getTestMode(): boolean {
+    const electionRow = this.client.one(
+      'select is_test_mode as isTestMode from election'
+    ) as { isTestMode: number } | undefined;
+
+    if (!electionRow) {
+      // test mode will be the default once an election is defined
```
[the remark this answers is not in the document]
adghayes: It will be the default "once" and "until" the election is defined, so correct either way.

apps/mark/backend/src/store.ts
```diff
@@ -163,4 +170,147 @@ export class Store {
...
+  /**
+   * Gets the current ballots printed count
+   */
+  getBallotsPrintedCount(): number {
+    const electionRow = this.client.one(
+      'select ballots_printed_count as ballotsPrintedCount from election'
+    ) as { ballotsPrintedCount: number } | undefined;
+
+    if (!electionRow) {
+      // 0 will be the default once an election is defined
```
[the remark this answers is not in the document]
adghayes: Same

apps/mark/backend/src/app.ts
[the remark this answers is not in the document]
kshen0: Makes sense

apps/mark/backend/src/store.ts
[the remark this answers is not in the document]
kshen0: Ah I understand now, the comment is pointing out that test mode is the default once an election is defined, so we also make it the default when no election is defined

apps/mark/frontend/test/helpers/mock_api_client.tsx
```diff
@@ -73,6 +80,10 @@ export function createApiMock() {
     );
   }
 
+  const electionStateRef: { current: ElectionState } = {
+    current: initialElectionState,
+  };
```
kshen0, line 85: Any reason to keep this state as a variable? It looks like we only modify this state right before setting the resolved value in `expectGetElectionState`, which already has access to the desired resolved value as an argument. Would it be cleaner to just pass the desired resolved value as an argument like `setUsbDriveStatus`?

adghayes: It maintains a history of "last election state" which allowed me, when writing tests, to only pass in the "diff" of the election state. If we just passed the desired resolved value in a test where the polls status were changing, you'd have pass the `precinctSelection` in every time as well. I think there is some clarity to that but it's also an annoyance and a footgun of a different kind.

kshen0: Ah your comment below makes sense now, thanks

kshen0 (approved):

PR #4244, opened 2023-11-21
Update CI to Debian 12. Remove converter-ms-sems

amcmanus:
This PR was initially scoped to update our CircleCI config to use new Debian 12 images. That simple change resurfaced an issue with converter-ms-sems python dependency, which is more involved under Debian 12 compared to 11. We've been planning to remove that service for awhile, so it was included in this PR. The bulk of that removal was handled by @adghayes.


kofi-q (approved):
LGTM!

PR #4245, opened 2023-11-21
Remove VxMarkScan Frontend State

### LINKED ISSUE #3899: Migrate VxMark and VxMarkScan frontend state to backend store

adghayes:
Closes #3899. Removes frontend persistent state from VxMarkScan and therefore from all of VxSuite!


adghayes:
@kshen0 in all the little choices in this PR, I chose to align `mark-scan` as closely to `mark` as I could, even if that meant replacing code with identical code in another part of the file, etc.

apps/mark-scan/frontend/src/pages/admin_screen.tsx
```diff
@@ -24,23 +24,22 @@ import {
...
+  setTestMode,
 } from '../api';
 
 export interface AdminScreenProps {
+  appPrecinct?: PrecinctSelection;
   ballotsPrintedCount: number;
   electionDefinition: ElectionDefinition;
-  isLiveMode: boolean;
-  toggleLiveMode: VoidFunction;
+  isTestMode: boolean;
```
kshen0, line 42: Do we have a preference for whether often-queried values are queried at each component or passed down from `app_root`? Seems like we are passing almost the entire `ElectionState` object down in props. When do we want to query from the subcomponent vs pass down as props?

kshen0 (commented):
Thanks for doing this! 🙏

apps/mark-scan/frontend/src/pages/admin_screen.tsx
adghayes: I would lean toward keeping the changes in this PR as-is because of consistency with `mark`, but I would also think that generally it would be better to move the query down a level rather than pass as props. 

The main disadvantage of this is that you have to insert loading behavior for subcomponents you wouldn't have to otherwise, when the needed data is coming through props. I avoided it for that reason! But you could also just make an assertion that the queries are not empty, making that disadvantage moot.

apps/mark-scan/backend/src/app.test.ts
```diff
@@ -252,3 +208,143 @@ test('usbDrive', async () => {
...
+  const singlePrecinctSelection = singlePrecinctSelectionFor('23');
+  await apiClient.setPrecinctSelection({
+    precinctSelection: singlePrecinctSelection,
+  });
+  await expectElectionState({
+    precinctSelection: singlePrecinctSelection,
+  });
+});
+
+test('printing a ballot incrementes the printed ballot count', async () => {
```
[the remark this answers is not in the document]
adghayes: a6726eb35

kshen0 (approved):

PR #4247, opened 2023-11-21
Replace NH Overvote Flag

### LINKED ISSUE #3896: Make nh-overvote env variable to precinct scan adjudication reason
New Hampshire has a unique law that does not allow overvotes to be casted into the scanner at all. Ballots are rejected and voters only have the option to return their ballot. Once returned, they are instructed to place in the auxiliary/emergency slot if the ballot is not spoiled and remarked. This frontend restriction is enabled through a build environment flag today `REACT_APP_VX_DISALLOW_CASTING_OVERVOTES=TRUE`.

Having this setting implemented as an environment variable requires us to create a new image and reimage devices if we would like to use/demo VxScan with a traditional overvote experience shown. @ginvdr most recently [ran into this problem](https://votingworks.slack.com/archives/CJU9MSC6S/p1693162875047239) at election center.

Instead of an environment variable, we should treat this as a separate precinct scanner adjudication. Instead of enabling `overvote` in the precinct scanning adjudication reasons, the election package would be populated with `overvote-nh` or `overvote-reject` (open to reason alternative reason names) for the reject-only no-casting experience. NH users should also be able to enable this in VxDesign when specifying election adjudication reasons.

adghayes:
## Overview

Closes #3896. Moves the NH overvote behavior (disallow casting with overvotes) from an environment flag to a system setting configurable from VxDesign. It's an absolutely minimal setting in VxDesign - just a stray checkbox - to get it in quickly for testing and code freeze. In the future, we can set it automatically based on state.

Screen Shot 2023-11-21 at 1 58 17 PM


apps/scan/frontend/src/screens/scan_warning_screen.tsx
```diff
@@ -74,20 +72,21 @@ function ConfirmModal({ content, onConfirm, onCancel }: ConfirmModalProps) {
...
   overvotes,
   undervotes,
 }: MisvoteWarningScreenProps): JSX.Element {
   const returnBallotMutation = returnBallot.useMutation();
   const acceptBallotMutation = acceptBallot.useMutation();
-  const allowCastingOvervotes = !isFeatureFlagEnabled(
-    BooleanEnvironmentVariableName.DISALLOW_CASTING_OVERVOTES
-  );
+  const allowCastingOvervotes =
+    !systemSettings.precinctScanDisallowCastingOvervotes;
```
jonahkagan, line 89: Am I right in thinking that this logic will only come into play if the overvote adjudication reason is turned on? If so, we might want to add some validation to system settings to make sure that you don't turn on `precinctScanDisallowCastingOvervotes` without also having the adjudication reason. Not blocking for this PR though

jonahkagan (approved):

Issue #4164, opened 2023-11-03
Central system: lenovo smartcard reader app support


mattroe:
![Image](https://github.com/votingworks/vxsuite/assets/1476320/811df6ae-db9c-4f72-96bd-c1451f7852b9)

mattroe:
https://votingworks.slack.com/archives/C060FEVB2L8/p1700513990181259

PR #4250, opened 2023-11-21
support both possible thinkpad card readers

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/4164

The smartcard reader got swapped out when we ordered more thinkpads so supporting both the old and new 

## Testing Plan
tested on real hardware

## Checklist


adghayes (approved):
0x2c3e3,0x9563 and 0x0588f,0x9540 are the two pairs, I assume, but logic also allows 0x2c3e3,0x9540, for example.

I assume that doesn't matter but noting in case it does.

PR #4257, opened 2023-11-22
Central system design polish

jonahkagan:
## Overview

Adds a variety of small tweaks to VxCentralSystem and VxAdmin to improve how they use the desktop theme.

## Demo Video or Screenshot
Central Scan - adjudication
Screenshot 2023-11-22 at 11 24 31 AM

Admin - Election Manager - Election screen (formerly Ballots screen)
Screenshot 2023-11-22 at 1 04 58 PM

Admin - Election Manager - tallies and marking as official
https://github.com/votingworks/vxsuite/assets/530106/54dd0dbd-f0c6-4028-9bc2-9bbc808b0148

## Checklist

### DIFF
--- a/apps/admin/frontend/src/components/app_routes.tsx
+++ b/apps/admin/frontend/src/components/app_routes.tsx
@@ -15,8 +15,7 @@ import {
 import type { ManualResultsVotingMethod } from '@votingworks/admin-backend';
 import { AppContext } from '../contexts/app_context';
 import { routerPaths } from '../router_paths';
-import { DefinitionScreen } from '../screens/election_screen';
-import { BallotListScreen } from '../screens/ballot_list_screen';
+import { ElectionScreen } from '../screens/election_screen';
 import { PrintTestDeckScreen } from '../screens/print_test_deck_screen';
 import { UnconfiguredScreen } from '../screens/unconfigured_screen';
 import { TallyScreen } from '../screens/tally_screen';
@@ -44,7 +43,7 @@ import { PrecinctBallotCountReport } from '../screens/reporting/precinct_ballot_
 import { VotingMethodBallotCountReport } from '../screens/reporting/voting_method_ballot_count_report_screen';
 import { FullElectionTallyReportScreen } from '../screens/reporting/full_election_tally_report_screen';
 
-export function ElectionManager(): JSX.Element {
+export function AppRoutes(): JSX.Element {
   const { electionDefinition, configuredAt, auth, hasCardReaderAttached } =
     useContext(AppContext);
   const election = electionDefinition?.election;
@@ -94,7 +93,7 @@ export function ElectionManager(): JSX.Element {
       return (
         <React.Fragment>
           <Switch>
-            <Route exact path={routerPaths.electionDefinition}>
+            <Route exact path={routerPaths.election}>
               <UnconfiguredScreen />
             </Route>
             <Route exact path={routerPaths.settings}>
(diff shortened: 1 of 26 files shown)


apps/admin/frontend/src/components/app_routes.tsx
adghayes, line 1: someone finally renamed this file!

apps/admin/frontend/src/screens/election_screen.tsx
```diff
@@ -46,14 +53,27 @@ export function DefinitionScreen(): JSX.Element {
...
+          <React.Fragment>
+            <P>
+              Save the Ballot Package to USB to configure VxCentralScan or
+              VxScan.
+            </P>
+            <P>
+              <ExportElectionBallotPackageModalButton />
+            </P>
+          </React.Fragment>
+        )}
```
adghayes, line 76: In development, I've always wanted to export the ballot package as the system admin 🤷 , but making a decision on that doesn't need to happen in this PR / code freeze.

apps/admin/frontend/src/components/app_routes.tsx
```diff
@@ -145,8 +144,8 @@ export function ElectionManager(): JSX.Element {
   // Election manager UI
   return (
     <Switch>
-      <Route exact path={routerPaths.ballotsList}>
-        <BallotListScreen />
```
adghayes, line 149: Is this a file that could be deleted now?

adghayes (approved):
LGTM! I like pretty much all the changes, and agree with @mattroe's feedback provided over Slack. I do like the arrangement of the buttons above the table. And the mark official flow is better than its ever been.

I don't have a view of the central-scan test banner without getting on your branch, but assuming that it's fine. The admin test banner on the Tally page is the only thing that I wanted to play with a bit:

Screen Shot 2023-11-22 at 4 38 18 PM

I am fine with shipping it as-is, but I would be curious to see it with the test mode banner above the CVRs header rather than below it. It feels like the "Load CVRs" and "Remove CVRs" button belong below the header and are being split by the banner.

apps/admin/frontend/src/components/app_routes.tsx
adghayes: The file containing `BallotsListScreen`

jonahkagan: Yup, meant to do that, thanks

apps/admin/frontend/src/screens/election_screen.tsx
jonahkagan: Oh interesting. I agree we can discuss outside of this PR

jonahkagan:
> I don't have a view of the central-scan test banner without getting on your branch, but assuming that it's fine.

Yeah, no visual change to that - just changed the implementation to take advantage of `Card.color`.

jonahkagan:
> I would be curious to see it with the test mode banner above the CVRs header rather than below it.

I agree it looks a bit nicer this way. I'm down to change it.
Screenshot 2023-11-22 at 3 03 25 PM

We should also consider how it looks after marking results as official (though I'm not sure if anyone will do that during L&A)

Screenshot 2023-11-22 at 3 03 01 PM

adghayes:
The test mode + mark official state is one that we have to support (I was told when editing report titles), for a complete parallel between normal and test mode, but we don't have to worry about how it looks since we believe no one uses it and, if they do, they know it's just for testing. My instinct is to have the test banner always come first, like in these two cards, but given rarity of this screen I don't think it really matters.

PR #4262, opened 2023-11-25
fix(randomize-cvrs): use `execFile('cp')` instead of `fs.copyFile`

benadida:
`fs.copyFile` has issues on real USB drives when `src` and `dest` are both on USB.


arsalansufi (approved):
Thanks for digging into this - LGTM!

I'm interested in better understanding why `fs.copyFile` errs but shelling out does not (and why our dev VMs, even with real USBs, are fine). No need for you to look into that - I can do some research to satisfy my own curiosity

Update: Just saw [your post in Slack](https://votingworks.slack.com/archives/C060FEVB2L8/p1700953538619239), which provides more context!

Issue #4077, opened 2023-10-11
Looks like we still have rendering bugs on big seal svgs :lolsob:


carolinemodic:
Looks like we still have rendering bugs on big seal svgs :lolsob:

Issue automatically created from slack feedback reported by caroline 
https://votingworks.slack.com/archives/CJU9MSC6S/p1697049242642429

jonahkagan:
In my experience, one way to fix this is to ensure that the SVG does not have `width`/`height` attributes but does have `viewbox="0 0 {width} {height}"`. There are probably ways we can make our code handle SVGs with width/height attributes better, but this is a workaround that I've used in the meantime.

PR #4267, opened 2023-11-27
Update NH Seal

adghayes:
_review as a whole, not by commit_

## Overview

Changes from #4077 fixed appearance for many seals but broke it for others, like the NH seal. We are keeping the change because at least now, it is consistent between VxDesign and the rest of the system, and we believe it makes better assumptions about SVGs. This PR fixes the seal in NH fixtures and converters.

## Screenshot, Fixed

Screen Shot 2023-11-27 at 10 13 04 AM

## Testing Plan

Manual


libs/fixtures/src/data/systemSettings.json.ts
adghayes, line 1: I think this was changed in another PR and was out of sync.

benadida (approved):

PR #4270, opened 2023-11-27
Add back explicit process.exit(0) calls to smart card scripts

arsalansufi:
_Easiest reviewed by commit_

## Overview

I recently removed explicit `process.exit(0)` calls from all scripts that I'd written to let processes exit naturally ([relevant comment thread](https://github.com/votingworks/vxsuite/pull/4225#discussion_r1396449308)). But looks like smart card scripts require an explicit exit or else they hang. This PR adds `process.exit(0)` calls back to those scripts.

While I was at it, I also moved CAC/RAVE scripts to a sub-directory, just to keep RAVE code neatly separated from VxSuite code. That second commit should have no effect on VxSuite, and I'll cherry-pick it onto RAVE.

## Testing Plan

- [x] Tested all relevant scripts before and after this PR


amcmanus (approved):

eventualbuddha:
FWIW, it's typically better to release whatever resource is causing node to not exit on its own than to force an exit.

PR #4271, opened 2023-11-27
feat(ui): add selector to `useQueryChangeListener`

eventualbuddha:
## Overview

Allows selecting a subset of the data and only triggering the change handler when that subset changes.

## Demo Video or Screenshot
n/a

## Testing Plan
New automated test for the hook.

### DIFF
--- a/libs/ui/src/hooks/use_change_listener.ts
+++ b/libs/ui/src/hooks/use_change_listener.ts
@@ -30,6 +30,26 @@ export function useExternalStateChangeListener<State>(
   }, [currentState, changeHandler]);
 }
 
+function identity<T>(value: T): T {
+  return value;
+}
+
+/**
+ * Callback for {@link useQueryChangeListener} when query data changes.
+ */
+export type UseChangeListenerChangeHandler<Data> = (
+  newData: Data,
+  previousData?: Data
+) => void | Promise<void>;
+
+/**
+ * Selector for {@link useQueryChangeListener} to map query data to a different
+ * type, typically to select a subset of the data.
+ */
+export type UseChangeListenerDataSelector<Data, SelectedData> = (
+  data: Data
+) => SelectedData;
+
 /**
  * Registers a handler that will be called whenever the result of a useQuery
  * hook changes (based on a deep equality check).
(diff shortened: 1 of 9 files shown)


libs/ui/src/hooks/use_change_listener.ts
jonahkagan, line 1: What's your motivation for this change? IMO, the selector pattern makes call sites a bit less readable, so I've generally avoided using it unless necessary.

eventualbuddha: I was in here looking at what I need to change in order to do auto-scanning a la PDI. I noticed the historical bug with the multiple sound effects due to triggering on a change of the whole status object, but not the part you cared about. This seemed like a good way to avoid that sort of bug and a lot of the guarding around e.g. the `logged_out` state transitions.

jonahkagan: Ah gotcha. I can definitely see the benefit of that.

I'm a little worried that changing the existing use cases here will change their logic. Even if it's theoretically stricter and therefore safer, what if somehow one of these cases was relying on the existing behavior? Maybe it's unlikely enough that it doesn't matter.

eventualbuddha: I can revert the other places and just do it in the couple I thought would for sure benefit.

libs/ui/src/hooks/use_change_listener.ts
```diff
@@ -47,14 +67,67 @@ export function useExternalStateChangeListener<State>(
...
+ * the on-screen UI (e.g. similar to how we can put an onChange handler on, say,
+ * a button, for a user event that is triggered in the UI).
+ *
+ * It can be combined with the refetchInterval option of useQuery to make the
+ * query poll for new data regularly.
+ *
+ * Example use cases:
+ * - Playing a sound when the scanner accepts a ballot
+ * - Redirecting to a new URL when a background task is complete
+ */
```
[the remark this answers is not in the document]
eventualbuddha: I switched to an options object so this is no longer duplicated.

PR #4273, opened 2023-11-27
Adjudication Status Ballot Reports

### LINKED ISSUE #4070: VxAdmin Reports: Blank Ballot Counts (and other adjudication statuses?)
Expose blank ballot count filtering in the ballot count report builder. It is VVSG required. Along the way, we may want to also expose overvoted / write-in ballot counts.

adghayes:
_review by commit_

## Overview

Closes #4070.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/37960853/9c91b43e-d2d2-4a17-bebe-61390df5b5ef

## Testing Plan

Automated testing, plus a manual test from end to end with a custom data set.

## Checklist

### DIFF
--- a/apps/admin/backend/src/cast_vote_records.ts
+++ b/apps/admin/backend/src/cast_vote_records.ts
@@ -282,10 +282,12 @@ export async function importCastVoteRecords(
         return err({ ...validationResult.err(), index: castVoteRecordIndex });
       }
 
-      // Add an individual cast vote record to the import
       const votes = convertCastVoteRecordVotesToTabulationVotes(
         castVoteRecordCurrentSnapshot
       );
+      // Currently, we only support filtering on initial adjudication status,
+      // rather than post-adjudication status. As a result, we can just calculate
+      // now, during import.
       const adjudicationFlags = getCastVoteRecordAdjudicationFlags(
         votes,
         electionDefinition
--- a/apps/admin/backend/src/cast_vote_records.ts
+++ b/apps/admin/backend/src/cast_vote_records.ts
@@ -43,6 +43,7 @@ import {
   CvrFileMode,
   ImportCastVoteRecordsError,
 } from './types';
+import { getCastVoteRecordAdjudicationFlags } from './util/cast_vote_records';
 
 /**
  * Validates that the fields in a cast vote record and the election definition correspond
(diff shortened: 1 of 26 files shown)


apps/admin/backend/src/cast_vote_records.ts
```diff
@@ -285,6 +286,10 @@ export async function importCastVoteRecords(
       const votes = convertCastVoteRecordVotesToTabulationVotes(
         castVoteRecordCurrentSnapshot
       );
+      const adjudicationFlags = getCastVoteRecordAdjudicationFlags(
+        votes,
+        electionDefinition
+      );
```
jonahkagan, line 292: What happens if the votes are later changed by write-in adjudication?

adghayes: @mattroe and I discussed, and decided that we are not handling that for now. If a user wanted to see a report on blank ballots, they _probably_ are more interested in the number of blank ballots before adjudication rather than after. The latter case we would probably meet with another set of flags, when we come to that.

libs/types/src/index.ts
```diff
@@ -1,4 +1,5 @@
 /* istanbul ignore file */
+export * as Admin from './admin';
```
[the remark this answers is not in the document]
adghayes: There is VxAdmin specific code in `libs/ui`, yes. Since our reports share components (e.g. headers and tally tables), it made sense when adding VxAdmin specific reports to add them in `libs/ui`. That way, they are grouped with all the other reports and can benefit from the development tools (storybook) set up in `libs/ui`. 

In the future, I envision report components living in `libs/printing`, which `admin-backend` would depend on, so then it would make even more sense to define it elsewhere.

Open to other thoughts here. Splitting `Admin` types across `admin-backend` and `libs/types` is not very pretty, but it works.

apps/admin/backend/src/cast_vote_records.ts
jonahkagan: Ok, do we want to make that clearer in the UI and/or in the code?

libs/types/src/index.ts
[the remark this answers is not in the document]
jonahkagan: Ok, I can live with it for ease of using Storybook

jonahkagan (approved):

apps/admin/backend/src/cast_vote_records.ts
adghayes: I added a comment, but I don't think this is important enough to make explicit in the UI. It's still a bit of a corner case since we only have write-in adjudication and not general purpose adjudication.

- Overvote status cannot change, because we cannot yet adjudicate overvotes
- Blank ballot and write-in status could _arguably_ change if we validate a UWI, but not worth worrying about this NH edge case
- The undervote status can change significantly with write-in adjudication, so that could be confusing. But I'd wager that is the least sought after report anyway.

There's a case for having the post-adjudication status eventually and, at that point, we'll have to find a way to be clear about it. But for now, not an issue IMO.

adghayes: related discussion: https://votingworks.slack.com/archives/CEL6D3GAD/p1699381326254249

PR #4281, opened 2023-11-28
[i81n/Audio] Add language tags to the BallotStyle schema

kofi-q:
## Overview

_Relevant spec available [here](https://docs.google.com/document/d/1a2GRj1g4aOawEssPdTCgbv-qs28d-M_dyMEfzm4I4nM/edit#heading=h.7g3k1e4s9ri)._

Currently, a unique ballot style is defined by a combination of its relevant precincts, districts, and/or party. This change enables additional disambiguation via the language/languages used in rendering the ballot content.

- Adding the repeated `Language` field to the CDF ballot definition schema, containing `LanguageCode` strings.
- Adding a corresponding `languages` field to the VXF election schema.
- Adding shared utilities for generating unique ballot style IDs based on the ballot style parameters, as well as for grouping and managing lists of ballot styles that only differ in language(s).
  - These allow us to switch between ballot styles whenever a voter switches their preferred display language.

## Testing Plan
- New unit tests for the ballot style utils
- Removed a redundant snapshot assertion in the schema conversion tests, since there are already tests asserting specific results of the conversions back-and-forth

## Checklist


libs/utils/src/ballot_styles.ts
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Returns a language-specific ballot style, for the given
+ * {@link destinationBallotStyleLanguage}, that belongs to the same
+ * language-agnostic ballot style group as the given
+ * {@link sourceBallotStyleId}.
+ */
+export function getRelatedBallotStyle(params: {
+  ballotStyles: readonly BallotStyle[];
+  destinationBallotStyleLanguage: LanguageCode;
```
arsalansufi, line 67: nit: I always find myself going back and forth between `destination` and `target` 😅, but in this case, I find `targetBallotStyleLanguage` to read more intuitively than `destinationBallotStyleLanguage`

libs/utils/src/ballot_styles.ts
```diff
@@ -0,0 +1,128 @@
...
+  );
+  if (!destinationBallotStyle) {
+    return err('destination ballot style not found');
+  }
+
+  return ok(destinationBallotStyle);
+}
+
+/**
+ * Returns only English-language ballot styles from the given list.
```
arsalansufi, line 92: Worth noting that this includes English-plus-a-second-language ballot styles, too

libs/utils/src/ballot_styles.ts
```diff
@@ -0,0 +1,128 @@
...
+}): Result<BallotStyle, string> {
+  const { ballotStyles, destinationBallotStyleLanguage, sourceBallotStyleId } =
+    params;
+
+  const ballotStyleGroups = getBallotStyleGroups(ballotStyles);
+  const groupId = extractBallotStyleGroupId(sourceBallotStyleId);
+
+  const matchingGroup = ballotStyleGroups[groupId];
+  if (!matchingGroup) {
+    return err('source ballot style group not found');
```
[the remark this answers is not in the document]
arsalansufi: ^ Got it! Yeah I think it's fine to use `err` however best fits your needs. And here, I can see why we just want to supply a string to log

Issue #4232, opened 2023-11-17
Upon selecting an option in a contest with a PAT device, focus should jump to the "Next" button


kshen0:
Advancing focus to the "Next" button avoids the voter having to sip+puff through a very long candidate list and redundant buttons in the footer

PR #4284, opened 2023-11-28
Polish PAT focus behavior

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/4233
https://github.com/votingworks/vxsuite/issues/4232

* Fixes a bug on post-print review page where a PAT device couldn't scroll through the contest cards
* When a contest vote is made, automatically moves focus from contest option to "Next" button
  * Happens on both the initial contest screen and contest screen reached from the review screen
  * Previously there was no special handling of focus. PAT users would need to scroll through all remaining contest options to get to the "Next" button

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/dc86169b-35f3-47f1-94c2-2b5f133ca773

## Testing Plan
- [x] Updated existing tests
- [ ] Add tests for new focus behavior when we confirm behavior in video is the final behavior

## Checklist


apps/mark-scan/backend/schema.sql
```diff
@@ -39,3 +39,7 @@ create table ui_string_audio_ids (
   data text not null, -- JSON blob - see libs/types/UiStringAudioIdsSchema
   foreign key (language_code) references languages(code)
 );
+
+create table hardware_config (
+  is_pat_device_connected integer
+);
```
kshen0, line 45: This state was previously stored in the state machine `context`, but since we now need it on the frontend this felt like an appropriate place to put it.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -636,11 +635,11 @@ export function buildMachine(
               await auth.endCardlessVoterSession(
                 constructAuthMachineState(context.workspace)
               );
+              context.workspace.store.setIsPatDeviceConnected(false);
```
kshen0, line 638: Kind of a tricky situation - resetting connection status when the state machine resets is required so a shared PAT device (uncommon, but possible if the precinct supplies one) is recalibrated for each voter. It's possible a jam would happen before the ballot is printed, and therefore the same voter would need to go through device calibration twice. I think we should err toward over-calibration rather than under-calibration.

apps/mark-scan/frontend/src/lib/gamepad.ts
```diff
@@ -1,16 +1,17 @@
 import { Button } from 'react-gamepad';
 import { mod } from '../utils/mod';
 
+enum QuerySelector {
+  Audio = 'button:not([aria-hidden="true"]):not([disabled])',
+  PAT = 'button:not([aria-hidden="true"]):not([disabled]):not([tabindex="-1"]), [role="button"]:not([aria-hidden="true"]):not([disabled]):not([tabindex="-1"])',
```
kshen0, line 6: @jonahkagan @kofi-q any thoughts on separating out these query selectors? With audio focus logic still in flux I don't want to step on any toes, but I do think our ultimate goal should be to unite screen reader and PAT focus behavior.

libs/mark-flow-ui/src/pages/contest_page.tsx
```diff
@@ -92,15 +103,60 @@ export function ContestPage(props: ContestPageProps): JSX.Element {
...
+      : getReviewPageUrl();
+    history.push(to);
+    // Need to blur or the focus on "Next" button will be retained for next contest,
+    // increasing chance the voter will accidentally skip a contest
+    nextContestButtonRef?.current?.blur();
+  }, [getContestUrl, getReviewPageUrl, history, nextContest, nextContestIndex]);
+
   const nextContestButton = (
-    <LinkButton
+    <Button
```
kshen0, line 118: Converted to `Button` so we can call `focus()` and `blur()` on it. Consequently needed to reimplement `to` behavior. Checked with @kofi-q and this is aligned with our direction for `LinkButton`, which we want to reserve for true web browser contexts only.

kshen0: Without code changes, the only way that might focus `LinkButton` out of the box is to pass the `autoFocus` prop. I found that doing that adds highlight styles but doesn't focus it ie. sending the PAT equivalent of "Enter" doesn't actually navigate to the next contest.

If we wanted to make changes to `LinkButton` we could use refs like `Button` does to handle focus and blur, or convert `LinkButton` to a class component so we can expose methods that wrap `Button.focus()` and `Button.blur()`. So the question is whether we want to invest in `LinkButton`, which is related to your second question.

Based on my conversation with @kofi-q, my understanding is `LinkButton` main benefit is when the user is in a regular web browser and we want to support actions like using the back button or `ctrl + click` for a new tab. Since we're not using a regular browser for the machine apps the preferred direction is to move away from `LinkButton` and back to `Button`. Therefore, while we could build out focus/blur handling in `LinkButton`, it's preferable to use `Button` which already has that handling.

libs/mark-flow-ui/src/pages/contest_page.tsx
```diff
@@ -26,6 +33,9 @@ interface ContestPageProps {
   precinctId?: PrecinctId;
   updateVote: ContestProps['updateVote'];
   votes: VotesDict;
+  // Moves focus to the "Next" button after voter selects an option in a contest.
+  // This helps reduce repetitive action for accessibility device users.
+  moveFocusAfterUpdateVote?: boolean;
```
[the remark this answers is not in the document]
kshen0: I think it's a design decision - I thought it was visually distracting to apply the dotted border styles to the "Next" button as a touch screen user. But I could also see an argument for usability, since it does hint the user what to press next. Do you have a strong opinion?

libs/mark-flow-ui/src/pages/contest_page.tsx
```diff
@@ -92,15 +103,34 @@ export function ContestPage(props: ContestPageProps): JSX.Element {
...
-    </LinkButton>
+    </Button>
+  );
+
+  const handleUpdateVote: ContestProps['updateVote'] = useCallback(
+    (contestIdProp: ContestId, voteProp: OptionalVote) => {
+      if (moveFocusAfterUpdateVote) {
+        nextContestButtonRef?.current?.focus();
+      }
+      updateVote(contestIdProp, voteProp);
```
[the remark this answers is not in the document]
kshen0: Good catch, we shouldn't advance focus until the nth selection is made. I'll update.

apps/mark-scan/backend/schema.sql
kshen0: That's reasonable. I think it's unintuitive to the reader if that state is stored in memory in the state machine context, but I think it makes sense as a class property of `Store`.

libs/mark-flow-ui/src/pages/contest_page.tsx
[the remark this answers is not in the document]
jonahkagan: Ah that makes sense. I agree with you it's nice to only show the focus outlines if the user is using focus-based navigation.

That's making me think it might be nice to have two behavior groups:
- Assistive devices/keyboard nav
- Touch screen nav

If we can distinguish which of these cases triggered a button (which is possible using the event object), then we can trigger further focus behavior. I'm thinking that it's possible to switch between the various navigation modes even when the devices are connected. Example here: https://github.com/votingworks/vxsuite/blob/main/libs/ui/src/radio_group.tsx#L133-L144

I'm thinking this might be a nice approach to both make behavior consistent and simplify the architecture.

libs/mark-flow-ui/src/pages/contest_page.tsx
jonahkagan: Thanks for sharing all of that context.

Since `LinkButton` is implemented with `Button` under the hood, I think it should be as simple as forwarding the ref to the `Button` to gain support for `focus` and `blur`. IMO, that's basically equivalent to what you've done here, which reimplements the same logic as inside `LinkButton` (and I'm pretty sure still supports the back button implicitly by way of the `history` object). I think there may be a slight misunderstanding since our `LinkButton` isn't using an `a` tag under the hood, which would support `ctrl+click` and such.

apps/mark-scan/frontend/src/lib/gamepad.ts
jonahkagan: Hmm, weird that `role="button"` didn't work. Oh well

I'm not familiar enough with our screen reader to know the answer to your second question. @kofi-q do you have any ideas?

It's also fine to just have both for now if we're not sure

libs/mark-flow-ui/src/pages/contest_page.tsx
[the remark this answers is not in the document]
kshen0: Can you say more about what situations could be handled by checking `e.nativeEvent.pointerType`?

1. Focus `Next` or `Review` button after pressing button to make a contest vote
2. Blur `Next` button after selected
3. Something I missed?

1 is complicated because we want to change focus state on an element other than the one that was just pressed. We handle the contest button press pretty far down in ContestPage > Contest > {CandidateContest, YesNoContest, ...} > ContestChoiceButton. So we'd have to pass the event up from [`ContestChoiceButton.onPress()`](https://github.com/votingworks/vxsuite/blob/main/libs/ui/src/contest_choice_button.tsx#L79) to at least ContestPage, which is the component that's aware of the `Next` and `Review` buttons. Or we'd pass a function `(event: Event) => void` down from ContestPage to ContestChoiceButton.

2 is simpler because we just need to blur the focus on the same element that received the event, so we can handle similarly to `RadioGroup` and avoid passing an additional `onPress` handler to `LinkButton`.

I'm not sure case 1 becomes any simpler if we switch over to checking `nativeEvent.pointerType`, but probably worth doing case 2. Am I missing something that makes case 1 simpler?

libs/mark-flow-ui/src/pages/contest_page.tsx
kshen0: Reverted back to LinkButton and used `forwardRef` in [5b74493](https://github.com/votingworks/vxsuite/pull/4284/commits/5b74493d7373f84c16c864de5a9319cb3a51bdaf).

Back button was my own assumption but you're right it would be handled by `history` anyway.

libs/mark-flow-ui/src/pages/contest_page.tsx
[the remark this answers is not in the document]
jonahkagan: I was imagining that we'd just modify `Button` to pass the event to its `onPress` handler. Then the `handleUpdateVote` function could take a parameter derived from the button's event to determine whether to focus the Next button.

The reason I think this is an improvement is that then the frontend doesn't to query the backend to know if the PAT device is connected (at least, based on what I'm reading here - but I could be missing something). We'd be letting the browser event tell us what interaction method is being used, and removes the need to store and expose the PAT device status.

If that doesn't seem like an improvement though, I'm game to keep it as is.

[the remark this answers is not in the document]
kshen0: > The reason I think this is an improvement is that then the frontend doesn't to query the backend to know if the PAT device is connected 

That's true, and I think that simplification is worth the (IMO) messier function passing in the frontend. Let me confirm that moving focus around isn't an anti pattern for screen readers. Then I'll make these changes.

apps/mark/frontend/src/setupTests.tsx
```diff
@@ -69,6 +69,9 @@ function mockSpeechSynthesis() {
 beforeEach(() => {
   mockSpeechSynthesis();
   fetchMock.mock();
+  // PointerEvent is unimplemented by js-dom
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (window as any).PointerEvent = Event;
```
kshen0, line 74: This is necessary because `contest_page` checks `instanceof PointerEvent` [here](https://github.com/votingworks/vxsuite/pull/4284/files#diff-f506ba67da9120a6f6d7f47a6988fec4355469efa7b5932ee71eb0aa3e4ad5b1R50). I was torn on this; feels like a trapdoor to require callers in other projects to add this to their test setup, and for some reason the resulting test failure from not having this polyfill results in an infinite loop and OOM errors in CI.

An alternative is to be less type safe:

```
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  return (nativeEvent as any).pointerType === '';
```

Also open to suggestions here.

apps/mark-scan/frontend/src/lib/gamepad.ts
kofi-q: Missed these threads earlier, but also don't have too much historical context on the existing gamepad implementation -- my understanding is that it's essentially trying to simulate browser `tab`/`shitft+tab` behaviour for the up/down buttons. I did wonder if we would be better off mapping the gamepad to keyboard `tab`/`shift+tab` key presses at the system level instead, which would save us from re-implementing this browser functionality.

That being said, we only expect UI controls like buttons and options to be focusable for audio readouts (we do have an initial focus on an introductory text div with `id="audiofocus"` on page load, but that's separate from the accessible controller navigation). That makes me think uniting the focus selectors should be fine/is probably what we want.

(Also, how much control do we have over the PAT -> keyboard mapping? Maybe we can map the focus control to `tab` key instead?)

eventualbuddha: > Also, how much control do we have over the PAT -> keyboard mapping? Maybe we can map the focus control to `tab` key instead?

I believe we can simulate any keypress we want, so `tab` and `shift+tab` should work fine. I'm not sure whether @kshen0 had a reason to use something else, though.

apps/mark-scan/backend/src/store.ts
```diff
@@ -25,10 +25,16 @@ const SchemaPath = join(__dirname, '../schema.sql');
  * Manages a data store for imported election definition and system settings
  */
 export class Store {
+  // We don't want PAT connection status to persist across machine restarts,
+  // but we want it to be accessed through the store instead of state machine context
```
eventualbuddha, line 29: Why do we want that? Also, storing it like this means that the value will be lost when a backend restart happens (such as when reloading after a file changes in dev). Is that a problem?

apps/mark-scan/backend/src/store.ts
```diff
@@ -313,4 +319,12 @@ export class Store {
...
     );
   }
+
+  createHardwareConfig(): void {
+    this.client.run('delete from hardware_config');
+    this.client.run(
+      'insert into hardware_config (is_pat_device_connected) values (?)',
+      0
+    );
+  }
```
eventualbuddha, line 329: Accidentally left in here?

apps/mark-scan/frontend/src/pages/contest_screen.tsx
```diff
@@ -21,7 +21,7 @@ function getStartPageUrl() {
   return '/';
 }
 
-export function ContestScreen(): JSX.Element {
+export function ContestScreen(): JSX.Element | null {
```
eventualbuddha, line 24: Leftover change?

libs/mark-flow-ui/src/components/candidate_contest.test.tsx
```diff
@@ -249,9 +249,12 @@ describe('supports write-in candidates', () => {
     userEvent.click(modal.getByText('Accept'));
     expect(screen.queryByRole('alertdialog')).not.toBeInTheDocument();
 
-    expect(updateVote).toHaveBeenCalledWith(candidateContestWithWriteIns.id, [
-      { id: 'write-in-lizardPeople', isWriteIn: true, name: 'LIZARD PEOPLE' },
-    ]);
+    expect(updateVote).toHaveBeenCalledWith(
+      candidateContestWithWriteIns.id,
+      [{ id: 'write-in-lizardPeople', isWriteIn: true, name: 'LIZARD PEOPLE' }],
+      // expect.objectContaining({ nativeEvent: expect.any(Object) })
```
eventualbuddha, line 255: Is this a TODO, or should we just delete it?

libs/mark-flow-ui/src/components/candidate_contest.tsx
```diff
@@ -103,31 +103,31 @@ export function CandidateContest({
     }
   }, [deselectedCandidate]);
 
-  function addCandidateToVote(id: string) {
+  function addCandidateToVote(id: string, event: SyntheticEvent) {
```
eventualbuddha, line 106: Feels a little odd to me to be passing this event around, but I do see why. If the only reason you need it is to determine the update method, maybe make that explicit here?

```ts
enum VoteUpdateMethod {
  Touch = 'touch',
  Other = 'other',
}

function addCandidateToVote(id: string, method: VoteUpdateMethod) {
```

libs/ui/src/link_button.tsx
```diff
@@ -1,38 +1,52 @@
 import { useHistory } from 'react-router-dom';
-import { Button, ButtonProps } from './button';
+import { forwardRef, Ref, SyntheticEvent } from 'react';
+import { Button, ButtonProps, ClickHandler, TypedClickHandler } from './button';
 
 /**
  * Props for {@link LinkButton}.
  */
 export interface LinkButtonProps extends Omit<ButtonProps, 'onPress'> {
+  // export type LinkButtonProps = ButtonProps & {
```
eventualbuddha, line 9: Delete?
```suggestion
```

libs/ui/src/link_button.tsx
```diff
@@ -1,38 +1,52 @@
...
   goBack?: boolean;
   primary?: boolean;
   to?: string;
+  ref?: Ref<Button<SyntheticEvent>>;
+  // onPress not optional for <Button> but optional here. Its behavior
```
eventualbuddha, line 14: Sentence fragment.

libs/ui/src/link_button.tsx
```diff
@@ -1,38 +1,52 @@
...
+  onPress?: ClickHandler | TypedClickHandler<unknown>;
```
eventualbuddha, line 15: This discards the feature of `Button` where if `onPress` takes a value `T` then the `value` prop is required.

apps/mark-scan/backend/src/store.ts
kshen0: PAT devices are usually brought by the voter so we expect the device to be disconnected at the end of a voting session. In the event of a machine restart mid-session, we don't want to assume the same voter has returned to the same machine. It wouldn't be _bad_ to store PAT connection status in the db and update its value on restart, but it's unnecessary.

Re: backend restarts, it's a slight inconvenience (requires 1 extra screen + button click to skip calibration) but I wouldn't say it's worth it to keep this bit in the db for dev purposes only.

apps/mark-scan/backend/src/store.ts
kshen0: Sorry, should have done a last pass for this type of cleanup before opening for review.

https://github.com/votingworks/vxsuite/pull/4284/commits/f977e7e976ffbf77a2e882c749ee8352e1fe3ffe

apps/mark-scan/frontend/src/pages/contest_screen.tsx
kshen0: https://github.com/votingworks/vxsuite/pull/4284/commits/f977e7e976ffbf77a2e882c749ee8352e1fe3ffe

libs/mark-flow-ui/src/components/candidate_contest.test.tsx
kshen0: https://github.com/votingworks/vxsuite/pull/4284/commits/f977e7e976ffbf77a2e882c749ee8352e1fe3ffe

Updated to use `syntheticEventMatcher`

libs/ui/src/link_button.tsx
kshen0: https://github.com/votingworks/vxsuite/pull/4284/commits/f977e7e976ffbf77a2e882c749ee8352e1fe3ffe

libs/ui/src/link_button.tsx
kshen0: https://github.com/votingworks/vxsuite/pull/4284/commits/f977e7e976ffbf77a2e882c749ee8352e1fe3ffe

kshen0: https://github.com/votingworks/vxsuite/pull/4284/commits/f5d6a6b0e18c065a8aa91b3b043a1a7d71cdac2c#diff-4ed014966e4ba63b7cc24fa7d03553eb0d1ded7d73a9ef1641d6c8df5e6a8c5eL9

libs/ui/src/link_button.tsx
kshen0: We never actually use `TypedClickHandler` (no existing usage of `LinkButton` passes a value, so there is no case where we want to pass a value through to `Button`). In [f5d6a6b](https://github.com/votingworks/vxsuite/pull/4284/commits/f5d6a6b0e18c065a8aa91b3b043a1a7d71cdac2c) I deleted `TypedClickHandler` from allowable types for `LinkButton.onPress` to make this explicit. This also avoids the problem you described.

libs/mark-flow-ui/src/components/candidate_contest.tsx
kshen0: Added in [8f2df03](https://github.com/votingworks/vxsuite/pull/4284/commits/8f2df03b9d6dd25b17789d68620c13a21987ec42). I like how this makes it more clear why we need to pass a value up and keeps the `UpdateVoteFunction` type simple. The downside is conversion from `Event` to `VoteUpdateInteractionMethod` isn't centralized. Overall felt like the best approach but open to suggestions.

apps/mark-scan/frontend/src/lib/gamepad.ts
kshen0: That's a good point. Here's what we could support if we send `tab`, `shift+tab`, and `enter` directly from the code where we handle input signals:

1. All behavior for the built-in PAT input
2. Some behavior for the accessible controller. For example, current left and right button behavior would be unsupported because they must rely on querying the DOM
3. No behavior for the USB PAT input switch, used in development only

(3) is a luxury but (2) is needed for production, so we'd need this gamepad abstraction in some form no matter what. Given that, I think it's cleaner to keep all focus behavior in one place - thoughts?

kofi-q: I'm thinking the ideal end state is that we no longer need `gamepad.ts` if accessible controllers can map their "focus next/previous" controls to `tab`/`shift+tab` and their "next/previous screen" controls to something like `ArrowLeft/ArrowRight`, then relevant screens can implement handlers for the left/right keys as needed.

After reading through the code, the expectation that a screen will render buttons with `id="previous"`/`id="next"` to get left/right working feels a little brittle. And I noticed the current gamepad behaviour is broken for contest cards after the recent `tabindex=-1` change (it focuses on the "change" buttons instead of the full contest card) - that's what worries me a bit about trying to re-implement browser focus functionality.

Even if we do want to move in the direction of relying on `tab`/`shift+tab`, I think it's fine to add to `gamepad.ts` for now (unifying the query selectors) and then refactoring as a whole later as a separate project.

apps/mark-scan/backend/src/store.ts
jonahkagan: Is this relevant anymore? It seems like we're not using this field now that it's just stored in the state machine context and it can just be removed

apps/mark/frontend/src/setupTests.tsx
jonahkagan: I think when I originally tried this approach I just checked for `'pointerType' in nativeEvent` which is both type-safe and doesn't require the `PointerEvent` class to be defined.

libs/ui/src/link_button.tsx
```diff
@@ -1,38 +1,52 @@
...
 
 /**
  * Props for {@link LinkButton}.
  */
 export interface LinkButtonProps extends Omit<ButtonProps, 'onPress'> {
+  // export type LinkButtonProps = ButtonProps & {
   goBack?: boolean;
   primary?: boolean;
   to?: string;
+  ref?: Ref<Button<SyntheticEvent>>;
```
jonahkagan, line 12: It doesn't seem like `SyntheticEvent` should be passed in as the value type of the `Button` - isn't that type parameter only used when storing a `value` on the button?

apps/mark/frontend/src/setupTests.tsx
jonahkagan: I just did a little more research and experimentation and it looks like we can just use a combo of two top-level fields `event.detail` (which is '0' for keyboard or touch and '1' for mouse) and `event.touches` (which is an array for touch and not defined for mouse/keyboard). That way we don't need to access the `nativeEvent` at all

libs/ui/src/button.tsx
```diff
@@ -43,8 +43,8 @@ const buttonVariants = {
 export type ButtonVariant = keyof typeof buttonVariants;
 export const BUTTON_VARIANTS = Object.keys(buttonVariants) as ButtonVariant[];
 
-type ClickHandler = () => void;
-type TypedClickHandler<T> = (value: T) => void;
+export type ClickHandler = (event: SyntheticEvent) => void;
+export type TypedClickHandler<T> = (value: T, event: SyntheticEvent) => void;
```
jonahkagan, line 47: Maybe a more specific event type would be appropriate here. E.g.:

```ts
type ButtonPressEvent = React.TouchEvent<ButtonElement> | React.MouseEvent<ButtonElement>
```

That way the consumer knows what fields it can access more precisely.

kshen0: Good suggestion, thanks! [635c445](https://github.com/votingworks/vxsuite/pull/4284/commits/635c44510fb64fa932b7781a8880fd7c76667521)

libs/ui/src/link_button.tsx
kshen0: Summarizing chat: `SyntheticEvent` value type was passed here to avoid `Type 'Ref<Button<undefined>>' is not assignable to type 'LegacyRef<Button<SyntheticEvent<Element, Event>>> | undefined'.` error in `LinkButton`. Not sure exactly how this type inference was working, but passing `never` fixes the problem. I believe this is correct because `LinkButton` never passes an `onClick` that requires a default value.

https://github.com/votingworks/vxsuite/pull/4284/commits/2046a88da5ba818c4b6ecf87d7149b0f11ae2710#diff-4ed014966e4ba63b7cc24fa7d03553eb0d1ded7d73a9ef1641d6c8df5e6a8c5eR44

apps/mark-scan/frontend/src/lib/gamepad.ts
kshen0: Agreed offline that sending `tab`, `shift+tab`, and `enter` from lower-level device signal handlers is better. We can reduce `gamepad.ts` down to handling just the USB PAT input switch (used in development only).

libs/test-utils/src/matchers.ts
```diff
@@ -0,0 +1,5 @@
+export function buttonPressEventMatcher():
+  | React.TouchEvent<HTMLButtonElement>
+  | React.MouseEvent<HTMLButtonElement> {
+  return expect.objectContaining({ detail: expect.any(Number) });
+}
```
kshen0, line 5: Could see an argument for putting this in `libs/ui`, closer to where it's defined

apps/mark-scan/frontend/src/lib/gamepad.ts
kshen0: [9f6b023](https://github.com/votingworks/vxsuite/pull/4284/commits/9f6b0236a04e9335159cbf2649f00c1ff1a390cf)

libs/mark-flow-ui/src/config/types.ts
```diff
@@ -11,16 +11,44 @@ import {
...
+      /* istanbul ignore next - react-testing-library/userEvent pointer API was added in 14.0.0; we are on 13.x.x */
+      if ((event as React.TouchEvent<HTMLButtonElement>).touches) {
+        return VoteUpdateInteractionMethod.Touch;
+      }
+      return VoteUpdateInteractionMethod.AssistiveTechnologyDevice;
+    case 1:
+      return VoteUpdateInteractionMethod.Mouse;
+    default:
+      /* istanbul ignore next */
+      return VoteUpdateInteractionMethod.Unknown;
```
jonahkagan, line 41: My read on `detail` is that it counts the number of interactions. So if someone double clicked you'd get `detail: 2` (see https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail). So it might be safe to just say `Mouse` in all cases over 0 if that makes life simpler.

libs/test-utils/src/matchers.ts
jonahkagan: No strong opinion - this is fine with me

libs/mark-flow-ui/src/pages/contest_page.tsx
```diff
@@ -92,39 +104,81 @@ export function ContestPage(props: ContestPageProps): JSX.Element {
...
+      nextContestButtonRef?.current?.blur();
+    }
+  }, []);
+
   const nextContestButton = (
     <LinkButton
       id="next"
       rightIcon="Next"
       variant={isVoteComplete ? 'primary' : 'neutral'}
+      onPress={handleOnPressNext}
```
jonahkagan, line 125: It feels a little weird to me that LinkButton now takes an optional `onPress` callback that can get called, because it's not immediately clear how that interacts with the actual navigation logic (which one run firsts? does `onPress` override `to` or complement it?)

A few alternative ideas:
- Potentially using a `key` prop with the contest id somewhere would force React to rerender the elements rather than reuse the existing elements (e.g. on this button, or on the whole screen - not sure exactly where the right place is)
- Switch from `LinkButton` back to `Button`. I know I suggested sticking with `LinkButton` in the first place, but I didn't realize the context of where we wanted to call `blur` from. Sorry for not reading more closely!

jonahkagan (approved):
A few small notes but overall things are looking good! Thanks for working through the complexity here. I appreciate the commitment to making PAT behavior work well and improving our tooling around accessibility

libs/mark-flow-ui/src/config/types.ts
kshen0: [ace489f](https://github.com/votingworks/vxsuite/pull/4284/commits/ace489f9e174e71a7f1eb75a4608183cef28e7ed)
thanks!

libs/mark-flow-ui/src/pages/contest_page.tsx
kshen0: I see the argument; I had initially called the prop `additionalOnPressBehavior` but thought it introduced more confusion than clarity. Switched to `key` in [d17ac75](https://github.com/votingworks/vxsuite/pull/4284/commits/d17ac7555b4b8ca9cfbf4dccd07219f13cd9e8f9).

This also revealed a rough edge where focus resumes at the "Next" button, instead at the top of the page, when advancing from one contest to another:

https://github.com/votingworks/vxsuite/assets/1060688/bc2a5966-5f8e-4640-b43e-bc3824e6aa62

This is fixable by adding an `autoFocus` prop to the various contest types and `ContestChoiceButton` but I think we should actually piggyback off the `id="audiofocus"` mentioned above: https://github.com/votingworks/vxsuite/pull/4284#discussion_r1431641061

Can tackle this in another PR, since this one is dragging out.

jonahkagan: Nice catch on the Next button focus issue. I agree addressing in a follow up makes sense.

Another potential way to handle that would be to have a child component for the contest screen (including buttons) for an individual contest that is rendered with `key={contest.id}`. Would that reset the focus state for the whole screen each time you move between contests? If so, that would also solve the link button focus issue from earlier in this thread.

PR #4293, opened 2023-11-28
Fix various VxScan mode-switching bugs

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4285

This PR fixes various VxScan bugs around mode switching (i.e. switching between test and official mode). The bugs include the following:

1. Switching mode after polls have been opened (and even subsequently closed) is supposed to return the polls to their initial closed state. But VxScan is currently only "half" switching - the backend is updating but the frontend is not.
2. Switching mode does not result in the creation of a new export directory for continuous export, when it should.
3. The only way to currently clear the CVR sync required screen is to click the screen's sync button. But other actions can also render a sync to USB drive unnecessary. One such action is switching from test mode to official mode, since switching mode clears scanned ballots. These other actions should also clear the CVR sync required screen.

This PR fixes all of the above.

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

### DIFF
--- a/apps/scan/backend/src/store.ts
+++ b/apps/scan/backend/src/store.ts
@@ -657,6 +657,7 @@ export class Store {
       this.client.transaction(() => {
         this.resetPollsState();
         this.setBallotCountWhenBallotBagLastReplaced(0);
+        this.setExportDirectoryName(undefined);
       });
     }
     // delete batches, which will cascade delete sheets
@@ -941,16 +942,18 @@ export class Store {
   /**
    * Stores the name of the directory that we'll be continuously exporting to
    */
-  setExportDirectoryName(exportDirectoryName: string): void {
+  setExportDirectoryName(exportDirectoryName?: string): void {
     this.client.run('delete from export_directory_name');
-    this.client.run(
-      `
+    if (exportDirectoryName !== undefined) {
+      this.client.run(
+        `
       insert into export_directory_name (
         export_directory_name
       ) values (?)
       `,
-      exportDirectoryName
-    );
+        exportDirectoryName
+      );
+    }
   }
 
   getCastVoteRecordRootHash(): string {
--- a/apps/scan/backend/src/store.ts
+++ b/apps/scan/backend/src/store.ts
@@ -38,6 +38,7 @@ import {
   RejectedSheet,
   Sheet,
   UiStringsStore,
+  clearDoesUsbDriveRequireCastVoteRecordSyncCachedResult,
   createUiStringStore,
 } from '@votingworks/backend';
 import {
@@ -657,9 +658,15 @@ export class Store {
       this.client.transaction(() => {
         this.resetPollsState();
         this.setBallotCountWhenBallotBagLastReplaced(0);
+
+        // Reset all export-related metadata
         this.setExportDirectoryName(undefined);
+        this.clearCastVoteRecordHashes();
+        this.setIsContinuousExportOperationInProgress(false);
+        clearDoesUsbDriveRequireCastVoteRecordSyncCachedResult();
       });
     }
+
     // delete batches, which will cascade delete sheets
     this.client.run('delete from batches');
     // reset autoincrementing key on "batches" table
--- a/apps/scan/backend/src/store.ts
+++ b/apps/scan/backend/src/store.ts
@@ -659,6 +659,11 @@ export class Store {
         this.resetPollsState();
         this.setBallotCountWhenBallotBagLastReplaced(0);
 
+        // Delete batches, which will cascade delete sheets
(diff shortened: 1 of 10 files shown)


apps/scan/frontend/src/api.ts
```diff
@@ -234,6 +234,8 @@ export const setTestMode = {
     return useMutation(apiClient.setTestMode, {
       async onSuccess() {
         await queryClient.invalidateQueries(getConfig.queryKey());
+        // Changing the mode after polls open resets polls to closed
+        await queryClient.invalidateQueries(getPollsInfo.queryKey());
```
arsalansufi, line 238: The fix for bug 1 in the PR description

> Switching mode after polls have been opened (and even subsequently closed) is supposed to return the polls to their initial closed state. But VxScan is currently only "half" switching - the backend is updating but the frontend is not.

apps/scan/backend/src/store.ts
```diff
@@ -657,6 +657,7 @@ export class Store {
       this.client.transaction(() => {
         this.resetPollsState();
         this.setBallotCountWhenBallotBagLastReplaced(0);
+        this.setExportDirectoryName(undefined);
```
arsalansufi, line 660: The fix for bug 2 in the PR description

> Switching mode does not result in the creation of a new export directory for continuous export, when it should.

apps/scan/frontend/src/app_root.tsx
```diff
@@ -316,11 +308,17 @@ export function AppRoot({
...
   }
 
-  if (isCastVoteRecordSyncRequiredScreenUp) {
+  if (
+    usbDrive.doesUsbDriveRequireCastVoteRecordSync ||
+    // This ensures that we don't immediately transition away from the CVR sync success message.
+    // We can't rely on doesUsbDriveRequireCastVoteRecordSync because it becomes false as soon as
+    // the sync completes.
+    shouldStayOnCastVoteRecordSyncRequiredScreen
+  ) {
```
arsalansufi, line 317: The fix for bug 3 in the PR description

> The only way to currently clear the CVR sync required screen is to click the screen's sync button. But other actions can also render a sync to USB drive unnecessary. One such action is switching from test mode to official mode, since switching mode clears scanned ballots. These other actions should also clear the CVR sync required screen.

I previously wasn't using `usbDrive.doesUsbDriveRequireCastVoteRecordSync` directly because I needed the CVR sync required screen to stay open after sync completion to display the success message. But my solution for that ended up introducing the mentioned bug. Now, I'm using `usbDrive.doesUsbDriveRequireCastVoteRecordSync` directly, with a separate solution for the success message - best of both worlds

apps/scan/backend/src/store.ts
```diff
@@ -941,16 +942,18 @@ export class Store {
...
       insert into export_directory_name (
         export_directory_name
       ) values (?)
       `,
-      exportDirectoryName
-    );
+        exportDirectoryName
+      );
+    }
   }
```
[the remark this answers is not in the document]
arsalansufi: Yes thanks for checking!

-----

> Is deleting the directory name on unconfigure already happening?

Yup this is happening. Full unconfigure is great because it just wipes the entire workspace / store - we don't have to explicitly update it every time we add state to the store:

https://github.com/votingworks/vxsuite/blob/a539e3089979d579008f473eb7dfa9c5ef95be08/apps/scan/backend/src/app.ts#L214-L216

https://github.com/votingworks/vxsuite/blob/a539e3089979d579008f473eb7dfa9c5ef95be08/apps/scan/backend/src/util/workspace.ts#L76

https://github.com/votingworks/vxsuite/blob/a539e3089979d579008f473eb7dfa9c5ef95be08/apps/scan/backend/src/store.ts#L196

https://github.com/votingworks/vxsuite/blob/a539e3089979d579008f473eb7dfa9c5ef95be08/libs/db/src/client.ts#L310

https://github.com/votingworks/vxsuite/blob/a539e3089979d579008f473eb7dfa9c5ef95be08/libs/db/src/client.ts#L258-L259

-----

> Does changing the precinct during polls open reset the export directory name? Does it need to?

It doesn't need to because the export directory name is only set _after_ the first ballot is cast and the first CVR is exported, and the precinct can only be changed _before_ polls open and ballot casting. That said, we are still resetting it by virtue of `setPrecinctSelection` calling `workspace.resetElectionSession`, which is where I added the clearing logic:

https://github.com/votingworks/vxsuite/blob/a539e3089979d579008f473eb7dfa9c5ef95be08/apps/scan/backend/src/app.ts#L218-L227

PR #4294, opened 2023-11-28
fix(logs): use exec cp instead of fs-extra, for USB reasons

benadida:


carolinemodic (approved):
remmeber to merge to the release branch too

PR #4309, opened 2023-11-30
refactor: use `IteratorPlus` in a few more places

eventualbuddha:
## Overview
I looked for places where we were building an array using `.push` inside a `for` loop and found some that I thought could be improved using `IteratorPlus`, either with `iter` directly or one of the helpers like `integers`. There are more, but this seemed like a good starting point of the ones that I felt were fairly clearly better this way.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests.

### DIFF
--- a/libs/auth/src/cast_vote_record_hashes.test.ts
+++ b/libs/auth/src/cast_vote_record_hashes.test.ts
@@ -2,6 +2,7 @@ import fs from 'fs';
 import { sha256 } from 'js-sha256';
 import path from 'path';
 import { dirSync } from 'tmp';
+import { iter } from '@votingworks/basics';
 import { Client } from '@votingworks/db';
 import { CastVoteRecordExportFileName } from '@votingworks/types';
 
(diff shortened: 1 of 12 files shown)


eventualbuddha:
Oh, and for posterity, this is how I found the code to possibly refactor:

```js
const { default: traverse } = require("@babel/traverse");
const { parse } = require("@babel/parser");
const { readFileSync } = require("fs");
const m = require("@codemod/matchers");
const assert = require("assert");

const accumulatorId = m.capture(m.identifier());
const forBodyMatcher = 
      m.blockStatement(
        m.anyList(
          m.slice({ min: 0 }),
          m.expressionStatement(
            m.callExpression(
              m.memberExpression(accumulatorId, m.identifier("push"))
            )
          ),
          m.slice({ min: 0 })
        )
      );

const matcher = m.blockStatement(
  m.anyList(
    m.slice({ min: 0 }),
    m.variableDeclaration(undefined, [
      m.variableDeclarator(accumulatorId),
    ]),
    m.slice({ min: 0 }),
    m.or(
      m.forStatement(undefined, undefined, undefined, forBodyMatcher),
      m.forInStatement(undefined, undefined, forBodyMatcher),
      m.forOfStatement(undefined, undefined, forBodyMatcher),
    ),
    m.slice({ min: 0 })
  )
);

for (const arg of process.argv.slice(2)) {
  const ast = parse(readFileSync(arg, "utf8"), {
    sourceType: "module",
    plugins: ["typescript"],
    sourceFilename: arg,
  });

  traverse(ast, {
    BlockStatement(path) {
      const { node } = path;
      m.match(matcher, { accumulatorId }, node, ({ accumulatorId }) => {
        const loc = accumulatorId.loc;
        assert(loc);
        console.log(`${loc.filename}:${loc.start.line}:${loc.start.column + 1}`);
      });
    },
  });
}
```

Searching for the code using a regex is hard/impossible.

jonahkagan (approved):
Looks good to me. I think if you want this style of thinking to catch on, a tech talk might help! Otherwise people will probably keep writing loops

PR #4315, opened 2023-11-30
refactor(card-reader): simplify card reader detection

eventualbuddha:
## Overview
Rather than checking a list of USB devices to determine whether a supported card reader is attached, just let `libs/auth` tell us whether we have a supported card reader or not. If not, we show the existing "No Card Reader Detected" screens we already have.

This removes the need for the `REACT_APP_VX_DISABLE_CARD_READER_CHECK` flag since using mock cards will automatically prevent `libs/auth` from reporting that there are no card readers.

## Demo Video or Screenshot
n/a

## Testing Plan
Manually tested.


apps/scan/frontend/src/app_root.tsx
```diff
@@ -128,7 +124,7 @@ export function AppRoot({
   const pollsInfo = pollsInfoQuery.data;
   const { pollsState } = pollsInfo;
 
-  if (!cardReader) {
+  if (!scannerStatus.isCardReaderConnected) {
```
jonahkagan, line 127: I wonder if we should surface `no_card_reader` as an auth logged out reason? Then we could just do all of this from the auth status object. That case has to be handled in the auth lib anyway

eventualbuddha: [fcdec12](https://github.com/votingworks/vxsuite/pull/4315/commits/fcdec1219a52aa429076392d1a624c94fbc6b4d4)

arsalansufi (approved):
Focused on the libs/auth changes, which look good! Glad we could consolidate!

PR #4319, opened 2023-11-30
[i18n/Audio] Add string translations for BMD paper ballots

kofi-q:
## Overview

Converting text in the `BmdPaperBallot` component to language-aware `<UiString>`s and translating based on the given ballot style language.

The ballot style language takes precedence over the app-wide language setting, if any, when rendering the ballot content - this will enable printouts of ballot styles in different languages for L&A purposes, for example.

For all non-English ballot styles, secondary English strings are printed alongside the primary language strings, where relevant, to enable auditing by English-only-speakers (_VVSG 5.1-C -- Vote Records_). Additionally, certain strings are rendered as English-only (e.g. candidate names, ballot metadata, etc).

Visual layout of the ballot may need some tweaking over time for clarity, but this at least establishes which strings are rendered in dual-language vs single-language mode.

## Demo Video or Screenshot
| Non-English Ballot Style  | English Ballot Style |
| Screenshot 2023-11-30 at 15 56 05 | Screenshot 2023-11-30 at 15 56 17 |

## Testing Plan
- New test cases for single and dual-language ballots
- Manual testing via storybookjs

## Checklist


libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -145,44 +149,148 @@ const VoteLine = styled.span`
...
+                electionParties={election.parties}
+              />
+              )
+            </DualLanguageText>
+          )}
+          {candidate.isWriteIn && (
+            <DualLanguageText
+              primaryLanguage={primaryBallotLanguage}
+              englishTextWrapper={AdjacentText}
+            >
```
arsalansufi, line 278: Nice! I like this `DualLanguageText` + `englishTextWrapper={...}` construct

libs/ui/src/bmd_paper_ballot.dual_language.test.tsx
```diff
@@ -0,0 +1,397 @@
...
+    act(() => getLanguageContext()!.setLanguage(LanguageCode.SPANISH));
+    await waitFor(() =>
+      expect(getLanguageContext()?.currentLanguageCode).toEqual(
+        LanguageCode.SPANISH
+      )
+    );
+
+    expect(container).not.toHaveTextContent(
+      new RegExp(`\\[${LanguageCode.SPANISH}\\]`)
+    );
```
arsalansufi, line 372: Since this check is tied to the implementation of `generateMockUiString` above, might be worth putting construction of this `[languageCode]` prefix in a shared function that `generateMockUiString` also uses. Wouldn't want us to change the output of `generateMockUiString` and then have this check still succeed, but for the wrong reason, potentially masking bugs.

arsalansufi (approved):
LGTM!

This may result in some BMD ballots taking more than one page, which I don't believe we currently support. But we can always revisit the layout when we hit that case. The framework looks great

kofi-q (commented):
> This may result in some BMD ballots taking more than one page

Yeah, I'm a little worried about that - hoping to revisit the formatting of the ballot as a whole to see if we can squeeze in more content and use some of the horizontal space we have in there 🤞

PR #4320, opened 2023-12-01
Standardize "election package" copy

jonahkagan:
## Overview

Use "election package" everywhere we currently use "ballot package" or "setup package."

Most of this PR is find-and-replace, with a few manual tweaks.

It's a bit of a beast to review. I separated out the user-facing copy changes into the first commit.

## Demo Video or Screenshot
Screenshot 2023-11-30 at 4 35 57 PM

## Testing Plan
Relying on automated tests here.

## Checklist


apps/admin/frontend/src/screens/election_screen.tsx
```diff
@@ -66,8 +66,8 @@ export function ElectionScreen(): JSX.Element {
         {isElectionManagerAuth(auth) && (
           <React.Fragment>
             <P>
-              Save the Ballot Package to USB to configure VxCentralScan or
-              VxScan.
+              Save the election package to the USB drive to configure
+              VxCentralScan or VxScan.
             </P>
```
adghayes, line 71: I wonder if we should add VxMark to this list, because it now requires an election package, or if we should be non-specific, as different customers will have different machines.

apps/admin/frontend/src/screens/unconfigured_screen.tsx
```diff
@@ -123,24 +123,23 @@ export function UnconfiguredScreen(): JSX.Element {
...
         <FileInputButton
-          accept=".json,application/json"
-          onChange={handleVxElectionFile}
+          accept=".zip,application/zip"
+          onChange={handleSetupPackageFile}
+          buttonProps={{ variant: 'primary' }}
         >
-          Select Existing Election Definition File
+          Select Election Package
         </FileInputButton>
```
adghayes, line 131: So we're not distinguishing between "Setup Package" and "Election Package"? And they were the same the whole time?

adghayes (approved):

apps/admin/frontend/src/screens/election_screen.tsx
jonahkagan: @mattroe thoughts?

mattroe: "VxSuite components"?

jonahkagan: Updated

PR #4323, opened 2023-12-01
Ballot Count Sheet Counts CSV + Frontend

### LINKED ISSUE #4274: VxAdmin: Ballot Count Report Sheet Breakdown

adghayes:
_review by commit_

## Overview

Closes #4274.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/37960853/85e66bcf-1ac3-48ef-a458-6e7613f6eee2

## Testing Plan

The usual.

## Checklist

### DIFF
--- a/libs/fixtures/src/data/electionFamousNames2021/index.ts
+++ b/libs/fixtures/src/data/electionFamousNames2021/index.ts
@@ -1,4 +1,59 @@
+import { Election, GridLayout } from '@votingworks/types';
+import { election } from './election.json';
+import { asElectionDefinition } from '../../util';
+
 export { election, electionDefinition } from './election.json';
 export * as electionJson from './election.json';
 export * as machineMarkedBallotPage1 from './bmd-ballot-complete-p1.jpg';
 export * as machineMarkedBallotPage2 from './bmd-ballot-complete-p2.jpg';
+
+const partialMockGridPosition = {
+  type: 'option',
+  side: 'front',
+  column: 0,
+  row: 0,
+  contestId: 'mayor',
+  optionId: 'sherlock-holmes',
+} as const;
+
+const mockMultiSheetGridLayouts: GridLayout[] = [
+  {
+    ballotStyleId: '1',
+    optionBoundsFromTargetMark: {
+      left: 0,
+      right: 0,
+      top: 0,
+      bottom: 0,
+    },
+    gridPositions: [
+      {
+        ...partialMockGridPosition,
+        sheetNumber: 1,
+      },
+      {
+        ...partialMockGridPosition,
+        sheetNumber: 2,
+      },
+      {
+        ...partialMockGridPosition,
+        sheetNumber: 3,
+      },
+    ],
+  },
+];
+
+/**
+ * Election with mock multi-sheet grid layouts. The layouts are invalid and
+ * cannot be used for scanning, but the data structure is useful for testing.
+ */
+export const multiSheetElection: Election = {
+  ...election,
+  gridLayouts: mockMultiSheetGridLayouts,
+};
+
+/**
+ * Election definition with mock multi-sheet grid layouts. The layouts are invalid and
+ * cannot be used for scanning, but the data structure is useful for testing.
+ */
+export const multiSheetElectionDefinition =
+  asElectionDefinition(multiSheetElection);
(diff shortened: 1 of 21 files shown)


adghayes:
@jonahkagan @mattroe - need feedback on how I'm including "Other Options" in the ballot count report builder before I lock this in!

mattroe:
@adghayes What about just report by "Sheet" as an option? I recognize it's not a perfect dimension like the others, but I think the simplicity in the interface could outweigh the possible inconsistency.

adghayes:
I've never liked this idea because it doesn't align with how the reports are actually generated. The sheet counts and the group by parameters work fundamentally differently. For example, in the video you can see how the group by parameters force us to re-generate the report whereas the "by sheet" option just changes the report appearance. And FWIW, the checkbox would only appear for multi-sheet elections.

But you've said this several times in the past, and @jonahkagan gave it a thumbs up, so I will relent.

libs/fixtures/src/data/electionFamousNames2021/index.ts
```diff
@@ -1,4 +1,59 @@
...
+];
+
+/**
+ * Election with mock multi-sheet grid layouts. The layouts are invalid and
+ * cannot be used for scanning, but the data structure is useful for testing.
+ */
+export const multiSheetElection: Election = {
+  ...election,
+  gridLayouts: mockMultiSheetGridLayouts,
+};
```
jonahkagan, line 52: Could we use the existing HMPB fixtures for this? I think it's prudent to try to avoid proliferation of different election fixtures

libs/utils/src/elections.ts
```diff
@@ -1,7 +1,7 @@
 import { Optional, unique } from '@votingworks/basics';
 import { Election } from '@votingworks/types';
 
-export function getElectionSheetCount(election: Election): Optional<number> {
+export function getMaxSheetsPerBallot(election: Election): Optional<number> {
```
jonahkagan, line 4: Should this function just take the maximum sheet number value from the grid layouts instead of counting the length of unique sheet numbers?

jonahkagan (approved):

libs/fixtures/src/data/electionFamousNames2021/index.ts
adghayes: We need a multi-sheet fixture that can be used in both the frontend and the backend - would we want to import those fixtures from `hmpb/render-backend` into frontends? I also think this fixture is pretty low maintenance, since it's just based off of another fixture.

jonahkagan: Most of those ballot fixtures are multi-sheet.

I also don't think there's a point in maintaining `electionFamousNames` as a non-grid-layouts fixture. I guess what I'm ultimately recommending is that we convert all of our base fixtures to the VxDesign-exported election package.

jonahkagan: But that's not in scope for this PR

adghayes: Are you sure the fixtures are mostly multi-sheet? If I search the repo for `"sheetNumber": 1` I get thousands of hits, but if I search the repo for `"sheetNumber": 2` I get zero hits.

jonahkagan: I think it's because the elections are serialized in CDF. Try `"Sheet": 2`

adghayes: Oh, got it.

Still wondering about my first question - is it OK to import fixtures from `hmpb/render-backend` into the frontends?

jonahkagan: Good question. I think if it's just for tests it should be fine, right?

adghayes: For tests it should be fine, but for frontend tests there may be some issues. There are a couple of things to polyfill or stub. I just pushed up a commit with my understanding of how you want this to work. The issue I haven't worked out is how to make it work with `storybook`. It doesn't work with the fixtures from `render-backend` because I think it is trying to import two different copies of `react`, which causes problems.

jonahkagan: Oh shoot that's annoying, I didn't think about the issues with importing into a frontend test environment. I think it's nice to just avoid importing backend packages into frontend environments rather than have a stub/polyfill house of cards. That makes me think maybe we need to find a way to move the fixture data into a different package that doesn't have the actual Node code that generates the fixtures. What do you think about that? It also doesn't have to happen in this PR - I'm ok with either keeping the polyfill approach you have now or abandoning ship for the moment and going back to what you had originally.

PR #4324, opened 2023-12-02
fix(logs): create intermediate diretories for log exports

benadida:
Once we moved to `cp`, we also have to create intermediate directories.


carolinemodic (approved):
do we need to give permissions for the user to do mkdir in prod?

benadida:
> do we need to give permissions for the user to do mkdir in prod?

I don't believe so because of how we mount USB sticks, but I will be double-checking this before merging.

benadida:
confirmed on prod hardware by manually creating the directory as `vx-services` user, and then running the export logs function from the UI, and it works.

PR #4333, opened 2023-12-04
fix(eslint-plugin-vx): support `Result` type aliasing

eventualbuddha:
## Overview
Previously we only checked if the type as literally `Result`, but it didn't support type aliases like `type CustomResult = Result<T, E>;`. This fixes that.

## Demo Video or Screenshot
Function returns `Promise<ExportDataResult>`:
image

| Before (no error) | After (ESLint error) |
|-|-|
|image|image|

## Testing Plan
- [x] New automated tests
- [x] Manual testing in VS Code


jonahkagan (approved):
Nice catch

PR #4345, opened 2023-12-05
[i18n] BMD: Switch ballot styles when voter changes language

kofi-q:
## Overview

_Easer to review by commit_

With the recently added support for language-specific ballot styles, we need to update the active ballot style in the voter session whenever the voter changes their preferred display language to make sure the printed paper ballot and CVR records reflect the voter's chosen language.

This change adds the necessary server API and client logic to update the voter session auth object with the respective ballot style ID in response to language changes.

Also updating the ballot style selection section on the `PollWorkerScreen` to display only the default English ballot style variations. This initial selection will determine which related ballot style is chosen when the voter changes the display language.

## Demo Video or Screenshot
https://github.com/votingworks/vxsuite/assets/264902/7737b989-452a-44da-ae58-e0756570d1ab

## Testing Plan
- New tests for the new API endpoint and clients, shared utils, and updated UI pieces.
- Manual testing for language switching.

## Checklist


apps/mark-scan/frontend/src/api_auth.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const queryClient = new QueryClient({
+  defaultOptions: QUERY_CLIENT_DEFAULT_OPTIONS,
+});
+const mockBackendApi: ApiClient = {
+  ...createApiClient(),
+  updateCardlessVoterBallotStyle: jest.fn(),
+};
+
+function TestHookWrapper(props: { children: React.ReactNode }) {
```
arsalansufi, line 19: Nice! I know we haven't historically written tests for the react-query integration, directly (we've just relied on coverage through higher-level tests).

Instead of putting this test in an `api_auth.test.tsx` sub-file, we could add it to the existing `api.test.tsx` file, so that this boilerplate is available for other tests of the react-query integration, should anyone else choose to add any.

arsalansufi (approved):
LGTM!

apps/mark-scan/frontend/src/api_auth.test.tsx
arsalansufi: > i think what i actually want is a glue test for the frontends using useBallotStyleManager to verify the integration with the api client mutation

That makes more sense to me!

PR #4356, opened 2023-12-06
[i18n/Audio] Add backend store and API logic for audio clips

kofi-q:
## Overview

Fleshing out logic for loading audio clips into backend DBs and fetching via the backend APIs.

Simplifying the API method a bit to just return the list of matching audio clip records (instead of the previous mapping of audio ID to base64 data) - simplifies the store retrieval logic and will make it easier to implement batch fetching in the client in upcoming PRs.

## Testing Plan
- New tests for ballot package parsing, store get/set, and glue code tests for the app/lib integration.

## Checklist


arsalansufi (approved):
LGTM!

PR #4357, opened 2023-12-06
Integrate accessible controller

kshen0:
## Overview

Can be reviewed by commit.

https://github.com/orgs/votingworks/projects/23/views/75?pane=issue&itemId=47630897

* Adds a daemon to listen to button press signal from the accessible controller and emit keypress events when detected.
* Removes the `gamepad` implementation, which uses the now-irrelevant [Gamepad API](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API)

This PR includes changes from https://github.com/votingworks/vxsuite/pull/4393 that I squashed in a rebase to make rebasing on `main` easier.

Once https://github.com/votingworks/vxsuite/pull/4432 is merged this branch will probably need a rebase on `main` and updates to `Cargo.toml`

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/2220b4ab-32f2-4d9c-a7ad-c8462df4112a

## Testing Plan
- added tests to daemon
- repurposed existing gamepad tests to test for keypress handling

## Checklist

- [x I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.


apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -0,0 +1,408 @@
...
+        .open();
+
+    println!("Opened controller serial port at {DEVICE_PATH}");
+
+    // Create virtual device for keypress events
+    let mut device = create_virtual_device();
+    println!("Created virtual device");
+
+    // Wait for virtual device to register
+    thread::sleep(Duration::from_secs(1));
```
jonahkagan, line 307: Is there a way to be sure the virtual device registered?

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -0,0 +1,408 @@
...
+                    // Timeout error just means no event was sent in the current polling period
+                    Err(ref e) if e.kind() == io::ErrorKind::TimedOut => (),
+                    Err(e) => eprintln!("{e:?}"),
+                }
+            }
+        }
+        Err(e) => {
+            eprintln!(r#"Failed to open "{DEVICE_PATH}". Error: {e}"#);
+            ::std::process::exit(1);
+        }
```
jonahkagan, line 336: It seems like there are a few different ways we handle unknown errors in this series of match blocks:
- panic (L324)
- eprintln and keep looping (L329)
- eprintln and exit 1 (L335)

I just wanted to make sure that was intentional. My instinct would be to crash in all cases unless there's a specific reason to keep going (which maybe could be documented with a comment)

jonahkagan (approved):

apps/mark-scan/accessible-controller/src/controllerd.rs
eventualbuddha: I agree with this. Is exiting really the approach you want? I assume we want this daemon to always be running, so it'd be better to log the error and continue unless the error is truly unrecoverable. Have you thought about how this daemon will be run?

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -0,0 +1,408 @@
...
+            loop {
+                match port.read(serial_buf.as_mut_slice()) {
+                    Ok(size) => match handle_command(&mut device, &serial_buf[..size]) {
+                        Ok(_) => (),
+                        Err(e) => match e {
+                            CommandError::UnsupportedButton(_) => {
+                                println!("[warn] {e}")
+                            }
+                            _ => panic!("Error handling command: {e}"),
+                        },
```
eventualbuddha, line 325: This might be a bit more idiomatic:
```suggestion
                        Err(e) if matches!(e, CommandError::UnsupportedButton(_)) => {
                            println!("[warn] {e}")
                        }
                        Err(e) => {
                            panic!("Error handling command: {e}")
                        }
```

apps/mark-scan/accessible-controller/src/controllerd.rs
kshen0: Seems like it's possible with the underlying `uinput` C library but that functionality isn't ported to the `uinput` crate we use. It's also possible with just the filesystem, though that requires some parsing: https://stackoverflow.com/questions/34064047/how-to-get-name-path-of-uinput-created-device

I'll experiment with this a little when I'm in front of the hardware again.

apps/mark-scan/accessible-controller/src/controllerd.rs
kshen0: Yeah agree with this:
1. eprintln and exit should be replaced with panic for unrecoverable errors
2. we should only exit on truly unrecoverable errors

I think in practice this means we only panic when the device can't be opened after `n` retries, which would indicate a hardware failure. Command errors are likely to be recoverable so we should not panic on those.

PR #4361, opened 2023-12-07
perf(scan): stop copying scanned images in memory

eventualbuddha:
## Overview
I added the `trimBlackFromTopAndBottomOfImage` function when I first added the Custom state machine because the images we got back from the scanner had somewhat tall black header and footer portions that interfered with the interpreter as it was then. Maybe. It's not entirely clear because _this function doesn't seem to do what it says it does_. It's looking for rows at the top and bottom of the image that it can trim off, but in practice it always seems to trim zero rows on both ends, effectively just copying the images. We can tell it's not trimming because the `imageHeight` calculation of the trimmed image is wrong: it's using `image.imageBuffer.length / offsetPerRow`, but the numerator should be the _new_ length, not the _old_ length. If it actually were cropping any rows then the width × height of the trimmed image would not equal the number of byes in the buffer and `createImageData` would reject it. (In trying to make the function _actually_ do cropping, I had exactly this happen.)

I believe maybe the old TS interpreter didn't handle the header/footer but the Rust one does. I'm not sure, and I don't think it's worth digging too deeply into. This theory is supported by the fact that the normalized images are properly cropped even with this removed.

## Demo Video or Screenshot
| | Front | Back |
|-|-|-|
| Original | ee454c39-7855-484c-ad9b-7ba64721a6f8-front | ee454c39-7855-484c-ad9b-7ba64721a6f8-back |
| Normalized | fe189f4e-b57a-415a-b990-df22afb8e679-front | fe189f4e-b57a-415a-b990-df22afb8e679-back |

## Testing Plan
- [x] Manual testing with NH 7/12/22 "Test Ballot".

### DIFF
--- a/apps/scan/backend/src/scanners/custom/state_machine.ts
+++ b/apps/scan/backend/src/scanners/custom/state_machine.ts
@@ -11,7 +11,6 @@ import {
   FormMovement,
   FormStanding,
   ImageColorDepthType,
-  ImageFromScanner,
   ImageResolution,
   openScanner,
   ScannerStatus,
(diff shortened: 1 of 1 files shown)


jonahkagan (approved):
I saw that function a few times and thought it was odd that it was there in the TS given that we do all our heavy image manipulation in Rust. But I never looked closely. Nice find

PR #4364, opened 2023-12-07
[libs/ui] Use consistent disabled button styling for touch UIs

kofi-q:
## Overview

While working on some BMD UI changes, noticed the new disabled styling for voter-facing buttons now retains the filled variant colours in the disabled state. Seemed hard to distinguish between the active and inactive state, even for non-impaired vision, so proposing a return to the previous styling -- I'm open to pushback to leave as is, though.

Only tweaking the `touch` styling -- leaving `desktop` stlying intact.

### Touch Styling
| Before  | After |
| Screenshot 2023-12-07 at 08 30 48  | Screenshot 2023-12-07 at 08 31 12  |

### Desktop Styling
Screenshot 2023-12-07 at 08 50 13

### Touch SegmentedButton Styling
| Enabled  | Disabled |
| Screenshot 2023-12-11 at 10 37 07 | Screenshot 2023-12-11 at 10 37 16 |

## Testing Plan
- Manual + existing tests as regression guard

## Checklist

### DIFF
--- a/libs/ui/src/segmented_button.tsx
+++ b/libs/ui/src/segmented_button.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import styled, { css, useTheme } from 'styled-components';
 
-import { Button } from './button';
+import { Button, ButtonProps } from './button';
 
 /** Props for {@link SegmentedButton}. */
 export interface SegmentedButtonProps<T extends SegmentedButtonOptionId> {
@@ -73,6 +73,26 @@ const OptionsContainer = styled.span<OptionsContainerProps>`
   ${(p) => p.theme.colorMode === 'desktop' && desktopStyles}
 `;
 
+type OptionButtonProps<T extends SegmentedButtonOptionId> = ButtonProps<T> & {
+  selected: boolean;
+};
+
+const optionButtonTouchStyles = css<OptionButtonProps<SegmentedButtonOptionId>>`
+  &[disabled] {
+    background-color: ${(p) =>
+      p.selected ? p.theme.colors.primary : undefined};
+    color: ${(p) => (p.selected ? p.theme.colors.onPrimary : undefined)};
+  }
+`;
+
+// TODO(kofi): Worth extracting a shared OptionButton or ToggleButton for cases
+// like this (e.g. segmented button, checkbox button, radio button, etc).
+const OptionButton = styled(Button)<OptionButtonProps<SegmentedButtonOptionId>>`
+  ${(p) => p.theme.colorMode !== 'desktop' && optionButtonTouchStyles}
+` as unknown as new <T extends SegmentedButtonOptionId>() => React.Component<
+  OptionButtonProps<T>
+>;
+
 /**
  * Renders a list of options as labelled buttons, where only one option can be
  * selected at a time.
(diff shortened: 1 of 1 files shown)


libs/ui/src/button.tsx
```diff
@@ -294,9 +294,18 @@ function disabledStyles(p: ThemedStyledButtonProps): CSSObject {
...
+  const foregroundColor = isInverse(color)
+    ? colors.background
+    : colors.foreground;
+
   return {
+    backgroundColor,
     borderStyle: 'dashed',
-    borderColor: 'currentColor',
+    color: foregroundColor,
+    borderColor: foregroundColor,
```
jonahkagan, line 308: Thanks for calling this out. I think you're right that it doesn't look disabled enough when it maintains the existing background color.

I made this change because of components like `SegmentedButton` and `RadioGroup` that need to show which option was selected even when disabled. But I think we don't use those components in their disabled state very often (if at all) currently in touch screen mode, so I think it's more important to prioritize regular buttons looking disabled.

That being said, I do think that leaves a hole in our implementation of `SegmentedButton`/`RadioGroup` in touch mode for the time being.

libs/ui/src/button.tsx
```diff
@@ -294,9 +294,18 @@ function disabledStyles(p: ThemedStyledButtonProps): CSSObject {
...
 
+  const backgroundColor = isInverse(color)
+    ? colors.foreground
+    : colors.background;
+  const foregroundColor = isInverse(color)
+    ? colors.background
+    : colors.foreground;
+
   return {
+    backgroundColor,
```
jonahkagan, line 305: I wonder if we should just make the background transparent? I.e. convert it to `outline` fill mode? Not a huge difference in touch mode since there aren't other colors to layer it over.

libs/ui/src/button.tsx
```diff
@@ -294,9 +294,18 @@ function disabledStyles(p: ThemedStyledButtonProps): CSSObject {
     };
   }
 
+  const backgroundColor = isInverse(color)
+    ? colors.foreground
+    : colors.background;
+  const foregroundColor = isInverse(color)
+    ? colors.background
+    : colors.foreground;
```
jonahkagan, line 302: `colors.foreground` is deprecated. Sorry I didn't mark it deprecated - I was planning to just remove it soon. Instead use `colors.onBackground`.

jonahkagan (approved):

libs/ui/src/button.tsx
kofi-q: Ah, that makes sense! I've always thought of `role=option` elements to have different semantics from `button`s, so in this case, I'd propose that those components override the default styling to suit their semantics for the disabled state (in the same way we map their `selected` states to a `primary` button variant to begin with). Wouldn't want the `option` components to influence `Button` styling, IMO

Will take a look at those elements and see if it's low-lift enough to update them as part of this PR.

libs/ui/src/button.tsx
kofi-q: I think we want to guarantee that the text within the button has the desired contrast against the button background, so I'm hesitant to go with `transparent` on the off chance the button's rendered on top of some other background color - WDYT?

libs/ui/src/button.tsx
kofi-q: Thanks for the heads-up! Will add the deprecation notice while I'm here

libs/ui/src/button.tsx
jonahkagan: Yeah I think that makes sense!

libs/ui/src/button.tsx
jonahkagan: I think I've set up the color system roles to guarantee that won't be an issue. Otherwise `outline` fill buttons would always be at risk of having contrast issues. And many of the buttons utilize `outline` fill - it's the default.

But I don't think it's a big deal here either way because in touch mode we pretty much only have one background color that buttons go on (`colors.background`). I don't think we even use the inverse case (though I think it's still good to support it to be safe).

Issue #4322, opened 2023-12-01
CVR sync required screen sometimes spuriously appears (and disappears on its own)


arsalansufi:
https://votingworks.slack.com/archives/C060FEVB2L8/p1701461432646639

arsalansufi:
Merged a potential fix! Though I want to keep this issue open until I can verify that the issue really is no more in prod

arsalansufi:
On second thought, though time + QA will tell if the issues are truly solved, going to mark this as done! Can always open new issues if need be

PR #4365, opened 2023-12-07
Ensure that continuous export operations do not interleave and avoid spurious "CVR Sync Required" screens

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4322

This PR addresses various continuous export edge cases, that I think explain some of the odd behaviors that we've seen in prod. Unfortunately, the issues that we've seen, like unexpected crashes and spurious "CVR Sync Required" screens, have been hard to repro with consistency in dev (VxDev or VM). So I'm partly hypothesizing. There may be other sources of problems out there.

More on the specific fixes in GitHub comments.

## Testing Plan

- [x] Lots of manual testing
- [x] Updated automated tests

## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/custom/state_machine.ts
+++ b/apps/scan/backend/src/scanners/custom/state_machine.ts
@@ -39,6 +39,7 @@ import {
 import { SheetInterpretationWithPages } from '@votingworks/ballot-interpreter';
 import { exportCastVoteRecordsToUsbDrive } from '@votingworks/backend';
 import { UsbDrive } from '@votingworks/usb-drive';
+import { Mutex } from '@votingworks/utils';
 import { interpret as defaultInterpret, InterpretFn } from '../../interpret';
 import { Store } from '../../store';
 import {
@@ -370,6 +371,11 @@ function storeInterpretedSheet(
   return addedSheetId;
 }
 
+/**
+ * A mutex to ensure that continuous export operations happen sequentially and do not interleave
+ */
+const continuousExportMutex = new Mutex(0);
+
 async function recordAcceptedSheet(
   store: Store,
   usbDrive: UsbDrive,
@@ -389,14 +395,17 @@ async function recordAcceptedSheet(
     // Gets reset to false within exportCastVoteRecordsToUsbDrive
     store.setIsContinuousExportOperationInProgress(true);
   });
-  (
-    await exportCastVoteRecordsToUsbDrive(
+
+  const exportResult = await continuousExportMutex.withLock(() =>
+    exportCastVoteRecordsToUsbDrive(
       store,
       usbDrive,
       [assertDefined(store.getSheet(sheetId))],
       { scannerType: 'precinct' }
     )
-  ).unsafeUnwrap();
+  );
+  exportResult.unsafeUnwrap();
+
   debug('Stored accepted sheet: %s', sheetId);
 }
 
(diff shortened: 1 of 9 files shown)


apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -370,6 +371,11 @@ function storeInterpretedSheet(
   return addedSheetId;
 }
 
+/**
+ * A mutex to ensure that continuous export operations happen sequentially and do not interleave
+ */
+const continuousExportMutex = new Mutex(0);
```
eventualbuddha, line 377: Created https://github.com/votingworks/vxsuite/pull/4368 so you don't have to provide a dummy value.

eventualbuddha, line 377: This feels like it should be a class/object property rather than a module-level global variable.

apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -389,14 +395,17 @@ async function recordAcceptedSheet(
     // Gets reset to false within exportCastVoteRecordsToUsbDrive
     store.setIsContinuousExportOperationInProgress(true);
   });
-  (
-    await exportCastVoteRecordsToUsbDrive(
+
+  const exportResult = await continuousExportMutex.withLock(() =>
```
eventualbuddha, line 399: To test this, it feels like we'd want to call `recordAcceptedSheet` two or more times without awaiting the results, then verify that everything is sane after we've awaited them all in batch.

apps/scan/backend/src/scanners/custom/state_machine.ts
arsalansufi: Good call! Made it a local variable within the `buildMachine` function: https://github.com/votingworks/vxsuite/pull/4365/commits/ce81444b6d527d81a2286217017d8a8b9a243d64

arsalansufi (commented):
Oops, left some explanatory comments on the PR before opening that I forgot to submit!

apps/scan/backend/src/scanners/custom/state_machine.ts
arsalansufi: Good point - this is making me realize that the final export operation handled by the polls-close operation should also be mutex-gated. Let me move the mutex up to the level of the `Store` or the `Workspace`

libs/backend/src/cast_vote_records/file_system_utils.ts
```diff
@@ -29,11 +29,14 @@ export async function updateCreationTimestampOfDirectoryAndChildrenFiles(
     return entry.name;
   });
   for (const fileName of fileNames) {
+    // Use execFile instead of fs.copyFile because fs.copyFile has issues in prod when the source
+    // and destination paths are both on a USB drive
```
eventualbuddha, line 33: I would love to know why this is the case.

eventualbuddha (approved):

libs/backend/src/cast_vote_records/file_system_utils.ts
arsalansufi: Yeah, it remains a mystery to me, and Ben, too, who found the issue (https://github.com/votingworks/vxsuite/pull/4262) 🤔

apps/scan/backend/src/scanners/custom/state_machine.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/4365/commits/8fac56efb2b8f0ed34096da940f54665b61c072e

apps/scan/backend/src/scanners/custom/state_machine.ts
arsalansufi: Yup exactly what I was going for

PR #4368, opened 2023-12-07
fix(utils): ensure `Mutex` internal async queue is cleared

eventualbuddha:
## Overview
@arsalansufi discovered that the private `asyncQueue` property of `Mutex` only ever grew and was never drained. This isn't great, and represents a memory leak. The only question is whether it's a fast or slow memory leak, which @arsalansufi is investigating.

While I was here I added some more testing, especially testing that covers the cases we care about: either with a value or without a value. For the case without a value I made it so you can do `new Mutex()` and just get one whose value is `void`. This avoids needing to give it a dummy value like `0`.

## Demo Video or Screenshot
n/a

## Testing Plan
- [x] New automated tests.

### DIFF
--- a/libs/utils/src/mutex.ts
+++ b/libs/utils/src/mutex.ts
@@ -1,4 +1,4 @@
-import { deferred, Optional } from '@votingworks/basics';
+import { assertDefined, deferred, Optional } from '@votingworks/basics';
 
 interface LockResult<T> {
   value: T;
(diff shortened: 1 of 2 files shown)


libs/utils/src/mutex.ts
```diff
@@ -29,17 +31,27 @@ export class Mutex<T> {
...
           throw new Error('unlock called more than once');
         }
         unlockCalled = true;
         this.locked = false;
-        for (const resolve of this.asyncQueue) {
+
+        while (!this.locked && this.asyncQueue.length > 0) {
+          const resolve = assertDefined(this.asyncQueue.shift());
           resolve();
         }
```
arsalansufi, line 56: So one more thing that I'm wondering is, in the `asyncLock` function below:

https://github.com/votingworks/vxsuite/blob/52933c0c9e56e577515e48eece72da8a7aed2809/libs/utils/src/mutex.ts#L53-L66

Do we need to move the creation of the `deferred` promise into the `for` loop? I'm reasoning through the following case:

- Entity 0 has the lock, and entities 1 and 2 are waiting to acquire the lock.
- Entity 0 releases the lock, and the blocking promises are resolved.
- Entity 1 acquires the lock.
- Entity 2 fails to acquire the lock and calls `this.asyncQueue.push(resolve)`.
- But the `await promise` doesn't pause execution because the promise was already resolved.
- Infinite loop until entity 1 releases the lock.

I'm still confirming the behavior of `deferred` but thought I'd run the question by you while I verify, myself

arsalansufi (approved):
LGTM! Enjoyed pairing on this!

PR #4371, opened 2023-12-07
[VxDesign] Add node/util browser polyfill

kofi-q:
## Overview

Recently added a `readline` dependency in `libs/backend` -- this caused frontends to break due to their dependency on backend types. Only caught the ones that broke in CI, so I accidentally missed VxDesign.

## Testing Plan
- Local run to verify


jonahkagan (approved):
Thanks for fixing this

Issue #4216, opened 2023-11-15
Only re-sync the one CVR that failed to be written to the USB drive, instead of all CVRs, when possible


arsalansufi:

arsalansufi:
Covered in https://github.com/votingworks/vxsuite/pull/5486!

PR #4387, opened 2023-12-11
Address two more continuous export edge cases

arsalansufi:
_Easiest reviewed by commit_

## Overview

After merging https://github.com/votingworks/vxsuite/pull/4365, I thought of two more continuous export edge cases worth addressing.

### Edge case 1

In that previous PR, I was focused on avoiding spurious "CVR Sync Required" screens. But a specific race condition could also cause us to miss a case where we _should_ display the screen. Consider the following sequence:

1. Ballot 1 is accepted, and we set `isContinuousExportOperationInProgress` to true
2. Export operation 1 (for ballot 1) starts
3. Ballot 2 is accepted, and we set `isContinuousExportOperationInProgress` to true
4. Export operation 1 completes and sets `isContinuousExportOperationInProgress` to false
5. Export operation 2 (for ballot 2) starts
6. Export operation 2 fails

We'd fail to display the "CVR Sync Required" screen on reboot because `isContinuousExportOperationInProgress` is false. The issue is that we're using a single boolean to track the status of all of the continuous export operations. What we really want is a list of all running operations. On reboot, we can check if this list is empty or not.

Side note: This change will make it much easier to implement https://github.com/votingworks/vxsuite/issues/4216.

### Edge case 2

In https://github.com/votingworks/vxsuite/pull/4365, I updated a number of operations (polls close, full export, etc.) to first wait for any pending continuous export operations. I forgot to cover mode-switching in that update.

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually with a slow USB 2.0 drive, which makes all of these edge cases more likely

## Checklist

### DIFF
--- a/apps/scan/backend/src/app_export.test.ts
+++ b/apps/scan/backend/src/app_export.test.ts
@@ -1,0 +1,186 @@
+import {
+  getCastVoteRecordExportDirectoryPaths,
+  readCastVoteRecordExport,
+} from '@votingworks/backend';
+import { assertDefined, ok } from '@votingworks/basics';
+import { mockOf } from '@votingworks/test-utils';
+import { CVR } from '@votingworks/types';
+import {
+  BooleanEnvironmentVariableName,
+  convertCastVoteRecordVotesToTabulationVotes,
+  getCurrentSnapshot,
+  getFeatureFlagMock,
+} from '@votingworks/utils';
+
+import { scanBallot, withApp } from '../test/helpers/custom_helpers';
+import { configureApp } from '../test/helpers/shared_helpers';
+
+jest.setTimeout(30_000);
+
+const mockFeatureFlagger = getFeatureFlagMock();
+
+jest.mock('@votingworks/utils', (): typeof import('@votingworks/utils') => {
+  return {
+    ...jest.requireActual('@votingworks/utils'),
+    isFeatureFlagEnabled: (flag) => mockFeatureFlagger.isEnabled(flag),
+  };
+});
+
+beforeEach(() => {
+  mockFeatureFlagger.resetFeatureFlags();
+  mockFeatureFlagger.enableFeatureFlag(
+    BooleanEnvironmentVariableName.SKIP_ELECTION_PACKAGE_AUTHENTICATION
+  );
+});
+
+test('continuous CVR export, including polls closing', async () => {
+  await withApp(
+    {},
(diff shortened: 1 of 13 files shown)


apps/scan/backend/src/app_export.test.ts
```diff
@@ -0,0 +1,186 @@
...
+});
+
+beforeEach(() => {
+  mockFeatureFlagger.resetFeatureFlags();
+  mockFeatureFlagger.enableFeatureFlag(
+    BooleanEnvironmentVariableName.SKIP_ELECTION_PACKAGE_AUTHENTICATION
+  );
+});
+
+test('continuous CVR export, including polls closing', async () => {
```
arsalansufi, line 36: Copied these tests over, as is, from `app_config.test.ts` to `app_export.test.ts`. So no new code in this commit

apps/scan/backend/src/app_export.test.ts
```diff
@@ -162,6 +162,56 @@ test('continuous CVR export, including polls closing, followed by a full export'
   );
 });
 
+test('continuous CVR export with a mode switch in between', async () => {
```
arsalansufi, line 165: This test covers "Edge case 2" in the PR description. I'd ideally add a test case for "Edge case 1," too. But recreating the relevant race condition would require lots of mocking / tricky timing, so I decided to punt.

apps/scan/backend/src/app_export.test.ts
```diff
@@ -162,6 +162,56 @@ test('continuous CVR export, including polls closing, followed by a full export'
...
+          (castVoteRecordResult) =>
+            castVoteRecordResult.unsafeUnwrap().castVoteRecord
+        );
+        if (i === 0) {
+          // Directory before mode switch
+          expect(castVoteRecords).toHaveLength(2);
+        } else {
+          // Directory before mode switch
+          expect(castVoteRecords).toHaveLength(1);
+        }
```
adghayes, line 209: We know that they didn't interleave, in this case, simply because they ended up in the right directories?

adghayes (approved):

apps/scan/backend/src/app_export.test.ts
arsalansufi: Yup, exactly! And also just from the lack of errors / crashes. If they did interleave, lots of things could fail, e.g. one export process assumes that a directory is there, but then a second export process deletes it, etc.

PR #4230, opened 2023-11-17
add term description to tally reports

adghayes:
## Overview

Closes #4024. Adds term description, the same string which appears on the ballot, to tally reports. Also modifies the styling in the case where there is no term description to be slightly less cramped.

## Demo Video or Screenshot

|  Type  | Scanned Only  | Manual Too |
| ------------- | ------------- | --------- |
| Before | Screen Shot 2023-11-17 at 1 09 39 PM | Screen Shot 2023-11-17 at 1 09 31 PM |
| After, no term  | Screen Shot 2023-11-17 at 1 02 43 PM | Screen Shot 2023-11-17 at 1 02 36 PM |
| After, term | Screen Shot 2023-11-17 at 1 02 14 PM | Screen Shot 2023-11-17 at 1 01 52 PM |

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [x] I have added a screenshot and/or video to this PR to demo the change
- [x] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #4388, opened 2023-12-11
Add `Contest.termDescription` to CDF election definition

### LINKED ISSUE #4384: Add term to election definition
https://votingworks.slack.com/archives/C060FEVB2L8/p1702317572796779?thread_ts=1701988828.801589&cid=C060FEVB2L8

jonahkagan:
## Overview
Fixes #4384 

_Review by commit_

VxDesign supports specifying a free-text term label for candidate contests to be shown under the contest title (`Contest.termDescription`) - e.g. "For two years". Previously, we didn't include this field in the exported election definition, since, at the time, the rest of VxSuite didn't need it for anything.

However, we later wanted to add this term label to the tally reports in VxAdmin (#4230). Since the field was optional, we never noticed that it wasn't actually being exported from VxDesign.

This PR adds functionality to convert the `termDescription` field from VXF to CDF so that it will be exported from VxDesign. CDF has an `Office` object that can be associated with a `CandidateContest`. The `Office` object has an associated `Term` object with a `label` field, which is the closest match to our use case.

Note: `Term.label` is a plain string, not internationalized text. @kofi-q - we'll likely need to use the workaround you came up with for the election date for this field as well.

## Demo Video or Screenshot
N/A

## Testing Plan
- Updated fixtures and automated tests
- Manually exported an election from VxDesign and confirmed the term was included
## Checklist


libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -217,9 +221,30 @@ export function convertVxfElectionToCdfBallotDefinition(
...
+        '@type': 'BallotDefinition.Office',
+        '@id': officeId(contest.id),
+        Name: text(contest.title), // Not used, but required
+        Term: {
+          '@type': 'BallotDefinition.Term',
+          Label: assertDefined(contest.termDescription),
+        },
+      })
+    );
+
```
adghayes, line 242: Might it be vaguely confusing to another (hypothetical) consumer of this data that we're only creating offices when there's a term description?

libs/types/src/cdf/ballot-definition/vx-schema.json
```diff
@@ -63,6 +63,13 @@
         "IssuerAbbreviation": {
           "type": "string"
         },
+        "Office": {
+          "items": {
+            "$ref": "#/definitions/BallotDefinition.Office"
+          },
+          "minItems": 0,
+          "type": "array"
+        },
```
adghayes, line 72: We revived this from the original schema?

libs/utils/src/tabulation/tabulation.test.ts
```diff
@@ -975,7 +975,12 @@ describe('tabulateCastVoteRecords', () => {
 
   test('with expected groups', async () => {
     const resultsByMethodAndPrecinct = await tabulateCastVoteRecords({
-      cvrs: cvrs.slice(0, 1),
+      cvrs: [
+        cvrs.find(
+          (cvr) =>
+            cvr.precinctId === 'precinct-1' && cvr.votingMethod === 'precinct'
+        )!,
+      ],
```
adghayes, line 983: thank you for flxing my lazy flaky test :)

adghayes (approved):

libs/types/src/cdf/ballot-definition/vx-schema.json
jonahkagan: Yep, just copy-pasted in from `nist-schema.json`, which lives in this repo

libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan: That's fair. It seemed more confusing to me to have offices with no purpose. Without term they have no information. Do you think we should change it?

adghayes: Not really, just making sure it's intentional.

PR #4390, opened 2023-12-11
VxAdmin: Add District Filtering

### LINKED ISSUE #4275: VxAdmin: District Report Filters

adghayes:
## Overview

Closes #4275.  Adds ability to filter by "District" to the report builders. This could include things like wards, cities, sub-jurisdictions, etc.

Under the hood, it just converts a list of districts to a list of ballot styles. We don't have districts arranged in the election definition in such a way that we can do groupings by them, just filtering.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/37960853/d7062f5d-91ad-4beb-88cb-04afed43c278

## Testing Plan

I had trouble thinking of a good way to write resilient tests for this. At each endpoint in `app.ts` on the backend, we convert from the frontend filter format, which contains district IDs, to a the more basic version. Ideally testing would ensure this always happens, but it's laborious because there's four endpoints to test and no existing fixtures. I chose to use assertions so that manual testing would reveal any issues, but didn't do the work to write four extensive automated tests.

Ideally the conversion could happen at some middle layer, like logging, but that middle layer would have to be aware of the election definition.

## Checklist

### DIFF
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -85,6 +85,7 @@ import {
 } from './cast_vote_records';
 import { generateBallotCountReportCsv } from './exports/csv_ballot_count_report';
 import { adjudicateWriteIn } from './adjudication';
+import { convertFrontendFilter as convertFrontendFileUtil } from './util/filters';
 
 const debug = rootDebug.extend('app');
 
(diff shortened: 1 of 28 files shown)


apps/admin/backend/src/app.ts
```diff
@@ -85,6 +85,7 @@ import {
 } from './cast_vote_records';
 import { generateBallotCountReportCsv } from './exports/csv_ballot_count_report';
 import { adjudicateWriteIn } from './adjudication';
+import { convertFrontendFilter as convertFrontendFileUtil } from './util/filters';
```
kofi-q, line 88: typo - was this supposed to be `convertFrontendFilterUtil`?

kofi-q (approved):
LGTM!

apps/admin/backend/src/app.ts
adghayes: Yup! Just fixed, thank you.

PR #4393, opened 2023-12-12
various improvements

eventualbuddha:
Here are a variety of changes that you might want to incorporate. I didn't break them down by commit, but will be adding explanatory comments.

### DIFF
--- a/apps/mark-scan/accessible-controller/src/controllerd.rs
+++ b/apps/mark-scan/accessible-controller/src/controllerd.rs
@@ -1,14 +1,17 @@
 use serialport::{self, SerialPort};
-use std::{fmt::Display, io, thread, time::Duration};
+use std::{
+    io, thread,
+    time::{Duration, Instant},
+};
 use uinput::{event::keyboard, Device};
 
-const POLL_INTERVAL_MS: u64 = 10;
+const POLL_INTERVAL: Duration = Duration::from_millis(10);
+const MAX_ECHO_RESPONSE_WAIT: Duration = Duration::from_secs(1);
 const UINPUT_PATH: &str = "/dev/uinput";
 const DEVICE_PATH: &str = "/dev/ttyACM1";
 const DEVICE_BAUD_RATE: u32 = 9600;
-const EXPECTED_PACKET_LEGNTH: usize = 7;
 
-#[derive(Debug, PartialEq, thiserror::Error)]
+#[derive(Debug, thiserror::Error)]
 enum ParseError {
     #[error("Unexpected command data length: {0}")]
     UnexpectedPacketSize(usize),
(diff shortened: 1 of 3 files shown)


apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -1,14 +1,18 @@
...
-use std::{fmt::Display, io, thread, time::Duration};
+use std::{
+    io, thread,
+    time::{Duration, Instant},
+};
 use uinput::{event::keyboard, Device};
 
-const POLL_INTERVAL_MS: u64 = 10;
+const POLL_INTERVAL: Duration = Duration::from_millis(10);
+const MAX_ECHO_RESPONSE_WAIT: Duration = Duration::from_secs(1);
```
eventualbuddha, line 9: Rust has types for representing durations of time. This ensures we can't mess up the units.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -1,14 +1,18 @@
...
 const UINPUT_PATH: &str = "/dev/uinput";
 const DEVICE_PATH: &str = "/dev/ttyACM1";
 const DEVICE_BAUD_RATE: u32 = 9600;
-const EXPECTED_PACKET_LEGNTH: usize = 7;
+const EXPECTED_PACKET_LENGTH: usize = 7;
 
-#[derive(Debug, PartialEq, thiserror::Error)]
```
eventualbuddha, line 11: This `PartialEq` was only needed for the `assert_eq!` in tests, and was causing problems with wrapping `uinput::Error`, so I removed it and used different methods in the tests.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -19,39 +23,151 @@ enum ParseError {
     #[error("Invalid action received: {0}")]
     InvalidAction(u8),
     #[error("Error occurred when sending keypress event")]
-    KeypressError(),
+    KeypressError(#[from] uinput::Error),
     #[error("Button value invalid: {0}")]
-    Button(#[from] ButtonError),
+    InvalidButton(u8),
+    #[error("Button unsupported: {0:?}")]
+    UnsupportedButton(Button),
```
eventualbuddha, line 29: There seemed to be a distinction between "invalid" and "not yet supported". I'm not sure this error type is the right place for this, and maybe it becomes something other than `ParseError` if it stays here? 🤷

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -19,39 +23,151 @@ enum ParseError {
...
+    #[error("Invalid CRC16 value: expected {expected:x}, actual {actual:x}")]
+    InvalidCrc { expected: u16, actual: u16 },
 }
 
-#[derive(Debug, PartialEq, thiserror::Error)]
-struct ButtonError {
-    invalid_code: u8,
+#[derive(Debug, num_enum::TryFromPrimitive, Clone, Copy)]
```
eventualbuddha, line 34: Uses the `num_enum` crate to avoid manually re-iterating all the values.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -19,39 +23,151 @@ enum ParseError {
...
+        }
+
+        Ok(Self {
+            payload: bytes[PAYLOAD_OFFSET..PAYLOAD_OFFSET + payload_length as usize].to_vec(),
+        })
     }
 }
 
+#[derive(Debug)]
+struct ButtonStatusCommand {
```
eventualbuddha, line 104: This and `EchoCommand` help keep our byte wrangling isolated.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -19,39 +23,151 @@ enum ParseError {
...
+}
+
+impl From<ButtonStatusCommand> for Vec<u8> {
+    fn from(command: ButtonStatusCommand) -> Self {
+        let mut bytes = Self::new();
+        bytes.push(CommandId::ButtonStatus as u8);
+        bytes.extend_from_slice(&(2 as u16).to_be_bytes());
+        bytes.push(command.button as u8);
+        bytes.push(command.action as u8);
+        bytes.extend_from_slice(&crc16::State::<crc16::XMODEM>::calculate(&bytes).to_be_bytes());
```
eventualbuddha, line 116: Incorporated the crc16 checks because why not?

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -19,39 +23,151 @@ enum ParseError {
...
+        if bytes.len() < NON_PAYLOAD_BYTES {
+            return Err(ParseError::UnexpectedPacketSize(bytes.len()));
+        }
+
+        match CommandId::try_from(bytes[0]) {
+            Ok(CommandId::ButtonStatus) => (),
+            _ => return Err(ParseError::InvalidCommand(bytes[0])),
+        }
+
+        let payload_length = u16::from_be_bytes([bytes[1], bytes[2]]);
```
eventualbuddha, line 143: Not sure if it should be big-endian or little-endian, but this is a better way to do the conversion.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -178,65 +231,58 @@ fn handle_command(device: &mut Device, data: &[u8]) -> Result<(), ParseError> {
         }
     }
 
-    return send_key(device, key);
+    send_key(device, key)
```
eventualbuddha, line 234: Rust-ism: no final `return` when you can do it implicitly (flagged by `cargo clippy -- -W clippy::pedantic`).

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -178,65 +231,59 @@ fn handle_command(device: &mut Device, data: &[u8]) -> Result<(), ParseError> {
...
-        Err(error) => eprintln!("{:?}", error),
+        Err(error) => eprintln!("{error:?}"),
     }
 
     let mut serial_buf: Vec<u8> = vec![0; 20];
 
-    // Allow 1s for echo response
-    for _n in 1..(1000 / POLL_INTERVAL_MS) {
+    let start_time = Instant::now();
+    loop {
```
eventualbuddha, line 248: I can't try this out so I'm not sure it's right, but this seemed like a better way to poll for up to 1s.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -317,7 +359,6 @@ mod tests {
             0x00,
             0x00,
         ];
-        let result = handle_command(&mut device, &data);
-        assert!(result.is_ok());
```
eventualbuddha, line 321: This won't have a good error message. Better to just unwrap the `Result`.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -19,39 +22,152 @@ enum ParseError {
...
+
+        match CommandId::try_from(bytes[0]) {
+            Ok(CommandId::ButtonStatus) => (),
+            _ => return Err(ParseError::InvalidCommand(bytes[0])),
+        }
+
+        let payload_length = u16::from_be_bytes([bytes[1], bytes[2]]);
+        if payload_length as usize != 0x0002 {
+            return Err(ParseError::UnexpectedDataLength(payload_length));
         }
```
kshen0, line 146: Here we are checking the reported data length whereas in the `&[u8]` to `EchoCommand` conversion we check reported data length against actual data length. I think we can just copy [this block](https://github.com/votingworks/vxsuite/pull/4393/files?diff=split&w=0#diff-28962332f625636ccfbf5757f6bf5564fd28993866795fab136ddd4ed6895de1R80-R83) (or make it a function). I can make this change

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -178,65 +231,59 @@ fn handle_command(device: &mut Device, data: &[u8]) -> Result<(), ParseError> {
...
                 println!("Received valid echo command response");
                 return;
             }
             Err(ref e) if e.kind() == io::ErrorKind::TimedOut => (),
-            Err(e) => eprintln!("Error reading echo response: {:?}", e),
+            Err(e) => eprintln!("Error reading echo response: {e:?}"),
+        }
+
+        if start_time.elapsed() >= MAX_ECHO_RESPONSE_WAIT {
+            break;
```
kshen0, line 265: Thinking about this more, we should exit if no echo response is received in 1s. Is `panic!` the best way to do this? Seems like extra boilerplate to pass an error up to `main` if `main` is just going to exit from there.

kshen0 (approved):

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -19,39 +22,152 @@ enum ParseError {
...
+    UnsupportedButton(Button),
+    #[error("Invalid CRC16 value: expected {expected:x}, actual {actual:x}")]
+    InvalidCrc { expected: u16, actual: u16 },
 }
 
-#[derive(Debug, PartialEq, thiserror::Error)]
-struct ButtonError {
-    invalid_code: u8,
+#[derive(Debug, num_enum::TryFromPrimitive, Clone, Copy)]
+#[repr(u8)]
```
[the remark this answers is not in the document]
eventualbuddha: It's a hint to `num_enum`.

apps/mark-scan/accessible-controller/src/controllerd.rs
eventualbuddha: In general it's better to return `Result`. There are affordances to make it less annoying to pass them along.

Issue #4326, opened 2023-12-04
Address scanning edge cases (double-sheet w wind-up and wind-up-less double-sheet)


arsalansufi:

arsalansufi:
Though time + QA will tell if the issues are truly solved, going to mark this as done!

PR #4395, opened 2023-12-12
Protect against wind-up-less double-sheet

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4326

See **#issue-custom-scanner-state-machine-v3** in Slack for lots of additional context.

In our stress testing of the scanner state machine, we've identified edge cases that are unlikely to come up in real use but that we want to protect against regardless. One such issue we've dubbed "wind-up-less double-sheet".

When we prep a second ballot just in front of the scanner before accept of a first ballot, and the second ballot isn't close enough to trigger the sensors for the "remove your ballot" notice, the first ballot is ejected without the typical backwards wind-up. The second ballot can then be passed through the scanner without being counted.

Because we only consider the front of the scanner to have paper if all of the front sensors detect paper, the scanner state machine doesn't recognize the second ballot in the scanner if it only triggers a subset of the front sensors. But the scanner itself does see the paper and, on accept, doesn't wind-up first so as not to bump into it.

This PR mitigates the issue by tweaking the use of ANDs and ORs in paper status conditions.

## Testing Plan

- [x] Extensive manual testing

### DIFF
--- a/apps/scan/backend/src/scanners/custom/state_machine.ts
+++ b/apps/scan/backend/src/scanners/custom/state_machine.ts
@@ -181,45 +181,62 @@ async function closeCustomClient({ client }: Context) {
 function scannerStatusToEvent(
   scannerStatus: ScannerStatus
 ): ScannerStatusEvent {
-  const frontHasPaper =
+  const allFrontSensorsDetectPaper =
     scannerStatus.sensorInputLeftLeft === SensorStatus.PaperPresent &&
     scannerStatus.sensorInputCenterLeft === SensorStatus.PaperPresent &&
     scannerStatus.sensorInputCenterRight === SensorStatus.PaperPresent &&
     scannerStatus.sensorInputRightRight === SensorStatus.PaperPresent;
-  const backHasPaper =
-    scannerStatus.sensorOutputLeftLeft === SensorStatus.PaperPresent &&
-    scannerStatus.sensorOutputCenterLeft === SensorStatus.PaperPresent &&
-    scannerStatus.sensorOutputCenterRight === SensorStatus.PaperPresent &&
+  const someFrontSensorsDetectPaper =
+    scannerStatus.sensorInputLeftLeft === SensorStatus.PaperPresent ||
+    scannerStatus.sensorInputCenterLeft === SensorStatus.PaperPresent ||
+    scannerStatus.sensorInputCenterRight === SensorStatus.PaperPresent ||
+    scannerStatus.sensorInputRightRight === SensorStatus.PaperPresent;
+
+  const someBackSensorsUnderRollersDetectPaper =
+    scannerStatus.sensorOutputCenterLeft === SensorStatus.PaperPresent ||
+    scannerStatus.sensorOutputCenterRight === SensorStatus.PaperPresent;
+  const someBackSensorsPastRollersDetectPaper =
+    scannerStatus.sensorOutputLeftLeft === SensorStatus.PaperPresent ||
     scannerStatus.sensorOutputRightRight === SensorStatus.PaperPresent;
+  const someBackSensorsDetectPaper =
+    someBackSensorsUnderRollersDetectPaper ||
+    someBackSensorsPastRollersDetectPaper;
 
   if (scannerStatus.isScannerCoverOpen) {
     return { type: 'SCANNER_DISCONNECTED' };
   }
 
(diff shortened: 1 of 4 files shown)


apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -181,45 +181,62 @@ async function closeCustomClient({ client }: Context) {
...
       return { type: 'SCANNER_JAM_DOUBLE_SHEET' };
     }
+
     return { type: 'SCANNER_JAM' };
   }
 
-  if (!frontHasPaper && !backHasPaper) {
+  // Paper past the back rollers but caught on, say, the emergency ballot bag should not be
+  // considered still in the scanner. Continuing to spin the rollers won't clear the paper. Rather,
+  // scanning the next ballot will.
```
arsalansufi, line 223: This behavior isn't new; I'm just maintaining it more explicitly now

apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -181,45 +181,62 @@ async function closeCustomClient({ client }: Context) {
...
   }
 
-  if (!frontHasPaper && backHasPaper) {
+  if (!someFrontSensorsDetectPaper && someBackSensorsUnderRollersDetectPaper) {
     return { type: 'SCANNER_READY_TO_EJECT' };
   }
 
-  // If we get here frontHasPaper and backHasPaper are true
+  // The above conditions make it such that this condition is equivalent to:
+  // someFrontSensorsDetectPaper && someBackSensorsUnderRollersDetectPaper
```
arsalansufi, line 239: This last case is the one that's relevant for this issue, since the first ballot (the one to be accepted) is at the back of the scanner and the second ballot is at the front. Previously, this condition was only hit if `allFrontSensorsDetectPaper && allBackSensorsDetectPaper`

jonahkagan: Might be worth an explicit check of the condition here followed by a `throwIllegalValue` or something like that to ensure we've covered all the cases

apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -181,45 +181,62 @@ async function closeCustomClient({ client }: Context) {
...
+    someBackSensorsUnderRollersDetectPaper ||
+    someBackSensorsPastRollersDetectPaper;
 
   if (scannerStatus.isScannerCoverOpen) {
     return { type: 'SCANNER_DISCONNECTED' };
   }
 
   if (scannerStatus.isPaperJam || scannerStatus.isJamPaperHeldBack) {
-    if (!frontHasPaper && !backHasPaper) {
+    if (!someFrontSensorsDetectPaper && !someBackSensorsDetectPaper) {
```
jonahkagan, line 210: This is stricter than before - intentional?

jonahkagan (approved):

apps/scan/backend/src/scanners/custom/state_machine.ts
arsalansufi: It was intentional - I figured we'd want users to completely clear the scanner before considering the jam cleared. Though on second thought, we might not want to block on the back sensors past the rollers being clear, as per my comments elsewhere in this PR.

I think I'll switch to `!someFrontSensorsDetectPaper && !someBackSensorsUnderRollersDetectPaper`. Interestingly, with that change in place, we'll never be using the signal from the back sensors past the rollers.

jonahkagan: 🤔 does make you wonder what those sensors are good for

apps/scan/backend/src/scanners/custom/state_machine.ts
arsalansufi: Good call! Updated in https://github.com/votingworks/vxsuite/pull/4395/commits/f2dc1adb0e2eafd59e96bc838a85fe2ee053e6ae and added a fast-check test for extra confidence

apps/scan/backend/src/scanners/custom/state_machine.ts
arsalansufi: Updated in https://github.com/votingworks/vxsuite/pull/4395/commits/f2dc1adb0e2eafd59e96bc838a85fe2ee053e6ae

And yeah I've been wondering the same 🤔

PR #4396, opened 2023-12-12
Protect against double-sheet w wind-up

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4326

See **#issue-custom-scanner-state-machine-v3** in Slack for lots of additional context.

In our stress testing of the scanner state machine, we've identified edge cases that are unlikely to come up in real use but that we want to protect against regardless. One such issue we've dubbed "double-sheet w wind-up" (distinct from "[wind-up-less double sheet](https://github.com/votingworks/vxsuite/pull/4395)").

When a second ballot is inserted as a first ballot is being accepted, it's possible for the second ballot to be passed through the scanner without being counted. The timing has to be just right to escape the existing safeguards in the state machine. But there are other conditions that also have to be met that we don't quite understand. We've never repro'd with an NH test ballot, only with the (extremely dense) all-bubble ballot. And one can go long stretches of time without being able to repro and then all of a sudden be able to back-to-back. Though we've been suspicious of continuous export as the biggest diff between the m17b image (which we haven't been able to repro with) and current images, we've still been able to repro with continuous export operations commented out.

While we continue to investigate what changed to make this edge case more likely, we have a solid mitigation strategy. During ballot accept, we'll now stop the scanner rollers as soon as the ballot clears the scanner. The scanner rollers currently roll for ~1.3s on accept, which gives a second ballot plenty of time to pass through. Now, the rollers will only roll for as long as necessary, up to 1.3s. I haven't hardcoded a time; the rollers will stop dynamically, when the no-paper status is seen. I've also increased the paper status polling interval during ballot accept  to detect issues as quickly as possible.

## Testing Plan

- [x] Extensive manual testing

### DIFF
--- a/libs/custom-scanner/src/custom_a4_scanner.ts
+++ b/libs/custom-scanner/src/custom_a4_scanner.ts
@@ -247,6 +247,13 @@ export class CustomA4Scanner implements CustomScanner {
         this.withRetries(async () => {
           (await this.createJobInternal()).okOrElse(fail);
 
+          // To stop the motors, the createJobInternal call above is sufficient, as it ends the
+          // current job, if any, before creating a new one. This in and of itself stops the
+          // motors.
+          if (movement === FormMovement.STOP) {
+            return ok();
+          }
+
           return await this.channelMutex.withLock((channel) =>
             formMove(channel, ONLY_VALID_JOB_ID, movement)
           );
--- a/apps/scan/backend/src/scanners/custom/state_machine.ts
+++ b/apps/scan/backend/src/scanners/custom/state_machine.ts
@@ -111,6 +111,7 @@ export type Event = ScannerStatusEvent | CommandEvent;
 
 export interface Delays {
   DELAY_PAPER_STATUS_POLLING_INTERVAL: number;
+  DELAY_PAPER_STATUS_POLLING_INTERVAL_DURING_ACCEPT: number;
   DELAY_PAPER_STATUS_POLLING_TIMEOUT: number;
   DELAY_SCANNING_TIMEOUT: number;
   DELAY_ACCEPTING_TIMEOUT: number;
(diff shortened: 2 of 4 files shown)


libs/custom-scanner/src/custom_a4_scanner.ts
```diff
@@ -247,6 +247,13 @@ export class CustomA4Scanner implements CustomScanner {
         this.withRetries(async () => {
           (await this.createJobInternal()).okOrElse(fail);
 
+          // To stop the motors, the createJobInternal call above is sufficient, as it ends the
+          // current job, if any, before creating a new one. This in and of itself stops the
+          // motors.
```
arsalansufi, line 252: I spent a good bit of time with the scanner's "manuale comandi" before landing on this approach haha. I also looked into the CANC (cancel) and STOP (stop scan) commands, and while those will stop the motors during a scan operation, they expectedly don't stop the motors outside of that context. JOB end appears to be the best way to do so. And also nicely requires the fewest code modifications

jonahkagan: Nice find. Does creating this "empty" job have any side effects afterwards? Does the scanner expect us to send some movement command as part of the job?

jonahkagan (approved):

libs/custom-scanner/src/custom_a4_scanner.ts
arsalansufi: Good q! As far as I can tell, it doesn't. Here's what the command manual has to say about the JOB command:

<table>
job-command
</table>

The other question on my mind has been whether it's okay to leave jobs open like we do. But we've already been doing that, so this PR doesn't change anything on that front.

I've also been wondering whether it would make more sense for us to open a job once and keep using it instead of constantly opening and closing jobs. But what we've been doing has been working, and I don't want to make any risky changes.

eventualbuddha: I think the concept of "jobs" in the Custom scanner is essentially meaningless. It always returns the same job ID and is pretty loose about job handling. It does want to have _some_ handling of jobs, though. I tried to build it in a sane way initially, but it didn't work so I just copied what their driver code did.

PR #4398, opened 2023-12-12
refactor(scan): pass `auth` to `start`

eventualbuddha:
## Overview
We already resolve the other values in `main`, might as well do the same here. This is part of the changes I'm making for making the Custom state machine decide when to scan.

## Demo Video or Screenshot
n/a

## Testing Plan

### DIFF
--- a/apps/scan/backend/src/server.ts
+++ b/apps/scan/backend/src/server.ts
@@ -1,23 +1,13 @@
-import {
-  InsertedSmartCardAuth,
-  InsertedSmartCardAuthApi,
-  JavaCard,
-  MockFileCard,
-} from '@votingworks/auth';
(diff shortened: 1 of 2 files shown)


jonahkagan (approved):

apps/scan/backend/src/server.ts
```diff
@@ -35,27 +25,13 @@ export function start({
   usbDrive,
   workspace,
 }: StartOptions): void {
-  /* c8 ignore start */
-  const resolvedAuth =
-    auth ??
```
kshen0, line 40: What was the use case for supporting `undefined` auth?

eventualbuddha: I think for testing/dependency injection? It doesn't appear to be needed at any rate.

kshen0 (approved):

PR #4403, opened 2023-12-12
[VxScan] Decide when to scan on the backend

eventualbuddha:
## Overview
This change copies the logic from the frontend that controlled whether to trigger a scan into the backend. Most of that logic was in deciding which screen to render, with the last part on the `VoterScreen` itself. In the future it might make sense to render the screen based on the new backend status I'm calling "app flow state", but I wanted to make minimal changes to the frontend as part of this change.

## Demo Video or Screenshot

Note in this video that we never call the `scanBallot` endpoint (which no longer exists):

https://github.com/votingworks/vxsuite/assets/1938/e1615fe5-8162-41a4-a440-d03accb3c74c

## Testing Plan
- [x] Updated a bunch of existing backend tests.
- [x] Manual testing with the NH "Test Ballot" election.
- [x] Added some automated tests around "app flow state" and not triggering scans when we shouldn't (i.e. someone is logged in).

### DIFF
--- a/apps/scan/backend/src/app.ts
+++ b/apps/scan/backend/src/app.ts
@@ -365,11 +365,6 @@ export function buildApi(
       };
     },
 
-    scanBallot(): void {
-      assert(store.getPollsState() === 'polls_open');
-      machine.scan();
-    },
-
     acceptBallot(): void {
       machine.accept();
     },
--- a/apps/scan/backend/src/app.ts
+++ b/apps/scan/backend/src/app.ts
@@ -30,11 +30,7 @@ import {
   Result,
   throwIllegalValue,
 } from '@votingworks/basics';
-import {
-  InsertedSmartCardAuthApi,
-  InsertedSmartCardAuthMachineState,
-  LiveCheck,
-} from '@votingworks/auth';
+import { InsertedSmartCardAuthApi, LiveCheck } from '@votingworks/auth';
 import { UsbDrive, UsbDriveStatus } from '@votingworks/usb-drive';
 import {
   PrecinctScannerStateMachine,
@@ -43,23 +39,10 @@ import {
   PollsTransition,
   PrecinctScannerPollsInfo,
 } from './types';
+import { constructAuthMachineState } from './util/construct_auth_machine_state';
 import { Workspace } from './util/workspace';
 import { getMachineConfig } from './machine_config';
 import { getScannerResults } from './util/results';
-
-function constructAuthMachineState(
-  workspace: Workspace
-): InsertedSmartCardAuthMachineState {
-  const electionDefinition = workspace.store.getElectionDefinition();
-  const jurisdiction = workspace.store.getJurisdiction();
-  const systemSettings =
-    workspace.store.getSystemSettings() ?? DEFAULT_SYSTEM_SETTINGS;
-  return {
-    ...systemSettings.auth,
-    electionHash: electionDefinition?.electionHash,
-    jurisdiction,
-  };
-}
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export function buildApi(
@@ -73,7 +56,7 @@ export function buildApi(
 
   async function getUserRole() {
     const authStatus = await auth.getAuthStatus(
-      constructAuthMachineState(workspace)
+      constructAuthMachineState(workspace.store)
     );
     if (authStatus.status === 'logged_in') {
       return authStatus.user.role;
@@ -85,20 +68,20 @@ export function buildApi(
     getMachineConfig,
 
     getAuthStatus() {
-      return auth.getAuthStatus(constructAuthMachineState(workspace));
+      return auth.getAuthStatus(constructAuthMachineState(workspace.store));
     },
 
     checkPin(input: { pin: string }) {
-      return auth.checkPin(constructAuthMachineState(workspace), input);
+      return auth.checkPin(constructAuthMachineState(workspace.store), input);
     },
 
     logOut() {
-      return auth.logOut(constructAuthMachineState(workspace));
+      return auth.logOut(constructAuthMachineState(workspace.store));
     },
 
     updateSessionExpiry(input: { sessionExpiresAt: Date }) {
       return auth.updateSessionExpiry(
-        constructAuthMachineState(workspace),
+        constructAuthMachineState(workspace.store),
         input
       );
     },
@@ -128,7 +111,7 @@ export function buildApi(
 
     async ejectUsbDrive(): Promise<void> {
       const authStatus = await auth.getAuthStatus(
-        constructAuthMachineState(workspace)
+        constructAuthMachineState(workspace.store)
       );
       return usbDrive.eject(
         authStatus.status === 'logged_in' ? authStatus.user.role : 'unknown'
@@ -141,7 +124,7 @@ export function buildApi(
       assert(!store.getElectionDefinition(), 'Already configured');
 
       const authStatus = await auth.getAuthStatus(
-        constructAuthMachineState(workspace)
+        constructAuthMachineState(workspace.store)
       );
       const electionPackageResult = await readElectionPackageFromUsb(
         authStatus,
--- a/apps/scan/backend/src/app_flow.test.ts
+++ b/apps/scan/backend/src/app_flow.test.ts
@@ -1,4 +1,3 @@
-import { typedAs } from '@votingworks/basics';
 import { buildMockInsertedSmartCardAuth } from '@votingworks/auth';
 import { createMockUsbDrive } from '@votingworks/usb-drive';
 import {
@@ -12,8 +11,7 @@ import { electionGeneralDefinition } from '@votingworks/fixtures';
 import { ALL_PRECINCTS_SELECTION } from '@votingworks/utils';
 import { PrecinctScannerState, TEST_JURISDICTION } from '@votingworks/types';
 import { doesUsbDriveRequireCastVoteRecordSync } from '@votingworks/backend';
-import { getCurrentAppFlowState } from './app_flow';
-import { AppFlowState } from '.';
+import { isReadyToScan } from './app_flow';
 import { Store } from './store';
 import { BALLOT_BAG_CAPACITY } from './globals';
 
@@ -37,13 +35,13 @@ test('setup_card_reader', async () => {
   });
 
   expect(
-    await getCurrentAppFlowState({
+    await isReadyToScan({
       auth,
       store,
       usbDrive: mockUsbDrive.usbDrive,
       precinctScannerState: 'no_paper',
     })
-  ).toEqual(typedAs<AppFlowState>('setup_card_reader'));
+  ).toEqual(false);
 });
 
 test('login_prompt', async () => {
@@ -52,13 +50,13 @@ test('login_prompt', async () => {
   const mockUsbDrive = createMockUsbDrive();
 
   expect(
-    await getCurrentAppFlowState({
+    await isReadyToScan({
       auth,
       store,
       usbDrive: mockUsbDrive.usbDrive,
       precinctScannerState: 'no_paper',
     })
-  ).toEqual(typedAs<AppFlowState>('login_prompt'));
+  ).toEqual(false);
 });
 
 test('card_error', async () => {
@@ -72,13 +70,13 @@ test('card_error', async () => {
   });
 
   expect(
-    await getCurrentAppFlowState({
+    await isReadyToScan({
       auth,
       store,
       usbDrive: mockUsbDrive.usbDrive,
       precinctScannerState: 'no_paper',
     })
-  ).toEqual(typedAs<AppFlowState>('card_error'));
+  ).toEqual(false);
 });
 
 test('unlock_machine (system_administrator)', async () => {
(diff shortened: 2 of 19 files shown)


apps/scan/backend/src/app.ts
```diff
@@ -382,11 +365,6 @@ export function buildApi(
       };
     },
 
-    scanBallot(): void {
-      assert(store.getPollsState() === 'polls_open');
-      machine.scan();
-    },
-
```
eventualbuddha, line 389: The backend API method is gone.

apps/scan/backend/src/app_flow.ts
```diff
@@ -0,0 +1,155 @@
...
+  isElectionManagerAuth,
+  isPollWorkerAuth,
+  isSystemAdministratorAuth,
+} from '@votingworks/utils';
+import { BALLOT_BAG_CAPACITY } from './globals';
+import { Store } from './store';
+import { constructAuthMachineState } from './util/construct_auth_machine_state';
+import { AppFlowState } from './types';
+
+export async function getCurrentAppFlowState({
```
eventualbuddha, line 16: The structure of this function was taken from the frontend's `AppRoot`. Instead of returning e.g. `<SetupCardReaderScreen />` I'm returning an equivalent status of `'setup_card_reader'`.

apps/scan/backend/src/globals.ts
```diff
@@ -30,3 +30,5 @@ export const SCAN_WORKSPACE =
   (NODE_ENV === 'development'
     ? join(__dirname, '../dev-workspace')
     : undefined);
+
+export const BALLOT_BAG_CAPACITY = 7000;
```
eventualbuddha, line 34: For now, this is duplicated.

apps/scan/backend/src/scanners/custom/app_scan.test.ts
```diff
@@ -130,27 +139,19 @@ test('configure and scan hmpb', async () => {
   );
 });
 
-beforeEach(() => {
-  mockFeatureFlagger.enableFeatureFlag(
-    BooleanEnvironmentVariableName.SKIP_ELECTION_PACKAGE_AUTHENTICATION
-  );
-});
```
eventualbuddha, line 137: Moved up. Seemed weird to have a `beforeEach` after a `test`.

apps/scan/backend/src/scanners/custom/app_scan.test.ts
```diff
@@ -130,27 +139,19 @@ test('configure and scan hmpb', async () => {
...
-
 test('configure and scan bmd ballot', async () => {
   await withApp(
     {},
     async ({ apiClient, mockScanner, mockUsbDrive, logger, mockAuth }) => {
       await configureApp(apiClient, mockAuth, mockUsbDrive, { testMode: true });
 
       mockScanner.getStatus.mockResolvedValue(ok(mocks.MOCK_READY_TO_SCAN));
-      await waitForStatus(apiClient, { state: 'ready_to_scan' });
```
eventualbuddha, line 146: In general we never see `ready_to_scan` in tests anymore unless there's something preventing the paper from being scanned, like we need to sync CVRs or someone is logged in.

apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -516,6 +522,37 @@ function buildMachine({
...
+              // scan.
+              precinctScannerState: 'ready_to_scan',
+            });
+            if (flowState === 'ballot:waiting_to_scan') {
+              return { type: 'SCAN' };
+            }
+          }
+
+          return event;
+        })
```
eventualbuddha, line 548: This feels not super in the spirit of the state machine. Thoughts, @jonahkagan?

apps/scan/backend/test/helpers/shared_helpers.ts
```diff
@@ -96,6 +96,13 @@ export async function configureApp(
     type: 'open_polls',
     time: Date.now(),
   });
+
+  mockOf(mockAuth.getAuthStatus).mockImplementation(() =>
+    Promise.resolve({
+      status: 'logged_out',
+      reason: 'no_card',
+    })
+  );
```
eventualbuddha, line 105: The election manager never remembered to take out their card!

apps/scan/frontend/src/screens/voter_screen.tsx
```diff
@@ -33,16 +33,12 @@ export function VoterScreen({
...
+  // The backend waits to receive a command accept a ballot. The frontend
+  // controls when this happens for now. In the future, we should move this to
+  // the backend.
   const acceptBallotMutation = acceptBallot.useMutation();
   useQueryChangeListener(scannerStatusQuery, {
     onChange: (newScannerStatus) => {
-      if (newScannerStatus.state === 'ready_to_scan') {
-        scanBallotMutation.mutate();
-      }
       if (newScannerStatus.state === 'ready_to_accept') {
```
eventualbuddha, line 42: We still trigger accept on the frontend because it may require voter interaction.

apps/scan/backend/src/scanners/custom/state_machine.ts
kshen0: I think ideally `doScanWhenReady` would be replaced with a guarded transition ie. on L860

```
        ready_to_scan: {
          id: 'ready_to_scan',
          entry: [clearError, clearLastScan],
          invoke: pollPaperStatus(),
          on: {
            SCAN: 'scanning',
            SCANNER_NO_PAPER: 'no_paper',
            SCANNER_READY_TO_SCAN: [
              {
                target: 'scanning',
                // Only transition to 'scanning' if the guard (cond) evaluates to true
                cond: async (context, event) => await getCurrentAppFlowState({ ... }) === 'ballot:waiting_to_scan',
              },
              // do nothing if guarded transition fails
              { target: doNothing },
            ]
          },
        },
```

But it looks like async guards are not supported because there's not a good way to handle errors. There are various flavors of "make a new state instead" described [here](https://github.com/statelyai/xstate/issues/533#issuecomment-630627546). So something like 

```
          on: {
            ...
            SCANNER_READY_TO_SCAN: 'evaluate_should_start_scan',
        },
```

then `evaluate_should_start_scan` calls `await getCurrentAppFlowState` in its `entry` or `invoke` and transitions based on the result; basically just making the transition explicit instead of using an intermediate `SCAN` event. I _think_ that might result in ping ponging between `ready_to_scan` and `evaluate_should_start_scan`, so debatable if it's cleaner than the current implementation of `doScanWhenReady`.

kshen0 (commented):
Still reviewing but wanted to leave the comment about guarded transitions

apps/scan/backend/src/scanners/custom/state_machine.ts
jonahkagan: Yeah I think @kshen0's second suggestion about creating an intermediate "checking" state seems like the recommended solution based on [this comment](https://github.com/statelyai/xstate/issues/533#issuecomment-510459432) from the xstate library author. Ping-ponging would be the desired behavior while waiting - that seems right to me.

apps/scan/backend/src/app_flow.ts
jonahkagan: Is the vision that this function will drive the frontend eventually?

apps/scan/backend/src/scanners/custom/app_scan.test.ts
```diff
@@ -536,3 +509,48 @@ test("scannerStatusToEvent's cases are exhaustive and can all be reached", () =>
   expect(eventCounts.get('SCANNER_READY_TO_EJECT')).toBeGreaterThan(0);
   expect(eventCounts.get('SCANNER_READY_TO_SCAN')).toBeGreaterThan(0);
 });
+
+test('scanning paused when election manager card is inserted', async () => {
```
jonahkagan, line 513: Any other guard conditions worth adding a test for? A few that feel relevant to real-world conditions:
- poll worker card inserted
- ballot bag needs replacement

apps/scan/backend/src/app_flow.ts
```diff
@@ -0,0 +1,155 @@
...
+    case 'returning':
+    case 'returned':
+    case 'rejecting':
+    case 'rejected':
+    case 'jammed':
+    case 'both_sides_have_paper':
+    case 'recovering_from_error':
+    case 'double_sheet_jammed':
+    case 'unrecoverable_error':
+      return 'unknown';
```
jonahkagan, line 150: Is there a reason so many of these states (which have meaningful implications for app flow) are mapped to `unknown` here? Is it just that they aren't relevant for deciding if we should scan or not?

apps/scan/frontend/src/screens/voter_screen.tsx
jonahkagan: But this particular case where we don't need voter interaction could be moved to the backend, right? (We don't necessarily need to)

apps/scan/backend/src/app_flow.test.ts
```diff
@@ -0,0 +1,704 @@
...
+});
+
+test('setup_scanner', async () => {
+  const auth = buildMockInsertedSmartCardAuth();
+  const store = Store.memoryStore();
+  const mockUsbDrive = createMockUsbDrive();
+
+  store.setElectionAndJurisdiction({
+    electionData:
+      electionGridLayoutNewHampshireAmherstFixtures.electionDefinition
```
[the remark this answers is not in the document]
eventualbuddha: No reason. Switched.

apps/scan/backend/src/app_flow.ts
eventualbuddha: Yeah, they don't change the decision for whether we should scan or not. This is one of the things that made me think this whole thing is a little overwrought and it should just determine whether we're ready to scan or not. I'd like it to be the thing that controls the frontend state in the future, but we're not really there yet. Thoughts?

jonahkagan: I think it's interesting to think about having the entire app flow state controlled by the backend. On reading this PR, I started thinking harder about what that would be like. It does feel like a big jump (and would seriously set VxScan apart from the other apps). I considered whether we should then move all of that logic into the state machine.

I do think a smaller change where we just have a guard on the backend that checks all of the conditions we care about before initiating a scan seems less intimidating for now. I almost suggested it, but wanted to understand where you were going with this first. Given that you are feeling a similar hesitation, I think we might at least want to explore that option.

apps/scan/backend/src/scanners/custom/state_machine.ts
eventualbuddha: Thanks for pushing on this and the suggestions for improvements. I am pretty happy with the end result: https://github.com/votingworks/vxsuite/pull/4403/commits/36035fee58d1ded90388257e84f0ed568156b893#diff-39166310c5456c6b6d7ed5a61720aeb48b89c2be78b81a4eeebdef3d981f3945R837-R851

apps/scan/backend/src/app_flow.ts
eventualbuddha: Simplified this a bit in 69fcb1be.

apps/scan/backend/src/scanners/custom/app_scan.test.ts
eventualbuddha: Added some more tests in e667a11.

apps/scan/frontend/src/screens/voter_screen.tsx
eventualbuddha: That's true. I didn't change it because I thought having it split between the frontend and backend might be more confusing. That said, I didn't try it. What do you think?

apps/scan/backend/src/app_flow.ts
eventualbuddha: As discussed elsewhere in this PR, that was my initial goal. I've retreated from that with the subsequent changes.

apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -827,7 +834,21 @@ function buildMachine({
...
+        check_ready_to_scan: {
+          invoke: {
+            src: async () => isReadyToScan({ auth, store, usbDrive }),
+            onDone: [
+              {
+                target: 'scanning',
+                cond: (_context, event) => event.data,
+              },
+              {
+                target: 'ready_to_scan',
```
kshen0, line 849: The naming is a little confusing here - we transition back to `ready_to_scan` when `isReadyToScan` returns false. What do you think about delineating eg `hardware_ready_to_scan` and `appStateIsReadyToScan`?

apps/scan/backend/src/app_flow.ts
```diff
@@ -0,0 +1,119 @@
...
+    return false;
+  }
+
+  if (!electionDefinition) {
+    return false;
+  }
+
+  if (isElectionManagerAuth(authStatus)) {
+    return false;
+  }
```
kshen0, line 79: Could we consolidate some of these? It feels like these checks are leftover from when we had specific screens for each reason scanning wasn't ready. Now that it's just a boolean I think we could reduce the number of checks. eg the checks for `!isSystemAdministratorAuth`, `!isPollWorkerAuth`, and `!isElectionManagerAuth` seem redundant with L107 

```
assert(authStatus.status === 'logged_out' && authStatus.reason === 'no_card');
```

and the various logged out checks are redundant with [L54](https://github.com/votingworks/vxsuite/pull/4403/files#diff-d209b5ccb87bea87d1299afbc7f9559a495ae9f5247d16380b060d83ab5d1f8fR54-R56)

(On the other hand it may make sense to keep these very specific checks if we plan to eventually move some decisions about what to render to the backend)

eventualbuddha: Though I think it may be a worthwhile refactor, we're unlikely to do it in the near future. So I agree that simplifying it is best for now.

jonahkagan (approved):
I like how this turned out too!

kshen0 (approved):
LGTM!

PR #4406, opened 2023-12-13
fix(utils): correct `jsonStream` cycle false positive

eventualbuddha:
## Overview
Since I was keeping track of _all_ objects that had been serialized as part of the initial `jsonStream` call, this would fail to serialize:

```ts
const o = {};
jsonStream([o, o]);
```

That's not a cycle, so it should be fine. This change updates the cycle tracking to only look at the ancestors of the current object being serialized.

While I was making this change I remembered that `JSON.stringify` respects `toJSON` being defined on objects, so I added that feature.

## Testing Plan
- [x] New automated tests.

### DIFF
--- a/libs/utils/src/json_stream.ts
+++ b/libs/utils/src/json_stream.ts
@@ -60,12 +60,24 @@ export async function* jsonStream<T>(
   input: JsonStreamInput<T>,
   { compact = true }: JsonStreamOptions = {}
 ): AsyncGenerator<string> {
-  const visitedObjects = new WeakSet();
-
   async function* jsonStreamInternal(
     value: unknown,
-    indentLevel = 0
+    indentLevel: number,
+    ancestorObjects: ReadonlySet<unknown>
   ): AsyncGenerator<string> {
+    if (ancestorObjects.has(value)) {
+      throw new Error('circular reference');
+    }
+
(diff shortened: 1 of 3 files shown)


libs/utils/src/json_stream.ts
```diff
@@ -60,12 +60,24 @@ export async function* jsonStream<T>(
...
   ): AsyncGenerator<string> {
+    if (ancestorObjects.has(value)) {
+      throw new Error('circular reference');
+    }
+
+    if (typeof (value as { toJSON?: () => unknown })?.toJSON === 'function') {
+      const nextAncestorObjects = new Set(ancestorObjects).add(value);
+      return yield* jsonStreamInternal(
+        (value as { toJSON(): unknown }).toJSON(),
+        indentLevel,
```
adghayes, line 76: Does it matter that you're not incrementing the `indentLevel` here? If so, why not?

adghayes (approved):
Thank you for fixing this! I have run across this before but just rebuilt objects instead of fixing this 🙈

libs/utils/src/json_stream.ts
eventualbuddha: Substituting for a different value with `toJSON` doesn't create an additional level of nesting, just replaces the value with another value. We _do_ add the original value to the set of ancestors so we can properly cycle track, but the indent doesn't change.

PR #4407, opened 2023-12-13
Decrease VxSuite test flakiness

arsalansufi:
## Overview

VxSuite tests have gotten quite flaky lately. Using the list of flaky tests on the (super helpful) [CI insights page](https://app.circleci.com/insights/github/votingworks/vxsuite/workflows/test/tests?branch=main) as a guide, I've made a number of small tweaks to decrease flakiness. More on the specifics of the fixes in GitHub comments.

## Testing Plan

- [x] Tested some of the fixes by SSH-ing onto a CI machine for a more representative environment

### DIFF
--- a/apps/scan/backend/package.json
+++ b/apps/scan/backend/package.json
@@ -19,7 +19,7 @@
     "pre-commit": "lint-staged",
     "start": "node ./build/index.js",
     "test": "is-ci test:ci test:watch",
-    "test:ci": "jest --coverage --reporters=default --reporters=jest-junit --maxWorkers=6",
+    "test:ci": "jest --coverage --reporters=default --reporters=jest-junit --maxWorkers=2 # VxScan backend tests (which involve lots of polling and I/O operations) get flaky when we set --maxWorkers higher",
     "test:coverage": "jest --coverage",
     "test:debug": "node --inspect-brk $(which jest) --runInBand --no-cache",
     "test:watch": "jest --watch",
--- a/libs/auth/src/live_check.test.ts
+++ b/libs/auth/src/live_check.test.ts
@@ -64,6 +64,12 @@ test.each<{
       machineId,
       electionHash: isMachineConfiguredForAnElection ? electionHash : undefined,
     });
-    expect(qrCodeValue).toHaveLength(expectedQrCodeValueLength);
+    expect([
+      expectedQrCodeValueLength,
+      // There's a slight chance that the base64-encoded signature within the QR code value is 92
+      // characters long instead of 96. (The length of a base64-encoded string is always a multiple
+      // of 4.)
+      expectedQrCodeValueLength - 4,
+    ]).toContain(qrCodeValue.length);
   }
 );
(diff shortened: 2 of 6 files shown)


apps/scan/backend/package.json
```diff
@@ -19,7 +19,7 @@
     "pre-commit": "lint-staged",
     "start": "node ./build/index.js",
     "test": "is-ci test:ci test:watch",
-    "test:ci": "jest --coverage --reporters=default --reporters=jest-junit --maxWorkers=6",
+    "test:ci": "jest --coverage --reporters=default --reporters=jest-junit --maxWorkers=2 # VxScan backend tests (which involve lots of polling and I/O operations) get flaky when we set --maxWorkers higher",
```
arsalansufi, line 22: My fuzzy theory here is that the VxScan backend tests involve lots of polling and I/O operations (polling because of paper status checks and I/O operations before of continuous export). And setting `--maxWorkers`, i.e. parallelization via worker threads, too high seems to be interfering with those. Also worth noting that worker threads aren't really meant to help with I/O operations. Pulled straight from the [Node docs](https://nodejs.org/api/worker_threads.html):

> Workers (threads) are useful for performing CPU-intensive JavaScript operations. They do not help much with I/O-intensive work. The Node.js built-in asynchronous I/O operations are more efficient than Workers can be.

Our CI build machines seem to have 36 cores (determined by SSH-ing onto one and running `cat /proc/cpuinfo | grep processor | wc -l`), so 6 doesn't feel like it should be a problem. But I SSH-ed onto a build machine and ran some tests. I was able to repro the common failures that we see at 6, 5, and 4 workers. And wasn't able to (in my limited number of test runs) at 3 and below. I landed on 2 to play it safe.

The tradeoff is of course test run time, but even with this change, VxScan backend tests are not the slowest test suite. Not including setup and teardown, the actual test running with 2 workers takes about 90s whereas with 6 workers it took about 45s.

libs/auth/src/live_check.test.ts
```diff
@@ -64,6 +64,12 @@ test.each<{
...
       electionHash: isMachineConfiguredForAnElection ? electionHash : undefined,
     });
-    expect(qrCodeValue).toHaveLength(expectedQrCodeValueLength);
+    expect([
+      expectedQrCodeValueLength,
+      // There's a slight chance that the base64-encoded signature within the QR code value is 92
+      // characters long instead of 96. (The length of a base64-encoded string is always a multiple
+      // of 4.)
+      expectedQrCodeValueLength - 4,
+    ]).toContain(qrCodeValue.length);
```
arsalansufi, line 73: This is closely related to https://github.com/votingworks/vxsuite/pull/4401

apps/scan/backend/package.json
jonahkagan: Thanks for all this context! Seems reasonable to me

jonahkagan (approved):
Thanks for digging into these!

PR #4418, opened 2023-12-14
Avoid warning flicker when switching from test mode to official mode while a CVR sync is required

arsalansufi:
_Easiest reviewed by commit_

## Overview

When switching from test mode to official mode while a CVR sync is required, you may see a flicker of the warning message typically displayed when you try to switch in the reverse direction (from official mode to test mode) without syncing CVRs first.

To repro reliably in dev, I had to slow down USB drive status polling:

https://github.com/votingworks/vxsuite/assets/12616928/5cafee1d-bb67-43d8-8906-f060dc9a0ae2

I think we're seeing this more reliably in prod, probably because our prod machines aren't as performant as our dev machines.

This PR fixes the issue by invalidating the USB drive status React query on mode switch, instead of just waiting for the next poll.

We didn't catch this earlier because the warning used to live in a modal, only surfaced when you actually clicked the "Official Ballot Mode" toggle. Technically, if you clicked the toggle fast enough, you might've been able to catch a flicker of the warning.

## Testing Plan

- [x] Tested manually
- [x] Updated automated tests, though I couldn't repro the flicker easily in a test environment

## Checklist

### DIFF
--- a/apps/scan/frontend/jest.config.js
+++ b/apps/scan/frontend/jest.config.js
@@ -14,7 +14,7 @@ module.exports = {
   coverageThreshold: {
     global: {
       statements: 95,
-      branches: 87,
+      branches: 86,
       functions: 88,
       lines: 95,
     },
--- a/apps/scan/frontend/src/api.ts
+++ b/apps/scan/frontend/src/api.ts
@@ -184,6 +184,9 @@ export const unconfigureElection = {
         await queryClient.invalidateQueries(getConfig.queryKey());
         await queryClient.invalidateQueries(getPollsInfo.queryKey());
         await uiStringsApi.onMachineConfigurationChange(queryClient);
+
+        // If doesUsbDriveRequireCastVoteRecordSync was true, unconfiguring resets it back to false
+        await queryClient.invalidateQueries(getUsbDriveStatus.queryKey());
       },
     });
   },
@@ -196,6 +199,7 @@ export const setPrecinctSelection = {
     return useMutation(apiClient.setPrecinctSelection, {
       async onSuccess() {
         await queryClient.invalidateQueries(getConfig.queryKey());
+
         // Changing the precinct selection after polls open resets polls to closed
         await queryClient.invalidateQueries(getPollsInfo.queryKey());
       },
@@ -233,7 +237,14 @@ export const setTestMode = {
     const queryClient = useQueryClient();
     return useMutation(apiClient.setTestMode, {
       async onSuccess() {
+        // If doesUsbDriveRequireCastVoteRecordSync was true, switching from test mode to official
+        // mode resets it back to false. To avoid a flicker of the warning prompting you to sync
+        // CVRs before you can switch from official mode to test mode, we invalidate this query
+        // first.
+        await queryClient.invalidateQueries(getUsbDriveStatus.queryKey());
+
         await queryClient.invalidateQueries(getConfig.queryKey());
+
         // Changing the mode after polls open resets polls to closed
         await queryClient.invalidateQueries(getPollsInfo.queryKey());
       },
--- a/apps/scan/frontend/src/screens/election_manager_screen.test.tsx
+++ b/apps/scan/frontend/src/screens/election_manager_screen.test.tsx
@@ -370,7 +370,7 @@ test('switching to test mode when ballots have been counted', async () => {
   ).toEqual('false');
 });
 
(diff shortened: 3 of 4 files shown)


apps/scan/frontend/src/api.ts
```diff
@@ -233,7 +237,14 @@ export const setTestMode = {
     const queryClient = useQueryClient();
     return useMutation(apiClient.setTestMode, {
       async onSuccess() {
+        // If doesUsbDriveRequireCastVoteRecordSync was true, switching from test mode to official
+        // mode resets it back to false. To avoid a flicker of the warning prompting you to sync
+        // CVRs before you can switch from official mode to test mode, we invalidate this query
+        // first.
```
arsalansufi, line 243: I did in fact find that, if I placed this query invalidation after the other query invalidations, I was still able to repro. Zooming out, it would be nice if there was a way to batch query invalidations and make a group of them atomic

apps/scan/frontend/src/screens/election_manager_screen.tsx
```diff
@@ -151,10 +151,7 @@ export function ElectionManagerScreen({
   const ballotMode = (
     <P>
       <SegmentedButton
-        disabled={
-          setTestModeMutation.isLoading ||
-          (isCvrSyncRequired && scannerStatus.ballotsCounted > 0)
-        }
+        disabled={setTestModeMutation.isLoading || isCvrSyncRequired}
```
arsalansufi, line 154: `isCvrSyncRequired` can only be true if `ballotsCounted` is greater than 0:

https://github.com/votingworks/vxsuite/blob/d5c32154efabec320cc303c50f8648097c01b9e2/libs/backend/src/cast_vote_records/export.ts#L847-L850

apps/scan/frontend/src/screens/election_manager_screen.test.tsx
```diff
@@ -370,7 +370,7 @@ test('switching to test mode when ballots have been counted', async () => {
   ).toEqual('false');
 });
 
-test('machine cannot be switched to test mode if CVR sync is required and ballots have been counted', async () => {
+test('machine cannot be switched to test mode if CVR sync is required', async () => {
```
arsalansufi, line 373: Updating tests to make it clear that CVR sync can't be required if no ballots have been counted

apps/scan/frontend/src/screens/election_manager_screen.test.tsx
```diff
@@ -393,26 +393,42 @@ test('machine cannot be switched to test mode if CVR sync is required and ballot
...
+
+  apiMock.expectSetTestMode(false);
+  apiMock.expectGetConfig({ isTestMode: false });
+  apiMock.expectGetPollsInfo();
+  apiMock.expectGetUsbDriveStatus('mounted');
+  userEvent.click(officialBallotModeButton);
+  await screen.findByRole('option', {
+    name: 'Test Ballot Mode',
+    selected: false,
+  });
```
arsalansufi, line 424: I tried adding a check here to make sure that we don't see a flicker of the warning, but I'm finding that, with the old code, the warning just doesn't show for long enough for a test to register it

apps/scan/frontend/jest.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
   coverageThreshold: {
     global: {
       statements: 95,
-      branches: 87,
+      branches: 86,
```
arsalansufi, line 17: Removing the unneeded `scannerStatus.ballotsCounted > 0` check seems to have been just enough to bring this down from 87 to 86 (86.96 to be exact)

apps/scan/frontend/src/api.ts
jonahkagan: Hmm, that is interesting. I think the hard part that you're identifying is that key information is split between two queries. 

A few solution ideas:
- Don't render the warning while `mutation.isLoading` is true.
- Consolidate the info we need to a single query (this could be a new query that just answers the question of whether or not to show this warning)

I think the way you've solved it here is totally good enough for now too, but interesting problem!

jonahkagan (approved):
Thanks for digging in and cleaning this up! Tricky business

apps/scan/frontend/src/api.ts
arsalansufi: I like the idea of a consolidated query for situations like this! Basically accomplishes the batching that I describe in my original comment. (Will still merge as is for NH, though, to avoid a riskier change)

PR #4432, opened 2023-12-14
Extract `types-rs` for Rust projects

eventualbuddha:
## Overview
Extracts `types-rs` library with some common types for Rust projects. Review by commit.

## Demo Video or Screenshot
n/a

## Testing Plan


libs/ballot-interpreter/Cargo.toml
```diff
@@ -12,29 +12,25 @@ path = "src/hmpb-rust/lib.rs"
...
 [dev-dependencies]
-proptest = "1.0.0"
-tempfile = "3.3.0"
-
-[dependencies.neon]
-version = "0.10"
-default-features = false
-features = ["napi-6"]
+proptest = { workspace = true }
+tempfile = { workspace = true }
```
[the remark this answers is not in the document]
eventualbuddha: That I don't know. I _think_ it will only download/build the dependencies for the project you're building, but that's just anecdotal.

libs/ballot-interpreter/src/hmpb-rust/debug.rs
```diff
@@ -27,7 +33,7 @@ pub fn draw_qr_code_debug_image_mut(
     detection_areas: &[Rect],
 ) {
     for detection_area in detection_areas {
-        draw_hollow_rect_mut(canvas, (*detection_area).into(), ORANGE);
```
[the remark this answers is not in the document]
eventualbuddha: I'm assuming you're referring to this:
> One should avoid implementing `Into` and implement `From` instead. Implementing `From` automatically provides one with an implementation of `Into` thanks to the blanket implementation in the standard library.

That means that you should avoid `impl Into<A> for B` and instead do `impl From<B> for A` instead. Then you can use `Into::into` because of this stdlib implementation:

```rust
// From implies Into
#[stable(feature = "rust1", since = "1.0.0")]
impl<T, U> Into<U> for T
where
    U: From<T>,
{
    /// Calls `U::from(self)`.
    ///
    /// That is, this conversion is whatever the implementation of
    /// <code>[From]&lt;T&gt; for U</code> chooses to do.
    #[inline]
    fn into(self) -> U {
        U::from(self)
    }
}
```

The above automatically implements `Into<U> for T` whenever `U` has a `From<T>` implementation.

The reason I could no longer use `.into()` here is because of the so-called ["Orphan Rule"](https://github.com/Ixrec/rust-orphan-rules?tab=readme-ov-file#what-are-the-orphan-rules), which requires that we cannot implement e.g. `From` between two types if our crate doesn't define either of them. Previously we had defined `Rect` in `ballot-interpreter` and had a `impl From<Rect> for imageproc::Rect::Rect` [here](https://github.com/votingworks/vxsuite/pull/4432/files#diff-eae8c1e5358ad1d7777f1d998ed1c7f6c3ec9e820c763b3da27fb282e704c8beL210-L214) which allowed us to use `Into::into` to convert between the two. Now that `Rect` is defined in `types-rs`, the Orphan Rule says we cannot define `From` between it and `imageproc::rect::Rect` since we don't define either of those types. The typical workaround for this is either to define a wrapper type to convert through (since we do own that one) or to do what I did and just write a normal function to be called manually.

libs/types-rs/src/ballot_card.rs
jonahkagan, line 1: Where else are these getting used besides lib/ballot-interpreter?

jonahkagan (approved):

libs/ballot-interpreter/Cargo.toml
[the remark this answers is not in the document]
kshen0: Got it. I don't think we should hold up this improvement for specific machine limitations but good to know about this potential complication ahead of time.

libs/types-rs/src/ballot_card.rs
eventualbuddha: Nowhere at present. I could move them back, but at least a `BallotSide` would need to stay in `types-rs` since it's referenced from `election.rs`.

jonahkagan: And is `election.rs` used elsewhere? Just trying to wrap my head around the motivation for extraction

eventualbuddha: I tried moving it and found it works fine, so I will revert that bit and keep only `BallotSide` in `types-rs`.

eventualbuddha: Since `ballot-interpreter` is the only crate that exposes anything to non-Rust code, by definition it is currently the only user of all of the code in `types-rs`. That said, the types in `election.rs` are foundational types and will definitely be used by any future code that needs information from an election definition. I have in mind specifically the BMD ballot interpreter and its QR code votes & metadata encoder/decoder.

kshen0 (approved):

PR #4433, opened 2023-12-14
VxAdmin: Improve configure UI and configure on the backend

### LINKED ISSUE #4412: VxAdmin: election def/setup package file picking

jonahkagan:
## Overview

Closes #4412 

_Review by commit_

- List potential election package zip files on screen
- Remove the "Load Demo Election" function
- Allow using a file picker to select either an election.json or election-package.zip file
- Load and parse the election configuration files on the backend
- Prompt if no USB drive inserted

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/e2279130-ba92-407f-98e6-e8a9f50c0110

## Testing Plan
- Manual testing
- (TODO) update automated tests

## Checklist


eventualbuddha:
First reaction from watching the video: those rows don't have a clear call to action. I'd suggest that you add a "Load" button or whatever verb you think makes sense on each row so that there's a clear CTA.

apps/admin/backend/src/app.ts
```diff
@@ -325,35 +333,61 @@ function buildApi({
...
+
+    // `configure` and `unconfigure` handle changes to the election definition
+    async configure(input: {
+      electionFilePath: string;
+    }): Promise<Result<{ electionId: Id }, ElectionPackageError>> {
+      let electionPackage: ElectionPackage;
+
+      if (input.electionFilePath.endsWith('.json')) {
+        const electionDefinitionResult = safeParseElectionDefinition(
+          await fs.readFile(input.electionFilePath, 'utf8')
```
eventualbuddha, line 365: This reads a file without doing any validation that the path is where we expect it to be. Perhaps the API should concern itself only with filenames and not full paths?

jonahkagan:
>  those rows don't have a clear call to action
I think there's enough visual feedback with the mouse to figure out that the rows are clickable, no?

eventualbuddha:
> I think there's enough visual feedback with the mouse to figure out that the rows are clickable, no?

Yes, but only once you hover over them. They look informational, not selectable.

libs/backend/src/election_package/election_package_io.ts
```diff
@@ -146,10 +316,9 @@ export async function readElectionPackageFromUsb(
     return err('election_package_authentication_error');
   }
 
-  const electionPackage = await readElectionPackageFromBuffer(
-    await fs.readFile(filepathResult.ok())
-  );
-
+  const electionPackage = (
+    await readElectionPackageFromFile(filepathResult.ok())
+  ).unsafeUnwrap();
```
eventualbuddha, line 321: If the containing function returns a `Result`, why are we unwrapping here? Can we just wrap the error?

jonahkagan: This is just preserving the existing behavior. I think the reasoning is that once an election package is signed by VxAdmin, it shouldn't have parsing errors when imported into another machine.

apps/admin/backend/src/app.ts
eventualbuddha: Where do we expect those files to be? Perhaps we only allow files from mounted USB drives in prod?

jonahkagan: That seems like a reasonable constraint to me. @mattroe what do you think?

carolinemodic: Do you mean in the new UI or in the manual old file picker? I think the old file picker should support any path.

mattroe:
> > I think there's enough visual feedback with the mouse to figure out that the rows are clickable, no?
> 
> Yes, but only once you hover over them. They look informational, not selectable.

Agree that it's not the most obvious selectable interface at a glance, but once you navigate I think the feedback is clear through the cursor change and background color change.

apps/admin/backend/src/app.ts
eventualbuddha: The problem with the way it is now is that a compromised frontend can make the backend read any file on the system and load it into memory. This might also include `/dev/` files that would result in a crash of the system or potentially alter system state. For example, passing `/dev/urandom` would cause the backend to read for quite a while until all available memory and swap is used up, making the system less responsive and possibly crashing the backend and/or the frontend.

Personally, I think allowing any path in dev (or with some other flag) and restricting it to mounted USB drives in prod is a reasonable compromise. If we decide to leave this as-is for now I strongly recommend creating an issue to follow up with a fix later.

eventualbuddha (approved):
I proposed some changes here: https://github.com/votingworks/vxsuite/pull/4436. Ignoring the potential security issue I pointed out earlier, this PR LGTM.

apps/admin/backend/src/app.ts
carolinemodic: I'm confused was the reading of the file already on the backend before this change? Is it possible to not change where that is at least for the manual file picker for the purposes of the release? I'm also not understanding if you just mean filenames or relative paths to the root directory @eventualbuddha , we absolutely need to be able to load things from subdirectories on the usb. We do not need to load things outside of a mounted usb in prod, the file permissions in kiosk-browser have made that impossible in prod to date but is that not relevant if we're reading from the backend? We definitely can not introduce the ability to do that with this change.

benadida: Agreed, we need to limit location of files read to USB in prod.

arsalansufi: Picking this up...

If my understanding of this thread is correct, I think we just want to add a prod-specific guard to the backend method, e.g. in the form of an `assert`, as a defense-in-depth measure. Sounds like the kiosk-browser file picker is already configured to only display removable media in prod (curious to find where exactly this logic lives).

arsalansufi: ^ Will be sure to use a proper glob matching library, like we do in `Exporter`, so that we don't mistakenly allow a path like `/media/../dev`

carolinemodic: It displays non-removable media but today when you select it you get a permissions reading error about it, but I'm not sure where the guard for that lives and if moving things ot the backend here changes that

arsalansufi: https://github.com/votingworks/vxsuite/pull/4433/commits/6935fe154a187074c1a0cf260a511aac5ceed7da

PR #4434, opened 2023-12-15
style(admin): remove unneeded `async`

eventualbuddha:
This violates an ESLint rule in my editor, but doesn't seem to be a problem in CI. I don't understand why yet and we may want to look into it, but this at least fixes the issue for now.


eventualbuddha:
This turned out to be a problem on my computer, perhaps a bad ESLint config somehow? Unclear. At any rate, I thought `@typescript-eslint/require-await` would flag this but it doesn't since the return value is a `Promise`.

PR #4436, opened 2023-12-15
feat(admin): include election card in list

eventualbuddha:
## Overview

Shows more information about the election to load to make it easier to choose the right one.

## Demo Video or Screenshot

Screenshot 2023-12-14 at 8 28 38 PM

## Testing Plan

Tested manually with one election package I manually moved from the location that VxAdmin exported it to.


carolinemodic:
@eventualbuddha What happens if there is an error parsing the zip / its not a valid setup package? Or if its like 10gb and takes a long time to unzip? I feel like parsing and reading everything automatically on load adds a lot of possible ways for this change to go wrong.

eventualbuddha:
> @eventualbuddha What happens if there is an error parsing the zip / its not a valid setup package? Or if its like 10gb and takes a long time to unzip? I feel like parsing and reading everything automatically on load adds a lot of possible ways for this change to go wrong.

That's a fair concern. We should have a way to restrict how much data we're willing to read from a file throughout the app, and here in particular.

jonahkagan:
@eventualbuddha do you still want to pursue this? If so, I'm happy to review.

eventualbuddha:
I do, but I want to a @carolinemodic's concern. I was thinking that what I'd really want to do is have a placeholder election card and just load the election info on demand as the elements come into view. That'd require some UI support I don't think we have yet, but it might be a good thing for us to have as we push toward improving performance. cc @kofi-q

jonahkagan:
Gotcha. Yeah this is a bit of a complicated UI challenge. Might want to check in about prioritization before we dive in further. I'm starting to think it might not be important enough to put the time into it right now.

kofi-q:
> I was thinking that what I'd really want to do is have a placeholder election card and just load the election info on demand as the elements come into view.

Can't remember where else something like this came up, but I remember chatting with someone about placeholder loading components a while back. Seems like it would be worth investing in at some point. In the short term, some fixed-size box with a spinner could be a good stopgap

PR #4437, opened 2023-12-15
Improve UX for continuous export recovery

arsalansufi:
_Easiest reviewed by commit_

## Overview

Today, when a continuous export operation fails in the background, say because the USB drive is removed mid-write (highly unlikely in real circumstances but not impossible), we take the heavy handed approach of displaying the catch-all "Something went wrong" screen, which requires users to restart the machine. After restart, the machine displays the "CVR Sync Required" screen to get the USB drive back in sync.

In my testing, I'm finding that not all of these mid-write USB removals require a machine restart to recover from. So this PR switches us over to an alternative, more user-friendly flow.

Now, we'll go straight to the "CVR Sync Required" screen without requiring a restart (once the USB drive is re-inserted if it was removed). Then, only if that resync fails will we prompt users to try 1) restarting, with the help of an on-screen button, or 2) inserting a different USB drive (necessary if the USB drive was corrupted by the mid-write removal).

While I was at it, I also added a number of logs for continuous export that I've been wishing I had for prod QA debugging.

## Demo Video or Screenshot

Here, I remove a USB drive right after ballot accept, reconnect it, try to resync, and find that I do in fact need to restart or use a different USB drive:

https://github.com/votingworks/vxsuite/assets/12616928/f75dfb86-0b38-4c9b-b852-1cbb54f245c6

And here's a freeze frame of the "CVR Sync Required" screen error modal if the attempted resync fails:

error-modal

## Testing Plan

- [x] Updated automated tests
- [x] Tested various permutations of USB drive removal and recovery manually

## Checklist


arsalansufi:
Copying from Slack:

> as per in-person discussion, deciding to play it safe and not incorporate [this] PR (improved continuous export recovery UX) into the next rc. what we have now works, has been well tested, and should be rarely seen regardless.
>
> and I do see some risk in not forcing a restart even if syncing works fine without it (e.g. what if something strange happens at the OS level when a USB drive is removed mid-write that affects other parts of the system)

I do think that we should still get this into main for v4, though!

eventualbuddha:
Is it worth reviving this?

PR #4447, opened 2023-12-19
CVR CDF: remove `vxLayoutHash`

adghayes:
## Overview

Removes `vxLayoutHash` from our CVR reports - instead concatenates the layout hash to the image hash and places it within the CDF hash field for image data.


arsalansufi (approved):
LGTM!

Issue #4443, opened 2023-12-18
VxSuite: limit the size of files when using `readFile`


eventualbuddha:
Loading arbitrary files with unknown size is a bad idea. It can cause the process to become super slow and exhaust system memory, leading to slow UIs and swapping at best and crashes or unrecoverable freezing at worst. In general, we should not even call `readFile` most of the time and instead use a stream that reads the file a chunk at a time. There is some work to be done to make that more ergonomic.

In cases where it does make sense to call `readFile` to get all the file data, I propose that we be required to explicitly specify the maximum amount of data to read as an option to a custom `readFile` in `@votingworks/backend` that takes a `maxSize` option. We should add a custom ESLint rule prohibiting `readFile` and `readFileSync` from `fs`, `fs/promises`, and `fs-extra` with an exception for test files, hinting to use `readFile` from `@votingworks/backend` with a specific `maxSize` value.

Because the new `readFile` would be `async`, this would effectively prohibit the use of `readFileSync` in non-test code as a side-effect, which might be a minor inconvenience.

eventualbuddha:
I did a spike on this here: https://github.com/votingworks/vxsuite/compare/brian/feat/backend/readfile-maxsize

PR #4448, opened 2023-12-19
feat(basics): add `lines` iterator helper

eventualbuddha:
## Overview
The changes made as part of this PR are fairly trivial examples where we're working on relatively small strings, so at best this is a minor improvement in semantic clarity and in correctness (splits on `\r\n` and `\n`). However, the real benefit comes when working on large files where we would previously read the whole file because it's easier. See below for an example.

Helps us avoid running afoul of the problem described in #4443.

## Demo Video or Screenshot
```ts
// old
const contents = fs.readFileSync(path, { encoding: 'utf8' });
const lines = contents.split('\n');
for (const line of lines) {
  …
}

// new
const input = fs.createReadStream(path, { encoding: 'utf8' });
for await (const line of lines(input)) {
  …
}
```

## Testing Plan
- [x] New automated tests for the feature.
- [x] Existing tests covering the refactor.

### DIFF
--- a/libs/basics/src/iterators/lines.ts
+++ b/libs/basics/src/iterators/lines.ts
@@ -1,0 +1,136 @@
+import { Buffer } from 'buffer';
+import { AsyncIteratorPlusImpl } from './async_iterator_plus';
+import { IteratorPlusImpl } from './iterator_plus';
+import { AsyncIteratorPlus, IteratorPlus } from './types';
+
+interface ToString {
+  toString(): string;
+}
+
+function matchAllNewlines(str: string): Iterable<RegExpMatchArray> {
+  return str.matchAll(/\r?\n/g);
+}
+
+function linesSync(iterable: Iterable<ToString>): IteratorPlus<string> {
+  let buffer = '';
+  return new IteratorPlusImpl(
+    (function* gen() {
+      let hasSeenAnyChunks = false;
+
+      for (const element of iterable) {
+        const chunk = element.toString();
+        let offset = 0;
+        for (const newlineMatch of matchAllNewlines(chunk)) {
+          /* istanbul ignore next */
+          if (typeof newlineMatch.index !== 'number') {
+            throw new Error('expected index');
+          }
+
+          const line = buffer + chunk.slice(offset, newlineMatch.index);
+          buffer = '';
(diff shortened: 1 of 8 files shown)


libs/basics/src/iterators/types.ts
```diff
@@ -95,6 +95,24 @@ export interface IteratorPlus<T> extends Iterable<T> {
...
+   * ```ts
+   * const input = ['hello', ' world', '!', '\n', 'goodbye', 'world'];
+   *
+   * expect(iter(input).lines().toArray()).toEqual([
+   *   'hello world!\n',
+   *   'goodbye world'
+   * ]);
+   * ```
+   */
+  lines(): IteratorPlus<T extends string ? string : unknown>;
```
jonahkagan, line 114: Feels a little weird to have a string-specific method on `iter`, which is otherwise oblivious to the contents of the iterable. Since this has such a specific use case (converting a file stream to an iterable of lines), maybe it should just be a standalone function?

libs/basics/src/iterators/lines.ts
```diff
@@ -0,0 +1,136 @@
...
+  iterable: ToString | Iterable<ToString>
+): IteratorPlus<string>;
+
+/**
+ * Pipes elements from `iterable` by collecting them together into a string
+ * until a full line is formed, then yielding that line. The final line may or
+ * may not include a newline.
+ */
+export function lines(
+  iterable: ToString | Iterable<ToString> | AsyncIterable<ToString>
```
eventualbuddha, line 127: @jonahkagan this version is slightly different from the version that was a method on `IteratorPlus`. This one accepts anything that has a `toString(): string` method as well as iterables of the same. That means it can take `ReadStream`s directly, making `lines(fs.createReadStream(path))` work without any casting.

libs/basics/src/iterators/types.ts
jonahkagan: Ah, I forgot about `sum`!

libs/basics/src/iterators/lines.ts
jonahkagan: Cool! Maybe worth a test case to demonstrate that use case

jonahkagan (approved):

PR #4450, opened 2023-12-19
Remove `vxBallotType`

adghayes:
## Overview

When it comes to CDF compliance, there's the good, the bad, and the ugly. This change represents the very ugly. The CDF format completely falls flat when it comes to the basic use case of distinguishing absentee and non-absentee ballots. There's no appropriate field on the CVR and there's no "Other" field to put it either. So in this PR, I put it in the Other field of the CVRSnapshot, which isn't semantically appropriate but it's the closest we have.

As I'm writing this, there's one alternative - we could appropriate a semantically incorrect field (like "BallotPrePrintedId") which we would be using against the spec, but it would avoid a portion of the complexity and footgun-y ness in this PR. Let me know what you think.


libs/utils/src/cast_vote_records.test.ts
```diff
@@ -458,3 +460,95 @@ test('getExportedCastVoteRecordIds', async () => {
...
+    BallotImage: [
+      { '@type': 'CVR.ImageData', Location: 'front.jpg' },
+      { '@type': 'CVR.ImageData', Location: 'back.jpg' },
+    ],
+    CVRSnapshot: [],
+  };
+
+  function buildMockCvr(
+    otherStatus?: string,
+    status = [CVR.CVRStatus.Other]
```
arsalansufi, line 520: nit: Some of the tests that use this function were a bit hard for me to grok. I think I'd have an easier time with labeled inputs:

```ts
function buildMockCvr({
  otherStatus,
  status,
}: {
  otherStatus: string,
  status: string[],
}) {
```

libs/utils/src/cast_vote_records.ts
```diff
@@ -32,6 +38,57 @@ export function getCurrentSnapshot(cvr: CVR.CVR): Optional<CVR.CVRSnapshot> {
   );
 }
 
+/**
+ * Because there is no place for "Absentee" vs. "Precinct" ballot on the CDF
+ * for cast vote records, we shove it into the `OtherStatus` field of the
+ * `CVRSnapshot`. This tool formats that metadata, which can be spread into
+ * a `CVRSnapshot`.
+ */
+export function buildCVRSnapshotBallotStyleMetadata(
```
arsalansufi, line 47: Should this be called `buildCVRSnapshotBallotTypeMetadata` rather than `buildCVRSnapshotBallotStyleMetadata`, since ballot style has another meaning?

arsalansufi (approved):
As clunky as using the other field on the CVR snapshot is, I do prefer this to a semantically incorrect, non-other field because we never know when we may end up needing one of those fields.

Completely agree that this is all less than ideal though 😅

libs/utils/src/cast_vote_records.ts
adghayes: Yes!

libs/utils/src/cast_vote_records.ts
```diff
@@ -32,6 +38,57 @@ export function getCurrentSnapshot(cvr: CVR.CVR): Optional<CVR.CVRSnapshot> {
...
+    ),
+  };
+}
+
+/**
+ * Extracts the ballot type (e.g. absentee vs. precinct) from a cast vote
+ * record. We have the ballot type serialized within the `OtherStatus` field
+ * of `CVRSnapshot`, this helper is for easier access to that data. It returns
+ * undefined if the data is not present - callers are responsible for making
+ * against the data existing if necessary.
```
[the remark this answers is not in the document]
adghayes: Yup!

libs/utils/src/cast_vote_records.ts
```diff
@@ -32,6 +38,57 @@ export function getCurrentSnapshot(cvr: CVR.CVR): Optional<CVR.CVRSnapshot> {
   );
 }
 
+/**
+ * Because there is no place for "Absentee" vs. "Precinct" ballot on the CDF
+ * for cast vote records, we shove it into the `OtherStatus` field of the
+ * `CVRSnapshot`. This tool formats that metadata, which can be spread into
+ * a `CVRSnapshot`.
```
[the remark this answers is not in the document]
adghayes: I meant, like, you can use the spread operator to add the multiple necessary attributes into a `CVRSnapshot` object. Can add a usage example instead of trying to explain it.

PR #4452, opened 2023-12-20
Proof of concept for generating election packages using the background worker

arsalansufi:
A proof of concept for generating election packages using the background worker - will get this over the finish line in the new year!

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -8,7 +8,6 @@ import {
   BallotPaperSize,
   SystemSettings,
   BallotType,
-  ElectionPackageFileName,
 } from '@votingworks/types';
 import express, { Application } from 'express';
 import { assertDefined, find, groupBy, ok, Result } from '@votingworks/basics';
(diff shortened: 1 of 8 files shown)


apps/design/backend/src/app.ts
```diff
@@ -221,35 +221,22 @@ function buildApi({ store }: { store: Store }) {
...
-        electionDefinition.electionData
-      );
-      zip.file(
-        ElectionPackageFileName.SYSTEM_SETTINGS,
-        JSON.stringify(systemSettings, null, 2)
+    exportElectionPackage({ electionId }: { electionId: Id }): void {
+      const electionPackageTaskId = store.createBackgroundTask(
+        'generate_election_package',
+        { electionId }
       );
```
eventualbuddha, line 235: Should multiple requests with the same election ID generate multiple background tasks? Or do we want this endpoint to call something like `store.getOrCreateGenerateElectionPackageBackgroundTask(electionId)` that checks for an existing background task and sets it within a transaction so we don't have two requests where one of them wins and the other becomes a task without anyone using the result?

arsalansufi: Good call! Incorporated this into my actual PR: https://github.com/votingworks/vxsuite/pull/4462

arsalansufi:
Actual PR here: https://github.com/votingworks/vxsuite/pull/4462

PR #4453, opened 2023-12-20
load images from VxAdmin

eventualbuddha:
## Overview
Got image rendering working. This isn't the only way we can do this. We could instead embed the SVGs in the frontend so that there's no fetch required. But if we do have images to load, this could be a way to do it.

## Demo Video or Screenshot
[report.pdf](https://github.com/votingworks/vxsuite/files/13730681/report.pdf)
image


eventualbuddha:
@adghayes, I'm not planning to do anything with this. Just wanted to see what was possible on top of what @jonahkagan built. Feel free to close/integrate it with your stuff/ignore/whatever.

adghayes:
@eventualbuddha is this an idea for how to render the PDF on the frontend? I am thinking that the PDF would be returned in the fetch request, and then could be rendered with existing react PDF display libraries. Maybe that would be prohibitively expensive for the largest reports?

eventualbuddha:
This is just a small tweak to @jonahkagan's PoC that adds the ability to load images. I agree that what we should try is using something like this to generate the PDF on the backend and send it to the frontend for previews. I think it's unlikely that this would be too expensive from a backend-to-frontend communication overhead perspective unless there are many, many pages in the report or a significant number of repeated expensive elements like images. Even then, we could implement pagination if we really needed to.

Issue #4026, opened 2023-10-03
VxDesign: test decks


mattroe:
First version of the test deck feature:
- Export one package that has a test deck for for each ballot style
- Each test deck should contain:
  - Pre-voted ballots (covering every bubble with at least one valid vote)
  - 2 blank ballots
  - 1 overvoted ballot
- There should also be a way to export a tally report that has the expected cumulative results of tabulating the entire test deck package
 
Past test deck reference: https://docs.voting.works/vxsuite/l-and-a-testing/test-deck-printing

Future test deck features omitted for now:
- BMD ballots
- Precinct-specific test decks and tally reports

PR #4454, opened 2023-12-21
VxDesign: Test deck ballots

jonahkagan:
## Overview

Task: #4026 

_Review by commits_

Adds a button to export test decks to VxDesign. The test deck zip file contains a PDF for each precinct, which contains a set of ballots for each ballot style in that precinct. The logic to decide the votes for the ballots is reused from our existing VxAdmin test deck feature.

To complete this feature, @adghayes is working on rendering tally reports on the backend.

## Demo Video or Screenshot
It's quite slow, so I'm skipping the video, but there's not much to see. Plus this UI will change once we move all this work to a background task (#4452)
Screenshot 2023-12-20 at 5 23 14 PM

Sample test deck package
[test-decks-f389e3310f.zip](https://github.com/votingworks/vxsuite/files/13734036/test-decks-f389e3310f.zip)

## Testing Plan
Some manual testing, some updating of automated tests. Explained in more detail in commits/code comments.

## Checklist


jonahkagan:
@adghayes assigning review to you since @eventualbuddha is on vacation

apps/design/frontend/src/export_screen.tsx
```diff
@@ -25,6 +30,20 @@ export function ExportScreen(): JSX.Element {
     );
   }
 
+  function onPressExportTestDecks() {
+    exportTestDecksMutation.mutate(
+      { electionId },
+      {
+        onSuccess: ({ zipContents, electionHash }) => {
+          fileDownload(
```
eventualbuddha, line 38: I'm wondering whether this should be done as a standard browser file download, i.e. an endpoint that streams an `application/octet-stream` rather than using grout for this, which requires that we hold the entire thing in memory both in the frontend and the backend. I'm not sure it's worth it to do now, but perhaps as a follow-up task.

apps/design/backend/src/test_decks.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      const { document, gridLayout } = ballot;
+      const ballotStyle = getBallotStyleById(
+        electionDefinition,
+        gridLayout.ballotStyleId
+      );
+      const contests = getContests({
+        election: electionDefinition.election,
+        ballotStyle,
+      });
+      const maxContestOptions = Math.max(...contests.map(numBallotPositions));
```
eventualbuddha, line 38: While a finished election will presumably always have contest in each ballot style, it's possible that an in-progress election definition will not. In that case `contests` would be `[]` which means `maxContestOptions` will be `-Infinity`, causing the final result of `expectedTestDeckPages` to also be `-Infinity`. This is just a test helper, so it's not a huge deal as we're not dealing with arbitrary user input. However, I'd still suggest this which makes the "empty contests" handling more explicit:
```suggestion
      const maxContestOptions =
        iter(contests).map(numBallotPositions).max() ?? 0;
```

apps/design/backend/src/test_decks.test.ts
```diff
@@ -0,0 +1,136 @@
...
+// Once ballot rendering is faster, it might be nice to also have a snapshot
+// test for test deck PDFs to ensure it all comes together correctly.
+describe('createPrecinctTestDeck', () => {
+  test('for a precinct with one ballot style', () => {
+    const electionDefinition = electionGeneralDefinition;
+    const { election } = electionDefinition;
+    const precinctId = election.precincts[0].id;
+    assert(
+      getBallotStylesByPrecinctId(electionDefinition, precinctId).length === 1
+    );
```
eventualbuddha, line 66: Any reason not to use `expect(getBallotStylesByPrecinctId(electionDefinition, precinctId)).toHaveLength(1)`? It'd have a better error message on failure. There are other places this question applies too, such as `expect(foo.length).toEqual(bar)` instead of `expect(foo).toHaveLength(bar)`. Perhaps for consistency since you're doing other `.length`-based checks? Or perhaps because Jest would blow up trying to pretty-print the value passed to `expect`?

eventualbuddha (approved):

apps/design/frontend/src/export_screen.tsx
jonahkagan: Yeah, I think once we move all of these exports to the background worker (which is the next planned change), that will become necessary. I'm imagining we might upload the files to s3 and then have a direct download link from there. TBD!

apps/design/backend/src/test_decks.test.ts
jonahkagan: I use `assert` when I'm trying to enforce an invariant on the test data. I use `assert` instead of `expect` to differentiate my intention - i.e. the test is invalid if this invariant breaks. But it seems like that wasn't so clear, so I'm open to suggestions on how to communicate that better.

I'll look into using `toHaveLength` more consistently for other cases.

eventualbuddha: I did have an inkling of that rationale when reading it, so I don't think it was completely lost. I'm not really opposed to the way it is now–maybe include a comment? I don't feel strongly about it.

libs/hmpb/layout/src/mark_ballot.ts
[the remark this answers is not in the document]
jonahkagan: That does make some sense to me. The tricky part is that document elements use positioning relative to their parent, so you'd need to compute the absolute positions first. In the next version of the ballot rendering pipeline, I'd like to build on this abstraction so that we can let the renderer flow items to different positions based on their computed dimensions (similar to a browser's HTML rendering). That would mean that it wouldn't be possible to use the multi-layer approach at this intermediate stage.

So then I think the question is: should marks be added later to the rendered ballot PDF as a separate layer based on the grid layouts? Or added into this intermediate representation like I have now? I originally was excited about adding them to the PDF, but got tripped up by two issues:
- We don't have a way to determine the position of the timing marks in the PDF (other than using the interpreter)
- We don't have a great way to know where to put write-ins (maybe we could leverage the UWI detection area?)

What do you think? I'm happy to call this "good enough for now" but also interested if you have more thoughts

apps/design/backend/src/test_decks.test.ts
jonahkagan: Great catch, fixed here: 2a6da512e

libs/hmpb/layout/src/mark_ballot.ts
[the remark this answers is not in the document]
eventualbuddha: Yeah, tricky. I'm inclined to call this good enough. We can reevaluate later.

PR #4462, opened 2024-01-03
Move election package generation to VxDesign's background worker

arsalansufi:
_Easiest reviewed by commit_

## Overview

This PR moves election package generation to VxDesign's background worker. Election package generation is likely to get much slower once we start translating and generating audio clips, so it'll no longer be reasonable to perform it synchronously. The patterns established in this PR can also be used to handle ballot and test deck generation asynchronously.

## Demo Video or Screenshot

The UX is still pretty bare bones. We can add progress bars and other optimizations in future PRs. We may also want to distinguish the act of generating the election package and downloading it. Note that I've artificially slowed down election package generation for this video 🐌

https://github.com/votingworks/vxsuite/assets/12616928/a30e6738-9033-48b2-80ac-a69dac53f372

_Error handling_

https://github.com/votingworks/vxsuite/assets/12616928/eaeba134-2cfb-4226-a59a-e681a8de2f51

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

## Checklist

- [x] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ N/A

### DIFF
--- a/apps/design/backend/src/app.test.ts
+++ b/apps/design/backend/src/app.test.ts
@@ -11,6 +11,7 @@ import {
   layOutAllBallotStyles,
   LayoutOptions,
 } from '@votingworks/hmpb-layout';
+import { suppressingConsoleOutput } from '@votingworks/test-utils';
 import {
   AdjudicationReason,
   AnyContest,
@@ -244,7 +245,9 @@ test('Election package management', async () => {
   );
 
   // Complete an export
-  await processNextBackgroundTaskIfAny({ log: jest.fn(), workspace });
+  await suppressingConsoleOutput(() =>
+    processNextBackgroundTaskIfAny(workspace)
+  );
   const electionPackageAfterExport = await apiClient.getElectionPackage({
     electionId,
   });
@@ -303,7 +306,9 @@ test('Election package export', async () => {
   const { election: appElection } = await apiClient.getElection({ electionId });
 
   await apiClient.exportElectionPackage({ electionId });
-  await processNextBackgroundTaskIfAny({ log: jest.fn(), workspace });
+  await suppressingConsoleOutput(() =>
+    processNextBackgroundTaskIfAny(workspace)
+  );
 
   const electionPackage = await apiClient.getElectionPackage({
     electionId,
--- a/apps/design/backend/src/app.test.ts
+++ b/apps/design/backend/src/app.test.ts
@@ -1,6 +1,4 @@
-import fs from 'fs';
 import JsZip from 'jszip';
-import path from 'path';
 import { assert, assertDefined } from '@votingworks/basics';
 import {
   electionFamousNames2021Fixtures,
@@ -30,7 +28,11 @@ import {
   safeParseSystemSettings,
 } from '@votingworks/types';
 import { getBallotStylesByPrecinctId } from '@votingworks/utils';
-import { ApiClient, testSetupHelpers } from '../test/helpers';
+import {
+  ApiClient,
+  exportElectionPackage,
+  testSetupHelpers,
+} from '../test/helpers';
 import { hasSplits, Precinct } from './store';
 import { processNextBackgroundTaskIfAny } from './worker/worker';
 
@@ -305,20 +307,11 @@ test('Election package export', async () => {
   });
   const { election: appElection } = await apiClient.getElection({ electionId });
 
-  await apiClient.exportElectionPackage({ electionId });
-  await suppressingConsoleOutput(() =>
-    processNextBackgroundTaskIfAny(workspace)
-  );
-
-  const electionPackage = await apiClient.getElectionPackage({
+  const electionPackageContents = await exportElectionPackage({
+    apiClient,
     electionId,
+    workspace,
   });
-  const electionPackageFileName = assertDefined(
-    electionPackage.url?.match(/election-package-[0-9a-z]{10}\.zip$/)?.[0]
-  );
-  const electionPackageContents = fs.readFileSync(
-    path.join(workspace.assetDirectoryPath, electionPackageFileName)
-  );
 
   // Check contents of generated zip file
 
@@ -476,6 +469,44 @@ test('Export test decks', async () => {
     layoutOptions: DEFAULT_LAYOUT_OPTIONS,
   });
 });
+
+test('Consistency of election hash across exports', async () => {
+  const baseElectionDefinition =
+    electionFamousNames2021Fixtures.electionDefinition;
+  const { apiClient, workspace } = setupApp();
+
+  const electionId = (
+    await apiClient.createElection({
+      electionData: baseElectionDefinition.electionData,
+    })
+  ).unsafeUnwrap();
+
+  const allBallotsOutput = await apiClient.exportAllBallots({
+    electionId,
+  });
+
+  const testDecksOutput = await apiClient.exportTestDecks({
+    electionId,
+  });
+
+  const electionPackageContents = await exportElectionPackage({
+    apiClient,
+    electionId,
+    workspace,
+  });
+  const zip = await JsZip.loadAsync(electionPackageContents);
+  const electionDefinitionFromElectionPackage = safeParseElectionDefinition(
+    await assertDefined(zip.file('election.json')).async('text')
+  ).unsafeUnwrap();
+
+  expect([
+    ...new Set([
+      allBallotsOutput.electionHash,
+      testDecksOutput.electionHash,
+      electionDefinitionFromElectionPackage.electionHash,
+    ]),
+  ]).toHaveLength(1);
+}, 30_000);
 
 describe('Ballot style generation', () => {
   const districts: District[] = [
--- a/apps/design/backend/src/app.test.ts
+++ b/apps/design/backend/src/app.test.ts
@@ -1,4 +1,7 @@
-import { assert } from '@votingworks/basics';
+import fs from 'fs';
+import JsZip from 'jszip';
+import path from 'path';
+import { assert, assertDefined } from '@votingworks/basics';
 import {
   electionFamousNames2021Fixtures,
   electionTwoPartyPrimaryDefinition,
@@ -21,14 +24,14 @@ import {
   Parties,
   Party,
   PartyId,
+  SystemSettings,
   safeParseElectionDefinition,
   safeParseSystemSettings,
-  SystemSettings,
 } from '@votingworks/types';
-import JsZip from 'jszip';
 import { getBallotStylesByPrecinctId } from '@votingworks/utils';
 import { ApiClient, testSetupHelpers } from '../test/helpers';
 import { hasSplits, Precinct } from './store';
+import { processNextBackgroundTaskIfAny } from './worker/worker';
 
 const { setupApp, cleanup } = testSetupHelpers();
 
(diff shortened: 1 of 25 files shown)


apps/design/backend/src/app.ts
```diff
@@ -291,9 +270,10 @@ function buildApi({ store }: { store: Store }) {
 }
 export type Api = ReturnType<typeof buildApi>;
 
-export function buildApp({ store }: { store: Store }): Application {
+export function buildApp({ workspace }: { workspace: Workspace }): Application {
   const app: Application = express();
-  const api = buildApi({ store });
+  const api = buildApi({ workspace });
   app.use('/api', grout.buildRouter(api, express));
+  app.use(express.static(workspace.assetDirectoryPath));
```
arsalansufi, line 277: This lets the frontend retrieve files from the backend

apps/design/backend/src/worker/tasks.ts
```diff
@@ -0,0 +1,68 @@
...
+    ballotType: BallotType.Precinct,
+    ballotMode: 'test',
+    layoutOptions,
+  }).unsafeUnwrap();
+  const zip = new JsZip();
+  zip.file(ElectionPackageFileName.ELECTION, electionDefinition.electionData);
+  zip.file(
+    ElectionPackageFileName.SYSTEM_SETTINGS,
+    JSON.stringify(systemSettings, null, 2)
+  );
```
arsalansufi, line 38: This logic above is unchanged, just moved from `app.ts`.

The logic below is what's different. Whereas the backend used to send the entire zip contents to the frontend in its response, we now store the zip in the backend's file system and return a URL to the frontend that the frontend can use to download it.

When we eventually deploy VxDesign, we'll likely use S3 and S3 URLs.

apps/design/backend/src/worker/worker.ts
```diff
@@ -1,45 +1,46 @@
...
 
-async function processBackgroundTask(
-  _store: Store, // eslint-disable-line @typescript-eslint/no-unused-vars
-  _task: BackgroundTask // eslint-disable-line @typescript-eslint/no-unused-vars
-): Promise<void> {
-  // Simulate a long-running operation
-  // TODO(arsalan): Implement actual processing logic
-  await sleep(3000);
-}
+export async function processNextBackgroundTaskIfAny({
```
arsalansufi, line 6: Easier to review this commit using https://github.com/votingworks/vxsuite/pull/4462/commits/50816f451894720fb256c5c539e1d51e8517d11c?w=1

Doing some rearranging to facilitate testing. Rather than having app tests additionally start up a background worker, app tests can just call `processNextBackgroundTaskIfAny` as needed. You'll see this in the next commit

apps/design/backend/src/worker/worker.ts
```diff
@@ -1,45 +1,46 @@
...
+  log = console.log, // eslint-disable-line no-console
```
arsalansufi, line 7: Tests can override with this with a `jest.fn` to avoid cluttering test output

apps/design/backend/src/app.test.ts
```diff
@@ -201,7 +204,82 @@ test('Update layout options', async () => {
...
+  // task
+  await apiClient.exportElectionPackage({ electionId });
+  const electionPackageAfterInitiatingRedundantExport =
+    await apiClient.getElectionPackage({ electionId });
+  expect(electionPackageAfterInitiatingRedundantExport).toEqual(
+    electionPackageAfterInitiatingExport
+  );
+
+  // Complete an export
+  await processNextBackgroundTaskIfAny({ log: jest.fn(), workspace });
```
arsalansufi, line 247: As mentioned in the last commit, calling this allows us to incorporate background task processing into app tests without having to fully start up a background worker

apps/design/frontend/src/export_screen.tsx
```diff
@@ -1,22 +1,58 @@
...
+        previousElectionPackage?.task &&
+          !previousElectionPackage.task.completedAt &&
+          currentElectionPackage.task?.completedAt
+      );
+      if (taskJustCompleted) {
+        const { error } = assertDefined(currentElectionPackage.task);
+        if (error) {
+          setExportError(error);
+        } else {
+          downloadFile(assertDefined(currentElectionPackage.url));
```
arsalansufi, line 48: As noted in the PR description, we may want to distinguish the act of generating the election package and downloading it, down the road.

Right now, after the background task finishes, we auto-download the file. If you miss the background task's transition from being in progress to being finished, e.g., because you close your browser and reopen it, you'll just have to click the export button again.

The backend could be smart and take advantage of work from past runs so that election package generation doesn't do everything from scratch every time, unless it needs to because the election definition has changed. This would give us all the benefits of separate generate and download buttons, without the potential confusion.

kofi-q (approved):
LGTM!

apps/design/backend/src/worker/worker.ts
eventualbuddha: Alternatively just use `console.log` and [`suppressingConsoleOutput`](https://github.com/votingworks/vxsuite/blob/5fc65180bf45e1a4617095f3c7925531c82764e0/libs/test-utils/src/console.ts#L97-L101). No strong preference.

apps/design/backend/src/worker/worker.ts
```diff
@@ -1,45 +1,46 @@
...
-  while (true) {
-    const nextTask = store.getOldestQueuedBackgroundTask();
+  const nextTask = store.getOldestQueuedBackgroundTask();
 
-    if (!nextTask) {
-      await sleep(1000);
-      continue;
-    }
+  if (!nextTask) {
+    await sleep(1000);
```
eventualbuddha, line 18: This feels like something that should happen outside this function, i.e. in the main loop. Return whether a job was processed and sleep if not?

apps/design/backend/src/worker/worker.ts
```diff
@@ -1,45 +1,46 @@
...
+  store.completeBackgroundTask(nextTask.id);
+  log(`✅ Finished processing background task ${nextTask.id}`);
 }
 
-export function start({ store }: { store: Store }): void {
-  setTimeout(() => processBackgroundTasks(store));
+export function start({ workspace }: { workspace: Workspace }): void {
+  setTimeout(async () => {
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
eventualbuddha, line 42: Avoids the "constant condition" lint:
```suggestion
    for (;;) {
```

apps/design/backend/src/worker/worker.ts
```diff
@@ -1,45 +1,46 @@
...
+    return;
   }
+  store.completeBackgroundTask(nextTask.id);
+  log(`✅ Finished processing background task ${nextTask.id}`);
 }
 
-export function start({ store }: { store: Store }): void {
-  setTimeout(() => processBackgroundTasks(store));
+export function start({ workspace }: { workspace: Workspace }): void {
+  setTimeout(async () => {
```
eventualbuddha, line 40: Use `process.nextTick`? Functionally should be the same, just a little clearer to me that no delay is intended.

eventualbuddha (approved):
Nice! Have you considered how these files will be cleaned up over time, for example if an election is deleted? Maybe it's a non-issue since longer term we want to move things to S3.

apps/design/backend/src/app.test.ts
```diff
@@ -341,11 +429,6 @@ test('Export all ballots', async () => {
       layoutOptions,
     });
   }
-
-  const electionPackageResult = await apiClient.exportElectionPackage({
-    electionId,
-  });
-  expect(electionHash).toEqual(electionPackageResult.electionHash);
```
jonahkagan, line 348: This checks a fairly important invariant - the exported ballots should match the exported election package. I think we should still test it in some way (and same with test decks below). Maybe a helper function to run through the whole election package export process would be useful here?

apps/design/frontend/src/export_screen.tsx
```diff
@@ -1,22 +1,58 @@
...
+export function ExportScreen(): JSX.Element | null {
   const { electionId } = useParams<ElectionIdParams>();
+
+  const electionPackageQuery = getElectionPackage.useQuery(electionId);
   const exportAllBallotsMutation = exportAllBallots.useMutation();
-  const exportTestDecksMutation = exportTestDecks.useMutation();
   const exportElectionPackageMutation = exportElectionPackage.useMutation();
+  const exportTestDecksMutation = exportTestDecks.useMutation();
+
+  const [exportError, setExportError] = useState<string | undefined>(undefined);
```
jonahkagan, line 34: Do we need state for this? I think we can just use `electionPackageQuery.data.task?.error`

apps/design/frontend/src/export_screen.test.tsx
```diff
@@ -36,59 +45,127 @@ function renderScreen() {
...
   renderScreen();
   await screen.findByRole('heading', { name: 'Export' });
 
-  apiMock.exportElectionPackage.expectCallWith({ electionId }).resolves({
+  apiMock.exportAllBallots.expectCallWith({ electionId }).resolves({
     zipContents: Buffer.from('fake-zip-contents'),
     electionHash: '1234567890abcdef',
   });
 
-  screen.getButton('Export Election Package').click();
```
jonahkagan, line 48: Wait, how did this `.click()` even work before? I didn't even know that was possible. And I supposedly wrote these tests?? 🥴

apps/design/backend/src/worker/tasks.ts
```diff
@@ -0,0 +1,68 @@
...
+  workspace: Workspace,
+  { taskName, payload }: BackgroundTask
+): Promise<void> {
+  switch (taskName) {
+    case 'generate_election_package': {
+      const parsedPayload = safeParseJson(
+        payload,
+        z.object({ electionId: z.string() })
+      ).unsafeUnwrap();
+      await generateElectionPackage(workspace, parsedPayload);
```
jonahkagan, line 60: I wonder if we want some kind of transaction management for background tasks? In Arlo, the entire task is one transaction, so if any error occurs it can be rolled back. It doesn't feel like a blocking concern here since there's only one write and it's the last thing in the task, but thought I'd note it for the future.

jonahkagan (approved):

apps/design/frontend/src/export_screen.tsx
arsalansufi: Yeah good question! I landed on a separate state value because:

1. I didn't want the error message to be displayed on first load of the page if the last task, however long ago it was, failed
2. I wanted clicking to export again to immediately dismiss the error, without having to wait for relevant React Query state to update
3. I wanted clicking any export button, not just the export election package button, to dismiss the error

eventualbuddha, line 34: Some code golf:
```suggestion
  const [exportError, setExportError] = useState<string>();
```

jonahkagan: Makes sense!

apps/design/backend/src/worker/worker.ts
arsalansufi: Ah good reminder! Decided to use `suppressingConsoleOutput` so as not to introduce a second pattern for that: https://github.com/votingworks/vxsuite/pull/4462/commits/a366f85e00c50ae53442dce8f39237f5acbb89b8

apps/design/backend/src/worker/worker.ts
arsalansufi: Agreed: https://github.com/votingworks/vxsuite/pull/4462/commits/9b88b0bee4f87ddee6a7cf6efe4a9b722b3b9263

apps/design/backend/src/worker/worker.ts
arsalansufi: 👍 https://github.com/votingworks/vxsuite/pull/4462/commits/9b88b0bee4f87ddee6a7cf6efe4a9b722b3b9263

apps/design/backend/src/worker/tasks.ts
```diff
@@ -0,0 +1,68 @@
...
+  zip.file(ElectionPackageFileName.ELECTION, electionDefinition.electionData);
+  zip.file(
+    ElectionPackageFileName.SYSTEM_SETTINGS,
+    JSON.stringify(systemSettings, null, 2)
+  );
+
+  const displayElectionHash = getDisplayElectionHash(electionDefinition);
+  const fileName = `election-package-${displayElectionHash}.zip`;
+  const filePath = path.join(assetDirectoryPath, fileName);
+  await fs.writeFile(filePath, await zip.generateAsync({ type: 'nodebuffer' }));
```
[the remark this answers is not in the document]
arsalansufi: Good call! https://github.com/votingworks/vxsuite/pull/4462/commits/502a424feb46465ce4cfba169718e6e58eba3482

It does look like the `streamFiles: true` option comes with some risks, but since we're importing election packages in such controlled contexts, we should be okay.

<table>note</table>

[Reference](https://stuk.github.io/jszip/documentation/api_jszip/generate_async.html#streamfiles-option)

apps/design/backend/src/worker/worker.ts
arsalansufi: I'll stick with `while (true)` as per Jonah's line of reasoning

apps/design/frontend/src/export_screen.tsx
arsalansufi: :100: https://github.com/votingworks/vxsuite/pull/4462/commits/3e1d1fad50f4a9f8ea9ba7b3079f241dc85c2620

apps/design/frontend/src/export_screen.test.tsx
arsalansufi: Yeah definitely surprised me! I think it's just the native [HTMLElement click method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click). Only reason I noticed this was because the tests started displaying "need `act` wrapping" warnings as I added state updates to the code

apps/design/backend/test/helpers.ts
```diff
@@ -18,14 +19,16 @@ export function testSetupHelpers() {
   const servers: Server[] = [];
 
   function setupApp() {
+    const assetDirectoryPath = tmp.dirSync().name;
     const store = Store.fileStore(tmp.fileSync().name);
-    const app = buildApp({ store });
+    const workspace: Workspace = { assetDirectoryPath, store };
```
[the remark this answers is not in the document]
arsalansufi: Good call! https://github.com/votingworks/vxsuite/pull/4462/commits/4ba1b5e73981beb224092c6f58574f68f4b5ec37

apps/design/backend/src/worker/tasks.ts
arsalansufi: Good nudge! I'm going back and forth right now on whether to build an automatic mechanism vs. giving task handlers flexibility.

Looking ahead to the addition of language and audio files to election packages, I see some value in not automatically wrapping everything in a transaction. If, for example, one speech synthesis request fails because of a Google Cloud blip, we wouldn't want to roll back speech synthesis results already added to the cache in the DB.

I'll think some more about this as I add language and audio files!

apps/design/backend/src/app.test.ts
arsalansufi: Ah nice, thanks for pointing that out. Added a test to cover the invariant, as well as the suggested helper: https://github.com/votingworks/vxsuite/pull/4462/commits/f9722c0460e0ed4c06bcd3a5ac85cd365ce4fd11

arsalansufi:
> Have you considered how these files will be cleaned up over time, for example if an election is deleted? Maybe it's a non-issue since longer term we want to move things to S3.

@eventualbuddha Thanks for checking on this! My thinking is exactly what you mention re S3. Once we're using S3, no need to worry about deletion. And until then, we can just manually clean our `dev-workspace` directories.

Though probably not a bad idea to add some kind of safeguard for cleanup in dev in the meantime - can look into that in a subsequent PR. It's also possible that we'll still use the file system for local dev even after we've deployed VxDesign and started using S3, making this cleanup more worthwhile.

PR #4463, opened 2024-01-03
[i18n/Audio] Expose audio controls via the UiStringsContextProvider

kofi-q:
## Overview

Fleshing out the `UiStringsAudioContext` (renamed from `AudioContext` to avoid the name collision with the native web audio API) to manage global audio settings and expose controls for modifying them.

Adding a `useAudioControls` hook to externally expose the parts of the `UiStringsAudioContext` that need to be available outside of `libs/ui`. Future PRs will add the internal audio playback utilities that make use of the remaining context values.

## Testing Plan
- Unit tests for the context API and external hook
- Some manual testing with a local prototype 

## Checklist

### DIFF
--- a/libs/ui/src/hooks/use_audio_controls.ts
+++ b/libs/ui/src/hooks/use_audio_controls.ts
@@ -1,0 +1,33 @@
+import { useAudioContext } from '../ui_strings/audio_context';
+
+export interface AudioControls {
+  decreasePlaybackRate: () => void;
+  decreaseVolume: () => void;
+  increasePlaybackRate: () => void;
+  increaseVolume: () => void;
+  reset: () => void;
+  setIsEnabled: (enabled: boolean) => void;
+  togglePause: () => void;
+}
+
+function noOp() {}
+
+/**
+ * Provides an API for modifying UiString screen reader audio settings.
+ *
+ * Returns a stubbed-out no-op API for convenience/ease-of-testing, when used
+ * without a parent `UiStringsAudioContext`.
+ */
+export function useAudioControls(): AudioControls {
+  const audioContext = useAudioContext();
+
+  return {
+    decreasePlaybackRate: audioContext?.decreasePlaybackRate || noOp,
+    decreaseVolume: audioContext?.decreaseVolume || noOp,
+    increasePlaybackRate: audioContext?.increasePlaybackRate || noOp,
+    increaseVolume: audioContext?.increaseVolume || noOp,
+    reset: audioContext?.reset || noOp,
+    setIsEnabled: audioContext?.setIsEnabled || noOp,
+    togglePause: audioContext?.togglePause || noOp,
+  };
+}
--- a/libs/ui/src/hooks/use_audio_controls.ts
+++ b/libs/ui/src/hooks/use_audio_controls.ts
@@ -5,12 +5,26 @@ export interface AudioControls {
   decreaseVolume: () => void;
   increasePlaybackRate: () => void;
   increaseVolume: () => void;
+  replay: () => void;
   reset: () => void;
   setIsEnabled: (enabled: boolean) => void;
   togglePause: () => void;
 }
 
 function noOp() {}
+
+/**
+ * Replays the current or last-played audio by re-triggering a focus event on
+ * the currently active element, if any.
+ */
+function replay() {
+  const { activeElement } = window.document;
+
+  if (activeElement instanceof HTMLElement) {
+    activeElement.blur();
+    activeElement.focus();
+  }
+}
 
 /**
  * Provides an API for modifying UiString screen reader audio settings.
@@ -27,6 +41,7 @@ export function useAudioControls(): AudioControls {
     increasePlaybackRate: audioContext?.increasePlaybackRate || noOp,
     increaseVolume: audioContext?.increaseVolume || noOp,
     reset: audioContext?.reset || noOp,
+    replay,
     setIsEnabled: audioContext?.setIsEnabled || noOp,
     togglePause: audioContext?.togglePause || noOp,
   };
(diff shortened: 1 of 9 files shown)


libs/ui/src/ui_strings/audio_volume.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+/**
+ * Minimum allowed gain to achieve a minimum SPL of 20 dB, as prescribed in
+ * VVSG 2.0, section 7.1-K,
+ *
+ * Assumes a 0 dB gain represents the 65 dB SPL midpoint of the default 60-70
+ * dB range prescribed in the VVSG spec.
+ */
+export const MIN_GAIN_DB = -45;
```
arsalansufi, line 23: Your two worlds collide! 🎼 + 💻

libs/ui/src/ui_strings/audio_volume.ts
```diff
@@ -0,0 +1,44 @@
...
+export const DEFAULT_GAIN_DB = 0;
+
+/**
+ * Maximum allowed gain to achieve a maximum SPL of 100 dB, as prescribed in
+ * VVSG 2.0, section 7.1-K,
+ *
+ * Assumes a 0 dB gain represents the 65 dB SPL midpoint of the default 60-70
+ * dB range prescribed in the VVSG spec.
+ */
+export const MAX_GAIN_DB = 35;
```
arsalansufi, line 38: Very minor but could encode the math behind these numbers into this file, e.g.

```ts
const DEFAULT_SPL_DB = 65;
const MAX_SPL_DB = 100;
const MIN_SPL_DB = 20;

export const MAX_GAIN_DB = MAX_SPL_DB - DEFAULT_SPL_DB;
export const MIN_GAIN_DB = MIN_SPL_DB - DEFAULT_SPL_DB;
```

libs/ui/src/ui_strings/audio_context.tsx
```diff
@@ -3,28 +3,133 @@ import React from 'react';
...
+    setGainDb(DEFAULT_GAIN_DB);
+    setPlaybackRate(DEFAULT_PLAYBACK_RATE);
+    setIsPaused(false);
+  }, []);
+
+  React.useEffect(() => {
+    if (isEnabled) {
+      reset();
+    } else {
+      setIsPaused(true);
```
arsalansufi, line 71: Just checking my understanding: Are we pausing here because disabling the audio player necessitates pausing any in-flight audio?

arsalansufi (approved):
LGTM!

libs/ui/src/hooks/use_audio_controls.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+/**
+ * Provides an API for modifying UiString screen reader audio settings.
+ *
+ * Returns a stubbed-out no-op API for convenience/ease-of-testing, when used
+ * without a parent `UiStringsAudioContext`.
+ */
+export function useAudioControls(): AudioControls {
```
kofi-q, line 35: cc @kshen0 - heads-up this will be the react hook exposing the audio controls. Not wired up to the the actual audio playback yet, but this should be stable API to develop against. Also open to modifications in case a different format would work better for the integration with the VSAP controller

libs/ui/src/ui_strings/audio_context.tsx
```diff
@@ -3,28 +3,136 @@ import React from 'react';
...
+  const { api, children } = props;
+  const [isEnabled, setIsEnabled] = React.useState(false);
+  const [isPaused, setIsPaused] = React.useState(true);
+  const [playbackRate, setPlaybackRate] = React.useState<number>(
+    DEFAULT_PLAYBACK_RATE
+  );
+  const [gainDb, setGainDb] = React.useState<number>(DEFAULT_GAIN_DB);
+
+  const webAudioContextRef = React.useRef(
+    window.AudioContext ? new AudioContext() : undefined
```
eventualbuddha, line 58: I'm guessing this isn't an expensive operation, but I thought I'd ask just in case: is creating a new `AudioContext` expensive? Looking at the [Chromium source code for `AudioContext`](https://chromium.googlesource.com/chromium/blink/+/refs/heads/main/Source/modules/webaudio/AudioContext.cpp#28) it appears that it might spawn a new thread at the very least. Can we use a `useMemo` for this maybe?

libs/ui/src/hooks/use_audio_controls.ts
kshen0: Thank you! If I run into any issues with the API this week I'll let you know ASAP

libs/ui/src/ui_strings/audio_context.tsx
kofi-q: Yeah, while I was prototyping, it registered as a separate JS VM instance using up somewhere between 1-1.5mb (whenever there's a fully decoded clip playing), so not too bad.

The intention's to only ever have one instance per app load, though. Previously had this as a lazy-instantiated single global instance and moved it to a `ref` to make testing easier, but your comment makes me question my understanding of how often `ref`s are re-instantiated. Will check and maybe move back to the singleton

eventualbuddha: Agreed, that's not terrible. And it looks like the source hardcodes a max of 6 and throws an exception if it goes over, so we'd become aware of it if we were creating too many that weren't being cleaned up.

As for `ref`, that doesn't really have anything to do with it afaict. Whenever the component is rendered it'll create a new `AudioContext` which is passed to `useRef` and promptly dropped. So we'd pay the startup+teardown costs with every render. Normally I'd say it's not worth thinking about e.g. creating some JS data structure repeatedly only to have it be gc'd, but this might do significantly more so I'm glad you opened #4478 to fix this 🙏

PR #4465, opened 2024-01-03
Save voter session on ballot spoil

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3909

Improves the flow for spoiling a ballot in post-print ballot validation
* Rather than restarting the voter session, the session is saved. The voter is returned to the review screen after the poll worker removes the old ballot, loads a new sheet, and removes their card
* The ballot spoil flow now checks that the ballot is removed by the poll worker before allowing the poll worker to continue

## Demo Video or Screenshot

Happy path

https://github.com/votingworks/vxsuite/assets/1060688/75a28cb1-e35f-44ee-8213-bccd34d25a42

Lightly stress testing pulling paper early and repeated auth changes

https://github.com/votingworks/vxsuite/assets/1060688/a89d9fa6-8547-4c7f-9662-7bf8205a3abc

## Testing Plan
- updated component and e2e tests

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (no new actions or errors introduced)

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -90,7 +90,7 @@ type PaperHandlerStatusEvent =
   | { type: 'VOTER_INITIATED_PRINT'; pdfData: Buffer }
   | { type: 'PAPER_IN_OUTPUT' }
   | { type: 'PAPER_IN_INPUT' }
-  | { type: 'VOTER_CONFIRMED_INVALIDATED_BALLOT' }
+  | { type: 'POLLWORKER_CONFIRMED_INVALIDATED_BALLOT' }
   | { type: 'SCANNING' }
   | { type: 'SET_INTERPRETATION_FIXTURE' }
   | { type: 'VOTER_VALIDATED_BALLOT' }
@@ -585,13 +585,25 @@ export function buildMachine(
               NO_PAPER_ANYWHERE: 'resetting_state_machine_after_success',
             },
           },
-          // Ballot invalidation is a 2-stage process so the frontend can prompt the voter to get a pollworker
           waiting_for_invalidated_ballot_confirmation: {
-            on: {
-              VOTER_CONFIRMED_INVALIDATED_BALLOT: 'eject_to_front',
-              // Even if ballot is removed from front, we still want the frontend to require pollworker auth before continuing
-              NO_PAPER_ANYWHERE: undefined,
+            initial: 'paper_present',
+            states: {
+              paper_present: {
+                invoke: pollPaperStatus(),
+                on: {
+                  NO_PAPER_ANYWHERE: 'paper_nowhere',
+                },
+              },
+              paper_nowhere: {
+                on: {
+                  POLLWORKER_CONFIRMED_INVALIDATED_BALLOT: 'done',
+                },
+              },
+              done: {
+                type: 'final',
+              },
             },
+            onDone: 'accepting_paper',
           },
           // Eject-to-rear jam handling is a little clunky. It
           // 1. Tries to transition to success if no paper is detected
@@ -818,9 +830,13 @@ export async function getPaperHandlerStateMachine({
         case state.matches('voting_flow.interpreting'):
           return 'interpreting';
         case state.matches(
-          'voting_flow.waiting_for_invalidated_ballot_confirmation'
+          'voting_flow.waiting_for_invalidated_ballot_confirmation.paper_present'
         ):
-          return 'waiting_for_invalidated_ballot_confirmation';
+          return 'waiting_for_invalidated_ballot_confirmation.paper_present';
+        case state.matches(
+          'voting_flow.waiting_for_invalidated_ballot_confirmation.paper_nowhere'
+        ):
+          return 'waiting_for_invalidated_ballot_confirmation.paper_nowhere';
         case state.matches('voting_flow.presenting_ballot'):
           return 'presenting_ballot';
         case state.matches('voting_flow.eject_to_front'):
@@ -902,7 +918,7 @@ export async function getPaperHandlerStateMachine({
 
     confirmInvalidateBallot(): void {
       machineService.send({
-        type: 'VOTER_CONFIRMED_INVALIDATED_BALLOT',
+        type: 'POLLWORKER_CONFIRMED_INVALIDATED_BALLOT',
       });
     },
   };
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -591,10 +591,10 @@ export function buildMachine(
               paper_present: {
                 invoke: pollPaperStatus(),
                 on: {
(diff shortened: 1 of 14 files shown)


apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -90,7 +90,7 @@ type PaperHandlerStatusEvent =
   | { type: 'VOTER_INITIATED_PRINT'; pdfData: Buffer }
   | { type: 'PAPER_IN_OUTPUT' }
   | { type: 'PAPER_IN_INPUT' }
-  | { type: 'VOTER_CONFIRMED_INVALIDATED_BALLOT' }
+  | { type: 'POLLWORKER_CONFIRMED_INVALIDATED_BALLOT' }
```
kshen0, line 93: This was a misnomer; the voter is not actually the one who confirms the ballot is has been invalidated

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -585,13 +585,25 @@ export function buildMachine(
...
+                invoke: pollPaperStatus(),
+                on: {
+                  NO_PAPER_ANYWHERE: 'paper_nowhere',
+                },
+              },
+              paper_nowhere: {
+                on: {
+                  POLLWORKER_CONFIRMED_INVALIDATED_BALLOT: 'done',
+                },
+              },
```
kshen0, line 601: The copy changes based on whether paper is present

apps/mark-scan/frontend/src/app_cardless_voting.test.tsx
```diff
@@ -33,7 +33,7 @@ beforeEach(() => {
   apiMock = createApiMock();
   kiosk = fakeKiosk();
   window.kiosk = kiosk;
-  apiMock.setPaperHandlerState('waiting_for_ballot_data');
```
kshen0, line 36: These e2e tests misrepresented paper handler state, but it wasn't an issue before now because of the order we rendered `LoadPaperPage` in (ie. it used to be lower priority)

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -457,10 +457,24 @@ export function AppRoot({
...
+        stateMachineState ===
+          'waiting_for_invalidated_ballot_confirmation.paper_nowhere')
     ) {
+      return (
+        <BallotInvalidatedPage
+          authStatus={authStatus}
+          paperPresent={
+            stateMachineState ===
+            'waiting_for_invalidated_ballot_confirmation.paper_present'
+          }
```
kshen0, line 472: Seemed clearer to pass this prop down rather than query for state in the lower components, but open to feedback

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -508,17 +517,10 @@ export function AppRoot({
...
         if (stateMachineState === 'presenting_ballot') {
           ballotContextProviderChild = <ValidateBallotPage />;
         }
-        if (
-          stateMachineState === 'waiting_for_invalidated_ballot_confirmation'
-        ) {
-          ballotContextProviderChild = (
-            <BallotInvalidatedPage authStatus={authStatus} />
-          );
-        }
```
kshen0, line 521: Since we are no longer restarting the voting session, we don't need access to `BallotContext`

apps/mark-scan/frontend/src/pages/poll_worker_screen.tsx
```diff
@@ -379,11 +379,7 @@ export function PollWorkerScreen({
                           key={precinct.id}
                           aria-label={`Activate Voter Session for Precinct ${precinct.name}`}
                           onPress={() => {
-                            setAcceptingPaperStateMutation.mutate(undefined, {
-                              onSuccess() {
-                                setSelectedCardlessVoterPrecinctId(precinct.id);
-                              },
-                            });
```
kshen0, line 386: This was redundant with another call to `setAcceptingPaperStateMutation.mutate` that happens after ballot style selection

apps/mark-scan/frontend/src/pages/remove_invalidated_ballot_page.tsx
```diff
@@ -1,35 +1,44 @@
...
 
-  async function onPressContinue() {
-    // Reset session and ballot before changing the state machine state. If done
-    // in the reverse order the screen will flicker
-    await endVoterSession();
-    resetBallot();
+  const history = useHistory();
+  useEffect(() => {
+    history.push('/ready-to-review');
+  });
```
kshen0, line 19: As discussed before, not ideal to use react-router routing here. Should be simplified in https://github.com/votingworks/vxsuite/issues/3985

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
jonahkagan: paper_nowhere is a little confusing. Maybe no_paper or paper_absent?

apps/mark-scan/frontend/src/app_root.tsx
jonahkagan: seems fine to me!

apps/mark-scan/frontend/src/pages/ballot_invalidated_page.tsx
```diff
@@ -8,13 +8,19 @@ interface Props {
...
+export function BallotInvalidatedPage({
+  authStatus,
+  paperPresent,
+}: Props): JSX.Element {
   return (
     <AskPollWorkerPage
       authStatus={authStatus}
-      pollWorkerPage={<RemoveInvalidatedBallotPage />}
+      pollWorkerPage={
+        <RemoveInvalidatedBallotPage paperPresent={paperPresent} />
```
jonahkagan, line 22: I noticed in the video that the "Ask a Poll Worker for Help" text wraps awkwardly to two lines. I think we should look into moving the left-aligned text. Also, the "Voting Session in Progress" screen could use some more horizontal padding - it's butting right up against the sides. But these are minor issues that can be addressed in a future UI polish pass.

jonahkagan (approved):
Thanks for the clear PR!

apps/mark-scan/frontend/src/pages/ballot_invalidated_page.tsx
kshen0: The "Ask a Poll Worker for Help" screen uses the `CenteredPageLayout` component, which is used by a lot of the app. Agree we should do a design pass on that component and make sure we update all callers at once for consistency.

"Voting Session in Progress" padding was a one-off fix though:

Before

Screenshot 2024-01-09 at 11 48 21 AM


After

Screenshot 2024-01-09 at 11 48 11 AM

It has the same wrapping awkwardness with padding, but at least it's consistent so IMO it's the lesser of two evils. We can fix this and other screens in https://github.com/orgs/votingworks/projects/23/views/75?pane=issue&itemId=49471078

PR #4482, opened 2024-01-04
Change election title on ballots for NH school districts

### LINKED ISSUE #4369: VxDesign: visual differentiation b/w local and school ballots
We currently have the way to create separate styles for printing based on precinct splits, but they will both say the same election name. Locals in NH will want one style to say "School Board Election" and the other to say "Annual Town Election".

To support this, add a form field to precinct splits to specify the ballot election title (just stored in VxDesign - would not impact the election hash).

jonahkagan:
## Overview

Closes #4369 

NH runs simultaneous town and school district elections. The ballots for these two elections need to be scanned on the same scanner, meaning they must use the same election definition. However, they need to show a different election title on the ballot. For now, we add a special case in the ballot layout code to support this.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/34818b5e-81e5-4c97-aa86-bd9185387314

## Testing Plan
- Added unit test
- Manual test in VxDesign

## Checklist

### DIFF
--- a/libs/hmpb/layout/src/layout.ts
+++ b/libs/hmpb/layout/src/layout.ts
@@ -1,6 +1,7 @@
 import {
   assert,
   assertDefined,
+  find,
   iter,
   lines,
   ok,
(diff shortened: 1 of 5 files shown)


libs/hmpb/layout/src/layout.ts
```diff
@@ -360,6 +361,32 @@ export function TimingMarkGrid({ m }: { m: Measurements }): AnyElement {
...
 }
 
+// Special case hack for NH school election ballots (which need to be visually
+// differentiated from town election ballots but also need to be scanned on the
+// same scanner, and therefore must use the same election definition). if a
+// ballot style only has a single district that appears to correspond to a
+// school district, then change the election title. In the future, we might
+// support customization like this by allowing custom content for different
+// precinct splits or by using different ballot templates.
+function modifyElectionTitleForNhSchoolElections(
```
eventualbuddha, line 371: This name sounds like it's going to mutate its arguments, but it actually just extracts the election title with a special case for NH school elections. Maybe it should be called something like `getElectionTitleOrSchoolElectionTitle`?

eventualbuddha (approved):

libs/hmpb/layout/src/layout.ts
jonahkagan: Updated: 1cc25ab16

PR #4483, opened 2024-01-04
[i18n/audio] Add <ReadOnLoad> for reading intro text on page/route load

kofi-q:
## Overview

Replacing the existing pattern of assigning `id='audiofocus'` to introductory sections of pages with a dedicated `<ReadOnLoad>` component.

The upcoming pre-generated audio "screen reader" will get triggered by focus/click events in the same way the kiosk-browser screen reader does. This component mostly mirrors the functionality in the old BMD `FocusManager` by simulating a user click on its container element, but there's room later improvement if we want to have a way of queuing up multiple sections of a page with multiple `<ReadOnLoad>` instances.

Will be following up later with PRs to update the other `id='audiofocus'` instances.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/264902/64fc4f24-b0c0-42f4-924a-f853c5d48370

## Testing Plan
- Unit tests + manual verification with prototype audio player.

## Checklist

### DIFF
--- a/libs/ui/src/ui_strings/read_on_load.test.tsx
+++ b/libs/ui/src/ui_strings/read_on_load.test.tsx
@@ -1,0 +1,140 @@
+import { Optional, assert } from '@votingworks/basics';
+import { act } from 'react-dom/test-utils';
+import React from 'react';
+import { createMemoryHistory } from 'history';
+import { Route, Router } from 'react-router-dom';
+import { ReadOnLoad } from './read_on_load';
+import { render, screen } from '../../test/react_testing_library';
+import {
+  UiStringsAudioContextInterface,
+  UiStringsAudioContextProvider,
+  useAudioContext,
+} from './audio_context';
+import {
+  UiStringsReactQueryApi,
+  createUiStringsApi,
+} from '../hooks/ui_strings_api';
+
+const mockUiStringsApi: UiStringsReactQueryApi = createUiStringsApi(() => ({
+  getAudioClips: jest.fn(),
+  getAvailableLanguages: jest.fn(),
+  getUiStringAudioIds: jest.fn(),
+  getUiStrings: jest.fn(),
+}));
+
+function renderWithClickListener(ui: React.ReactNode) {
+  const mockOnClick = jest.fn();
+  const result = render(<div onClickCapture={mockOnClick}>{ui}</div>);
+
+  return {
+    mockOnClick,
+    result,
+  };
+}
+
+let audioContext: Optional<UiStringsAudioContextInterface>;
+function TestContextConsumer() {
+  audioContext = useAudioContext();
+
+  return null;
+}
+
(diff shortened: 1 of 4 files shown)


libs/ui/src/ui_strings/read_on_load.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * On initial render, this triggers an audio read-out of any descendant
+ * `UiString` elements rendered within (in order of appearance in the DOM), if
+ * audio playback is enabled.
+ *
+ * Re-triggers a read-out on subsequent route changes, if applicable, or whenever
+ * audio playback is newly enabled.
+ *
+ * NOTE: Intended for use as a single instance on any given app screen. If
+ * multiple instances are present on screen, audio will be played only for
+ * content within the last `ReadOnLoad` instance.
```
eventualbuddha, line 35: Would you classify multiple `ReadOnLoad` in a single render tree as a bug? If so, would it make sense to log/warn/throw?

eventualbuddha (commented):
Does this interact with modals appearing? Should it?

libs/ui/src/ui_strings/read_on_load.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+  return null;
+}
+
+afterEach(() => {
+  audioContext = undefined;
+});
+
+test('is no-op when audio context is absent', () => {
+  const { mockOnClick } = renderWithClickListener(
+    <ReadOnLoad>Bonjour!</ReadOnLoad>
```
arsalansufi, line 48: Completely unserious comment that doesn't actually require a code change, but I think that there's a space before exclamation marks in French 😝

arsalansufi (approved):
Other than Brian's comments/questions, LGTM!

kofi-q (commented):
> Does this interact with modals appearing? Should it?

No direct interaction -- when a modal with  a `ReadOnLoad` pops up, the modal content will get read out and when the modal is closed, focus returns to the last-focused element on the page (likely a button-like control) and if that element has readable content, it will then get read out.

There's a brief example in the [demo video here](https://drive.google.com/file/d/1YPvjNmuYKPxp1nC5k6KysJAQdZevT9_T/view?usp=sharing) showing that behaviour.

libs/ui/src/ui_strings/read_on_load.tsx
eventualbuddha: I don't think it's worth blocking. Go for it.

kofi-q: SG! Will follow up once everything's wired up. Been thinking about the flip side of this, WRT people forgetting to add a `ReadOnLoad` instance to a voter-facing screen and how we might surface those as warnings -- can try to address both issues in the same task

PR #4484, opened 2024-01-05
Remove unused SVG Assets

adghayes:
I checked whether these images were being referenced anywhere in the code before removing.


arsalansufi (approved):
LGTM!

PR #4488, opened 2024-01-08
[i18n/audio] Use <ReadOnLoad> component in BMD screens and modals

kofi-q:
## Overview

Using the `<ReadOnLoad>` component added in https://github.com/votingworks/vxsuite/pull/4483 on voter-facing screens and modals in the BMD apps. This replaces the previous `id="audiofocus"`/`id="modalaudiofocus"` tags and removes them altogether from non-voter-facing screens/modals.

Adding a default `<ReadOnLoad>` instance to the shared `<Modal>` component, which can be disabled for more complex modals (e.g. the write-in candidate modal, where we only want a portion of the modal content to be read on open). This is a no-op for apps with no audio and/or translations and also a no-op when there are no `<UiString>` instances in the modal. That being said, open to defaulting to autoplay disabled and having voter-facing modals explicitly enable it, if that feels more appropriate.

### `<ReadOnLoad>` used on Contest Screen and autoplay in overvote modal:

https://github.com/votingworks/vxsuite/assets/264902/25cfce89-d6ca-4230-92e6-f8d6b4057209

### Write-in modal with autoplay disabled and custom `<ReadOnLoad>` setup:

https://github.com/votingworks/vxsuite/assets/264902/f90849e1-c563-4fef-bd19-b01273e5df3b

## Testing Plan
- Existing tests as regression checks
- New test cases for modal autoplay audio functionality

## Checklist

### DIFF
--- a/libs/ui/src/modal.tsx
+++ b/libs/ui/src/modal.tsx
@@ -1,4 +1,4 @@
-import { ReactNode } from 'react';
+import React, { ReactNode } from 'react';
 import ReactModal from 'react-modal';
 import styled from 'styled-components';
 import { rgba } from 'polished';
@@ -8,6 +8,7 @@ import { assert } from '@votingworks/basics';
 import { Theme } from './themes';
 import { ButtonBar } from './button_bar';
 import { H2 } from './typography';
+import { ReadOnLoad } from './ui_strings/read_on_load';
 
 /**
  * Controls the maximum width the modal can expand to.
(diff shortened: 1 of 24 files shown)


libs/ui/src/modal.tsx
```diff
@@ -193,8 +195,11 @@ export function Modal({
       overlayClassName="_"
     >
       <ModalContent centerContent={centerContent} fullscreen={fullscreen}>
-        {title && <H2 as="h1">{title}</H2>}
-        {content}
+        {disableAutoplayAudio ? (
+          modalContent
+        ) : (
+          <ReadOnOpen>{modalContent}</ReadOnOpen>
```
arsalansufi, line 201: Double checking that using `ReadOnOpen` on non-voter facing machines, like VxAdmin, is safe. I imagine it'll just be a no-op on those machines

arsalansufi (approved):
LGTM!

libs/ui/src/modal.tsx
kofi-q: Yeah, it'll be a no-op in cases where there's no `UiStringsContextProvider` at the root, but I should probably make that a little more obvious in the code -- will add a conditional or comment to clarify

PR #4489, opened 2024-01-08
feat(basics): add `iter::toString`

eventualbuddha:
## Overview
It's become a common thing to do `.toArray().join('')` on `iter`s, so this makes that easier.

## Demo Video or Screenshot
```ts
# before
  const partsJoinedToString = (
    await iter(parts).filter(shouldIncludePart).toArray()
  ).join('');

# after
const partsJoinedToString = await iter(parts)
  .filter(shouldIncludePart)
  .toString();
```

## Testing Plan
- New automated tests.


eventualbuddha:
@arsalansufi just review afcadb8 as I stacked #4490 on top of this but accidentally merged it first.

arsalansufi (approved):
LGTM!

libs/basics/src/iterators/async_iterator_plus.ts
```diff
@@ -391,6 +391,10 @@ export class AsyncIteratorPlusImpl<T> implements AsyncIteratorPlus<T> {
     return set;
   }
 
+  async toString(separator = ''): Promise<string> {
```
kofi-q, line 394: opt: Would `.join(separator)` naming work better semantically then `.toString(separator)`? That might be closer to the syntax people are used to when working with arrays.

eventualbuddha: That's a fair point. As `toString` would otherwise have no useful implementation, what do you think of making it an alias for `join`?

kofi-q: Ah, that works too!

PR #4492, opened 2024-01-09
feat(basics): rename `iter.toString` to `iter.join`

eventualbuddha:
## Overview
Per https://github.com/votingworks/vxsuite/pull/4489#discussion_r1446390035.

Keeps `iter.toString` as an alias for `iter.join` as it's semantically nicer when there's no separator.

## Demo Video or Screenshot
n/a

## Testing Plan
- Updated tests.


kofi-q (approved):
:shipit:

PR #4496, opened 2024-01-10
Fix flaky property test for PickDateTimeModal

jonahkagan:
## Overview

Calling `aDate.set` with a time that is exactly the daylight savings time border changes the time zone offset, whereas the date returned from the modal retains the original offset.

## Checklist


arsalansufi (approved):
LGTM!

Issue #4241, opened 2023-11-20
vxmarkscan: accessible controller help


mattroe:
determine how we want to use this

kshen0:
v1 will replay current screen text

PR #4501, opened 2024-01-10
handle audio buttons on accessible controller

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/4240
https://github.com/votingworks/vxsuite/issues/4241

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/9813ea63-4cbb-44ac-a9aa-e3394a78214c

## Testing Plan
- [ ] add frontend tests for handler

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (no new user actions or errors)


apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -206,29 +207,39 @@ fn send_key(device: &mut Device, key: keyboard::Key) -> Result<(), CommandError>
...
-            Button::Help => {
-                key = keyboard::Key::Q;
+            Button::VolumeDown => {
+                key = uinput::event::Keyboard::Misc(keyboard::Misc::VolumeDown);
             }
-            _ => {
-                return Err(CommandError::UnsupportedButton(button));
+            Button::VolumeUp => {
+                key = uinput::event::Keyboard::Misc(keyboard::Misc::VolumeUp);
+            }
```
kshen0, line 240: @kofi-q the 2 handlers for Volume Up/Down are not actually being called because the Volume Up/Down keys are being handled by the OS. IMO this is potentially a feature if it means we can avoid implementing Volume Up/Down hooks in code. But if the hooks do something more sophisticated than just volume up/down, or if this will cause confusion because the implementation differs from VxMark, then I can change they key codes to avoid changing volume at the system level. Thoughts?

kofi-q: Ah, interesting! Also feels like a feature to me -- it could simplify things a bit, because we can then set the default OS volume directly to somewhere around midpoint and just have the controller increment/decrement as needed, rather than amplifying the raw signal from the screen reader audio.

We'd still have to do a bit of work either way to ensure that:
- we can reset the volume to the default level after every voter session
- the volume up/down buttons change the volume in steps of `< 10 db`
- we can clamp the volume to a minimum and maximum in the range specified in VVSG 2

kshen0: Doesn't point 3 mean we shouldn't send the volume up/down events directly from the controller daemon? It seems like that approach doesn't give us enough control to clamp.

Also realizing that this may not work outside GNOME - can verify if this works in OpenBox tomorrow.

kofi-q: > Doesn't point 3 mean we shouldn't send the volume up/down events directly from the controller daemon?

Hm, at the time, I was thinking we might be able to set a minimum and maximum at the system level, but that might not actually be as straightforward as I'd hoped. If that's the case, might be best to assign to the volume controls to regular keyboard keys that we handle within the Vx apps

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -197,7 +198,7 @@ enum Action {
     Pressed = 0x01,
 }
 
-fn send_key(device: &mut Device, key: keyboard::Key) -> Result<(), CommandError> {
+fn send_key(device: &mut Device, key: Keyboard) -> Result<(), CommandError> {
```
kshen0, line 201: @eventualbuddha requesting optional review for idiomatic Rust

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -206,29 +207,39 @@ fn send_key(device: &mut Device, key: keyboard::Key) -> Result<(), CommandError>
...
             }
             Button::Up => {
-                key = keyboard::Key::Up;
+                key = uinput::event::Keyboard::Key(keyboard::Key::Up);
             }
             Button::Down => {
-                key = keyboard::Key::Down;
+                key = uinput::event::Keyboard::Key(keyboard::Key::Down);
+            }
+            Button::Help => key = uinput::event::Keyboard::Misc(keyboard::Misc::Help),
```
kofi-q, line 228: Wondering if we can map these misc keys (`Help` and `Pause`) to regular keyboard keys so dev testing with a keyboard easier -- any downsides to doing that?

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -298,6 +300,35 @@ export function AppRoot({
     }
   }, [endCardlessVoterSessionMutateAsync]);
 
+  const handleKeyboardAudioEvent = useCallback(
```
kofi-q, line 303: Thinking we should keep the keyboard controls consistent across all apps that might need audio -- these handlers could go further upstream to `libs/ui`.

Just added [`<KeyboardShortcutHandlers>`](https://github.com/votingworks/vxsuite/blob/main/libs/ui/src/ui_strings/keyboard_shortcut_handlers.tsx) with a focus on dev convenience, but that could be expanded to include these handlers to centralise things. WDYT?

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -317,9 +328,6 @@ fn main() {
                 match port.read(serial_buf.as_mut_slice()) {
                     Ok(size) => match handle_command(&mut device, &serial_buf[..size]) {
                         Ok(_) => (),
-                        Err(e) if matches!(e, CommandError::UnsupportedButton(_)) => {
-                            println!("[warn] {e}")
-                        }
                         Err(e) => {
                             eprintln!("Unexpected error handling command: {e}")
                         }
```
eventualbuddha, line 333: Since there's now just `Ok` and `Err` cases and you're only doing anything for `Err`, this can be simplified:

```suggestion
                    Ok(size) => {
                        if let Err(e) = handle_command(&mut device, &serial_buf[..size]) {
                            eprintln!("Unexpected error handling command: {e}");
                        }
```

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -206,29 +207,39 @@ fn send_key(device: &mut Device, key: keyboard::Key) -> Result<(), CommandError>
...
     let ButtonStatusCommand { button, action } = data.try_into()?;
 
-    let key: keyboard::Key;
+    let key: Keyboard;
     match action {
         Action::Pressed => match button {
             Button::Select => {
-                key = keyboard::Key::Enter;
+                key = uinput::event::Keyboard::Key(keyboard::Key::Enter);
             }
```
eventualbuddha, line 215: Use the `Keyboard` that you brought in scope above here and the rest of this `match`?
```suggestion
                key = Keyboard::Key(keyboard::Key::Enter);
```

eventualbuddha (approved):
Rust stuff LGTM with some minor tweaks.

apps/mark-scan/accessible-controller/src/controllerd.rs
kshen0: Good point, I'll update to normal keyboard keys. I wanted to make these keys as specific as possible, since I don't love the arbitrary mapping of `[` to rate decrease, for example. Depending on the on-screen keyboard's implementation, I was also worried actual character keys might interfere with write-ins. But since we are using character keys for other functions anyway that's a moot point.

apps/mark-scan/frontend/src/app_root.tsx
kshen0: I like that idea. Looks like `KeyboardShortcutHandlers` is used in `UiStringsContextProvider` so should be de facto imported everywhere, but would it make sense to import in `app_base` or something more explicitly top-level?

apps/mark-scan/accessible-controller/src/controllerd.rs
kofi-q: Yeah, I've been a little worried about collisions -- might be worth figuring out a way to centralise all our keyboard shortcuts across apps. Fortunately don't have to worry about the `VirtualKeyboard` though, since it doesn't fire real key events.

apps/mark-scan/frontend/src/app_root.tsx
kofi-q: I think it could work well in `<AppBase>` if it's generalised beyond just language/audio shortcuts (as long as the relevant handlers are appropriately no-ops when there aren't language/audio contexts present), but that might also require rolling `UiStringsContextProvider` into `AppBase` to make sure `KeyboardShortcutHandlers` is rendered within context. Might be easier to reason about if  `UiStringsContextProvider` installs the keyboard shortcuts it needs

libs/ui/src/ui_strings/keyboard_shortcut_handlers.tsx
```diff
@@ -11,26 +12,47 @@ export function KeyboardShortcutHandlers(): React.ReactNode {
...
+  const audioControls = useAudioControls();
 
   React.useEffect(() => {
     function onKeyPress(event: KeyboardEvent) {
-      // Shift+L: Switch to next available display/audio language:
-      if (event.key === 'L') {
-        const currentLanguageIndex = availableLanguages.findIndex(
-          (l) => l === currentLanguageCode
-        );
+      const currentLanguageIndex = availableLanguages.findIndex(
```
kofi-q, line 19: nit: move L19-22 into the `case 'L'` block to co-locate the related logic

libs/ui/src/ui_strings/keyboard_shortcut_handlers.tsx
```diff
@@ -11,26 +12,47 @@ export function KeyboardShortcutHandlers(): React.ReactNode {
...
+        (l) => l === currentLanguageCode
+      );
+      const nextIndex = (currentLanguageIndex + 1) % availableLanguages.length;
 
-        const nextIndex =
-          (currentLanguageIndex + 1) % availableLanguages.length;
-
-        setLanguage(availableLanguages[nextIndex]);
+      switch (event.key) {
+        // Shift+L: Switch to next available display/audio language:
```
kofi-q, line 25: While you're here: might as well drop this comment, since the code's readable enough without it (along with the one in the jsdoc comment above)

libs/ui/src/app_base.tsx
```diff
@@ -93,6 +94,7 @@ export function AppBase(props: AppBaseProps): JSX.Element {
         screenType={screenType}
         sizeMode={sizeMode}
       >
+        <KeyboardShortcutHandlers />
```
kofi-q, line 97: Re: the previous thread -- this isn't guaranteed to render within a `UiStringsContextProvider`, so better to keep rendering it there for now

apps/mark-scan/accessible-controller/src/controllerd.rs
kshen0: Github was unable to commit this from UI but this change is included in https://github.com/votingworks/vxsuite/pull/4501/commits/2d46774fc3a0e9e0e1a0c6860b0cf2b900713d52

apps/mark-scan/accessible-controller/src/controllerd.rs
kshen0: Since we settled on common keyboard keys we don't need `Keyboard::Misc` any more and simplified the types in https://github.com/votingworks/vxsuite/pull/4501/commits/e1a104bec3a4ac6e5cd13463ee572eb4578678dc

libs/ui/src/ui_strings/keyboard_shortcut_handlers.tsx
kshen0: [bd0969b](https://github.com/votingworks/vxsuite/pull/4501/commits/bd0969b9f918d10a3701c3d26833c0c1646cdae7)

libs/ui/src/ui_strings/keyboard_shortcut_handlers.tsx
kshen0: [bd0969b](https://github.com/votingworks/vxsuite/pull/4501/commits/bd0969b9f918d10a3701c3d26833c0c1646cdae7)

apps/mark-scan/frontend/src/app_root.tsx
kshen0: Returned to rendering `KeyboardShortcutHandlers` in `UiStringsContextProvider` [f642f13](https://github.com/votingworks/vxsuite/pull/4501/commits/f642f13c320a4180cbb5858aa6c43747392618e0).

I think it's somewhat counterintuitive to render from there since the accessible controller and UI strings are separate concerns. But it feels more dangerous to scatter keypress handling across multiple places and possibly introduce conflicts.

kofi-q (approved):
LGTM -- thanks for the keyboard handler consolidation!

apps/mark-scan/frontend/src/app_root.tsx
kshen0: Got it - realizing now how closely the UI strings and screen reader are related, I think it makes sense to keep `KeyboardShortcutHandlers` within `libs/ui`. Plus relying on an implicit order for rendering feels like a big trap door.

PR #4503, opened 2024-01-11
Move repeated dotenv logic to libs/backend and add dotenv logic to VxDesign backend

arsalansufi:
## Overview

This PR moves repeated dotenv logic to `libs/backend` and adds dotenv logic to the VxDesign backend. This logic is necessary for our `.env*` file based feature flagging to work in app backends.

Doing this as a pre-req before I add a feature flag for skipping cloud translation and speech synthesis in VxDesign.

## Testing Plan

Manually tested that env vars in `.env` files are still properly recognized in app backends


eventualbuddha (approved):
🧹

PR #4504, opened 2024-01-11
[i18n/Audio] Update <VirtualKeyboard> for screen reader compatibility

kofi-q:
## Overview

Adding audio-compatible button labels to the `<VirtualKeyboard>` used in BMD write-in modals.

Letter keys are rendered and spoken in English, while punctuation keys are spoken in the user's language (space bar and delete key labels are also translated to user language).

### Follow-Up Improvements Needed:
- It's probably weird to scroll through letters in QWERTY order for audio-only users -- we might want to re-order the keys in the DOM in that case and assign flex ordering to maintain the visual ordering while providing a more intuitive order for audio-only users.
- PAT users have a lot of scrolling to do to get to the `Cancel`/`Accept` buttons, but don't have any immediate ideas for improvements there.
- Currently, pressing a key will just replay the key name - might be useful to say something like "Entered: M" instead.

https://github.com/votingworks/vxsuite/assets/264902/4b48c965-6708-4467-a503-945fe09b4777

## Testing Plan
- New/updated test cases for the utility components and virtual keyboard.
- Manual verification with test audio data.

## Checklist


libs/ui/src/ui_strings/with_alt_audio.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Disables the screen reader for {@link children} and renders
+ * {@link audioText} within a hidden {@link AudioOnly}.
+ *
+ * Useful when we need to provide assistive audio users with additional or
+ * extended information that isn't necessary for or relevant to visual-only
+ * users.
+ */
+export function WithAltAudio(props: WithAltAudioProps): React.ReactNode {
```
arsalansufi, line 37: Nice! I like this abstraction a lot, very clear when used

libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -402,6 +430,58 @@ export const appStrings = {
...
 
+  letterA: () => <UiString uiStringKey="letterA">A</UiString>,
+
+  letterB: () => <UiString uiStringKey="letterB">B</UiString>,
+
+  letterC: () => <UiString uiStringKey="letterC">C</UiString>,
+
+  letterD: () => <UiString uiStringKey="letterD">D</UiString>,
+
+  letterE: () => <UiString uiStringKey="letterE">E</UiString>,
```
arsalansufi, line 441: Was curious how Google Translate was going to handle these single-letter values and it did generate wonky results for some 😆

<table>
mistranslation-e
</table>

Not a big deal at all, just making me realize that even before we build proper override functionality, I'll want to bake some overrides in, maybe by preloading values into the VxDesign translation cache

arsalansufi (approved):
LGTM!

libs/ui/src/ui_strings/app_strings.tsx
kofi-q: Haha! I'd initially forgotten to force English rendering and ended up with a very funky looking keyboard 😅 

And good point -- I keep forgetting about the override functionality and how that'll interact with some of these edge cases. That makes me think we might need two levels of overrides: one for changes VotingWorks makes when we audit the cloud translations and another on top of that the the election officials manage (in case we want them to be able to undo/remove their overrides for some strings). Or maybe Vx overrides/fixes are just applied directly to the base translations layer, since we can always just re-generate from GCloud if needed.

arsalansufi: Yeah good point that these two overrides are conceptually distinct!

PR #4505, opened 2024-01-11
various improvements & PR feedback for #4502

eventualbuddha:
- use `test.each` rather than reimplementing it
- allow printing from a stream of data
- make `exec` return `Result`
- test `exec` using real `child_process.spawn`
- use `iter` for processing `lpinfo` stdout
- use `never` return type for never-returning function

### DIFF
--- a/libs/printing/package.json
+++ b/libs/printing/package.json
@@ -53,7 +53,6 @@
     "@types/debug": "4.1.8",
     "@types/jest": "^29.5.3",
     "@types/jest-image-snapshot": "^6.4.0",
-    "@types/memorystream": "^0.3.4",
     "@types/node": "16.18.23",
     "@types/react": "18.2.18",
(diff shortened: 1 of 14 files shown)


adghayes (approved):
Thanks for the changes! I wasn't aware of `lines` or `testCases`.

Why use `never` instead of `void`?

I like the facelift you gave `exec`. Good to test against real system commands. I see there's a CI failure on one test due to environment. We can swap out for a friendlier command.

Also, now that you revamped `exec.test`, we can remove the `memorystream` and `@types/memorystream` dependency.

Feel free to fix issues and merge, or merge and let me fix issues.

eventualbuddha:
> Why use `never` instead of `void`?

Using `never` communicates to tooling that the function call never returns, which enables things like unreachable code detection:

image

PR #4506, opened 2024-01-11
VxScan: print from backend rather than frontend

adghayes:
## Overview

Use the printing tools added in `libs/printing` to manage printing in one of our apps, VxScan. Currently, the backend tabulates results, sends it to the frontend, which formats the results into a report, and prints that report via `kiosk-browser`. Now, the backend tabulates results, formats the results, and prints it.

Eventually, VxScan will not use the printer management or printing methods of `libs/printing`, only the rendering ability. So some of this work will be changed in the next couple months. But migrating to the backend will have to happen, and it's good to get it out the way now. Also, VxScan served as a better test case for `libs/printing` than any other app.

I have to make a couple changes in `vxsuite-complete-system` to make sure that it supports `libs/printing` for VxScan:
- add `chromium` package
- remove Brother printers from `vxsuite-complete-system` autoconfig file so `kiosk-browser` is not also trying to add them


libs/ui/src/reports/precinct_scanner_tally_reports.tsx
```diff
@@ -22,9 +21,8 @@ export interface PrecinctScannerTallyReportsProps {
   pollsTransition: StandardPollsTransitionType;
   isLiveMode: boolean;
   pollsTransitionedTime: number;
+  reportPrintedTime: number;
```
adghayes, line 24: Besides removing the QR code page in this commit, I parameterized the `reportPrintedTime` for testing purposes.

apps/scan/backend/src/print_report.ts
```diff
@@ -0,0 +1,70 @@
...
+import { getDocument } from 'pdfjs-dist';
+import { Store } from './store';
+import { rootDebug } from './util/debug';
+import { getMachineConfig } from './machine_config';
+import { getScannerResults } from './util/results';
+import { getCurrentTime } from './util/get_current_time';
+
+const debug = rootDebug.extend('print-report');
+
+export async function printReport({
```
adghayes, line 17: Logic here is largely copied from the frontend and simplified.

apps/scan/frontend/src/api.ts
```diff
@@ -16,6 +16,8 @@ import {
   createUiStringsApi,
 } from '@votingworks/ui';
 
+const PRINTER_STATUS_POLLING_INTERVAL_MS = 100;
```
adghayes, line 19: Didn't want to put this in `libs/printing` because not sure whether we'll want the same interval for all apps.

apps/scan/frontend/src/app_tally_report_paths.test.tsx
adghayes, line 1: I found that everything in this file is covered elsewhere. I moved a couple tests to `poll_worker_screen.test.tsx` and the report contents are covered in `libs/ui`.

apps/scan/frontend/src/components/reprint_report_button.tsx
```diff
@@ -17,47 +14,24 @@ interface ReprintReportButtonProps {
...
-      pollsTransitionInfo: lastPollsTransition,
-      electionDefinition,
-      precinctSelection,
-      isLiveMode: !isTestMode,
-      machineConfig: machineConfigQuery.data,
-      scannerResultsByParty: scannerResultsByPartyQuery.data,
-      copies: 1,
-      numPagesCallback: afterPrint,
-    });
+    const numPages = await printTallyReportMutation.mutateAsync();
```
adghayes, line 26: I need to add error handling here.

adghayes: Actually, looks like it's already wrapped in error handling that serves just fine.

PR #4508, opened 2024-01-11
VxAdmin: `TallyCache`

adghayes:
## Overview

Simple implementation of caching to to avoid re-tabulating cast vote records for different export mediums. For example, today if you load a report, it calculates the tally once. If you then export the report as a CSV file, it calculates the tally all over again, a second time. This caches the results of the most computationally expensive part - pulling all CVRs out of the database and tabulating them all. 

It's not very important now, because we aren't trying to optimize CSV exports. But when printing is moved to the backend, previews, prints, and PDF exports will all be entirely different code pathways that need a way to share results.

This is not intended to be a general caching tool in any way.


eventualbuddha (approved):
I ended up making some changes in https://github.com/votingworks/vxsuite/pull/4515 to do more to check that the cache is not stale. Take what you want from that. Mostly I wanted to make it hard to accidentally have a stale cache and for this PR to add tests that would fail if the `tallyCache.clear()` lines are not present.

PR #4515, opened 2024-01-16
test: ensure stale caches get caught

eventualbuddha:
Possible changes to #4508

As long as a cached tally is used before it's cleared it'll be compared with the regenerated value, making sure that the value didn't change. This required an API change so that getting and setting the value happened in a single operation. As a result, it was simpler to always use a tally cache but make it an interface that could do whatever it needed, including not caching.


jonahkagan:
This change almost looks like `memoize(func, hashFn)` (e.g. https://lodash.com/docs/4.17.15#memoize). Might consider using the memoization conceptual model since it's fairly standard

PR #4518, opened 2024-01-17
Kevin/mark scan disable visual mode

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/3689

Enables audio-only mode on VxMarkScan

## Demo Video or Screenshot

Copy in the video isn't up to date; screenshots are.

https://github.com/votingworks/vxsuite/assets/1060688/149ee697-e785-475b-b4b0-a65832fc5199

Screenshot 2024-01-22 at 1 51 28 PM
Screenshot 2024-01-22 at 1 51 32 PM

Screenshot 2024-01-22 at 1 51 36 PM
Screenshot 2024-01-22 at 1 51 40 PM

## Testing Plan
- added tests for button behavior and rendering

## Checklist


libs/ui/src/display_settings_manager_context.tsx
```diff
@@ -13,12 +13,16 @@ export interface ThemeManagerContextInterface {
 
   /** Switches over to the specified size theme. */
   setSizeMode: (mode: SizeMode) => void;
+
+  /** Toggles a full screen overlay that visually blocks underlying elements but keeps them in the accessibility tree */
+  setIsVisualModeDisabled: (isDisabled: boolean) => void;
```
kshen0, line 18: This change snuck into this commit, and is related to a later commit https://github.com/votingworks/vxsuite/pull/4518/commits/d312ab52cdd5aaaecba67fda750586b04a71ec3d

libs/ui/src/display_settings/sensory_toggle_settings.tsx
```diff
@@ -0,0 +1,26 @@
+import { ThemeConsumer } from 'styled-components';
+import React from 'react';
+import { SettingsPane } from './settings_pane';
+import { Button } from '../button';
+import { DisplaySettingsManagerContext } from '../display_settings_manager_context';
+
+export interface SensoryToggleSettingsProps {}
+
+export function SensoryToggleSettings(): JSX.Element {
+  const themeManager = React.useContext(DisplaySettingsManagerContext);
```
kshen0: This is renamed later

libs/ui/src/display_settings_manager_context.tsx
kshen0, line 18: This change is related to a later commit https://github.com/votingworks/vxsuite/pull/4518/commits/d312ab52cdd5aaaecba67fda750586b04a71ec3d but accidentally committed here

libs/ui/src/display_settings/sensory_toggle_settings.tsx
```diff
@@ -0,0 +1,33 @@
+import { ThemeConsumer } from 'styled-components';
+import React from 'react';
+import { SettingsPane } from './settings_pane';
+import { Button } from '../button';
+import { DisplaySettingsManagerContext } from '../display_settings_manager_context';
+import { appStrings } from '../ui_strings';
+
+export interface SensoryToggleSettingsProps {}
+
+export function SensoryToggleSettings(): JSX.Element {
```
kshen0, line 10: I called this "SensoryToggle" because we expect to add an option for visual-only mode here

libs/ui/src/visual_mode_disabled_overlay.tsx
```diff
@@ -0,0 +1,50 @@
...
+`;
+
+export function VisualModeDisabledOverlay(): JSX.Element {
+  const displaySettingsManager = React.useContext(
+    DisplaySettingsManagerContext
+  );
+  const screenInfo = useScreenInfo();
+  return (
+    <Overlay aria-hidden>
+      <Header portrait={screenInfo.isPortrait}>
```
kshen0, line 35: Not sure if this is necessary. Using the factored-out `display_settings/header` component keeps the header consistent with the display settings page, but it might be overkill since we are guaranteed mark-scan screen orientation can't change.

apps/mark-scan/frontend/src/app.tsx
```diff
@@ -81,8 +84,21 @@ export function App({
                 api={uiStringsApi}
                 disabled={!enableStringTranslation}
               >
-                <AppRoot hardware={hardware} reload={reload} logger={logger} />
-                <SessionTimeLimitTracker />
+                <ThemeConsumer>
```
kofi-q, line 87: opt: heads-up that there's a `useCurrentTheme()` hook you could use that might make this code a little cleaner: https://github.com/votingworks/vxsuite/blob/main/libs/ui/src/hooks/use_current_theme.tsx

libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -623,6 +629,24 @@ export const appStrings = {
     <UiString uiStringKey="titleDisplaySettingsSize">Text Size</UiString>
   ),
 
+  titleDisplaySettingsSensoryToggle: () => (
+    <UiString uiStringKey="titleDisplaySettingsSensoryToggle">
+      Accessibility Modes
+    </UiString>
+  ),
+
+  titleEnableAudioOnlyMode: () => (
```
kofi-q, line 638: nit: for consistency, these should probably be `buttonEnableAudioOnlyMode`, `buttonExitAudioOnlyMode`

apps/mark-scan/frontend/src/app.tsx
```diff
@@ -81,8 +84,21 @@ export function App({
...
+                  {(currentTheme) => (
+                    <React.Fragment>
+                      {currentTheme.isVisualModeDisabled && (
```
kofi-q, line 90: Assuming this isn't behaviour specific to mark-scan alone, does it make sense to move this logic to `<AppBase>`?

libs/ui/src/ui_strings/app_strings.tsx
kshen0: Thanks. https://github.com/votingworks/vxsuite/pull/4518/commits/c035cbce847d1dd949e52a63a17865f2c8a2dd01 uses correct semantics for `button` and `title`

apps/mark-scan/frontend/src/app.tsx
kshen0: From my read of the product brief, we'd also need audio-only mode in VxScan. @mattroe can you confirm before we make this change?

kofi-q: FWIW, moving it to a common component like `<AppBase>` shouldn't affect apps that never set `isVisualModeDisabled` to `true`. The potentially confusing piece here is that it's a global setting at the shared theme level, but the corresponding behaviour is only being implemented in one app -- whereas, the toggle to control it is exposed in the shared `<DisplaySettings>` component, which is rendered currently in Mark, Scan, and Mark-Scan. (We might need to make the sensory toggle settings tab optionally enabled, for that matter.)

kshen0: Ah yeah I just wanted to clarify since I realized I don't know if we can make this assumption
> Assuming this isn't behaviour specific to mark-scan alone

Agreed on moving to `<AppBase>` in that case, and also hiding the toggle from apps (if any) that don't need audio-only mode.

kshen0: [b03bf3d](https://github.com/votingworks/vxsuite/pull/4518/commits/b03bf3d734b28c9f0e1c5d325db897dd85d7e0a1)

mattroe: @kshen0 We don't have an audio-tactile interface on VxScan so it's not applicable. It's overall a better experience to just cast an audio-tactile interface ballot on VSAP than to try to cast a ballot you may not be able to verify into a piece of precinct scanning hardware that you may not be able to see.

mattroe:
Minor: do we think we'll add anything to the "Accessibility Modes" menu other than enable/disable of audio/visual-only modes? I can't think of anything. As such, we might want to call the menu button "Audio/Video Only" to be more specific?

kshen0:
> Minor: do we think we'll add anything to the "Accessibility Modes" menu other than enable/disable of audio/visual-only modes? I can't think of anything. As such, we might want to call the menu button "Audio/Video Only" to be more specific?

Plan was just for audio/video modes only so yes I think that name change makes sense

apps/mark-scan/frontend/src/app.tsx
kshen0: https://github.com/votingworks/vxsuite/pull/4518/commits/84bda2dff6d86ee28252588391a46d6efe5de0da updates all naming to variations of "Audio/Video Only", `audioVideoOnlySettings` etc.

kshen0: @kofi-q heads up I reverted back to rendering `VisualModeDisabledOverlay` from `mark-scan/app.tsx` because

1. It's used in mark-scan only
2. Rendering from `app_base` means it takes priority over the `ErrorBoundary` in app-specific `app.tsx` files. So an error would be rendered under the overlay, but we probably want the error to be rendered over the overlay.

kofi-q: ^ ah, that makes sense -- SGTM!

kofi-q (approved):

libs/mark-flow-ui/src/components/display_settings_button.tsx
```diff
@@ -16,7 +16,10 @@ export function DisplaySettingsButton(): JSX.Element | null {
   const history = useHistory();
 
   return (
-    <Button onPress={history.push} value={Paths.DISPLAY_SETTINGS}>
+    <Button
+      onPress={(target: string) => history.push(target)}
+      value={Paths.DISPLAY_SETTINGS}
+    >
```
kshen0, line 22: Addresses https://github.com/votingworks/vxsuite/issues/4469

PR #4519, opened 2024-01-17
refactor: migrate simple system calls (reboot, reboot to BIOS, set clock, power down)

adghayes:
## Overview

Moves the system "mutations" currently called via `kiosk-browser` into `vxsuite`. Along the way, refactored `libs/ui` testing a little bit. 

Because our error boundary has a restart button, and that restart button now requires the API, the `ErrorBoundary` needs to be wrapped in an API provider. But then, we don't catch errors that might happen at the level of the `UiStrings` provider or other API setup. So I factored out the `AppErrorBoundary` and the apps with a restart button now have two `AppErrorBoundary`s - one outside of the API, without a restart button, and one within the API. This could be overkill, but we were testing against the error boundary catching API problems so I maintained it.

If you try to use the system calls in dev, you will hit the same backend errors that you would hit via `kiosk-browser` - the commands say they require "Interactive Authentication." I am a bit confused how the permissions work out correctly for these in prod, still trying to understand that, but should not be a blocker for these changes.

## Demo Video or Screenshot

VxScan error page before & after:

Screen Shot 2024-01-17 at 10 35 38 AM
Screen Shot 2024-01-17 at 10 35 19 AM

The other error boundary pages didn't change.

VxAdmin / VxCentralScan:
Screen Shot 2024-01-17 at 10 30 45 AM

VxMarkScan:
Screen Shot 2024-01-17 at 10 36 41 AM


adghayes:
paired `kiosk-browser` PR: https://github.com/votingworks/kiosk-browser/pull/163

kofi-q (approved):

PR #4520, opened 2024-01-17
Add mechanism for global Vx-managed translation overrides

arsalansufi:
## Overview

Cloud speech synthesis is great! Cloud translation is... good, but sometimes off. My favorite example is the Spanish translation of "Green Party" to "Fiesta Verde" 🟢 🎉 😆

We expect that election officials will proof translations and add their own overrides. But before we implement election-level overrides (which we will before cert), we should also have global Vx-managed overrides, to make the default translations as good as possible. This PR introduces a very rudimentary mechanism for that. We can build upon this rudimentary mechanism (e.g. with a UI and proper `Store` persistence) down the road.

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

## Checklist

### DIFF
--- a/apps/design/backend/src/language_and_audio/translation_overrides.ts
+++ b/apps/design/backend/src/language_and_audio/translation_overrides.ts
@@ -1,0 +1,17 @@
+import { LanguageCode, NonEnglishLanguageCode } from '@votingworks/types';
+
+export type TranslationOverrides = Record<
+  NonEnglishLanguageCode,
+  { [englishText: string]: string | undefined }
+>;
+
+/**
+ * Cloud translations are sometimes incorrect and need to be globally overridden.
+ */
+export const GLOBAL_TRANSLATION_OVERRIDES: TranslationOverrides = {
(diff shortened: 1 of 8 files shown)


apps/design/backend/src/language_and_audio/translation_overrides.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export type TranslationOverrides = Record<
+  NonEnglishLanguageCode,
+  { [englishText: string]: string | undefined }
+>;
+
+/**
+ * Cloud translations are sometimes incorrect and need to be globally overridden.
+ */
+export const GLOBAL_TRANSLATION_OVERRIDES: TranslationOverrides = {
```
arsalansufi, line 11: For now, we can update these global overrides by updating this object. As mentioned in the PR description, down the road, I imagine that we'll persist these overrides in the `Store` and expose some internal UI for updating them.

kofi-q (approved):
🚀

PR #4524, opened 2024-01-18
refactor(admin): memoize tabulation results for scanned CVRs

adghayes:
I thought @jonahkagan's [comment](https://github.com/votingworks/vxsuite/pull/4515#issuecomment-1897165330) on #4515 was clarifying - all I'm really trying to do is memoize a function, and also be able to clear the memoization cache of that function. I took the comment literally and am just using the `lodash` memoization function. If there's opposition to the `lodash` usage I could rewrite something of our own. But it works, and I think it is smaller, cleaner, and simpler than introducing a `TallyCache` object.

I made sure that a test will fail if any of the cache clearing is removed, consolidating the cache assertions in a single `app.caching.test.ts` file.


apps/admin/backend/src/tabulation/full_results.ts
```diff
@@ -82,6 +82,7 @@ const tabulateCastVoteRecordsMemoized = memoize(
  * Clears the memoized results of tabulating cast vote records.
  */
 export function clearTabulationCache(): void {
+  /* c8 ignore next 1 */
```
adghayes, line 85: I don't know why exactly `clear` is typed as an optional property of the memoized function's cache. I've tested and found it always exists, even immediately after the function is initialized.

Because `clear` doesn't work outside of its `this` context, we can't `assertDefined(...)()`. We could typecast, use an eslint ignored non-null assertion, or skip coverage, and this seemed the cleanest and safest.

adghayes:
@jonahkagan adding you for review since @eventualbuddha is out for the rest of the week!

apps/admin/backend/src/tabulation/full_results.ts
eventualbuddha: It's because you can change the caching mechanism to something that doesn't have `clear`, e.g. `WeakMap` (see [this commit](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/6448efea6b05d28d3f77f36c7c9e161af18788ea)).

apps/admin/backend/src/tabulation/full_results.ts
```diff
@@ -57,6 +59,33 @@ export function tabulateCastVoteRecords({
   });
 }
 
+/**
+ * Memoized version of `tabulateCastVoteRecords`.
+ */
+const tabulateCastVoteRecordsMemoized = memoize(
```
eventualbuddha, line 65: While I'm sure this works fine, do we actually need to memoize multiple values? Or do we just need to memoize _the most recent_ value?

apps/admin/backend/src/tabulation/full_results.ts
```diff
@@ -57,6 +59,33 @@ export function tabulateCastVoteRecords({
...
+        unorderedArrays: true,
+        unorderedObjects: true,
+      }
+    )
+);
+
+/**
+ * Clears the memoized results of tabulating cast vote records.
+ */
+export function clearTabulationCache(): void {
```
jonahkagan, line 84: Manual cache invalidation worries me a bit. It's easy to miss a spot, especially as code changes over time (e.g. someone adds a feature that impacts tabulation but forgets to add a call to clear the cache). How big would the impact be if we miss a spot?

A potential way to mitigate this risk would be to tie invalidation to the data itself. E.g. if the data used for tabulation had a `last_updated_at` field that was updated any time the data was changed, you could query that field before loading the full data set and use it to invalidate outdated cache records.

I also want to think through memory management since there are no limits on cache size that I can see. How large is the data being cached? How large can the cache grow before being cleared?

adghayes: The impact would be big (well, equivalent to the same sorts of issues we've had with react query caching) so it's a good call out.

I am also concerned that we could add a function, like deleting individual cast vote records, and not do validation. Two tables influence scanned CVR validation - `cvrs` and `vote_adjudications`. Or actually, we could just use `cvr_files` and `vote_adjudications`. We could have timestamps on rows in both those tables and scan them to clear the cache. Or we could add triggers that update a separately stored, singular timestamp whenever those tables are modified. I think that would be an improvement over the current invalidation.

I don't think we should worry about memory management here. The data being stored is just a JSON representation of the results of an election. There will be one for each report viewed. The amount of data will be less than the amount currently cached on the frontend. cc @eventualbuddha. I think the memory cost of caching here is low and it is worth caching more than one value.

adghayes: I thought about attaching timestamps to the existing data model, but I didn't think that would work as well. For example, say we introduce a feature to delete individual CVR files. You have three CVR files:

- CVR File 1, August 1st
- CVR File 2, August 2nd
- CVR File 3, August 3rd

Let's say you delete the middle file, then you have:

- CVR File 1, August 1st
- CVR File 3, August 3rd

We need some timestamp generated on deletion, which doesn't suit row-by-row timestamps as well.

eventualbuddha (approved):
I like the new database trigger version! However, this version will just keep cached results of calling this function until the process exits (likely due to a restart). We should be able to set a TTL for cached results, maybe for an hour or a day?

apps/admin/backend/schema.sql
```diff
@@ -225,3 +225,41 @@ create table settings (
 );
 
 insert into settings default values;
+
+-- to track data changes in order to invalidate cached data
+create table data_versions (
+  election_id varchar(36),
+  cvrs_data_version integer,
```
jonahkagan, line 232: Would this be simpler as a column on `elections`? Or is there a specific reason to have a separate table (e.g. tracking previous versions)?

apps/admin/backend/src/tabulation/full_results.ts
jonahkagan: Yeah, I think the solution you came up with makes sense.

I don't have a strong opinion about memory management. I'm sympathetic to @eventualbuddha's point that keeping all the past results cached is unnecessary and also to your point that it's a small amount of memory. Maybe a simple drop-in memoize that only keeps one value would be an easy solution https://github.com/alexreardon/memoize-one

apps/admin/backend/schema.sql
adghayes: No specific reason to keep it different other than it seemed semantically different.

apps/admin/backend/src/store.ts
```diff
@@ -2476,6 +2476,20 @@ export class Store {
...
+      `
+        select
+          cvrs_data_version as cvrsDataVersion
+        from data_versions
+        where election_id = ?
+      `,
+      electionId
+    ) as Optional<{ cvrsDataVersion: number }>;
+
+    return row ? row.cvrsDataVersion : 0;
```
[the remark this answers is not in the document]
adghayes: Why preference one style over another here? I find the existing easier to read.

apps/admin/backend/src/tabulation/full_results.ts
adghayes: The current implementation doesn't clear the cache at all, even if it becomes invalid. That's a change I will make before merging, because there's no reason to keep that data around.

But I don't want to move to a `memoize-one` model because it would get rid of some of the benefit. That would mean that if you print the full election report, go preview another report, and come back and print the full election report again, you'll expend the tabulation cost both times you print, whether that's half a second or half a minute. I'd like to avoid that, and keep more than one cached at a time.

jonahkagan: Good call!

eventualbuddha: I understand why you want to keep multiple cache values, but the problem as I see it is that the cache never expires. Imagine I call this function thousands of times with different keys, now we're storing thousands of results in memory and never discarding them until the program exits. There should be something that clears the data after a time, no?

eventualbuddha: > That's a change I will make before merging, because there's no reason to keep that data around.

Ah, missed this. SGTM as long as we're not building up an unbounded cache in memory.

adghayes: Well, it would still be unbounded as long as the CVR data doesn't change. But in order to hit the situation you're talking about - with thousands of results - the user would have to generate thousands of custom reports within a power cycle. I don't think that's going to happen.

apps/admin/backend/src/store.ts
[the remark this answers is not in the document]
eventualbuddha: Mostly because I don't have to repeat the `row` part, which in other places might be something non-trivial and possibly side-effecty.

apps/admin/backend/src/tabulation/full_results.ts
eventualbuddha: It would be a slow leak, but it's still a leak. And it doesn't have to be reports for a single election. If I never reboot my VxAdmin, I'll slowly leak memory as I generate reports over months and years in the extreme case.

Is this likely? No. But IMO we should not create opportunities for unbounded memory leaks in our application. The easiest solution might be to provide `memoize` with an LRU cache with some fixed maximum count.

apps/admin/backend/src/app.caching.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    writeInId,
+    type: 'invalid',
+  });
+  const resultsExportAfterAdjudication = await getParsedExport({
+    apiClient,
+  });
+  expect(tabulationSpy).toHaveBeenCalledTimes(4);
+  expect(resultsExportAfterAdjudication).not.toEqual(doubledResultsExport);
+
+  // adjudicating a mark a vote (by un-invalidating a write-in) should clear the cache
```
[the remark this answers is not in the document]
adghayes: f0942fa1d

apps/admin/backend/src/tabulation/full_results.ts
adghayes: Provided `memoize` with an `LRUMap`: e7aea2740

PR #4526, opened 2024-01-18
[i18n/audio] Add audio-specific data attributes to <UiString>s

kofi-q:
## Overview

Adding a `<WithAudio>` wrapper, used in `<UiString>` for decorating elements with i18n key and language code data attributes. These attributes will be used by the upcoming screen reader to map elements under focus to audio clips for playback.

`<WithAudio>` also pre-fetches associated audio clips on mount, allowing us to load audio clips into memory on page load to reduce the latency between focus/click actions and audio playback. (We're trying to keep this latency under 500ms, per VVSG 2.0 requirements.)

### Batched audio clips fetch on page render:
Screenshot 2024-01-17 at 16 31 44

### Data attributes on relevant elements:
Screenshot 2024-01-17 at 16 34 34

## Testing Plan
- New and updated automated tests
- Manual verification with local prototype screen reader

## Checklist


libs/ui/src/ui_strings/with_audio.tsx
```diff
@@ -0,0 +1,105 @@
...
+  api: UiStringsReactQueryApi;
+  i18nKey: string;
+  languageCode: LanguageCode;
+}) {
+  const { api, i18nKey, languageCode } = props;
+
+  const audioIdsQuery = api.getAudioIds.useQuery(languageCode);
+  const audioIdMappings = audioIdsQuery.isFetching
+    ? undefined
+    : audioIdsQuery.data;
```
arsalansufi, line 53: I know in most places, we check `isSuccess` instead of `isFetching`, e.g.

```ts
if (!audioIdsQuery.isSuccess) {
  return null;
}

const audioIdMappings = audioIdsQuery.data;
```

Is there a particular reason to use `isFetching` instead here?

arsalansufi (approved):
LGTM!

libs/ui/src/ui_strings/with_audio.tsx
kofi-q: Ohh, good question -- wrote this a while ago and now don't remember why 😂 
The `isSuccess` guard pattern's definitely the right way to go here - thanks

PR #4529, opened 2024-01-22
copy `getBatteryInfo` from `kiosk-browser`

adghayes:
## Overview

Copies `getBatteryInfo` from `kiosk-browser` to `vxsuite`. I only use it in VxScan because I want to hit the "no `kiosk-browser`" milestone in VxScan, and I want to test VxScan on a prod image before extending changes to other apps. 

Open question - VxScan is no longer using the `useDevices` hook from `libs/ui`. That hook does a good deal of logging - it logs on printer changes (this is now covered in `libs/printing`) and also logs on any device attach or detach, which is not covered anywhere else. @carolinemodic is logging every device attach and detach necessary for cert? Also - I think the OS is logging on device attach / detach and that ends up in `vx-logs` - if that's the case, is it necessary?

## Demo Video or Screenshot

Still works:
Screen Shot 2024-01-22 at 11 49 05 AM

## Testing Plan

- 100% test coverage for new code
- updated tests in `scan/frontend`
- manual test
- upcoming prod image test

## Checklist


adghayes:
Tagging @kofi-q for review as it's a continuation of the PR last week. Tagging @carolinemodic for answer to my logging question in the description.

libs/backend/src/system_call/get_battery_info.test.ts
adghayes, line 1: @kofi-q this testing - the expected values from the `uevent` file - are just copied from `kiosk-browser`. It's clear someone did some work understanding the file format there and I didn't want to change their work at all.

libs/ui/src/system_call_api.tsx
```diff
@@ -45,6 +47,17 @@ function createReactQueryApi(getApiClient: () => SystemCallApiClient) {
...
+      queryKey(): QueryKey {
+        return ['getBatteryInfo'];
+      },
+      useQuery() {
+        const apiClient = getApiClient();
+        return useQuery(this.queryKey(), () => apiClient.getBatteryInfo(), {
+          refetchInterval: BATTERY_POLLING_INTERVAL_GROUT,
+        });
+      },
+    },
```
adghayes, line 60: I'm adding this here so it's accessible in all the apps, but will only be active if the hook is mounted. So will use in VxScan but will not affect the other apps.

apps/scan/frontend/src/app_root.tsx
```diff
@@ -36,28 +31,25 @@ import {
...
   const configQuery = getConfig.useQuery();
   const pollsInfoQuery = getPollsInfo.useQuery();
   const usbDriveStatusQuery = getUsbDriveStatus.useQuery();
   const checkPinMutation = checkPin.useMutation();
 
-  const { computer } = useDevices({
-    hardware,
-    logger,
-  });
+  const batteryInfoQuery = systemCallApi.getBatteryInfo.useQuery();
```
adghayes, line 52: Previously we had the very light `computer` abstraction which was really just a wrapper / flags derived from the batteryInfo. I think it's fine to just use the raw battery info returned from the backend, and the frontends decide what they want to do with it.

kofi-q (approved):
🎉

carolinemodic:
Yes we need to log every device connection/disconnection. We log new usb devices we see on the OS level but its not a very legible log IIRC, I'm not sure that it covers detaching either. 

In the VVSG the langugae is technically "Removable media that is inserted into or removed
from the voting device" so I guess a printer may not count as that but it seems pretty straightforward to me that we should log a printer connection/disconnection. There is also 21.1-D which says "The voting system must log when a voter-facing scanner, electronic BMD, or other
component is connected or disconnected during the activated voting stage. " which is ... a little confusing in what it means

adghayes:
In the first one, I feel like we could interpret "removable media" as storage media i.e. USB drives, because that's normally what "removable media" means. The second one is confusing, yeah. I'm not sure we need to, but the cost of maintaining is fairly low and arguably useful for debugging, so I will tack onto this PR to keep it, but on the backend.

PR #4537, opened 2024-01-23
feat(basics): add `reduce` method to `iter`

eventualbuddha:
## Overview

This was a bit of a glaring omission since it's a fairly common operation on `Array`.

## Demo Video or Screenshot

```ts
// without an explicit initial value
expect(integers().take(0).reduce((a, b) => a + b)).toBeUndefined();
expect(integers().take(10).reduce((a, b) => a + b)).toEqual(45);

// with an explicit initial value
expect(integers().take(0).reduce((a, b) => a + b, 0)).toEqual(0);
```

## Testing Plan
Updated automated tests.

### DIFF
--- a/libs/basics/src/iterators/types.ts
+++ b/libs/basics/src/iterators/types.ts
@@ -1,4 +1,4 @@
-import { MaybePromise } from '../types';
+import { MaybePromise, Optional } from '../types';
 
 /**
  * An iterable with a number of convenience methods for chaining. Many methods are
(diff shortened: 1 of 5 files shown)


libs/basics/src/iterators/types.ts
```diff
@@ -1035,6 +1063,37 @@ export interface AsyncIteratorPlus<T> extends AsyncIterable<T> {
...
 
+  /**
+   * Reduces elements from `this` using `fn`. Consumes the entire contained
+   * iterable.
+   *
+   * @example
+   *
+   * ```ts
+   * const input = fs.createReadStream('file.txt', { encoding: 'utf8' });
+   * const longestLine = await lines(input).reduce((a, b) => a.length > b.length ? a : b);
```
jonahkagan, line 1074: Maybe good to have an example that's not better implemented with another iter method (in this case `maxBy`)?

jonahkagan (approved):

PR #4539, opened 2024-01-23
Log from mark scan app and daemons

kshen0:
## Overview
Adds logging to mark-scan backend and daemons

## Demo Video or Screenshot
N/A

## Testing Plan
Manually tested logs output in correct format

## Checklist


apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -303,42 +336,109 @@ fn create_virtual_device() -> Device {
...
-
-    println!("Opened controller serial port at {DEVICE_PATH}");
+    log(Log {
+        event_id: "process-started".to_string(),
+        message: format!("Opened controller serial port at {DEVICE_PATH}"),
+        event_type: EventType::SystemAction,
+        ..Default::default()
+    });
+
+    let running = Arc::new(AtomicBool::new(true));
```
kshen0, line 346: I don't think we need this thread-safe bool but using it is consistent with `ctrlc` docs, so might as well

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -303,42 +336,109 @@ fn create_virtual_device() -> Device {
...
+        disposition: Disposition::Success,
+        event_type: EventType::SystemAction,
+        ..Default::default()
+    });
+
+    log(Log {
+        event_id: "controller-connection-init".to_string(),
+        event_type: EventType::SystemAction,
+        ..Default::default()
+    });
```
kshen0, line 381: These 2 logs are redundant atm for debugging, but IMO better to be verbose in case we change the logic later.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -531,7 +532,10 @@ export function buildMachine(
                 // Heavier paper can fail to eject completely and trigger a jam state even though the paper isn't physically jammed.
                 // To work around this, we avoid ejecting to front. Instead, we present the paper so it's held by the device in a
                 // stable non-jam state. We instruct the poll worker to remove the paper directly from the 'presenting' state.
-                entry: async (context) => await context.driver.presentPaper(),
+                entry: async (context) => {
+                  await logger.log(LogEventId.BlankInterpretation, 'system');
```
kshen0, line 536: Was on the fence about this log since technically state transition logging should cover this. But blank interpretation is 1 of 2 flows that requires poll worker intervention and it felt safer to log these explicitly.

apps/mark-scan/backend/src/pat-input/logging.c
```diff
@@ -0,0 +1,51 @@
...
+void print_log_common(char *event_id, char *event_type, char *message, char *operation, char *disposition)
+{
+  printf(
+      "{\"eventId\": \"%s\", \"source\": \"vx-mark-scan-pat-input-daemon\", \"eventType\": \"%s\", \"user\": \"system\", \"message\": \"%s\", \"operation\": \"%s\", \"disposition\": \"%s\"}\n",
+      event_id,
+      event_type,
+      message,
+      operation,
+      disposition);
+}
```
kshen0, line 17: I wasn't sure if going from 0 to 1 C deps (for `json-c` or similar) would introduce build headaches. Since the logging needs are somewhat simple this felt acceptable, but open to feedback or ideas.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -303,42 +336,109 @@ fn create_virtual_device() -> Device {
...
+        message: format!("Opened controller serial port at {DEVICE_PATH}"),
+        event_type: EventType::SystemAction,
+        ..Default::default()
+    });
+
+    let running = Arc::new(AtomicBool::new(true));
+    let r = running.clone();
+
+    if let Err(e) = ctrlc::set_handler(move || {
+        r.store(false, Ordering::SeqCst);
```
eventualbuddha, line 350: I tend to prefer this style where I have a closure that needs a clone of an outer binding:

```suggestion
    if let Err(e) = ctrlc::set_handler({
        let running = running.clone();
        move || {
            running.store(false, Ordering::SeqCst);
        }
```
That way you keep the same name and you can remove the `r` from the outer scope that was only ever needed in this closure.

eventualbuddha (commented):
Adding some thoughts before I switch off to another task for a bit.

apps/mark-scan/accessible-controller/src/logging.rs
```diff
@@ -0,0 +1,60 @@
...
+            event_id: "".to_string(),
+            message: "".to_string(),
+            event_type: EventType::SystemStatus,
+            user: "system".to_string(),
+            disposition: Disposition::NA,
+        }
+    }
+}
+
+pub fn log(log: Log) -> () {
```
eventualbuddha, line 51: Remove unnecessary return type `()`.
```suggestion
pub fn log(log: Log) {
```

libs/logging/scripts/generate_event_id_toml.ts
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash';
```
kshen0, line 1: No need to review this file, but leaving in version control in case we need to change toml file generation

Also note lodash dep will be removed before merge

libs/logging/src/log_event_ids.ts
```diff
@@ -1,36 +1,34 @@
...
- */
-
+export interface LogDetails {
+  eventId: LogEventId;
+  eventType: LogEventType;
+  defaultMessage?: string;
+  documentationMessage: string;
+  // Only includes the log in the documentation file for the given apps, if not specified will be included in all apps documentation.
+  restrictInDocumentationToApps?: LogSource[];
+}
```
kshen0, line 17: Might move this type to a different file.

libs/logging/src/log_event_ids.ts
```diff
@@ -1175,43 +1178,47 @@ const UnknownError: LogDetails = {
 export function getDetailsForEventId(eventId: LogEventId): LogDetails {
   switch (eventId) {
     case LogEventId.ElectionConfigured:
-      return ElectionConfiguredEvent;
```
kshen0, line 962: The variable name of type `LogDetail` is private to this file and now autogenerated, so the enum member (eg. L1177) and variable name (eg. L1178) can no longer diverge.

libs/logging/src/log_event_ids.ts
```diff
@@ -1,36 +1,34 @@
...
+  documentationMessage: string;
+  // Only includes the log in the documentation file for the given apps, if not specified will be included in all apps documentation.
+  restrictInDocumentationToApps?: LogSource[];
+}
 export enum LogEventId {
-  // Election configuration logs
   ElectionConfigured = 'election-configured',
   ElectionUnconfigured = 'election-unconfigured',
-
-  // System level logs
```
kshen0, line 18: I'll manually sort and add these comments back the `.toml` file once we land on format

libs/logging/src/log_event_ids.ts
```diff
@@ -39,23 +37,15 @@ export enum LogEventId {
   UsbDriveMounted = 'usb-drive-mount-complete',
```
kshen0, line 37: There are a handful of events like this where the enum member (`UsbDriveMounted`) differs from the value (`'usb-drive-mount-complete'`) in more than just formatting (the addition of the `-complete` suffix). Ideally if formatting was the only difference, we could do away with the `eventId` field in the `.toml` file and just convert between title case and dash/kebab-case.

Since we'd have to update callsites of those logs it's probably safer to do it in another PR.

libs/logging/scripts/generate_types_from_toml.ts
```diff
@@ -0,0 +1,116 @@
...
+  configFilepath,
+  logEventIdsOutputFile,
+  logEventIdsTemplateFile,
+} from './filepaths';
+
+async function prepareOutputFile(): Promise<void> {
+  await fs.copyFile(logEventIdsTemplateFile, logEventIdsOutputFile);
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kshen0, line 14: Still figuring out the most elegant way to type parsed data.

apps/mark-scan/accessible-controller/src/logging.rs
```diff
@@ -0,0 +1,60 @@
...
+    #[serde(rename = "n/a")]
+    NA,
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Log {
+    pub source: String,
+    #[serde(rename = "event-id")]
+    pub event_id: String,
```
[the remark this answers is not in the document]
kshen0: I agree, https://github.com/votingworks/vxsuite/pull/4539/commits/ab493abeeb19cf8df9cb266869ad7eb6202d0727 generates a TOML file from the existing Typescript types, then generates Typescript types from that TOML file. The biggest pain point in this commit is lack of type safety in the TOML -> Typescript generation. Still trying to square TOML's lack of enums, the TOML parsing library's own types, and our own typing to squash use of `any`.

Also not sure if you'd consider this metaprogramming or compilation and what best practices are. It feels brittle to be writing TypeScript in string literals

libs/logging/src/log_event_ids.ts
```diff
@@ -1,36 +1,34 @@
+/**
```
kshen0, line 1: This file was generated by libs/logging/scripts/generate_types_from_toml.ts

Some diffs are just things moving around so I tried to highlight the main ones with comments

carolinemodic (commented):
lgtm from a logging perspective, will let Brian review the rest

kshen0:
Split into https://github.com/votingworks/vxsuite/pull/4569 and https://github.com/votingworks/vxsuite/pull/4563

PR #4543, opened 2024-01-25
[i18n/audio] Implement <PlayAudioClips> utility component

kofi-q:
## Overview

Fleshing out `<PlayAudioClips>`, which takes in a list of audio IDs and plays their respective audio in order.

This implementation assumes that we're only ever playing audio for one section of the UI at a time -- whenever the audio ID queue changes, playback is stopped for the previous queue and started for the new queue.

Setting up a stubbed interface for the audio player for now - will implement in upcoming PR.

## Testing Plan
- Unit tests for audio queueing and playback rate/volume management.
- Manual verification with prototype audio player

## Checklist

### DIFF
--- a/libs/ui/src/ui_strings/play_audio_clips.tsx
+++ b/libs/ui/src/ui_strings/play_audio_clips.tsx
@@ -1,23 +1,124 @@
-/* istanbul ignore file - pending implementation. */
-
 import React from 'react';
 
 import { LanguageCode } from '@votingworks/types';
+import { assertDefined } from '@votingworks/basics';
+
+import { useAudioContext } from './audio_context';
+import { AudioPlayer, newAudioPlayer } from './audio_player';
 
 export interface ClipParams {
   audioId: string;
   languageCode: LanguageCode;
 }
 
-export interface PlayAudioClipsProps {
+type PlayAudioClipProps = ClipParams & {
+  onDone: () => void;
+};
+
+function PlayAudioClip(props: PlayAudioClipProps) {
+  const { audioId, languageCode, onDone } = props;
+  const [audioPlayer, setAudioPlayer] = React.useState<AudioPlayer>();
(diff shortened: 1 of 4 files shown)


libs/ui/src/ui_strings/play_audio_clips.tsx
```diff
@@ -1,23 +1,124 @@
...
+};
+
+function PlayAudioClip(props: PlayAudioClipProps) {
+  const { audioId, languageCode, onDone } = props;
+  const [audioPlayer, setAudioPlayer] = React.useState<AudioPlayer>();
+  const { api, gainDb, playbackRate, webAudioContext } = assertDefined(
+    useAudioContext()
+  );
+
+  const audioOutputRef = React.useRef<AudioPlayer>();
```
arsalansufi, line 25: Is there a particular reason for calling this `audioOutputRef` instead of `audioPlayerRef`?

libs/ui/src/ui_strings/play_audio_clips.tsx
```diff
@@ -1,23 +1,124 @@
...
+  const gainDbRef = React.useRef<number>(gainDb);
+  const onDoneRef = React.useRef(onDone);
+  const playbackRateRef = React.useRef<number>(playbackRate);
```
arsalansufi, line 28: For my understanding, is the usage of refs throughout this component to allow for their use in effects without having the effects update in response to changes to them? A quick comment noting that could be helpful if so

arsalansufi (approved):
LGTM!

libs/ui/src/ui_strings/play_audio_clips.tsx
kofi-q: Good call -- will add a comment. Might also be able to re-organise the audio player interface a bit to make some of these unnecessary.

PR #4546, opened 2024-01-26
[i18n/audio] Implement screen reader audio playback

kofi-q:
## Overview

Adding an `AudioPlayer` implementation for the `UiStringsScreenReader`.

Currently using the `GrainPlayer` from [`Tone.js`](https://tonejs.github.io/), which is primarily intended for music applications, but provided the least glitchy implementation of pitch-invariant playback rate control that I found. It's working well as a v1 -- we could choose to write our own implementation later down the line if we find functional or performance issues once we've tested on prod hardware.

## Limitations/TODOs:
- After some idle time with no audio playing, the next clip played stutters a for a fraction of a second, so we may need to play some quiet warmup audio at regular intervals during moments of silence to smooth that out.
- Switching up from slower playback rates causes some audio to be lost, due to way data is buffered to handle rate changes.

https://github.com/votingworks/vxsuite/assets/264902/c51a9a57-b4d6-4f13-b45c-d68988f46042

## Testing Plan
- Unit tests with Tone.js mocked/stubbed out
- Manual tests with audio data from VxDesign

## Checklist

### DIFF
--- a/libs/ui/src/ui_strings/audio_player.ts
+++ b/libs/ui/src/ui_strings/audio_player.ts
@@ -46,11 +46,6 @@ const RATE_CONFIG_VERY_SLOW: RateBasedConfig = {
   grainSizeSeconds: 0.02,
 };
 
-const RATE_CONFIG_SLOW: RateBasedConfig = {
-  grainOverlapSeconds: 0.05,
-  grainSizeSeconds: 0.1,
-};
-
 const RATE_CONFIG_DEFAULT: RateBasedConfig = {
   grainOverlapSeconds: 0.05,
   grainSizeSeconds: 0.1,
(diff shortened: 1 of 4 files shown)


libs/ui/src/ui_strings/audio_player.ts
```diff
@@ -14,15 +24,154 @@ export interface AudioPlayer {
...
+
+const RATE_CONFIG_VERY_SLOW: RateBasedConfig = {
+  grainOverlapSeconds: 0.025,
+  grainSizeSeconds: 0.02,
+};
+
+const RATE_CONFIG_SLOW: RateBasedConfig = {
+  grainOverlapSeconds: 0.05,
+  grainSizeSeconds: 0.1,
+};
```
arsalansufi, line 52: I see that this config is the same as the default config. Okay with keeping it to make fine tuning easier but also okay with dropping it to avoid confusion, and just keeping the very-slow and default configs

arsalansufi (approved):
Cool to see all the fine tuning! LGTM!

libs/ui/src/ui_strings/audio_player.ts
kofi-q: Thanks! I'd missed the fact the last 2 configs had settled on the same params after tuning -- will drop this

PR #4549, opened 2024-01-30
[i18n/audio] Export number strings for speech synthesis

kofi-q:
## Overview

Add a "finalize" step in the app string catalog build process that generates and adds `numberN` catalog entries for a limited set of numbers for which we need audio generated.

Also throwing an error if `<NumberString>` is used in an audio context with a number with no audio available -- not ideal that it's a runtime check, but want to avoid missing audio content silently failing to play in a real-world situation. We'll eventually be asserting the more general case of all rendered audio strings needing to have corresponding clips, so this is mostly to provide a more specific error for this case.

## Testing Plan
- New/updates unit tests
- Manual verification with test audio data

## Checklist


libs/ui/scripts/finalize_app_strings_catalog.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import {
+  UiStringTranslations,
+  UiStringTranslationsSchema,
+  safeParseJson,
+} from '@votingworks/types';
+import { generateNumberStringsCatalog } from '../src/ui_strings/number_strings';
+
+const APP_STRINGS_CATALOG_FILE_PATH =
+  './src/ui_strings/app_strings_catalog/latest.json';
```
arsalansufi, line 11: This path makes it such that this script has to be run from the base of `libs/ui`, which is fine since we'll almost always be running this through `pnpm build:app-strings-catalog`. But for flexibility, could use `__dirname` here, e.g. `path.join(__dirname, '../src/ui_strings/app_strings_catalog/latest.json')`

libs/ui/scripts/finalize_app_strings_catalog.ts
```diff
@@ -28,7 +28,9 @@ export function main(): void {
   appStringsCatalog = addNumberStrings(appStringsCatalog);
 
   const sortedAppStringsCatalog: UiStringTranslations = {};
-  for (const appStringKey of Object.keys(appStringsCatalog).sort()) {
+  for (const appStringKey of Object.keys(appStringsCatalog).sort((a, b) =>
+    a.localeCompare(b, undefined, { numeric: true })
```
arsalansufi, line 32: TIL about `{ numeric: true }`!

libs/ui/src/ui_strings/number_strings.ts
```diff
@@ -10,19 +10,22 @@ import { Dictionary, UiStringTranslations } from '@votingworks/types';
 export const MAXIMUM_SUPPORTED_NUMBER_FOR_TTS = 45;
 export const NUMBER_STRINGS_BASE_I18N_KEY = 'number';
 
+export function getI18nKeyForNumber(value: number): string {
+  return `${NUMBER_STRINGS_BASE_I18N_KEY}${value}`;
```
arsalansufi, line 14: > Flatten number string keys for VxDesign compatibility

Ah yes that's my bad! I assumed the two levels of nesting for all files but `appStrings.json`

libs/ui/scripts/finalize_app_strings_catalog.ts
kofi-q: ah, thanks - good call

libs/ui/scripts/finalize_app_strings_catalog.ts
kofi-q: same 😅

libs/ui/src/ui_strings/number_strings.ts
kofi-q: No worries -- thought of making the change in vxdesign instead, but decided I kind of like the simplicity of keeping the `appStrings.json` flat, so considering it a feature now 😁

PR #4550, opened 2024-01-30
[i18n/audio] Improve audio feedback for candidate selection/deselection.

kofi-q:
## Overview

Add a "votes remaining in this contest" audio readout after a voter selects or deselects a candidate. This provides audio-only votes feature parity with the corresponding "votes remaining" status  rendered at the top of the contest page.

## Demo Video or Screenshot [🔊]

https://github.com/votingworks/vxsuite/assets/264902/3f5cc03d-93ef-4abe-9888-87d6cc2bd1ac

## Testing Plan
- Additional test assertions for the new behaviour
- Manual run with test audio data

## Checklist

### DIFF
--- a/libs/mark-flow-ui/src/components/candidate_contest.tsx
+++ b/libs/mark-flow-ui/src/components/candidate_contest.tsx
@@ -92,30 +92,43 @@ export function CandidateContest({
   const [writeInCandidateModalIsOpen, setWriteInCandidateModalIsOpen] =
     useState(false);
   const [writeInCandidateName, setWriteInCandidateName] = useState('');
-  const [deselectedCandidate, setDeselectedCandidate] = useState('');
+  const [recentlyDeselectedCandidate, setRecentlyDeselectedCandidate] =
+    useState('');
+  const [recentlySelectedCandidate, setRecentlySelectedCandidate] =
+    useState('');
 
   const screenInfo = useScreenInfo();
 
   useEffect(() => {
-    if (deselectedCandidate !== '') {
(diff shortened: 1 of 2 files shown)


libs/mark-flow-ui/src/components/candidate_contest.tsx
```diff
@@ -92,30 +92,43 @@ export function CandidateContest({
...
       return () => clearTimeout(timer);
     }
-  }, [deselectedCandidate]);
+  }, [recentlyDeselectedCandidate]);
+
+  useEffect(() => {
+    if (recentlySelectedCandidate !== '') {
+      const timer = setTimeout(() => {
+        setRecentlySelectedCandidate('');
+      }, 100);
```
arsalansufi, line 115: If the `<AudioOnly>` element is removed while the audio is still being read out, the audio will still be allowed to finish, is that right? I imagine so given the video in the PR description - just double checking

arsalansufi (approved):
Nice!

libs/mark-flow-ui/src/components/candidate_contest.tsx
kofi-q: Yeah, good question - the screen reader does check to see if the element's still attached to the DOM before playing audio, but only at the beginning, so this just allows enough time for the audio to start, at which point the element can be removed without affecting playback

arsalansufi: Got it, thanks for explaining!

PR #4552, opened 2024-01-31
WIA Backend Report Printing

adghayes:
## Checklist


apps/admin/backend/src/app.results.test.ts
```diff
@@ -84,159 +81,3 @@ test('card counts', async () => {
     },
   ]);
 });
-
-test('election write-in adjudication summary', async () => {
```
adghayes, line 88: replaced by new test file

apps/admin/backend/src/app.ts
```diff
@@ -898,6 +905,35 @@ function buildApi({
       });
     },
 
+    async printWriteInAdjudicationReport(): Promise<void> {
+      return printWriteInAdjudicationReport({
+        store,
+        printer,
+        logger,
+        userRole: assertDefined(await getUserRole()),
```
adghayes, line 913: having to pass the `userRole` around in order to pass it to the logger is getting a little silly. not for this PR, but we ought to refactor this to logger perhaps has that state, or a hook into it. @jonahkagan has brought this up before

apps/admin/backend/src/app.ts
```diff
@@ -898,6 +905,35 @@ function buildApi({
       });
     },
 
+    async printWriteInAdjudicationReport(): Promise<void> {
```
adghayes, line 908: I'm encountering a general testing dilemma with rendering PDFs on the backend. Above this method, there's the `getElectionWriteInSummary` method. Its purpose is to provide the data for building a WIA report. But now the frontend doesn't need it - yet I'm not removing it.

It's nice to test at the app-level, but if the app only exposes PDFs, it's trickier to test than data blobs. There are some tests in `app.write_ins.test.ts` that depend heavily on this endpoint. But... the app doesn't use it so there's a risk of drift between the actual interface and the testing interface. I can think of a few ways to deal with them, none too bad or too good:

1. Let the endpoint stick around and use it for testing, the risk of divergence is small.
2. Develop means to read the text content of PDFs and parse it, so that we can actually extract the information from the PDFs. This is annoying and fragile testing, but it would do the best job of testing the actual output of the app. I explored this a bit, leveraging the Linux `pdftotext` command-line call. It seemed possible, but like a good deal of work parsing the text output.
3. Test this stuff a level up (or down?), not at the `app` level. I like the sound of that, but I think these sorts of tests can be finicky in a different way, because between the `store` and the `app` we don't have a clear intermediate layer defined to test at. Tests would involve sort of cobbling together helpers, store functions, and some logic that needs to be extracted from `app.ts`.

Suggestions or other ideas? I would argue it's fine to leave it as-is for now, but I want to think through the problem.

apps/admin/backend/src/reports/__image_snapshots__/wia-report-adjudicated-1.png
adghayes, line 1: I think it's pretty neat we now have living PDFs of our reports in our repo.

libs/ui/src/themes/render_with_themes.tsx
```diff
@@ -110,6 +110,14 @@ export interface VxQueryFunctions {
...
     opts?: ButtonQueryOptions
   ): ReturnType<queries.FindByRole>;
+  queryButton(
+    buttonText: string | RegExp,
+    opts?: ButtonQueryOptions
+  ): ReturnType<queries.QueryByRole>;
+  queryAllButtons(
+    buttonText: string | RegExp,
+    opts?: ButtonQueryOptions
+  ): ReturnType<queries.AllByRole>;
```
adghayes, line 120: We have all the others (get, getAll, find, findAll) so we might as well have all 6.

apps/admin/frontend/.vscode/settings.json
```diff
@@ -1,6 +1,5 @@
 {
   "editor.tabSize": 2,
-  "eslint.packageManager": "pnpm",
```
adghayes, line 3: this is deprecated now, `eslint` detects package manager

apps/admin/frontend/package.json
```diff
@@ -86,12 +86,14 @@
     "normalize.css": "^8.0.1",
     "pagedjs": "^0.1.40",
     "path": "^0.12.7",
+    "pdfjs-dist": "3.11.174",
```
adghayes, line 89: We need to serve the frontend a version of the `pdf.worker` file, from `pdfjs-dist`. The recommended `react-pdf` way wasn't working for me, so I did what @eventualbuddha had previously done.

We can't use the `pdfjs-dist` version installed in other parts of the repo, because the version mismatch with `react-pdf`'s version causes the worker to fail. So I installed `pdfjs-dist` here in `admin/frontend`, with a version matching `react-pdf`. But now I'm running afoul of monorepo validation. @eventualbuddha is there a blacklist for that?

We could upgrade `pdfjs-dist` across the repo but I haven't scoped that out, it's a major version change.

Alternatively, we could just try to serve the `pdf.worker.min.js` file out of the public folder. It would be copy-pasta'd, but it might be easier than explicitly hosting it from the prodserver.

apps/admin/frontend/src/api.ts
```diff
@@ -468,6 +470,24 @@ export const listPotentialElectionPackagesOnUsbDrive = {
...
+  useQuery() {
+    const apiClient = useApiClient();
+    return useQuery(
+      this.queryKey(),
+      () => apiClient.getWriteInAdjudicationReportPreview(),
+      {
+        cacheTime: 0,
+        staleTime: 0,
+        refetchOnWindowFocus: false,
+      }
```
adghayes, line 486: backend only has one preview at a time, so frontend only tracks a preview if it's mounting. the assumption here is that we only have one preview at a time. if we chose to change that, it would break (but could be easily fixed).

apps/admin/frontend/src/components/print_button.tsx
adghayes, line 1: this `print_button` uses the backend status rather than the kiosk-derived status

adghayes: otherwise, this component is mostly copy pasted and I don't want to dig into it too much now. the deprecated version will be removed before merging.

apps/admin/frontend/src/components/reporting/pdf_viewer.tsx
```diff
@@ -0,0 +1,127 @@
...
+      <Document
+        file={`${window.location.protocol}//${window.location.host}/${pdfResourcePath}`}
+        loading={<PreviewLoading />}
+        onLoadSuccess={(page) => setNumPages(page.numPages)}
+      >
+        <ThumbnailContainer>
+          <Page
+            pageNumber={pageNumber}
+            scale={2.3}
+            renderMode="svg"
```
adghayes, line 93: SVG render mode is deprecated, but the default `canvas` mode makes the fonts look grainy. Interestingly, this is only in `kiosk-browser`, not Chrome. So maybe there's some difference in `canvas` font handling across Chrome versions? Anyway, using SVG mode generates a bunch of console warnings but it works great. I've troubleshot the `canvas` mode by playing with dimensions and scaling, but couldn't get it to look clean.

apps/admin/backend/src/reports/helpers.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const emptyDir = promisify(emptyDirCallback);
+
+/**
+ * Returns the path of a temporary file in the workspace's preview directory.
+ * A PDF file should be written to the path, which will be served to the
+ * frontend through a static route.
+ *
+ * On each call, the previous preview file is deleted. This aims to prevent
+ * memory leaks and to make stale previews fail loudly.
```
eventualbuddha, line 14: Technically this wouldn't be a memory leak, but a series of files on disk that never get cleaned up. I guess a more general term might be "resource leak", but that feels dissatisfying. Maybe "storage leak"?

apps/admin/backend/src/app.ts
```diff
@@ -891,6 +905,35 @@ function buildApi({
...
+    }): Promise<ExportDataResult> {
+      return exportWriteInAdjudicationReportPdf({
+        path: input.path,
+        store,
+        logger,
+        userRole: assertDefined(await getUserRole()),
+      });
+    },
+
+    async getWriteInAdjudicationReportPreview(): Promise<string> {
```
eventualbuddha, line 928: Are we writing the PDF to disk because of a Grout limitation? Can we just use the browser's built-in PDF rendering rather than going to great lengths to incorporate pdfjs? Returning a `Buffer` wouldn't work because Grout turns it into something the browser can't just use. Perhaps its time we added an escape hatch to Grout to allow returning a `Response` that directly controls the HTTP response.

apps/admin/backend/src/util/get_current_time.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Simply a wrapper for `Date.now()`, returns the current time in Unix time.
+ *
+ * Separated out into its own file so that it can be mocked in tests.
```
eventualbuddha, line 4: Using `jest` to mock the date doesn't work for some reason?

apps/admin/backend/src/app.ts
eventualbuddha: Could you call `this.getElectionWriteInSummary()` in this function to ensure you're using the same data?

apps/admin/frontend/src/components/deprecated_print_button.tsx
```diff
@@ -0,0 +1,119 @@
...
+    />
+  );
+}
+
+interface PrintButtonProps extends StyledButtonProps {
+  print: () => Promise<void>;
+  useDefaultProgressModal?: boolean;
+}
+
+export function DeprecatedPrintButton({
```
eventualbuddha, line 62: There's also the `@deprecated` JSDoc tag that shows up as a strikethrough in VS Code. Also, for deprecated things I like to add a comment that explains when and how it should be removed. Either an explanation inline or a link to a Github issue.

apps/admin/frontend/src/components/reporting/export_wia_report_pdf_button.tsx
```diff
@@ -0,0 +1,85 @@
...
+  disabled,
+}: {
+  disabled?: boolean;
+}): JSX.Element {
+  const { electionDefinition, isOfficialResults } = useContext(AppContext);
+  assert(electionDefinition);
+  const { election } = electionDefinition;
+
+  const [isSaveModalOpen, setIsSaveModalOpen] = useState(false);
+  const [exportDate, setExportDate] = useState<Date>();
```
eventualbuddha, line 27: Could use `useNow` hook.

eventualbuddha (commented):
I'm not sure whether building the whole PDF viewer client-side is the right approach. Did you consider using the browser's built-in rendering and just respond with the PDF data as a `application/pdf` response? Also, why do we write the PDFs to a file to be served later instead of just returning a `Buffer` (or letting the browser handle it as per the previous question)?

apps/admin/backend/src/app.ts
jonahkagan: FWIW I'm also working on PDF ballot previews and I've been experimenting with just sending the Buffer via grout. It seems to work pretty well.

I think using the browser's built-in PDF rendering would be cool, but from what I've seen you can't customize the UI at all. In Chrome there's a toolbar with a bunch of extra functionality (e.g. printing, downloading) that we probably don't want.

apps/admin/backend/src/app.ts
adghayes: That's a cool idea!

apps/admin/backend/src/util/get_current_time.ts
adghayes: It made more sense on the frontend because, when using fake timers, the time advances within test even if you mock it to begin with. So you have to remember to mock the date before every render. But on the backend, when renders are more controlled, could probably just go back to `jest`.

apps/admin/backend/src/app.ts
adghayes: I only went with the file approach because I'm already introducing loading latency in this PR, and I thought that the base64 encoding and decoding would make it even slower. But I was probably overthinking it, a `Buffer` should be fine and easier to manage. 

I'm hesitant about using built-in functionality because I think that would continue to tie us to a specific browser rather than a generic, simple renderer. My understanding is that different browsers roll their own PDF interface and they are different, and something like WebKitGTK might not have one at all.

apps/admin/frontend/src/components/deprecated_print_button.tsx
adghayes: I don't intend for this deprecated component to ever be merged to main, this is mostly for me.

apps/admin/frontend/src/components/reporting/export_wia_report_pdf_button.tsx
adghayes: Oh cool, didn't know about that.

PR #4553, opened 2024-01-31
VxAdmin: remove L&A functionality

adghayes:
## Overview

Ballot printing was moved to VxDesign, then HMPB test deck ballot printing was moved, and then test deck tally report printing was moved. BMD test deck ballot still hasn't been moved to VxDesign. But L&A logic in VxAdmin is already hidden behind flags that mean it won't appear for VxDesign elections. Stripping it out of the app frees us up to move in the direction of our ideal architecture - doing the heavily lifting and system calls on the backend.


apps/admin/frontend/src/app.test.tsx
```diff
@@ -629,32 +533,6 @@ test('election manager cannot auth onto machine with different election hash', a
   );
 });
 
-test('primary election flow', async () => {
```
adghayes, line 632: the only purpose of this test was to test L&A documents for a primary settings, so it's also no longer necessary

arsalansufi (approved):
mr-clean-lgtm

PR #4569, opened 2024-02-02
Kevin/mark scan logging 2

kshen0:
## Overview
* Adds logging to mark-scan frontend, backend, and daemons
* Migrates PAT input daemon from C to Rust to converge on Rust logging library

## Demo Video or Screenshot
N/A

## Testing Plan
- [ ] Add automated tests

## Checklist


apps/mark-scan/frontend/src/pages/remove_invalidated_ballot_page.tsx
```diff
@@ -18,7 +20,11 @@ export function RemoveInvalidatedBallotPage(props: Props): JSX.Element {
     history.push('/ready-to-review');
   });
 
-  function onPressContinue() {
+  async function onPressContinue() {
+    await logger.log(
+      LogEventId.PollWorkerConfirmedBallotRemoval,
+      'poll_worker'
+    );
```
kshen0, line 27: Most behavior requiring logs is common to VxMark or covered in the backend

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -303,42 +328,103 @@ fn create_virtual_device() -> Device {
 }
 
 fn main() {
-    // Open the serial port
-    let port = serialport::new(DEVICE_PATH, DEVICE_BAUD_RATE)
-        .timeout(POLL_INTERVAL)
-        .open();
-
-    println!("Opened controller serial port at {DEVICE_PATH}");
+    set_app_name(APP_NAME.to_string());
```
eventualbuddha, line 331:
```suggestion
    set_app_name(APP_NAME);
```

apps/mark-scan/frontend/src/pages/ballot_invalidated_page.test.tsx
```diff
@@ -29,6 +32,7 @@ function renderWithAuthAndBallotContext(
 ): VxRenderResult {
   return render(
     <BallotInvalidatedPage
+      logger={logger}
```
eventualbuddha, line 35: Feels like we need a `useLogger` hook.

apps/mark-scan/pat-device-input/src/patinputd.rs
```diff
@@ -0,0 +1,138 @@
...
+    // Create virtual device for keypress events
+    log!(
+          EventId::CreateVirtualUinputDeviceInit;
+          EventType::SystemAction
+    );
+    let mut device = create_virtual_device();
+    // Wait for virtual device to register
+    thread::sleep(Duration::from_secs(1));
+    log!(
+
```
eventualbuddha, line 83:
```suggestion
```

apps/mark-scan/pat-device-input/src/patinputd.rs
```diff
@@ -0,0 +1,138 @@
...
+        event_id: EventId::ConnectToPatInputInit,
+        message: format!("Connected to PAT with initial values [is_connected={is_connected}], [signal_a={signal_a}], [signal_b={signal_b}]")
+    );
+
+    loop {
+        if !running.load(Ordering::SeqCst) {
+            log!(
+                event_id: EventId::ProcessTerminated,
+                event_type: EventType::SystemAction
+            );
```
eventualbuddha, line 116:
```suggestion
            log!(EventId::ProcessTerminated; EventType::SystemAction);
```

apps/mark-scan/pat-device-input/src/patinputd.rs
```diff
@@ -0,0 +1,138 @@
...
+            message: e.to_string(),
+            event_type: EventType::SystemStatus
+        );
+    }
+
+    // Create virtual device for keypress events
+    log!(
+          EventId::CreateVirtualUinputDeviceInit;
+          EventType::SystemAction
+    );
```
eventualbuddha, line 78:
```suggestion
    log!(EventId::CreateVirtualUinputDeviceInit; EventType::SystemAction);
```

apps/mark-scan/pat-device-input/src/patinputd.rs
```diff
@@ -0,0 +1,138 @@
...
+    SIGNAL_B_PIN.set_up()?;
+    Ok(())
+}
+
+fn main() {
+    set_app_name(APP_NAME);
+    log!(
+        EventId::ProcessStarted;
+        EventType::SystemAction
+    );
```
eventualbuddha, line 57:
```suggestion
    log!(EventId::ProcessStarted; EventType::SystemAction);
```

apps/mark-scan/pat-device-input/src/patinputd.rs
```diff
@@ -0,0 +1,138 @@
...
+    // is_connected is unused for keypresses in this daemon but it's important to export
+    // the pin so it can be read by the mark-scan backend
+    let is_connected = IS_CONNECTED_PIN.is_active();
+    let mut signal_a = SIGNAL_A_PIN.is_active();
+    let mut signal_b = SIGNAL_B_PIN.is_active();
+
+    log!(
+        event_id: EventId::ConnectToPatInputInit,
+        message: format!("Connected to PAT with initial values [is_connected={is_connected}], [signal_a={signal_a}], [signal_b={signal_b}]")
+    );
```
eventualbuddha, line 109:
```suggestion
    log!(
        EventId::ConnectToPatInputInit,
        "Connected to PAT with initial values [is_connected={is_connected}], [signal_a={signal_a}], [signal_b={signal_b}]"
    );
```

apps/mark-scan/pat-device-input/src/patinputd.rs
```diff
@@ -0,0 +1,138 @@
...
+            SIGNAL_A_PIN.tear_down();
+            SIGNAL_B_PIN.tear_down();
+            exit(0);
+        }
+
+        let new_signal_a = SIGNAL_A_PIN.is_active();
+        let new_signal_b = SIGNAL_B_PIN.is_active();
+
+        if new_signal_a && !signal_a {
+            send_key(&mut device, keyboard::Key::_1).unwrap();
```
eventualbuddha, line 127: Do we want these to crash if it fails? Or just log an error?

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn export(&self) -> io::Result<()> {
+        fs::write(EXPORT_PIN_FILEPATH, self.to_string())
+    }
+
+    fn unexport(&self) -> io::Result<()> {
+        fs::write(UNEXPORT_PIN_FILEPATH, self.to_string())
+    }
+
+    fn set_direction_in(&self) -> io::Result<()> {
+        let filepath = format!("/sys/class/gpio/gpio{}/direction", self);
```
eventualbuddha, line 31:
```suggestion
        let filepath = format!("/sys/class/gpio/gpio{self}/direction");
```

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -0,0 +1,84 @@
...
+            "1" => false,
+            _ => {
+                log!(
+                    EventId::PatDeviceError,
+                    "Unexpected value for pin #{self}: {string_value}",
+                );
+                false
+            }
+        }
+    }
```
eventualbuddha, line 77: Consider not crashing on errors and simplifying the value comparison:

```suggestion
    pub fn is_active(&self) -> bool {
        let filepath = format!("/sys/class/gpio/gpio{self}/value");
        let Ok(mut file) = File::open(filepath) else {
            log!(
                EventId::PatDeviceError,
                "Failed to open file for pin {self}",
            );
            return false;
        };

        let mut buf = [0; 1];
        if let Err(e) = file.read_exact(&mut buf) {
            log!(
                EventId::PatDeviceError,
                "Failed to read file for pin {self}: {e}",
            );
            return false;
        }

        // Pin status does not follow boolean convention. We translate to typical
        // boolean convention here.
        // 1 is the default state, 0 is actioned state
        // connection status: 1 when PAT device is not plugged in, 0 when plugged in
        // A/B signal: 1 when no signal is sent from device, 0 when signal is sent
        match buf[0] {
            b'0' => true,
            b'1' => false,
            _ => {
                log!(
                    EventId::PatDeviceError,
                    "Unexpected value for pin #{self}: {char}",
                    char = buf[0] as char,
                );
                false
            }
        }
    }
```

eventualbuddha (commented):
Needs CircleCI config updates.

apps/mark-scan/pat-device-input/src/patinputd.rs
kshen0: Initially I was thinking this should crash because it's useless without successful keypress events, but I suppose the errors could be intermittent. In that case we wouldn't want to crash. So far I haven't seen uinput fail in any way, so it's hard to say what type of error we'd get.

All said, it's probably safer to log the error and continue. Process restarts can always be achieved by restarting the machine.

apps/mark-scan/frontend/src/pages/ballot_invalidated_page.test.tsx
kshen0: In line with discussion in https://github.com/votingworks/vxsuite/pull/4592#discussion_r1488741427, [579f510](https://github.com/votingworks/vxsuite/pull/4569/commits/579f510b785d3fe76d0162cdfec727c36210606b) moves logging to the backend

apps/mark-scan/pat-device-input/src/patinputd.rs
kshen0: [779745a](https://github.com/votingworks/vxsuite/pull/4569/commits/779745a8a06502e8e4b235adce8509d2536ad1e9)

kshen0:
> Needs CircleCI config updates.

Do you mind if I do this in a follow up PR? I didn't realize `circleci.yml` is autogenerated. From a quick look, the generation script only handles pnpm packages atm and needs to be extended to handle Rust crates.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -394,17 +393,23 @@ fn main() {
...
-                            log!(
-                                event_id: EventId::UnknownError,
-                                message: format!(
-                                    "Unexpected error when handling controller command: {e}"
-                                ),
-                                event_type: EventType::SystemStatus
-                            );
+                    Ok(size) => match handle_command(&serial_buf[..size]) {
+                        Ok(key_option) => {
+                            if key_option.is_some() {
```
kshen0, line 398: Refactored `handle_command` to avoid passing the `device`, which is cumbersome to create in CI because it requires special permissions. This seemed easier than configuring udev rules for CI.

The downside is I introduced this nested conditioning to handle when the command should no-op ie. when the controller indicates one of its buttons was released. We only want to send keyboard events when the button is depressed.

To support this, `handle_command` returns `Some(keyboard::Key)` on controller button press, and `None(keyboard::Key)` on button release.

 Open to feedback on how to handle these checks more cleanly/inline.

eventualbuddha: This can be improved by being more specific about the types you're looking for in your match arms:

```rust
                        Ok(Some(key)) => {
                            if let Err(err) = send_key(&mut device, key) {
                                log!(EventId::UnknownError, "Error sending key: {err}");
                            }
                        }
                        Ok(None) => {}
```

In general, you don't want to use `is_some()` or `is_ok()` et al when you want the value being wrapped. Use `match`/`if let`/`let/else` destructuring

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -374,7 +463,6 @@ mod tests {
 
     #[test]
     fn test_handle_command_success() {
-        let mut device = create_virtual_device();
         // In prod we wait 1s for the device to register.
         // We can afford to be riskier to speed up tests.
         thread::sleep(DEVICE_WAIT_DURATION);
```
eventualbuddha, line 468: This isn't needed, right?

apps/mark-scan/frontend/src/pages/remove_invalidated_ballot_page.test.tsx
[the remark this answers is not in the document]
kshen0: Somewhere along the way this code path was reported as uncovered so I added this test.

eventualbuddha, line 1: Not supposed to be here?

apps/mark-scan/frontend/src/pages/remove_invalidated_ballot_page.tsx
eventualbuddha, line 1: Not supposed to be here?

apps/mark-scan/pat-device-input/src/patinputd.rs
```diff
@@ -0,0 +1,138 @@
...
+    log!(EventId::ConnectToPatInputInit);
+
+    // Export pins and set direction
+    set_up_pins().unwrap_or_else(|error| {
+        log!(
+            event_id: EventId::ConnectToPatInputComplete,
+            disposition: Disposition::Failure,
+            message: format!("An error occurred during GPIO pin connection: {error}")
+        );
+    });
```
eventualbuddha, line 91: When you're not using the value it seems a little weird to use `unwrap_or_else`. Maybe do this?

```suggestion
    if let Err(error) = set_up_pins() {
        log!(
            event_id: EventId::ConnectToPatInputComplete,
            disposition: Disposition::Failure,
            message: format!("An error occurred during GPIO pin connection: {error}")
        );
    }
```

Also, do we want to continue after this?

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -0,0 +1,97 @@
...
+    io::{self, Read},
+    str::from_utf8,
+};
+
+use vx_logging::{log, EventId};
+
+const EXPORT_PIN_FILEPATH: &str = "/sys/class/gpio/export";
+const UNEXPORT_PIN_FILEPATH: &str = "/sys/class/gpio/unexport";
+
+pub struct Pin {
```
eventualbuddha, line 13: Write a `///` doc comment?

apps/mark-scan/pat-device-input/src/patinputd.rs
```diff
@@ -0,0 +1,138 @@
+use pin::Pin;
```
eventualbuddha, line 1: Write a `//!` module-level doc comment? I know a fair amount of the existing Rust code doesn't do this, but I'm trying to encourage better habits here.

eventualbuddha (approved):
A couple minor things, but LGTM!

apps/mark-scan/pat-device-input/src/patinputd.rs
kshen0: [5171668](https://github.com/votingworks/vxsuite/pull/4569/commits/517166896e8fd00ab8869627f3026539370d59ee)
good catch, thank you

apps/mark-scan/pat-device-input/src/patinputd.rs
kshen0: [a75cf84](https://github.com/votingworks/vxsuite/pull/4569/commits/a75cf8485fb1ec441b67502eb474b7038219a9c8)

doc comments added to both daemons and `Pin` struct

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -10,33 +10,54 @@ use vx_logging::{log, EventId};
...
+    /// Removes access to the pin from userspace.
     pub fn tear_down(&self) {
         self.unexport()
             .expect("Unexpected failure to tear down pin");
     }
 
+    /// Makes the pin accessible from userspace. If the pin is already set up due to
+    /// a previous `set_up` call, or from other callers of the sysfs interface, `set_up`
+    /// will tear down the pin and attempt one more time to set up. This may cause
+    /// interruption to other processes attempting to read the pin value.
```
kshen0, line 59: This brings up the question of whether we should just continue when we see the pin is already exported. IMO since we expect this daemon to be the only exporter it should never rely on the unexpected state of the pin already being exported. ie. it should always be exercising the path to export the pin.

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -51,6 +72,9 @@ impl Pin {
         self.set_direction_in()
     }
 
+    /// Reads the pin value and returns the boolean inverse of the pin's raw value.
+    /// # Example
+    /// When the pin's value is b'0', `is_active` returns true.
```
kshen0, line 76: I believe this convention is specific to VSAP's implementation, so it could arguably live in the daemon code. But since we have no planned use for this module outside the PAT integration it feels cleaner to leave this translation in the pin module.

PR #4588, opened 2024-02-07
feat(logging): `log!` macro and various other improvements

eventualbuddha:
## Overview
Follow-up to #4563.

## Demo Video or Screenshot
```rust
use vx_logging::{log, set_app_name, Disposition, EventId};

set_app_name("my-app");

// shorthand versions: event ID and optional message
log!(EventId::MachineBootComplete);
log!(EventId::AuthLogin, "User {user} logged in", user = "test-user");

// full version: specify any set of fields by name
log!(
   message: format!("User {user} blocked!", user = "bad-user"),
   event_id: EventId::AuthLogin,
   disposition: Disposition::Failure
);
```

## Testing Plan
New automated tests.


libs/logging/bacon.toml
```diff
@@ -0,0 +1,78 @@
+# This is a configuration file for the bacon tool
+# More info at https://github.com/Canop/bacon
```
kshen0, line 2: Not familiar with bacon but it looks like an editor config tool? Should this be added to `.gitignore`?

kshen0 (approved):
Thanks for these improvements!

libs/logging/bacon.toml
eventualbuddha: It's actually an auto test/clippy/check runner that I recommend installing. I'd like to have a way to share these settings throughout the workspace, but I just copied this from `ballot-interpreter` for now.

PR #4590, opened 2024-02-08
VxDesign 3.1: Render nested HTML formatting tags

jonahkagan:
## Overview

Previously, we were just rendering the outermost tag. Now we traverse the tree and render nested tags.

## Checklist


eventualbuddha (approved):
Would be nice to have a test that only works with this change.

jonahkagan:
I'll add one before changing this code again, but skipping for expediency right now

PR #4591, opened 2024-02-09
Dev Dock Printer

adghayes:
## Overview

Add printer to the dev dock for VxAdmin and VxScan (for a limited time only).

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/37960853/a665f2bd-6963-4b1c-b4af-55b0b471d5da

Screen Shot 2024-02-09 at 4 00 38 PM


libs/dev-dock/backend/src/dev_dock_api.ts
```diff
@@ -121,6 +137,20 @@ function buildApi(devDockFilePath: string, machineType: MachineType) {
       usbHandler.clearData();
     },
 
+    getPrinterStatus(): PrinterStatus {
+      return printerHandler.getPrinterStatus();
+    },
+
+    connectPrinter(): void {
+      const config = DEFAULT_PRINTERS[machineType];
```
jonahkagan, line 145: Instead of passing in the machine type and casing on it inside the dev dock, maybe it would be cleaner to pass in a printer config to the dev dock

adghayes: We could pass in the config to the dev dock, maybe that would be a bit cleaner. But I think we do want to have the machine type available on the frontend. When we hopefully add more product-specific mocks, it'll be nice to have a consistent way to determine whether to show a feature or not. Rather than having each feature (e.g. printing mock) handle it in its own way. With the machine type being there anyway, I don't think the way we determine the printing config (passing in vs. having this lookup within the dev dock) matters.

jonahkagan: Could which features are turned on also be configured with options passed into dev dock? I generally try to build interfaces that don't need to know their calling context (in this case, the machine type) so that components can be composed more easily.

I don't think it's super important in this particular case to hold this interface to that standard, so I wouldn't block on it, but wanted to at least raise the general principle.

adghayes: We could configure features with passed options, definitely, but that configuration still needs to be passed to the frontend for the frontend to know what to render, or alternatively, each relevant endpoint needs to find a way to communicate whether it should be active or not. Passing around the machine type seems to me like a compact and natural way to do the same thing.

I think I understand the principle, but machine type seems important enough to be passed as context (in our prod builds, we make it available as context at an OS level) and since it's the dev dock... I'm not thinking too hard about it.

libs/dev-dock/frontend/src/dev_dock.tsx
```diff
@@ -453,6 +458,79 @@ function ScreenshotControls({
...
+  });
+  const disconnectPrinterMutation = useMutation(apiClient.disconnectPrinter, {
+    onSuccess: async () =>
+      await queryClient.invalidateQueries(['getPrinterStatus']),
+  });
+
+  const status = getPrinterStatusQuery.data ?? undefined;
+
+  function onPrinterClick() {
+    if (status?.connected === true) {
```
jonahkagan, line 503: `=== true` seems unnecessary

jonahkagan (approved):
Thanks for adding this

Issue #3871, opened 2023-08-23
Swap ballot box at 200 ballots


kshen0:
Blocked by https://github.com/votingworks/vxsuite/issues/3899 or else we'll do duplicate work, once to build in frontend state where ballot printed count exists today and once to port to backend store

PR #4592, opened 2024-02-12
Kevin/mark scan ballot box swap

kshen0:
## Overview
Can be reviewed by commit.

https://github.com/votingworks/vxsuite/issues/3871

* Prompts a poll worker to empty the ballot box after 200 ballots have been cast.
* Left-aligns text in all `AskPollWorkerPage` instances
* Fixes a warning caused by `P` and `H1` tags nested under `P`

## Demo Video or Screenshot

The ballot box maximum was set to 1 for this video.

https://github.com/votingworks/vxsuite/assets/1060688/a776efc0-8c2e-4361-a2e2-a721b3a7cba9

VxScan for reference:

Screenshot 2024-02-05 at 1 26 40 PM

Screenshot 2024-02-05 at 1 27 01 PM

## Testing Plan
- added tests

## Checklist


libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -642,6 +642,12 @@ export const appStrings = {
     </UiString>
   ),
 
+  noteBmdBallotBoxIsFull: () => (
+    <UiString uiStringKey="noteBmdBallotBoxIsFull">
+      A poll worker must empty the full ballot box.
+    </UiString>
```
kshen0, line 648: We need different strings than the existing "Replace Ballot Box" text because the ballot box is not replaced like in VxScan; rather, it's opened, emptied, and then closed again.

apps/mark-scan/frontend/src/pages/ask_poll_worker_page.tsx
```diff
@@ -17,16 +15,17 @@ interface Props {
 // 1. Voter stage: tells the voter to ask a poll worker for help. Does not allow the voter to advance the state.
 // 2. Poll worker stage: after auth, gives instructions and a button to advance the state.
 export function AskPollWorkerPage(props: Props): JSX.Element {
-  const { authStatus, children, pollWorkerPage } = props;
+  const { authStatus, children, hideTitle, pollWorkerPage } = props;
```
kshen0, line 18: Supporting hidden titles allows us to render `Icon`s (or any other element) above the title. Without this change the `Icon` is rendered between the title and body which looks awkward.

apps/mark-scan/frontend/src/pages/empty_ballot_box_page.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+});
+
+test('requires poll worker auth', () => {
+  const electionDefinition = electionGeneralDefinition;
+  const authStatus = fakeCardlessVoterLoggedInAuth(electionDefinition);
+  render(<EmptyBallotBoxPage authStatus={authStatus} />, {
+    apiMock,
+  });
+
+  userEvent.click(screen.getByText('Insert a poll worker card to continue.'));
```
jonahkagan, line 32: Why click on this?

apps/mark-scan/backend/schema.sql
```diff
@@ -7,6 +7,7 @@ create table election (
   is_test_mode boolean not null default true,
   polls_state text not null default "polls_closed_initial",
   ballots_printed_count integer not null default 0,
+  ballots_cast_since_last_ballot_box_change integer not null default 0,
```
jonahkagan, line 10: Any reason to not use the same schema as apps/scan here? (`ballot_count_when_ballot_bag_last_replaced`)? IIRC correctly, we landed on that because it only needed to be updated when the ballot bag actually gets replaced (rather than on every ballot), so there's a little less chance for something to go wrong.

apps/mark-scan/frontend/src/pages/empty_ballot_box_page.tsx
```diff
@@ -34,7 +36,10 @@ function ConfirmBallotBoxEmptied(): JSX.Element {
       <ConfirmButtonWrapper>
         <Button
           variant="primary"
-          onPress={() => confirmBallotBoxEmptiedMutation.mutate()}
+          onPress={async () => {
+            await logger.log(LogEventId.BallotBoxEmptied, 'poll_worker');
+            confirmBallotBoxEmptiedMutation.mutate();
+          }}
```
jonahkagan, line 42: Maybe log this on the backend? I'm imagining a world in the future where we might be able to get rid of frontend logging completely, since all the important user actions are handled on the backend. Plus, it's probably better to only log it if it actually succeeds

apps/mark-scan/frontend/src/app_states.test.tsx
```diff
@@ -99,7 +99,7 @@ test('`waiting_for_invalidated_ballot_confirmation` state renders ballot invalid
     'waiting_for_invalidated_ballot_confirmation.paper_present'
   );
   apiMock.setAuthStatusCardlessVoterLoggedInWithDefaults(electionDefinition);
-  await screen.findByText('Ask a Poll Worker for Help');
```
jonahkagan, line 102: I would consider "ask a poll worker for help" a bit plainer than "notify a poll worker." I know it's a matter of opinion but generally important to think about (more on this here: https://civicdesign.org/topics/plain-language/)

apps/mark-scan/frontend/src/constants.ts
```diff
@@ -1,4 +1,4 @@
 /* istanbul ignore file */
-export const STATE_MACHINE_POLLING_INTERVAL_MS = 300;
+export const STATE_MACHINE_POLLING_INTERVAL_MS = 200;
 // Overrides shorter libs/ui polling interval
-export const AUTH_STATUS_POLLING_INTERVAL_MS_OVERRIDE = 300;
+export const AUTH_STATUS_POLLING_INTERVAL_MS_OVERRIDE = 400;
```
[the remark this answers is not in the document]
jonahkagan: Could this be addressed by changing the precedence order that screens are rendered in? I'm not sure this solution of changing the polling intervals will be reliable enough.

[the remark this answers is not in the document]
kshen0: Unfortunately the "Insert card" screen is already the last screen in the render order. The logic can be summarized as `if (any auth including logged out) && state === 'empty_ballot_box' then render EmptyBallotBoxPage, else if logged out then render InsertCardPage`. So if the frontend sees `authStatus` is `logged_out` the only thing that will avoid flicker is if the frontend already has the updated state.

If we think the polling interval change is too hacky, we could change the last few states of the machine. I'll push a commit with that fix and we can see if it feels more reliable without too much complexity added.

apps/mark-scan/backend/schema.sql
kshen0: If I understand correctly, you're suggesting that we write `ballot_count_when_ballot_box_last_replaced` with the value of `ballots_printed_count` each time the box is emptied? This could work in practice but would be less accurate.

mark-scan allows voters to invalidate a ballot, so if 5 of 200 voters invalidate their ballot and end up printing 2 ballots each, there would be 200 ballots printed and 195 ballots in the box and 5 spoiled ballots elsewhere.

Since this approach would cause the ballot box to be emptied sooner rather than later, we don't risk overfilling the box, but we can't make any guarantees about the number of ballots in the box. I thought it might be a potential cause for alarm if we say the box limit is 200, a poll worker changes the box, and later they find only 195 ballots from that box change. With correct ballot accounting one would be able to attribute the other 5 to invalidated ballots.

Do you have a sense for how important it is that the number of ballots in the box at change time equals exactly the stated limit?

apps/mark-scan/frontend/src/pages/empty_ballot_box_page.tsx
kshen0: I think @carolinemodic mentioned we might want to log button presses on the frontend for actions that require poll worker intervention. Is that still the case? If not, we can delete this entirely because the automatic state transition logging will cover this case.

apps/mark-scan/frontend/src/app_states.test.tsx
kshen0: Agreed - I made this change to avoid wrapping a single word to the next line in the title. Any ideas for more concise, plain language that could fit on one line? Maybe "Ask a Poll Worker" or "Poll Worker Needed"?

apps/mark-scan/frontend/src/pages/empty_ballot_box_page.tsx
carolinemodic: We are supposed to log any action a user (poll worker, election admin) takes and whether or not it succeed/fails

apps/mark-scan/frontend/src/app_states.test.tsx
jonahkagan: Hmm, could we just shrink the font size a tiny bit or change the padding? I don't have a visual reference so if that's a bad idea, ignore me.

apps/mark-scan/frontend/src/pages/empty_ballot_box_page.test.tsx
kshen0: Typo fixed in [754ca8d](https://github.com/votingworks/vxsuite/pull/4592/commits/754ca8d15e92ae7613d047874462024d9d47bfb0)

apps/mark-scan/frontend/src/pages/empty_ballot_box_page.tsx
kshen0: [d9675c1](https://github.com/votingworks/vxsuite/pull/4592/commits/d9675c13f4c6f957dc2386af22b9af507abcc8d8)

jonahkagan (approved):

apps/mark-scan/frontend/src/app_states.test.tsx
kshen0: [6207da7](https://github.com/votingworks/vxsuite/pull/4592/commits/6207da7bac4decbe3609ec8c5f8d8ee13b7a617a) reduces the font size

apps/mark-scan/frontend/src/constants.ts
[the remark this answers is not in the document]
kshen0: [357453a](https://github.com/votingworks/vxsuite/pull/4592/commits/357453a8650bb0b143ce6f621c9c0b7147761f92) renders `BallotSuccessfullyCastPage` (the screen for the previous state) regardless of auth status to avoid the flicker.

PR #4597, opened 2024-02-13
admin(feat): battery status

adghayes:
## Overview

Makes VxAdmin battery-aware, via three separate features:
1. Adds a battery display to the nav bar which indicates a variably filled battery icon, the percent, and a charge icon if the laptop is charging.
2. Creates a modal alert when the battery level drops to a warning level. Currently I have **dismissable** warnings at 15%, 10%, and 5%, and a non-dismissable warning at 1%.
3. Adds the same battery info the diagnostics screen.

#### Battery Display

Screen Shot 2024-02-12 at 12 57 23 PM

Screen Shot 2024-02-12 at 12 56 58 PM

Screen Shot 2024-02-12 at 12 56 31 PM

#### Battery Warning

Screen Shot 2024-02-13 at 11 00 30 AM

Screen Shot 2024-02-13 at 11 00 09 AM

#### Diagnostics Screen

Screen Shot 2024-02-13 at 11 41 23 AM

Screen Shot 2024-02-13 at 11 06 13 AM

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
+++ b/apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
@@ -1,0 +1,50 @@
+import { Font, H2, Icons, P } from '@votingworks/ui';
+
+import { format } from '@votingworks/utils';
+import { NavigationScreen } from '../components/navigation_screen';
+import { getPrinterStatus, systemCallApi } from '../api';
+import { Loading } from '../components/loading';
+
+export function HardwareDiagnosticsScreen(): JSX.Element {
+  const batteryInfoQuery = systemCallApi.getBatteryInfo.useQuery();
+  const printerStatusQuery = getPrinterStatus.useQuery();
+
+  if (!batteryInfoQuery.isSuccess || !printerStatusQuery.isSuccess) {
+    return (
+      <NavigationScreen title="Hardware Diagnostics">
+        <Loading isFullscreen />
+      </NavigationScreen>
+    );
+  }
+
+  const batteryInfo = batteryInfoQuery.data;
+  const printerStatus = printerStatusQuery.data;
+
+  return (
+    <NavigationScreen title="Hardware Diagnostics">
+      <H2>Laptop</H2>
+      <P>
+        <Font weight="semiBold">Power Source:</Font>{' '}
+        {batteryInfo
+          ? batteryInfo.discharging
+            ? 'Battery'
+            : 'External Power Supply'
+          : '-'}
+      </P>
+      <P>
+        <Font weight="semiBold">Battery Level:</Font>{' '}
+        {batteryInfo ? format.percent(batteryInfo.level) : '--%'}
+      </P>
+      <H2>Printer</H2>
+      {printerStatus.connected ? (
+        <P>
+          <Icons.Done color="success" /> Connected
+        </P>
+      ) : (
+        <P>
+          <Icons.Info /> No compatible printer detected
+        </P>
+      )}
+    </NavigationScreen>
+  );
+}
--- a/libs/ui/src/battery_display.tsx
+++ b/libs/ui/src/battery_display.tsx
@@ -1,0 +1,71 @@
+import styled from 'styled-components';
+import type { BatteryInfo as BatteryInfoType } from '@votingworks/backend';
+import { format } from '@votingworks/utils';
+import { assert } from '@votingworks/basics';
+import { IconProps, Icons } from './icons';
+import { Font } from './typography';
+import { useSystemCallApi } from './system_call_api';
+
+const BatteryInfo = styled.div`
+  align-self: center;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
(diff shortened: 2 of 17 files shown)


adghayes:
@mattroe adding you for feedback

libs/ui/src/battery_low_alert.tsx
```diff
@@ -0,0 +1,94 @@
...
+        return;
+      }
+
+      // if there is an applicable warning, show it
+      const applicableWarning = BATTERY_WARNINGS.find(
+        (w) => w.level === newLevel
+      );
+      if (applicableWarning) {
+        setCurrentWarning(applicableWarning);
+      }
```
adghayes, line 62: The logic of this component is that it gives a warning when the battery is at specific levels, rather than ranges. It keeps it slightly simpler and means we manage less state. But it means that, for example, if you booted up your computer at 14% you wouldn't get a warning immediately, not until the battery drops to 10%. We also are making the assumption that battery decline is smooth and passes through all percentiles, rather than dropping and skipping levels. That is the behavior of my development machine battery, not sure if that's universal.

apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <H2>Printer</H2>
+      {printerStatus.connected ? (
+        <P>
+          <Icons.Done color="success" /> Connected
+        </P>
+      ) : (
+        <P>
+          <Icons.Info /> No compatible printer detected
+        </P>
+      )}
```
adghayes, line 47: Printer content is just a placeholder for now, will make it richer in a subsequent PR.

libs/ui/src/icons.tsx
```diff
@@ -94,7 +100,7 @@ function iconColor(theme: UiTheme, color?: IconColor) {
   }[color];
 }
 
-function FaIcon(props: InnerProps): JSX.Element {
+export function FaIcon(props: InnerProps): JSX.Element {
```
kofi-q, line 103: is this `export` from an older approach? looks like it's not used externally

libs/ui/src/battery_display.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <BatteryInfo>
+      <IconContainer>
+        {batteryInfo ? getBatteryIcon(batteryInfo) : <Icons.BatteryFull />}
+        {batteryInfo && !batteryInfo.discharging && (
+          <Icons.Bolt style={{ fontSize: '0.8em' }} />
+        )}
+      </IconContainer>
+      <Font weight="semiBold" style={{ fontSize: '0.8em', lineHeight: '0.9' }}>
```
kofi-q, line 66: nit: seeing the use of the `style` prop a few times here, which is functionally fine, but breaks from the pattern of using styled components - non-blocking, but consider using the latter for consistency in the codebase

libs/ui/src/battery_low_alert.tsx
```diff
@@ -0,0 +1,94 @@
...
+    <Modal
+      actions={
+        currentWarning.canDismiss ? (
+          <Button onPress={() => setCurrentWarning(undefined)}>Dismiss</Button>
+        ) : undefined
+      }
+      content={
+        <React.Fragment>
+          <H1>
+            <Icons.BatteryQuarter color="danger" /> Low Battery Warning
```
kofi-q, line 84: nit: use title prop for styling consistency

kofi-q (commented):
UX looks great!
One concern with the alert triggering, but all other comments non-blocking

libs/ui/src/icons.tsx
adghayes: yup, removing, thank you for calling out

adghayes: b1cfb2eb9

libs/ui/src/battery_display.tsx
adghayes: I'm don't think we're being consistent currently. I've seen @jonahkagan use `style` quite a bit lately, and I understand why, because if it's only one or two properties I think it's more readable to have it inline, TailwindCSS style. I think we would have to discuss to decide what our ideal pattern is.

But I think within-file consistency makes sense - I moved the font props up into the styled component in 370f8098a, but left the `Icons.Bolt` style because it seems like a reasonable use case for inline styles - I'd rather have `fontSize` inline than make a new `styled` component or do something tricky with selectors to add the one-off property.

libs/ui/src/battery_low_alert.tsx
adghayes: Fair point. I realized this was how it worked when I was writing testing but wanted some other input. Fixing this up now.

adghayes:
We are starting to wrap (we may have already been) in cases where the title is too long and the session time limit comes up (edge case) but I don't think it looks too bad:

Screen Shot 2024-02-13 at 3 57 16 PM

kofi-q (approved):
🚀

PR #4599, opened 2024-02-14
[VxDesign] Add language_code field to the speech_synthesis_cache table

kofi-q:
## Overview

Realised after adding the numeric number strings for translation/audio generation that we're now getting speech synthesis cache hits across different languages, so the numbers were getting spoken in the same language, regardless of language setting. Adding a `language_code` field to the primary key to distinguish audio between text that might be written the same in different languages, but pronounced differently.

## Testing Plan
- Updated cache test case

## Checklist


arsalansufi (approved):
Good catch!

PR #4605, opened 2024-02-16
Kevin/rust circleci config

kshen0:
## Overview
Includes Rust crates in circleci config generation.

## Demo Video or Screenshot
n/a

## Testing Plan
Ran manually and confirmed CI jobs are added

## Checklist

### DIFF
--- a/libs/monorepo-utils/src/cargo.ts
+++ b/libs/monorepo-utils/src/cargo.ts
@@ -5,7 +5,8 @@ import { getAbsoluteRootPath } from './dependencies';
 const exec = promisify(execFile);
 
 // Ballot interpreter tests are already run by the pnpm package test job
-const EXCLUDED_PACKAGE_IDS = ['ballot-interpreter'];
+// patinputd has no tests currently
+const EXCLUDED_PACKAGE_IDS = ['ballot-interpreter', 'patinputd'];
 
 /**
  * Get all Rust crate paths.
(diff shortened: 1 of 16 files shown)


libs/monorepo-utils/src/cargo.ts
```diff
@@ -4,6 +4,9 @@ import { getAbsoluteRootPath } from './dependencies';
 
 const exec = promisify(execFile);
 
+// Ballot interpreter tests are already run by the pnpm package test job
+const EXCLUDED_PACKAGE_IDS = ['ballot-interpreter'];
```
kshen0, line 8: This avoids spinning up a container to re-test `ballot-interpreter`, which is already tested by the pnpm test job for that package. There are a few ways to avoid duplicate test runs:

1. Remove `test:rust-addon` from the `pnpm test` command in `libs/ballot-interpreter/package.json`
2. Special-case `ballot-interpreter` to run in CI with a different command eg. `pnpm test:all-but-rust`
3. This approach

Each has disadvantages
1. Contributors expect to run `pnpm test` to run all tests in the package. Removing `test:rust-addon` from the command may mean contributors are unaware of the Rust tests until they push to CI
2. We need to special-case pnpm test job generation for one repo
3. Essentially special-casing in the opposite direction. Existence of other `test-crate-*` CI jobs implies there should be one for every Rust package

kshen0:
`controllerd` tests fixed in https://github.com/votingworks/vxsuite/pull/4569

eventualbuddha (approved):

PR #4607, opened 2024-02-20
VxAdmin Diagnostics: Show Disk Space

adghayes:
## Overview

Shows available disk space on the VxAdmin Diagnostics page.

We only want to show users diagnostic information that is meaningful to them and would make sense to include in a "readiness report." So most standard computer diagnostics do not make the cut - you don't need to see the RAM performance because there's nothing the user (or us) can do about. Used disk space is technically actionable, although users normally will not want to delete election data to free up space. I think it will be a somewhat useful indicator to see how much disk space we actually use in production, and to catch serious memory leaks. If VxAdmin goes to a multi-election format, disk space _could_ be a regular concern.

I made a few choices in how to define "disk space" here:
- I'm not simply taking the operating system's report of the disk space availability at the given paths. Because that will count a lot of system and other application files, e.g. our code, against the "used" disk space. If we did that, the user would never see disk space be 100% available - a new VxAdmin might be at 93% available or 98% available. This is not abnormal - it's the same with how 128GB phones come with less than 128GB free space. But I thought that approximating _application_ disk space would give a better sense of the status, and be at or near 100% after unconfiguring. The downside is that it adds some state complexity, and it's theoretically possible that due to some OS or frontend events, the application discovers _more_ available space over time, so one day it's 71GB / 71.9GB and the next day it's 71GB / 72.1GB. But I am assuming that the emptiest the file system will be is at first boot up, and this actually isn't a possible event.
- I'm not including `/tmp` storage. I don't want to lump it in, because it would introduce volatility to an otherwise stable metric. And I don't want to show it separately - if it were just for me, I might, to catch `/tmp` memory leaks, but for users it doesn't mean anything (although we could prompt them to restart the computer?) Could add this in if we wanted.
This PR also makes VxAdmin run a database vacuum after deleting an election, so disk space is actually reclaimed from SQLite. It may make unconfiguring a little slower.

## Demo Video or Screenshot

Screen Shot 2024-02-20 at 3 32 35 PM

## Testing Plan
Automated testing + manually checked that disk space results reflect actions on my filesystem.

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
+++ b/apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
@@ -84,7 +84,7 @@ export function HardwareDiagnosticsScreen(): JSX.Element {
             : 'External Power Supply'
           : '-'}
       </P>
-      <H6 as="h3">Memory</H6>
+      <H6 as="h3">Storage</H6>
       <P>
         {storageAvailableLevel < DISK_SPACE_WARN_LEVEL ? (
           <Icons.Warning color="warning" />
--- a/apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
+++ b/apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
@@ -17,7 +17,7 @@ function roundToGigabytes(kilobytes: number): number {
   return Math.round(kilobytes / 100_000) / 10;
 }
 
-const DISK_SPACE_WARN_LEVEL = 0.05;
+const FREE_DISK_SPACE_RATIO_WARN_THRESHOLD = 0.05;
 
 function BatteryStatusIcon({ discharging, level }: BatteryInfo): JSX.Element {
   if (discharging && level < 0.1) {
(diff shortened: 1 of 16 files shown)


apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
```diff
@@ -67,6 +84,17 @@ export function HardwareDiagnosticsScreen(): JSX.Element {
             : 'External Power Supply'
           : '-'}
       </P>
+      <H6 as="h3">Memory</H6>
```
eventualbuddha, line 87: Not sure how regular users think about this, but "Memory" could be interpreted as RAM.
```suggestion
      <H6 as="h3">Storage</H6>
```

apps/admin/frontend/src/screens/hardware_diagnostics_screen.tsx
```diff
@@ -1,13 +1,24 @@
...
+  systemCallApi,
+} from '../api';
 import { Loading } from '../components/loading';
 import { PrintTestPageButton } from '../components/print_diagnostic_button';
 
+function roundToGigabytes(kilobytes: number): number {
+  return Math.round(kilobytes / 100_000) / 10;
+}
+
+const DISK_SPACE_WARN_LEVEL = 0.05;
```
kofi-q, line 20: nit: measurement unit would be useful here for readability (`DISK_SPACE_RATIO_WARN_LEVEL` maybe?)

kofi-q (approved):

PR #4609, opened 2024-02-22
Drew/readiness report

adghayes:
## Checklist


adghayes:
I thought the check circles looked out of place in print, and the boxes look a little better in the UI too because they're more visually distinct from the info circle.

libs/ui/src/reports/tally_report_metadata.tsx
```diff
@@ -7,13 +7,11 @@ import { Text } from '../text';
 interface Props {
   election: Election;
   generatedAtTime?: Date;
-  footer?: JSX.Element;
```
adghayes, line 10: This was unrelated, I just noticed it's no longer used.

libs/ui/src/reports/admin_tally_report.tsx
```diff
@@ -60,54 +59,52 @@ export function AdminTallyReport({
   };
 
   return (
-    <ThemeProvider theme={tallyReportThemeFn}>
-      <TallyReport data-testid={testId}>
-        <ReportSection>
```
adghayes, line 65: I know that combining these doesn't change the styling due to VxAdmin PDF snapshots remaining stable.

libs/logging/src/log_documentation.test.ts
```diff
@@ -26,7 +26,6 @@ describe('test cdf documentation generation', () => {
     expect(structuredData.DeviceModel).toEqual('VxAdmin 1.0');
     expect(structuredData.GeneratedDate).toEqual('2020-07-24T00:00:00.000Z');
     expect(structuredData.EventTypeDescription).toHaveLength(5);
-    expect(structuredData.EventIdDescription).toHaveLength(43);
```
adghayes, line 29: These assertions are annoying and useless because we just change them when they fail.

apps/admin/backend/src/__image_snapshots__/readiness-report-1.png
adghayes, line 1: In a later commit, I change the circles to squares.

mattroe:
Report artifact looks beautiful

libs/ui/test/expect_text_with_icon.ts
```diff
@@ -0,0 +1,16 @@
+import { screen, within } from './react_testing_library';
+
+export async function expectTextWithIcon(
+  text: string,
+  icon: string
+): Promise<void> {
+  const textElement = await screen.findByText(text);
+  expect(
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    within(textElement.closest('p')!)
```
kofi-q, line 10: nit: this `closest('p')` might be a too-specific/brittle assumption for a general test helper - maybe `within(textElement.parentNode)`? (although, i guess the text could be nested in `span`s or broken up across elements)

kofi-q (approved):
Some minor comments -- LGTM overall!

apps/admin/backend/src/app.diagnostics.test.ts
```diff
@@ -99,3 +127,60 @@ test('test print', async () => {
...
+  );
+
+  const printPath = mockPrinterHandler.getLastPrintPath();
+  assert(printPath !== undefined);
+
+  await expect(printPath).toMatchPdfSnapshot({
+    customSnapshotIdentifier: 'readiness-report',
+  });
+
+  const pdfContents = await pdfToText(printPath);
```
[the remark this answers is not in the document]
adghayes: I'm actually pretty bought into this `pdfToText` approach and here's why. I think, for static printed documents, snapshots make more sense than they do for UI components. But I think there's a risk that in big codebase changes, a developer updates a lot of snapshots at once without truly validating each change. I think that checks like this help ensure we are printing the right content / document. We could have checks on the component (like you point out we already do) but checking in the PDF is the final, user-consumed output, which ensures we're putting the right components into the right document.

libs/ui/src/reports/layout.tsx
```diff
@@ -0,0 +1,43 @@
...
+  & > div {
+    margin-top: 0;
+  }
+`;
+
+export const PrintedReport = styled.section`
+  font-size: 12px;
+  page-break-after: always;
+
+  @media print {
```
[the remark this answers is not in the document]
adghayes: I think that both do?

A bunch or report components are defined in `em` so setting the base font size of the report does affect what is rendered.

The page break is important because otherwise, in the few cases where we're rendering multiple reports into one document, reports always start on a new page. For example, for a primary tally report, we wouldn't want the Fish Party Tally Report header and report to start at the bottom of the Mammal Party Tally Report page. Without this style, I think that would happen.

libs/ui/test/expect_text_with_icon.ts
adghayes: Yeah, I agree. I didn't think of it as a general purpose test helper, just a helper for the diagnostics page format. But I was repeating it and wanted to extract it, which put it in the `test` folder. Do you think a scoping comment would make it more acceptable?

libs/ui/src/reports/layout.tsx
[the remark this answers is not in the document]
adghayes: Ah but, for this particular report, I'm not sure if it does anything. But I can test with snapshots to see.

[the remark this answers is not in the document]
adghayes: looks like I'm using `em` in the metadata so will keep something like this wrapper for now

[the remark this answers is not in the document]
kofi-q: ah, I was referring to just the `@media print {}` section -- was wondering if the PDF is rendered within a `@print` context or `@screen` and whether the `@print` styles have any effect. wasn't clear why we have to declare `font-size` twice

libs/ui/test/expect_text_with_icon.ts
kofi-q: gotcha -- maybe move this file into the `src/diagnostics/` package? seems like it's only used there at the moment

libs/ui/src/reports/layout.tsx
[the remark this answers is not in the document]
adghayes: it's rendered in a print context. now that all reports are rendered in a print context, the non-print styling could be removed. but while some other things are still printing on the frontend and the transition is not complete, i'll leave the double styling

libs/ui/test/expect_text_with_icon.ts
adghayes: If I can, I'd like to avoid the pattern of having test helper files in `src`, which either need to be excluded from coverage or be `.test.tsx` files with exports. I've seen (and used) both but using the `test` directory seems preferable. I just added a comment, which will hopefully act as a disclaimer.

PR #4614, opened 2024-02-22
perf(fs): use `opendir` instead of `readdir`

eventualbuddha:
## Overview
If we don't need to list all of the contents of a directory we shouldn't be forced to.

## Demo Video or Screenshot
n/a

## Testing Plan
Updated automated tests.

### DIFF
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -392,23 +392,30 @@ function buildApi({
     async listPotentialElectionPackagesOnUsbDrive(): Promise<
       Result<FileSystemEntry[], ListDirectoryOnUsbDriveError>
     > {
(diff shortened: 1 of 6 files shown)


adghayes (approved):
My reading is that `listDirectory` now doesn't need to list the entire directory, but in both of our use cases we are listing the entire directory. Is the perf boost for other, to-be-written use cases? Or does the asynchronous nature have a performance benefit over `readdir`?

eventualbuddha:
> My reading is that `listDirectory` now doesn't need to list the entire directory, but in both of our use cases we are listing the entire directory. Is the perf boost for other, to-be-written use cases? Or does the asynchronous nature have a performance benefit over `readdir`?

Yeah, the existing uses do require the whole thing. In particular, sorting the values means that we can't stream values as we might like to. This does allow future uses of the underlying `listDirectory` function to stop early and/or not load the entire set of file metadata in memory at once.

PR #4624, opened 2024-02-27
user aware logging - refactor

adghayes:
## Overview

Refactor `Logger` so it is aware of the currently logged in user. The former `Logger` becomes `BaseLogger`, which is not user aware. Setting up `Logger` in the backends requires that the `workspace` and `auth` API are set up first, which forces some re-ordering. 

`BaseLogger` is now in three main places:
- some `libs`
- frontends, although in a bonus commit I pull it out of `admin/frontend`
- entry to backend apps - `server`, `index`, etc.

The third example is necessary due to `auth` and `workspace` bootstrapping. The second is changeable, but requires more refactor work. `libs` is changeable too - most of the `libs` are now being passed a user-aware `Logger` class but don't have their types updated. There's more to this transition but doing as much as I've done here gets us to a point where we're not doing a clunky passing of the `userRole` everywhere on the backend.


adghayes:
@jonahkagan tagging you as main reviewer. @carolinemodic and @arsalansufi, you were auto-tagged as `logging` and `auth` owners respectively.

libs/logging/src/logger.ts
```diff
@@ -1,77 +1,30 @@
...
-    } else {
-      // eslint-disable-next-line no-console
-      console.log(JSON.stringify(logLine));
-    }
+  static from(
+    baseLogger: BaseLogger,
+    getUserRole: () => Promise<LoggingUserRole>
+  ): Logger {
+    const logger = new Logger(baseLogger.getSource(), getUserRole);
+    logger.log = baseLogger.log;
```
jonahkagan, line 27: why is this line necessary?

libs/logging/src/logger.ts
```diff
@@ -1,77 +1,30 @@
...
-  getSource(): LogSource {
-    return this.source;
+    source: LogSource,
+    private readonly getUserRole: () => Promise<LoggingUserRole>
+  ) {
+    super(source);
   }
 
-  async log(
+  async logAsCurrentUser(
```
jonahkagan, line 14: Is it possible to have this method just be called `log`? I'm thinking it might be more obvious to use as the main method of the logger that way (e.g. when using auto-complete).

apps/admin/backend/src/server.ts
```diff
@@ -76,9 +82,16 @@ export async function start({
       config: {
         allowElectionManagersToAccessUnconfiguredMachines: false,
       },
-      logger,
+      logger: baseLogger,
     });
 
+    const logger = Logger.from(baseLogger, () =>
+      getUserRole(auth, assertDefined(resolvedWorkspace))
+    );
```
jonahkagan, line 90: Another way you could work around the bootstrapping problem:

```ts
// Start with a Logger that always returns `system`
const systemLogger = new Logger(LogSource.VxAdminService, () => 'system');

// Once auth is defined, make another Logger with auth
const appLogger = new Logger(LogSource.VxAdminService, () => getUserRole(auth, resolvedWorkspace))
```

I think the advantage would be avoiding the confusion of having to understand the difference between `BaseLogger` and `Logger` when reading this file.

apps/admin/frontend/src/screens/settings_screen.tsx
```diff
@@ -13,14 +13,21 @@ import {
...
 import { LiveCheckButton } from '../components/live_check_button';
 
-export function SettingsScreen(): JSX.Element {
-  const { auth, logger, usbDriveStatus } = useContext(AppContext);
+export interface SettingsScreenProps {
+  logger?: BaseLogger;
+}
+
+export function SettingsScreen({
+  logger = new BaseLogger(LogSource.VxAdminFrontend, window.kiosk),
```
jonahkagan, line 28: Does this need to be a prop?

jonahkagan (approved):
This is awesome! Thanks for taking it on. Maybe worth a PSA once it's merged to both announce the change and help align everybody on moving logging to the backend.

libs/logging/src/logger.ts
jonahkagan: I'm also thinking that maybe we don't need to have the `BaseLogger.log` method available on `Logger`

libs/logging/src/logger.ts
adghayes: This is helpful in tests where we're passing in a mocked `BaseLogger` that is converted to a logger with `Logger.from`. I want the created `Logger` to still be mocked.

libs/logging/src/logger.ts
adghayes: I made it different because, to make this refactor easier, I have `Logger` extend `BaseLogger`, which allows a lot of tests to remain unchanged. But I can't change the signature of of a function in an extended class.

apps/admin/frontend/src/screens/settings_screen.tsx
adghayes: I think I make it a prop in a later commit? Or in my follow-up PR, it's totally removed. so however it is here, I'll leave it.

apps/admin/backend/src/server.ts
adghayes: `libs/auth` is sometimes logging with user roles, like for card programming, it's logging as `system_administrator` because it knows that's the only role that can program cards.

And I think that the `app` could have use cases where it wants to log as `system`, which is distinct from `unknown`, so I still see a need for the `Logger` to have two methods rather than just one `log` method.

jonahkagan: Gotcha. That all makes sense. I'm not suggesting we get rid of `BaseLogger` altogether, just not using it for this particular use case. What do you think of that idea?

adghayes: OK, I think I see what you mean. You mean, not starting with `BaseLogger`, but rather with a systemLogger so all the loggers you see are `Logger`? I like that idea. I might change `logAsCurrentUser` to `logAsCurrentRole` in that case. And in your example,  I would want to make `appLogger` with `Logger.from` for testing purposes. How does that sound?

jonahkagan: Ah, I didn't think about the testing needs for using `Logger.from`. A comment might help explain that pattern, since it's a little confusing.

adghayes: I'm still feeling fuzzy on what the best pattern here is so I'm going to leave it for now and re-evaluate once logging's pretty exclusively on the backend.

Issue #4576, opened 2024-02-06
Make "focus Next" a11y behavior for PAT only


kshen0:
See: https://votingworks.slack.com/archives/CEL6D3GAD/p1707243437605809?thread_ts=1707179092.052939&cid=CEL6D3GAD

PR #4625, opened 2024-02-27
[BMD] Restrict "next" button autofocus to PAT voting sessions only

kofi-q:
## Overview

Fixes https://github.com/votingworks/vxsuite/issues/4576

On BMD contest screens, we were previously automatically moving focus to the "next" button whenever the user makes the final available selection in the current contest -- this makes it easier for sip-and-puff voters to move to the next contest without having to scroll through the remaining contest options, but also prevented the screen reader from providing selection feedback for audio-only voters using the accessible button controller.

This limits that behaviour to only when a PAT device is connected to try to provide both PAT voters and audio-only-button-controller voters with an optimal experience.

### Before - Non-PAT Voter [ 🔊 ]:

https://github.com/votingworks/vxsuite/assets/264902/0bdbd308-b88b-4ee6-8bcd-016874353727

### After - Non-PAT Voter [ 🔊 ]:

https://github.com/votingworks/vxsuite/assets/264902/08f5699e-3084-464d-a0cf-127a304fa9df

## Testing Plan
- Updated existing tests, added PAT-specific test case
- Manual verification

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/lib/assistive_technology.test.tsx
+++ b/apps/mark-scan/frontend/src/lib/assistive_technology.test.tsx
@@ -145,7 +145,7 @@ it('auto-focuses "next" button on contest screen after voting', async () => {
   // Confirm first contest only has 1 seat
   expect(contest0.seats).toEqual(1);
 
-  // Test navigation by accessible controller keyboard event interface
+  // Test navigation by PAT input keyboard event interface
   userEvent.keyboard('1');
   expect(getActiveElement()).toHaveTextContent(contest0candidate0.name);
   userEvent.keyboard('1');
@@ -158,6 +158,6 @@ it('auto-focuses "next" button on contest screen after voting', async () => {
     selected: true,
   });
 
-  // Focus should have jumped to the "Next" button because we're using keyboard nav
+  // Focus should have jumped to the "Next" button because we're using PAT nav
   expect(await screen.findByRole('button', { name: 'Next' })).toHaveFocus();
 });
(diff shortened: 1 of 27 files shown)


apps/mark-scan/frontend/src/lib/assistive_technology.test.tsx
```diff
@@ -117,3 +113,51 @@ it('accessible controller handling works', async () => {
...
+  expect(getActiveElement()).toHaveTextContent(contest0candidate1.name);
+
+  // select candidate
+  userEvent.keyboard('2');
+  await screen.findByRole('option', {
+    name: new RegExp(contest0candidate1.name),
+    selected: true,
+  });
+
+  // Focus should have jumped to the "Next" button because we're using keyboard nav
```
kshen0, line 161:
```suggestion
  // Focus should have jumped to the "Next" button because we're using PAT nav
```

kshen0: nit: I know this is an old comment, but since we're now distinguishing the different types of keyboard nav, might be more accurate to specify the device

kshen0 (approved):
I also manually tested this with a PAT device on hardware and it works as intended

apps/mark-scan/frontend/src/lib/assistive_technology.test.tsx
kofi-q: ah, thanks for catching!

kshen0 (approved):

PR #4626, opened 2024-02-28
Model Election.date without time

### LINKED ISSUE #4611: Fix election date timezone bug in VxSuite
The election date in the ballot def CDF is just a date without a specific time. When VxSuite reads it in and configures for the election it convert the date as it it were midnight UTC time, and then when displaying the date in various components (Election Info footer, reports, etc.) it translates the UTC timestamp to the time that the prod machine is configured for (pacific, eastern, etc.). This causes the date to then get displayed as the day PRIOR to when it is set in the election definition. For example an election package with an election date of February 22nd will display in the UI as February 21.

jonahkagan:
## Overview

Fixes #4611 

Change `Election.date` to use a new `DateWithoutTime` class instead of a datetime string. Since the election date doesn't have a specific time that's meaningful (i.e. we only ever show the date itself), we shouldn't model it with a time. This helps prevent bugs related to timestamps being interpreted differently based on the timezone of the machine.

## Demo Video or Screenshot
N/A - no user facing change other than fixing the bug

## Testing Plan
- Updated automated tests
- Removed `TZ=UTC` from automated tests (see commit for details)
- Some basic manual testing of VxAdmin to ensure this actually fixes the bug and shows the correct date for a given election

## Checklist


jonahkagan:
@arsalansufi @carolinemodic @eventualbuddha @adghayes you all got auto tagged as code owners. Could one of you comment to claim this review so everybody else knows not to review it?

adghayes:
@jonahkagan I'll review tomorrow morning!

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -532,7 +521,7 @@ export function convertVxfElectionToCdfBallotDefinition(
     // we were to use the current date, it would cause changes every time we
     // hash the object. We want hashes to be based on the content of the
     // election, not the date generated.
-    GeneratedDate: dateTimeString(new Date(vxfElection.date)),
+    GeneratedDate: `${vxfElection.date.toISOString()}T00:00:00Z`,
```
adghayes, line 524: Is there a reason you're not using the `toMidnight` helper here, as you do below?

libs/fixtures/data/electionFamousNames2021/bmd-ballot-complete-p1.jpg
adghayes, line 1: @jonahkagan for this fixture, it looks like the previous fixture, with June 6th, is correct, but the new fixture, with June 5th, is incorrect. Why might that be?

libs/auth/certs/test/set-1/election-manager/card-vx-admin-private-key.pem
adghayes, line 1: Why did these have to be regenerated? I didn't know these certs were related to the election definition.

apps/admin/backend/package.json
adghayes, line 1: Great rationale, love it!

adghayes (commented):
There's one snapshot that doesn't look right to me but other than that, looks great.

After doing a repo-wide data model change I'm curious what you think the highest yield change is that we need to make to make this sort of thing easier (e.g. test assertions, fixture generation...)

adghayes:
@jonahkagan I forgot to review this PR through the lens of compatibility. This PR fixes the issue across `vxsuite`. But since NH machines have frozen code, will this introduce file incompatibility between `main` VxDesign and NH machines?

jonahkagan:
> But since NH machines have frozen code, will this introduce file incompatibility between main VxDesign and NH machines?
The v3.1 machines in the field are targeted by VxDesign v3.1, which is on its own release branch. `main` is now targeting 4.0.

libs/fixtures/data/electionFamousNames2021/bmd-ballot-complete-p1.jpg
jonahkagan: Nice catch. I think I maybe recreated this ballot fixture before I totally fixed the error. Then got distracted over the weekend and forgot I needed to do it again. Thanks for catching!

libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan: Yeah, in this case I want it to be the same no matter what the system timezone is so it's stable for tests. This date doesn't actually matter for anything, it's just a required field in CDF.

jonahkagan:
> After doing a repo-wide data model change I'm curious what you think the highest yield change is that we need to make to make this sort of thing easier (e.g. test assertions, fixture generation...)

Good question. It was actually pretty easy already since I was able to work through type-checking and test failures one-by-one. The most annoying part was regenerating the BMD ballot image fixture by hand.

adghayes (approved):

eventualbuddha (approved):

libs/fixtures/data/electionFamousNames2021/bmd-ballot-complete-p1.jpg
jonahkagan: 1e2e95482

apps/design/frontend/src/elections_screen.tsx
```diff
@@ -97,9 +98,11 @@ export function ElectionsScreen(): JSX.Element | null {
                     <td>{election.title || 'Untitled Election'}</td>
                     <td>
                       {election.date &&
-                        new Date(election.date).toLocaleDateString(undefined, {
-                          timeZone: 'UTC',
-                        })}
+                        format.localeLongDate(
+                          new Date(
```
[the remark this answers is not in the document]
jonahkagan: nice catch!

PR #4629, opened 2024-02-28
fix: VxCentralScan Blank Ballot Detection

adghayes:
Some tweaks to blank ballot handling, follow up to #3407.

### DIFF
--- a/apps/central-scan/backend/src/store.ts
+++ b/apps/central-scan/backend/src/store.ts
@@ -599,13 +599,10 @@ export class Store {
     [front, back]: SheetOf<PageInterpretationWithFiles>
   ): string {
     try {
-      const finishedAdjudicationAt =
-        front.interpretation.type === 'InterpretedHmpbPage' &&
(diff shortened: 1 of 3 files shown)


apps/central-scan/backend/src/store.ts
```diff
@@ -599,13 +599,10 @@ export class Store {
     [front, back]: SheetOf<PageInterpretationWithFiles>
   ): string {
     try {
-      const finishedAdjudicationAt =
-        front.interpretation.type === 'InterpretedHmpbPage' &&
-        back.interpretation.type === 'InterpretedHmpbPage' &&
-        !front.interpretation.adjudicationInfo.requiresAdjudication &&
-        !back.interpretation.adjudicationInfo.requiresAdjudication
-          ? DateTime.now().toISOTime()
-          : undefined;
```
adghayes, line 608: Because the finished date was using different logic than the flag, it could lead to situations in which the ballot was immediately marked as needing adjudication and adjudicated.

eventualbuddha (approved):
🎉🧹

PR #4630, opened 2024-02-28
VxDesign 3.1: Tweaks from ballot creation

jonahkagan:
## Overview

I was keeping this customizations on a separate branch, but I realized it's too confusing and error-prone, so I want to merge them into the release branch.

The changes are:
- Hackily including Loudon's zoning ballot measure image
- Addressing the off-by-one election date bug by changing the date shown on the ballot to also be off-by-one

## Checklist


jonahkagan:
Given that the off-by-one date error is in the v3.1 VxSuite release, I think it makes sense to merge the corresponding change to VxDesign, since we're locked into that workaround.

The ballot image change I feel less strongly about, but was mainly thinking that if anybody besides me wants to generate ballots, it would be less confusing if this code was already merged to the release branch.

I could leave a comment explaining this thinking, but I don't imagine reverting. Does that seem reasonable to you?

As for future support, it's a good question. I'm hoping the new ballot rendering stack can support embedded images more easily, so maybe we'll just get it for cheap.

eventualbuddha:
Ah, I missed the little `UTC`-related change in there. That part makes sense, for sure. I ask about reverting because I assumed we wouldn't want to clutter things up with more and more images if this does become a recurring thing. If we wanted to generate historical ballots we'd use the corresponding version of the software, maybe?

jonahkagan:
Sure, I think it's reasonable to revert and then go back to a historical commit if needed. I don't think the cost is super high to leaving the image hack in right now, but maybe a compromise is to revert it the next time we need to change that code. I can leave a comment to that effect.

PR #4631, opened 2024-02-28
logging: migrate some from frontend to backend

adghayes:
## Overview

Migrate more logging from frontend to backend. Adds some logging in VxScan.

### DIFF
--- a/apps/mark-scan/backend/src/app.ts
+++ b/apps/mark-scan/backend/src/app.ts
@@ -281,6 +281,7 @@ export function buildApi(
 
     ...createSystemCallApi({
       usbDrive,
+      logger,
       machineId: getMachineConfig().machineId,
     }),
 
--- a/apps/mark-scan/backend/src/app.ts
+++ b/apps/mark-scan/backend/src/app.ts
@@ -313,7 +313,7 @@ export function buildApi(
         }
       })();
 
-      await logger.logAsCurrentUser(logEvent, { disposition: 'success' });
+      await logger.logAsCurrentRole(logEvent, { disposition: 'success' });
     },
 
     setTestMode(input: { isTestMode: boolean }) {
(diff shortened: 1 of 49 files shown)


apps/scan/backend/src/util/logging.ts
adghayes, line 1: Previously we weren't doing these logs at all in `scan`, so adding this helper modeled off of the others.

apps/mark-scan/backend/src/app.ts
```diff
@@ -321,11 +322,20 @@ export function buildApi(
...
+      assert(electionDefinition);
       store.setPrecinctSelection(input.precinctSelection);
       store.setBallotsPrintedCount(0);
+      await logger.logAsCurrentUser(LogEventId.PrecinctConfigurationChanged, {
+        disposition: 'success',
+        message: `User set the precinct for the machine to ${getPrecinctSelectionName(
+          electionDefinition.election.precincts,
+          input.precinctSelection
+        )}`,
+      });
```
adghayes: agnostic as to doing it in this PR, but little logging bits like this could be broken out into logging helpers in `utils` to make it a little less verbose in app code.

jonahkagan: I think we should add an API middleware that logs every request, its inputs, and whether it succeeded or failed. That would cut out a lot of this verbosity.

apps/scan/backend/src/scanners/custom/app_scan_precinct_config.test.ts
```diff
@@ -89,62 +88,3 @@ test('bmd ballot is accepted if precinct is set for the right precinct', async (
     }
   );
 });
-
-test('hmpb ballot is rejected when scanned for wrong precinct', async () => {
```
jonahkagan, line 93: Isn't this test still important because it tests that the precinct configuration is correctly applied during interp?

adghayes: Isn't that covered by the test above which checks that a BMD ballot of the wrong precinct is rejected, since that confirms precinct state is being passed to interpreter?

jonahkagan: Oh, didn't see that one. That does make me feel better about it. I still think it's useful to not rely on implementation details for this test (which tests an important property to enforce for HMPB), but I don't feel so strongly to block.

jonahkagan (approved):

PR #4636, opened 2024-02-29
refactor / logging: VxCentralScan cleanup + `grout` transition Part 1 of 2

adghayes:
## Overview

Navigating the state logic on the `central-scan` frontend is a little thorny, some state should really be in the backend. The backend has a lot of no-op code. Here's a substantial refactor. Key points:
- moves scan logging from frontend to backend
- move scan endpoints to `grout`
- cleans up backend logic and testing
- fixes a bug found along the way

## Checklist


apps/central-scan/frontend/src/app_root.tsx
```diff
@@ -80,48 +80,6 @@ export function AppRoot({
...
-        // Scanning a sheet failed and needs adjudication.
-        void logger.log(LogEventId.ScanSheetComplete, userRole, {
-          disposition: 'failure',
-          message: 'Sheet rejected while scanning.',
-          result:
-            'Sheet not tabulated, user asked to intervene in order to proceed.',
-          batchId: currentScanningBatchId,
-          sheetCount: currentBatch.count,
-        });
-        return;
```
adghayes, line 108: I opted to just remove this logging because it is always followed by an "AdjudicationInfo" log. The problem is not really that the scan failed, but that the scan went to adjudication.

apps/central-scan/backend/src/fujitsu_scanner.ts
```diff
@@ -15,9 +15,6 @@ const debug = makeDebug('scan:scanner');
 
 export interface BatchControl {
   scanSheet(): Promise<SheetOf<string> | undefined>;
-  acceptSheet(): Promise<boolean>;
-  reviewSheet(): Promise<boolean>;
-  rejectSheet(): Promise<boolean>;
```
adghayes, line 20: If there's a reason to keep these around that I don't know about, please do tell, but it seemed like leftover code to me.

apps/central-scan/backend/src/sheet_requires_adjudication.ts
jonahkagan, line 1: Is there a regression test for this fix?

adghayes: Nope, I chose not to since it was a side quest of a side quest of a side quest. But your callout is well-taken, if there's any time to write a test for this, it is now.

eventualbuddha (approved):
Thanks for making this change! Loving the cleanups. I tested it with the NH Test Election (7/12) and it works great.

adghayes:
@eventualbuddha thank you for reviewing and for even manual testing!

Issue #4618, opened 2024-02-26
vxmarkscan: a11y controls (pause, help, rate) do nothing


mattroe:
https://votingworks.slack.com/archives/CEL6D3GAD/p1707179141268099?thread_ts=1707179092.052939&cid=CEL6D3GAD

kshen0:
Tested rate up/down and they work as expected. Not sure what was happening during initial QA. Pause and Help covered in https://github.com/votingworks/vxsuite/pull/4646

PR #4646, opened 2024-03-04
Send Shift keypress event when sending alphabet key codes

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/4618

Sends a `Shift` key event along with `r` and `p` keypresses to correctly trigger keypress handlers in the apps. These keypress events map to `Pause` and `Replay` functionality.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/e7a3bffd-343f-4d60-8bb4-adf653c96b05

## Testing Plan
- updated test for `handle_command` in controller daemon

## Checklist

### DIFF
--- a/apps/mark-scan/accessible-controller/src/controllerd.rs
+++ b/apps/mark-scan/accessible-controller/src/controllerd.rs
@@ -18,7 +18,7 @@ use std::{
     time::{Duration, Instant},
 };
 use uinput::{
-    event::{keyboard, Keyboard, Kind},
+    event::{keyboard, Keyboard},
     Device,
 };
 use vx_logging::{log, set_app_name, Disposition, EventId, EventType};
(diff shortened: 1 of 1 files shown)


apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -476,6 +487,26 @@ mod tests {
...
+    #[test]
+    fn test_handle_command_success_no_shift() {
+        let data = [
+            0x30,
+            0x00,
+            0x02,
+            Button::Left as u8,
+            Action::Pressed as u8,
+            0xd8,
+            0x09,
```
kshen0, line 505: The last 2 values are the [CRC checksum](https://en.wikipedia.org/wiki/Cyclic_redundancy_check); the important difference in this test is that we are sending `Button::Left`, which as an arrow key requires no `Shift` key.

kofi-q (approved):
🎉

Issue #3974, opened 2023-09-14
Pulling poll worker card during paper load state, before paper is loaded, should end the session


kshen0:
If the poll worker card is removed during the paper load state, the frontend advances to the "Begin Voting" screen. 

Ignore the note below. Now that we poll for auth in the state machine this is a fast fix.

--

~It's a little tricky to handle this in the state machine because the state machine doesn't subscribe to auth changes. We can handle in the frontend by checking for the `accepting_paper` state in `app_root` where we render the `Ballot` like~

```
if (stateMachineState === 'accepting_paper') {
  return <NoPaperLoadedPage />;
}

return <BallotContext.Provider> ...
</BallotContextProvider>
```

PR #4647, opened 2024-03-04
Reset session if pollworker card is pulled during paper load prompt/intake

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/3974

During either paper loading state, resets the voter session if poll worker auth is ended before paper is done loading.

* If card is pulled before paper load is attempted, the app just returns to the login page
* If card is pulled during physical paper load, the app displays a message for 5 seconds then resets

Previously, the code allowed a poll worker to pull their card either
1. before paper loading was attempted, resulting in a cardless voter session that would eventually fail at print time
2. in the middle of paper loading. In the event of a jam during paper load, the poll worker would be unaware, so we want the poll worker to remain until loading is successful.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/eb0aee15-e98d-4339-952b-632f89f18a38

## Testing Plan

- [x] add backend tests
- [x] add frontend tests

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (logging is already covered by existing auth and state transition logging)

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -464,7 +464,7 @@ export function buildMachine(
               on: {
                 PAPER_READY_TO_LOAD: 'loading_paper',
                 PAPER_PARKED: 'waiting_for_ballot_data',
-                AUTH_STATUS_CARDLESS_VOTER: 'not_accepting_paper',
+                AUTH_STATUS_CARDLESS_VOTER: 'resetting_state_machine_no_delay',
               },
             },
             loading_paper: {
@@ -726,6 +726,10 @@ export function buildMachine(
                 },
               },
             },
+            resetting_state_machine_no_delay: {
+              entry: ['resetContext', 'endCardlessVoterAuth'],
+              always: 'not_accepting_paper',
+            },
             resetting_state_machine_after_jam: {
               entry: ['resetContext', 'endCardlessVoterAuth'],
               after: {
@@ -947,6 +951,8 @@ export async function getPaperHandlerStateMachine({
           return 'jammed';
         case state.matches('voting_flow.jam_physically_cleared'):
           return 'jam_cleared';
+        case state.matches('voting_flow.resetting_state_machine_no_delay'):
+          return 'resetting_state_machine_no_delay';
         case state.matches('voting_flow.resetting_state_machine_after_jam'):
           return 'resetting_state_machine_after_jam';
         case state.matches('voting_flow.ballot_accepted'):
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -464,8 +464,7 @@ export function buildMachine(
               on: {
                 PAPER_READY_TO_LOAD: 'loading_paper',
                 PAPER_PARKED: 'waiting_for_ballot_data',
-                AUTH_STATUS_CARDLESS_VOTER:
-                  'poll_worker_auth_ended_unexpectedly',
+                AUTH_STATUS_CARDLESS_VOTER: 'not_accepting_paper',
               },
             },
             loading_paper: {
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -925,6 +925,9 @@ export async function getPaperHandlerStateMachine({
 
       switch (true) {
         case state.matches('voting_flow.not_accepting_paper'):
+        case state.matches('voting_flow.resetting_state_machine_no_delay'):
+          // Frontend has nothing to render for resetting_state_machine_no_delay
+          // so to avoid flicker we just return the state it's guaranteed to transition to
           return 'not_accepting_paper';
         case state.matches('voting_flow.accepting_paper'):
           return 'accepting_paper';
@@ -956,9 +959,6 @@ export async function getPaperHandlerStateMachine({
           return 'jammed';
         case state.matches('voting_flow.jam_physically_cleared'):
           return 'jam_cleared';
-        case state.matches('voting_flow.resetting_state_machine_no_delay'):
-          /* istanbul ignore next - nonblocking state can't be reliably asserted on. Assert on business logic eg. jest mock function calls instead */
-          return 'resetting_state_machine_no_delay';
         case state.matches('voting_flow.resetting_state_machine_after_jam'):
           return 'resetting_state_machine_after_jam';
         case state.matches('voting_flow.ballot_accepted'):
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -697,9 +697,7 @@ export function buildMachine(
             },
             eject_to_front: {
               invoke: pollPaperStatus(),
-              entry: async (context) => {
-                await context.driver.ejectPaperToFront();
-              },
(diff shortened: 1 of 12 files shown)


apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -43,8 +43,7 @@ import {
   DEVICE_STATUS_POLLING_INTERVAL_MS,
   DEVICE_STATUS_POLLING_TIMEOUT_MS,
   MAX_BALLOT_BOX_CAPACITY,
-  RESET_AFTER_JAM_DELAY_MS,
-  SUCCESS_NOTIFICATION_DURATION_MS,
+  NOTIFICATION_DURATION_MS,
```
kshen0, line 46: This file's diff is mostly whitespace; recommend turning whitespace diff off to view

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -439,6 +440,18 @@ export function AppRoot({
     return <JamClearedPage stateMachineState={stateMachineState} />;
   }
 
+  const isInPaperLoadingStage =
+    stateMachineState === 'accepting_paper' ||
+    stateMachineState === 'loading_paper';
```
jonahkagan, line 445: In the second video, it seemed a little weird to me that the "start voting" screen is shown while the paper is still being loaded and then if the poll worker removes their card, this error can occur which makes them have to start over. I feel like the "start voting" screen indicates that it's safe to remove the poll worker card. Maybe a "loading paper... please wait" screen would be helpful?

Going to hold off on technical review for now til we resolve the product discussion.

jonahkagan: Cool, just let me know when you want me to take another look

kshen0: Uploaded the new video and re-requested review 🙏

jonahkagan: One more UX question: Is there a reason to have a delay and show a message when the poll worker card is removed before loading paper? Kind of seems like it could just cancel and go back to the Insert Card screen immediately

kshen0: No strong reason - I kept them the same for consistency but we can also just end the session. The more I think about it, it's probably less confusing if we skip the message.

To be clear, are you ok with keeping the message for when the poll worker card is pulled during the physical paper load stage?

jonahkagan: I'm ok with keeping it, though I'd be curious to see how it feels to you without it.

kshen0: I think in that case, it'll be more surprising to the poll worker if we end the session. IMO we should leave the message in for the actual paper loading stage, or remove both the message and auth check (so a poll worker can pull their card once the paper has been grasped by the printer).

Of those 2 options, the former is safer, so inclined to leave it as is.

jonahkagan: Sounds good to me!

kshen0:
Must have accidentally requested a review early - not sure how. @jonahkagan this is now ready for review

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -403,383 +402,396 @@ export function buildMachine(
...
-              await auth.endCardlessVoterSession(
-                constructAuthMachineState(context.workspace)
-              );
-
-              assign({
-                interpretation: undefined,
-                scannedImagePaths: undefined,
-                isPatDeviceConnected: false,
-              });
+            poll_worker_auth_ended_unexpectedly: {
```
jonahkagan, line 751: What happens if the poll worker reinserts their card during this state or during the `not_accepting_paper` state that comes after?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -403,383 +402,396 @@ export function buildMachine(
...
+              entry: ['resetContext', 'endCardlessVoterAuth'],
+              after: {
+                // The frontend needs time to idle in this state so the user can read the status message
+                [NOTIFICATION_DURATION_MS]: 'not_accepting_paper',
```
jonahkagan, line 755: Watching the video, it feels like the paper should be ejected while this message is showing, rather than after. Though I think it's totally fine to leave as is since this is likely an uncommon case.

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -545,6 +561,9 @@ export function AppRoot({ reload }: Props): JSX.Element | null {
       }
     }
 
+    // Implicitly map resetting_state_machine_no_delay to return InsertCardScreen
+    // because that state implies we are returning to the starting (default) state.
```
jonahkagan, line 565: Could this be made explicit (e.g. with an `assert`) or by masking `resetting_state_machine_no_delay` on the backend by returning a state that would go to the insert card screen?

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -513,7 +523,13 @@ export function AppRoot({ reload }: Props): JSX.Element | null {
...
 
     if (pollsState === 'polls_open') {
-      if (isCardlessVoterAuth(authStatus)) {
+      if (
+        isCardlessVoterAuth(authStatus) &&
+        // accepting_paper expects poll worker auth. If the frontend sees accepting_paper but has cardless voter auth,
+        // it means the state hasn't caught up to auth changes. We check that edge case here to avoid flicker ie.
+        // rendering the ballot briefly before rendering the correct "Insert Card" screen
+        stateMachineState !== 'accepting_paper'
+      ) {
```
jonahkagan, line 532: I feel like we've run into these cases before where the backend and frontend have out of sync state. It's pretty confusing. Do you think we could make it more straightforward if the backend was solely responsible for deciding what state we should show and the frontend only rendered screens based on the state machine state (rather than also trying to incorporate auth state)?

kshen0: Yeah I didn't foresee this issue when adding auth polling to the backend. I agree it's pretty awful. In my mind the fix is related to https://github.com/votingworks/vxsuite/issues/3985, which would move the frontend to an exhaustive `switch` to handle all state machine states (and _only_ states, no auth).

cc @arsalansufi who had a question about #3985 - for all states that involve auth, we'll need to patch like this.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Good point. It's a quick change so not a problem to add even if it's uncommon

[8b2a91a](https://github.com/votingworks/vxsuite/pull/4647/commits/8b2a91a33d4702ac7ef4fa9f2723a8aab9f9376d)

apps/mark-scan/frontend/src/app_root.tsx
kshen0: [80b0193](https://github.com/votingworks/vxsuite/pull/4647/commits/80b0193548f15f61b0685553efcd675edfe4155b)

jonahkagan (approved):
Thanks for working through the changes!

apps/mark-scan/frontend/src/app_root.tsx
arsalansufi: Thanks for the ping / example! Does make the motivation for https://github.com/votingworks/vxsuite/issues/3985 clear

PR #4650, opened 2024-03-06
interpreter blank paper diagnostic

adghayes:
## Overview

Introduces a blank paper diagnostic to use on scanned images. This PR does _not_ wire it through to the app and UI. 

The goal of this diagnostic is:

- it passes when the image shows a blank sheet of paper
- it fails when the image shows streaking

There's a tricky balance to strike here between false positives and streak detection. If we turn the dials too far, we might fail the diagnostic too often. If we don't turn the dials far enough, we never catch streaking issues. We don't have a large set of training images from actual usage, we don't even have a small set. For lack of that, I lean toward making the test less sensitive.

I look for streaks / marks by breaking up the page into a grid with bubble-sized cells and seeing if significant markings are detected in any cells. I crop out 20 pixels on all edges to avoid the black edges that can appear in scans. This seems to be plenty to prevent false positives at the edge for the Fujitsu, but unclear how this will play out if we eventually apply this to the precinct scanner, where skew is worse. 

Blank:
e5c7b476-7d39-4716-9f75-dfb2f1c35ea0-front_debug_diagnostic

Streaked - see how there are more streaks we could catch if we wanted to be even more particular, but currently we only want to catch the obvious:
0dc29646-3c6a-4abd-9d2d-ae1b03a3b4ad-front_debug_diagnostic

Not Blank:
all-bubble-side-a_debug_diagnostic


libs/ballot-interpreter/src/hmpb-rust/diagnostic.rs
```diff
@@ -0,0 +1,182 @@
...
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
+use types_rs::geometry::Rect;
+
+use crate::{
+    ballot_card::load_ballot_scan_bubble_image,
+    debug::{draw_diagnostic_cells, ImageDebugWriter},
+    image_utils::{ratio, BLACK},
+};
+
+const FAIL_SCORE: f32 = 0.05;
```
adghayes, line 15: Note that you can bring this down to 0.03 and all tests will pass, but I wanted to give it some leeway due to unpredictability of field data.

libs/ballot-interpreter/src/hmpb-rust/diagnostic.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    cells
+}
+
+fn inspect_cells(
+    img: &GrayImage,
+    cells: &[Rect],
+    foreground_threshold: u8,
+) -> (Vec<Rect>, Vec<Rect>) {
+    let (failed_cells, passed_cells) = cells.par_iter().partition(|cell| {
```
eventualbuddha, line 58: 🎉 glad to see the use of `par_iter` here to do the work in parallel!

eventualbuddha (approved):
With #4652, this looks great! I'm glad to see more Rust where it's appropriate.

benadida:
high-level question – are we expecting to have a feature where we ask them to scan a blank sheet to test?

Ideally, I'd love to detect streaking on a normal ballot scan. I realize that's harder.

Just wondering about our overall plan.

adghayes:
We've discussed detecting streaking on normal scanning (ballots, during polls open) but haven't prioritized or fully defined that. 

If I'm understanding your question correctly, that is merged as #4653 for VxCentralScan. Planned but not implemented yet for VxScan.

adghayes:
And @benadida to clarify - these are currently just planned as diagnostics behind System Administrator auth. If you're suggesting doing it as part of a configuration flow, I think that's a potentially good idea (we're planning on doing something similar by asking people to print a test report on the printer) but we haven't narrowed in on it yet.

benadida:
👍 gotcha. For diagnostics. Makes sense.

My ideal world, which could be hard, is that as we're scanning ballots, the VxScan says "wow, we've got some streaking, ask a poll worker for help." And then poll worker does cleaning sheet, and then we're done and life is beautiful, angels sing, and world peace is within our grasp.

eventualbuddha:
> My ideal world, which could be hard, is that as we're scanning ballots, the VxScan says "wow, we've got some streaking, ask a poll worker for help."

While detecting _all_ instances of this is probably hard, there are versions of it that I don't think would be difficult or cost too much time to implement. Just check for "at least N consecutive columns of entirely black pixels surrounded by at least one column not entirely black on each side", where N is probably 5-10?

PR #4654, opened 2024-03-07
we forgot to make this tweak in the env file when we changed var names

adghayes:


kshen0 (approved):
Sorry about that

adghayes:
np, I forgot too and the file should really be linted

PR #4657, opened 2024-03-07
[i18n/audio] Generate a unique ballot style definitions/IDs for each language

kofi-q:
## Overview

Adding language codes to ballot styles generated in VxDesign to be matched up with BMD language selections (and eventually HMPB content as well).

Multi-language ballot style generation is currently hidden behind the `REACT_APP_VX_ENABLE_CLOUD_TRANSLATION_AND_SPEECH_SYNTHESIS` flag, since it needs to hand-in-hand with multi-language ballot content generation.

### TODO:
- Synthesised audio for the ballot style IDs sounds a little awkward and isn't useful for audio-only users in its current form. We may need to add a custom election string to use as alt audio text for this.

### VxDesign:

Screenshot 2024-03-07 at 14 29 50

### BMD:

https://github.com/votingworks/vxsuite/assets/264902/5f8aca41-fa67-448e-9c0f-fdf5980472a5

## Testing Plan
- Updated ballot generation  test cases and related tests.
- Manually verified list of ballot styles in VxDesign screens and ballot exports.
- Manually verified BMD integration with voter language switcher.

## Checklist

### DIFF
--- a/apps/design/backend/src/store.ts
+++ b/apps/design/backend/src/store.ts
@@ -3,34 +3,49 @@ import {
   Optional,
   assert,
   find,
-  groupBy,
   iter,
-  throwIllegalValue,
-  unique,
 } from '@votingworks/basics';
 import { Client as DbClient } from '@votingworks/db';
 import {
   DEFAULT_LAYOUT_OPTIONS,
   LayoutOptions,
-  NhCustomContent,
   NhCustomContentByBallotStyle,
 } from '@votingworks/hmpb-layout';
 import {
   Id,
   Iso8601Timestamp,
   Election,
-  DistrictId,
-  PrecinctId,
-  BallotStyleId,
   DEFAULT_SYSTEM_SETTINGS,
   SystemSettings,
   safeParseSystemSettings,
-  CandidateContest,
-  PartyId,
   LanguageCode,
 } from '@votingworks/types';
 import { join } from 'path';
 import { v4 as uuid } from 'uuid';
+import {
+  BooleanEnvironmentVariableName,
+  isFeatureFlagEnabled,
+} from '@votingworks/utils';
+import {
+  BallotLanguageConfig,
+  BallotLanguageConfigs,
+  BallotStyle,
+  Precinct,
+  convertToVxfBallotStyle,
+  hasSplits,
+} from './types';
+import { generateBallotStyles } from './ballot_styles';
+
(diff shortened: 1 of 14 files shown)


apps/design/backend/src/store.ts
```diff
@@ -3,34 +3,49 @@ import {
...
+} from './types';
+import { generateBallotStyles } from './ballot_styles';
+
+export function getTempBallotLanguageConfigsForCert(): BallotLanguageConfigs {
+  const translationsEnabled = isFeatureFlagEnabled(
+    BooleanEnvironmentVariableName.ENABLE_CLOUD_TRANSLATION_AND_SPEECH_SYNTHESIS
+  );
+  return translationsEnabled
+    ? Object.values(LanguageCode).map(
+        (l): BallotLanguageConfig => ({ languages: [l] })
```
arsalansufi, line 45: If I'm remembering right, this is an array because, for multi-language HMPBs, we'll have two languages on the ballot at once, e.g., English and Spanish.

These single-language configs are then for the BMD, is that correct?

Once we build support for multi-language HMPBs, will the ballot language configs include both `[English, Spanish]` (HMPB) and `[Spanish]` (BMD)?

All seems fine! Just checking my understanding

arsalansufi (approved):
LGTM!

apps/design/backend/src/store.ts
kofi-q: That's right, this is allowing for both single-language and multi-language HMPBs - the current assumption is that only the BMD would use non-English single-language ballot styles, but I could see a world where we move to only specifying one language in the ballot style definition and always treating that as multi-language (English + specified language) when generating both the HMPBs and BMDPBs.

apps/design/backend/src/types.ts
```diff
@@ -0,0 +1,81 @@
+// We create new types for precincts that can be split, since the existing
+// election types don't support this. We will likely want to extend the existing
+// types to support it in the future, but doing it separately for now allows us
+// to experiment and learn more first. We'll store these separately in the
```
jonahkagan, line 4: Comment got cut off here

kofi-q: Oh, weird - thanks! Must've gotten broken up by the auto-imports  - will fix up

apps/design/backend/src/store.ts
arsalansufi: Perfect, thanks for confirming!

PR #4661, opened 2024-03-08
refactor(mark-scan): get battery info from backend, not `kiosk-browser`

adghayes:
## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/app_root.tsx
+++ b/apps/mark-scan/frontend/src/app_root.tsx
@@ -13,7 +13,6 @@ import {
 import { useHistory } from 'react-router-dom';
 import { IdleTimerProvider } from 'react-idle-timer';
 import {
-  Hardware,
   isElectionManagerAuth,
   isCardlessVoterAuth,
   isPollWorkerAuth,
@@ -22,12 +21,8 @@ import {
   isFeatureFlagEnabled,
   BooleanEnvironmentVariableName,
 } from '@votingworks/utils';
-
-import { BaseLogger } from '@votingworks/logging';
-
 import {
   SetupCardReaderPage,
-  useDevices,
   UnlockMachineScreen,
   VoterSettingsManagerContext,
   useAudioControls,
@@ -54,6 +49,7 @@ import {
   startCardlessVoterSession,
   updateCardlessVoterBallotStyle,
   unconfigureMachine,
+  systemCallApi,
 } from './api';
 
 import { Ballot } from './components/ballot';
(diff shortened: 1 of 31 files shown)


apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -564,7 +553,7 @@ export function AppRoot({
         <InsertCardScreen
           appPrecinct={precinctSelection}
           electionDefinition={optionalElectionDefinition}
-          showNoChargerAttachedWarning={!computer.batteryIsCharging}
+          showNoChargerAttachedWarning={!!battery && battery.discharging}
```
kshen0, line 556: Is battery every not defined if the query succeeds? Should we `assertDefined(battery)` after the check for `batteryQuery.isSuccess`?

apps/mark-scan/frontend/src/pages/diagnostics_screen.test.tsx
```diff
@@ -86,120 +90,6 @@ describe('System Diagnostics screen: Computer section', () => {
   });
 });
 
-describe('System Diagnostics screen: Printer section', () => {
```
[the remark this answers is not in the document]
kshen0: Thanks!

apps/mark-scan/frontend/src/pages/diagnostics_screen.test.tsx
```diff
@@ -245,22 +135,6 @@ describe('System Diagnostics screen: Accessible Controller section', () => {
     unmount();
   });
 
-  it('shows when the controller is disconnected', () => {
```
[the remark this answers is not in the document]
kshen0: Controller could be disconnected internally or the daemon could have crashed, but it's fine to remove this for now.

kshen0 (commented):
What battery level is the backend reading in the context of mark-scan? Is it the UPS?

apps/mark-scan/frontend/src/app_root.tsx
adghayes: It's possible that battery is not defined. In fact, to your main question, it's probably not defined for VxMarkScan. We check battery status by polling the file at `/sys/class/power_supply/BAT0/uevent` or the equivalent with `BAT1`. That file may not exist for a number of reasons. 

I didn't think about the particulars of the VxMarkScan hardware. I'm curious what you'd find if you checked for those files, but I'm guessing they don't exist. Does the VxMarkScan have an onboard UPS? Or are we planning on using an external one? In either case, we probably don't have access to battery level.

If you unplug the VxMarkScan, does it just turn off?

kshen0: Summarizing offline discussion - there is no onboard battery. Power requirements are in flux right now but we might want a data connection from a UPS, so we’re going to leave this in to avoid unnecessary work of removing/adding later.

kshen0 (approved):

Issue #3762, opened 2023-07-27
Prepare a script to generate a production sys admin card for a given jurisdiction


carolinemodic:
this one's for @arsalan (not now, no rush) – we should have a script that is roughly "generate sysadmin card for a given jurisdiction" so we don't have to have a vxadmin machine to do it. (in prod)

Issue automatically created from slack feedback reported by ben 
https://votingworks.slack.com/archives/CJU9MSC6S/p1690340303852449

PR #4662, opened 2024-03-08
Add script for programming a prod sys admin Java Card without a VxAdmin

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3762

This PR adds a script for programming a prod sys admin Java Card without a VxAdmin. This is useful for preparing backup
cards after we've shipped a VxAdmin.

This is possible because the VxAdmin cert on a card does not have to exactly match the cert on the VxAdmin; it just needs to have the same jurisdiction (and be signed by VotingWorks, of course).

## Testing Plan

- [x] Tested manually

### DIFF
--- a/libs/auth/scripts/program_production_system_administrator_java_card_without_vx_admin.ts
+++ b/libs/auth/scripts/program_production_system_administrator_java_card_without_vx_admin.ts
@@ -1,0 +1,83 @@
+import fs from 'fs/promises';
+import path from 'path';
+import { DirResult, dirSync } from 'tmp';
+import { extractErrorMessage } from '@votingworks/basics';
+
+import { PROD_VX_CERT_AUTHORITY_CERT_PATH } from '../src';
+import { CERT_EXPIRY_IN_DAYS, constructMachineCertSubject } from '../src/certs';
+import { createCert } from '../src/cryptography';
+import { getRequiredEnvVar } from '../src/env_vars';
+import { JavaCard } from '../src/java_card';
+import {
+  generatePrivateKey,
+  programSystemAdministratorJavaCard,
+} from './utils';
+
+const jurisdiction = getRequiredEnvVar('VX_MACHINE_JURISDICTION');
+const vxPrivateKeyPath = getRequiredEnvVar('VX_PRIVATE_KEY_PATH');
+
+async function instantiateJavaCardWithOneOffVxAdminPrivateKeyAndCertAuthorityCert(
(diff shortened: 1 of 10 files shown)


libs/auth/scripts/program_production_system_administrator_java_card_without_vx_admin.ts
```diff
@@ -0,0 +1,83 @@
...
+const vxPrivateKeyPath = getRequiredEnvVar('VX_PRIVATE_KEY_PATH');
+
+async function instantiateJavaCardWithOneOffVxAdminPrivateKeyAndCertAuthorityCert(
+  tempDirectoryPath: string
+): Promise<JavaCard> {
+  const vxAdminPrivateKey = await generatePrivateKey();
+  const vxAdminPrivateKeyPath = path.join(
+    tempDirectoryPath,
+    'vx-admin-private-key.pem'
+  );
```
benadida, line 26: can this be explicitly deleted rather than rely on tmp cleanup later?

libs/auth/scripts/program_production_system_administrator_java_card_without_vx_admin.ts
```diff
@@ -0,0 +1,83 @@
...
+    await programSystemAdministratorJavaCard({
+      card,
+      jurisdiction,
+      isProduction: true,
+    });
+  } catch (error) {
+    console.error(`❌ ${extractErrorMessage(error)}`);
+    exitCode = 1;
+  } finally {
+    tempDirectory?.removeCallback();
```
benadida, line 79: is this doing the removal of tmp files?

benadida (approved):

libs/auth/scripts/program_production_system_administrator_java_card_without_vx_admin.ts
arsalansufi: Ah yup, so we are explicitly removing, in the `finally` block

libs/auth/scripts/program_production_system_administrator_java_card_without_vx_admin.ts
arsalansufi: Yes exactly. And because we use `{ unsafeCleanup: true }` when creating the directory, we ensure that the files in it are deleted, too. Without that flag, `removeCallback` fails if the directory still contains files in it

Added a comment for clarity!

Issue #4641, opened 2024-03-04
Update VxDesign to support NH state elections with existing ballot design


arsalansufi:
https://votingworks.slack.com/archives/C02Q3M874SV/p1709153283534609

PR #4664, opened 2024-03-08
image-utils: Add toImageBuffer

jonahkagan:
## Overview

Adds `toImageBuffer` to convert `ImageData` to a PNG or JPEG image `Buffer`. This enables easy usage with `toMatchImageSnapshot`, which expects a `Buffer`.

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #4672, opened 2024-03-11
v3.1: Convert NH state primary elections and encode QR code metadata

jonahkagan:
## Overview

Task: #4641

This PR extends our v3.1 release branch to:
- Parse some newly provided NH state election definitions. The biggest change here is to support primary elections by combining election definitions for multiple parties.
- Generate encoded metadata (for now, just in QR code form) for adding to the bottom of NH state ballots. Previously, we've scanned ballots with the metadata already encoded. Now, we want to encode them on our own.

## Testing Plan
- Added some new tests to ensure the QR code metadata can be decoded
- Manually tested by converting election definitions, adding the QR code to ballots (manually editing the PDF), and then interpreting them using the lib/ballot-interpreter `interpret` script

## Checklist


libs/converter-nh-accuvote/src/convert/convert_election_definition_header.ts
```diff
@@ -139,22 +139,15 @@ export function convertElectionDefinitionHeader(
...
-  }
-  const cleanedPrecinctId = rawPrecinctId.replace(/[^-_\w]/g, '');
-  const precinctId = `town-id-${townId}-precinct-id-${cleanedPrecinctId}`;
+  const rawPrecinctId =
+    root.getElementsByTagName('PrecinctID')[0]?.textContent?.trim() ||
+    undefined;
+  const cleanedPrecinctId = rawPrecinctId?.replace(/[^-_\w]/g, '');
+  const precinctId = cleanedPrecinctId
+    ? `town-id-${townId}-precinct-id-${cleanedPrecinctId}`
+    : `town-id-${townId}-precinct`;
```
jonahkagan, line 149: I noticed that in our sample definition.xml files, the `PrecinctID` contained just a newline, so it was mistakenly passing the check that was guarding against empty `PrecinctID` (previously L143). I made this a bit more defensive and added a default case of ending in `-precinct` instead of accidentally ending up with `-precinct-id-`). This ended up having downstream effects since it impacted the existing election fixtures, which you'll encounter in later commits.

Similarly, I changed the district ID below to differentiate it from the precinct ID.

libs/image-utils/src/image_data.test.ts
jonahkagan, line 1: This commit is cherry-picked from #4664

eventualbuddha (approved):

PR #4673, opened 2024-03-11
[BMD] Set system volume to max at startup

kofi-q:
## Overview

Since we'll be managing the volume for screen reader audio in the client, this sets the system volume to a fixed point at startup to allow us to calibrate a [consistent default volume](https://github.com/votingworks/vxsuite/blob/aa06fcc/libs/ui/src/ui_strings/audio_volume.ts#L42-L50) and increase/decrease by explicit dB amounts.

Setting the startup volume to 100%, since that reduces the likelihood of having to amplify audio past the point of distortion when the voter increases playback volume.

Related: 

## Testing Plan
- New unit test cases + manual run to verify system volume is reset to max at startup.

## Checklist


libs/backend/src/initialize_system_audio.ts
```diff
@@ -0,0 +1,29 @@
...
+ * system volume to a fixed level and adjust the amplification for audio clips
+ * during playback by explicit decibel amounts, instead of adjusting the system
+ * volume by percentage amounts.
+ *
+ * Setting the system volume to 100% allows us to limit how much positive gain
+ * we apply to audio and avoid signal distortion at higher volumes.
+ */
+export async function initializeSystemAudio(): Promise<void> {
+  // Allow skipping when there's no sound card available (e.g. for integration
+  // tests in CI environments).
```
arsalansufi, line 21: If the only context that we can think of where we need to skip system audio setup is integration tests, we could use `isIntegrationTest` instead of adding a feature flag:

https://github.com/votingworks/vxsuite/blob/7d4143febe0f345c44ad2a97fbeb1f44f0a681bf/libs/utils/src/environment_variable.ts#L258-L265

I'm okay with the explicitness of a feature flag too, though

arsalansufi (approved):
LGTM!

libs/backend/src/initialize_system_audio.ts
kofi-q: Oh, perfect! Exactly what I was looking for, but not in the right places, apparently 😆  - thanks

PR #4679, opened 2024-03-13
[`controllerd`] more testable and idiomatic

eventualbuddha:
## Overview

- Improves `controllerd`'s testability by introducing a `VirtualKeyboard` trait.
- Separates the code into modules.
- Various minor improvements.

## Demo Video or Screenshot
n/a

## Testing Plan
As I can't actually test this on the real hardware easily, I'd like @kshen0 to give this the green light.

### DIFF
--- a/apps/mark-scan/accessible-controller/src/port.rs
+++ b/apps/mark-scan/accessible-controller/src/port.rs
@@ -95,8 +95,17 @@ impl io::Read for Port {
 impl fmt::Debug for Port {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Port")
-            .field("name", &self.inner.name())
-            .field("baud_rate", &self.inner.baud_rate())
+            .field(
+                "name",
+                &self.inner.name().unwrap_or_else(|| "n/a".to_string()),
+            )
+            .field(
+                "baud_rate",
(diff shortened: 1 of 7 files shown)


kshen0 (commented):
Thanks for refactoring! Code changes look good, will test on hardware tomorrow to verify.

eventualbuddha:
> Thanks for refactoring! Code changes look good, will test on hardware tomorrow to verify.

Feel free to merge if everything seems in order tomorrow.

apps/mark-scan/accessible-controller/src/port.rs
```diff
@@ -95,8 +95,17 @@ impl io::Read for Port {
 impl fmt::Debug for Port {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Port")
-            .field("name", &self.inner.name())
-            .field("baud_rate", &self.inner.baud_rate())
+            .field(
+                "name",
+                &self.inner.name().unwrap_or_else(|| "n/a".to_string()),
```
kshen0, line 100: I know your previous commit removed usage of `unwrap` but IIUC that was for better error handling. In this case I don't think we want to propagate the error up, since the name of the port is somewhat inconsequential but we need to provide a `None` alternative.

kshen0 (approved):

PR #4681, opened 2024-03-14
mark-scan: Instructions for early ballot removal

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/4542

Gives instructions to the voter when the ballot is removed during the presentation stage. On mark-scan the ideal path is for the voter to use the interface to confirm their ballot is correct, which causes the app to pull the paper back in and cast it to the rear box. But it's likely that some voters will remove the ballot during the presentation stage, especially if they realize other voters are casting their HMPBs at a precinct scanner.

In the case of early removal we should give instructions to cast at a precinct scanner rather than just ending the session.

## Demo Video or Screenshot
Changes start at 0:20

https://github.com/votingworks/vxsuite/assets/1060688/eaff182f-f523-4685-aed5-7eb6a0102dbc

## Testing Plan
- added tests

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (logging should be free with state machine state transition logging)

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
@@ -23,7 +23,11 @@ import {
   TEST_JURISDICTION,
   safeParseSystemSettings,
 } from '@votingworks/types';
-import { singlePrecinctSelectionFor } from '@votingworks/utils';
+import {
+  BooleanEnvironmentVariableName,
+  getFeatureFlagMock,
+  singlePrecinctSelectionFor,
+} from '@votingworks/utils';
 import { assert } from 'console';
 import {
   PaperHandlerStateMachine,
@@ -93,6 +97,14 @@ const webDevice: MinimalWebUsbDevice = {
 };
 
 const precinctId = electionGeneralDefinition.election.precincts[1].id;
+const featureFlagMock = getFeatureFlagMock();
+jest.mock('@votingworks/utils', (): typeof import('@votingworks/utils') => {
+  return {
+    ...jest.requireActual('@votingworks/utils'),
+    isFeatureFlagEnabled: (flag: BooleanEnvironmentVariableName) =>
+      featureFlagMock.isEnabled(flag),
+  };
+});
 
 beforeEach(async () => {
   logger = mockBaseLogger();
@@ -136,6 +148,7 @@ beforeEach(async () => {
 });
 
 afterEach(async () => {
+  featureFlagMock.resetFeatureFlags();
   await machine.cleanUp();
   jest.resetAllMocks();
 });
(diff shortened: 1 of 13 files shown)


apps/mark-scan/frontend/.codemod.bookmark
```diff
@@ -1 +0,0 @@
-./build/assets/index.c1222e2b.js:31135
```
kshen0, line 1: `codemod` util file that was previously committed by accident

apps/mark-scan/frontend/src/app_root.tsx
```diff
@@ -525,10 +529,14 @@ export function AppRoot({ reload }: Props): JSX.Element | null {
...
-        // rendering the ballot briefly before rendering the correct "Insert Card" screen
-        stateMachineState !== 'accepting_paper'
+        // Handle states that expect poll worker auth. If the frontend sees one of these states but has cardless voter auth,
+        // it means the state hasn't caught up to auth changes. We check that edge case here to avoid flicker eg.
+        // rendering the ballot briefly before rendering the correct "Insert Card" screen.
+        // This problem is caused by conditioning on auth in both the frontend and backend. The long term fix is to
+        // move auth entirely to the backend.
+        // https://github.com/votingworks/vxsuite/issues/3985
+        stateMachineState !== 'accepting_paper' &&
+        stateMachineState !== 'not_accepting_paper'
```
kshen0, line 539: Now that we've seen this is a consistent pattern for any state that changes auth in the backend, IMO we should bump the priority of #3985 ahead of adding any additional states with auth changes @arsalansufi @jonahkagan @mattroe

apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
```diff
@@ -283,6 +296,45 @@ test('voting flow happy path', async () => {
   await waitForStatus('not_accepting_paper');
 });
 
+describe('removing ballot during presentation state', () => {
+  test('is a no op if USE_MOCK_PAPER_HANDLER=true', async () => {
```
kshen0, line 300: Need this to no-op because paper handler status always reports `NO_PAPER_ANYWHERE` when using mock paper handler.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
```diff
@@ -391,6 +443,14 @@ test('getRawDeviceStatus', async () => {
...
 });
 
+describe('paperHandlerStatusToEvent', () => {
+  test('PAPER_IN_OUTPUT', () => {
+    expect(paperHandlerStatusToEvent(getPaperInRearStatus())).toEqual({
+      type: 'PAPER_IN_OUTPUT',
+    });
+  });
+});
+
```
kshen0, line 453: Added for coverage. I couldn't find where branch coverage went down with my changes but this brings it back up.

jonahkagan (approved):

apps/mark-scan/frontend/src/app_root.tsx
arsalansufi: That makes sense! I brought the issue up to the medium pri section on the VxMarkScan board. (I don't have a great sense of which issues would require "adding any additional states with auth changes")

PR #4693, opened 2024-03-19
fix(design): omit empty place parts

eventualbuddha:
## Overview

If someone wants to leave e.g. state blank we shouldn't put a trailing comma after the county name. I realize this is going to be replaced soonish. Consider it a request to preserve this fix in the new version.

## Demo Video or Screenshot
image

## Testing Plan
Generated some sample ballots.


jonahkagan (approved):
noted!

PR #4695, opened 2024-03-20
Allow daemons to run without hardware; make more robust for different device configs

kshen0:
## Overview
Allows `controllerd` and `patinputd` to run when no hardware is found. The daemons will report failure to connect then run in a no-op loop with a heartbeat logged every 5s.

## Demo Video or Screenshot

PAT input running with no hardware

Screenshot 2024-03-22 at 6 30 48 PM

Controller daemon running with no hardware

Screenshot 2024-04-01 at 3 36 41 PM

Controller daemon successfully connecting

Screenshot 2024-04-01 at 3 35 21 PM

## Testing Plan
* Manually tested both daemons run on VM
* Manually tested both hardware devices work on real hardware

## Checklist

### DIFF
--- a/apps/mark-scan/accessible-controller/src/controllerd.rs
+++ b/apps/mark-scan/accessible-controller/src/controllerd.rs
@@ -36,7 +36,7 @@ const KPB_200_FW_PID: u16 = 16392;
 #[derive(Parser, Debug)]
 #[command(version, about, long_about = None)]
 struct Args {
-    // Whether to allow the daemon to run if no hardware is found.
+    /// Allow the daemon to run if no hardware is found.
     #[arg(short, long)]
     skip_hardware_check: bool,
 }
@@ -44,6 +44,7 @@ struct Args {
 fn main() -> color_eyre::Result<()> {
     color_eyre::install()?;
 
+    let args = Args::parse();
     set_app_name(APP_NAME);
     log!(
         EventId::ProcessStarted;
@@ -87,7 +88,6 @@ fn main() -> color_eyre::Result<()> {
         exit(0);
     }
 
-    let args = Args::parse();
     if args.skip_hardware_check {
         run_no_op_event_loop(&running);
         exit(0);
--- a/apps/mark-scan/accessible-controller/src/controllerd.rs
+++ b/apps/mark-scan/accessible-controller/src/controllerd.rs
@@ -30,6 +30,8 @@ mod device;
 mod port;
 
 const APP_NAME: &str = "vx-mark-scan-controller-daemon";
+const KPB_200_FW_VID: u16 = 10445;
+const KPB_200_FW_PID: u16 = 16392;
 
 #[derive(Parser, Debug)]
 #[command(version, about, long_about = None)]
@@ -80,7 +82,27 @@ fn main() -> color_eyre::Result<()> {
         EventType::SystemAction
     );
 
-    match Port::open() {
+    if let Some(port) = get_port() {
+        run_event_loop(keyboard, port, &running);
+        exit(0);
+    }
+
+    let args = Args::parse();
+    if args.skip_hardware_check {
+        run_no_op_event_loop(&running);
+        exit(0);
+    }
+
+    log!(
+        event_id: EventId::ProcessTerminated,
+        event_type: EventType::SystemStatus,
+        message: "Could not connect to ATI controller".to_string()
+    );
+    exit(1);
+}
+
+fn get_port() -> Option<Port> {
+    match Port::open_by_ids(KPB_200_FW_VID, KPB_200_FW_PID) {
         Ok(port) => {
             log!(
                 event_id: EventId::ControllerConnectionComplete,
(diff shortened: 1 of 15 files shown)


apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -28,6 +30,15 @@ mod device;
 mod port;
 
 const APP_NAME: &str = "vx-mark-scan-controller-daemon";
+const POSSIBLE_DEVICE_PATHS: [&str; 2] = ["/dev/ttyACM0", "/dev/ttyACM1"];
```
kshen0, line 33: I went back and forth on using `available_ports` to list ports rather than just hardcode. I'll try out `available_ports` on hardware at next availability and possibly update this.

apps/mark-scan/accessible-controller/src/port.rs
```diff
@@ -18,15 +17,52 @@ pub struct Port {
...
             .open()?;
         let mut port = Self { inner };
         port.validate_connection()?;
         Ok(port)
     }
 
+    fn get_device_path(vendor_id: u16, product_id: u16) -> Result<String, io::Error> {
+        match available_ports() {
+            Ok(ports_info) => {
+                for port_info in ports_info {
```
kshen0, line 37: Wasn't sure if `iter` is more idiomatic but this is how `serialport` examples do it

eventualbuddha: No, this is the typical way to do it in Rust. `for` loops desugar into a `loop` with manual handling of an `Iterator`:

```rust
// this…
for i in 0..2 {
    println!("{}", i);
}

// becomes (more or less) this:
{
    let _t =
        match IntoIterator::into_iter(Range { start: 0, end: 2 }) {
            mut iter =>
            loop {
                match iter.next() {
                    None => break,
                    Some(i) => {
                        println!("{}", i);
                    },
                }
            },
        };
    _t
}
```

`IntoIterator::into_iter` consumes `self`, which in your code is `ports_info`. This is what you want since you don't need to use `ports_info` later. If you did, you could do `for port_into in ports_info.iter()`, which creates an `Iterator<Item = &SerialPortInfo>` instead of `Iterator<Item = SerialPortInfo>`, making `port_info` a reference instead of an owned value.

apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -28,10 +30,21 @@ mod device;
 mod port;
 
 const APP_NAME: &str = "vx-mark-scan-controller-daemon";
+const KPB_200_FW_VID: u16 = 10445;
+const KPB_200_FW_PID: u16 = 16392;
```
eventualbuddha, line 34: These are frequently displayed as hex values rather than decimal values (e.g. in the output of `lsusb` and often in device documentation). Consider using hex?
```suggestion
const KPB_200_FW_VID: u16 = 0x28cd;
const KPB_200_FW_PID: u16 = 0x4008;
```

apps/mark-scan/daemon-utils/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread::sleep,
+    time::{Duration, Instant},
+};
+use vx_logging::{log, EventId, EventType};
+
+const HEARTBEAT_INTERVAL: Duration = Duration::new(5, 0);
+const NOOP_LOOP_INTERVAL: Duration = Duration::new(1, 0);
```
eventualbuddha, line 12: Use factory functions that make the value clearer?
```suggestion
const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
const NOOP_LOOP_INTERVAL: Duration = Duration::from_secs(1);
```

apps/mark-scan/pat-device-input/Cargo.toml
```diff
@@ -11,5 +11,7 @@ path = "src/patinputd.rs"
 
 [dependencies]
 ctrlc = { workspace = true }
-vx-logging = { workspace = true }
+daemon-utils = { workspace = true }
```
eventualbuddha, line 14: Alphabetical?

eventualbuddha (approved):

apps/mark-scan/accessible-controller/src/port.rs
kshen0: Thank you for the clarification!

apps/mark-scan/accessible-controller/src/controllerd.rs
kshen0: https://github.com/votingworks/vxsuite/pull/4695/commits/454ff9f6d85379e1317715dbadcd666586cb24f0

apps/mark-scan/pat-device-input/Cargo.toml
kshen0: https://github.com/votingworks/vxsuite/pull/4695/commits/2a9d30f17aee158de8148258252cd165f47910bd

PR #4696, opened 2024-03-20
Modify pdfToImages to output an image Buffer instead of ImageData

jonahkagan:
## Overview

Avoids an extra round trip from Canvas->Buffer->Canvas in all of the places we actually use pdfToImages. Also removes the toImageBuffer function that was being used to do this round trip, which is no longer needed.

## Checklist


jonahkagan:
Realized there are actually some important places where we use the existing pdfToImages functionality, so punting on this PR since it would optimize tests at the expense of app code

PR #4698, opened 2024-03-20
fix(auth): properly parse the challenge response

eventualbuddha:
## Overview

Rather than hardcoding a fixed number of bytes to trim we now properly parse the nested response TLV structures. While the old way didn't bite us in practice, if the signature returned was ever larger than 0x80 bytes the TLV would use two bytes to encode the length and the signature would be corrupted.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests, plus manual testing with a sysadmin card.

### DIFF
--- a/libs/auth/src/java_card.ts
+++ b/libs/auth/src/java_card.ts
@@ -561,7 +561,16 @@ export class JavaCard implements Card {
         ),
       })
     );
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
Good catch!

PR #4699, opened 2024-03-21
chore(auth): remove unnecessary `async` in tmpfile cleanup

eventualbuddha:
## Overview

`FileResult::removeCallback` doesn't return a `Promise`, so there's no sense in `await`ing the result. See the value of `removeCallback` here: https://github.com/raszi/node-tmp/blob/7ae22ed2d56c10d425a66e99fe8bc10c925442e6/lib/tmp.js#L181.

## Demo Video or Screenshot
n/a

## Testing Plan
Updated automated test that incorrectly mocked using a `Promise`.


arsalansufi (approved):
Good catch!

PR #4700, opened 2024-03-21
fix(auth): parse all TLV responses properly

eventualbuddha:
## Overview

Rather than trimming metadata by ignoring fixed-length prefixes and suffixes, we now properly parse the TLV structures. Additionally, this changes the `parseTlv` function to expect the data to be entirely consumed by the TLV structure. If it isn't, use `parseTlvPartial` instead, which returns a remainder.

## Demo Video or Screenshot
n/a

## Testing Plan
Updated tests for where we had the expected responses wrong (just getting certs for CAC's).

### DIFF
--- a/libs/auth/src/apdu.ts
+++ b/libs/auth/src/apdu.ts
@@ -286,12 +286,24 @@ export function constructTlv(
 }
 
 /**
- * The inverse of constructTlv, splits a TLV into its tag, length, and value
+ * A Tag-Length-Value (TLV) structure.
+ *
+ * See https://en.wikipedia.org/wiki/Type-length-value for more information.
  */
-export function parseTlv(
+export interface Tlv {
+  tag: Buffer;
+  length: Buffer;
+  value: Buffer;
+}
+
+/**
+ * Parses a TLV from a buffer, returning the remainder of the buffer and the parsed TLV. If the tag
+ * does not match the expected tag, an error is thrown.
+ */
+export function parseTlvPartial(
   tagAsByteOrBuffer: Byte | Buffer,
   tlv: Buffer
-): [tag: Buffer, length: Buffer, value: Buffer] {
+): [remainder: Buffer, tlv: Tlv] {
   const expectedTag = Buffer.isBuffer(tagAsByteOrBuffer)
     ? tagAsByteOrBuffer
     : Buffer.of(tagAsByteOrBuffer);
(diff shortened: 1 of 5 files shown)


libs/auth/src/apdu.ts
```diff
@@ -343,8 +355,23 @@ export function parseTlv(
...
+ * The inverse of constructTlv, splits a TLV into its tag, length, and value. If the tag does not
+ * match the expected tag, an error is thrown. If the entire TLV is not consumed, an error is
+ * thrown.
+ */
+export function parseTlv(tagAsByteOrBuffer: Byte | Buffer, data: Buffer): Tlv {
+  const [remainder, tlv] = parseTlvPartial(tagAsByteOrBuffer, data);
+  assert(
+    remainder.length === 0,
+    `TLV was not fully consumed: remainder=${remainder.toString('hex')}`
+  );
```
arsalansufi, line 373: :100:

libs/auth/src/java_card.ts
```diff
@@ -496,10 +497,25 @@ export class JavaCard implements Card {
...
+      'Expected cert info to be uncompressed'
+    );
+    const certErrorDetectionCode = parseTlv(
+      PUT_DATA.ERROR_DETECTION_CODE_TAG,
+      certInfoRemainder
+    ).value;
+    assert(
+      certErrorDetectionCode.length === 0,
+      'Expected no error detection code'
+    );
```
arsalansufi, line 517: Great call being more explicit with the parsing here

libs/auth/src/apdu.ts
```diff
@@ -286,12 +286,24 @@ export function constructTlv(
...
+/**
+ * Parses a TLV from a buffer, returning the remainder of the buffer and the parsed TLV. If the tag
+ * does not match the expected tag, an error is thrown.
  */
-export function parseTlv(
+export function parseTlvPartial(
   tagAsByteOrBuffer: Byte | Buffer,
   tlv: Buffer
-): [tag: Buffer, length: Buffer, value: Buffer] {
+): [remainder: Buffer, tlv: Tlv] {
```
arsalansufi, line 306: nit: Very minor preference for putting the remainder after the tlv in the return value as that feels a bit more intuitive than having the remainder come first

libs/auth/src/cac/common_access_card.ts
```diff
@@ -412,8 +413,23 @@ export class CommonAccessCard implements CommonAccessCardCompatibleCard {
...
-    const [, , certInDerFormat] = parseTlv(PUT_DATA.CERT_TAG, certTlv);
+    const [certInDerFormatRemainder, { value: certInDerFormat }] =
+      parseTlvPartial(PUT_DATA.CERT_TAG, data);
+    const [certInfoRemainder, { value: certInfo }] = parseTlvPartial(
+      PUT_DATA.CERT_INFO_TAG,
+      certInDerFormatRemainder
+    );
+    // NOTE: CACs seem to return 0x01 for this. I'm not sure what it means, but
+    // it's not compressed.
+    assert(certInfo[0] === 0x01, 'Expected cert info to be uncompressed');
```
arsalansufi, line 424: Interesting! According to the page 35 of the [NIST PIV standard](https://nvlpubs.nist.gov/nistpubs/specialpublications/nist.sp.800-73-4.pdf):

> Thus, for a certificate encoded in uncompressed form CertInfo shall be
> 0x00, and for a certificate encoded using GZIP compression CertInfo shall be 0x01.

Odd that it's not actually compressed

arsalansufi (approved):
Thanks for cleaning this up!

libs/auth/src/cac/common_access_card.ts
eventualbuddha: ae2551ddbb097efe0eb4dcc19f1e8540af4797d2

libs/auth/src/apdu.ts
eventualbuddha: 4df01ada191004de047ec78e3d04998569db33e1

PR #4701, opened 2024-03-21
refactor(scan): polls flow and poll worker screen

adghayes:
## Overview

The `PollWorkerScreen` in VxScan was (still is) a large and unwieldy component. Also, the poll transitions are a bit nuanced with lots of branching logic. The biggest change I ended up making was making open polls, close polls, pause voting, and resume voting completely different paths, where reasonable. When I originally made the transition polls logic, I tried to model everything as a single `transitionPolls` action which behaves differently based on the type of polls transition. But differences between each transition continue to crop up, so it actually feels safer now to just treat each transition as a distinct code path at most levels, which does add some redundancy, rather than have a single code path with lots of exceptions.

I also refactored the frontend:
- move some functions to the backend where possible (triggering CVR export, logging)
- change the state management of `PollWorkerScreen` to be more unified
- pull out subscreens into their own components where reasonable


apps/scan/backend/src/app_config.test.ts
```diff
@@ -182,124 +181,6 @@ test('setPrecinctSelection will reset polls to closed', async () => {
   });
 });
 
-test('ballot batching', async () => {
```
adghayes, line 185: moved to `app_polls.test.ts`

apps/scan/backend/src/export.ts
adghayes, line 1: moved from `app.ts`

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -143,21 +316,11 @@ export function PollWorkerScreen({
...
-  // lastPollsTransition is really the currentPollsTransition.
-  const lastPollsTransition: Optional<PollsTransition> =
-    transitionPollsMutation.variables
-      ? {
-          ...transitionPollsMutation.variables,
-          ballotCount: scannedBallotCount,
-        }
-      : pollsInfo.pollsState !== 'polls_closed_initial'
-      ? pollsInfo.lastPollsTransition
-      : undefined;
```
adghayes, line 158: Wanted to get rid of this weird piece of state management.

apps/scan/backend/src/polls.ts
```diff
@@ -67,12 +67,15 @@ export async function closePolls({
...
+    const exportResult = await exportCastVoteRecordsToUsbDrive({
       mode: 'polls_closing',
       workspace,
       usbDrive,
       logger,
-    }));
+    });
+    exportResult.assertOk(
+      'Failed to finish cast vote record export to USB drive.'
+    );
```
adghayes, line 78: Making this throw on fail, which matches the previous behavior.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -106,17 +278,18 @@ const ButtonGrid = styled.div`
 `;
 
 export function PollWorkerScreen({
+  initialPollsInfo,
```
jonahkagan, line 281: What's the motivation for passing in an initial polls info and then also querying polls info within the component?

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -169,268 +332,191 @@ export function PollWorkerScreen({
...
-      const pollsTransitionTime = getCurrentTime();
-      await transitionPollsMutation.mutateAsync({
-        type: pollsTransitionType,
-        time: pollsTransitionTime,
+  async function openPolls() {
+    // In compliance with VVSG 2.0 1.1.3-B, confirm there are no scanned
+    // ballots before opening polls, even though this should be an impossible
+    // state in production.
+    if (scannedBallotCount > 0) {
+      setIsShowingBallotsAlreadyScannedScreen(true);
```
jonahkagan, line 352: This could also move to the backend at some point and just show an error message if it occurs

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -169,268 +332,191 @@ export function PollWorkerScreen({
...
-              afterPrint={() =>
-                setPollWorkerFlowState('polls_transition_complete')
-              }
-            />
-          </P>
-        </CenteredLargeProse>
-      </Screen>
-    );
+  if (isShowingBallotsAlreadyScannedScreen) {
+    return BallotsAlreadyScannedScreen;
```
jonahkagan, line 438: Wait, how does this work? Shouldn't it be `<BallotsAlreadyScannedScreen />`?

jonahkagan (approved):
Much clearer!

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: Because I want to set the initial component state (`pollWorkerFlowState`). Setting it after the query completes would force me to use an effect, I thought. Do you have an idea for a better pattern for this? I agree it's weird.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: True. We have the assertion on the backend but we could just pass an `err` `Result` and render the screen accordingly. I'd like that because it would move this logging to the backend, but punted on it here.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: Good catch! I guess it has been working, because it was the same before. React must accept / handle this, although perhaps not exactly as we expect. Will switch it to JSX syntax.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
jonahkagan: My go-to pattern is to use a parent component that has the query and renders a loading state until the data is loaded. Then the child component has a required prop for the result of the query. (Which is basically what was already set up here, with `AppRoot` as the parent).

Sometimes it's nice for testing to have a screen component manage all its own queries though, in which case I just make the screen component the parent and add a child as needed. E.g.:

```ts
function PollWorkerScreen() {
  const pollsInfoQuery = getPollsInfo.useQuery()
  if (!pollsInfoQuery.isSuccess) return null
  return <PollWorkerScreenContents pollsInfo={pollsInfo} />
}
```
}

apps/scan/frontend/src/screens/poll_worker_screen.tsx
jonahkagan: Oh I just looked at the definition of `BallotsAlreadyScannedScreen` - it's a JSX value, not a component function, so that explains it.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: I added a wrapper in my latest commit, but it brought me back to why I originally did it in this weird way.

Do you remember this discussion? https://github.com/votingworks/vxsuite/pull/4222#discussion_r1396218990

Some states of the poll worker screen depend on the `pollsInfo` being up to date. E.g., by the time we go to print a report, we expect `pollsInfo` to be updated with the latest polls transition. We can use `async` `onSuccess` invalidations to guarantee this, but not if the mutation and the query are in different components, which is what happens when we use the suggested pattern. 

To address this, using a wrapper, I added that state that we were asserting exists into the `FlowState`.

62ea498fb

Issue #4474, opened 2024-01-04
vxmarkscan: main ballot template


kshen0:
Review BMD ballot layout and font size. In particular, use VVSG min paper font size and ensure all content is visible to voter for review after printing.

See https://votingworks.slack.com/archives/C03Q42V9MJ5/p1710783453461679 for current state.

The current template shows all the content to the voter (as desired) with a 2" top margin. VVSG2 requires that paper records have a font-size of at least 3.5mm (10 points). We should confirm that this is in place and update as needed.

Relatedly, we have the unsolved problem of https://github.com/votingworks/vxsuite/issues/4473 that we'll solve by adding in a second ballot layout template rather than supporting scanning of multiple sheets for a given voter on the BMD. To understand when that template should trigger, we should size the number of contests that can fit on the main template. If reducing the main template to 10 pt (from something larger) significantly increases the number of contests we can fit, we should consider making that change.

Issue #4473, opened 2024-01-04
vxmarkscan: large-N contest printed ballot template


kshen0:
Related to: https://github.com/votingworks/vxsuite/issues/4474

In the case of a high (exact threshold to be determined) number of contests, we should print vxmarkscan ballots in a different template to fit more content on the ballot. These changes should include: 
- removing the top padding and therefore hiding the header from the voter.
- reducing the font size below the 10pt number to what we deem the minimal readable number. this requires a bit of subjective analysis and we should test options.

We may also want to incorporate a three column layout to maximize the amount of screen real estate used by voter selections.

In the course of making these adjustments, we should calculate what the maximum number of selections a voter can make that can be printed on one page. There is no specific minimum number to meet, but real world usage can see over 50 and maybe up to 100.

PR #4723, opened 2024-03-25
Update BMD paper ballot single-page layout to fit more contests

kofi-q:
## Overview

Resolves https://github.com/votingworks/vxsuite/issues/4474

Applying the following changes to increase the number of contests we can fit on a single page of the BMD paper ballot:
- Decrease font size to the VVSG 2.0 minimum: `10pt`
- For VSAP prints, decrease the size of the top margin, so that the ballot header is covered by the VSAP in-feed hood, while keeping the first row of contests visible.
- Increase the number of contest columns as the number of columns increases.
- Drop party names from candidate selections when rendering 3+ columns.
- Shrink contest titles.
- Shrink English text on multi-language ballots further (75% -> 60%).
- Inline English text whenever possible on multi-language ballots (we were previously rendering on a new line in some cases).
- Reduce spacing where possible.

We don't yet put a cap on how many contests this will attempt to render on a single page - we'll be addressing how split contests across multiple sheets later in https://github.com/votingworks/vxsuite/issues/4473.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/264902/87926a3e-2213-4158-84d7-3996767bcad4

## Testing Plan
- Added/updated unit test cases.
- Manual testing and tuning for layout change thresholds.

## Checklist

### DIFF
--- a/libs/ui/src/bmd_paper_ballot.tsx
+++ b/libs/ui/src/bmd_paper_ballot.tsx
@@ -39,36 +39,62 @@ import {
   NumberString,
 } from './ui_strings';
 
-export type BmdBallotPrintType = 'vxMark' | 'vsap';
+export type MachineType = 'mark' | 'markScan';
 
 /**
  * Max margin required to keep the ballot header visible when the page is
  * partially covered by the VSAP infeed hood.
  */
-export const MAX_VSAP_TOP_MARGIN = '1.75in';
+export const MAX_MARK_SCAN_TOP_MARGIN = '1.75in';
 
 /**
  * Min margin required to keep the first row of contests visible when the page
  * is partially covered by the VSAP infeed hood.
  */
-export const MIN_VSAP_TOP_MARGIN = '0.5625in';
+export const MIN_MARK_SCAN_TOP_MARGIN = '0.5625in';
 
 interface Layout {
+  /**
+   * Whether or not to hide candidate party names in candidate contests to save
+   * space in denser layouts.
+   */
   hideParties: boolean;
+
+  /** Maximum number of contest rows per column to print on the ballot. */
   maxRows: number;
+
+  /**
+   * Minimum number of contests required for this layout to be used in printing
+   * a ballot.
+   */
   minContests: number;
-  topMargin?: typeof MAX_VSAP_TOP_MARGIN | typeof MIN_VSAP_TOP_MARGIN;
+
+  /**
+   * Additional top margin to apply to the print to ensure all important content
+   * is visible - necessary for MarkScan, where printouts are partially covered
+   * by the hood of the paper tray when being reviewed by the voter.
+   */
+  topMargin?: typeof MAX_MARK_SCAN_TOP_MARGIN | typeof MIN_MARK_SCAN_TOP_MARGIN;
 }
 
-export const BMD_BALLOT_LAYOUTS: Record<BmdBallotPrintType, Layout[]> = {
-  vsap: [
+/**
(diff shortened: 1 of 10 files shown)


libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -39,12 +39,50 @@ import {
   NumberString,
 } from './ui_strings';
 
-const Ballot = styled.div`
+export type BmdBallotPrintType = 'vxMark' | 'vsap';
```
jonahkagan, line 42: It might be clearer to have these be `mark` vs `mark-scan` to correspond with the names of the apps in the rest of the codebase

libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -39,12 +39,50 @@ import {
...
+
+/**
+ * Max margin required to keep the ballot header visible when the page is
+ * partially covered by the VSAP infeed hood.
+ */
+export const MAX_VSAP_TOP_MARGIN = '1.75in';
```
jonahkagan, line 48: Similarly, maybe replace references to VSAP with VxMarkScan in constant names/comments

libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -393,10 +418,17 @@ export function BmdPaperBallot({
     }
   }, [onRendered]);
 
+  const layout = find(
+    [...BMD_BALLOT_LAYOUTS[printType]].reverse(),
+    (l) => contests.length >= l.minContests
+  );
```
jonahkagan, line 424: Does this rely on `BMD_BALLOT_LAYOUTS` being specified in a certain order? Might be more defensive to just sort here before selecting a layout

libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -39,12 +39,50 @@ import {
...
+ * is partially covered by the VSAP infeed hood.
+ */
+export const MIN_VSAP_TOP_MARGIN = '0.5625in';
+
+interface Layout {
+  hideParties: boolean;
+  maxRows: number;
+  minContests: number;
+  topMargin?: typeof MAX_VSAP_TOP_MARGIN | typeof MIN_VSAP_TOP_MARGIN;
+}
```
jonahkagan, line 61: It's nice how these layout specs are defined as data. Very cool

I'm a little confused still by what `maxRows` means

jonahkagan (approved):

libs/ui/src/bmd_paper_ballot.tsx
kofi-q: Ah, good call

libs/ui/src/bmd_paper_ballot.tsx
kofi-q: +1

libs/ui/src/bmd_paper_ballot.tsx
kofi-q: Yeah, good point - makes me think all of these fields could use documentation for clarity. Will update

libs/ui/src/bmd_paper_ballot.tsx
kofi-q: There's a unit test asserting the order of the layouts, but maybe this would be less confusing if we just explicitly sort here or add more obvious naming/documentation -- will update

PR #4724, opened 2024-03-25
Fujitsu Thermal Printer Initial Integration

adghayes:
## Overview

We did a preliminary software validation of the Fujitsu thermal printer for VxScan a few months ago. This is a more extensive, app-integrated proof-of-concept that will allow for paper path testing. I'm trying to strike a balance between something extensive enough to merge to `main` and to highlight any rough edges to the integration, but minimal enough that I don't waste cycles building something that won't work in the actual paper path / enclosure.

PR Includes:
- Basic Driver
- Driver CLI
- Print Success / Failure Tracking
- Basic UX Loading Flow
- Basic UX Printing Flow
- Printing flow when multiple reports are involved. I decided this was important to include in an initial POC because it is so different than a single report flow
- Happy path testing

PR Excludes:
- Fujitsu Mock for dev environments. Everything is behind a feature flag, and if the feature flag is on, the machine expects the hardware to be available at startup or will fail.
- Testing for all failure cases

One of the goal requirements is to leave V3 hardware printing support untouched. But because an embedded roll printer is such a different interface than a standard printer, supporting them both is a bit clunky. As you review the PR, note whether there are ways you think the two paths could co-exist better with better abstractions, or if it seems just too unwieldy, cutting out V3 support is not off the table for a future PR. cc @arsalansufi @benadida 

#### Paper Loading

https://github.com/votingworks/vxsuite/assets/37960853/3102f8d6-58e0-4ca6-b544-5735e99af1af

#### Print Single Report

https://github.com/votingworks/vxsuite/assets/37960853/13f43423-cf99-480c-88b0-f5310eee7020

#### Print Multipart Report

https://github.com/votingworks/vxsuite/assets/37960853/bb2189aa-51c5-4577-b113-ee4171832a2f

#### Print Fails (No Paper)

https://github.com/votingworks/vxsuite/assets/37960853/1bb75984-bf72-4038-ac6f-a2f3f21ee53d

#### Print Fails (Error)

https://github.com/votingworks/vxsuite/assets/37960853/02fab842-0797-43f5-a681-6552a78daf05

## Checklist


apps/scan/backend/src/printing/printer.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export type PrintResult =
+  | {
+      scheme: 'hardware-v3';
+      pageCount: number;
+    }
+  | {
+      scheme: 'hardware-v4';
+      result: FujitsuPrintResult;
+    };
```
adghayes, line 56: To be able to have a singular `Printer` in the app, and a single `getPrinterStatus` endpoint and `printReport` endpoint, creating all these discriminated types.

apps/scan/backend/test/helpers/custom_helpers.ts
```diff
@@ -74,6 +86,7 @@ export async function withApp(
     workspace: Workspace;
     mockUsbDrive: MockUsbDrive;
     mockPrinterHandler: MemoryPrinterHandler;
+    mockFujitsuPrinterHandler: MemoryFujitsuPrinterHandler;
```
adghayes, line 89: This is a bit awkward - the testing helper will return a print handler for both types of printers, but only one will actually be active depending on what printer the app is using. Could make a sort of discriminated union of the two but then you'd have to assert within each test, instead of just picking your poison.

mattroe:
UX LGTM and I think this is the right balance for the level of integration we want to build at this point. nice job @adghayes

adghayes:
@kofi-q I noticed in the multi-page flow that the buttons were "Print Previous Report" and "Print Next" - I just removed "Report" from all buttons in the multi-page flow so it's just "Print Previous" and "Print Next," and the "Print Next" is more appropriately sized relative to the other button. Just noting here because I haven't updated the videos.

libs/fujitsu-thermal-printer/src/printing.ts
```diff
@@ -0,0 +1,351 @@
...
+  debug('printed page successfully');
+  return ok();
+}
+
+export async function print(
+  driver: FujitsuThermalPrinterDriverInterface,
+  pdfData: Uint8Array
+): Promise<Result<void, RawPrinterStatus>> {
+  const pdfImages = pdfToImages(Buffer.from(pdfData), {
+    scale: 200 / 72,
```
kofi-q, line 320: I think I can guess these numbers are, but could we add comments or constant definitions for clarity?

libs/fujitsu-thermal-printer/src/printing.ts
```diff
@@ -0,0 +1,351 @@
...
+
+  await driver.setReplyParameter(IDLE_REPLY_PARAMETER);
+  return ok();
+}
+
+export async function printFixture(
+  pdfFixturePath: string,
+  driver: FujitsuThermalPrinterDriver
+): Promise<void> {
+  const pdfData: Uint8Array = readFileSync(pdfFixturePath);
```
kofi-q, line 346: nit: Should we be moving to the `readFileSync` util from `libs/fs` that Brian added? This looks like it's not in the main app path, so maybe less important

kofi-q (approved):
🖨️

apps/scan/backend/src/printing/print_report_section.ts
```diff
@@ -0,0 +1,100 @@
...
+    isLiveMode,
+    pollsTransition: pollsTransitionType,
+    pollsTransitionedTime: pollsTransitionTime,
+    reportPrintedTime: getCurrentTime(),
+    precinctScannerMachineId: machineId,
+    electionResultsByParty: scannerResultsByParty,
+  });
+}
+
+const getReportSectionsMemoized = memoizeOne(getReportSections);
```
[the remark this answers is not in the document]
adghayes: I didn't run into perf problems while making this PR, but we did run into perf problems when testing the all bubble ballot. With so many contests, it took almost a minute to create a report for 1,000's of ballots. The expensive part is not the DB fetches or the React initializations, I believe, it's the tabulation itself. The all bubble ballot is an edge case, but I think an informative one. If we had an equivalently large election that was a primary, I'd rather the wait be all up front rather than repeated between each print.

PR #4727, opened 2024-03-27
PDI scanner driver

jonahkagan:
## Overview

Adds the PDI driver written in Rust by @eventualbuddha. I want to get this merged into `main` so that I can start building on top of it in smaller chunks. I squashed all the WIP commits into one, then did some minor cleanup and added a bit of documentation.

Note that the TS client is still in an unfinished prototype state - I plan to work on fleshing that out next in a separate PR.

## Checklist


libs/pdi-scanner/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+                    stdin_tx.send(command).unwrap();
+                    // println!("received message: {command:?}");
+                }
+
+                Err(e) => {
+                    tracing::error!("failed to read from stdin: {e:?}");
+                }
+            }
+        }
+    });
```
jonahkagan, line 152: What's the advantage to having a separate thread for checking stdio for new commands vs just doing it in the main loop?

libs/pdi-scanner/src/protocol/packets.rs
```diff
@@ -0,0 +1,1217 @@
...
+    /// (`n3a20`).
+    GetDoubleFeedDetectionDoubleSheetCalibrationValueRequest,
+
+    /// Requests the double feed detection threshold value (`n3a10`).
+    GetDoubleFeedDetectionDoubleSheetThresholdValueRequest,
+
+    RawPacket(Vec<u8>),
+}
+
+macro_rules! checked {
```
jonahkagan, line 571: Just out of curiosity, why does this need to be a macro rather than a function?

libs/pdi-scanner/src/protocol/parsers.rs
```diff
@@ -0,0 +1,1297 @@
...
+    map(
+        packet((tag(b"Q"), le_u8, le_u8, le_u8)),
+        |(_, byte0, byte1, byte2)| {
+            Status::new(
+                byte0 & 0b0000_0001 != 0,
+                byte0 & 0b0000_0010 != 0,
+                byte0 & 0b0000_0100 != 0,
+                byte0 & 0b0000_1000 != 0,
+                byte0 & 0b0001_0000 != 0,
+                byte0 & 0b0100_0000 != 0,
```
jonahkagan, line 779: I noticed these two lines skip over `0b0010_0000`. Is that intentional?

libs/pdi-scanner/src/protocol/parsers.rs
```diff
@@ -0,0 +1,1297 @@
...
+simple_response!(fpga_out_of_date_event, b"#0F");
+simple_response!(calibration_ok_event, b"#10");
+simple_response!(calibration_short_calibration_document_event, b"#11");
+simple_response!(calibration_document_removed_event, b"#12");
+simple_response!(calibration_pixel_error_front_array_black, b"#13");
+simple_response!(calibration_pixel_error_front_array_white, b"#19");
+simple_response!(calibration_timeout_error, b"#1A");
+simple_response!(calibration_speed_value_error, b"#1B");
+simple_response!(calibration_speed_box_error, b"#1C");
+// TODO: fill out the rest
```
jonahkagan, line 1252: What are the rest?

libs/pdi-scanner/src/rusb_async/error.rs
```diff
@@ -0,0 +1,59 @@
...
+//
+//     THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+//     EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+//     MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+//     NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+//     LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+//     OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+//     WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+// Modifications have been made to better support the use case of this project.
```
jonahkagan, line 26: Not totally sure how to review these modded rusb files. Suggestions?

libs/pdi-scanner/src/scanner.rs
```diff
@@ -0,0 +1,265 @@
...
+                        Ok(StopMode::WaitUntilTransfersComplete) => {
+                            tracing::debug!(
+                                "Scanner thread received stop signal; cancelling incoming transfers and waiting for outgoing transfers to complete"
+                            );
+                            is_stopping = true;
+                            transfer_pool.cancel_all_for_endpoint(ENDPOINT_IN_PRIMARY);
+                            transfer_pool.cancel_all_for_endpoint(ENDPOINT_IN_IMAGE_DATA);
+                        }
+                        Err(_) => {}
+                    }
```
jonahkagan, line 127: Why have two stop modes? It's not totally clear to me how to choose one over the other

libs/pdi-scanner/src/scanner.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        (
+            host_to_scanner_tx,
+            host_to_scanner_ack_rx,
+            scanner_to_host_rx,
+        )
+    }
+
+    /// Stop the scanner. Blocks until cleanup is complete.
+    pub fn stop(&mut self, mode: StopMode) {
```
jonahkagan, line 256: I could use a little more info on what `stop` does and when it should be called. From my initial experimentation, it seems important to call this method (e.g. if I didn't call it, I couldn't open a new connection to the scanner in pdi-tui)

jonahkagan (commented):
@eventualbuddha I reviewed the Rust files (excluding bin/pdi-tui and parse_traffic.rs, since they are just dev tools)

libs/pdi-scanner/src/client.rs
```diff
@@ -0,0 +1,921 @@
...
+
+    /// Connects to the scanner using the same commands that were captured by
+    /// wireshark from the `scan_demo` program. Some of them may not be
+    /// necessary, but they're included for now.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if any of the commands fail to
+    /// validate or if the response is not received within the timeout.
+    pub fn send_connect(&mut self) -> Result<()> {
```
jonahkagan, line 745: What do you think we should do with `send_connect` and `send_enable_scan_commands`? Should we keep using them or try going without and adding in the bare minimum necessary?

(accidentally missed this file in my first review because it was collapsed)

libs/pdi-scanner/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "author": "VotingWorks Eng <eng@voting.works>",
+  "main": "build/index.js",
+  "types": "build/index.d.js",
+  "files": [
+    "build"
+  ],
+  "scripts": {
+    "build": "pnpm install:rust-addon && pnpm build:rust-addon && pnpm build:ts",
+    "build:ts": "tsc --build tsconfig.build.json",
+    "build:rust-addon": "cargo build --release",
```
amcmanus, line 16: For trusted build purposes, we should add `--offline` to this build command, i.e. `cargo build --release --offline`

libs/pdi-scanner/src/rusb_async/error.rs
eventualbuddha: Well, we could:
1. Say we want to keep them exactly as-is but still vendor them.
2. Refer to the git commit as a crate dependency to again keep the original versions as-is.
3. Review the diff manually and then just treat them as "our" files, but still vendor them.
4. Fork `rusb` from the point of the commit and make the changes on our fork. This could either be essentially a git subtree or a separate repo.

I'm kinda inclined toward #3, but I don't have a strong preference.

libs/pdi-scanner/src/client.rs
eventualbuddha: I went with this approach initially because it minimized the sources of issues that could arise when building our own driver. To the extent we think we understand the interaction model I think we can move to customize this a bit. For now I'm inclined to leave it, but I don't feel strongly about it.

libs/pdi-scanner/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+    log_level: tracing::Level,
+}
+
+fn setup(config: &Config) -> color_eyre::Result<()> {
+    color_eyre::install()?;
+    setup_logging(config)?;
+    Ok(())
+}
+
+fn setup_logging(config: &Config) -> color_eyre::Result<()> {
```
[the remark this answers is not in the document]
eventualbuddha: The logging uses a very flexible crate called `tracing`, and we set it up in this function to go to `stderr` because we're otherwise using `stdout` as one half of a communication channel.

libs/pdi-scanner/src/main.rs
eventualbuddha, line 1: I'm not really enamored with this approach. I think it's worth having _you_ do an evaluation of whether this separate-binary-controlled-over-stdio or something using either `neon` or `napi-rs` would be better.

libs/pdi-scanner/src/scanner.rs
eventualbuddha: We could make a call that we always use one. In your debugging I think you found that canceling the pending transfers actually caused issues, right? So maybe we just don't.

libs/pdi-scanner/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+base64 = { workspace = true }
+bitter = { workspace = true }
+chrono = "0.4.31"
+clap = { workspace = true }
+color-eyre = { workspace = true }
+crc16 = { workspace = true }
+csv = { workspace = true }
+ctrlc = { workspace = true }
+crossterm = "0.27.0"
+image = { version = "0.24.7", default-features = false, features = [
```
eventualbuddha, line 27: Update to 0.25.1.
```suggestion
image = { version = "0.25.1", default-features = false, features = [
```

libs/pdi-scanner/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+    "bmp",
+    "jpeg",
+    "png",
+] }
+libc = { workspace = true }
+nom = "7.1.3"
+psutil = "3.3.0"
+rand = "0.8.5"
+ratatui = "0.23.0"
```
eventualbuddha, line 36: Update to 0.26.1.
```suggestion
ratatui = "0.26.1"
```

libs/pdi-scanner/src/main.rs
arsalansufi: My quick take is that we should at least get this PR merged with the current stdio approach and can then explore alternatives in follow-up PRs. Might already be what we're planning to do, just wanted to explicitly state!

libs/pdi-scanner/src/rusb_async/error.rs
jonahkagan: I agree option 3 sounds best. I briefly tried to diff the files when I first reviewed this PR but it was a bit complex and I couldn't immediately identify what changes you added (although I was just using `diff`, so there was no colors/syntax highlighting). If you remember and could give an overview, that might help. Or maybe I just need a better diffing tool

libs/pdi-scanner/src/client.rs
jonahkagan: Leaving it for now sounds good to me. I wonder if this is an area where PDI could provide some insight so we're not guessing as much?

libs/pdi-scanner/src/main.rs
eventualbuddha: Reading from stdin is a synchronous operation and does not have "read with timeout" semantics, so the loop below would be unable to process events coming from the scanner until something came in over stdin to wake it up.

libs/pdi-scanner/src/protocol/packets.rs
eventualbuddha: The answer to this is actually fairly instructive. You can write the function like so and it compiles fine as-is:

```rust
fn checked(command: Command, parser: fn(&[u8]) -> nom::IResult<&[u8], ()>) -> Vec<u8> {
    if let Ok(([], _)) = parser(&command.to_bytes()) {
        command.to_bytes()
    } else {
        panic!("Command {:?} failed to parse", command)
    }
}
```

However, as you start converting `checked!(…)` to `checked(…)` you notice that some of them become compile errors. This is because the type of `parser` doesn't match:

```
error[E0308]: mismatched types
   --> libs/pdi-scanner/src/protocol/packets.rs:622:17
    |
620 |             Self::SetSerialNumberRequest(serial_number) => checked(
    |                                                            ------- arguments to this function are incorrect
621 |                 Command::new(b"*").with_data(serial_number),
622 |                 parsers::set_serial_number_request,
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected fn pointer, found fn item
    |
    = note: expected fn pointer `for<'a> fn(&'a _) -> std::result::Result<(&'a _, ()), nom::Err<nom::error::Error<&'a _>>>`
                  found fn item `for<'a> fn(&'a _) -> std::result::Result<(&'a _, &'a [u8; 8]), nom::Err<nom::error::Error<&'a _>>> {set_serial_number_request}`
note: function defined here
   --> libs/pdi-scanner/src/protocol/packets.rs:582:4
    |
582 | fn checked(command: Command, parser: fn(&[u8]) -> nom::IResult<&[u8], ()>) -> Vec<u8> {
    |    ^^^^^^^                   --------------------------------------------
```

You might then try to make `checked` be generic, but then you get this error:

```
error[E0308]: mismatched types
   --> libs/pdi-scanner/src/protocol/packets.rs:622:17
    |
620 |             Self::SetSerialNumberRequest(serial_number) => checked(
    |                                                            ------- arguments to this function are incorrect
621 |                 Command::new(b"*").with_data(serial_number),
622 |                 parsers::set_serial_number_request,
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ one type is more general than the other
    |
    = note: expected fn pointer `for<'a> fn(&'a _) -> std::result::Result<(&'a _, _), nom::Err<nom::error::Error<&'a _>>>`
                  found fn item `for<'a> fn(&'a _) -> std::result::Result<(&'a _, &'a [u8; 8]), nom::Err<nom::error::Error<&'a _>>> {set_serial_number_request}`
note: function defined here
   --> libs/pdi-scanner/src/protocol/packets.rs:582:4
    |
582 | fn checked<T>(command: Command, parser: fn(&[u8]) -> nom::IResult<&[u8], T>) -> Vec<u8> {
    |    ^^^^^^^                      -------------------------------------------
```

Perhaps there is a way to get past this, but I'm not sure what it is.

libs/pdi-scanner/src/protocol/parsers.rs
eventualbuddha: The "Combined PS5 UltraScan Software Spec.pdf" file defines the response to the Request Scanner Status ('Q') command in section 2.1.51. It lists the first byte, bit 5 as being reserved for future use.

libs/pdi-scanner/src/protocol/parsers.rs
eventualbuddha: There are more, but it's not clear to me that they will be of use to us in this driver. I'll remove this comment.

libs/pdi-scanner/src/rusb_async/error.rs
eventualbuddha: I'm going to use the `difft` (difftastic) tool to remind myself what the differences are and then comment based on that.

libs/pdi-scanner/src/rusb_async/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+
+use std::convert::TryInto;
+use std::ptr::NonNull;
+
+use std::sync::atomic::{AtomicBool, Ordering};
+
+mod error;
+mod pool;
+
+pub use error::{Error, Result};
```
eventualbuddha, line 38: The original did not export this publicly, but we need them in order to use them with e.g. `match`.

libs/pdi-scanner/src/rusb_async/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+pub use pool::TransferPool;
+
+struct Transfer {
+    ptr: NonNull<ffi::libusb_transfer>,
+    buffer: Vec<u8>,
+}
+
+impl Transfer {
+    // Invariant: Caller must ensure `device` outlives this transfer
+    #[allow(unused)]
```
eventualbuddha, line 48: This was added to make `clippy` happy for each of the methods. In reality, we do use `bulk` but not the others so we could delete the others and remove this. I opted to keep it closer to the original instead.

libs/pdi-scanner/src/rusb_async/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+        value: u16,
+        index: u16,
+        data: &[u8],
+    ) -> Self {
+        let mut buf = Vec::with_capacity(data.len() + ffi::constants::LIBUSB_CONTROL_SETUP_SIZE);
+
+        let length = data.len() as u16;
+
+        ffi::libusb_fill_control_setup(
+            buf.as_mut_ptr(),
```
eventualbuddha, line 104: There was previously a ` as *mut u8` cast here, but that was a useless cast.

libs/pdi-scanner/src/rusb_async/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    unsafe fn control(
+        device: *mut ffi::libusb_device_handle,
+
+        request_type: u8,
+        request: u8,
+        value: u16,
+        index: u16,
+        data: &[u8],
+    ) -> Self {
+        let mut buf = Vec::with_capacity(data.len() + ffi::constants::LIBUSB_CONTROL_SETUP_SIZE);
```
eventualbuddha, line 99: Rather than importing all the constants individually I just reference them through the namespace.

libs/pdi-scanner/src/rusb_async/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+        Err(err)
+    }
+}
+
+/// Invariant: transfer must not be pending
+impl Drop for Transfer {
+    fn drop(&mut self) {
+        unsafe {
+            drop(Box::from_raw(
+                self.transfer().user_data.cast::<AtomicBool>(),
```
eventualbuddha, line 319: Added this `.cast::<AtomicBool>()`, presumably because `rustc` wanted it.

libs/pdi-scanner/src/rusb_async/error.rs
eventualbuddha, line 1: This file changed from a manual `fmt::Display` implementation to using `thiserror`:

image

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,269 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// Represents a pool of asynchronous transfers, that can be polled to completion
+pub struct TransferPool<C: UsbContext> {
+    device: Arc<DeviceHandle<C>>,
+    pending: VecDeque<Transfer>,
+}
+
+impl<C: UsbContext> TransferPool<C> {
+    pub fn new(device: Arc<DeviceHandle<C>>) -> Self {
```
eventualbuddha, line 46: This previously returned a `Result` but it was always `Ok`:

```rust
    pub fn new(device: Arc<DeviceHandle<C>>) -> Result<Self> {
        Ok(Self {
            device,
            pending: VecDeque::new(),
        })
    }
```

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,269 @@
...
+        Self {
+            device,
+            pending: VecDeque::new(),
+        }
+    }
+
+    #[allow(unused)]
```
eventualbuddha, line 53: Same directives to make clippy happy as in `mod.rs`.

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,269 @@
...
+        // dropped before `device` is freed.
+        unsafe {
+            let mut transfer = Transfer::iso(self.device.as_raw(), endpoint, buf, iso_packets);
+            transfer.submit()?;
+            self.pending.push_back(transfer);
+            Ok(())
+        }
+    }
+
+    pub fn poll(&mut self, timeout: Duration) -> Result<CompletedTransfer> {
```
eventualbuddha, line 127: This used to just return a `Result<Vec<u8>>`, but that meant I couldn't identify which endpoint the data came from. The `CompletedTransfer` struct gives me both the data and the endpoint.

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,269 @@
...
+            let mut transfer = self.pending.pop_front().unwrap();
+            let endpoint = transfer.transfer().endpoint;
+            let data = transfer.handle_completed()?;
+            Ok(CompletedTransfer::new(endpoint, data))
+        } else {
+            Err(Error::PollTimeout)
+        }
+    }
+
+    pub fn poll_endpoint(&mut self, endpoint: u8, timeout: Duration) -> Result<Vec<u8>> {
```
eventualbuddha, line 139: New function to make polling by an endpoint easier.

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /// Returns the number of async transfers pending
+    pub fn pending(&self) -> usize {
+        self.pending.len()
+    }
+}
+
+unsafe impl<C: UsbContext> Send for TransferPool<C> {}
+// unsafe impl<C: UsbContext> Sync for TransferPool<C> {}
```
eventualbuddha, line 178: This wasn't commented out in the original, but I think I ran into problems with it so I removed it? I don't really remember.

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,269 @@
...
+        &self.data
+    }
+
+    #[allow(unused)]
+    pub fn into_vec(self) -> Vec<u8> {
+        self.data
+    }
+}
+
+/// This is effectively `libusb_handle_events_timeout_completed`, but with
```
eventualbuddha, line 216: A couple markdown fixes were made in this documentation comment.

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,269 @@
...
+///
+/// This design is based on
+/// [this](https://libusb.sourceforge.io/api-1.0/libusb_mtasync.html#threadwait).
+///
+/// Returns `true` when `completed` becomes true, `false` on timeout, and panics on
+/// any other libusb error.
+fn poll_completed(ctx: &impl UsbContext, timeout: Duration, completed: &AtomicBool) -> bool {
+    let deadline = Instant::now() + timeout;
+
+    let mut err = ffi::constants::LIBUSB_SUCCESS;
```
eventualbuddha, line 229: Reference the appropriate constant rather than hardcoding 0.

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,269 @@
...
+        let timeval = libc::timeval {
+            tv_sec: remaining.as_secs().try_into().unwrap(),
+            tv_usec: remaining.subsec_micros().into(),
+        };
+        unsafe {
+            if ffi::libusb_try_lock_events(ctx.as_raw()) == ffi::constants::LIBUSB_SUCCESS {
+                if !completed.load(Ordering::SeqCst)
+                    && ffi::libusb_event_handling_ok(ctx.as_raw()) != ffi::constants::LIBUSB_SUCCESS
+                {
+                    err = ffi::libusb_handle_events_locked(ctx.as_raw(), addr_of!(timeval));
```
eventualbuddha, line 244: Per clippy, use `addr_of!` rather than casting as `*const _`.

libs/pdi-scanner/src/scanner.rs
eventualbuddha: I changed this around a bit so that `connect` no longer returns a `Scanner` alongside the `Client`. Instead, the `Client` holds a reference to the scanner and `Scanner` will call `stop` when it is dropped.

libs/pdi-scanner/src/client.rs
eventualbuddha: Probably so, yeah.

libs/pdi-scanner/src/scanner.rs
eventualbuddha: Changed this to not have `StopMode`.

eventualbuddha:
@jonahkagan I made some updates. PTAL?

libs/pdi-scanner/src/main.rs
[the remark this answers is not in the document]
jonahkagan: Could you add some instructions for turning on the debug logging to the README?

libs/pdi-scanner/src/protocol/packets.rs
jonahkagan: Interesting! Thanks for walking me through it. Doesn't seem worth digging deeper to me

libs/pdi-scanner/src/protocol/parsers.rs
jonahkagan: Ah, I see. Thanks for the pointer. Maybe a comment here would be nice to help remember

libs/pdi-scanner/src/rusb_async/pool.rs
```diff
@@ -0,0 +1,261 @@
...
+        } else {
+            Err(Error::PollTimeout)
+        }
+    }
+
+    pub fn poll_endpoint(&mut self, endpoint: u8, timeout: Duration) -> Result<Vec<u8>> {
+        for (index, transfer) in self.pending.iter_mut().enumerate() {
+            if transfer.transfer().endpoint == endpoint {
+                if poll_completed(self.device.context(), timeout, transfer.completed_flag()) {
+                    let mut transfer = self.pending.remove(index).unwrap();
```
jonahkagan, line 143: Is it safe to remove from `pending` while iterating over it?

jonahkagan:
@eventualbuddha I left a couple more comments about adding documentation/comments, and also one code question. Looks good to me generally! I appreciate your help in cleaning this up!

eventualbuddha (approved):

PR #4734, opened 2024-03-29
chore(rust): run `cargo update`

eventualbuddha:
## Overview
Updates dependencies in a manner compatible with the versions specified in `Cargo.toml` files. (See [docs](https://doc.rust-lang.org/cargo/commands/cargo-update.html)).

### Why?
This PR only updates the versions of packages that `cargo` decides are semver-compatible with the versions specified in `Cargo.toml` files transitively. If we assume that care was taken in choosing the versions appropriately, I believe the risk of this is low. It is of course possible that an update could have a change that unintentionally breaks our code. In the worst case, this breakage is subtle and isn't caught by our tests. This risk has to be weighed against the risk of stagnation.

In the Rust ecosystem, the emphasis on staying up to date is even stronger than many others. Many crates have policies that look something like "we support the last N minor revisions of Rust", where N is often less than 10.  For context, 10 minor versions from the current 1.77.0 is 1.67.0, which was released on January 2023. This MSRV (minimum supported Rust version, a per-crate config) is typically ratcheted up over time, so you need to keep both your Rust version and crate dependencies updated with some frequency to avoid a situation where your workspace becomes "stale" and you can no longer install or update crates without looking for a compatible older version that supports your older version of Rust and your existing set of crate dependencies.

This dynamic exists in the NodeJS ecosystem but is often much less codified, with packages failing to declare which versions of NodeJS they support in a tooling-readable way.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,9 +25,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.6"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
LGTM! Thanks for all the notes on the Rust ecosystem's update philosophy

PR #4735, opened 2024-03-29
VxMarkScan - Save Diagnostics Report

### LINKED ISSUE #4721: vxmarkscan: allow exporting diagnostics report to USB
Allow VxMarkScan readiness report to be saved to an attached USB. I think this will be easier than involving the print flow.

adghayes:
## Overview

Closes #4721.

## Demo Video or Screenshot

Screen Shot 2024-03-29 at 9 26 58 AM

## Checklist


adghayes:
@kofi-q I'll be out for two weeks when you review this, feel free to go ahead and merge, make tweaks, or ignore it if there's too much to change.

apps/mark-scan/backend/src/app.diagnostics.test.ts
```diff
@@ -127,12 +136,82 @@ test('getApplicationDiskSpaceSummary', async () => {
...
   );
 });
+
+const reportPrintedTime = new Date('2021-01-01T00:00:00.000');
+jest.mock('./util/get_current_time', () => ({
+  getCurrentTime: () => reportPrintedTime.getTime(),
+}));
+
+test('saving the readiness report', async () => {
+  jest.useFakeTimers().setSystemTime(reportPrintedTime.getTime());
```
kofi-q, line 155: nit: Not clear why we need both this jest time mock and a mock of the `get_current_time` module - does this not affect the result of `Date.now()`?

kofi-q (approved):

apps/mark-scan/backend/src/app.diagnostics.test.ts
adghayes: It does, but there's something in our mocks that does not play nicely with mocking `Date.now()` - the test will freeze and never complete. That is why `getCurrentTime` exists at all - because mocking with `jest` breaks our tests. But mocking during certain set up actions is OK. 

It's been a while since I've looked at this - ideally I would have a better understanding of what exactly is hanging when using fake timers. I'm just going to merge this and make a ticket for the question, since it's not worth me digging all the way into now.

adghayes: It looks like the problem is that `playwright` can't properly launch Chromium when fake timers are running, so anytime we use `libs/printing` for rendering, we can't be running fake timers.

PR #4738, opened 2024-04-02
fix(mark-scan): `clap` usage fixes

eventualbuddha:
- use doc comments (triple-slash) on `clap::Parser` struct fields to have them show up both in intellisense and in `--help` output
- unconditionally run `Args::parse()` early to allow handling of `--help` and `--version` flags

```
Usage: patinputd [OPTIONS]

Options:
  -s, --skip-hardware-check  Allow the daemon to run if no hardware is found
  -h, --help                 Print help
  -V, --version              Print version
```


kshen0 (approved):
Thanks for the fixes

Issue #4623, opened 2024-02-27
vxmarkscan: ATI onboarding / help


mattroe:
VxMarkScan provides general on-screen guidance for a visual-touch user and also a guided device-setup flow for a dual-switch/limited-dexterity user. The audio-tactile interface experiences leaves much more to be desired.

The audio interface should initially provide audio instructions for how to use the tactile controller. Specifically it should instruct a user how to:
- Navigate and select contests
- Change volume
- Change rate of audio
- Pause audio
- Repeat content
- Seek more help with the controller

The help button on the controller should be remapped to an optional interactive flow (see below). As discussed, it's acceptable for the means to repeat content to be simply navigating "back" to the contest - as long as the means to do so is communicated. We do not need and explicit "replay" action on the audio tactile controller.

The interactive help feature should prompt the user to touch the controls on the tactile controller to learn their functions. When pressed, the audio track should state their function. A graphic of the controller can be shown on screen during this function. A user can be prompted to exit the help menu by clicking the help button again.

Slack context: https://votingworks.slack.com/archives/CJU9MSC6S/p1709052941623499

kshen0:
Seems like @adghayes handled this in https://github.com/votingworks/vxsuite/pull/4676

PR #4740, opened 2024-04-03
[libs/ui] Consolidate keybindings across apps into shared definition

kofi-q:
## Overview

Related to: https://github.com/votingworks/vxsuite/issues/4623

Moving navigation, language, and audio keybindings to a shared location in libs/ui in preparation for use in the BMD accessible controller help flow.

Also fixing a couple controller navigation bugs:
- Focus navigation on the VxMark "Review" screen was previously broken - this fixes the query selectors used to find focusable elements.
- Previously, voters could accidentally navigate to a different contest while an overvote modal or write-in modal was active - this fixes the page navigation logic to no-op when a modal is active.

## Demo Video or Screenshot
Also includes updates to the `MarkScanControllerIllustration` to use theme-aware styling for the interactive states - prototype of the help flow here:
https://drive.google.com/file/d/1nsAstb3RIabvBU6FgqhFOr2pH-Y4h1-K/view?usp=drive_link

## Testing Plan
- New tests for accessible navigation functions, updated relevant existing tests.

## Checklist

### DIFF
--- a/apps/mark/frontend/src/lib/gamepad.ts
+++ b/apps/mark/frontend/src/lib/gamepad.ts
@@ -1,81 +1,31 @@
 import { Button } from 'react-gamepad';
-import { mod } from '../utils/mod';
-
-export function getActiveElement(): HTMLElement {
-  return document.activeElement as HTMLElement;
-}
-
-function getFocusableElements(): HTMLElement[] {
-  const tabbableElements = Array.from(
-    document.querySelectorAll(
-      'button:not([aria-hidden="true"]):not([disabled])'
-    )
-  );
-  const ariaHiddenTabbableElements = Array.from(
-    document.querySelectorAll('[aria-hidden="true"] button')
-  );
-  return tabbableElements.filter(
-    (element) => ariaHiddenTabbableElements.indexOf(element) === -1
-  ) as HTMLElement[];
-}
-
-function handleArrowUp() {
-  const focusable = getFocusableElements();
-  const currentIndex = focusable.indexOf(getActiveElement());
-  /* istanbul ignore else */
-  if (focusable.length) {
-    if (currentIndex > -1) {
-      focusable[mod(currentIndex - 1, focusable.length)].focus();
-    } else {
-      focusable[focusable.length - 1].focus();
-    }
-  }
-}
-
-function handleArrowDown() {
-  const focusable = getFocusableElements();
-  const currentIndex = focusable.indexOf(getActiveElement());
-  /* istanbul ignore else */
-  if (focusable.length) {
-    focusable[mod(currentIndex + 1, focusable.length)].focus();
-  }
-}
-
-function handleArrowLeft() {
-  const prevButton = document.getElementById('previous');
-  /* istanbul ignore else */
-  if (prevButton) {
-    prevButton.click();
-  }
-}
-
-function handleArrowRight() {
-  const nextButton = document.getElementById('next');
-  /* istanbul ignore else */
-  if (nextButton) {
-    nextButton.click();
-  }
-}
+import {
+  Keybinding,
+  PageNavigationButtonId,
+  advanceElementFocus,
+  triggerPageNavigationButton,
+} from '@votingworks/ui';
 
 function handleClick() {
-  const activeElement = getActiveElement();
-  activeElement.click();
+  if (document.activeElement instanceof HTMLElement) {
+    document.activeElement.click();
+  }
 }
 
 export function handleGamepadButtonDown(buttonName: Button): void {
   switch (buttonName) {
     case 'DPadUp':
-      handleArrowUp();
+      advanceElementFocus(-1);
       break;
     case 'B':
     case 'DPadDown':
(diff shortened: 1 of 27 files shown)


libs/ui/src/accessible_controllers/mark_scan_controller_illustration.tsx
```diff
@@ -1,42 +1,69 @@
...
 
 interface MarkScanControllerIllustrationProps {
   highlight?: MarkScanControllerButton;
 }
 
-const HIGHLIGHT_FILL = '#885fce';
-export const MARK_SCAN_CONTROLLER_ILLUSTRATION_HIGHLIGHT_FILL = HIGHLIGHT_FILL;
+const BUTTON_CLASS_NAME = 'markScanControllerIllustration--Button';
+const BUTTON_CLASS_NAME_HIGHLIGHTED = `${BUTTON_CLASS_NAME}--highlighted`;
+const BUTTON_FOREGROUND_CLASS_NAME = `${BUTTON_CLASS_NAME}Foreground`;
```
kshen0, line 12: any reason to not delineate with `--` like the other class names?

libs/ui/src/keybindings.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Keyboard keybindings for VxSuite apps.
+ *
+ * These are mapped to features/behaviors in various apps (some triggered via
+ * accessible controllers) and are consolidated here to encourage consistency
+ * and avoid accidental collisions.
+ */
```
kshen0, line 7: 🎉

libs/ui/src/accessible_controllers/navigation.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Simulates browser `Tab`/`Shift+Tab` behavior by moving focus to the next
+ * focusable element in the document, in the specified {@link direction}.
+ */
+export function advanceElementFocus(direction: 1 | -1): void {
+  const hiddenTabEnabledElements = getHiddenTabEnabledElements();
+
+  const focusableElements = Array.from(
+    document.querySelectorAll<HTMLElement>(FOCUSABLE_ELEMENT_SELECTORS)
+  ).filter((e) => !hiddenTabEnabledElements.has(e));
```
kshen0, line 31: Are `hiddenTabEnabledElements` already filtered by the `not([aria-hidden="true"])` clauses in `FOCUSABLE_ELEMENT_SELECTORS`?

apps/mark/frontend/src/lib/gamepad.ts
```diff
@@ -88,23 +38,23 @@ export function handleGamepadButtonDown(buttonName: Button): void {
...
+    case Keybinding.PAGE_PREVIOUS:
+      triggerPageNavigationButton(PageNavigationButtonId.PREVIOUS);
       break;
-    case 'ArrowRight':
-      handleArrowRight();
+    case Keybinding.PAGE_NEXT:
+      triggerPageNavigationButton(PageNavigationButtonId.NEXT);
       break;
-    case ']':
+    case Keybinding.LEGACY_PAT_SELECT:
```
kshen0, line 54: Tracking updating the daemon in https://github.com/votingworks/vxsuite/issues/4742 so we can get rid of the `LEGACY` keybindings later.

(Now that I think about it, we probably don't need to handle PAT input in VxMark because there is no PAT input on those machines, but the comment stands for `mark-scan/.../assistive_technology`)

libs/ui/src/accessible_controllers/navigation.ts
kofi-q: This could probably use a more explicit naming - it's more along the lines of `tabEnabledElementsInHiddenBlocks`, which don't get caught by the `not([aria-hidden="true"])` and have to be queried via an ancestor query: `[aria-hidden="true"] button`

Will update naming and add clarifying comments to avoid confusion

apps/mark/frontend/src/lib/gamepad.ts
kofi-q: Ah, perfect! Would be great to standardise on the `FOCUS_NEXT` and `SELECT` keybindings - wasn't sure if we had control over the key mapping for the PAT input.

And thanks for the heads-up on VxMark - will drop the legacy bindings altogether if there's no PAT support there anyway

libs/ui/src/accessible_controllers/mark_scan_controller_illustration.tsx
kofi-q: Ah, didn't mean to have the `--` in the `BUTTON_CLASS_NAME` definition - should be `markScanControllerIllustrationButton` there, whereas the `--` suffixes are intended to signify style variations for the base class. Will fix

libs/ui/src/accessible_controllers/navigation.ts
kshen0: Thanks for clarifying!

kshen0 (approved):

PR #4746, opened 2024-04-04
[MarkScan] Add the accessible controller help flow

kofi-q:
## Overview

Resolves https://github.com/votingworks/vxsuite/issues/4623

Adding an accessible controller sandbox screen that allows the user to try out the various buttons without triggering the default behaviour - instead, they'll receive audio feedback on what each button does.

The sandbox will be triggered by the "help" button (currently mapped to the old "replay" keybinding, `Shift+R` -- we can clean that up later) and can be exited with another two successive presses on the "help" button (the first press describes the button and tells the voter they can press it again to exit).

A little long, so recorded with increased speech rate to speed things up a bit:

https://github.com/votingworks/vxsuite/assets/264902/cc3ffc23-c8a5-4d7d-8bad-2cf38cdf0395

## Testing Plan
- New unit tests for the sandbox component and trigger hook
- Broke out a `VoterFlow` component from `app_root` to make testing easier and make it easier to scope the trigger to just voter sessions 

## Checklist

### DIFF
--- a/libs/ui/src/accessible_controllers/accessible_controller_sandbox.test.tsx
+++ b/libs/ui/src/accessible_controllers/accessible_controller_sandbox.test.tsx
@@ -1,0 +1,156 @@
+import {
+  advanceTimers,
+  fakeUseAudioControls,
+  mockOf,
+} from '@votingworks/test-utils';
+import React from 'react';
+import { assert } from '@votingworks/basics';
+import userEvent from '@testing-library/user-event';
+import { UiString } from '../ui_strings/ui_string';
+import {
+  AccessibleControllerHelpStrings,
+  AccessibleControllerIllustrationProps,
+  AccessibleControllerSandbox,
+} from './accessible_controller_sandbox';
+import { act, screen } from '../../test/react_testing_library';
+import { Keybinding } from '../keybindings';
+import { newTestContext } from '../../test/test_context';
+
+const mockAudioControls = fakeUseAudioControls();
+
+jest.mock(
+  '../ui_strings/ui_string',
+  (): typeof import('../ui_strings/ui_string') => ({
+    ...jest.requireActual('../ui_strings/ui_string'),
+    UiString: jest.fn(),
+  })
+);
+
+jest.mock(
+  '../hooks/use_audio_controls',
+  (): typeof import('../hooks/use_audio_controls') => ({
+    ...jest.requireActual('../hooks/use_audio_controls'),
+    useAudioControls: () => mockAudioControls,
+  })
+);
+
+function newRenderer() {
+  let lastReadElement: HTMLElement;
+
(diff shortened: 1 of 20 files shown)


kshen0:
Nice! Button descriptions sound good. Should the pause button also say it can be pressed again to resume playback? There's no `Play` button so the behavior might be slightly unintuitive.

libs/ui/src/accessible_controllers/accessible_controller_sandbox.tsx
```diff
@@ -0,0 +1,92 @@
...
+      const key = event.key as T;
+      const audioStringKey = feedbackStringKeys[key];
+      if (!audioStringKey) {
+        return;
+      }
+
+      // Waiting to update the audio string key in the next tick allows the
+      // previous state to get cleared and for the screen reader to stop before
+      // triggering new audio. This enables repeated audio feedback on the
+      // repeated pressed of the same button.
```
kshen0, line 62: Typo? Did you mean something like
> This enables repeated audio feedback on the repeated pressing of the same button

libs/ui/src/accessible_controllers/accessible_controller_sandbox.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  return {
+    expectLastMockScreenReaderContent: (textContent: string) =>
+      expect(lastReadElement).toHaveTextContent(textContent),
+    renderWithMockScreenReader,
+  };
+}
+
+function simulateKeyPress(key: string) {
+  userEvent.keyboard(key.length === 1 ? key : `[${key}]`);
```
kshen0, line 59: nit: factor out for DRYness? Similar implementation to https://github.com/votingworks/vxsuite/pull/4740/files#diff-4d3153775fb1fec32f94bd89603d0bcf0828ed2d79f72cefcc61d807dc23b5d6R26

kshen0 (approved):

libs/ui/src/accessible_controllers/accessible_controller_sandbox.tsx
kofi-q: Thanks for catching! Yeah, meant to write `repeated audio feedback on repeated presses`

libs/ui/src/accessible_controllers/accessible_controller_sandbox.test.tsx
kofi-q: +1 - will do

kofi-q (commented):
> Should the pause button also say it can be pressed again to resume playback?

Ah, good catch - I accidentally cut that audio short in the video. The full audio text mentions the unpause functionality: https://github.com/votingworks/vxsuite/blob/main/libs/ui/src/ui_strings/app_strings.tsx#L244

PR #4747, opened 2024-04-04
chore: remove `@votingworks/ballot-encoder` where it's unused

eventualbuddha:
Some cleanup I did this morning while waiting for the power to come back on.


kshen0 (approved):
Not super familiar with `ballot-encoder`, but I'm assuming it's used now only in VxDesign.

eventualbuddha:
It's also used in decoding, but only in the backends.

PR #4748, opened 2024-04-05
Try GPIO connection with address offset

kshen0:
## Overview
* Adds `Pin` trait and implements for `GpioPin` and `MockPin`
* Adds `safe_export` function that, on `export` error, will try to `unexport` and `export` the pin. This solves for the case where the daemon crashes and is restarted without cleanup, leaving pins exported. Also adds convenience when developer has exported the pin manually.
* Adds `PatInputReader` to separate pins from daemon logic
* Attempts to connect to `pin address + 512` first

VxMarkScan machines running prod images fail to connect to the PAT input GPIO pins. After some experimentation I found the pins were addressed at an offset of 512 eg. pin 478 is addressable instead at 990.

I think this is because pin address assignment may differ between kernel versions.

Long term we should move this daemon to use `libpiod` and the character device API, the accepted replacement for `/sys/class/gpio`, but that's a larger undertaking (more details [here](https://docs.google.com/document/d/1b8o34e7Rseb8L8eHbtOqe34cdlz5tCTENPZw_v-OxH4/edit)).

## Demo Video or Screenshot

Exporting new pin addresses

IMG_5041

Unexport/re-export when pin is already exported
IMG_5042

Manual testing of existing behavior. Note some logging in this video is out of date; see screenshots above for most current logging.

https://github.com/votingworks/vxsuite/assets/1060688/93f0eced-e466-4176-acd0-284e03dffe8e

## Testing Plan
- [x] validate against hardware
- [x] add automated tests

## Checklist


apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,73 @@
...
+const SIGNAL_B_PIN_ADDRESS: u16 = 476;
+
+pub struct PatInputReader {
+    is_connected_pin: Pin,
+    signal_a_pin: Pin,
+    signal_b_pin: Pin,
+}
+
+impl PatInputReader {
+    pub fn new() -> Result<Self, io::Error> {
```
kshen0, line 26: I was skeptical of returning `Result` from a constructor but it seems to be an accepted pattern per https://users.rust-lang.org/t/idiomatic-error-handling-in-new-and-builder-pattern/86361 and https://users.rust-lang.org/t/should-i-return-a-result-from-new/17372

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -53,6 +53,16 @@ impl Pin {
             .expect("Unexpected failure to tear down pin");
     }
 
+    /// Attempts to export the pin. If successful, cleans up by unexporting
+    /// the pin and returning an empty Result. Returns an error if the pin
+    /// does not exist, the program lacks permission to export, or an
+    /// unexpected error occurs.
```
kshen0, line 59: Sort of self-explanatory but I wanted to enumerate the expected vs unexpected errors

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -113,7 +103,57 @@ impl Pin {
...
+
+    // Unexports the pin, removing access from userspace.
+    fn unexport(&self) -> io::Result<()> {
+        fs::write(UNEXPORT_PIN_FILEPATH, self.to_string())
+    }
+
+    // Sets the pin direction to "in" so its value is readable. Must be called after
+    // `export`
+    fn set_direction_in(&self) -> io::Result<()> {
+        let filepath = format!("/sys/class/gpio/gpio{self}/direction");
```
eventualbuddha, line 151: It's not obvious that the `self.fmt(…)` result should be interpolated into the file path without knowing that its implementation just uses `self.address`.

```suggestion
        let filepath = format!("/sys/class/gpio/gpio{}/direction", self.address);
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+            signal_b_pin: None,
+        }
+    }
+
+    pub fn connect(&mut self) -> Result<(), io::Error> {
+        // Prod is expected to use the offset so attempt with offset first.
+        let mut offset: u16 = POSSIBLE_PIN_OFFSET;
+
+        if let Some(probe_pin) = &self.is_connected_pin {
+            // let probe_pin_inner = probe_pin;
```
eventualbuddha, line 39:
```suggestion
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+            // Earlier kernel versions may succeed without the offset.
+            if let Err(err) = probe_pin.probe() {
+                log!(
+                    EventId::Info,
+                    "Failed to connect to pin {} with error {}. Retrying without offset.",
+                    probe_pin,
+                    err
```
eventualbuddha, line 46:
```suggestion
                    "Failed to connect to pin {probe_pin} with error {err}. Retrying without offset."
```

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -113,7 +103,57 @@ impl Pin {
...
+            );
+
+            self.unexport()?;
+            self.export()?;
+        }
+
+        log!(
+            event_id: EventId::ConnectToGpioPinComplete,
+            disposition: vx_logging::Disposition::Success,
+            message: format!("Successfully exported pin {}", self)
```
eventualbuddha, line 129:
```suggestion
            message: format!("Successfully exported pin {self}")
```

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -113,7 +103,57 @@ impl Pin {
...
+        );
+        Ok(())
+    }
+
+    // Exports the pin to be globally accessible from userspace. Returns an error if the
+    // pin has already been exported.
```
eventualbuddha, line 135:
```suggestion
    /// Exports the pin to be globally accessible from userspace. Returns an error if the
    /// pin has already been exported.
```

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -113,7 +103,57 @@ impl Pin {
...
+    fn export(&self) -> io::Result<()> {
+        fs::write(EXPORT_PIN_FILEPATH, self.to_string())?;
+        // Without this delay subsequent pin operations may fail with a permission error
+        sleep(EXPORT_DELAY);
+        Ok(())
+    }
+
+    // Unexports the pin, removing access from userspace.
```
eventualbuddha, line 143:
```suggestion
    /// Unexports the pin, removing access from userspace.
```

apps/mark-scan/pat-device-input/src/pin.rs
```diff
@@ -113,7 +103,57 @@ impl Pin {
...
+    fn unexport(&self) -> io::Result<()> {
+        fs::write(UNEXPORT_PIN_FILEPATH, self.to_string())
+    }
+
+    // Sets the pin direction to "in" so its value is readable. Must be called after
+    // `export`
```
eventualbuddha, line 149:
```suggestion
    /// Sets the pin direction to "in" so its value is readable. Must be called after
    /// `export`
```

libs/monorepo-utils/src/cargo.ts
```diff
@@ -2,8 +2,8 @@ import { assertDefined, lines } from '@votingworks/basics';
 import { spawn } from 'node:child_process';
 import { getAbsoluteRootPath } from './dependencies';
 
-// patinputd has no tests currently
-const EXCLUDED_PACKAGE_IDS = ['patinputd'];
+// Add untested packages here
+const EXCLUDED_PACKAGE_IDS: string[] = [];
```
eventualbuddha, line 6: Just remove this construct?

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+mod tests {
+    use vx_logging::set_app_name;
+
+    use super::*;
+
+    fn get_default_mocked_reader() -> PatInputReader<MockPin> {
+        get_mocked_reader(None)
+    }
+
+    fn get_mocked_reader(probe_err: Option<io::Error>) -> PatInputReader<MockPin> {
```
eventualbuddha, line 162:
```suggestion
    fn get_mocked_reader(connection_error: Option<io::Error>) -> PatInputReader<MockPin> {
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    pub fn is_connected(&self) -> bool {
+        match &self.is_connected_pin {
+            Some(pin) => {
+                return pin.is_active();
+            }
+            None => {
+                return false;
+            }
+        }
```
eventualbuddha, line 86:
```suggestion
        match &self.is_connected_pin {
            Some(pin) => pin.is_active(),
            None => false,
        }
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    pub fn is_signal_a_active(&self) -> bool {
+        match &self.signal_a_pin {
+            Some(pin) => {
+                return pin.is_active();
+            }
+            None => {
+                return false;
+            }
+        }
```
eventualbuddha, line 97:
```suggestion
        match &self.signal_a_pin {
            Some(pin) => pin.is_active(),
            None => false,
        }
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    pub fn is_signal_b_active(&self) -> bool {
+        match &self.signal_b_pin {
+            Some(pin) => {
+                return pin.is_active();
+            }
+            None => {
+                return false;
+            }
+        }
```
eventualbuddha, line 108:
```suggestion
        match &self.signal_b_pin {
            Some(pin) => pin.is_active(),
            None => false,
        }
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+            IS_CONNECTED_PIN_ADDRESS
+        );
+    }
+
+    #[test]
+    fn test_is_connected() {
+        let mut reader = get_default_mocked_reader();
+
+        assert_eq!(reader.is_connected(), false);
+        reader.is_connected_pin.as_mut().unwrap().active = true;
```
eventualbuddha, line 210:
```suggestion
        assert!(!reader.is_connected());
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+        assert_eq!(reader.is_connected(), true);
+    }
```
eventualbuddha, line 212:
```suggestion
        assert!(reader.is_connected());
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    #[test]
+    fn test_signal_a() {
+        let mut reader = get_default_mocked_reader();
+
+        assert_eq!(reader.is_signal_a_active(), false);
+        reader.signal_a_pin.as_mut().unwrap().active = true;
```
eventualbuddha, line 219:
```suggestion
        assert!(!reader.is_signal_a_active());
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+        assert_eq!(reader.is_signal_a_active(), true);
+    }
```
eventualbuddha, line 221:
```suggestion
        assert!(reader.is_signal_a_active());
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    #[test]
+    fn test_signal_b() {
+        let mut reader = get_default_mocked_reader();
+
+        assert_eq!(reader.is_signal_b_active(), false);
+        reader.signal_b_pin.as_mut().unwrap().active = true;
+        assert_eq!(reader.is_signal_b_active(), true);
+    }
```
eventualbuddha, line 230:
```suggestion
        assert!(reader.is_signal_b_active());
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+        reader.signal_a_pin.as_mut().unwrap().active = true;
+        assert_eq!(reader.is_signal_a_active(), true);
+    }
+
+    #[test]
+    fn test_signal_b() {
+        let mut reader = get_default_mocked_reader();
+
+        assert_eq!(reader.is_signal_b_active(), false);
+        reader.signal_b_pin.as_mut().unwrap().active = true;
```
eventualbuddha, line 228:
```suggestion
        assert!(!reader.is_signal_b_active());
```

eventualbuddha (approved):

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
kshen0: I think this suggestion is off by a line but will incorporate manually

eventualbuddha: Grr. Blame VS Code.

libs/monorepo-utils/src/cargo.ts
kshen0: Seemed useful to keep around for experimental or pre-prod packages but no strong opinion. I can remove it.

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+pub struct PatInputReader<T: Pin> {
+    is_connected_pin: Option<T>,
+    signal_a_pin: Option<T>,
+    signal_b_pin: Option<T>,
+}
+
+impl<T: Pin> PatInputReader<T> {
+    pub fn new() -> Self {
+        Self {
+            is_connected_pin: Some(T::new(IS_CONNECTED_PIN_ADDRESS + POSSIBLE_PIN_OFFSET)),
```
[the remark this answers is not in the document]
kshen0: Agreed, I liked that the default (of using the offset) is explicit in `new` but it is weird to put that assumption in 2 separate places. Will update.

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -0,0 +1,231 @@
...
+            signal_a_pin: None,
+            signal_b_pin: None,
+        }
+    }
+
+    pub fn connect(&mut self) -> Result<(), io::Error> {
+        // Prod is expected to use the offset so attempt with offset first.
+        let mut offset: u16 = POSSIBLE_PIN_OFFSET;
```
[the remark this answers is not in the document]
kshen0: `probe` is just a way to recover if the pin doesn't exist. If we fail the `offset = 0` case then we don't need to recover. In that case we want to bubble the error up from L52

```
is_connected_pin.set_up()?;
```

But thinking about it more, this behavior isn't as clear to the reader as probing in a loop and returning an error if both probes fail. I'll update this to fail in a clearer way.

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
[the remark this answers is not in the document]
kshen0: Ah I remember now. In `test_connect_offset` we need `offset = 512` to fail so it tries `offset = 0`. We do this by making `reader.connect()` probe `self.is_connected_pin`. Then in the test we set `is_connected_pin` to a `MockPin` that fails on `probe()`. Dependency injection works here because `connect` assumes `new` has already set up `is_connected_pin` for its first attempt. If we make `is_connected_pin = None` to start and instead `connect()` just calls `T::new` for a function-scoped `Pin` to probe, then we need to be able to control the mock without having a reference to the `Pin`.

I'm not sure if there's a preferred way to do this without dependency injection. I could create a third implementer of `Pin`, eg `MockZeroOffsetPin`, whose `probe()` fails specifically for `address = 990` and have the test use a `PatInputReader<MockZeroOffsetPin>`. But that seems oddly specific. Or I could make a `PinFactory` layer that `reader` uses to make `Pins`, like (pseudocode):

```
trait PinFactory {
  create(&self) -> Pin;
}

impl RealPinFactory for PinFactory {
  fn create(&self, addr: u16) -> GpioPin {
    GpioPin::new(addr)
  }

struct TestPinFactory {
  addrs_that_fail: [u16]
}

impl TestPinFactory for PinFactory {
  fn create(&self, addr: u16) -> MockPin {
    if self.addrs_that_fail.contains(addr) {
      MockPin { address: addr, active: false, probe_error: Some(io::Error::new(...)) }
    }
    MockPin { address: addr, active: false, probe_error: None }
  }
}
```

but this fundamentally has the same clunkiness of creating one or more mock implementations for a single test. Will think more on how to do this tomorrow.

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -25,39 +25,51 @@ pub struct PatInputReader<T: Pin> {
...
-
-        self.is_connected_pin = Some(is_connected_pin);
-        self.signal_a_pin = Some(signal_a_pin);
-        self.signal_b_pin = Some(signal_b_pin);
-
-        Ok(())
+        match offset {
+            None => Err(io::Error::new(
+                io::ErrorKind::ConnectionRefused,
+                format!("Unable to connect to probe pins at offsets: {:?}", offsets),
```
eventualbuddha, line 56:
```suggestion
                format!("Unable to connect to probe pins at offsets: {offsets:?}"),
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -134,83 +146,125 @@ impl fmt::Display for MockPin {
...
+        );
     }
 
     #[test]
     fn test_connect_no_offset() {
         set_app_name("test");
-        let mut reader =
-            get_mocked_reader(Some(io::Error::new(io::ErrorKind::NotFound, "Test Error")));
-        reader.connect().unwrap();
+        let mut reader: PatInputReader<MockNoOffsetPin> = PatInputReader::new();
```
eventualbuddha, line 233: Not sure which version you find more readable, but you could avoid repeating the struct name:
```suggestion
        let mut reader = PatInputReader::<MockNoOffsetPin>::new();
```

apps/mark-scan/pat-device-input/src/pat_input_reader.rs
```diff
@@ -134,83 +146,125 @@ impl fmt::Display for MockPin {
...
+        let result = reader.connect();
+        assert!(result.is_ok());
```
eventualbuddha, line 235: I think what you had before is better for a test. This will just complain that `false` is not `true`, whereas the `unwrap` version will panic with the `Err` from `connect`.

eventualbuddha (approved):

PR #4759, opened 2024-04-15
apps/scan: Remove ultrasonic flag

jonahkagan:
## Overview

Removes the `supportsUltrasonic` flag, since all of our scanners (Custom and PDI) support multiple sheet detection, so this is an unnecessary flag. Also changes the language used in the code to refer to this feature from "ultrasonic" to "multi sheet detection" in order to better describe what the feature does, not how it works.

## Demo Video or Screenshot
N/A

## Testing Plan
Existing tests

## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/custom/state_machine.ts
+++ b/apps/scan/backend/src/scanners/custom/state_machine.ts
@@ -1423,10 +1423,6 @@ export function createPrecinctScannerStateMachine({
       machineService.send('RETURN');
     },
 
-    supportsUltrasonic: () => {
-      return true;
-    },
-
     stop: () => {
       machineService.stop();
     },
--- a/apps/scan/backend/src/scanners/custom/state_machine.ts
+++ b/apps/scan/backend/src/scanners/custom/state_machine.ts
@@ -312,13 +312,14 @@ async function reset({ client }: Context): Promise<void> {
 async function scan({ client, workspace }: Context): Promise<SheetOf<string>> {
   assert(client);
   debug('Scanning');
-  const isUltrasonicDisabled = workspace.store.getIsUltrasonicDisabled();
+  const isMultiSheetDetectionDisabled =
+    workspace.store.getIsMultiSheetDetectionDisabled();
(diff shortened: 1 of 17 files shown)


apps/scan/backend/src/scanners/custom/state_machine.ts
```diff
@@ -312,13 +312,14 @@ async function reset({ client }: Context): Promise<void> {
...
-  const isUltrasonicDisabled = workspace.store.getIsUltrasonicDisabled();
+  const isMultiSheetDetectionDisabled =
+    workspace.store.getIsMultiSheetDetectionDisabled();
   const scanResult = await client.scan({
     wantedScanSide: ScanSide.A_AND_B,
     resolution: ImageResolution.RESOLUTION_200_DPI,
     imageColorDepth: ImageColorDepthType.Grey8bpp,
     formStandingAfterScan: FormStanding.HOLD_TICKET,
-    doubleSheetDetection: isUltrasonicDisabled
+    doubleSheetDetection: isMultiSheetDetectionDisabled
```
eventualbuddha, line 322: While we're unifying naming, what do you think of "double sheet" vs. "multi sheet"?

eventualbuddha (approved):
LGTM with one caveat about "double" vs "multi".

apps/scan/backend/src/scanners/custom/state_machine.ts
jonahkagan: I think double sheet is a little clearer, but chose multi sheet because I've been seeing the MSD acronym used a lot in discussion, which made me think that was the term of art. But I don't feel strongly and would be down to standardize on either. What do you think?

eventualbuddha: I think MSD is fine. The question here was really just "do you think we should also change double to multi?"

jonahkagan: I almost changed it, but decided since it was full encapsulated within the custom scanner code, it wasn't a high value change moving forward

PR #4764, opened 2024-04-16
apps/scan/backend: Use isFeatureFlag for Fujitsu printer env var

jonahkagan:
## Overview

`getEnvironmentVariable` doesn't actually convert it to a boolean, so
the check was always true.

## Checklist


adghayes (approved):
does getEnvironmentVariable need to be exported if we don't want to use it?

jonahkagan:
> does getEnvironmentVariable need to be exported if we don't want to use it?

I considered hiding it, but it still needs to get exported for tests. We could not export it outside of libs/utils, but it didn't seem worth the effort to me

Issue #4762, opened 2024-04-16
vxmarkscan: 4/16 review polish


mattroe:
- [ ] - show ballot style as just one on the poll worker screen / not as english
- [x] - align post-printing review screen verbiage with ballot: undervotes, votes remaining
- [x] - remove icon from incorrect ballot
- [x] - remove incorrect ballot screens should look consistently wrong
- [x] - audio repeating final content after casting
- [x] - settings button broken on post-scan review page
- [x] - remove "reset accessible controller"
- [x] - only enable pat calibration in a voter state
- [x] - maintain voter state after paper jam
- [x] - can't exit controller help menu

PR #4765, opened 2024-04-16
[MarkScan] Fix exit key sequence for the accessible controller sandbox

kofi-q:
## Overview

Related to: https://github.com/votingworks/vxsuite/issues/4762

The accessible controller sandbox requires two consecutive keypresses of the `TOGGLE_HELP` keybinding, but previously only worked if the `Shift` key was held across both key presses - whereas in real-world scenarios, the controller fires key press events `Shift`, `r`, `Shift`, `r` for a double press of the help button.
This change ignores the `Shift` key when determining what the last keypress was, to avoid having the `Shift` key break the exit sequence.

### Before:

https://github.com/votingworks/vxsuite/assets/264902/d353ef0d-a440-41b9-8bb5-2522dd945419

### After:

https://github.com/votingworks/vxsuite/assets/264902/37e5dfd7-5f9b-483a-865d-44311398246e

## Testing Plan
- Updated unit tests to assert this behaviour

## Checklist

### DIFF
--- a/libs/ui/src/accessible_controllers/test_utils.tsx
+++ b/libs/ui/src/accessible_controllers/test_utils.tsx
@@ -1,5 +1,5 @@
 import userEvent from '@testing-library/user-event';
 
 export function simulateKeyPress(key: string): void {
-  userEvent.keyboard(key.length === 1 ? key : `[${key}]`);
+  userEvent.keyboard(key.length === 1 ? key : `{${key}}`);
 }
--- a/libs/ui/src/accessible_controllers/use_accessible_controller_help_trigger.test.tsx
+++ b/libs/ui/src/accessible_controllers/use_accessible_controller_help_trigger.test.tsx
@@ -25,6 +25,7 @@ test('toggles "on" to "off" for two consecutive keypresses', () => {
   expect(result.current.shouldShowControllerSandbox).toEqual(true);
 
   simulateKeyPress(Keybinding.TOGGLE_HELP);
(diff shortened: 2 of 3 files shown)


libs/ui/src/accessible_controllers/test_utils.tsx
```diff
@@ -1,5 +1,5 @@
 import userEvent from '@testing-library/user-event';
 
 export function simulateKeyPress(key: string): void {
-  userEvent.keyboard(key.length === 1 ? key : `[${key}]`);
+  userEvent.keyboard(key.length === 1 ? key : `{${key}}`);
```
kofi-q, line 4: The square brace syntax didn't seem to work for the `Shift` key, so had to switch to the curly brace syntax (https://testing-library.com/docs/user-event/keyboard/)

kshen0 (approved):
LGTM though I wonder if it might be simpler just to send and expect all lowercase letters?

kofi-q:
> I wonder if it might be simpler just to send and expect all lowercase letters

Yeah, that could work if we pick infrequently used keys. Ran into a few cases where there would be collisions (like, with the old VxMark screen reader, hitting `ctrl+r` to refresh would trigger the the screen reader, because it was mapped to `r`, so using the modifier keys reduced the chances of accidentally triggering a keybinding) - but, that's really only relevant for dev environments, where we're using the keyboard more often, so maybe not a big deal

PR #4770, opened 2024-04-18
fix: `sync` after exporting logs

### LINKED ISSUE #4685: Investigate report that "Logs Saved" modal is appearing before logs have finished exporting to the USB drive
Full email from SLI:

> Hi all,
>
> This is a low priority observation and is not tied to any requirement failure
>
> I noticed that when using the “Save Log File” feature, the “Logs Saved” popup displays immediately after pressing “Save”. In a few instances now, I have selected “Close” and removed the USB drive from the device, and the logs were missing from the USB.
>
> I have tested this on the VxScan, Admin, and CentralScan, and it appears that if the USB is removed too quickly after selecting “Close”, the logs are not saved to the USB. I believe that the “Logs Saved” popup may be appearing before the logs are fully saved to the USB drive. I have not narrowed down a timing, but have noticed that if I remove the USB one to three seconds after selecting “Close” from the “Logs Saved” popup, the logs are not on the USB drive. Waiting longer than five seconds after selecting close, though, the logs are consistently saved to the USB as expected.
>
> Thanks!

https://votingworks.slack.com/archives/CJU9MSC6S/p1709652583628519

adghayes:
## Overview

Closes #4685. I have also seen the same issue described in the same ticket when saving logs from `tty2`. I'm not sure what about saving logs is more prone to delays than other things, but in any case we can just run `sync` on the mount point to ensure the save completes before freeing up the modal / indicating the save is complete to the user.

## Testing Plan
- Automated testing confirms we are running `sync`
- I manually confirmed that the export still works after this change, although I'm not able to reproduce the original issue on my dev hardware, so I can't be 100% sure it fixes the issue, just 99% sure

### DIFF
--- a/libs/backend/src/system_call/export_logs_to_usb.ts
+++ b/libs/backend/src/system_call/export_logs_to_usb.ts
@@ -49,7 +49,7 @@ async function exportLogsToUsbHelper({
   try {
     await execFile('mkdir', ['-p', machineNamePath]);
     await execFile('cp', ['-r', LOG_DIR, destinationDirectory]);
-    await execFile('sync', [status.mountPoint]);
+    await execFile('sync', ['-f', status.mountPoint]);
   } catch {
     return err('copy-failed');
   }
--- a/libs/backend/src/system_call/export_logs_to_usb.ts
+++ b/libs/backend/src/system_call/export_logs_to_usb.ts
@@ -49,6 +49,7 @@ async function exportLogsToUsbHelper({
   try {
     await execFile('mkdir', ['-p', machineNamePath]);
     await execFile('cp', ['-r', LOG_DIR, destinationDirectory]);
+    await execFile('sync', [status.mountPoint]);
   } catch {
     return err('copy-failed');
   }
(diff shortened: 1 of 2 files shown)


libs/backend/src/system_call/export_logs_to_usb.ts
```diff
@@ -49,6 +49,7 @@ async function exportLogsToUsbHelper({
   try {
     await execFile('mkdir', ['-p', machineNamePath]);
     await execFile('cp', ['-r', LOG_DIR, destinationDirectory]);
+    await execFile('sync', [status.mountPoint]);
```
arsalansufi, line 52: I know elsewhere we use `sync -f ...`, e.g.

https://github.com/votingworks/vxsuite/blob/d48791757a5ad2e15c5980b2cd7320ec5d6c557e/libs/backend/src/exporter.ts#L159-L161

```
$ sync --help
Usage: sync [OPTION] [FILE]...
Synchronize cached writes to persistent storage

If one or more files are specified, sync only them,
or their containing file systems.

  -d, --data             sync only file data, no unneeded metadata
  -f, --file-system      sync the file systems that contain the files
      --help        display this help and exit
      --version     output version information and exit
```

Just double checking that we don't need that here. I don't _think_ we do since `status.mountPoint` is the root of the USB drive file system.

arsalansufi (approved):
Thanks for looking into this!

libs/backend/src/system_call/export_logs_to_usb.ts
adghayes: I'm confident we don't need `-f` for the reason you pointed it, but I added it in for consistency with the `exporter` code.

PR #4771, opened 2024-04-18
pdi-scanner: Support command queueing in TS client

jonahkagan:
## Overview

The Rust code already queues commands and processes them one a time, so we can adapt the TS client to match this behavior and receive responses in the order that requests were sent. That way, the consuming code in the state machine doesn't need to worry about waiting for previous requests to complete before sending a new request (which can be an issue when transitioning from a state that polls scanner status to another state that sends an accept command or the like).

## Demo Video or Screenshot
N/A

## Testing Plan
Manual testing with the state machine for now

## Checklist

### DIFF
--- a/libs/pdi-scanner/ts/src/scanner_client.ts
+++ b/libs/pdi-scanner/ts/src/scanner_client.ts
@@ -2,11 +2,10 @@ import * as path from 'path';
 import { spawn } from 'child_process';
 import { createInterface } from 'readline';
 import {
-  Deferred,
   Result,
   assert,
   assertDefined,
-  deferred,
+  deferredQueue,
   err,
   ok,
   throwIllegalValue,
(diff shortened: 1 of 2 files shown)


libs/pdi-scanner/ts/src/scanner_client.ts
```diff
@@ -146,9 +146,9 @@ export function createPdiScannerClient() {
     }
   }
 
-  // pdictl only processes one command at a time, so once we send a command, we
-  // can wait on a single response.
-  let pendingResponse: Deferred<PdictlResponse> | undefined;
+  // pdictl queues the commands it receives and only processes one command at a
+  // time, so we track the commands we sent in a queue.
+  const pendingResponseQueue: Array<Deferred<PdictlResponse>> = [];
```
eventualbuddha, line 151: This may or may not be useful to you: [`DeferredQueue`](https://github.com/votingworks/vxsuite/blob/d48791757a5ad2e15c5980b2cd7320ec5d6c557e/libs/basics/src/deferred.ts#L41).

eventualbuddha (approved):
I didn't test this out myself, but the code looks reasonable.

libs/pdi-scanner/ts/src/scanner_client.ts
jonahkagan: Ooh, very cool. Just tried out using it, but realized that there's no way to tell if the `DeferredQueue` has any pending promises waiting (which we use in the error case to figure out if we received an error in response to a message or as an event).

This does tempt me to distinguish responses from events in pdictl's JSON format...

jonahkagan: https://github.com/votingworks/vxsuite/pull/4776

jonahkagan: Updated to use `DeferredQueue` after making that change. It's very clean now!

eventualbuddha (approved):

PR #4773, opened 2024-04-18
pdi-scanner: Detect disconnect and send event

jonahkagan:
## Overview

Detect when the scanner is disconnected (e.g. unplugged) while polling the scanner for incoming events. This manifests as `rusb_async::Error:Stall`. Propagate this error back up to the main thread and send a disconnect error event to stdout.

We also break from the loop that is detecting events when we receive an error, which causes the scanner_to_host channel to close, so we handle that case in the main thread by dropping the client (allowing the user to reconnect if desired).
## Demo Video or Screenshot
```sh
$ pdictl
# Start with scanner plugged in
{"type":"connect"} # command
{"type":"ok"} # response
# Unplug scanner
{"type":"error","code":"disconnected","message":null} # event
{"type":"connect"} # command
{"type":"error","code":"disconnected","message":null} # response
# Plug scanner back in
{"type":"connect"} # command
{"type":"ok"} # response
```

## Testing Plan
Manual tested `pdictl` as outlined above, testing with state machine

## Checklist

### DIFF
--- a/libs/pdi-scanner/src/scanner.rs
+++ b/libs/pdi-scanner/src/scanner.rs
@@ -9,11 +9,7 @@ use std::{
 
 use rusb::UsbContext;
 
-use crate::{
-    protocol::parsers,
-    rusb_async::{self, Error},
-    Result,
-};
+use crate::{protocol::parsers, rusb_async, Result};
 
 use super::protocol::packets::{self, Incoming};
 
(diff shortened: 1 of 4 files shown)


libs/pdi-scanner/src/scanner.rs
```diff
@@ -219,7 +213,10 @@ impl Scanner {
     /// dropped.
     pub fn stop(&mut self) {
         if let Some(stop_tx) = self.stop_tx.take() {
-            stop_tx.send(()).unwrap();
+            // send will only error if its receiver has been dropped, which
+            // means the scanner is already cleaned up, so it's safe to ignore
+            // the error here.
+            stop_tx.send(()).ok();
```
eventualbuddha, line 219: From what I've seen, the idiomatic way to ignore a result is like so:
```suggestion
            let _ = stop_tx.send(());
```

eventualbuddha (approved):

PR #4778, opened 2024-04-22
chore: remove unused deps

eventualbuddha:
## Overview
Mostly removes `zod` from a number of packages, but also a couple others plus some intra-monorepo deps.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests.

### DIFF
--- a/libs/ballot-encoder/package.json
+++ b/libs/ballot-encoder/package.json
@@ -31,12 +31,9 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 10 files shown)


jonahkagan (approved):
Did you use any tool to detect these? Just curious for future cleanup

eventualbuddha:
> Did you use any tool to detect these? Just curious for future cleanup

`npx depcheck` pointed me in the right direction. Has false positives, but it at least narrows the field.

PR #4784, opened 2024-04-23
Generalize mark-scan diagnostic section component

kshen0:
## Overview
Generalizes a `MarkScanDeviceDiagnosticSection` component for use by the accessible controller, paper handler, PAT device, and headphone diagnostics.

## Demo Video or Screenshot
No changes, but manually verified the page renders as expected in a VM

Screenshot 2024-04-23 at 5 37 55 PM

## Testing Plan
* For exported readiness report, ran snapshot tests and verified imperceptible difference. Updated to add tolerance.
* Updated frontend tests and manually verified render above

## Checklist

- [ ] (N/A) I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.


.circleci/config.yml
```diff
@@ -241,6 +247,8 @@ jobs:
             JEST_JUNIT_OUTPUT_DIR: ./reports/
       - store_test_results:
           path: apps/mark-scan/backend/reports/
+      - store_artifacts:
+          path: apps/mark-scan/backend/src/__image_snapshots__/__diff_output__/
```
adghayes, line 251: Thank you for adding this!

`__image_snapshots__` directories may appear in other places like `apps/mark-scan/backend/src/reports/__image_snapshots__/__diff_output__/`. They appear in the same directory as the test. I just did a little research and it looks like this config doesn't support a glob pattern, so I think what we'll want to do is manually add paths as they come up or change the snapshot config. 

In either case, I don't think you need to overthink or block on that. Thank you for adding this across apps.

apps/mark-scan/backend/src/__image_snapshots__/readiness-report-1.png
adghayes, line 1: I can't see any difference here - is this a Deb 11 / Deb 12 issue that still needs to be fixed? Or is there a good reason that this is changing?

apps/mark-scan/backend/src/app.diagnostics.test.ts
```diff
@@ -149,6 +149,9 @@ jest.mock('./util/get_current_time', () => ({
 }));
 
 test('saving the readiness report', async () => {
+  // The VotingWorks logo has a small diff we weren't able to identify the root cause of
+  const failureThreshold = 0.001;
+
```
adghayes, line 154: Same Q - is this a VM issue?

libs/monorepo-utils/src/circleci.ts
```diff
@@ -21,6 +21,7 @@ function generateTestJobForNodeJsPackage(
   }
 
   const hasPlaywrightTests = existsSync(`${pkg.path}/playwright.config.ts`);
+  const hasSnapshotTests = existsSync(`${pkg.path}/src/__image_snapshots__`);
```
adghayes, line 24: Ohhh, now I see how you added it across libs. You could make this logic more robust to catch all the `__image_snapshots__` folders. But that seems low pri compared to VSAP work if its a hassle - can you make a ticket for us to track?

adghayes (commented):
@kshen0 looks good, thanks for generalizing this component. I think it is probably generalizable beyond VxMarkScan (for e.g. VxScan also) but fine to leave named as is until someone else needs to further generalize.

Not fully approving for now based on my questions about the snapshot changes / VM differences.

libs/monorepo-utils/src/circleci.ts
kshen0: Yup! https://github.com/orgs/votingworks/projects/23/views/15?pane=issue&itemId=62790720

apps/mark-scan/backend/src/__image_snapshots__/readiness-report-1.png
kshen0: VM issue, thanks for catching. Updated to remove the diff and failure threshold, but I left the infra in to provide a failure threshold.

adghayes (approved):

PR #4785, opened 2024-04-23
[MarkScan] Improve styling consistency across centered layouts

kofi-q:
## Overview

Related to: https://github.com/votingworks/vxsuite/issues/4762

After running through E2E tests, we noticed layout inconsistencies between pages, particularly the ones with minimal, centred content. This attempts to move those pages into a more consistent format, preserving the centred layout inherited from the older VxMark, while avoiding hard-to-read centred paragraph text.

Also adding icons to make the pages look less empty and provide from visual context cues.

## Demo Video or Screenshot

| First Header  | Second Header |
| ms-24 | ms-9 |
| ms-14 | ms-1 |
| ms-15 | ms-2 |
| ms-25 | ms-10 |
| ms-21 | ms-6 |
| ms-27 | ms-26 |
| ms-23 | ms-7 |
| ms-22 | ms-8 |
| ms-16 | ms-11 |
| ms-17 | ms-12 |
| ms-18 | ms-13 |
| ms-28 | ms-5 |
| ms-19 | ms-3 |
| ms-20 | ms-4 |

## Testing Plan
- Existing tests as regression guard
- Manual verification

## Checklist


kofi-q:
cc @mattroe for design tweaks/feedback

mattroe:
Much better - overall looks good to me. Appreciate the poll worker specific callouts on the relevant screens as well. My one question/piece of feedback on the design: is the border around the messages necessary? No strong preference to remove it, but it seems different from the styling we've implemented for similar messages on VxScan.

kofi-q:
> My one question/piece of feedback on the design: is the border around the messages necessary?

Yeah, went back and forth on that - what pushed me in this direction was that some of the less verbose content looked strange floating left-aligned in the middle and even more jarring with the centred icon above them. Will post a screenshot without it and see what y'all think.

kshen0:
I like how the border helps anchor the icons. I think it's ok to diverge from VxScan given the different screen orientation will present different visual design challenges.

kofi-q:
> I like how the border helps anchor the icons.

^ Yeah, that's a good to put it.

Some examples without borders:

mattroe:
Sounds good - thanks for the explanation. Agree it works better after seeing the alternative.

mattroe:
How many distinct styles did we have on these message screens before lol?

kofi-q:
Lol surprisingly, they were already all using a common template, but with enough variation in how they were configured that we ended up with lots of different-looking layouts over time 😅

apps/mark-scan/frontend/src/pages/validate_ballot_page.tsx
```diff
@@ -80,7 +80,7 @@ export function ValidateBallotPage(): JSX.Element | null {
             {appStrings.buttonBallotIsIncorrect()}
           </Button>
           <Button
-            id={PageNavigationButtonId.PREVIOUS}
```
kshen0, line 83: Thanks for catching this

kshen0:
IIRC the `warn` color is set to black/dark grey, but the `success` color is green. Is there a reason for that?

kshen0 (approved):

kofi-q:
> IIRC the warn color is set to black/dark grey (as shown in these screenshots), but the success color is green. Is there a reason for that?

Ah yeah, in the default color mode, we couldn't find an amber/yellow shade that satisfies the `10:1` contrast ratio requirement (without looking like a dark brown), so only low-contrast-mode users benefit from the warning colour at the moment. We treat the colors as secondary to the shape-based visual cues for that reason and don't rely on them to convey information

PR #4787, opened 2024-04-23
apps/scan: Detect when scanner cover is open

jonahkagan:
## Overview

Show a blocking message when the scanner cover is open. Implemented for the PDI scanner only.

Context from @mattroe:
>I think we should populate a distinct message from the “internal connection error” screen that calls out the scanner cover is open and for a poll worker to ensure its closed.
>
> Realistically, this should only happen in the following cases:
> - cleaning (intentional)
> - not closing it all the way after cleaning (unintentional)
> - transport movement (unintentional and unlikely, but possible)

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/79679efa-e721-436f-a820-d9b4786dee2d

## Testing Plan
Manual testing. Added one frontend test for coverage. Will add backend/driver tests later.

## Checklist

### DIFF
--- a/apps/scan/frontend/src/screens/scanner_cover_open_screen.tsx
+++ b/apps/scan/frontend/src/screens/scanner_cover_open_screen.tsx
@@ -1,0 +1,13 @@
+import { CenteredLargeProse, H1, P, appStrings } from '@votingworks/ui';
+import { ScreenMainCenterChild } from '../components/layout';
+
+export function ScannerCoverOpenScreen(): JSX.Element {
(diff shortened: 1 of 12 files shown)


apps/scan/frontend/src/screens/scanner_cover_open_screen.tsx
jonahkagan, line 1: modeled after SetupScannerScreen

eventualbuddha (approved):

PR #4788, opened 2024-04-24
VxMarkScan Backend BMD Rendering

adghayes:
## Overview

Moves BMD rendering to the backend in VxMarkScan, a necessary step toward being able to render BMD ballots for test decks and also to be able to deprecate `kiosk-browser` features.


apps/mark-scan/backend/src/app.test.ts
```diff
@@ -360,20 +376,8 @@ test('setting precinct', async () => {
   );
 });
 
-test('printing a ballot increments the printed ballot count', async () => {
```
adghayes, line 363: Covered below.

apps/mark-scan/backend/src/app.test.ts
```diff
@@ -394,46 +398,47 @@ test('empty ballot box', async () => {
 });
 
 test('getPaperHandlerState returns state machine state', async () => {
-  await configureForTestElection();
+  await configureForTestElection(electionGeneralDefinition);
   await apiClient.setAcceptingPaperState();
   expect(await apiClient.getPaperHandlerState()).toEqual('accepting_paper');
 });
 
-test('setAcceptingPaperState is a no-op when USE_MOCK_PAPER_HANDLER flag is on', async () => {
```
adghayes, line 402: I removed this as a no-op but don't have a good understanding of why it was no-op, so may have to revert.

apps/mark-scan/backend/src/app.test.ts
```diff
@@ -487,3 +482,88 @@ test('setPatDeviceIsCalibrated', async () => {
...
+
+    return vote.slice().sort();
+  });
+}
+
+function expectVotesEqual(expected: VotesDict, actual: VotesDict) {
+  expect(sortVotes(actual)).toEqual(sortVotes(expected));
+}
+
+test('printing ballots', async () => {
```
adghayes, line 504: With the changes to the interpretation fixture flow, we can generate and interpret ballots dynamically, allowing us to catch more in backend testing than we previously could.

apps/mark-scan/backend/src/app.ts
```diff
@@ -180,16 +179,7 @@ export function buildApi(
...
 
     setAcceptingPaperState(): void {
-      if (
-        isFeatureFlagEnabled(
-          BooleanEnvironmentVariableName.USE_MOCK_PAPER_HANDLER
-        )
-      ) {
-        return;
-      }
-
```
adghayes, line 190: I didn't understand why this was here. I needed to remove it because I wanted to use the mock paper handler flag in tests to use the printing --> straight to interpretation mocked behavior, but I needed to advance the state machine into a `waiting_on_ballot_data` state to print at all.

apps/mark-scan/backend/src/app.ts
```diff
@@ -202,27 +192,16 @@ export function buildApi(
...
-        if (!stateMachine) {
-          return;
-        }
-
-        // Mock print behavior when no paper handler is connected.
-        // Skips the print state, sets the scanned ballot filepaths to
-        // a fixture for Sample General Election, North Springfield, ballot style 5,
-        // and continues to the interpretation state.
-        stateMachine.setInterpretationFixture();
-      }
```
adghayes, line 222: I moved this logical fork into the state machine itself.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
```diff
@@ -453,24 +453,6 @@ test('voting flow happy path', async () => {
...
-      scannedBallotFixtureFilepaths
-    );
-
-    setMockDeviceStatus(getPaperInFrontStatus());
-    await expectStatusTransitionTo('presenting_ballot');
-
-    setMockDeviceStatus(getDefaultPaperHandlerStatus());
-    await expectStatusTransitionTo('presenting_ballot');
-  });
-
```
adghayes, line 473: Removed this one too 🤷  I'm guessing I probably broke something.

apps/mark-scan/frontend/src/app_contest_ms_either_neither.test.tsx
```diff
@@ -68,151 +45,8 @@ const pickOneContest = find(
 assert(eitherNeitherContest.type === 'yesno');
 assert(pickOneContest.type === 'yesno');
 
-const ballotStyleId = '1';
 const precinctId = '6526';
 
-function expectPrintedVotes(
```
adghayes, line 74: All this testing was essentially just testing that the rendered ballot reflected various either / neither possibilities. These possibilities are even special cased in the BmdPaperBallot component so this sort of basic logic - does the ballot reflect the passed votes - is tested elsewhere.

apps/mark/frontend/src/pages/print_page.tsx
```diff
@@ -16,12 +17,27 @@ export function PrintPage(): JSX.Element {
     resetBallot,
   } = useContext(BallotContext);
   assert(electionDefinition, 'electionDefinition is not defined');
-  assert(typeof ballotStyleId === 'string', 'ballotStyleId is not defined');
-  assert(typeof precinctId === 'string', 'precinctId is not defined');
 
   const printerTimer = useRef(0);
 
-  function onPrintStarted() {
+  async function print() {
```
adghayes, line 23: had to move this from `mark-flow-ui` to here until it is moved to the backend in the next PR

adghayes (commented):
I'd like @kofi-q and @kshen0 both to take a look at this because it touches UI String and paper handler logic respectively.

@kshen0 the main issue I struggled with in `mark-scan/backend` is the mocking. We're mocking the paper handler in lots of ways - some times by `jest.mock`ing the lib, sometimes by using `MockPaperHandlerDriver`, sometimes by using the feature flag. I think that we will probably want to get down to a single mocking interface at some point. In order to allow me to test printing, I removed some of the no-ops - I don't know if those are necessary on hardware. 

Also @kshen0 I know you have other PRs out that will conflict with this one - I'll take your lead on order. I'll be in the office this week and will test on hardware myself / with you.

apps/mark-scan/frontend/package.json
```diff
@@ -93,7 +93,6 @@
     "@types/fetch-mock": "^7.3.2",
     "@types/history": "4.7.11",
     "@types/jest": "^29.5.3",
-    "@types/kiosk-browser": "workspace:*",
```
kofi-q, line 96: 🎉

kofi-q (commented):
Deferring to Kevin for the state machine pieces, but rendering/language framework's looking good - one thing to address on making sure we're using election package content over hardcoded strings.

apps/mark-scan/backend/src/util/render_ballot.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const electionDefinition = assertDefined(store.getElectionDefinition());
+  const isLiveMode = !store.getTestMode();
+
+  const uiStringTranslations =
+    store.getUiStringsStore().getUiStrings(languageCode) ?? undefined;
+
+  const ballot = (
+    <BackendLanguageContextProvider
+      languageCode={languageCode}
+      uiStringTranslations={uiStringTranslations}
```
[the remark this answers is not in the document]
adghayes: Makes total sense, will make this change!

apps/mark-scan/backend/src/app.ts
```diff
@@ -202,27 +192,16 @@ export function buildApi(
...
-
-      assert(stateMachine);
-      void stateMachine.printBallot(input.pdfData);
+      const pdfData = await renderBallot({
+        store,
+        ballotId: randomBallotId(),
+        ...input, // any passed ballotId will override randomBallotId()
```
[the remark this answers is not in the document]
adghayes: You're right, it is just to support testing. I can switch the mock strategy.

apps/mark-scan/backend/src/util/render_ballot.tsx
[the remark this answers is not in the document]
adghayes: @kofi-q can I get a quick glance at 5bf18630a?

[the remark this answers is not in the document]
kofi-q: Looks good - thanks!

kofi-q (approved):
Render/print logic LGTM

apps/mark-scan/backend/src/app.ts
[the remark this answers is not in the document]
adghayes: 39e2bea2e

kshen0:
> @kshen0 the main issue I struggled with in mark-scan/backend is the mocking. We're mocking the paper handler in lots of ways - some times by jest.mocking the lib, sometimes by using MockPaperHandlerDriver, sometimes by using the feature flag. I think that we will probably want to get down to a single mocking interface at some point. In order to allow me to test printing, I removed some of the no-ops - I don't know if those are necessary on hardware.

Sorry about the state of that. Adding a clear mock direction should be outside the scope of your PR so don't feel blocked on doing the "right" thing wrt. mock direction - there isn't a clear one.

My thinking was the feature flag should be used for VM development only to instantiate a MockPaperHandlerDriver instead of a real one. The intent was for this to behave like a typical mock, where mark-scan is unaware of whether it has a mock or real driver and can just call methods on the interface they share.

This gets muddy because we use MockPaperHandlerDriver in some places where Jest ergonomics were inconvenient. A clearer way to separate concerns is

* MockPaperHandlerDriver is completely built out (currently many methods are unimplemented)
* `USE_MOCK_PAPER_HANDLER` is used exactly once in production code, on mark-scan startup to decide whether to make a PaperHandlerDriver or MockPaperHandlerDriver
* The flag should be set in tests except where we test that _not_ passing the flag results in starting the state machine with a real PaperHandlerDriver
* Jest is used only to spy on MockPaperHandlerDriver and assert on calls
  * Any failure behavior we want to test should be built into `MockPaperHandlerDriver`
  * The philosophy here is that any behavior we want to test automatically, like jams, disconnect, bad response from the printer etc. should also be testable manually.

Also noting that we discussed mocking at the `application_driver` level, which might require some work on that layer to make it a true layer instead of bag of helper functions.

kshen0 (approved):

apps/mark-scan/frontend/src/pages/print_page.tsx
```diff
@@ -1,47 +1,28 @@
...
   const printBallotMutation = printBallot.useMutation();
 
-  async function printElementToCustomPaperHandler(
-    element: JSX.Element,
-    options: PrintOptions
-  ): Promise<void> {
-    debug(`Ignoring print options with keys: ${Object.keys(options)}`);
-    const pdfData = Buffer.from(await printElementToPdf(element));
-    printBallotMutation.mutate({ pdfData });
+  // eslint-disable-next-line @typescript-eslint/require-await
```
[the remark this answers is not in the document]
adghayes: Exactly.

kofi-q:
@adghayes heads-up - I'm seeing empty prints after this change:


Once every few attempts, I get a print where the text shows up, so I'm guessing there's a race condition WRT page content being loaded completely and the PDF getting rendered.

adghayes:
@kofi-q thanks for catching, will look into this today.

adghayes:
@kofi-q the margins here look like a VxMark ballot - I'm less concerned if it messed up the VxMark rendering because I'll fix that in the next PR and it's not an app in production.

adghayes:
Ah, this is a high-density ballot, got it.

kofi-q:
> the margins here look like a VxMark ballot

Ah, we started shrinking the top margin whenever we go over a certain number of contests on the ballot, which is why it might look different from what you're used to, but this is from MarkScan. Switching gears to do some user testing prep, but will take another look later today or early tomorrow

kofi-q:
Finally sorted it out! Looks like I just had a bad merge that reverted the `@media screen` styling change in `BmdPaperBallot`. Sorry for the false alarm - realising now that the server wasn't hot-reloading the changes to `BmdPaperBallot` so cause and effect weren't really clear while I was troubleshooting

adghayes:
@kofi-q no problem! I'm just relieved I don't need to troubleshoot it :)

PR #4789, opened 2024-04-24
pdi-scanner: TS client tests

jonahkagan:
## Overview

Adds tests for the TS client to pdi-scanner. Also reorganizes the files to make the test config simpler and fixes one small issue in the Rust code identified by testing.

## Demo Video or Screenshot
N/A

## Testing Plan
I'm still a little unsure what the best way to test the Rust code is, so open to any thoughts on that. I will probably move on to testing the state machine next, but I do think it's worth writing some tests for the Rust code as well, since it's quite intricate.

## Checklist


libs/pdi-scanner/src/ts/scanner_client.ts
```diff
@@ -42,11 +45,11 @@ export interface ScannerStatus {
   /**
    * @deprecated Not used by PageScan 6, always false.
    */
-  frontM5SensorCovered: never;
+  frontM5SensorCovered: false;
   /**
    * @deprecated Not used by PageScan 6, always false.
    */
-  frontRightSensorCovered: never;
+  frontRightSensorCovered: false;
```
jonahkagan, line 52: Turns out there's no way to construct a valid object literal if you have required fields with the `never` type

eventualbuddha (approved):

PR #4795, opened 2024-04-24
apps/scan: Use machines instead of observables in PDI state machine

jonahkagan:
## Overview

This will allow us to use xstate's simulated clock for testing, since all delays will use xstate machinery (rather than using rxjs for the polling delays).

## Demo Video or Screenshot
N/A
## Testing Plan
Manual test
## Checklist

### DIFF
--- a/apps/scan/backend/package.json
+++ b/apps/scan/backend/package.json
@@ -70,7 +70,7 @@
     "rxjs": "7.8.1",
     "tmp": "^0.2.1",
     "uuid": "9.0.1",
-    "xstate": "^4.32.1",
+    "xstate": "^4.33.0",
     "zod": "3.14.4"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


apps/scan/backend/package.json
```diff
@@ -70,7 +70,7 @@
     "rxjs": "7.8.1",
     "tmp": "^0.2.1",
     "uuid": "9.0.1",
-    "xstate": "^4.32.1",
+    "xstate": "^4.33.0",
```
jonahkagan, line 73: Upgraded xstate to resolve a specific bug where a delayed event from a child machine wasn't getting canceled after the child machine was stopped, which caused a warning to be logged every time we stopped polling

eventualbuddha (approved):
Nice, I hope this leads to more testable state machines.

PR #4808, opened 2024-04-30
libs/ballot-interpreter: Bugfix for binarized scanner images

jonahkagan:
## Overview

If the image is binarized by the scanner before reaching the interpreter, the Otsu threshold will be computed as 0 (pure black). We were only accepting bottom timing marks whose luminance value was less than the threshold, which was impossible with a threshold of 0. To fix this, we allow matching the threshold exactly.
## Testing Plan
Manual testing

## Checklist


arsalansufi (approved):
Great find!

PR #4809, opened 2024-05-01
Generate blank page image instead of using image file

kshen0:
## Overview
Generates a 1x1 blank page image instead of using an image file fixture.

Should solve an issue where `backend/src/custom-paper-handler/fixtures` is not included in prod builds, resulting in hanging interpretation when we try to read the nonexistent blank page image file. The blank page image file is used to pass a blank back page to the interpreter because the mark-scan scanner only scans the front page.

## Demo Video or Screenshot

No behavior change, but here's QA of this change on hardware.

**Interpreting a valid ballot**

**Interpreting blank front page**

## Testing Plan
- updated mark-scan/backend test mocks to return blank image
- manually tested on hardware

## Checklist


libs/ballot-interpreter/src/interpret.ts
```diff
@@ -9,7 +9,8 @@ import {
   ok,
   Ok,
 } from '@votingworks/basics';
-import { loadImageData } from '@votingworks/image-utils';
+import { ImageData } from 'canvas';
```
kshen0, line 12: Breaks encapsulation a little bit :/

apps/mark-scan/backend/jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   ],
   coverageThreshold: {
     global: {
-      branches: -15,
+      branches: -16,
```
kshen0, line 26: Overlap of files changed in this PR and the files with < [100 branch coverage](https://app.circleci.com/pipelines/github/votingworks/vxsuite/15440/workflows/0d0c50ae-bb85-4d19-ad4e-9d7015ef4b8b/jobs/573102) is just `state_machine.ts`. However, the uncovered branches reported in the `lcov` report do not correspond to any changes in this PR. Not sure how the uncovered branches increased, but feels like we shouldn't block on this as this review is somewhat time sensitive.

kofi-q (approved):
🚀

PR #4813, opened 2024-05-01
apps/scan: Prevent hanging when second ballot inserted during accept

jonahkagan:
## Overview

Previously, we were waiting for the scanner to no longer have any document in the scanner to declare accepting complete. It's possible to trick the scanner by feeding a second ballot in while the motors are running to accept the first ballot, which results in either:
- The first ballot successfully ejected, while the second ballot is held in the front
- The first ballot successfully ejected, while the second ballot is held in the middle (covering front and rear sensors)
- The first ballot jammed, while the second ballot is held in the front

We handle this case with a few updates:
- Only declare accepting complete when the rear is cleared
- If the front is covered before the rear is cleared (very unlikely, but possible), go back to the warning to remove the second ballot before deciding whether the first ballot was successfully accepted

## Demo Video or Screenshot
Common case: The first ballot successfully ejected, while the second ballot is held in the front

https://github.com/votingworks/vxsuite/assets/530106/dd3642dc-97a3-4edf-9832-5b028a73c55d

Rare case: The first ballot successfully ejected, while the second ballot is held in the middle (covering front and rear sensors)

https://github.com/votingworks/vxsuite/assets/530106/be070149-079b-4791-b72c-62a96e4b04d5

I didn't have enough hands to actually create the third case (the first ballot jammed, while the second ballot is held in the front), but I've tested jams separately.

## Testing Plan
Manual testing + automated tests

## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/pdi/pdi_app_scan_double_sheet.test.ts
+++ b/apps/scan/backend/src/scanners/pdi/pdi_app_scan_double_sheet.test.ts
@@ -9,7 +9,8 @@ import {
   DEFAULT_SYSTEM_SETTINGS,
   SheetInterpretation,
 } from '@votingworks/types';
-import { typedAs } from '@votingworks/basics';
+import { Result, deferred, ok, typedAs } from '@votingworks/basics';
+import { ScannerError } from '@votingworks/pdi-scanner';
 import {
   ballotImages,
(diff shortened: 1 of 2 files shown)


arsalansufi (approved):
LGTM!

PR #4816, opened 2024-05-01
Fix signing of large election packages in prod

arsalansufi:
## Overview

With the introduction of audio clips to election packages, election packages can now be substantially larger than in the past (e.g., 10+ MB instead of 50 KB). Testing with such an election package on a prod VxAdmin has surfaced an issue; digital signing of the election package hangs indefinitely. This issue is specific to prod and use of the TPM.

This PR fixes the issue by separating hashing and signing so that we can hash using the main processor and only sign using the TPM. Hashing large inputs with the TPM can be extremely slow as the TPM processor isn't particularly powerful. It's really only meant for small cryptographic operations.

## Testing Plan

- [x] Updated `cryptography.test.ts`
- [x] Updated `cryptography.end_to_end.test.ts`, which tests signing and verification commands end-to-end
- [x] Copied the updated code to my non-locked-down prod-imaged ThinkPad and verified that election package export was successful

### DIFF
--- a/libs/auth/src/cryptography.ts
+++ b/libs/auth/src/cryptography.ts
@@ -1,7 +1,7 @@
 import { Buffer } from 'buffer';
 import fs from 'fs/promises';
 import path from 'path';
-import { Stream } from 'stream';
+import { Readable, Stream } from 'stream';
 import { FileResult, fileSync } from 'tmp';
 import { z } from 'zod';
 import { throwIllegalValue } from '@votingworks/basics';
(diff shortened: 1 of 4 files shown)


benadida (approved):

libs/auth/src/cryptography.ts
```diff
@@ -474,23 +474,34 @@ export function parseSignMessageInputExcludingMessage(
 export async function signMessageHelper({
   signingPrivateKey,
 }: SignMessageInputExcludingMessage): Promise<Buffer> {
-  const signParams: OpensslParam[] = (() => {
+  // While it's possible to hash and sign with a single OpenSSL command, we separate the two
+  // actions. Hashing large inputs with the TPM can be extremely slow. So in production, we hash
+  // using the main processor and only sign using the TPM.
+  const hash = await openssl(['dgst', '-sha256', '-binary'], {
+    stdin: process.stdin,
+  });
```
eventualbuddha, line 482: Is this because we're trying to keep it as similar as possible to what it was before? I don't see anything here that indicates why it isn't just using `node:crypto`, so I'd be tempted to make that change in the future. This doesn't provide any context on that.

benadida: I would suggest keeping the hash in OpenSSL so all of this can be done in the FIPS-compliant component.

eventualbuddha: Add a comment to that effect?

arsalansufi: Comment updated: https://github.com/votingworks/vxsuite/pull/4816/commits/4b9fb5ed69af331f0b94471110870b43278720f1

PR #4820, opened 2024-05-02
apps/scan: Fix pdictl process leak

jonahkagan:
## Overview

When the scanner is disconnected, we create a new client each time we
reconnect, which spawns a new pdictl process. This process is not killed
when the previous client goes out of scope.

To fix the immediate problem, we exit the client explicitly before
creating a new one. There may be a cleaner design for the client that
will solve this in a better way, but I wanted to fix the immediate
problem as a starting point.

## Testing Plan
Manually tested, updated automated tests

## Checklist


eventualbuddha (commented):
Does it make sense to have what we used to have for plustekctl where we politely ask it to exit, wait for it to do so, then send a more aggressive `kill` signal?

eventualbuddha (approved):

jonahkagan:
> Does it make sense to have what we used to have for plustekctl where we politely ask it to exit, wait for it to do so, then send a more aggressive `kill` signal if it doesn't exit?

I've yet to see `pdictl` fail to exit when asked politely, so I'm not worried about it currently.

PR #4822, opened 2024-05-02
[mark-scan] Prod image hardware peripheral fixes

kshen0:
## Overview

Fixes 2 bugs with hardware daemons in prod images.
*  `PatInputConnectionStatusReader` did not check for GPIO address offset present in Debian 12, resulting in the app being unable to trigger the PAT device calibration flow on Debian 12
* `controllerd` did not receive an echo response the majority of the time on startup, likely because the device was not ready yet

## Demo Video or Screenshot

No behavior change. We won't be able to QA on Debian 12 until we build a prod image, but here's a video of backwards compatibility of the PAT change on Debian 11.

https://github.com/votingworks/vxsuite/assets/1060688/c00071b7-9fd4-47e4-a3a4-f6ddc0f6f865

## Testing Plan

- updated tests

## Checklist

### DIFF
--- a/apps/mark-scan/accessible-controller/src/controllerd.rs
+++ b/apps/mark-scan/accessible-controller/src/controllerd.rs
@@ -15,6 +15,8 @@ use std::{
         atomic::{AtomicBool, Ordering},
         Arc,
     },
+    thread::sleep,
+    time::Duration,
 };
 use uinput::event::keyboard;
 use vx_logging::{log, set_app_name, Disposition, EventId, EventType};
@@ -32,6 +34,7 @@ mod port;
 const APP_NAME: &str = "vx-mark-scan-controller-daemon";
 const KPB_200_FW_VID: u16 = 0x28cd;
 const KPB_200_FW_PID: u16 = 0x4008;
+const STARTUP_SLEEP_DURATION: Duration = Duration::from_millis(1500);
 
 #[derive(Parser, Debug)]
 #[command(version, about, long_about = None)]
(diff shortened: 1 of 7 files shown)


apps/mark-scan/accessible-controller/src/controllerd.rs
```diff
@@ -32,6 +34,7 @@ mod port;
 const APP_NAME: &str = "vx-mark-scan-controller-daemon";
 const KPB_200_FW_VID: u16 = 0x28cd;
 const KPB_200_FW_PID: u16 = 0x4008;
+const STARTUP_SLEEP_DURATION: Duration = Duration::from_millis(1500);
```
arsalansufi, line 37: Curious how much extra padding this 1.5 second sleep includes. Just want to make sure this sleep is sufficient to resolve the problem every time. And am wondering if there's any harm in bumping up just a bit more, e.g., to 3 seconds?

arsalansufi (approved):
LGTM! Thanks for the quick investigation and fixes

arsalansufi:
Also noting that I'll create a new VxMarkScan image once this is merged

apps/mark-scan/accessible-controller/src/controllerd.rs
kshen0: Since we saw the daemon initialize correctly on startup 1/3 times, I'm guessing the race condition is <100ms. But the risk of increasing to 3s is pretty low, so went ahead and did that.

PR #4826, opened 2024-05-06
fix(design): drop the definite article from "Green Party" translation

eventualbuddha:
The Spanish version of the name, like the English one, does not appear to contain the leading "El"/"The" definite article. This isn't really meant as the place to put all our translations and is really just an example of how to do overrides (replacing the obviously incorrect "Fiesta Verde"), but I'd still like it to be correct.

Context from Slack: https://votingworks.slack.com/archives/CEL6D3GAD/p1715015367963289


arsalansufi (approved):
🟩 🎉

PR #4830, opened 2024-05-07
[BMD] Render prints within the main React tree

kofi-q:
## Overview

Noticed string translations weren't getting propagated to the `BmdPaperBallot`, since it was getting rendered in a separate React tree without any of the context data from the main tree.

Looks like that was done to enable rendering print content under a different root element in the DOM, since the main `#root` element is hidden in `@media print` mode -- updating that logic to use a React portal instead, so both DOM trees can share the same React context.

### VxMarkScan:

| Before | After |
| markscan-before | markscan-after |

### VxMark:
| Before | After |
| vxmark-before | vxmark-after |

## Testing Plan
- New/updated automated test cases
- Manual verification in BMDs

## Checklist


kofi-q:
Can hold off on this for now - we'll be able to scrap this altogether once https://github.com/votingworks/vxsuite/pull/4788 and the follow up for VxMark are merged

PR #4831, opened 2024-05-08
Fujitsu Printer Reconnect & Power Settings

adghayes:
Commit 1: Currently VxScan just crashes if the printer is not attached on startup, and it cannot handle disconnects and reconnects. While this was OK for the happy path, it's not OK for the sad path and development. Changing the layers so that we are just checking for the device on every status call.

Commit 2: Finishes removing the speed setting (started in commit 1) and also specifies the print quality setting on printer connect. This is to set "Auto-Division" which we've found that we need to set to prevent spikes in power draw that go beyond our v4 PSU capabilities.

### DIFF
--- a/libs/fujitsu-thermal-printer/src/printer.ts
+++ b/libs/fujitsu-thermal-printer/src/printer.ts
@@ -1,10 +1,10 @@
-import { Optional, Result, err, ok } from '@votingworks/basics';
+import { Optional, Result, assert, err, ok } from '@votingworks/basics';
 import { Buffer } from 'buffer';
 import { print } from './printing';
 import {
+  FujitsuThermalPrinterDriver,
   FujitsuThermalPrinterDriverInterface,
-  SpeedSetting,
-  getFujitsuThermalPrinterDriver,
+  getDevice,
 } from './driver';
 import { rootDebug } from './debug';
 import { IDLE_REPLY_PARAMETER, LINE_FEED_REPLY_PARAMETER } from './globals';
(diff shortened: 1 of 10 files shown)


adghayes:
@jonahkagan normally @kofi-q has been reviewing this integration but since he's out, passing you to. Setting up the reconnect logic we talked about.

libs/fujitsu-thermal-printer/src/printer.ts
jonahkagan, line 1: The logic in this file feels a little complex to me. There's a strategy for managing the connection state of the driver that's implemented across two methods: `getDriver` and `getStatus`, and then all of the other methods have to remember to start by calling `getStatus` (which feels a little brittle).

Could this code be reorganized such that in order to access the driver, each method needs to call some function that checks the connection status? A sketch:

```ts
async print(data: Buffer) {
  const driverResult = await getDriver();
  if (driverResult.isErr()) { // If the driver is disconnected or otherwise in a non-ready state
     return driverResult;
  }

  const printResult = await print(this.driver, data); 
  ...
```

(Ideally in this scheme, there would be no `this.driver` - the only way to access it would be through `getDriver`)

adghayes: @jonahkagan I ended up not re-organizing the code, but dumbing down my assumptions. I have never seen the printer be unresponsive while still attached, which we've seen with scanners. As long as it's connected, it responds to commands. So within `print` and `advancePaper`, I'm going to assume that the driver exists and is available. `getStatus` is the only endpoint which handles connection and disconnection. It's not actually plausible that we would jump straight to invoke an action - the app should always be checking for device status before starting an action.

libs/fujitsu-thermal-printer/src/printer.ts
```diff
@@ -22,40 +22,74 @@ const WAIT_FOR_ADVANCE_APPEAR_PER_CM_MS = 250;
...
-    const status = await this.driver.getStatus();
-    return summarizeRawStatus(status);
+    // try to initialize the driver if it does not yet exist
+    if (!this.driver) {
+      this.driver = await this.getDriver();
+      // if we failed to initialize the driver, the device is likely not connected
+      if (!this.driver) {
+        return { state: 'error', type: 'disconnected' };
+      }
+    }
```
jonahkagan, line 68: Could also write as:

```ts
this.driver ??= await this.getDriver();
 // if we failed to initialize the driver, the device is likely not connected
if (!this.driver) {
  return { state: 'error', type: 'disconnected' };
}
```

which is slightly simpler for me to read

libs/fujitsu-thermal-printer/src/printer.ts
```diff
@@ -22,40 +22,74 @@ const WAIT_FOR_ADVANCE_APPEAR_PER_CM_MS = 250;
...
-  private readonly driver: FujitsuThermalPrinterDriverInterface;
+  private driver?: FujitsuThermalPrinterDriverInterface;
 
-  constructor(_driver: FujitsuThermalPrinterDriverInterface) {
-    this.driver = _driver;
-  }
+  /**
+   * Initializes and returns a new driver instance.
+   */
+  private async getDriver(): Promise<Optional<FujitsuThermalPrinterDriver>> {
```
jonahkagan, line 36: maybe rename to `initializeDriver` to be slightly more descriptive?

jonahkagan (approved):

PR #4832, opened 2024-05-08
Improve PDI scanner error handling

jonahkagan:
## Overview

Makes a variety of changes to libs/pdi-scanner and apps/scan-backend to improve the behavior in edge case error scenarios. The main improvements:
- Improved reliability of recovering from unexpected errors
- Better discernment of when the scanner is in a borked state (so we can show the message to restart)

Details in the commits.

## Testing Plan
Added tests to cover each piece of new behavior, and manually tested.

## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/pdi/pdi_app_scan.test.ts
+++ b/apps/scan/backend/src/scanners/pdi/pdi_app_scan.test.ts
@@ -13,6 +13,7 @@ import {
 import { ScannerError } from '@votingworks/pdi-scanner';
 import {
   ballotImages,
+  createMockPdiScannerClient,
   mockStatus,
   simulateScan,
   withApp,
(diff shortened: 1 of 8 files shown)


libs/pdi-scanner/src/rust/main.rs
```diff
@@ -218,7 +201,33 @@ fn main() -> color_eyre::Result<()> {
...
     let mut client: Option<Client<Scanner>> = None;
     let mut raw_image_data = RawImageData::new();
-    let mut scan_in_progress = false;
+
+    // We reject sending a command while a scan is in progress because it will
+    // interrupt the scan. To ensure this flag gets reset whenever a scan stops
+    // (whether successfully or with an error or disconnection), we manage this
+    // flag in the send_response/send_event functions, since they are called in
+    // basically every case when the scanner state changes.
+    let scan_in_progress = Cell::new(false);
```
jonahkagan, line 210: I'm not sure if this is the most idiomatic way to implement the functionality I want (factoring out the logic to update this flag into a few different functions), so open to suggestions for other patterns

libs/logging/src/test_utils.ts
```diff
@@ -18,7 +19,27 @@ export function mockLogger(
...
+        eventId,
+        eventType: eventSpecificDetails.eventType,
+        user,
+        message,
+        disposition,
+        ...additionalData,
+      };
+      outerDebug(logLine);
+    }
+  });
```
jonahkagan, line 42: excerpted from the actual definition of `logger.log`

libs/pdi-scanner/src/rust/main.rs
eventualbuddha: I think this is okay. `Cell` gives us what we need in terms of mutability from multiple closures, so I'd keep this as-is.

apps/scan/backend/src/scanners/pdi/pdi_app_scan.test.ts
```diff
@@ -490,6 +498,12 @@ test('if scanning times out, ballot rejected', async () => {
         },
         expect.any(Function)
       );
+
+      // Make sure event listener was added to new client (regression test for
+      // bug where we weren't adding the root event listener to the new client)
```
eventualbuddha, line 503: I appreciate the regression test.

eventualbuddha (approved):

PR #4841, opened 2024-05-08
apps/scan: Scan ballot inserted during accepted state

jonahkagan:
## Overview

Previously, we would reject a ballot inserted during the accepted state. Instead, we still wait until the accepted state is done, but then scan the next ballot. This is especially important for testing use cases where we want to feed many ballots in a row.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/aea505fc-7b45-4eb5-884b-c78aa53bf2e7

## Testing Plan

## Checklist


eventualbuddha (approved):
Verified that the new test fails when run against the old code 👍

PR #4843, opened 2024-05-09
apps/scan: Enable PDI scanner eject pause feature

jonahkagan:
## Overview

When a ballot is being accepted and a second ballot is inserted, we don't want the second ballot to be mistakenly sucked through the scanner. Though we have guards in place, if there's enough resistance against the first ballot, it's still possible that this can occur.

Luckily, the PDI scanner has an internal "eject pause" feature that is designed to detect a second ballot in the front and pause the ejection. It's more effective than our previous approach since it has lower level control over the motors and can act quicker.

This PR enables the eject pause feature (only for rear ejections - i.e. ballot
accepts) and adds the corresponding events for eject pause/eject resume.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/b7a54dc6-d990-40ef-b42e-7b3b5b5ed8fa

Note that in the 4th scan in this video, there's a 2s delay before the "remove your ballot" message pops up. This is the case where the scanner status request times out (detailed in one of the commits). We could potentially change the status request timeout in this context to make this faster, but given that it's an extreme edge case requiring precise timing to reproduce, I opted to let it be slow and keep things simpler.

## Testing Plan

## Checklist

### DIFF
--- a/libs/pdi-scanner/src/rust/client.rs
+++ b/libs/pdi-scanner/src/rust/client.rs
@@ -90,7 +90,7 @@ impl<T> Client<T> {
             .iter()
             .filter(|packet| !packet.is_event())
             .collect::<Vec<_>>();
-        if unhandled_non_event_packets.len() > 0 {
+        if unhandled_non_event_packets.is_empty() {
             tracing::debug!("clearing unhandled packets: {unhandled_non_event_packets:?}");
             self.unhandled_packets.retain(Incoming::is_event);
         }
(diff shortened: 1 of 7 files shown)


libs/pdi-scanner/src/rust/client.rs
```diff
@@ -73,6 +74,28 @@ impl<T> Client<T> {
...
+    // delay sending a response to a command (e.g. when the "eject pause"
+    // feature pauses the scanner, it will queue commands received during that
+    // time).
+    fn clear_unhandled_packets_except_events(&mut self) {
+        let unhandled_non_event_packets = self
+            .unhandled_packets
+            .iter()
+            .filter(|packet| !packet.is_event())
+            .collect::<Vec<_>>();
+        if unhandled_non_event_packets.len() > 0 {
```
eventualbuddha, line 93: Per clippy:

```suggestion
        if !unhandled_non_event_packets.is_empty() {
```

eventualbuddha (approved):

libs/pdi-scanner/src/rust/client.rs
jonahkagan: thanks, changed.

maybe we should run clippy automatically?

PR #4846, opened 2024-05-10
[mark-scan] Handle status transmission command

kshen0:
## Overview
Buffering print data (or any data) to the generic transfer-out buffer may invoke the `real-time status transmission` command if the buffered data contains a byte sequence matching that command (`0x10 0x04`). This causes unsolicited data to be passed to the generic transfer-in buffer, which breaks parsing the next time the driver issues a command that responds over the generic transfer-in buffer.

See comment here: https://github.com/votingworks/vxsuite/pull/4846/files#diff-6d4416eebc285362e1f02d2f6252d35972e5c0db67a2db365540cb87a2bea431R388 and [Slack](https://votingworks.slack.com/archives/C03Q42V9MJ5/p1715296095534459) for more.

## Demo Video or Screenshot

Successful print of ballot with problematic byte sequence

https://github.com/votingworks/vxsuite/assets/1060688/7a52b519-e07a-4bba-b86c-ecf444566709

## Testing Plan

- added tests
- manually verified ballot with problematic byte sequence can be printed

## Checklist

- [ ] (N/A) I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -845,13 +845,15 @@ function setUpLogging(
     .onEvent(async (event) => {
       // To protect voter privacy, we only log the event type (since some event
       // objects include ballot interpretations)
-      await logger.log(
-        LogEventId.MarkScanStateMachineEvent,
-        'system',
-        { message: `Event: ${event.type}` },
-        /* istanbul ignore next */
-        (logLine: LogLine) => debugEvents(logLine.message)
-      );
+      if (event.type !== 'PAT_DEVICE_NO_STATUS_CHANGE') {
+        await logger.log(
+          LogEventId.MarkScanStateMachineEvent,
+          'system',
+          { message: `Event: ${event.type}` },
+          /* istanbul ignore next */
+          (logLine: LogLine) => debugEvents(logLine.message)
+        );
+      }
     })
     .onChange(async (context, previousContext) => {
       if (!previousContext) return;
--- a/libs/custom-paper-handler/src/driver/coders.ts
+++ b/libs/custom-paper-handler/src/driver/coders.ts
@@ -84,12 +84,7 @@ export type SensorStatusRealTimeExchangeResponse = CoderType<
   typeof SensorStatusRealTimeExchangeResponse
 >;
 
-export const PrinterStatusRealTimeExchangeResponse = message({
-  startOfPacket: literal(0x82),
-  requestId: uint8(),
-  token: literal(TOKEN),
-  returnCode: uint8(),
-  optionalDataLength: literal(6),
+const printerStatusMessage = {
   // Byte 0 fixed
   dle: literal(0x10),
   // Byte 1 fixed
@@ -121,6 +116,15 @@ export const PrinterStatusRealTimeExchangeResponse = message({
   eepromError: uint1(),
   ramError: uint1(),
   byte3Padding1: padding(2),
+} as const;
+
+export const PrinterStatusRealTimeExchangeResponse = message({
+  startOfPacket: literal(0x82),
+  requestId: uint8(),
+  token: literal(TOKEN),
+  returnCode: uint8(),
+  optionalDataLength: literal(6),
+  ...printerStatusMessage,
 });
 export type PrinterStatusRealTimeExchangeResponse = CoderType<
   typeof PrinterStatusRealTimeExchangeResponse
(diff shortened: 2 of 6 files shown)


apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -851,13 +851,15 @@ function setUpLogging(
...
       // To protect voter privacy, we only log the event type (since some event
       // objects include ballot interpretations)
-      await logger.log(
-        LogEventId.MarkScanStateMachineEvent,
-        'system',
-        { message: `Event: ${event.type}` },
-        /* istanbul ignore next */
-        (logLine: LogLine) => debugEvents(logLine.message)
-      );
+      if (event.type !== 'PAT_DEVICE_NO_STATUS_CHANGE') {
```
kshen0, line 854: This event is noisy and has little value in the logs

libs/custom-paper-handler/src/driver/driver.test.ts
```diff
@@ -640,3 +641,82 @@ test('bufferChunk', async () => {
...
+      .mockReturnValueOnce(
+        getMockTransferInResponse(
+          Buffer.of(ReturnCodes.INVALID_STATUS_TRANSMISSION_ARGUMENT)
+        )
+      )
+      .mockReturnValueOnce(
+        getMockTransferInResponse(
+          Buffer.of(ReturnCodes.POSITIVE_ACKNOWLEDGEMENT)
+        )
+      );
```
kshen0, line 690: As far as I can tell, the paper handler will buffer responses in discrete chunks. So when 2 commands are executed without reading the transfer in buffer after the first command, we can still expect reading the transfer in buffer to return a single response rather than mashing them all together.

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -505,6 +549,7 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
   }
 
   async scanAndSave(pathOut: string): Promise<ImageFromScanner> {
+    debug('setting scan direction');
```
kshen0, line 552: Found these useful so I left them in

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -369,27 +375,65 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
...
-        throw new Error(`uninterpretable acknowledgement: ${code}`);
+    if (typeof code === 'number') {
+      switch (code) {
+        case ReturnCodes.POSITIVE_ACKNOWLEDGEMENT:
+          debug('positive acknowledgement');
+          return true;
+        case ReturnCodes.NEGATIVE_ACKNOWLEDGEMENT:
+          debug('negative acknowledgement');
+          return false;
+        case ReturnCodes.INVALID_STATUS_TRANSMISSION_ARGUMENT:
```
arsalansufi, line 423: Is it possible that a blob might contain a _valid_ status transmission request? Would we need to handle those responses as well here?

arsalansufi: This is already handled - @kshen0 walked me through it in person

That said, we're now taking a step back and leaning toward the likely safer alternative of just always reading/clearing the in buffer after the print command, instead of catching unexpected responses from a previous print command while reading the response for a subsequent command

libs/custom-paper-handler/src/driver/coders.ts
```diff
@@ -137,6 +137,41 @@ export type RealTimeExchangeResponseWithoutData = CoderType<
   typeof RealTimeExchangeResponseWithoutData
 >;
 
+export const RealTimeStatusTransmission = message({
```
adghayes, line 140: In what way is this different from the `PrinterStatusRealTimeExchangeResponse` above?

libs/custom-paper-handler/src/driver/driver.ts
adghayes: This makes to me sense provided we aren't worried about other endpoints accidentally triggering the same command.

libs/custom-paper-handler/src/driver/coders.ts
kshen0: It's the same but lacks the real-time response header. I can factor it out and use spread syntax.

adghayes: Your call, just wanted to make sure the code duplication was intentional.

libs/custom-paper-handler/src/driver/driver.test.ts
kshen0: Confirmed this by running `enablePrint`, `printChunk` with the bad chunk, and `setScanDirection` with 5s sleep between each. Reading the transfer in buffer after all commands are run results in 3 discrete, parsable responses of `0x06`, the `RealTimeStatusTransmission` response, and `0x06`.

libs/custom-paper-handler/src/driver/driver.ts
kshen0: For the record it is possible to send a well-formed status transmission request `0x10 0x04 0x14` and we do handle it appropriately.

We looked into reading the buffer after every print command, but with the WebUSB API once a `transferInGeneric` is initiated it can't be canceled. So in the vast majority of cases we would block on the `transferInGeneric` command.

We can avoid blocking the caller with standard use of `Promise.race` as we do with `clearGenericInBuffer`. But that still initiates a `transferInGeneric` request, so the next response to a valid command over that buffer will be read and discarded by the `transferInGeneric` request that "timed out"`*`

For simplicity I believe it's best to `transferInGeneric` only when we are guaranteed a response on the transfer-in buffer, and handle unexpected extra data as it comes up.

`*` Example in more detail:

* `printChunk` is called with unproblematic data
* Initiate a `transferInGeneric` (call this `transfer in A`) that unblocks the caller after a timeout but does not (cannot) cancel the transfer in request
* After timeout, state machine continues to issue next generic command `setScanDirection('backward')`
  * This and every other generic command involves 2 steps: issuing the command over the generic transfer out buffer and reading the response over the generic transfer in buffer (call this `transfer in B`)
* After execution of `setScanDirection`, device responds on transfer in buffer with `0x06`
* `transfer in A` request resolves by reading `0x06` from the transfer in buffer. The buffer is now empty
* `transfer in B` is now blocked until the subsequent command responds over the transfer in buffer

kshen0: > This makes to me sense provided we aren't worried about other endpoints accidentally triggering the same command.

The only command that sends arbitrary data is `bufferChunk` and its caller `printChunk`.

libs/custom-paper-handler/src/driver/coders.ts
kshen0: Good call, [5802d37](https://github.com/votingworks/vxsuite/pull/4846/commits/5802d37fbb73a3e26c00135302c8f4a12fc44a6e)

arsalansufi (approved):
LGTM!

Issue #4601, opened 2024-02-15
printer detection blip in dev


adghayes:
In development, I've noticed that with the new `libs/printing`, the printer is usually lost and found again when printing. The reason for the `vxsuite` blip is the underlying CUPS blip. `lpinfo` loses the device URI for small fractions of a second. This seems to be made more frequent by triggering a print or polling excessively (although that may just be due to larger samples).

I checked on a prod image and _did not_ see the same behavior. For now, I want to track the blip on this issue and keep an eye on the issue in prod, but not change the printer detection yet. I think this is a VM issue.

If we used the `usb` connection and disconnection events, which appear to be stable, we wouldn't have this problem. So that is plan B.

adghayes:
Closing as it has never been reported in prod.

PR #4851, opened 2024-05-15
VxMark Print Revamp

adghayes:
## Overview

Using the backend rendering strategy developed for #4788, render and print ballots from the backend on VxMark. Use printer status from `libs/printing` rather than `kiosk-browser`. This allows us to remove a chunk of duplicated printing code and the whole `printElement` infrastructure which was somewhat prone to difficult to detect visual bugs. There's still a bunch of `kiosk-browser` printing-related stuff but it'll be easier and cleaner to remove that after removing the final `kiosk-browser` device-related pieces (also in VxMark) in a follow-up. And now we have a "mock printer" in the dev dock when developing in VxMark (although we don't really need to do that much, nice to have).

For the diagnostics screen printer status, I just deleted what we had and dropped in what was developed for VxAdmin. There's one use case there that I will need to create a follow-up ticket for (negative toner levels).

Generally, since we're not investing in VxMark, I didn't follow the printing refactor with other clean up in `mark` that might be possible now. Not worth the time.

## Testing Plan

Tested with hardware (in dev) and tested with the mock printer. Confirmed that printer disconnection still resets a voting session. 

When you print, there's a blip where the printer is not detected after print. I've already seen and documented this #4601 but I think it's only in a VM. Now that we have a new VxAdmin image flashed at the bakery, I can confirm tomorrow.

## Visual Change

Printer Diagnostics After:

Screen Shot 2024-05-14 at 11 02 25 PM

Printer Diagnostics Before:

Screen Shot 2024-05-14 at 11 02 44 PM

## Checklist

Logging about printer disconnection lost from `libs/ui` is already covered in `libs/printing`.


apps/mark/backend/src/util/print_ballot.tsx
```diff
@@ -0,0 +1,62 @@
...
+  languageCode,
+}: PrintBallotProps): Promise<void> {
+  const electionDefinition = assertDefined(store.getElectionDefinition());
+  const isLiveMode = !store.getTestMode();
+
+  const uiStringsPackage: UiStringsPackage = {};
+  for (const language of store.getUiStringsStore().getLanguages()) {
+    uiStringsPackage[language] = assertDefined(
+      store.getUiStringsStore().getUiStrings(language)
+    );
```
adghayes, line 36: @kofi-q question about this assert - if `getLanguages` returns a language code, does than mean we have `UiStrings` for that language? Or might it return "English" but not have overridden English strings, because that's our default language. Affects whether this assert is safe or not.

Alternatevely, could add a method to the UiStringsStore to get the whole package rather than reconstructing it.

apps/mark/frontend/src/pages/diagnostics_screen.test.tsx
```diff
@@ -87,115 +100,23 @@ describe('System Diagnostics screen: Computer section', () => {
...
-      state: 'idle',
-      stateReasons: ['none'],
-      markerInfos: [mockMarkerInfo({ level: -2 })],
-    });
-    renderScreen({ hardware });
-
-    const tonerLevelText = await screen.findByText('Toner level: Unknown');
+    const tonerLevelText = screen.getByText('Toner Level: 2%');
     expectToHaveWarningIcon(tonerLevelText);
   });
```
adghayes, line 121: @jonahkagan this is a case (negative toner levels) that I never handled in the print diagnostic for VVSG. Is this reproducible on our HP printers, or only applies to other printers?

apps/mark/backend/src/util/print_ballot.tsx
eventualbuddha: IMO, this does feel like the kind of thing the `UiStringsStore` should be able to do and is not unique to printing a ballot.

apps/mark/frontend/test/helpers/mock_api_client.tsx
```diff
@@ -31,6 +35,16 @@ import { mockMachineConfig } from './mock_machine_config';
 import { initialElectionState } from '../../src/app_root';
 import { ApiProvider } from '../../src/api_provider';
 
+// the below is copied from libs/printing to avoid importing a backend package
+export const MOCK_PRINTER_CONFIG: PrinterConfig = {
+  label: 'HP LaserJet Pro M404n',
+  vendorId: 1008,
+  productId: 49450,
+  baseDeviceUri: 'usb://HP/LaserJet%20Pro%20M404-M405',
```
eventualbuddha, line 43: VID/PID are typically written in hex notation. Searching the internet for the hex values yields better results.
```suggestion
  vendorId: 0x03f0,
  productId: 0xc12a,
```
I realize this was likely just copied from the JSON file where we can't use hex notation. I don't think there's actually a good reason to have `configs.json` as JSON. We should probably just make it a TS file.

eventualbuddha (approved):

apps/mark/backend/src/util/print_ballot.tsx
kofi-q: Good question. This assertion makes sense - we only add languages when loading strings for that language. I do like the idea of a `getAllUiStrings`-type store method, now that you mention it, but whatever's simpler - can't think of other use cases for it at this point

apps/mark/frontend/src/pages/diagnostics_screen.test.tsx
jonahkagan: Unfortunately I have no memory of this, but I think I originally built this based on output from an HP printer, so it seems likely that's where I saw this behavior

apps/mark/backend/src/util/print_ballot.tsx
adghayes: Helper added.

apps/mark/frontend/src/pages/diagnostics_screen.test.tsx
adghayes: Making a note for myself: https://github.com/votingworks/vxsuite/issues/4854

apps/mark/frontend/test/helpers/mock_api_client.tsx
adghayes: I think part of the reason is that it used to exist in `vxsuite-complete-system` and be read in as a config file for `kiosk-browser`. Now it doesn't need to be. I do like how it's co-located with the `.ppd` files that it references, outside of the `src` directory. Going to leave as-is for now to move on to next PR.

PR #4853, opened 2024-05-15
[BMD] Reduce paper ballot text size as contest count increases

kofi-q:
## Overview

Resolves https://github.com/votingworks/vxsuite/issues/4473

Making a couple updates to increase the number of contests we can fit on a single page of the BMD paper ballot:
- Reducing overall spacing where possible
- Gradually reducing font size once we go past 30 contests (which is a conservative limit based on how many contests we can fit on a fairly complex multi-language ballot at the VVSG minimum 10pt font size).

Limiting the minimum font size to `5pt`, which seems to be as small as we can go before seriously sacrificing readability on the prints -- will need to do some more testing on the VSAP thermal printer to make sure 5pt font is readable at that resolution and adjust accordingly.

### Max supported contests (30) at 10pt font, with wiggle room for higher-complexity ballots/more verbose languages:
Screenshot 2024-05-15 at 15 25 00

### Font scaling based on contest count

https://github.com/votingworks/vxsuite/assets/264902/57070b31-1121-487f-8da5-3fa67d42d48e

## Testing Plan
- Manual via both storybook and printing to PDF/physical sheets 

## Checklist


libs/ui/src/bmd_paper_ballot.test.tsx
```diff
@@ -117,9 +117,6 @@ test('BmdPaperBallot renders votes for candidate contests and yes-no contests',
   screen.getByText(
     hasTextAcrossElements(/Question B: Separation of Powers.?No/)
   );
-
-  // Use a snapshot to avoid unintentional regressions to general layout
-  expect(container).toMatchSnapshot();
```
kshen0, line 122: Just making sure I understand, we can remove this because rendering and snapshot testing was moved to the backend?

kshen0 (approved):

libs/ui/src/bmd_paper_ballot.test.tsx
kofi-q: yeah, it's usefulness was a little questionable to begin with, since it was just comparing html and css code (we've been stripping out this type of snapshot from tests whenever possible) and it's redundant now that we have pixel-based diffing in the server-side tests now

PR #4857, opened 2024-05-16
transition from eject_to_front when paper is in input

kshen0:
## Overview
Fixes an issue where ejecting to front may sometimes result in the paper being held in the front by friction with the paper tray. In this case, the input sensors are tripped and the state machine gets stuck in `eject_to_front` state because we currently only check for complete absence of paper. 

This PR adds transitions for when paper triggers some or all input sensors, allowing the state machine to exit `eject_to_front` state.

## Demo Video or Screenshot

Note in these videos I have to manually stop the paper because the paper tray of the dev unit has less friction than the paper tray of the prod unit. On the dev unit, the paper will slide all the way out of the input and avoid tripping input sensors.

Before: when paper trips input sensors, state machine is stuck in `ejecting_to_front` state.

https://github.com/votingworks/vxsuite/assets/1060688/da1b4036-7540-41ab-b3c5-36ff84b205b7

After: when paper trips input sensors but not internal sensors, state machine transitions to reset the session and then back to initial `not_accepting_paper` state. 

https://github.com/votingworks/vxsuite/assets/1060688/be1f9eb9-8ef6-4c64-b8e1-7a467851c887

## Testing Plan
- added test for state transition
- tested manually

## Checklist

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/scanner_status.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/scanner_status.ts
@@ -28,7 +28,7 @@ export function isPaperInScanner(
   );
 }
 
-export function isPaperInInput(
+export function isPaperPartiallyInInput(
   paperHandlerStatus: PaperHandlerStatus
 ): boolean {
   return (
@@ -53,7 +53,7 @@ export function isPaperAnywhere(
   paperHandlerStatus: PaperHandlerStatus
 ): boolean {
   return (
-    isPaperInInput(paperHandlerStatus) ||
+    isPaperPartiallyInInput(paperHandlerStatus) ||
     isPaperInOutput(paperHandlerStatus) ||
     isPaperInScanner(paperHandlerStatus)
   );
--- a/apps/mark-scan/backend/src/custom-paper-handler/scanner_status.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/scanner_status.ts
@@ -28,7 +28,7 @@ export function isPaperInScanner(
   );
 }
 
-export function isPaperPartiallyInInput(
+export function isPaperInInput(
   paperHandlerStatus: PaperHandlerStatus
 ): boolean {
   return (
@@ -53,7 +53,7 @@ export function isPaperAnywhere(
   paperHandlerStatus: PaperHandlerStatus
 ): boolean {
   return (
-    isPaperPartiallyInInput(paperHandlerStatus) ||
+    isPaperInInput(paperHandlerStatus) ||
     isPaperInOutput(paperHandlerStatus) ||
     isPaperInScanner(paperHandlerStatus)
   );
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
@@ -249,6 +249,25 @@ describe('not_accepting_paper', () => {
   });
 });
 
+describe('eject_to_front', () => {
+  test.each([
+    {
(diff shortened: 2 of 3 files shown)


apps/mark-scan/backend/src/custom-paper-handler/scanner_status.ts
```diff
@@ -28,7 +28,7 @@ export function isPaperInScanner(
   );
 }
 
-export function isPaperInInput(
+export function isPaperPartiallyInInput(
```
adghayes, line 31: I don't care too much about naming, but IMO this rename is more confusing than previously, because if you told me the paper was partially in the input I would assume that some of the sensors are not firing, but the paper fully being in the input does trigger `isPaperPartiallyInInput`.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -705,6 +705,11 @@ export function buildMachine(
               entry: ['ejectPaperToFront'],
               on: {
                 NO_PAPER_ANYWHERE: 'resetting_state_machine_after_success',
+                // Sometimes paper ejected to front is not held by the motors but
+                // will still trigger input sensors.
+                // In this case we still want to allow the machine to progress.
+                PAPER_READY_TO_LOAD: 'resetting_state_machine_after_success',
+                PAPER_IN_INPUT: 'resetting_state_machine_after_success',
```
adghayes, line 712: These events will file before the paper is fully ejected. In theory, I might want to check that paper is detected in the input but also that paper is no longer detected inside the scanners (CIS sensors, etc.) to exit the `ejectPaperToFront` state, otherwise we will prematurely exit the state. Practically, may not matter.

It definitely doesn't matter for solving the problem you're trying to solve, so not blocking on it.

apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
```diff
@@ -249,6 +249,25 @@ describe('not_accepting_paper', () => {
...
+describe('eject_to_front', () => {
+  test.each([
+    {
+      description: 'paper is absent',
+      setStatus: () => setMockDeviceStatus(getDefaultPaperHandlerStatus()),
+    },
+    {
+      description: 'paper is in front input',
+      setStatus: () => setMockDeviceStatus(getPaperInFrontStatus()),
+    },
```
adghayes, line 261: Do you want coverage for the partially detected state?

adghayes (approved):
Approved with some questions!

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Great catch. Recapping offline conversation - we should fix this in the long run but it introduces complication to the state machine because
* the way we convert printer/scanner status to events makes it difficult to rely solely on event type for union and intersection statuses
* checking in the state machine requires adding a new state because guarded transitions don't support returning a promise, and getting device status is async

apps/mark-scan/backend/src/custom-paper-handler/scanner_status.ts
kshen0: Reverted naming!

apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
kshen0: [20e077b](https://github.com/votingworks/vxsuite/pull/4857/commits/20e077b8debd58ecbe9e70ba98f9b89b18437746)

PR #4861, opened 2024-05-17
Update Print Margins on Fujitsu

### LINKED ISSUE #4839: Fujitsu Printer Integration: Print Margin
We must adjust the top and bottom print margin on the pages printed via the Fujitsu, otherwise it will have too much white space at the top and too little at the bottom, due to the distance between the printer and the tear bar.

adghayes:
## Overview

Closes #4839.

```
/**
 * While loaded, the paper must be fed through the paper output slot with the
 * tear bar. There is a distance between the output slot and the printhead, however,
 * which means that a certain chunk at the top of each page is unprintable. To
 * account for this, we redistribute a certain amount of the top margin to the
 * bottom margin. This must be calibrated based off of the hardware.
 */
```

This initial margin is based off a measurement from the CAD provided by Pump Studios. We may want to tweak it (increase if anything).


adghayes:
cc @mattroe noting that I'm fixing this issue now but, at the next opportunity to run our software on the v4 unit, we can see if we need to further adjust it

kofi-q (approved):
🖨️

PR #4865, opened 2024-05-20
apps/scan: Only listen for scanner events at root

jonahkagan:
## Overview

We don't actually want to have multiple active listeners at a time for scanner events. The listener at the root will send a state machine event that can be caught in any state. Registering an additional listener within a state will just create a duplicate state machine event. This hasn't been causing problems because we've been transitioning away from the states with listeners before that happens, but it still presents a risk and is unnecessary.

## Checklist


eventualbuddha (approved):
Seems sane so long as you've done some manual testing.

jonahkagan:
Yes, I manually tested

PR #4875, opened 2024-05-22
apps/scan: Set PDI scanner paper length to prevent back-to-back ballots

jonahkagan:
## Overview
Task: #4825 

Previously, if you timed it just right, it was possible to trick the PDI scanner into thinking it was scanning one long ballot by inserting two ballots in quick succession back to back. In this case, the first ballot would be ejected out of the rear of the scanner into the ballot box before the issue would be detected by the interpreter.

To prevent this, we set the max paper length setting provided by the PDI scanner. After scanning (approximately) this length, the PDI scanner declares a jam and stops the motors. I found that the optimal length for stopping the motors soon enough to prevent the first ballot from being ejected was 0.5in shorter than the actual length of the ballot.
## Demo Video or Screenshot
Before

https://github.com/votingworks/vxsuite/assets/530106/e41ee457-4bd1-4639-850d-a9c52efefa36

After

https://github.com/votingworks/vxsuite/assets/530106/4b64265e-269c-4c1b-b47c-e1a0969198c1

## Testing Plan
- Updated automated tests
- Manual test with letter and legal ballots

## Checklist

### DIFF
--- a/libs/pdi-scanner/src/rust/client.rs
+++ b/libs/pdi-scanner/src/rust/client.rs
@@ -768,6 +768,7 @@ impl<T> Client<T> {
     pub fn send_enable_scan_commands(
         &mut self,
         double_feed_detection_mode: DoubleFeedDetectionMode,
+        paper_length_inches: f32,
     ) -> Result<()> {
         let timeout = Duration::from_secs(5);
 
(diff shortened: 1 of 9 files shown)


libs/pdi-scanner/src/rust/client.rs
```diff
@@ -808,12 +809,18 @@ impl<T> Client<T> {
         self.set_threshold(Side::Bottom, ClampedPercentage::new_unchecked(75), timeout)?;
         // OUT SetRequiredInputSensorsRequest { sensors: 2 }
         self.set_required_input_sensors(2)?;
-        // OUT SetLengthOfDocumentToScanRequest { length_byte: 32, unit_byte: None }
-        self.set_length_of_document_to_scan(0.0)?;
+
+        // Set the max lenght of the document to scan to 0.5" less than the
```
eventualbuddha, line 813:
```suggestion
        // Set the max length of the document to scan to 0.5" less than the
```

eventualbuddha (approved):

PR #4877, opened 2024-05-22
fix(ui): cache modal `appElement` to fix tests

eventualbuddha:
## Overview

In tests an element with id "root" typically does not exist, so `Modal` falls back to `document.body.firstElementChild`. However, `ReactModal` (under our `Modal`) adds an element with attribute `data-react-modal-body-trap` at the root. So if the `Modal` is ever re-rendered, it'll get a different `appElement` the next time around. This is important because when the modal first appears `ReactModal` adds `aria-hidden` to the `appElement`, and then when the modal goes away it removes the `aria-hidden` attribute. However, if the `appElement` changes between renders then you will end up with the `aria-hidden` attribute sticking around, which means all the `screen.findByRole` calls will fail because they ignore anything inside an element with `aria-hidden`.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests.

### DIFF
--- a/apps/admin/frontend/src/components/smartcard_modal/smartcard_modal.test.tsx
+++ b/apps/admin/frontend/src/components/smartcard_modal/smartcard_modal.test.tsx
@@ -378,9 +378,7 @@ test('Programming election manager and poll worker smartcards', async () => {
     await waitFor(() =>
       expect(screen.queryByRole('alertdialog')).not.toBeInTheDocument()
     );
-    // For some reason, we need to use the "hidden" option here, even though the
-    // heading is not hidden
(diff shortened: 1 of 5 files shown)


kofi-q (approved):
🎉  glad this is finally fixed!

Should also be able to remove these now: https://github.com/search?q=repo%3Avotingworks%2Fvxsuite+hackActuallyCleanUpReactModal&type=code

eventualbuddha:
Just for fun, I decided to see when the bug this fixed was introduced. Looks like April 2019: 38a83ef4a25c112d912af52c677555dbba7d118e.

PR #4878, opened 2024-05-23
test: remove `react-modal` workaround

eventualbuddha:
Follow-up to #4877. It wasn't an issue with `react-modal`, but with our usage of it.


adghayes (approved):
glad we finally made sense of this

PR #4883, opened 2024-05-28
apps/scan: Prevent test flakiness by combining ready to scan states

jonahkagan:
## Overview
I saw a [flaky test](https://app.circleci.com/pipelines/github/votingworks/vxsuite/15707/workflows/a9bed89d-e04f-44af-8327-6bfe135134f0/jobs/589301) where there was an "unexpected event" error with the paper status `SCANNER_READY_TO_SCAN` while the test was waiting for the `scanning` state. So I looked directly before the `scanning` state and found a loop between the two states `hardware_ready_to_scan` (which has paper status handlers) and `check_app_ready_to_scan` (which doesn't). I wasn't able to repro the error, but my guess is that the event was somehow caught in the `check_app_ready_to_scan` state.

I also realized that the way these states are currently set up, there's no delay in polling the paper status, since we loop back and forth between the two states immediately.

I decided to combine the states into a single state with a single child state that self-loops. Hopefully that will prevent the test flakiness and preserve the paper status polling interval.

## Checklist


apps/scan/backend/test/helpers/custom_helpers.ts
```diff
@@ -228,10 +228,6 @@ export function simulateScan(
 ): void {
   let didScan = false;
   mockScanner.getStatus.mockImplementation(async () => {
-    // Simulate an actual async call so that we don't get trapped in a sync
-    // execution loop switching between hardware_ready_to_scan and
-    // check_app_ready_to_scan states.
-    await sleep(1);
```
jonahkagan, line 234: no longer necessary now that there's a delay

eventualbuddha (approved):

Issue #4689, opened 2024-03-18
vxmarkscan: update audio instructions for dual switch user


arsalansufi:
To address the edge case where a dual switch user relies on audio feedback, we'll need to use different audio for the contest navigation instructions played on each contest page when a PAT device is plugged in. Current instructions are very specific to the tactile controller and don't make sense for dual switch inputs. This should be set based on detecting a connection to the dual switch input.

PR #4885, opened 2024-05-28
[MarkScan] Switch between instructions for tactile controller and PAT device

kofi-q:
## Overview

Resolves https://github.com/votingworks/vxsuite/issues/4689

Adding an `AssistiveTechInstructions` component to switch between audio instructions for either the tactile controller or PAT devices, based on whether or not a PAT device is connected. Using the switch component in the relevant parts of the MarkScan flow.

## Testing Plan
- Unit tests for the new helper components
- Manual verification locally with generated audio

## Checklist


libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -308,6 +314,19 @@ export const appStrings = {
...
   ),
 
+  instructionsBmdBallotNavigationMarkScanPatDevice: () => (
+    <UiString uiStringKey="instructionsBmdBallotNavigationMarkScanPatDevice">
+      When voting with the text-to-speech audio, use your personal assistive
+      device to navigate your ballot. To navigate through contest choices, use
+      the move input. To select or unselect a contest choice as your vote, use
+      the select input. To navigate through the contests, use the move input to
+      navigate to the controls labelled "next" and "back" and use the select
+      input to navigate to a new contest. To get started, use the move input to
```
kshen0, line 324: Are "to navigate through the contest" and "to navigate to a new contest" redundant? Or maybe subtly different because the latter informs the voter that using the "next" and "back" buttons are the specific actions to take in order to perform the former?

kshen0: Also, I can't recall if focus will jump to "Next" when a voter is using both PAT input and speech-to-text. Do you think it's worth mentioning here if so?

libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -341,6 +360,14 @@ export const appStrings = {
     </UiString>
   ),
 
+  instructionsBmdContestNavigationPatDevice: () => (
+    <UiString uiStringKey="instructionsBmdContestNavigationPatDevice">
+      To navigate through the contest choices, use the move input. To move to
```
kshen0, line 365: "To move" could be confusing with our terminology for the "move" input. What do you think about "To advance to the next contest ..."?

libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -407,6 +441,14 @@ export const appStrings = {
     </UiString>
   ),
 
+  instructionsBmdReviewPageChangingVotesPatDevice: () => (
+    <UiString uiStringKey="instructionsBmdReviewPageChangingVotesPatDevice">
+      To change your vote in any contest, use the select button to navigate to
+      that contest. When you are finished making your ballot selections and
+      ready to print your ballot, use the right button to print your ballot.
```
kshen0, line 448: I think "right button" is left over from the ATI controller instructions.

libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -415,19 +457,42 @@ export const appStrings = {
...
   instructionsBmdWriteInFormNavigation: () => (
     <UiString uiStringKey="instructionsBmdWriteInFormNavigation">
       Use the up and down buttons to navigate between the letters of a standard
       keyboard. Use the select button to select the current letter.
     </UiString>
   ),
 
+  instructionsBmdWriteInFormNavigationPatDevice: () => (
+    <UiString uiStringKey="instructionsBmdWriteInFormNavigationPatDevice">
+      Use the move input to navigate between the letters of a standard keyboard.
```
kshen0, line 491: I'm not actually sure if QWERTY keyboards are standard for PAT users. They very well could be, but it might be more specific to say QWERTY keyboard (if our TTS can handle that).

libs/ui/src/accessible_controllers/pat_device_context.tsx
```diff
@@ -0,0 +1,35 @@
...
+export function useIsPatDeviceConnected(): boolean {
+  return usePatDeviceContext()?.isPatDeviceConnected || false;
+}
+
+export interface PatDeviceContextProviderProps {
+  children: React.ReactNode;
+  isPatDeviceConnected: boolean;
+}
+
+export function PatDeviceContextProvider(
```
kshen0, line 25: Thanks for adding this!

libs/ui/src/ui_strings/app_strings.tsx
kofi-q: Yeah, good point WRT the autofocus behaviour - will try to write something up that doesn't sound too confusing. Will also see if I can get reword to get rid of the repeated "to navigate"

libs/ui/src/ui_strings/app_strings.tsx
kofi-q: Good call - thanks

libs/ui/src/ui_strings/app_strings.tsx
kofi-q: Thanks for catching - forgot to modify this string

libs/ui/src/ui_strings/app_strings.tsx
kofi-q: Good point -- the audio-only virtual keyboard experience is currently incomplete/not ideal for both tactile controller voters and PAT device voters and we've talked about changing the scroll order to alphabetical for audio-only users, or implementing some T9-type UX for them. We'll revisit these instructions at that point and can treat these as short-term stand-ins for now

kofi-q:
cc @mattroe for copy review -- this file's probably the easiest to scan through for the new PAT strings: https://github.com/votingworks/vxsuite/pull/4885/files#diff-ea19341f7e7f0831bf1d83d80797d6dd4c3b642ea4262cf0f6c62daa6ce4c1b1

kshen0 (approved):

PR #4752, opened 2024-04-12
Paper handler diagnostic

kshen0:
Recommend reviewing by commit.

## Overview
https://github.com/votingworks/vxsuite/issues/4718

Adds a diagnostic flow for the paper handler.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/33ab6d57-5f19-4c53-af9d-533527573555

Screenshot 2024-04-22 at 6 03 32 PM

Screenshot 2024-04-22 at 6 03 38 PM

## Testing Plan
- added tests

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [x] I have added a screenshot and/or video to this PR to demo the change
- [x] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #4887, opened 2024-05-29
[mark-scan] Clean up mock ballots

kshen0:
## Overview

PR 1 of ? in breaking up https://github.com/votingworks/vxsuite/pull/4752.

Now that we have backend ballot rendering, arguably the best solution is to replace these mock files with ballots rendered on the fly. But with the goal of closing out the diagnostic tasks, I'd like to leave that for later. This PR at least removes mock ballots we don't need and isolates the mock ballots we do need to a clear `test_utils` dir.

## Demo Video or Screenshot
No user facing change

## Testing Plan
These files are used by tests only

## Checklist
N/A


apps/mark-scan/backend/src/custom-paper-handler/cli/state_machine_cli.ts
```diff
@@ -1,129 +0,0 @@
-/* istanbul ignore next */
```
kshen0, line 1: This CLI is the only caller of `apps/mark-scan/backend/src/custom-paper-handler/fixtures/ballot-pdf-data.bin`. It doesn't add value since the state machine is more complicated now, but it requires occasional maintenance when types etc. change. Since we have to maintain it but get no value from it we should just delete it.

apps/mark-scan/backend/src/custom-paper-handler/filepaths.ts
```diff
@@ -1,9 +0,0 @@
-import { join } from 'path';
-
-export function getSampleBallotFilepath(): string {
-  return join(
-    __dirname,
-    'fixtures',
-    'bmd-ballot-general-north-springfield-style-5.jpg'
-  );
-}
```
kshen0, line 9: Moved to `test_utils/utils.ts`

PR #4890, opened 2024-05-29
[mark-scan] Add backend for paper handler diagnostic

kshen0:
## Overview
Backend support for the paper handler diagnostic flow in VxMarkScan. Part 3 of 4 in splitting up https://github.com/votingworks/vxsuite/pull/4752.

Still need to add test coverage but this is ready for a first pass if you want to get a head start.

* Adds diagnostic flow to the state machine
* Adds an API endpoint to start the diagnostic flow

## Demo Video or Screenshot
* No frontend changes

## Testing Plan
- [ ] add tests

## Checklist


apps/mark-scan/backend/src/custom-paper-handler/diagnostic/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+): Promise<Result<ElectionDefinition, Error>> {
+  const electionFileReadResult = await readFile(path, {
+    maxSize: FILE_MAX_SIZE,
+  });
+  if (electionFileReadResult.isErr()) {
+    return err(new Error(`Failed to read election file at ${path}`));
+  }
+  const electionData = electionFileReadResult.ok().toString();
+  return safeParseElectionDefinition(electionData);
+}
```
jonahkagan, line 37: could you use `readElection` from `libs/fs` here?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -82,12 +92,38 @@ interface Context {
...
+  diagnosticError?: DiagnosticError;
 }
 
 function assign(arg: Assigner<Context, any> | PropertyAssigner<Context, any>) {
   return xassign<Context, any>(arg);
 }
 
+class DiagnosticError extends Error {
+  constructor(
+    userFacingMessage: string,
```
jonahkagan, line 105: I think we've generally been trying to follow the pattern of having the frontend/UI define the error message copy and having backends just use a string error code to indicate which error happened.

jonahkagan (approved):

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: Agree that the backend defining error messages is hard to follow.

In this implementation, `DiagnosticError.userFacingMessage` eventually gets stored on the `DiagnosticRecord`. Since there's an existing pattern of messages being written to that table (from the frontend), I think there's a bigger refactor that would need to happen where

1. All writers of `DiagnosticRecord` pass an error code
2. We have a mapping of code to user-facing string
3. We reference those strings from report and diagnostic screen components

To avoid muddling the error landscape further I just replaced `DiagnosticError` with `Error`. Raw error messages will still be logged but no user-facing error will be saved to `DiagnosticRecord`.

PR #4892, opened 2024-05-29
apps/scan: PDI scanner double feed detection calibration

### LINKED ISSUE #4729: vxscan: pdi msd calibration

jonahkagan:
## Overview

Closes #4729 

Adds double feed detection calibration to the Election Manager screen. Our flow is a straightforward translation of the PDI scanner's recommended calibration flow:
first insert two blank sheets at once, then insert one blank sheet.

Note that there's no way to cancel calibration once the scanner has been told to expect the sheets to be inserted, so we can't allow for canceling. Instead, we have to rely on the PDI scanner's internal timeout (~10s) to exit the flow. In that case, we simply show a failure message to indicate to the user that the calibration wasn't completed.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/530106/f884a21b-ab24-47a7-beaa-685df61085ea

## Testing Plan
Manually tested and added automated tests

## Checklist


libs/pdi-scanner/src/rust/client.rs
```diff
@@ -649,6 +659,7 @@ impl<T> Client<T> {
         &mut self,
         calibration_type: DoubleFeedDetectionCalibrationType,
     ) -> Result<()> {
+        self.set_double_feed_detection_mode(DoubleFeedDetectionMode::Disabled)?;
```
jonahkagan, line 662: Some docs from PDI indicated that it was a "common sequence" to disable DFD before calibrating, so I replicated that here to be safe

libs/pdi-scanner/src/rust/client.rs
```diff
@@ -722,6 +733,26 @@ impl<T> Client<T> {
         )
     }
 
+    pub fn get_double_feed_detection_calibration_config(
+        &mut self,
+        timeout: Duration,
```
eventualbuddha, line 738: We haven’t been super rigorous with our timeout values, but the way this function is implemented means that it could actually take up to `< timeout * 4` because each of the sub-calls is allowed to take up to `timeout`.

eventualbuddha (approved):

libs/pdi-scanner/src/rust/client.rs
jonahkagan: Yeah, do you think that's an issue? Or just a little confusing?

Issue #4889, opened 2024-05-29
vxmarkscan: replace ballot image fixtures with on-the-fly ballot generation


kshen0:
MarkScan now supports backend ballot rendering so we can generate ballot images on the fly in tests. We should replace ballot image fixtures, which are hard to maintain and prone to drift, with autogenerated ballots.

PR #4911, opened 2024-05-31
Render diagnostic ballots on the fly

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/4889

Renders ballots used exclusively by the hardware diagnostic on the fly instead of loading them from fixture images. This should reduce maintenance and make it easier to change the diagnostic test ballot in the future.

## Demo Video or Screenshot

TODO

## Testing Plan

* Moved tests from `state_machine.test.ts` to `app.diagnostics.test.ts`. An unknown problem was preventing `chromium` from launching in the former.

Verified the diagnostic still works on hardware.

## Checklist
N/A


apps/mark-scan/backend/src/app.diagnostics.test.ts
```diff
@@ -227,3 +273,93 @@ test('failure saving the readiness report', async () => {
...
+    await apiClient.startPaperHandlerDiagnostic();
+    await waitForStatus('paper_handler_diagnostic.prompt_for_paper');
+
+    mockDriverStatus(getPaperInFrontStatus());
+    await waitForStatus('paper_handler_diagnostic.load_paper');
+
+    mockDriverStatus(getPaperParkedStatus());
+    await waitForStatus('paper_handler_diagnostic.print_ballot_fixture');
+    // Chromium, used by print_ballot_fixture, needs ~150ms to spin up.
+    // 75ms with retries is long enough.
```
adghayes, line 327: Update comment after `fixup`?

apps/mark-scan/backend/src/app.diagnostics.test.ts
```diff
@@ -227,3 +273,93 @@ test('failure saving the readiness report', async () => {
...
 });
+
+function mockDriverStatus(status: PaperHandlerStatus) {
+  mockOf(driver.getPaperHandlerStatus).mockResolvedValue(status);
+}
+
+describe('paper handler diagnostic', () => {
+  test('success', async () => {
+    const electionDefinition = (
+      await readElection(DIAGNOSTIC_ELECTION_PATH)
```
adghayes, line 284: Why do you load an election definition from file, and accommodate that in your build step, rather than just load an election from `libs/fixtures`?

apps/mark-scan/backend/test/ballot_helpers.ts
```diff
@@ -0,0 +1,19 @@
+import { InterpretFileResult } from '@votingworks/ballot-interpreter';
+import { SheetOf } from '@votingworks/types';
+
+export const MOCK_IMAGE: ImageData = {
+  colorSpace: 'srgb',
```
adghayes, line 5: Note this may have a logical merge conflict with #1732.

apps/mark-scan/backend/src/custom-paper-handler/diagnostic/utils.ts
```diff
@@ -1,44 +1,44 @@
...
+  );
 }
 
-export async function getMockBallotPdfData(): Promise<Buffer> {
-  const result = await readFile(DIAGNOSTIC_MOCK_BALLOT_PDF_PATH, {
-    maxSize: FILE_MAX_SIZE,
-  });
-  // Error must be handled by caller
-  return result.unsafeUnwrap();
+export async function getDiagnosticMockBallotImagePath(): Promise<string> {
```
adghayes, line 31: A comment that this is only for testing would be helpful.

adghayes (approved):
LGTM! Some code nits.

apps/mark-scan/backend/src/app.diagnostics.test.ts
kshen0: A few reasons:

1. This is a user-facing mock election so we wanted the contests to make it abundantly clear it's a mock election with names like "Candidate A" and "Choice B". I believe our existing fixtures all use real-sounding names.

2. I was concerned that fixtures are associated with testing and don't have a stable contract. We'd be in danger of changes to fixtures accidentally breaking this production flow. If that's wrong or if we have a construct for mock elections used in production, I'd be happy to fold this into that construct!

adghayes: I think reason #1 makes sense and is a good enough reason. I think our testing should guard against #2. We don't have another construct but if we need an election definition for another ballot, would be excellent to use the one you've created here.

PR #4915, opened 2024-06-03
Create new i18next instance for every backend render

jonahkagan:
## Overview

Instead of relying on a single global i18next instance and updating all of its resources every render, create a new instance every render. This is just a little cleanup to make this flow less confusing. Now each render more clearly loads the latest translations from the store and creates a language context to be used by the React components.
## Demo Video or Screenshot
N/A

## Testing Plan
Relying on existing automated tests
## Checklist

### DIFF
--- a/libs/ui/src/ui_strings/language_context.tsx
+++ b/libs/ui/src/ui_strings/language_context.tsx
@@ -1,9 +1,14 @@
 import React from 'react';
-import i18next, { i18n } from 'i18next';
-import { initReactI18next, useSSR, useTranslation } from 'react-i18next';
+import i18next, { InitOptions, i18n } from 'i18next';
+import { initReactI18next, useTranslation } from 'react-i18next';
 
 import { LanguageCode, UiStringsPackage } from '@votingworks/types';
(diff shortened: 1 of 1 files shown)


libs/ui/src/ui_strings/language_context.tsx
```diff
@@ -147,31 +154,28 @@ export function BackendLanguageContextProvider({
...
+      resources: mapObject(uiStringsPackage, (translations) => ({
+        [DEFAULT_I18NEXT_NAMESPACE]: translations,
+      })),
+    })
+    .use(initReactI18next);
+  // Since we have already loaded the translation resources and are passing
+  // them in explicitly above, we can set initImmediate to false to tell i18next to
+  // initialize synchronously. However, we still need to deal with the Promise
+  // returned by init(), but it will never reject so it's safe to ignore.
+  void i18nextInstance.init({ initImmediate: false });
```
adghayes, line 170: Nice option.

adghayes (approved):
LGTM, thank you for the improvement. Having global state is confusing enough in the frontend, glad you got rid of it on the backend!

PR #4916, opened 2024-06-04
clear controller in-buffer before echo command

kshen0:
## Overview

The controller daemon sends a health check command to the controller and expects a response back. If buttons on the controller have been pressed before the daemon has started up, the in-buffer will have data in it. That data is returned alongside the health check response. The concatenated response is unparsable by the daemon and crashes.

This PR clears the in-buffer before sending the health check command to avoid the problem.

## Demo Video or Screenshot

N/A

## Testing Plan

Manually tested both cases on hardware. No automated tests since this implementation is likely going away soon.

* starting daemon without any buttons pressed
* pressing buttons on controller and then starting daemon

### DIFF
--- a/apps/mark-scan/accessible-controller/src/port.rs
+++ b/apps/mark-scan/accessible-controller/src/port.rs
@@ -67,13 +67,40 @@ impl Port {
         let echo_command = EchoCommand::new(&[0x01, 0x02, 0x03, 0x04, 0x05]);
         let echo_command: Vec<u8> = echo_command.into();
         let port = &mut self.inner;
+        let mut serial_buf = [0; 1024];
+
+        // Clear the controller in-buffer. The in-buffer will contain data if buttons on the
+        // controller are pressed before the echo command is sent to the controller. That data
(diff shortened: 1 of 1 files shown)


apps/mark-scan/accessible-controller/src/port.rs
```diff
@@ -67,13 +67,40 @@ impl Port {
...
+        let mut serial_buf = [0; 1024];
+
+        // Clear the controller in-buffer. The in-buffer will contain data if buttons on the
+        // controller are pressed before the echo command is sent to the controller. That data
+        // will break parsing of the echo response below.
+        match port.bytes_to_read() {
+            Ok(0) => (),
+            Ok(num_bytes) => {
+                match port.clear(serialport::ClearBuffer::Input) {
+                    Ok(()) => log!( event_id: EventId::Info,
```
eventualbuddha, line 79:
```suggestion
                    Ok(()) => log!(
                        event_id: EventId::Info,
```

eventualbuddha (approved):

Issue #4510, opened 2024-01-12
vxdesign: ballot style readiness reports


mattroe:
1.1.2-J – Ballot style readiness reports
The voting system must provide the capability to produce pre-election reports that include:
1. the allowable number of votes in each contest;
2. the tabulation method for each contest;
3. the inclusion or exclusion of contests as the result of precinct splits; and
4. samples of all final ballot styles.

No VVSG testing assertions defined.

adghayes:
Adding some context from synchronous discussion with @mattroe:

We'd like to, and plan to, implement this as an export from VxDesign. That makes the most product sense and requires the least amount of work (exporting from VxAdmin would require the election package to contain ballot PDFs). It could simply be a report added within the Ballot PDF export that operates as a manifest and readiness report.

PR #4917, opened 2024-06-05
[VxDesign] Add a ballot style readiness report to the ballot PDF zip export

kofi-q:
## Overview

Resolves https://github.com/votingworks/vxsuite/issues/4510

_[VVSG 2.0 - 1.1.2-J – Ballot style readiness reports]_
Adding a ballot style readiness report in the "all ballots" export from VxDesign.
Lists out each ballot style along with associated metadata (id, language, party, precincts, districts) and contests.

Also adding an explicit `print` theme - hoping to make it easier to preview and iterate on prints in the browser using consistent sizing and layout dimensions. Will be following up later with a print preview helper component for use in the libs/ui storybook instance.

## Demo Video or Screenshot
See snapshots in commits cd3ccd1211d62e83e1bef3480071dc4b86cedfa6 and 87c7b368e8c2a54b4d15847381b357d7ae576080

ballot-style-reports-test-ts-pdf-layout-regression-test-1-snap

## Testing Plan
- Unit tests for report details, snapshot test for layout regression
- Manual testing via VxDesign export and with hardcopy prints

## Checklist


libs/printing/src/render.tsx
```diff
@@ -58,27 +58,39 @@ function getContentHeight(page: Page): Promise<number> {
   });
 }
 
+export type DocumentRenderer = () => JSX.Element | JSX.Element[];
+
 export interface RenderSpec {
-  document: JSX.Element | JSX.Element[];
+  document: JSX.Element | JSX.Element[] | DocumentRenderer;
```
adghayes, line 64: What is the rationale for adding deferred rendering rather than just passing an element from `design/backend`? I think it's fine but don't see why it's necessary.

adghayes (approved):
Looks good to me, some small code callouts in my comments. Thank you very much for setting up the print theming. Re: the output report, nits:

Screen Shot 2024-06-07 at 1 10 25 PM

1. I would prefer the title without "VxSuite." I don't think it's adding any additional information with the logo already there. And because it's used so sparsely throughout the system and in marketing materials, users might be unfamiliar / confused by it.
2. The way that the election header is formatted in the report "Mock Primary Election, 0ed1e43e4" looks to me like a subsection header that should be one among many subsections, but it indicates the election for the whole report. One option would be to add election information to the title, but that could get crammed. I'd suggest trying to add the election name and hash as metadata similar to **Date** rather than the current sub-header, and to see how that looks.
3. If the list of precincts or districts gets long, I'm not how it would wrap just from looking at the CSS. Would want to make sure it doesn't look wonky.

libs/ui/src/app_base.test.tsx
```diff
@@ -83,6 +83,34 @@ test('renders with selected themes', () => {
...
+      <AppBase defaultColorMode="desktop" defaultSizeMode="desktop">
+        <div>foo</div>
+      </AppBase>
+    )
+  );
+
+  expect(styleSheet.getStyleTags()).toMatch('font-size:16px !important');
+});
+
+test('skips legacy print media styles when using print theme', () => {
```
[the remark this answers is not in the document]
kofi-q: Oops, thanks - copy-paste error

libs/printing/src/render.tsx
kofi-q: Ah, leftover from a previous approach that I don't need anymore -- I tried removing the `<ThemeProvider>` wrapper from one of the existing reports earlier and tests were failing due to missing theme context and I mistakenly thought the components were getting mounted somewhere else before getting used here, but the actual fix was just to move [this line](https://github.com/votingworks/vxsuite/pull/4917/files#diff-14c71a93ae4f41c491a64dd5750495b30d6b47331a893fd4c2a2734c5588790cL117) up into the theme wrapper. Will revert this

kofi-q (commented):
> I would prefer the title without "VxSuite." I don't think it's adding any additional information with the logo already there. And because it's used so sparsely throughout the system and in marketing materials, users might be unfamiliar / confused by it.

SGTM!

> The way that the election header is formatted in the report "Mock Primary Election, 0ed1e43e4" looks to me like a subsection header that should be one among many subsections, but it indicates the election for the whole report. One option would be to add election information to the title, but that could get crammed. I'd suggest trying to add the election name and hash as metadata similar to Date rather than the current sub-header, and to see how that looks.

+1 - started going down the path of adding it to the metadata row earlier and ended up dropping. Will resurrect that code

> If the list of precincts or districts gets long, I'm not how it would wrap just from looking at the CSS. Would want to make sure it doesn't look wonky.

+1 - here's a screenshot from some earlier tests in storybook (we could probably fit in another 1 or 2 columns for the districts/precincts if we need to):

ballot-style-readiness-report 2

Issue #4622, opened 2024-02-26
vxscan: enable scanning official ballots in test mode


carolinemodic:
New Hampshire ballot testing indicates that we need a way to test with official ballots, i.e., scan official ballots while in test mode. This could be a [system setting](https://github.com/votingworks/vxsuite/blob/924f1722b273afb80f3de7f58ca6e30fb693e965/libs/types/src/system_settings.ts#L54-L65) that we turn on for NH elections.

Issue automatically created from slack feedback reported by ben 
https://votingworks.slack.com/archives/CJU9MSC6S/p1708985168783059?thread_ts=1708985168.783059&cid=CJU9MSC6S

PR #4940, opened 2024-06-06
[VxScan] Optionally allow scanning official ballots in test mode

kofi-q:
## Overview

Related to https://github.com/votingworks/vxsuite/issues/4622

Adding an optional `allowOfficialBallotsInTestMode` system setting -- when enabled, always interpret official ballots as test-mode ballots when scanning in test mode.

## Testing Plan
- New unit tests for the `normalizeBallotMode` validation helper and test cases to verify it's getting used in each code path (BMD ballots, NH HMPBs, Vx HMPBs).
- Existing tests as regression guard

## Checklist

### DIFF
--- a/libs/ballot-interpreter/src/interpret.ts
+++ b/libs/ballot-interpreter/src/interpret.ts
@@ -20,7 +20,6 @@ import {
   InvalidTestModePage,
   mapSheet,
   PageInterpretation,
-  PrecinctSelection,
   SheetOf,
   AdjudicationInfo,
   AnyContest,
@@ -39,7 +38,6 @@ import {
   Rect,
   WriteInAreaStatus,
   WriteInId,
-  MarkThresholds,
   AdjudicationReason,
   PageInterpretationWithFiles,
   Id,
@@ -67,19 +65,10 @@ import {
 import { interpret as interpretVxBmdBallotSheet } from './bmd';
 import { getAllPossibleAdjudicationReasons } from './adjudication_reasons';
 import { saveSheetImage } from './save_images';
+import { InterpreterOptions } from './types';
+import { normalizeBallotMode } from './validation';
 
 const debug = makeDebug('ballot-interpreter:scan:interpreter');
-
-/**
- * Options for interpreting a sheet of ballot images.
- */
-export interface InterpreterOptions {
-  electionDefinition: ElectionDefinition;
-  precinctSelection: PrecinctSelection;
-  testMode: boolean;
-  markThresholds: MarkThresholds;
-  adjudicationReasons: readonly AdjudicationReason[];
-}
 
 /**
  * Result of interpreting a sheet of ballot images.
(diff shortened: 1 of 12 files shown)


libs/ballot-interpreter/src/interpret_vx_hmpb.test.ts
```diff
@@ -183,6 +193,39 @@ describe('HMPB - Famous Names', () => {
     expect(frontResult.interpretation.type).toEqual('InvalidTestModePage');
     expect(backResult.interpretation.type).toEqual('InvalidTestModePage');
   });
+
+  test('normalizes ballot mode', async () => {
+    const ballotImagePaths = await ballotPdfToPageImages(blankBallotPath);
+    expect(ballotImagePaths.length).toEqual(2);
```
eventualbuddha, line 199: I think this would provide a nicer error message:
```suggestion
    expect(ballotImagePaths).toHaveLength(2);
```

libs/ballot-interpreter/src/interpret.ts
```diff
@@ -503,17 +492,18 @@ export async function convertRustInterpretResult(
...
-    precinctSelection: PrecinctSelection;
-    testMode: boolean;
-  }
-): SheetOf<InterpretFileResult> {
-  return mapSheet(results, ({ interpretation, normalizedImage }) => {
+  } = options;
+
+  return mapSheet(results, (result) => {
+    const { normalizedImage } = result;
+    const interpretation = normalizeBallotMode(result.interpretation, options);
```
eventualbuddha, line 505: Why normalize the test mode rather than just allowing both when the flag is set?

eventualbuddha (commented):
Code looks fine, but I wonder about the approach. Are we not preserving the original test mode value because it'd have larger downstream consequences?

libs/ballot-interpreter/src/interpret.ts
kofi-q: Yeah, good question - started out with just skipping the test mode validation, but wasn't sure if the `isTestMode` metadata on the eventual CVR would be used anywhere else downstream. Noticed the NH HMPB code path sets the ballot's `isTestMode` based on the interpreter's test mode and went that route to be safe, but maybe it's not an issue and we can use the simpler approach here. Do you have any context here?

cc @arsalansufi - wondering if we're there are any downstream side-effects to having non-test-mode CVRs in test-mode CVR exports (guessing we only ever reference the test mode on the batch export metadata in the new continuous export world?)

kofi-q: +1 -- also leaning this way after thinking about it some more. Don't foresee a need to know the original physical ballot mode later down the line and if we do end up wanting that, we could add more specific metadata

libs/ballot-interpreter/src/interpret_vx_hmpb.test.ts
kofi-q: 💯 - will update all the instances in this file

kofi-q:
@arsalansufi - mind doing a quick approval pass while Brian's out?

arsalansufi (approved):
LGTM!

Issue #4163, opened 2023-11-03
Upgrade to Node 20 (or 18)


carolinemodic:
Node 16 is now EOL'd

amcmanus:
This is no longer part of the nh-march-freeze. We will revisit afterward.

eventualbuddha:
Node v22 is now current.

eventualbuddha:
Slack thread from the last look at this: https://votingworks.slack.com/archives/CEL6D3GAD/p1700198455681619

kofi-q:
All seems to be good in dev envs, CI, and on prod images (although prod's only tested on VxMark so far, but these should be fairly good signal). Will keep an eye out for anything that pops up that might be related and if I'm OOO while that goes down, the last two PRs linked above are the ones to roll back.

PR #4942, opened 2024-06-06
chore(deps): update `canvas` to v2.11.2

eventualbuddha:
## Overview

Refs https://github.com/votingworks/vxsuite/issues/4163

v2.9.1 is incompatible with NodeJS v20. Also, v2.11.2 is incompatible with NodeJS v22, but that's a problem for future us.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests, plus I tested manually building an image using `createCanvas` and it worked:

test

```js
const canvas = require("canvas").createCanvas(100, 100);
const ctx = canvas.getContext("2d");
ctx.fillStyle = "red";
ctx.fillRect(20, 20, 40, 40);
const png = canvas.toBuffer("image/png");
fs.writeFileSync("test.png", png);
```

### DIFF
--- a/libs/ballot-interpreter/src/hmpb-ts/interpret.test.ts
+++ b/libs/ballot-interpreter/src/hmpb-ts/interpret.test.ts
@@ -1,4 +1,5 @@
 import { assertDefined, iter, ok, unique } from '@votingworks/basics';
+import { ImageData } from 'canvas';
 import { electionGridLayoutNewHampshireTestBallotFixtures } from '@votingworks/fixtures';
 import { Election, ElectionDefinition, SheetOf } from '@votingworks/types';
 import { interpret } from './interpret';
--- a/apps/admin/frontend/package.json
+++ b/apps/admin/frontend/package.json
@@ -66,7 +66,6 @@
     "base64-js": "1.5.1",
     "browserify-zlib": "^0.2.0",
     "buffer": "^6.0.3",
-    "canvas": "2.11.2",
     "debug": "4.3.4",
     "dompurify": "^2.0.12",
     "dotenv": "16.3.1",
--- a/apps/admin/frontend/package.json
+++ b/apps/admin/frontend/package.json
@@ -66,7 +66,7 @@
     "base64-js": "1.5.1",
     "browserify-zlib": "^0.2.0",
     "buffer": "^6.0.3",
-    "canvas": "2.9.1",
+    "canvas": "2.11.2",
     "debug": "4.3.4",
     "dompurify": "^2.0.12",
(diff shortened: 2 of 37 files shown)


libs/ballot-interpreter/src/hmpb-ts/interpret.test.ts
```diff
@@ -1,4 +1,5 @@
 import { assertDefined, iter, ok, unique } from '@votingworks/basics';
+import { ImageData } from 'canvas';
```
adghayes, line 2: Why did you need to add in all these type imports?

adghayes (approved):

libs/ballot-interpreter/src/hmpb-ts/interpret.test.ts
eventualbuddha: The types of `canvas.ImageData` and `globalThis.ImageData` used to be compatible but no longer are:

```ts
// canvas@2.11.2
export class ImageData {
	constructor(sw: number, sh: number);
	constructor(data: Uint8ClampedArray, sw: number, sh?: number);
	readonly data: Uint8ClampedArray;
	readonly height: number;
	readonly width: number;
}

// lib.dom.d.ts
/**
 * The underlying pixel data of an area of a <canvas> element. It is created using the ImageData() constructor or creator methods on the CanvasRenderingContext2D object associated with a canvas: createImageData() and getImageData(). It can also be used to set a part of the canvas by using putImageData().
 *
 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ImageData)
 */
interface ImageData {
    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ImageData/colorSpace) */
    readonly colorSpace: PredefinedColorSpace;
    /**
     * Returns the one-dimensional array containing the data in RGBA order, as integers in the range 0 to 255.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ImageData/data)
     */
    readonly data: Uint8ClampedArray;
    /**
     * Returns the actual dimensions of the data in the ImageData object, in pixels.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ImageData/height)
     */
    readonly height: number;
    /**
     * Returns the actual dimensions of the data in the ImageData object, in pixels.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ImageData/width)
     */
    readonly width: number;
}
```

Notably, `colorSpace` is present in the DOM definition but not in the `canvas` one. So to avoid type errors we always use the one from `canvas`.

PR #4951, opened 2024-06-12
Increase HMPB margins to match printer

jonahkagan:
## Overview

To make it easier to print ballots without having to adjust the scale, I matched the paper margins to the minimum supported margins of the HP LaserJet m404 (our standard printer). Of course, ballots will often be printed by a print shop, but this at least improves our lives when testing on our printers.

## Demo Video or Screenshot
The change is almost too small to tell the difference, though maybe you can see in the snapshot diffs in this PR
## Testing Plan
- Updated automated tests
- Manually tested that the print dialog defaults to 100% scale and that the ballot prints fully
## Checklist


libs/fixtures/data/electionGeneral/ui-strings.json
```diff
@@ -1748,7 +1748,7 @@
     "hmpbOfficialProvisionalBallot": "Boleta provisional oficial",
     "hmpbPage": "Página",
     "hmpbPageIntentionallyBlank": "Esta página se dejó en blanco intencionalmente.",
-    "hmpbSampleAbsenteeBallot": "Ejemplo de boleta de voto en ausencia",
+    "hmpbSampleAbsenteeBallot": "Modelo de boleta de voto ausente",
```
jonahkagan, line 1751: Making it consistent with the other sample ballot translations (and a bit shorter to avoid line wrapping)

kofi-q (approved):

Issue #4720, opened 2024-03-22
vxmarkscan: diagnostic for headphone input


adghayes:
Create a simple diagnostic flow for confirming the headphone jack is properly functioning. The flow could be:

1. Press button to start flow
2. Prompt to plug in headphones (if not attached)
3. Start playing sound on repeat
4. Give user opportunity to confirm success or failure
5. If failed, suggest testing again with another pair of headphones

Will have to add headphone detection functionality (which we may need for other reasons).

The section on the diagnostics screen should show a header, a connected status with icon, a last tested string with icon, and a button to do the diagnostic. It can be modeled off of the other diagnostic sections.

PR #4955, opened 2024-06-12
headphone diagnostic

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/4720

Adds a diagnostic for headphones. Though the copy says it's for the front headphone input, nothing about the code enforces that.

We previously discussed leaving out connection status for the headphone diagnostic because implementation might change later, but seeing as headphone connection status is already exported nicely for a previous use case, I could be convinced to add it here. On the other hand, it's not super clear what "Connected" means in the diagnostic use case for easily-removed peripherals eg. when exporting the readiness report, does the user want to know or care if headphones or a PAT device are connected at that moment?

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/0857cff5-c51c-4daa-b0d8-3bcb44a050b0

## Testing Plan

- added tests
- manually tested on VM

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced. (should be free with API logging)


kofi-q (approved):
✅

PR #4957, opened 2024-06-12
libs/hmpb: Wait for HTML to load during rendering

jonahkagan:
## Overview

When inserting HTML content into the DOM, we want to wait for it to finish rendering before doing anything with it (e.g. taking measurements).

I found the `MutationObserver` API, which registers an observer to call a callback when various types of changes happen to a DOM node. Specifically, the one that is useful to us is a change in the node's children.
## Demo Video or Screenshot
N/A

## Testing Plan
Manually tested that the layout bugs that were previously happening (due to measuring before content was finished rendering) are solved by this.

## Checklist

### DIFF
--- a/libs/hmpb/src/renderer.tsx
+++ b/libs/hmpb/src/renderer.tsx
@@ -3,7 +3,7 @@ import { Buffer } from 'buffer';
 import type { Page as PlaywrightPage } from 'playwright';
 import ReactDomServer from 'react-dom/server';
 import { ServerStyleSheet } from 'styled-components';
-import { assert, sleep } from '@votingworks/basics';
+import { assert } from '@votingworks/basics';
 import { PixelMeasurements } from './types';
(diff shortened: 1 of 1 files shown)


kofi-q (approved):

jonahkagan:
@adghayes this might be useful for libs/printing as well

adghayes:
@jonahkagan in `libs/printing` we're just waiting for the content to load with a built-in Playwright argument: https://github.com/votingworks/vxsuite/blob/ea38f4e613def04fdc81cffbf1f9d69b0a70c552/libs/printing/src/render.tsx#L164

That seems to work and be simpler, but from a quick glance at the code here it looks like you may have more requirements, such as inserting content at a particular node (rather than just replacing all the content).

jonahkagan:
@adghayes Ah ok yeah seems different. Glad libs/printing already has a solution to this

Issue #4551, opened 2024-01-31
vxscan: make resync cvr modal require pollworker or higher auth


carolinemodic:
Make resync cvr modal require pollworker or higher auth

Issue automatically created from slack feedback reported by caroline 
https://votingworks.slack.com/archives/CJU9MSC6S/p1706661864331329

carolinemodic:
Context: 
https://votingworks.slack.com/archives/CE7LM9LQ2/p1706731059831519?thread_ts=1706656747.812099&cid=CE7LM9LQ2

Initial plan: Change resync required modal to just be a blocking modal or screen telling you to insert pollworker card, make the pollworker screen have you resync as a first action if resync is required

PR #4968, opened 2024-06-12
Move Resync CVR Flow Behind PollWorker Auth

carolinemodic:
## Overview
Task: #4551 
Move resyncing of CVRs on USB behind pollworker auth based on SLI feedback. 

Previously the CVR sync required screen would only come up when polls were open, the pollworker screen would supersede this screen so other pollworker actions were not blocked on this sync occuring. However closing of polls specifically had a blocking modal added to prevent that occuring before sync. 

Now this is updated so there is a voter-facing screen that appears when the old screen did: only when polls are open and the usb is out of sync. We could display it when polls are paused/closed as well but I figured that then the next action would be a pollworker inserting their card anyway. 

When a pollworker authenticates and the USB is out of sync the old screen will come up and block all other pollworker actions, as discussed with @mattroe. This means the previous closing polls blocking modal is no longer needed because you can't get to that point without syncing CVRs anyway. 

I am open to making the copy in the potentially voter facing first screen more generic, I primarily copied how the ballot bag full screen works which does give that context. 

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/14897017/fc1076e4-431e-4a4a-8fd0-108ce7a62c10

## Testing Plan
Tested manually as shown in video above, ran updated unit tests. 

## Checklist

- [ N/A ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.


mattroe (commented):
LGTM. The jump to the polls closed screen after completing the sync is a bit awkward, but I think it's fine given the low-likelihood of this issue occurring in the field.

Super-nit: let's not capitalize the `Cast` of cast vote record on the first screen

libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -1363,6 +1369,12 @@ export const appStrings = {
     </UiString>
   ),
 
+  warningCvrsNeedSync: () => (
+    <UiString uiStringKey="warningCvrsNeedSync">
+      A poll worker must sync Cast vote records (CVRs) to the USB drive.
+    </UiString>
+  ),
```
arsalansufi, line 1376: > I am open to making the copy in the potentially voter facing first screen more generic, I primarily copied how the ballot bag full screen works which does give that context.

Yeah I'm a bit torn here since "CVRs need to be synced" is more in the weeds than "the ballot bag is full". But ultimately, I like the consistency with the ballot bag screen. And I think the immediate context for poll workers, before card insertion, is nice. (I know poll workers won't be able to read the initial message if a voter has changed the language, and the poll worker doesn't speak that language, but we don't expect voters to change the language on this screen often.)

+1 to Roe's suggestion to lower-case the C in cast vote records

arsalansufi (approved):
LGTM!

Issue #4659, opened 2024-03-07
Ensure that VxScan can scan both HMPBs and BMD ballots


kshen0:
See: https://votingworks.slack.com/archives/CEL6D3GAD/p1709841884384079

arsalansufi:
Noting that, until we address this, VxScan will only be able to scan HMPBs and not BMD ballots. We need it to be able to scan both, so this is actually high pri.

PR #4978, opened 2024-06-18
fix(ballot-interpreter): allow interpreting both BMD and HMPB ballots

eventualbuddha:
## Overview

Refs #4659

This will increase the time taken to scan BMD ballots for elections that have a grid layout defined. We can do better by unifying the BMD and HMP interpretation code paths, which I will do later. This is a smaller change that removes the current limitation while I work on the bigger unification.
## Testing Plan
- [x] Add automated tests using automatically-generated BMD and HMP ballots.
- [ ] Manually test VxScan election with both BMD and HMP ballots.
- [ ] Manually test VxCentralScan election with both BMD and HMP ballots.


libs/ballot-interpreter/src/bmd/interpret.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    electionFamousNames2021Fixtures.electionDefinition,
+    card
+  );
+  const { ballot, summaryBallotImage, blankPageImage } = result.unsafeUnwrap();
+  expect(ballot).toMatchSnapshot();
+
+  // don't use Jest `toEqual` matcher because it tries to pretty print the
+  // ImageData objects, which is slow and causes the test to time out
+  assert(summaryBallotImage === card[0]);
+  assert(blankPageImage === card[1]);
```
jonahkagan, line 47: I've used `jest-image-matcher` for image comparison, though if this is just for object identity then maybe you don't need it

libs/ballot-interpreter/src/bmd/interpret.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    await sampleBallotImages.blankPage.asImageData(),
+    await sampleBallotImages.blankPage.asImageData(),
+  ];
+  const result = await interpret(
+    electionFamousNames2021Fixtures.electionDefinition,
+    card
+  );
+  expect(result).toEqual(
+    err(
+      typedAs<InterpretError>({
```
jonahkagan: Could you use the `toEqual` type parameter here like your recent PR?

libs/ballot-interpreter/src/interpret_bmd_ballots.test.ts
```diff
@@ -1,175 +1,75 @@
...
-
 beforeEach(() => {
   mockOf(normalizeBallotMode).mockImplementation((input) => input);
 });
 
 describe('VX BMD interpretation', () => {
   const fixtures = electionFamousNames2021Fixtures;
   const { electionDefinition } = fixtures;
-  const bmdSummaryBallotPage = fixtures.machineMarkedBallotPage1.asFilePath();
-  const bmdBlankPage = fixtures.machineMarkedBallotPage2.asFilePath();
```
jonahkagan, line 171: 🎉 amazing

jonahkagan (approved):
Thanks for making all these test improvements!

libs/ballot-interpreter/src/bmd/interpret.test.tsx
eventualbuddha: As there doesn't appear to be anything set up to compare two images instead of one image vs a snapshot, I'm going to leave this as is for now.

jonahkagan: Here is an example: https://github.com/votingworks/vxsuite/blob/main/libs/hmpb/src/generate_fixtures.test.ts#L30-L33

jonahkagan: But I agree that it's fine to leave as is

Issue #4812, opened 2024-05-01
VxMark: Casting of previously marked/removed ballot


mattroe:
Multiple users in all of our user testing to-date have removed their ballot after printing, but before validating their ballot and casting.

Although voters will have the option to scan their ballot on VxScan at a polling place, this is a common enough occurence that we want to address the problem on VxMark/VSAP itself.

We should address this problem in two ways:
1. explicitly guide voters to NOT remove their ballot
2. enable inserting and casting a previously printed valid ballot

On voter guidance:
- when initially confirming their selections and waiting for the ballot to print, the BMD should communicate to voters that they will need to verify their ballot and should not remove their ballot from the machine

On inserting a valid printed ballot:
- Valid ballots that are inserted should be assumed to have been previously validated and selections do not need to be reconfirmed before casting.
- Invalid ballot sheets should be rejected as they are rejected on VxScan.
- A voter/poll worker should be able to insert a valid previously printed ballot at the start of a voter session instead of a blank ballot.
- If a voter removes their ballot after printing, the user should have a period of time where the scanner accepts sheets (30s?) before the voter session times out and a poll worker needs to intervene to assist. They should be prompted that they can reinsert the sheet into the BMD for casting OR scanning at the precinct scanner (as instructions call out today).

PR #4979, opened 2024-06-18
[MarkScan] Expand mock paper handler functionality

kofi-q:
## Overview

Motivated by https://github.com/votingworks/vxsuite/issues/4812

Adding a bit more functionality to the `MockPaperHandlerDriver` to enable usage in dev with minimal logical branching and to help speed up VM dev cycles.
Have been maintaining a hacky version of this locally, so cleaning up and merging to avoid that overhead while preparing to add a new flow re-inserting removed ballots.

Will follow up with a separate PR to add the mock management UI to the dev dock later.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/264902/f642a711-8b3d-44ac-9b67-16d3aedee80b

## Testing Plan
- New/updated unit tests for the mock and consumers
- Manual verification locally with a 

## Checklist


apps/mark-scan/backend/src/app.ts
```diff
@@ -195,6 +200,12 @@ export function buildApi(
     setAcceptingPaperState(): void {
       assert(stateMachine);
       stateMachine.setAcceptingPaper();
+
+      // TODO(kofi): Remove once we've added mock paper handler functionality to the
+      // dev dock:
+      if (isMockPaperHandler(paperHandler)) {
```
kofi-q: This gets moved client-side in commit https://github.com/votingworks/vxsuite/pull/4979/commits/5de0640c2ddb661b4ab753ee4fe515b353bb7174 - sorry meant to rebase and clean up this commit

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -579,17 +562,6 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
         (luminance << 24) | (luminance << 16) | (luminance << 8) | 255;
     }
     await writeImageData(pathOut, colorResult);
-
-    const imageMetadata: ImageFromScanner = {
```
kofi-q, line 583: Looked like this return isn't currently getting used anywhere, so removed it to simplify both here and in the mock - LMK if it's something we actually need to keep around.

kofi-q (commented):
Also, apologies for the PR size - the tests and cascading changes ballooned more than I expected. No rush on this -- won't be actively developing on top of this until end-of-week

apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
```diff
@@ -86,28 +80,19 @@ const TEST_POLL_INTERVAL_MS = 50;
 // detect `ballot_accepted` state
 const TEST_NOTIFICATION_DURATION_MS = 150;
 
-jest.mock('@votingworks/custom-paper-handler');
+// jest.mock('@votingworks/custom-paper-handler');
```
kshen0, line 83: delete?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1286,29 +1272,11 @@ export async function getPaperHandlerStateMachine({
...
-      ) {
-        const pdfPages = iter(pdfToImages(pdfData));
-        const pdfPage = await pdfPages.first();
-        assert(pdfPage);
-
-        const jpgPath = tmpNameSync({ postfix: '.jpg' });
-        await writeImageData(jpgPath, pdfPage.page);
-
-        machineService.send({
-          type: 'SET_INTERPRETATION_FIXTURE',
```
kshen0, line 1303: noting that this state can probably be replaced with usage of [`setMockPaperContents`](https://github.com/votingworks/vxsuite/pull/4979/files#diff-a61779136fdb9187d8346134eb8b69030bb10931fc271581e01e8a03afe47c2cR318) in the future (doesn't need to be in this PR or by you in particular)

libs/custom-paper-handler/src/driver/mock_driver.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    ...jest.requireActual('@votingworks/image-utils'),
+    writeImageData: jest.fn(),
+  })
+);
+
+function newMockDriver() {
+  const mockDriver = new MockPaperHandlerDriver();
+
+  const operationDelayMs = 1000;
+  mockDriver.setMockOperationDelayMs(operationDelayMs);
```
kshen0, line 26: I'm trying to understand the intended usage of this delay. Is the idea that we just need some known delay so we know when to expect states when manipulating Jest fake timers?

apps/mark-scan/backend/src/custom-paper-handler/state_machine.test.ts
kofi-q: thanks!

libs/custom-paper-handler/src/driver/mock_driver.ts
```diff
@@ -39,302 +25,310 @@ function makeUsbOutTransferResult(
...
-  readonly webDevice: MinimalWebUsbDevice = mockMinimalWebUsbDevice();
+  private statusRef: PaperHandlerStatus = defaultPaperHandlerStatus();
+  private mockStatus: MockPaperHandlerStatus = 'noPaper';
+  private operationDelayMs: number = 0;
+  private mockPaperContents?: ImageData;
+  private pendingStatusChangeTimerId?: NodeJS.Timeout;
 
-  statusRef: PaperHandlerStatus = defaultPaperHandlerStatus();
+  constructor() {
+    void this.setMockStatus('noPaper');
```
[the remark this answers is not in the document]
kofi-q: oops, had `setMockStatus` as an async function previously - will update

libs/custom-paper-handler/src/driver/mock_driver.test.ts
kofi-q: Starting to rethink whether this is necessary - this was more intended as a way to simulate the real-world delay between sending a command and the paper handler actually physically completing the task. In some cases, the mock was transitioning through states too quickly to be able to see in-between states in the UI, and in particular, I was testing forcing a jam state while a print/scan was in progress. But these feel a little too niche to work into the mock/dev dock, so might leave out for now

kshen0: Yeah, I haven't been very happy with our ability to reliably and exhaustively test states. Our current state machine tests use `xstate`'s `onTransition`, which behaves unreliably in that not all states trigger the `onTransition` handler. It looks like others have had this issue and `onTransition` was removed in v5 in favor of `subscribe`: https://stately.ai/docs/migration#use-subscribe-for-changes-not-ontransition.

That said, since it's way out of scope in the foreseeable future to validate and upgrade to xstate v5, the mock delay approach seemed fine with me. But I'll defer to you on whether that approach would be more trouble to maintain than it's worth.

kshen0 (approved):

PR #4982, opened 2024-06-20
libs/hmpb: Rework fix to wait for HTML to load during rendering

jonahkagan:
## Overview

I had tried out using a MutationObserver to wait for DOM content to load after being inserted (https://github.com/votingworks/vxsuite/pull/4957). However, the same issue occurred again, where some content was still changing after the observer fired. I think it only appeared to work by delaying things slightly without actually solving the problem.

Here is a new approach where we use requestAnimationFrame to wait for the next browser repaint. This seems to work, though similarly to last time I can't be 100% confident that it's solving the problem, so we'll need to wait and see.
## Demo Video or Screenshot
N/A

## Testing Plan
Manual testing with an example that was buggy before, and now works.
## Checklist

### DIFF
--- a/libs/hmpb/src/renderer.tsx
+++ b/libs/hmpb/src/renderer.tsx
@@ -43,13 +43,17 @@ export function createDocument(page: Page) {
             throw new Error(`No element found with selector: ${selector}`);
           }
 
+          node.innerHTML = content;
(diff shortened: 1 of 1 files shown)


kofi-q (approved):
🤞

Issue #4939, opened 2024-06-06
VxCentralScan imprinting


arsalansufi:
See https://github.com/votingworks/roadmap/issues/31

mattroe:
https://docs.google.com/document/d/1meTdI7JqkvVZS8JzW7Z8Q6lL7nZKMpxYHDrwv6w8QcE/edit?usp=sharing

PR #4983, opened 2024-06-20
Increase default margins from VxDesign to accomodate imprinting 

carolinemodic:
## Overview
#4939 
When imprinting in the central scanner we want to imprint in the margins of an HMPB. In order to, by default, have enough margin to accomodate this I slightly bumped the margins to 5mm on the left and right sides. 

There are two tests in apps-design-backend that got broken by this change with translated ballots for the famous names election now not fitting on a letter sized page. I tried changing the election to electionGeneral which has translation strings defined but was still hitting the issue (although at least there the error is thrown when using that election locally manually as well), so I settled for just changing the paper size to Legal in the tests to work around the issue. The need for legal size only exists when there aren't translated strings for some of the content defined which is true in the tests but not when using the election manuallly. 

## Demo Video or Screenshot
N/A

## Testing Plan
Ran tests, manually printed ballots from webserver and imprinted on them in all orientations. 

## Checklist


apps/central-scan/backend/src/__image_snapshots__/__diff_output__/readiness-report-1-diff.png
jonahkagan, line 1: I wouldn't expect this diff file to be committed. Is this test passing?

apps/design/backend/src/app.test.ts
```diff
@@ -290,6 +291,18 @@ test('Election package management', async () => {
       electionData: baseElectionDefinition.electionData,
     })
   ).unsafeUnwrap();
+  const election = await apiClient.getElection({ electionId });
+
+  await apiClient.updateElection({
```
jonahkagan, line 296: Maybe add the same explanatory comment here

jonahkagan (approved):

apps/central-scan/backend/src/__image_snapshots__/__diff_output__/readiness-report-1-diff.png
carolinemodic: must be a relick of a past issue, test passes when remove so will do that

PR #4986, opened 2024-06-22
Support controller and sip & puff on BMD 150 hardware

kshen0:
## Overview
Solves https://github.com/votingworks/vxsuite/issues/4960 and https://github.com/votingworks/vxsuite/issues/4961

Adds support for the FAI-100 USB device which encompasses both ATI controller and sip & puff input on the BMD 150. The implementation is forked from the BMD 155 controller daemon.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/6c7447cb-fd44-487b-8f04-4afdb8d663f8

https://github.com/user-attachments/assets/24df0b63-03bf-4256-bd06-59fc4c8bc5fe

## Testing Plan
- rewrote existing tests for BMD 155 controller
- added sip and puff tests

## Checklist

### DIFF
--- a/apps/mark-scan/fai-100-controller/src/commands.rs
+++ b/apps/mark-scan/fai-100-controller/src/commands.rs
@@ -11,10 +11,8 @@ pub enum CommandError {
     FailedSliceConversion,
     #[error("Unexpected data length: {0}")]
     UnexpectedDataLength(usize),
-    #[error("Unexpected command ID received: {0}")]
+    #[error("Unexpected command ID received: {0:x}")]
     UnexpectedCommandId(u8),
-    #[error("Unknown command received: {0}")]
-    UnknownCommand(u8),
     #[error("Error occurred when sending keypress event")]
     KeypressError(#[from] uinput::Error),
     #[error("Error occurred when converting button value from raw response")]
@@ -222,7 +220,7 @@ impl TryFrom<&[u8]> for VersionResponse {
 
         // Validate correct command ID was returned
         if response[0] != CommandId::GetFirmwareVersion as u8 {
-            return Err(CommandError::UnknownCommand(bytes[25]));
+            return Err(CommandError::UnexpectedCommandId(response[0]));
         }
 
         // Version is the only chunk of data that is big endian according to docs.
--- a/apps/mark-scan/fai-100-controller/src/commands.rs
+++ b/apps/mark-scan/fai-100-controller/src/commands.rs
@@ -1,0 +1,346 @@
+use num_enum::TryFromPrimitiveError;
+use std::io;
+
+#[derive(Debug, thiserror::Error)]
+pub enum CommandError {
+    #[error("Unexpected command data length: {0}")]
+    UnexpectedPacketSize(usize),
+    #[error("Unknown sender ID: {0}")]
+    UnknownSenderId(u32),
+    #[error("Failed to convert slice to unsigned int type")]
+    FailedSliceConversion(),
+    #[error("Unexpected data length: {0}")]
+    UnexpectedDataLength(usize),
+    #[error("Unexpected command ID received: {0}")]
+    UnexpectedCommandId(u8),
+    #[error("Unknown command received: {0}")]
+    UnknownCommand(u8),
+    #[error("Error occurred when sending keypress event")]
+    KeypressError(#[from] uinput::Error),
+    #[error("Error occurred when converting button value from raw response")]
+    InvalidButton(TryFromPrimitiveError<ButtonSignal>),
+    #[error("Error occurred when converting sip/puff signal status value from raw response")]
+    InvalidSipPuffSignalStatus(TryFromPrimitiveError<SipAndPuffSignalStatus>),
+    #[error("Error occurred when converting sip/puff connection status value from raw response")]
+    InvalidSipPuffDeviceStatus(TryFromPrimitiveError<SipAndPuffDeviceStatus>),
+    #[error("I/O error when parsing command response")]
+    IOError(#[from] io::Error),
+}
+
+impl From<TryFromPrimitiveError<ButtonSignal>> for CommandError {
+    fn from(err: TryFromPrimitiveError<ButtonSignal>) -> Self {
+        Self::InvalidButton(err)
+    }
+}
+
+impl From<TryFromPrimitiveError<SipAndPuffSignalStatus>> for CommandError {
+    fn from(err: TryFromPrimitiveError<SipAndPuffSignalStatus>) -> Self {
+        Self::InvalidSipPuffSignalStatus(err)
+    }
+}
+
+impl From<TryFromPrimitiveError<SipAndPuffDeviceStatus>> for CommandError {
+    fn from(err: TryFromPrimitiveError<SipAndPuffDeviceStatus>) -> Self {
+        Self::InvalidSipPuffDeviceStatus(err)
+    }
+}
+
+const SENDER_ID: u32 = 0xD0BB_55AA;
+
+#[derive(Debug, num_enum::TryFromPrimitive, Clone, Copy)]
+#[repr(u8)]
+pub enum CommandId {
+    GetFirmwareVersion = 0x50,
+    GetNotificationValues = 0xc1,
+}
+
+/// An outgoing command sent from host to device.
+#[derive(Debug)]
+pub struct Command {
+    pub command_id: CommandId,
+}
+
+#[macro_export]
+macro_rules! create_command {
+    ($id:ident) => {
+        $crate::commands::Command {
+            command_id: $crate::commands::CommandId::$id,
+        }
+    };
+}
+
+impl From<Command> for Vec<u8> {
+    fn from(command: Command) -> Self {
+        const EXPECTED_LENGTH: usize = 27;
+        let mut bytes = Self::with_capacity(EXPECTED_LENGTH);
+        // All commands start with device sender ID
+        bytes.extend_from_slice(&u32::to_le_bytes(SENDER_ID));
+        // Unused padding
+        bytes.append(&mut vec![0; 16]);
+        // LE u32 size of upcoming noise + data = 3 bytes
+        bytes.push(0x03);
+        bytes.append(&mut vec![0; 3]);
+        // 2 bytes of noise
+        bytes.append(&mut vec![0xff; 2]);
+        // 1 byte of data (just the command ID)
+        bytes.push(command.command_id as u8);
+
+        bytes
+    }
+}
+
+#[derive(Debug, num_enum::TryFromPrimitive, PartialEq, Eq, Hash)]
+#[repr(u8)]
+pub enum ButtonSignal {
+    Help = 0x30,
+    VolumeDown = 0x31,
+    Down = 0x32,
+    RateDown = 0x33,
+    Left = 0x34,
+    Select = 0x35,
+    Right = 0x36,
+    VolumeUp = 0x37,
+    Up = 0x38,
+    RateUp = 0x39,
+    Play = 0x3a,
+    Pause = 0x3b,
+    NoButton = 0xff,
+}
+
+#[derive(Debug, num_enum::TryFromPrimitive, PartialEq, Eq)]
+#[repr(u8)]
+pub enum SipAndPuffSignalStatus {
+    // Documented as Active = 0x01, Idle = 0x00, but based on experimentation the signals are flipped
+    // when a PAT device is plugged in. TODO handle this in event loop
+    Active = 0x00,
+    Idle = 0x01,
+}
+
+#[derive(Debug, num_enum::TryFromPrimitive, PartialEq, Eq)]
+#[repr(u8)]
+pub enum SipAndPuffDeviceStatus {
+    Connected = 0x01,
+    Disconnected = 0x00,
+}
+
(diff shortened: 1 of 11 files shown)


apps/mark-scan/fai-100-controller/src/usb_device.rs
```diff
@@ -0,0 +1,114 @@
+use std::{io, time::Duration};
+
+use color_eyre::eyre::Result;
+use rusb::{Context, DeviceHandle, UsbContext};
+
+const INTERFACE_NUMBER: u8 = 0x00;
+const CONFIG_NUMBER: u8 = 0x01;
+const SETTING_NUMBER: u8 = 0x00;
+const ENDPOINT_ADDRESS_IN: u8 = 0x81;
+const ENDPOINT_ADDRESS_OUT: u8 = 0x01;
```
kshen0, line 10: Per conversations with the manufacturer `0x81` and `0x01` are fixed and can be hard coded.

apps/mark-scan/fai-100-controller/src/virtual_keyboard.rs
```diff
@@ -0,0 +1,44 @@
+use std::{thread, time::Duration};
```
kshen0, line 1: This implementation is identical to https://github.com/votingworks/vxsuite/blob/main/apps/mark-scan/accessible-controller/src/device.rs#L33-L67

apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
```diff
@@ -0,0 +1,555 @@
...
+        },
+    ) {
+        log!(EventId::UnknownError, "Error sending key: {err}");
+    }
+}
+
+/// Checks new status for changed statuses in button press, sip, puff, and sip & puff device connection.
+/// If changing from inactive to active, sends the appropriate event (keypress or system file update).
+/// If changing from active to inactive or no status change, does nothing.
+fn handle_status_response(
```
kshen0, line 317: Similar to `controllerd/handle_command` (https://github.com/votingworks/vxsuite/blob/main/apps/mark-scan/accessible-controller/src/commands.rs#L166), but moved up to the top level because it seemed odd that the command parsing layer also was aware of the `uinput` device and business logic of which controller keys map to which keyboard keys. It also makes boundaries in testing a little more clean. I'm open to considering other options though.

apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
```diff
@@ -0,0 +1,555 @@
...
+        }
+    }
+    current_status.sip_puff_device_connected = new_connection_status;
+
+    // Only check for sip & puff actions when the device is connected. When the device is not connected
+    // the values switch meaning.
+    // ie. When a sip & puff is connected, 0x01 is sent to host when a sip
+    // is happening and 0x00 is sent when no sip is happening.
+    // When no sip & puff is connected, 0x00 is sent when no sip is happening.
+    // Therefore we must ignore all sip & puff signals unless we know the sip & puff is connected.
```
kshen0, line 359: I tested on hardware and it seems like we still catch stray sip/puff signals in the moment the device is plugged or unplugged. Will need to figure out another way to handle them.

apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
```diff
@@ -0,0 +1,555 @@
...
+            event_type: EventType::SystemAction
+        ),
+        Err(e) => log!(
+            event_id: EventId::UnknownError,
+            message: format!("validate_connection unexpected error when writing command: {e:?}"),
+            event_type: EventType::SystemStatus
+        ),
+    }
+
+    let mut buf: [u8; BUFFER_MAX_BYTES] = [0; BUFFER_MAX_BYTES];
```
eventualbuddha, line 202: No type annotation needed:
```suggestion
    let mut buf = [0; BUFFER_MAX_BYTES];
```

apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
eventualbuddha: Could maybe have a delay in handling events after a connection if the first N ms it's sending junk?

apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
```diff
@@ -0,0 +1,555 @@
...
+    // Only check for sip & puff actions when the device is connected. When the device is not connected
+    // the values switch meaning.
+    // ie. When a sip & puff is connected, 0x01 is sent to host when a sip
+    // is happening and 0x00 is sent when no sip is happening.
+    // When no sip & puff is connected, 0x00 is sent when no sip is happening.
+    // Therefore we must ignore all sip & puff signals unless we know the sip & puff is connected.
+    if current_status.sip_puff_device_connected == SipAndPuffDeviceStatus::Connected {
+        // Send keypress for new sip event
+        if new_sip_status == SipAndPuffSignalStatus::Active
+            && current_status.sip == SipAndPuffSignalStatus::Idle
```
eventualbuddha, line 363: Do we know whether there's any jitter in these? Will the device handle fluctuation in the sip/puff strength and do the thresholding itself to ensure it always sends a single clean transition event? Or do we need to be concerned about transitions like Active → Active → Active → Idle → Active → Idle → Active → Idle → Idle → Idle?

eventualbuddha (commented):
Looking pretty good so far. Lots of minor suggestions and a couple thoughts on the contract with the sip/puff device.

apps/mark-scan/fai-100-controller/src/commands.rs
```diff
@@ -0,0 +1,346 @@
...
+        bytes.extend_from_slice(&u32::to_le_bytes(SENDER_ID));
+        // Unused padding
+        bytes.append(&mut vec![0; 16]);
+        // LE u32 size of upcoming noise + data = 3 bytes
+        bytes.push(0x03);
+        bytes.append(&mut vec![0; 3]);
+        // 2 bytes of noise
+        bytes.append(&mut vec![0xff; 2]);
+        // 1 byte of data (just the command ID)
+        bytes.push(command.command_id as u8);
```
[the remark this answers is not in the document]
kshen0: Thanks!

apps/mark-scan/fai-100-controller/src/commands.rs
```diff
@@ -0,0 +1,346 @@
...
+                .map_err(|_| CommandError::FailedSliceConversion())?,
+        );
+
+        // Length including 2 bytes of noise we don't need to read
+        let response_length: usize =
+            usize::try_from(data_length - 2).expect("Failed to parse data length");
+        if response_length != 5 {
+            return Err(CommandError::UnexpectedDataLength(response_length));
+        }
+        let response: &[u8] = &bytes[26..26 + response_length];
```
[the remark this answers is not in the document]
kshen0: TIL thanks

apps/mark-scan/fai-100-controller/src/usb_device.rs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        None
+    }
+
+    fn claim_interface<T: UsbContext>(handle: &mut DeviceHandle<T>) -> Result<()> {
+        let has_kernel_driver = match handle.kernel_driver_active(INTERFACE_NUMBER) {
+            Ok(true) => {
+                handle.detach_kernel_driver(INTERFACE_NUMBER).ok();
```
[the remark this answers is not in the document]
kshen0: Good catch. This step is necessary so failing detach should error and be handled.

apps/mark-scan/fai-100-controller/src/usb_device.rs
```diff
@@ -0,0 +1,114 @@
...
+                    address: ENDPOINT_ADDRESS_OUT,
+                };
+
+                Ok(Self {
+                    endpoint_in,
+                    endpoint_out,
+                    handle,
+                })
+            }
+            None => panic!("could not find device {vendor_id:04x}:{product_id:04x}"),
```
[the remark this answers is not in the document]
kshen0: If the device can't be found there's no reason for the daemon to run so we should exit. `bail!`ing and handling the error in main is probably more hygienic than panicking here so I can update that.

I also considered retrying but we found that waiting or retrying with the BMD 155 controller didn't fix the issue when the device was not found. Only exiting and allowing systemctl to restart the daemon worked. It's a completely different device but a hard restart of the daemon does feel less error prone for that reason.

apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
kshen0: I don't actually have a sip & puff on hand at the moment (the one I had is being used for demos) so I'm not sure. I've been testing with buttons that are much more clearly binary.

The report period of the device seems to be 100ms though, so the situation you described is partially mitigated. It would depend on how fast the fluctuation happens and probably varies device to device. It's a good call out that we should test when we have access to a sip and puff device.

apps/mark-scan/fai-100-controller/src/usb_device.rs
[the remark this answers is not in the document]
kshen0: Ended up using the `?` operator. `clippy` seemed a little confused in its recommendation - [here](https://app.circleci.com/pipelines/github/votingworks/vxsuite/16242/workflows/2e7b96dd-f82f-47da-b2ce-0cd656625073/jobs/622433) it suggests explicitly matching over the `Ok` unit type but `--fix` omits handling the `Ok` type altogether. I went with the `--fix` output for simplicity.

apps/mark-scan/fai-100-controller/src/commands.rs
```diff
@@ -0,0 +1,339 @@
...
+#[derive(Debug)]
+pub struct NotificationStatusResponse {
+    pub button_pressed: ButtonSignal,
+    pub sip_status: SipAndPuffSignalStatus,
+    pub puff_status: SipAndPuffSignalStatus,
+    pub sip_puff_device_connection_status: SipAndPuffDeviceStatus,
+}
+
+// Not all possibleresponses are 31 bytes, but all the ones we use are
+pub const RESPONSE_BYTE_LENGTH: usize = 31;
```
eventualbuddha, line 130:
```suggestion
// Not all possible responses are 31 bytes, but all the ones we use are
```

eventualbuddha (approved):
Overall this is safe to try, though I believe it's likely that your `sleep` to "skip" inputs is not going to work.

apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
```diff
@@ -328,6 +328,30 @@ fn handle_status_response(
...
+    if new_connection_status != current_status.sip_puff_device_connected {
+        current_status.sip_puff_device_connected = new_connection_status;
+        current_status.sip = SipAndPuffSignalStatus::Idle;
+        current_status.puff = SipAndPuffSignalStatus::Idle;
+        current_status.button_pressed = ButtonSignal::NoButton;
+
+        // Write device connection status to system file so mark-scan app is aware
+        write_pat_connection_status(new_connection_status, workspace_path)?;
+
+        sleep(2 * POLL_INTERVAL);
```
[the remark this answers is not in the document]
kshen0: I've tested this on hardware and it seems to work well.

The buffer doesn't queue anything unless we send the `GetNotificationStatus` command. That command itself is a misnomer, because there are no notifications or events. Rather, the device responds with a snapshot of the currently activated inputs and PAT connection status. It sends this snapshot exactly once in response to `GetNotificationStatus`, so status reports should map exactly 1:1 with `GetNotificationStatus` commands sent and we shouldn't have to worry about enqueued status reports.

When we detect PAT connection status at has changed we do 2 things

1. Wait for a period so `sip status` and `puff status` can settle
2. Throw out the signals from the previous response because we know we can't trust it

I think to make this more readable and clear, we can change `handle_status_response` to return a `bool` describing whether the next poll should be delayed. We can check that and sleep right before the next `GetNotificationStatus` command is sent. This locates the `sleep` right next to the thing we actually want to delay.

apps/mark-scan/fai-100-controller/src/commands.rs
kshen0: [90b0721](https://github.com/votingworks/vxsuite/pull/4986/commits/90b0721f6a4885cf46812249a3074e13014a1378)

apps/mark-scan/fai-100-controller/src/usb_device.rs
```diff
@@ -39,19 +40,19 @@ impl UsbDevice {
 
             if device_desc.vendor_id() == vendor_id && device_desc.product_id() == product_id {
                 match device.open() {
-                    Ok(handle) => return Some(handle),
-                    Err(e) => panic!("Device found but failed to open: {e}"),
+                    Ok(handle) => return Ok(handle),
+                    Err(e) => bail!("Device found but failed to open: {e}"),
                 }
```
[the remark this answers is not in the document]
kshen0: [90b0721](https://github.com/votingworks/vxsuite/pull/4986/commits/90b0721f6a4885cf46812249a3074e13014a1378)

Issue #4912, opened 2024-06-03
Separate election hash and ballot hash


arsalansufi:
[Product spec](https://docs.google.com/document/d/1FVAOWReHFzd75C5bqdgrh5gl22MFpPyGephc3TqWgIQ/edit)

mattroe:
Thinking on packages: https://votingworks.slack.com/archives/CEL6D3GAD/p1715398497176849?thread_ts=1705526372.720269&cid=CEL6D3GAD

PR #4992, opened 2024-06-25
Add translated ballot strings to election.json

jonahkagan:
## Overview

Refs: https://github.com/votingworks/vxsuite/issues/4912

Adds the translated election strings and HMPB app strings to election.json so that they will be included in the hash that's encoded in the HMPB QR code. If these strings change, the ballot changes, so we need to hash them to prevent out of date ballots from being tabulated.

To implement this, this PR adds a new field `ballotStrings` to the `Election` type. VxDesign adds translated strings to this field instead of the `electionStrings.json` file, which is no longer needed.

## Demo Video or Screenshot
N/A

## Testing Plan
I updated all of the election fixtures to have some ballot strings, which required then updating many tests that snapshotted the election hash in one way or another. I also updated some logical tests that ensure the translated strings are correctly loaded from the election package.

## Checklist


libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan, line 1: Most of this file was copied from libs/utils/src/extract_cdf_ui_strings.ts, which you'll see is deleted below. I made a few changes that I'll call out below, since the diff is not showing them

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -30,6 +31,326 @@ function officeId(contestId: Vxf.ContestId): string {
...
+/**
+ * Returns all the language codes in use in a CDF election based on the election
+ * title internationalized text.
+ */
+function electionLanguageCodes(cdfElection: Cdf.BallotDefinition): string[] {
+  const electionTitleStrings = assertDefined(cdfElection.Election[0]).Name.Text;
+  return electionTitleStrings
+    .map((string) => string.Language)
+    .filter((languageCode) => SUPPORTED_LANGUAGES.has(languageCode));
+}
```
jonahkagan, line 112: This function is new

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -30,6 +31,326 @@ function officeId(contestId: Vxf.ContestId): string {
...
+            new Intl.DisplayNames([languageCode], {
+              type: 'language',
+              style: 'narrow',
+              fallback: 'none',
+            }).of(languageCode)
+          ),
+        })
+      ),
+    });
+  },
```
jonahkagan, line 138: We previously didn't do anything in the extractor function for ballot language, since we could rely on electionStrings.json to provide the translation. Now, if we get a CDF election, we won't have an additional electionStrings.json, so we have to make do and fill in a reasonable default here. I emulated the approach taken in VxDesign's translation code to translate ballot language.

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -30,6 +31,326 @@ function officeId(contestId: Vxf.ContestId): string {
...
+            (languageCode): Cdf.LanguageString => ({
+              '@type': 'BallotDefinition.LanguageString',
+              Language: languageCode,
+              Content: termDescription,
+            })
+          ),
+        });
+      }
+    }
+  },
```
jonahkagan, line 222: Similarly, I needed a default for term description since it doesn't have internationalization support in CDF. In this case, JS doesn't have any native way to translate the arbitrary string, so I couldn't actually do anything but use the English. It's not really a passable solution, so I would say we just don't support term description translations when using CDF, which seems ok since CDF itself doesn't support them.

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -30,6 +31,326 @@ function officeId(contestId: Vxf.ContestId): string {
...
+            year: 'numeric',
+          }).format(
+            new DateWithoutTime(
+              assertDefined(cdfElection.Election[0]).StartDate
+            ).toMidnightDatetimeWithSystemTimezone()
+          ),
+        })
+      ),
+    });
+  },
```
jonahkagan, line 284: Another default I needed to fill in. This time I was able to use JS again

apps/design/backend/src/language_and_audio/ballot_strings.ts
```diff
@@ -29,5 +30,10 @@ export async function translateBallotStrings(
     machineVersion,
     ballotLanguageConfigs
   );
-  return mergeUiStrings(electionStrings, appStrings);
+  // Temporary hack: Only pass the HMPB app strings to the renderer.
+  // TODO: construct and translate these as a separate package
```
jonahkagan, line 34: In my next PR, I'll be building a nicer solution for HMPB app strings, likely a separate UI strings module rather than including them in `appStrings`

jonahkagan:
> Why do the test cert keys need to be regenerated? Because of the election hashes changing? I guess I'm not really sure what those are even for.

Yes, I think that's due to an election hash fixture changing

libs/types/src/cdf/ballot-definition/convert.ts
jonahkagan: Yeah, a warning sounds like a good idea.

> If they import CDF then they should still be able to add term afterwords right?
In VxDesign yes, but not in the rest of VxSuite

PR #4993, opened 2024-06-25
Kevin/paper handler cli set scan dir

kshen0:
## Overview

* Adds support for `setScanDirection` to the driver CLI for manual testing.
* Moves the call to set default scan direction out of `driver.scanAndSave` and into the application layer. Always setting scan direction to `backward` in `scanAndSave` doesn't make a lot of sense because any previous call to `setScanDirection` would be overwritten right before the scan actually happens.
* Fixes a typo where the `in_park` scan direction was encoded with `0x02` instead of `0x03`

## Demo Video or Screenshot

N/A

## Testing Plan

* Ran CLI and confirmed the setting works

## Checklist

- [ ] (No new actions) I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
@@ -10,6 +10,7 @@ import {
   imageDataToBinaryBitmap,
   isPaperAnywhere,
   isMockPaperHandler,
+  ScanDirection,
 } from '@votingworks/custom-paper-handler';
 import { pdfToImages } from '@votingworks/image-utils';
 import { tmpNameSync } from 'tmp';
@@ -36,6 +37,10 @@ export async function setDefaults(
   debug('set line spacing to 0');
   await driver.setPrintingSpeed('slow');
   debug('set printing speed to slow');
+
+  const scanDirection: ScanDirection = 'backward';
+  await driver.setScanDirection(scanDirection);
+  debug('set scan direction to', scanDirection);
 }
 
 export async function printBallotChunks(
--- a/libs/custom-paper-handler/src/driver/coders.ts
+++ b/libs/custom-paper-handler/src/driver/coders.ts
@@ -207,7 +207,7 @@ export enum ConfigureScannerOptionSensorConfig {
 export enum ConfigureScannerFlags {
   NONE = 0x00,
   SCAN_BACKWARDS = 0x01,
-  SCAN_IN_PARK = 0x02,
+  SCAN_IN_PARK = 0x03,
 }
 export const ConfigureScannerCommand = message({
   command: literal(0x1c, 'SPC'),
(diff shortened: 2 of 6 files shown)


apps/mark-scan/backend/src/custom-paper-handler/application_driver.ts
```diff
@@ -36,6 +37,10 @@ export async function setDefaults(
   debug('set line spacing to 0');
   await driver.setPrintingSpeed('slow');
   debug('set printing speed to slow');
+
+  const scanDirection: ScanDirection = 'backward';
+  await driver.setScanDirection(scanDirection);
+  debug('set scan direction to', scanDirection);
```
kshen0, line 43: For https://github.com/votingworks/vxsuite/issues/4812 you may need to move `setScanDirection` into individual states, but you could also leave this default, change the direction for the re-cast flow, and change it back after. Up to you.

libs/custom-paper-handler/src/driver/scanner_config.ts
```diff
@@ -9,7 +9,8 @@ export type PaperMovementAfterScan =
 export type ScanLight = 'red' | 'green' | 'blue' | 'white';
 export type ScanDataFormat = 'BW' | 'grayscale';
 export type Resolution = 100 | 150 | 200 | 250 | 300;
-export type ScanDirection = 'forward' | 'backward' | 'in_park';
+export const scanDirections = ['forward', 'backward', 'in_park'] as const;
```
kshen0, line 12: Allows printing of all valid scan directions in the CLI here: https://github.com/votingworks/vxsuite/pull/4993/files#diff-9b400756f5e8f9849559bce70d1e273e357200c02fea2d5e808e58e33f702ad4R133

libs/custom-paper-handler/src/driver/coders.ts
```diff
@@ -207,7 +207,7 @@ export enum ConfigureScannerOptionSensorConfig {
 export enum ConfigureScannerFlags {
   NONE = 0x00,
   SCAN_BACKWARDS = 0x01,
-  SCAN_IN_PARK = 0x02,
+  SCAN_IN_PARK = 0x03,
```
kofi-q, line 210: For my understanding: this is essentially setting both the `0x01` and `0x02` flags, right?

kofi-q (approved):
🎉

libs/custom-paper-handler/src/driver/scanner_config.ts
```diff
@@ -9,7 +9,8 @@ export type PaperMovementAfterScan =
...
+export type ScanDirection = (typeof scanDirections)[number];
```
kofi-q, line 13: nit: Actually, could you also add a jsdoc comment to `ScanDirection` describing the behaviour of each direction setting?

libs/custom-paper-handler/src/driver/coders.ts
kshen0: Correct!

PR #4994, opened 2024-06-25
Add imprinting to the central scanner 

carolinemodic:
## Overview

When an imprinter module is attached to the scanner and you scan a batch, a new BallotAuditID in the form <Batch_ID>_<Sequence> will be imprinted on each sheet and exported in the CVR, where <Sequence> is a sequential ID starting with 0000, then 0001, etc. The largest scanner we use in the central system scans in max batches of 300 so 4 digit should be plenty for this ID. The ID gets reset on each subsequent batch, and is notably different from the BallotSequenceIdx currently included in the exported CVR in that it increments on rejected sheets, and all sheets, rejected ballots, unreadables, etc. are imprinted. 

BallotAuditId is an existing CDF field we are not currently using.

When an imprinter module is not attached to the scanner everything behaves identically to how to does today. Nothing is imprinted, and the CVR does not include a BallotAuditId. 

## Demo Video or Screenshot
The UI doesn't look any different then today, but here are photos of two sample ballots scanned with the imprinter module attached along with the associated CVRs. 
IMG_1391
[cast-vote-record-report.json](https://github.com/user-attachments/files/15976098/cast-vote-record-report.json)

IMG_1392
[cast-vote-record-report.json](https://github.com/user-attachments/files/15976100/cast-vote-record-report.json)

## Testing Plan

- Manually tested behavior with imprinter attached, scanning multiple batches, batches with unreadbles, etc. to verify all expected behavior
- Manually tested behavior without imprinter attached, verified it matches existing behavior
- Ran tests 

## Checklist

### DIFF
--- a/apps/central-scan/backend/src/fujitsu_scanner.ts
+++ b/apps/central-scan/backend/src/fujitsu_scanner.ts
@@ -136,7 +136,7 @@ export class FujitsuScanner implements BatchScanner {
     if (imprintIdPrefix !== undefined) {
       args.push('--endorser=yes');
       // Imprint the prefix followed by a sequential index for each page in the batch
-      args.push('--endorser-string', `${imprintIdPrefix}_%04d`);
+      args.push('--endorser-string', `${imprintIdPrefix}_%04ud`);
     }
 
     const MM_PER_INCH = 25.3967;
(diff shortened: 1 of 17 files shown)


mattroe (commented):
LGTM!

apps/central-scan/backend/src/fujitsu_scanner.ts
```diff
@@ -82,9 +89,35 @@ export class FujitsuScanner implements BatchScanner {
...
   }
 
+  async isImprinterAttached(): Promise<boolean> {
+    return new Promise((resolve) => {
+      const process = streamExecFile('scanimage', [
+        '-d',
+        'fujitsu',
+        '--endorser=yes',
+        '--format=jpeg',
+        '-n',
```
eventualbuddha, line 99: We should probably do this with the other short options as well, but this one in particular I had to look up to make sure this was doing what I expected.
```suggestion
        '--dont-scan',
```

eventualbuddha (approved):

PR #4995, opened 2024-06-25
Move HMPB strings out of `appStrings`

jonahkagan:
## Overview

Refs: #4912 
Follow up to: #4992 

Moves HMPB strings into a separate string catalog so that they can be easily included in `Election.ballotStrings`. This also has the nice side effect of not generating audio for these strings, since audio is not needed.

## Demo Video or Screenshot
N/A

## Testing Plan
Updated tests

## Checklist


kofi-q (approved):
Nice!

Issue #4578, opened 2024-02-06
vxscan: improve resiliency against skew


carolinemodic:
Improve resiliency against crazy skew situations

Issue automatically created from slack feedback reported by caroline 
https://votingworks.slack.com/archives/CJU9MSC6S/p1707252500947249

carolinemodic:
Some ideas of guards where we should just reject as unreadable:
- Increase the number of timing marks we have to recognize on each side - does this apply to the top/bottom as well as left/right? If not it should. 
- Reject if the skew is over some percent - shouldn't happen often
- Reject if the top timing marks are way above any of the found left/right timing marks
- Look for the pixels where a bubble should be to be black in the bubble locations and reject if some percent of bubbles aren't above some threshold to avoid looking at bubbles in the middle of text.

eventualbuddha:
> Test set to look at https://votingworks.slack.com/archives/CJU9MSC6S/p1707252693335049?thread_ts=1707252500.947249&cid=CJU9MSC6S

All the test set examples in this batch are now rejected with https://github.com/votingworks/vxsuite/pull/5015.

eventualbuddha:
With #5015 I think we can call this done for now. Let's create specific tickets for issues that come up in future testing.

PR #4996, opened 2024-06-25
feat(ballot-interpreter): add more debug information for partial timing marks

eventualbuddha:
## Overview

Adds more debug information to the partial timing mark debug image and includes more specific reasons why the timing marks could not be found. This should not change the behavior of interpretation at all, just output more debug info along with some minor refactors.

Refs #4578 

## Demo Video or Screenshot
image

## Testing Plan
- [x] Existing automated tests

### DIFF
--- a/libs/types-rs/src/geometry.rs
+++ b/libs/types-rs/src/geometry.rs
@@ -5,6 +5,8 @@ use std::{
 
 use serde::{Deserialize, Serialize};
 
+use crate::f32_newtype::f32_newtype;
+
 /// A unit of length in timing mark grid, i.e. 1 `GridUnit` is the logical
 /// distance from one timing mark to the next. This does not map directly to
 /// pixels.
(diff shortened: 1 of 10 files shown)


libs/types-rs/src/geometry.rs
```diff
@@ -280,95 +394,42 @@ pub enum IntersectionBounds {
...
-    diff.min(PI - diff)
+    let diff = normalize_angle(Radians::new(a.0 - b.0));
+    Radians::new(f32::min(diff.0, PI - diff.0))
 }
 
 /// Normalize angle to [0, PI). This means that two angles that are
 /// equivalent modulo PI will be equal, e.g. 90° and 270°, even though
 /// they are not equal in the mathematical sense.
 #[must_use]
 pub fn normalize_angle(angle: Radians) -> Radians {
```
jonahkagan, line 408: Do we still need this function if it's a method on `Radians` now?

jonahkagan (approved):

libs/types-rs/src/geometry.rs
eventualbuddha: Nope, good catch. I'd removed this in an earlier iteration but it came back when I re-built it in commits.

PR #4999, opened 2024-06-27
feat(ballot-interpreter): reject ballots whose timing mark borders are way off

eventualbuddha:
## Overview

Refs #4578 

When the timing mark border lines are determined to be more than 10% of width inset from the image edge we now reject the ballot. This should prevent a variety of situations where we might accidentally infer the wrong set of shapes as the best set of timing marks for a given edge.

To make that inference less likely, this change also is smarter about picking the subset of rectangles we determine is the set of timing marks for a given side. To do that we now favor subsets of rectangles that are closest to the edge we're looking for while also having a favorable count (i.e. closest to expected value).

## Demo Video or Screenshot
In this example all of the candidate subsets are equivalent when looking for the top line under the old approach and so the last one was used (the bottom one). The new approach uses the one closest to the top that is no worse in terms of timing mark count.

| Before | After |
|-|-|
| rotated-back_debug_partial_timing_marks-before | rotated-back_debug_partial_timing_marks |

## Testing Plan
- [x] Tested with m15 backup.
- [x] Tested with @arsalansufi's latest test images from 6/26/24.


libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -443,48 +459,130 @@ pub fn find_partial_timing_marks_from_candidate_rects(
...
+    }
+
     let half_height = (geometry.canvas_size.height / 2) as PixelPosition;
     let half_width = (geometry.canvas_size.width / 2) as PixelPosition;
     let top_half_rects = rects
         .iter()
         .filter(|r| r.top() < half_height)
         .copied()
-        .collect::<Vec<Rect>>();
+        .collect::<Vec<_>>();
```
eventualbuddha, line 482: This doesn't change the functionality. It just tells Rust "you figure out the type where `_` appears".

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -443,48 +459,130 @@ pub fn find_partial_timing_marks_from_candidate_rects(
...
+        match a.len().cmp(&b.len()) {
+            std::cmp::Ordering::Equal => {
+                // if the counts are equal, sort by bottom (where we want the largest value)
+                average(a.iter().map(|r| r.bottom())).cmp(&average(b.iter().map(|r| r.bottom())))
+            }
+            cmp => cmp,
+        }
+    };
+
+    let cmp_left_line_candidates = |a: &Vec<&Rect>, b: &Vec<&Rect>| -> std::cmp::Ordering {
```
eventualbuddha, line 519: This needs to be a closure because functions cannot capture surrounding variables (i.e. `geometry`). I made all the `cmp_*` functions closures just for consistency even though the above two could be regular `fn`s.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -443,48 +459,130 @@ pub fn find_partial_timing_marks_from_candidate_rects(
...
+        // we can do this because the right side should be completely filled
+        let a_diff_from_expected = (a.len() as i32 - geometry.grid_size.height).abs();
+        let b_diff_from_expected = (b.len() as i32 - geometry.grid_size.height).abs();
+
+        match a_diff_from_expected.cmp(&b_diff_from_expected) {
+            std::cmp::Ordering::Equal => {
+                // if the counts are equal, sort by right (where we want the largest value)
+                average(a.iter().map(|r| r.right())).cmp(&average(b.iter().map(|r| r.right())))
+            }
+            // swap the ordering because we're using max_by and we want to minimize the diff
```
eventualbuddha, line 547: The number of comments I added to this effect indicates this may be a bit of a code smell. I'm open to suggestions for improvements.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -443,48 +459,130 @@ pub fn find_partial_timing_marks_from_candidate_rects(
...
-    let mut right_line = find_largest_subset_intersecting_line(
-        &right_half_rects,
-        VERTICAL_ANGLE,
-        MAX_BEST_FIT_LINE_ERROR,
-    );
+    )
+    .max_by(cmp_bottom_line_candidates)?
+    .into_iter()
+    .copied()
+    .collect::<Vec<_>>();
```
eventualbuddha, line 569: `cargo fmt` yuno format this one the same way?! 😒

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -705,6 +803,38 @@ pub enum FindCompleteTimingMarksError {
...
     MissingCorners,
+
+    /// One of the timing mark border sides is invalid.
+    #[error("Invalid timing mark side: {side_marks:?}")]
+    InvalidTimingMarkSide { side_marks: SideMarks },
+}
+
+#[derive(Debug)]
+pub enum SideMarks {
+    #[allow(dead_code)]
```
eventualbuddha, line 814: This is here because we only ever debug-print this `enum` and don't actually access the variant properties.

libs/types-rs/src/geometry.rs
```diff
@@ -432,74 +431,6 @@ pub fn find_largest_subset_intersecting_line(
                     .collect::<Vec<_>>(),
             )
         })
-        // Pick the list of rectangles that is the longest.
-        .max_by_key(Vec::len)
```
eventualbuddha, line 436: The previous version hardcoded the criterion for picking the best subset to "most number of rectangles". Since we want the criterion to vary depending on which edge we're looking at, I changed this function to return an iterator over all the subsets of rectangles.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -715,49 +845,115 @@ pub fn find_complete_timing_marks_from_partial_timing_marks(
     partial_timing_marks: &Partial,
     debug: &ImageDebugWriter,
 ) -> FindCompleteTimingMarksResult {
-    let top_line = &partial_timing_marks.top_rects;
-    let bottom_line = &partial_timing_marks.bottom_rects;
-    let left_line = &partial_timing_marks.left_rects;
-    let right_line = &partial_timing_marks.right_rects;
+    const ALLOWED_TIMING_MARK_INSET_PERCENTAGE: UnitIntervalValue = 0.15;
```
eventualbuddha, line 848: Here's what a 15% inset looks like:
image

I tried going lower than this (10%), but one of our regression tests began failing (`test_inferred_missing_metadata_from_one_side`). Admittedly, that test image is *very* skewed and has a large inset.

eventualbuddha: I measured it, and the max timing mark border rotation is 1.83º in this fixture:
image
The most I saw in the set of images @arsalansufi gave me from the latest hardware was about 1.3º.

eventualbuddha:
Adding @benadida back as a reviewer for high-level strategy. @kshen0 still reviewing for code etc.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
benadida: Given your description of that one regression test, I would suggest killing it and reducing to 10%.

benadida (approved):
high level strategy feels safe and solid.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: Here's the debug image for the fixture with a dark red rect showing the inset:
image

I'm somewhat inclined to leave it at 15%, but if we want to try 10% I can run it against the latest images from @arsalansufi and the m15 backups.

benadida: I'm shocked that image works at all! Impressive.

I would be very comfortable, in fact I'd encourage us to not attempt to scan this image, and if that means we take less risk in our extrapolation, then that's a win.

eventualbuddha: On second though and after looking at this debug image for several of @arsalansufi's latest images, I want to try 10%.

eventualbuddha: Alright, here's the worst rotation from the latest images (top rotation 1.49º):
2479d422-714f-4d6e-b79a-45571f5439d1-front_debug_partial_timing_marks

The timing marks are still fairly comfortably within the 10% inset, so let's go with 10%.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: Yes, but the idiomatic approach in Rust is usually to let the compiler infer types when it can. In this case the inferred `_` has to be `Rect` and is specified elsewhere, so there’s no need to duplicate it.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: I think they'd be able to access `average` even if they were `fn`s.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -443,48 +470,130 @@ pub fn find_partial_timing_marks_from_candidate_rects(
...
+            }
+            cmp => cmp,
+        }
+    };
+
+    let cmp_left_line_candidates = |a: &Vec<&Rect>, b: &Vec<&Rect>| -> std::cmp::Ordering {
+        // compare by the difference from the expected number of timing marks.
+        // we can do this because the left side should be completely filled
+        let a_diff_from_expected = (a.len() as i32 - geometry.grid_size.height).abs();
+        let b_diff_from_expected = (b.len() as i32 - geometry.grid_size.height).abs();
```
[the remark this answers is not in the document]
eventualbuddha: I'm not sure. It could theoretically be `u32` instead.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: We could, but we'd need to swap the earlier `cmp` calls.

kshen0 (approved):

libs/types-rs/src/geometry.rs
```diff
@@ -397,29 +397,28 @@ pub fn angle_diff(a: Radians, b: Radians) -> Radians {
...
         // Map to lists of rectangles in line with each pair.
-        .filter_map(|(rect, other_rect)| {
+        .filter_map(move |(rect, other_rect)| {
             let line_angle = Radians::new(
                 (other_rect.center().y - rect.center().y)
                     .atan2(other_rect.center().x - rect.center().x),
             );
             if angle_diff(line_angle, angle) > tolerance {
                 // The line between the two rectangles is not within the
-                // tolerance of the desired angle, so skip this pair.
```
jonahkagan, line 422: Seems like this comment is cut off mid-sentence after this deletion @eventualbuddha

eventualbuddha: https://github.com/votingworks/vxsuite/pull/5060

PR #5002, opened 2024-06-28
fix(ballot-interpreter): handle smudges on timing marks better

### LINKED ISSUE #3828: Interp robustness: handle timing mark smudges from m17
https://votingworks.slack.com/archives/C055XBPE7D1/p1690738872777129

eventualbuddha:
## Overview

Fixes #3828 

When a smudge that's connected to a timing mark causes the contours algorithm to report a rectangle containing both the timing mark and the adjacent smudge, we might filter out the timing mark because it's too big but fail to filter out some smaller rectangle that's in the smudge. This smaller rectangle might then be used as the timing mark, leading to incorrect interpretation. This commit ignores timing marks that are too different in size from their immediately surrounding neighbors, letting them be inferred when generating the complete set of timing marks.

### Before
image
image

### After
image

## Testing Plan
Tested on the m15 backup and @arsalansufi's latest hardware test images.


libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -680,6 +699,38 @@ pub fn find_partial_timing_marks_from_candidate_rects(
     Some(partial_timing_marks)
 }
 
+/// Filters rectangles where they are not within the sizes of their immediately
+/// surrounding neighbors, within a certain error tolerance. This is useful for
+/// filtering out rectangles that are not part of the timing marks.
+fn filter_rects_not_following_size_progression_of_neighbors(
```
carolinemodic, line 705: What actually is counted as a neighbor? If a timing mark is surrounded by a lot of little smudges that get seen as little rectangles would the timing mark be the rectangle thats filtered out instead of the little ones?

carolinemodic (approved):
Are there other ballots in the backup sets that have timing marks filtered from this?

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: I tried this out on the m15 and v4 hardware test backups. It performed reasonably well, though I think it would be vulnerable to this type of issue where multiple consecutive small rects would not all be filtered out. I experimented with a few different approaches to filtering and ended up with a version that filters based on the standard deviation for a particular side. This version worked well except that it filtered out the top left or top right corner (depending on which way the paper was rotated) in about half the ballots processed. It would be subsequently inferred in the right spot, but I still felt this should be avoided.

To address the problem of removing the corner, I changed the filtering code to ignore the first and last on the top and bottom edges. This works much better, only filtering rects in 10 out of 906 images in the m15 backup and 2 out of 60 images in the v4 hardware test backup. In addition to the smudge that prompted this change, the only other shapes filtered out were artifacts at the corners of the image like this:

image

I feel pretty confident in this new approach and will post the code after cleaning it up a bit.

eventualbuddha: Here's the smudge ballot with a new debug image around filtering timing marks:
image

PR #5005, opened 2024-06-29
support 8" sheet printing and scanning

kshen0:
## Overview

* Adds a feature flag to toggle 150 vs 155 hardware config. 155 is the default.
* Adds support for 8" width printing.

Setting page width to exactly 8" causes an overflow of dots, which pushes those dots to the next line like so:

IMG_5402

Setting printable page width to 7.795" fixes the issues. I'd still like to investigate for the root cause but this will unblock 8" printing currently.

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/1060688/1d36b2ba-17d3-4757-badf-f44bef18b685

IMG_5403

## Testing Plan
- Tested manually with the printer/scanner diagnostic
- I didn't add tests for 150/155 hardware model branching because those decisions are still in flux

## Checklist

- [x] (N/A) I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.


apps/mark-scan/backend/src/globals.ts
```diff
@@ -31,3 +32,11 @@ export const MARK_SCAN_WORKSPACE =
   (NODE_ENV === 'development'
     ? join(__dirname, '../dev-workspace')
     : undefined);
+
+/**
+ * What kind of hardware is the app running on?
+ */
+export const MARK_SCAN_BMD_MODEL: BmdModelNumber = process.env
```
kshen0, line 39: The feature flag module isn't initialized at this point so I check the environment variable directly. Not sure if this is the best way to do it, but seemed relatively harmless since the flag library and environment variables are very closely coupled.

libs/custom-paper-handler/src/printing.ts
```diff
@@ -150,10 +149,6 @@ export function imageDataToBinaryBitmap(
 export function chunkBinaryBitmap(
   binaryBitmap: BinaryBitmap
 ): PaperHandlerBitmapExt[] {
-  if (binaryBitmap.width > DEVICE_MAX_WIDTH_DOTS) {
-    throw new Error("binaryBitmap width exceeds printer's allowed size");
-  }
```
kshen0, line 155: This is already checked [in driver.printChunk()](https://github.com/votingworks/vxsuite/blob/f09310f2cf6bb1d1f4b43650cf335ee86930839c/libs/custom-paper-handler/src/driver/driver.ts#L810-L813).

apps/mark-scan/backend/src/globals.ts
kofi-q: Could replace with a `getMarkScanBmdModel()` function instead if there are no constraints WRT this being a constant.

kofi-q (approved):
LGTM!

Issue #4927, opened 2024-06-05
vxmarkscan: highlight button on overvote modal when in controller mode


mattroe:
https://votingworks.slack.com/archives/C03Q42V9MJ5/p1717449420443549?thread_ts=1717449283.634269&cid=C03Q42V9MJ5

The core problem here is that we don't auto-focus the "Okay" button when the modal is surfaced. This problem likely applies to other modals, too

PR #5011, opened 2024-07-02
Fix bug to autofocus okay button on overvote modal

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/4927 
Because of the way React automatically handles focusing of Modals using the autofocus html5 attribute on a button inside a Modal does not work and the "Okay" to dismiss the overvote warning modal does not get properly autofocused. Instead we have to manually focus the button ref in order for it to autofocus as implemented here. Note the onAfterOpen was not triggering at the right time in the Modal component unless I passed through the isOpen parameter I know we don't otherwise use. 

## Testing Plan
Manually triggered overvotes saw "okay" focus 

## Checklist

### DIFF
--- a/libs/mark-flow-ui/src/components/candidate_contest.tsx
+++ b/libs/mark-flow-ui/src/components/candidate_contest.tsx
@@ -29,6 +29,7 @@ import {
   electionStrings,
   ReadOnLoad,
   AssistiveTechInstructions,
+  PageNavigationButtonId,
 } from '@votingworks/ui';
 import { assert } from '@votingworks/basics';
 
@@ -381,8 +382,8 @@ export function CandidateContest({
               {appStrings.warningOvervoteCandidateContest()}
               <AudioOnly>
                 <AssistiveTechInstructions
-                  controllerString={appStrings.instructionsBmdSelectToContinue()}
-                  patDeviceString={appStrings.instructionsBmdSelectToContinuePatDevice()}
+                  controllerString={appStrings.instructionsBmdNextToContinue()}
+                  patDeviceString={appStrings.instructionsBmdMoveToSelectToContinuePatDevice()}
                 />
               </AudioOnly>
             </P>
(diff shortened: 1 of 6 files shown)


kofi-q (commented):
This approach might cancel out the screen reader audio that's supposed to play when the modal opens (so, instead of reading out the overvote message, it just reads the button audio instead).

I think the more complete/pattern-consistent fix would be to update the audio message to say `"Press the right button to continue"` instead of `"Press the select button to continue"` and then assign the ["next" navigation id](https://github.com/votingworks/vxsuite/blob/273a2d9/libs/ui/src/language_settings/language_settings_screen.tsx#L118) to the Okay button. The issue there is that the old navigation logic assumes theres only one "next" id button on the page at a time, but the contest page underneath the modal also has a "next" button. We'd need to update the [navigation logic](https://github.com/votingworks/vxsuite/blob/8565b9a/libs/ui/src/accessible_controllers/navigation.ts#L60) to use `querySelectorAll` instead of `getElementById` and filter out the hidden one with the [`getTabEnabledElementsInHiddenBlocks`](https://github.com/votingworks/vxsuite/blob/8565b9a/libs/ui/src/accessible_controllers/navigation.ts#L20) utility.

carolinemodic:
Ah you're right good catch

libs/mark-flow-ui/src/components/candidate_contest.tsx
```diff
@@ -392,6 +393,7 @@ export function CandidateContest({
               variant="primary"
               autoFocus
               onPress={closeAttemptedVoteAlert}
+              id={PageNavigationButtonId.NEXT}
             >
               {appStrings.buttonOkay()}
               <AudioOnly>
```
carolinemodic, line 399: Kept these next audio descriptions unchanged as at that point you are focusing on the Okay button

carolinemodic:
back to you @kofi-q

kofi-q (approved):
LGTM!

Issue #4930, opened 2024-06-05
vxmarkscan: prevent accidental casting when using ATI


mattroe:
https://votingworks.slack.com/archives/C03Q42V9MJ5/p1717449364279619?thread_ts=1717449283.634269&cid=C03Q42V9MJ5

arsalansufi:
We _think_ what's happening here is clicking the right button on the review screen is auto-casting. If that's the case, we should fix that.

PR #5019, opened 2024-07-03
Prevent accidental ballot casting with accessible controller 

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/4930

In order to prevent pressing the right arrow too quickly and accidental casting ballots this PR makes the right arrow on the review screen simply focus on the "Print my Ballot" button rather then clicking the button. So the user will now need to press the right button and the select button, unless they are using the touch screen and just press the button. I added an additional audio instruction to confirm your selections and print to the Print My Ballot when focused as well. 

## Testing Plan
Manually Tested with keyboard. Ran tests for navigation helper function changes. 

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/lib/assistive_technology.ts
+++ b/apps/mark-scan/frontend/src/lib/assistive_technology.ts
@@ -18,7 +18,10 @@ export function handleKeyboardEvent(event: KeyboardEvent): void {
       triggerPageNavigationButton(PageNavigationButtonId.PREVIOUS);
       break;
     case Keybinding.PAGE_NEXT:
-      triggerPageNavigationButton(PageNavigationButtonId.NEXT);
+      triggerPageNavigationButton(
+        PageNavigationButtonId.NEXT,
+        PageNavigationButtonId.NEXT_AFTER_CONFIRM
+      );
       break;
     case Keybinding.FOCUS_PREVIOUS:
       advanceElementFocus(-1);
--- a/libs/mark-flow-ui/src/pages/review_page.tsx
+++ b/libs/mark-flow-ui/src/pages/review_page.tsx
@@ -53,11 +53,17 @@ export function ReviewPage(props: ReviewPageProps): JSX.Element {
   const printMyBallotButton = (
     <LinkButton
       to={printScreenUrl}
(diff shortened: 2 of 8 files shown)


libs/mark-flow-ui/src/pages/review_page.tsx
kofi-q, line 1: We should also make the change to this final confirmation screen: https://github.com/votingworks/vxsuite/blob/main/apps/mark-scan/frontend/src/pages/validate_ballot_page.tsx#L84 - initially thought the feedback was referring to that screen, but regardless, it does feel like double confirmation is useful on this first review page as well

libs/ui/src/accessible_controllers/navigation.ts
```diff
@@ -53,20 +54,29 @@ export function advanceElementFocus(direction: 1 | -1): void {
...
+    (navigationButtons[0] as HTMLElement).click();
     return;
   }
+  const navigationOnConfirmButtons = Array.from(
+    document.querySelectorAll(`#${navigationOnConfirmId}`)
+  ).filter((e) => !hiddenElements.has(e) && e instanceof HTMLElement);
 
-  (visibleButtons[0] as HTMLElement).click();
+  if (navigationOnConfirmButtons.length >= 1) {
+    console.log('we are focusing');
```
kofi-q, line 79: stray debug line

apps/mark-scan/frontend/src/lib/assistive_technology.ts
```diff
@@ -18,7 +18,10 @@ export function handleKeyboardEvent(event: KeyboardEvent): void {
       triggerPageNavigationButton(PageNavigationButtonId.PREVIOUS);
       break;
     case Keybinding.PAGE_NEXT:
-      triggerPageNavigationButton(PageNavigationButtonId.NEXT);
+      triggerPageNavigationButton(
```
kofi-q, line 21: Worth making the same change to https://github.com/votingworks/vxsuite/blob/main/apps/mark/frontend/src/lib/gamepad.ts#L28 just so that instance of the "print my ballot" button isn't in a broken state on VxMarkOld

kofi-q (approved):
🚢

PR #5020, opened 2024-07-03
chore(deps): update to `typescript` v5.5.3

eventualbuddha:
See the release notes here: https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/

I updated the few places I could easily find where we can benefit from TS v5.5's new ability to automatically infer type narrowing functions (e.g. better `filter` type narrowing). We also may benefit from a tooling perspective in the future if we turn on `isolatedDeclarations`, but this isn't straightforward to do right now.

### DIFF
--- a/libs/ballot-encoder/tsconfig.json
+++ b/libs/ballot-encoder/tsconfig.json
@@ -4,6 +4,7 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "lib": ["dom", "dom.iterable", "esnext"],
+    "skipLibCheck": true,
     "module": "node16",
     "moduleResolution": "node16",
     "noUnusedParameters": true,
--- a/libs/ui/src/modal.test.tsx
+++ b/libs/ui/src/modal.test.tsx
@@ -115,10 +115,12 @@ describe('Modal', () => {
 
   it('handles after open', () => {
     // Work around a react-modal bug: https://github.com/reactjs/react-modal/issues/903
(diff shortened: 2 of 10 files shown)


libs/ballot-encoder/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "lib": ["dom", "dom.iterable", "esnext"],
+    "skipLibCheck": true,
```
eventualbuddha, line 7: This works around this issue: https://stackoverflow.com/questions/75758154/how-to-fix-conflicting-abortsignal-in-node-modules-types-node-globals-d-ts-and. I didn't want to update `@types/node` because we're still using an old version of NodeJS and didn't want TS to promise APIs that aren't available.

libs/ui/src/modal.test.tsx
```diff
@@ -115,10 +115,12 @@ describe('Modal', () => {
 
   it('handles after open', () => {
     // Work around a react-modal bug: https://github.com/reactjs/react-modal/issues/903
-    jest.spyOn(window, 'requestAnimationFrame').mockImplementation((cb) => {
```
eventualbuddha, line 118: TS v5.5 flags `cb` as having type `any`, presumably because `window.requestAnimationFrame` is overloaded?

kshen0 (approved):

Issue #5029, opened 2024-07-09
Change smart card auth to use election ID and date instead of hash


jonahkagan:
[Product spec](https://docs.google.com/document/d/1FVAOWReHFzd75C5bqdgrh5gl22MFpPyGephc3TqWgIQ/edit)

PR #5030, opened 2024-07-09
Add `Election.id` field

jonahkagan:
## Overview

Task: https://github.com/votingworks/vxsuite/issues/5029

For the above task, we want to change smart cards to encode and verify the election id and date rather than the ballot hash. As a first step towards this goal, this PR adds an `Election.id` field to the election definition. This id is generated by VxDesign.

Though we don't expect an election's id to change, we include it in the election definition (and thus the ballot hash) to defensively enforce this assumption. That way, if an election id ever does change, it will be very obvious.

## Demo Video or Screenshot

N/A

## Testing Plan
Updated existing tests

## Checklist


libs/cvr-fixture-generator/README.md
jonahkagan, line 1: Side quest to make it easier to regenerate CVR fixtures (and know when they get out of date)

libs/cvr-fixture-generator/README.md
```diff
@@ -26,6 +26,14 @@ Optional flags:
   the CDF). Since ballot ids are incrementing numbers starting from 1, they will
   have id collisions in VxAdmin if they are not distinguished by a prefix.
 
+## Saved Fixtures
```
carolinemodic, line 29: ty

carolinemodic (approved):

PR #5031, opened 2024-07-09
[MarkScan] Screens for invalid ballot re-insertion

kofi-q:
## Overview

Related to https://github.com/votingworks/vxsuite/issues/4812

Adding voter-facing error screens for the ballot re-insertion flow - shown after the voter pulls a ballot after it's printed and re-inserts a sheet that isn't valid BMD ballot for the current election/precinct/test mode.

Needed to remove the valid interpretion assertion guard in the `getInterpretation` API to do this - moved the assertion into the `ValidateBallotPage`.

Also adding clearer instructions to the `ValidateBallotPage` while here, per product requirements.

### Error Screens:

_(May have tweaked the strings a bit since these screenshots)_

Screenshot 2024-07-09 at 17 29 55

## Testing Plan
- Test cases for the higher-level glue screen, `ReinsertedInvalidBallotScreen`

## Checklist


libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -1083,6 +1103,30 @@ export const appStrings = {
...
+
+  titleBmdInvalidBallotWrongPrecinct: () => (
+    <UiString uiStringKey="titleBmdInvalidBallotWrongPrecinct">
+      Wrong Precinct
+    </UiString>
+  ),
+
+  titleBmdInvalidBallotWrongTestMode: () => (
+    <UiString uiStringKey="titleBmdInvalidBallotWrongTestMode">
+      Wrong Test Mode
```
kshen0, line 1126: "Test Mode" seems oddly opinionated when we use this to refer to both test and live ballot mode, but we don't really have a way to refer to that mode in a neutral way. If we were to change this, maybe "Ballot Mode"? Just "Mode"?

kshen0 (approved):

apps/mark-scan/backend/src/app.test.ts
```diff
@@ -554,7 +554,7 @@ test('printing ballots', async () => {
   await waitForStatus('presenting_ballot');
   const interpretationChinese = await apiClient.getInterpretation();
   assert(interpretationChinese);
-  expect(interpretationChinese.type).toEqual('InterpretedBmdPage');
+  assert(interpretationChinese.type === 'InterpretedBmdPage');
```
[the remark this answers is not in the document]
kofi-q: Moving the interpretation type assertion out of the API method meant that the `.votes` field access below started failing the type check (since jest `toEqual` doesn't provide compile-time assertion)

libs/ui/src/ui_strings/app_strings.tsx
kofi-q: Ah, SGTM

PR #5037, opened 2024-07-10
[MarkScan] UI Screens for Invalid New Sheet Insertion

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/4812

Adding error screens for the upcoming changes to scan new sheets at the start of the session.
These are shown if a newly inserted sheet contains a valid BMD ballot for a different election/precinct/ballot mode, or contains a BMD ballot with an unreadable QR code.

## Demo Video or Screenshot

Screenshot 2024-07-10 at 17 11 35

## Testing Plan
- Tests cases for UI branching logic 
- Manual verification locally

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/pages/inserted_unreadable_ballot_screen.tsx
+++ b/apps/mark-scan/frontend/src/pages/inserted_unreadable_ballot_screen.tsx
@@ -1,0 +1,25 @@
+import { Caption, Icons, P } from '@votingworks/ui';
+
+import { CenteredCardPageLayout } from '../components/centered_card_page_layout';
+
+export function InsertedUnreadableBallotScreen(): JSX.Element {
+  return (
+    <CenteredCardPageLayout
+      icon={<Icons.Warning color="warning" />}
+      title="Unable to Read Ballot"
+      voterFacing={false}
+    >
+      <P>
+        There was a problem while reading the ballot information on the inserted
+        sheet.
+      </P>
+      <P>
+        Please remove the sheet and insert a valid BMD ballot for the configured
+        election.
+      </P>
+      <Caption>
+        <Icons.Info /> Insert a blank sheet to start a new voting session.
+      </Caption>
+    </CenteredCardPageLayout>
+  );
+}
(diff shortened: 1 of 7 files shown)


apps/mark-scan/frontend/src/pages/inserted_unreadable_ballot_screen.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <P>
+        There was a problem while reading the ballot information on the inserted
+        sheet.
+      </P>
+      <P>
+        Please remove the sheet and insert a valid BMD ballot for the configured
+        election.
+      </P>
+      <Caption>
+        <Icons.Info /> Insert a blank sheet to start a new voting session.
```
kshen0, line 21: I _think_ if you inserted a blank page you would get to this page because the interpretation doesn't currently distinguish between a printed page with no QR code and a truly blank page. Does this mean that one could get to this page by inserting a blank sheet, then can get to a new session by removing and reinserting the same blank sheet?

kofi-q: IIUC, the interpreter returns `BlankPage` interpretation types in both the blank-page and no-qr-code cases, so that goes down a different code path and doesn't render this page (asserted [here](https://github.com/votingworks/vxsuite/pull/5037/files#diff-c90f0fc6a88782e4d7b577daabd030768455a1796bac82c69a3d929a8fe104c8R14)).

Inserting a blank sheet will start a new session as it currently does - verified while testing out the prototype (with the mock paper handler, admittedly, so I might be missing an edge case here)

kshen0: Ah ok I missed this - BMD interpretation returns `UnreadablePage` when there are multiple QR codes.

https://github.com/votingworks/vxsuite/blob/main/libs/ballot-interpreter/src/interpret.ts#L630-L646

kshen0 (approved):

Issue #3527, opened 2023-05-30
VxAdmin: Handle BMD Write-Ins


adghayes:
## Summary

Currently VxAdmin does not have any handling of BMD write-ins. We had [a discussion](https://votingworks.slack.com/archives/CEL6D3GAD/p1683580468063749) on how we'd like to handle BMD write-ins. We'd like BMD ballots to have images, like HMPBs. Ideally these images are also zoomable on the contest in question, like HMPBs. In order to accomplish that, we'd have to have BMD ballot layouts calculated and exported with the BMD ballot images, which makes this a substantial project.

## Requirements

- Adjudicate BMD write-ins in existing VxAdmin write-in interface
- Include and show image of ballot for adjudication flow
- (Flexible) Enable zooming in on the BMD contest, as you can with HMPBs

adghayes:
@mattroe and @adghayes discussed a few weeks ago with more specifics about scope. We will not enable any kind of zooming on the ballot. We _will_ show the ballot image, just without zoom. The voter transcribed value can be presented within the sidebar, with the administrator then selecting an adjudicated value (as they might for an HMPB ballot). Optionally, we may implement auto-selection if the transcribed value perfectly matches an adjudication option.

adghayes:
Building and validating this feature requires scanning BMD ballots for `gridlayouts` elections, which is unfortunately broken on `main`: https://github.com/votingworks/vxsuite/issues/4659. This task is blocked on that issue. cc @arsalansufi

arsalansufi:
Thanks for noting @adghayes! I think it's fine to put your focus elsewhere for now. Planning to have Brian address that issue more fully (pending CACVote timelines). There's also a quick fix that we can implement to unblock (try interpreting as an HMPB, catch if there's an error and try interpreting as a BMDB). It doesn't need to be you who does that, just documenting for our future selves.

PR #5039, opened 2024-07-10
Support BMD Write Ins in VxAdmin

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/3527 

Handles BMD write-ins in VxAdmin. All of the necessary info is already exported from the scanners. This PR writes the text the voter entered in VxMark to the DB with importing CVRs and includes those write ins in the adjudication queue. WHen BMD ballots are shown in WIA they do not support zoom and do show the text entered by the voter. Additionally that text auto populates the add new candidate typeahead. 

## Demo Video or Screenshot

https://github.com/votingworks/vxsuite/assets/14897017/65b3b7e3-6778-438d-8c91-7cd3a8a161d7

## Testing Plan
Manually tested as in video above. 
Added test cases. 

## Checklist


apps/admin/backend/src/types.ts
```diff
@@ -332,7 +332,13 @@ export type WriteInAdjudicationAction =
 /**
  * Information necessary to display a write-in on the frontend.
  */
-export interface WriteInImageView {
+export type WriteInImageView = HmpbWriteInImageView | BmdWriteInImageView;
+
+/**
+ * Information necessary to display a hmpb write-in on the frontend.
+ */
+
```
jonahkagan, line 340: Extra blank line

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -117,6 +117,7 @@ const BallotViews = styled.div`
 const AdjudicationControls = styled.div`
   display: flex;
   flex-direction: column;
+  width: 575px;
```
jonahkagan, line 130: I wonder if this should be set as a proportion of the screen width in case we ever change screen sizes? And/or whether it should be in `rem` rather than pixels for consistency with the rest of our styles?

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -87,6 +92,11 @@ const ContestTitle = styled(H2)`
   margin: 0 !important;
 `;
 
+const ContestSubTitle = styled.div`
+  padding-top: 0.5rem;
+  word-break: break-word;
+`;
```
jonahkagan, line 98: I wonder if this should be styled to pop more, since we want the user to focus on it when viewing a BMD write-in.

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
```diff
@@ -306,6 +316,17 @@ export function WriteInsAdjudicationScreen(): JSX.Element {
...
+  const writeInAdjudicationContext = writeInAdjudicationContextQuery.data;
+  const currentWriteIn = writeInAdjudicationContext?.writeIn;
+  const currentWriteInMarkedInvalid =
+    currentWriteIn &&
+    currentWriteIn.status === 'adjudicated' &&
+    currentWriteIn.adjudicationType === 'invalid';
+
+  const writeInImageView = writeInImageViewQuery.data;
+  const hmpbWriteInImageView = writeInImageView as HmpbWriteInImageView;
+  const bmdWriteInImageView = writeInImageView as BmdWriteInImageView;
```
jonahkagan, line 328: I found it a little confusing to cast `writeInImageView` to both `Hmpb` and `Bmd` types and then check their fields as a way of determining which one it is.

Wondering if we could do something like:

```ts
const isHmpb = 'ballotCoordinates' in writeInImageView;
```

and then use `isHmpb` lower down when deciding what to render

apps/admin/backend/src/app.write_ins.test.ts
```diff
@@ -337,6 +345,67 @@ test('getWriteInImageView', async () => {
...
+    })
+  ).unsafeUnwrap();
+
+  // look at a contest that can have multiple write-ins per ballot
+  const contestId = 'zoo-council-mammal';
+  const writeInIds = await apiClient.getWriteInAdjudicationQueue({
+    contestId,
+  });
+  expect(writeInIds).toHaveLength(24);
+  const [writeInIdA, writeInIdB] = writeInIds.slice(0, 2);
```
jonahkagan, line 367: Technically, you don't need the `slice` here since the destructuring will omit the rest of the array

apps/admin/backend/src/app.write_ins.test.ts
```diff
@@ -337,6 +345,67 @@ test('getWriteInImageView', async () => {
...
+  const { imageUrl: actualImageUrl, machineMarkedText: machineMarkedTextA } =
+    writeInImageViewA as BmdWriteInImageView;
+
+  expect(machineMarkedTextA).toEqual('Mock Write-In');
+
+  const expectedImage = await loadImage(
+    join(
+      reportDirectoryPath,
+      '02daef18-5915-423a-baba-4130920662d9',
+      '02daef18-5915-423a-baba-4130920662d9-front.jpg'
```
jonahkagan, line 385: Will this identifier change when regenerating CVR fixtures? If so, might be worth thinking about how to make the test resilient to that change

apps/admin/backend/src/app.write_ins.test.ts
```diff
@@ -337,6 +345,67 @@ test('getWriteInImageView', async () => {
...
+  );
+  expect(actualImageUrl).toEqual(expectedImageUrl);
+
+  // check the second write-in image view, which should have the same image
+  // but different writeInCoordinates
+  const writeInImageViewB1 = await apiClient.getWriteInImageView({
+    writeInId: writeInIdB,
+  });
+
+  // contest and ballot coordinates should be the same, but write-in coordinates are different
```
jonahkagan, line 403: these comments about coordinates don't totally make sense to me

jonahkagan (approved):
Looks great!

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
carolinemodic: Yeah I wasn't completely sure what the best way to do this was, will play with that idea.

apps/admin/backend/src/app.write_ins.test.ts
carolinemodic: oh I see now the test I was copying this from was using a manual export... I don't think this actually needs to be tested as its the same code path that is used for hmpb for this piece so will just remove

apps/admin/frontend/src/screens/write_ins_adjudication_screen.tsx
carolinemodic: adding primary color styling after experimenting

PR #5047, opened 2024-07-11
Validate smart card election using ID + date rather than hash

### LINKED ISSUE #5029: Change smart card auth to use election ID and date instead of hash

jonahkagan:
## Overview

Closes #5029 

VxSuite's authentication system ensures that smart cards programmed by VxAdmin are only used on machines that are configured for the same election as that VxAdmin. Currently, we enforce this by encoding the configured election's hash in the cert stored in each java card. As we include more data in the election hash (e.g. translations of ballot content), it makes less sense to force reprogramming of cards whenever any bit of the election content changes.

Instead, this PR moves to a model where we use a "election key" to identify the election. This key is intended to be stable for the life of the election, so that smart cards can still safely be tied to the election without needing to be reprogrammed when election content changes. The stable key we've landed on for now is a combination of the election ID (which should never change and be unique) and the election date (which also should never change).

## Demo Video or Screenshot
N/A

## Testing Plan
Updated automated tests
TODO some sort of manual testing - need to figure out what that should be

## Checklist


libs/types/src/auth/auth.ts
```diff
@@ -17,6 +17,16 @@ export interface ElectionKey {
   date: DateWithoutTime;
 }
 
+/**
+ * Create an {@link ElectionKey} from an {@link Election}.
+ */
+export function electionAuthKey(election: Election): ElectionKey {
```
arsalansufi, line 23: Extreme nit that you're free to ignore: Could call this `constructElectionKey` or `electionKeyFromElection` instead of `electionAuthKey` since "election auth key" sounds like something distinct from an "election key"

libs/auth/applets/OpenFIPS201-v1.10.2-with-vx-mods.cap
arsalansufi, line 1: Technically, the applet doesn't need to be regenerated for this change - it only needs to be regenerated when https://github.com/votingworks/OpenFIPS201 changes. But running the update script does no harm

arsalansufi (approved):
LGTM!

apps/admin/frontend/src/components/smartcard_modal/card_details_view.tsx
```diff
@@ -17,14 +22,17 @@ import {
 } from './status_message';
 import { userRoleToReadableString } from './user_roles';
 
-function checkDoesCardElectionHashMatchMachineElectionHash(
+function checkDoesCardElectionMatchMachineElection(
   programmedUser: User,
   electionDefinition: ElectionDefinition
 ): boolean {
-  if (!('electionHash' in programmedUser)) {
+  if (!('electionKey' in programmedUser)) {
```
carolinemodic, line 29: oh I guess this will supersede my previous comment and we will get invalid card instead of something went wrong?

carolinemodic (approved):

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -359,24 +360,24 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
         return pin;
       }
       case 'election_manager': {
-        assert(electionHash !== undefined);
+        assert(electionKey !== undefined);
```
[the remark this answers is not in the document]
jonahkagan: I think `electionKey` on this line is coming from the machine state (i.e. the configured election package), not the card, so it should be fine?

libs/types/src/auth/auth.ts
jonahkagan: Reasonable feedback! 6eb372e88

apps/admin/frontend/src/components/smartcard_modal/card_details_view.tsx
arsalansufi: For programming purposes, the card will just show up as unprogrammed and allow you to program it. For auth purposes, the card will show up as an invalid card.

The full sequence of events will be as follows:

- Card cert parsing will error because of an incompatibility with [this schema](https://github.com/votingworks/vxsuite/blob/6eb372e8874825c60dcbb157f0b5ff08a7e90f38/libs/auth/src/certs.ts#L157-L168)
- The `JavaCard` layer [will return `undefined` card details](https://github.com/votingworks/vxsuite/blob/6eb372e8874825c60dcbb157f0b5ff08a7e90f38/libs/auth/src/java_card.ts#L445)
- The auth layer will return `invalid_user_on_card` if you try to auth with the card ([link]( https://github.com/votingworks/vxsuite/blob/6eb372e8874825c60dcbb157f0b5ff08a7e90f38/libs/auth/src/dipped_smart_card_auth.ts#L626-L628)) or that the card is unprogrammed in programming contexts ([link](https://github.com/votingworks/vxsuite/blob/6eb372e8874825c60dcbb157f0b5ff08a7e90f38/libs/auth/src/dipped_smart_card_auth.ts#L61-L80))

arsalansufi: This condition is really just detecting sys admin cards, which have no election key. The backend will never expose an old-style election manager or poll worker user to the frontend

Issue #5040, opened 2024-07-10
VxMark: Scan fails on mostly empty ballots


kshen0:
On the BMD 150 model I noticed that about half the time, immediately after issuing the `scan ticket` command, the state machine will receive a response `0x12` from the printer/scanner. This breaks the state machine because it expects a scan ticket response header starting with `IMG...`.

I'm not sure what is causing the `0x12` response, but this is similar to the issue we saw with the print buffer: https://votingworks.slack.com/archives/C03Q42V9MJ5/p1715296095534459

kshen0:
Confirmed this is the same issue that resurfaced because we handle it in `transferInAcknowledgement`, which is _not_ used by a few commands. We should instead handle it one level lower in `transferInGeneric`.

PR #5052, opened 2024-07-12
Handle `realtime status transmission` response in `transferInGeneric`

kshen0:
## Overview
Issue https://github.com/votingworks/vxsuite/issues/5040

See explanation of issue in https://github.com/votingworks/vxsuite/pull/4846

We used to handle unexpected data sent from paper handler to host in `transferInAcknowledgement`. That function is called by most, but not all, commands. It has 2 jobs:
1. read data from generic in buffer
2. parse and handles a simple success/fail response code
 
This worked because printing was always followed by a call to `setScanDirection`, which does call `transferInAcknowlegdement`.

Recently we removed the call to `setScanDirection`, so the next call after printing that reads from the generic in buffer is `scan`. This call does not expect an acknowledgement. Rather, it expects a 16-byte header and a subsequent large block of scan data. Therefore `scan` does not call `transferInAcknowledgement`.

Since `transferInAcknowledgement` is no longer called, the generic in buffer is never cleared of unexpected data. This causes scan parsing to fail.

To fix this I moved handling of unexpected data to `transferInGeneric`. That function is the closest to the metal we can get and is eventually called by all driver functions that want to read the generic in buffer.

## Demo Video or Screenshot

N/A

## Testing Plan
* reproduced issue on hardware, installed fix, and verified it fixed the problem
* rewrote tests for `transferInGeneric`

## Checklist

### DIFF
--- a/libs/custom-paper-handler/src/driver/driver.ts
+++ b/libs/custom-paper-handler/src/driver/driver.ts
@@ -426,7 +426,7 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
     const { data } = transferInResult;
     assert(data);
 
-    const result = AcknowledgementResponse.decode(Buffer.from(data.buffer));
+    const result = AcknowledgementResponse.decode(bufferFromDataView(data));
     if (result.isErr()) {
       debug(`Error decoding transferInGeneric response: ${result.err()}`);
       return false;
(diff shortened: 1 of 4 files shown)


eventualbuddha (approved):

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -110,13 +123,12 @@ export const REAL_TIME_ENDPOINT_IN = 3;
 export const REAL_TIME_ENDPOINT_OUT = 4;
 export const PACKET_SIZE = 65536;
 
+// The number of times transferInGeneric will retry after receiving unexpected data.
+const MAX_TRANSFER_IN_ATTEMPTS = 20;
```
kshen0, line 127: We've never seen the bad byte sequence twice in a row so 20 feels like a sufficiently high limit.

PR #5061, opened 2024-07-16
switch to pactl wrapper with sudo

amcmanus:
This PR uses a wrapper script for the `pactl` command with sudo permissions. This is necessary because `pulseaudio` runs unique instances per user. The `vx-ui` user is responsible for our production X session and accompanying audio configuration. To determine if headphones are present, the `vx-services` user needs to access the `vx-ui` session, and this approach makes that possible.


arsalansufi (approved):
:shipit:

PR #5062, opened 2024-07-16
[MarkScan] Implement pre-printed ballot insertion in PollWorkerScreen

kofi-q:
## Overview

Related to https://github.com/votingworks/vxsuite/issues/4812

Enabling support (currently behind the `MARK_SCAN_ENABLE_BALLOT_REINSERTION` feature flag) for re-inserting a printed ballot after a voter session has ended or timed out.

- Adding a new  Poll Worker screen section to bypass ballot style selection (pre-printed ballots will still be accepted after ballot style selection, though).
- Hooking up the new state machine states to poll-worker-facing screens.
- Updating load paper prompt

## Demo Video or Screenshot

(Full flow [here](https://www.figma.com/design/hvboQDWXqE1PbpGbWfb9iU/%5BVxMarkScan%5D-Printed-Ballot-Scan%2FRe-Scan-Flows?node-id=2-78&t=l4h4NmR73jzOW8Zz-1)):

### Ballot Insertion Flow:
Screenshot 2024-07-16 at 13 13 49

## Testing Plan
- New unit test cases
- Manual testing with mock paper handler (will need to verify fully on real VSAP later)

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/pages/inserted_preprinted_ballot_screen.tsx
+++ b/apps/mark-scan/frontend/src/pages/inserted_preprinted_ballot_screen.tsx
@@ -1,0 +1,72 @@
+import React from 'react';
+
+import { Button, Icons, P } from '@votingworks/ui';
+import { BallotStyleId, PrecinctId, VotesDict } from '@votingworks/types';
+import { assert } from '@votingworks/basics';
+
+import * as api from '../api';
+import { CenteredCardPageLayout } from '../components/centered_card_page_layout';
+
+export interface InsertedPreprintedBallotScreenProps {
+  activateCardlessVoterSession: (
+    precinctId: PrecinctId,
+    ballotStyleId: BallotStyleId
+  ) => void;
+  setVotes: (votes: VotesDict) => void;
+}
+
+export function InsertedPreprintedBallotScreen(
+  props: InsertedPreprintedBallotScreenProps
+): React.ReactNode {
+  const { activateCardlessVoterSession, setVotes } = props;
+
+  const startSessionWithPreprintedBallot =
+    api.startSessionWithPreprintedBallot.useMutation().mutate;
+
+  const returnPreprintedBallot =
+    api.returnPreprintedBallot.useMutation().mutate;
+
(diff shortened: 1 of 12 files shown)


apps/mark-scan/frontend/src/pages/inserted_preprinted_ballot_screen.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const interpretation = interpretationQuery.data;
+
+  const onPressReview = React.useCallback(() => {
+    assert(interpretation);
+    assert(interpretation.type === 'InterpretedBmdPage');
+
+    const { ballotStyleId, precinctId } = interpretation.metadata;
+
+    activateCardlessVoterSession(precinctId, ballotStyleId);
+    startSessionWithPreprintedBallot();
```
kshen0, line 39: It looks like the state machine can end cardless voter auth [here](https://github.com/votingworks/vxsuite/blob/main/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts#L619). Would it be cleaner to make the state machine responsible for all cardless voter auth changes in this flow?

kofi-q: Not a hardened opinion, but leaning toward this path in the short term -- the UI is current the only place where we activate sessions (granted, because it depends on information from users). So, this feels consistent with that, but would be open to larger change to move that logic as well.

I think even if we moved the `activateCardlessVoterSession` call to the state machine (assuming that's what were referring to), it would still make sense to keep the choice of `precinctId` and `ballotStyleId` here in the UI, at which point, the only change we'd be making would be to move the logic of calling out to the auth library from the backend API into the state machine (and my long-term bias is to separate the concerns of auth state, UI state, and machine state anyway, to make things easier to reason about).

kshen0: That sounds good - thanks for laying out the reasoning

kshen0 (approved):

Issue #5004, opened 2024-06-28
VxScan: Printing a test page in production crashes the app


arsalansufi:
Printing a test page in production crashes the app

Issue automatically created from Slack feedback reported by arsalan: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1719612688094639

You can access the "Print Test Page" button on the election manager screen under the "Printer" tab

carolinemodic:
repros in VxDev because it can't find the test page pdf, I assume its the same issue in prod but it might be a little different how it should work there

PR #5066, opened 2024-07-17
Fix something went wrong printing test page

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/5004

Fixes the something went wrong screen when printing a test page in production. There is a custom step in the pnpm build for scan-backend that copies the printed pdf to the build directory. This ensures that the backend is build through that command and the pdf gets copied over, making sure it is there when printing needs to occur. 

## Demo Video or Screenshot
N/A

## Testing Plan
Tested on VxDev 

## Checklist

### DIFF
--- a/apps/scan/frontend/Makefile
+++ b/apps/scan/frontend/Makefile
@@ -8,6 +8,7 @@ build: FORCE
 	pnpm install && \
 		pnpm --dir ../../../libs/ballot-interpreter build && \
 		pnpm --dir ../../../libs/pdi-scanner build && \
+		pnpm --dir ../backend build && \
 		pnpm build
 
 bootstrap: install build


apps/scan/frontend/Makefile
```diff
@@ -8,6 +8,7 @@ build: FORCE
 	pnpm install && \
 		pnpm --dir ../../../libs/ballot-interpreter build && \
 		pnpm --dir ../../../libs/pdi-scanner build && \
+		pnpm --dir ../backend build && \
```
arsalansufi, line 11: I am still a bit confused as to why this isn't getting run already. For `libs/ballot-interpreter` and `libs/pdi-scanner`, I understand that they're not referenced by source code as typical dependencies. We just use the Rust binaries that those libs build. So makes some sense that they get missed. For the `apps/scan/backend`, seems like we're definitely building it but maybe not using its `pnpm build` 🤔

arsalansufi (approved):
A classic 1-line prod environment fix haha - LGTM!

apps/scan/frontend/Makefile
carolinemodic: yeah we're definitely building it just...differently somehow lol

carolinemodic: I'm only 98% sure this fix will apply to prod but if I make images in the next day or two for the screen recorder I can double check this as well

PR #5067, opened 2024-07-17
refactor(fai-100-controller): let `clap` handle MarkScan workspace

eventualbuddha:
Rather than accessing the `MARK_SCAN_WORKSPACE` env var when we need it, read it upfront using `clap`. This has the additional effect of allowing it to be passed as a CLI arg instead: `--mark-scan-workspace /path/to/workspace`.

Also, changes `write_pat_connection_status` to improve readability by accepting the status directly rather than passing a boolean.


kshen0 (approved):
Thanks for making these changes! I agree the ergonomics are better.

Issue #5003, opened 2024-06-28
vxscan: app freeze on power down


mattroe:
https://votingworks.slack.com/archives/C062Y0ZFKFT/p1719600160010549

Issue #4904, opened 2024-05-30
App crashes (“Something Went Wrong”) after setting time from sys admin or election manager screen in production


arsalansufi:
While we saw this on a production VxMark, it likely affects all machines in production. We recently moved clock changes from kiosk-browser to app backends and, in the new code path, are likely either missing sudo calls or missing an entry in the sudoers file to allow password-less sudo calls.

Issue #4899, opened 2024-05-30
Restart button on "Something Went Wrong" page does nothing in production


arsalansufi:
While we saw this on a production VxMark, it likely affects all machines in production. We recently moved power-down/reboot functionality from kiosk-browser to app backends and, in the new code path, are either missing sudo calls or missing an entry in the sudoers file to allow password-less sudo calls.

PR #5070, opened 2024-07-17
Fix libs/backend system calls in prod

arsalansufi:
## Overview

Issue links: https://github.com/votingworks/vxsuite/issues/4904 + https://github.com/votingworks/vxsuite/issues/5003 + https://github.com/votingworks/vxsuite/issues/4899

When we moved the system-call API from `kiosk-browser` to `libs/backend`, we missed a few pieces of configuration for production, namely some `sudo` calls and the relevant `sudoers` file update to allow those `sudo` calls to be made without having to provide the `sudo` password.

The result has been that setting the clock and interacting with reboot buttons cause VxSuite apps to crash in production.

This PR, paired with https://github.com/votingworks/vxsuite-complete-system/pull/384, remedies this.

## Testing Plan

Reproduced the crashes on a production-imaged non-locked-down VxAdmin, then copied updated source code over, rebuilt, and verified that the errors were no more.

## Checklist

### DIFF
--- a/libs/backend/src/intermediate-scripts/set-clock
+++ b/libs/backend/src/intermediate-scripts/set-clock
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+usage () {
+    echo 'Usage: set-clock <iana-zone> <datetime-string>'
+    exit 1
+}
+
+if ! [[ "${#}" -eq 2 ]]; then
+    usage
+fi
+
+LOCALTIME_SYMLINK=`readlink /etc/localtime`
+TIMEZONE_DIRECTORY=/usr/share/zoneinfo
+
+# timedatectl sets timezone by symlinking /etc/localtime to a file in /usr/share/zoneinfo, but
+# because /etc is read-only on locked-down images, the setup-machine.sh script sets up a level
+# of indirection:
+#
+# /etc/localtime (read-only symlink) -->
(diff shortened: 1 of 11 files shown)


libs/backend/src/intermediate-scripts/set-clock
```diff
@@ -0,0 +1,33 @@
...
+# of indirection:
+#
+# /etc/localtime (read-only symlink) -->
+# /vx/config/localtime (writable symlink) -->
+# /usr/share/zoneinfo/...
+#
+# We account for that indirection, if relevant, here.
+#
+if [[ ! "${LOCALTIME_SYMLINK}" =~ "${TIMEZONE_DIRECTORY}".* ]]; then
+    ln -sf "${TIMEZONE_DIRECTORY}/${1}" "${LOCALTIME_SYMLINK}"
```
arsalansufi, line 28: This bit of logic (translated over from [vxsuite-complete-system/config/admin-functions/timedatectl](https://github.com/votingworks/vxsuite-complete-system/blob/14878282277ee2782b3379e10b39c067ad84d235/config/admin-functions/timedatectl)) was also lost in the transition from `kiosk-browser` to `libs/backend`.

The lack of this logic hasn't crashed our non-locked-down images, but I'm 99% sure that it will crash a locked-down image, so I'm getting ahead of it now.

adghayes (approved):
🫠

amcmanus (approved):

PR #5078, opened 2024-07-22
Mock Fujitsu Printer

adghayes:
## Overview

Adds mocking capabilities for the Fujitsu thermal printer used in our V4 hardware. I think this is important as we will have to iterate on the UX for the printer while not everyone has the hardware and some hardware states are difficult to reach. 

Uses the same file-based mock as auth and standard printers.

The dev dock is getting pretty big - I would be in favor of making the card selection a simple drop down to take up less space and, in general, shoot for a lo-fi UX. I want to be able to iterate on mocks - which might vary across products - without getting hung up on layout.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/b2880632-af3f-4c89-a07c-897a0681fcfb

## Testing Plan

Automated tests added and manually tested.

### DIFF
--- a/libs/dev-dock/frontend/src/api_client.tsx
+++ b/libs/dev-dock/frontend/src/api_client.tsx
@@ -1,0 +1,18 @@
+import * as grout from '@votingworks/grout';
+import type { Api } from '@votingworks/dev-dock-backend';
+import React from 'react';
+
+export type ApiClient = grout.Client<Api>;
+
+export const ApiClientContext = React.createContext<ApiClient | undefined>(
+  undefined
+);
+
+export function useApiClient(): ApiClient {
+  const apiClient = React.useContext(ApiClientContext);
+  // istanbul ignore next
+  if (!apiClient) {
+    throw new Error('ApiClientContext.Provider not found');
+  }
+  return apiClient;
+}
--- a/libs/dev-dock/frontend/src/dev_dock.tsx
+++ b/libs/dev-dock/frontend/src/dev_dock.tsx
@@ -9,7 +9,12 @@ import {
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
 import styled from 'styled-components';
 import * as grout from '@votingworks/grout';
-import { assert, assertDefined, uniqueBy } from '@votingworks/basics';
+import {
+  assert,
+  assertDefined,
+  throwIllegalValue,
+  uniqueBy,
+} from '@votingworks/basics';
 import type {
   Api,
   DevDockUserRole,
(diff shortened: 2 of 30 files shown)


libs/fujitsu-thermal-printer/src/mocks/file_printer.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  constructor(printPollingInterval = FILE_FUJITSU_PRINTER_POLLING_INTERVAL_MS) {
+    this.printPollingInterval = printPollingInterval;
+  }
+
+  getStatus(): Promise<PrinterStatus> {
+    return Promise.resolve(readFromMockFile());
+  }
+
+  async print(data: Uint8Array): Promise<PrintResult> {
```
adghayes, line 125: This is more elaborate than mocks usually should be - it actually sleeps and introduces real delay to similar printing. This could present issues testing in an integration setting in the future. HOWEVER I think that it's really valuable to have the pause because it allows the developer to change printer state mid-print, which is the only way to enter some software states.

libs/dev-dock/frontend/src/api_client.tsx
adghayes, line 1: Needed to extract this file to re-use in `fujitsu_printer_mock.tsx`.

libs/dev-dock/frontend/src/dev_dock.tsx
```diff
@@ -557,16 +570,16 @@ const Container = styled.div`
...
     /* Move the handle down a bit to compensate for sliding up extra to hide the
      * shadow */
     #handle {
-      top: 20px;
+      top: 60px;
     }
   }
   /* Slide the dock down when open */
-  transition: top 0.15s ease-out;
+  transition: all 0.15s ease-out;
```
adghayes, line 582: Now the height of the dev dock is variable, so we couldn't use translations based on fixed px values.

libs/dev-dock/frontend/src/dev_dock.tsx
```diff
@@ -660,9 +674,14 @@ function DevDock() {
...
             <ScreenshotControls containerRef={containerRef} />
-            {APPS_WITH_PRINTER.includes(machineType) && <PrinterMockControl />}
+            {printerType === 'standard' && <PrinterMockControl />}
           </Column>
         </Row>
+        {printerType === 'fujitsu' && (
+          <Row>
+            <FujitsuPrinterMockControl />
+          </Row>
+        )}
```
adghayes, line 684: It made more sense to break this out in a row because if it's in a column it takes up an obnoxious amount of horizontal space with the dropdown values.

libs/fujitsu-thermal-printer/src/mocks/file_printer.ts
```diff
@@ -0,0 +1,194 @@
...
+import {
+  FujitsuThermalPrinterInterface,
+  PrintResult,
+  PrinterStatus,
+} from '../types';
+
+export const MOCK_FUJITSU_PRINTER_STATE_FILENAME = 'state.json';
+export const MOCK_FUJITSU_PRINTER_OUTPUT_DIRNAME = 'prints';
+export const DEFAULT_MOCK_FUJITSU_PRINTER_DIR = '/tmp/mock-fujitsu-printer';
+export const DEV_MOCK_FUJITSU_PRINTER_DIR = join(
```
carolinemodic, line 23: Might be worth adding something to the Dev Dock readme with the basics of how to use this mock, and things like where the prints get saved to

carolinemodic (approved):

Issue #4967, opened 2024-06-12
VxScan seems to hang indefinitely on "Configuring VxScan from USB drive" if you have the wrong election package on your USB from the election manager card you inserted


arsalansufi:
VxScan seems to hang indefinitely on "Configuring VxScan from USB drive" if you have the wrong election package on your USB from the election manager card you inserted

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1718222351013889?thread_ts=1718222351.013889&cid=CJU9MSC6S

PR #5079, opened 2024-07-22
Trigger configuration from usb properly on EM screen

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/4967

The issue was that the configureMutation call was never being made previously if the usb was mounted BEFORE the election manager card was inserted. Causing it to hang forever here and never show the failure screens when the usb didn't have the right election package for the election manager card. This makes it so that its triggered properly if the usb is mounted either before or after the EM card is authenticated. 

For some reason the happy path when there was an election package on the card always worked properly even if the usb was previously mounted which I don't totally understand. 

## Testing Plan
Manually Tested
- Mount usb and then insert EM card - in success case and in error cases saw correct behavior 
- Insert EM card then mount usb - in success case and in error cases saw correct behavior 

## Checklist

### DIFF
--- a/apps/scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
+++ b/apps/scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
@@ -24,6 +24,7 @@ export function UnconfiguredElectionScreenWrapper(
   // TODO move watching for USB drive to configure to the backend
   useEffect(() => {
     if (
+      isElectionManagerAuth &&
       usbDriveStatusQuery.isSuccess &&
       usbDriveStatusQuery.data.status === 'mounted'
     ) {
@@ -34,6 +35,7 @@ export function UnconfiguredElectionScreenWrapper(
     configureMutation.mutate,
     usbDriveStatusQuery.isSuccess,
     usbDriveStatusQuery.data?.status,
+    isElectionManagerAuth,
   ]);
 
   const error = configureMutation.data?.err();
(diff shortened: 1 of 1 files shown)


apps/scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
```diff
@@ -26,13 +22,20 @@ export function UnconfiguredElectionScreenWrapper(
...
-    },
-  });
+  useEffect(() => {
+    if (
+      usbDriveStatusQuery.isSuccess &&
+      usbDriveStatusQuery.data.status === 'mounted'
+    ) {
+      configureMutation.mutate();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
kshen0, line 32: Which dep is missed?

kshen0 (commented):
Would this allow configuration without an election manager ever authorizing?

apps/scan/frontend/src/screens/unconfigured_election_screen_wrapper.tsx
carolinemodic: It thinks configureMutation even though I include configureMutation.mutate and if you try to include the entire mutation a different lint rule tells you to just include the .mutate function.

carolinemodic:
@kshen0 I'll add another auth check on this call just for defense in depth but no we aren't currently calling this when the election manager isn't authenticated, when the user is logged out app_root renders other screens that supersede this one. There are also checks to make sure the auth status is right in the backend call.

carolinemodic:
There is kind of duplicated logic right now around the unconfigured screens as this isElectionManager prop gets passed through to potentially render a "insert card" screen but app_root is already rendering that instead of this anyway

kshen0 (approved):

PR #5083, opened 2024-07-22
Support 13.25" paper

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/4924

Supports printing on 8x13.25" paper.

## Demo Video or Screenshot

IMG_5460

## Testing Plan
- printed on 13.25" paper on hardware, ensuring printable area extends past 11"

## Checklist

- [x] N/A I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.


libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -103,40 +103,40 @@ export const ORDERED_BMD_BALLOT_LAYOUTS: Readonly<
...
-      maxRows: 17,
+      maxRows: 18,
       hideParties: true,
       topMargin: '0.5625in',
     },
     {
       minContests: 86,
       fontSizePt: 5,
-      maxRows: 20,
+      maxRows: 21,
```
kshen0, line 139: Adding 2.25" printable space -> expect room for ~1 more row

libs/printing/src/render.tsx
```diff
@@ -180,7 +182,7 @@ export async function renderToPdf(
         height: inchesToText(
           height === Infinity
             ? Math.max(PAPER_DIMENSIONS.Letter.height, contentHeight)
-            : PAPER_DIMENSIONS.Letter.height
```
kshen0, line 183: I'm not sure if we meant to hard code this, but I believe this would have broken legal paper size (or any non-11" paper)

libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -103,40 +103,40 @@ export const ORDERED_BMD_BALLOT_LAYOUTS: Readonly<
...
     {
       minContests: 31,
       fontSizePt: 9,
-      maxRows: 11,
+      maxRows: 12,
       hideParties: true,
       topMargin: '0.5625in',
     },
     {
       minContests: 46,
```
kshen0, line 116: Not increasing `minContests` for any layouts - see reasoning in [Slack](https://votingworks.slack.com/archives/C03Q42V9MJ5/p1721712074225429)

libs/printing/src/render.tsx
```diff
@@ -115,8 +115,7 @@ export async function renderToPdf(
         (width - horizontalMargin) * PLAYWRIGHT_PIXELS_PER_INCH
       ),
       height: Math.floor(
-        (PAPER_DIMENSIONS.Letter.height - verticalMargin) *
-          PLAYWRIGHT_PIXELS_PER_INCH
+        (height - verticalMargin) * PLAYWRIGHT_PIXELS_PER_INCH
```
jonahkagan: Just was looking at the context for this change and noticed that L113 has a misspelling of "noninteger" as "nonintenger." Not related to this change, but maybe fix while we're here looking at it

libs/printing/src/render.tsx
jonahkagan: Seems like the right change to me

libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -174,7 +174,7 @@ const Ballot = styled.div<BallotProps>`
         /* istanbul ignore next - hardware specs still in flux */
         case 'bmd150':
           /* istanbul ignore next - hardware specs still in flux */
-          return '7.975in 11in';
+          return '7.975in 13.25in';
```
jonahkagan, line 177: Out of curiosity, why do we use 7.975in instead of 8in for the height here and in the printing dimensions?

libs/printing/src/render.tsx
```diff
@@ -177,7 +180,9 @@ export async function renderToPdf(
         path: outputPath,
         width: inchesToText(width),
         height: inchesToText(
-          height === Infinity ? Math.max(height, contentHeight) : height
+          height === Infinity
+            ? Math.max(PAPER_DIMENSIONS.Letter.height, contentHeight)
+            : height
```
jonahkagan, line 185: Maybe replace this with `Math.max(viewportHeight, contentHeight)` now that you've handled the `height === Infinity` case above when defining `viewportHeight`

libs/printing/src/render.tsx
```diff
@@ -109,13 +109,16 @@ export async function renderToPdf(
     // be in the PDF, which allows us to determine the necessary height to fit
     // the page to the content. viewport height here is irrelevant, but we have to
     // set something.
+    const viewportHeight =
+      // Viewport height can't be infinity
+      height === Infinity ? PAPER_DIMENSIONS.Letter.height : height;
```
jonahkagan, line 114: When is `height` set to `Infinity` and why? It seems weird that libs/printing decides internally to use letter-sized paper in that case.

jonahkagan (approved):

libs/ui/src/bmd_paper_ballot.tsx
kshen0: 8in results in a width of exactly 1600 dots. The math is correct but for some reason 1600 dots overflows when printing on actual hardware. The extra dots overflow to the next line and cause a mostly blank line to be printed, resulting in stripes in the printed document.

I haven't had a chance to find the root cause yet unfortunately.

jonahkagan: Weird! Maybe leave a comment explaining this issue?

libs/printing/src/render.tsx
kshen0: `height` is `Infinity` when printing with `LetterRoll` size paper. I'm not very familiar with our uses of letter roll paper but I wanted to preserve the behavior.

PR #5084, opened 2024-07-23
mock paper handler driver: give mock ImageData height

kshen0:
## Overview

MarkScan run with a mock paper handler was crashing because the scan data returned had no height. The mock ImageData returned by the mock driver was given a width of 2 and no height; height is computed by dividing the length of pixel data by width, so we can either provide 4 total pixels' worth of data or explicitly set width (and height, while we're at it) to 1.

## Demo Video or Screenshot

No user facing change

## Testing Plan

Manually reproduced issue and verified fix.


carolinemodic (approved):
tested locally and this fixes my issue!

PR #5090, opened 2024-07-23
Ignore Pat Connection on Pollworker Screen

carolinemodic:
## Overview
Currently if you insert the pat device while on the pollworker screen, or insert a PW card with a pat device connected you end up on a white screen. This PR fixes that and handles those cases to just go to the pollworker screen as expected. 

## Testing Plan
Connected/Disconnected PAT at various moments in PW screen flows and saw no impact, saw PAT test flow come up once voter session started (if still plugged in) as expected
Inserted PW card again once in PAT connection flow and saw PW screen display as expected. 

## Checklist

### DIFF
--- a/apps/mark-scan/backend/src/app.diagnostics.test.ts
+++ b/apps/mark-scan/backend/src/app.diagnostics.test.ts
@@ -225,6 +225,7 @@ test('saving the readiness report', async () => {
   const exportPath = exportResult.ok()![0];
   await expect(exportPath).toMatchPdfSnapshot({
     customSnapshotIdentifier: 'readiness-report',
+    failureThreshold: 0.001,
   });
 
   const pdfContents = await pdfToText(exportPath);
--- a/apps/mark-scan/backend/src/app.test.ts
+++ b/apps/mark-scan/backend/src/app.test.ts
@@ -50,7 +50,7 @@ import {
 import { PatConnectionStatusReader } from './pat-input/connection_status_reader';
 import { createWorkspace } from './util/workspace';
 
-const TEST_POLLING_INTERVAL_MS = 5;
+const TEST_POLLING_INTERVAL_MS = 15;
 
 jest.mock('./pat-input/connection_status_reader');
 
--- a/apps/mark-scan/backend/src/app.test.ts
+++ b/apps/mark-scan/backend/src/app.test.ts
@@ -45,7 +45,6 @@ import { ElectionState } from './types';
 import {
   mockElectionManagerAuth,
   mockPollWorkerAuth,
-  mockCardlessVoterAuth,
 } from '../test/auth_helpers';
 import { PatConnectionStatusReader } from './pat-input/connection_status_reader';
 import { createWorkspace } from './util/workspace';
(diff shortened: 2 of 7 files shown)


apps/mark-scan/backend/src/app.diagnostics.test.ts
```diff
@@ -225,6 +225,7 @@ test('saving the readiness report', async () => {
   const exportPath = exportResult.ok()![0];
   await expect(exportPath).toMatchPdfSnapshot({
     customSnapshotIdentifier: 'readiness-report',
+    failureThreshold: 0.001,
```
carolinemodic, line 228: I have some snapshot tests fail locally because of a 1 pixel difference on main, this fixes that

apps/mark-scan/backend/src/app.test.ts
```diff
@@ -45,11 +45,12 @@ import { ElectionState } from './types';
 import {
   mockElectionManagerAuth,
   mockPollWorkerAuth,
+  mockCardlessVoterAuth,
 } from '../test/auth_helpers';
 import { PatConnectionStatusReader } from './pat-input/connection_status_reader';
 import { createWorkspace } from './util/workspace';
 
-const TEST_POLLING_INTERVAL_MS = 5;
+const TEST_POLLING_INTERVAL_MS = 15;
```
carolinemodic, line 53: I'm not really sure why I needed this but it fixed my issue with the mockLoadAndPrint function failing looking for 'presenting_ballot' instead of 'scanning'

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -360,6 +362,16 @@ function buildPatDeviceConnectionStatusObservable() {
     return timer(0, devicePollingIntervalMs).pipe(
       switchMap(async () => {
         try {
+          // We only try to connect a PAT device if we are currently in a voter session.
+          const authStatus = await auth.getAuthStatus(
+            constructAuthMachineState(workspace)
+          );
```
kshen0, line 368: Is it possible to accomplish this with a [guarded transition](https://stately.ai/docs/xstate-v4/xstate/transitions-and-choices/guards) where we handle `PAT_DEVICE_(DIS)CONNECTED` ([src](https://github.com/votingworks/vxsuite/blob/main/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts#L490-L497))?

This would mean we don't have to make the observable aware of `auth` and keeps the division of responsibility between the observable as an event creator and state machine as an event consumer.

carolinemodic:
@kshen0 Back to you on this. See https://votingworks.slack.com/archives/CJU9MSC6S/p1721936151123989 for a related bug found in testing of this. Fixing that might involve changing this a bit, maybe we should only poll for pat device in certain states? But for the purposes of the UAT code freeze this is a more likely scenario to occur so I think we should just get this in for now.

kshen0:
> maybe we should only poll for pat device in certain states

Agreed this is the direction we should go after UAT but for now this LGTM

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -502,6 +504,7 @@ export function buildMachine(
             isPatDeviceConnected: false,
           }),
           target: 'pat_device_disconnected',
+          cond: (context) => context.isInVoterSession,
```
kshen0, line 507: I was thinking through whether we can avoid this small duplication of logic and I don't think we can in the current setup. I was imagining we'd check `auth` directly but I had forgotten that guards can't await promises: https://github.com/statelyai/xstate/issues/533

The post says the recommended workaround if you need to evaluate a promise as part of your guard is to use an intermediate state to contain the guard logic. This won't work in our case because the PAT events will continually fire and we'll ping pong between states constantly.

Sorry for not thinking of this rough edge before suggesting the change - given our options, this small duplication of logic is an ok tradeoff vs the larger task of restructuring when to invoke PAT polling.

kshen0 (approved):

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
carolinemodic: Yeah I did this workaround since I couldn't await in a cond.

PR #5093, opened 2024-07-23
fix(basics): improve types of conditional `iter` methods

eventualbuddha:
## Overview
When it only makes sense to call certain `iter` methods based on the contained type `T`, the better approach is to use an explicit type for `this` in the method definition rather than a ternary type resulting in `unknown` when the type doesn't match. This way actually gives us TypeScript errors and is clearer.

## Demo Video or Screenshot
image

## Testing Plan
Automated testing, including new tests to ensure the error is triggered appropriately.


jonahkagan (approved):
Nice! I was wondering about this issue once, seems like a solid solution

PR #5096, opened 2024-07-24
feat(db): allow using prepared statements

eventualbuddha:
For situations where we insert many records over and over (i.e. CVRs) this could speed things up a lot.

### DIFF
--- a/apps/admin/backend/src/store.ts
+++ b/apps/admin/backend/src/store.ts
@@ -11,7 +11,7 @@ import {
   isResult,
   assert,
 } from '@votingworks/basics';
-import { Bindable, Client as DbClient } from '@votingworks/db';
+import { Bindable, Client as DbClient, Statement } from '@votingworks/db';
 import {
   AnyContest,
   BallotId,
@@ -2347,23 +2347,18 @@ export class Store {
     }
 
     if (writeInCandidateIds.length > 0) {
-      const params: Bindable[] = [];
-      const questionMarks: string[] = [];
-      for (const writeInCandidateId of writeInCandidateIds) {
-        params.push(manualResultsRecordId, writeInCandidateId);
-        questionMarks.push('(?, ?)');
-      }
-
       // insert new write-in candidate references
-      this.client.run(
-        `
-          insert into manual_result_write_in_candidate_references (
-            manual_result_id,
-            write_in_candidate_id
-          ) values ${questionMarks.join(', ')}
-        `,
-        ...params
-      );
+      const statement: Statement<
+        [typeof manualResultsRecordId, (typeof writeInCandidateIds)[number]]
+      > = this.client.prepare(`
+        insert into manual_result_write_in_candidate_references (
+          manual_result_id,
+          write_in_candidate_id
+        ) values (?, ?)`);
+
+      for (const writeInCandidateId of writeInCandidateIds) {
+        this.client.run(statement, manualResultsRecordId, writeInCandidateId);
+      }
     }
 
     // delete write-in candidates that may have only been included on the
--- a/apps/central-scan/backend/src/store.ts
+++ b/apps/central-scan/backend/src/store.ts
@@ -644,7 +644,7 @@ export class Store {
         back.imagePath
       );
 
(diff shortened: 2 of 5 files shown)


apps/admin/backend/src/store.ts
```diff
@@ -2347,23 +2347,18 @@ export class Store {
...
-      const questionMarks: string[] = [];
+      // insert new write-in candidate references
+      const statement: Statement<
+        [typeof manualResultsRecordId, (typeof writeInCandidateIds)[number]]
+      > = this.client.prepare(`
+        insert into manual_result_write_in_candidate_references (
+          manual_result_id,
+          write_in_candidate_id
+        ) values (?, ?)`);
+
```
adghayes, line 2358: I don't understand why this would result in a performance improvement. I thought that preparing statements would be cheaper than actually running the query, and this change goes from (1 prepare + 1 query) to (1 prepare & n queries). 

The new way certainly reads better though.

adghayes (approved):
In the initial use case I don't understand why this would be a performance gain, but I see how it could be in CVR import use case and I think we should try it. If I can carve out a little time next week I will.

apps/admin/backend/src/store.ts
eventualbuddha: I'm okay leaving this out. I mostly wanted to make _something_ use it as a demonstration. I'm guessing the performance difference in this case is negligible since the N is likely quite low. I can do some microbenchmarking to see if I can parse out any signal in the noise.

eventualbuddha: Looks like the "one big SQL string" approach probably is faster here, though the absolute time difference is fairly negligible even up to N=10,000. As I expected, the "prepared statement run repeatedly" approach is faster than "repeatedly parsing SQL" approach. Values in the right three columns are operations per second:

| N | prepared statement run repeatedly | one big sql string | repeatedly parsing sql |
| -- | -- | -- | -- |
| 1 | 181,402 | 154,925 | 305,350 |
| 10 | 57,116 | 64,977 | 33,039 |
| 100 | 8,709 | 9,891 | 3,135 |
| 1,000 | 778 | 1,056 | 316 |
| 10,000 | 85 | 104 | 28 |

<summary>
Click here for benchmark file
<details>

```ts
import { Bench } from 'tinybench';
import assert from 'assert';
import { Store } from './store.js';

const bench = new Bench({ time: 100 });
const store = Store.memoryStore();
const client = store['client'];
const N = process.argv[2] ? parseInt(process.argv[2], 10) : 1000;

assert(N > 0);

client.exec('create table people (first_name text, last_name text, age int)');

bench
  .add(`prepared statement run repeatedly (N=${N})`, () => {
    const statement = client.prepare('insert into people values (?, ?, ?)');

    for (let i = 0; i < N; i += 1) {
      client.run(statement, `first${i}`, `last${i}`, i);
    }

    client.run('delete from people');
  })
  .add(`one big SQL string (N=${N})`, () => {
    const sql = `insert into people values ${Array.from(
      { length: N },
      () => `(?, ?, ?)`
    ).join(', ')}`;
    const values = [];

    for (let i = 0; i < N; i += 1) {
      values.push(`first${i}`, `last${i}`, i);
    }
    client.run(sql, ...values);
    client.run('delete from people');
  })
  .add(`repeatedly parsing SQL string (N=${N})`, () => {
    for (let i = 0; i < N; i += 1) {
      client.run(
        'insert into people values (?, ?, ?)',
        `first${i}`,
        `last${i}`,
        i
      );
    }
  });

await bench.warmup();
await bench.run();

console.log(bench.table());
```

</details>
</summary>

eventualbuddha: Opted to keep this one as-is because it should only cost a handful of `ms` even with 10,000 records and I don't think we'll ever get near that number.

adghayes: I think that's the right call - it's more readable and the performance impact is negligible, I just wanted to confirm my understanding. The test results are promising for the CVR case, where the "one big sql string" approach is not an option.

PR #5097, opened 2024-07-24
fix(pdi-scanner): use a larger buffer when receiving

eventualbuddha:
The PDI scanner doesn't seem to stream as it receives image data from its sensors, so it sends the data all in one go if it can. This reduces the number of incoming transfers for the image data to just one, which should reduce the odds of any image corruption that might happen if e.g. a packet is dropped. It might also make it a tiny bit faster, but I doubt the effect would be noticeable.

This change was suggested by PDI's engineering team. See the email thread from @arsalansufi dated 2024-07-15. See [this Slack thread](https://votingworks.slack.com/archives/CEL6D3GAD/p1721854880450379) for exploration related to this.

### DIFF
--- a/libs/pdi-scanner/src/rust/scanner.rs
+++ b/libs/pdi-scanner/src/rust/scanner.rs
@@ -72,7 +72,17 @@ impl Scanner {
         /// receive image data.
         const ENDPOINT_IN_IMAGE_DATA: u8 = 0x86;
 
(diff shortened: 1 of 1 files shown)


jonahkagan (approved):
Promising! 🤞

arsalansufi (approved):
Nice investigation!! Fingers crossed 🤞

PR #5104, opened 2024-07-25
dev-dock: Delay before screenshotting to ensure dev dock is hidden

jonahkagan:
## Overview

I noticed that most of the time, the dev dock wasn't actually hidden fast enough to be excluded from the screenshot, so just added a little delay.

## Testing Plan
Manually tested

## Checklist


arsalansufi (approved):
I recently ran into this too and hacked my way around the issue - thanks for fixing!

PR #5107, opened 2024-07-25
chore: remove `converter-nh-accuvote`

eventualbuddha:
There has been a decent amount of work done on this library on `v3.1.0-release-branch`, and it's fairly out of sync with `main`. To avoid confusion, we're removing it from `main` for the time being. We'll re-evaluate with v4.


apps/scan/backend/package.json
```diff
@@ -42,7 +42,6 @@
     "@votingworks/ballot-interpreter": "workspace:*",
     "@votingworks/basics": "workspace:*",
     "@votingworks/bmd-ballot-fixtures": "workspace:*",
-    "@votingworks/converter-nh-accuvote": "workspace:*",
```
arsalansufi, line 45: Did the VxScan backend not actually depend on anything from the converter module?

libs/cvr-fixture-generator/package.json
```diff
@@ -26,7 +26,6 @@
     "@votingworks/auth": "workspace:*",
     "@votingworks/backend": "workspace:*",
     "@votingworks/basics": "workspace:*",
-    "@votingworks/converter-nh-accuvote": "workspace:*",
```
arsalansufi, line 29: Same question here as above but for `libs/cvr-fixture-generator`

arsalansufi (approved):
Double checking two things, otherwise LGTM! 🧹

apps/scan/backend/package.json
eventualbuddha: I believe it used to offer the conversion in the app before the backends diverged, but the feature was removed. We just missed the dependency.

libs/cvr-fixture-generator/package.json
eventualbuddha: We used to generate ballot page layouts in `converter-nh-accuvote` (see 5d64f735cf30efcb5e31e01a6f75058944be8560). It looks like we then removed the reference in 88836d906c2de65b149b78da28a087bc34897052.

PR #5119, opened 2024-07-26
Use PID file for controller connection status

kshen0:
## Overview

We currently check for controller connection by running `child_process.execFile('ps aux | grep controller[d])`. `execFile` doesn't support I/O redirect so this command will always fail and report no controller connection.

~`exec` behaves similarly but supports I/O redirect at the slight performance cost of starting (and ending) a shell session.~

Based on feedback, this PR changes controller daemons to write a PID file to be read by the mark-scan backend. The reported PID is checked with `kill` signal 0 to see if the daemon process is still running.

## Demo Video or Screenshot
IMG_5468

## Testing Plan

Manual testing
- Ran daemon, confirmed backend can detect controller
- Killed daemon, confirmed backend can no longer detect controller
- Automated tests

## Checklist
N/A not user facing, no new actions

### DIFF
--- a/apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
+++ b/apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
@@ -37,7 +37,7 @@ const APP_NAME: &str = "vx-mark-scan-fai-100-controller-daemon";
 const FAI_100_VID: u16 = 0x28cd;
 const FAI_100_PID: u16 = 0x4002;
 const POLL_INTERVAL: Duration = Duration::from_millis(100);
-const EVENT_LOOP_LOG_INTERVAL: Duration = Duration::from_secs(1);
+const EVENT_LOOP_LOG_INTERVAL: Duration = Duration::from_secs(5 * 60); // 5 mins; from_mins is a nightly build feature
 const BUFFER_MAX_BYTES: usize = 64;
 const PAT_CONNECTION_STATUS_FILENAME: &str = "_pat_connection.status";
 use std::io::Write;
--- a/apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
+++ b/apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
@@ -11,10 +11,10 @@ use commands::{
     ButtonSignal, CommandError, NotificationStatusResponse, SipAndPuffDeviceStatus,
     SipAndPuffSignalStatus, VersionResponse, RESPONSE_BYTE_LENGTH,
 };
-use daemon_utils::run_no_op_event_loop;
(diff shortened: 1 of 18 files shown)


apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
```diff
@@ -37,7 +37,7 @@ const APP_NAME: &str = "vx-mark-scan-fai-100-controller-daemon";
 const FAI_100_VID: u16 = 0x28cd;
 const FAI_100_PID: u16 = 0x4002;
 const POLL_INTERVAL: Duration = Duration::from_millis(100);
-const EVENT_LOOP_LOG_INTERVAL: Duration = Duration::from_secs(1);
+const EVENT_LOOP_LOG_INTERVAL: Duration = Duration::from_secs(5 * 60); // 5 mins; from_mins is a nightly build feature
```
kshen0, line 40: Unrelated but in debugging this I found the logs to be spammy. Now that active development on the FAI 100 is coming to a close, we can reduce the frequency of this log for production.

jonahkagan (approved):
Unrelated but noticed from the screenshot, "Connected" might be a bit more specific status text than "Detected"

kshen0:
> Unrelated but noticed from the screenshot, "Connected" might be a bit more specific status text than "Detected"

Thanks, will update.

@eventualbuddha suggested writing and checking a PID file instead of our current approach. I'll make that change as well and re-request review.

kshen0:
> Unrelated but noticed from the screenshot, "Connected" might be a bit more specific status text than "Detected"

https://github.com/votingworks/vxsuite/pull/5119/files#diff-232d82be01734fa5fa4cf4b2be793a8110e95c2b36f9bc84b2f520762e59fc43R52

libs/ui/src/diagnostics/mark_scan_readiness_report.tsx
```diff
@@ -49,6 +49,7 @@ export function MarkScanReadinessReportContents(
         {...accessibleControllerProps}
         diagnosticType="mark-scan-accessible-controller"
         title={DiagnosticSectionTitle.AccessibleController}
+        detectedLabel="Connected"
```
jonahkagan, line 52: Why not have them all say "Connected"?

kshen0: It felt wrong for the PAT input because there's a difference between the jack being detected and a sip & puff device being connected. But for everything else that's supposed to be permanently connected, using "Connected" makes sense. Thinking about it now, I agree for simplicity we should just go with "Connected" for all.

eventualbuddha (approved):

PR #5122, opened 2024-07-29
Election package hash

### LINKED ISSUE #4912: Separate election hash and ballot hash

jonahkagan:
## Overview

Closes #4912 

Adds a hash of the entire election package to help users detect configuration mismatches. While the ballot hash prevents tabulation errors due to misconfigured machines, it doesn't prevent errors due to other aspects of the election package (e.g. system settings) not getting propagated to all machines.

As a starting point, we simply display the election package hash alongside the ballot hash, allowing users to manually compare the hash to ensure all machines have the same election package. The hashes are concatenated with a hyphen (and shortened to 7 chars each, similar to Github commit hashes).

## Demo Video or Screenshot

Screenshot-VxAdmin-2024-07-25T00:12:39 815Z
Screenshot-VxCentralScan-2024-07-25T23:53:58 972Z
Screenshot-VxMark-2024-07-25T22:09:50 180Z
Screenshot-VxScan-2024-07-26T00:23:58 954Z

## Testing Plan
- Updated existing tests and added some tests where needed
- Manually tested

## Checklist


libs/test-utils/src/zip.ts
```diff
@@ -1,43 +1,17 @@
...
-    void Object.entries(files)
-      .reduce(
-        (last, [name, data]) => last.then(() => addFile(zip, name, data)),
-        Promise.resolve()
-      )
-      .then(() => {
-        zip.finalize();
-      });
-  });
+  const zip = new JsZip();
```
carolinemodic, line 7: nice this seems simpler

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -67,7 +67,9 @@ export interface ScannerStoreBase {
   clearCastVoteRecordHashes(): void;
   getBatches(): BatchInfo[];
   getCastVoteRecordRootHash(): string;
-  getElectionDefinition(): ElectionDefinition | undefined;
+  getElectionRecord():
+    | { electionDefinition: ElectionDefinition; electionPackageHash: string }
```
carolinemodic, line 71: why not have some kind of ElectionDefinitionWithHash interface for this type you are using around?

apps/design/backend/src/worker/generate_election_package.ts
```diff
@@ -111,13 +111,18 @@ export async function generateElectionPackage(
...
+  const zipContents = await zip.generateAsync({
+    type: 'nodebuffer',
+    streamFiles: true,
+  });
+  const electionPackageHash = sha256(zipContents);
+  const combinedHash = formatElectionHashes(
+    electionDefinition.ballotHash,
+    electionPackageHash
   );
+  const fileName = `election-package-${combinedHash}.zip`;
```
carolinemodic, line 123: nice this is helpful

carolinemodic (approved):
Obivously out of scope for this PR but those screenshots make me realize how out of date the central scan UI is and how we should maybe standerdize the footer UI. 

Excited about this, will definitely be much easier for us to understand whats happening and manage the IDs. Reviewing this is also making me think a bit about how we explain this to people, I imagine its going to be a little tricky and confusing for people to grasp the two different sides of the ID. Don't have any great ideas on how to make things clearer though so I suppose we just have to do our best in training/docs.

libs/backend/src/cast_vote_records/export.ts
jonahkagan: Yeah, I considered doing that but wasn't sure how often I would need it. I think in retrospect it's worthwhile, I'll add it

Issue #5033, opened 2024-07-10
VxMark: Language picker usability improvements


mattroe:
From https://votingworks.slack.com/archives/C03Q42V9MJ5/p1720209700026399:
- Language icon should be changed to a globe
- When the language button is focused, the audio should say "Would you like to change the language? Current language English"

From https://votingworks.slack.com/archives/C03Q42V9MJ5/p1720209700026399:
- Make sure touch screen clicks properly focus language options so that they're read out

carolinemodic:
On the last point in the description, i just tested a bit and looked at the code and we do read out "Selected: Language" on each language when you select it. It seems like there can be a race condition where it gets cut off sometimes if it starts to read the button before it renders with the "selected: " prefix added? I'm not entirely sure if thats what is happening but that is my guess, although we don't seem to have any problem when doing the same thing for candidate buttons on a candidate contest page

PR #5123, opened 2024-07-29
[fix/5033] Fix inconsistent audio feedback on touch events

kofi-q:
## Overview

Related to https://github.com/votingworks/vxsuite/issues/5033

We noticed in manual tests that touch-based "click" events don't trigger audio in the same way that "click" events with the accessible controller do.

This change fixes our handling of touch events in the `<Button>` component by triggering an actual `click` event on the button when we detect a touch-based button press. This allows the click/focus-based screen reader to properly detect button presses.

__Context:__ 
Our `<Button>` component overrides the default browser handling of touch taps by intercepting `touchStart` and `touchEnd` and treating the pair as a button press even if the user's finger moved slightly in-between. It then triggers the `onPress` callback directly and blocks the default browser handling, which prevents any downstream `click` listeners from being triggered. Not entirely sure if the custom touch detection logic is necessary, but leaving it intact, assuming it was added to solve a specific issue.

### Before:

https://github.com/user-attachments/assets/bbd9a1a7-636f-41a8-9a89-27dd91ffac18

### After:

https://github.com/user-attachments/assets/3c95abdf-9361-4b62-9867-e9617e167e50

## Testing Plan
- Manual (+ existing tests as regression checks)

## Checklist

### DIFF
--- a/libs/ui/src/button.tsx
+++ b/libs/ui/src/button.tsx
@@ -3,7 +3,7 @@ import React, { PureComponent } from 'react';
 import styled, { css, CSSObject, DefaultTheme } from 'styled-components';
 import { SizeMode, SizeTheme, UiTheme } from '@votingworks/types';
 
-import { throwIllegalValue } from '@votingworks/basics';
+import { assertDefined, throwIllegalValue } from '@votingworks/basics';
 import { rgba } from 'polished';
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
LGTM! Was passing by and this was small enough figured I might as well approve haha

PR #5125, opened 2024-07-29
Convert cast vote record optimization env vars to system settings

### LINKED ISSUE #4234: Switch from environment variables to system settings for the CAST_VOTE_RECORD_OPTIMIZATION_EXCLUDE_REDUNDANT_METADATA and CAST_VOTE_RECORD_OPTIMIZATION_EXCLUDE_ORIGINAL_SNAPSHOTS feature flags
https://github.com/usnistgov/CastVoteRecords/issues/45

jonahkagan:
## Overview

Closes #4234 

Switch the cast vote record optimization environment variables to system settings.

## Demo Video or Screenshot
N/A

## Testing Plan
Updated existing tests
## Checklist


libs/types/src/system_settings.ts
```diff
@@ -64,6 +64,16 @@ export interface SystemSettings {
...
+  /**
+   * Excludes original snapshots in cast vote record reports, decreasing export size and
+   * import/export time
+   */
+  readonly castVoteRecordOptimizationExcludeOriginalSnapshots?: boolean;
+  /**
+   * Excludes redundant metadata in cast vote record reports, decreasing export size and
+   * import/export time
+   */
+  readonly castVoteRecordOptimizationExcludeRedundantMetadata?: boolean;
```
arsalansufi, line 76: Can we make it such that, if you don't include these fields in the `systemSettings.json`, we default to the behavior that we want in the overwhelming majority of cases (i.e. excluding this data)? (We really only want to include this data during cert and, even then, for just a tiny part of cert.)

Could do that by flipping these booleans, e.g., `castVoteRecordOptimizationIncludeOriginalSnapshots` and `castVoteRecordOptimizationIncludeRedundantMetadata`

arsalansufi (commented):
One quick thought!

libs/types/src/system_settings.ts
arsalansufi: I guess `Optimization` doesn't really make sense in the names anymore if you flip haha

jonahkagan: Would it be worth also just collapsing them into one flag, like `castVoteRecordStrictCdf` or something?

arsalansufi: Yeah I like that!

arsalansufi: Actually, on second thought, let's keep them separate. Just to have the flexibility. There may be cases where a customer wants original snapshots (for auditing purposes) but not the redundant metadata.

It's also not clear to me reading the CVR CDF spec whether or not original snapshots are required or optional.

libs/types/src/system_settings.ts
```diff
@@ -65,15 +65,15 @@ export interface SystemSettings {
   readonly precinctScanDisallowCastingOvervotes: boolean;
   readonly precinctScanEnableShoeshineMode?: boolean;
   /**
-   * Excludes original snapshots in cast vote record reports, decreasing export size and
-   * import/export time
+   * Includes original snapshots in cast vote record reports, decreasing export size and
```
arsalansufi, line 68: > ~decreasing~ increasing export size

I think you got this below but just missed here

arsalansufi (approved):
LGTM!

Issue #4953, opened 2024-06-12
VxAdmin: Strict CDF


jonahkagan:
- VxAdmin should support importing a strict CDF election definition, meaning it will need to fill in reasonable defaults or turn off features that aren't supported in strict CDF. It also needs to support exporting a strict CDF election definition.
- We should also convert VxDesign to output VXF election definitions instead of our extended CDF that we've been using up til now. We'll need to figure out another place to store election content translations.
- Switch from an environment variable to a system setting for the [REACT_APP_VX_CAST_VOTE_RECORD_OPTIMIZATION_EXCLUDE_REDUNDANT_METADATA feature flag](https://github.com/votingworks/vxsuite/blob/31e23bc76e7f5905efed4c462299499b206ee880/libs/utils/src/environment_variable.ts#L41-L43), which is technically necessary for strict CDF compliance. (See https://github.com/usnistgov/CastVoteRecords/issues/45.)

PR #5129, opened 2024-07-30
Strict ballot definition CDF

jonahkagan:
## Overview
Closes: #4953 

Updates our support for the ballot definition CDF to be strictly compatible with NIST's CDF by removing custom extension fields that we added. Instead, we either use a reasonable default or disable the behavior that was supported by each custom field (details in commits).

## Demo Video or Screenshot
VxAdmin and VxMarkScan with no seal
Screenshot-VxAdmin-2024-07-30T20:07:39 211Z
Screenshot-VxMark-2024-07-30T20:09:23 087Z

## Testing Plan
Updated existing tests, manually tested

## Checklist

### DIFF
--- a/libs/types/src/cdf/ballot-definition/convert.ts
+++ b/libs/types/src/cdf/ballot-definition/convert.ts
@@ -536,7 +536,6 @@ export function convertVxfElectionToCdfBallotDefinition(
           {
             '@type': 'BallotDefinition.PhysicalContest',
             BallotFormatId: 'ballot-format',
-            vxOptionBoundsFromTargetMark: gridLayout.optionBoundsFromTargetMark,
             PhysicalContestOption: gridLayout.gridPositions
               .filter((position) => position.contestId === contest.id)
               .map(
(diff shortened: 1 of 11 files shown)


libs/ui/src/seal.tsx
```diff
@@ -38,7 +38,9 @@ export function Seal({
   maxWidth,
   style,
   inverse,
-}: SealProps): JSX.Element {
+}: SealProps): JSX.Element | null {
```
carolinemodic, line 41: Are we sure a Seal is only used in this component?

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -1097,11 +1096,16 @@ export function convertCdfBallotDefinitionToVxfElection(
...
       ).map((ballotStyle): Vxf.GridLayout => {
         const orderedContests = assertDefined(ballotStyle.OrderedContent);
-        const optionBoundsFromTargetMark =
-          orderedContests[0].Physical[0].vxOptionBoundsFromTargetMark;
         return {
           ballotStyleId: ballotStyle.ExternalIdentifier[0].Value,
-          optionBoundsFromTargetMark,
+          // Since there's no CDF field for this, we set a default based on what
+          // generally works well for our HMPBs.
+          optionBoundsFromTargetMark: {
```
carolinemodic, line 1106: slight preference to define a const for these values

carolinemodic (approved):

libs/ui/src/seal.tsx
jonahkagan: Good question - I just searched through the repo and the answer is yes!

PR #5132, opened 2024-07-31
perf(scan): optimize writing of scanned images to disk

eventualbuddha:
## Overview

Previously we were writing both front & back images to disk twice: once before interpretation (as a kind of staging area, I guess?) and once after. This removes the first write (and the subsequent read+decode). In addition, this ensures that the PNG encoding (which is the slowest part of saving the images) happens in parallel.

## Demo Video or Screenshot
Testing on my Core i5, I'm seeing the time spend just encoding images and writing to disk going from ~450ms down to 250-300ms. Here's a chart of the total time for _all of intake, transfer, interpret, write and record_ before and after:

image

The median scan time went from 3.47s to 2.865s, a decrease of about 23%.

## Testing Plan
Manual testing, new automated testing to prevent regression of the PNG encoding parallelism.

### DIFF
--- a/apps/scan/backend/src/scanners/pdi/state_machine.ts
+++ b/apps/scan/backend/src/scanners/pdi/state_machine.ts
@@ -1,12 +1,22 @@
+import { InsertedSmartCardAuthApi } from '@votingworks/auth';
 import { assertDefined, throwIllegalValue } from '@votingworks/basics';
-import { ImageData } from 'canvas';
+import { BaseLogger, LogEventId, LogLine } from '@votingworks/logging';
 import {
   ScannerClient,
   ScannerError,
   ScannerEvent,
   ScannerStatus,
 } from '@votingworks/pdi-scanner';
+import {
+  SheetInterpretation,
+  SheetOf,
+  ballotPaperDimensions,
+} from '@votingworks/types';
+import { UsbDrive } from '@votingworks/usb-drive';
+import { time, Timer } from '@votingworks/utils';
 import assert from 'assert';
+import { ImageData } from 'canvas';
+import { v4 as uuid } from 'uuid';
 import {
   ActorRef,
   BaseActionObject,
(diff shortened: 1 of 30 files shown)


apps/scan/backend/src/scanners/pdi/state_machine.ts
```diff
@@ -19,52 +29,33 @@ import {
...
   PrecinctScannerStateMachine,
 } from '../../types';
+import { rootDebug } from '../../util/debug';
+import { Workspace } from '../../util/workspace';
 import { recordAcceptedSheet, recordRejectedSheet } from '../shared';
-import { isReadyToScan } from '../../app_flow';
 
 const debug = rootDebug.extend('state-machine');
 
+let scanAndInterpretTimer: Timer | undefined;
```
jonahkagan, line 47: Just making sure you want to leave the timer in

libs/image-utils/src/image_data.ts
```diff
@@ -96,38 +101,103 @@ export function fromGrayScale(
...
+  const canvas = createCanvasWithImageData(image);
   return mimeType === 'image/jpeg'
     ? canvas.toDataURL(mimeType)
     : canvas.toDataURL(mimeType);
 }
 
 /**
  * Writes an image to a file.
+ *
+ * @param buffer - Whether to buffer the image before writing it to disk.
```
jonahkagan, line 127: I don't see a `buffer` param - is this outdated?

jonahkagan (approved):
Long time coming!

libs/image-utils/src/image_data.ts
eventualbuddha: Ah yep. Will remove 👍

apps/scan/backend/src/scanners/pdi/state_machine.ts
eventualbuddha: I do. I think we're going to want to be able to get this timing information in the future.

Issue #5112, opened 2024-07-26
Auto-rotate BMDB images during interpretation so that they're oriented correctly in WIA UI


arsalansufi:
https://votingworks.slack.com/archives/CEL6D3GAD/p1721927741119439

PR #5165, opened 2024-07-31
Rotate bmd ballots if the qr code is on the bottom

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/5112
Now that we display BMD ballots in WIA in VxAdmin we should make sure they always save right-side up. Conveniently we already detect if the QR code is on the top or bottom which will tell us whether or not to rotate. 

## Testing Plan
Manually scanned bmd ballots in all orientations and saw the ballot images save in the CVR properly oriented. 

Ran tests and visually verified new snapshots were right side up and not corrupted. 

## Checklist


libs/ballot-interpreter/src/bmd/utils/rotate.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    // Copy the pixel data from the original image to the new position in the output image
+    output.data[newIndex * 4] = imageData.data[i * 4] || 0; // R
+    output.data[newIndex * 4 + 1] = imageData.data[i * 4 + 1] || 0; // G
+    output.data[newIndex * 4 + 2] = imageData.data[i * 4 + 2] || 0; // B
+    output.data[newIndex * 4 + 3] = imageData.data[i * 4 + 3] || 0; // A
+  }
+
+  return output;
+}
```
eventualbuddha, line 28: This version is much faster:

```suggestion
/**
 * Rotate an image 180 degrees in place.
 */
export function rotateImageData180(imageData: ImageData): void {
  const { data } = imageData;
  const channels = 4;

  for (
    let head = 0, tail = data.length - channels;
    head < tail;
    head += channels, tail -= channels
  ) {
    const r = data[head] as number;
    data[head] = data[tail] as number;
    data[tail] = r;

    const g = data[head + 1] as number;
    data[head + 1] = data[tail + 1] as number;
    data[tail + 1] = g;

    const b = data[head + 2] as number;
    data[head + 2] = data[tail + 2] as number;
    data[tail + 2] = b;

    const a = data[head + 3] as number;
    data[head + 3] = data[tail + 3] as number;
    data[tail + 3] = a;
  }
}
```

It should be fine that we do the transform in place since we never really need to keep both versions. Here's the benchmark results:

```js
[
  {
    'Task Name': 'rotateImageData180',
    'ops/sec': '23',
    'Average Time (ns)': 42768779.32995558,
    Margin: '±1.74%',
    Samples: 12
  },
  {
    'Task Name': 'rotateImageData180 (new)',
    'ops/sec': '117',
    'Average Time (ns)': 8493360.658823434,
    Margin: '±1.56%',
    Samples: 59
  }
]
```

eventualbuddha (approved):

Issue #5121, opened 2024-07-29
Holistically consider VxMark and VxScan settings, and when they should and shouldn't be reset


arsalansufi:
A parent ticket for all remaining issues related to settings reset:

- https://github.com/votingworks/vxsuite/issues/5034
- https://github.com/votingworks/vxsuite/issues/5087
- https://github.com/votingworks/vxsuite/issues/5055
- https://github.com/votingworks/vxsuite/issues/4947

Some recent PRs involving settings reset include:

- https://github.com/votingworks/vxsuite/pull/5085
- https://github.com/votingworks/vxsuite/pull/5089
- https://github.com/votingworks/vxsuite/pull/5092

Issue #4947, opened 2024-06-11
VxMark: Don't reset voter settings after poll worker reloads paper that was incorrectly loaded


arsalansufi:
Slack thread: https://votingworks.slack.com/archives/C03Q42V9MJ5/p1717783760439829

<video src='https://github.com/votingworks/vxsuite/assets/12616928/6df4a5e6-f7ab-42bd-80ff-5a2d1eeeed97' />

kshen0:
This might be a race condition. I've been able to reproduce but unreliably. Adding console logs seems to increase the chance that `useSessionSettingsManager` is able to restore settings after the poll worker card is removed.

PR #5171, opened 2024-08-01
[Mark] Tighten up voter settings save/restore logic

kofi-q:
## Overview

Fixes https://github.com/votingworks/vxsuite/issues/4947
Fixes https://github.com/votingworks/vxsuite/issues/5121

Attempting to make the `useSessionSettingsManager` hook a little less brittle and more readable:
- Moving the responsibility of resetting settings at the end of the session to the apps (VxMark, VxMarkScan), so that it's more explicit and more tightly coupled with the `resetBallot` action in each app.
- Break up the effect into smaller effects/callbacks to make sure each piece is running only when necessary.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/7089c5a5-552c-4be9-af7b-095b48cd2862

## Testing Plan
- Updated associated unit tests
- Manual testing:
  - Insert and remove card at various points during the voter flow - verify settings are restored when card is removed.
  - Deactivate voter session at various points during the voter flow - verify settings stay in default state when card is removed or when a new session is started.

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/app.test.tsx
+++ b/apps/mark-scan/frontend/src/app.test.tsx
@@ -29,6 +29,10 @@ let apiMock: ApiMock;
 beforeEach(() => {
   jest.useFakeTimers();
   apiMock = createApiMock();
+
+  mockOf(useSessionSettingsManager).mockReturnValue({
+    onSessionEnd: jest.fn(),
+  });
 });
 
 afterEach(() => {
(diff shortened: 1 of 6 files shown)


carolinemodic (approved):
Nice this is a lot cleaner. Are there other audio settings like playback speed that should be getting reset too?

Issue #3049, opened 2023-02-28
Seal-less BMD ballots may not find the QR code properly


eventualbuddha:
## Description
BMD ballots printed without a seal may adjust the position of the QR code away from where it is expected to be such that our cropping doesn't find it.

[Slack context](https://votingworks.slack.com/archives/CEL6D3GAD/p1677608658764839)

## Repro

| PIC-1677608507-001.jpg | PIC-1677608507-001.jpg |
|-|-|
|![PIC-1677608507-001](https://user-images.githubusercontent.com/1938/221951784-f5657183-7477-4ec2-b046-ae052ddf5852.jpg)|![PIC-1677608507-002](https://user-images.githubusercontent.com/1938/221951812-fbadd8d9-7850-473b-859d-d104e70f73e5.jpg)|

1. Download the above images.
2. Use the latest `main` that has `libs/fixtures/data/electionSampleCDF.json`.
3. In `libs/ballot-interpreter-vx`, run: `./bin/ballot-interpreter-vx.js interpret -e ../../libs/fixtures/data/electionSampleCDF.json -b PIC-1677608507-001.jpg -b PIC-1677608507-002.jpg` (adjusting the paths to the images as appropriate)

This is the output @jonahkagan got when he ran it:
```
MetadataDecodeError: Expected QR code not found.
    at detect (/home/jonahkagan/code/vxsuite/libs/ballot-interpreter-vx/src/metadata.ts:103:11)
    at Interpreter.normalizeImageDataAndMetadata (/home/jonahkagan/code/vxsuite/libs/ballot-interpreter-vx/src/interpreter/index.ts:407:26)
    at Interpreter.interpretBallot (/home/jonahkagan/code/vxsuite/libs/ballot-interpreter-vx/src/interpreter/index.ts:247:24)
    at Object.run (/home/jonahkagan/code/vxsuite/libs/ballot-interpreter-vx/src/cli/commands/interpret.ts:285:20)
```

When I ran it with `DEBUG=*`, I saw this:

```
  ballot-interpreter-vx:qrcode detect: checking 2547ˣ3318 image +0ms
  ballot-interpreter-vx:qrcode cropping top to check for QR code: { x: 0, y: 0, width: 637, height: 415 } +0ms
  ballot-interpreter-vx:qrcode scanning with qrdetect +1ms
  ballot-interpreter-vx:qrcode qrdetect found no raw QR codes in top +5ms
[...]

PR #5173, opened 2024-08-01
[Mark] Swap positions of ballot QR code and metadata for safer detection

kofi-q:
## Overview

Fixes https://github.com/votingworks/vxsuite/issues/3049

Fixing a QR code detection issue with the QR code getting pushed too far inward when the ballot is rendered with no state seal. Swapping the metadata text with the QR code image, so that it's consistently closest to the edge of the sheet.

## Demo Video or Screenshot

_See commit for print image diff._

## Testing Plan
- Existing tests
- Manual run-through to verify scanning works with and without seals present

## Checklist


carolinemodic (approved):
There is code for the bmd paths in libs/ballot-interpreter that uses lib/bmd-ballot-fixture-generator (or something like that) to generate fixtures for interpretation testing. Will this change also update those tests? I would expect to see a snapshot test update on the snapshot test I landed here (https://github.com/votingworks/vxsuite/pull/5165) if so  (I think we should make sure the interpreter tests are covering this)

kofi-q:
Hm, would've expected tests to fail in CI if there was more snapshots that needed updating, but it could be that we've set the tolerance high enough in those cases to miss them. Might try going through and manually deleting them so they get re-generated

kofi-q:
Ohh, just looked at the PR - if these are new snapshots you just merged, I'll need to rebase first. Will see it that does the trick

PR #5174, opened 2024-08-01
feat(image-utils): add `toMatchImage` Jest matcher

eventualbuddha:
## Overview

Replaces `jest-match-image` in `libs/hmpb` with our own version that operates on raw `ImageData` rather than requiring a round-trip through PNG encoding. Speeds up the `hmpb` "generate fixtures" test quite a bit, and can be used anywhere we'd like to compare images but currently do in a janky way because `expect(image1).toEqual(image2)` makes Jest explode on mismatch.

## Demo Video or Screenshot
An example diff image output:

image

## Testing Plan
Added tests for `toMatchImage` and transitioned existing uses to the new one.


libs/image-utils/src/jest_match_image.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  return {
+    message: () =>
+      `Expected the images to be equal, but they differ by ${diff} pixels.${
+        /* istanbul ignore next */
+        options.diffPath ? ` Diff image saved to ${options.diffPath}.` : ''
+      }`,
+    pass,
+  };
```
jonahkagan, line 99: `options.diffPath` should be defined at this point due to return on L75

jonahkagan (approved):
Very exciting! I was conflicted when I added these tests because they were so slow

PR #5175, opened 2024-08-01
refactor(image-utils): replace `Image` APIs with `ImageData` APIs

eventualbuddha:
We don't need the `Image` APIs, so this simplifies things. Also removes the redundant `toImageBuffer` function.


libs/image-utils/src/jest_pdf_snapshot.ts
```diff
@@ -33,7 +33,7 @@ export async function toMatchPdfSnapshot(
     typeof received === 'string' ? await readFile(received) : received;
   const pdfPages = pdfToImages(pdfContents, { scale: 200 / 72 });
   for await (const { page, pageNumber } of pdfPages) {
-    const imageBuffer = toImageBuffer(page);
+    const imageBuffer = await encodeImageData(page, 'image/png');
```
eventualbuddha, line 36: Somehow this is causing this sort of failure:

image

Feels like a race condition of some sort.

jonahkagan (approved):

libs/image-utils/src/jest_pdf_snapshot.ts
eventualbuddha: I reverted this change to avoid this problem.

PR #5192, opened 2024-08-01
Update copy on casting ballot

carolinemodic:
## Overview

@jonahkagan @arsalansufi I wasn't sure the best practice here on updating the translations overrides file for VxDesign and/or generated fixtures that have ui_strings files. I opted not to update translations_overrides which had an existing spanish override for 'My Ballot is Correct'. I also did not update any of the fixtures which will override the text back to "My Ballot is Correct" with the current ui_strings files (same is true for the current UAT election package) 

## Demo Video or Screenshot
Before
before

After
after

## Testing Plan
Viewed page. 

## Checklist


libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -78,9 +78,7 @@ export const appStrings = {
   buttonBack: () => <UiString uiStringKey="buttonBack">Back</UiString>,
 
   buttonBallotIsCorrect: () => (
-    <UiString uiStringKey="buttonBallotIsCorrect">
-      My Ballot is Correct
-    </UiString>
+    <UiString uiStringKey="buttonBallotIsCorrect">Cast my Ballot</UiString>
```
jonahkagan, line 81: I think capitalize "My" (to match "Print My Ballot" on an earlier screen)

jonahkagan (approved):
I don't think the fixtures necessarily need to be updated, but maybe the override should be removed?

arsalansufi:
@carolinemodic Yeah don't worry about the translation overrides for now! That's actually how I'm going to load in vendor-generated translations. So I'll take care of that once strings are stable and I can request a final batch of translations

I will admit that the translation overrides are a bit clunky right now and long-term a JSON file might be cleaner than a TS file, but it works for now

Issue #5153, opened 2024-07-31
VxMark: If you insert a PAT device while a ballot is printing, it will crash the app


arsalansufi:
VxMarkScan: If you insert a PAT device while a ballot is printing it will crash the app. I noticed this testing changes related to this to fix the bug with inserting a PAT device on the pollworker screen causing a white screen. But my changes did not cause this and the crash exists on main right now as well so I'm not planning to fix in the scope of this other bug fix.

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1721936151123989?thread_ts=1721936151.123989&cid=CJU9MSC6S

PR #5195, opened 2024-08-02
Only poll PAT device status from voter states

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/5153

Generally, changes the state machine to only poll for PAT changes during the voter flow. This replaces the transition guards that would no-op if auth wasn't cardless voter auth.

Fixes an issue where connecting a PAT device during printing would cause the app to crash.

## Demo Video or Screenshot

Before

https://github.com/user-attachments/assets/d2ee63d8-bd44-45bb-ac82-40b36df4a8a1

---

After

https://github.com/user-attachments/assets/6fcf30ef-8f72-44ba-a2be-27f78047aaf4

## Checklist


apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -686,9 +682,7 @@ export function buildMachine(
             },
 
             waiting_for_ballot_data: {
-              entry: assign({
-                isInVoterSession: true,
-              }),
+              invoke: pollPatDeviceConnectionStatus(),
```
arsalansufi, line 685: Just checking my understanding, we're now polling PAT device connection in specific voter states as opposed to all voter states, so specifically:
- `waiting_for_ballot_data`: The voting UI
- `presenting_ballot`: The confirmation UI
- `waiting_for_ballot_reinsertion`: That start of the reinsertion UI
- `waiting_for_invalidated_ballot_confirmation`: A follow-up step in the confirmation UI

Was the issue with the previous approach that we were polling PAT device connection status in too many states and surfacing the PAT configuration UI at points where it could interrupt operations like printing?

arsalansufi (approved):
LGTM!

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
kshen0: That's correct. I'm not exactly sure why the transition guards didn't work in those cases, but we wanted to make this refactor anyway so I didn't spend too much time investigating.

Since the state machine has many states but only these 4 are voter states where the voter is in control, it made more sense to opt in to PAT polling than opt out with the guarded transition.

PR #5206, opened 2024-08-05
build(ballot-interpreter): fix build in Rust v1.80.0

eventualbuddha:
The latest Rust seems stricter about unused structs. Let's remove them!


kofi-q (approved):
🙏

Issue #4956, opened 2024-06-12
VxScan: Investigate PDI ejection failure issue


mattroe:
https://votingworks.slack.com/archives/C070FAWK71P/p1719878129956169

https://votingworks.slack.com/archives/CEL6D3GAD/p1716517550599449

arsalansufi:
Related: https://github.com/votingworks/vxsuite/issues/5051

arsalansufi:
Consolidating this issue and https://github.com/votingworks/vxsuite/issues/5051 and closing https://github.com/votingworks/vxsuite/issues/5051

PR #5217, opened 2024-08-05
VxScan: Improve handling of jam on ballot accept with PDI scanner

jonahkagan:
## Overview
Fixes https://github.com/votingworks/vxsuite/issues/4956

Previously, if a ballot jammed during the accept command, we tried to reject it back out of the front. However, due to an unknown bug (likely a race condition), the scanner did not actually turn on the motor when it received the command to eject the ballot to the front. This caused the ballot to be recorded as rejected even though it remained in the back of the scanner (and could easily fall into the ballot box as it wasn't held by the rollers). This would result in a ballot in the ballot box that was not counted.

To ensure that a ballot in the box is always counted, we change the approach taken. Since the ballot in this case is already ejected enough from the rear of the scanner to fall into the ballot box, we always record it as accepted. Then, when we prepare to start waiting for a new ballot to be inserted, if the rear sensors are still covered (meaning the last ballot didn't fully make it into the ballot box), we declare a jam and show a message that the ballot box opening is obstructed (since this is likely to occur only when the ballot box is overfull).

## Demo Video or Screenshot
(note that the copy is changed from what you see in this video, "Ballot Box Opening Blocked," to "Ballot Box Opening is Blocked")
https://github.com/user-attachments/assets/9c562dda-7e3c-4b7d-9625-4a86b6325bbc

## Testing Plan
- Updated automated tests
- Manually tested by obstructing ballots on accept
## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/pdi/state_machine.ts
+++ b/apps/scan/backend/src/scanners/pdi/state_machine.ts
@@ -360,9 +360,18 @@ function buildMachine({
         invoke: pollScannerStatus,
         on: {
           SCANNER_STATUS: [
+            /**
+             * If the ballot jams during accept, it probably means the ballot box is
+             * full or something is blocking the opening. Since the ballot is already
+             * in the rear of the scanner when we start accepting, we declare it
+             * accepted even during a jam. Though it may still be covering the rear
+             * sensors, the rollers will have released it, so it can fall into the
+             * box. This jam will be identified and displayed to the user as an
+             * outfeed_blocked jam before any more ballots can be scanned.
+             */
             {
               cond: (_, { status }) => status.documentJam,
-              target: '#rejecting',
+              target: '#accepted',
             },
             {
               cond: (_, { status }) => !anyRearSensorCovered(status),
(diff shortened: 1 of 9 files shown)


apps/scan/backend/src/scanners/pdi/state_machine.ts
```diff
@@ -850,7 +850,10 @@ function buildMachine({
           on: {
             SCANNER_STATUS: [
               {
-                cond: (_, { status }) => !status.documentJam,
+                cond: (_, { status }) =>
+                  !status.documentJam &&
+                  !anyRearSensorCovered(status) &&
+                  !anyFrontSensorCovered(status),
```
jonahkagan, line 856: Since we're declaring a jam above when we detect a paper in the rear even though the scanner itself may not have set the `documentJam` flag, we need to make sure the scanner is entirely clear of paper before we exit the jammed state. With the previous approach, we would exit the jammed state immediately since no `documentJam` flag was set.

eventualbuddha (approved):

PR #5219, opened 2024-08-05
VxScan: Accept ballots on backend after interpretation

### LINKED ISSUE #5162: PDI state machine: Auto-accept on the backend instead of going back and forth between the frontend and backend
https://votingworks.slack.com/archives/CEL6D3GAD/p1722368016499089

jonahkagan:
## Overview

Closes #5162 

There's no need to wait for the frontend to send an accept command. Note that I chose to _not_ check auth status and whatnot before accepting. While we've had that guard in the past (by waiting for the frontend accept command), I figured that the time between starting the scan and accepting is so short that it seems reasonable to just rely on the check we already do before starting the scan.

## Demo Video or Screenshot
No change, just faster 😄 

## Testing Plan
- Updated automated tests
- Manually tested with each scanner

## Checklist


eventualbuddha (approved):
🎉

Issue #4913, opened 2024-06-03
VxScan: Validate corner timing mark identification


jonahkagan:
When testing scanning ballots without the bottom row of timing marks, we realized that in some cases, this caused the interpreter to incorrectly identify the corner timing marks. Details here: https://votingworks.slack.com/archives/C073KBS0G91/p1717102793424839

We should improve the robustness of the interpreter by validating that the corner timing marks match the end of both the top/bottom rows and the left/right columns.

Also relevant is false positive scans of ballots with the corner folded: https://votingworks.slack.com/archives/C070FAWK71P/p1721062576578069

arsalansufi:
@eventualbuddha Is this issue still relevant given the other interpretation updates that we've made?

PR #5220, opened 2024-08-06
fix(ballot-interpreter): reject corners we are not sure about

eventualbuddha:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/4913.

We were not being very careful about ensuring the edges first/last timing marks agreed and were thus quite likely to be corners. However, naively ensuring that we always found the corner timing marks rejected a lot of ballots due to the shape of those marks sometimes being incorrect. This might occur when the corners are quite close to the edge.

To address this, I re-used the logic to determine which of the bottom timing marks are present/absent and expanded it to look for white surrounding the black of the timing mark itself. This ensures that images from when a ballot corner is folded down are properly rejected, but minor scanning artifacts do not interfere with inferring the corner timing marks.

## Demo Video or Screenshot

A bottom-right corner timing mark that is not initially found when rejecting naively because of the black along the edge, but will be inferred because it matches the black surrounded by white pattern. This was already handled properly on `main`, but required going beyond a naive rejection of non-agreeing right and bottom sides:
image

A corner-folded ballot that was previously accepted due to an erroneously inferred top-left corner timing mark:

image

## Testing Plan
Added new tests for cases where we _do_ and _don't_ want to infer corners. Updated existing tests to account for the stricter corner finding.


libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -327,12 +314,12 @@ pub fn interpret_ballot_card(
     };
 
     let (side_a_grid_result, side_b_grid_result) = par_map_pair(
-        (&side_a.image, &mut side_a_debug),
-        (&side_b.image, &mut side_b_debug),
-        |(image, debug)| {
```
eventualbuddha, line 332: Changes along these lines are just to support passing a `BallotImage` instead of a `GrayImage` in order to get the threshold alongside the image.

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -699,10 +692,21 @@ mod test {
...
-            ("scan-skewed-side-a.jpeg", "scan-skewed-side-b.jpeg"),
+        let (mut side_a_image, side_b_image, options) = load_ballot_card_fixture(
+            "2023-05-09-nh-moultonborough",
+            ("write-ins-front.jpeg", "write-ins-back.jpeg"),
         );
+
+        let detected = qr_code::detect(&side_a_image, &ImageDebugWriter::disabled()).unwrap();
+        let qr_code_bounds = detected.bounds();
+
+        // white out the QR code on side A
```
eventualbuddha, line 703: Rather than relying on knowing that the image has a QR code that isn't found, I deliberately create an image without the QR code.

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -787,8 +819,11 @@ mod test {
     #[test]
     fn test_high_rotation_is_rejected() {
         let (side_a_image, side_b_image, options) = load_ballot_card_fixture(
-            "2021-06-06-lincoln-test-ballot",
-            ("high-rotation-front.png", "high-rotation-back.png"),
+            "nh-test-ballot",
+            (
+                "template-rotated-3deg-front.jpeg",
```
eventualbuddha, line 824: I created these images manually by rotating them beyond the allowed rotation limit (2º).

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -363,34 +376,148 @@ pub fn find_timing_mark_grid(
...
-                rect_within_image.width(),
-                rect_within_image.height(),
-            );
-            let rect_area = rect.width() * rect.height();
-            if rect_sub_image
-                .pixels()
-                .filter(|(_, _, luma)| luma.0[0] <= threshold)
-                .count()
-                > (rect_area / 2) as usize
-            {
```
eventualbuddha, line 393: This previously logic only checked that the expected timing mark area was black(ish). The replacement checks that the area is black(ish) and surrounded by white(ish).

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -513,16 +506,16 @@ pub fn interpret_ballot_card(
...
                     BallotStyleId::from(format!("card-number-{}", front_metadata.card)),
                 ),
                 (
                     BallotPageTimingMarkMetadata::Back(_),
                     BallotPageTimingMarkMetadata::Front(front_metadata),
                 ) => (
-                    side_b,
-                    side_a,
+                    (side_b.0, side_b.1.image, side_b.2, side_b.3),
+                    (side_a.0, side_a.1.image, side_a.2, side_a.3),
```
jonahkagan, line 518: It looks like `side_a` and `side_b` are only used here. I think this bit would be more readable if we never constructed the `side_a` and `side_b` tuples and just used their components directly.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -18,12 +15,28 @@ use types_rs::{election::UnitIntervalValue, geometry::IntersectionBounds};
...
+pub struct BallotPage {
+    pub ballot_image: BallotImage,
+    pub geometry: Geometry,
+}
+
+pub struct BallotCard {
+    pub side_a: BallotImage,
+    pub side_b: BallotImage,
+    pub geometry: Geometry,
+}
```
jonahkagan, line 38: Maybe these could live in the `ballot_card` module? That seems slightly more thematically appropriate

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -363,34 +393,148 @@ pub fn find_timing_mark_grid(
...
+        .pixels()
+        .filter(|(_, _, luma)| luma.0[0] <= threshold)
+        .count();
+    let black_pixel_score = black_pixel_count as f32 / rect_area as f32;
+
+    let mut white_pixel_count = 0;
+    let mut possible_white_pixel_count = 0;
+
+    // crop the area to the left of the timing mark whose width is the same as
+    // the timing mark's height, then look for white pixels in the rectangle
```
jonahkagan, line 409: Why do we use the timing mark's height here to extend the _width_ to the left? Seems like an odd mix of dimensions that has me a tad confused. Something like `width/2` or `width/4` would be more what I would expect

jonahkagan: Oh, is it because we're looking above/below by `height` later?

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -363,34 +393,148 @@ pub fn find_timing_mark_grid(
...
-            );
-            let rect_area = rect.width() * rect.height();
-            if rect_sub_image
-                .pixels()
-                .filter(|(_, _, luma)| luma.0[0] <= threshold)
-                .count()
-                > (rect_area / 2) as usize
-            {
+            let score = score_timing_mark_at_rect(ballot_image, rect)?;
+            if score > MIN_REQUIRED_MARK_SCORE {
```
jonahkagan, line 521: Pedantic nit that probably doesn't matter in practice. If this is the "min required" score I would expect `>=` here

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -1234,6 +1388,69 @@ pub fn find_complete_timing_marks_from_partial_timing_marks(
         );
     }
 
+    const MIN_REQUIRED_CORNER_MARK_SCORE: f32 = 0.5;
+    let mut missing_corners = vec![];
+
+    if partial_timing_marks.top_left_rect.is_none() {
```
jonahkagan, line 1380: I wonder if we should score every left and right inferred timing mark to give us more confidence.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -1234,6 +1388,69 @@ pub fn find_complete_timing_marks_from_partial_timing_marks(
         );
     }
 
+    const MIN_REQUIRED_CORNER_MARK_SCORE: f32 = 0.5;
```
jonahkagan: How did you pick this threshold? What's the general range of scores you've seen?

jonahkagan (approved):
Seems like a solid solution! Glad to see all the additional test cases

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: Maybe. I wonder what we'd do if any didn't meet the threshold. Maybe reject if there's more than a handful?

jonahkagan: I was thinking fail if any don't meet the threshold. I feel like every side timing mark should easily meet it, unless they've been ripped out or colored in-between. Protecting against ripping out a chunk of the ballot seems useful, though maybe not the most realistic worry.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: I mostly just left it as it was, though I realize that the thing being scored is different. In practice, for clearly defined timing marks with a small amount of noise around them, the score is 60-90%:

image

image

image

jonahkagan: Ok, great!

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: I'd like to consider this as a future enhancement. I think it could be a good idea, but I'd want to do more testing. cc @arsalansufi

PR #5221, opened 2024-08-06
VxScan: Fix erroneous jam error message after double sheet detected

### LINKED ISSUE #5143: VxScan: MSD error quickly transitions to generic jam error
_Update: This might be specific to a single unit, currently in Ben's possession. Roe has not been able to repro on the current Atrium unit. Feel free to pause on this if you can't easily repro._

When scanning multiple sheets you get the multi sheet error and then it quickly changes to the generic jam error. I’ve also noticed the jam kickback is sometimes forceful enough that it doesn’t detect the paper anymore and then just shows insert ballot instead of the error message (maybe ok? But in the multi sheet case the error is useful as the voter might not even realize they have two pages stuck together) 

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1721250303259169?thread_ts=1721250303.259169&cid=CJU9MSC6S

jonahkagan:
## Overview

Fixes #5143 

Details in the commit messages

## Demo Video or Screenshot
Before

https://github.com/user-attachments/assets/1564047e-c214-44dc-a1f7-51f75b913401

After

https://github.com/user-attachments/assets/dd87a13a-9733-4c4d-b461-6b71d2999330

## Testing Plan
- Added regression test
- Manually tested

## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/pdi/pdi_app_scan_double_sheet.test.ts
+++ b/apps/scan/backend/src/scanners/pdi/pdi_app_scan_double_sheet.test.ts
@@ -316,7 +316,7 @@ test('insert two sheets back-to-back as if they were one long sheet', async () =
   );
 });
 
-test('insert two sheets at once', async () => {
+test('insert two sheets at once - scanFailed event', async () => {
   await withApp(
     async ({ apiClient, mockScanner, mockUsbDrive, mockAuth, clock }) => {
       await configureApp(apiClient, mockAuth, mockUsbDrive);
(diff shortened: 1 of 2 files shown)


apps/scan/backend/src/scanners/pdi/pdi_app_scan_double_sheet.test.ts
```diff
@@ -344,6 +344,39 @@ test('insert two sheets at once', async () => {
...
+      await expectStatus(apiClient, { state: 'scanning' });
+      // In this case, the scanner seems to set the documentJam flag as well
+      mockScanner.setScannerStatus({
+        ...mockStatus.documentInFront,
+        documentJam: true,
+      });
+      mockScanner.emitEvent({ event: 'error', code: 'doubleFeedDetected' });
+      mockScanner.emitEvent({
+        event: 'scanComplete',
+        images: await ballotImages.completeHmpb(),
```
eventualbuddha, line 365: In reality we'd likely be getting really short images, right? Do you want to just create a pair of ~100px tall images for this?

eventualbuddha (approved):

apps/scan/backend/src/scanners/pdi/pdi_app_scan_double_sheet.test.ts
jonahkagan: I've noticed a variety of behavior actually, from stopping immediately to scanning the entire ballot (granted, that was in a case where I intentionally mis-calibrated double sheet detection by using printed ballots instead of blank sheets)

PR #5226, opened 2024-08-07
VxScan: Set Fujitsu thermal printer (V4) as default in development and builds

adghayes:
## Overview

Sets it as the default, just as the PDI scanner is now the default.


apps/scan/backend/src/__image_snapshots__/legacy-polls-closed-report-1.png
adghayes, line 1: I'm not sure why git didn't catch all of these snapshot changes as renames, but I've visually compared them and they are the same.

jonahkagan (approved):

Issue #5048, opened 2024-07-12
Better understand idle time performance and make sure we aren't unnecessarily re-rendering in response to polling data that isn't actually changing


arsalansufi:
https://votingworks.slack.com/archives/C062Y0ZFKFT/p1719347442239229

Also make sure we aren't doing anything strange re reads/writes to the USB drive on VxScan: https://votingworks.slack.com/archives/C070FAWK71P/p1719874964559819

Update: One more relevant thread: https://votingworks.slack.com/archives/CEL6D3GAD/p1721677108669309

arsalansufi:
Relevant change that could close this out: https://github.com/votingworks/vxsuite/pull/5227

arsalansufi:
Closing this out as we've seen https://github.com/votingworks/vxsuite/pull/5227 help

PR #5227, opened 2024-08-07
Reduce React Query re-renders across apps

adghayes:
I wanted to check in on re-renders on VxMark. With the React Dev Tools, it looks like the app is re-rendering at the app root on every auth poll when authenticated, every 300ms. This is a small cost on my dev machine in Chrome - like 10ms total of extra rendering per second. But on the slow VxMark, that could be a lot greater.

We've fixed this elsewhere by updating the `structuralSharing` method. I just sub in our fix at the across-app-level React Query settings. This fixes the polling problem and also should address it for future polling added.

The reason this is necessary is because the default `react-query` comparison method only considers two objects equal if they both have the plain object prototype.

### DIFF
--- a/libs/ui/src/react_query.ts
+++ b/libs/ui/src/react_query.ts
@@ -1,4 +1,5 @@
 import type { DefaultOptions } from '@tanstack/react-query';
+import { persistDataReferenceIfDeepEqual } from '@votingworks/utils';
 
 /**
  * A custom retry function
@@ -45,6 +46,12 @@ export const QUERY_CLIENT_DEFAULT_OPTIONS: DefaultOptions = {
     // phase so it will propagate up to the nearest error boundary. Consumers
     // are responsible for defining a global error boundary.
     useErrorBoundary: true,
+
+    // react-query's default method here, `replaceEqualDeep`, does not consider
+    // two objects equal unless they both have the same plain object prototype.
+    // This means that two identical Dates, for example, would not be considered
+    // equal, so we use our own helper leveraging lodash's `deepEqual`.
+    structuralSharing: persistDataReferenceIfDeepEqual,
   },
   mutations: {
     networkMode: 'always',
--- a/libs/utils/src/structural_sharing.ts
+++ b/libs/utils/src/structural_sharing.ts
@@ -1,0 +1,21 @@
+import { deepEqual } from '@votingworks/basics';
+
+/**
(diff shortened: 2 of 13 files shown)


arsalansufi:
Nice!! This is roughly the kind of change that I had in mind for us to explore when I opened https://github.com/votingworks/vxsuite/issues/5048

libs/utils/src/structural_sharing.ts
```diff
@@ -0,0 +1,21 @@
...
+  newData: T
+): T {
+  if (!oldData) {
+    return newData;
+  }
+
+  // Prevent unnecessary re-renders of dependent components
+  const isUnchanged = deepEqual(oldData, newData);
+  return isUnchanged ? oldData : newData;
+}
```
jonahkagan, line 21: Here's the function we're replacing: https://github.com/TanStack/query/blob/1fc6124bfd22a26095ab1fca20caa39b4036fd86/packages/query-core/src/utils.ts#L244 (call site: https://github.com/TanStack/query/blob/1fc6124bfd22a26095ab1fca20caa39b4036fd86/packages/query-core/src/utils.ts#L357)

Maybe you could test out running that function on some of our returned data to figure out why this isn't working out of the box? (Also, I didn't track down the exact version of react-query that we're using, so it might be worth doing that before testing)

adghayes: Thank you for the pointers! I followed the code and found that their default function only considers objects equal if they both have the plain object prototype. This means that our `DateWithoutTime` and even the inbuilt `Date` objects were causing equality checks to fail.

adghayes:
Note that the third commit is just reverting the second, so you only have to review the first and last commits.

libs/ui/src/react_query.ts
```diff
@@ -45,6 +46,12 @@ export const QUERY_CLIENT_DEFAULT_OPTIONS: DefaultOptions = {
     // phase so it will propagate up to the nearest error boundary. Consumers
     // are responsible for defining a global error boundary.
     useErrorBoundary: true,
+
+    // react-query's default method here, `replaceEqualDeep`, does not consider
+    // two objects equal unless they both have the same plain object prototype.
+    // This means that two identical Date's, for example, would not be considered
```
jonahkagan, line 52: remove apostrophe in "Date's"

jonahkagan (approved):
🎉 great digging!

Issue #2867, opened 2023-01-04
Implement challenge-based Signed Hash Validation in every component


benadida:
## Summary

Every component (VxAdmin, VxMark, etc.) should provide a proof of integrity via LiveCheck, as follows:
- Prompt user to launch https://check.voting.works on their phone
- Prompt for challenge code to be entered
- Display a QR code containing a LiveCheck signature

## Tasks
- [x] core functions in library
- [x] add to VxAdmin
- [x] add to VxScan
- [x] add to VxCentralScan
- [x] add to VxMark
- [x] Incorporate system hash (computable via https://github.com/votingworks/vxsuite-complete-system/blob/main/config/admin-functions/hash-signature.sh) into signature
- [x] Add some kind of loading indicator since computing system hash can take time
- [x] Make sure we render the button for all relevant user types on all machines (https://votingworks.slack.com/archives/CJU9MSC6S/p1722349141192579)

arsalansufi:
Using https://github.com/votingworks/vxsuite/issues/5412 to track the last change outside the certified system

PR #5241, opened 2024-08-08
Replace references to Live Check with signed hash validation (and remove feature flag)

arsalansufi:
## Overview

I'm finally returning to https://github.com/votingworks/vxsuite/issues/2867 and closing out some loose ends. A while back, we decided to rebrand Live Check ➡️ signed hash validation. While I'd previously updated user-facing copy, I hadn't updated all internal references (variable names, file names, etc.). This PR takes care of that, before I move onto actual functional changes.

This PR also removes the no longer needed feature flag for the feature. We've been building with this feature flag enabled for quite some time now.

## Testing Plan

Ran `git grep` to make sure that no references to any of the following remain:

- [x] `LIVE_CHECK`
- [x] `LIVECHECK`
- [x] `LiveCheck`
- [x] `liveCheck`
- [x] `live_check`
- [x] `Live Check` (minus one intentional reference in a comment)
- [x] `live check`


apps/admin/frontend/src/components/signed_hash_validation_button.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react';
```
arsalansufi, line 1: Not actually a new file - just the existing `live_check_buttton.tsx` renamed and then edited via find-and-replace

Same comment elsewhere in this PR

apps/scan/frontend/src/components/signed_hash_validation_button.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react';
```
kshen0, line 1: Should these buttons be factored out to a common component?

kshen0 (approved):

apps/scan/frontend/src/components/signed_hash_validation_button.tsx
arsalansufi: Yes good q! I'm working on exactly that (and other changes) as part of my next PR

PR #5255, opened 2024-08-12
Script to generate large elections

jonahkagan:
## Overview

Adds a script to generate large elections for scale testing. This is just a very basic first draft that will likely be modified as we continue scale testing.

To avoid adding yet another lib, I reworked the `cvr-fixture-generator` lib to be a home for all fixture generators (named `fixture-generators`) and added the script to it.

## Demo Video or Screenshot
Example election: 
[big-election.json](https://github.com/user-attachments/files/16592650/big-election.json)

## Testing Plan
Ran the script and imported the resulting election into VxAdmin

## Checklist


libs/fixture-generators/src/generate-election/config.ts
```diff
@@ -0,0 +1,62 @@
...
+    contestBallotMeasureOptionLabel: positiveInt,
+  }),
+});
+
+export type GenerateElectionConfig = z.infer<
+  typeof GenerateElectionConfigSchema
+>;
+
+/**
+ * TODO once we set our system limits, we should set this default config to
```
carolinemodic, line 36: This comment makes sense to me and makes me wonder if we should also then have a basic text documentation for it autogenerated from that code when we get there.

libs/fixture-generators/src/generate-election/generate_election.ts
```diff
@@ -0,0 +1,184 @@
...
+  Precinct,
+  safeParse,
+} from '@votingworks/types';
+import { customAlphabet } from 'nanoid';
+import { defaultConfig, GenerateElectionConfig } from './config';
+import { ballotMeasureText, seal, words } from './source_text';
+
+const generateId = customAlphabet('0123456789abcdefghijklmnopqrstuvwxyz', 12);
+
+function chooseRandom<T>(array: T[]): T {
```
carolinemodic, line 27: some of these util functions are probably also useful for other fixture generators, maybe there should be a general utils file they all share

libs/fixture-generators/src/generate-election/generate_election.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  function generatePrecinct(index: number): Precinct {
+    return {
+      id: `precinct-${index}`,
+      name: randomString(maxStringLengths.precinctName, words.locations),
+    };
+  }
+  const precincts = range(0, config.numPrecincts).map(generatePrecinct);
+
+  function generateBallotStyle(index: number): BallotStyle {
```
carolinemodic, line 82: I wonder if this will create wacky elections with failures we wouldn't see in the real world by having districts/precincts without corresponding ballot styles

libs/fixture-generators/src/generate-election/generate_election.ts
```diff
@@ -0,0 +1,184 @@
...
+    } as const;
+    if (Math.random() < 0.5) {
+      return {
+        ...baseContest,
+        type: 'candidate',
+        termDescription: randomString(
+          maxStringLengths.contestTermDescription,
+          words.durations
+        ),
+        seats: config.maxContestVoteFor,
```
carolinemodic, line 125: I wonder if we will want some random variation here but I get that you want to ensure there is at least 1 at the max

libs/fixture-generators/src/generate-election/generate_election.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+  const parties = range(0, config.numParties).map(generateParty);
+
+  function generateContest(index: number): AnyContest {
+    const baseContest = {
+      id: `contest-${index}`,
+      districtId: chooseRandom(districts).id,
+      title: randomString(maxStringLengths.contestTitle, words.offices),
+    } as const;
+    if (Math.random() < 0.5) {
```
carolinemodic, line 117: You could just make the config specify numCandidateContests and numYesNoContests too

carolinemodic (approved):
I like having the fixture-generators generalized together. 

Left some ideas/thoughts but don't consider any of it blocking given that this is just a helper script

libs/fixture-generators/src/generate-election/config.ts
jonahkagan: That could be cool!

libs/fixture-generators/src/generate-election/generate_election.ts
jonahkagan: Good point. I'd be down to see if that happens just out of curiosity. But I do think I added a filter in VxDesign to prevent that, so maybe it's not so relevant to test.

libs/fixture-generators/src/generate-election/generate_election.ts
jonahkagan: Yeah, I agree we may want to change it. This was my naive first cut

jonahkagan:
> I like having the fixture-generators generalized together.
> 
> Left some ideas/thoughts but don't consider any of it blocking given that this is just a helper script

Thanks! I'm gonna merge as-is for now, but I think we should definitely consider making some of the changes you suggested as we learn what's useful for scale testing

PR #5256, opened 2024-08-13
Optimize retrieval of election key during auth polling

### LINKED ISSUE #4790: Explore alternatives to constantly fetching entire election def for auth lib
When an election definition is large, say, because it has a large seal, this seems to bog the system down.

Copying from Slack:

> I know we're constantly refetching the election def from the store [to pass to the auth lib](https://github.com/votingworks/vxsuite/blob/b3aaa3f7c0531719a85e9cb7243dd32e074b5a79/apps/scan/backend/src/util/auth.ts#L10-L21). we initially planned to cache the relevant data in the auth lib, but as we all know caching is hard / error prone. and ultimately decided that the election def was too small for this to matter. a big seal definitely breaks that assumption. for the auth lib use case, all we really need is the election hash. we could explicitly persist that in the DB and just refetch that

Link to Slack thread: https://votingworks.slack.com/archives/C06H2PBP0MU/p1712798609099009?thread_ts=1712790170.563049&cid=C06H2PBP0MU

jonahkagan:
## Overview

Fixes #4790 

During scale testing with a large election definition, I noticed that auth polling was very slow due to loading the election definition from the database on every polling request. This caused the entire app to lag.

To fix this, I changed the auth polling to extract only the relevant fields for the election key (id and date) from the JSON at the database level.

## Demo Video or Screenshot
Before

https://github.com/user-attachments/assets/5859693d-b25c-4959-bdb6-f5572ded0fd3

After

https://github.com/user-attachments/assets/087eb45d-d86b-4f42-8838-ece523939fe7

## Testing Plan
Manual testing

## Checklist

### DIFF
--- a/apps/admin/backend/src/app.test.ts
+++ b/apps/admin/backend/src/app.test.ts
@@ -225,6 +225,12 @@ test('configuring with a CDF election', async () => {
   expect(currentElectionMetadata?.electionDefinition.ballotHash).toEqual(
     ballotHash
   );
+
+  // Ensure loading auth election key from db works
+  mockElectionManagerAuth(auth, electionGeneral);
+  expect(await apiClient.getAuthStatus()).toMatchObject({
+    status: 'logged_in',
+  });
 });
 
 test('configuring with an election not from removable media in prod errs', async () => {
--- a/apps/admin/backend/src/store.ts
+++ b/apps/admin/backend/src/store.ts
@@ -10,6 +10,7 @@ import {
   typedAs,
   isResult,
   assert,
+  DateWithoutTime,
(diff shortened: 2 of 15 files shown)


carolinemodic (approved):
This PR makes me wonder if we could be sharing more code for the auth pieces across apps. Seems a little silly you had to do this 4 times

jonahkagan:
@carolinemodic turns out I missed a case - CDF elections - since I was bypassing the election parsing logic. I added handling for that and regression tests. Requested a new review

carolinemodic (approved):
You have some tests failing but code lgtm

PR #5257, opened 2024-08-13
refactor(converter-nh-accuvote): separate XML parsing

eventualbuddha:
Similar to #5065.

### DIFF
--- a/libs/converter-nh-accuvote/src/convert/accuvote.ts
+++ b/libs/converter-nh-accuvote/src/convert/accuvote.ts
@@ -1,0 +1,274 @@
+import {
+  assert,
+  err,
+  ok,
+  Optional,
+  Result,
+  resultBlock,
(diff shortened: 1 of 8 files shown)


libs/converter-nh-accuvote/src/convert/accuvote.ts
```diff
@@ -0,0 +1,274 @@
...
+function getOptionalChild(
+  element: Element,
+  tagName: string
+): Optional<Element> {
+  for (const child of Array.from(element.childNodes)) {
+    if (child.nodeType === 1 && child.nodeName === tagName) {
+      return child as Element;
+    }
+  }
+}
```
jonahkagan, line 102: Minor nit: could use `getOptionalChild` in `getOptionalChildText`, since there's some repeated logic

jonahkagan (approved):
Readability++

Issue #5245, opened 2024-08-12
VxMark: Make sure printing isn't stalling


arsalansufi:
https://votingworks.slack.com/archives/CEL6D3GAD/p1723128095890309

arsalansufi:
https://github.com/votingworks/vxsuite/pull/5258

PR #5258, opened 2024-08-13
Launch a single Chromium instance on VxMark startup for BMDB printing

arsalansufi:
## Overview

This PR should address https://github.com/votingworks/vxsuite/issues/5245 per my testing. After considering a few explanations that turned out to be red herrings, we ultimately found that printing sporadically stalling in production involved the Chromium `newPage` call hanging:

https://github.com/votingworks/vxsuite/blob/034edbcc5e89f33397961c7e88636af51385b81f/libs/printing/src/render.tsx#L96

We found [this relevant Playwright issue](https://github.com/microsoft/playwright/issues/3939). The poster notes that they're constantly launching new Chromium instances, like we are, and saw the issue go away after switching to a single persisted instance (among other changes). We, too, have been wanting to shift away from launching and closing a new Chromium instance every time we print, primarily for speed gains. But now we have another reason. The exact reason why repeatedly launching and closing new Chromium instances has this effect is still a bit unclear, maybe something to do with memory usage.

## Testing Plan

With this change (hackily) applied to two production 150s, I've not been able to repro print stalling across 40 prints (30 on one machine, 10 on the other). I was definitely reproing this more often than once every 40 prints before so feeling optimistic that this will do the trick.

### DIFF
--- a/apps/mark-scan/backend/src/app.ts
+++ b/apps/mark-scan/backend/src/app.ts
@@ -81,7 +81,7 @@ export function buildApi(
   usbDrive: UsbDrive,
   logger: Logger,
   workspace: Workspace,
-  chromium: Browser,
+  browser: Browser,
   stateMachine?: PaperHandlerStateMachine,
   paperHandler?: PaperHandlerDriverInterface
 ) {
@@ -234,7 +234,7 @@ export function buildApi(
 
       const pdfData = await renderBallot({
         store,
-        chromium,
+        browser,
         ...input,
       });
       stateMachine.printBallot(pdfData);
@@ -449,7 +449,7 @@ export function buildApp(
   logger: Logger,
   workspace: Workspace,
   usbDrive: UsbDrive,
-  chromium: Browser,
+  browser: Browser,
   stateMachine?: PaperHandlerStateMachine,
   paperHandler?: PaperHandlerDriverInterface
 ): Application {
@@ -459,7 +459,7 @@ export function buildApp(
     usbDrive,
     logger,
     workspace,
-    chromium,
+    browser,
     stateMachine,
     paperHandler
   );
--- a/apps/mark-scan/backend/src/app.ts
+++ b/apps/mark-scan/backend/src/app.ts
@@ -46,6 +46,7 @@ import {
   MockPaperHandlerStatus,
   PaperHandlerDriverInterface,
 } from '@votingworks/custom-paper-handler';
+import { Browser } from '@votingworks/printing';
 import { getMachineConfig } from './machine_config';
 import { Workspace } from './util/workspace';
 import {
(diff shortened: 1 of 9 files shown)


apps/mark-scan/backend/src/app.ts
```diff
@@ -80,6 +81,7 @@ export function buildApi(
   usbDrive: UsbDrive,
   logger: Logger,
   workspace: Workspace,
+  chromium: Browser,
```
eventualbuddha, line 84: Slight preference for `browser` or `renderer` rather than `chromium`.

apps/mark-scan/backend/src/util/render_ballot.tsx
```diff
@@ -22,6 +23,7 @@ import { getMarkScanBmdModel } from './hardware';
 
 export interface RenderBallotProps {
   store: Store;
+  chromium: Browser;
```
eventualbuddha, line 26: Ditto here.

apps/mark-scan/backend/test/app_helpers.ts
```diff
@@ -90,6 +91,7 @@ interface MockAppContents {
   stateMachine: PaperHandlerStateMachine;
   patConnectionStatusReader: PatConnectionStatusReaderInterface;
   driver: MockPaperHandlerDriver;
+  chromium: Browser;
```
eventualbuddha, line 94: Ditto here.

libs/printing/src/render.tsx
```diff
@@ -61,6 +61,15 @@ function getContentHeight(page: Page): Promise<number> {
   });
 }
 
+export async function launchChromium(): Promise<Browser> {
```
eventualbuddha, line 64: I could see an argument for leaving this with "Chromium" in the name, but I'd also be happy if this were `launchBrowser`.

eventualbuddha (approved):

eventualbuddha:
Looks like you need to rebase to get the `ci/circleci: test-libs-fixture-generators` job.

adghayes (approved):
@arsalansufi this looks good to me, and I'm glad the `browserOverride` option was already there to make this easier. 

We probably want to apply this change across apps in a follow-up. Could `libs/printing` just own a singleton `chromium` instance, and then we don't have to manage it at the app level and pass it around in tests? This is good to merge to unblock rc3, but noting for v4.0 final.

apps/mark-scan/backend/src/app.ts
arsalansufi: Yeah I was going back and forth on this one - will switch!

arsalansufi:
Yeah the existence of `browserOverride` made this super easy! And agreed that we should apply this change elsewhere, too - will ticket that

`libs/printing` owning a single Chromium instance is an interesting idea! Would have to play around with that to see what that ends up looking like. My hunch is that we'd still end up needing some kind of reference that we keep track of in the app, e.g., a `Printing` class. But maybe that higher level is preferable to a `Browser` class, which is a lower-level detail. Regardless, will leave a note in the ticket that I create

PR #5267, opened 2024-08-15
feat(converter-nh-accuvote): write printing and proofing ballots

eventualbuddha:
## Overview

When converting we now write out two ballot PDFs for each ballot style/precinct combination. The first is prefixed with PRINT and has the QR code to allow it to be scanned, the second overlays information about where the bubbles and contest options are encoded in the election definition. This allows one to quickly determine whether the generated election definition is correct.

## Demo Video or Screenshot

| PRINT Ballot Example | PROOF Ballot Example |
|-|-|
| image | image
 |

Note that the PROOF ballot image above clearly shows that the contest options are in the wrong place.

## Testing Plan
Some new automated tests, manually tested with a few different NH elections.


libs/converter-nh-accuvote/src/convert/convert_election_definition.ts
```diff
@@ -277,6 +293,8 @@ async function convertCardDefinition(
           ),
         },
       ],
+      },
```
jonahkagan, line 296: This looks like the auto-formatting didn't run here - I would expect the contents of this object literal to be indented

libs/converter-nh-accuvote/src/drawing.ts
```diff
@@ -0,0 +1,95 @@
...
+    truncatedText = truncatedText.slice(0, -1);
+    truncatedTextWidth = font.widthOfTextAtSize(
+      truncatedText + ellipsis,
+      maxFontSize
+    );
+  } while (truncatedTextWidth > size.width && truncatedText.length > 0);
+
+  if (truncatedText.length === 0) {
+    // even the smallest font size is too large
+    return undefined;
```
jonahkagan, line 87: maybe throw an error here?

libs/converter-nh-accuvote/src/proofing.ts
```diff
@@ -0,0 +1,535 @@
...
+    );
+    const maximumContestNameWidth = contestTextConfig.font.widthOfTextAtSize(
+      'Modest Contest Name Length',
+      contestTextConfig.maxFontSize
+    );
+
+    const textColor = rgb(1.0, 1.0, 1.0);
+    const textBackgroundColor = rgb(0.5, 0.8, 0.5);
+    const textBackgroundOpacity = 0.8;
+    const textBackgroundPadding = 2;
```
jonahkagan, line 240: A little more luminance contrast between text color and background would improve readability (and also for write-in labels)

jonahkagan (approved):
Looks great! Seems really helpful

libs/converter-nh-accuvote/src/proofing.ts
eventualbuddha: Yeah, fair. It is pretty light. Updated:

image

PR #5280, opened 2024-08-19
ballot-interpreter: Use generated HMPB fixtures instead of saved images

jonahkagan:
## Overview

To avoid having to manually recreate ballot image fixtures in the libs/ballot-interpreter test, use the auto-generated HMPB fixtures from libs/hmpb, modifying them programmatically as needed.

## Checklist


libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -815,26 +859,6 @@ mod test {
         }
     }
 
-    #[test]
-    fn test_skewed_ballot_scoring_write_in_areas_with_write_ins() {
```
jonahkagan, line 819: Decided to remove this test because it doesn't assess whether or not the skew is impacting the write-in scores and we have other tests that already test that UWI detection works.

eventualbuddha (approved):

Issue #5265, opened 2024-08-15
VxScan: Footer election info not updating properly


arsalansufi:
Not sure if this is a bubble ballot election bug or more widespread but I noticed the following behavior on VxScan UAT-RC2

Machine was previously configured for Sample Primary Election. Unconfigured as Election Manager and reconfigured with a new election package for all bubble ballot. Noticed the footer still had the old election so I thought I saved the wrong election package. Re-did this flow a few times with resaving the package and saw the same thing. Eventually realized that it was strange my bubble ballot-election manager card was even working if that wasn't the election so I opened the polls. This crashed (known problem reported elsewhere) and upon reboot the footer now properly reflected the bubble ballot election.

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1723567650346349?thread_ts=1723567650.346349&cid=CJU9MSC6S

carolinemodic:
Repros in Dev switching between BubbleElection and MaximumElection

carolinemodic:
Struggling to debug whats happening it seems like the election title values are getting refreshed properly when console logging but the html is still showing the wrong thing

carolinemodic:
@kofi-q I was able to repro this consistently in dev when unconfiguring and reconfiguring between Jonah's maximum election and the all bubble election. The eelction title would not update but the election hash would.

PR #5285, opened 2024-08-20
[UiStrings] Clear i18next cache when backend translations are removed

kofi-q:
## Overview

Fixes https://github.com/votingworks/vxsuite/issues/5265

Whenever machines are unconfigured, we clear the backend DB and frontend query cache, but our translation library, i18next, has a cache of its own that we weren't clearing yet. That meant stale translations (including English strings) in the voter-facing bits of the UI. These would ordinarily get overwritten when a new election is configured, along with the new translations, but not when we use elections without translations.

Making sure we reset the i18next cache whenever the query cache changes (which happens on unconfigure).

## Testing Plan
- Added a new test case to repro and verify the fix
- Manual verification

## Checklist

### DIFF
--- a/libs/ui/src/ui_strings/language_context.test.tsx
+++ b/libs/ui/src/ui_strings/language_context.test.tsx
@@ -16,7 +16,13 @@ import {
   DEFAULT_LANGUAGE_CODE,
 } from './language_context';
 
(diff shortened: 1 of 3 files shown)


carolinemodic (approved):
Nice

Issue #4964, opened 2024-06-12
VxScan: Adjust ballot scanned count screen location for non-multi-lingual definitions


arsalansufi:
The Voter facing language/accessibility settings button makes the screen broken based on ballot count placement

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1718220434398769?thread_ts=1718220434.398769&cid=CJU9MSC6S

mattroe:
Bumping this up because I can't stand seeing this anymore

arsalansufi:
This one is ultimately not critical for UAT as we'll be using a multi-language definition for UAT

arsalansufi:
@kofi-q I'm going to move this ticket back out of the UAT milestone as we're using a multi-language election package for UAT, and some of the new finds from QA are higher priority for UAT

PR #5286, opened 2024-08-20
[Scan] Left-align ballot count when there's no header title

kofi-q:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/4964

Removing the title placeholder in the screen header when there's no title, so that the scanned ballot count falls to the left in those cases.

## Demo Video or Screenshot
| Single-Language  | Multi-Language |
| Screenshot 2024-08-22 at 12 34 29 | Screenshot 2024-08-22 at 12 34 08 |
| Screenshot 2024-08-22 at 12 34 38 | Screenshot 2024-08-22 at 12 34 45 |

## Testing Plan
- Manual

## Checklist

### DIFF
--- a/apps/scan/frontend/src/components/layout.tsx
+++ b/apps/scan/frontend/src/components/layout.tsx
@@ -138,7 +138,11 @@ export function Screen(props: ScreenProps): JSX.Element | null {
         </ButtonBar>
       )}
       <Header>
-        <TitleContainer>{title && <H1>{title}</H1>}</TitleContainer>
+        {title && (
+          <TitleContainer>
+            <H1>{title}</H1>
+          </TitleContainer>
+        )}
         {!hideBallotCountFromProps && ballotCount !== undefined && (
           <ScannedBallotCount count={ballotCount} />
         )}
--- a/apps/scan/frontend/src/components/layout.tsx
+++ b/apps/scan/frontend/src/components/layout.tsx
@@ -64,6 +64,10 @@ const ButtonBar = styled.div`
     height: 100%;
     width: 100%;
   }
+
+  & > :only-child {
+    grid-column: 1 / span 2;
+  }
 `;
 
 const Header = styled.div`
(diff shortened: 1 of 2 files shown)


arsalansufi:
cc @mattroe (who I know has had opinions on this for a while hah and might've been preferring a full-width button in this case)

kofi-q:
> cc @mattroe (who I know has had opinions on this for a while hah and might've been preferring a full-width button in this case)

Ah, that could work too - @mattroe, LMK if either of these is what you're thinking:

### Full-size settings, leaving ballot count right-aligned:
Screenshot 2024-08-20 at 17 07 50

### Full-size settings, changing ballot count to left-aligned when there's no screen title:
Screenshot 2024-08-20 at 17 08 26

mattroe:
My vote is full-width button right-aligned. I think it's valuable to *not* change the count location depending on state for poll workers. The whitespace also feels a bit weird to me keeping the same button width, but that may just be because I'm used to the space being utilized at this point.

carolinemodic:
@kofi-q I agree with keeping it right aligned too fwiw, should I wait to review this?

kofi-q:
> My vote is full-width button right-aligned.

Perfect - that's where I was leaning too. Glad we get to keep the right alignment consistent

kofi-q:
K, @carolinemodic, updated the code and screenshots - ready for review ✅

jonahkagan:
Would it be possible to have a regular width button on the left and the count on the right in the same row? I feel like the full width button is so long that it doesn't necessarily read as a button anymore.

mattroe:
> Would it be possible to have a regular width button on the left and the count on the right in the same row? I feel like the full width button is so long that it doesn't necessarily read as a button anymore.

That would be my preference - I just assumed it was a challenge to implement considering we didn't start there, but maybe bad assumption.

carolinemodic (approved):

kofi-q:
Ahh, gotcha - will iterate on that

kofi-q:
@carolinemodic - sorry for the thrash! Few more changes after slack design iteration - ready for one last sanity check

carolinemodic (approved):
no problem!

Issue #5263, opened 2024-08-15
VxScan: Scanner crashes on polls open for accuracy test definition


arsalansufi:
:bug: VxScan: Something went wrong upon opening polls for all bubble ballot (haven’t tried a different election since bumping to rc1) 

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1722975768922219?thread_ts=1722975768.922219&cid=CJU9MSC6S

carolinemodic:
does not repro when using a mock printer in dev

Issue #4835, opened 2024-05-08
Fujitsu Printer Integration: Rewrite Image Conversion in Rust


adghayes:
Image conversion is currently in TS and is very slow, resulting in the print pausing between the first and second print blocks of a page. If it were in `rust`, it would be faster, and printing would be visibly smoother. There are other solutions to this - we could wait to print the first block until the second block is processed - but better just to do this computationally intensive work in `rust`.

PR #5290, opened 2024-08-21
Break pages up when long enough on tally report

carolinemodic:
## Overview
Fixes https://github.com/votingworks/vxsuite/issues/5263

On the all bubble ballot the pdf we generate is so long that canvas seg faults when converting to an image. To avoid this problem we can just break up the page when it streches beyond 100 inches, which will never happen in anything but extreme situations. If the content height is longer then a standard paper length but shorter then 100 inches we will still truncate the length of the report to the length of the content (and we will never make it smaller then a standard page). 

Note this change should really only affect the all bubble ballot.

## Demo Video or Screenshot
PDF of printed all bubble polls open report 
[print-job-2024-08-21T00:36:50.587Z.pdf](https://github.com/user-attachments/files/16696348/print-job-2024-08-21T00.36.50.587Z.pdf)

Note the contest starts on the second page because of an unrelated annoyance with how tally reports render for this election. 

## Testing Plan
Used mock printer and inspected pdfs, ran tests and saw them still pass, added new test. 

## Checklist

### DIFF
--- a/libs/printing/src/render.tsx
+++ b/libs/printing/src/render.tsx
@@ -24,7 +24,7 @@ export const PAPER_DIMENSIONS = {
   // their own, followed by a mostly blank line. This causes stripes in the printed page.
   Bmd150: { width: 7.975, height: 13.25 },
   Letter: { width: 8.5, height: 11 },
-  LetterRoll: { width: 8.5, height: Infinity },
+  LetterRoll: { width: 8.5, height: 100 }, // If we make the height infinite the canvas conversion to an image can seg fault. Break into pages beyond 100inches.
 } satisfies Record<string, PaperDimensions>;
 
 export interface MarginDimensions {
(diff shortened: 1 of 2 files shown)


libs/printing/src/render.tsx
```diff
@@ -24,7 +24,7 @@ export const PAPER_DIMENSIONS = {
   // their own, followed by a mostly blank line. This causes stripes in the printed page.
   Bmd150: { width: 7.975, height: 13.25 },
   Letter: { width: 8.5, height: 11 },
-  LetterRoll: { width: 8.5, height: Infinity },
+  LetterRoll: { width: 8.5, height: 100 }, // If we make the height infinite the canvas conversion to an image can seg fault. Break into pages beyond 100inches.
```
adghayes, line 27: Space between "100" and "inches"

adghayes (approved):
LGTM! See one comment.

I'm glad we're fixing the all bubble ballot, but I do hope that we can eventually fix this with #4835 rather than building around `canvas`'s specific limitations.

PR #5292, opened 2024-08-21
Add high level logging to DB Client

carolinemodic:
## Overview
Adds basic logging when we connect to the db, create the db and destroy the db. I used the BaseLogger here since the workspace with the db needs to be created for the auth status polling used in the Logger class. This means that the logs will have the user "system" rather then the currently logged in user. I think this is okay / makes sense for low-level actions like creating and connecting to the database. The currently logged in user could be determined via context with other logs in the timeline. 

## Testing Plan
Ran apps, saw initial connection logs. 

## Checklist


kofi-q (approved):
🪵

PR #5284, opened 2024-08-20
feat(converter-nh-accuvote): add `correct-definition` command

eventualbuddha:
## Overview

Add another command to `converter-nh-accuvote`: `correct-definition`. This command takes pairs of XML and PDF and uses the bubble positions from the PDF to fix the `OX`/`OY` coordinates in the AccuVote definition in the same way our regular conversion does. It outputs the corrected AccuVote definition and a proofing PDF to manually verify the coordinates.

## Demo Video or Screenshot

```
❯ ./bin/correct-definition -h
Usage: correct-definition --config <config-path>

Corrects an AccuVote definition file based on the PDF.

Config File Format:

{
  "cards": [
    {
      // Required: path to the AccuVote definition file
      "definitionPath": "definition.xml",

      // Required: path to the ballot PDF file
      "pdfPath": "ballot.pdf",

      // Required: output directory
      "outputDir": "output"

      // Optional: page numbers (1-based, defaults to 1 & 2,
      // useful for PDFs containing multiple ballot cards)
      "frontPage": 3,
      "backPage": 4,
    }
  ]
}

All paths are relative to the config file or absolute.
```

| PROOF Front | PROOF Back |
|-|-|
| image | image |

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

Issue #5117, opened 2024-07-26
Productionize/generalize the changes in the September ballot creation branch for November ballot creation


arsalansufi:
@eventualbuddha and I made a number of changes to NH state ballot creation logic, all captured in some way or another on [this branch](https://github.com/votingworks/vxsuite/tree/arsalan/nh-ballot-creation-final).

Let's productionize/generalize the changes on that branch, and port them over to the [VxDesign v3.1.0 release branch](https://github.com/votingworks/vxsuite/tree/vxdesign-v3.1.0-release-branch) for November ballot creation.

There are also some scripts and notes that we added directly to the `election-definitions` repo [here](https://github.com/votingworks/election-definitions/tree/main/2024-09-10-nh-state-primary) that belong on the VxDesign v3.1.0 release branch as well.

One thing that we didn't tackle this go around is using polling places as precinct names, to avoid the wonky looking "Town, Town, NH" strings in the app footers. This data is available on the [NH SoS website](https://app.sos.nh.gov/statelistclerkandpolling). Let's see what else we can pull from that site that isn't available in the XML files. Another example that comes to mind is district names, which show up on final reports. We should cross-check NH SoS website data with the [data that we're collecting from towns ourselves](https://docs.google.com/spreadsheets/d/1Imb-OgQSi6mH9K4EV0PApWMuhtvW4hpB_o4SzNo9WsY/edit?usp=sharing).

Once we complete all of the above, let's update the [top-level checklist](https://docs.google.com/document/d/1LSTaOb_YxZ9MtOpDbEiPAP7XZhGaevGBFFb8jQbA9Bw/edit?usp=sharing).

arsalansufi:
Also important to consider: https://votingworks.slack.com/archives/CJU9MSC6S/p1724382115247439

PR #5304, opened 2024-08-23
[vxdesign-v3.1.0-release-branch] Improve ballot proofing and multi-card election support

eventualbuddha:
## Overview
Replaces https://github.com/votingworks/vxsuite/pull/5284 by incorporating the correction facility within the conversion process. Gets much of the way toward completing https://github.com/votingworks/vxsuite/issues/5117 but doesn't quite finish. Notably, however, we no longer need any special hacks to make multi-card elections (i.e. Rochester) work properly.

There's more we will likely want to do in a follow-up, such as making the generated IDs friendlier and more legible. The hashed suffix isn't really a problem, but the part before it should be meaningful.

## Demo Video or Screenshot

**PROOF Example**
image

**Test Deck Example**
image

## Testing Plan
Tested on all jurisdictions for the upcoming NH election (including multi-precinct Rochester). Tested on the Hudson ballot. Tested on the NH Test Ballot.


libs/converter-nh-accuvote/src/convert/convert_election_definition.ts
```diff
@@ -3,257 +3,208 @@ import {
...
-
-    const frontMetadata = frontGridAndBubbles.metadata;
     const ballotStyleParty = getPartyForBallotStyle({
       ballotStyleId: ballotStyle.id,
       election,
     });
-    const partyPrefix = ballotStyleParty ? `${ballotStyleParty.abbrev}-` : '';
-    const cardNumber =
-      frontMetadata?.side === 'front' ? frontMetadata.cardNumber : 1;
-    const ballotStyleId = `${partyPrefix}card-number-${cardNumber}`;
```
eventualbuddha, line 195: Removing this broke interpretation of timing mark ballots. The problem is that this logic is essentially duplicated in the ballot interpreter code. Both assume that for a given election only one ballot style will use a given card number, and that the ballot style ID can be generated knowing _only_ that card number. With the changes in this branch, that's no longer true. We'll need to decide what we want to do about this.

jonahkagan (approved):

libs/converter-nh-accuvote/src/convert/convert_election_definition.ts
jonahkagan: Could we use an incrementing index for card number when generating the ballot style ids?

eventualbuddha: We could, but it wouldn't help since they wouldn't necessarily correspond to the `cardNumber` from the timing mark-encoded metadata.

Issue #5214, opened 2024-08-05
VxSuite: Remove power down buttons from all machines that have physical power buttons AND remove restart buttons from "Something went wrong" screens


arsalansufi:
Let's remove the UI power down buttons from all machines that have physical power buttons. This means that only VxScan needs such a power down button.

We should also remove the restart buttons from "Something went wrong" screens as those buttons require the backend to be running and the backend may have (in fact likely) crashed. So in most cases, those buttons won't do anything.

Relevant Slack threads:
- https://votingworks.slack.com/archives/CJU9MSC6S/p1722453881850639?thread_ts=1722453881.850639&cid=CJU9MSC6S
- https://votingworks.slack.com/archives/CJU9MSC6S/p1722975786885349

PR #5307, opened 2024-08-26
Remove power down and restart buttons

jonahkagan:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/5214

- Remove power down button from all apps except  Every other app has a physical power button, so this on-screen button is unnecessary.
- Remove restart button from error boundary screen. The error boundary is mostly triggered by server crashes, in which case the restart button doesn't actually work (since it goes through the server).

## Testing Plan
Updated automated tests

## Checklist


adghayes (commented):
In the `app.tsx` pages, we now have two identical error boundaries, one inside the `ApiProvider` and one outside. I recommend removing the ones inside the `ApiProvider`s. I added those purely to have an error boundary with access to the API, for the purpose of rebooting, but now that's no longer necessary.

eventualbuddha (approved):

jonahkagan:
> In the `app.tsx` pages, we now have two identical error boundaries, one inside the `ApiProvider` and one outside. I recommend removing the ones inside the `ApiProvider`s. I added those purely to have an error boundary with access to the API, for the purpose of rebooting, but now that's no longer necessary.

Thanks for this context. I was wondering why there were two error boundaries. Will update

adghayes (approved):

Issue #5238, opened 2024-08-07
VxScan: V4 printing flow is broken for paused / resumed poll reports for primary elections


arsalansufi:
VxScan: When configured with a primary election, printing the “Voting Paused” report claims to print a report for a single party, but does not say anything about the party on the report. When you click “Print Next” to print the next party’s voting paused report, you get a “Something Went Wrong” screen

Issue automatically created from Slack feedback reported by jonah: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1723073635419579?thread_ts=1723073635.419579&cid=CJU9MSC6S

PR #5310, opened 2024-08-26
bug fix: fix primary election polls paused and resumed reports

adghayes:
## Overview

These are broken according to #5238.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/5ef99980-8a60-4d67-a2b5-bf3159440ace

## Testing Plan

Added new test to act as regression test, confirmed flow now works in dev.


jonahkagan (approved):
sus

PR #5314, opened 2024-08-27
feat(converter-nh-accuvote): add general and primary bubble layouts

eventualbuddha:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/5117

Adds two new strategies for matching ballot card definition contest options from the XML to the discovered bubble positions from the PDF:
- `PartyColumns`: for general elections. Assumes contests span the width of the ballot and all candidates for a party share a column. Uses only the definition's candidate order and not their coordinates.
- `ContestColumns`: for primary elections. Assumes contests are laid out in columns top to bottom and left to right. Uses only the definition's candidate order and not their coordinates.

These are now the default bubble layouts, replacing `RelativeSpacial` (which does use the definition's candidate coordinates) unless you specifically ask for it in the configuration. Using `ContestColumns` on the Conway fixture now matches everything up correctly, which wasn't the case with `RelativeSpacial`.

## Demo Video or Screenshot

| Test Ballot w/`PartyColumns` | Conway w/`ContestColumns` | Rochester Ward 3 (cropped) w/`ContestColumns` |
|-|-|-|
| image | image | image |

## Testing Plan

Added some unit tests. Tested on NH Test Ballot, Hudson, and Conway fixtures. Tested on the upcoming 9/10/24 NH election with _unedited XML files_.


jonahkagan (approved):
Looks great!

Issue #5193, opened 2024-08-02
VxScan: Speaker System Diagnostic


adghayes:
Press a button "Test Sound" which will play a chime "Playing sound..." and then display a modal for the user to say they heard it, or not. The sound could simply be our ballot accepted chime via the `useSound` utility. Potential modal copy:

**Sound Test**
Did you hear the sound played over the speakers?
Yes | No

The result creates a diagnostic record and a log entry. The diagnostic report now shows the result of the most recent sound test.

PR #5317, opened 2024-08-27
[Scan] Add audio diagnostic to system admin screen

kofi-q:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/5193

Adding a speaker/audio test and recent result to the diagnostics screen (and associated generated PDF).

https://github.com/user-attachments/assets/b5c62b9c-a831-4b3b-8cde-109868c6bbb6

readiness-report__000__2024-08-27_10-42-05

## Testing Plan
- New + updated unit tests
- Manual verification

## Checklist


kofi-q:
cc @adghayes for optional product review

adghayes (commented):
Nit: I'd prefer we name the section "Speaker" rather than "Sound Test," as I think it aligns better with the general pattern we've established of the header being the hardware (e.g. "Front Headphone Input") rather than the test, and "Sound test" is repeated on the next line.

Overall UX LGTM!

Did not review code.

kofi-q:
> Nit: I'd prefer we name the section "Speaker" rather than "Sound Test," as I think it aligns better with the general pattern we've established of the header being the hardware (e.g. "Front Headphone Input") rather than the test, and "Sound test" is repeated on the next line.

Ah, good point! Misread the task and used the modal copy instead - will update

apps/scan/frontend/src/components/audio_diagnostic_modal_button.tsx
```diff
@@ -0,0 +1,73 @@
...
+export interface AudioDiagnosticModalProps {
+  setVisible: (visible: boolean) => void;
+}
+
+export function AudioDiagnosticModal(
+  props: AudioDiagnosticModalProps
+): JSX.Element {
+  const { setVisible } = props;
+  const playSound = useSound('success');
+  React.useEffect(playSound, [playSound]);
```
adghayes, line 16: I've been under the impression that we're moving away from `useEffect` when possible, instead triggering things on user action (clicking of the button / opening of the modal). I truly do not care, but noting for the future / alignment across engineers.

apps/scan/frontend/src/components/audio_diagnostic_modal_button.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const { isLoading, mutate: logOutcome } =
+    api.logAudioDiagnosticOutcome.useMutation();
+
+  const onConfirm = React.useCallback(
+    (outcome: DiagnosticOutcome) => {
+      logOutcome({ outcome });
+      setVisible(false);
+    },
+    [logOutcome, setVisible]
+  );
```
adghayes, line 27: IMO it's not necessary to `useCallback` on these tiny functions because it's unclear if there's a performance gain and it sacrifices some readability.

adghayes (approved):
I just went ahead and reviewed the code - LGTM! All my comments are nits / curiosities and can be ignored.

apps/scan/frontend/src/components/audio_diagnostic_modal_button.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+    mockLogOutcome,
+    mockPlaySound,
+    render: (ui: React.ReactNode) =>
+      baseRender(
+        <TestErrorBoundary>
+          <ApiClientContext.Provider value={apiMock as ApiClient}>
+            <QueryClientProvider client={createQueryClient()}>
+              {ui}
+            </QueryClientProvider>
+          </ApiClientContext.Provider>
```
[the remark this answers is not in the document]
kofi-q: Can't even remember how I ended up here anymore, but I did try using `provideApi` initially and had some trouble. I think the issue was something else and I can revert back to that now

apps/scan/frontend/src/components/audio_diagnostic_modal_button.tsx
kofi-q: Ah, I hadn't hard that - worth a team discussion, because not sure I can get behind a blanket approach without the full context (there are a few patterns we use that are equivalent to "do something when this component loads", like our data fetching, or screen reader logic). +1 to being consistent across, though

apps/scan/frontend/src/components/audio_diagnostic_modal_button.tsx
kofi-q: Yeah, force of habit. Also hadn't considered that it might be a negative for readability, since I've gotten so used to it, so that's good perspective - will update.
Related to the previous comment on consistency, though - there could be some benefit to making the default pattern the one that adds cumulative value (we've already seen perf hits from lots of unnecessary re-renders when stuff changes at the top of the tree, which is more of a death-by-a-thousand-cuts sort of thing)

Issue #5006, opened 2024-07-01
Export CDF format for logs


carolinemodic:
Will be slightly tricky to figure out converting compressed older logs. There is a base implementation for converting a single file in libs/logging but it was disabled in https://github.com/votingworks/vxsuite/issues/4152

PR #5319, opened 2024-08-28
Allow CDF and Error-Only Log Exports

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/5006
https://github.com/votingworks/vxsuite/issues/5333
Adds the ability to export logs as an error-only report or as CDF. The CDF conversion code was pre-existing. This just adds the UI logic and the logic to uncompress/recompress old files in addition to the error-only filtering. 

On moderately large log files this can take a minute or so for the CDF conversion, the error filtering is instantaneous which I find surprising since both need to uncompress / parse json / recompress, so I may play around a bit and see if there is something dumb we are doing. Otherwise will test on larger data from Ben and on a real image after merging. 
## Demo Video or Screenshot

https://github.com/user-attachments/assets/5384cb26-290d-4da8-aceb-43daf1089569

## Testing Plan
Copied logs from a prod machine to the appropriate location in my VM, then exported logs in all formats and inspected the results. 

## Checklist

### DIFF
--- a/libs/backend/src/system_call/export_logs_to_usb.ts
+++ b/libs/backend/src/system_call/export_logs_to_usb.ts
@@ -1,28 +1,114 @@
-import { Result, err, ok } from '@votingworks/basics';
+import { Result, err, ok, throwIllegalValue } from '@votingworks/basics';
 import { UsbDrive } from '@votingworks/usb-drive';
 import * as fs from 'fs/promises';
 import { join } from 'path';
 
-import { LogEventId, Logger } from '@votingworks/logging';
+import {
+  LogEventId,
+  LogExportFormat,
+  Logger,
+  buildCdfLog,
+  filterErrorLogs,
+} from '@votingworks/logging';
+import { createReadStream, createWriteStream } from 'fs';
+import { pipeline } from 'stream/promises';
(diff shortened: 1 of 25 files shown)


libs/backend/src/system_call/export_logs_to_usb.ts
```diff
@@ -1,28 +1,114 @@
...
+
+  // Create CDF for all compressed vx-logs files
+  for (const file of files) {
+    if (file.match(COMPRESSED_VX_LOGS_NAME_REGEX)) {
+      const cdfFileName = file.replace('vx-logs', 'vx-logs.cdf');
+      await pipeline(
+        createReadStream(join(logDir, file)),
+        createGunzip(),
+        (inputStream: AsyncIterable<string>) =>
+          buildCdfLog(logger, inputStream, machineId, 'DEV'),
```
kofi-q, line 59: is `'DEV'` here meant to be `codeVersion` like above?

kofi-q (approved):

libs/backend/src/system_call/export_logs_to_usb.ts
carolinemodic: yes good catch

Issue #3966, opened 2023-09-13
VxScan: Proactive cleaning notification


mattroe:
https://votingworks.slack.com/archives/CEL6D3GAD/p1722264404949679?thread_ts=1722263246.452699&cid=CEL6D3GAD

PR #5326, opened 2024-08-28
VxScan: Show warning when scanner needs cleaning

jonahkagan:
## Overview

Closes: https://github.com/votingworks/vxsuite/issues/3966

Detects vertical streaks in scanned ballots, rejecting the ballots and showing a warning that the scanner needs to be cleaned.

The heuristic we're using to detect streaks is to look for columns of pixels with more than 75% black pixels that contain gaps of no more than 15 contiguous white pixels.

## Demo Video or Screenshot
Screenshot-VxScan-2024-08-28T18:00:07 690Z

scan-1_debug_vertical_streaks

## Testing Plan
- Added new test for streak detection
- Existing interp tests pass
- Manually tested with a corpus of scanned ballots that contained about 50% streaked ballots to tune the parameters. Ensured we caught most of the streaks but no false positives.

## Checklist


libs/ballot-interpreter/test/fixtures/nh-test-ballot/folded-corner-back.png
jonahkagan, line 1: I whited out a portion of the streaks at the top left so this image could still be used for the folded ballot test without triggering streak detection

libs/ballot-interpreter/src/hmpb-rust/image_utils.rs
```diff
@@ -278,6 +285,93 @@ pub fn find_scanned_document_inset(image: &GrayImage, threshold: u8) -> Option<I
...
+        if let Some(last_column) = combined_streaks.pop() {
+            let (last_x, _, _) = last_column;
+            let (x, _, _) = column;
+            if x - last_x != 1 {
+                combined_streaks.push(last_column);
+            }
+        }
+        combined_streaks.push(column);
+        combined_streaks
+    });
```
eventualbuddha, line 363: Minor suggestion for some improvements:

```suggestion
    let streak_columns = binarized_columns.filter_map(|(x, column)| {
        let num_black_pixels = column.iter().filter(|is_black| **is_black).count();
        let percent_black_pixels = num_black_pixels as f32 / height as f32;
        if percent_black_pixels < PERCENT_BLACK_PIXELS_IN_STREAK {
            return None;
        }

        let longest_white_gap_length = column
            .into_iter()
            .group_by(|is_black| *is_black)
            .into_iter()
            .filter(|(is_black, _)| !*is_black)
            .map(|(_, white_gap_pixels)| white_gap_pixels.count() as PixelUnit)
            .max()
            .unwrap_or(0);

        if longest_white_gap_length <= MAX_WHITE_GAP_PIXELS {
            Some((x, percent_black_pixels, longest_white_gap_length))
        } else {
            None
        }
    });

    // If there are adjacent streak columns, just pick one to represent the streak.
    let streaks = streak_columns
        .coalesce(|(x1, p1, l1), (x2, p2, l2)| {
            if x2 - x1 == 1 {
                Ok((x2, p2, l2))
            } else {
                Err(((x1, p1, l1), (x2, p2, l2)))
            }
        })
        .collect::<Vec<_>>();
```

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -313,6 +323,24 @@ pub fn interpret_ballot_card(
...
+        (SIDE_B_LABEL, &side_b, &mut side_b_debug),
+    ]
+    .par_iter_mut()
+    .map(|(label, side, debug)| {
+        let streaks = detect_vertical_streaks(&side.image, side.threshold, debug);
+        if streaks.is_empty() {
+            Ok(())
+        } else {
+            Err(Error::VerticalStreaksDetected {
+                label: (**label).to_string(),
```
eventualbuddha, line 337: I don't think the derefs are needed.
```suggestion
                label: label.to_string(),
```

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -313,6 +323,24 @@ pub fn interpret_ballot_card(
         None => ImageDebugWriter::disabled(),
     };
 
+    [
+        (SIDE_A_LABEL, &side_a, &mut side_a_debug),
+        (SIDE_B_LABEL, &side_b, &mut side_b_debug),
+    ]
+    .par_iter_mut()
+    .map(|(label, side, debug)| {
+        let streaks = detect_vertical_streaks(&side.image, side.threshold, debug);
```
eventualbuddha, line 332: Do we think this should be optional? I'm wondering whether there should be an escape hatch to turn this feature off in case of something we don't foresee causing false positives (maybe scanner image sensor failure?). I can also see the argument that if there really are issues with the image that match streaks that we really don't want to interpret them. A question for @mattroe and @benadida, probably.

libs/ballot-interpreter/src/interpret_vx_hmpb.test.ts
```diff
@@ -219,6 +219,38 @@ describe('HMPB - Famous Names', () => {
...
+  test('streaks on ballot', async () => {
+    const images = asSheet(await pdfToPageImages(blankBallotPath).toArray());
+    const [frontImage, backImage] = images;
+    const canvas = createCanvas(frontImage.width, frontImage.height);
+    const context = canvas.getContext('2d');
+    context.imageSmoothingEnabled = false;
+    context.putImageData(frontImage, 0, 0);
+    context.fillStyle = 'black';
+    context.fillRect(canvas.width / 2, 0, 1, canvas.height);
+    const streakImage = context.getImageData(0, 0, canvas.width, canvas.height);
```
eventualbuddha, line 232: This being a test the performance doesn't matter _that_ much, but if you want to directly edit the image data this is how I'd do it:

```suggestion
    const bytesPerRow = frontImage.width * 4;
    for (
      let offset = bytesPerRow / 2;
      offset < frontImage.data.length;
      offset += bytesPerRow
    ) {
      frontImage.data[offset] = 0;
      frontImage.data[offset + 1] = 0;
      frontImage.data[offset + 2] = 0;
      frontImage.data[offset + 3] = 255;
    }
```

eventualbuddha (approved):
Glad this came out as well as it did!

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
jonahkagan: I didn't have them originally but clippy requested them

libs/ballot-interpreter/src/interpret_vx_hmpb.test.ts
jonahkagan: Seems more complicated to me since you have to understand the image data format

libs/ballot-interpreter/src/hmpb-rust/image_utils.rs
jonahkagan: Nice! I had something similar to your approach for calculating the gap length, but it was pretty slow. Though I didn't know about `group_by`, so I had implemented my own version that required materializing the iterator into a vector, which slowed things down a lot. I didn't really expect that `group_by` would be the name for this logic, since I generally associate it with grouping the entire set, not just consecutive runs. Anyway, this version seems fast, yay!

`coalesce` is also exactly what I was looking for but couldn't find, thanks!

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
benadida: great question. Can we test and see how likely it is to get a false positive before we do extra work to make it togglable? I can be convinced either way easily.

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
jonahkagan: Made it a tad better: 5bf0b0b41

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
jonahkagan: It's a good question! I'm gonna go ahead and merge this PR and we can discuss in Slack. I'd prefer to implement an escape hatch for this as a follow up anyway

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
eventualbuddha: I'm surprised that clippy isn't complaining for me. What's your Rust version? Here's mine:

```
❯ rustc --version
rustc 1.80.0 (051478957 2024-07-21)
```

libs/ballot-interpreter/src/hmpb-rust/image_utils.rs
eventualbuddha: I hear you on `group_by`–I had the same thought. 😬

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
jonahkagan: Same:
```
$ rustc --version
rustc 1.80.0 (051478957 2024-07-21)
```

eventualbuddha: And now I do see it 🤷 

```
error: calling `to_string` on `&&str`
   --> libs/ballot-interpreter/src/hmpb-rust/interpret.rs:337:24
    |
337 |                 label: label.to_string(),
    |                        ^^^^^^^^^^^^^^^^^ help: try dereferencing the receiver: `(*label).to_string()`
    |
    = help: `&str` implements `ToString` through a slower blanket impl, but `str` has a fast specialization of `ToString`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string
```

PR #5331, opened 2024-08-28
docs(basics): add an example of using `lines` with `stream.pipeline`

eventualbuddha:
Also adds a test to ensure it really works.


carolinemodic (approved):
oh perfect! ty!

Issue #5278, opened 2024-08-19
VxAdmin: Large tally reports, with lots of groups, fail to generate


arsalansufi:
With a large election, the “All Precincts Tally Report” fails to generate (spins for a while, then goes to “something went wrong”)

Log output:

```
[backend:run] {"source":"vx-admin-service","eventId":"election-report-previewed","eventType":"user-action","user":"election_manager","message":"User previewed a tally report.","disposition":"success"}
[backend:run] page.setContent: Target page, context or browser has been closed
[backend:run] Call log:
[backend:run]   - setting frame content, waiting until "load"
[backend:run] 
[backend:run]     at renderToPdf (/home/jonahkagan/code/vxsuite/libs/printing/build/render.js:87:20)
[backend:run]     at async generateTallyReportPreview (/home/jonahkagan/code/vxsuite/apps/admin/backend/build/reports/tally_report.js:80:14)
[backend:run]     at /home/jonahkagan/code/vxsuite/apps/admin/backend/async /home/jonahkagan/code/vxsuite/libs/grout/build/server.js:92:32
```

Issue automatically created from Slack feedback reported by jonah: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1723747765653159?thread_ts=1723747765.653159&cid=CJU9MSC6S

----

This issue seems to affect any report generation involving a group-by on a field with many distinct values. We can try to fix this, but just blocking report generation for unreasonably large group-by's is fine, too. We can record this as one of our system limits.

[Another example here](https://docs.google.com/document/d/1uGyi0L66Vdj2XltwteUpJLI48TKLdmtCbMs-NWgPyec/edit?usp=sharing):

> VxAdmin something went wrongs when reporting (tally report) grouped by batch when there are an insane (10,000) number of batches. Ballot count report grouped by batch succeeds after several minutes

PR #5343, opened 2024-08-29
VxAdmin: Disable PDF preview/export for large reports

jonahkagan:
## Overview

Fixes: https://github.com/votingworks/vxsuite/issues/5278

When report PDFs are very very long (~1000 pages), Chromium will crash when trying to render them. In addition, the frontend PDF preview can crash at ~100 pages.

To prevent this, we set two limits:
- A limit on the HTML content length that we'll try to render as a PDF
- A limit on the page length for PDF previews

This PR also fixes two small bugs:
- Don't render empty reports
- Reset page count when PDF data changes

## Demo Video or Screenshot

https://github.com/user-attachments/assets/116bcb9f-9d01-425c-961b-d1b91e7fcf1f

## Testing Plan
To set the limits, I tested a variety of content lengths and chose conservative numbers. I also updated automated tests

## Checklist

### DIFF
--- a/apps/admin/backend/src/reports/ballot_count_report.ts
+++ b/apps/admin/backend/src/reports/ballot_count_report.ts
@@ -75,8 +75,8 @@ type BallotCountReportPreviewProps = BallotCountReportSpec & {
  * PDF data for a ballot count report alongside any potential warnings.
  */
 export interface BallotCountReportPreview {
-  pdf: Buffer;
-  warning: BallotCountReportWarning;
+  pdf?: Buffer;
+  warning?: BallotCountReportWarning;
 }
 
 /**
@@ -93,9 +93,12 @@ export async function generateBallotCountReportPreview({
     message: `User previewed a ballot count report.`,
     disposition: 'success',
   });
+  const pdf = await renderToPdf({ document: report });
   return {
-    pdf: await renderToPdf({ document: report }),
-    warning: getBallotCountReportWarning(reportProps),
+    pdf: pdf.ok(),
+    warning: pdf.isErr()
+      ? { type: pdf.err() }
+      : getBallotCountReportWarning(reportProps),
   };
 }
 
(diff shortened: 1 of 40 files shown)


apps/admin/backend/src/reports/ballot_count_report.ts
```diff
@@ -88,14 +88,19 @@ export async function generateBallotCountReportPreview({
 
   ...reportProps
 }: BallotCountReportPreviewProps): Promise<BallotCountReportPreview> {
-  const report = buildBallotCountReport(reportProps);
   await logger.logAsCurrentRole(LogEventId.ElectionReportPreviewed, {
     message: `User previewed a ballot count report.`,
     disposition: 'success',
```
carolinemodic, line 93: is it possible to make this log as a failure if there is an error in getting the report preview?

apps/admin/backend/src/reports/tally_report.ts
```diff
@@ -121,12 +120,18 @@ export async function generateTallyReportPreview({
     message: `User previewed a tally report.`,
     disposition: 'success',
```
carolinemodic, line 121: same logging comment

apps/admin/backend/src/reports/write_in_adjudication_report.ts
```diff
@@ -41,20 +41,36 @@ interface WriteInAdjudicationReportPreviewProps {
...
 export async function generateWriteInAdjudicationReportPreview({
   logger,
 
   ...reportProps
-}: WriteInAdjudicationReportPreviewProps): Promise<Buffer> {
+}: WriteInAdjudicationReportPreviewProps): Promise<WriteInAdjudicationReportPreview> {
   const report = buildWriteInAdjudicationReport(reportProps);
   await logger.logAsCurrentRole(LogEventId.ElectionReportPreviewed, {
     message: `User previewed the write-in adjudication report.`,
     disposition: 'success',
```
carolinemodic, line 67: failure logging

carolinemodic (approved):
In a few places you are added a potential error condition after a "success" log occurs so it would be better to move the log below there and have it log either a "success" or "failure" as appropriate. Otherwise, lgtm.

apps/admin/backend/src/reports/ballot_count_report.ts
jonahkagan: good call, thanks 40d199cfa

Issue #4716, opened 2024-03-22
VxAdmin: Import ERR CDF


mattroe:
Thread reference: https://votingworks.slack.com/archives/CE7LM9LQ2/p1711126552821739

We need to add some ability to import ERR CDF. If we allow generically importing ERR CDF alongside CVRs, we encounter a lot of trouble. First of all, what cut does the ERR CDF represent? If it's only whole election results, then we can no longer generate many of our report cuts and would have to just disable them. Also, we would have to build a whole new representation of the ERR results on the backend. Even if we disable uploading CVRs at the same time, we still would have to suss out the structure / cuts of the ERR data and toggle behavior accordingly unless we just always assume it's full election results. Our manual results, however, already exist, and were designed to address some of these concerns. Manual results require you to enter the ballot style, precinct, and ballot type ahead of time so we work it into our cuts. As a result, we can shoehorn the ERR CDF import into our manual results flow and avoid the complexity of an entirely new source of results. 

UX-wise, here's what this could look like:
- On the manual tally summary page, in addition to the "Add Tallies" button there would also be an "Upload Tallies" button. "Add Tallies" could be renamed "Enter Tallies." When you click the "Upload Tallies" button, a modal pops up describing the import requirements (ERR). You can select a file.
- The backend attempts to parse the file and store as a manual results blob on the backend. We'll have to start indicating which manual results are from upload and which are not. Parsing failure is surfaced to the user.
- The summary page now shows the uploaded data in the list. It has a remove button but, unlike the manually entered tallies, it has a "View" button instead of an "Edit" button. 
- If you click "View," you're taken to the manual tally entry page but all form entry is disabled. 

In this way, we can add a meaningful (if niche) ERR import functionality without touching the core tally logic this close to certification.

mattroe:
@adghayes to flesh this out with more specific direction

adghayes:
@mattroe @kshen0 I updated the description with my vision for this feature - please let me know if you have suggested tweaks or concerns.

PR #5344, opened 2024-08-30
ERR CDF import

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/4716

Adds basic support for ERR file import

No support for write-ins yet.

## Demo Video or Screenshot

Screenshots with updated styles and copy.

Screenshot 2024-09-04 at 8 41 02 PM

Screenshot 2024-09-04 at 8 41 10 PM

Screenshot 2024-09-04 at 8 41 23 PM

Screenshot 2024-09-04 at 8 41 28 PM
\
This video shows the flow, but with old copy and styles.

https://github.com/user-attachments/assets/522740b2-51bc-496e-ab3a-d34d16c897ae

## Testing Plan
- added automated tests
- manually tested flow in VM

## Checklist


libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -0,0 +1,269 @@
...
+} from '.';
+import {
+  isBallotMeasureContest,
+  isCandidateContest,
+  isRetentionContest,
+} from './types';
+
+type CandidateNameRecord = Record<CandidateId, Candidate['name']>;
+
+function assertGetDeepValue(root: unknown, valuePath: string) {
```
kshen0, line 22: This is used somewhat widely because many fields in the ERR schema are not technically required and therefore are optional. In many of these cases, the file would be useless without such fields. For example, a `CandidateSelection` has multiplicity `0..*` for the `Candidate` prop. Votes described by a `CandidateSelection` can't be mapped to a candidate without that prop, so we `assertGetDeepValue` rather than eg. use and handle a `Result`.

libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  return records;
+}
+/**
+ * Converts an ElectionReport to an instance of ManualElectionResults.
+ * @param electionReport
+ * @returns an instance of ManualElectionResults.
+ */
+export function getManualResultsFromErrElectionResults(
```
jonahkagan, line 285: nit: maybe use "convert" in the name of this function

libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -0,0 +1,269 @@
...
+} from './types';
+
+type CandidateNameRecord = Record<CandidateId, Candidate['name']>;
+
+function assertGetDeepValue(root: unknown, valuePath: string) {
+  return assertDefined(
+    getDeepValue(root, valuePath),
+    `Couldn't get value: ${valuePath}`
+  );
+}
```
jonahkagan, line 33: I'm a bit worried that using this `getDeepValue` function bypasses type checking, which may make it easier to make a mistake if we ever change this code. Do you feel like there's a tradeoff that makes it worthwhile?

libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -0,0 +1,269 @@
...
+
+function getBallotCounts(contest: ResultsReporting.CandidateContest): {
+  overvotes: number;
+  undervotes: number;
+  total: number;
+} {
+  const otherCounts = contest.OtherCounts && contest.OtherCounts[0];
+
+  let ballotCount =
+    0 + (otherCounts?.Overvotes || 0) + (otherCounts?.Undervotes || 0);
```
jonahkagan, line 160: Are we sure that overvotes and undervotes represent individual ballots? They may represent a number of bubbles. E.g. undervotes may count each bubble that was not filled in. This is at least how we did things in our CVR CDF implementation. cc @adghayes

I don't see anything in the documentation about how to interpret this, but looking at our validation for manually entered tallies, it seems like we treat the overvotes/undervotes as vote counts, not ballot counts: https://github.com/votingworks/vxsuite/blob/main/apps/admin/frontend/src/screens/manual_data_entry_screen.tsx#L284-L304

I think the equivalency implied by that code is:
```
total ballots * votes allowed = candidate votes + overvotes + undervotes
```
Which would mean in this case, we should use this equation:
```
total ballots = (candidate votes + overvotes + undervotes) / votes allowed
```

libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -0,0 +1,269 @@
...
+    contest.ContestSelection
+  ) as ResultsReporting.BallotMeasureSelection[];
+  const yesOption = findBallotMeasureSelectionWithContent(
+    /yes/i,
+    contestSelections
+  );
+  const noOption = findBallotMeasureSelectionWithContent(
+    /no/i,
+    contestSelections
+  );
```
jonahkagan, line 117: I don't think there's necessarily a guarantee that these contest options will have "yes" or "no" in them. It might be safer to just use the order of the contest options (assign the first to yes, the second to no). There's not really any meaning to which one is assigned yes vs no in our internal schema.

libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -0,0 +1,269 @@
...
+}
+
+function findTotalVoteCounts(
+  voteCounts: readonly ResultsReporting.VoteCounts[]
+): number {
+  return assertDefined(
+    voteCounts.find((vc) => vc.Type === CountItemType.Total),
+    'Could not find total vote count'
+  ).Count;
```
jonahkagan, line 91: consider using `find` from `libs/basics`, also above on L55

apps/admin/frontend/src/components/import_err_file_modal.tsx
```diff
@@ -0,0 +1,190 @@
...
+        onOverlayClick={onClose}
+        actions={<Button onPress={onClose}>Close</Button>}
+      />
+    );
+  }
+
+  if (currentState.state === 'success') {
+    return (
+      <Modal
+        title="ERR File Added"
```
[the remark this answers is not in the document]
jonahkagan: And maybe remove the "Success" copy from the `content` field

apps/admin/frontend/src/screens/manual_data_summary_screen.tsx
jonahkagan, line 1: I saw these notes in the original task:
* The summary page now shows the uploaded data in the list. It has a remove button but, unlike the manually entered tallies, it has a "View" button instead of an "Edit" button.
* If you click "View," you're taken to the manual tally entry page but all form entry is disabled.

Did you decide to change it so that you can edit the imported results? Or should we change it to read-only like @adghayes originally suggested?

apps/admin/backend/src/app.ts
```diff
@@ -701,6 +703,7 @@ function buildApi({
       }
     ): Promise<void> {
       const electionId = loadCurrentElectionIdOrThrow(workspace);
+      debug(JSON.stringify(input.manualResults, null, 2));
```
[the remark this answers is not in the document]
kshen0: Thanks for catching. Deleting

apps/admin/backend/src/tabulation/election_results_reporting.ts
```diff
@@ -0,0 +1,44 @@
+import { Result, err } from '@votingworks/basics';
+import { readFile, ReadFileError } from '@votingworks/fs';
+import { LogEventId, Logger } from '@votingworks/logging';
+import { safeParseJson, ResultsReporting } from '@votingworks/types';
+import z from 'zod';
+
+const MAX_ERR_FILE_SIZE_BYTES = 5 * 1024 * 1024; // 5 MB
```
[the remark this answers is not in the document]
kshen0: Agreed, terminology is unfortunately _very_ overloaded in this context. `Result(s)` can refer to both the `Result<Ok, Err>` construct or election results, and ERR can refer to either errors or Election Results Reporting. For the most part I tried to disambiguate by spelling out "Election Results Reporting" unless it made variables and methods too long. Might be worth just always spelling it out though.

apps/admin/frontend/src/screens/manual_data_summary_screen.tsx
adghayes: @jonahkagan we decided to skip the read-only requirement, I forgot to update the description. Even though we agreed that it made sense from a product perspective, we wanted to descope as much as possible considering that ERR I/O might exist in a totally different part of the product (i.e. not wedded to manual results) in the future.

libs/types/src/cdf/election-results-reporting/convert.ts
adghayes: +1 to @jonahkagan's second formula

apps/admin/frontend/src/components/import_err_file_modal.tsx
```diff
@@ -0,0 +1,190 @@
...
+          <React.Fragment>
+            {window.kiosk && (
+              <FileInputButton
+                data-testid="manual-input"
+                onChange={processElectionResultReportingFileFromFilePicker}
+                accept=".json"
+              >
+                Select Import…
+              </FileInputButton>
+            )}
```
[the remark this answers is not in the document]
adghayes: I'd say only if it's very easy, since this feature doesn't need polish _and_ even with the USB inserted the file picker UX is still bad (you have to select the bookmark on the side with the low-level name of the USB drive like `usb-drive-1` or something).

apps/admin/frontend/src/components/import_err_file_modal.tsx
```diff
@@ -0,0 +1,190 @@
...
+            <Button onPress={onClose}>Cancel</Button>
+          </React.Fragment>
+        }
+      />
+    );
+  }
+
+  if (usbDriveStatus.status === 'mounted') {
+    return (
+      <Modal
```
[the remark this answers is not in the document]
adghayes: My fear with adding this button where it is (and I do think it is in the right place) is that someone expects it to be more useful than it actually is. They have a file in some format (maybe even a scanned PDF) and expect us to automagically import it.

I am in support of keeping the intermediate modal to clue in the user to the fact that they would have to be using a very specific file type, which they almost certainly don't have, to make this feature work. We will put it in the docs but if we can be explicit about it in the product too I think that could save a headache down the road. So for that landing modal, I propose -

Title: Import Results File (matches the button title @jonahkagan suggested)
Content: Results may be imported as an Election Results Reporting Common Data Format (ERR CDF) file. Choose an ERR CDF file to import.

Outside of this reference to the file format itself, I'd prefer to use more generic terms than "ERR" or "Election Results Reporting" as those are not universally used terms, like "Results" or "Results File" per other @jonahkagan comments.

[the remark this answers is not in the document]
jonahkagan: Sounds good to me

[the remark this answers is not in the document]
kshen0: In the current implementation it's just a convenient place to put status messages. If we got rid of the intermediate modal we'd have to put them on the main page.

apps/admin/frontend/src/screens/manual_data_summary_screen.tsx
```diff
@@ -305,28 +331,49 @@ export function ManualDataSummaryScreen(): JSX.Element {
...
-                  Add Tallies
-                </LinkButton>
+                <ButtonBar>
+                  <LinkButton icon="Add" disabled>
+                    Enter Tallies
+                  </LinkButton>
+                  <LinkButton icon="Import" variant="secondary" disabled>
+                    Upload Tallies
+                  </LinkButton>
+                </ButtonBar>
```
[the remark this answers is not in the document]
kshen0: Seems like the ternary was there to allow dereference of `selectedBallotStyle` et al. I moved the ternary down to where we dereference to simplify and make that more clear.

libs/types/src/cdf/election-results-reporting/convert.ts
kshen0: I also didn't like the bypassed type checking. Even losing tab completion when adding this feature it was a little annoying.

If we don't use `getDeepValue` then we would use optional chaining `?`. We'd need to handle `undefined` where realistically we'd never see that specific prop be `undefined` eg. `Candidate` on a `CandidateSelection` object. We'd also need to add coverage for those unrealistic `undefined` cases or ignore lines with optional chaining from coverage.

If we wanted to keep type safety and coverage, we could break out the existence checks to avoid `?` altogether. We'd get those 2 benefits at the cost of lots of boilerplate in the code.

IMO losing type safety in this context isn't _as_ bad as normal because we only use `getDeepValue` on objects in the ERR schema, and that schema is unlikely to change. But agreed it's still not ideal.

I added `assertGetDeepValue` pretty early in development so I'm not sure how messy the code will get if we switch to optional chaining, but I can experiment with it to see which feels better.

libs/types/src/cdf/election-results-reporting/convert.ts
kshen0: Is there a strong convention that yes is always first and no is always second? I'm really hesitant to enforce a convention like that with a file format that is so flexible.

By explicitly looking for "yes" and "no", and throwing if no matches are found, the worst that happens is election staff need to manually enter results.

By enforcing ordering, the worst that happens is we accidentally flip the results of a contest.

apps/admin/frontend/src/screens/manual_data_summary_screen.tsx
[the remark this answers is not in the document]
jonahkagan: Ohh, gotcha. That makes sense

libs/types/src/cdf/election-results-reporting/convert.ts
jonahkagan: > We'd need to handle undefined where realistically we'd never see that specific prop be undefined
In this case, maybe `assertDefined` instead of optional chaining?

libs/types/src/cdf/election-results-reporting/convert.ts
jonahkagan: Hmm I didn't think about it needing to match the existing contest options in the election definition. I think this is a tricky problem, because it's very possible to give the options labels that don't have yes/no in them. Maybe we do a hybrid approach where we look for yes/no to cover the common cases, but fall back to order if not? In my experience, it's common for option order to be pretty reliable, especially for ballot measures.

adghayes: Out of fear for an extreme edge case where we flip contest results, I support @kshen0's suggestion of looking for "yes" and "no" and, if they are not found, failing the import.

jonahkagan: Ok, I guess the mitigation if the contest options had labels without yes/no in them would be to edit the ERR file and replace those labels with yes/no, so I can get behind that.

apps/admin/backend/src/app.ts
```diff
@@ -749,6 +752,63 @@ function buildApi({
...
+      }
+    ): Promise<Result<void, Error>> {
+      const parseResult = await parseElectionResultsReportingFile(
+        input.filepath,
+        logger
+      );
+
+      if (parseResult.isErr()) {
+        // Logging is handled by parseElectionResultsReportingFile
+        return err(new Error('Could not parse file'));
```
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/5344/commits/630406a50889e000b485b8cfb4055b650e3f3196

apps/admin/backend/src/app.ts
```diff
@@ -749,6 +752,63 @@ function buildApi({
...
+        await logger.logAsCurrentRole(LogEventId.ParseError, {
+          message: 'Error converting ERR file to VX format',
+          error: wrappedManualResults.err().message,
+        });
+        return err(new Error('Could not convert ERR file'));
+      }
+
+      const manualResults = wrappedManualResults.ok();
+
+      await store.withTransaction(() => {
```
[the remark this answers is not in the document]
kshen0: This pattern is copied from the existing `setManualResults` [here](https://github.com/votingworks/vxsuite/blob/cb4bcce897209b7f94d97a8a42c702319e5fb09f/apps/admin/backend/src/app.ts#L704-L718). Both `handleEnteredWriteInCandidateData` and `setManualResults` write to the store so I'm guessing that's why the caller is responsible for handling the transaction.

When write-in support is added for ERR import we'll want to factor out as much as possible and move the transaction handling down to the store layer. Agreed we also should look into whether this needs to be async. I'll have a better idea of how feasible that is once we start that work, so I think it makes sense to leave this as is for now.

libs/types/src/cdf/election-results-reporting/convert.ts
kshen0: https://github.com/votingworks/vxsuite/pull/5344/commits/50142fcac09c7a7fb3ddb895c441740401d3c164

libs/types/src/cdf/election-results-reporting/convert.ts
kshen0: https://github.com/votingworks/vxsuite/pull/5344/commits/56fdcc88c10c5ca012721f8db07f3f4a0fc42721

libs/types/src/cdf/election-results-reporting/convert.ts
kshen0: https://github.com/votingworks/vxsuite/pull/5344/commits/65424132fcfc26c83553d11402f721cb46e9e4d4

libs/types/src/cdf/election-results-reporting/convert.ts
kshen0: https://github.com/votingworks/vxsuite/pull/5344/commits/676d679685eceab7daeb635ff5782050af8a11d1

apps/admin/frontend/src/screens/manual_data_summary_screen.tsx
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/5344/files#diff-091e628ff6b2cb050f622e33dbf23113f49961f41f9f0f97e526b5908a80d6a3R341

apps/admin/frontend/src/screens/manual_data_summary_screen.tsx
```diff
@@ -305,28 +331,49 @@ export function ManualDataSummaryScreen(): JSX.Element {
...
-                  variant="primary"
-                  to={routerPaths.manualDataEntry({
-                    ballotStyleId: selectedBallotStyle.id,
-                    precinctId: selectedPrecinct.id,
-                    votingMethod: selectedVotingMethod,
-                  })}
-                >
-                  Add Tallies
-                </LinkButton>
+                <ButtonBar>
```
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/5344/commits/d1c283ca9b32c0d508b4bdd80b77433e26fed6b6

apps/admin/frontend/src/components/import_err_file_modal.tsx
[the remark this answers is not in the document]
kshen0: The page was already structured in a way that made disabling easy:

https://github.com/votingworks/vxsuite/pull/5344/files#diff-965242a7809fefa3e2a8ae22d9e69c4046cecf247220b6e03f9d6b772641eaddR156

apps/admin/frontend/src/components/import_err_file_modal.tsx
```diff
@@ -0,0 +1,190 @@
...
+    usbDriveStatus.status === 'no_drive' ||
+    usbDriveStatus.status === 'ejected' ||
+    usbDriveStatus.status === 'error'
+  ) {
+    return (
+      <Modal
+        title="No USB Drive Detected"
+        content={
+          <P>
+            <UsbImage src="/assets/usb-drive.svg" alt="Insert USB Image" />
```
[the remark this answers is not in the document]
kshen0: https://github.com/votingworks/vxsuite/pull/5344/commits/fe75049be36766b5fd8c24af646fd36019b98fbc

apps/admin/frontend/src/components/import_err_file_modal.tsx
```diff
@@ -0,0 +1,190 @@
...
+}: Props): JSX.Element | null {
+  const { usbDriveStatus, electionDefinition, auth } = useContext(AppContext);
+  const importElectionResultReportingFileMutation =
+    importElectionResultsReportingFile.useMutation();
+
+  assert(electionDefinition);
+  assert(isElectionManagerAuth(auth) || isSystemAdministratorAuth(auth));
+  const [currentState, setCurrentState] = useState<ModalState>({
+    state: 'init',
+  });
```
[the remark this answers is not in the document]
kshen0: Missed this when forking and paring down a different file import modal. Thanks for catching.

https://github.com/votingworks/vxsuite/pull/5344/commits/fe75049be36766b5fd8c24af646fd36019b98fbc#diff-e2202cb4b116fe8ab041c171383789cd0499f02eaa45d2e5b3859871a95c01e7L35

apps/admin/frontend/src/components/import_err_file_modal.tsx
```diff
@@ -57,62 +45,59 @@ export function ImportElectionsResultReportingFileModal({
...
+        content={<P>{message}</P>}
         onOverlayClick={onClose}
         actions={<Button onPress={onClose}>Close</Button>}
       />
     );
   }
 
-  if (currentState.state === 'success') {
+  /* istanbul ignore next */
+  if (importElectionResultReportingFileMutation.isError) {
```
jonahkagan: Will this error just get caught by the error boundary due to the react-query default settings?

apps/admin/frontend/src/components/import_err_file_modal.tsx
```diff
@@ -26,6 +29,20 @@ const Content = styled.div`
...
+  errorCode: ImportElectionResultsReportingError
+): string {
+  switch (errorCode.type) {
+    case 'parsing-failed':
+      return 'The results file could not be parsed.';
+    case 'conversion-failed':
+      return 'The contents of the file could not be converted.';
+    /* istanbul ignore next - compile time check */
+    default:
+      return 'An error occurred.';
```
jonahkagan: should we have `throwIllegalValue(errorCode)` here?

apps/admin/frontend/src/components/import_err_file_modal.tsx
```diff
@@ -26,6 +29,20 @@ const Content = styled.div`
...
 `;
 
+function errorCodeToMessage(
+  errorCode: ImportElectionResultsReportingError
+): string {
+  switch (errorCode.type) {
+    case 'parsing-failed':
+      return 'The results file could not be parsed.';
+    case 'conversion-failed':
+      return 'The contents of the file could not be converted.';
```
jonahkagan: I'm a little worried these error messages don't recommend any course of action for the user. I'm not totally sure how I would improve them though. What should they do if either of these errors occurs?

libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -154,19 +154,24 @@ function getBallotCounts(contest: ResultsReporting.CandidateContest): {
...
     const totalCount = findTotalVoteCounts(voteCounts);
-    ballotCount += totalCount;
+    totalCandidateVotes += totalCount;
   }
 
+  const totalBallots =
+    (totalCandidateVotes +
+      (otherCounts?.Overvotes || 0) +
+      (otherCounts?.Undervotes || 0)) /
+    contest.VotesAllowed;
```
jonahkagan, line 155: Should we validate that these divide evenly? Or do you think it's safe to assume that the math works out?

apps/admin/frontend/src/components/import_election_results_reporting_file_modal.tsx
```diff
@@ -0,0 +1,200 @@
...
+        }
+        onOverlayClick={onClose}
+        actions={
+          <React.Fragment>
+            <FileInputButton
+              data-testid="manual-input"
+              onChange={processElectionResultReportingFileFromFilePicker}
+              accept=".json"
+            >
+              Select Import…
```
jonahkagan, line 189: Copy suggestion: "Select File…" (here and above)

jonahkagan (approved):
Thanks for working through all that feedback! It's looking good. Just a few final notes

apps/admin/frontend/src/components/import_err_file_modal.tsx
kshen0: Good point - [a6cc1c0](https://github.com/votingworks/vxsuite/pull/5344/commits/a6cc1c0da44cf8db044956ea99b016705c37d632)

apps/admin/frontend/src/components/import_err_file_modal.tsx
kshen0: [a6cc1c0](https://github.com/votingworks/vxsuite/pull/5344/commits/a6cc1c0da44cf8db044956ea99b016705c37d632)

apps/admin/frontend/src/components/import_err_file_modal.tsx
kshen0: Added some context in [a6cc1c0](https://github.com/votingworks/vxsuite/pull/5344/commits/a6cc1c0da44cf8db044956ea99b016705c37d632)

We can't make assumptions about where the ERR files are coming from so our help text unfortunately has to be a little vague. The text "Please re-export the results file from your system and import in VxAdmin again" would address file corruption or eg. pulling the USB drive out before writing finishes. It won't help with misconfiguration of or bugs in the exporting system, but without more information we can't really offer much advice.

libs/types/src/cdf/election-results-reporting/convert.ts
kshen0: [9498992](https://github.com/votingworks/vxsuite/pull/5344/commits/94989928be322c2f4b7e6a7acf61e875e527db6b)

apps/admin/frontend/src/components/import_err_file_modal.tsx
kshen0: Also worth noting that the difference between "could not be parsed" and "could not be converted" have different technical implications, but the action for the user is the same. Parsing error == could not read file or file contents could not be parsed as JSON. Conversion error == file contents are valid JSON but did not match the ERR schema.

For conversion errors we could tell the user to look at the file and ensure it's a valid ERR file, but I'm not sure if we expect users to have fluency with the ERR format.

jonahkagan: That makes me think for parsing errors we could say something like: "File is unreadable. Try exporting it again." and for conversion errors we could say something like "File is not a valid Election Results Reporting CDF file. Please ensure you are using the correct file format."

Issue #5334, opened 2024-08-29
Fix bug with logs from rust code


carolinemodic:
Logs from rust (specifically the pat input daemon but maybe other places) are logging with hyphenated keys like event-id, event-type etc. instead of camelcase eventId, eventType. This breaks CDF conversion. 

See: https://votingworks.slack.com/archives/CEL6D3GAD/p1724883987916719

PR #5345, opened 2024-08-30
camelcase log keys

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/5334

Uses camelcase for log event keys when logging with Rust.

## Demo Video or Screenshot
n/a

## Testing Plan
Ran Rust daemon and manually confirmed camelcase output.

```
{"source":"vx-mark-scan-fai-100-controller-daemon","eventId":"create-virtual-uinput-device-init","message":"","eventType":"system-action","user":"system","disposition":"n/a"}
```

### DIFF
--- a/libs/logging/types-rust/src/types.rs
+++ b/libs/logging/types-rust/src/types.rs
@@ -44,10 +44,10 @@ pub enum User {
 #[serde(deny_unknown_fields)]
 pub struct Log {
     pub source: String,
-    #[serde(rename = "event-id")]
+    #[serde(rename = "eventId")]
     pub event_id: EventId,
     pub message: String,
-    #[serde(rename = "event-type")]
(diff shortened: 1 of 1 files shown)


libs/logging/types-rust/src/types.rs
```diff
@@ -44,10 +44,10 @@ pub enum User {
 #[serde(deny_unknown_fields)]
 pub struct Log {
     pub source: String,
-    #[serde(rename = "event-id")]
+    #[serde(rename = "eventId")]
```
eventualbuddha, line 47: Rather than doing this, you should apply `#[serde(rename_all = "camelCase")]` to the `Log` struct.

eventualbuddha (approved):

Issue #5342, opened 2024-08-29
Scan - Make sure we log sheet result as expected


carolinemodic:
Log on reject/accept user action for adjudication ballots. Make sure we have a general success casting log.

PR #5355, opened 2024-09-04
Improve state machine logging

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/5342
Improvements after audit state machine logging in mark and scan. We were unneccesarily downgrading the logger to BaseLogger in the function calls so I fixed that in both. I also am logging new states on the state transition logs as an explicit key in the json blob. Now that we are using Logger, I will log with the current role rather then 'system' for "event" logs that come from a user action, on VxScan this will log as "voter" if no one is currently authenticated. This makes sures on the precinct scanner that we log with the user as voter: 
1. Initiating a scan with the "scanStart" log
2. Accepting a ballot after adjudication
3. Returning a ballot after adjudication. 

Things like calibration will now be logged with the right user as well. Similar logic applied to MarkScan to make sure we log the happy path voter actions: 
1. Initiating a print 
2. Invalidating a ballot after review
2. Casting a ballot after review

And the pollworker will have logs for
1. Loading the paper at the begining of the flow 

There were already explicit other logs for the validated/invalidate voter-actions in mark-scan that I decided to leave in place as those logs are a bit easier to find but this change is still useful to get the "user" on more other mark-scan logs.  

## Testing Plan
Tested in dev setup, ran test cases, planning to test more on real vsap at atrium for more mark-scan flows will make other PRs if I need to tweak things

## Checklist

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -26,6 +26,7 @@ import {
   StateSchema,
   State,
   sendParent,
+  EventObject,
 } from 'xstate';
 import { Buffer } from 'buffer';
 import { Optional, assert, assertDefined } from '@votingworks/basics';
@@ -39,7 +40,7 @@ import {
   InterpretFileResult,
   interpretSimplexBmdBallot,
 } from '@votingworks/ballot-interpreter';
-import { LogEventId, LogLine, BaseLogger } from '@votingworks/logging';
+import { LogEventId, LogLine, Logger } from '@votingworks/logging';
 import { InsertedSmartCardAuthApi } from '@votingworks/auth';
 import {
   BooleanEnvironmentVariableName,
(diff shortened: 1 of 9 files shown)


apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -150,6 +151,25 @@ export type PaperHandlerStatusEvent =
...
+    'PAT_DEVICE_CONNECTED',
+    'PAT_DEVICE_DISCONNECTED',
+    'VOTER_CONFIRMED_PAT_DEVICE_CALIBRATION',
+    'VOTER_CONFIRMED_SESSION_END',
+    'POLL_WORKER_CONFIRMED_BALLOT_BOX_EMPTIED',
+    'RESET',
+    'START_SESSION_WITH_PREPRINTED_BALLOT',
+    'SYSTEM_ADMIN_STARTED_PAPER_HANDLER_DIAGNOSTIC',
+    'PAPER_READY_TO_LOAD', // paper was inserted by a user
+  ].includes(event.type);
```
kofi-q, line 170: Non-blocker: thinking this could go stale fairly easily - could be worth switching to an exhaustive `Record<X, bool>` or similar to make sure people remember to classify when adding a new event?

kofi-q (approved):
LGTM!

PR #5357, opened 2024-09-04
Log lockout events

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/5341
Adds specific logs instead of just general incorrect pin logs when an incorrect pin causes an account lockout. @arsalansufi can you think of any other weird auth cases to make sure to log? 

## Testing Plan
Caused lockout, noticed logs, added tests. 

## Checklist


libs/auth/src/inserted_smart_card_auth.ts
```diff
@@ -104,14 +104,26 @@ async function logAuthEvent(
...
-            }
-          );
+          if (newAuthStatus.lockedOutUntil) {
+            await logger.log(
+              LogEventId.AuthPinEntryLockout,
+              newAuthStatus.user.role,
+              {
+                disposition: LogDispositionStandardTypes.Failure,
+                message: `User entered incorrect PIN. Maximum attempts exceeded, locked out until: ${newAuthStatus.lockedOutUntil.toString()}`,
+                lockedOutUntil: newAuthStatus.lockedOutUntil.toString(),
```
arsalansufi, line 114: Just curious: What's the thinking behind putting this in a separate field when it's also already included in the message?

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -130,10 +130,22 @@ async function logAuthEventIfNecessary(
...
+                message: `User entered incorrect PIN. Maximum attempts exceeded, locked out until: ${newAuthStatus.lockedOutUntil.toString()}`,
+                lockedOutUntil: newAuthStatus.lockedOutUntil.toString(),
+              }
+            );
+          } else {
+            await logger.log(LogEventId.AuthPinEntry, newAuthStatus.user.role, {
+              disposition: LogDispositionStandardTypes.Failure,
+              message: 'User entered incorrect PIN.',
+            });
+          }
```
arsalansufi, line 148: > can you think of any other weird auth cases to make sure to log?

The one other one that comes to mind is hitting a session time limit. We do log that you got logged out in that case but don't indicate why.

We could probably accomplish this by changing the `reason` field here:

https://github.com/votingworks/vxsuite/blob/44e8bd8c4fae2b02b59ac41ab7e79174e96c331f/libs/auth/src/dipped_smart_card_auth.ts#L432-L436

And adding a reason field to this log:

https://github.com/votingworks/vxsuite/blob/44e8bd8c4fae2b02b59ac41ab7e79174e96c331f/libs/auth/src/dipped_smart_card_auth.ts#L169-L172

Like we have for the failed login log:

https://github.com/votingworks/vxsuite/blob/44e8bd8c4fae2b02b59ac41ab7e79174e96c331f/libs/auth/src/dipped_smart_card_auth.ts#L108-L110

Open to other approaches too of course!

arsalansufi (approved):
LGTM! Thanks for adding these - I should've added them way back when I worked on this 😅

libs/auth/src/inserted_smart_card_auth.ts
carolinemodic: I have gotten in the habit of doing this in case we add a UI for filtering by keys someday whenever I think I might want to pull out a specific data item but its probably not really necessary

carolinemodic:
@arsalansufi ended up being a bit more complex to get the session expiry stuff working, lmk what you think

arsalansufi (approved):
Ah yes I see how this ended up being a bit complex. But I don’t have any immediate alternative suggestions so this LGTM!

Issue #4896, opened 2024-05-30
VxDesign: Properly account for HTML in propositions when generating translations and audio


arsalansufi:

jonahkagan:
Mostly addressed this in #5359 (for audio). The translation API [docs](https://cloud.google.com/translate/docs/basic/translating-text) claim that HTML tags should be preserved. However, I've noticed that sometimes it strips HTML tags. This is likely just a bug in the API, so I'm not sure what we can do to fix it.

arsalansufi:
Consensus during discussion with @mattroe and @adghayes: We'll solve this with our eventual support of user-entered translation overrides

Issue #4895, opened 2024-05-30
VxMark: Support VxDesign rich text in propositions


arsalansufi:
While VxDesign HMPB ballot generation knows how to handle VxDesign rich text in propositions, the VxMarkScan UI does not. So you'll end up with funky looking prompts, where

```
{
  "@type": "BallotDefinition.LanguageString",
  "Language": "en",
  "Content": "Retain <b>Robert Demergue</b> as Chief Justice of the Supreme Court?"
},
```

is rendered as:

> Retain ,Robert Demergue, as Chief Justice of the Supreme Court?

Related issue: https://github.com/votingworks/vxsuite/issues/4896

jonahkagan:
Noting that in v3, I don’t think we ever supported markdown, just a limited subset of HTML.

For v4, with the new ballot rendering stack, I haven't yet built any support for rich text (though it will be quite easy to do), so we can decide what kind of markup language we want.

I would lean towards sticking with a limited subset of HTML, since that will give us the most flexibility in the long term.

jonahkagan:
Also noting that I do think we need to implement this (rather than just work around the markup by stripping it, for example), since many ballot measures use the formatting to convey their meaning (e.g. striking through text that is being removed from a law)

PR #5359, opened 2024-09-05
Rich text in ballot measures

jonahkagan:
## Overview

Task: https://github.com/votingworks/vxsuite/issues/4895

Adds a rich text editor for ballot measure descriptions in VxDesign (outputting HTML). Ensures that rich text can be rendered in HMPB and VxMark, and also accounts for HTML when generating audio. Only supports inline formatting and lists for now, not images or tables (I'll add them in a future PR).

Note that translations seem to mostly work, preserving HTML (as described in the [docs](https://cloud.google.com/translate/docs/basic/translating-text)), though I've seen it mess up and strip some tags. Will leave a note on https://github.com/votingworks/vxsuite/issues/4896.

Details in commits

## Demo Video or Screenshot

VxDesign rich text input

https://github.com/user-attachments/assets/2d177259-7557-4612-a874-ef9103bb120a

VxMark
(Note that I added the audio cue for underlined text after taking this video, so it's not shown) 

https://github.com/user-attachments/assets/4626375e-71ad-4d23-b3b9-a7132641ff0a

HMPBs
Screenshot 2024-09-05 at 12 30 46 PM

## Testing Plan
Manual testing and updated automated tests

## Checklist


apps/design/frontend/src/rich_text_editor.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+// userEvent.type only works when the editor is empty for some reason - this
+// appends to existing content
+function type(editor: HTMLElement, text: string) {
+  fireEvent.change(editor.querySelector('.tiptap')!, {
+    target: {
+      textContent: editor.querySelector('.tiptap p')!.textContent + text,
+    },
+  });
+}
```
eventualbuddha, line 13: I looked into why this happens. It seems that versions of `@testing-library/user-event` prior to 14.0.0 crashed when attempting to type into a `[contenteditable]` element that was not empty due to a bug with trying to set the selection. It would throw a `DOMException` in `jsdom` like so:

```ts
  if (offset > nodeLength(node)) {
    throw DOMException.create(node._globalObject, ["Offset out of bound.", "IndexSizeError"]);
  }
```

This happened because `offset` was the length of the text content of the `[contenteditable]` element (i.e. 7 when the text content is "Content") and `nodeLength(node)` is computed as the child count of `node` (i.e. 1 when the content of the `[contenteditable]` element is `<p>Content</p>`).

This workaround happens to poke `tiptap` just the right way to make it do what we want, but it feels kinda brittle. I investigated whether a newer version of `@testing-library/user-event` fixes this issue and it _sort of_ does. In https://github.com/testing-library/user-event/pull/822, basic support for updates to `[contenteditable]` elements is added and released in v14.0.0. However, when I updated to v14.5.2 (latest) and changed to `userEvent.type` in the "bold formatting" test below, it failed because the content ended up as `<p><strong> bold</strong>Content</p>`. This is because the selection defaults to the beginning of the content area. I tried using `userEvent.type(editor, ' bold', { initialSelectionStart: 7 })` or `userEvent.type(editor, '[End]')` beforehand, but in each case I get [this error](https://github.com/testing-library/user-event/blob/d0362796a33c2d39713998f82ae309020c37b385/src/event/selection/setSelectionRange.ts#L33):

> Not implemented. The result of this interaction is unreliable.

This happens because `@testing-library/user-event` only sets selection ranges for `input`, `textarea`, and `[contenteditable]` where there is a single child node and it's just text.

In summary, `@testing-library/user-event` does not appear to support testing with `[contenteditable]` in anything but the most basic way. I suggest we replace most or all of the tests in this file with Playwright tests.

apps/design/frontend/src/rich_text_editor.tsx
```diff
@@ -0,0 +1,163 @@
...
+    <Button
+      color={isActive ? 'primary' : 'neutral'}
+      fill={isActive ? 'tinted' : 'transparent'}
+      {...props}
+    />
+  );
+}
+
+function Toolbar({ editor }: { editor: Editor | null }) {
+  if (!editor) return null;
```
eventualbuddha, line 79: Consider pushing the conditional outside this component:
```suggestion
function Toolbar({ editor }: { editor: Editor }) {
```

eventualbuddha (approved):
I didn't see any comments explaining where the XSS prevention code is located. How do we know that VxDesign or VxMark won't run `<script>` or `onclick` JavaScript? I verified that in a very simple attack that the code is _not_ run, but it took a fair amount of digging to convince myself that it was doing the right thing. This should be documented inline wherever a reasonable person would be wondering about XSS mitigation.

jonahkagan:
> I didn't see any comments explaining where the XSS prevention code is located. 
Thanks for raising this, and sorry it wasn't more clearly documented. I think the current XSS prevention is happening in react-i18next, which strips HTML tags except the allow-listed ones. But I need to change that approach anyway in the next PR to allow images, which aren't supported. I'll update this PR with the new approach and some HTML sanitization so we can discuss.

apps/design/frontend/src/rich_text_editor.test.tsx
jonahkagan: Thanks for looking into this. I agree that the workaround is a bit brittle and that Playwright tests would be a better overall solution. I briefly considered switching to Playwright tests when I first realized `user-event` wasn't working. I want to stick with this for now because it's working and I don't want to put more time into it, given that VxDesign is still an internal tool and won't be part of the upcoming release. How does that sound to you?

jonahkagan:
>  I'll update this PR with the new approach and some HTML sanitization so we can discuss.
Actually, these changes are a little bit entangled with some changes in the next PR. It'll make my life a little bit easier to just include them in the next PR, so we can discuss them there.

apps/design/frontend/src/rich_text_editor.tsx
jonahkagan: [b899ddc](https://github.com/votingworks/vxsuite/pull/5359/commits/b899ddc2211ab59ca355a9ba432cb88e236aee76)

apps/design/frontend/src/rich_text_editor.test.tsx
eventualbuddha: I don't have a problem with that, but I'd suggest adding a comment with a link to a Github issue to change to Playwright. Unless you think we _shouldn't_ change it in the future?

jonahkagan: Sounds good: e69a916a4

PR #5370, opened 2024-09-10
[Node Upgrade] Revert packages using v8 coverage back to babel

kofi-q:
## Overview

Prepping for a Node upgrade, 16 -> (hopefully) 20 and `v8` coverage breaks in v20 (https://github.com/nodejs/node/issues/51251). Moving back to the `babel` coverage provider in packages that we previously moved to `v8`

Had to do some minor refactoring in a few cases to be able to mark the right blocks with `istanbul ignore`, but no functional changes.

### DIFF
--- a/apps/admin/backend/src/app.tally_report_data.test.ts
+++ b/apps/admin/backend/src/app.tally_report_data.test.ts
@@ -9,6 +9,8 @@ import {
 } from '@votingworks/utils';
 import { assert, find } from '@votingworks/basics';
 import { Tabulation } from '@votingworks/types';
+import { initializeGetWorkspaceDiskSpaceSummary } from '@votingworks/backend';
+import { mockOf } from '@votingworks/test-utils';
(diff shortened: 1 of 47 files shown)


apps/admin/backend/src/app.tally_report_data.test.ts
```diff
@@ -35,6 +45,13 @@ beforeEach(() => {
   featureFlagMock.enableFeatureFlag(
     BooleanEnvironmentVariableName.SKIP_CAST_VOTE_RECORDS_AUTHENTICATION
   );
+  mockOf(initializeGetWorkspaceDiskSpaceSummary).mockReturnValue(() =>
```
kofi-q, line 48: unrelated, but this test kept flaking locally, so copied this pattern from other tests that depend on this function to de-flake

carolinemodic (approved):

Issue #3873, opened 2023-08-23
VxMark: Alarm when printer cover is opened while polls are open


kshen0:
The takeaway from [All Things 12.1](https://docs.google.com/document/d/1Zp9_qVn5fzC9FgMiD2RyGemOEYXIFxlRdAA2vzpub0o/edit) was that we need to add an alert on VxMark when the printer/scanner cover is opened when polls are open (in an activated state).

See the doc above for all the specific requirements we must meet for this issue. It notably includes:

- The visual styling requirements we have implemented on VxScan warnings
- An audio alert over the VSAP speaker that measures 60db output from 4 feet away. The nature of the sound is not specified in VVSG, but it should be a continuous alert until a user authenticates (see below).

The alarm should not go off in an activated state when an a user (pw, em, sa) is authenticated to enable unjamming the printer. As part of this change, we should add copy to the jam screen on VxMark to prompt poll workers to authenticate before unjamming.

The alarm should also be disabled and screen as soon as a user authenticates.

Slack discussion of how we got here: https://votingworks.slack.com/archives/C062Y0ZFKFT/p1715781514012669

kshen0:
On pause since we may be getting subcomponents that won't include the optical sensor/plastic cover

amcmanus:
@kofi-q, I've done some initial tests with this via the command line, and it will just be a set of `amixer` commands to enable playing via the on-board speaker, even if headphones are connected. 

1. `amixer set -c0 Speaker 100% unmute`
2. play the alarm
3. `amixer set -c0 Speaker 0% mute`

If the headphones are connected, the above settings will result in the alarm playing through both. If we'd rather not have it also play in the headphones, it will require adding the appropriate `amixer` commands to mute headphones, play the alarm, unmute the headphones. Those commands would be: 
```
amixer set -c0 Headphone 0% mute
amixer set -c0 Headphone 100% unmute
```

The `amixer` commands will need to be run as the `vx-ui` user since it owns the audio session used by the app.

kofi-q:
Finally managed to test this on a prod image and the audio output switching between headphones and speaker is working great and it's just about hitting the 60 dB mark - although, VxAustin is fairly echoey and the readings might be higher here.

[...]

PR #5377, opened 2024-09-11
[MarkScan] Switch audio to speakers and play alarm sound when cover is open

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/3873

Adding alarm sound asset and playback component to the "cover open" alarm screen in the frontend.
Also unmuting the speaker whenever we enter the alarm state and re-muting it when we exit.

Using a combination of instructions from [here](https://github.com/votingworks/vxsuite/issues/3873#issuecomment-2284961742) and the patter used in the headphone detection script [here](https://github.com/votingworks/vxsuite-complete-system/blob/7d9661f/app-scripts/pactl.sh#L16). It's slow to execute, since we're going through the shell, and sometimes gets the headphone port into a weird unmuted-but-no-sound state, but time-boxing the effort, since this screen will be fairly rare.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/b31bc495-6b69-4d69-afab-6cbe4027a4dd

## Testing Plan
- Manual

## Checklist


apps/mark-scan/backend/src/audio/outputs.ts
```diff
@@ -0,0 +1,59 @@
+/* istanbul ignore file */
+import { LogEventId, Logger } from '@votingworks/logging';
+import { exec } from '@votingworks/backend';
+import { NODE_ENV } from '../globals';
+
+const VX_UI_USERNAME = 'vx-ui';
+const CMD_RUN_AS_VX_UI = `sudo -u ${VX_UI_USERNAME} XDG_RUNTIME_DIR=/run/user/$(id -u ${VX_UI_USERNAME})`;
```
carolinemodic, line 7: we probably need to add something to the sudoers file in complete-system for this to work on a prod image cc @amcmanus

carolinemodic (approved):

apps/mark-scan/backend/src/audio/outputs.ts
kofi-q: Yeah, worth double-checking - Adam, for context, these commands will be running in the mark-scan backend service (vx-services). Using a similar pattern to the `pactl.sh` script we execute for the headphones check , but running the command directly instead of prxying through a script - LMK if there could be any issues there

amcmanus: To make `CMD_RUN_AS_VX_UI` work, we'll need to make a couple of changes to our `sudoers` config. 

To enable setting the `XDG_RUNTIME_DIR` env var, we need to add 
```
Defaults!/usr/bin/amixer env_keep += "XDG_RUNTIME_DIR"
```

To allow the vx-services user to run amixer as the vx-ui user, we need to add
```
vx-services ALL=(vx-ui:ALL) NOPASSWD: /usr/bin/amixer
```

With that said, it may not be necessary to pass the `XDG_RUNTIME_DIR` env var to amixer commands. That was specific to the `pactl` command to gain access to the pulseaudio config running under the `vx-ui` user, and a reason for using a script rather than a similar sudo config as above. 

I think we should get your code running on a VSAP (not via VxDev) and test both the `sudoers` and `amixer` configs. It's not really possible for me to simulate the audio config on any hardware other than a VSAP, and I want to be sure we're not missing something specific to a production environment. I think there's a chance `amixer` may need to include the `Master` audio channel based on behavior we saw with the startup chime and headphone support. (My primary concern is that the alarm could then put the system in a state where headphone audio is no longer audible.)

kofi-q: Ah, thanks - I see the corresponding suoders config for `pactl.sh` now. Was only able to test the "run as vx-ui" piece while logged in as vx-admin on the prod vsap, so will try just copying this code over and running the actual app to make sure it all works end-to-end. Good to know we can probably do this without setting `XDG_RUNTIME_DIR` - will drop it for the test to verify

kofi-q:
Thanks for the guidance here! Will test this on a prod image after merge and verify everything's working.

Issue #5320, opened 2024-08-28
VxAdmin: Expose election package export button in sys admin UI


arsalansufi:
We have three key user roles in VxSuite, aside from voters: system administrators, election managers, and poll workers. Sys admins have the highest privileges. They configure the election definition on VxAdmin, which is then used to configure VxCentralScan, VxMark, and VxScan. Sys admins' smart cards are accordingly election-agnostic. Election managers and poll workers, on the other hand, have election-specific smart cards and perform operations within the context of an election.

With that context in mind, VxSuite configuration currently looks like this:
- System administrator unlocks VxAdmin with their smart card and configures VxAdmin with an election package
- System administrator locks the machine
- Election manager unlocks VxAdmin with their smart card and re-exports a signed version of the election package to a USB drive
- That USB drive is then used to configure the other VxSuite machines

In practice, system administrators and election managers are often the same person. We had to separate the two roles and their corresponding smart cards to meet certain VVSG2 requirements.

Since the two are often the same person, and system administrators have higher privileges than election managers, let's give system administrators the same ability to re-export a signed version of the election package that we give to election managers.

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1724786491850729?thread_ts=1724786491.850729&cid=CJU9MSC6S

adghayes:
Note:
- not creating ticket for automatically signing package at this time, as it could create complications and confusion
- not creating ticket for SA to switch fully into EM role

PR #5379, opened 2024-09-11
Allow for system admins to save election packages in same way election manager can

nbhatia823:
## Overview

https://github.com/votingworks/vxsuite/issues/5320

## Demo Video or Screenshot

Current System Admin

Screenshot 2024-09-11 at 2 35 40 PM

Current Election Manager

Screenshot 2024-09-11 at 2 36 40 PM

System Admins + Election Manager without explanations

Screenshot 2024-09-11 at 2 20 00 PM
Screenshot 2024-09-11 at 2 40 41 PM

OUTDATED: Systems Admins with short explanations (Not using this approach)

Screenshot 2024-09-10 at 9 18 23 PM

## Testing Plan

Test added

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/election_screen.tsx
+++ b/apps/admin/frontend/src/screens/election_screen.tsx
@@ -61,17 +61,16 @@ export function ElectionScreen(): JSX.Element {
           </P>
         </div>
       </ElectionCard>
-      <div style={{ display: 'flex', gap: '1rem' }}>
-        {(isSystemAdministratorAuth(auth) || isElectionManagerAuth(auth)) && (
+      {isSystemAdministratorAuth(auth) && (
+        <div style={{ display: 'flex', gap: '1rem' }}>
           <ExportElectionPackageModalButton />
-        )}
-        {isSystemAdministratorAuth(auth) && (
           <UnconfigureMachineButton
             isMachineConfigured
             unconfigureMachine={unconfigureMachine}
           />
-        )}
-      </div>
+        </div>
+      )}
+      {isElectionManagerAuth(auth) && <ExportElectionPackageModalButton />}
     </NavigationScreen>
   );
 }
--- a/apps/admin/frontend/src/screens/election_screen.tsx
+++ b/apps/admin/frontend/src/screens/election_screen.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import { useContext } from 'react';
 import styled from 'styled-components';
 import {
   format,
@@ -61,23 +61,17 @@ export function ElectionScreen(): JSX.Element {
           </P>
         </div>
       </ElectionCard>
-      {isSystemAdministratorAuth(auth) && (
-        <UnconfigureMachineButton
-          isMachineConfigured
-          unconfigureMachine={unconfigureMachine}
-        />
-      )}
-      {isElectionManagerAuth(auth) && (
-        <React.Fragment>
-          <P>
-            Save the election package to the USB drive to configure VxSuite
-            components.
-          </P>
-          <P>
-            <ExportElectionPackageModalButton />
-          </P>
-        </React.Fragment>
-      )}
+      <div style={{ display: 'flex', gap: '1rem' }}>
+        {(isSystemAdministratorAuth(auth) || isElectionManagerAuth(auth)) && (
+          <ExportElectionPackageModalButton />
+        )}
+        {isSystemAdministratorAuth(auth) && (
+          <UnconfigureMachineButton
+            isMachineConfigured
+            unconfigureMachine={unconfigureMachine}
+          />
+        )}
+      </div>
     </NavigationScreen>
   );
 }
--- a/apps/admin/frontend/src/screens/election_screen.tsx
+++ b/apps/admin/frontend/src/screens/election_screen.tsx
@@ -30,6 +30,7 @@ export function ElectionScreen(): JSX.Element {
   const history = useHistory();
   const unconfigureMutation = unconfigure.useMutation();
 
+  assert(isSystemAdministratorAuth(auth) || isElectionManagerAuth(auth));
   assert(electionDefinition && typeof configuredAt === 'string');
   const { election } = electionDefinition;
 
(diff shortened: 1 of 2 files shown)


apps/admin/frontend/src/screens/election_screen.tsx
```diff
@@ -61,23 +61,17 @@ export function ElectionScreen(): JSX.Element {
...
+        {(isSystemAdministratorAuth(auth) || isElectionManagerAuth(auth)) && (
+          <ExportElectionPackageModalButton />
+        )}
+        {isSystemAdministratorAuth(auth) && (
+          <UnconfigureMachineButton
+            isMachineConfigured
+            unconfigureMachine={unconfigureMachine}
+          />
+        )}
+      </div>
```
nbhatia823, line 74: This implementation is for Option 2. Review commit by commit to compare Option 1 and Option 2.

apps/admin/frontend/src/screens/election_screen.tsx
```diff
@@ -61,22 +61,30 @@ export function ElectionScreen(): JSX.Element {
...
-        </React.Fragment>
+        </div>
+      )}
+      {isSystemAdministratorAuth(auth) && (
+        <div
+          style={{
+            display: 'flex',
+            gap: '1.25rem',
+            alignItems: 'center',
+            paddingTop: '1rem',
```
nbhatia823: Whoops, would update to `marginTop`

apps/admin/frontend/src/screens/election_screen.tsx
jonahkagan: I think it's clearer without the text. I'd be curious to see what it's like with the "unconfigure machine" button aligned to the right edge of the election card

nbhatia823: Screenshot 2024-09-11 at 2 58 17 PM

I think it's a bit less common of an alignment for buttons, but interested in what you think

nbhatia823: And would be fine with removing the text for Election Managers as well, or is it more helpful for them?

jonahkagan: Ok yeah, let's stick with them both on the left. Sometimes I like to put destructive actions a bit out of the main flow, but it doesn't seem necessary here.

jonahkagan: > And would be fine with removing the text for Election Managers as well, or is it more helpful for them?

I think it's best to remove the text in both and have them consistent

nbhatia823: Oh right, that's a helpful thing to keep in mind reg. destructive actions. Will revert to left-aligned buttons for this.

apps/admin/frontend/src/screens/election_screen.tsx
```diff
@@ -61,23 +61,17 @@ export function ElectionScreen(): JSX.Element {
...
-      )}
-      {isElectionManagerAuth(auth) && (
-        <React.Fragment>
-          <P>
-            Save the election package to the USB drive to configure VxSuite
-            components.
-          </P>
-          <P>
-            <ExportElectionPackageModalButton />
-          </P>
```
nbhatia823, line 78: Any reason why we would have wrapped the Button in a `P` previously?

jonahkagan: It's a legacy pattern, feel free to remove. The only reason to keep it would be if it's needed for spacing

apps/admin/frontend/src/screens/election_screen.tsx
```diff
@@ -61,17 +61,16 @@ export function ElectionScreen(): JSX.Element {
...
-        {isSystemAdministratorAuth(auth) && (
           <UnconfigureMachineButton
             isMachineConfigured
             unconfigureMachine={unconfigureMachine}
           />
-        )}
-      </div>
+        </div>
+      )}
+      {isElectionManagerAuth(auth) && <ExportElectionPackageModalButton />}
```
nbhatia823, line 73: Made a small code reorg to have it be rendering buttons by user type. Either way one line of code was duplicated (`isSystemAdmin type check` or `ExportElectionPackageModalButton`). This way might be a bit easier to read, but it's inconsequential.

nbhatia823:
Screenshot 2024-09-11 at 3 29 33 PM
Screenshot 2024-09-11 at 3 29 18 PM

Updated screenshots, approach w/o explanations.

apps/admin/frontend/src/screens/election_screen.tsx
```diff
@@ -61,23 +61,17 @@ export function ElectionScreen(): JSX.Element {
...
-          <P>
-            Save the election package to the USB drive to configure VxSuite
-            components.
-          </P>
-          <P>
-            <ExportElectionPackageModalButton />
-          </P>
-        </React.Fragment>
-      )}
+      <div style={{ display: 'flex', gap: '1rem' }}>
```
jonahkagan: I think a 0.5rem gap might look better here (and IIRC that's what we've been doing elsewhere in VxAdmin)

apps/admin/frontend/src/screens/election_screen.tsx
jonahkagan: Given that this screen will only ever be rendered for system admins or election managers, it should be safe to just check for the sys admin auth to conditionally render the unconfigure button.

If we wanted to be extra sure about that assumption, we could add something like this at the top of the screen component:
```ts
assert(isSystemAdministrator(auth) || isElectionManager(auth))
```

jonahkagan (approved):
Looks great! Just a few small notes

apps/admin/frontend/src/screens/election_screen.tsx
nbhatia823: Ah that makes complete sense given that this screen is only rendered for those two user types. Will keep that in mind going forward.

And that `assert` makes sense to provide the certainty - in reality do we want to add that check here or was it more of an instructional thought?

jonahkagan: I'm in favor of adding the `assert` to be defensive, though the stakes feel fairly low here so I wouldn't insist on it.

apps/admin/frontend/src/screens/election_screen.tsx
nbhatia823: Ah right, the gap between the header buttons is 0.5rem, will update

adghayes:
Another reason I'm excited about this change is that it no longer seems like the "Unconfigure" button is the primary button (it didn't have a primary look, but it was the _only_ button), so we might see fewer people accidentally unconfigure after L&A, which we just saw in NH a couple times.

PR #5380, opened 2024-09-11
libs/usb-drive: Add `sync` method to enable mocking

jonahkagan:
## Overview

Adds a `sync` method to `libs/usb-drive` so that we can easily mock sync calls in tests. This will hopefully help prevent test flakiness due to hanging sync calls in `apps/scan/backend`.

To enable easy mocking across all of the `apps/scan/backend` tests, adds a new mock function method `expectOptionalRepeatedCallsWith`.

## Testing Plan
Updated tests. Hope it solves flakiness but can't be sure - ran ~20 runs with no flake.

## Checklist

### DIFF
--- a/libs/backend/src/exporter.test.ts
+++ b/libs/backend/src/exporter.test.ts
@@ -36,6 +36,7 @@ afterEach(() => {
     tmpDir.removeCallback();
   }
   tmpDirs.length = 0;
+  mockUsbDrive.assertComplete();
 });
 
 test('exportData with string', async () => {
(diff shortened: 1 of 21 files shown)


arsalansufi (approved):
LGTM! Thanks for digging into the flaky tests

libs/backend/src/exporter.test.ts
```diff
@@ -148,14 +148,14 @@ test('exportDataToUsbDrive happy path', async () => {
   usbDrive.status
     .expectCallWith()
     .resolves({ status: 'mounted', mountPoint: tmpDir });
+  usbDrive.sync.expectCallWith().resolves();
```
kofi-q, line 151: do these cases need an `assertComplete()` at the end to be equivalent to the previous code?

jonahkagan: Ah thank you, I just assumed it was already there

kofi-q (approved):
🚢

PR #5394, opened 2024-09-12
Converter updates for NH November 2024

eventualbuddha:
## Overview
Based on preliminary data from the state, here's a few changes to make the conversion process better with the data provided.

## Demo Video or Screenshot
image

## Testing Plan
Manually ran conversion on the data from the state.

### DIFF
--- a/libs/converter-nh-accuvote/src/convert/convert_election_definition_header.ts
+++ b/libs/converter-nh-accuvote/src/convert/convert_election_definition_header.ts
@@ -82,6 +82,22 @@ export function convertElectionDefinitionHeader(
       paperSize = BallotPaperSize.Legal;
       break;
 
+    case '8.5X17':
+      paperSize = BallotPaperSize.Custom17;
+      break;
+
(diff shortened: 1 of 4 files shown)


libs/converter-nh-accuvote/src/convert/convert_election_definition_header.ts
```diff
@@ -82,6 +82,22 @@ export function convertElectionDefinitionHeader(
       paperSize = BallotPaperSize.Legal;
       break;
 
+    case '8.5X17':
+      paperSize = BallotPaperSize.Custom17;
+      break;
+
+    case '8.5X18':
+      paperSize = BallotPaperSize.Custom18;
```
eventualbuddha, line 90: The data from the state says the ballot is 8.5X18, but it's not. We only warn on mismatch, but we do fail to parse the data if the ballot size isn't recognized at all.

arsalansufi (approved):
LGTM! Thank you for testing and making these edits

PR #5408, opened 2024-09-16
unmarked write-in accounting tweaks

### LINKED ISSUE #5382: VxAdmin: UWIs in WIA Report
If an unmarked write-in is marked as invalid, it will disappear from the WIA report, which presents an accounting problem. They should stay in the WIA report.

See: https://votingworks.slack.com/archives/C07LBG41B3M/p1726016113081819https://votingworks.slack.com/archives/C07LBG41B3M/p1726016113081819>

adghayes:
## Overview

Closes #5382. Does two things:
- makes sure unmarked write-ins appear in the write-in adjudication report
- undoes the WIA adjudication queue ordering that always puts UWIs last, which can be confusing if a UWI and a proper WI are on the same ballot contest 

## Testing Plan

Updated existing test coverage to match new logic.

### DIFF
--- a/apps/admin/backend/src/store.ts
+++ b/apps/admin/backend/src/store.ts
@@ -1836,13 +1836,13 @@ export class Store {
    * Gets write-in tallies specifically for tabulation, filtered and and
    * grouped by cast vote record attributes.
    *
-   * When using write-in tallies to hydrate tally results, unadjudicated and
-   * unmarked write-ins should not be included because they are not associated
-   * with an original or adjudicated mark. When using write-in tallies as a
-   * summary of write-in adjudication, they should be included because they
-   * exist in the adjudication flow just as marked write-ins do. The behavior
-   * can be controlled with the `includeUnadjudicatedAndInvalidUnmarkedWriteIns`
-   * flag.
+   * When using write-in tallies to hydrate tally results, unmarked write-ins
+   * that haven't been adjudicated or are invalid should not be included
+   * because they are not associated with an original or adjudicated mark. Whe
+   * using write-in tallies as a summary of write-in adjudication, they should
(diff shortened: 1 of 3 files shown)


adghayes:
@arsalansufi re-assign review if needed, but wanted you tagged for review since this is a candidate for the potential v3.1.1 general patch, whether it's being tracked or not already.

apps/admin/backend/src/store.ts
```diff
@@ -1835,17 +1835,27 @@ export class Store {
...
    * Gets write-in tallies specifically for tabulation, filtered and and
    * grouped by cast vote record attributes.
+   *
+   * When using write-in tallies to hydrate tally results, unadjudicated and
+   * unmarked write-ins should not be included because they are not associated
+   * with an original or adjudicated mark. When using write-in tallies as a
+   * summary of write-in adjudication, they should be included because they
+   * exist in the adjudication flow just as marked write-ins do. The behavior
+   * can be controlled with the `includeUnadjudicatedAndInvalidUnmarkedWriteIns`
+   * flag.
```
arsalansufi, line 1845: Should "unadjudicated and unmarked write-ins" in this comment specifically read "unadjudicated and *invalid* unmarked write-ins"?

Even with that change, there's a bit of ambiguity as this could mean "unadjudicated _marked_ write-ins and invalid _unmarked_ write-ins". I think "unmarked write-ins that haven't been adjudicated or are invalid" might be clearer

adghayes: Sounds good, change incorporated.

I can see someone making the same complaint about the variable name, but I think it reads a bit clearer as a variable name, so I am inclined to leave it.

arsalansufi (approved):
LGTM!

PR #5421, opened 2024-09-19
Allow System Admins to access diagnostics in VxScan and MarkScan

nbhatia823:
## Overview

https://github.com/votingworks/vxsuite/issues/5266

## Demo Video or Screenshot

**Scan - Election Manager access of Diagnostics under System Settings**

https://github.com/user-attachments/assets/5350db3a-dfa2-42ee-b1c8-3943367120cd

- Note: One non-ideal functionality shown in the video is that after a user closes the Diagnostics, they return to the Configuration tab rather than remaining in the System Settings tab. This is because the `TabbedSection` is not being rendered, so state is reset. We could fix this by pulling the state to the `ElectionManagerScreen`, but I am not sure if we feel this is needed.

**MarkScan - Election Manager access of Diagnostics**

https://github.com/user-attachments/assets/9e57a4a7-68d2-4966-911a-ca11d3ae6188

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/pages/admin_screen.tsx
+++ b/apps/mark-scan/frontend/src/pages/admin_screen.tsx
@@ -18,6 +18,7 @@ import {
   UnconfigureMachineButton,
   ExportLogsButton,
   SignedHashValidationButton,
+  Button,
 } from '@votingworks/ui';
 import {
   ElectionDefinition,
@@ -33,6 +34,7 @@ import {
   setTestMode,
   useApiClient,
 } from '../api';
+import { DiagnosticsScreen } from './diagnostics/diagnostics_screen';
 
 export interface AdminScreenProps {
   appPrecinct?: PrecinctSelection;
@@ -63,6 +65,8 @@ export function AdminScreen({
   const ejectUsbDriveMutation = ejectUsbDrive.useMutation();
   const setPrecinctSelectionMutation = setPrecinctSelection.useMutation();
   const setTestModeMutation = setTestMode.useMutation();
+  const [isDiagnosticsScreenOpen, setIsDiagnosticsScreenOpen] =
+    React.useState(false);
 
   async function unconfigureMachineAndEjectUsb() {
     try {
@@ -72,6 +76,14 @@ export function AdminScreen({
     } catch (error) {
       // Handled by default query client error handling
     }
+  }
+
+  if (isDiagnosticsScreenOpen) {
+    return (
+      <DiagnosticsScreen
+        onBackButtonPress={() => setIsDiagnosticsScreenOpen(false)}
+      />
+    );
   }
 
   return (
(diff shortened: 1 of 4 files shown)


apps/mark-scan/frontend/src/pages/admin_screen.tsx
```diff
@@ -184,6 +196,12 @@ export function AdminScreen({
         <P>
           <SignedHashValidationButton apiClient={apiClient} />
         </P>
+        <H6 as="h2">Diagnostics</H6>
+        <P>
+          <Button onPress={() => setIsDiagnosticsScreenOpen(true)}>
+            View Diagnostics
```
nbhatia823, line 202: I added `View` because all the other buttons were actions, but happy to remove it if it seems unneeded

apps/scan/frontend/jest.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
   ],
   coverageThreshold: {
     global: {
-      branches: -54,
+      branches: -55,
```
nbhatia823, line 16: Adding this so we can use this thread to discuss solutions -

I am seeing the branch coverage for `admin_screen` [go down from 100% to 97.91%](https://app.circleci.com/pipelines/github/votingworks/vxsuite/17088/workflows/a492505d-4edc-475c-bda8-2e88ffe5a519/jobs/673529). I think it is because the branch where there is no diagnostic button (because `hardware v3`), then the branch for the `diagnostic_screen` cannot open. So line 244 evaluates to `false/null`, and then line 309 always goes down the same `false` branch.

I was able to fix the other jest errors, but this one has me a little stumped, unless the explanation I gave above is accurate. In that case, this would be an impossible state that would require this bump to the `coverageThreshhold`. Thanks for any advice!

apps/mark-scan/frontend/src/pages/admin_screen.tsx
jonahkagan: Either works for me

apps/scan/frontend/jest.config.js
jonahkagan: `printerStatusQuery.data` can't be undefined on L244 because we check `printerStatusQuery.isSuccess` earlier in the component. So the `?` causes an impossible branch.

A few tips for figuring out coverage issues:
- The terminal coverage report gives precedence to listing uncovered line numbers (last column, in red) before listing uncovered branch line numbers (in yellow). If you mark all of the uncovered lines with ignore comments, then you can see the uncovered branches (in this case, L244).
- Viewing the HTML report in the `coverage` directory is often a quick way to get more info - I think it shows uncovered branches IIRC

jonahkagan (approved):
LGTM!

apps/scan/frontend/jest.config.js
nbhatia823: Ah okay, thanks for the tips, super helpful. 

I was able to see what you mean by commenting out the uncovered line, showing that the L244 was uncovered. That being said, I am a bit surprised that this PR would affect the branch coverage in that case, since it was unrelated to that line of code.

apps/mark-scan/frontend/src/pages/admin_screen.tsx
nbhatia823: Screenshot 2024-09-20 at 9 02 25 AM

Renaming the title to "System" and the button to "Diagnostics", as this is more similar verbiage to what is used in the other apps. I figure we can change them all in the future if we  need, but better if we stay consistent for now.

Issue #5349, opened 2024-09-04
Hardware Diagnostics: Scanner Mark Thresholds


adghayes:
Per 1.1.6-G, we need to be able to show mark detection thresholds on a report:

> The voting system must have the capability to provide a report of the mark detection thresholds that have been used to program the scanner so that the information is available upon request.

For both VxScan and VxCentralScan, we need to add a section to the diagnostic reports that includes the mark thresholds, including the write-in threshold. Like the "Configuration" section, this section will be empty until configuration because the mark thresholds come from the system settings file.

PR #5427, opened 2024-09-23
Show mark thresholds for central-scan and scan

nbhatia823:
## Overview

https://github.com/votingworks/vxsuite/issues/5349

### CentralScan

Screenshot 2024-09-23 at 2 25 46 PM
Screenshot 2024-09-23 at 2 26 53 PM

### Scan

Screenshot 2024-09-23 at 2 24 08 PM
Screenshot 2024-09-23 at 2 24 25 PM

## Checklist


apps/scan/frontend/src/screens/diagnostics_screen.tsx
```diff
@@ -137,6 +141,7 @@ export function DiagnosticsScreen({
         mostRecentAudioDiagnostic={
           mostRecentAudioDiagnosticQuery.data ?? undefined
         }
+        markThresholds={systemSettings.markThresholds}
```
nbhatia823, line 144: Do we have a process for updating the `readiness-report` [mocks](https://github.com/votingworks/vxsuite/blob/main/apps/scan/backend/src/__image_snapshots__/readiness-report-1.png) that we use for testing?

kshen0: First confirm that the diff output is what you expected. The path is in the test output: `apps/scan/backend/src/__image_snapshots__/__diff_output__/readiness-report-1-diff.png`

Then to update the snapshot run `npm run test:ci -- -u`, or run the test and press `u` in the watcher tool.

libs/ui/src/diagnostics/configuration_section.tsx
```diff
@@ -89,6 +92,16 @@ export function ConfigurationSection({
...
         </P>
       )}
+      {markThresholds?.definite && (
+        <P>
+          <SuccessIcon /> Mark Threshold: {markThresholds.definite}
+        </P>
+      )}
+      {markThresholds?.writeInTextArea && (
+        <P>
+          <SuccessIcon /> Write-in Threshold: {markThresholds.writeInTextArea}
```
kshen0, line 102: Maybe truncate these to 4 decimals? I don't expect mark thresholds will be very long decimals but doesn't hurt to be safe.

apps/scan/frontend/src/screens/diagnostics_screen.tsx
nbhatia823: Awesome, thanks for the clear instructions. It looks as expected, so I've pushed the updated snapshot

apps/scan/backend/src/printing/readiness_report.ts
```diff
@@ -34,6 +34,7 @@ export async function saveReadinessReport({
   const { store } = workspace;
   const generatedAtTime = new Date(getCurrentTime());
   const electionRecord = store.getElectionRecord();
+  const markThresholds = store.getSystemSettings()?.markThresholds ?? undefined;
```
kshen0, line 37: Since `store.getSystemSettings()` [returns `SystemSettings | undefined`](https://github.com/votingworks/vxsuite/blob/65e2910634bc3cab072f4f40cd986432b4dca88c/apps/scan/backend/src/store.ts#L919), `store.getSystemSettings()?.markThresholds` will always evaluate to a truthy value or `undefined`. The nullish coalescing operator `??` isn't necessary in this case.

apps/central-scan/backend/src/readiness_report.ts
```diff
@@ -31,6 +31,7 @@ export async function saveReadinessReport({
   const generatedAtTime = new Date(getCurrentTime());
   const { electionDefinition, electionPackageHash } =
     store.getElectionRecord() ?? {};
+  const markThresholds = store.getSystemSettings()?.markThresholds ?? undefined;
```
kshen0, line 34: Same as other comment

apps/scan/frontend/src/screens/diagnostics_screen.tsx
kshen0: No problem. Some of this information is in the test output, though that output is admittedly noisy.

libs/ui/src/diagnostics/configuration_section.test.tsx
```diff
@@ -99,3 +99,27 @@ test('election, single precinct selected', () => {
...
+        definite: 0.07,
+        marginal: 0.05,
+        writeInTextArea: 0.05,
+      }}
+    />
+  );
+
+  screen.getByText(`Precinct: Precinct 1`);
+  screen.getByText(`Mark Threshold: 0.07`);
+  screen.getByText(`Write-in Threshold: 0.05`);
```
kshen0, line 124: Add a similar test for truncation?

apps/scan/backend/src/printing/readiness_report.ts
nbhatia823: Ah, true. Will update accordingly

libs/ui/src/diagnostics/configuration_section.test.tsx
```diff
@@ -123,3 +123,27 @@ test('election, mark threshold provided', () => {
...
+        definite: 0.070707,
+        marginal: 0.05,
+        writeInTextArea: 0.05050505,
+      }}
+    />
+  );
+
+  screen.getByText(`Precinct: Precinct 1`);
+  screen.getByText(`Mark Threshold: 0.0707`);
+  screen.getByText(`Write-in Threshold: 0.0505`);
```
nbhatia823, line 148: Actually, maybe this will match even in the case of the non-truncated value showing. Will consider alternative tomorrow

kshen0: `getByText` will look for an exact match unless you use a regexp or pass additional args: https://testing-library.com/docs/queries/about/#textmatch-examples

So this is fine. You could also make your test value nonrepeating eg 0.1234567

kshen0 (approved):

libs/ui/src/diagnostics/configuration_section.test.tsx
nbhatia823: With those examples I don't feel convinced this will work as behaved. Because whether the screen says `Mark Threshold: 0.0707` or `Mark Threshold: 0.07070707`, I can anticipate `Mark Threshold: 0.0707` matching them both, as that substring is an exact match for both. It seems regex could actually be better, to indicate the end of the string. Sorry if I'm overcomplicating it!

And I don't think the nonrepeating number would change this

kshen0: Talked offline; my read of the docs is `getByText` performs an exact string match so `Mark Threshold: 0.0707` will not match `Mark Threshold: 0.07070707`.

Nonrepeating numbers is just a small readability improvement and catches any bugs in `truncate` that would result in eg. truncating the first part of decimal instead of the last.

nbhatia823: Can confirm @kshen0's understanding was correct, as `getByText` performs full string match. Tested [here](https://app.circleci.com/pipelines/github/votingworks/vxsuite/17182/workflows/6879182f-247b-43f7-a4af-e57a6375af62/jobs/679277), which showed that `Mark Threshold: 0` was not picked up when the screen showed `Mark Threshold: 0.0707`. So, I will leave the test as is, except for changing the number pattern to be non-repeating.

PR #5429, opened 2024-09-24
[eslint-plugin-vx] Remove self-dependency to fix packaging with npm v10

kofi-q:
## Overview

`npm pack` is failing for `eslint-plugin-vx` after upgrading to node20/npm10.
Removing the self-dependency fixes it, so taking that out for now until we figure out a way for that package to safely lint itself.

## Testing Plan
- Tested with vxsuite-complete-system build script


eventualbuddha:
I'd rather not do this if we can avoid it, unless it doesn't actually affect the linting experience on the CLI and in-editor. Per our Slack discussion, maybe we can just use `pnpm pack` instead?

kofi-q:
> I'd rather not do this if we can avoid it, unless it doesn't actually affect the linting experience on the CLI and in-editor. Per our Slack discussion, maybe we can just use pnpm pack instead?

Yeah, was planning on looking into `pnpm pack` to replace the `npm pack` step - just seemed like a slightly larger surface area to attempt to change right now to unblock the build (in a part of the build code I'm not familiar with), so thought we'd start with the more isolated fix.

Didn't think this would affect linting in the rest of the codebase, but haven't checked explicitly - can take a look. Anything in particular to be sure to check?

eventualbuddha:
I suspect it'd only affect `pnpm lint` and editor auto-fix/formatting within `eslint-plugin-vx`. If so, that's acceptable as long as we come back to it at some point. If it affects either of those things in other packages I'd say that's a non-starter.

kofi-q:
Yeah, linting and auto-fixing seem to be working fine in other packages I've spot-checked (both in the IDE and command line). I believe this just affects `eslint-plugin-vx` linting.

arsalansufi (approved):
Thanks for confirming that this doesn't affect linting elsewhere! I'm okay with merging this, and if you could ticket the follow up to look at using `pnpm pack` instead of `npm pack`, that'd be great! Feel free to tag the issue with the v4.0 milestone and we can opportunistically get to it

PR #5431, opened 2024-09-24
Print through a single chromium instance across apps

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/5260
Updates our handling of chromium to only ever launch one browser at startup instead of launching a new one everytime we render. Original context in  [previous pr
](https://github.com/votingworks/vxsuite/pull/5258#pullrequestreview-2236077592) by @arsalansufi :
There was a comment in that PR to see if there was a way to abstract this out so that libs/printing would handle holding the browser and the individual apps/tests could be more oblivious to it. Ultimately I struggled to really make this a reality but decided to keep the browser inside of a new SimpleRenderer class in case we need to add more logic / things in memory in the future. It didn't make sense to combine this with the printer handlers since those are different by app and not all apps physically print so it seemed like it increased the scope of this PR too much to refactor all of that. I hit some issues in naming  and thinking about this with the existing hmpb and VxDesign Renderer classes and concepts. Open to feedback on better ideas.... 

## Demo Video or Screenshot
n/a

## Testing Plan
Apps compiled and ran, ran tests. Did a few sanity test prints. 

## Checklist


jonahkagan:
Checked this PR out as I've done a bit of exploration of this before with the HMPB renderer. I also ran into some problems there, but they were specific to the complexities of rendering HMPBs. FWIW, I think that you shouldn't worry about the HMPB renderer - we should just keep it completely separate for now since it has pretty different usage.

What problems did you run into with caching the browser inside libs/printing? I just did a quick test of that approach and it seemed to work at first glance, though I'm likely just missing what it breaks. https://github.com/votingworks/vxsuite/compare/jonah/cache-printing-browser.

libs/printing/src/__image_snapshots__/roll-document-2.png
adghayes, line 1: Why did this new snapshot page get created?

apps/central-scan/backend/src/app.diagnostics.test.ts
```diff
@@ -23,20 +28,36 @@ jest.mock(
...
 );
 
+jest.mock(
+  '@votingworks/printing',
+  (): typeof import('@votingworks/printing') => ({
+    ...jest.requireActual('@votingworks/printing'),
+    createSimpleRenderer: jest.fn(),
+  })
+);
+
```
adghayes, line 38: Is there a way to avoid mocking this, but instead toggle the `jest` timers at appropriate times in the test to get the desired behavior?

adghayes (commented):
I still think it would be nice to see the renderer cached in `libs/printing`, although I haven't looked at @jonahkagan's approach yet. It could avoid all the passing around the renderer that's necessary in the apps now. 

I can see a case for having access to the `renderer` in the apps if we need to restart / manipulate the instance, but we haven't seen a need for that yet. 

In the current approach, it looks like we're using `cleanup` in some apps' test suites but not others (I don't see it in VxAdmin tests). Not sure if it matters, just noting.

carolinemodic:
I guess I wasn't totally sure how to cache it in libs/printing, will play with Jonah's version of that more and report back

eventualbuddha:
> I can see a case for having access to the `renderer` in the apps if we need to restart / manipulate the instance, but we haven't seen a need for that yet.

I suggest we keep this use case in mind for this. Having a long-lived Chromium instance probably will get stuck eventually, and having some way to kill and recreate it would be nice. This probably doesn't need to be exposed to the apps, though.

> In the current approach, it looks like we're using `cleanup` in some apps' test suites but not others (I don't see it in VxAdmin tests). Not sure if it matters, just noting.

This is part of the reason I actually think I'd prefer the more explicit management of the renderer that this PR has now. We already don't do a great job of cleaning up resources when we're done with them, and I fear that making some global state in `@votingworks/printing` be the owner of the renderer will just exacerbate this problem.

carolinemodic:
ok I feel stupid this is much easier then I had originally made it. Still a few things to discuss but I'm going to open a fresh PR.

PR #5440, opened 2024-09-26
Print through a single chromium instance cached in libs/printing

carolinemodic:
## Overview
After discussion in https://github.com/votingworks/vxsuite/pull/5431 look into this caching approach. Overall worked well with the main issue being test cleanup. I initially cleaned up the cached browser through the jest globalTeardown function but found that was still giving me warning about things being open (despite logging confirming it was closing the browser) and failing CI in bmd_ballot_fixtures. When I instead cleaned up in a global afterAll that was resolved and I noticed that is how we are already cleaning up other state in VxAdminBackend so decided that that seemed like the safer approach. 

Note the safeguard I add in the caching for the isConnected() check is in case a browser was closed mistakenly, I have never seen this happen in testing (unless I had a bug) but it seemed like a good safety to make sure we don't just break in production if that happens. Open to removing though. 

## Demo Video or Screenshot
N/A

## Testing Plan
Compiled apps, ran tests, manual testing of prints 

## Checklist

### DIFF
--- a/libs/bmd-ballot-fixtures/package.json
+++ b/libs/bmd-ballot-fixtures/package.json
@@ -19,7 +19,7 @@
     "stylelint:run": "stylelint 'src/**/*.{js,jsx,ts,tsx}'",
     "stylelint:run:fix": "stylelint 'src/**/*.{js,jsx,ts,tsx}' --fix",
     "test": "is-ci test:ci test:watch",
-    "test:coverage": "TZ=America/Anchorage jest --coverage",
+    "test:coverage": "TZ=America/Anchorage jest --coverage --detectOpenHandles",
     "test:ci": "TZ=America/Anchorage jest --coverage --reporters=default --reporters=jest-junit --maxWorkers=6",
     "test:watch": "TZ=America/Anchorage jest --watch",
     "pre-commit": "lint-staged"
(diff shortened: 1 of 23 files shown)


libs/bmd-ballot-fixtures/package.json
```diff
@@ -19,7 +19,7 @@
     "stylelint:run": "stylelint 'src/**/*.{js,jsx,ts,tsx}'",
     "stylelint:run:fix": "stylelint 'src/**/*.{js,jsx,ts,tsx}' --fix",
     "test": "is-ci test:ci test:watch",
-    "test:coverage": "TZ=America/Anchorage jest --coverage",
+    "test:coverage": "TZ=America/Anchorage jest --coverage --detectOpenHandles",
```
eventualbuddha, line 22: Intentional?

carolinemodic: oops no, thanks

adghayes (approved):
LGTM!

I think @eventualbuddha's reservations are reasonable, but if we can manage test cleanup with global state I think this is safe to try. If we get to a point where we have to restart or manage the browser, we'd have to make changes regardless of the approach it is just slightly harder with a cached approach. Ultimately, I appreciate that this reduces clutter (and code!).

Issue #5158, opened 2024-07-31
VxAdmin: If you import CVRs from VxScan into VxAdmin, then go back to VxScan and scan one more ballot, you can't re-import that CVR export and just import the new CVRs


arsalansufi:
This works in v3.1 but doesn't seem to in v4: https://votingworks.slack.com/archives/CJU9MSC6S/p1722037210319939

Update: I (Arsalan) was able to do this just fine last week, so not sure if something else is at play here. If we can't repro, I'm fine with closing

arsalansufi:
I was able to do this just fine last week, so not sure if something else is at play here

PR #5443, opened 2024-09-26
VxAdmin: Allow continuous CVR exports to be re-imported

jonahkagan:
## Overview
Fixes https://github.com/votingworks/vxsuite/issues/5158

Previously, if you imported a continuous CVR export from VxScan into VxAdmin, then scanned more ballots and tried to re-import the same export, the "Load" button would be disabled. This fixes the logic for disabling that button to check the export timestamp, which changes when new ballots are exported (instead of just checking the file name, which doesn't change). The backend already knows how to only import the newly added CVRs, so the only fix needed is enabling the button.
## Demo Video or Screenshot
Skipping since this is fixing a regression

## Testing Plan
Manual test, updated automated tests

## Checklist


nbhatia823 (approved):
LGTM!

Issue #5368, opened 2024-09-09
Fix ERR CDF to pass NIST CDF Validator


carolinemodic:
In checking other CDF files I noticed I was getting errors running the NIST CDF validator against the ERR CDF. I generated my ERR CDF by importing a election definition (in cdf format as I was also testing that) and exporting the cdf of results, with no data actually loaded (not sure if that matters to my errors). 

Attached my sample ERR file and the output from the nist tool. 
Nist tool github: https://github.com/usnistgov/cdf-test-method?tab=readme-ov-file (download from releases tab) 
Usage instructions (you have to do sudo bash ./testcdf): https://pages.nist.gov/cdf-test-method/USAGE.html 

You can just `cat` the report file to see all of the metadata.

  
[Sample ERR File](https://github.com/user-attachments/files/16936847/err.json)

[Sample output in slack (can't upload html to github) ](https://votingworks.slack.com/files/U01DC19BN6N/F07M0MUL533/vx-err-report.output.html)

carolinemodic:
@adghayes @kshen0

adghayes:
@kshen0 can this be closed?

PR #5445, opened 2024-09-27
CDF ERR export: conform to NCName and ID uniqueness constraints

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/5368

Fixes ERR validation errors on exported CDF ERR files from VxAdmin. The errors fall into 2 categories:

1. IDs are `NCName`s ([spec](https://www.w3.org/TR/xml-names/#NT-NCName)) which can only start with letters, `_`, or `.`. They also can't have the colon character, though our IDs are unlikely to include this character.
2. IDs must be globally unique, but we used the same candidate ID for both `Candidate` and `CandidateSelection` objects. 

The fixes are:
1. Prefix all IDs with `vx_` and remove the `:` character.
2. Use a concatenation of `contest` and `candidate` IDs for `CandidateSelection`. `CandidateSelection` describes the tally for a single candidate in in a single contest so this concatenation seemed like the most accurate option.

## Demo Video or Screenshot

Not a new feature, but exporting ERR files from VxAdmin still works:

https://github.com/user-attachments/assets/ff07850c-5638-4bfc-884b-ee511035ac36

Here's an example ERR file and the output of running the validator tool on that file.

[err.json](https://github.com/user-attachments/files/17168511/err.json)

```
<?xml version="1.0" encoding="UTF-8"?>
<reports xmlns="http://www.xproc.org/ns/xvrl">
   <metadata xmlns:tm="http://itl.nist.gov/ns/voting/test-method">
      <creator name="NIST Common Data Format Test Method" version="1.0.1"/>
      <timestamp>2024-09-27T11:21:40.61-07:00</timestamp>
      <summary>Please review each report to determine conformance to the applicable test assertions / requirements</summary>
      <supplemental tm:xproc-engine-version="1.2" tm:psvi-supported="false" tm:xproc-engine-vendor="&lt;xml-project /&gt; Achim Berndzen" tm:xproc-engine-name="MorganaXProc-IIIse">
         <tm:validity>valid</tm:validity>
         <tm:validity-reason/>
         <tm:runcode>Episode-1727461299329</tm:runcode>
      </supplemental>
   </metadata>
   <report>
      <metadata>
         <timestamp>2024-09-27T11:21:39.31-07:00</timestamp>
         <document href="file:///path/to/cdf-test-method-v1.0.1/err.json"/>
         <schema href="http://itl.nist.gov/ns/voting/1500-100/v2/json" schematypens="JsonSchema"/>
         <validator name="networknt/json-schema-validator"/>
      </metadata>
      <digest/>
   </report>
   <report>
      <metadata>
         <timestamp>2024-09-27T11:21:40.05-07:00</timestamp>
         <document href="file:///path/to/cdf-test-method-v1.0.1/err.json"/>
         <schema href="http://itl.nist.gov/ns/voting/1500-100/v2" schematypens="http://www.w3.org/2001/XMLSchema"/>
         <validator name="org.apache.xerces.jaxp.validation.XMLSchemaFactory"/>
      </metadata>
      <digest/>
   </report>
   <report>
      <metadata>
         <timestamp>2024-09-27T11:21:40.6-07:00</timestamp>
         <document href="file:///path/to/cdf-test-method-v1.0.1/err.json"/>
         <schema href="http://itl.nist.gov/ns/voting/1500-100/v2/sch" schematypens="http://purl.oclc.org/dsdl/schematron"/>
         <validator name="SchXslt"/>
      </metadata>
      <digest/>
   </report>
</reports>
```

## Testing Plan

Updated existing correctness test.

### DIFF
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -932,9 +932,7 @@ function buildApi({
             isTestMode,
             writeInCandidates,
             machineConfig: getMachineConfig(),
-          }),
-          null,
-          2
+          })
         ),
       });
 
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -921,6 +921,7 @@ function buildApi({
       assert(electionResults);
 
       debug('exporting CDF election results report JSON file: %o', input);
+
       const exportFileResult = await exportFile({
         path: input.path,
         data: JSON.stringify(
@@ -931,7 +932,9 @@ function buildApi({
             isTestMode,
             writeInCandidates,
             machineConfig: getMachineConfig(),
-          })
+          }),
+          null,
+          2
         ),
       });
 
--- a/apps/admin/backend/src/app.err_cdf_export.test.ts
+++ b/apps/admin/backend/src/app.err_cdf_export.test.ts
@@ -200,28 +200,28 @@ it('exports results and metadata accurately', async () => {
     GeneratedDate: expect.anything(),
     IsTest: true,
     Issuer: 'Test Ballot',
-    IssuerAbbreviation: '00701',
+    IssuerAbbreviation: 'vx_00701',
     SequenceEnd: 1,
     SequenceStart: 1,
     Status: 'unofficial-complete',
     VendorApplicationId: 'VxAdmin, version dev',
   });
   expect(GpUnit?.map((gpUnit) => gpUnit['@id'])).toEqual([
-    'nh',
-    '00701',
-    'town-id-00701-precinct-id-default',
+    'vx_nh',
+    'vx_00701',
+    'vx_town-id-00701-precinct-id-default',
   ]);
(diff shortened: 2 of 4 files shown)


apps/admin/backend/src/app.ts
```diff
@@ -931,7 +932,9 @@ function buildApi({
             isTestMode,
             writeInCandidates,
             machineConfig: getMachineConfig(),
-          })
+          }),
+          null,
+          2
```
jonahkagan, line 937: Is this just for debugging? Or did you want the exported JSON to be more readable? It could inflate the size a good amount

jonahkagan (approved):

apps/admin/backend/src/app.ts
kshen0: I wanted the exported JSON to be more readable, but I don't feel strongly about it.

PR #5446, opened 2024-09-27
bug: handling double endorsement tabulation

adghayes:
## Bug Description  

We've noticed that on a ballots with two ballot positions for the same candidate (due to a double party endorsement) we hit a couple interrelated tabulation bugs. First, when both positions are filled out, the scanner accepts the ballot (correct) but tabulates it as an overvote:

valid_double

Second, we are not detecting overvotes when the second position is filled in, but the first is not. For example, this is not rejected as an overvote:

overvote_2

But this is correctly rejected as an overvote:

overvote_1

## Bug Cause

It seems that when we introduced double endorsements we broke with an underlying assumption that each candidate had only one ballot position, which we did not completely root out.

First, when we convert marks to votes, we were not deduplicating candidates. That is why the first example was eventually tabulated as an overvote.

Second, when we analyze marks for adjudication reasons, we just analyze the first mark we see each candidate option. That's why the second example was adjudicated properly, but the third example was not. 

## Bug Fix

When we convert marks to votes, we now deduplicate candidates. That is the first commit and addresses improper tabulation.

When we analyze marks for adjudication reasons, if there are multiple marks for a candidate, we now base our analysis off of the _more filled in mark_ rather than the _first_ mark. I'm not 100% sure this is the right handling for all cases, but I think within our current adjudication framework it is sufficient. If either mark is filled in, we count it. If neither mark is filled in, we don't count it. If neither is filled in, but either is marginal, we flag it as a marginal mark. We don't actually have marginal mark adjudication but I wanted to keep behavior consistent.

## Testing

After the changes, I confirmed that all the examples above scan as expected, as tabulated as expected, and appear in the CVR as expected. I've added automated testing to enforce expected behavior.

## Small Loose Ends

We may need to revisit this issue insofar as how these marks appear in the CVR files. Currently, with each position we specify in the CVR we specify an `OptionPosition` which is an index of the candidate in the race. Due to the single-candidate-single-position assumption, we only ever track this as the _first_ position in the race. For example, say a candidate is positions 0 and 1. Both sets of mark metadata in the original snapshot will show 0, and only the 0 position will appear in the modified snapshot.

First, we will need to fix this so the correct option positions appear. This may involve tracking more metadata about each mark, so could be more involved and gets closer to the interpreter.

Second, we want to consider the behavior we want in the modified snapshot in the first example. Do we want the modified snapshot to have both marks, and VxAdmin sorts it out? Or do we just choose one?

### DIFF
--- a/libs/ballot-interpreter/src/adjudication_reasons.test.ts
+++ b/libs/ballot-interpreter/src/adjudication_reasons.test.ts
@@ -9,7 +9,7 @@ import {
   YesNoContest,
 } from '@votingworks/types';
 import { electionTwoPartyPrimaryDefinition } from '@votingworks/fixtures';
-import { assert, typedAs } from '@votingworks/basics';
+import { assert, find, typedAs } from '@votingworks/basics';
 import { allContestOptions } from '@votingworks/utils';
 import {
   getAllPossibleAdjudicationReasons,
@@ -43,13 +43,17 @@ const ballotMeasure3 = electionTwoPartyPrimaryDefinition.election.contests.find(
   ({ id }) => id === 'fishing'
 ) as YesNoContest;
 
+type GetAllPossibleAdjudicationReasonsOptionStatus = Parameters<
+  typeof getAllPossibleAdjudicationReasons
+>[1][0];
+
 function generateMockContestOptionScores(
   contest: AnyContest,
   overrides: Record<
     ContestOptionId,
     { markStatus?: MarkStatus; writeInAreaStatus?: WriteInAreaStatus }
   >
-): Parameters<typeof getAllPossibleAdjudicationReasons>[1] {
+): GetAllPossibleAdjudicationReasonsOptionStatus[] {
   return [...allContestOptions(contest)].map((option) => ({
     option,
     markStatus: overrides[option.id]?.markStatus ?? MarkStatus.Unmarked,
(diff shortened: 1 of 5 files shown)


adghayes:
@arsalansufi test is failing due to snapshots changing. The interpretation snapshots lose one of their two references to the selection, since we're deduplicating. 

I think that I should just update the snapshots - agree?

Perhaps there is a good reason for the selections being duplicated in the interpretation blob. @eventualbuddha do you have any thoughts? I want to deduplicate before we hit the interpretation blob because we have multiple consumers (VxScan tabulation & CVR export) and don't want to worry about this in more places. Perhaps we should have an abstraction over marks closer to the Next interpreter so that we're not worrying about this in the two places we already are.

arsalansufi:
Agreed re updating snapshots!

libs/ballot-interpreter/src/adjudication_reasons.test.ts
```diff
@@ -299,3 +303,116 @@ test('an unmarked write-in can trigger the overvote reason', () => {
...
+      zebraOptionB: marginalZebraOption,
+      expected: [marginalAdjudicationReason, undervoteAdjudicationReason],
+    },
+    {
+      description: 'unmarked, unmarked',
+      zebraOptionA: unmarkedZebraOption,
+      zebraOptionB: unmarkedZebraOption,
+      expected: [undervoteAdjudicationReason],
+    },
+  ])(
```
arsalansufi, line 407: Would be nice to add some kind of test for overvote detection, too. I see that that might require a different contest, e.g, a vote for 1

arsalansufi (approved):
LGTM!! Thanks for investigating on fixing on short notice!

libs/ballot-interpreter/src/adjudication_reasons.ts
```diff
@@ -7,10 +7,31 @@ import {
...
+      return 1;
+    case MarkStatus.Unmarked:
+      return 0;
+    /* istanbul ignore next */
+    default:
+      throwIllegalValue(markStatus);
+  }
+}
+
+function sortMarkStatusDescending(
```
eventualbuddha, line 28: nit: I'd name this `compareMarkStatusDescending` since it's not actually sorting anything.

eventualbuddha (approved):

arsalansufi:
@adghayes There's a new [v3.1.2-release-branch](https://github.com/votingworks/vxsuite/tree/v3.1.2-release-branch) for you to merge into! You can change the base branch of this PR to that branch and then merge

PR #5452, opened 2024-09-30
Update VxAdmin Readiness Report Ballot Style List

carolinemodic:
## Overview
Updates the ballot style list in the admin readiness report to list the parent ballot style id along with all supported languages of that ballot style rather then a list of all language specifc ballot style ids. 

If we can assume all ballot styles have the same languages we could make the UI much cleaner here but I don't think we can make that assumption, let me know if I'm wrong there. Open to best copy formatting ideas the parenthetical on Spanish including (US) makes this look bad but that is build into the js Intl.DisplayNames call we are making to format the strings. 

First commit is making the electionPrimaryWithPrecinctSplits fixture election multi-language for usage in this and my next PR. Following current practice I removed grid layoutes from the libs/fixtures election.json. 

## Demo Video or Screenshot
Screenshot-VxAdmin-2024-09-30T21:22:07 211Z

## Testing Plan
Viewed readiness report, saved and verified pdf, ran tests 

## Checklist


libs/hmpb/fixtures/primary-election/election.json
jonahkagan, line 1: I'm wondering if we should make one of the other primary election fixtures multi-language instead of this one so that the libs/hmpb fixtures aren't affected. The general election HMPB fixture is already multi-language, so that base is covered. That might simplify things a bit. Or did you pick this one intentionally?

libs/ui/src/diagnostics/configuration_section.tsx
```diff
@@ -93,7 +132,10 @@ export function ConfigurationSection({
       {!(expectPrecinctSelection && !precinctSelection) && (
         <P>
           <SuccessIcon /> Ballot Styles:{' '}
-          {getBallotStyleIds(election, precinctSelection).join(', ')}
+          {getBallotStylesConfigurationDetails(
+            election,
+            precinctSelection
+          ).join(', ')}
```
jonahkagan, line 138: This might be a bit easier to read as a list or a table, e.g.:

Ballot Styles:
- 1 - English, Chinese, Spanish (US)
- 2 - English, Chinese, Spanish (US)

 Ballot Styles
| ID | Language |
| --- | --------- |
| 1  | English, Chinese, Spanish (US) |
| 2  | English, Chinese, Spanish (US) |

libs/ui/src/diagnostics/configuration_section.tsx
```diff
@@ -32,18 +35,54 @@ function getPrecinctSelectionName(
...
+      const languageCodes = new Set<LanguageCode>();
+      // If there is only one language per ballot style we don't need to display it.
+      if (ballotStylesInGroup.size <= 1) {
+        return bsParentId;
+      }
+      for (const bs of ballotStylesInGroup) {
+        if (bs.languages) {
+          for (const lang of bs.languages) languageCodes.add(lang);
+        }
+      }
```
jonahkagan, line 75: If you'd like, there's `iter.toSet` or `unique` from @votingworks/basics

jonahkagan (approved):

libs/hmpb/fixtures/primary-election/election.json
carolinemodic: I thought the precinct splits might be a good confusing edge case in combination with languages but after playing with it I don't think that really actually matters

libs/ui/src/diagnostics/configuration_section.tsx
carolinemodic: Yeah I thought about this but wasn't sure with printing the report if it would be too long, will mock up and get Matt's perspective though

carolinemodic:
Updating to a table layout after consulting with @mattroe 
Screenshot 2024-10-01 at 2 36 30 PM

libs/hmpb/fixtures/primary-election/election.json
carolinemodic: I tried experimenting with other elections and they each seemed to be more and more difficult in terms of the number of broken tests and things to untangle so leaving this for now, may revisit in my next pr

Issue #5455, opened 2024-10-01
Make sure v3.1.+ patches (or versions of those patches) make it onto main


arsalansufi:
- [x] https://github.com/votingworks/vxsuite/pull/5453
- [x] https://github.com/votingworks/vxsuite/pull/5446
- [x] https://github.com/votingworks/vxsuite/issues/5398
- [ ] https://github.com/votingworks/vxsuite/issues/5397 ➡ Tracking via https://github.com/votingworks/vxsuite/issues/5555

Generally check https://github.com/votingworks/vxsuite/tree/v3.1.2-release-branch for anything of relevance

arsalansufi:
Tracking the last unchecked box in its own issue: https://github.com/votingworks/vxsuite/issues/5555

PR #5456, opened 2024-10-01
bug: handling double endorsement tabulation (`main`)

adghayes:
Part of #5455. See #5446 for details.

### DIFF
--- a/libs/ballot-interpreter/src/__snapshots__/interpret_nh_hmpb.test.ts.snap
+++ b/libs/ballot-interpreter/src/__snapshots__/interpret_nh_hmpb.test.ts.snap
@@ -6024,14 +6024,6 @@ exports[`HMPB - m17 backup Interprets all ballots correctly: ballotId=af6d30a5-9
               "Republican-f0167ce7",
             ],
           },
-          {
-            "id": "Edward-Randolph-bf4c848a",
-            "name": "Edward Randolph",
-            "partyIds": [
-              "Democratic-aea20adb",
-              "Republican-f0167ce7",
-            ],
-          },
         ],
       },
     },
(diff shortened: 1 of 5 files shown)


adghayes:
# Discussion of Future Changes

This addresses the bug and any tabulation concerns, but there's an open question as to how best represent this use case in the CVR. 

Currently, the original snapshot will show two contest selections for the same candidate. It will not have the correct `OptionPosition` because those are calculated based off of the election definition, where a doubly endorsed candidate will only appear once. In other words, the `OptionPosition` field is potentially wrong for all contest selections in races with a doubly endorsed candidates. `OptionPosition` field needs to be calculated based on the ballot layout rather than the election definition.

The modified snapshot will only show one contest selection for the contest. It will not be apparent that there are two marks from looking at the modified snapshot. Keep in mind - the modified snapshots can, according to the spec, contain "virtual" contest selections for options that were never marked directly, such as in a straight party vote in jurisdictions which allow that. So compressing two marks into one selection is fair game, as far as I'm concerned.

If we do show two selections on the modified snapshot, it's unclear how to assign them standard CDF statuses. The both obviously are `HasIndication.Yes` but are they `IsAllocable.Yes`? Which one is allocable if both are marked? If both are allocable, isn't that a bit confusing? I think the most accurate thing would be to make them both allocable and give them both special statuses.

We probably will have to go that direction. Currently, the vote data that gets into the modified snapshot is the vote summary distilled from the mark data. Eventually, for a fuller fledged adjudication flow, we need higher fidelity mark data available in VxAdmin. Getting there requires touching the data pipeline at a lot of points - how it's stored in VxScan, how it's recorded in the CVR file, how it's stored in VxAdmin - that make it a far larger project.

eventualbuddha (approved):

Issue #5348, opened 2024-09-03
VxScan: Display settings do not reset on polls close


arsalansufi:
Can't change any of the settings back while polls are closed

Issue automatically created from Slack feedback reported by pius: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1725400046299259?thread_ts=1725399990.498999&cid=CJU9MSC6S

PR #5462, opened 2024-10-01
Reset settings on poll close for VxScan

nbhatia823:
## Overview

https://github.com/votingworks/vxsuite/issues/5348

## Demo Video or Screenshot

Old behavior: after polls close and poll worker card removal, old settings/themes persist

https://github.com/user-attachments/assets/dc8584a4-4940-4571-bb2b-17562d73aba8

New behavior: after polls close and poll worker card removal, settings/themes reset

https://github.com/user-attachments/assets/06417712-b23e-413d-9678-6feb433710b3

## Checklist

### DIFF
--- a/apps/scan/frontend/src/components/voter_settings_manager.tsx
+++ b/apps/scan/frontend/src/components/voter_settings_manager.tsx
@@ -7,7 +7,7 @@ import {
 } from '@votingworks/ui';
 import { DefaultTheme, ThemeContext } from 'styled-components';
 import { LanguageCode } from '@votingworks/types';
-import { getAuthStatus, getScannerStatus } from '../api';
+import { getAuthStatus, getPollsInfo, getScannerStatus } from '../api';
 
 /**
  * Side-effect component for monitoring for auth and voter session changes and
@@ -19,6 +19,7 @@ export function VoterSettingsManager(): JSX.Element | null {
 
   const authStatusQuery = getAuthStatus.useQuery();
   const scannerStatusQuery = getScannerStatus.useQuery();
+  const pollsInfoQuery = getPollsInfo.useQuery();
 
   const { reset: resetLanguage, setLanguage } = useLanguageControls();
   const currentLanguage = useCurrentLanguage();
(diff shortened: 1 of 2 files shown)


kshen0 (commented):
LGTM!

kshen0 (approved):

apps/scan/frontend/src/components/voter_settings_manager.tsx
```diff
@@ -30,6 +31,19 @@ export function VoterSettingsManager(): JSX.Element | null {
...
+    onChange: (newState) => {
+      // Reset to default theme when polls close
+      if (newState === 'polls_closed_final') {
+        voterSettingsManager.resetThemes();
+        resetLanguage();
+        setVoterSessionTheme(null);
+        setVoterLanguage(null);
+      }
+    },
+  });
```
jonahkagan, line 45: Sorry to come in with a late review after this was already merged - I only just saw it and was curious. In general, it's a best practice to put changes to frontend-state in the handler for the user action that triggered them. In this case, that would be the [`closePolls` function](https://github.com/votingworks/vxsuite/blob/main/apps/scan/frontend/src/screens/poll_worker_screen.tsx#L335-L349). `useQueryChangeListener` is only intended to be used as a workaround for when we can't implement this pattern since the user action isn't triggered via the screen (e.g. if they are interacting with a different piece of hardware). I recommend [this article from the React docs](https://react.dev/learn/you-might-not-need-an-effect) that digs into these concepts more.

nbhatia823: No worries, I appreciate it! Happy to make a subsequent PR. Thanks for explaining and for linking the article. 

In this case, I was seeing that `VoterSettingsManager` has the following comment: ` Side-effect component for monitoring for auth and voter session changes and resetting/restoring voter display settings and language choice as needed.`

And it seems to be self-contained in managing settings, and I didn't see a streamlined way to update the settings within the `PollWorkerScreen` component, so I assumed the side-effect reaction might be our expected approach. Anything in that thinking that you could help reframe?

jonahkagan: I think this is a limitation of the side-effect component idea. Now `PollWorkerScreen` needs a way to access the voter settings state but can't get to it. The state needs to be [lifted up to a common parent](https://react.dev/learn/sharing-state-between-components) of whatever is consuming the voter settings state and `PollWorkerScreen`.

A few ideas:
- Since there's already a `VotingSettingsManagerContext`, we could move the state to whatever component creates that context, and then expose some functions in the context to modify that state. I'd maybe model this as a stack, since the behavior we want is to be able to push and pop themes as users log in/log out.
- We could move the state (and associated listeners) to `AppRoot`.

nbhatia823: Ah gotcha, so we are talking about making a slightly larger structural change. Those two approaches both make sense to me, I'll do some playing around and come up with a solution that feels fitting!

Issue #5190, opened 2024-08-01
VxAdmin: Manually added tallies are listed in the CVR table instead of under manual tallies


arsalansufi:
VxAdmin: manually tallies are listed in the CVR table instead of under manual tallies. When you try to remove cvrs, the manual tallies don’t get removed. I was so confused

Issue automatically created from Slack feedback reported by jonah: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1722534585904659?thread_ts=1722534585.904659&cid=CJU9MSC6S

PR #5464, opened 2024-10-01
VxAdmin: Redesign tally screen with tabs for CVRs and manual tallies

jonahkagan:
## Overview

Fixes https://github.com/votingworks/vxsuite/issues/5190

The existing tally screen was a bit confusing, listing manual tallies in the CVR table and then linking to a different screen to view the manual tally details. This PR redesigns this screen to treat CVRs and manual tallies as two separate parallel types of tallies, organizing them in two tabs. It also adds a safeguard to ensure that users that remove CVRs after finishing an L&A test also remember to clear the tallies.

## Demo Video or Screenshot
Loading and removing CVRs and manual tallies
https://github.com/user-attachments/assets/9c09e814-2272-419d-8b37-42f13c9b0f41

After marking results official
https://github.com/user-attachments/assets/621c3636-8771-4c1c-a36f-18501c2dec50

## Testing Plan

## Checklist


apps/admin/frontend/src/screens/tally/manual_tallies_tab.tsx
jonahkagan, line 1: hide whitespace for this one

apps/admin/frontend/src/screens/tally/confirm_remove_cvrs_modal.tsx
```diff
@@ -0,0 +1,119 @@
...
+  function removeCvrs() {
+    clearCastVoteRecordFilesMutation.mutate(undefined, {
+      onSuccess: onClose,
+    });
+  }
+
+  async function removeCvrsAndManualResults() {
+    await Promise.all([
+      clearCastVoteRecordFilesMutation.mutateAsync(),
+      deleteAllManualResultsMutation.mutateAsync(),
```
eventualbuddha, line 43: How do you feel about combining these to make a new mutation endpoint? I don't really like the idea of one failing and the other succeeding, though I _think_ that'd be okay here if a bit confusing.

eventualbuddha (approved):

apps/admin/frontend/src/screens/tally/confirm_remove_cvrs_modal.tsx
jonahkagan: Hmm, good question. I like that idea.

I peeked at the backend and there would also need to be some work done to manage the transactions. We're already not great about that (e.g. `clearCastVoteRecords` does two different db writes in separate transactions). 

Given this PR is focused on frontend changes, I think it's probably out of scope. But maybe there's a larger tech improvement to make in terms of transaction management?

Issue #5183, opened 2024-08-01
VxSuite: Review ballot style lists and only expose parent/english ballot style


arsalansufi:
VxAdmin: I find it slightly confusing in the manual tally flow how you have to select a ballot style but we have different ballot styles for each language. I know why we made those choices independently, but together it strikes me as odd as I'm guessing when people are entering manual tallies in general they are not trying to record the distinction of which tallies came from which languages?

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1722451195735169?thread_ts=1722451195.735169&cid=CJU9MSC6S

adghayes:
- [ ] Manual Tally Creation
- [ ] VxAdmin Report Generation
- [ ] VxMark Ballot Style Selection

PR #5467, opened 2024-10-02
VxAdmin: Update tabulation logic (manual tallies and reports) to use ballot style group

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/5183
In adding multi-language ballot styles as "sub" ballot styles in our election format we made the ballot style filters/groups for reporting in VxAdmin too specific. To protect privacy insitead of reporting on ballot styles athat are language specific like 1-Ma_en we should instead report on the aggregated grouping 1-Ma. 

In an effort to make it as clear as possible what is happening where in the codebase I did a pretty large sweeping renaming changes to hopefully always refer to it as ballotStyleGroupId when referencing the parent of a group and ballotStyleId refers to the specific sub-language style id. 

Manual tallies will now be entered on a "group" level as well. 

## Demo Video or Screenshot
Ballot styles replaced with language - agnostic ids through VxAdmin, screenshots not particularly illustrative for this change. 

## Testing Plan
Manually tested with a multi-language primary election getting reports in all formats (pdf, csv, err) for full tally report, and available formats (pdf, csv) for filtered and grouped tally reports with filters and groups by ballot style, party, etc. Manually tested basic WIA flow. Manually tested adding manual tallies and verifying tabulation results are correct with them. Etc. Tested both with generated cvrs and manually created ones with cvrs across different languages for the same grouped ballot style. 

## Checklist


carolinemodic:
cc @adghayes

libs/types/src/tabulation.ts
```diff
@@ -76,7 +76,7 @@ export interface Filter {
  * Attributes that always exist for every cast vote record.
  */
 export interface CastVoteRecordAttributes {
-  readonly ballotStyleGroupId: BallotStyleGroupId;
+  readonly ballotStyleGroupId: Id;
```
jonahkagan, line 79: Why this change?

libs/types/src/election.ts
```diff
@@ -308,15 +308,26 @@ export const PrecinctsSchema = z
...
 }
+
+export type BallotStyleGroupId = NewType<string, 'BallotStyleGroupId'>;
+export interface ParentBallotStyle {
+  readonly id: BallotStyleGroupId;
+  readonly precincts: readonly PrecinctId[];
+  readonly districts: readonly DistrictId[];
+  readonly partyId?: PartyId;
+  readonly languages?: readonly LanguageCode[]; // TODO(kofi): Make required.
+}
```
jonahkagan, line 329: I notice that in some places we use default language ballot styles and in some places we use `ParentBallotStyle`. I think having a mix of both concepts makes things a bit confusing. Can we pick one?

Also, if we go with parent ballot styles, maybe we could call it `BallotStyleGroup` instead and remove the languages field (keeping only the fields that are common across all of the ballot styles in the group).

jonahkagan (commented):
Thanks for separating out the different commits. Very helpful!

(Side note: there are now so many files that I can only view the PR commit by commit, Github freezes when I try to view the full diff)

libs/types/src/tabulation.ts
carolinemodic: I forget exactly but I think there were insurmountable other issues if this isn't a string as all other fields here are strings and that is utilized in other parts of the code base that generalize the keys

jonahkagan (approved):
💥 :shipit:

PR #5470, opened 2024-10-03
allow import of vx-exported CDF ERR

kshen0:
## Checklist


libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -117,10 +132,10 @@ function convertToYesNoContest(
   const undervotes = otherCounts?.Undervotes || 0;
 
   return {
-    contestId: contest['@id'],
+    contestId: trimVxIdPrefix(contest['@id']),
     contestType: 'yesno',
-    yesOptionId: yesOption['@id'],
-    noOptionId: noOption['@id'],
+    yesOptionId: trimVxIdPrefix(yesOption['@id']),
+    noOptionId: trimVxIdPrefix(noOption['@id']),
```
kshen0, line 138: I wrestled with whether I should include getters that wrap this pattern but in the end felt like it was overwrought. Open to feedback if this feels dangerous though.

libs/types/src/cdf/election-results-reporting/convert.ts
```diff
@@ -378,10 +399,25 @@ function validateCandidateIds(
...
+
+        const candidateId = trimVxIdPrefix(
+          assertDefined(
+            selection.CandidateIds,
+            `No candidate ID on selection: ${JSON.stringify(
+              selection,
+              null,
+              2
+            )}`
+          )[0]
```
kshen0, line 414: `CandidateIds` is not required by the schema but I can't think of any other way to relate `CandidateSelection` to `Candidate` except packing that info in `CandidateSelection['@id']` and then parsing the ID string. `CandidateIds` feels like it was designed for this purpose. If it's not de-factor required by industry convention, I think we should at least require it in our own products.

jonahkagan: I think this is the way the format is intended to be used. My guess is that maybe `CandidateIds` is not required because a write-in selection might not align to any candidate.

I've noticed in CDF docs that UML fields that are linked by ID usually don't have the `ID` suffix in the docs. I think that's a pattern.

libs/types/src/cdf/election-results-reporting/convert.test.ts
```diff
@@ -241,6 +280,56 @@ describe('getManualResultsFromErrElectionResults', () => {
     expect(results.ok()).toEqual(expected);
   });
 
+  test('yes no ContestSelections defaults "yes" to first ContestSelection', () => {
```
kshen0, line 283: FYI @adghayes we discussed avoiding this behavior in a previous PR, but even our fixture election couldn't be matched on `yes` and `no` strings (IIUC it uses "For" and "Against"). I added this in since it's seeming more like a necessary feature, but LMK if you have any concerns.

PR #5471, opened 2024-10-03
Software mitigation strategies for crashes and potentially time-consuming CVR resyncs following sporadic USB drive disconnects

arsalansufi:
## Overview

It's long been known that a USB drive disconnect in the middle of a CVR export operation causes a crash. What's been less known is the likelihood of a sporadic disconnect without physical USB drive removal. We now have real data on that front and saw two sporadic USB drive disconnects in the field this past election. That's high enough that we've decided to improve the recovery flow for this case.

The current recovery flow is as follows:
1. Machine accepts or rejects ballot
2. Machine saves CVR to its own drive
3. Machine saves CVR to USB drive
4. USB drive disconnects mid-write
5. Machine crashes and displays the "Something went wrong" screen
6. User restarts
7. Machine detects that the USB drive is out of sync on boot and prompts the user to sync CVRs
8. Machine simplistically syncs all CVRs from scratch

| Resync screen |
| -- |
| resync |

This PR makes a few critical changes.

## Smart/Fast Recovery Sync

First, the PR attempts a smart/fast recovery sync before falling back to a full sync, which can be slow if many ballots have been cast (~4 minutes for 1000 ballots in the most recent election). We actually already have a lot of infrastructure in place for this. We already maintain a queue of pending continuous export operations in durable storage for example. The recovery sync performs necessary cleanup, syncs the few CVRs still in the queue, and confirms that the Merkle tree hash of cast vote records on the USB drive matches that on the machine. The cleanup accounts for failures during random timestamp shuffling. And the final confirmation step ensures that we detect if the USB drive was swapped and needs a full sync, or if the export has been generally corrupted.

## Allowing Pausing Continuous Export

Second, the PR allows users to pause continuous export, via the election manager menu. This completely removes the USB drive from the critical voter path and presents a heavy-handed solution for the worst-case scenario that a USB drive repeatedly fails, and a replacement isn't available or an election official can't wait for a full sync because the voter line will grow too long. This setting allows you to remove the USB drive completely without triggering the "No USB drive detected" screen. We also don't write to the USB drive on polls close. The thinking is that you can export CVRs manually via the election manager menu after polls close and the release of some time pressure.

| Pause | Resume |
| -- | -- |
| pause | resume |

## Tweaking "Something Went Wrong" Screen

Third, this PR replaces the "Restart" button on the "Something went wrong" screen with a "Power Down" button. Users will then unplug and replug the machine to power it back on. Experience in the field plus @jdVotingWorks's (awesome) research/testing has found that a complete removal of power is necessary to get these USB drives to properly reregister. A restart, I'm learning, doesn't fully remove power.

## Other Misc Changes

There are other small changes included in this PR, namely for improved logging and making sure that errors during polls close are properly surfaced.

I'll be porting over all of the changes that are relevant here to v4 as well.

## Testing Plan

- [x] Lots of manual testing, see #issue-v311-crashes
- [x] Automated testing

## Checklist

### DIFF
--- a/libs/backend/src/cast_vote_records/export.ts
+++ b/libs/backend/src/cast_vote_records/export.ts
@@ -1,7 +1,9 @@
 import crypto from 'crypto';
+import { existsSync } from 'fs';
 import fs from 'fs/promises';
 import path from 'path';
 import {
+  computeCastVoteRecordRootHashFromScratch,
   computeSingleCastVoteRecordHash,
   HashableFile,
   prepareSignatureFile,
@@ -53,7 +55,10 @@ import {
 } from './build_cast_vote_record';
 import { buildCastVoteRecordReportMetadata as baseBuildCastVoteRecordReportMetadata } from './build_report_metadata';
 import { CanonicalizedSheet, canonicalizeSheet } from './canonicalize';
-import { updateCreationTimestampOfDirectoryAndChildrenFiles } from './file_system_utils';
+import {
+  recoverAfterInterruptedCreationTimestampUpdate,
+  updateCreationTimestampOfDirectoryAndChildrenFiles,
+} from './file_system_utils';
 import { readCastVoteRecordExportMetadata } from './import';
 import { buildElectionOptionPositionMap } from './option_map';
 
@@ -130,6 +135,10 @@ interface PrecinctScannerOptions {
    * support full exports.)
    */
   isFullExport?: boolean;
+  /**
+   * An export performed to recover after an error, e.g., a sporadic USB disconnect
+   */
+  isRecoveryExport?: boolean;
 }
 
 /**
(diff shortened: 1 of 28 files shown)


libs/ui/src/error_boundary.tsx
```diff
@@ -28,6 +28,8 @@ export class ErrorBoundary extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {};
+
+    this.handleUnhandledRejection = this.handleUnhandledRejection.bind(this);
```
arsalansufi, line 32: This is an important one for us to up-port to v4. I noticed errors in kiosk-browser because of this function's lack of binding when it's called. Causing the error boundary itself to hang in rare circumstances. Will document more fully in a standalone PR for v4!

libs/backend/src/cast_vote_records/export.ts
```diff
@@ -679,12 +698,36 @@ export async function exportCastVoteRecordsToUsbDrive(
     scannerStore.clearCastVoteRecordHashes();
   }
 
-  const exportDirectoryPathRelativeToUsbMountPoint =
-    await getExportDirectoryPathRelativeToUsbMountPoint(exportContext);
+  const isRecoveryExport =
+    exportOptions.scannerType === 'precinct' && exportOptions.isRecoveryExport;
```
nbhatia823, line 702: What are we confirming by checking the `scannerType` is `precinct`?

libs/ui/src/error_boundary.tsx
nbhatia823: good find!

nbhatia823 (approved):
Looks good to me! Look forward to testing in person

libs/backend/src/cast_vote_records/export.ts
arsalansufi: Yup exactly! VxCentralScan doesn't use continuous export, so the notion of a recovery export doesn't make sense on it

PR #5472, opened 2024-10-03
Caro/add ballot style group id election def

carolinemodic:
## Overview
Based on https://github.com/votingworks/vxsuite/pull/5467 continues by adding BallotStyleGroupId as an explicit field in the election definition rather then parsing it out of the string of the BallotStyleId

## Checklist


libs/utils/src/ballot_styles.ts
```diff
@@ -141,11 +130,36 @@ export function getDefaultLanguageBallotStyles(
  *
  *
  */
-export function getParentBallotStyles(
+
+export function getGroupedBallotStyles(
```
jonahkagan, line 134: Nit: I think `getBallotStyleGroups` would be a slightly clearer name.

Also need to update comment for this function.

libs/utils/src/ballot_styles.ts
```diff
@@ -141,11 +130,36 @@ export function getDefaultLanguageBallotStyles(
...
+          englishBallotStyle = ballotStyle;
+        } else if (!ballotStyle.languages) {
+          legacyBallotStyle = ballotStyle;
+        }
+      }
+
+      const defaultBallotStyleInfo = assertDefined(
+        englishBallotStyle || legacyBallotStyle,
+        'Expected at least one English language ballot style per ballot style group.'
+      );
```
jonahkagan, line 157: All ballot styles in a group should only differ in the `languages` field - they should otherwise be identical. So it may just suffice in this function to pick a random ballot style from the group.

libs/types/src/election.ts
```diff
@@ -324,7 +324,7 @@ export interface BallotStyle {
 export type BallotStyleGroupId = NewType<string, 'BallotStyleGroupId'>;
 export const BallotStyleGroupIdSchema =
   IdSchema as unknown as z.ZodSchema<BallotStyleGroupId>;
-export interface ParentBallotStyle {
+export interface BallotStyleGroup {
```
jonahkagan, line 327: Remove `languages` field from `BallotStyleGroup`? I'm not sure what it would mean so safer to remove.

libs/utils/src/tabulation/lookups.ts
```diff
@@ -93,8 +93,8 @@ export const getBallotStyleById = createElectionMetadataLookupFunction(
 export const getParentBallotStyleById = createElectionMetadataLookupFunction(
```
jonahkagan, line 93: maybe find and replace "parentBallotStyle" with "ballotStyleGroup"

jonahkagan (approved):
Looks awesome overall! I only thoroughly reviewed logic updates, not tests/snapshots. Approving but with that caveat

libs/types/test/regression/elections.test.ts
jonahkagan, line 1: As a follow up, we should remove the `maintainBackwardsCompatibility` function that these were testing. Doesn't have to be you, but just noting

PR #5473, opened 2024-10-04
Reorganize management of voter settings in Scan

nbhatia823:
## Overview

Follow up to https://github.com/votingworks/vxsuite/pull/5462. Thread conversation [here](https://github.com/votingworks/vxsuite/pull/5462#discussion_r1785414877)

Main issue:
Our current approach of `VoterSettingsManager` being a component that reacts to side effects is not the best frontend practice.

First, as Jonah points out in the thread linked above, we are having a user action (ex. closing polls) that could directly trigger a state change (ex. resetting settings), instead trigger a side effect. But, right now `<VoterSettingsManager>` manages state that is not exposed to the app and actually is created all the way in `<App>`.

Second, the idea of a component that has no UI but only manages state indicates that it may be a better fit as a custom hook, as hooks can encapsulate state.

So, this PR suggests converting the component into a hook, which is created in `App_root` and provides the needed state management function to `PollWorkerScreen` to reset the settings.

Note: One upside about out current approach is that we do have all resetting/restoring of settings due to log ins/outs in the same place, so it is easy to reason about. With this change, we are following best practices for directly triggering state updates rather than relying on side effects, but we are also no longer co-locating these updates. 

## Demo Video or Screenshot

Behavior should not change from this PR. The following video shows: 
1. we maintain voter settings after returning from an auth session
2. we reset voter settings after polls close

https://github.com/user-attachments/assets/588f5127-49d9-4555-b33c-887b8d4c2e3b

## Testing Plan

_Will add tests if we decide this is an approach we want to take._

## Checklist


apps/scan/frontend/src/utils/use_voter_settings_controls.ts
```diff
@@ -0,0 +1,85 @@
...
+    select: ({ state }) => state,
+    onChange: (newState, previousState) => {
+      // If we transition from paused to no_paper we are just returning from an election official screen
+      if (
+        previousState &&
+        previousState !== 'no_paper' &&
+        previousState !== 'paused' &&
+        newState === 'no_paper'
+      ) {
+        resetVoterSettings();
```
nbhatia823, line 79: This line is the only change of the listeners from line 42-82 from `<VoterSettingsManager>`. Previously `setVoterSessionTheme(null)` and `setVoterSessionLanguage(null)` were not called, though I think the behavior remains the same if I understand these comments and the code properly 

That being said, it seems like this may also be logic we can move out of a listener and into the UI

apps/scan/frontend/src/app_root.tsx
```diff
@@ -40,12 +40,14 @@ import { SystemAdministratorScreen } from './screens/system_administrator_screen
...
 import { PrinterCoverOpenScreen } from './screens/printer_cover_open_screen';
 import { ScannerDoubleFeedCalibrationScreen } from './screens/scanner_double_feed_calibration_screen';
+import { useVoterSettingsControls } from './utils/use_voter_settings_controls';
 
 export function AppRoot(): JSX.Element | null {
   const [
     shouldStayOnCastVoteRecordSyncRequiredScreen,
     setShouldStayOnCastVoteRecordSyncRequiredScreen,
   ] = useState(false);
+  const voterSettingsControls = useVoterSettingsControls();
```
nbhatia823, line 50: probably rename this to something like `useScanVoterSettings`, because im wondering if "controls" makes it seem optional, but it is actually managing required state regardless if the controls are used

apps/scan/frontend/src/utils/use_voter_settings_controls.ts
```diff
@@ -0,0 +1,85 @@
...
+import { DefaultTheme, ThemeContext } from 'styled-components';
+import { LanguageCode } from '@votingworks/types';
+import { getAuthStatus, getScannerStatus } from '../api';
+
+interface VoterSettingsControls {
+  resetVoterSettings: () => void;
+}
+
+/**
+ * useVoterSettingsControls adds Scan-specific settings to the base settings provided by VoterSettingsManagerContext
```
nbhatia823, line 17: really it aggregates three different contexts, will change comment

apps/scan/frontend/src/utils/use_voter_settings_controls.ts
jonahkagan: I agree it makes sense to clear the saved voter settings when a voter is done.

I don't think we can change this from being a listener, since the voter doesn't interact with the screen, they just interact with the scanner hardware (via inserting a ballot).

apps/scan/frontend/src/utils/use_voter_settings_controls.ts
```diff
@@ -0,0 +1,85 @@
...
+    },
+  });
+
+  // [VVSG 2.0 7.1-A] Reset to default theme and language when voter is done scanning. We
+  // have chosen to interpret that as whenever paper leaves the scanner (either
+  // into the ballot box, or retrieved by the user after a ballot rejection).
+  useQueryChangeListener(scannerStatusQuery, {
+    select: ({ state }) => state,
+    onChange: (newState, previousState) => {
+      // If we transition from paused to no_paper we are just returning from an election official screen
```
jonahkagan: Since the `paused` state is triggered whenever an election official logs in and ends whenever they log out, I think we can unite these two listeners to just use the `paused` state instead of also checking auth. A slightly riskier change, so worth looking into whether there is good test coverage if we decide to pursue it.

apps/scan/frontend/src/utils/use_voter_settings_controls.ts
jonahkagan: I would like to see both comments changed to describe what this piece of code actually does, which IMO is:
- saving and reapplying a voter's theme and language settings when their voting session is interrupted by an election official
- resetting the settings after a voter finishes

apps/scan/frontend/src/app_root.tsx
```diff
@@ -228,6 +230,7 @@ export function AppRoot(): JSX.Element | null {
     return (
       <PollWorkerScreen
         electionDefinition={electionDefinition}
+        onPollsClose={() => voterSettingsControls.resetVoterSettings()}
```
jonahkagan, line 233: I think it would be clearer if `PollWorkerScreen` used the `useVoterSettingsControl` hook directly rather than passing in a generic `onPollsClose` handler, since we don't really need a generic handler.

That being said, the way the hook is currently written, that will also create a side effect of registering the change listeners again. I'm not sure if that would cause problems or not, but I'm a bit concerned about double side effects. I think to protect against this we should separate out the creation of the API to control the voter settings (keeping this in the hook) from the change listeners that uses the API (putting these directly in `AppRoot`).

apps/scan/frontend/src/utils/use_voter_settings_controls.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  // Queries for listeners
+  const authStatusQuery = getAuthStatus.useQuery();
+  const scannerStatusQuery = getScannerStatus.useQuery();
+
+  // Voter session specific settings, saved to return to after auth-ed sessions
+  const [voterSessionTheme, setVoterSessionTheme] =
+    React.useState<DefaultTheme | null>(null);
+  const [voterSessionLanguage, setVoterSessionLanguage] =
+    React.useState<LanguageCode | null>(null);
```
jonahkagan, line 33: Maybe rename to `savedVoterSessionTheme` and `savedVoterSessionLanguage` for clarity

jonahkagan (commented):
Thanks for exploring this refactor! I like the direction

apps/scan/frontend/src/app_root.tsx
nbhatia823: Great point, I agree with that. Moving the listeners out of the hook into `AppRoot` feels right. I also didn't like that the hook wasn't reusable across components without re-registering listeners, so this addresses that, as it felt like this hook was owning too much functionalit. Will push a commit soon with these changes

apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -348,7 +348,7 @@ function PollWorkerScreenContents({
       isAfterPollsTransition: true,
       printResult,
     });
-    onPollsClose();
+    voterSettingsControls.resetVoterSettings();
```
nbhatia823, line 351: This approach won't actually work right now, because `voterSettingsControls.resetVoterSettings()` will not reset the voter session state that is managed by the invocation of `useVoterSettingsControl` from `app_root`

apps/scan/frontend/src/utils/use_session_settings_manager.ts
```diff
@@ -0,0 +1,63 @@
...
+  useCurrentLanguage,
+  useLanguageControls,
+} from '@votingworks/ui';
+import { DefaultTheme, ThemeContext } from 'styled-components';
+import { LanguageCode } from '@votingworks/types';
+
+export interface SessionSettingsManagerProps {
+  resetVoterSettings: () => void;
+  cacheAndResetVoterSettings: () => void;
+  restoreVoterSessionsSettings: () => void;
```
jonahkagan, line 13: Another way to think about these that might be clearer:

```ts
startNewSession
pauseSession
resumeSession
```

apps/scan/frontend/src/app.test.tsx
jonahkagan, line 1: I'm wondering if there's a way to test the user-facing behavior we want rather than relying on mocking the hook. E.g. start a voter session, change the theme/language, insert a poll worker card, check that the theme/language are back to default, remove the card, check that the theme/language are back to the voter's choice.

If we could do that, we could even get rid of the tests for the hook. My general philosophy is to mock as little as possible to ensure that all the pieces are working together and to make the tests resilient to changes in implementation. I'll still do small unit tests if there are intricate behaviors in a function or subcomponent that will be easier to test there, but for main behavior I like to test at the highest level that covers the entirety of the behavior. Here's [an article](https://kentcdodds.com/blog/write-tests) that I think lays out a compelling argument for this approach (except the point about coverage).

I'm also ok with what you've got, so gonna approve the PR, in case you'd rather move forward than rework these.

jonahkagan (approved):

apps/scan/frontend/src/app.test.tsx
nbhatia823: Thanks for the thoughtful point. I like that concept, though I feel like I need to write the tests and see what they  look like to feel confident about it. My concern is that it might make tests too complicated, but I am very open to it because I do see how it could actually make it simpler to test the core flows that matter.

That being said, I think I will merge this PR and then work on subsequent PR that reworks these tests. Thinking that so 1) we can QA these changes on Scan because writing these tests will still take me some time (I was running into some roadblocks earlier when I tried to use less mocks), and I'd love to do it properly without being rushed so i can learn along the way and 2) since it will only be updating tests and not behavior, it seems okay to have the test rework PR not make the certification cut.

Let me know if something in that logic seems off to you, otherwise I'll move forward with it

apps/scan/frontend/src/utils/use_session_settings_manager.ts
nbhatia823: Agreed

apps/scan/frontend/src/app.test.tsx
jonahkagan: sounds good to me!

PR #5475, opened 2024-10-05
refactor(central-scan): move "next sheet" endpoint to grout

eventualbuddha:
## Overview

Moves our last VxCentralScan JSON API to use Grout.

## Demo Video or Screenshot
n/a

## Testing Plan
TODO


apps/central-scan/frontend/src/screens/ballot_eject_screen.tsx
```diff
@@ -90,6 +90,8 @@ export function BallotEjectScreen({ isTestMode }: Props): JSX.Element | null {
 
   const systemSettingsQuery = getSystemSettings.useQuery();
   const continueScanningMutation = continueScanning.useMutation();
+  const getNextSheetToReviewQuery = getNextSheetToReview.useQuery();
```
jonahkagan, line 93: I don't see anything that invalidates this query, so I'm a little concerned it would reuse cached data. Have you tested that?

eventualbuddha: Should be fixed now.

jonahkagan: I think `staleTime: 0` will help, but I think it's still possible to have a momentary flash of the cached data while the new data is being fetched (since the refetch won't be initiated until the component mounts). Are you seeing that? I think checking `isFetching` could help catch this case. Or maybe setting `cacheTime: 0` as well

PR #5483, opened 2024-10-07
utils: Add `getBallotStyleGroup` helper

jonahkagan:
## Overview

For cases where we have a group ID and want to look it up. I realized I needed this when I was rebasing another PR after the recent refactor to ballot styles.
## Demo Video or Screenshot
N/A

## Testing Plan
Added new tests

## Checklist

### DIFF
--- a/libs/utils/src/ballot_styles.ts
+++ b/libs/utils/src/ballot_styles.ts
@@ -13,6 +13,7 @@ import {
   LanguageCode,
   BallotStyleGroup,
   Party,
+  Election,
 } from '@votingworks/types';
 
 const ID_LANGUAGES_SEPARATOR = '_';
@@ -36,6 +37,7 @@ export function generateBallotStyleId(params: {
   languages: LanguageCode[];
   party?: Party;
 }): BallotStyleId {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
   return [generateBallotStyleGroupId(params), ...params.languages].join(
     ID_LANGUAGES_SEPARATOR
   ) as BallotStyleId;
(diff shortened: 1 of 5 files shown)


libs/utils/src/ballot_styles.ts
```diff
@@ -36,6 +37,7 @@ export function generateBallotStyleId(params: {
   languages: LanguageCode[];
   party?: Party;
 }): BallotStyleId {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
carolinemodic, line 40: did you need to add this for it to clear lint? You didn't change the next line so checking. I was having issues with this lint rule on my PR (I dont remember if it was here specifically)

libs/utils/src/tabulation/lookups.ts
```diff
@@ -149,12 +149,10 @@ export function determinePartyId<T>(
...
-    electionDefinition.election.ballotStyles
-  ).find((bs) => bs.id === group.ballotStyleGroupId);
-  if (!ballotStyle) return undefined;
-
-  return ballotStyle.partyId;
+  const ballotStyleGroup = getBallotStyleGroup({
+    election: electionDefinition.election,
+    ballotStyleGroupId: group.ballotStyleGroupId as BallotStyleGroupId,
+  });
+  return ballotStyleGroup?.partyId;
```
carolinemodic, line 157: I thought it was against the lint rules to use the non-null ? assertion. Did I just totally make that up?

carolinemodic (approved):

libs/utils/src/ballot_styles.ts
jonahkagan: Yeah I couldn't pass lint without this comment

libs/utils/src/tabulation/lookups.ts
jonahkagan: We can't use `!` but we can use `?`

PR #5485, opened 2024-10-07
Update dev dock to have full list of fixtures elections

carolinemodic:
## Overview
The dev dock today just hard codes a few fixtures elections, I find this frustrating when I'm working with other elections so I am proposing we just generate this list from the fixtures/data directory and look for any subfolders with an election*.json file inside. 

Next step would be to add a way to autopopulate the mock usb drive with the election package for the current selected election (maybe with different system settings options) 

## Demo Video or Screenshot
Screenshot 2024-10-07 at 3 51 26 PM

## Testing Plan
Used dev dock

### DIFF
--- a/libs/dev-dock/backend/src/dev_dock_api.ts
+++ b/libs/dev-dock/backend/src/dev_dock_api.ts
@@ -109,6 +109,29 @@ function buildApi(devDockFilePath: string, machineType: MachineType) {
       return readDevDockFileContents(devDockFilePath).electionInfo;
     },
 
+    getCurrentFixtureElectionPaths(): DevDockElectionInfo[] {
+      const baseFixturePath = join(__dirname, '../../../../libs/fixtures/data');
+      return fs
+        .readdirSync(baseFixturePath, {
+          withFileTypes: true,
+        })
+        .filter((item) => item.isDirectory())
+        .map((item) => {
+          const filesInDir = fs.readdirSync(join(baseFixturePath, item.name));
+          const electionFile = filesInDir.find((file) =>
+            /^election\.json$/.test(file)
+          );
+          if (electionFile) {
+            return {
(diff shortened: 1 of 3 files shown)


libs/dev-dock/backend/src/dev_dock_api.ts
```diff
@@ -109,6 +109,29 @@ function buildApi(devDockFilePath: string, machineType: MachineType) {
...
+        })
+        .filter((item) => item.isDirectory())
+        .map((item) => {
+          const filesInDir = fs.readdirSync(join(baseFixturePath, item.name));
+          const electionFile = filesInDir.find((file) =>
+            /^election\.json$/.test(file)
+          );
+          if (electionFile) {
+            return {
+              path: join(join(baseFixturePath, item.name), 'election.json'),
```
jonahkagan, line 126: I think `join` can take as many arguments as you want to join, so this could be simplified to a single `join` call

libs/dev-dock/frontend/src/dev_dock.tsx
```diff
@@ -63,33 +63,18 @@ const ElectionControlSelect = styled.select`
...
   const apiClient = useApiClient();
   const getElectionQuery = useQuery(
     ['getElection'],
     async () => (await apiClient.getElection()) ?? null
   );
+  const currentFixturesQuery = useQuery(
+    ['getCurrentFixtureElectionPaths'],
+    async () => (await apiClient.getCurrentFixtureElectionPaths()) ?? []
+  );
+  const fixturesElections = currentFixturesQuery.data || [];
```
jonahkagan, line 77: Do we need two different fallbacks to `[]`?

jonahkagan (approved):
Good call!

I like the idea of putting an election package on the USB as well, though we'd need to think through how to incorporate the other elements of the package (translations, system settings, etc)

PR #5487, opened 2024-10-08
Kevin/bmd adjudication flags

kshen0:
## Overview

* Evaluates `Undervote` and `BlankBallot` adjudication reasons for BMD ballots and reports them on `AdjudicationReasonInfo`
* Handles `AdjudicationReasonInfos` in `scan/backend` for BMD ballots so the frontend can call for user action

## Demo Video or Screenshot

Currently no hardware setup for me to demo this

## Testing Plan

Added tests at library and app level

## Checklist

### DIFF
--- a/libs/ballot-interpreter/src/adjudication_reasons.ts
+++ b/libs/ballot-interpreter/src/adjudication_reasons.ts
@@ -79,11 +79,16 @@ export function getAllPossibleAdjudicationReasonsForBmdVotes(
           break;
         case 'yesno':
           for (const option of actualVotes as YesNoVote) {
+            // Should never be executed because a yes/no contest should have at
+            // most 1 vote so an undervote would be 0 votes. But the type
+            // doesn't enforce this, so we should handle the case where the
+            // contest has > 1 vote.
+            /* istanbul ignore next */
             optionIds.push(option);
           }
           break;
+        /* istanbul ignore next */
         default:
-          /* istanbul ignore next */
           throwIllegalValue(contestType);
       }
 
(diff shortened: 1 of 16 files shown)


libs/ballot-interpreter/src/adjudication_reasons.ts
```diff
@@ -31,8 +35,83 @@ function compareMarkStatusDescending(
...
+          for (const option of actualVotes as CandidateVote) {
+            optionIds.push(option.id);
+          }
+          break;
+        case 'yesno':
+          for (const option of actualVotes as YesNoVote) {
+            // Should never be executed because a yes/no contest should have at
+            // most 1 vote so an undervote would be 0 votes. But the type
+            // doesn't enforce this, so we should handle the case where the
+            // contest has > 1 vote.
```
arsalansufi, line 85: Got it, I think I follow. The comment's wording took me a second to parse, but maybe something like this would be easier to follow:

> Because you can't overvote on a BMD, and the only way to undervote on a yes-no contest is to leave it blank, we'll never have any option IDs to populate here

arsalansufi (commented):
The logic all looks sound and consistent with HMPB adjudication! Will wait till this is out of draft state and tests are all passing (I see you're close!) to fully approve

arsalansufi (approved):
Other than my comment above re comment wording, LGTM!

libs/ballot-interpreter/src/adjudication_reasons.ts
kshen0: Clarified in [37fc820](https://github.com/votingworks/vxsuite/pull/5487/commits/37fc82086abf81e3d8d36c0afb36934234058b36) and removed no-op code

arsalansufi: Ah nice, good call

kshen0: At first I left the no-op code in because the type technically allowed for it, but on second thought I didn't like leaving in code that we never expect to execute and that has undefined business logic.

Issue #3826, opened 2023-08-11
Audit vxsuite & kiosk-browser dependecies and upgrade/prune where possible 


carolinemodic:
Audit all of our packages installed by kiosk-browser/vxsuite (i.e. anything not covered by: https://github.com/votingworks/vxsuite/issues/3825 ) and determine what we should upgrade before cert and then execute on those upgrades.

eventualbuddha:
How are we planning to decide what should be upgraded? Should we do a security audit a la `npm audit`?

arsalansufi:
Discussed at planning and like the idea of running a `pnpm audit fix` to guide our updates here! I don't think we need to address everything surfaced but should probably address the critical ones.

Many of the finds may not apply to us in our offline, locked-down context. And any update will come with some risk re invalidating prior QA. But we can flag specific updates and discuss as needed

PR #5489, opened 2024-10-08
chore: remove various create-react-app remnants

eventualbuddha:
We used to use [Create React App](https://create-react-app.dev/) to manage our VxSuite React apps. Some time ago, we [ejected](https://create-react-app.dev/docs/available-scripts/#npm-run-eject), which essentially "inlined" CRA into the apps where we used it. We cleared out most of the stuff we didn't need, but some of the files–such as `test.js`, `paths.js`, etc.–that were added as part of the eject process have lingered. This finally removes those extra files, plus a dependency we do not use (`react-dev-utils`).


eventualbuddha:
It turns out there were vulnerabilities in transitive dependencies of `react-dev-utils`, so this is relevant to my current task: https://github.com/votingworks/vxsuite/issues/3826.

jonahkagan (approved):
✂️

PR #5491, opened 2024-10-08
mark-scan: go to unrecoverable error for interpretation errors

kshen0:
## Overview

Sends the state machine to a final `unrecoverable_error` state when we encounter an unexpected error with scanning or interpretation. This can happen when
* interpreting when first loading a ballot sheet. We use the interpretation result to decide whether a blank sheet or previously-printed sheet has been inserted
* interpreting immediately after a ballot has been printed in the course of normal voting

## Demo Video or Screenshot

Before

https://github.com/user-attachments/assets/4d6588c6-a484-4c16-9222-af63762c3f8d

After

https://github.com/user-attachments/assets/76dddc10-95f3-4efa-870f-93da705ffb44

## Testing Plan

* [x] Manual testing on VM
* [x] Updated and added automated tests

## Checklist


libs/ui/src/ui_strings/app_strings.tsx
```diff
@@ -1378,6 +1378,18 @@ export const appStrings = {
     </UiString>
   ),
 
+  unrecoverableError: () => (
+    <UiString uiStringKey="unrecoverableError">
+      An unrecoverable error occurred.
```
jonahkagan, line 1383: Since this is voter facing, maybe use plainer language. Could even just be "Something went wrong" since we already use that

jonahkagan (approved):

PR #5493, opened 2024-10-08
Add ability to use bin/interpret with path to cvr file directory

carolinemodic:
## Overview
Adds the ability to run the bin/interpret command line tool with the options: <election-path> <system-settings-path> <cvr-directory-path> where <cvr-directory-path> is assumed to be a folder that we save in the "cast-vote-records" folder on usb drives in production. These folders each contain a subfolder with front and back images that the script will pick out and run against the interpret tool with the other specified arguements. Should make it easier to quickly get reports after generating cvrs in prod. 

## Testing Plan
Ran script against a recent cvr export


eventualbuddha (approved):
Added a test in https://github.com/votingworks/vxsuite/pull/5498.

PR #5494, opened 2024-10-08
VxAdmin: Clear query cache on unconfigure

jonahkagan:
## Overview

I noticed that VxAdmin would crash after the following sequence:
- Configure with election A
- Create manual tallies
- Unconfigure
- Configure with election B
- Go to the manual tallies screen

The error was happening because we were [looking up a precinct ID](https://github.com/votingworks/vxsuite/blob/c517faaf2666328be9ac157999db0a57fc2a7f30/apps/admin/frontend/src/screens/tally/manual_tallies_tab.tsx#L414-L417) from the manual tallies in the election, but not finding it. This made me think that there was still cached manual tally data from the previous election that was being rendered while the new election's manual tally data was being fetched.

To prevent this from happening, I changed the logic to clear the cache after unconfiguring instead of just invalidating the queries.

## Demo Video or Screenshot
Skipping because it's not very interesting

## Testing Plan
Manual test, added regression test

## Checklist

### DIFF
--- a/apps/admin/frontend/src/app.test.tsx
+++ b/apps/admin/frontend/src/app.test.tsx
@@ -23,7 +23,10 @@ import { eitherNeitherElectionDefinition } from '../test/render_in_app_context';
 import { buildApp } from '../test/helpers/build_app';
 import { ApiMock, createApiMock } from '../test/helpers/mock_api_client';
 
-import { mockCastVoteRecordFileRecord } from '../test/api_mock_data';
+import {
+  mockCastVoteRecordFileRecord,
+  mockManualResultsMetadata,
+} from '../test/api_mock_data';
 import { MARK_RESULTS_OFFICIAL_BUTTON_TEXT } from './components/mark_official_button';
 
 let apiMock: ApiMock;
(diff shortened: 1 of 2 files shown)


apps/admin/frontend/src/app.test.tsx
```diff
@@ -272,6 +281,34 @@ test('removing election resets cvr and manual data files', async () => {
...
+  apiMock.expectConfigure(electionPackage.path);
+  apiMock.expectGetCurrentElectionMetadata({ electionDefinition });
+  userEvent.click(screen.getByText(electionPackage.name));
+  await screen.findAllByText(electionDefinition.election.title);
+
+  await apiMock.logOut();
+  await apiMock.authenticateAsElectionManager(electionDefinition);
+
+  // Manual tallies reset
+  // Note that this test of manually tally data specifically is a regression
```
kshen0, line 296: nit: "manual tally data"

kshen0 (approved):

apps/admin/frontend/src/app.test.tsx
jonahkagan: whoops, thanks! fixed

Issue #5468, opened 2024-10-03
CDF ERR import should dedupe write-in candidate IDs


kshen0:
Attempting to import two ERR files that specify the exact same write in candidate for the same contest will result in a uniqueness constraint violation for the `write_in_candidates` table:

`unique (election_id, contest_id, name)`

Converted ERR data is eventually processed through `handleEnteredWriteInCandidateData`. It seems like `handleEnteredWriteInCandidateData` assumes any write in candidates in the data need a `write_in_candidates` row created for them. I haven't fully investigated the solution but I'm guessing we should pre-process the converted ERR data:

1. Look for an existing write-in candidate for this `{election, contest, name}` combination
2. If it exists, modify the tally so the duplicate write-in candidate fails [this check](https://github.com/votingworks/vxsuite/blob/b64b4dbe826429045286e0ef82666fc016daf8d5/apps/admin/backend/src/util/manual_results.ts#L31)
3. Ensure tallies for the candidate are correctly summed

PR #5510, opened 2024-10-12
allow import of candidate with same name

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/5468

Fixes a bug where CDF ERR import would hard fail if a duplicate write-in candidate was specified in its tallies. That is, if a candidate for the given `election`, `contest`, and `name` already exists in the `write_in_candidates` table (as it would when a popular 3rd party option is written in), the app would crash. 

## Demo Video or Screenshot

Before:

https://github.com/user-attachments/assets/9866a9b5-efa1-4de8-87b4-c2dc18f80424

After:

https://github.com/user-attachments/assets/b2770cfa-6ba2-4629-8e4a-7c9a051307df

## Testing Plan
- manual testing
- added test for double import

## Checklist

- [ x I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

### DIFF
--- a/apps/admin/backend/src/util/manual_results.ts
+++ b/apps/admin/backend/src/util/manual_results.ts
@@ -46,12 +46,33 @@ function handleEnteredWriteInCandidateData({
           } else if (
             candidateId.startsWith(AdminTypes.TEMPORARY_WRITE_IN_ID_PREFIX)
           ) {
-            // for temp-write-in candidates, create records and substitute ids
-            const writeInCandidateRecord = store.addWriteInCandidate({
+            const { contestId } = contestResults;
+            // Check for existing write-in candidate before creating a new one.
+            // The main reason for this check is for CDF ERR file import. All
+            // write-ins in an ERR file will be tagged with TEMPORARY_WRITE_IN_ID_PREFIX
+            // even if they conflict with existing write-in candiate rows.
+            //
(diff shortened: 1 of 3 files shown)


apps/admin/backend/src/util/manual_results.ts
```diff
@@ -46,12 +46,33 @@ function handleEnteredWriteInCandidateData({
...
+            // write-ins in an ERR file will be tagged with TEMPORARY_WRITE_IN_ID_PREFIX
+            // even if they conflict with existing write-in candiate rows.
+            //
+            // The manual results frontend only tags with TEMPORARY_WRITE_IN_ID_PREFIX
+            // if it determines no conflicting write-in candidate exists. Therefore,
+            // this check should no-op for manual results entered through the VxAdmin UI.
+            const existingWriteInCandidateRecords = store.getWriteInCandidates({
               electionId,
-              contestId: contestResults.contestId,
-              name: candidateTally.name,
```
adghayes, line 53: Before, it was failing due to the uniqueness requirement in the DB?

adghayes (approved):

apps/admin/backend/src/util/manual_results.ts
kshen0: Correct, it was failing on the combination of `{election, contest, name}`

PR #5518, opened 2024-10-14
Remove Reboot To BIOS options

adghayes:
## Overview

Removed Reboot to BIOS as this is now a vendor function.

## Demo Video or Screenshot

Screen Shot 2024-10-14 at 12 32 28 PM
Screen Shot 2024-10-14 at 12 45 04 PM
Screen Shot 2024-10-14 at 12 39 12 PM
Screen Shot 2024-10-14 at 12 33 02 PM

## Testing Plan

Checked the screens (screenshotted above).


kshen0 (commented):
Can you explain more what it means for this to now be a vendor function?

adghayes:
This is now behind the vendor menu (a.k.a tty2) as of https://github.com/votingworks/vxsuite-complete-system/pull/400, so it doesn't need to be available to any customer roles. They shouldn't ever need to reboot into BIOS, and the option can just be confusing.

kshen0 (approved):
Thanks for the context!

Issue #5515, opened 2024-10-14
On the maximum election we don’t show the ballot measure text in extra large font mode 


arsalansufi:
On the maximum election we don’t show the ballot measure text in extra large font mode 

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1728594021050789?thread_ts=1728594021.050789&cid=CJU9MSC6S

The middle section of the UI that displays the prop is essentially. It probably just needs a min-height.

PR #5521, opened 2024-10-15
fix disappearing measure description

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/5515

Fixes a bug where ballot measure description is not shown when using XL display settings.

This change always shows the contest options and allows scrolling of the measure title and description (rather than include contest options in the scroll box). I thought this was preferable because it would cut down on repetitive scrolling for voters who know how they want to vote based on contest title alone.

## Demo Video or Screenshot

**Medium**
Screenshot 2024-10-15 at 2 58 09 PM

**Large**
Screenshot 2024-10-15 at 3 45 45 PM

**XL**
Screenshot 2024-10-15 at 3 45 52 PM

**Video of scrolling behavior**

https://github.com/user-attachments/assets/11d10f2c-d81b-46b0-896f-0e88c9f895ae

## Testing Plan
* Manual testing

## Checklist

### DIFF
--- a/libs/mark-flow-ui/src/components/contest_header.tsx
+++ b/libs/mark-flow-ui/src/components/contest_header.tsx
@@ -13,16 +13,12 @@ import {
 import { Contest, District } from '@votingworks/types';
 import { MsEitherNeitherContest } from '../utils/ms_either_neither_contests';
 
-interface ContainerProps {
-  horizontalPadding?: boolean;
-}
-
 export interface ContestHeaderProps {
   breadcrumbs?: BreadcrumbMetadata;
   children?: React.ReactNode;
   contest: Contest | MsEitherNeitherContest;
   district: District;
-  styleOverrides?: ContainerProps;
+  className?: string;
 }
 
 export interface BreadcrumbMetadata {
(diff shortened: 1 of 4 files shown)


libs/mark-flow-ui/src/components/yes_no_contest.test.tsx
```diff
@@ -50,9 +50,11 @@ test('voting for both yes and no', () => {
...
   screen.getByRole('heading', { name: contest.title });
 
-  within(screen.getByTestId(MOCK_WITH_SCROLL_BUTTONS_TEST_ID)).getByText(
-    contest.description
-  );
+  const descriptions = within(
+    screen.getByTestId(MOCK_WITH_SCROLL_BUTTONS_TEST_ID)
+  ).getAllByText(contest.description);
+  // Expect once for AudioOnly component and once for visual component
+  expect(descriptions.length).toEqual(2);
```
kshen0, line 57: This test selects the `WithScrollButtons` component and finds the contest description within it. The description is rendered twice as the comment describes. Previously the `WithScrollButtons` component only wrapped one occurrence, but now it wraps both, so the test needed to be changed.

libs/mark-flow-ui/src/components/contest_header.tsx
```diff
@@ -48,10 +53,16 @@ export function Breadcrumbs(props: BreadcrumbMetadata): React.ReactNode {
...
 
 export function ContestHeader(props: ContestHeaderProps): JSX.Element {
-  const { breadcrumbs, children, contest, district } = props;
+  const {
+    breadcrumbs,
+    children,
+    contest,
+    district,
+    styleOverrides = { horizontalPadding: true },
+  } = props;
```
jonahkagan, line 62: If you pass in a `className` prop and pass it to `Container`, you can use styled-components to add custom CSS styling and override the padding. I suggest since custom styling props can become a bit convoluted.

Alternatively I'd be into just removing the horizontal padding from this component and adding padding around it as needed at various render sites.

jonahkagan (approved):
Nice, seems like a good compromise

libs/mark-flow-ui/src/components/contest_header.tsx
kshen0: Thanks! Ended up overriding with `className` to favor a smaller diff.

PR #5535, opened 2024-10-16
allow system admin to access system even in hardware error states

### LINKED ISSUE #5523: The system admin screen is inaccessible in VxMark when the “Internal Connection Problem” screen is showing
The system admin screen is inaccessible in VxMark when the “Internal Connection Problem” screen is showing. I think we could use some reworking of screen precedence like @caroline did for VxScan.

https://votingworks.slack.com/archives/CJU9MSC6S/p1728934500514529

adghayes:
## Overview

Closes #5523. Allows the vendor and system admin screens to be accessed in hardware error states.

### DIFF
--- a/apps/mark-scan/frontend/src/app_root.tsx
+++ b/apps/mark-scan/frontend/src/app_root.tsx
@@ -351,23 +351,11 @@ export function AppRoot(): JSX.Element | null {
   const machineConfig = machineConfigQuery.data;
   const usbDriveStatus = usbDriveStatusQuery.data;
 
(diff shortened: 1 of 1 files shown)


adghayes:
@kshen0 can you think through whether this is a save change? We want to avoid situations where we can't pull off logs on the devices, which would currently happen if one of the moved error states were to happen.

kshen0 (approved):
This should be fine. I was concerned the paper handler diagnostic assumes the paper handler is connected but we actually handle the `no_hardware` state in the API.

PR #5536, opened 2024-10-17
show no usb drive detected screen even when continuous export disabled

### LINKED ISSUE #5526: Accept ballots when continuous export is disabled and no USB drive is connected
In v4, if you disable continuous export and remove the USB drive, we properly don't display the "No USB drive detected" screen. But we also don't accept any ballots until a USB drive is reconnected. This works as expected in v3 and you can scan without a USB drive connected when you pause continuous export. I missed a case in this [new isReadyToScan function](https://github.com/votingworks/vxsuite/blob/326edf3e3d1c7776d2ef571a6811d8042a92c2eb/apps/scan/backend/src/app_flow.ts#L51-L54) in my up-port from v3 to v4.

Issue automatically created from Slack feedback reported by arsalan: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1728610432894339?thread_ts=1728610432.894339&cid=CJU9MSC6S

adghayes:
## Overview

Closes #5526. Shows the "No USB Drive Detected" message on VxScan even when continuous export is disabled.

### DIFF
--- a/apps/scan/frontend/src/app_root.tsx
+++ b/apps/scan/frontend/src/app_root.tsx
@@ -112,7 +112,6 @@ export function AppRoot(): JSX.Element | null {
     isTestMode,
     precinctSelection,
     isSoundMuted,
-    isContinuousExportEnabled,
   } = configQuery.data;
   const scannerStatus = scannerStatusQuery.data;
   const usbDrive = usbDriveStatusQuery.data;
@@ -267,11 +266,7 @@ export function AppRoot(): JSX.Element | null {
     );
   }
 
-  if (
-    isContinuousExportEnabled &&
-    usbDrive.status !== 'mounted' &&
-    pollsState !== 'polls_closed_final'
-  ) {
+  if (usbDrive.status !== 'mounted' && pollsState !== 'polls_closed_final') {
     return <InsertUsbScreen />;
   }
 


adghayes:
@arsalansufi I know we haven't had a chance to discuss this, so feel free to wait until we do.

arsalansufi:
Thanks for the ping @adghayes! I just shared my thinking on this with @mattroe, and it's actually intentional in v3.1.2 that we do not surface the "No USB drive detected" screen when continuous export is disabled. The thinking there is as follows:

- The point of this feature is to remove the need for a working USB drive from the voter path.
- It's completely feasible that a USB drive that errors when written to also fails to be detected.
- If we keep the "No USB drive detected" screen, you'll still need to find a working USB drive to unblock voting, which could take time if you don't have a spare easily accessible.
- When continuous export is disabled, we also don't export CVRs on polls close. We'll instead have you use the "Save CVRs" button in the EM menu after polls close. That too is intentional. We don't want USB drive finickiness to get in the way of polls close, the last time-sensitive operation of the day. Given this, there's really no utility in requiring you to have a USB drive plugged in.

So all that said, I can own getting https://github.com/votingworks/vxsuite/issues/5526 over the finish line, switching back to the original intent. But happy to discuss further of course!

adghayes:
@arsalansufi thanks for explaining, we didn't know that saving CVRs is disabled on polls closed, I agree we should do #5526.

arsalansufi:
Perfect, I'll close this PR out and open my own!

Issue #5330, opened 2024-08-28
Make it possible to support additional languages without a new software image


arsalansufi:
This should be quick - just a matter of removing some hardcoded supported language lists in VxSuite code. Hardcoded supported language lists are still fine and necessary in VxDesign.

PR #5545, opened 2024-10-22
[Multi-Language] Move all hardcoded language assumptions out of machine images

kofi-q:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/5330

Lot of little changes, but most of them repetitive and grouped in commits WRT which find-and-replace operations were done:
- First 5 commits are pure find/replace via vscode, along with its built-in refactoring kicking in for dropping unused imports, so those are quickly scannable.
- Last 3 commits have actual manual code changes and some logic tweaks here and there, so I'd focus on those.

This removes all hardcoded assumptions about "valid" language codes from everything except VxDesign code, in order to enable adding/configuring supported languages without creating new images. We'll still need one more revision to support RTL languages and add the necessary config to flip, but this avoids a revision for each added language.

Moved the `LanguageCode` enum from `libs/types` to `apps/design`, where it's used exclusively now.
Ended up adding a `TestLanguageCode` to `libs/test-utils` though, to make some of the test code replacements easier, as well as a `DEFAULT_LANGUAGE_CODE` constant to `libs/types` for a few places where we do make valid assumptions about the `en` language code (might've missed some, but not critical).

## Testing Plan
- Existing tests as regression guards
- Manual test: VxDesign export -> VxAdmin -> VxMarkScan

## Checklist


libs/ui/test/test_ui_strings.tsx
```diff
@@ -55,6 +55,7 @@ export const testUiStrings = {
 // Using `testUiStrings.numPlanets()` directly in Storybook, doesn't work, for
 // some reason. This is a convenience wrapper to avoid the failure, which seems
 // related to rendering the count param object as a React component child.
+// istanbul ignore next
```
nbhatia823, line 58: For my understanding, what is this doing?

kofi-q: Ah eyah, [`istanbul`](https://istanbul.js.org/) is what we use for instrumenting our source code for test coverage reporting (it comes in as a dependency with `jest`). It reads these comment directives and skips code lines or blocks based on `ignore <type>` - in this particular case, this function is now only used in `libs/ui/src/ui_strings/ui_string.stories.tsx`, so it's not a tested path anymore

nbhatia823: Ah gotcha, thank you!

PR #5552, opened 2024-10-24
fix(ballot-interpreter): reject over-stretched ballots

eventualbuddha:
## Overview

Any stretching is likely to show up in the form of a over-large gap between timing marks in the direction of scanning (i.e. vertical). ~In addition to rejecting ballots with too much stretching, this PR also improves the algorithm for matching candidate timing marks. Instead of finding the next timing mark only in one direction, we now find them in both directions. This makes it better handle a single gap that isn't too big.~ **Edit:** I opted to revert the change to the timing mark inference algorithm because it was causing regressions, particularly in matching bottom timing marks. It wasn't the "two directions" change that was an issue. Instead, it was narrowing the allowed timing mark spacing error too much so that bottom timing marks were not properly found, especially in situations where the bottom left corner timing mark was partially cut off.

## Demo Video or Screenshot

image

Output after this change with the above stretched ballot:

```
Error interpreting ballot:
{
  "type": "missingTimingMarks",
  "rects": […],
  "reason": "Timing marks are too far apart, likely due to stretching: border=Left, index=6 distance=182.06866px"
}
```

## Testing Plan

I attempted to reproduce the original stretching by holding the paper mid-scan on the PDI scanner, but was never able to do so. Instead, I used the one example @nbhatia823 was able to generate, which the partial ballots above are from. I'm omitting the full images because they are from ballots for an upcoming election. I checked in two censored images as test cases for the stretch rejection.


jonahkagan:
I won't have time to fully review this until next week, but a quick question: How does this interact with our existing ability to infer timing marks (e.g. if one gets cut off or messed up for some reason)?

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -1538,17 +1545,22 @@ fn infer_missing_timing_marks_on_segment(
...
+    let segment_backward = segment_forward.reversed();
+
     let mut inferred_timing_marks = vec![];
-    let mut current_timing_mark_center = segment.start;
-    let next_point_vector = segment.with_length(expected_distance).vector();
+    let mut current_timing_mark_center_forward = segment_forward.start;
+    let mut current_timing_mark_center_backward = segment_backward.start;
+    let next_point_vector_forward = segment_forward.with_length(expected_distance).vector();
+    let next_point_vector_backward = segment_backward.with_length(expected_distance).vector();
     let maximum_error = expected_distance / 2.0;
```
jonahkagan, line 1597: How should `maximum_error` here relate to the max gap allowed in the resulting timing marks?

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -1298,6 +1316,29 @@ pub fn find_complete_from_partial(
         );
     }
 
+    // We only check for stretching along the vertical axis because that's the
+    // direction we scan ballots in. If we ever begin scanning horizontally, we
+    // should also check for stretching along the horizontal axis.
+    let max_allowed_vertical_distance = median_vertical_distance * 1.5;
```
jonahkagan, line 1322: It seems high risk to try to interpret when there's any stretching, and I think it happens pretty rarely, so I'd suggest shortening the max gap allowed

jonahkagan (approved):

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: Shrunk to 120%.

libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
eventualbuddha: Consolidated with the other change.

eventualbuddha:
Whoops, forgot to disable auto-merge. @jonahkagan, I wanted you to take a look because I reverted the change to the inferred timing mark detection algorithm and wanted your take.

jonahkagan:
What motivated that change?

jonahkagan:
Ah, I missed the updated description, thanks.

I'm a little confused still. If the tighter tolerances were the cause of the regressions, should we still be worried about those tighter tolerances increasing rejected ballots that we want to interpret? Or was it somehow the combination of the tighter tolerances with the bi-directional algorithm specifically? Would reversing the direction of the single-direction algorithm cause the same regressions?

eventualbuddha:
We don't enforce the timing mark spacing on the top/bottom, only the left/right. Changing the inference algorithm affected all four sides.

jonahkagan:
Ah, gotcha! That all makes sense!

PR #5554, opened 2024-10-25
Automated Screenshot Generation (VxAdmin)

adghayes:
## Overview

I'm experimenting with automatically generating screenshots. Instead of a painstaking process of generating screenshots manually, which are then invalidated with whatever subsequent copy or design changes, I want to have a way to regenerate all important screenshots for the app at once. At the same time, it serves as an integration test that clicks through most of our screens. The screenshots are saved to the `./test-results/screenshots` directory, which can be accessed on CircleCI or copied when generated locally.

There are two concerns with this approach IMO:
- Flakiness - I believe I've worked out the flaky kinks in the tests as they currently are, but it's _very_ possible for Playwright tests to be flaky if you're not careful
- Maintenance - now these tests will have to be updated when the flows change

If the tests become cumbersome, I think that they can just be skipped in the future as we've skipped integration tests for months at a time in the past, and revived at later dates. I see this as safe to try. Although, I hope that the burden is actually low and it provides an integrated way for engineers to create screenshots for a PR while also making that screenshot generation sustainable. In cases where a PR changes the appearance of the app without changing flows, I think it will be even more useful as a tool as it allows you to holistically review the impact of the changes on the app and iterate on it systematically, something that previously was not possible.


adghayes:
Attached is an archive of the produced screenshots in the tests' current form.

[admin-screenshots.zip](https://github.com/user-attachments/files/17535402/admin-screenshots.zip)

adghayes:
@jonahkagan @arsalansufi is one of you interested in being primary reviewer here?

arsalansufi:
I know @jonahkagan's pretty heads down on Arlo reviews right now, so I can take a look! But maybe one of @eventualbuddha or @kofi-q (who are still working on VxSuite loose ends) could also take a look

Can discuss review assignments at planning today!

kofi-q (commented):
Some high-level thoughts before digging in:

Excited about the direction and not having to manually take screenshots (as someone who's had to do multiple runs of that recently 😅). I think having something that runs regularly and automatically is the right call.

From the description, it sounds like this is maybe doing too many things at once (or we're hoping it does).

If we're primarily focused on maintaining up-to-date screenshots for documentation, my nudge would be to leverage something closer to the `PreviewApp` we have in VxScan, given that there's already a lot of functional coverage in the other tests. That would have the advantage of being less flaky (0% flaky?) and a lot faster to run, debug, and update.

In a former life, I worked on a project where we had a setup similar to `PreviewApp` with test data/props for each app state we cared about (no backend needed) and the headless browser would just page through a bunch of links and take screenshots, with a few bits of interaction here and there - made it easier to add a new screen/states and easier to debug and fix too, being able to open up the failing mocked screen directly and not have to run the test or get the app into a particular state.

Would be a bigger upfront cost though, so not pushing for it, but worth thinking about, WRT maintenance, especially if we plan to do this more broadly for all the apps.

apps/admin/integration-testing/package.json
```diff
@@ -25,8 +25,9 @@
     ]
   },
   "dependencies": {
-    "@playwright/test": "^1.37.1",
+    "@playwright/test": "^1.48.1",
```
kofi-q, line 28: Why the bump here? Wondering if we should pick a version that sort of matches our kiosk-browser electron version.

And while you're here, might be worth making all the Playwright versions consistent across the repo, so we don't have multiple chromium installations floating around - I'm seeing `1.40.1` in our printing packages

apps/admin/integration-testing/playwright.config.ts
```diff
@@ -12,6 +12,7 @@ dotenv.config({ path: resolve(__dirname, '.env') });
  */
 export default defineConfig({
   testDir: './e2e',
+  timeout: 60_000,
```
kofi-q, line 15: Might need to go higher - seeing flaky timeouts in the latest CircleCI run

kofi-q (approved):

apps/admin/integration-testing/package.json
adghayes: It was get to access to the `clock` API. The monorepo validation script reminded me to update it elsewhere, but not in libs/printing or design, so I'll update those and look into why monorepo didn't check there.

apps/admin/integration-testing/playwright.config.ts
adghayes: Where?

adghayes:
@kofi-q thanks for reviewing. I appreciate the suggestion on a `PreviewApp` type entity, I think that would be cleaner. I'd like to see how this iteration of the test sits with the repo / flakiness an go from there. One advantage of the current full integration test approach is that it can generate artifacts, like PDF reports, that are not generated on the frontend.

apps/admin/integration-testing/playwright.config.ts
kofi-q: If you open up the `admin-integration-test` -> `Test` step, it has the logs of two timed out attempts and a third successful one: https://app.circleci.com/pipelines/github/votingworks/vxsuite/17636/workflows/807e1e8b-f7be-49e6-a307-bb8889e440a9/jobs/706679

apps/admin/integration-testing/package.json
kofi-q: Ah, makes sense it wouldn't catch those, since those are importing the general `plapywright` package

kofi-q:
> One advantage of the current full integration test approach is that it can generate artifacts, like PDF reports, that are not generated on the frontend.

Don't we have image comparison tests for those already? And if it's something we're displaying in the UI, that could fall under the category of test data as well, FWIW

apps/admin/integration-testing/playwright.config.ts
adghayes: good catch! i think the issue is not the timeout but a point where the test can sometimes hang, which I can fix.

adghayes:
@kofi-q it's true, it could be test data. In a frontend-only screenshot generation setup, we'd have to have the PDFs pre-generated. It could be sourced by the backend snapshots, however, so that could potentially be easy.

apps/admin/integration-testing/playwright.config.ts
kofi-q: Side note: also now wondering if retries got turned on automatically in newer versions of Playwright (or if I've just never noticed that before as the default setting) - it's great for getting through CI without intervention, but not so great for our visibility into flaky tests, so I'm torn on whether or not we should leave it on

Issue #4567, opened 2024-02-02
Run OpenSSL in FIPS-compliant mode


arsalansufi:
See https://votingworks.slack.com/archives/CEY45Q4JX/p1706881337663999

General context here: https://github.com/openssl/openssl/blob/master/README-FIPS.md

arsalansufi:
Current plan is for @amcmanus to think through how to best build FIPS-compliant OpenSSL in the context of Trusted Build and @arsalansufi to reason through and update application code, e.g., adding `-provider fips` to relevant calls

PR #5556, opened 2024-10-28
Surface more specific error messages when smart cards are rejected because of out-of-sync clocks

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/5043

This PR should address one of the most common sources of internal confusion when preparing smart cards. If a VxAdmin and a VxScan's clocks are out of sync, specifically if a VxAdmin has a time ahead of a VxScan, you can run into the following confusing sequence of events:

- Program a new smart card on VxAdmin.
- Try to use it on VxScan.
- Card is considered invalid because the card has a cert start date/time at a time in the future. UI surfaces a generic "card is invalid" screen without a clear indication of what to do next.
- Wait however long you need for VxScan's clock to catch up to the time on the VxAdmin at the time of programming, try again, and everything works.

This PR avoids this confusion by surfacing a distinct error message for this case.

This PR also surfaces a distinct error message for the opposite case that a card is expired (or a machine has a misconfigured date/time far in the future past a card's expiration date).

## Demo Video or Screenshot

| Card cert not yet valid | Card cert expired |
| - | - |
| cert-not-yet-valid | cert-expired |

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

## Checklist

- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced~ ➡️ Relying on existing logging


libs/auth/src/card.ts
```diff
@@ -90,15 +111,15 @@ export function areElectionManagerCardDetails(
...
+  return cardDetails.user?.role === 'poll_worker';
 }
 
 /**
  * A sub-type of CardStatus
  */
 export interface CardStatusReady<T = CardDetails> {
   status: 'ready';
-  cardDetails?: T;
+  cardDetails: T;
```
arsalansufi, line 122: We previously used an `undefined` `cardDetails` object to indicate that a card is unprogrammed or invalid. Now, we always have a `cardDetails` object when a card is inserted and registered, but set `cardDetails.user` to `undefined` when a card is unprogrammed or invalid. Many of the changes in this and subsequent commits are accounting for this change

libs/auth/src/java_card.ts
```diff
@@ -438,19 +455,33 @@ export class JavaCard implements Card {
...
-  private async safeReadCardDetails(): Promise<Optional<CardDetails>> {
+  private async safeReadCardDetails(): Promise<CardDetails> {
     try {
       return await this.readCardDetails();
-    } catch {
-      return undefined;
+    } catch (error) {
+      const errorMessage = extractErrorMessage(error);
+      const reason = (() => {
+        if (errorMessage.includes('certificate has expired')) {
```
arsalansufi, line 464: Openssl shell commands don't seem to return reliably documented error codes, nor do we have code in place to propagate such error codes, so we inspect error messages

libs/auth/src/java_card.ts
```diff
@@ -438,19 +455,33 @@ export class JavaCard implements Card {
...
+          return 'certificate_expired';
+        }
+        /* istanbul ignore next: It's hard to create test certs with start dates in the future
```
eventualbuddha, line 467: Could we create one that isn't valid for 100 years?

libs/auth/src/java_card.ts
eventualbuddha: This feels fragile. If we want to keep this instead of using the OpenSSL API directly, could we at least do a version check and assert that it's an acceptable version?

eventualbuddha (approved):

libs/auth/src/java_card.ts
arsalansufi: So that won't quite do it because that'll just bring up the cert end date, when what we want is to push out the cert start date

I do in fact create a cert that is valid for a very short duration to test the `certificate_expired` case: https://github.com/votingworks/vxsuite/blob/3a84241d7a70c99c1fb63d3b7d339fa00ff82b42/libs/auth/scripts/src/generate_dev_keys_and_certs.ts#L288-L292

libs/auth/src/java_card.ts
arsalansufi: I like the version check idea! I'll also add that we'll get another layer of safety once Adam and I complete https://github.com/votingworks/vxsuite/issues/4567, which will be making it into the release. To run OpenSSL in FIPS compliant mode, we're going to be building OpenSSL from source. That'll be part of Trusted Build, so in production, we'll be pinned to an exact, reproducible build of OpenSSL. And I'm planning to keep dev and prod consistent here, so that build from source will be part of dev env setup, too.

An automated test for this case would also help, but as noted in the code comment you later reference, it's hard to create test certs with start dates in the future. There are ways to do it to be clear, but they require either messing around with the system time using `sudo`, which doesn't lend itself nicely to a script that can be used by others, or switching the cert creation logic from `openssl x509` to `openssl ca`, which I don't want to do this late in the game. But would like to update post cert freeze!

libs/auth/src/java_card.ts
eventualbuddha: Sorry, restating my comment above more clearly:

Could we create a certificate whose validity window doesn't start until 100 years from now?

arsalansufi: Ah yup, so I posted some thinking on creating certs with a start date in the future above:

> there are ways to do it to be clear, but they require either messing around with the system time using sudo, which doesn't lend itself nicely to a script that can be used by others, or switching the cert creation logic from openssl x509 to openssl ca, which I don't want to do this late in the game. But would like to update post cert freeze!

`openssl x509 -req` always sets the cert start time to the current time.
`openssl ca -in` supports a `-startdate` param, but I don't want to make that switch right before cert.

The idea of using `sudo` to temporarily change the time isn't terrible, but also requires you to toggle some settings around time syncing.

I could also make such a cert in a one-off capacity, but it would then be hard for others to regenerate it should the need ever arise. It'd be more like a manual fixture. Which is maybe fine if I label it as such, but I'm not a huge fan of that idea.

eventualbuddha: Okay, that makes sense. This PR is really strongly making me feel that using the `openssl` binary directly is _not_ a great idea. I understand the FIPS requirement, but I think this should be on the list of things to consider changing in the future. For now, this is fine.

arsalansufi: Yup I totally agree! This PR convinced me of that as well. Would be great to have proper error codes.

I wanted to get these updated error screens into this release because they come up so often but wouldn't want us to continue down the route of conditioning on error message strings for too long.

Funnily enough, there's discussion on changing the OpenSSL error message for this case to clue users into clock mismatches 😄 https://github.com/openssl/openssl/issues/14771

Issue #3769, opened 2023-07-28
Update machine certs to include the machine ID, as certified by VxCertifier


carolinemodic:
We need to update the way we do machine certs to include the machine ID, as certified by VxCertifier. Otherwise a machine can just claim to be whatever it wants to be. This is just an extra field in the certificate, and a tweak to how we run VxCertifier. cc @arsalan

Issue automatically created from slack feedback reported by ben 
https://votingworks.slack.com/archives/CJU9MSC6S/p1690554554157169?thread_ts=1690554554.157169&cid=CJU9MSC6S

Issue #3322, opened 2023-04-20
Auth: Reconsider how we access VxAdmin's jurisdiction and all machine's machine IDs, given their presence in machine certs


arsalansufi:
We currently pull these fields from env vars which further pull from static files in vxsuite-complete-system:

- [VX_MACHINE_ID](https://github.com/votingworks/vxsuite-complete-system/blob/78f0858f525def8bfc0c08d84a77a4808649cc7b/config/read-vx-machine-config.sh#L43)
- [VX_MACHINE_JURISDICTION](https://github.com/votingworks/vxsuite-complete-system/blob/78f0858f525def8bfc0c08d84a77a4808649cc7b/config/read-vx-machine-config.sh#L55-L57)

We could always pull them directly from the machine cert. Or as a middle ground, we could pull them from the machine cert to populate the env vars and continue to use the env vars. Lots of permutations, all with their own tradeoffs.

Relevant past discussions:
- https://github.com/votingworks/vxsuite-complete-system/pull/291#discussion_r1166861087
- https://github.com/votingworks/vxsuite/pull/5557#discussion_r1818373350

Also just noting that there's no security issue here as, where it matters for security, we always use the machine cert for this data, and that too only after having verified the cert.

PR #5557, opened 2024-10-28
Update machine certs to include machine ID

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3769

Corresponding vxsuite-complete-system change: https://github.com/votingworks/vxsuite-complete-system/pull/401

This PR updates machine certs to include machine ID, e.g., SC-09-001. A field which previously only existed outside the machine cert. This gives us a way to verify the authenticity of machine IDs.

We use this new cert field to:
1. Verify that the machine ID in a CVR export directory name is authentic (we display machine IDs in the VxAdmin CVR import modal using these names)
2. Retrieve the machine ID for signed hash validation

## Testing Plan

- [x] Updated automated tests
- [ ] Manually tested

## Checklist


libs/auth/src/jurisdictions.ts
```diff
@@ -2,7 +2,7 @@
  * The jurisdiction in all dev certs. Follows the format of a prod cert jurisdiction
  * ({state-2-letter-abbreviation}.{county-or-town}).
  */
-export const DEV_JURISDICTION = 'st.dev-jurisdiction';
+export const DEV_JURISDICTION = 'vx.test';
```
arsalansufi, line 5: As noted in the commit message, I'm switching the dev jurisdiction to match what we use for prod QA images (which in a separate PR I'm going to make more similar to dev, in terms of keys/certs used).

I'm adding this to this PR because I'm already having to regenerate dev keys/certs for the machine ID change. And I wanted to lump in this jurisdiction change to avoid yet another change to dev keys/certs after this PR.

Though I think that few engineers are actually still using physical smart cards in dev (as opposed to the dev dock), I'll still be sure to note that dev smart cards will need to be reprogrammed after this PR.

libs/auth/src/artifact_authentication.ts
```diff
@@ -164,35 +165,15 @@ function deserializeArtifactSignatureBundle(
...
-        certDetails.component === 'admin',
-        'Signing machine cert for election package should be a VxAdmin cert'
-      );
-      break;
-    }
-    /* istanbul ignore next: Compile-time check for completeness */
-    default: {
-      throwIllegalValue(artifact, 'type');
-    }
-  }
```
arsalansufi, line 191: Moved to the `performArtifactSpecificAuthenticationChecks` function below, which I think is a better fit for this logic

libs/auth/src/artifact_authentication.ts
```diff
@@ -247,10 +232,31 @@ function constructSignatureFilePath(artifact: ArtifactToImport): string {
...
+      assert(
+        exportDirectoryNameComponents !== undefined,
+        `Error parsing export directory name: ${exportDirectoryName}`
+      );
+      assert(
+        exportDirectoryNameComponents.machineId ===
+          signingMachineDetails.machineId,
+        `Machine ID in export directory name doesn't match machine ID in signing machine cert: ` +
+          `${exportDirectoryNameComponents.machineId} != ${signingMachineDetails.machineId}`
+      );
```
arsalansufi, line 258: I contemplated hopping a level deeper and verifying the various machine ID related fields in the CVR files themselves but chose not to for a few reasons:

- The CDF fields that house machine ID are designed to hold multiple machine IDs in the case of a consolidated export, so not just the exporting machine ID
- We pull from the export directory name, not the file contents, to surface machine ID in the VxAdmin import modal
- We already validate that the contents of the CVR files are unaltered

From a UI perspective, with this change in place, a user will still see an altered machine ID in the VxAdmin import modal. But on click to import, will be shown our existing "Error authenticating cast vote records" error message.

apps/admin/backend/src/app.ts
```diff
@@ -254,12 +254,11 @@ function buildApi({
 
     /* istanbul ignore next */
     async generateSignedHashValidationQrCodeValue() {
-      const { codeVersion, machineId } = getMachineConfig();
+      const { codeVersion } = getMachineConfig();
```
arsalansufi, line 257: `getMachineConfig` does still read machine ID from the `VX_MACHINE_ID` env var which reads from the `/vx/config/machine-id` file. We could eventually get rid of the `/vx/config/machine-id` file and always pull the machine ID from the machine cert. But for the few touch points where it truly matters, we are reading from the cert (after verifying the cert) (as you'll see below).

It ballooned scope too much under our time constraints to completely get rid of the `/vx/config/machine-id` file. But I will ticket that future change and same for the analogous change for jurisdiction ID on VxAdmin.

libs/fixture-generators/bin/generate-cvr-fixtures
```diff
@@ -4,18 +4,14 @@
 
 export NODE_ENV=development
 
-rm -rf ../fixtures/data/electionGridLayoutNewHampshireTestBallot/castVoteRecords/generated
-./bin/generate-cvrs --electionDefinition ../fixtures/data/electionGridLayoutNewHampshireTestBallot/election.json --outputPath ../fixtures/data/electionGridLayoutNewHampshireTestBallot/castVoteRecords/generated
+rm -rf ../fixtures/data/electionGridLayoutNewHampshireTestBallot/castVoteRecords/machine_0000__2024-01-01_00-00-00
```
arsalansufi, line 7: Had to rename the CVR fixture directory names to ensure that they pass the artifact auth check of machine ID

libs/fixture-generators/bin/generate-cvr-fixtures
```diff
@@ -4,18 +4,14 @@
...
-./bin/generate-cvrs --electionDefinition ../fixtures/data/electionPrimaryPrecinctSplits/election.json --outputPath ../fixtures/data/electionPrimaryPrecinctSplits/castVoteRecords/generated
+rm -rf ../fixtures/data/electionPrimaryPrecinctSplits/castVoteRecords/machine_0000__2024-01-01_00-00-00
+./bin/generate-cvrs --electionDefinition ../fixtures/data/electionPrimaryPrecinctSplits/election.json --outputPath ../fixtures/data/electionPrimaryPrecinctSplits/castVoteRecords/machine_0000__2024-01-01_00-00-00
 
 echo
-<<<<<<< HEAD
 echo "Note: You need to run \`pnpm build:resources && pnpm build\` in libs/fixtures for the new fixtures to register"
-=======
-echo "Note: You need to run \`pnpm build:resources && pnpm build\` in libs/fixtures for the new fixtures to register in this library's tests"
->>>>>>> a96ae87fc (regen fixtures after rebase)
```
arsalansufi, line 21: Looks like some missed merge conflict markers from a past code change

arsalansufi:
Most relevant commits for security review @benadida are:
- https://github.com/votingworks/vxsuite/pull/5557/commits/8ea5bce0b0f318e9f58f8cb43c7ca75afee92986
- https://github.com/votingworks/vxsuite/pull/5557/commits/873e70dd22595c5931301e2f0e6ed094bf936f37
- https://github.com/votingworks/vxsuite/pull/5557/commits/3fa893c0878408d78c5e5e280c4912b93ad82fe2

libs/auth/src/certs.ts
```diff
@@ -44,23 +47,28 @@ export const STANDARD_CERT_FIELDS = [
...
 }
 
 interface VxMarkScanCustomCertFields {
   component: 'mark-scan';
+  machineId: string;
 }
 
 interface VxScanCustomCertFields {
   component: 'scan';
+  machineId: string;
```
benadida, line 71: should these interfaces now extend a base interface?

libs/auth/src/certs.ts
```diff
@@ -370,20 +393,39 @@ export function constructCardCertSubjectWithoutJurisdictionAndCardType(): string
...
 
+/**
+ * Parses the provided cert and returns machine details. Throws an error if the cert doesn't follow
+ * VotingWorks's machine cert format.
+ */
+export async function parseMachineDetailsFromCert(
+  cert: Buffer
+): Promise<MachineDetails> {
+  const certDetails = await parseCert(cert);
+  assert(certDetails.component !== 'card', 'Cert is not a machine cert');
```
benadida, line 404: is this the right assert, or can we future proof with `component==='machine'` (or maybe we don't have a component for machines?)

libs/auth/src/certs.ts
```diff
@@ -370,20 +393,39 @@ export function constructCardCertSubjectWithoutJurisdictionAndCardType(): string
...
+  const { component, machineId } = certDetails;
+  return {
+    machineType: component,
+    machineId,
+    jurisdiction: component === 'admin' ? certDetails.jurisdiction : undefined,
```
benadida, line 409: should we be stricter and fail if there's a jurisdiction and it's not a `vxadmin` ?

libs/auth/src/certs.ts
```diff
@@ -370,20 +393,39 @@ export function constructCardCertSubjectWithoutJurisdictionAndCardType(): string
...
-export function constructMachineCertSubject(
-  machineType: MachineType,
-  jurisdiction?: string
-): string {
+export function constructMachineCertSubject({
+  machineType,
+  machineId,
+  jurisdiction,
+}: MachineDetails): string {
   assert(
```
benadida, line 421: we may want to assert a format of `machineId` so we don't get in trouble later if it's got weird characters.

libs/auth/src/machine_ids.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * The machine ID in dev certs
+ */
+export const DEV_MACHINE_ID = '0000';
```
benadida, line 4: suggest a machine ID that actually says DEV

libs/auth/src/artifact_authentication.ts
```diff
@@ -164,35 +165,15 @@ function deserializeArtifactSignatureBundle(
...
-  }
+  signingMachineCert: Buffer
+): Promise<MachineDetails> {
   await verifyFirstCertWasSignedBySecondCert(
     signingMachineCert,
     config.vxCertAuthorityCertPath
   );
+  const signingMachineDetails =
+    await parseMachineDetailsFromCert(signingMachineCert);
+  return signingMachineDetails;
```
benadida, line 176: in the process of this change you've made an improvement – no trying to parse until the signature is verified. That is important to protect against potential malicious data formats that try to screw up our parsing.

libs/auth/src/artifact_authentication.ts
benadida: consider an attacker who has access to the key for `machine-1`. They thus can only create a CVR directory that says `machine-1`. But can the CVRs falsely say they come from `machine-2`? What will the final report say in VxAdmin? Will it ever surface `machine-2`? Or maybe there's a later check?

apps/admin/backend/src/app.ts
benadida: as long as we're verifying the machine ID in certs at signature verification time, I'm not worried about this. Yes, for simplicity and consistency, good to fix this later, but not urgent in this PR.

libs/auth/src/signed_hash_validation.ts
```diff
@@ -91,6 +91,9 @@ export async function generateSignedHashValidationQrCodeValue(
     : '';
   const date = new Date();
 
+  const machineCert = await fs.readFile(config.machineCertPath);
+  const { machineId } = await parseMachineDetailsFromCert(machineCert);
```
benadida, line 95: I don't think you need the `machineId` in the payload anymore for the signed QR code, it's part of the cert.

benadida (approved):
security looks good modulo a couple of small comments, and one topic to discuss – what happens if the CVRs reference a different machine ID?

libs/auth/src/certs.ts
eventualbuddha: Along these lines, `machineId` could then be typed `MachineId`, a "newtype"'d `string`.

libs/auth/src/machine_ids.ts
eventualbuddha: Related, we seem to hardcode this value in many places. Can they reference this version instead?

apps/admin/backend/src/app.ts
eventualbuddha: Can we mark the property as deprecated, and link to the issue you (will have) created?

```ts
export interface MachineConfig {
  /**
   * @deprecated Use the machine ID from the machine's certificate instead.
   */
  machineId: string;
```

libs/auth/src/certs.ts
eventualbuddha: `component` is otherwise the type of machine, like `mark-scan`, `admin`, etc.

eventualbuddha (approved):
I don't see any red flags here. I commented sporadically where we might consider changing something.

libs/auth/src/artifact_authentication.ts
arsalansufi: The attacker won't be able to display any machine ID other than `machine-1` in the CVR import modal. But yes, they could put an arbitrary machine ID inside the CVRs (per our assumption that they have access to machine 1's private key and can sign arbitrary data blobs).

I didn't want to over-rotate on this case because we eventually will have cases where a CVR contains machine IDs other than the signing machine, namely when VxAdmin will export aggregated post-adjudication CVRs.

But it seems like a fair assumption for now that VxScan will never export anything but its own CVRs. So I can add this check.

libs/auth/src/certs.ts
arsalansufi: Yeah Brian covered what I was going to say. There is no catch-all `machine` component, just component types for every machine.

`component !== card` does narrow the type down to the other components, all of which are machines. I could alternatively delineate every machine, e.g., `component === 'admin' || component === 'central-scan' || ...`.

Regardless, on the next line, we run:

```
const { component, machineId } = certDetails;
```

Which only compiles successfully if `certDetails` has been narrowed down to a machine cert details object (card certs don't have a machine ID). So the current code is safe / future-proof as is

libs/auth/src/certs.ts
arsalansufi: This should already be enforced when we parse with our Zod schema, which indicates that VxAdmin has a jurisdiction field while other machines don't. Though I may need to switch the schemas to use a `strictObject` instead of an `object` to fully guarantee this, as a `strictObject` disallows extra fields. Will confirm

Regardless, no assertion is necessary at this stage as `certDetails.jurisdiction` is only defined if `component === 'admin'` or `component === 'card'`

libs/auth/src/certs.ts
arsalansufi: We enforce the structure of the machine ID in the vxsuite-complete-system script that sets it, so I'm going to continue to rely on that for now: https://github.com/votingworks/vxsuite-complete-system/blob/78f0858f525def8bfc0c08d84a77a4808649cc7b/config/vendor-functions/choose-vx-machine-id.sh#L13-L23

libs/auth/src/machine_ids.ts
arsalansufi: I initially did that (in fact, I even went a step further and started by giving every machine type a unique dev machine ID) but then saw how much of the code/tests/snapshots use this existing 0000 ID. I'm now running quite short on time so will likely leave as is.

I noticed the hardcoding outside of this lib and considered changing it. Happy to do that

libs/auth/src/signed_hash_validation.ts
arsalansufi: Oh this is a good catch!

libs/auth/src/artifact_authentication.ts
arsalansufi: Ah nice! That makes sense

libs/auth/src/artifact_authentication.ts
arsalansufi: Added valildation for the field used to determine and persist machine/scanner ID on import: https://github.com/votingworks/vxsuite/pull/5557/commits/857bfaf43179d48f6b9b25d264bbb28f5fdf9fc9

libs/auth/src/machine_ids.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/5557/commits/e9e280f61942a38408054fb2849fb6d637deed1b

libs/auth/src/certs.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/5557/commits/aeadd1da5d6c1a432fed683f73648da4bfa5e7a4

libs/auth/src/signed_hash_validation.ts
arsalansufi: Noting that this will require an update to https://github.com/votingworks/livecheck-web, to extract machine ID from the cert

Will ticket that to make sure that we don't forget

arsalansufi: https://github.com/votingworks/vxsuite/pull/5557/commits/a49f1967f2240c563309d693206b5af6d5fccba7

Follow-up issue: https://github.com/votingworks/vxsuite/issues/5575

libs/auth/src/certs.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/5557/commits/4ee780457b37dafeac6a0509ffde989f9075f6dc

apps/admin/backend/src/app.ts
arsalansufi: Looking at this again, I'm not necessarily thinking of this field as deprecated. It's okay for us to group machine ID with code version in this interface, and to have people access machine ID through that interface. We just want to be populating the field differently, extracting from the machine cert instead of relying on the VX_MACHINE_ID env var, which is currently populated by the `machine-id` file in vxsuite-complete-system:

https://github.com/votingworks/vxsuite-complete-system/blob/78f0858f525def8bfc0c08d84a77a4808649cc7b/config/read-vx-machine-config.sh#L43

That said, we may still arrive at the conclusion that we want to remove the field from this interface.

I think it's a bit too early for me to tell without considering how I want to approach this holistically, across machines, machine ID + jurisdiction, and vxsuite + vxsuite-complete-system.

So I'm not going to add the deprecated marker yet.

I've grouped this together with the existing ticket I had for jurisdiction and updated accordingly: https://github.com/votingworks/vxsuite/issues/3322

Issue #1439, opened 2022-02-17
Improving the repo's licensing


cjerdonek:
While confirming this repo's license, I noticed that while the repo's `README` does say "GPLv3" at the bottom, the repo doesn't seem to contain a `COPYING` (or `LICENSE`) file, at least at the top level.

This page contains instructions on licensing a program under the GPL or one of its variants: https://www.gnu.org/licenses/gpl-howto.html

It includes, for example:

* Add a `COPYING` file.
* Give each file the proper copyright notices.
* Put a license notice in each file.
* Possibly including an "or (at your option) any later version" clause in the license notice, as discussed in the following linked page: https://www.gnu.org/licenses/identify-licenses-clearly.html

(Note there is a difference between the copyright notice and the license notice in each file.)

cjerdonek:
PS - I think the other repos should also be checked. For example, even though `vxsuite-complete-system` may not itself contain much code, it contains some. Thus, it should probably contain a `COPYING` (or `LICENSE`) file, too.

carolinemodic:
@benadida Can you own this?

adghayes:
Added `LICENSE` files, `package.json` tage, rust metadata, and README descriptions for consistent GPLv3 only licensing across the repo's used in the certified system:

- `vxsuite`: #5558 
- `vxsuite-complete-system`: https://github.com/votingworks/vxsuite-complete-system/pull/402
- `kiosk-browser`: https://github.com/votingworks/kiosk-browser/pull/166
- `vxsuite-build-system`: https://github.com/votingworks/vxsuite-build-system/pull/136

I followed the pattern typical to GitHub, which varies a little from GNU's standard instructions:
- using `LICENSE` files rather than `COPYING` files - `COPYING` is just a GNU convention rather than a requirement, and `LICENSE` is the one that Github automatically recognizes
- using a `LICENSE` file at the root to cover the whole repo. some projects have license disclaimers in every file but it's not required and a root `LICENSE` is common practice

PR #5558, opened 2024-10-28
Improve GPL Licensing

adghayes:
## Overview

Addresses #1439 for `vxsuite`, although I will need to do something similar for the other repo's.

Currently we have inconsistent licensing patterns:
- `LICENSE` files in some packages, not others
- References to GPL-v3 in some README's and `package.json`s but not all
- Some packages use GPL-v3 some use AGPL-v3

This PR attempts to standardize by:
- adding `LICENSE` file to the root and removing it everywhere else, as is common practice
- for Typescript packages, standardizing on a `license` field that is one of the SPDX official labels
- for Rust packages, setting the packages to inherit license from the `workspace` manifest
- standardizing on GPL-v3 rather than AGPL
- removing references to GPL from all README's except the root README, because I think that's a hard precendent to maintain

Let me know if you have any objections / suggestions to do this differently.


adghayes:
Assigning @eventualbuddha as primary reviewer, @benadida for visibility.

eventualbuddha (approved):
I have no concerns about the mechanics of making this change. Assuming @benadida is okay with the implications of AGPL → GPL, I think this is fine.

The only suggestion I have is to update the `validate-monorepo` script to ensure that the fields are set as expected for all packages.

benadida (approved):
I am good with standardizing on GPL 3.0 only (which means not "or later").

adghayes:
@eventualbuddha can you take a look at the monorepo validation I added?

Another inconsistency in our packages - we sometimes have an `author` specified as `VotingWorks Engineering <eng@voting.works>`, sometimes we don't. Do either of you have an inclination to standardize either way on that?

libs/@types/jsfeat/package.json
```diff
@@ -1,5 +1,6 @@
 {
     "name": "@types/jsfeat",
```
adghayes, line 2: @eventualbuddha does this package need to exist? I don't see it imported anywhere.

eventualbuddha: It does not. We stopped using `jsfeat` when we switched the ballot interpreter to Rust.

eventualbuddha:
> Another inconsistency in our packages - we sometimes have an `author` specified as `VotingWorks Engineering <eng@voting.works>`, sometimes we don't. Do either of you have an inclination to standardize either way on that?

I don't really care. It's probably easier to leave it out, and as long as the tooling we use doesn't complain then let's remove it.

eventualbuddha (approved):
`validate-monorepo` changes LGTM 👍

Issue #5516, opened 2024-10-14
Make sure that default BMDB font size is large enough to meet printed paper requirements


arsalansufi:
https://votingworks.slack.com/archives/CJU9MSC6S/p1728483547121629

PR #5567, opened 2024-10-30
[BMD] Bump up print font size so capital letters are >= 3.5mm

kofi-q:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/5516

Capital letters on BMD prints are currently measuring at `~2.65-2.70mm`, with full font size measuring at `~3.45-3.5mm`. VVSG 2.0 sets a minimum requirement of 10pt (~3.5mm) font, but it's unclear whether this refers to standard 10pt prints (where the full font height - ascender to descender is 3.5mm) or to the capital letter height needing to be 3.5mm, as it calls out specifically in the section on screen font size.

To be safe, applying the same scaling we use for on-screen fonts (scale letters up so that the capital letter height matches the target font size) on the BMD paper ballot. Basically: `10pt  * (1 / 0.725)`, or (for capital letters): `2.65mm * (1/0.725) = ~3.65mm`.

Also enabling the BMD-150 paper dimensions for the MarkScan tests, to match prod. Looks like the `bmd-ballot-fixtures` code uses a different rendering path, so wasn't able to update that easily - leaving as is for now.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/a64c9564-cdc6-4be6-97a4-265fa6bd552b

## Testing Plan
- Local, on-screen testing - not the most accurate, but should give a rough enough idea 
- Haven't tested on a prod 150 yet - don't have a machine  at the moment, so will have someone else test when we build our next image 

## Checklist


libs/ui/src/bmd_paper_ballot.tsx
```diff
@@ -308,7 +343,6 @@ const ContestTitle = styled.div`
   display: flex;
   flex-wrap: wrap;
   gap: 0.25em;
-  font-size: 1.125em;
```
kshen0, line 311: Was this being overridden or otherwise unused?

kshen0 (approved):
Thanks @kofi-q  for the detailed PR, fixture updates, and offline conversation about this methodology.

Point and call: "full font size" height is measured as distance between ascender and descender eg. bottom of the stem on `p` to top of the stem on `d`. Therefore, while `10 pt font == 3.5mm` from stem to stem, a capital letter `I` is significantly smaller. VVSG 2.0 uses capital letter `I` as its metric so just using `10 pt` font isn't sufficient to achieve a 3.5mm capital letter `I`.

To achieve a 3.5mm capital letter `I` we scale font point size by the ratio of full font size to actual capital letter size. That is, the height of a capital `I` is 72.5% the height of the full font size from ascender to descender, so we use a ratio of `1 / 0.725` defined in `make_theme.ts`.

The result of scaling the font is just under `13.8 pt` font, so we need to adjust our ballot layouts accordingly to avoid overflow.

libs/ui/src/bmd_paper_ballot.tsx
kofi-q: Ah, forgot to call this out - decided to reduce the font size of the contest titles to match the selections, to create a bit more room on the ballot. With the extra `~3pts` we're applying, they were starting to take up too much space and making it tough to fit enough contests at the default size

kofi-q (commented):
Yes! Thanks for the helpful summary. And for the offline conversation - helped remind myself about some of the details of the initial screen font scaling investigation

PR #5570, opened 2024-10-30
fix(central-scan): prevent `scanimage` race

eventualbuddha:
## Overview
Ever since #4994 we call `scanimage` twice when starting a batch, once to determine whether an imprinter is present and once to actually start the scan. Both invocations require communicating with the scanner, which means both need to open the USB device. Because we previously resolved the `isImprinterAttached` promise early in the case where the expected error message is present, the first `scanimage` instance could still be running and holding the USB device open when we try to run the second `scanimage` instance. When that happens, the second instance would fail to open the USB device and we'd get a zero-sheet batch and this error message: `scanimage: open of device fujitsu failed: Invalid argument`.

In my testing on macOS host with a Debian VM with the attached fi-7180, this happens on every scan attempt. It may be that different configurations behave differently, as the outcome of the race is non-deterministic. This change fixes the issue by waiting until the `scanimage` process exits to resolve the `isImprinterAttached` promise.

Additionally, I changed a few other things:
- expand the shorthand `scanimage` option `-d` to `--device-name`
- collect `scanimage` stderr before checking for the expected error message in case it is split across multiple chunks (unlikely but possible)
- update tests to fail if the `isImprinterAttached` promise resolves early or if we don't see the error message split across multiple stderr chunks

## Demo Video or Screenshot
| Before (0-sheet batch) | After (3-sheet batch) |
|-|-|
| CleanShot 2024-10-30 at 16 39 34@2x | CleanShot 2024-10-30 at 16 40 33@2x |

## Testing Plan
Tested manually with the fi-7180 on macOS host / Debian VM

### DIFF
--- a/apps/central-scan/backend/src/fujitsu_scanner.ts
+++ b/apps/central-scan/backend/src/fujitsu_scanner.ts
@@ -140,7 +140,7 @@ export class FujitsuScanner implements BatchScanner {
       `--batch=${join(directory, `${dateStamp()}-ballot-%04d.${this.format}`)}`,
       `--batch-print`,
       `--batch-prompt`,
-      // If the sheet is smaller then the given size fill in extra image space with black
+      // If the sheet is smaller than the given size fill in extra image space with black
       `--bgcolor=black`,
     ];
 
--- a/apps/central-scan/backend/src/fujitsu_scanner.ts
+++ b/apps/central-scan/backend/src/fujitsu_scanner.ts
@@ -93,10 +93,10 @@ export class FujitsuScanner implements BatchScanner {
     );
   }
 
-  async isImprinterAttached(): Promise<boolean> {
+  isImprinterAttached(): Promise<boolean> {
     return new Promise((resolve) => {
       const process = streamExecFile('scanimage', [
-        '-d',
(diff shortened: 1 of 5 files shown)


apps/central-scan/backend/src/fujitsu_scanner.ts
```diff
@@ -93,27 +93,32 @@ export class FujitsuScanner implements BatchScanner {
...
         '--dont-scan',
       ]);
 
       assert(process.stderr);
+      let stderr = '';
       process.stderr.on('data', (data: string) => {
-        // If there is no imprinter attached a message will be sent to stderr
-        // that the endorser parameter is readonly.
-        if (data.includes(EXPECTED_IMPRINTER_UNATTACHED_ERROR)) {
-          resolve(false);
```
eventualbuddha, line 111: Here is the problematic early `resolve` call. This PR changes it so we always resolve in the `close` event handler.

In addition to happening too early, the way this is set up also means that `resolve` gets called twice. This is actually allowed by the JavaScript spec and the second call will just be ignored, but it is confusing to read.

carolinemodic (approved):

PR #5573, opened 2024-10-31
chore(deps): upgrade various vulnerable packages

eventualbuddha:
Upgrades various packages past known vulnerable versions:

- https://github.com/advisories/GHSA-67hx-6x53-jw92
- https://github.com/advisories/GHSA-f9xv-q969-pqx4
- https://github.com/advisories/GHSA-74fj-2j2h-c42q
- https://github.com/advisories/GHSA-c24v-8rfc-w8vw
- https://github.com/advisories/GHSA-87hq-q4gp-9wr4

None of these were clearly exploitable, but the risk of the upgrades should be low as they're mostly development dependencies. The one that we do use in the application is `react-pdf`, but I kept the upgrade as minimal as possible (7.7.0 → 7.7.3, current is 9.1.1).

We still have a number of known vulnerable dependencies with `high`, `moderate`, and `low` severity. None of the remaining `high` ones are clearly exploitable:

- https://github.com/advisories/GHSA-p6mc-m468-83gw – `lodash.set` and `lodash.setWith`: used in tests and with no user-generated property paths
- https://github.com/advisories/GHSA-wgrm-67xf-hhpq – arbitrary code execution in `pdfjs-dist` for malicious PDFs: we only show PDFs we generate
- https://github.com/advisories/GHSA-grv7-fg5c-xmjg – uncontrolled resource consumption in `braces`: this is a transitive dependency of `lint-staged` and `stylelint`, which take our code and paths as input
- https://github.com/advisories/GHSA-3h5v-q93c-6h6q – crash in `ws` given too many headers: we don't use websockets, and we don't accept requests outside `localhost`
- https://github.com/advisories/GHSA-c2qf-rxjj-qqgw – `semver` DoS, but not in the way it is transitively used in our codebase
- https://github.com/advisories/GHSA-2p57-rm9w-gvfp – `ip` vulnerability, used via development-only `@storybook/*` packages
- https://github.com/advisories/GHSA-qwcr-r2fm-qrc7 – `body-parser` vulnerability, used via development-only `@storybook/*` packages
- https://github.com/advisories/GHSA-9wv6-86v2-598j – `path-to-regexp` vulnerability, used by `fetch-mock` in tests, `react-router` with fixed vendor-provided strings, and `@storybook/cli` in development


jonahkagan (approved):

arsalansufi:
Thank you for the detailed consideration here Brian! I agree with your assessments and think that these changes are sufficient for v4

I may end up merging this on your behalf before I make images today, FYI

PR #5576, opened 2024-11-03
style: separate formatting & linting

eventualbuddha:
## Overview
Running `prettier` as part of `eslint` isn't great for a couple reasons:
- seeing red error squiggles when the formatting just needs an autofix is annoying and potentially confusing
- any ESLint autofixes that result in invalid formatting will cause ESLint to re-run linting multiple times until there are no more changes to be made, causing slow performance

## Demo Video or Screenshot
n/a

## Testing Plan
As part of running `git commit` to create this PR, worked through any issues that `lint-staged` flagged with the configuration and fixed them.

### DIFF
--- a/apps/admin/backend/package.json
+++ b/apps/admin/backend/package.json
@@ -27,17 +27,6 @@
     "test:watch": "TZ=America/Anchorage jest --watch",
     "type-check": "tsc --build"
   },
-  "lint-staged": {
-    "*.+(css|graphql|json|less|md|mdx|sass|scss|yaml|yml)": [
-      "prettier --write"
-    ],
-    "*.+(js|jsx|ts|tsx)": [
-      "eslint --quiet --fix"
-    ],
-    "package.json": [
-      "sort-package-json"
-    ]
-  },
   "dependencies": {
     "@votingworks/auth": "workspace:*",
     "@votingworks/backend": "workspace:*",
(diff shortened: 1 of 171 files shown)


apps/admin/backend/package.json
```diff
@@ -27,17 +27,6 @@
     "test:watch": "TZ=America/Anchorage jest --watch",
     "type-check": "tsc --build"
   },
-  "lint-staged": {
-    "*.+(css|graphql|json|less|md|mdx|sass|scss|yaml|yml)": [
-      "prettier --write"
-    ],
-    "*.+(js|jsx|ts|tsx)": [
-      "eslint --quiet --fix"
-    ],
```
jonahkagan, line 36: How was prettier being run as part of eslint before? I see how the new pattern now runs it on JS/TS files, but I can't tell how it was being run on those files before, so I'm having trouble evaluating how we know it's no longer being run as part of eslint.

jonahkagan (approved):

apps/admin/backend/package.json
eventualbuddha: I went looking for the answer and realized that we were still including `eslint-plugin-prettier` in `vx/react` contexts. I fixed that and removed `eslint-plugin-prettier` in [4cb71cc](https://github.com/votingworks/vxsuite/pull/5576/commits/4cb71cc1cd9ebb3942827e8b52369e8179612e06). The same change was made in `vx/recommended` to disable prettier within eslint.

PR #5580, opened 2024-11-06
always allow polls transitions when continuous export is disabled

### LINKED ISSUE #5579: Cannot Change Polls w/ Continuous Export Disabled & No USB Drive
The purpose of disabling continuous CVR export is to allow voting to proceed without a connected USB drive. This works during scanning, but poll worker transitions (open, close, pause, resume) are still disabled partially defeating the point of the feature.

adghayes:
## Overview

Closes #5579.

### DIFF
--- a/apps/scan/frontend/src/screens/poll_worker_screen.tsx
+++ b/apps/scan/frontend/src/screens/poll_worker_screen.tsx
@@ -23,7 +23,6 @@ import type {
   PrecinctScannerPollsInfo,
   PrintResult,
 } from '@votingworks/scan-backend';
-import type { UsbDriveStatus } from '@votingworks/usb-drive';
 import {
   getUsbDriveStatus,
   printReport,
@@ -34,6 +33,7 @@ import {
   resumeVoting as resumeVotingApi,
   getPollsInfo,
   useApiClient,
+  getConfig,
 } from '../api';
 import { FullScreenPromptLayout } from '../components/full_screen_prompt_layout';
 import {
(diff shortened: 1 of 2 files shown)


apps/scan/frontend/src/screens/poll_worker_screen.tsx
```diff
@@ -148,13 +148,20 @@ function OpenPollsPromptScreen({
...
+            }
           >
             Open Polls
           </Button>
         </P>
         <PrinterAlertText printerSummary={printerSummary} />
-        <UsbDriveAlertText usbDriveStatus={usbDriveStatus} />
+        <UsbDriveAlertText
+          mustInsertUsbDriveToContinue={mustInsertUsbDriveToContinue}
+        />
```
arsalansufi, line 164: nit: Rather than having `UsbDriveAlertText` internally render `null` when `mustInsertUsbDriveToContinue` is `false`, I think the code may read clearer if we pull that check out of the component, i.e.

```tsx
{mustInsertUsbDriveToContinue && <UsbDriveAlertText />}
```

arsalansufi (approved):
Thank you for fixing!

Also just confirming that I did test that, on v3.1.2, polls can be closed without a USB drive inserted if continuous export is disabled, per the [v3.1.2 QA checklist](https://docs.google.com/document/d/1AQ6EQNiIBm2G86capEsJajgBHf1eCHhk1Z5QbSdj4mY/edit?usp=sharing).

On v3.1.2, I did not test polls open and polls pause without a USB drive and continuous export disabled, so it's possible that those still require a USB drive. But ultimately, it's polls close that's the most important operation to make sure is unblocked.

apps/scan/frontend/src/screens/poll_worker_screen.tsx
adghayes: addressed in follow-up commit

PR #5581, opened 2024-11-06
Remove Stray `assert` to allow programming sys admin cards when VxAdmin is unconfigured

### LINKED ISSUE #5578: Programming System Administrator Cards w/o Election Configured is Broken
Even though we allow programming SA cards when the machine is unconfigured (we did not allow that before) we have an `assert` on the frontend checking for an election definition. We need to fix that and make sure this is covered by testing.

adghayes:
Closes #5578.


jonahkagan (approved):
Nice catch!

Issue #5591, opened 2024-11-09
VxMark crashing on skewed ballot insert


arsalansufi:
https://votingworks.slack.com/archives/CJU9MSC6S/p1731098194439429

arsalansufi:
There may still be some crashing here, particularly on jam right at the time of insert. There's a separate #machine-feedback Slack thread about that. Will use that during the next issue triage session to create a v4.1 ticket

But ultimately, we're not too concerned about this on actual hardware, using the infeed to guide paper into the slot

Issue #5590, opened 2024-11-09
VxMark sometimes hanging when loading ballot sheet


arsalansufi:
https://votingworks.slack.com/archives/CJU9MSC6S/p1731097545442529

Issue #5589, opened 2024-11-09
Crashes around pre-printed ballot insertion flows


arsalansufi:
https://votingworks.slack.com/archives/CJU9MSC6S/p1731095966851349

PR #5599, opened 2024-11-11
Address various VxMark hangs/crashes

arsalansufi:

## Overview

Issue links:
1. https://github.com/votingworks/vxsuite/issues/5586
2. https://github.com/votingworks/vxsuite/issues/5589
3. https://github.com/votingworks/vxsuite/issues/5590
4. https://github.com/votingworks/vxsuite/issues/5591

In our final QA, we've unearthed a variety of VxMark hangs/crashes. This PR addresses/alleviates those issues and makes the app generally more stable.

Note that there are corner cases around paper handling/jams where we still crash, but we do so quickly, and a restart always results in a recovery. This is fine behavior. The main thing that we want to avoid is hanging.

### Commit 1: Mutex issues

The first find is that we're using a deprecated, one-off mutex implementation. I wasn't able to dig deep into whether this implementation has any bugs, but I switched over to the improved/shared mutex implementation in `libs/utils`, which I know is well-tested and has had bugs in the past which have since been addressed, e.g., this one that @eventualbuddha and I worked on: https://github.com/votingworks/vxsuite/pull/4368.

I made some further tweaks to ensure that we release the mutex even in error cases, by using the `withLock` method, which uses a `try` `finally` under the hood. I did find at least one case where we're not releasing the mutex even in the success case, though this is in a method that seems to be unused.

### Commit 2: No more hanging after jam clearing

With lots of logging on real hardware, I identified the following sequence after a jam is cleared:

- Scan-reset command is issued and returns success, but as documented in code, the reset doesn't actually complete until after a delay
- We continue polling the paper handler status
- Those requests seem to succeed, until the scanner reset actually completes under the hood
- The last polling request, which had acquired the mutex, never completes, probably because scanner state was cleared. So the mutex is never released and the app hangs

The fix was to stop polling paper handler status while resetting the scanner after a jam is cleared, which required some minimal state machine rearrangement.

### Commit 3: Addressing crash involving preprinted ballot insertion edge case

Currently, if you insert a preprinted ballot, return it, and immediately try reinserting that preprinted ballot without returning to the initial poll worker screen, the app crashes.

This is because `context.acceptedPaperTypes` is cleared when you return the preprinted ballot but needs to be populated for paper loading and validation to work correctly.

The fix was to drop you off at the initial poll worker screen rather than the paper loading screen after returning your preprinted ballot, which does make for a better UX anyway. The initial poll worker screen always sets `context.acceptedPaperTypes` as needed.

## Testing Plan

- [x] Tons of manual testing across Frankenstein's VSAP and the East Bay VSAP, with QA images


libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -23,6 +23,7 @@ import {
   writeImageData,
 } from '@votingworks/image-utils';
 import { Rect } from '@votingworks/types';
+import { Mutex } from '@votingworks/utils';
```
arsalansufi, line 26: This commit is easiest reviewed ignoring whitespace: https://github.com/votingworks/vxsuite/pull/5599/commits/5976abda49a6df6fb492e94a4e1d245a5ac05240?w=1. See `Commit 1: Mutex issues` in the PR description for more details

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -280,18 +278,17 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
     // According to the manual, "REAL-TIME USB PROTOCOL FORMAT" supports transferring out optional data.
     // The prototype doesn't support it, so leave out support from this function until needed.
   ): Promise<Result<T, CoderError>> {
-    await this.realTimeLock.acquire();
+    return await this.realTimeLock.withLock(async () => {
```
arsalansufi, line 281: As noted in the PR description, using this `withLock` method is nice because it uses a `try` `finally` under the hood to make sure we release the mutex even in error cases.

Previously, the mutex would not be released if there was an error before the `.release()` call

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -463,12 +460,13 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
...
-    const { data } = transferInResult;
-    assert(data);
-    return parseScannerCapability(data);
+    return this.genericLock.withLock(async () => {
+      await this.transferOutGeneric(GetScannerCapabilityCommand, undefined);
+      const transferInResult = await this.transferInGeneric();
+      const { data } = transferInResult;
+      assert(data);
+      return parseScannerCapability(data);
+    });
```
arsalansufi, line 469: This is the block that was missing a `.release()` call altogether. Though this method appears to be unused

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1172,6 +1156,27 @@ export function buildMachine(
             },
           },
         },
+        // This state is intentionally placed outside of the voting_flow block because we don't
+        // want to poll the paper handler status while we're resetting the paper handler. Doing so
+        // can result in unspecified erroneous behavior. The voting_flow block consistently polls
+        // paper handler status by way of polling the cover open status.
```
arsalansufi, line 1165: See `Commit 2: No more hanging after jam clearing` in the PR description for more details

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -740,7 +740,10 @@ export function buildMachine(
                     'resetContext',
                     'endCardlessVoterAuth',
                   ],
-                  target: 'accepting_paper',
+                  // Return to the initial poll worker screen so that the poll worker has to
+                  // explicitly choose whether they want to start a new session with a blank sheet
+                  // or insert a preprinted ballot again
+                  target: 'not_accepting_paper',
```
arsalansufi, line 746: See `Commit 3: Addressing crash involving preprinted ballot insertion edge case` in the PR description for more details

adghayes (approved):
@arsalansufi looks good to me - your descriptions align exactly with the code and it seems like as minimal an intervention as possible. I don't have history on the state machine so can't say I fully understand the implications of commit 2, but we've both tested it and it seems to work well.

arsalansufi:
Merging but please feel free to add any comments after the fact @kshen0 and @kofi-q!

libs/custom-paper-handler/src/driver/driver.ts
kshen0: Thanks for upgrading this.

Issue #5594, opened 2024-11-09
VxCentralScan: 22" HMPBs sometimes getting cutoff and failing to interpret as a result


arsalansufi:
https://votingworks.slack.com/archives/CJU9MSC6S/p1731102639092259

PR #5601, opened 2024-11-11
VxCentralScan: Ensure that HMPB timing marks aren't clipped

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/5594

clipping

With 22" ballots on VxCentralScan, we've sporadically seen interpretation fail because timing marks are clipped, ~circled~ squared in red above.

To address this, we're adding some buffer to the page height param that we specify to the Fujitsu. Scanning extra is safe because the HMPB interpreter trims any extra scan area. In fact, we already rely on this trimming when scanning letter HMPBs, as we specify the BMDB height for that case (13.25 inches > 11 inches).

## Testing Plan

- [x] Manually tested with 22" HMPBs

### DIFF
--- a/apps/central-scan/backend/src/fujitsu_scanner.test.ts
+++ b/apps/central-scan/backend/src/fujitsu_scanner.test.ts
@@ -115,9 +115,9 @@ test('fujitsu scanner can scans with expected params on letter size election', (
     'scanimage',
     expect.arrayContaining([
       '--page-width',
-      '215.872',
+      '215.9',
       '--page-height',
-      '336.506',
+      '336.55',
       '--bgcolor=black',
     ])
   );
@@ -144,9 +144,9 @@ test('fujitsu scanner can scan with expected params on legal size election', () 
     'scanimage',
     expect.arrayContaining([
       '--page-width',
-      '215.872',
+      '215.9',
       '--page-height',
-      '355.554',
+      '361.95',
       '--bgcolor=black',
     ])
   );
--- a/apps/central-scan/backend/src/fujitsu_scanner.ts
+++ b/apps/central-scan/backend/src/fujitsu_scanner.ts
@@ -150,19 +150,29 @@ export class FujitsuScanner implements BatchScanner {
       args.push('--endorser-string', `${imprintIdPrefix}_%04ud`);
     }
 
-    const MM_PER_INCH = 25.3967;
+    /**
+     * We've occasionally seen the Fujitsu return images where HMPB timing marks are ever so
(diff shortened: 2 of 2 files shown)


apps/central-scan/backend/src/fujitsu_scanner.test.ts
```diff
@@ -115,9 +115,9 @@ test('fujitsu scanner can scans with expected params on letter size election', (
     'scanimage',
     expect.arrayContaining([
       '--page-width',
-      '215.872',
+      '215.9',
       '--page-height',
-      '336.506',
+      '336.55',
```
arsalansufi, line 120: We expectedly don't see a meaningful change in the height param for letter HMPBs since, as noted in the PR description, we specify the BMDB height for that case (13.25 inches > 11 inches)

apps/central-scan/backend/src/fujitsu_scanner.ts
```diff
@@ -150,19 +150,29 @@ export class FujitsuScanner implements BatchScanner {
...
     }
 
-    const MM_PER_INCH = 25.3967;
+    /**
+     * We've occasionally seen the Fujitsu return images where HMPB timing marks are ever so
+     * slightly clipped, so we add some buffer to the page height to ensure that we capture
+     * everything. Scanning extra is safe because the HMPB interpreter trims any extra scan area.
+     */
+    const HMPB_HEIGHT_BUFFER_INCHES = 0.25;
+    const MM_PER_INCH = 25.4;
```
arsalansufi, line 159: I'm not sure why we'd previously set the mm per inch conversion factor to 25.3967, but I did confirm that the conversion factor is in fact exactly 25.4. This difference is so small that changing it alone doesn't solve the larger clipping issue.

apps/central-scan/backend/src/fujitsu_scanner.ts
```diff
@@ -150,19 +150,29 @@ export class FujitsuScanner implements BatchScanner {
...
+    const { height: bmdbHeight } = ballotPaperDimensions(
       BmdBallotPaperSize.Vsap150Thermal
     );
     args.push(
       '--page-width',
       toMillimeters(width),
       '--page-height',
-      toMillimeters(Math.max(height, bmdThermalHeight))
+      toMillimeters(
+        Math.max(hmpbHeight + HMPB_HEIGHT_BUFFER_INCHES, bmdbHeight)
```
arsalansufi, line 174: BMDB QR codes are so far from the edges of the sheet, they don't need this buffer

adghayes (approved):

Issue #5597, opened 2024-11-10
Decrease log volume on VxMark


arsalansufi:
https://votingworks.slack.com/archives/CJU9MSC6S/p1731189519700209

PR #5603, opened 2024-11-11
Decrease log noise on VxMark

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/5597

This PR decreases log noise on VxMark by suppressing two highly-frequent low-utility events:

- `COVER_STATUS`: Just tells us that we polled cover status and does not include the cover status itself. The transitions from cover-closed to cover-open and vice versa are already covered elsewhere.
- `NO_PAPER_ANYWHERE`: Just tells us that we polled for paper and found none.

As a kinda funny stat, we're currently writing ~1MB of logs every 5 minutes on VxMark. Given the very limited storage capacity on VxMark (~5GB), it would take 17 days to fill up the entire capacity of VxMark with logs. Of course rebooting compresses the logs down, and I think they compress pretty well. But still, this PR should render that possibility moot.

## Testing Plan

- [x] Tested manually on Frankenstein's VSAP

### DIFF
--- a/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
+++ b/apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
@@ -1425,19 +1425,32 @@ export function buildMachine(
   );
 }
 
+const HIGHLY_FREQUENT_EVENT_TYPES: Array<PaperHandlerStatusEvent['type']> = [
+  'COVER_STATUS',
+  'NO_PAPER_ANYWHERE',
+  'PAT_DEVICE_NO_STATUS_CHANGE',
+];
+
+function isHighlyFrequentEvent(event: EventObject): boolean {
+  return HIGHLY_FREQUENT_EVENT_TYPES.includes(
+    event.type as PaperHandlerStatusEvent['type']
+  );
+}
+
 function setUpLogging(
   machineService: Interpreter<Context, any, PaperHandlerStatusEvent, any, any>,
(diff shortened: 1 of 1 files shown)


apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
```diff
@@ -1425,19 +1425,32 @@ export function buildMachine(
   );
 }
 
+const HIGHLY_FREQUENT_EVENT_TYPES: Array<PaperHandlerStatusEvent['type']> = [
+  'COVER_STATUS',
```
adghayes, line 1429: The other two make sense to me. Will excluding `COVER_STATUS` leave us without logs for cover open and close events, or are those covered elsewhere?

adghayes (approved):

apps/mark-scan/backend/src/custom-paper-handler/state_machine.ts
arsalansufi: Covered elsewhere! And in fact the `COVER_STATUS` log doesn't even include the open/close status. Per the PR description:

> `COVER_STATUS`: Just tells us that we polled cover status and does not include the cover status itself. The transitions from cover-closed to cover-open and vice versa are already covered elsewhere.

Look for the "Transitioned to" logs in Kofi's original PR: https://github.com/votingworks/vxsuite/pull/5360

PR #5607, opened 2024-11-14
Ignore an initial ack in custom printer driver scan command

carolinemodic:
## Overview
Ignores an initial "ack" if seen on the transfer in endpoint when scanning. We have sporadically seen this in weird situations and ignoring it seems harmless. Note that @eventualbuddha has seen issues with scanning the image after this in dev but we have not repro'd those testing with images. 

## Checklist

### DIFF
--- a/libs/custom-paper-handler/src/driver/driver.ts
+++ b/libs/custom-paper-handler/src/driver/driver.ts
@@ -525,6 +525,11 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
       while (scanStatus === OK_CONTINUE) {
         const rawResponse = await this.transferInGeneric();
         assert(rawResponse?.data);
+
+        // We sometimes find an unexpected ack in the USB buffer when we go to read the scan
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
Thank you both for investigating/testing! I'll pull this onto the release branch before building images

libs/custom-paper-handler/src/driver/driver.ts
```diff
@@ -527,6 +527,12 @@ export class PaperHandlerDriver implements PaperHandlerDriverInterface {
     while (scanStatus === OK_CONTINUE) {
       const rawResponse = await this.transferInGeneric();
       assert(rawResponse?.data);
+      const isAck = AcknowledgementResponse.decode(
```
arsalansufi, line 530: Might be worth adding a quick comment that we see an unexpected ack sometimes so are just ignoring it for now

arsalansufi: Per discussion in Slack, I can add this and merge on your behalf

PR #5611, opened 2024-11-14
[BMD] Add retries to pulse audio command for output selection

kofi-q:
## Overview

We have some signal that attempting to run the `pactl set-sink-port` command too early after boot-up (which happens if we start up with the printer cover open), the `pulse audio` service may not be running yet or fully spun up.

This adds two retries with delays to attempt to wait until `pulse audio` is ready to run the command.

## Testing Plan
- Patch was tested on a VSAP 150 that seemed to be running into the issue and verified that it at least prevents crashes on startup.
- Unit tests for the retry logic

## Checklist


arsalansufi (approved):
LGTM!

PR #5614, opened 2024-11-15
start contributors

benadida:
Add a contributors list and in particular include VotingWorks pioneers, the election administrators who piloted VotingWorks when it was just a prototype.


arsalansufi (approved):
Love it!

Issue #5612, opened 2024-11-15
Fix Confusing VxMark PAT Calibration Copy


adghayes:
In #5519, @product-team changed the copy in the VxMark calibration flow to be more confusing rather than less confusing due to a misunderstanding of how PAT devices work. See discussion: https://votingworks.slack.com/archives/CJU9MSC6S/p1731623193918399

The copy should be reverted to what it previously was or something similar.

PR #5519, opened 2024-10-14
V4 Copy Pass Pt. 4 (VxMark Focused)

adghayes:
## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #5619, opened 2024-11-18
[BMD] Revert PAT calibration page copy changes

kofi-q:
## Overview

Fixes https://github.com/votingworks/vxsuite/issues/5612

Reverting a few copy changes from https://github.com/votingworks/vxsuite/pull/5519 related to the PAT device calibration voter/diagnostic flows. There was a misunderstanding in that PR about whether or not this flow enables the voter to choose which control triggers which action. This re-aligns the copy with the current behaviour.

### TODO:
- Long-term, we may re-visit and update this flow to actually enable the user to map inputs to actions.
- Also noticed I had to press the controls twice in some cases for it to register as one press - it was only happening in the voter flow, but not in the sys admin diagnostic flow. I couldn't find an obvious root cause, so I'll keep digging and open a separate PR if I find anything.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/e542929e-bc7a-4e57-b82f-90e8c4637901

## Testing Plan
- Manual verification
- Updated unit tests

## Checklist


kshen0 (approved):
Thanks for taking this on!

PR #5620, opened 2024-11-18
Application-level tweaks for FIPS-compliant OpenSSL

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4567

Related PRs:
- https://github.com/votingworks/vxsuite-complete-system/pull/418
- Adam's vxsuite-build-system PR once up

What a roller coaster this has been 😅 Full investigation details in [this Slack thread](https://votingworks.slack.com/archives/C07QH5HUBH9/p1731458142552579) for future reference.

VVSG2 requires that cryptographic modules be FIPS-compliant. When using our TPM chips, we get FIPS compliance through hardware - our TPM chips are all FIPS-certified. But all other OpenSSL operations, like hash generation and signature verification, should also be FIPS-compliant. To accomplish this, we have to build and run OpenSSL in FIPS mode.

This at first seemed easy but quickly grew complex because enabling FIPS mode interfered with our TPM-related commands. Further research uncovered the following:

1. Our default OpenSSL config file should enable FIPS mode. For TPM commands, we need to one-off disable FIPS mode since FIPS mode locks you into using the FIPS provider. Our narrative here is solid. We're always either using FIPS-compliant software or FIPS-compliant hardware.
2. The trickiest command sequence to figure out has been the card programming command sequence. Namely because it requires both the TPM provider and the default provider in a single command.
3. One of my key findings last week was that for, both to be able to work, it wasn't enough to override `default_properties = fips=yes` with `default_properties = fips=no`; I needed to remove the `fips=` reference altogether. I found a nice way to do accomplish this override in a sub-config using `default_properties = ''`.
4. I ended up stuck in between a rock and a hard place with the two OpenSSL commands for cert creation. `openssl x509` doesn't support a `-config` override param, and in all my testing, the `OPENSSL_CONF` env var doesn't seem to affect it either. I tried looking at the source code and docs for more intel and couldn't find a clear mention of when that env var applies. So all I have to go off of right now is my empirical findings. `openssl ca` does support a `-config` override param. But it doesn't support the `-force_pubkey` param that `openssl x509` supports. This param is necessary for card cert creation because we don't have access to the card private keys nor is there an obvious API for getting the card private keys to sign a CSR. So we have to create a dummy CSR and overwrite the public key on cert creation with `-force_pubkey`. I tried forcing in the right public key during CSR creation rather than cert creation. But that ultimately didn't work because the resulting CSR ends up mismatched/unverifiable.
5. To pre-empt a comment from Brian :stuck_out_tongue_closed_eyes: I'm fully aware that this is yet another example of the limitations of shelling out to OpenSSL, where we're bound by the quirks of the CLI, instead of using OpenSSL bindings directly. I'm still very sold on that switch and want to make that change for the next release. I still think that's too big/risky an undertaking in the last week before TRR.
6. The above has left me with an option that we were trying to avoid - shuffling around the default OpenSSL config file. I do have a proof-of-concept working, with a method that should work on locked-down images, given judicious use of symlinks. To keep this as safe as possible, I'm using the `-config` param everywhere I can and using this alternate method in the one place I need it on VxAdmin, for card cert creation. I'm using a `try` `finally` to make sure that we always undo the change. And as an added layer of safety, I'm proactively resetting the config on VxAdmin boot in case we ever crash before the `finally` can run.

## Testing Plan

- [x] Tons of iteration and testing on real hardware, imaged with Adam's test image that includes relevant `vxsuite-build-system` changes
- [x] Updated automated tests and added one "integration-y" test

Before I merge, planning to run some final E2E tests since there are so many moving pieces here.


libs/auth/config/openssl.vx.cnf
```diff
@@ -0,0 +1,12 @@
+# In production, this include is replaced with `.include /etc/ssl/openssl.default.cnf` to avoid a
+# circular dependency. See src/intermediate-scripts/manage-openssl-config for more details.
+.include /etc/ssl/openssl.cnf
```
arsalansufi, line 3: Overall happy with how this piece turned out! No more copying the default OpenSSL config file into the repo and editing it.

This gives `vxsuite-build-system` the ability to define a default OpenSSL config file that's FIPS-compliant, while still allowing local dev without FIPS-compliant OpenSSL.

I also like that this clearly separates application-specific OpenSSL config file logic (like Vx OIDs) from the rest of the config file logic.

libs/auth/src/intermediate-scripts/manage-openssl-config
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+# Requires sudo
+
+# In production, we use the following symlink arrangement:
+#
+# /etc/ssl/openssl.cnf --> /vx/config/openssl.cnf -->
+#   EITHER /etc/ssl/openssl.default.cnf
+#   OR     /vx/code/vxsuite/libs/auth/config/openssl.vx-tpm.cnf
```
arsalansufi, line 9: This symlink arrangement is prepared in the `vxsuite-complete-system` `setup-machine.sh` script: https://github.com/votingworks/vxsuite-complete-system/pull/418

libs/auth/src/intermediate-scripts/manage-openssl-config
```diff
@@ -0,0 +1,48 @@
...
+      "${IS_INTEGRATION_TEST:-}" == "true" ||
+      "${IS_INTEGRATION_TEST:-}" == "1" ]]; then
+    echo "Skipping in non-production environment" >&2
+    exit 0
+fi
+
+if [[ "${1}" == "override-for-tpm-use" ]]; then
+    ln -fs /vx/code/vxsuite/libs/auth/config/openssl.vx-tpm.cnf /vx/config/openssl.cnf
+elif [[ "${1}" == "restore-default" ]]; then
+    ln -fs /etc/ssl/openssl.default.cnf /vx/config/openssl.cnf
```
arsalansufi, line 47: I'm taking some time in my testing to reason through what the permissions on `/vx/config/openssl.cnf`, whether this script actually needs sudo in prod, what the permissions will be if it is run with sudo, and whether any permission resetting needs to happen

Will update as needed after testing!

benadida (approved):
Lgtm!

I'm assuming the move away from await for calls to openSSL are okay.

One comment I would add: it's safe to do this symlinking through /vx/config because that writable partition is now TPM encrypted, and so can't be modified by an attacker easily.

arsalansufi:
> I'm assuming the move away from await for calls to openSSL are okay.

Yup, those changes are a no-op. The following two functions are equivalent; both are returning a promise. The internal await would only be necessary if there were multiple commands in the function. I made the change for consistency with other cases like this in `libs/auth`.

```
async function fn(): Promise<Data> {
  return await innerFn();
}
```

```
function fn(): Promise<Data> {
  return innerFn();
}
```

> One comment I would add: it's safe to do this symlinking through /vx/config because that writable partition is now TPM encrypted, and so can't be modified by an attacker easily.

Thanks for noting!

Issue #3796, opened 2023-08-08
Handle slight clock skew during smart card auth and validation of machine certs


arsalansufi:
@benadida shared the following observation in Slack:

> I've noticed a small issue that I think I've diagnosed: sometimes if I use a freshly programmed card, it's not valid on a separate machine.... and then 1-2 minutes later it is. I'm wondering if those cards are programmed "in the future" and thus aren't yet valid because of clock drift. If that's right, maybe we should program cards to be valid 5-10 minutes into the past, just to prevent this otherwise confusing error state.

arsalansufi:
Let's apply the same to machine cert creation, now that we're validating machine certs during the config wizard and this same issue has a tendency to come up

PR #5621, opened 2024-11-19
Handle slight clock divergences between machines during smart card auth

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3796

The following situation comes up fairly often:
- A smart card is programmed on VxAdmin.
- That card is immediately taken to VxScan, and a message appears indicating that the card is not yet valid (or in earlier versions of the code, the message just said that the card was invalid with no explanation).
- Wait a while, maybe a minute or two. And VxScan then accepts the card as valid.

The issue here is that the VxAdmin has a time slightly ahead of the VxScan, so the card cert start date is in the future when the card is first taken to the VxScan. Only once the VxScan clock catches up to the card cert start date is the card accepted.

I'd for so long been thinking that the solution was to mint certs with a cert start time a couple mins in the past. But this has been hard given the constraints of the OpenSSL CLI in the context of VxSuite and which cert-issuing commands actually support that ability. Just dawned on me last night that I could instead change the _verification_ command to account for this time mismatch! `openssl verify` has an `-attime` param that let's you pick a time other than the system time to assume for verification. 

## Testing Plan

- [x] Verified that automated tests still pass.
- [x] Patched my VxAdmin ThinkPad, programmed a card, adjusted the time back a couple of minutes, and confirmed that the card was still accepted. Adjusted the time back even further and confirmed that the card was rejected.

### DIFF
--- a/libs/auth/src/cryptography.ts
+++ b/libs/auth/src/cryptography.ts
@@ -187,6 +187,8 @@ export function extractPublicKeyFromCert(
   return openssl(['x509', '-noout', '-pubkey', '-in', cert]);
 }
 
+const VERIFICATION_TIME_MARGIN_MINUTES = 10;
+
 /**
  * Verifies that the first cert was signed by the second cert. Both certs should be in PEM format.
  * Throws an error if signature verification fails.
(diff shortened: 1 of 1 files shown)


libs/auth/src/cryptography.ts
```diff
@@ -187,6 +187,8 @@ export function extractPublicKeyFromCert(
   return openssl(['x509', '-noout', '-pubkey', '-in', cert]);
 }
 
+const VERIFICATION_TIME_MARGIN_MINUTES = 10;
```
arsalansufi, line 190: 10 mins of wiggle room felt about right to me. Any more and you probably do want a reminder to check your clocks.

benadida (approved):

PR #5622, opened 2024-11-19
fix(scan): improve acceptance rate of skewed ballots

eventualbuddha:
## Overview

When a ballot has a significant skew (>= 1 degree), the previous implementation of `find_scanned_document_inset` would cause too much of the ballot to be cropped out of the image. Because it would only stop cropping once the white pixels were half of a given row, the portion of a skewed ballot that would be cropped was likely to contain timing marks. At the very least, it was likely to partially or completely crop the corner timing mark, which we require in order to proceed with interpretation.

The new implementation takes a configurable ratio of white to black pixels at which to stop cropping, which we now set to 10% (vs the previous 50%). This makes the cropping much more conservative while still cropping off any edge rows/columns of pure black. This preserves the corner timing marks even with skew around 3 degrees, which is as high as we're likely to be able to interpret successfully.

## Demo Video or Screenshot
| Original | Before (missing top-left corner) | After (top-left corner present) |
|-|-|-|
|image|image|image|

## Testing Plan
Tested with the corpus @adghayes provided from TRR. Ignoring the ballots that were rejected for non-skew reasons (defaced timing marks or QR codes, mainly), the original results were like so:

- 968 accepted
- 61 rejected
- 6% rejection rate (30/500)

With the changes in this PR, the new results are:

- 1005 accepted (37 newly accepted)
- 24 rejected (0 newly rejected)
- 2.3% rejection rate (11.6/500)

Looking at the remaining ones that are failing, the pattern seems to be poor corner finding:

CleanShot 2024-11-19 at 10 27 07@2x

I have another change that may fix this, but it's a more significant change so I'll be doing it separately.

### DIFF
--- a/libs/ballot-interpreter/src/hmpb-rust/image_utils.rs
+++ b/libs/ballot-interpreter/src/hmpb-rust/image_utils.rs
@@ -233,7 +233,8 @@ pub fn expand_image(
 }
 
 /// Finds the inset of a scanned document in an image such that each side of the
-/// inset has more than half of its pixels above the given threshold.
+/// inset has more than `min_ratio_above_threshold` of its pixels above the
+/// given threshold.
 #[allow(clippy::similar_names)]
 pub fn find_scanned_document_inset(
     image: &GrayImage,
--- a/libs/ballot-interpreter/src/hmpb-rust/image_utils.rs
+++ b/libs/ballot-interpreter/src/hmpb-rust/image_utils.rs
@@ -235,7 +235,11 @@ pub fn expand_image(
 /// Finds the inset of a scanned document in an image such that each side of the
 /// inset has more than half of its pixels above the given threshold.
 #[allow(clippy::similar_names)]
-pub fn find_scanned_document_inset(image: &GrayImage, threshold: u8) -> Option<Inset> {
+pub fn find_scanned_document_inset(
+    image: &GrayImage,
+    threshold: u8,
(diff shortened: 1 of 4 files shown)


libs/ballot-interpreter/src/hmpb-rust/image_utils.rs
```diff
@@ -235,33 +235,37 @@ pub fn expand_image(
 /// Finds the inset of a scanned document in an image such that each side of the
 /// inset has more than half of its pixels above the given threshold.
```
jonahkagan, line 236: This comment could use an update, specifically "more than half of its pixels"

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -235,7 +235,7 @@ pub fn prepare_ballot_card_images(
 /// Return the image with the black border cropped off.
 pub fn crop_ballot_page_image_borders(mut image: GrayImage) -> Option<BallotImage> {
     let threshold = otsu_level(&image);
-    let border_inset = find_scanned_document_inset(&image, threshold)?;
+    let border_inset = find_scanned_document_inset(&image, threshold, 0.1)?;
```
jonahkagan, line 238: Maybe leave a comment explaining how `0.1` was chosen so we can easily remember

jonahkagan (approved):
Great find and fix!

PR #5628, opened 2024-11-20
style(scan): fix clippy lints

eventualbuddha:
Fix some lints pointed out by clippy @ Rust 1.81.0.

### DIFF
--- a/libs/ballot-interpreter/src/hmpb-rust/interpret.rs
+++ b/libs/ballot-interpreter/src/hmpb-rust/interpret.rs
@@ -728,7 +728,7 @@ mod test {
             serde_json::from_reader(BufReader::new(File::open(election_path).unwrap())).unwrap();
 
         let bubble_template = load_ballot_scan_bubble_image().unwrap();
-        let side_a_path = fixture_path.join(format!("blank-ballot-p{}.jpg", starting_page_number));
+        let side_a_path = fixture_path.join(format!("blank-ballot-p{starting_page_number}.jpg"));
         let side_b_path =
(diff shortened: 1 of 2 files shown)


libs/ballot-interpreter/src/hmpb-rust/timing_marks.rs
```diff
@@ -1604,7 +1604,10 @@ fn infer_missing_timing_marks_on_segment(
             );
 
         // if the closest timing mark is close enough, use it
-        if closest_rect.center().distance_to(&current_timing_mark_center) <= maximum_error
+        if closest_rect
+            .center()
+            .distance_to(&current_timing_mark_center)
+            <= maximum_error
```
eventualbuddha, line 1610: This was just reformating.

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -874,7 +874,7 @@ mod test {
     fn test_vertical_streaks() {
         let (mut side_a_image, mut side_b_image, options) =
             load_hmpb_fixture("general-election/letter", 1);
-        let thin_complete_streak_x = side_a_image.width() * 1 / 5;
+        let thin_complete_streak_x = side_a_image.width() / 5;
```
jonahkagan, line 877: I think the original was more readable, but if clippy insists I guess we do it

jonahkagan, line 877: I think the original was more readable, but if clippy insists I guess we do it

jonahkagan (approved):

PR #5632, opened 2024-11-21
Load final batch of vendored translations for v4

arsalansufi:
Issue link: https://github.com/votingworks/vxsuite/issues/5569

While only relevant to VxDesign and therefore outside the certified system, I still figured that it'd be good to get and load a final batch of vendored translations for v4. Last time we gathered translations was before UAT, and some copy has changed since then.


kofi-q (approved):
Oh, that was a quicker turnaround than I expected! ✅

PR #5637, opened 2024-11-26
chore(kiosk-browser): remove uses of non-trivial APIs

### LINKED ISSUE #4531: tech improvement: stop using `kiosk-browser` APIs
# Overview
Originally, `kiosk-browser` (i.e. our electron wrapper) was our only way of accessing many system APIs. Now we have backends for every app that do that for us. We want to stop using any `kiosk-browser` APIs because we want the option of using a more lightweight alternative to electron that simply renders web content. Even if we, for some reason, can't replace `kiosk-browser`, moving things from `kiosk-browser` to the backend aligns better with our architectural model.

Apps vary in how much they're using `kiosk-browser`. Separately, the dev dock uses a few `kiosk-browser` APIs. The removals, by app, are:

### Dev Dock

- [ ] Screenshots - uses `screenshot` and `writeFile` endpoints. We could simply disable this or provide an alternative.
- [ ] Fixture Selection - uses `showOpenDialog` to pick a file. Most browser replacements would offer this

### VxScan

No usage.

### VxCentralScan

- [x] Scanner Connection Status - needs to use polling through the backend

###  VxMarkScan

- [x] Battery Info  (#4533)
- [x] Speech Synthesis
- [x] Rendering ballot to PDF

###  VxMark

- [x] Battery Info  (#4533)
- [x] Accessible Controller Connection Status (VxMark)
- [x] Speech Synthesis
- [x] Rendering & printing ballot to PDF
- [x] Printer Status

### VxAdmin

- [x] Report Printing - uses `kiosk.print` and file system endpoints (#4532)
- [x] Printer Status (#4532)
- [ ] Custom Location Saving - uses `showSaveDialog` (we could also just axe this)
- [ ] Custom Select File - uses `showOpenDialog`

### Miscellaneous

A drop-in replacement for `kiosk-browser` would also have to handle:
- [ ] logging to console, unless all logging is moved to backend
- [ ] `quit`? A shortcut to close the frontend in VxDev

We will gradually move away from these endpoints as our work touches them. In the meantime, the path is cleared on VxScan (and pretty much on VxCentralScan too) to begin experimenting with a replacement. The replacement would have to meet other production image and security requirements that are not listed here, but it's likely that overall a slimmer replacement would be easier to work with than Electron.

eventualbuddha:
## Overview

Refs #4531

Removes all uses of the following `kiosk-browser` APIs:
- ~`captureScreenshot` + `saveAs` (replaced with `dev-dock/backend` API + `gnome-screenshot`)~ this was restored because we need a way to capture the web rendering area and not the entire screen
- ~`showOpenDialog` (selecting election package to load in dev is removed)~ this was restored because we still need to pick election definitions
- `showSaveDialog` ("Save As" in a few places is removed)
- `saveAs` ("Save As" in a few places is removed, and writing the screenshot is now done on the backend)

`kiosk.quit` and `kiosk.log` remain unchanged.

## Demo Video or Screenshot

**Edit 11/27:** The video below shows the screenshot flow using a `wry`-based version of `kiosk-browser`. This no longer works with the reversion to using `kiosk.captureScreenshot`, but the flow outlined in the video is the same.

https://github.com/user-attachments/assets/b84b2c8c-6379-4ad4-bf5d-b9c883ef6a08

<details>
<summary>Click to view `wry` based app code</summary>

```rust
use clap::Parser;
use serde::Deserialize;
use tao::{
    event::{ElementState, Event, StartCause, WindowEvent},
    event_loop::{ControlFlow, EventLoop},
    keyboard::ModifiersState,
    window::{Fullscreen, WindowBuilder},
};
use wry::WebViewBuilderExtUnix;
use wry::{http::Uri, WebViewBuilder};

#[derive(Debug, Parser)]
struct Options {
    #[clap(short, long, default_value = "false")]
    allow_devtools: bool,

    #[clap(default_value = "http://localhost:3000/")]
    url: Uri,
}

#[derive(Debug, Deserialize)]
#[serde(tag = "type")]
enum IpcCommand {
    /// Log a message to stdout.
    #[serde(rename = "log")]
    Log { message: String },

    /// Quit the application.
    #[serde(rename = "quit")]
    Quit,
}

const INIT_JS: &str = "
    globalThis.kiosk = Object.freeze({
        quit() {
            window.ipc.postMessage(JSON.stringify({ type: 'quit' }));
        },

        log(message) {
            window.ipc.postMessage(JSON.stringify({ type: 'log', message }));
        },
    });
";

fn main() -> wry::Result<()> {
    let options = Options::parse();
    let event_loop = EventLoop::new();
    let window = WindowBuilder::new()
        .with_fullscreen(Some(Fullscreen::Borderless(None)))
        .with_title("kiosk-browser")
        .build(&event_loop)
        .unwrap();
    let builder = WebViewBuilder::new()
        .with_url(options.url.to_string())
        .with_initialization_script(INIT_JS)
        // allow devtools if the flag is set
        .with_devtools(options.allow_devtools)
        // deny clipboard access
        .with_clipboard(false)
        // deny downloads
        .with_download_started_handler(|_, _| false)
        // do not record browsing history, cookies, cache, etc.
        .with_incognito(true)
        // handle IPC messages, i.e. `kiosk.log` et al
        .with_ipc_handler(
            |ipc| match serde_json::from_str::<IpcCommand>(&ipc.body()) {
                Ok(IpcCommand::Log { message }) => println!("{message}"),
                Ok(IpcCommand::Quit) => std::process::exit(0),
                Err(err) => {
                    eprintln!(
                        "failed to parse IPC message: {body}\nError: {err}",
                        body = ipc.body()
                    );
                }
            },
        );

    let webview = {
        use tao::platform::unix::WindowExtUnix;
        let vbox = window.default_vbox().unwrap();
        builder.build_gtk(vbox)?
    };

    let mut modifiers = Default::default();
    event_loop.run(move |event, _, control_flow| {
        *control_flow = ControlFlow::Wait;

        match event {
            Event::NewEvents(StartCause::Init) => println!("kiosk-browser has started!"),
            Event::WindowEvent {
                event: WindowEvent::ModifiersChanged(new_modifiers),
                ..
            } => {
                modifiers = new_modifiers;
            }
            Event::WindowEvent {
                event: WindowEvent::KeyboardInput { event, .. },
                ..
            } => match (modifiers, event.state, event.text) {
                (ModifiersState::CONTROL, ElementState::Released, Some("w")) => {
                    println!("Ctrl+W pressed, closing the window");
                    *control_flow = ControlFlow::Exit;
                }
                (ModifiersState::CONTROL, ElementState::Released, Some("r")) => {
                    println!("Ctrl+R pressed, reloading the page");
                    webview.evaluate_script("location.reload()").unwrap();
                }
                (
                    ModifiersState::CONTROL | ModifiersState::SHIFT,
                    ElementState::Released,
                    Some("i"),
                ) => {
                    if options.allow_devtools {
                        println!("Ctrl+Shift+I pressed, toggling devtools");
                        if webview.is_devtools_open() {
                            webview.close_devtools();
                        } else {
                            webview.open_devtools();
                        }
                    }
                }
                _ => {
                    // ignore
                }
            },
            Event::WindowEvent {
                event: WindowEvent::CloseRequested,
                ..
            } => *control_flow = ControlFlow::Exit,
            _ => (),
        }
    });
}
```
</details>

## Testing Plan
Tested in the original `kiosk-browser` and the new prototype `wry`-based `kiosk-browser` (see above). Standard browser usage remains unchanged since it doesn't have these APIs anyway.


libs/dev-dock/backend/src/dev_dock_api.ts
```diff
@@ -168,6 +169,14 @@ function buildApi(devDockFilePath: string, machineType: MachineType) {
       usbHandler.clearData();
     },
 
+    async captureScreenshot({ appName }: { appName: string }): Promise<string> {
+      assert(/^[a-z0-9]+$/i.test(appName));
+      const downloadsPath = join(homedir(), 'Downloads');
+      const fileName = `Screenshot-${appName}-${new Date().toISOString()}.png`;
+      await execFile('gnome-screenshot', ['-f', join(downloadsPath, fileName)]);
+      return fileName;
+    },
```
jonahkagan, line 178: One of the nice features about the previous screenshot functionality was that it captured the browser window only when using device emulation, which is important for screenshotting UI since you want to have the proper screen dimensions. I'm assuming `gnome-screenshot` takes a full-screen picture?

libs/dev-dock/frontend/src/dev_dock.tsx
```diff
@@ -83,22 +82,11 @@ function ElectionControl(): JSX.Element | null {
...
-        properties: ['openFile'],
-      });
-      if (dialogResult.canceled) return;
-      const selectedPath = dialogResult.filePaths[0];
-      if (selectedPath) {
-        setElectionMutation.mutate({ path: selectedPath });
-      }
-    } else {
-      setElectionMutation.mutate({ path });
-    }
```
jonahkagan, line 101: This feature is currently quite important in dev dock if you want to use mock smart cards with any election other than the fixtures. I don't think we can get rid of it without an alternative.

jonahkagan (changes requested):
I called out two important dev dock features removed in this PR that I think we need to maintain or find another way to implement. Happy to help research/brainstorm alternatives

libs/dev-dock/frontend/src/dev_dock.tsx
eventualbuddha: Added this feature back using the same approach, so we still need `kiosk.showOpenDialog`.

libs/dev-dock/backend/src/dev_dock_api.ts
eventualbuddha: Added this feature back using the same approach, so we still require `kiosk.captureScreenshot`. The only difference is that it's the backend that writes the file rather than the frontend. This lets us remove `kiosk.saveAs`.

jonahkagan (approved):

PR #5639, opened 2024-11-26
Fix bug in contest layout for HMPBs

jonahkagan:
## Overview

When laying out multiple contest sections on the same page, the code calculates whether there is enough vertical space left before adding a new section. That calculation forgot to include the gap between sections. This caused it to believe there was more space available than there actually was. Occasionally, this would result in adding more contests than would fit on the page, causing the page content to grow a bit longer and eat into the bottom margin of the page.

## Demo Video or Screenshot

Before - the bottom row of timing marks overflows into the bottom page margin
Screenshot 2024-11-26 at 2 41 59 PM

After - Proposition 1 is correctly bumped to the next page and the bottom timing marks stay in bounds
Screenshot 2024-11-26 at 2 42 21 PM

## Testing Plan
Manual test with the election that caused the bug. Our HMPB fixtures don't cover this case.

## Checklist

### DIFF
--- a/libs/hmpb/src/vx_default_ballot_template.tsx
+++ b/libs/hmpb/src/vx_default_ballot_template.tsx
@@ -788,8 +788,10 @@ async function BallotPageContent(
   const pageSections: JSX.Element[] = [];
   let heightUsed = 0;
 
-  // TODO is there a better way to incorporate gutter width here?
-  const gutterWidthPx = 0.75 * 16; // Assuming 16px per 1rem
+  // TODO is there some way we can use rem here instead of having to know the
+  // font size and map to px?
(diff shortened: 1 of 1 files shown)


libs/hmpb/src/vx_default_ballot_template.tsx
```diff
@@ -841,11 +841,12 @@ async function BallotPageContent(
       break;
     }
 
-    heightUsed += height;
+    // Add vertical gap to account for space between sections
+    heightUsed += height + verticalGapPx;
```
jonahkagan, line 845: This is the bug fix. The rest is just renaming for clarity

kofi-q (approved):
✅

PR #5641, opened 2024-12-03
Replace `kiosk-browser` in development with system webview wrapper app

eventualbuddha:
## Overview

Adds a native Linux GTK app using the Rust bindings for GTK that wraps a webview, providing a binary compatible with `kiosk-browser`. It provides the same `window.kiosk` API, and respects the same environment variables for configuration. This does _not_ change what we're using in production yet. That will happen in a follow-up change.

This is part of the effort to minimize and replace `kiosk-browser` the electron app with something else that uses the system webview. This allows us to stop building Electron/Chromium as part of trusted build, which has been a big source of pain.

The only `kiosk-browser` APIs that vxsuite still uses are:
- `log`
- `quit`
- `captureScreenshot`
- `showOpenDialog`

These are all implemented in terms of `webkit2gtk`.

In comparison to the Electron-based `kiosk-browser`, this one:
- has a fullscreen mode (default in prod) and a windowed mode (default in dev)
- sets the webview to the correct size for whichever app you're developing
- uses Webkit (Safari) instead of Blink (Chrome) for the rendering engine
- is 1.4MB vs 132MB for Electron

## Demo Video or Screenshot

The `run-dev` script (used by apps' `pnpm start`) now runs the new browser with the right resolution:

https://github.com/user-attachments/assets/bf0fc7b5-74fb-4493-9b3d-d56632282244

## Testing Plan

- [x] sanity check `cargo run --bin browser`
  - [x] runs fullscreen by default
  - [x] `--dev` runs in windowed mode
  - [x] `--device mark` works
  - [x] `--device scan` works
  - [x] `--device admin` works
  - [x] `--device central-scan` works
- [x] `pnpm start` runs with the right device resolution for each app
- [x] dev-dock screenshots work
- [x] disallows devtools by default
- [x] no obvious visual differences (was not extensively tested)


libs/browser/src/keybindings.rs
```diff
@@ -0,0 +1,45 @@
...
+                }
+                gdk::keys::constants::r if modifiers.contains(gdk::ModifierType::CONTROL_MASK) => {
+                    device.reload();
+                    gtk::glib::Propagation::Stop
+                }
+                gdk::keys::constants::Escape => {
+                    if options.is_dev() {
+                        if let Err(e) = device.unfullscreen() {
+                            eprintln!("Error unfullscreening: {e}");
+                        }
```
jonahkagan, line 27: Maybe crash as well as log since it's hard to see this log if you're in fullscreen

libs/browser/src/ui/header.rs
```diff
@@ -0,0 +1,114 @@
...
+    //     let device = device.clone();
+    //     move |index| {
+    //         config_selection.update_by_index(index);
+    //         let config = &DEVICE_CONFIGS[index];
+    //         device_button_label.set_text(config.name);
+    //         if let Err(e) = device.set_configuration(config) {
+    //             eprintln!("Failed to set device configuration: {e}");
+    //         }
+    //     }
+    // };
```
jonahkagan, line 55: Delete?

libs/browser/src/ui/header.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let popup_menu = gtk::Menu::new();
+
+    for (index, device) in config_selection.values().iter().enumerate() {
+        let menu_item = gtk::RadioMenuItem::builder()
+            .label(device.name)
+            .active(*device == config_selection.current())
+            .build();
+        menu_item.connect_activate({
+            // let device_button_label = device_button_label.clone();
```
jonahkagan, line 65: delete

libs/browser/src/device.rs
```diff
@@ -0,0 +1,222 @@
...
+};
+
+use gtk::prelude::*;
+use webkit2gtk::{WebInspectorExt, WebView, WebViewExt};
+
+/// Configuration of a VX device.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Configuration {
+    pub name: &'static str,
+    pub dimensions: (i32, i32),
```
jonahkagan, line 16: I also have set in my current device emulation configs two other settings: device pixel ratio and user agent. Are either of those relevant here?
Screenshot 2024-12-03 at 3 15 03 PM

libs/browser/src/device.rs
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    /// Resizes the window according to the current configuration.
+    pub fn reset_size(&self) -> anyhow::Result<()> {
+        let (width, height) = self.configuration()?.dimensions;
+        // reset minimum size
+        self.window.resize(1, 1);
+        // set the size we want for the webview; this will resize the window
+        self.webview.set_size_request(width, height);
```
jonahkagan, line 150: What happens if the monitor isn't big enough for the webview? Chrome's device emulation gives zoom options to deal with this

jonahkagan (approved):
Looks great!

libs/browser/src/ipc.rs
```diff
@@ -0,0 +1,263 @@
...
+                    }
+                });
+            }
+        }
+    });
+
+    // Add an object to the JavaScript global object that we can use to send IPC
+    // messages. Registers after setting up the IPC message handler per the
+    // documentation.
+    manager.register_script_message_handler("ipc");
```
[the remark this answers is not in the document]
eventualbuddha: https://webkitgtk.org/reference/webkit2gtk/2.25.2/WebKitUserContentManager.html#webkit-user-content-manager-register-script-message-handler

libs/browser/src/ipc.rs
```diff
@@ -0,0 +1,263 @@
...
+        .user_content_manager()
+        .context("Failed to get user content manager")?;
+    manager.connect_script_message_received(None, {
+        let window = window.clone();
+        let webview = webview.clone();
+        move |_, message| {
+            if let Some(value) = message.js_value() {
+                gtk::glib::spawn_future_local({
+                    let window = window.clone();
+                    let webview = webview.clone();
```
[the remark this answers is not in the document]
eventualbuddha: Yeah, it's the `move`s and the fact that the `connect_script_message_received` callback can be called multiple times. Each time it's called, we need to do these inner clones to pass off to the `async` block. The clone is, as you guessed, just copying a pointer around. It may be using something similar to an `Rc` or `Arc` where it's doing reference counting, but I'm not sure.

libs/browser/kiosk.js
```diff
@@ -0,0 +1,78 @@
...
+          resolve(data);
+        }
+      };
+
+      window.addEventListener('message', listener);
+      postMessage(message);
+    });
+  }
+
+  window.kiosk = {
```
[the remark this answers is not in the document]
eventualbuddha: b761273587c84703b975012c707a18a1bba7c2e7

libs/browser/src/device.rs
eventualbuddha: I'm not sure exactly what the device pixel ratio does. Is it just changing the scaling factor of the webview? I'm not sure how to do that, but we may be able to add something to that effect. We don't actually care about user agent, right?

libs/browser/src/device.rs
eventualbuddha: I believe it just makes the window large enough that it doesn't fit within the display. I have it set up so that Parallels just draws the VM monitor at a high resolution but scaled down.

libs/browser/src/device.rs
jonahkagan: I'm also not sure about what device pixel ratio does. I feel like @kofi-q might know.

I don't know that user agent matters. In Chrome, it will change the cursor to a touch circle, which is minorly useful in making sure touch targets are big enough and remember that it's a touch screen app. It may also change how clicks are translated into browser events. Neither of those seem like things we rely on heavily, but I might be missing something.

libs/browser/src/device.rs
jonahkagan: Interesting, I guess I'll have to try it and see how it looks on my laptop screen

libs/browser/src/device.rs
eventualbuddha: Ah, good points. @kofi-q, mind taking a little time to try this out? I'm not sure what is actually important vs a nice to have for UI development in kiosk-browser.

arsalansufi:
Thank you for investigating @eventualbuddha! Obviously not the outcome we were hoping for, but I don't think that we could've reached this conclusion without actually trying out this alternative. Mind summarizing or pointing me to the findings that resulted in the conclusion not to proceed with this switch? The misc findings in https://votingworks.slack.com/archives/C082ALXQBEZ/p1733336582320649 I think?

PR #5684, opened 2024-12-04
scan/backend: Refactor PDI scanner client creation

jonahkagan:
## Overview

Instead of passing a function to create a PDI scanner client to the state machine, create the client in the server.ts module and pass it into the state machine. In previous scanner integrations, we wanted to create new clients after failures. However, in this integration, we don't do that and instead just crash the app (part of our "fail fast" approach), so we don't ever need to recreate the client.

This refactor also prepares for introducing a mock PDI scanner client, which will also be a singleton. Since it will be referenced by the dev dock as well, it will need to be created in server.ts.

## Demo Video or Screenshot
N/A
## Testing Plan
Existing automated tests
## Checklist

### DIFF
--- a/apps/scan/backend/src/scanners/pdi/state_machine.ts
+++ b/apps/scan/backend/src/scanners/pdi/state_machine.ts
@@ -219,20 +219,19 @@ export const delays = {
 } satisfies Delays;
 
 function buildMachine({
-  createScannerClient,
+  scannerClient,
   workspace,
   usbDrive,
   auth,
   logger,
 }: {
-  createScannerClient: () => ScannerClient;
+  scannerClient: ScannerClient;
   workspace: Workspace;
   usbDrive: UsbDrive;
   auth: InsertedSmartCardAuthApi;
   logger: Logger;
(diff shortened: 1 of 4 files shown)


apps/scan/backend/src/scanners/pdi/state_machine.ts
```diff
@@ -242,10 +240,10 @@ function buildMachine({
 
   function createPollingChildMachine(
     id: string,
-    queryFn: (context: Pick<Context, 'client'>) => Promise<Event>,
+    queryFn: () => Promise<Event>,
```
carolinemodic, line 243: why remove the type here?

carolinemodic (approved):

apps/scan/backend/src/scanners/pdi/state_machine.ts
jonahkagan: The query function has access to the singleton scanner client in the outer function scope, so it doesn't need to get it from the context anymore

PR #5690, opened 2024-12-05
test(basics): convert tests from `jest` to `vitest`

### LINKED ISSUE #5707: libs/basics

eventualbuddha:
Closes #5707

### DIFF
--- a/libs/basics/src/assert.test.ts
+++ b/libs/basics/src/assert.test.ts
@@ -1,3 +1,4 @@
+import { expect, test } from 'vitest';
 import {
   assert,
   assertDefined,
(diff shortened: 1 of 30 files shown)


libs/basics/src/assert.test.ts
```diff
@@ -90,4 +91,5 @@ test('assertFalsy', () => {
   expect(() => assertFalsy(truthyValue)).toThrowError(
     `Unexpected truthy value: ${truthyValue}`
   );
+  expect(() => assertFalsy(0)).not.toThrowError();
```
eventualbuddha, line 94: This was flagged as a case that wasn't covered. Different `istanbul` version from what `jest` had, I guess?

kofi-q (approved):
🚀

PR #5691, opened 2024-12-05
Update all-bubble ballot generation to avoid bubble squishing and include cycle index for cycling test deck

arsalansufi:
## Overview

In prepping ballots for cert TRR, we identified an issue with the all-bubble ballot, namely that the bubbles on the front pages were squished such that they diverged from integer points on the timing mark grid. This likely happened when we added footer metadata below the tinted rectangles. This squishing resulted in other quirks, like the cycling pattern itself breaking.

<table><tbody><tr><td>
close-up
</td></tr></tbody></table>

To address this, we chose to move the footer metadata (and only the ballot hash in particular) into the tinted rectangle, just for the all-bubble ballot. While we were at it, we also included a cycle index value in the footer metadata for the cycling test deck to facilitate sheet management during the accuracy test.

| Before | After |
| - | - |
| before | after |

## Testing Plan

- [x] Automated tests
- [x] Manual e2e testing of the newly generated all-bubble election package and ballot


apps/design/backend/src/test_decks.test.ts
```diff
@@ -104,7 +108,9 @@ describe('createPrecinctTestDeck', () => {
       precinctId,
       ballots,
     });
-    await expect(testDeckDocument).toMatchPdfSnapshot();
+    await expect(testDeckDocument).toMatchPdfSnapshot({
+      failureThreshold: PDF_SNAPSHOT_DIFF_THRESHOLD,
+    });
```
arsalansufi, line 113: Had to add some wiggle room here as was already the case for the report snapshot comparison below. Will post the diff images in a moment to confirm that nothing of substance changed (I'm a bit surprised that anything changed at all)

libs/hmpb/jest.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   ...shared,
   coverageThreshold: {
     global: {
-      branches: -16,
+      branches: -18,
```
arsalansufi, line 10: The generated fixtures currently don't cover the case that we include customer footer metadata _without_ condensing the metadata. I did try this out manually, and it worked just fine.

libs/hmpb/src/all_bubble_ballot_fixtures.tsx
```diff
@@ -176,6 +180,8 @@ function BallotPageFrame({
             precinctId={election.precincts[0].id}
             pageNumber={pageNumber}
             totalPages={totalPages}
+            condenseFooterMetadata
+            includeCustomFooterMetadataSlot
```
jonahkagan, line 184: I'm a little hesitant to add these custom props to the footer element. It feels a little overcomplicated.

Two ideas for other ways to approach this:
1. If we want the all bubble ballot to have the same footer as the default ballot template, then we might consider having one less row of bubbles in the all bubble ballot template. Did you look into removing a row of bubbles (and updating the election definition accordingly)?
2. Since there's this new cycle index field that you added, maybe the all bubble ballot could have its own footer component that meets its specific needs (QR code, ballot hash, and cycle index).

libs/hmpb/src/all_bubble_ballot_fixtures.tsx
```diff
@@ -332,6 +338,13 @@ export const allBubbleBallotFixtures = (() => {
               blankBallot,
               votesForSheet
             );
+            await injectCustomFooterMetadata(
+              sheetDocument,
+              <FooterMetadataContent
+                label="Cycle Index"
+                value={sheetIndex + 1}
```
jonahkagan, line 345: Unless I'm misreading something, cycle index should be computable from `pageNumber`, which the footer has as a prop. Something like: `Math.floor(pageNumber / 2) + 1`. That means it doesn't need to be injected after the initial render, but rather could be computed in the footer component.

arsalansufi: Ah yes, I should've added a note about this as it's not intuitive. But the all-bubble ballot is still just a 1-sheet (2-page) ballot. The cycling test deck just includes 6 sheets of this 1-sheet ballot. So the page number will always be 1 or 2.

| Cycling test deck sheet | Page number |
| - | - |
| 1 | 1 |
| | 2|
| 2 | 1 |
| | 2|
| 3 | 1 |
| | 2|
| 4 | 1 |
| | 2|
| 5 | 1 |
| | 2|
| 6 | 1 |
| | 2|

So the "Cycle Index" has to be injected at the test-deck-generation level as opposed to the individual ballot-generation level.

libs/hmpb/src/all_bubble_ballot_fixtures.tsx
arsalansufi: Agreed that current params result in an abstraction that doesn't quite make sense. I like the idea of a unique footer component for the all-bubble ballot!

I did consider removing a row of bubbles but was looking for a quick solution last week and rearranging the footer proved to be quite simple. Happy to look into that now for a more permanent fixture/solution

libs/hmpb/src/all_bubble_ballot_fixtures.tsx
arsalansufi: One more reason for the injection was to only add the cycle index for the cycling test deck but not the blank all-bubble ballot. Though there's an argument to be made that we should always have the cycle index field there and just set it to, say, N/A when rendering the blank or filled all-bubble ballot.

jonahkagan: Ah right, ok I just read the code more fully and refreshed myself. Thanks for the explanation.

I remember previously just using the first marked bubble in the top row as my indicator of cycle index. But a printed number is a bit easier to use.

I think injecting it is fine given the requirements.

arsalansufi: > I remember previously just using the first marked bubble in the top row as my indicator of cycle index.

Yeah I've done the same. I don't have a great read on how exactly SLI manages and scans their all-bubble ballots for the accuracy test, but the general sense I got from Matt is that this would be helpful and make testing easier for SLI

arsalansufi:
Incorporated into https://github.com/votingworks/vxsuite/pull/6015 so closing this out

PR #5692, opened 2024-12-05
feat(test-utils): work with either `jest` or `vitest`

eventualbuddha:
## Overview
Follow-up to https://github.com/votingworks/vxsuite/pull/5690.

Changes `@votingworks/test-utils` to look for either a `jest` global or import from `vitest`, mostly for creating mock functions (i.e. `fn`). For simplicity, keeps the `jest` API for mock functions because they're so similar (but not actually TS-compatible). The possibility of `import`ing whenever we call `fn` within `test-utils` means that anything that creates a mock function now needs to be async. This isn't awesome and it caused a bunch of changes in other packages, but I think it's worth it to allow the `jest` → `vitest` migration to occur in pieces.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests.


libs/test-utils/src/console.test.ts
```diff
@@ -1,41 +1,31 @@
...
   'suppressingConsoleOutput replaces console.%s with a mock',
-  (method) => {
-    suppressingConsoleOutput(() => {
+  async (method) => {
+    const original = console[method];
+    await suppressingConsoleOutput(() => {
       console[method]('test');
       expect(console[method]).toHaveBeenNthCalledWith(1, 'test');
     });
-    expect('mock' in console[method]).toEqual(false);
```
eventualbuddha, line 11: This may still be true with `vitest`, but felt like testing an implementation detail. Replaced it with what we actually care about.

libs/test-utils/src/console.test.ts
```diff
@@ -1,41 +1,31 @@
...
 });
 
 test('suppressingConsoleOutput resolves to the value returned by the callback', async () => {
   await expect(
     suppressingConsoleOutput(async () => await Promise.resolve('test'))
   ).resolves.toEqual('test');
-  expect('mock' in console.log).toEqual(false);
-});
-
-test('suppressingConsoleOutput throws the error thrown by the callback', () => {
```
eventualbuddha, line 27: Now that `suppressingConsoleOutput` always returns a `Promise`, this test is redundant.

libs/test-utils/src/console.ts
```diff
@@ -127,28 +130,29 @@ export function suppressingConsoleOutput<T>(
...
     errorStats.count += 1;
   });
 
   function cleanup() {
-    jest.spyOn(console, 'log').mockRestore();
-    jest.spyOn(console, 'warn').mockRestore();
-    jest.spyOn(console, 'error').mockRestore();
+    logSpy.mockRestore();
+    warnSpy.mockRestore();
+    errorSpy.mockRestore();
```
eventualbuddha, line 147: Seeming behavior difference: `jest` uses the original spy with this second `spyOn` call, then restores it. `vitest` seems to replace the first spy with a second spy, then immediately restores back to the first spy.

eventualbuddha:
Closing in favor of https://github.com/votingworks/vxsuite/pull/5703.

PR #5695, opened 2024-12-05
Election Packages in Fixtures - Proof of Concept 

carolinemodic:
## Overview

Note: This is a draft PR meant to share my progress and current plan of action
to gather feedback and give others a change to interject if they question the
direction I'm heading this code change itself is not intended for formal review
or merging.

Recommend perusing by commit. The commit that adds the script to generate
election packages is the closest to being ready to review (but still not ready
for detailed line by line feedback).

Creates a script that generates election packages to be used in fixtures from
base election json files. Commit `f019195` shows how existing code can import
the new election package instead of an election.json file, I will likely update
res-to-ts or create other helpers to make this easier before migrating
everything. There is a test added in libs/fixture-generators that will fail if
the election package needs to be regenerated. Commit `f4b863e` is not intended
for merging at any point and illustrates how changing the electionBase will
cause the new test to fail. Commit `3e88969` regenerates fixtures after commit
`f4b863e` which allows the new test to pass again.

I have not updated libs/hmpb yet to stop regenerating this same election.json
with grid layouts, that is a planned future step (see below).

### Plan of future PRs:

I have been thinking about my next steps in terms of how to break things up in
future PRs, some including work you see here others not.

- Move apps/design/backend/src/language_and_audio to its own library - Open for
  feedback on the best way to do this, if there is a better location then
  creating a new library but I do not like that I am importing
  apps/design/backend from libs/fixture-generators and would like to move this
  to a shared location so I do not need to do that.
- Refactor the way I am changing the language_and_audio related code in this PR
  to be more sensible / have better names for things / not copy and paste code
  like the class methods to actually talk to google around / etc. to set up and
  give the helpers needed for this generate election package script.
- Add the election package generator script and associated tests/logic,
  including the initial example of electionPrimaryPrecinctSplits without fully
  fixing all uses of that fixture to move to the election package. Add helpers
  or logic in res-to-ts to make working with these election packages more
  straightforward. Make sure I'm deleting and cleaning up old files when
  regenerating.
- Fix all references to electionPrimaryPrecinctSplits to use the new files and
  not the deprecated baseElection. Update generate-fixtures in libs/hmpb to only
  generate ballot images and just use the generated election.json file produced
  by this script.
- For each other election fixture:
  - Add to the generation script / test and migrate all usages.
- Update dev docks knowledge of election files to use the generated
  election.json.
- Experiment with a way to load the election package in the dev dock or on a usb
  when configuring VxAdmin

### Questions I'm thinking about:

- Should I move where the ballot images generated by generate-fixtures in
  lib/hmpb to be in lib/fixtures? I think this might make sense in the long term
  but I am trying to keep it out of scope for now.
- Should the hashes be in the generated election-package and election.json
  filenames? This means you have to manually update the fixtures index.ts file
  if this changes which is a little annoying, but people should be aware when
  they are causing it to change. Maybe we could expose it as a variable more
  easily so that tests currently referencing these hardcoded values can at least
  be fixed and make this the ONLY place you need to update. I probably will want
  to keep these files in a subfolder to make deletion easier instead of what I'm
  doing right now.
- Do we need to be able to generate election packages with arbitrary system
  settings? For now I'm just using the default system settings file, not sure
  the best folder structure in libs/fixtures/data if there may be different
  election packages with different system settings files generated.
- Should I add smarter caching for the audio synthesis? Its more complicated
  then speech as a given piece of text can be represented by multiple audio
  clips. I could create a mapping of string | ElectionStringKey to an audio base
  64 to use but the election string key structure is complicated as it can nest.
  Currently it is quite slow when you need to regenerate audio which happens
  even if you just change 1 string in 1 election definition.
- The election packages generated are just over 50mb in what I've seen so far
  triggering git to tell me to use git lfs - probably need to figure that out.


libs/fixture-generators/src/generate-election-package/generate_election_package.ts
```diff
@@ -0,0 +1,229 @@
...
+  const electionPackageHash = sha256(zipContents);
+  const combinedHash = formatElectionHashes(
+    electionDefinition.ballotHash,
+    electionPackageHash
+  );
+  const fileName = `election-package-${combinedHash}.zip`;
+  const packageFilePath = path.join(assetDirectoryPath, fileName);
+  const electionFilePath = path.join(
+    assetDirectoryPath,
+    `generated-election-${combinedHash}.json`
```
jonahkagan, line 223: > Should the hashes be in the generated election-package and election.json
filenames? This means you have to manually update the fixtures index.ts file
if this changes which is a little annoying, but people should be aware when
they are causing it to change. Maybe we could expose it as a variable more
easily so that tests currently referencing these hardcoded values can at least
be fixed and make this the ONLY place you need to update. I probably will want
to keep these files in a subfolder to make deletion easier instead of what I'm
doing right now.

Hmm, trying to think through the pros/cons to having hashes in the name. I guess the pro is that you can use it to distinguish which package is which if you are using multiple. That feels like it could come in handy. Is there some hack we could use like a symlink or something so that the index.ts file can always load the package even if the file name changes? I.e. when we generate the package we also create a symlink from `election-package.zip` to `election-package-{hash}.zip`?

I like the idea of exposing the hash in a variable as well. Though I think we already have code to do that when we parse election packages, so maybe any test that uses it will just get it through that?

libs/fixture-generators/src/generate-election-package/generate_election_package.ts
```diff
@@ -0,0 +1,229 @@
...
+      // Each ballot style will have exactly one grid layout regardless of precinct, ballot type, or ballot mode
+      // So we just need to render a single ballot per ballot style to create the election definition
+      election.ballotStyles.map((ballotStyle) => ({
+        election,
+        ballotStyleId: ballotStyle.id,
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        precinctId: ballotStyle.precincts[0]!,
+        ballotType: BallotType.Precinct,
+        ballotMode: 'test',
+      })),
```
jonahkagan, line 101: > Should I move where the ballot images generated by generate-fixtures in
lib/hmpb to be in lib/fixtures? I think this might make sense in the long term
but I am trying to keep it out of scope for now.

I do think there could be some value in this. We could potentially just move the hmpb fixture generation into `libs/fixtures-generators`. But I think it can definitely be a separate project.

How are you planning to handle the many different elections in the hmpb fixtures (specifically in `general-election` where we test all the paper sizes)?

jonahkagan (approved):
Very exciting, thanks for sharing the draft!

apps/admin/backend/src/store.test.ts
```diff
@@ -316,14 +320,26 @@ function expectArrayMatch<T>(a: T[], b: T[]) {
...
-    systemSettingsData: JSON.stringify(DEFAULT_SYSTEM_SETTINGS),
-    electionPackageFileContents: Buffer.of(),
-    electionPackageHash: 'test-election-package-hash',
+  let store: Store;
+  let electionId: string;
+  let election: Election;
+
+  beforeEach(async () => {
+    const electionFileContents =
+      electionPrimaryPrecinctSplitsFixtures.electionPackageExport.asBuffer();
```
jonahkagan, line 329: Are you imagining most tests will use the election package fixture? My guess is that many will just need the `Election` from within it, so it might be nice to have an easy way to access that.

libs/fixture-generators/src/generate-election-package/generate_election_package.ts
```diff
@@ -0,0 +1,229 @@
...
+    priorElectionPackage.uiStringAudioIds &&
+    priorElectionPackage.uiStringAudioClips
+  ) {
+    zip.file(
+      ElectionPackageFileName.AUDIO_IDS,
+      JSON.stringify(priorElectionPackage.uiStringAudioIds, null, 2),
+      { date: FIXTURES_FILE_DATE }
+    );
+    zip.file(
+      ElectionPackageFileName.AUDIO_CLIPS,
```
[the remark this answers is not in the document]
arsalansufi: Agreed re mock audio clips! I think we should have representative data structures but the contents of the audio in particular don't feel as important. Even just one fixture with real audio would suffice I think. Or if it simplifies things to start, just relying on manual election package generation for true testing of audio, when we need it, is fine by me

carolinemodic:
Closing this proof of concept now in favor of the real PRs

PR #5697, opened 2024-12-06
style(clippy): fix some lints

eventualbuddha:
Identified by Rust 1.83.0.


eventualbuddha:
Looks like 1.83 has more support for operations in `const fn`s than the version we're currently using.

PR #5704, opened 2024-12-10
Refactor language and audio generic code outside of VxDesign

carolinemodic:
## Overview
In order to set up for: https://github.com/votingworks/vxsuite/pull/5695 
This PR extracts out all the code in VxDesign /src/language_and_audio to shared locations, largely libs/backend with some types moved to lib/types. The Translator and SpeechSynthesizer classes that hold VxDesign specific - store knowledge still exist within the design backend and the more generic parent class that does no caching is exposed in libs/backend. I originally made this an abstract class but decided to make it real for ease of testing. Outside of tests though these classes should generally be subclasses with caching or vendor override logic. 

Most of this code was only tested in large end to end VxDesign tests so added some basic more specific testing to get coverage up. 

## Demo Video or Screenshot
N/A

## Testing Plan
Ran Tests 

## Checklist


libs/utils/src/hmpb/hmpb_strings.ts
jonahkagan, line 1: This feels like a weird place for these to live. I think this libs/utils/src/hmpb directory is a remnant of the old hmpb code (when it was part of VxAdmin). Maybe they should just live with the language and audio code?

jonahkagan (approved):

PR #5705, opened 2024-12-10
test(ballot-interpreter): switch from `jest` to `vitest`

### LINKED ISSUE #5708: libs/ballot-interpreter

eventualbuddha:
## Overview

Closes #5708
Switch from `jest` to `vitest` in `libs/ballot-interpreter`.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.


vitest.config.shared.ts
```diff
@@ -0,0 +1,28 @@
+import * as vitest from 'vitest/config';
+
+export const base: vitest.ViteUserConfig = {
+  test: {
+    coverage: {
+      thresholds: { 100: true },
```
jonahkagan, line 6: Should we set thresholds for just `lines` and `branches`, since that's the convention we adopted previously?

libs/test-utils-vitest/package.json
```diff
@@ -18,7 +18,7 @@
     "lint:fix": "pnpm type-check && eslint . --fix",
     "pre-commit": "lint-staged",
     "test": "is-ci test:ci test:watch",
-    "test:ci": "vitest --coverage --reporter junit --outputFile reports/junit.xml",
+    "test:ci": "vitest --coverage",
     "test:coverage": "vitest run --coverage",
```
jonahkagan: I think these should be switched - we want run mode in CI and watch mode in local dev, right? (This is my mistake, I'm just noticing it now)

vitest.config.shared.ts
```diff
@@ -0,0 +1,28 @@
...
+  config: vitest.ViteUserConfig
+): vitest.ViteUserConfig {
+  const { test, ...rest } = config;
+  return vitest.defineConfig({
+    ...base,
+    ...rest,
+    test: {
+      ...base.test,
+      ...test,
+    },
```
jonahkagan, line 26: Should we do a recursive object extension here so you can override individual fields (e.g. change coverage thresholds without also needing to specify `provider` and `include`)?

libs/ballot-interpreter/vitest.config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default defineConfig({
+  test: {
+    setupFiles: ['src/setupTests.ts'],
+    coverage: {
+      thresholds: {
+        lines: 51,
+        functions: 71,
+        statements: 51,
+        branches: 47,
```
jonahkagan, line 11: Same idea about just doing lines and branches

jonahkagan (approved):

vitest.config.shared.ts
eventualbuddha: Done.

libs/test-utils-vitest/package.json
eventualbuddha: Done, though if `CI=true` then `vitest` already doesn't use watch mode by default.

vitest.config.shared.ts
eventualbuddha: Done.

libs/test-utils-vitest/package.json
jonahkagan: Oh, that's good to know. In that case I don't care as much

vitest.config.shared.ts
jonahkagan: Just remembered we have a `mergeObjects` function in libs/basics

eventualbuddha: We could copy it, I guess. Probably don't want to try to import it from here.

jonahkagan: Oh yeah I guess it doesn't make sense to import. Nevermind

PR #5723, opened 2024-12-11
test: consolidate test utils

eventualbuddha:
Consolidates `test-utils` and `test-utils-vitest` by parameterizing the functions that generate `fn`-style mock objects. Stops using `jest.fn` where we only used it incidentally. Reverts as much of the use of `.mts` as possible.


jonahkagan (approved):
This looks great. Nice touch with removing the unnecessary mock functions

PR #5727, opened 2024-12-11
test(logging): fix local test running

eventualbuddha:
The tests were assuming UTC timezone, so running in any other timezone fails. CI runs in UTC by default, so that was going fine.

### DIFF
--- a/libs/logging/package.json
+++ b/libs/logging/package.json
@@ -22,8 +22,8 @@
     "lint": "pnpm type-check && eslint .",
     "lint:fix": "pnpm type-check && eslint . --fix",
     "test": "is-ci test:ci test:watch",
-    "test:watch": "jest --watch",
-    "test:coverage": "jest --coverage",
+    "test:watch": "TZ=UTC jest --watch",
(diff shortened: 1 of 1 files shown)


libs/logging/package.json
```diff
@@ -22,8 +22,8 @@
     "lint": "pnpm type-check && eslint .",
     "lint:fix": "pnpm type-check && eslint . --fix",
     "test": "is-ci test:ci test:watch",
-    "test:watch": "jest --watch",
-    "test:coverage": "jest --coverage",
+    "test:watch": "TZ=UTC jest --watch",
+    "test:coverage": "TZ=UTC jest --coverage",
```
jonahkagan, line 26: I think we should use `Alaska/Anchorage` instead of `UTC` for consistency with the rest of VxSuite

jonahkagan (approved):

libs/logging/package.json
eventualbuddha: https://github.com/votingworks/vxsuite/issues/5733

PR #5732, opened 2024-12-11
Migrate election primary precinct split fixture to use generated version

carolinemodic:
## Overview
Migrates all usages of electionPrimaryWithPrecinctSplits to use newly generated fixtures with grid layouts and translations. I am not migrating everything to use election packages rather then election.json files. There is code in VxAdmin especially (but also other places) constructing election packages out to election files for tests and ideally those will instead just configure from the election package now exported. However I think it will be easier to update those tests once all of the fixtures they are using are converted AND I don't think its strictly necessary to migrate them all so I'm holding off on that for now. 

I considered adding a asElectionPackage() function automatically added to the generated .zip.ts file in res-to-ts but the code that handles election package parsing right now lives in libs/backend and imports helpers from libs/utils neither of which are current dependencies of libs/fixtures. Ultimately its not that hard to just call another helper function when you are using the fixture and I didn't want to over-complicate the dependency graph so I did not add that. 

The toElectionPackage() function that is added to election.json.ts functions I hope to remove at the end of this work and migrate anything using that function to instead load in the ElectionPackage from the .zip export. I'm intentionally NOT exporting that from the fixtures index.ts file to make sure any usages are updated, for this particular election there were none. 

## Demo Video or Screenshot
N/A

## Testing Plan
Tests passing


libs/hmpb/src/render_ballot.tsx
```diff
@@ -476,11 +477,19 @@ export async function renderAllBallotsAndCreateElectionDefinition<
...
     switch (electionSerializationFormat) {
-      case 'vxf':
-        return electionWithGridLayouts;
+      case 'vxf': {
+        // Re-parse the election to ensure it is being saved in a consistent format
+        const sortedElectionWithGridLayouts = safeParseElection(
+          JSON.stringify(electionWithGridLayouts)
+        ).unsafeUnwrap();
+        return sortedElectionWithGridLayouts;
+      }
```
jonahkagan, line 492: Is this specifically about the field order? If, so maybe include that in the comment

Why does this matter now and didn't matter before?

jonahkagan (approved):

libs/hmpb/src/render_ballot.tsx
carolinemodic: I'm not entirely sure what this is mattering now and not before other then this specific flow of creating/reading elections not existing previously. Before this specific flow rewrote a new election json instead of relying on an old one matching. Basically the issue is without this change the generate-election-packages script would plop gridLayouts at the end of the JSON. When libs/hmpb generate-fixtures was reading it in and creating new ballots it would get a new hash of the election after reading in where gridLayouts are now higher up in the order of keys so the hashes don't match even though the contents are identical.

carolinemodic: will update the comment

PR #5736, opened 2024-12-11
refactor(ui): remove `useCancelablePromise`

eventualbuddha:
## Overview

All this hook does is try to prevent component callbacks that have awaited something from continuing to do anything after they are unmounted. In practice, the official guidance from React on this seems to be "don't worry about it".

## Demo Video or Screenshot
n/a

## Testing Plan
Automated testing.


kofi-q (approved):
✅

PR #5738, opened 2024-12-12
refactor(jest): inline calls to `advanceTimers`

eventualbuddha:
This way we can replace the `jest` with `vi` at individual callsites as we get to them.


jonahkagan (approved):
Love this idea!

Should we rebase to `main` rather than cherry-pick?

eventualbuddha:
Yeah, I'll do the rebase now.

PR #5739, opened 2024-12-12
test(dev-dock/backend): move from `jest` to `vitest`

### LINKED ISSUE #5735: libs/dev-dock/backend

eventualbuddha:
Closes #5735 

- [x] Merge #5738


libs/dev-dock/backend/src/dev_dock_api.test.ts
```diff
@@ -36,11 +37,13 @@ import { Api, useDevDockRouter, MockSpec } from './dev_dock_api';
...
-  isFeatureFlagEnabled: (flag: BooleanEnvironmentVariableName) =>
-    featureFlagMock.isEnabled(flag),
-}));
+vi.mock(
+  '@votingworks/utils',
+  async (importActual): Promise<typeof import('@votingworks/utils')> => ({
+    ...(await importActual<typeof import('@votingworks/utils')>()),
+    isFeatureFlagEnabled: (flag) => featureFlagMock.isEnabled(flag),
+  })
+);
```
jonahkagan, line 46: What's this new pattern of passing in `importActual` rather than just using `vi.importActual`? Also what are all the type annotations for?

jonahkagan (approved):

libs/dev-dock/backend/src/dev_dock_api.test.ts
eventualbuddha: This is what @kofi-q's script generated, which seemed a little nicer than having to repeat the string. The type annotation on the `importActual` call tells it to treat the return type as the import of that module, and the type annotation at the return position of the call to `vi.mock`'s callback tells it to expect an object of that type. This should ensure we don't miss anything or typo an export name.

jonahkagan: Nice, thanks for explaining!

PR #5743, opened 2024-12-12
Migrate Famous Names election fixture to have grid layouts / strings generated

carolinemodic:
## Overview
Depends on https://github.com/votingworks/vxsuite/pull/5732

Next step of cleaning up election fixtures, migrates famous names to generate an election with strings & grid layouts and use that across tests. toElectionPackage was already not exported for this election. I am leaving this election as single-language for now, there are many tests that use a random single language election to test single language behavior and I don't want to disrupt that so I'm going to leave everything how it currently exists in terms of single/multi language for now and then when everything is migrated consider if some of the single language fixtures should change to be multi language. 

## Testing Plan
Ran Tests


libs/fixture-generators/bin/regenerate-election-packages
```diff
@@ -9,8 +9,10 @@ export NODE_ENV=development
 # Check if FORCE_RETRANSLATE is set to true
 if [ -z "$FORCE_RETRANSLATE" ]; then
```
jonahkagan, line 10: Nit: This comment is a little confusing since this is actually checking if `FORCE_RETRANSLATE` is unset, right?

jonahkagan (approved):
Exciting to see this come together!

PR #5760, opened 2024-12-16
Switch scanning for write-ins

kshen0:
## Overview

Adds scan panels to the candidate write-in screen. See [this doc](https://docs.google.com/document/d/1nan1Gwrb0CCC757h1UYK4pdGVmiUwSfuuZ2-VNZ_hpQ/edit?tab=t.0#heading=h.77fs298fyke8) for more details.

There are a few necessary follow-ups that I think should be in separate PRs to keep diff size down:
* audio strings for rows and scan panels
* resetting focus state after
  * tabbing through all scan panels in a row twice
  * tabbing through all keys in a scan panel twice
* remapping ATI controller direction pad for the default virtual keyboard

Note that the row with `Space` and `Delete` is a row of a single scan panel, so a voter needs to select the row, then the scan panel, and finally the desired key. I went back and forth on this decision but landed this way because
* it maintains consistency with the row -> panel -> key hierarchy of other rows
* it reduces special-casing in the code

Open to discussion on this interaction pattern because it has a major downside of forcing more actions from the voter than strictly necessary. The alternative is to render a row of keys, skipping the intermediate scan panel.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/83005670-e9b5-48e8-9d36-19ebf0e87f5d

Screenshot 2024-12-16 at 6 48 21 PM
Screenshot 2024-12-16 at 6 48 28 PM
Screenshot 2024-12-16 at 6 48 36 PM
Screenshot 2024-12-16 at 6 48 55 PM

Non-default font sizes:

Screenshot 2024-12-16 at 7 58 55 PM
Screenshot 2024-12-16 at 7 59 01 PM
Screenshot 2024-12-16 at 7 59 05 PM
Screenshot 2024-12-16 at 7 59 15 PM
Screenshot 2024-12-16 at 7 59 20 PM
Screenshot 2024-12-16 at 7 59 25 PM
Screenshot 2024-12-16 at 7 59 38 PM
Screenshot 2024-12-16 at 7 59 43 PM
Screenshot 2024-12-16 at 7 59 49 PM

## Testing Plan
-  [x] added automated tests

## Checklist

### DIFF
--- a/libs/ui/src/virtual_keyboard/scan_panel_virtual_keyboard.test.tsx
+++ b/libs/ui/src/virtual_keyboard/scan_panel_virtual_keyboard.test.tsx
@@ -1,0 +1,245 @@
+import userEvent from '@testing-library/user-event';
+
+import {
+  hasTextAcrossElements,
+  mockOf,
+  TestLanguageCode,
+} from '@votingworks/test-utils';
+import { assertDefined } from '@votingworks/basics';
+
+import { act, render, screen, waitFor } from '../../test/react_testing_library';
+import { newTestContext as newUiStringsTestContext } from '../../test/test_context';
+import { AudioOnly } from '../ui_strings/audio_only';
+import { useCurrentLanguage } from '../hooks/use_current_language';
+import {
+  ScanPanelVirtualKeyboard,
+  US_ENGLISH_SCAN_PANEL_KEYMAP,
+} from './scan_panel_virtual_keyboard';
+
+jest.mock(
+  '../ui_strings/audio_only',
+  (): typeof import('../ui_strings/audio_only') => ({
+    ...jest.requireActual('../ui_strings/audio_only'),
+    AudioOnly: jest.fn(),
+  })
+);
+
+const { ENGLISH, SPANISH } = TestLanguageCode;
+
+const firstRow = 'QWERTYUIOP';
+const secondRow = `ASDFGHJKL'"`;
+const thirdRow = 'ZXCVBNM,.-';
+const fourthRow = 'space delete';
+
+function getMockAudioOnlyTextPrefix(languageCode: string) {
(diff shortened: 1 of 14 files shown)


libs/ui/src/virtual_keyboard/scan_panel_virtual_keyboard.test.tsx
```diff
@@ -0,0 +1,245 @@
+import userEvent from '@testing-library/user-event';
```
kshen0, line 1: Most tests in this file are adapted from `virtual_keyboard.test.tsx` unless noted

libs/ui/src/virtual_keyboard/scan_panel_virtual_keyboard.tsx
```diff
@@ -0,0 +1,471 @@
...
+const Keyboard = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+
+  & button {
+    ${buttonStyles}
+  }
+
+  /* Overrides buttonStyles's justify-content.
+   * https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity
```
kofi-q, line 21: 😮 TIL

kofi-q (approved):
This looks really good!! Such a big improvement 🙌 

All LGTM - as a nice-to-have, could be useful (if feasible) to autofocus on the first panel/button in a group whenever we drop down a level, to shave off one action for the voter in getting to target button

kshen0:
> This looks really good!! Such a big improvement 🙌
> 
> All LGTM - as a nice-to-have, could be useful (if feasible) to autofocus on the first panel/button in a group whenever we drop down a level, to shave off one action for the voter in getting to target button

Great call out, thanks! Will include in a follow up PR that changes focus behavior

PR #5765, opened 2024-12-17
test(logging): replace `jest` with `vitest`

### LINKED ISSUE #5725: libs/logging
- [x] replace `jest` with `vitest`
- [ ] remove `fn` option from `mockLogger` et al

eventualbuddha:
Closes #5725 

In addition to changing the tests for `libs/logging` itself, changes `mockLogger` et al to require a mock function. Some of the `istanbul ignore` directives no longer work because `esbuild` at the version `vitest` uses doesn't support preserving comments preceding a `case` or `default` (see [this](https://github.com/evanw/esbuild/issues/3838)).

### DIFF
--- a/libs/logging/package.json
+++ b/libs/logging/package.json
@@ -22,9 +22,9 @@
     "lint": "pnpm type-check && eslint .",
     "lint:fix": "pnpm type-check && eslint . --fix",
     "test": "is-ci test:ci test:watch",
-    "test:watch": "TZ=UTC vitest",
-    "test:coverage": "TZ=UTC vitest --coverage",
-    "test:ci": "pnpm build && TZ=UTC vitest run --coverage && pnpm build:generate-typescript-types --check && pnpm build:generate-rust-types --check",
+    "test:watch": "TZ=America/Anchorage vitest",
+    "test:coverage": "TZ=America/Anchorage vitest --coverage",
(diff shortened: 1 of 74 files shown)


libs/logging/package.json
```diff
@@ -22,9 +22,9 @@
     "lint": "pnpm type-check && eslint .",
     "lint:fix": "pnpm type-check && eslint . --fix",
     "test": "is-ci test:ci test:watch",
-    "test:watch": "TZ=UTC jest --watch",
-    "test:coverage": "TZ=UTC jest --coverage",
-    "test:ci": "pnpm build && pnpm test:coverage --reporters=default --reporters=jest-junit --maxWorkers=6 && pnpm build:generate-typescript-types --check && pnpm build:generate-rust-types --check",
+    "test:watch": "TZ=UTC vitest",
+    "test:coverage": "TZ=UTC vitest --coverage",
+    "test:ci": "pnpm build && TZ=UTC vitest run --coverage && pnpm build:generate-typescript-types --check && pnpm build:generate-rust-types --check",
```
jonahkagan, line 27: Maybe a good time to switch to `America/Anchorage` timezone?

jonahkagan (approved):
Bummer about the esbuild issue. Do you foresee that being something we can fix in the future?

eventualbuddha:
> Bummer about the esbuild issue. Do you foresee that being something we can fix in the future?

I'm guessing that eventually vitest will either use a new enough version of esbuild or switch to oxc (where hopefully they'll have a solution to this).

PR #5768, opened 2024-12-17
Add script for generating machine label QR code contents

arsalansufi:
## Overview

This PR adds the script necessary for the plan laid out [Labeling VotingWorks Equipment](https://docs.google.com/document/d/1FSi9ghWlSZpQQYEkYc7R0Coh3THR4x_iIalJuTsWx2Q/edit?usp=sharing).

_Help message_

```
$ ./scripts/generate-machine-label-qr-code-contents --help
Options:
  --private-key-path  The path to the VotingWorks label QR codes private key.
                      This is not the same key as our root private key. [string]
  --serial-number     The machine serial number, also referred to as the machine
                      ID                                                [string]

Examples:
  $ generate-machine-label-qr-code-contents --help
  $ generate-machine-label-qr-code-contents \
  --private-key-path path/to/private-key.pem \
  --serial-number SC-00-000
```

_Basic usage_

```
$ ./scripts/generate-machine-label-qr-code-contents --private-key-path /media/psf/shared/keys/vx-label-qr-codes-private-key.pem --serial-number TEST-00-000
Enter pass phrase for /media/psf/shared/keys/vx-label-qr-codes-private-key.pem:
https://vxqr.org/sn/TEST-00-000#<signature>
```

_Batch usage_

```
$ cat machine-ids.txt 
1
2
3
$ cat machine-ids.txt | xargs -I {} ./scripts/generate-machine-label-qr-code-contents --private-key-path /media/psf/shared/keys/vx-label-qr-codes-private-key.pem --serial-number {} > machine-label-qr-code-contents.txt
Enter pass phrase for /media/psf/shared/keys/vx-label-qr-codes-private-key.pem:
Enter pass phrase for /media/psf/shared/keys/vx-label-qr-codes-private-key.pem:
Enter pass phrase for /media/psf/shared/keys/vx-label-qr-codes-private-key.pem:
$ cat machine-label-qr-code-contents.txt
https://vxqr.org/sn/1#<signature>
https://vxqr.org/sn/2#<signature>
https://vxqr.org/sn/3#<signature>
```

_Input validation_

```
$ ./scripts/generate-machine-label-qr-code-contents --private-key-path /media/psf/shared/keys/vx-label-qr-codes-private-key.pem --serial-number abc$
❌ Serial number must contain only uppercase letters, numbers, and dashes
```

## Testing Plan

- [x] Tested manually
- [x] Updated automated tests where relevant
- [x] Baked a verification check into the script itself


arsalansufi:
A nice follow-up would probably be a script that allows us to check the correctness of these URLs using the corresponding cert / public key

libs/auth/scripts/src/generate_machine_label_qr_code_contents.ts
```diff
@@ -0,0 +1,124 @@
...
+): Promise<CommandLineArgs> {
+  const argParser = yargs()
+    .options({
+      'private-key-path': {
+        description:
+          'The path to the VotingWorks label QR codes private key. ' +
+          'This is not the same key as our root private key.',
+        type: 'string',
+      },
+      'private-key-passphrase-file-path': {
```
benadida, line 26: do we need this? I don't love encouraging the passphrase to be in a file on disk.

libs/auth/scripts/src/generate_machine_label_qr_code_contents.ts
```diff
@@ -0,0 +1,124 @@
...
+  serialNumber,
+}: CommandLineArgs): Promise<void> {
+  const signingPrivateKey: FileKey = {
+    source: 'file',
+    path: privateKeyPath,
+    passphraseFilePath: privateKeyPassphraseFilePath,
+  };
+  const message = constructPrefixedMessage(
+    'signed-serial-number',
+    `sn=${serialNumber}`
```
benadida, line 97: let's assert a serial number format that is only chars, digits, and dashes.

benadida (changes requested):
I question the passphrase in a file... Encourages not ideal practice.

Also, I think a couple of tests to sign and verify a URL would be good.

libs/auth/scripts/src/generate_machine_label_qr_code_contents.ts
arsalansufi: Ah yeah I was thinking of the batch generation use case where you'd need to generate URLs for say a hundred machine IDs and thereby need to enter in the passphrase a hundred times. But I'm okay with that.

arsalansufi: https://github.com/votingworks/vxsuite/pull/5768/commits/c6a00dd55f08fd1b98dddcacca1f01dee575eff3

libs/auth/scripts/src/generate_machine_label_qr_code_contents.ts
arsalansufi: https://github.com/votingworks/vxsuite/pull/5768/commits/75f00c9a9c55d8d5944dffc194f0be19d3859070

arsalansufi:
Decided to bake a verification check into the script itself to ensure that we're signing using the correct key. This required me to commit the Vx Label QR Codes CA cert into the repo. https://github.com/votingworks/vxsuite/pull/5768/commits/a6d2748711407db369d028c4b6a31061fac3e126

I want to add more tests for all the scripts in this repo but for now this is a good start

benadida (approved):
love it

PR #5769, opened 2024-12-18
refactor(fixtures): load fixture files directly

eventualbuddha:
Rather than generating `.ts` files that embed the contents of data files as base64, just load the files directly when required. Doing this means we can't just have bare values exported and have to export functions, hence the fairly massive diff.

This means we can also get rid of `res-to-ts` and just use some simple helpers to create easy-to-use fixtures. For example:

```ts
// an election.json file
export const electionJson = builders.election(
  'data/electionTwoPartyPrimary/election.json'
);

// a basic file
export const systemSettings = builders.file(
  'data/systemSettings.json'
);

// an image file
export const scanMarkedFront = builders.image(
  'data/electionGridLayoutNewHampshireTestBallot/scan-marked-front.jpeg'
);

// a directory
const castVoteRecords = builders.directory(
  'data/electionGridLayoutNewHampshireTestBallot/castVoteRecords'
);
```


libs/printing/src/__image_snapshots__/roll-document-2.png
kofi-q, line 1: Might be an accidental add? I think this is the snapshot that's been causing issues locally for us

kofi-q (approved):
🎉

libs/printing/src/__image_snapshots__/roll-document-2.png
eventualbuddha: Removed.

PR #5773, opened 2024-12-19
test(backend): replace `jest` with `vitest`

### LINKED ISSUE #5742: libs/backend

eventualbuddha:
Closes #5742 

This library provides various test utilities to the app backends, which are still using jest. So this change makes those utilities work with either vitest or jest in a similar way to what we've done before. The main difference this time is that we needed to pass global hooks/helpers through to some helpers, which can be a little tricky typing-wise. To help with that, I'm importing `@jest/globals` on both sides of the call to ensure we get the same types and that everything is happy.

Additionally, the mock Google service classes were changed to functions because TS overloading of class constructors is very much not fun.


apps/admin/backend/test/app.ts
```diff
@@ -131,7 +131,7 @@ export async function configureMachine(
 export function buildMockLogger(
   auth: DippedSmartCardAuthApi,
   workspace: Workspace
-): Logger {
+): MockLogger {
```
eventualbuddha, line 134: The result of this function is passed to `testDetectDevices`, which now requires a `MockBaseLogger`.

apps/central-scan/backend/src/server.test.ts
```diff
@@ -35,10 +36,10 @@ test('logs device attach/un-attach events', async () => {
     onListening?.();
     return undefined as unknown as Server;
   });
-  jest.spyOn(console, 'log').mockImplementation();
+  jest.spyOn(console, 'log').mockReturnValue();
```
eventualbuddha, line 39: The `@jest/globals` version of `mockImplementation` wants an implementation. 🤷

apps/mark-scan/backend/src/server.test.ts
```diff
@@ -23,7 +24,7 @@ jest.mock(
   '@votingworks/backend',
   (): typeof import('@votingworks/backend') => ({
     ...jest.requireActual('@votingworks/backend'),
-    initializeSystemAudio: jest.fn(),
+    initializeSystemAudio: jest.fn<() => Promise<void>>(),
```
eventualbuddha, line 27: Some slight typing differences between `@types/jest` and `@jest/globals`.

libs/backend/src/cast_vote_records/build_cast_vote_record.test.ts
```diff
@@ -760,7 +761,7 @@ test('buildCastVoteRecord - HMPB ballot with unmarked write-in', () => {
   });
 });
 
-describe('hash manipulation', () => {
+test('hash manipulation', () => {
```
eventualbuddha, line 764: This was actually incorrect before. The assertions were running outside the context of a test.

libs/backend/src/cast_vote_records/build_report_metadata.test.ts
```diff
@@ -196,8 +201,6 @@ test('builds well-formed cast vote record report', () => {
       ])
     );
   }
-
-  jest.useRealTimers();
```
eventualbuddha, line 200: Moved this up to an `afterEach` because a failed assertion would mean this never gets called.

jonahkagan (approved):

PR #5776, opened 2024-12-20
test(types): replace `jest` with `vitest`

### LINKED ISSUE #5775: libs/types

eventualbuddha:
Closes #5775


libs/types/src/cdf/ballot-definition/index.ts
eventualbuddha, line 1: This and the other large diffs are auto generated.

jonahkagan (approved):

Issue #5706, opened 2024-12-10
Move from `jest` to `vitest`


eventualbuddha:
(See also the [official guide](https://vitest.dev/guide/migration#migrating-from-jest))

The motivation for this is not so much any particular problem with `jest` as the desire to move from targeting CommonJS modules (CJS) to EcmaScript modules (ESM) in our build system:
- CJS modules interoperate poorly with ESM packages from NPM, of which there are a growing number. This has caused us to use older versions of various packages instead of the latest.
- Faster native-speed tooling for building TypeScript tends to have poor or non-existent support for CJS. Targeting CJS therefore locks us into using slower build tools.

`vitest` has a very similar API and DX to `jest`, but has much better support for ESM. It uses faster native-speed tooling-esbuild at the moment-though the gains appear to be [uneven][perf]. So I wouldn't necessarily consider this to be a net performance win.

[perf]: https://dev.to/thejaredwilcurt/vitest-vs-jest-benchmarks-on-a-5-year-old-real-work-spa-4mf1

## Progress

- [x] #5722
- [x] #5840
- [x] apps/admin/frontend/prodserver
- [x] apps/admin/integration-testing
- [x] #5825
- [x] #5831
- [x] apps/central-scan/frontend/prodserver
- [x] apps/central-scan/integration-testing
- [x] #5843
- [x] #5855
- [x] #5859
- [ ] apps/mark-scan/frontend
- [x] apps/mark-scan/frontend/prodserver
- [x] apps/mark-scan/integration-testing
- [x] #5750
- [x] #5835
- [x] apps/mark/frontend/prodserver
- [x] apps/mark/integration-testing
- [x] #5827
- [x] #5726
- [x] apps/scan/frontend/prodserver
- [x] #5754
- [x] #5790
- [x] #5740
- [x] #5742
- [x] #5714
- [x] #5708
- [x] #5707
- [x] #5846
- [x] #5848
- [x] #5857
- [x] #5803
- [x] #5729
- [x] #5735
- [x] #5720
- [x] #5833
- [x] #5850
- [x] #5766
- [x] #5756
- [x] #5805
- [x] #5763
- [x] #5761
- [x] #5807
- [x] #5710
- [ ] #5725
- [x] #5715
- [x] #5717
- [x] #5771
- [x] #5809
- [x] #5724
- [ ] #5709
- [x] #5775
- [ ] #5745
- [x] #5800
- [x] #5712
- [x] script

PR #5777, opened 2024-12-20
test: switch a bunch of packages from `jest` to `vitest`

eventualbuddha:
Refs https://github.com/votingworks/vxsuite/issues/5706

The work was all done with a bunch of individual pull requests on the `vitest` branch.


libs/ui/src/virtual_keyboard/scan_panel_virtual_keyboard.test.tsx
eventualbuddha, line 1: Looks like this was accidentally deleted. Sorry, @kshen0!

libs/auth/scripts/src/generate_machine_label_qr_code_contents.ts
```diff
@@ -1,134 +0,0 @@
-import path from 'node:path';
```
arsalansufi, line 1: Looks like this and a few other related changes in libs/auth for https://github.com/votingworks/vxsuite/pull/5768 accidentally got lost, too - will bring them back!

Issue #5698, opened 2024-12-06
VxDesign: Hosting & Authentication


adghayes:
[Hosted VxDesign Brief](https://docs.google.com/document/d/1GMlUX3Jjka7ACW2ltp7iYlYEMUoogYhPn3I5GUoOoLs/edit?usp=sharing)

adghayes:
SLI does not need authentication, so hosting could be spun up before accounts + authentication.

arsalansufi:
Noting that it's okay for us to host without auth for SLI. So we can phase this, hosting first, then auth on top of that

arsalansufi:
@mattroe is planning to spec out the long-term vision for the auth model

arsalansufi:
Closing out this mega ticket! Nice work Kofi!! 👏

PR #5791, opened 2025-01-06
Set up VxDesign container-based Heroku deployment

kofi-q:
## Overview

Adding config for building and deploying a VxDesign docker container to Heroku.

- Add `Dockerfile.vxdesign`, which currently copies the repo and runs a `pnpm install` and `pnpm build:prod`. Not yet doing any pruning, but we can add that in if we find a need.
- Add `heroku.yml` and `app.json` for Heroku configuration (unclear if the `app.json` is really needed - Heroku's declarative vs CLI/UI configuration delineation is still confusing to me, but will sort out the details later).
- Update VxDesign server to serve static files from the frontend bundle build.

### TODO
- Migrate DB to Postgres
- Figure out scaling needs
- Look into auth options

## Demo Video or Screenshot

Dashboard for the WIP Heroku app is available here: https://dashboard.heroku.com/apps/tmp-vxdesign-dev

Link to the test deployment will be posted in the project Slack channel (also available via the "Open App" action in Heroku).

## Testing Plan
- Fair amount of parity between the behaviour of Heroku's container engine vs my local Docker instance, so currently testing out changes locally before attempting them on Heroku (also because the latter takes quite a while to complete).
- Will be testing out PDF rendering a bit to see what the load looks like for a single dyno, to inform scaling decisions.


.dockerignore
```diff
@@ -0,0 +1,10 @@
+**/*.local
+**/*.tsbuildinfo
+**/build
+**/node_modules
+**/target
+Dockerfile
+Dockerfile.*
+heroku.yml
+libs/fixtures/data
+libs/hmpb/fixtures
```
kshen0, line 10: Can this ignore other apps?
```
apps/admin
apps/scan
apps/central-scan
apps/mark-scan
```

package.json
```diff
@@ -10,6 +10,9 @@
     "configure-vxdev-env": "node libs/utils/build/scripts/generate_env_file.js --isVxDev -o /vx/config/.env.local",
     "run-dev": "script/run-dev"
   },
+  "engines": {
+    "node": "20.16.0"
```
kshen0, line 14: Crazy idea but how trivial would it be to upgrade to node v22.x?

kshen0 (approved):

.dockerignore
kofi-q: Ah, great call - worth a shot! Our pnpm workspace configuration uses glob patterns, so hopefully, there aren't any direct references to these

package.json
kofi-q: I wish - sadly, it's a can of worms. Tried upgrading to v22 back when we went from v16 to v20, but there's a breaking change to the native node-api that affects a lot of our native  dependencies.

Technically, VxDesign should only need `node-canvas`, as far as I'm aware (which I think has a newer 22-compatible version), so we could've theoretically gotten away with v22 just for this deployment if we had an easy way to pnpm install and build only the pieces needed for VxDesign

.dockerignore
```diff
@@ -6,5 +6,8 @@
 Dockerfile
 Dockerfile.*
 heroku.yml
-libs/fixtures/data
```
kofi-q: I think the recent changes for the move to vitest have made it so there's a transitive dependency on the fixture data somewhere - had to restore this. Will track down later and see if it's easy to pull out of the dependency chain

PR #5799, opened 2025-01-09
vxdesign structured name input

kshen0:
## Overview

* Adds structured first, middle, and last name inputs to VxDesign candidate contest creation
* Prefers to sort by last name, first name, and finally full concatenated name

## Demo Video or Screenshot

Screenshot 2025-01-09 at 11 14 26 AM

https://github.com/user-attachments/assets/0cac2142-99a1-4f10-a2b9-e22978e20faf

## Testing Plan
- Added tests to `contests_screen.test.tsx` and `candidate_rotation.test.ts`

## Checklist

- [x] have added a screenshot and/or video to this PR to demo the change

### DIFF
--- a/libs/types/src/election.ts
+++ b/libs/types/src/election.ts
@@ -112,6 +112,10 @@ export interface Candidate {
   readonly partyIds?: readonly PartyId[];
   readonly isWriteIn?: boolean;
   readonly writeInIndex?: number;
+  // Structured name properties are supported only in VxDesign.
+  readonly firstName?: string;
+  readonly middleName?: string;
+  readonly lastName?: string;
 }
 export const CandidateSchema: z.ZodSchema<Candidate> = z
   .object({
--- a/apps/design/backend/src/candidate_rotation.test.ts
+++ b/apps/design/backend/src/candidate_rotation.test.ts
@@ -149,4 +149,96 @@ describe('rotateCandidates', () => {
       },
     ]);
   });
+
+  test('rotates candidates according to structured last name data', () => {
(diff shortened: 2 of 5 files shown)


libs/types/src/election.ts
```diff
@@ -112,6 +112,10 @@ export interface Candidate {
   readonly partyIds?: readonly PartyId[];
   readonly isWriteIn?: boolean;
   readonly writeInIndex?: number;
+  // Structured name properties are supported only in VxDesign.
+  readonly firstName?: string;
+  readonly middleName?: string;
+  readonly lastName?: string;
```
kshen0, line 118: I thought about extending `Candidate` to a type used only in VxDesign eg. `VxDesignCandidate` found that muddied the existing `CandidateContest` type. We'd have to loosen `CandidateContest.candidates` to allow `Candidate[] | VxDesignCandidate[]` which IMO causes a lot more headache downstream in other VX apps than just having these 3 optional fields.

kofi-q: Makes sense for now, for expediency sake.

After looking through the VxDesign code a bit though, it does feel like it's outgrowing the data types we use in the rest of the product (and we've already started splitting out a few VxDesign-specific types like ballot styles and precinct splits), so could make sense, longer term, to represent stuff like this in separate types for VxDesign, which would then handle translating those into the election package format when exporting. Then we also wouldn't need to expose unused fields to downstream consumers and reduce confusion.

kofi-q (approved):

PR #5802, opened 2025-01-10
test(usb-drive): replace `jest` with `vitest`

### LINKED ISSUE #5800: libs/usb-drive

eventualbuddha:
Closes #5800 
Part of #5706

### DIFF
--- a/libs/usb-drive/package.json
+++ b/libs/usb-drive/package.json
@@ -15,13 +15,14 @@
     "lint:fix": "pnpm type-check && eslint . --fix",
     "pre-commit": "lint-staged",
     "test": "is-ci test:ci test:watch",
-    "test:ci": "pnpm build && pnpm test:coverage --reporters=default --reporters=jest-junit --maxWorkers=6",
-    "test:coverage": "jest --coverage",
-    "test:watch": "jest --watch",
+    "test:ci": "pnpm build && pnpm test:coverage --run",
+    "test:coverage": "vitest --coverage",
+    "test:watch": "vitest",
     "type-check": "tsc --build"
   },
   "dependencies": {
     "@votingworks/basics": "workspace:*",
+    "@votingworks/fixtures": "workspace:*",
     "@votingworks/fs": "workspace:*",
     "@votingworks/logging": "workspace:*",
     "@votingworks/test-utils": "workspace:*",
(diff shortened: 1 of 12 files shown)


libs/usb-drive/package.json
```diff
@@ -15,13 +15,14 @@
...
-    "test:coverage": "jest --coverage",
-    "test:watch": "jest --watch",
+    "test:ci": "pnpm build && pnpm test:coverage --run",
+    "test:coverage": "vitest --coverage",
+    "test:watch": "vitest",
     "type-check": "tsc --build"
   },
   "dependencies": {
     "@votingworks/basics": "workspace:*",
+    "@votingworks/fixtures": "workspace:*",
```
eventualbuddha, line 25: Added this dependency because I wanted to use the easier tmp file setup/cleanup it provides.

libs/usb-drive/src/list_directory_on_usb_drive.ts
```diff
@@ -38,7 +38,6 @@ export async function* listDirectoryOnUsbDrive(
       yield* listDirectory(join(usbDriveStatus.mountPoint, relativePath));
       break;
 
-    // istanbul ignore next
```
eventualbuddha, line 41: As before, these can't apply to a `switch` case.

libs/usb-drive/src/mocks/memory_usb_drive.test.ts
eventualbuddha, line 1: Since I changed the mock USB drive functionality a little bit I added some tests to sanity check it.

libs/usb-drive/src/mocks/memory_usb_drive.ts
```diff
@@ -18,18 +18,6 @@ export interface MockUsbDrive {
   removeUsbDrive(): void;
 }
 
-// Clean up any mock USB drives directories that were created during tests
-if (typeof jest !== 'undefined') {
```
eventualbuddha, line 22: We can't (afaik) cleanly conditionally run code in the vitest environment, so I replaced this with an assumption that the `@votingworks/fixtures` cleanup function would be running for any test that uses this.

jonahkagan (approved):

libs/usb-drive/src/mocks/memory_usb_drive.ts
```diff
@@ -39,6 +27,9 @@ if (typeof jest !== 'undefined') {
  * automatically be updated to return the correct status.
  *
  * Warning: if you use both interfaces, they may get out of sync.
+ *
+ * Requires that `setupTemporaryRootDir()` from `@votingworks/fixtures` has been
+ * called.
```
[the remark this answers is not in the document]
eventualbuddha: It actually will crash if you haven't with an error message telling you that you need to call it. See https://github.com/votingworks/vxsuite/blob/393fab5fb10e44821917ae4ff6f2c1396cb48155/libs/fixtures/src/tmpdir.ts#L43-L46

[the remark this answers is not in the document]
jonahkagan: Oh, perfect! I forgot you already had that guard

PR #5808, opened 2025-01-10
test(hmpb): replace `jest` with `vitest`

### LINKED ISSUE #5807: libs/hmpb

eventualbuddha:
Closes #5807 
Part of #5706


libs/hmpb/src/render_ballot.tsx
```diff
@@ -496,9 +496,10 @@ export async function renderAllBallotsAndCreateElectionDefinition<
       }
       case 'cdf':
         return convertVxfElectionToCdfBallotDefinition(electionWithGridLayouts);
-      default:
-        /* istanbul ignore next */
+      default: {
+        /* istanbul ignore next - @preserve */
```
jonahkagan, line 500: Does this work here?

jonahkagan (approved):

libs/hmpb/src/render_ballot.tsx
jonahkagan: Gotcha. Should we convert to this pattern in general (move the ignore comments into the default blocks)?

eventualbuddha: It's probably the best option we have, though once we can use negative thresholds I'm not sure it'll matter much.

PR #5816, opened 2025-01-14
[VxDesign] Wire up Postgres integration

kofi-q:
## Overview

Migrating VxDesign backend logic over from SQLite to Postgres:
- Update store layer to use clients from the postgres pool (managed by the pg wrapper in `src/db`)
- Async-ify store consumers, now that all queries are async
- Add a `TestStore` helper for initialising a per-test-file Postgres schema so tests can run concurrently against the same DB
- Update CircleCI config to include an additional Postgres server image for `design-backend` that tests can run against
  - Also sneaking in an update I've been meaning to make to remove the unneccesary `nodejs-browsers` image (was only used back when integration test ran on `selenium`)

## Testing Plan
- Manually tested locally and on Heroku (still need to do manual testing for the translation/TTS worker logic - my local GCloud config is a weird state at the moment).
- Relying on updated automated tests as regression guards

## Checklist


apps/design/backend/src/db/db.ts
```diff
@@ -36,11 +42,26 @@ export class Db {
...
   async withClient<T>(fn: (client: Client) => Promise<T>): Promise<T> {
     const poolClient = await this.pool.connect();
-    const client = new Client(await this.pool.connect());
+    const client = new Client(poolClient);
 
     try {
+      // Enable test suites to run concurrently on separate DB schemas.
+      // The default schema search path needs to be set on a per-connection
+      // basis.
+      // [TODO] Figure out if there's a better way to do this.
```
kshen0, line 58: What do you feel are the disadvantages to how this is currently done?

.circleci/config.yml
```diff
@@ -3,23 +3,25 @@
...
-  nodejs:
+
+  nodejs_postgres:
     docker:
       - image: votingworks/cimg-debian12:4.1.0
         auth:
           username: $VX_DOCKER_USERNAME
           password: $VX_DOCKER_PASSWORD
 
+      - image: cimg/postgres:14.0
```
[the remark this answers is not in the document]
kofi-q: Good catch! Initially copied from some Circle docs and never updated - thanks for flagging

apps/design/backend/src/db/db.ts
kofi-q: Thought there might be an environment-wide flag we can set somewhere, instead of having to do this every time a client's pulled from the pool, but it's not looking likely - and this is only in the test path anyway, so not really a big deal, now that I think about it

kshen0 (approved):

PR #5817, opened 2025-01-15
feat(test-utils): add `disposable` helpers

eventualbuddha:
## Overview
Adds `disposable` and `asyncDisposable` for wrapping resources that have explicit cleanup functions that do not run automatically.  Uses a new feature of JavaScript: the `using` and `await using` declaration kinds.

Most resources in the NodeJS ecosystem don't support this feature yet, so this is a way of adding it specifically for use in tests where we might otherwise fail to clean up a resource because of an expectation failure.

## Demo Video or Screenshot
```ts
test('example', () => {
  using resource = disposable(createResource(), (r) => r.close());
 // Use `resource` here.
});
```

## Testing Plan
Added tests for this feature, and am using it on the branch that originally inspired these helpers.

### DIFF
--- a/libs/test-utils/src/disposable.test.ts
+++ b/libs/test-utils/src/disposable.test.ts
@@ -1,0 +1,81 @@
+/* eslint-disable no-lone-blocks */
+import { expect, test } from 'vitest';
+import { asyncDisposable, disposable } from './disposable';
+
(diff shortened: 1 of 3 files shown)


kofi-q (approved):
🎉

Issue #5702, opened 2024-12-09
Identify procedures for managing v3 vs. v4 VxDesign and election packages / ballots


arsalansufi:
Context:
- We can't produce enough v4 units to get everyone onto v4 in time for NH local elections in March, so some existing customers will stay on v3.
- Everyone will be using a deployed version of v4 VxDesign for data entry and proofing. VotingWorks will still manage the final exporting of election packages and ballots and can accordingly handle the conversion to v3 for the customers that need it.
- The more automated the solution the better, of course, but parts of the process can be manual if need be. We'll be doing this for around 15 towns.
- The election package conversion need not be perfect in terms of VxMark support, as we won't be piloting VxMark in a v3 town.

PR #5818, opened 2025-01-15
feat(hmpb): fix preview

eventualbuddha:
Builds on top of #5817, in service of #5702.

The HMPB dev preview feature was broken in `libs/hmpb`. This fixes it and protects against future regressions.

CleanShot 2025-01-14 at 17 22 59@2x


jonahkagan (approved):
Thanks for fixing this! I like the change to put the hmpb strings catalog in libs/hmpb - that makes a lot more sense organizationally too!

Issue #5793, opened 2025-01-07
VxDesign: proofing workflow


mattroe:

jonahkagan:
Implemented a basic button to finalize all ballots. I think we should add a follow up that locks all editing of the election after finalizing. And also a way to undo using internal tools.

jonahkagan:
Closing in favor of https://github.com/votingworks/vxsuite/issues/5838

PR #5820, opened 2025-01-16
VxDesign: Button to finalize ballots

jonahkagan:
## Overview

Task: #5793 

First step towards a ballot proofing flow. Adds a button to "finalize ballots" to the Ballots screen. Currently, there's no way to undo finalizing ballots, but eventually we can build this into our internal tools. Also, finalizing ballots should lock the rest of the election to prevent further edits, but that is punted for now as well.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/f9cf7c11-f8bd-4010-ae52-a590941dfe4b

## Testing Plan
Added new automated tests

## Checklist

### DIFF
--- a/apps/design/frontend/src/ballots_screen.test.tsx
+++ b/apps/design/frontend/src/ballots_screen.test.tsx
@@ -7,7 +7,7 @@ import {
   MockApiClient,
 } from '../test/api_helpers';
 import { generalElectionRecord, primaryElectionRecord } from '../test/fixtures';
-import { render, screen, within } from '../test/react_testing_library';
+import { render, screen, waitFor, within } from '../test/react_testing_library';
 import { withRoute } from '../test/routing_helpers';
 import { BallotsScreen } from './ballots_screen';
 import { routes } from './routes';
@@ -43,6 +43,7 @@ describe('Ballot styles tab', () => {
     apiMock.getElection
       .expectCallWith({ electionId })
       .resolves(generalElectionRecord);
+    apiMock.getBallotsFinalizedAt.expectCallWith({ electionId }).resolves(null);
     renderScreen(electionId);
     await screen.findByRole('heading', { name: 'Ballots' });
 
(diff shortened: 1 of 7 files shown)


apps/design/frontend/src/ballots_screen.tsx
```diff
@@ -142,104 +171,180 @@ function BallotStylesTab(): JSX.Element | null {
...
+                color="primary"
+                fill="outlined"
+                disabled={
+                  ballotsFinalizedAt !== null ||
+                  setIsBallotProofingCompleteMutation.isLoading
+                }
+                onPress={() => setIsConfirmingFinalize(true)}
+              >
+                Finalize Ballots
+              </Button>
```
kshen0, line 208: I think this structure is a little hard to reason about because (1) it has nested rows and (2) I'd expect a `Row` to take the full width of its container, so I'm surprised the `Button` appears to the right of the text contained by the `Row`. Is there a strong reason to use this structure over `flex` styling?

jonahkagan:
> In the "Ballots are Finalized" state the checkmark icon is offset from the "Ballots are Finalized" text. Is it possible to align them?



Whoops, I already fixed this but forgot to take a new video I guess. Here's what it looks like:
Screenshot 2025-01-16 at 11 40 08 AM

apps/design/frontend/src/ballots_screen.test.tsx
```diff
@@ -43,6 +43,7 @@ describe('Ballot styles tab', () => {
     apiMock.getElection
       .expectCallWith({ electionId })
       .resolves(generalElectionRecord);
+    apiMock.getBallotsFinalizedAt.expectCallWith({ electionId }).resolves(null);
```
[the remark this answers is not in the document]
jonahkagan: I think it could be added to the `beforeEach`, but I thought it might be a little more confusing to read the test case that way

apps/design/frontend/src/ballots_screen.tsx
jonahkagan: All `Row` does is apply `display: flex; flex-direction: row`. Are you suggesting just applying those styles inline instead?

kshen0:
Ah I misinterpreted Row as a Table component. Current implementation makes
sense - sorry!

kshen0 (approved):

PR #5824, opened 2025-01-17
test(admin): replace `jest` with `vitest`

### LINKED ISSUE #5722: apps/admin/backend

eventualbuddha:
Closes #5722 
Part of #5706 

The only tricky part of this was that `authenticateArtifactUsingSignatureFile` was being mocked in tests, but was still being called. This is because only the direct imports of `@votingworks/auth` were being intercepted; those in built files were not. To fix this, we direct vitest to import the TS directly for the `backend` and `auth` libs.


apps/admin/backend/vitest.config.ts
```diff
@@ -9,6 +9,10 @@ export default defineConfig({
       './test/setup_custom_matchers.ts',
     ],
     coverage: {
+      thresholds: {
+        lines: 100,
+        branches: 98,
```
jonahkagan, line 14: not bad!

jonahkagan (approved):

PR #5839, opened 2025-01-21
test(mark/frontend): replace `jest` with `vitest`

### LINKED ISSUE #5835: apps/mark/frontend

eventualbuddha:
Closes #5835
Part of #5706

The main difference from jest to vitest is to use fake timers that automatically advance the current time. Without this, `screen.findBy*` calls would simply time out (h/t @jonahkagan for the pointer).


apps/mark/frontend/src/apimachine_config.test.tsx
```diff
@@ -26,7 +26,7 @@ test('machineConfig is fetched from api client by default', async () => {
   apiMock.expectGetElectionState({
     precinctSelection: ALL_PRECINCTS_SELECTION,
   });
-  render(<App reload={jest.fn()} apiClient={apiMock.mockApiClient} />);
+  render(<App reload={vi.fn()} apiClient={apiMock.mockApiClient} />);
   apiMock.setAuthStatusPollWorkerLoggedIn(electionDefinition);
-  await screen.findByText('mock-code-version');
+  await vi.waitFor(() => screen.getByText('mock-code-version'));
```
jonahkagan, line 31: I think I ran into something similar and a workaround I found was to use `vi.useFakeTimers({ shouldAdvanceTime: true })`. I think it's worth trying that out since `findBy*` is a very useful and readable pattern.

Alternatively, do we need fake timers everywhere?

apps/mark/frontend/src/pages/admin_screen.test.tsx
```diff
@@ -23,7 +24,8 @@ import {
 let apiMock: ApiMock;
 
 beforeEach(() => {
-  jest.useFakeTimers().setSystemTime(new Date('2020-10-31T00:00:00.000'));
+  vi.useRealTimers();
+  // vi.useFakeTimers().setSystemTime(new Date('2020-10-31T00:00:00.000'));
```
jonahkagan, line 28: Remove comment.

Also, why do we need `vi.useRealTimers` here? I don't see anywhere that we're setting fake timers as a default for all tests.

apps/mark/frontend/vitest.config.ts
```diff
@@ -0,0 +1,29 @@
...
+      exclude: [
+        'src/config',
+        'src/stubs',
+        'src/**/*.d.ts',
+        'src/index.tsx',
+        'src/contexts/ballot_context.ts',
+        '**/*.test.{ts,tsx}',
+      ],
+      thresholds: {
+        branches: 98,
```
jonahkagan, line 19: add line coverage threshold

apps/mark/frontend/src/pages/admin_screen.test.tsx
eventualbuddha: Remove.

jonahkagan (approved):

Issue #5795, opened 2025-01-07
VxDesign: Port NH specific functionality from v3 branch


mattroe:
https://docs.google.com/spreadsheets/d/1IIZcgyTyNaSRPhNe5m4Dkw8ps-jrI9hhu4UAPL7UcN4/edit?gid=0#gid=0

PR #5811, opened 2025-01-10
add features provider

kshen0:
## Overview

Related to https://github.com/votingworks/vxsuite/issues/5795

Adds a `FeaturesContext` to expose feature configuration by state. Right now it's just NH vs not-NH but eventually we'll want to support different configurations for NH, MS, and VX users.

I added the "Bubble Position" SegmentedControl for illustration, but it's not wired up to anything yet.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/0c2ec8a1-79d0-4779-9988-5c1cf1ca627c

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #5852, opened 2025-01-22
right align bubbles for NH

kshen0:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/5795

Moves bubbles to the right for the v4 NH template and updates write-in area.

## Demo Video or Screenshot

Screenshot 2025-01-21 at 5 28 15 PM

## Testing Plan

- Manually tested with `libs/hmpb` ballot preview tool
- Updated snapshot test in VxDesign

## Checklist


libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
   );
 }
 
+function CandidateContest({
```
kshen0, line 246: Almost all of this is copy + pasted from `vx_default_ballot_template` except for styling and DOM tree position of `<Bubble>` and `<BubbleWrapper>` components, noted below

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+            const optionInfo: OptionInfo = {
+              type: 'write-in',
+              contestId: contest.id,
+              writeInIndex,
+              writeInArea: {
+                top: 0.8,
+                right: -0.8,
+                bottom: 0.3,
+                left: 8.8,
+              },
```
kshen0, line 345: `writeInArea` has been updated

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+                    style={{
+                      borderBottom: `1px solid ${Colors.BLACK}`,
+                      height: '1.25rem',
+                    }}
+                  />
+                  <div style={{ fontSize: '0.8rem' }}>
+                    <WriteInLabel />
+                  </div>
+                </div>
+                <BubbleWrapper isWriteIn optionInfo={optionInfo} />
```
kshen0, line 370: `<BubbleWrapper` moved after `WriteInLabel`

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+              >
+                <div style={{ flex: 1, textAlign: 'right' }}>
+                  <strong>{candidate.name}</strong>
+                  {partyText && (
+                    <DualLanguageText delimiter="/">
+                      <div>{partyText}</div>
+                    </DualLanguageText>
+                  )}
+                </div>
+                <BubbleWrapper optionInfo={optionInfo} />
```
kshen0, line 329: `BubbleWrapper` moved after candidate name

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+                  i !== 0 ? `1px solid ${Colors.DARK_GRAY}` : undefined,
+              }}
+            >
+              <div
+                style={{
+                  display: 'flex',
+                  gap: '0.5rem',
+                }}
+              >
+                <div style={{ flex: 1, textAlign: 'right' }}>
```
kshen0, line 321: Right aligned text

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+              style={{
+                padding: '0.375rem 0.5rem',
+                borderTop: `1px solid ${Colors.LIGHT_GRAY}`,
+              }}
+            >
+              <div
+                style={{
+                  display: 'flex',
+                  gap: '0.5rem',
+                  paddingTop: '0.25rem',
```
kshen0, line 432: Added padding to vertically center

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+                }}
+              >
+                <strong
+                  style={{ flex: 1, textAlign: 'right', marginTop: '-0.25rem' }}
```
kshen0, line 436: Right aligned and added marginTop

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+                    {electionStrings.contestOptionLabel(option)}
+                  </DualLanguageText>
+                </strong>
+                <Bubble
+                  optionInfo={{
+                    type: 'option',
+                    contestId: contest.id,
+                    optionId: option.id,
+                  }}
+                />
```
kshen0, line 448: `Bubble` moved to after contest options

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+    case 'candidate':
+      return <CandidateContest election={election} contest={contest} />;
+    case 'yesno':
+      return <BallotMeasureContest contest={contest} />;
+    default:
+      return throwIllegalValue(contest);
+  }
+}
+
+async function BallotPageContent(
```
kshen0, line 475: This is entirely copy + pasted from the default template. I considered adding a parameter to the default template's `BallotPageContent` to take a `Contest` component, which I may still do, but this was the fastest to unblock.

libs/hmpb/vitest.config.ts
```diff
@@ -11,6 +11,8 @@ export default defineConfig({
       exclude: [
         // tested by src/preview.test.ts, but no coverage is collected
         'src/preview',
+        // tested in VxDesign, the only consumer, but coverage should be added here to reduce fragility
+        'src/ballot_templates/nh_ballot_template.tsx',
```
kshen0, line 15: https://github.com/orgs/votingworks/projects/23/views/85?pane=issue&itemId=94670983

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
arsalansufi: > I considered adding a parameter to the default template's `BallotPageContent` to take a `Contest` component

I do like this idea! While I agree with the decision to keep the NH and Vx ballot templates separate instead of branching within a single template, I do think that we should keep duplication to a minimum where possible. I think there's a very high chance that a future fix in one template accidentally doesn't make it to the other template. Not blocking for the training of course but would be good to adjust while context is fresh.

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -218,6 +243,374 @@ function BallotPageFrame({
...
+            contests: contestSectionsLeftToLayout.flat(),
+          },
+        }
+      : undefined;
+
+  return {
+    currentPageElement,
+    nextPageProps,
+  };
+}
```
arsalansufi, line 612: Very possibly already on your radar but given that the NH template will always have right-aligned bubbles now, I think we can probably remove the placeholder toggle and feature flag we added for left/right bubble alignment in https://github.com/votingworks/vxsuite/pull/5811

arsalansufi (approved):
LGTM for the training!

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
kshen0: Thanks for the reminder!

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
jonahkagan: The write-in area looks a little bit closer to the write-in line than the top of the box. That can sometimes be an issue when the ballot is skewed (as the line can overlap with the box). I might move it up a tiny bit. Though I'd be curious how it compares to the VX default ballot template - probably they should match.

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
jonahkagan: The approach I envisioned is that any component shared between templates goes into ballot_components.tsx to make it clear that they're shared.

I think figuring out the right abstractions will actually be a little tricky until we have a few templates and can see which parts differ, so I'm more ok with keeping copy-pasting as a starting point, even if it results in some duplication.

kshen0: I see. I'll open a follow up PR to move the components exported from `vx_default_ballot_template` to `ballot_components` while we decide how to handle `*Content` components.

PR #5853, opened 2025-01-22
VxDesign: Refactor to find precinct split from ballot style in ballot preview

jonahkagan:
## Overview
If a precinct has splits, VxDesign creates a separate ballot style for each split. When previewing a particular ballot style, we know the precinct ID and the ballot style ID, so we can figure out the relevant split ID (if there is one) from that information. If the precinct has splits, then we look up which split this ballot style corresponds to.

This refactor helps set up for factoring out the logic to add the NH-specific ballot props in all of the other places where we render ballots.

## Testing Plan
Updated existing tests

## Checklist

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -16,6 +16,7 @@ import express, { Application } from 'express';
 import {
   assertDefined,
   DateWithoutTime,
+  find,
   groupBy,
   iter,
   ok,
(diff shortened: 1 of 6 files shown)


apps/design/backend/src/app.ts
```diff
@@ -330,20 +330,23 @@ function buildApi({ workspace, translator }: AppContext) {
...
-          );
-          extraProps.electionTitleOverride = split.electionTitleOverride;
-          extraProps.clerkSignatureImage = split.clerkSignatureImage;
-          extraProps.clerkSignatureCaption = split.clerkSignatureCaption;
-        }
+        const { splitId } = find(
+          ballotStyle.precinctsOrSplits,
+          (p) => p.precinctId === input.precinctId
+        );
+        const split = find(precinct.splits, (s) => s.id === splitId);
```
kshen0, line 344: Consider factoring out to a helper like `getSplitForBallotStyleAndPrecinct`? I think it'd be helpful to readers to encapsulate the logic.

kshen0 (approved):

apps/design/backend/src/app.ts
jonahkagan: Good idea, doing this in the follow up PR that refactors this code

PR #5865, opened 2025-01-22
VxDesign: Factor out selecting template and creating ballot props

jonahkagan:
## Overview

Factors out logic to select the ballot template based on the state and create the ballot props for that template (including any custom props needed). The new function `selectTemplateAndCreateBallotProps` creates props for every possible ballot, allowing consumers to filter down to the props they need in their specific situation.

I also removed the functions from each template file that just render enough ballots to create an election definition and replaced it with a function `renderMinimalBallotsToCreateElectionDefinition`, which encapsulates that bit of logic but is agnostic to which template is used.

## Testing Plan
Add automated tests (TODO)

## Checklist

### DIFF
--- a/apps/design/backend/src/app.test.ts
+++ b/apps/design/backend/src/app.test.ts
@@ -46,10 +46,8 @@ import {
 import { countObjectLeaves, getObjectLeaves } from '@votingworks/test-utils';
 import {
   BallotMode,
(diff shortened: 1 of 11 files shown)


jonahkagan:
Closing and folding into a different PR for v3-compatible ballots

PR #5874, opened 2025-01-22
Refine VxDesign ballot order info screen

arsalansufi:
## Overview

This PR refines the VxDesign ballot order info screen. It specifically:

- Better aligns inputs and helper text
- Renames the "Paper Color for Polling Place Ballots" field "Paper Color for Ballots" - the values entered into this field will be used for both precinct (polling place) and absentee ballots

| Before | After |
| - | - |
| before | after |

## Testing Plan

- [x] Tested manually
- [x] Updated automated tests

## Checklist

### DIFF
--- a/apps/design/frontend/src/ballot_order_info_screen.tsx
+++ b/apps/design/frontend/src/ballot_order_info_screen.tsx
@@ -16,8 +16,22 @@ import { getElection, updateBallotOrderInfo } from './api';
 import { Form, FormActionsRow, InputGroup } from './layout';
 import { ElectionNavScreen } from './nav_screen';
 
+export const Input = styled.input`
+  width: 25rem;
+`;
+
 export const Annotation = styled.div`
+  line-height: 1.25rem;
   margin-top: -1rem;
+  width: 25rem;
+`;
+
+export const CheckboxButtonContainer = styled.div`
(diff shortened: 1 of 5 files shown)


apps/design/frontend/src/ballot_order_info_screen.tsx
```diff
@@ -16,8 +16,22 @@ import { getElection, updateBallotOrderInfo } from './api';
...
   margin-top: -1rem;
+  width: 25rem;
+`;
+
+export const CheckboxButtonContainer = styled.div`
+  width: 25rem;
+
+  button {
+    width: 100%;
+  }
```
jonahkagan, line 34: I'm fine with this approach, but maybe there's also a slightly more concise way to set style rules to control the width on `Form` and then make all the inputs and annotations full-width?

jonahkagan (approved):

apps/design/frontend/src/ballot_order_info_screen.tsx
arsalansufi: Good catch! Adjusted ✅ https://github.com/votingworks/vxsuite/pull/5874/commits/349e0d7119022cd2b93a2c0f8fd8fec5dda46f3d

PR #5877, opened 2025-01-23
VxDesign: Assign ballot template to each election

jonahkagan:
## Overview

Task: https://github.com/votingworks/vxsuite/issues/5702

To support towns that require a v3-compatible ballot vs a v4 ballot, adds a ballot template id field to the election in the database. We default all new elections to the NH ballot template. When loading an election, we pick a template based on the `Election.state` field. Internal VX users can change the template for an election via a dropdown on the Export screen (temporarily placed there until we have internal tools).

## Demo Video or Screenshot

https://github.com/user-attachments/assets/412ad5fc-98e9-43c5-abb4-d1f8da64a820

## Testing Plan
Updated existing automated tests

## Checklist

### DIFF
--- a/apps/design/backend/src/ballots.ts
+++ b/apps/design/backend/src/ballots.ts
@@ -73,6 +73,7 @@ export function createBallotPropsForTemplate(
       return baseBallotProps;
 
     default:
+      /* istanbul ignore next - @preserve */
       throwIllegalValue(templateId);
   }
 }
--- a/apps/design/backend/src/ballots.ts
+++ b/apps/design/backend/src/ballots.ts
@@ -72,8 +72,9 @@ export function createBallotPropsForTemplate(
     case 'VxDefaultBallot':
       return baseBallotProps;
 
-    default:
+    default: {
       /* istanbul ignore next - @preserve */
       throwIllegalValue(templateId);
+    }
   }
 }
--- a/apps/design/backend/src/ballots.ts
+++ b/apps/design/backend/src/ballots.ts
@@ -1,0 +1,78 @@
+import { Election } from '@votingworks/types';
+import {
+  allBaseBallotProps,
+  BallotTemplateId,
+  BaseBallotProps,
+  NhBallotProps,
+} from '@votingworks/hmpb';
+import { find, throwIllegalValue } from '@votingworks/basics';
+import {
+  BallotStyle,
+  hasSplits,
+  normalizeState,
+  Precinct,
+  PrecinctSplit,
+  PrecinctWithSplits,
+  UsState,
+} from './types';
+
+function getPrecinctSplitForBallotStyle(
+  precinct: PrecinctWithSplits,
+  ballotStyle: BallotStyle
+): PrecinctSplit {
+  return find(precinct.splits, (split) =>
+    ballotStyle.precinctsOrSplits.some(
+      (precinctOrSplit) =>
+        precinctOrSplit.precinctId === precinct.id &&
+        precinctOrSplit.splitId === split.id
+    )
+  );
+}
+
+export function defaultBallotTemplate(state: string): BallotTemplateId {
+  switch (normalizeState(state)) {
+    case UsState.NEW_HAMPSHIRE:
+      return 'NhBallot';
+    case UsState.MISSISSIPPI:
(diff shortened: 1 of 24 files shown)


libs/hmpb/src/render_ballot.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    BALLOT_MODES.map((ballotMode) => ({ ballotMode })),
+  ]);
+  expect(allBallotProps).toHaveLength(expectedPropCombos.length);
+  for (const expectedPropCombo of expectedPropCombos) {
+    const expectedProps: BaseBallotProps = { ...expectedPropCombo, election };
+    expect(allBallotProps).toContainEqual(expectedProps);
+  }
+  for (const actualProps of allBallotProps) {
+    expect(actualProps.watermark).toBeUndefined();
+  }
```
kshen0, line 50: I love the coverage but I usually have 2 reservations with completely generated tests like this:

1. Makes it harder for readers to understand what is being tested (and thus update)
2. Increases risk of the test drifting in the future since generation of test case input and expectations are untested

I think this test is great for the situation because a large part of `allBaseBallotProps` is producing ballot props for all combinations of inputs, but could we add a second test for a single combination of `'ballotStyleId' | 'precinctId' | 'ballotType' | 'ballotMode'`? That would at least address concern #1 directly.

libs/hmpb/src/render_ballot.tsx
```diff
@@ -578,3 +585,53 @@ export async function renderAllBallotsAndCreateElectionDefinition<
...
   };
 }
+
+/**
+ * Creates a list of the {@link BaseBallotProps} for all possible ballots -
+ * every combination of ballot style, precinct, ballot type (precinct/absentee),
+ * and ballot mode (official/test/sample).
+ */
+export function allBaseBallotProps(election: Election): BaseBallotProps[] {
+  const ballotTypes = [BallotType.Precinct, BallotType.Absentee];
```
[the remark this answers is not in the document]
jonahkagan: I don't remember exactly why, but we don't need to produce provisional ballots for these elections.

libs/hmpb/src/render_ballot.test.tsx
jonahkagan: Added: 7e3a145091d5d3678979e9d94a208a9051edf937

apps/design/backend/src/ballots.ts
```diff
@@ -73,6 +73,7 @@ export function createBallotPropsForTemplate(
       return baseBallotProps;
 
     default:
+      /* istanbul ignore next - @preserve */
```
eventualbuddha, line 76: This won't be honored unless it's within a block statement, so do `default: { … }`.

apps/design/backend/src/types.test.ts
```diff
@@ -52,3 +52,24 @@ test('convertToVxfBallotStyle()', () => {
...
+    },
+    {
+      input: ['ms', 'MS', 'Mississippi', 'MISSISSIPPI'],
+      expectedState: UsState.MISSISSIPPI,
+    },
+    {
+      input: ['State of Hamilton'],
+      expectedState: UsState.UNKNOWN,
+    },
+  ])('normalizes state string: $state', ({ input, expectedState }) => {
```
eventualbuddha, line 70:
```suggestion
  ])('normalizes state string: $expectedState', ({ input, expectedState }) => {
```

libs/hmpb/src/render_ballot.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return arrays.reduce(
+    (acc, array) =>
+      acc.flatMap((accItem) =>
+        array.map((arrayItem) => ({ ...accItem, ...arrayItem }))
+      ),
+    [{}]
+  ) as T[];
+}
+
+test('allBaseBallotProps creates props all possible ballots for an election', () => {
```
eventualbuddha, line 25: Maybe `creates props for all possible ballots for an election`? Right now it feels weirdly phrased.

eventualbuddha (approved):

kshen0 (approved):

PR #5878, opened 2025-01-23
move shared components to ballot_components.tsx

kshen0:
## Overview

Moves shared components from the default HMPB template to `ballot_components.tsx` and updates imports.

## Demo Video or Screenshot

N/A refactor

## Testing Plan

- [x] Verify ballots still render with preview tool

## Checklist

### DIFF
--- a/libs/hmpb/src/ballot_components.tsx
+++ b/libs/hmpb/src/ballot_components.tsx
@@ -226,18 +226,17 @@ export function Bubble({
 
 export function BubbleWrapper({
   optionInfo,
-  isWriteIn,
+  style = {},
 }: {
   optionInfo: OptionInfo;
-  isWriteIn?: boolean;
+  style: React.CSSProperties;
 }): React.ReactElement {
   return (
     <div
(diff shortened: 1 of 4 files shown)


libs/hmpb/src/ballot_components.tsx
```diff
@@ -196,6 +224,27 @@ export function Bubble({
...
+  optionInfo: OptionInfo;
+  isWriteIn?: boolean;
+}): React.ReactElement {
+  return (
+    <div
+      style={{
+        display: 'flex',
+        alignItems: 'center',
+        // Match line-height of text to align bubble to center of first line of option label or write-in candidate name
+        height: isWriteIn ? '1.25rem' : '1.2rem',
```
jonahkagan, line 240: It seems like a bit of a leaky abstraction to have this height defined here. Maybe it should be defined by the consumer (either via `styled-components` or a `style` prop)?

jonahkagan (approved):
Thanks for tackling this change!

PR #5898, opened 2025-01-27
VxDesign: Sentry instrumentation

jonahkagan:
## Overview

Adds Sentry instrumentation to both server and client. Captures uncaught exceptions and also enables web transaction tracing (useful for performance profiling).

## Demo Video or Screenshot
N/A

## Testing Plan
Manual testing

## Checklist

### DIFF
--- a/apps/design/frontend/src/index.tsx
+++ b/apps/design/frontend/src/index.tsx
@@ -1,7 +1,19 @@
+import * as Sentry from '@sentry/browser';
 import React from 'react';
 import { createRoot } from 'react-dom/client';
 import { assert } from '@votingworks/basics';
 import { App } from './app';
+
+/* istanbul ignore next - @preserve */
+if (process.env.NODE_ENV === 'production') {
+  Sentry.init({
+    // Note: This isn't a secret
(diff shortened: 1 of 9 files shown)


apps/design/frontend/src/index.tsx
```diff
@@ -1,8 +1,19 @@
+import * as Sentry from '@sentry/browser';
 import React from 'react';
 import { createRoot } from 'react-dom/client';
 import { assert } from '@votingworks/basics';
 import { App } from './app';
 
+if (process.env.NODE_ENV === 'production') {
```
jonahkagan, line 7: @kofi-q Do you know if `process.env.NODE_ENV` will be available in client-side code? I saw that we use it in some other places, but couldn't find any configuration that made me sure

arsalansufi: I _believe_ it's set here for the frontend build: https://github.com/votingworks/vxsuite/blob/5cd82a14263f62d89753950ddfde1d2570445a96/Dockerfile.vxdesign#L70-L71

kofi-q: +1 - `vite` should be passing that through to the bundle, so it's safe to use here

kofi-q (approved):
✅

Issue #5895, opened 2025-01-27
Backend validation for basic form data correctness


eventualbuddha:

eventualbuddha:
Not staging the work this way.

PR #5911, opened 2025-01-29
feat(design): allow removing images

eventualbuddha:
## Overview

Refs #5895

Allows removing any images. This is mostly to allow removing images where the images are optional, such as clerk signatures.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/c2cdf621-faa1-4ea4-b368-be78b7cf4af4

## Testing Plan

Added an automated test, manually tested with the election seal image and clerk signature.

## Checklist

### DIFF
--- a/apps/design/frontend/src/image_input.tsx
+++ b/apps/design/frontend/src/image_input.tsx
@@ -1,6 +1,7 @@
 import { Buffer } from 'node:buffer';
 import sanitizeHtml from 'sanitize-html';
 import {
+  Button,
   Callout,
   FileInputButton,
   FileInputButtonProps,
@@ -209,7 +210,7 @@ export function ImageInput({
   minHeightPx,
 }: {
   value?: string;
-  onChange: (value: string) => void;
+  onChange: (value?: string) => void;
   buttonLabel: string;
   disabled?: boolean;
   className?: string;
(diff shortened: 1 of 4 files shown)


jonahkagan:
Oh, I see that this applies to multiple forms. I do think it makes sense in the precinct split case, since we want to make the image optional. Maybe we should have the Remove button only show up for optional image fields?

apps/design/frontend/src/image_input.tsx
```diff
@@ -177,6 +177,19 @@ export function ImageInput({
...
+        <Button
+          onPress={() => onChange(undefined)}
+          disabled={disabled}
+          variant="danger"
+          icon="Delete"
+          style={{
+            marginLeft: '0.5rem',
+          }}
+        >
+          Remove Image
```
jonahkagan, line 190: I think this text should be a prop, similar to `buttonLabel`, so they can be consistent with each other. Or we standardize them to Upload Image and Remove Image

eventualbuddha: 96e4d3d8dd

arsalansufi:
Rebased and made some simple tweaks, including:
- Not displaying a remove option when an image is required
- More consistent add/remove text
- Using less strong styling (fill="outline") to avoid conflicting with even more destructive delete operations on pages

https://github.com/votingworks/vxsuite/pull/5911/commits/8256278513504bedd8b24329e545aa4d95f02729

screenshot

arsalansufi (approved):
LGTM! Made some edits (see the above comment) and happy for us to make further edits but otherwise merging for initial launch

PR #5912, opened 2025-01-29
feat(design): improve form a11y

### LINKED ISSUE #5788: [Form Entry] Enter/submit key in forms is broken
We'll need to override and block the default submit action to prevent it from trying to reload the page.

eventualbuddha:
Closes #5788

Found it annoying to not be able to submit forms with Enter while working on #5895.

### DIFF
--- a/apps/design/frontend/src/ballot_order_info_screen.tsx
+++ b/apps/design/frontend/src/ballot_order_info_screen.tsx
@@ -41,7 +41,7 @@ function BallotOrderInfoForm({
     useState<BallotOrderInfo>(savedBallotOrderInfo);
   const updateBallotOrderInfoMutation = updateBallotOrderInfo.useMutation();
 
-  function onSaveButtonPress() {
+  function onSubmit() {
     updateBallotOrderInfoMutation.mutate(
       {
         electionId,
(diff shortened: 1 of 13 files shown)


apps/design/frontend/src/ballot_order_info_screen.tsx
```diff
@@ -52,7 +52,7 @@ function BallotOrderInfoForm({
   }
 
   return (
-    <StyledForm>
+    <StyledForm onKeyUp={(e) => e.key === 'Enter' && onSaveButtonPress()}>
```
kofi-q, line 55: This can be a later iteration, but usually better for a11y to:

1. use a `<button type="submit">` variant of the button element in the form and
2. override the `onSubmit` prop of the `<form>` element here and perform the action while also doing an `event.preventDefault()` to stop the browser from performing a default submit.

Enables default integration with things like mobile devices and some a11y devices.

eventualbuddha: Thanks @kofi-q. I changed it to do as you suggested, including using `<button type="reset">` for the Cancel button.

apps/design/frontend/src/tabulation_screen.test.tsx
```diff
@@ -66,17 +66,17 @@ test('mark thresholds', async () => {
 
   userEvent.click(screen.getByRole('button', { name: 'Edit' }));
 
-  // Due to some weirdness with the tests, we can't clear the input before
-  // typing, so we have to just append
-  userEvent.type(definiteInput, '9');
-  userEvent.type(marginalInput, '8');
```
eventualbuddha, line 72: This was actually causing invalid data to be entered (because `step={0.1}`) and so the form was not submitting in the tests. Now it properly prevents submission, so I had to change this to backspace the "7" before typing the "8".

kofi-q (approved):
🎉  thanks for the extra-mile fix!

PR #5915, opened 2025-01-29
feat(design): validate Election Info screen inputs

eventualbuddha:
## Overview

Adds both client and server-side validation. The client validation is fairly simple, just trimming the text and ensuring something non-blank is set. The server-side validation does a tiny bit more, and for now just crashes if it fails.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/7cb28a3c-efc1-49e5-a618-1964de451eca

## Testing Plan
Mostly manual testing, plus a couple new automated tests.

## Checklist

### DIFF
--- a/apps/design/backend/src/app.test.ts
+++ b/apps/design/backend/src/app.test.ts
@@ -20,7 +20,6 @@ import {
   BallotType,
   CandidateContest,
   DEFAULT_SYSTEM_SETTINGS,
-  DistrictId,
   Election,
   ElectionStringKey,
   SystemSettings,
@@ -33,6 +32,7 @@ import {
   ElectionPackageFileName,
   unsafeParse,
   ElectionIdSchema,
+  DistrictIdSchema,
 } from '@votingworks/types';
 import {
   BooleanEnvironmentVariableName,
@@ -148,7 +148,7 @@ test('CRUD elections', async () => {
       ballotStyles: [],
       contests: [],
       county: {
-        id: '',
+        id: 'county-id',
         name: '',
       },
       date: expect.any(DateWithoutTime),
@@ -187,7 +187,7 @@ test('CRUD elections', async () => {
     election2Definition.election.precincts.map((vxfPrecinct) => ({
       id: vxfPrecinct.id,
       name: vxfPrecinct.name,
-      districtIds: ['district-1' as DistrictId],
+      districtIds: [unsafeParse(DistrictIdSchema, 'district-1')],
     }));
   const expectedBallotStyles: BallotStyle[] = generateBallotStyles({
     ballotLanguageConfigs: election2.ballotLanguageConfigs,
@@ -234,6 +234,56 @@ test('CRUD elections', async () => {
   await apiClient.deleteElection({ electionId });
 
   expect(await apiClient.listElections()).toEqual([election2]);
+});
+
+test('update election info', async () => {
+  const { apiClient } = await setupApp();
+  const electionId = unsafeParse(ElectionIdSchema, 'election-1');
+  (await apiClient.createElection({ id: electionId })).unsafeUnwrap();
+
+  await apiClient.updateElectionInfo({
+    electionId,
+    // trim text values
+    title: '   Updated Election  ',
+    jurisdiction: '   New Hampshire   ',
+    state: '   NH   ',
+    seal: '\r\n<svg>updated seal</svg>\r\n',
+    type: 'primary',
+    date: new DateWithoutTime('2022-01-01'),
+  });
+
+  const record = await apiClient.getElection({ electionId });
+  const { election } = record;
+  expect(election.title).toEqual('Updated Election');
+  expect(election.type).toEqual('primary');
+  expect(election.state).toEqual('NH');
+  expect(election.county.name).toEqual('New Hampshire');
+  expect(election.seal).toEqual('\r\n<svg>updated seal</svg>\r\n');
+  expect(election.date).toEqual(new DateWithoutTime('2022-01-01'));
+
+  // empty string values do overwrite existing values
+  await apiClient.updateElectionInfo({
+    electionId,
+    type: 'primary',
+    title: '',
+    jurisdiction: '  ',
+    state: '',
+    seal: '',
+    date: new DateWithoutTime('2022-01-01'),
+  });
+
+  expect(await apiClient.getElection({ electionId })).toEqual(
+    expect.objectContaining({
+      election: expect.objectContaining({
+        title: '',
+        county: expect.objectContaining({
+          name: '',
+        }),
+        state: '',
+        seal: '',
+      }),
+    })
+  );
 });
 
 test('Updating contests with candidate rotation', async () => {
--- a/apps/design/backend/src/app.test.ts
+++ b/apps/design/backend/src/app.test.ts
@@ -49,7 +49,11 @@ import {
   mockCloudTranslatedText,
   readElectionPackageFromFile,
 } from '@votingworks/backend';
-import { countObjectLeaves, getObjectLeaves } from '@votingworks/test-utils';
+import {
+  countObjectLeaves,
+  getObjectLeaves,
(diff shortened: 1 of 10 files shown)


arsalansufi:
As we pre-create elections for every town, we'll be pre-populating an election title, e.g., "March 11, 2025 Election". With this change in place, we'll need to either 1) pre-fill out all form fields, including a placeholder seal, or 2) use a script to bypass the validation to just populate some of the fields.

Not a problem either way! We should optimize for NH users and not our internal bootstrapping, just noting

apps/design/backend/src/app.test.ts
```diff
@@ -236,6 +236,56 @@ test('CRUD elections', async () => {
...
+  const record = await apiClient.getElection({ electionId });
+  const { election } = record;
+  expect(election.title).toEqual('Updated Election');
+  expect(election.type).toEqual('primary');
+  expect(election.state).toEqual('NH');
+  expect(election.county.name).toEqual('New Hampshire');
+  expect(election.seal).toEqual('\r\n<svg>updated seal</svg>\r\n');
+  expect(election.date).toEqual(new DateWithoutTime('2022-01-01'));
+
+  // empty string values do overwrite existing values
```
jonahkagan, line 264: What's the use case for this functionality if all fields are required?

apps/design/frontend/src/api.ts
```diff
@@ -67,6 +67,38 @@ export const getElection = {
...
+    );
+  },
+} as const;
+
+async function invalidateElectionQueries(
+  queryClient: QueryClient,
+  electionId: ElectionId
+) {
+  await queryClient.invalidateQueries(getElection.queryKey(electionId));
+  await queryClient.invalidateQueries(getElectionInfo.queryKey(electionId));
```
jonahkagan, line 87: Does `listElections` also need to be invalidated?

apps/design/frontend/src/election_info_screen.tsx
```diff
@@ -107,15 +117,18 @@ function ElectionInfoForm({
           type="text"
           value={electionInfo.title}
           onChange={onInputChange('title')}
+          onBlur={onInputBlur('title')}
           disabled={!isEditing}
+          autoComplete="off"
+          required
```
jonahkagan, line 123: Nice, I like using the built-in validation methods

apps/design/frontend/src/election_info_screen.tsx
```diff
@@ -1,88 +1,98 @@
...
+  savedElectionInfo: ElectionInfo;
 }): JSX.Element {
   const [isEditing, setIsEditing] = useState(
     // Default to editing for newly created elections
-    hasBlankElectionInfo(savedElection)
+    hasBlankElectionInfo(savedElectionInfo)
   );
-  const [electionInfo, setElectionInfo] = useState<ElectionInfo>(savedElection);
-  const updateElectionMutation = updateElection.useMutation();
+  const [editedElectionInfo, setEditedElectionInfo] = useState<ElectionInfo>();
```
jonahkagan, line 45: What made you change from using the saved election info as the default state?

jonahkagan (commented):
Looks great overall! Should we shortcut by just doing the frontend validation for the rest of the forms?

apps/design/frontend/src/election_info_screen.tsx
eventualbuddha: I think I'd wanted to get rid of `isEditing`, but then found I needed to keep it. I'll try to revert this.

apps/design/frontend/src/api.ts
eventualbuddha: Going back to the main elections list screen after editing does seem to reflect the changes, but I can invalidate it here just in case.

apps/design/backend/src/app.test.ts
eventualbuddha: This was left over from before we decided to require everything. Will fix.

eventualbuddha: [9fd6bd3](https://github.com/votingworks/vxsuite/pull/5915/commits/9fd6bd3c4b3f21a67c38f5e5b5575e69e4679d6e)

apps/design/frontend/src/api.ts
eventualbuddha: [7576da8](https://github.com/votingworks/vxsuite/pull/5915/commits/7576da8e8fc73c04d023e95cb0d12619f01921b7)

apps/design/frontend/src/election_info_screen.tsx
eventualbuddha: [0236383](https://github.com/votingworks/vxsuite/pull/5915/commits/0236383d9f1ee4aa68ce727969824e285a0ede86)

apps/design/backend/src/app.test.ts
```diff
@@ -261,28 +265,19 @@ test('update election info', async () => {
...
+      apiClient.updateElectionInfo({
+        electionId,
+        type: 'primary',
         title: '',
-        county: expect.objectContaining({
-          name: '',
-        }),
+        jurisdiction: '  ',
         state: '',
         seal: '',
```
jonahkagan, line 277: I think we're still accepting an empty seal based on my read of the schema

apps/design/frontend/src/election_info_screen.tsx
```diff
@@ -1,95 +1,106 @@
...
-  const [electionInfo, setElectionInfo] = useState<ElectionInfo>(savedElection);
-  const updateElectionMutation = updateElection.useMutation();
+  const [editedElectionInfo, setEditedElectionInfo] =
+    useState(savedElectionInfo);
+  const updateElectionInfoMutation = updateElectionInfo.useMutation();
   const deleteElectionMutation = deleteElection.useMutation();
   const history = useHistory();
+  const electionInfo = isEditing
+    ? editedElectionInfo ?? savedElectionInfo
+    : savedElectionInfo;
```
jonahkagan, line 59: I think we can also remove this intermediate variable now and just use `editedElectionInfo` - similar to what we had before

apps/design/frontend/src/election_info_screen.tsx
```diff
@@ -1,95 +1,106 @@
...
+    setEditedElectionInfo(savedElectionInfo);
+  }
 
   function onSubmit() {
-    updateElectionMutation.mutate(
-      { electionId, election: { ...savedElection, ...electionInfo } },
-      { onSuccess: () => setIsEditing(false) }
-    );
+    updateElectionInfoMutation.mutate(electionInfo, {
+      onSuccess: resetForm,
```
jonahkagan, line 68: I'm a little worried that calling `resetForm` here will call `setEditedElectionInfo(savedElectionInfo);` with the old `savedElectionInfo` and overwrite the changes the user made. Maybe that's not actually happening, but seems safer to not do this and revert to the logic we had before. Unless I'm misreading

apps/design/backend/src/app.test.ts
eventualbuddha: Yeah, we do accept empty string as a seal value. Arguably we shouldn't, but I don't think we're going to change it right now.

jonahkagan: Ah forgot about that, cool!

apps/design/frontend/src/election_info_screen.tsx
eventualbuddha: [ad7f4c3](https://github.com/votingworks/vxsuite/pull/5915/commits/ad7f4c34da68ccf1d4299e342afc47c8a34958fb)

apps/design/frontend/src/election_info_screen.tsx
eventualbuddha: [ad7f4c3](https://github.com/votingworks/vxsuite/pull/5915/commits/ad7f4c34da68ccf1d4299e342afc47c8a34958fb)

jonahkagan (approved):

Issue #5917, opened 2025-01-29
Misc tweaks to NH interface for first ballot production run


arsalansufi:
- [x] Hide/disable create election button
- [x] Hide/disable delete election button
- [x] Lock basics of town/school precinct split setup

jonahkagan:
Tackling the first two points in #5919

Issue #5868, opened 2025-01-22
Hide ballot length options greater than 14 inches from NH users, with note to reach out to us if a larger size is preferred


arsalansufi:
NH users shouldn't be able to see the other options, just a note to reach out to us if a larger size is preferred. Vx users should be able to select any options.

PR #5919, opened 2025-01-29
VxDesign: Hide some functionality unless user is a VX support user

jonahkagan:
## Overview

Split the features context into user features vs election features:
- User features are features that should be enabled based on the user's organization - i.e. differentiating between what VX support users can do and what election officials can do.
- Election features are features that should be enabled based on the organization of the election currently being viewed. VX support users and election officials should all have the same functionality for these features when viewing a specific election.

This model of feature flagging means we won't need any sort of feature for support users to "log in as" election official users.

Uses this feature flagging to implement a few tasks:
- https://github.com/votingworks/vxsuite/issues/5810
- https://github.com/votingworks/vxsuite/issues/5868
- https://github.com/votingworks/vxsuite/issues/5917 (first two points)

## Demo Video or Screenshot

https://github.com/user-attachments/assets/17f4e7db-16a7-4150-93ea-e139e9aab3d3

## Testing Plan
Manual testing - skipping automated tests since auth integration broke them all

## Checklist

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -585,11 +585,6 @@ function buildApi({ auth, workspace, translator }: AppContext) {
     },
 
     /* istanbul ignore next - @preserve */
-    getVotingWorksOrgId(): string {
-      return votingWorksOrgId();
-    },
-
-    /* istanbul ignore next - @preserve */
     getAllOrgs(): Promise<Org[]> {
       return auth.allOrgs();
     },
@@ -664,6 +659,7 @@ export function buildApp(context: AppContext): Application {
     const userInfo: ReturnType<Api['getUser']> = {
       orgId: user.org_id,
       orgName: org.displayName,
+      isVotingWorksUser: user.org_id === votingWorksOrgId(),
     };
 
     res.set('Content-type', 'application/json');
--- a/apps/design/frontend/src/features_context.tsx
+++ b/apps/design/frontend/src/features_context.tsx
@@ -29,6 +29,10 @@ enum UserFeature {
    * Allow the user to create an election.
    */
   CREATE_ELECTION = 'CREATE_ELECTION',
+  /**
+   * Allow the user to delete an election.
+   */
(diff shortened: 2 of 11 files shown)


jonahkagan:
@kshen0 tagging you for initial review of implementation, even though it doesn't have real auth yet

apps/design/backend/src/app.ts
```diff
@@ -126,6 +126,10 @@ function buildApi({ workspace, translator }: AppContext) {
   const { store } = workspace;
 
   return grout.createApi({
+    getUser(): { orgId: string } {
+      return { orgId: 'VX_ORG_ID' };
+    },
```
jonahkagan, line 131: placeholder

apps/design/backend/src/store.ts
```diff
@@ -144,6 +145,7 @@ function hydrateElection(row: {
   ).unsafeUnwrap();
 
   return {
+    orgId: 'VX_ORG_ID',
```
jonahkagan, line 148: placeholder

jonahkagan:
@kshen0 I'm gonna rework this now that the auth changes are merged, so no need to review

apps/design/frontend/src/features_context.tsx
```diff
@@ -1,79 +1,181 @@
...
+  );
+}
+
 interface FeaturesProviderProps {
   children: React.ReactNode;
-  electionId?: ElectionId;
+  electionId: ElectionId;
 }
 
 export function FeaturesProvider({
```
kshen0, line 153: Might be helpful to copy or adapt the comments from `UserFeature` and `ElectionFeature` enums to explain when to use `FeaturesProvider` vs `UserFeaturesProvider`. Something simple like "Use `UserFeaturesProvider` for screens where there isn't a specific election" could help differentiate at a glance in vscode autocomplete.

kshen0 (approved):

apps/design/frontend/src/features_context.tsx
jonahkagan: Good idea!

PR #5923, opened 2025-01-30
[VxDesign] Use user org info client-side

kofi-q:
## Overview

Adding API for user org info and using client-side for querying and creating elections.
- Add `org_id` field to the `elections` table
- Read and write `org_id` when retrieving/creating elections
- "Get" requests include a `user` field in the input, containing the user's org ID - this is used for filtering election query results
- "Set" requests include both the `user` field and an `orgId` field in the input - the former is used to check access to the latter and the latter is written to the DB election record.
- Moving "known org ID"s to env vars, since they'll differ for each deploy env.
- Also adding an org picker in the "create election" flow client-side, when logged in to the known VotingWorks org.

### Logged in to VotingWorks Org:

https://github.com/user-attachments/assets/86c065ab-f130-46b1-b01f-452f2b37eccb

### Logged in to non-VotingWorks Org (with creation enabled):

https://github.com/user-attachments/assets/5a04d5c4-ea7c-4bed-8d23-ee37a5c653bc

## Testing Plan
- Manual for now

## Checklist


apps/design/backend/src/env.d.ts
```diff
@@ -1,13 +1,15 @@
...
   export interface ProcessEnv {
+    readonly AUTH_ENABLED?: string;
     readonly AUTH0_CLIENT_ID?: string;
     readonly AUTH0_ISSUER_BASE_URL?: string;
     readonly AUTH0_SECRET?: string;
     readonly BASE_URL?: string;
     readonly CI?: string;
     readonly DEPLOY_ENV: 'development' | 'staging' | 'production';
     readonly NODE_ENV: 'development' | 'production' | 'test';
+    readonly ORG_ID_VOTINGWORKS?: string;
```
jonahkagan, line 12: For feature flagging, we're gonna need both the VotingWorks and NH ideas available in the frontend. Or I guess we could assume any org that's not VotingWorks is NH. I was imagining we just hardcode the org IDs into the codebase for now. What do you think makes the most sense? Env vars for both?

apps/design/frontend/src/app.tsx
```diff
@@ -69,6 +76,23 @@ function ElectionScreens(): JSX.Element {
...
+      <Screen>
+        <Main centerChild>
+          <LoadingAnimation />
+        </Main>
+      </Screen>
+    );
+  }
+
+  return children;
+}
```
jonahkagan, line 94: Is it generally slow enough loading the user that we see the loading animation?

jonahkagan (approved):

apps/design/backend/src/env.d.ts
kofi-q: Yeah, I was thinking we can just assume all orgs are NH for now, unless explicitly special-cased (VotingWorks, maybe SLI).
Only using an env var here because the org ID will be different across dev/staging/prod Auth0, but we could go with hardcoded prog IDs for the NH orgs, if that's easier - might just be tough keeping them in sync with what we create in Auth0 over the next few days

apps/design/frontend/src/app.tsx
kofi-q: Yeah, good question - it was, occasionally, but might just be a dev thing. Might end up being better to leave it out in prod to avoid flicker; wasn't sure what latency spikes/user connection quirks to expect in practice, but those should be out of the ordinary anyway - will make a note to revisit

apps/design/backend/src/env.d.ts
jonahkagan: Ok, I'll try building on top of what you have here. I think it should work, but we can see how it goes.

Issue #5900, opened 2025-01-28
Support bootstrapping a local VxDesign with a prod DB snapshot


arsalansufi:
The production database, registered under the production app in Heroku, can be found here: https://dashboard.heroku.com/apps/vxdesign-production/resources

There's a placeholder section for the bootstrapping process in this doc: https://docs.google.com/document/d/1jPkf9qqBJIRobthKh8cRSs4lQGAt1hNROw4FoaGn7Mc/edit?tab=t.0#heading=h.t3v76acya0n

PR #5924, opened 2025-01-30
VxDesign db restore script and downgrade to Postgres 16.6

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/5900

Adds a script and instructions for restoring a dev db from snapshot.

## Demo Video or Screenshot

N/a

## Testing Plan

- exported from prod and restored locally

### DIFF
--- a/apps/design/backend/README.md
+++ b/apps/design/backend/README.md
@@ -3,38 +3,3 @@
 This backend is used by the [VxDesign frontend](../frontend) and isn't intended
 to be run on its own. The best way to develop on the backend is by running the
 frontend.
-
-## PostgreSQL Version
-
-VxDesign expects PostgreSQL version 16.6. Debian 12
-[defaults](https://packages.debian.org/bookworm/postgresql) to PostgreSQL
-version 15.
-
-To upgrade, configure `apt` to use the PostgreSQL Apt Repository. From the
-official [PostgreSQL docs](https://www.postgresql.org/download/linux/debian/)
-run
-
-```bash
-sudo apt install -y postgresql-common
-sudo /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh
-```
-
-then install `postgresql-16`:
-
-```bash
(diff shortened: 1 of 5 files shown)


apps/design/backend/scripts/restore_db_backup.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+BACKUP=${1:-latest.dump}
+PGPASSWORD=design dropdb -U design -h localhost design
+PGPASSWORD=design createdb -U design -h localhost design
```
kshen0, line 4: Explicitly dropping the db causes warnings like

```
pg_restore: error: could not execute query: ERROR:  relation "public.elections" does not exist
```

When running `pg_restore` because the dump doesn't assume the db has been dropped. We drop it explicitly to reduce the chance of existing db state causing problems with restore.

apps/design/backend/README.md
```diff
@@ -3,3 +3,38 @@
 This backend is used by the [VxDesign frontend](../frontend) and isn't intended
 to be run on its own. The best way to develop on the backend is by running the
 frontend.
+
+## PostgreSQL Version
```
arsalansufi, line 7: Great instructions!

I know this may not be intuitive, but mind moving them to `apps/design/frontend`. In doing our README passes pre-cert, we decided to have the `apps/<app-name>/frontend` READMEs be the source of truth READMEs with dev tips and all, e.g., https://github.com/votingworks/vxsuite/tree/main/apps/central-scan/frontend.

And then every `apps/<app-name>/backend` README just points to the frontend, and includes nothing else.

And while you're at it, would you mind fixing this broken link in the VxDesign frontend README? I know it's been broken unrelated to this PR. Just remembered us talking about it a while back.

https://github.com/votingworks/vxsuite/blob/da20f07c856054ec18d01b540199e2aba2eb2aed/apps/design/frontend/README.md#L20

arsalansufi (approved):
Nice!!

Issue #5903, opened 2025-01-28
Handle "Contest is too tall to fit on page" error


arsalansufi:
This is one of the few errors that we can't catch with validation before rendering the ballot. Currently, it causes the app to "Something went wrong"

User action in response to this is to pick a larger ballot size

PR #5925, opened 2025-01-30
VxDesign: Show expected errors during ballot preview

jonahkagan:
## Overview

Closes: #5903 

Currently, there's just one type of expected error: when a contest is too tall to fit on the page.

To deliver this error to the frontend during ballot proofing, we catch the error within the ballot template so we know what contest was too long and pass it up the stack using a `Result`.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/8b8a2437-0c7c-4514-8a4e-faabfb0682bc

## Testing Plan
Manual test

## Checklist

### DIFF
--- a/apps/design/frontend/src/ballot_screen.tsx
+++ b/apps/design/frontend/src/ballot_screen.tsx
@@ -1,6 +1,6 @@
 import fileDownload from 'js-file-download';
 import { useParams } from 'react-router-dom';
-import { assertDefined, range } from '@votingworks/basics';
+import { assert, assertDefined, range } from '@votingworks/basics';
 import {
   getPrecinctById,
   getBallotStyle,
@@ -18,7 +18,7 @@ import styled from 'styled-components';
 import { z } from 'zod';
 import {
   Button,
-  Card,
+  Callout,
   H1,
   Icons,
   LinkButton,
@@ -299,6 +299,8 @@ export function BallotScreen(): JSX.Element | null {
             const ballotResult = getBallotPreviewPdfQuery.data;
 
             if (ballotResult.isErr()) {
+              const err = ballotResult.err();
+              assert(err.error === 'contestTooLong');
               return (
                 <Row
                   style={{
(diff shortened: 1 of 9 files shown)


apps/design/frontend/src/ballot_screen.tsx
```diff
@@ -307,10 +309,12 @@ export function BallotScreen(): JSX.Element | null {
...
                   }}
                 >
-                  <Card color="danger">
-                    Error:{' '}
-                    {ballotResult.err().message ?? 'Something went wrong'}
-                  </Card>
+                  <Callout color="danger" icon="Danger">
+                    <span>
+                      Contest &quot;{err.contest.title}&quot; was too long to
+                      fit on the page. Try a longer paper size.
```
kshen0, line 315: nit: if the user is already trying the longest paper size what are their options?

libs/hmpb/src/ballot_templates/vx_default_ballot_template.tsx
```diff
@@ -548,6 +550,17 @@ async function BallotPageContent(
...
+  const contestsLeftToLayout = contestSectionsLeftToLayout.flat();
+  if (
+    contests.length > 0 &&
+    contestsLeftToLayout.flat().length === contests.length
+  ) {
+    return err({
+      error: 'contestTooLong',
+      contest: contestsLeftToLayout[0],
+    });
+  }
```
kshen0, line 562: factor out to helper?

kshen0 (approved):

apps/design/frontend/src/ballot_screen.tsx
jonahkagan: Contact support. I figured they'd resort to that by default so didn't include in the message

libs/hmpb/src/ballot_templates/vx_default_ballot_template.tsx
jonahkagan: I feel like there's a larger factoring out needed for contest layout logic that would include this, but I don't have time to do it now and I don't think there's much to be gained just from factoring out these few lines, since they only make sense in context

PR #5926, opened 2025-01-30
VxDesign: Deploy worker

arsalansufi:
## Overview

The minimal changes necessary to actually deploy the VxDesign worker as part of our VxDesign Heroku deploy. In a separate PR, I'll include the changes to actually allow the worker to be used. Locally, the server and worker use the shared file system to pass files to each other, which naturally won't work in a deployed environment. In prod, I'll be using S3.

## Testing Plan

- [x] Tested by deploying my branch to Heroku


kofi-q (approved):
🎉

PR #5931, opened 2025-01-31
[VxDesign] DRAFT scripts to help with user/org management

kofi-q:
To run:
- Add relevant env vars to `.env.local`:
```sh
# Details in the 'vxdesign' tenant in the Auth0 dashboard
AUTH_ENABLED=TRUE
AUTH0_CLIENT_ID='xxxxx'
AUTH0_CLIENT_DOMAIN='vxdesign.us.auth0.com'
AUTH0_SECRET='xxxxx
```
- From `apps/design/backend`:
  - Create orgs with `pnpm create-org "City of Vx"`
    - The new org ID and short name will be printed out to the console
  - Create users with `pnpm create-user --orgId=<orgId from previous step> "someone@example.com"`

Once we have a sense of what we want we can update this to iterate over a list of org names and user emails. Each operation is pretty slow (~5 seconds per org, ~10-20 seconds per user) - might dig into that later.

### DIFF
--- a/apps/design/backend/src/auth/client.ts
+++ b/apps/design/backend/src/auth/client.ts
@@ -2,6 +2,7 @@
 
 import { ManagementClient, AuthenticationClient } from 'auth0';
 import { assertDefined } from '@votingworks/basics';
+import crypto from 'node:crypto';
 import {
   auth0ClientDomain,
   auth0ClientId,
@@ -9,6 +10,8 @@ import {
   votingWorksOrgId,
 } from '../globals';
 import { Auth0User, Org, User } from '../types';
+
+type ConnectionType = 'Username-Password-Authentication' | 'google-oauth2';
 
 export class AuthClient {
   constructor(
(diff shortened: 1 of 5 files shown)


kofi-q:
Sorry for the size of this one - a good chunk of it are scripts and docs (commit 2), so hopefully not too bad 🙏🏿

apps/design/backend/src/auth/client.ts
```diff
@@ -40,6 +123,123 @@ export class AuthClient {
...
+
+    return res.data.map((c) => ({ id: c.id, name }))[0];
+  }
+
+  async createOrg(params: {
+    displayName: string;
+    enableGoogleAuth?: boolean;
+    logoUrl?: string;
+  }): Promise<Org> {
+    const VX_PURPLE = '#8d24ce';
```
kshen0, line 139: This might be useful as a constant in `libs/ui`

kshen0 (approved):

apps/design/backend/src/auth/client.ts
kofi-q: Oh, good point - just went searching, thinking it might already exist, but looks like the purple shades are defined in CSS `hsl` format, so will see about reconciling that later

PR #5933, opened 2025-01-31
Always clean up renderer

arsalansufi:
## Overview

Thinking through error handling flows and noticed that we don't clean up the renderer if the call to `renderBallotPreviewToPdf` fails. Figured we should always be doing that. Maybe this will address the error I posted about in https://votingworks.slack.com/archives/C085YT4798C/p1738280647190929 - but I'm not sure. I'm not able to reproduce that error locally unfortunately.


jonahkagan (approved):
Seems like a good idea

PR #5935, opened 2025-01-31
VxDesign: Only apply NH rotation rules if using NH template

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/5904

This PR updates VxDesign to only apply NH rotation rules if using an NH template. When using the default Vx template, we'll now use the order as entered into the system. Note that if you start with the default Vx template, and switch to an NH template, we won't automatically rotate. We'll only do so on the next save operation. This feels like a fine limitation for now.

Context here is that we're going to be prepping some ballots for SLI where the candidate order has been defined in a spec, and we don't want to apply NH rotation rules.

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -288,12 +288,16 @@ function buildApi({ auth, workspace, translator }: AppContext) {
       electionId: ElectionId;
       election: Election;
     }): Promise<void> {
-      const { election } = await store.getElection(input.electionId);
+      const { election, ballotTemplateId } = await store.getElection(
+        input.electionId
+      );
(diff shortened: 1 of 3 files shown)


kofi-q (approved):

eventualbuddha (approved):
This looks okay, but it took a little while to determine that this was safe because the name `rotateCandidates` implies that it may not be idempotent, and there is no comment or test that ensures that it is.

arsalansufi:
That makes sense @eventualbuddha! I threw in a test to check that the NH rotation logic is idempotent

Issue #5866, opened 2025-01-22
VxDesign: Validate image uploads


jonahkagan:
For seals:
- Background is transparent (if not SVG) --> Can punt on this one
- Minimum size: 200x200px 
- Maximum size: TBD - investigate this as part of this task

Reject the upload and return an error message if any of the conditions isn't met.

Also consider:
- Clerk signatures
- Ballot measure images

Generally safe to assume: 1 seal, 2 clerk signatures, 1 ballot measure image

Clerk signature notably doesn't go into the election package.

Note that non-seal images are converted to base64 SVGs, so they grow in size. Current limits on upload end (5MB) apply to original upload, not converted file. Let's check the converted file.

Main guiding principle: Don't crash VxDesign or slow down VxSuite with gigantic election defs. Ballot looks okay.

PR #5936, opened 2025-01-31
optionally validate image dimensions

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/5866

## Demo Video or Screenshot

https://github.com/user-attachments/assets/9bc47ff9-19e1-4215-99b3-8db2f8cb68bb

Screenshot 2025-01-31 at 3 18 20 PM

Screenshot 2025-01-31 at 3 18 22 PM

Screenshot 2025-01-31 at 3 18 07 PM

Screenshot 2025-01-31 at 3 18 13 PM

Styling when a valid seal has been previously uploaded:

Screenshot 2025-01-31 at 3 17 40 PM

(The NH seal pictured is valid and remains after I attempted to upload an invalid seal).

## Testing Plan
- manually tested

## Checklist

### DIFF
--- a/apps/design/frontend/src/image_input.tsx
+++ b/apps/design/frontend/src/image_input.tsx
@@ -1,9 +1,11 @@
 import { Buffer } from 'node:buffer';
 import sanitizeHtml from 'sanitize-html';
-import { FileInputButton, FileInputButtonProps } from '@votingworks/ui';
-import { assert, assertDefined } from '@votingworks/basics';
+import { FileInputButton, FileInputButtonProps, P } from '@votingworks/ui';
+import { assert, assertDefined, err, ok, Result } from '@votingworks/basics';
+import { useEffect, useState } from 'react';
+import { safeParseNumber } from '@votingworks/types';
 
-const MAX_IMAGE_UPLOAD_BYTES = 5 * 1_000 * 1_000; // 5 MB
+const MAX_IMAGE_UPLOAD_BYTES = 1 * 1000 * 1_000; // 1 MB
 
 const ALLOWED_IMAGE_TYPES = ['image/svg+xml', 'image/png', 'image/jpeg'];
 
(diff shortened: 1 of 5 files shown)


apps/design/frontend/src/rich_text_editor.tsx
```diff
@@ -147,6 +147,8 @@ function Toolbar({ editor }: { editor: Editor }) {
               })
               .run();
           }}
+          // eslint-disable-next-line no-console
+          onError={(err) => console.error(err)}
```
kshen0, line 151: This is existing behavior; since RichTextEditor doesn't specify any minimum dimensions it can only throw on unexpected errors or hitting the image size limit.

apps/design/frontend/src/image_input.tsx
```diff
@@ -113,39 +115,131 @@ async function loadBitmapImageAndConvertToSvg(file: File): Promise<string> {
...
+        const svgImage =
+          file.type === 'image/svg+xml'
+            ? await loadSvgImage(file)
+            : await loadBitmapImageAndConvertToSvg(file);
+
+        const validateResult = validateSvgDimensions(
+          svgImage,
+          minHeightPx,
+          minWidthPx
+        );
```
kshen0, line 236: I debated whether we needed to enforce image dimensions on SVGs but decided to for a few reasons

* It's simpler for the user if all image types are enforced in the same way
* I'm not sure what the limits of SVG scaling are eg. can an SVG with a defined 1x1 size scale up to 200x200? Probably, but without time to investigate properly I don't want to risk complications

apps/design/frontend/src/image_input.tsx
```diff
@@ -1,9 +1,11 @@
...
 import sanitizeHtml from 'sanitize-html';
-import { FileInputButton, FileInputButtonProps } from '@votingworks/ui';
-import { assert, assertDefined } from '@votingworks/basics';
+import { FileInputButton, FileInputButtonProps, P } from '@votingworks/ui';
+import { assert, assertDefined, err, ok, Result } from '@votingworks/basics';
+import { useEffect, useState } from 'react';
+import { safeParseNumber } from '@votingworks/types';
 
-const MAX_IMAGE_UPLOAD_BYTES = 5 * 1_000 * 1_000; // 5 MB
+const MAX_IMAGE_UPLOAD_BYTES = 1 * 1000 * 1_000; // 1 MB
```
eventualbuddha, line 8: Why reducing to 1MB?

apps/design/frontend/src/image_input.tsx
eventualbuddha: Personally, I would have stuck with just validating the size of non-SVG images. It's much easier than trying to guess whether an SVG has the size we want. For example, an SVG that says it's 1x1 pixel could possibly be scaled up to 200x200 just fine. It's also possible that a 1000x1000 pixel SVG could contain a very low-res bitmap image. Without going quite deep on it, we don't really know.

apps/design/frontend/src/image_input.tsx
```diff
@@ -158,14 +252,36 @@ export function ImageInput({
...
   value?: string;
   onChange: (value: string) => void;
   buttonLabel: string;
   disabled?: boolean;
   className?: string;
   required?: boolean;
+  minWidthPx?: number;
+  minHeightPx?: number;
 }): JSX.Element {
+  const [error, setError] = useState<Error | undefined>(undefined);
```
eventualbuddha, line 267:
```suggestion
  const [error, setError] = useState<Error>();
```

apps/design/frontend/src/image_input.tsx
```diff
@@ -158,14 +252,36 @@ export function ImageInput({
...
+  // Clear error if the parent component has stopped interacting with this one
+  useEffect(() => {
+    if (disabled) {
+      setError(undefined);
+    }
+  }, [disabled]);
+
+  function onError(e: Error) {
+    setError(e);
+  }
```
eventualbuddha, line 278: This doesn't actually prevent the form from being submitted. Is that intentional? If so, use a warning symbol and coloring?

eventualbuddha (approved):
Some questions, but overall okay for now.

apps/design/frontend/src/image_input.tsx
kshen0: Whoops, left in `color: 'red'`. Meant to use a red consistent with the rest of the product.

I'll update this + form locking.

apps/design/frontend/src/image_input.tsx
kshen0: Initially reduced to avoid unnecessarily large images in seal and signature fields, but doing the math on ballot measure images 5 MB may be more appropriate:

width = ~7 inches (accounting for ballot frame, padding etc) * 300 ppi = 2100 px
height = ~half a legal page max = 9 inches * 300 ppi = 2700 px

Not unreasonable for a 2100x2700 image to be up to 5 MB. Maybe in the future we want to add image size as a prop but not very high priority right now

apps/design/frontend/src/image_input.tsx
kshen0: Good points. Removed validation of SVG dimensions in

[ec27aca](https://github.com/votingworks/vxsuite/pull/5936/commits/ec27aca0b2903dee1559fd6786f28abeea6cbcde)

apps/design/frontend/src/image_input.tsx
kshen0: Used native form validation and `Callout` instead of `P`: https://github.com/votingworks/vxsuite/pull/5936/commits/2a0354a0bbb26e2bc4bc891a086c18cd2e0bb3a3#diff-00afb2dfbef2411d546c95d8c995ca5938024b7b9ad8d483e59c43472b3ebd13R306

PR #5937, opened 2025-01-31
Get election package exports working in deployed environments by using S3 for file storage

arsalansufi:
## Overview

This PR should get election package exports working in deployed environments by using S3 for file storage. With our current local setup, the background worker writes the election package to the local file system, and the server then pulls that file from that same file system as it's running on the same machine. In deployed environments, the the server and background worker are of course running on different machines so we need another means of sharing the file between the two. I've opted for S3.

To keep local and production code paths as close together as possible, I've set up a generic `FileStorageClient` interface, with two implementations, one for deployed environments `S3FileStorageClient` and one for local dev `LocalFileStorageClient`. Eventually, local dev could also use S3, with its own bucket and credentialing setup.

## Testing Plan

- [x] Configured my local dev env to point to an S3 bucket and tested end-to-end
- [x] Deploy to staging and test

Punting on automated tests for now given the time constraints

## Checklist

- [x] Set up VxDesign S3 buckets, one for production and one for staging
- [x] Set up VxDesign IAM roles with access to those buckets
- [x] Set up VxDesign AWS env vars in Heroku

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -17,6 +17,7 @@ import {
   ElectionIdSchema,
   DateWithoutTimeSchema,
   unsafeParse,
+  ElectionDefinition,
 } from '@votingworks/types';
 import express, { Application } from 'express';
 import {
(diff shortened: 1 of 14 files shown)


apps/design/backend/src/file_storage_client.ts
```diff
@@ -0,0 +1,71 @@
...
+    const data = await this.s3Client.send(
+      new GetObjectCommand({
+        Bucket: process.env.AWS_S3_BUCKET_NAME,
+        Key: filePath,
+      })
+    );
+    return (data.Body ?? Readable.from([])) as Readable;
+  }
+
+  async writeFile(filePath: string, contents: Buffer): Promise<void> {
```
arsalansufi, line 41: Because we're not generating language and audio files this ballot production run, the election packages will be reasonably small. So I think it's okay to hold them in memory for writing purposes. Down the road, we'll want to use streaming APIs. This PR does use streaming APIs for election package downloads.

apps/design/backend/src/app.ts
```diff
@@ -494,13 +494,15 @@ function buildApi({ auth, workspace, translator }: AppContext) {
...
+    }: WithUserInfo<{
       electionId: ElectionId;
       electionSerializationFormat: ElectionSerializationFormat;
-    }): Promise<void> {
+    }>): Promise<void> {
       return store.createElectionPackageBackgroundTask(
         electionId,
-        electionSerializationFormat
+        electionSerializationFormat,
+        user.orgId
```
arsalansufi, line 505: Following the pattern for user info used elsewhere in the app

apps/design/backend/src/app.ts
```diff
@@ -674,6 +676,25 @@ export function buildApp(context: AppContext): Application {
...
+    const userOrg = await context.auth.org(user.org_id);
+    if (!userOrg) {
+      res.status(500).send('No org found for user');
+      return;
+    }
+
+    const { orgId, fileName } = req.params;
+    if (orgId !== userOrg.id && userOrg.id !== votingWorksOrgId()) {
+      res.status(404).send('File not found');
+    }
```
arsalansufi, line 690: I'm siloing files by org, but still making it such that Vx users can access any org's files.

apps/design/backend/src/file_storage_client.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  async readFile(filePath: string): Promise<Readable> {
+    const data = await this.s3Client.send(
+      new GetObjectCommand({
+        Bucket: process.env.AWS_S3_BUCKET_NAME,
+        Key: filePath,
+      })
+    );
+    return (data.Body ?? Readable.from([])) as Readable;
```
kshen0, line 38: We might want to consider using `Result<Readable, FileStorageClientError>` in the future. I'm assuming `data.Body` is undefined or null when the file does not exist, so it's not quite dire enough to throw. But it might make debugging and error handling more confusing to silently return no data like this.

kshen0 (approved):

apps/design/backend/src/file_storage_client.ts
arsalansufi: I actually don't know if `data.Body` will be undefined when the file doesn't exist and am running out of time to test, but your suggestion re error handling is a good one! Set up the scaffolding here: https://github.com/votingworks/vxsuite/pull/5937/commits/53a62d70936e7952822d4605c02a85f87a25029d

Issue #5928, opened 2025-01-30
Placeholder for post-release cleanup tasks


arsalansufi:
- [ ] Rename "BallotOrderInfoScreen" "OrderBallotsScreen" to align with new sidebar title
- [ ] Rename "BallotsScreen" "ProofBallotsScreen" to align with new sidebar title
- [ ] [Auth TODO](https://github.com/votingworks/vxsuite/blob/c1c2e4bdd9b1e836832688640a827052c53d6348/apps/design/backend/src/app.ts#L191-L195)
- [ ] Decide whether or not to wrap up https://github.com/votingworks/vxsuite/pull/5958
- [x] DB migration support

adghayes:
@arsalansufi this relevant?

arsalansufi:
We can close - all the listed items are either no longer relevant, already addressed, or not clearly worth pursuing

PR #5958, opened 2025-02-04
perf(design): move ballot preview generation out of request handler

eventualbuddha:
## Overview
Moves ballot preview generation out of the `express`/`grout` request handler and into a background task. We shouldn't be doing anything this heavyweight inside a request handler, both in terms of time and memory usage.

The first commit updates the background task infrastructure to support a returned "result" rather than just an error, and the second uses that to return the generated PDF.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/a76e12c6-074a-4178-b391-ca5036c1de4a

## Testing Plan
Tested manually as automated testing is temporarily disabled.


arsalansufi:
Added an item to https://github.com/votingworks/vxsuite/issues/5928 so we don't lose track of this after the ballot production run ✅

arsalansufi:
Yup fine to close! While this may still be worthwhile down the road, we haven't seen major perf issues while rendering ballots. And if we were to move to a background worker, we'd want to first setup proper infrastructure to process more than one background task at once. Background tasks are currently processed serially. Definitely don't want previews blocked on someone else's exports, etc.

PR #5976, opened 2025-02-05
feat(design): allow per-precinct split seal images

eventualbuddha:
## Overview

Allows setting an election seal image per precinct split that overrides the one for the election as a whole. One must still be provided at the election level, and is optional at the precinct split level.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/588765fe-e85a-4e69-afad-098cc329c160

## Testing Plan
Manually tested only as automated tests are still mostly disabled.

## Checklist

### DIFF
--- a/apps/design/frontend/src/image_input.tsx
+++ b/apps/design/frontend/src/image_input.tsx
@@ -202,6 +202,18 @@ export function ImageInputButton({
   );
 }
 
+export interface ImageInputProps {
+  value?: string;
+  onChange: (value?: string) => void;
+  buttonLabel: string;
+  removeButtonLabel: string;
+  disabled?: boolean;
+  className?: string;
+  required?: boolean;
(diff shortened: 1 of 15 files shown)


apps/design/frontend/src/image_input.tsx
```diff
@@ -243,6 +243,15 @@ export function ImageInput({
     onChange(newValue);
   }
 
+  function onRemoveButtonClick() {
+    // Prevent the button underlying <input> from also being clicked, which
+    // would trigger a file selection dialog immediately after the remove button
+    // is clicked.
```
arsalansufi, line 249: Huh interesting - I don't think I've seen this behavior but good to know!

arsalansufi (approved):
LGTM! Thanks for hopping on this!

PR #5980, opened 2025-02-06
fix(design): prevent crash on contest delete

eventualbuddha:
## Overview

Fixes https://votingworks.sentry.io/issues/6266281369/

### Steps to Reproduce
1. Create a contest.
2. Delete the contest.

### Expected Result
The contest should be deleted and the user should be redirected back to the contest list.

### Actual Result
Most of the time, it happens exactly as expected. However, the order in which `savedElection` is updated and the `updateElectionMutation` `onSuccess` callback is called is not defined. If `savedElection` is updated first it will no longer have the contest that was just deleted and the `find` call will panic.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

### DIFF
--- a/apps/design/frontend/src/contests_screen.tsx
+++ b/apps/design/frontend/src/contests_screen.tsx
@@ -358,9 +358,9 @@ function ContestForm({
   savedElection: Election;
 }): JSX.Element | null {
   const savedContests = savedElection.contests;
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
Thank you for investigating these Sentry alerts and patching!

Issue #5975, opened 2025-02-05
Add user-defined ballot images to election.json


kshen0:
Motivation: election hash should update when user-defined images change

- clerk signature
- precinct split seal override

Should work for v3 and v4

PR #5991, opened 2025-02-07
Make precinct split config affect election hash

kshen0:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/5975

Ensures that changes to precinct split-specific fields (text or image) update the election hash.

* Adds `additionalHashInput` to the shared `Election` type and the `election.json` file exported by VxDesign via test deck or election package export. Into this property we add hashes of precinct split signature and seal override images.
* Adds `electionTitleOverride` and `clerkSignatureCaption` to `hmpbStrings` in `election.json`

nb. I decided to write to `election.json` the hashes of the images, rather than the SVGs themselves, since we have seen images up to our max 5 MB and one election may have several precinct splits each with their own seal override and signature. See [discussion](https://votingworks.slack.com/archives/C014MGC0WRE/p1738905661981559) that confirms this does not meaningfully increase chance of hash collision.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/1ba0dda7-8281-4a06-a3bf-5902d297ba57

## Testing Plan
- added test for new utility functions
- manually verified election.json output

## Checklist

### DIFF
--- a/apps/design/backend/src/ballots.ts
+++ b/apps/design/backend/src/ballots.ts
@@ -1,4 +1,11 @@
-import { Election, UiStringsPackage } from '@votingworks/types';
+import {
+  Election,
+  hasSplits,
+  PrecinctSplit,
+  PrecinctWithSplits,
+  SplittablePrecinct,
+  UiStringsPackage,
+} from '@votingworks/types';
 import {
   allBaseBallotProps,
   BallotTemplateId,
@@ -7,16 +14,7 @@ import {
 } from '@votingworks/hmpb';
 import { find, throwIllegalValue } from '@votingworks/basics';
 import { sha256 } from 'js-sha256';
-import {
-  BallotStyle,
-  hasSplits,
-  normalizeState,
-  Precinct,
-  PrecinctSplit,
-  PrecinctWithSplits,
-  User,
-  UsState,
-} from './types';
+import { BallotStyle, normalizeState, User, UsState } from './types';
 
 function getPrecinctSplitForBallotStyle(
   precinct: PrecinctWithSplits,
@@ -52,7 +50,7 @@ export function defaultBallotTemplate(
 export function formatElectionForExport(
   election: Election,
   ballotStrings: UiStringsPackage,
-  precincts: Precinct[]
+  precincts: SplittablePrecinct[]
 ): Election {
   const splitPrecincts = precincts.filter((p) => hasSplits(p));
 
@@ -89,7 +87,7 @@ export function formatElectionForExport(
 export function createBallotPropsForTemplate(
   templateId: BallotTemplateId,
   election: Election,
-  precincts: Precinct[],
+  precincts: SplittablePrecinct[],
   ballotStyles: BallotStyle[]
 ): BaseBallotProps[] {
   function buildNhBallotProps(props: BaseBallotProps): NhBallotProps {
--- a/apps/design/backend/src/ballots.ts
+++ b/apps/design/backend/src/ballots.ts
@@ -1,4 +1,4 @@
-import { Election } from '@votingworks/types';
+import { Election, UiStringsPackage } from '@votingworks/types';
 import {
   allBaseBallotProps,
   BallotTemplateId,
@@ -6,6 +6,7 @@ import {
   NhBallotProps,
 } from '@votingworks/hmpb';
 import { find, throwIllegalValue } from '@votingworks/basics';
+import { sha256 } from 'js-sha256';
 import {
   BallotStyle,
   hasSplits,
@@ -46,6 +47,43 @@ export function defaultBallotTemplate(
     default:
       return 'VxDefaultBallot';
   }
+}
+
+export function formatElectionForExport(
+  election: Election,
+  ballotStrings: UiStringsPackage,
+  precincts: Precinct[]
+): Election {
+  const splitPrecincts = precincts.filter((p) => hasSplits(p));
+
+  const signatureImageBySplit = splitPrecincts.flatMap((p) =>
+    p.splits.flatMap((split) =>
+      split.clerkSignatureImage
+        ? [[`${p.id}-${split.id}`, sha256(split.clerkSignatureImage)]]
+        : []
+    )
+  );
+  const sealOverrideBySplit = splitPrecincts.flatMap((p) =>
+    p.splits.flatMap((split) =>
+      split.electionSealOverride
+        ? [[`${p.id}-${split.id}`, sha256(split.electionSealOverride)]]
+        : []
+    )
+  );
+
+  const additionalHashInput: Record<string, Record<string, string>> = {
+    precinctSplitSeals: Object.fromEntries(sealOverrideBySplit),
+    precinctSplitSignatureImages: Object.fromEntries(signatureImageBySplit),
+  };
+
+  return {
+    ...election,
+    ballotStrings,
+    additionalHashInput: {
+      ...(election.additionalHashInput || {}),
+      ...additionalHashInput,
+    },
+  };
 }
 
 export function createBallotPropsForTemplate(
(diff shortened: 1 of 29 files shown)


apps/design/backend/src/ballots.ts
```diff
@@ -40,6 +41,43 @@ export function defaultBallotTemplate(state: string): BallotTemplateId {
...
+        : []
+    )
+  );
+  const sealOverrideBySplit = splitPrecincts.flatMap((p) =>
+    p.splits.flatMap((split) =>
+      split.electionSealOverride
+        ? [[`${p.id}-${split.id}`, sha256(split.electionSealOverride)]]
+        : []
+    )
+  );
```
arsalansufi, line 64: I think we should actually include all of the NH split override options, as all of these affect the rendered ballot:

https://github.com/votingworks/vxsuite/blob/6554fab5e4df912c76cbeb43cf5ef35588c4e71a/libs/hmpb/src/ballot_templates/nh_ballot_template.tsx#L58-L63

arsalansufi (commented):
Looks great conceptually! Just one suggestion

apps/design/backend/src/ballots.ts
kshen0: Huh I thought these were already added to `hmpbStrings` but looks like I deprioritized merging https://github.com/votingworks/vxsuite/pull/5837.

I'll try to fold those changes into this PR

arsalansufi: Oh interesting, yeah missed that other PR!

kshen0: Called here: https://github.com/votingworks/vxsuite/pull/5991/files#diff-c268bf2996342639e8bfd885b518b620095551a7b47c8517f93bab768c57bfd0R150

apps/design/backend/src/worker/generate_election_package_and_ballots.ts
```diff
@@ -144,11 +147,19 @@ export async function generateElectionPackageAndBallots(
     JSON.stringify(metadata, null, 2)
   );
 
+  const userDefinedHmpbStrings = getUserDefinedHmpbStrings(precincts);
+  // Combine predefined and user-defined HMPB strings here because they can be
+  // translated in the same path.
+  const combinedHmpbStringsCatalog: Record<string, string> = {
+    ...hmpbStringsCatalog,
+    ...userDefinedHmpbStrings,
+  };
```
jonahkagan, line 156: Should this logic be part of `getAllStringsForElectionPackage` so it's executed at other call sites as well? (I think currently there's just one in the fixture-generators library)

apps/design/backend/src/worker/generate_election_package_and_ballots.ts
```diff
@@ -158,16 +169,19 @@ export async function generateElectionPackageAndBallots(
...
 
   const ballotStrings = mergeUiStrings(electionStrings, hmpbStrings);
-  const electionWithBallotStrings: Election = {
-    ...election,
+  const formattedElection = formatElectionForExport(
+    election,
     ballotStrings,
-  };
+    precincts
+  );
```
jonahkagan, line 176: Similar to my other comment, where else does this logic need to be executed? Should the fixture-generators lib also include it? What about test deck generation?

Anywhere we call `renderAllBallotsAndCreateElectionDefinition` or `renderMinimalBallotsToCreateElectionDefinition` is probably a good place to check

We also might want to think about factoring out some of the logic here even more to avoid missing a spot in the future

jonahkagan: Good point. I think we could consider libs/fixture-generators part of VxDesign, since it's simulating what VxDesign does. And test deck generation is already in VxDesign. So it might not require moving the type fully into the rest of VxSuite

jonahkagan: Also, I think the test deck part is the only correctness concern. libs/fixture-generators is more of a nice to have, so I'm fine omitting that.

kshen0: Summary of changes:

* VxDesign's `Precinct` is factored out to `SplittablePrecinct` in libs/types ([src](https://github.com/votingworks/vxsuite/pull/5991/commits/bfa65191b9d011cbe8b283eea5b3a5a7c134408b#diff-f64086a9f63d69967e91c55e86467e75be5dcb8aa355f76ef921cbd00f529006R1016))
* I realized `translateBallotStrings` and `getAllStringsForElectionPackage` are very similar, but just different enough to warrant factoring out the shared `translateElectionAndHmpbStrings` to keep existing signatures ([src](https://github.com/votingworks/vxsuite/pull/5991/commits/c318a451269dbc2c57a1ccd7ba7bf71f2f731a22))
  * Extend existing signatures to expect `SplittablePrecinct[]`
* `exportTestDecks` passes precincts to `translateBallotStrings` so precinct split strings are included in ballot strings ([src](https://github.com/votingworks/vxsuite/pull/5991/commits/bfa65191b9d011cbe8b283eea5b3a5a7c134408b#diff-6771324e32d54bc6f040273b5f7d8381ad6227b7031f87a1b8726aefcc9481dfR500-R505))
  * Also now calls `formatElectionForExport` so precinct split image hashes are included in the election
* `generateElectionPackageAndBallots` passes precincts to `getAllStringsForElectionPackage`([src](https://github.com/votingworks/vxsuite/pull/5991/files#diff-c268bf2996342639e8bfd885b518b620095551a7b47c8517f93bab768c57bfd0R149-R154))


Because fixture elections read by `fixture-generator` don't have precinct splits, the call to `getAllStringsForElectionPackage` is made compatible [here](https://github.com/votingworks/vxsuite/pull/5991/files#diff-707fbc1f6e73e18f1b5990c39a710d1f1b012e4613fa5c0797930bc1c43c1debR68). The fixture elections should have no change because there are no precinct split options to add to the election. I believe we can support precinct splits in `fixture-generator` when
* `SplittablePrecinct` replaces `Precinct` and is integrated with the main `Election` type
* precinct splits are added to the source `election.json` files [here](https://github.com/votingworks/vxsuite/blob/main/libs/fixture-generators/bin/regenerate-election-packages#L11-L15)

libs/types/src/test_utils.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Election, SplittablePrecinct } from './election';
+
+// istanbul ignore next - @preserve
+export function generateSplittablePrecinctsForTest(
+  sourceElection: Election
+): SplittablePrecinct[] {
+  return sourceElection.precincts.map(
+    (p): SplittablePrecinct => ({
+      ...p,
+      districtIds: [assertDefined(sourceElection.districts[0]).id],
```
jonahkagan, line 11: For this to be technically correct, we'd need to go through the ballot styles to find the district list for each precinct and create splits accordingly. Not sure how important it is for the tests that are using this to have the correct precincts and splits

libs/backend/src/language_and_audio/ballot_strings.ts
```diff
@@ -77,13 +77,16 @@ export async function translateHmpbStrings(
  * Translates (or loads from the translation cache) a UI strings package for
  * HMPB rendering. Includes all election strings and app strings.
  */
-export async function translateBallotStrings(
+export async function translateElectionAndHmpbStrings(
```
jonahkagan, line 80: Nice factoring out of shared logic 👏

jonahkagan (approved):

libs/types/src/test_utils.ts
kshen0: Currently it's not important because we're only using it to unlock compatibility where tests call existing translation functions that expect `SplittablePrecincts` to be passed in.

String extraction itself is tested here: https://github.com/votingworks/vxsuite/pull/5991/files#diff-2e159f5332cddf7352b9c7a86bbde3761202da561cacf3198bc921fd4242a58cR133

Issue #5953, opened 2025-02-03
Gradually bring VxDesign tests back


arsalansufi:
We needed to disable a number of tests to get deploys over the finish line before the launch. But can now piecemeal bring them back to restore some safety to dev work during the ballot production run.

We'll finish this work no matter what after the ballot production run, but there's utility in the meantime, too. Other tasks take precedence, but we should chip away at this when we have cycles.

PR #5995, opened 2025-02-10
test(design/frontend): replace `jest` with `vitest`

eventualbuddha:
## Overview

Closes #5585
Refs #5953

Only converts the tests that are enabled. This unblocks re-enabling the rest of the tests that depend on `@votingworks/design-backend` which uses an ESM module that Jest doesn't like.

## Demo Video or Screenshot
n/a

## Testing Plan
n/a


apps/design/frontend/test/setupTests.ts
```diff
@@ -9,26 +10,23 @@ import {
...
-jest.mock('styled-components', () =>
-  jest.requireActual('styled-components/dist/styled-components.browser.cjs.js')
-);
+expect.extend(matchers);
 
 // Deterministic ID generation
 const idFactory = makeIdFactory();
 
 beforeEach(() => {
+  cleanup();
```
kshen0, line 19: Should this be run in `beforeEach` or `afterEach`? https://testing-library.com/docs/react-testing-library/api/#cleanup

If a test fails and this suite short-circuits will we violate the recommendation " If not, you will need to call cleanup() after each test."? I suppose if all suites call `cleanup()` in `beforEach` it doesn't matter.

kshen0 (approved):

apps/design/frontend/test/setupTests.ts
eventualbuddha: I'll switch it per their recommendation. However, I don't think it actually matters most of the time if ever.

PR #6001, opened 2025-02-10
fix(design): work around V3 custom yes/no label bug

eventualbuddha:
## Overview
V3 has a bug where the tally reports do not use the custom label for yes/no contest options (aka ballot measures or articles). As a result, the ballot will have e.g. "A" and "B" but the results will say "Yes" and "No". To work around this we transform yes/no contests in V3 elections that contain custom labels into candidate contests with candidates named using the custom label.

I've verified that the various tally reports show the correct information and include the contests where they were originally defined within the election definition.

## Demo Video or Screenshot
image
**Interpretation is correct.**

image
**Overvotes work the same way.**

CleanShot 2025-02-10 at 15 38 40@2x
**Articles appear in the proper order in reports.**

## Testing Plan
Added basic automated test. Tested with draft version of Londonderry's election definition.

### DIFF
--- a/libs/hmpb/src/render_ballot.test.tsx
+++ b/libs/hmpb/src/render_ballot.test.tsx
@@ -3,7 +3,12 @@ import {
   electionGeneralFixtures,
 } from '@votingworks/fixtures';
 import { test, expect } from 'vitest';
-import { BallotType, HmpbBallotPaperSize } from '@votingworks/types';
+import {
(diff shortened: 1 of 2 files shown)


benadida (approved):

arsalansufi (approved):
LGTM! Thanks for checking all the downstream implications!

PR #5856, opened 2025-01-22
test(design/frontend): replace `jest` with `vitest`

eventualbuddha:
Closes #5855 
Part of #5706

PR #6003, opened 2025-02-11
test(design): re-enable some tests

eventualbuddha:
## Overview

Refs #5953 
Obsoletes #5856

Re-enables one test file per commit. Note that I changed the ballot finalization API to make testing it easier (explained in the commit).

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests 🎉

### DIFF
--- a/apps/design/frontend/src/ballots_screen.tsx
+++ b/apps/design/frontend/src/ballots_screen.tsx
@@ -31,7 +31,7 @@ import styled from 'styled-components';
 import {
   getElection,
   getBallotsFinalizedAt,
-  setBallotsFinalizedAt,
+  finalizeBallots,
   updateElection,
 } from './api';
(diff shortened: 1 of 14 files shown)


apps/design/frontend/src/ballots_screen.tsx
```diff
@@ -238,11 +237,8 @@ function BallotStylesTab(): JSX.Element | null {
                   <Button
                     icon="Done"
                     onPress={() =>
-                      setIsBallotProofingCompleteMutation.mutate(
-                        {
-                          electionId,
-                          finalizedAt: new Date(),
-                        },
```
kshen0, line 245: 👍 makes sense to remove this as a parameter for easier testing, especially if we don't have a use case to set `finalizedAt` from the frontend

kshen0 (approved):

Issue #5999, opened 2025-02-10
VxDesign: Add ability to edit all system settings to Tabulation tab (and maybe rename the tab or split it)


arsalansufi:
We've started adding a few system settings here but need to give SLI the ability to edit all of them. Actual customers should still not be able to edit any of these settings.

No need to create the SLI org as part of this task, we'll do that later. But conceptually, the Vx org and SLI org will be quite similar in terms of which toggles they'll have access to.

PR #6019, opened 2025-02-13
[VxDesign] Edit all system settings

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/5999

* Renames `Tabulation` tab to `System Settings`
* Supports editing of all system settings values

## Demo Video or Screenshot
Screenshot 2025-02-18 at 10 26 15 AM
Screenshot 2025-02-18 at 10 26 20 AM
Screenshot 2025-02-18 at 10 30 18 AM

https://github.com/user-attachments/assets/54de0397-5477-40aa-813c-65ca22dcaba1

## Testing Plan
- added automated tests
- manually verified form works and updated fields are present in systemSettings.json

## Checklist

### DIFF
--- a/apps/design/frontend/src/system_settings_screen.tsx
+++ b/apps/design/frontend/src/system_settings_screen.tsx
@@ -317,7 +317,7 @@ export function SystemSettingsForm({
           ballotTemplateId !== 'NhBallotV3Compact' && (
             <React.Fragment>
               <Card>
-                <H2>Auth</H2>
+                <H2>Authentication</H2>
                 <Column style={{ gap: '1.5rem' }}>
                   <CheckboxButton
                     label="Enable Poll Worker PINs"
--- a/apps/design/frontend/src/system_settings_screen.tsx
+++ b/apps/design/frontend/src/system_settings_screen.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import {
   H1,
   H2,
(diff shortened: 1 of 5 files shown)


apps/design/frontend/src/system_settings_screen.tsx
```diff
@@ -228,22 +315,211 @@ export function SystemSettingsForm({
...
+                      options={[
+                        { value: '3', label: '3' },
+                        { value: '4', label: '4' },
+                        { value: '5', label: '5' },
+                        { value: '6', label: '6' },
+                        { value: '7', label: '7' },
+                        { value: '8', label: '8' },
+                        { value: '9', label: '9' },
+                        { value: '10', label: '10' },
+                      ]}
```
kshen0, line 378: I made this a dropdown because

1. 3-10 might be an unintuitive range for the user, and this makes it more clear than a number spinner what the options are
2. It avoids an annoying edge case where typing `10` is impossible because validation happens immediately on change and `1` is an invalid option. This can be worked around but the dropdown felt simpler.

apps/design/frontend/src/system_settings_screen.tsx
```diff
@@ -0,0 +1,559 @@
...
+                />
+              </InputGroup>
+            )}
+          </Column>
+        </Card>
+        {ballotTemplateId !== 'NhBallotV3' &&
+          ballotTemplateId !== 'NhBallotV3Compact' && (
+            <React.Fragment>
+              <Card>
+                <H2>Auth</H2>
```
kofi-q, line 320: Since this is user-facing, `Authentication` might be clearer?

apps/design/frontend/src/system_settings_screen.tsx
```diff
@@ -0,0 +1,559 @@
...
+                      ariaLabel="Inactive Session Time Limit"
+                      isMulti={false}
+                      isSearchable={false}
+                      value={systemSettings.auth.inactiveSessionTimeLimitMinutes.toString()}
+                      options={[
+                        { value: '10', label: '10 minutes' },
+                        { value: '15', label: '15 minutes' },
+                        { value: '20', label: '20 minutes' },
+                        { value: '30', label: '30 minutes' },
+                        { value: '360', label: '6 hours' },
```
kofi-q, line 349: Would it simplify things a bit to use number `value`s for these dropdowns? I think that's supported with `SearchSelect` (unless there are complications with going that route)

kofi-q (approved):

apps/design/frontend/src/system_settings_screen.tsx
kshen0: `SearchSelect` option values must extend `string`: https://github.com/votingworks/vxsuite/blob/ac70a182ffbda2960c6e0e8328389a4d97fbfbb9/libs/ui/src/search_select.tsx#L60-L63

In retrospect, it might have been worth it to relax this type and build number support into `SearchSelect`, but probably out of scope for this change.

Issue #6021, opened 2025-02-13
Feature flagging tweaks for SLI


arsalansufi:
As we prep for SLI to use VxDesign throughout cert, we need to make a few feature flagging tweaks for them. As a user, SLI falls somewhere between Vx internal users and NH customers. They should see what Vx internal users see, but with these adjustments:
- [ ] They should only be able to see their elections, not all elections across orgs - this _should_ just work out of the box given how the code is set up.
- [ ] They should only be able to select the default Vx template - none of the NH templates should be available to them. Maybe that means we should just hide the template selector from them. Probably need to adjust [this block](https://github.com/votingworks/vxsuite/blob/efc424502f3cd91c6312cb9216b997b9f7466d80/apps/design/backend/src/app.ts#L234-L236) as part of this.
- [ ] We should hide the "Marginal Mark Threshold" input from them as that doesn't yet do anything. We'll be building support for that as part of v4.1 so good to keep it for us for internal testing.
- [ ] "Export Test Decks" either needs to be moved to the background worker or hidden as it'll currently timeout. SLI doesn't need this so we can temporarily hide it for them and bring it back when it's ready.

We could also hide the ballot order info tab from them, but I'd say don't bother and just leave as is. We may be end up removing that altogether as we think through long-term process flows. And it's not a problem for SLI to be able to see that.

I've created the SLI org in both staging and production and added myself and @kshen0 to it:
- Staging org ID: `org_hFI2hvkzlwqooLCT`
- Production org ID: `org_yi4RLwKwS9nZndfv`

mattroe:
Overall LGTM. Just noting that I think we're going to want to expose the test deck functionality to them when its ready to expedite the marking process.

PR #6025, opened 2025-02-19
add SLI feature flags

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/6021

Adds feature config and flagging for SLI.

## Demo Video or Screenshot

**Hide elections from other orgs**
Screenshot 2025-02-18 at 5 14 09 PM

**Hide Marginal Mark Threshold**
Screenshot 2025-02-18 at 5 15 49 PM

**Hide ballot template selector and "Export Test Decks"**
Screenshot 2025-02-18 at 5 15 52 PM

**Default to VX template (NH has full-width ballot measures)**
Screenshot 2025-02-18 at 5 16 04 PM

## Testing Plan
- [x] Add automated tests
- [x] Test in staging

## Checklist


apps/design/backend/src/app.test.ts
```diff
@@ -94,11 +94,13 @@ vi.setConfig({
 const vxUser: User = {
   orgId: 'votingworks',
   isVotingWorksUser: true,
+  isSliUser: false,
 };
 
 const nonVxUser: User = {
   orgId: '123',
   isVotingWorksUser: false,
+  isSliUser: false,
```
kshen0, line 103: Open to feedback on this approach. Seemed simplest to add this field and while it's not elegant for n+1 orgs, we don't expect to have more than 2 special-case orgs in the near future. Plus a rework of feature flags to send the flags down from the backend rather than these `is*User` bits might make any reorganization in the frontend obsolete soon anyway.

apps/design/backend/src/auth/client.ts
```diff
@@ -99,7 +105,8 @@ export class AuthClient {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       userFromRequest(_: Express.Request): Auth0User | undefined {
         return {
-          org_id: votingWorksOrgId(),
+          // org_id: votingWorksOrgId(),
```
arsalansufi, line 108: Lingering comment

arsalansufi (approved):
LGTM!

apps/design/backend/src/app.test.ts
arsalansufi: That makes sense to me, that the rework of feature flags to send them down from the backend will address this

PR #6029, opened 2025-02-19
test(mark-scan): replace `jest` with `vitest`

### LINKED ISSUE #5859: apps/mark-scan/backend

eventualbuddha:
Closes #5859
Part of #5706 

Converts both frontend and backend.


eventualbuddha:
@kofi-q for vitest review, @kshen0 for a quick glance at anything amiss with `mark-scan` updates.

apps/mark-scan/backend/src/app.diagnostics.test.ts
```diff
@@ -150,7 +147,10 @@ test('diagnostic records', async () => {
       diagnosticType: 'mark-scan-accessible-controller',
     })
   ).toBeNull();
-  jest.useFakeTimers().setSystemTime(0);
+  vi.useFakeTimers({
+    shouldAdvanceTime: true,
```
kofi-q, line 151: Does this have the potential for introducing non-determinism, given the `timestamp: 0` assertion below? Or does something else break if the fake timers stay frozen for this portion of the test?

kofi-q (approved):

kshen0 (approved):
Nothing looks wrong from the business logic side for mark-scan assuming the currently failing `printing ballots` test failure is an intermittent test timeout that can be fixed with a slightly higher timeout.

apps/mark-scan/backend/src/app.diagnostics.test.ts
eventualbuddha: It seems to work, but I take your point and it seems to work without it, so I removed it.

Issue #5745, opened 2024-12-12
libs/ui


jonahkagan:

jonahkagan:
I made a clean up PR on `main` that should make converting libs/ui a bit easier. But pausing there to switch projects. Rebase to main before working more on this

PR #6032, opened 2025-02-20
Completely replace `jest` with `vitest`

### LINKED ISSUE #5706: Move from `jest` to `vitest`

eventualbuddha:
Closes #5706 
Closes https://github.com/votingworks/vxsuite/issues/5745


libs/ui/src/invalid_card_screen.tsx
```diff
@@ -71,7 +71,7 @@ export function InvalidCardScreen({
             return 'election manager card';
           case 'poll_worker':
             return 'poll worker card';
-          /* istanbul ignore next */
+          /* istanbul ignore next - @preserve */
```
jonahkagan, line 74: Does this need to be moved inside the `default` block?

libs/ui/package.json
```diff
@@ -124,9 +118,10 @@
     "stream-browserify": "^3.0.0",
     "styled-components": "^5.3.11",
     "tmp": "^0.2.1",
-    "ts-jest": "29.1.1",
     "util": "^0.12.4",
-    "vite": "4.5.2"
+    "vite": "4.5.2",
+    "vitest": "^2.1.8",
+    "vitest-styled-components": "^1.0.0"
```
jonahkagan, line 124: I didn't realize you got this working! Nice!

libs/eslint-plugin-vx/src/configs/recommended.ts
```diff
@@ -175,12 +175,7 @@ export = {
     },
     {
       files: ['**/*.test.ts', '**/*.test.tsx'],
-      plugins: ['jest'],
       rules: {
-        'jest/max-nested-describe': ['error', { max: 1 }],
-        'jest/no-identical-title': 'error',
-        'jest/no-focused-tests': 'error',
-        'jest/valid-expect': ['error', { alwaysAwait: true }],
```
jonahkagan, line 183: Do we want any similar lint rules for `vitest`?

libs/auth/src/test_utils.ts
```diff
@@ -7,83 +7,42 @@ import type { Mocked, vi } from 'vitest';
...
-): Mocked<DippedSmartCardAuthApi>;
-
 /**
  * Builds a mock dipped smart card auth instance for application-level tests
  */
 export function buildMockDippedSmartCardAuth(
-  fn: jfn | vfn = jest.fn
-): jest.Mocked<DippedSmartCardAuthApi> | Mocked<DippedSmartCardAuthApi> {
+  fn: typeof vi.fn
+): Mocked<DippedSmartCardAuthApi> {
```
jonahkagan, line 15: Do you think we'll eventually want to remove the `fn` argument from all of these mock builders and just import `vi.fn` directly from `vitest` in the defining module? Obviously less work to leave as is now that they are set up this way, but just wondering what your thoughts are

jonahkagan (approved):
Very exciting!

libs/auth/src/test_utils.ts
eventualbuddha: Probably? If we did it that way we'd probably make `vitest` a `peerDependency`, right?

libs/eslint-plugin-vx/src/configs/recommended.ts
eventualbuddha: Working on adding `eslint-plugin-vitest`.

libs/auth/src/test_utils.ts
jonahkagan: Yeah that should work. I can't remember if we had it as a peer before, given that all the packages already have vitest as a dev dependency already.

libs/eslint-plugin-vx/src/configs/recommended.ts
eventualbuddha: [8b93ef6](https://github.com/votingworks/vxsuite/pull/6032/commits/8b93ef6b763fe2271e27aba555d9412022f14598)

libs/ui/package.json
eventualbuddha: Yep! For anything that didn't quite work identically I just disabled it with a comment.

libs/ui/src/invalid_card_screen.tsx
eventualbuddha: [c97fcfb](https://github.com/votingworks/vxsuite/pull/6032/commits/c97fcfb9b4fc5a423630c2a45863d37c7dff06d8)

Issue #5985, opened 2025-02-06
VxDesign: Decrease memory usage in production by using a render pool instead of rendering at max parallelizability


arsalansufi:
https://votingworks.slack.com/archives/C085YT4798C/p1738866316895089

I actually recently encountered a situation where even my Mac couldn't handle the extreme number of ballot styles. I addresses this by replacing the Promise.all with a loop: https://github.com/votingworks/vxsuite/commit/accdb3f16d7ea501b7fcb41aea987cfedf1644d4

I think we can probably find a middle ground, stil parallelizing but limiting the parallelization. Worth doing some testing in staging with different resource allocations for the background worker.

arsalansufi:
Closing this ticket out as it's gotten a bit confusing and has conflated some problems.

- First potential problem is exports taking so long to finish that it seems like they're hanging. More work is required to check if this is actually the case and where the code is hanging, if it really is. One theory is that this may have to do with overly high concurrency that Node can't handle.
- Second problem is very high memory usage. Early signs suggest that that's independent of concurrency.

After taking some time to investigate some more, I'm going to open new issues.

PR #6037, opened 2025-02-21
limit render concurrency

kshen0:
## Overview

Mitigates but does not close https://github.com/votingworks/vxsuite/issues/5985.

Uses a `for` loop instead of unbounded `Promise.all` to render ballots in `renderAllBallotsAndCreateElectionDefinition`. Rendering in parallel seemed to cause resource contention and slow down overall ballot and election package export time.

## Demo Video or Screenshot

N/A

## Testing Plan
- manually tested render of large ballot package locally

### DIFF
--- a/libs/hmpb/src/render_ballot.tsx
+++ b/libs/hmpb/src/render_ballot.tsx
@@ -541,25 +541,24 @@ export async function renderAllBallotsAndCreateElectionDefinition<
   const { election } = ballotProps[0];
   assert(ballotProps.every((props) => props.election === election));
 
-  const ballotsWithLayouts = await Promise.all(
-    ballotProps.map(async (props) => {
-      // We currently only need to return errors to the user in ballot preview -
-      // we assume the ballot was proofed by the time this function is called.
-      const document = (
-        await renderBallotTemplate(renderer, template, props)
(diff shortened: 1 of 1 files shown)


libs/hmpb/src/render_ballot.tsx
```diff
@@ -541,25 +541,24 @@ export async function renderAllBallotsAndCreateElectionDefinition<
...
+      document,
+      props.ballotStyleId,
+      template
+    );
+    ballotsWithLayouts.push({
+      document,
+      gridLayout,
+      props,
+    });
+  }
```
nikhilb4a, line 561: If we want to allow but limit concurrency, another simple option could be to use [Bluebird's](http://bluebirdjs.com/docs/api/promise.map.html) `Promise.map` which takes in a `concurrency` param. But we may not want to introduce another dep

nikhilb4a (approved):
One thought but LGTM!

libs/hmpb/src/render_ballot.tsx
kshen0: Good point - a previous version of this PR used `p-limit` to similar effect, but @arsalansufi and I found in testing that no concurrency was more stable than limited concurrency.

PR #6039, opened 2025-02-22
flag translation by org ID and global flag

kshen0:
## Overview

Multilanguage ballot and audio support was previously gated by a global feature flag that would turn multilang on for all users of VxDesign. This PR adds an additional check for org ID while keeping the global flag, so a user must be in the VX or SLI org AND the global flag must be on.

I debated removing the global flag but it's useful for dev. It's also a nice safety and convenience feature to keep around so we can turn off multilanguage export without a code change and deploy eg. in case long-running exports are impacting performance of customers' export jobs.

## Demo Video or Screenshot

No feature change

## Testing Plan
- [x] validate this works in staging
- [x] updated tests

## Checklist


arsalansufi (approved):
> I debated removing the global flag but it's useful for dev. It's also a nice safety and convenience feature to keep around so we can turn off multilanguage export without a code change and deploy eg. in case long-running exports are impacting performance of customers' export jobs.

Agreed on this intuition for now as well - LGTM!

For those passing by this PR, this is just a stopgap while we start getting SLI set up but still have NH users accessing their ballots. For language and audio, we'll ultimately have user-facing toggles. And re feature flagging as a whole, I do still think that we'll move that logic to the backend, and have the backend provide to the frontend the list of enabled features.

apps/design/backend/src/worker/generate_election_package_and_ballots.ts
```diff
@@ -181,7 +186,13 @@ export async function generateElectionPackageAndBallots(
     JSON.stringify(systemSettings, null, 2)
   );
 
+  const isCloudTranslationAndSpeechSynthesisEnabled =
+    isVxOrSliOrg(orgId) &&
```
arsalansufi, line 190: One small gotcha here is that this org ID appears to be the user org ID rather than the election org ID. So this means that when we, as Vx, go to export an election package for an NH org, we will end up generating audio.

I think there's a simple fix here, to use the election org ID when generating the background job rather than the user org ID. I'll make that tweak. Not a big deal in the meantime since we're done with NH exports (knock on wood) for this ballot production run.

arsalansufi: https://github.com/votingworks/vxsuite/pull/6041

PR #6042, opened 2025-02-24
VxScan: Electrical testing scaffolding

arsalansufi:
## Overview

[ [Product spec](https://docs.google.com/document/d/1rQw5QtJIORgLRdi3MKVSlLTD43Fp8AWOB_UOFRynYBw/edit?usp=sharing) ]

This PR sets up scaffolding for VxScan electrical testing. During electrical testing, SLI will image VxScan with an alternate image running this code.

This code isn't part of our certified system, so we have some flexibility in approach. I ultimately opted for an env var within that enables a different backend and frontend. This approach is nice because vxsuite-complete-system won't have to change at all. When I build electrical testing images, I'll enable the special env var - an env var that will otherwise always be false for our standard builds.

I went ahead and handled a few of the requirements while setting up the scaffolding:

- [x] Continuous card reads
- [x] Continuous USB drive writes
- [x] Continuous sound playing
- [x] Continuous display animation

I also set up a super minimal means of presenting status information on the frontend.

I'll be opening a PR for the analogous scaffolding in VxMark next.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/63059203-63ec-416c-89aa-209fb3acd5ed

## Testing Plan

- [x] Tested manually


eventualbuddha (approved):

apps/scan/backend/src/store.ts
```diff
@@ -1063,4 +1063,37 @@ export class Store {
...
+        status_message as statusMessage,
+        updated_at as updatedAt
+      from electrical_testing_status_messages
+      order by component asc
+      `
+    ) as Array<{ component: string; statusMessage: string; updatedAt: string }>;
+  }
+
+  setElectricalTestingStatusMessage(
+    component: string,
```
kshen0, line 1085: Since this is the primary key, maybe worth making the component type an enum?

kshen0 (approved):

apps/scan/backend/src/store.ts
arsalansufi: Makes sense to me! Adjusted: https://github.com/votingworks/vxsuite/pull/6042/commits/22324acbfd0249145c3111d0826d1eff22a4002d

PR #6043, opened 2025-02-24
ci: set test timeout to 10s in CI only

eventualbuddha:
## Overview

Removes ad hoc test timeout configs that were already 10s, assuming they are set to that for CI purposes.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.


arsalansufi (approved):
Thanks for removing the ad-hoc timeouts as we make this switch - LGTM!

PR #6045, opened 2025-02-24
ci: perform recursive builds sequentially

eventualbuddha:
## Overview

Something with `tsc` gets tripped up when multiple processes try to build the same package simultaneously. We haven't seen this be a problem in dev, so I'm only making this change in CI. It will slow CI down a bit, but it should reduce flakiness. Long term, I don't think we want this recursive `pnpm` package-based build setup, so I don't feel bad disabling it here.

An alternative, if we don't like how messy this makes the `package.json` files, would be to simply always do these builds sequentially. For example, on my M4 MacBook Pro `apps/scan/frontend` builds from a clean state in 45s without `CI=true` and in 50s with `CI=true`.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

### DIFF
--- a/apps/admin/backend/package.json
+++ b/apps/admin/backend/package.json
@@ -10,7 +10,9 @@
     "schema.sql"
   ],
   "scripts": {
-    "build": "pnpm --filter $npm_package_name... build:self",
+    "build": "is-ci build:ci build:dev",
+    "build:ci": "pnpm --filter $npm_package_name... --sequential build:self",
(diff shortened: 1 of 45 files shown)


eventualbuddha:
@jonahkagan unassigned @arsalansufi and @adghayes to make it clear who should be reviewing.

jonahkagan (approved):
I'm good with this approach!

Issue #5530, opened 2024-10-16
VxDesign: A clone/duplicate button to create a copy an existing election


arsalansufi:
Note that the copy should have a different top-level election ID. The database requires that field to be unique.

Issue automatically created from Slack feedback reported by caroline: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1728425577683799

PR #6051, opened 2025-02-26
[VxDesign] Add API for cloning elections

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/5530

Adding server-side API and client query for cloning elections, guarded by access to both the source and destination orgs.
Copies over election data, system settings, and ballot template ID; initialises everything else to defaults.

## Testing Plan
- New test cases

## Checklist


apps/design/backend/src/app.ts
```diff
@@ -240,6 +240,40 @@ function buildApi({ auth, workspace, translator }: AppContext) {
...
+      const { election, ballotTemplateId, orgId, precincts, systemSettings } =
+        await store.getElection(input.srcId);
+
+      if (!auth.hasAccess(input.user, orgId)) {
+        throw new grout.GroutError('Access denied', {
+          cause: 'Cannot clone election: invalid source organization.',
+        });
+      }
+      if (!auth.hasAccess(input.user, input.destOrgId)) {
+        throw new grout.GroutError('Access denied', {
```
eventualbuddha, line 259: A little odd that this function both returns a `Result` and `throw`s. Why not do this here?

```ts
return err(new grout.GroutError('Access denied', {
  cause: 'Cannot clone election: invalid destination organization.',
}));
```

apps/design/backend/src/app.ts
```diff
@@ -240,6 +240,40 @@ function buildApi({ auth, workspace, translator }: AppContext) {
...
+        input.destOrgId,
+        {
+          ...election,
+          id: input.destId,
+        },
+        precincts,
+        ballotTemplateId,
+        systemSettings
+      );
+      return ok(input.destId);
```
kshen0, line 274: I know `createElection` also allows the frontend to specify the new election's ID, but why not generate in the backend for both?

kshen0 (approved):

apps/design/backend/src/app.ts
kofi-q: Ah, thanks for catching - no reason; just copied from `createElection` without double-checking.
Are you tied to going the `Result` route? I'm thinking just change that to a normal return to simplify, since this edge case is likely a bad actor or broken client anyway

apps/design/backend/src/app.ts
kofi-q: Yeah, I started down that path, but punted for later and just stayed consistent for now - definitely feels like we should be generating IDs server-side, so we could update all the APIs together in a batch

PR #6061, opened 2025-02-27
[VxDesign] Add UI for cloning elections

kofi-q:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/5530

(Some alternatives/future ideas considered in this thread: https://votingworks.slack.com/archives/C085YT4798C/p1740606709518789)

Adding UI for cloning elections from the election list screen (API added in https://github.com/votingworks/vxsuite/pull/6051). Uses the same pattern from `CreateElectionButton` - Vx users get an org selection modal; single-org users get a new election created immediately) - we could explore adding a confirmation for the latter eventually, but easy enough to delete accidental copies at the moment.

### Vx Users:

https://github.com/user-attachments/assets/163c3257-18b9-4ecc-8e24-201a2dcc667b

### Single-Org Users (Creation Enabled):

https://github.com/user-attachments/assets/fcc04aca-7e14-4048-ba3c-3ca0350b9cf6

### Creation Disabled:

Screenshot 2025-02-27 at 10 58 40

## Checklist

### DIFF
--- a/apps/design/frontend/src/clone_election_button.test.tsx
+++ b/apps/design/frontend/src/clone_election_button.test.tsx
@@ -61,7 +61,7 @@ afterEach(() => {
   apiMock.assertComplete();
 });
 
-test('clones immediately for Vx users', async () => {
+test('clones immediately for non-Vx users', async () => {
   mockUseUserFeatures.mockReturnValue(userFeatureConfigs.nh);
 
   const { election } = generalElectionRecord(nonVxUser.orgId);
--- a/apps/design/frontend/src/clone_election_button.test.tsx
+++ b/apps/design/frontend/src/clone_election_button.test.tsx
@@ -1,0 +1,132 @@
+import { afterEach, beforeEach, expect, test, vi } from 'vitest';
+import { createMemoryHistory } from 'history';
+import { User } from '@votingworks/design-backend';
+import userEvent from '@testing-library/user-event';
+import { ElectionId } from '@votingworks/types';
+import { sleep } from '@votingworks/basics';
+import {
+  createMockApiClient,
+  MockApiClient,
+  nonVxUser,
+  provideApi,
+  vxUser,
+} from '../test/api_helpers';
+import { generalElectionRecord } from '../test/fixtures';
+import { render, screen, within } from '../test/react_testing_library';
+import { CloneElectionButton } from './clone_election_button';
+import { userFeatureConfigs, useUserFeatures } from './features_context';
+import { generateId } from './utils';
+import { withRoute } from '../test/routing_helpers';
+import { routes } from './routes';
+import * as api from './api';
+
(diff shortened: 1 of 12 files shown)


apps/design/frontend/vitest.config.ts
```diff
@@ -9,10 +9,15 @@ export default defineConfig({
 
     coverage: {
       thresholds: {
-        lines: 92,
-        branches: 72,
+        lines: 84.9,
```
kofi-q, line 12: Guessing test files with higher coverage were skewing the higher before - dropped a bit after filtering out test files from coverage

apps/design/frontend/src/clone_election_button.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const history = useHistory();
+  const features = useUserFeatures();
+  const user = api.getUser.useQuery().data;
+
+  const [orgId, setOrgId] = React.useState<string | undefined>(user?.orgId);
+  const [modalActive, setModalActive] = React.useState(false);
+
+  const cloneMutation = api.cloneElection.useMutation();
+  const mutateCloneElection = cloneMutation.mutate;
```
kshen0, line 72: Just curious - why not inline in L76 as `cloneMutation.mutate(...`?

apps/design/frontend/src/clone_election_button.tsx
```diff
@@ -0,0 +1,140 @@
...
+  return (
+    <React.Fragment>
+      <ButtonContainer>
+        <Tooltip role="tooltip">{`Make a copy of ${election.title}`}</Tooltip>
+        <Button
+          variant={variant}
+          onPress={features.ACCESS_ALL_ORGS ? setModalActive : cloneElection}
+          aria-label={`Make a copy of ${election.title}`}
+          value
+          disabled={modalActive}
```
kshen0, line 95: Should this be disabled if `cloneMutation.isLoading`?

apps/design/frontend/src/clone_election_button.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+beforeEach(() => {
+  apiMock = createMockApiClient();
+});
+
+afterEach(() => {
+  apiMock.assertComplete();
+});
+
+test('clones immediately for Vx users', async () => {
```
kshen0, line 64: Should read "non-Vx users"?

kshen0 (approved):

apps/design/frontend/src/clone_election_button.tsx
kofi-q: Ah, mostly working around an overly eager eslint check - would ideally put `cloneMutation.mutate` in the dependency list and then use it directly in the callback, but that triggers an eslint requiring `cloneMutation` to be in the deps list

PR #6063, opened 2025-02-27
Various stress test app fixes

eventualbuddha:
- make Mario show up in scan and mark-scan
- fix PDF parse error in scan


arsalansufi (approved):
Nice!

Issue #5656, opened 2024-12-03
Write-in adjudication handling when overvotes are allowed


adghayes:
Relevant threads:
- https://votingworks.slack.com/archives/CJU9MSC6S/p1732143593055329
- https://votingworks.slack.com/archives/C07QH5HUBH9/p1732298174667139

nikhilb4a:
### Overview

**Key finding**
As previously found, for both marked and unmarked write-in, we see that post-adjudication the write-in candidate has a count of 1 when it should have a count of 0. Existing candidate vote correctly has count of 0. Overvote correctly has count of 1.

**Additional oddities**
1. For marked write-in overvote, on polls closed we show the write-in count is 0 (and overvote is 1), but on Admin pre-adjudication we show the unadjudicated write-in count is 1 (and overvote is 1).
2. For unmarked write-in overvote, on ingest we warn the overvote will not count, but the tally report shows a vote only for marked existing candidate without an overvote. This is likely correct because Drew mentioned that before a write-in is adjudicated, we consider it an undervote.
3. For unmarked write-in overvote, the transition from pre to post adjudication is strange when there is a valid write-in. Pre-adjudication existing candidate A has vote=1, overvote =0, write-in=0. Post-adjudication, candidate A has vote =0, overvote=1, write-in=1. In reality this is the same key bug, as the only issue is that write-in should be 0. It is just especially odd right now that all 3 counts flip.

### Testing

**Marked write-ins in an overall overvote (marked existing candidate A + marked write-in)**

1. VxScan ballot input
- On ingest, warning that there is an overvote and the vote will not be counted. Cast ballot as is.
2. VxScan pre-adjudication tally report printed on polls close
- 1 ballot cast. Marked existing candidate A count is 0. Write-in count is 0. Overvote count is 1.
3. VxAdmin pre-adjudication tally report
- 1 ballot cast. Marked existing candidate A count is 0. *Unadjudicated* write-in count is 1. Overvote is 1.
4. VxAdmin write-in report pre-adjudication
- 1 total write-ins. 1 not adjudicated.

*If write-in is marked as undervote:*

1. VxAdmin post-adjudication tally report
- Marked existing candidate A count 1. Write-in count is 0. Overvote count is 0.
2. VxAdmin write-in report post-adjudication
- 1 total write-in, 1 invalid

*If write-in is marked as valid write-in candidate “Test Candidate”:*

1. VxAdmin post-adjudication tally report
[...]

PR #6070, opened 2025-03-01
Handle unadjudicated unmarked write-ins

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/5656

As I started working with Unmarked Write-Ins today, I realized I had previously mishandled filtering overvotes for _Unadjudicated and Unmarked_ write-ins from the tally report. 

**Context:**
Our initial write-in tally code had a bug in it where we would still add tally marks for Write-In candidates on the tally report even when the vote was part of an overvote.

I then [made a fix](https://github.com/votingworks/vxsuite/pull/5815) for this so we filter those votes out. This worked, but it mishandled Unmarked Write-ins, as those were never tabulated to begin.

**Example**
1. Two ballot contests are cast, both Overvotes. The first votes for candidate A and Write-in, the second votes for candidate A and Unmarked Write-in.
2. Pre-adjudication, the tally report should show an Overvote from ballot 1, and a vote for Candidate A on ballot 2. This is because the UWI from ballot 2 is counted as an undervote before adjudication. NOTE: The **initial** bug was that the Write-In from ballot 1 would contribute a count of 1 to Unadjudicated Write-in, even though it should not since it is part of an overvote. The **current** bug that this PR fixes is that Write-In count would actually show -1, because the pending unmarked write-in would incorrectly decrement from the Unadjudicated write-in count, even though it never contributed to it since we already ignore unmarked write-ins.
3. We adjudicate both write-ins as valid, thus now the second ballot contest is registered as an overvote. Now the tally report should show as 2 overvotes, and no additional votes.

## Demo Video or Screenshot

Screenshot 2025-02-28 at 4 02 04 PM
This screenshot shows an incorrect Write-In count of 0, because a pending unmarked write-in from an overvote incorrectly decremented a count from Unadjudicated Write-In, when there should be 1. You can confirm because the tally votes do not add up to seats * ballots.

Screenshot 2025-02-28 at 4 01 40 PM
This screenshot shows the fixed count, now with an Unadjudicated Write-In count of 1.

## Testing Plan

Added test case and manually tested.

## Checklist

### DIFF
--- a/apps/admin/backend/src/tabulation/write_ins.test.ts
+++ b/apps/admin/backend/src/tabulation/write_ins.test.ts
@@ -596,6 +596,15 @@ test('filterOvervoteWriteInsFromElectionResults', async () => {
       votes: { 'zoo-council-mammal': ['lion'] },
       card: { type: 'bmd' },
     },
+    {
+      ballotStyleGroupId: '1M' as BallotStyleGroupId,
+      batchId: 'batch-1-1',
+      scannerId: 'scanner-1',
+      precinctId: 'precinct-1',
+      votingMethod: 'precinct',
+      votes: { 'zoo-council-mammal': ['write-in-unmarked', 'lion'] },
+      card: { type: 'bmd' },
+    },
   ];
   addMockCvrFileToStore({ electionId, mockCastVoteRecordFile, store });
 
@@ -617,7 +626,7 @@ test('filterOvervoteWriteInsFromElectionResults', async () => {
             votesAllowed: 1,
             overvotes: 2,
             undervotes: 0,
-            ballots: 3,
+            ballots: 4,
             tallies: {
               [writeInId || Tabulation.GENERIC_WRITE_IN_ID]: {
                 tally: writeInTally,
@@ -635,7 +644,7 @@ test('filterOvervoteWriteInsFromElectionResults', async () => {
           },
         },
         cardCounts: {
-          bmd: 3,
+          bmd: 4,
           hmpb: [],
         },
       },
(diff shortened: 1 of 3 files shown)


apps/admin/backend/src/tabulation/write_ins.ts
```diff
@@ -455,6 +455,7 @@ export function filterOvervoteWriteInsFromElectionResults({
     const isOvervote = cvrContestVotes.length > contestResult.votesAllowed;
     if (isOvervote) {
       if (writeIn.status === 'pending') {
+        if (writeIn.isUnmarked) continue; // pending unmarked write-ins do not contribute to tallies
```
nikhilb4a, line 458: This is the one line change

apps/admin/backend/src/tabulation/write_ins.test.ts
```diff
@@ -596,6 +596,15 @@ test('filterOvervoteWriteInsFromElectionResults', async () => {
...
     },
+    {
+      ballotStyleGroupId: '1M' as BallotStyleGroupId,
+      batchId: 'batch-1-1',
+      scannerId: 'scanner-1',
+      precinctId: 'precinct-1',
+      votingMethod: 'precinct',
+      votes: { 'zoo-council-mammal': ['write-in-unmarked', 'lion'] },
+      card: { type: 'bmd' },
+    },
```
nikhilb4a, line 607: This test adds an Unmarked Write-in overvote. It should increase the count of Lion by 1 before adjudication, and should have no effect on counts after adjudication + filtering overvotes. 

That is why line 657 does increment by 1, as Lion has 1 more vote before adjudication, but 771 does not increment, as Lion vote count remains the same after adjudication.

And that is why checks on 668 and 766 do not change, as the unmarked write-in does not contribute to tallies before adjudication nor after adjudication/overvote filtering.

kshen0 (approved):

Issue #4476, opened 2024-01-04
Prevent contest navigation w/ accessible controller when in write-in modal


kshen0:

adghayes:
@kshen0 is this still relevant?

arsalansufi:
Cleaning up some old tickets. I believe addressed as part of https://github.com/votingworks/vxsuite/pull/6071

PR #6071, opened 2025-03-01
VxMark: rework ATI controller for write ins

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/4476

Changes bindings on the write-in virtual keyboard to support navigating between rows using the up and down keys of the ATI controller.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/2cba988f-6823-48c9-9907-611848bbb237

## Testing Plan
- [x] add automated tests
- [x] manually tested in mark-scan
- [x] manually tested in mark for backwards compatibility:

Screenshot 2025-03-28 at 5 36 26 PM
Screenshot 2025-03-28 at 5 35 49 PM

## Checklist


libs/ui/src/virtual_keyboard.tsx
```diff
@@ -1,275 +0,0 @@
-import React from 'react';
```
kshen0, line 1: Duplicate of `libs/ui/src/virtual_keyboard/virtual_keyboard.tsx` that was never removed in a previous refactor.

libs/ui/src/virtual_keyboard/virtual_keyboard.test.tsx
```diff
@@ -246,35 +262,16 @@ test('navigation with up and down arrow', async () => {
...
-  await pressKeyAndExpectFocus('[ArrowDown]', 'A');
-  await pressKeyAndExpectFocus('[ArrowRight]', 'S');
-  await pressKeyAndExpectFocus('[ArrowRight]', 'D');
-  await pressKeyAndExpectFocus('[ArrowRight]', 'F');
-  await pressKeyAndExpectFocus('[ArrowRight]', 'G');
-  await pressKeyAndExpectFocus('[ArrowRight]', 'H');
-  await pressKeyAndExpectFocus('[ArrowRight]', 'J');
-
-  // Now on the right side of the keyboard; go up and wrap around to start
-  await pressKeyAndExpectFocus('[ArrowDown]', 'N');
```
kshen0: It seems like the new method of using rendered bounding boxes to find the x-coordinate of the target key in the prev/next row doesn't play nicely with tests. I don't think there's a quick solution so removed these tests.

libs/ui/src/virtual_keyboard/virtual_keyboard.tsx
```diff
@@ -206,29 +227,266 @@ export const SPACE_BAR_KEY: Key = {
...
+          );
+          const targetRow = rowRefs.current[targetRowIndex];
+          if (targetRow) {
+            const buttons = Array.from(targetRow.querySelectorAll('button'));
+            buttons[targetKeyIndex]?.focus();
+            setFocusedRowIndex(targetRowIndex);
+          }
+          preventBrowserScroll(event);
+          break;
+        }
```
jonahkagan, line 404: It might be a little more React-y to compute where to focus next from data rather than from the DOM. Since `keyMapWithActions` is a 2d array representing the keyboard, you could track the current focused element as a coordinate `(row, column)` in state and then pick out the next button to focus using those coordinates and the colspans defined in the key map. You could even attach a ref to each key in the key map.

I'm also fine with how it's currently implemented. It works and is clear enough. Just thought I'd throw out the idea

libs/ui/src/virtual_keyboard/virtual_keyboard.tsx
```diff
@@ -206,29 +227,266 @@ export const SPACE_BAR_KEY: Key = {
...
 }: VirtualKeyboardProps): JSX.Element {
+  const [focusedRowIndex, setFocusedRowIndex] = useState(-1);
+
+  const keyMapWithActions: KeyMap = useMemo(() => {
+    const actions = enableWriteInAtiControllerNavigation
+      ? [
+          [SPACE_BAR_KEY, DELETE_KEY],
+          [CANCEL_KEY, ACCEPT_KEY],
+        ]
+      : // Cancel and Accept keys are rendered outside this component when ATI Controller navigation is off
```
jonahkagan, line 366: Any reason not to just always render them in this component? Seems like it would remove some complexity

apps/mark-scan/frontend/src/pages/contest_screen.tsx
```diff
@@ -23,13 +25,34 @@ function getStartPageUrl() {
...
 
+  // handleKeyboardEvent is defined in mark-scan and must be unbound here rather
+  // than in libs/ui
+  const onOpenWriteInKeyboard = useCallback(() => {
+    /* istanbul ignore next - @preserve */
+    if (!isPatDeviceConnected) {
+      document.removeEventListener('keydown', handleKeyboardEvent);
+      setKeyEventListenerWasUnbound(true);
+    }
+  }, [isPatDeviceConnected]);
```
jonahkagan, line 45: Do we have to add the keydown listeners for the write-in keyboard to the document, or could we add them just to the keyboard component? I'm thinking if they worked just on the keyboard component, then we could call `event.stopPropagation` to prevent this document-level listener from triggering, which would remove the need to manage it here.

jonahkagan (commented):
The navigation looks really smooth now! Nice!

I left some technical implementation questions

libs/ui/src/virtual_keyboard/virtual_keyboard.tsx
kshen0: I tried this initially but couldn't figure out a way to account for the gap between the keys defined [here](https://github.com/votingworks/vxsuite/pull/6071/files#diff-455a38a3cc5ca97747ac07039f626ce029a92b02f3f6f75f59e08af01e9deafeR32). Since we're mixing units (abstract column spans for width and pixels for gap) it seemed much easier to let the browser take care of rendering and then make the calculation after.

I agree the render-first approach isn't very React-y. The only way I could think to do the calculation before render would be to abandon the column span approach and give keys their own fixed pixel width, which has its own set of issues.

libs/ui/src/virtual_keyboard/virtual_keyboard.tsx
kshen0: Mostly wanted to avoid reimplementing

1. action keys in ScanPanelVirtualKeyboard (the sip & puff keyboard)
2. the landscape-orientation sidebar for action keys in VxMark v2

apps/mark-scan/frontend/src/pages/contest_screen.tsx
kshen0: Great suggestion! This unlocks a pretty large simplification

https://github.com/votingworks/vxsuite/pull/6071/commits/a4b1ff3a65160a99718bfa470da3786c8eb474e8

libs/ui/.vscode/settings.json
```diff
@@ -15,10 +15,10 @@
     "editor.formatOnSave": false
   },
   "[typescript]": {
-    "editor.formatOnSave": false
+    "editor.formatOnSave": true
   },
   "[typescriptreact]": {
-    "editor.formatOnSave": false
+    "editor.formatOnSave": true
```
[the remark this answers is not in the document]
kshen0: I'm not sure how these made it in, thanks for catching!
[e27f954](https://github.com/votingworks/vxsuite/pull/6071/commits/e27f95497a9adf977489bfb597bf5383901cb6d2)

libs/ui/src/virtual_keyboard/virtual_keyboard.tsx
jonahkagan: Ah yeah I see what you mean. I just looked more closely at the keyboard and realized that the keys don't line up easily into columns. I think given that design, I see why we're constrained to this solution. I do think there's a case to make that the keys could be instead visually grouped into columns for clarity in navigation (since the normal keyboard layout is optimized for fingers, not controllers), but given that it's already finished I don't think it's worth changing.

apps/mark-scan/frontend/src/pages/contest_screen.tsx
jonahkagan: Glad to hear it! It took me a while to find this solution so I'm glad it ended up working haha

jonahkagan (approved):
🚢

Issue #6079, opened 2025-03-04
VxDesign: Make it easy to deploy review apps


jonahkagan:
- Specify a database to provision (I think you can use app.json for this)
- Parse the db creds from the Heroku-provided DATABASE_URL env var
- Figure out a way to default BASE_URL to the review app URL automatically
- Figure out a way to bypass or configure Auth0 (since it currently requires knowing the app’s callback URL, which includes the hostname)

jonahkagan:
Documenting that I was able to use a wildcard in the Auth0 callback URL

PR #6081, opened 2025-03-04
design/backend: Configure PG client with `DATABASE_URL` env var

jonahkagan:
## Overview
Task: #6079 

This is the default env var set by Heroku's Postgres add-on, so we can use that instead of having to manually set `PGUSER`, `PGPASSWORD`, etc. This will be particularly helpful for creating review apps.
## Demo Video or Screenshot
N/A

## Testing Plan
Will test in staging by removing `PG...` env vars

## Checklist

### DIFF
--- a/apps/design/backend/package.json
+++ b/apps/design/backend/package.json
@@ -26,7 +26,7 @@
     "pre-commit": "lint-staged",
     "start": "./scripts/start.sh",
     "test": "is-ci test:ci test:watch",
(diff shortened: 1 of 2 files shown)


kofi-q (approved):
🚀

jonahkagan:
Tested that this works in staging (and I removed the `PG...` env vars). cc @arsalansufi this change will roll out to prod whenever the next deploy happens, and then we can remove the env vars there as well

PR #6084, opened 2025-03-05
Fix flaky test

nikhilb4a:
## Overview

This test has been flaky, with this [line](https://github.com/votingworks/vxsuite/blob/main/apps/admin/backend/src/tabulation/write_ins.ts#L467) not being consistently hit. 

I was able to reproduce locally about half the time. The ordering of the `writeInRecords` returned from the store was inconsistent, so I updated the test to adjudicate based on optionId, instead of assuming a certain ordering. This ordering mattered because if the "unmarked" write-in gets adjudicated as "invalid", then that cvr has no overvote, so the invalid overvote case is never hit.

## Checklist


jonahkagan (approved):
Thanks for digging into this!

PR #6085, opened 2025-03-06
apps/design: Move feature flag config to backend

jonahkagan:
## Overview

Closes: #6074 

Moves the feature flag config from the frontend to the backend, since the special cased org IDs (VX/SLI) are defined there. Adds endpoints to load this config on the frontend. This will also allow us to implement guards for features on the backend as well if needed.

## Demo Video or Screenshot
N/A

## Testing Plan
- Add minimal new test coverage for features endpoints
I chose not to update coverage for uncovered feature flag consumers because I now think we may change some of that soon anyways as we move towards having support tools.

## Checklist


jonahkagan:
@kshen0 I'd appreciate an initial review as I work on adding more test coverage

apps/design/frontend/src/create_election_button.tsx
```diff
@@ -58,6 +57,11 @@ export function CreateElectionButton(
     );
   }, [history, mutateCreateElection, orgId]);
 
+  if (!getUserFeaturesQuery.isSuccess) {
+    return null;
```
kshen0, line 61: I wonder if it's worth it to keep `FeaturesContext` around just to wrap this call. On error, it could return a top-level error page or default to a standard no-extras feature config. It would save us from this pattern and relieve individual consumers of the responsibility of deciding how to report errors.

apps/design/frontend/test/api_helpers.tsx
```diff
@@ -11,6 +17,55 @@ export function createMockApiClient(): MockApiClient {
...
+  EXPORT_TEST_DECKS: true,
+  MARGINAL_MARK_THRESHOLD: true,
+  ONLY_LETTER_AND_LEGAL_PAPER_SIZES: true,
+  CREATE_ELECTION: true,
+  DELETE_ELECTION: true,
+  CREATE_DELETE_DISTRICTS: true,
+  CREATE_DELETE_PRECINCTS: true,
+  CREATE_DELETE_PRECINCT_SPLITS: true,
+  BALLOT_LANGUAGE_CONFIG: true,
+};
```
kshen0, line 34: 💯 this is so much clearer than how tests did this before

apps/design/frontend/src/create_election_button.tsx
jonahkagan: This pattern is for loading, not errors. I believe errors will still use our default error boundary.

I could see a case for using a context to factor out the loading pattern. I considered it when starting this change, but decided against it for two reasons:
1. I thought it would make it more complicated to reason about the sequence of rendering/add a different pattern of data management to think about
2. I wasn't sure if every page would want to wait for features to load before displaying anything

I don't feel super strongly about those reasons, so I'm open to arguments for changing it if you think there are compelling advantages to a context

kshen0: Ah you are right. Since we already have standardized error handling I don't think there's a strong enough reason to push for a less flexible context at this point.

apps/design/frontend/src/clone_election_button.test.tsx
```diff
@@ -62,54 +54,48 @@ afterEach(() => {
...
 const VX_ORG = {
-  id: vxUser.orgId,
+  id: user.orgId,
   name: 'votingworks',
   displayName: 'VotingWorks',
 } as const;
 
 const NON_VX_ORG = {
-  id: nonVxUser.orgId,
+  id: user.orgId,
```
kshen0, line 84: Nit: could just hardcode these to avoid confusion from using same orgId for separate test orgs (where presumably the tests don't care about the IDs of the orgs)

apps/design/frontend/src/clone_election_button.test.tsx
```diff
@@ -62,54 +54,48 @@ afterEach(() => {
...
   name: 'not-voting-works',
   displayName: 'Not VotingWorks',
 } as const;
 
 test('shows org picker for Vx users', async () => {
```
kshen0, line 89: The test names might now be inconsistent with behavior. Is the user's org now the thing that determines whether the org picker is shown? Or is it the `ACCESS_ALL_ORGS` flag response from the backend?

kshen0 (approved):

apps/design/frontend/src/clone_election_button.test.tsx
jonahkagan: good call

PR #6088, opened 2025-03-07
[VxDesign] Add toggle for audio export

kshen0:
## Checklist


apps/design/frontend/src/features_context.tsx
```diff
@@ -69,6 +69,11 @@ enum UserFeature {
    * Allow the user to select ballot languages.
    */
   BALLOT_LANGUAGE_CONFIG = 'BALLOT_LANGUAGE_CONFIG',
+  /**
+   * Show the user a toggle that allows them to include audio files in
+   * ballot and election package export.
+   */
+  EXPORT_AUDIO_TOGGLE = 'EXPORT_AUDIO_TOGGLE',
```
jonahkagan, line 76: Do we need a feature flag for this? Seems like it's part of the feature of accessing the export screen

apps/design/backend/src/app.test.ts
```diff
@@ -929,11 +933,8 @@ vi.mock(import('@votingworks/hmpb'), async (importActual) => {
 });
 
 test.skip('Export all ballots', async () => {
-  // This test runs unnecessarily long if we're generating exports for all
-  // languages, so disabling multi-language support for this case:
-  mockFeatureFlagger.disableFeatureFlag(
-    BooleanEnvironmentVariableName.ENABLE_CLOUD_TRANSLATION_AND_SPEECH_SYNTHESIS
-  );
+  // TODO: This test runs unnecessarily long if we're generating exports for all
+  // languages, so disable multi-language support for this case.
```
jonahkagan, line 937: When/how do you think these TODOs will get resolved?

apps/design/backend/src/worker/generate_election_package_and_ballots.ts
```diff
@@ -191,11 +188,7 @@ export async function generateElectionPackageAndBallots(
...
   );
 
-  const isCloudTranslationAndSpeechSynthesisEnabled =
-    isVxOrSliOrg(orgId) &&
-    isFeatureFlagEnabled(
-      BooleanEnvironmentVariableName.ENABLE_CLOUD_TRANSLATION_AND_SPEECH_SYNTHESIS
-    );
+  const isCloudTranslationAndSpeechSynthesisEnabled = shouldExportAudio;
   const { uiStringAudioIds, uiStringAudioClips } = generateAudioIdsAndClips({
     isCloudTranslationAndSpeechSynthesisEnabled,
```
jonahkagan, line 193: Should we just omit the audio files completely if `shouldExportAudio` is false rather than returning an empty file?

apps/design/frontend/src/features_context.tsx
kshen0: I thought it might be useful if we eventually do a general release of the export page, but we can simplify now and cross that bridge when we get there.

apps/design/backend/src/app.test.ts
```diff
@@ -885,7 +895,7 @@ test.skip('Election package export', async () => {
   //
   // Check uiStringAudioIds.json
   //
-  expect(countObjectLeaves(uiStringAudioIds)).toEqual(
+  expect(countObjectLeaves(uiStringAudioIds)).toBeGreaterThanOrEqual(
```
kshen0, line 898: Still figuring out why this went up

apps/design/backend/src/app.test.ts
```diff
@@ -745,11 +747,16 @@ test.skip('Election package export', async () => {
...
   });
 
+  const contents = assertDefined(
+    fileStorageClient.getRawFile(
+      join(electionRecord.orgId, electionPackageFilePath)
+    )
+  );
   const { electionPackage, electionPackageHash } = (
-    await readElectionPackageFromFile(electionPackageFilePath)
+    await readNestedElectionPackageFromBuffer(contents)
```
kshen0, line 759: Since this test was disabled, election package export has changed. Instead of exporting `election-package-<hash>.zip` it now exports a zip with 2 zips:
* `election-package-<hash>.zip`
* `ballots-<hash>.zip`

apps/design/backend/src/app.test.ts
kshen0: Sorry, left this as a reminder to myself then forgot to follow up. TLDR these TODOs can be deleted.

`ENABLE_CLOUD_TRANSLATION_AND_SPEECH_SYNTHESIS` is now a misnomer because it covers speech synthesis only. Cloud translation is on by default, but selection of languages other than English is not. So in all tests
* cloud text translation is off by default because test elections default to English-only
  * I re-enabled [this test](https://github.com/votingworks/vxsuite/pull/6088/commits/5687aeede41513924cc158d22047fa2e6e4007ac#diff-f727b4656b5559f4f63b491188fb8bc2db591dd63fa5384f7ecdca56acf509deL801) and removed the expectation for translated text strings
* speech synthesis is controlled just by the `shouldExportAudio` job payload param [here](https://github.com/votingworks/vxsuite/pull/6088/commits/5687aeede41513924cc158d22047fa2e6e4007ac#diff-c268bf2996342639e8bfd885b518b620095551a7b47c8517f93bab768c57bfd0R191), which is set by the frontend [here](https://github.com/votingworks/vxsuite/pull/6088/files#diff-b6a0a477a5d9762712d944e9bee4f62823033e0d7cc1d0a4881c79ecd01caf29R112)
* TODOs and flag configuration removed in https://github.com/votingworks/vxsuite/pull/6088/commits/753358e7fa9a74815a9af2211410276d9575430b

apps/design/backend/src/worker/generate_election_package_and_ballots.ts
kshen0: Caller logic simplified [here](https://github.com/votingworks/vxsuite/pull/6088/commits/5687aeede41513924cc158d22047fa2e6e4007ac#diff-c268bf2996342639e8bfd885b518b620095551a7b47c8517f93bab768c57bfd0R191) and param removed [here](https://github.com/votingworks/vxsuite/pull/6088/commits/5687aeede41513924cc158d22047fa2e6e4007ac#diff-1022e77fe3d00d85799195dd201761ed822cfc114b1548a88587b1549dbf5ec1L39).

apps/design/backend/src/app.test.ts
jonahkagan: Awesome, sounds like a big simplification!

jonahkagan (approved):

apps/design/backend/src/app.test.ts
kshen0: [e1beae4](https://github.com/votingworks/vxsuite/pull/6088/commits/e1beae4e2f9b89bbd9c72444022bf811d69c15fd)

Issue #6080, opened 2025-03-04
VxDesign: Finish backend form validation


jonahkagan:
Instead of having a single `updateElection` endpoint for all election content, make a separate endpoint for each form on the frontend. Then add backend validation to each endpoint using schemas. We already started implementing this pattern with the election info screen, but need to finish by converting all of the other screens.

PR #6093, opened 2025-03-12
VxDesign: Backend validation for districts form

jonahkagan:
## Overview

Task: #6080 

Adds backend validation to the districts form. Adds new CRUD API methods for districts and validates input on the backend. Since there is also frontend validation, we don't actually expect invalid input to be sent to the backend, so we just throw errors if invalid input occurs.

## Demo Video or Screenshot
N/A - no functionality changed

## Testing Plan
Updated existing tests

## Checklist

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -26,7 +26,6 @@ import {
 } from '@votingworks/types';
 import express, { Application } from 'express';
 import {
-  assert,
   assertDefined,
   DateWithoutTime,
   find,
@@ -342,8 +341,7 @@ function buildApi({ auth, workspace, translator }: AppContext) {
     async listDistricts(input: {
       electionId: ElectionId;
     }): Promise<readonly District[]> {
-      const { election } = await store.getElection(input.electionId);
-      return election.districts;
+      return store.listDistricts(input.electionId);
     },
 
     async createDistrict(input: {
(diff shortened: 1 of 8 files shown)


apps/design/backend/src/app.ts
```diff
@@ -335,6 +340,88 @@ function buildApi({ auth, workspace, translator }: AppContext) {
...
+          };
+        }
+        return {
+          ...precinct,
+          districtIds: precinct.districtIds.filter(
+            (id) => id !== input.districtId
+          ),
+        };
+      });
+      await store.updatePrecincts(input.electionId, updatedPrecincts);
```
kshen0, line 422: Should the calls to `updateElection` and `updatePrecincts` be done as a single transaction? Alternatively, should the check on L390 be removed so this operation is idempotent and can be retried?

jonahkagan: Good call, updated: b6bca6b1e

Noting that the rest of the app doesn't really use transactions at all yet, so there's probably more work do be done on that front

apps/design/backend/src/app.ts
```diff
@@ -335,6 +339,55 @@ function buildApi({ auth, workspace, translator }: AppContext) {
...
+      });
+      const { election } = await store.getElection(input.electionId);
+      assert(
+        election.districts.some((d) => d.id === input.districtId),
+        'District not found'
+      );
+      const updatedDistricts = election.districts.map((d) =>
+        d.id === input.districtId ? district : d
+      );
+      await store.updateElection(input.electionId, {
```
eventualbuddha, line 378: Should we move this operation into a `Store::updateDistrict` method? It's a little odd for this RPC method to use `updateElection` and the `deleteDistrict` one to use `Store::deleteDistrict`. Moving toward having an abstraction around that will enable us to move toward having dedicated tables for districts/precincts/etc as you suggested earlier.

eventualbuddha (approved):

apps/design/backend/src/app.ts
jonahkagan: I agree that's the eventual direction we should go. I don't see a strong reason to do it now, but also don't have a strong reason not to do it now haha. So I'm down to do it now.

To be clear, under the hood `Store::updateDistrict` would be exactly the same code as is currently in `App::updateDistrict`. Is that what you're imagining too?

kshen0 (approved):

apps/design/backend/src/app.ts
jonahkagan: 174e6f3af

PR #6095, opened 2025-03-12
VxDesign: Backend validation for precincts form 

jonahkagan:
## Overview

Task: #6080 

Builds on the patterns established for districts in #6093 to validate precincts.

## Demo Video or Screenshot
N/A - no change in functionality

## Testing Plan
Updated existing tests

## Checklist

### DIFF
--- a/apps/design/frontend/src/geography_screen.test.tsx
+++ b/apps/design/frontend/src/geography_screen.test.tsx
@@ -534,11 +534,7 @@ describe('Precincts tab', () => {
     });
 
     apiMock.updatePrecinct
-      .expectCallWith({
-        electionId,
-        precinctId: changedPrecinct.id,
-        updatedPrecinct: changedPrecinct,
-      })
+      .expectCallWith({ electionId, updatedPrecinct: changedPrecinct })
       .resolves();
     apiMock.listPrecincts
       .expectCallWith({ electionId })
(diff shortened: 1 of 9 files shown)


apps/design/frontend/src/geography_screen.test.tsx
```diff
@@ -308,26 +316,25 @@ describe('Precincts tab', () => {
...
+        newPrecinct,
       })
       .resolves();
-    apiMock.getElection
-      .expectCallWith({ user, electionId })
-      .resolves(electionWithNewPrecinctRecord);
+    apiMock.listPrecincts
+      .expectCallWith({ electionId })
+      .resolves([newPrecinct]);
+    // Districts haven't changed, but we are using coarse-grained invalidation
```
kshen0, line 328: Is it necessary to use coarse invalidation ie. `invalidateElectionQueries` when only precincts have changed? Changing districts may change precincts, but not vice versa.

OTOH I see the value of simplifying because precincts and districts both are stored in the `elections` table, cache invalidation is hard, and the performance change (if any) is probably not noticeable.

jonahkagan: I was thinking along the lines of your second paragraph - it's hard to get cache invalidation right, so it makes sense to start with the safest thing and optimize as needed.

kshen0 (approved):

PR #6097, opened 2025-03-13
VxDesign: Refactor districts/precincts forms

jonahkagan:
## Overview

I realized that these forms could be a bit simpler and more straightforward if they didn't take in the entire list of saved entities, but rather only the one that was being edited.
## Demo Video or Screenshot
N/A

## Testing Plan
Existing test coverage
## Checklist

### DIFF
--- a/apps/design/frontend/src/api.ts
+++ b/apps/design/frontend/src/api.ts
@@ -255,6 +255,7 @@ export const createDistrict = {
     return useMutation(apiClient.createDistrict, {
       async onSuccess(_, { electionId }) {
         await invalidateElectionQueries(queryClient, electionId);
+        await queryClient.refetchQueries(listDistricts.queryKey(electionId));
       },
     });
   },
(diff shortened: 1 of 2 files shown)


apps/design/frontend/src/api.ts
```diff
@@ -255,6 +255,7 @@ export const createDistrict = {
     return useMutation(apiClient.createDistrict, {
       async onSuccess(_, { electionId }) {
         await invalidateElectionQueries(queryClient, electionId);
+        await queryClient.refetchQueries(listDistricts.queryKey(electionId));
```
jonahkagan, line 258: Since the `listDistricts` query is no longer mounted on the Add District form, we have to refetch it during the mutation so that it's up to date with the new district before navigating back to the districts list (otherwise there's a lag before it shows up)

kshen0 (approved):

PR #6101, opened 2025-03-17
VxDesign: Backend validation for ballot layout form

jonahkagan:
## Overview
Task: #6080 

Continues the pattern of making separate endpoints for each form to validate the ballot layout form. 

## Demo Video or Screenshot
N/A - no functional change

## Testing Plan
Updated existing tests, added some new ones

## Checklist

### DIFF
--- a/apps/design/frontend/src/api.ts
+++ b/apps/design/frontend/src/api.ts
@@ -168,17 +168,30 @@ export const listContests = {
   },
 } as const;
 
+export const getBallotPaperSize = {
+  queryKey(id: ElectionId): QueryKey {
+    return ['getBallotPaperSize', id];
+  },
+  useQuery(id: ElectionId) {
+    const apiClient = useApiClient();
+    return useQuery(this.queryKey(id), () =>
+      apiClient.getBallotPaperSize({ electionId: id })
+    );
+  },
+} as const;
+
 async function invalidateElectionQueries(
   queryClient: QueryClient,
   electionId: ElectionId
 ) {
+  await queryClient.invalidateQueries(listElections.queryKey());
   await queryClient.invalidateQueries(getElection.queryKey(electionId));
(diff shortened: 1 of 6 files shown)


apps/design/frontend/src/api.ts
```diff
@@ -168,17 +168,30 @@ export const listContests = {
...
 ) {
+  await queryClient.invalidateQueries(listElections.queryKey());
   await queryClient.invalidateQueries(getElection.queryKey(electionId));
   await queryClient.invalidateQueries(getElectionInfo.queryKey(electionId));
   await queryClient.invalidateQueries(listDistricts.queryKey(electionId));
   await queryClient.invalidateQueries(listPrecincts.queryKey(electionId));
   await queryClient.invalidateQueries(listParties.queryKey(electionId));
   await queryClient.invalidateQueries(listContests.queryKey(electionId));
-  await queryClient.invalidateQueries(listElections.queryKey());
+  await queryClient.invalidateQueries(getBallotPaperSize.queryKey(electionId));
```
eventualbuddha, line 194: Perhaps we should change the query keys so we don't have to keep adding to this list. Per the [docs](https://tanstack.com/query/latest/docs/framework/react/guides/query-invalidation), if we used a query key like `['election', id, 'getBallotPaperSize']` for all the queries for an election then we can just invalidate `['election', id]` to invalidate everything for the election.

eventualbuddha (approved):

apps/design/frontend/src/api.ts
jonahkagan: Yeah, I like that idea. It's been on the back of my mind as well. I haven't done it yet because I'm not sure if we'll want to stick with the pattern of invalidating all election queries at once or whether we'll want to do more fine-grained invalidation. Though this change wouldn't preclude that. Created an issue to track both together so we don't forget to revisit this: https://github.com/votingworks/vxsuite/issues/6106

PR #6103, opened 2025-03-17
[Design] Add DB schema/data migration support

kofi-q:
## Overview

Replacing `schema.sql` with an initial migration script and an integration with `node-pg-migrate` to manage migrations.

Integration points:
- pnpm scripts (`db:migrations:create`, `db:migrations:run`, `db:migrations:undo-last`, along with `-dev` variants).
  - The `db:reset` script now uses `pnpm db:migrations:run` in place of the old schema script.
- Using the programmatic API in the DB code for setting up temporary test schemas.
- Heroku `release` step, which reuses the `web` docker image and runs `pnpm db:migrations:run` to set up/update the schema before starting the server and worker.

## Testing Plan
- Tested locally on blank and pre-populated DBs to verify no data is overwritten
- Tested on a fresh DB instance in Heroku (review app)
- Tested data persistence after initial migration in [staging](https://dashboard.heroku.com/apps/vxdesign-staging/activity/releases/fcda3dc0-3b51-4904-8a7c-7ab49c334152) 

## Checklist


apps/design/backend/package.json
```diff
@@ -21,7 +19,12 @@
     "clean:self": "rm -rf build && tsc --build --clean tsconfig.build.json",
     "create-org": "node -r esbuild-runner/register ./scripts/create_org.ts",
     "create-user": "node -r esbuild-runner/register ./scripts/create_user.ts",
-    "db:reset": "./scripts/db_reset_dev.sh",
+    "db:migrations:create": "node-pg-migrate create",
+    "db:migrations:run": "node-pg-migrate up",
+    "db:migrations:run-dev": "DATABASE_URL='postgresql://design:design@localhost:5432/design' node-pg-migrate up",
+    "db:migrations:undo-last": "node-pg-migrate down",
```
jonahkagan, line 25: How do the down migrations work? I see that node-pg-migrate claims it can infer them, but I imagine there are a bunch of cases where it's not obvious without writing a reverse migration how to undo.

In Arlo we just decided to forgo down migrations and have never needed them. I don't think it's something we need to decide right now, but mostly just wondering your thoughts

apps/design/backend/package.json
```diff
@@ -21,7 +19,12 @@
     "clean:self": "rm -rf build && tsc --build --clean tsconfig.build.json",
     "create-org": "node -r esbuild-runner/register ./scripts/create_org.ts",
     "create-user": "node -r esbuild-runner/register ./scripts/create_user.ts",
-    "db:reset": "./scripts/db_reset_dev.sh",
+    "db:migrations:create": "node-pg-migrate create",
+    "db:migrations:run": "node-pg-migrate up",
+    "db:migrations:run-dev": "DATABASE_URL='postgresql://design:design@localhost:5432/design' node-pg-migrate up",
```
jonahkagan, line 24: I wonder if we could combine each of these command pairs into a single command that uses the env's `DATABASE_URL` if set, otherwise uses the default dev db URL. Might just simplify usage a tad

apps/design/frontend/README.md
```diff
@@ -41,6 +36,90 @@ then install `postgresql-16`:
...
+## Running the app
+
+In `apps/design/frontend`, run the following:
+
+```sh
+pnpm start
+```
+
+The server will be available at http://localhost:3000.
+
```
jonahkagan, line 122: A+ docs updates 🙇

apps/design/backend/schema.sql
jonahkagan, line 1: Sometimes I think it's nice to also have a declarative schema to document the expected result of running all of the migrations. It makes it easier to read and grok the data model while developing.

The downside is that it's a second source of truth, but you can always write a test to mitigate the risk of them becoming out of sync. With that test in place, it also serves as a guard against making a mistake while writing a migration, because that way you have an expected result schema to compare to.

apps/design/backend/migrations/1742235043624_init.js
```diff
@@ -0,0 +1,101 @@
...
+      id: { type: 'text', primaryKey: true },
+      task_name: { type: 'text', notNull: true },
+      payload: { type: 'text', notNull: true },
+      created_at: {
+        type: 'timestamp',
+        notNull: true,
+        default: pgm.func('current_timestamp'),
+      },
+      started_at: { type: 'timestamp' },
+      completed_at: { type: 'timestamp' },
```
jonahkagan, line 24: should we take this opportunity to standardize on `timestamptz` or `timestamp`?

apps/design/backend/migrations/1742248030415_add-ballot-language-codes.js
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @param pgm {import('node-pg-migrate').MigrationBuilder}
+ * @param run {() => void | undefined}
+ * @returns {Promise<void> | void}
+ */
+exports.up = (pgm) => {
+  pgm.addColumn(
+    'elections',
+    {
+      ballot_language_codes: {
```
jonahkagan, line 15: I noticed that this column is already in the initial migration. Why have a separate migration for it as well?

jonahkagan (approved):
This looks great, very excited to have this tooling in place

apps/design/backend/package.json
jonahkagan: Cool, explicitness is a good enough reason for me!

apps/design/backend/schema.sql
jonahkagan: That sounds like a good solution to me

apps/design/backend/migrations/1742235043624_init.js
jonahkagan: Ah, makes sense

PR #6116, opened 2025-03-20
Add support for a remote card certification flow

arsalansufi:
## Overview

A little side project to take a task off my plate haha

We have a remote machine certification flow that we've used quite extensively to certify machines that are in the possession of someone who doesn't have acccess to the root VotingWorks private key. We don't yet have an analogous flow for card certification. This PR adds support for that and will make it such that others aren't blocked on the few of us with possession to the latest key when they need smart cards compatible with v4.0.1+ software images. Right now, we've been getting by by certifying and shipping cards around the country, which works but isn't great in a pinch.

## Demo Video or Screenshot

Person without access to the root VotingWorks private key:

```
NODE_ENV=production VX_PRIVATE_KEY_PATH=remote WORKING_DIRECTORY=/media/psf/Shared ./scripts/configure-java-card

[ Skipping irrelevant output lines ]

------------------------------------------------
🔏 Creating and storing card VotingWorks cert...
------------------------------------------------
A card-specific public key has been written to /media/psf/Shared/public-key-26337f.pem.
Share this file with someone who has the appropriate signing key.
They will return to you a cert-26337f.pem file.
Place that file in /media/psf/Shared and press enter. 
No /media/psf/Shared/cert-26337f.pem file found. Make sure that it exists and press enter to try again. 
--------
✅ Done!
--------
```

Person with access to the root votingWorks:

```
$ CERT_PUBLIC_KEY_PATH=/media/psf/Shared/public-key-26337f.pem VX_PRIVATE_KEY_PATH=/path/to/private-key.pem ./scripts/remote-card-vx-certifier 
Enter pass phrase for /path/to/private-key.pem:
✅ Cert written to /media/psf/Shared/cert-26337f.pem and can be shared with the sender of the public key
```

## Testing Plan

- [x] Added automated tests
- [x] Tested manually

## Checklist

### DIFF
--- a/libs/auth/src/java_card.ts
+++ b/libs/auth/src/java_card.ts
@@ -1,6 +1,9 @@
-import { Buffer } from 'node:buffer';
-import * as fs from 'node:fs/promises';
 import { sha256 } from 'js-sha256';
+import { Buffer } from 'node:buffer';
+import { existsSync } from 'node:fs';
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+import readline from 'node:readline/promises';
 import { v4 as uuid } from 'uuid';
 import {
   assert,
@@ -330,6 +333,7 @@ export class JavaCard implements Card {
     if (user.role === 'vendor') {
       assert(this.cardProgrammingConfig.configType === 'vx');
       const { vxPrivateKey } = this.cardProgrammingConfig;
+      assert(vxPrivateKey.source === 'file');
 
       // Store card identity cert
       const cardIdentityCert = await createCert({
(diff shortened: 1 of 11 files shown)


libs/auth/src/java_card.ts
```diff
@@ -854,24 +858,61 @@ export class JavaCard implements Card {
...
+    if (this.cardProgrammingConfig.vxPrivateKey.source === 'remote') {
+      assert(remoteCertParams !== undefined);
+      const { randomId, workingDirectory } = remoteCertParams;
+      const certPublicKeyPath = path.join(
+        workingDirectory,
+        `public-key-${randomId}.pem`
+      );
+      const certPath = path.join(workingDirectory, `cert-${randomId}.pem`);
+
+      await fs.writeFile(certPublicKeyPath, certPublicKey);
```
arsalansufi, line 883: For reference Ben, I opted to write a public key instead of a CSR for the exchange because, for cards, we're creating dummy CSRs with a throwaway private key and then using the -force_pubkey param to explicitly override the public key. If we did use the dummy CSR for the exchange here, we'd still need to write the card public key, too. At that point, might as well just drop the dummy CSR

libs/auth/src/java_card.ts
```diff
@@ -854,24 +858,61 @@ export class JavaCard implements Card {
...
+
+      const rl = readline.createInterface({
+        input: process.stdin,
+        output: process.stdout,
+      });
+      await rl.question(`A card-specific public key has been written to ${certPublicKeyPath}.
+Share this file with someone who has the appropriate signing key.
+They will return to you a ${path.basename(certPath)} file.
+Place that file in ${workingDirectory} and press enter. `);
```
arsalansufi, line 892: You'll see in the next commit the script that'll turn this card public key into a card cert

benadida (approved):

Issue #5989, opened 2025-02-06
v4.0.2 image creation


arsalansufi:
PRs tagged v4.0.2-candidate:
- [vxsuite](https://github.com/votingworks/vxsuite/issues?q=label%3Av4.0.2-candidate)
- [vxsuite-complete-system](https://github.com/votingworks/vxsuite-complete-system/issues?q=label%3Av4.0.2-candidate)

arsalansufi:
Closing this as we're now planning to bump up to main for v4.0.2 - this was mainly to track the (what would have been very involved cherry picking)

PR #6120, opened 2025-03-20
Fix bug with multi-language CDF imports to VxSuite

carolinemodic:
## Overview
When an election is exported in VxDesign in CDF format and is multi-language the IDs are created in the style: `{ballotStyleGroupID}_{languageCode}`. If we detect ballots in this format from VxDesign we should parse out the group ID from it so that ballots that are intended to be the same but different language varients of one another are handled properly. In order to use a multi-language election in CDF format as intended in the VxSuite system the ExternalIdentifiers must be specified in this format. 

## Demo Video or Screenshot
N/A 

## Testing Plan
Loaded a multi-language CDF election into VxAdmin, configured VxMark and VxScan and scanned hmpbs and verified language selector worked as expected. 

## Checklist

### DIFF
--- a/libs/types/src/cdf/ballot-definition/convert.ts
+++ b/libs/types/src/cdf/ballot-definition/convert.ts
@@ -1076,25 +1076,17 @@ export function convertCdfBallotDefinitionToVxfElection(
 
       // Check if this is a VxDesign formatted Id_LanguageCode Ballot Id.
       // If so extract the group ID from the first part of the Id.
-      if (
+      const useExtractedGroupId =
         ballotStyle.Language &&
         ballotStyle.Language.length === 1 &&
         idParts.length === 2 &&
-        idParts[1] === ballotStyle.Language[0]
-      ) {
-        return {
-          id: ballotStyleId as Vxf.BallotStyleId,
-          groupId: idParts[0] as Vxf.BallotStyleGroupId,
-          districts: districtIds,
-          precincts: precinctIds,
-          partyId: ballotStyle.PartyIds?.[0] as Vxf.PartyId | undefined,
-          languages: ballotStyle.Language,
-        };
-      }
+        idParts[1] === ballotStyle.Language[0];
 
       return {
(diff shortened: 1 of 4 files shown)


libs/types/test/election.ts
```diff
@@ -127,7 +127,7 @@ export const primaryElection: Election = {
     ...election.ballotStyles.map((bs) => ({
       ...bs,
       id: `${bs.id}R`,
-      groupId: `${bs.id}D`,
+      groupId: `${bs.id}R`,
```
carolinemodic, line 130: unrelated bug in a test fixture that was being hidden by the normalizeVxf function changing ballot style ids

libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -1070,10 +1070,31 @@ export function convertCdfBallotDefinitionToVxfElection(
...
+        idParts[1] === ballotStyle.Language[0]
+      ) {
+        return {
+          id: ballotStyleId as Vxf.BallotStyleId,
+          groupId: idParts[0] as Vxf.BallotStyleGroupId,
+          districts: districtIds,
+          precincts: precinctIds,
+          partyId: ballotStyle.PartyIds?.[0] as Vxf.PartyId | undefined,
+          languages: ballotStyle.Language,
+        };
```
arsalansufi, line 1092: nit: Since it's only the `groupId` that varies across the two cases (VxDesign-formatted ballot ID vs. not), maybe we just vary that instead of a fully separate return. Just to avoid accidentally updating one return and not the other down the road when we want to edit one of the other fields.

```ts
const groupId = isVxDesignFormattedBallotId ? idParts[0] : ballotStyleId;
```

arsalansufi (approved):
Yay! Glad this ended up being not too bad

Also thank you for adding the 4.0.2-candidate label! Going to retroactively apply that tag and mention it in the [4.0.2 tracking ticket](https://github.com/votingworks/vxsuite/issues/5989)

Issue #6139, opened 2025-03-24
Clown car refinements


arsalansufi:
[Slack thread](https://votingworks.slack.com/archives/CEL6D3GAD/p1742831083130979?thread_ts=1742514375.879069&cid=CEL6D3GAD)

- [x] Pausing/resuming of individual functions
- [x] Last key press indicator
- [x] (nice-to-have) A safe shutdown button
- [x] Simple viewer for scanned images

arsalansufi:
Variety of changes merged, latest being https://github.com/votingworks/vxsuite/pull/6209

We'll want to make new images at some point, but can otherwise consider this ticket completed

PR #6143, opened 2025-03-24
feat(scan): allow pausing clown car app components

eventualbuddha:
## Overview

Refs https://github.com/votingworks/vxsuite/issues/6139

Rather than having a single "stop test" button, we now have pause buttons for each of the components that we're testing. Additionally, I added a "Power Off" button to make safe shutdowns easier.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/1d65e6da-cf43-4327-be8f-a545dad93a5a

### Last Key Press

CleanShot 2025-03-24 at 16 21 28@2x

## Testing Plan

Mostly tested manually, with a few additional automated tests thrown in.


apps/scan/backend/src/electrical_testing/simple_scanner_client.ts
```diff
@@ -30,7 +30,7 @@ export function createSimpleScannerClient(): SimpleScannerClient {
 
     async disconnect() {
       assert(client, 'Scanner client is not connected');
-      (await client.disconnect()).unsafeUnwrap();
+      (await client.exit()).unsafeUnwrap();
```
arsalansufi, line 33: Ooh good find! Out of curiosity, do we face a similar issue anywhere in the real state machine? Using disconnect and not cleaning up pdictl processes?

apps/scan/backend/src/electrical_testing/task_controller.test.ts
```diff
@@ -0,0 +1,25 @@
+import { describe, expect, test } from 'vitest';
+import { TaskController } from './task_controller';
+
+describe('TaskController', () => {
+  test('is initially not running', () => {
+    const loop = new TaskController();
```
arsalansufi, line 6: nit: Let's switch from `loop` to `task` here like we did in the application code for consistency

apps/scan/frontend/src/electrical_testing/electrical_testing_screen.tsx
```diff
@@ -0,0 +1,353 @@
...
+            <TaskStatusButton
+              taskStatus={taskStatus}
+              onPress={onToggleRunning}
+            />
+          </Column>
+        )}
+      </Row>
+    </Card>
+  );
+}
```
arsalansufi, line 167: Not necessary right now but just noting that the VxMark electrical testing app may also get the same updates at which point we can move some of these components to a shared location

apps/scan/backend/src/electrical_testing/loops/card_read_and_usb_drive_write_loop.ts
```diff
@@ -0,0 +1,75 @@
...
+  while (true) {
+    // Exit the loop if both tasks are stopped.
+    if (cardTask.isStopped() && usbDriveTask.isStopped()) {
+      break;
+    }
+
+    // Wait for at least one task to be running.
+    await Promise.race([cardTask.running(), usbDriveTask.running()]);
+
+    if (cardTask.isRunning()) {
```
arsalansufi, line 42: Took me a bit of time to grok the difference between `.running()` and `.isRunning()`. Looking at the implementation of `TaskController` cleared my confusion. Maybe some name changes might make the external interface easier to understand without looking at the implementation, e.g.,

```ts
cardTask.isRunning()

await cardTask.waitUntilIsRunning()
```

arsalansufi (approved):
A few suggestions but otherwise LGTM!

apps/scan/backend/src/electrical_testing/simple_scanner_client.ts
eventualbuddha: I'm not sure. We don't appear to be calling `exit` in the PDI state machine. Perhaps @jonahkagan knows whether the reconnect logic makes it so the underlying `pdictl` process gets cleaned up properly?

jonahkagan: IIRC, it turned out that for all of the disconnections we were handling, it was possible to reconnect by simply calling `client.connect()` again instead of exiting and creating a new client. So the `pdictl` process stays alive. Any situation where the `pdictl` process crashes is treated as an unexpected failure, not a disconnection.

But it's probably worth having someone else take a look to confirm. I just checked but a second pair of eyes wouldn't hurt

jonahkagan: This comment explains some of how the client works in this regard: https://github.com/votingworks/vxsuite/blob/main/libs/pdi-scanner/src/ts/scanner_client.ts#L215-L218

(Though the comment should say 'exit' error code, not 'disconnected' error code)

PR #6145, opened 2025-03-25
fix(scan): print less during stress test

eventualbuddha:
## Overview

We want to stop printing a whole 8.5x11" sheet's worth of paper every time we print during the stress test. Instead, we print a single line of text. This required some changes to the underlying `fujitsu-thermal-printer` library since it only really exposed printing a PDF document, not arbitrary images. This adds the ability to print images and then uses that in the stress test app to print a single "Testing printer" line every time printing happens.

## Demo Video or Screenshot

image

## Testing Plan

Ran manually for a while, seemed to work as expected.


arsalansufi (approved):
LGTM!

PR #6117, opened 2025-03-20
Jonah/relational election tables

jonahkagan:
## Overview

Task: #6096 

Migrates VxDesign's database schema to have separate tables for election entities (districts, precincts, precinct splits, parties, contests, and candidates). These tables model the form data used in VxDesign's forms, rather than the eventual exported election definition (these mostly align, but differ in a few places, e.g. modeling candidate names as first/middle/last vs a single name).

Currently this draft PR has just the migration and a bit of in-progress store query conversion. I'd like to get early feedback on the schema before I update all the queries.

## Demo Video or Screenshot
N/A

## Testing Plan
TODO

## Checklist

- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate to any new user actions, system updates such as file reads or storage writes, or errors introduced.

- [ ] I have added a screenshot and/or video to this PR to demo the change
- [ ] I have added the "user_facing_change" label to this PR to automate an announcement in #machine-product-updates

PR #6148, opened 2025-03-25
VxDesign: Regenerate IDs when loading/cloning elections

jonahkagan:
## Overview

To ensure that IDs are unique across elections, when loading/cloning elections, generate a new ID for each district, precinct, precinct split, party, contest, and candidate. This will allow us to use the IDs as primary keys once we switch to having separate db tables for each entity.

## Demo Video or Screenshot
N/A

## Testing Plan
- Updated existing tests
- The `regenerateIds` function came from #6117, where it was tested against database primary key indexes, ensuring that all of the necessary IDs are regenerated

## Checklist


apps/design/backend/src/app.ts
```diff
@@ -146,6 +148,90 @@ export function convertVxfPrecincts(election: Election): SplittablePrecinct[] {
...
+): {
+  districts: District[];
+  precincts: SplittablePrecinct[];
+  parties: Party[];
+  contests: AnyContest[];
+} {
+  const idMap = new Map<string, string>();
+  function replaceId<T extends string>(id: T): T {
+    if (!idMap.has(id)) {
+      idMap.set(id, generateId());
```
kofi-q, line 163: Looks like a fair amount of synchronous crypto happening here - is this is safe for the main thread? Wondering if we should use an async version of the ID generator (probably fine given the number of elements and the size of the IDs, but just double-checking)

kofi-q (approved):
Nice! LGTM - minor, non-blocking comments

apps/design/backend/src/app.ts
jonahkagan: Good q, I didn't think about that!

It looks like it's pretty fast according to the [benchmarks](https://github.com/ai/nanoid?tab=readme-ov-file#benchmark) (and we're also using a shorter ID length than the default).

There is also a [non-secure option](https://github.com/ai/nanoid?tab=readme-ov-file#non-secure) that is faster that we could consider. Or making it async would be fine too.

I'm also thinking about how rare it is to clone/load an election, so that also makes me less worried.

PR #6149, opened 2025-03-25
VxDesign: Migrate duplicate IDs in election content

jonahkagan:
## Overview

Detects elections with duplicate IDs (for districts/precincts/parties/contests) and regenerates their IDs. This migration can probably be removed once it's been run, since we now have a change in place to prevent duplicate IDs from being saved in the first place.

## Demo Video or Screenshot
N/A

## Testing Plan
Ran on a db snapshot and then ran the migration from #6117 to confirm that there are no more unique index errors for any of the election content.

## Checklist

### DIFF
--- a/apps/design/backend/migrations/1742941957696_regenerate-duplicate-ids.js
+++ b/apps/design/backend/migrations/1742941957696_regenerate-duplicate-ids.js
@@ -1,0 +1,95 @@
+/* eslint-disable vx/gts-object-literal-types */
+/* eslint-disable @typescript-eslint/no-var-requires */
+const { hasSplits } = require('@votingworks/types');
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+const { regenerateElectionIds } = require('../build/utils.js');
+
+/**
+ * @type {import('node-pg-migrate').ColumnDefinitions | undefined}
+ */
+exports.shorthands = undefined;
+
+/**
+ * @param pgm {import('node-pg-migrate').MigrationBuilder}
+ * @param run {() => void | undefined}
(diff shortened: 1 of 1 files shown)


apps/design/backend/migrations/1742941957696_regenerate-duplicate-ids.js
```diff
@@ -0,0 +1,95 @@
...
+      `
+      UPDATE elections
+      SET election_data = '${JSON.stringify(updatedElection).replaceAll(
+        "'",
+        "''"
+      )}',
+          precinct_data = '${JSON.stringify(updatedPrecincts).replaceAll(
+            "'",
+            "''"
+          )}'
```
jonahkagan, line 90: There's no quoting support in `pgm.sql`, but you have to use `pgm.sql` to include a database command in the migration's transaction. Super annoying, but seems like manually escaping the single quote solves it.

kofi-q (commented):
Interesting point about the migration script being more of a temporary one-off - makes me wonder if we want to use a different pattern for one-off, purely data migration tasks like this - i.e. maybe we write a separate script using the in-app DB client instead of a `node-pg-migrate` script. The latter requires the script to stay put indefinitely - deleting it after running would break the record keeping (although there are ways to manually fix that)

kofi-q (approved):
Actually, nvm - changed my mind soon as I hit send 😅 
I think this is fine to leave in place indefinitely - it'll get skipped in prod/staging and existing dev DBs after it's been run anyway and it'll be a no-op for fresh DBs

jonahkagan:
Ah right I forgot about the record keeping of the migrations. I think your reasoning makes sense about why it's fine to keep this migration around.

Issue #6167, opened 2025-03-26
VxPollbook: Fix inverseWarning Icon color


carolinemodic:
We hardcoded a color for the inverseWarning icon color treatment in VxPollbook. We should make this part of the regular theming.

PR #6169, opened 2025-03-26
Vxpollbook integration branch with fixes reference dup

carolinemodic:
PR Reference that shows the changes between the base integration of vxpollbook commit history: https://github.com/votingworks/vxsuite/pull/6168

And my PR for actual merging to the repo: https://github.com/votingworks/vxsuite/pull/6170

This PR is useful to review by commit to see all of the changes I made after applying the git commit. history in order to clean up pollbook/vxsuite code as appropriate. See the notes/comments left in the primary PR for more context.


libs/ui/src/icons.tsx
```diff
@@ -135,7 +135,7 @@ function iconColor(theme: UiTheme, color?: IconColor) {
     danger: colors.dangerAccent,
     inverse: colors.onInverse,
     inversePrimary: colors.inversePrimary,
-    inverseWarning: 'rgb(255 163 84)',
+    inverseWarning: 'rgb(255 163 84)', // TODO-POLLBOOK-MERGE
```
jonahkagan: I'm happy to take on adding this color to the design system (as we'll need to make sure it meets contrast requirements). Can you assign a follow up to me?

libs/printing/src/render.tsx
```diff
@@ -216,7 +216,7 @@ export async function renderToPdf(
     const headerHtml =
       headerTemplate &&
       (() => {
-        const { elementHtml, styleElement } =
```
jonahkagan, line 219: This looks like potentially a regression. I think we should use the `styleElement` from the rendering of the `headerTemplate` rather than using the variable of the same name from the outer scope.

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -680,7 +682,7 @@ export class DippedSmartCardAuth implements DippedSmartCardAuthApi {
       return err('wrong_jurisdiction');
     }
 
-    if (!(user.role === 'election_manager' || user.role === 'poll_worker')) {
+    if (this.config.allowedUserRoles.indexOf(user.role) === -1) {
```
jonahkagan, line 684: could use `this.config.allowedUserRoles.includes(user.role)`

libs/ui/src/search_select.tsx
```diff
@@ -82,6 +82,7 @@ export interface SearchSelectMultiProps<T extends string = string>
   value: T[];
   onChange: (values: T[]) => void;
   disabled?: boolean;
+  menuPortalTarget?: HTMLElement;
```
jonahkagan, line 83: We might want to consider changing this everywhere, but I think this is the right decision for now. Maybe we track as a follow up

apps/pollbook/backend/src/index.ts
```diff
@@ -65,12 +65,6 @@ function main(): Promise<number> {
   return Promise.resolve(0);
 }
 
-process.on('unhandledRejection', (reason, promise) => {
-  console.error('Unhandled rejection:', reason);
-  console.error(JSON.stringify(reason));
-  console.error(JSON.stringify(promise));
-});
```
jonahkagan, line 71: Do we not want this handler anymore?

libs/printing/src/render.tsx
carolinemodic: yeah I found this bug while testing something else just now too haha. I have a seperate branch with a fix and some added tests to pollbook to catch this in the future automatically. I'll pull the primary fix in and then keep the added tests and other changes on another branch to send out post-merge

libs/ui/src/icons.tsx
carolinemodic: https://github.com/votingworks/vxsuite/issues/6167 ty!

libs/ui/src/search_select.tsx
carolinemodic: yeah I didn't totally understand this change to be honest. It seemed to inexplictly make the font weight lighter on the selections in some vxsuite search select use cases

carolinemodic: https://github.com/votingworks/vxsuite/issues/6177 didn't assign anyone yet but feel free to assign yourself if you want to look into it

PR #6170, opened 2025-03-26
Integrate VxPollbook Into VxSuite

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/6153
Integrates the newly created apps for VxPollbook along with changes to libs/ that were made over the course of the pollbooks project to vxsuite. This is the PR that I intend to actually merge into main and is based off of main, however the PRs below may be useful references in reviewing/iterating on updates: 

Reference that is just the result of applying git filter-repo and bringing the commit history into vxsuite.
https://github.com/votingworks/vxsuite/pull/6168

Reference PR to see just the changes AFTER the branch above where I changed things / fixed issues, recommend at least skimming by commit:  https://github.com/votingworks/vxsuite/pull/6169

A few specific changes that I did in this PR but would be happy to split out into a seperate PR if useful:
- VxPollbook requires EM/PW auth for DippedSmartCardAuth. I went ahead and incorporated those changes, updating the existing DippedSmartCard auth to take a config of allowed user types rather then assuming the list of vendor, sys_admin, em. If we are ok temporarily breaking auth in VxPollbook this could be reverted to vxsuite main and done in a followup PR. 
- We updated the versions of some of the font awesome icons in VxPollbook I believe in order to resolve a type issue in the usb icon added. I just updated everything to 6.7.2 in vxsuite which did not seem to cause changes/problems but if we think that is safer to revert we could just use a different placeholder icon in VxPollbook and file a followup ticket. 
- We made a change in printing to only show a printer as disconnected after a certain period has passed right after printing occurs. I think this is safe to be true for everyone, but we could special case to poll books, or break for now and file a followup to reimplement. 

Other Notes
- RadioGroup changes did not seem to cause any noticeable impact on VxSuite from basic testing but I don’t understand the changes particularly well. 
- SearchSelect changes did have a noticeable impact on styles in vxsuite so I special cased them to poll book only 
- I'm not entirely sure why we changed adding asserts around time zone in various places, but it all seems safe
- I think some of the changes we made to loosen election parsing in the dev dock to support the smaller pollbook elections are safe/fine to merge since they only apply to mocks / dev dock. Filed https://github.com/votingworks/vxsuite/issues/6173 to test the pollbook use case of this more and fix as needed post-merge.
- Checkbox Icon is switched to the vxsuite original default of returning a filled checkmark. Followup ticket to fix pollbook uses: https://github.com/votingworks/vxsuite/issues/6174 

In addition to skimming #6169 by commit, I think browsing the file changes on this PR, particularly for the libs/ changes (or anything else not in apps/pollbook) would be useful to make sure there is not something unintentionally modified. 

## Demo Video or Screenshot
N/A 

## Testing Plan
Sanity testing of vxsuite apps and pollbook apps for basic functionality but largely relying on our test infrastructure for vxsuite impact. 

## Checklist


jonahkagan (approved):
I skimmed https://github.com/votingworks/vxsuite/pull/6169 and left comments there. The only blocking one is a potential regression in the printing logic. I also skimmed the final diff for libs in this PR and left a few more comments.

Overall, this looks really good. Thanks for taking on this big job!

PR #6200, opened 2025-03-27
Add test & fix bug with backup paper checklist

carolinemodic:
## Overview
Adds a basic test for the backup paper checklist in VxPollbook. 
Discovers and fixes a bug where we were only including the new registration sheet in the backup for a given letter if there were voters currently registered with a last name starting with that letter. 
Does some reorganization / setup of general test helpers

## Demo Video or Screenshot
N/A

## Testing Plan
Ran Tests


apps/pollbook/backend/src/store.ts
```diff
@@ -500,25 +500,34 @@ export class Store {
...
     const sortedVoters = sortedByVoterName(Object.values(voters), {
       useOriginalName: true,
     });
     const groupedVoters = groupBy(sortedVoters, (v) =>
       v.lastName[0].toUpperCase()
-    ).map(([, voterGroup]) => voterGroup);
+    ).reduce<Record<string, Voter[]>>((acc, [key, voterGroup]) => {
+      acc[key] = voterGroup;
+      return acc;
+    }, {});
```
jonahkagan, line 514: I think you could use `Object.fromEntries` here

apps/pollbook/backend/vitest.config.ts
```diff
@@ -7,8 +7,8 @@ export default defineConfig({
     clearMocks: true,
     coverage: {
       thresholds: {
-        lines: 0,
-        branches: 0,
+        lines: 30,
+        branches: 30,
```
jonahkagan: 🎉

jonahkagan (approved):
Nice catch!!

Issue #6006, opened 2025-02-12
VxAdmin: Single Contest Editing – Digital representation of ballot state


nikhilb4a:

adghayes:
Closed by #6073

PR #6210, opened 2025-03-28
Contest adjudication allowing for vote and write-in adjudication

nikhilb4a:
## Overview

Features:
- New write-in entry
- Existing official candidate selection via search
- Existing write-in candidate selection via search
- Write-in entry filtering on type
- Double vote prevention
- Only persists changes on Save
- Unmarked write-in support
- Captions indicating adjudications
- Ballot zoom out

[Project requirements - Single Contest Editing](https://docs.google.com/document/d/1rfK6BX375Tjiv3eaPYaYNr3VkPGIIF6TQctHIEUHnho/edit?tab=t.36eix9bht40a#heading=h.cz2wi2i7smod)

https://github.com/votingworks/vxsuite/issues/6006

Screenshot 2025-03-28 at 4 16 11 PM

To see the original commit history, view the [original draft PR](https://github.com/votingworks/vxsuite/pull/6073).

## Demo Video or Screenshot

Demo video:
Note: some copy is stale in this video

https://github.com/user-attachments/assets/d89e1941-784d-434c-aa96-c1e35ed5d9a9

## Testing Plan

Manual testing done by multiple team members
Automated test coverage for all new code

## Checklist


libs/ui/src/search_select.tsx
```diff
@@ -6,20 +6,27 @@ import Select, {
...
+  const { selectProps } = props;
   return (
     <components.DropdownIndicator {...props}>
       <Button
         fill="transparent"
         icon="CaretDown"
-        // The react-select DropdownIndicator component has its own click
-        // handler. It seems to work fine with the button inside it, so we just
-        // put a dummy handler on the button itself.
-        onPress={() => {}}
```
nikhilb4a, line 22: The dropdown button did nothing before when clicked on, not even opening/closing the menu

apps/admin/backend/src/app.ts
```diff
@@ -672,12 +713,85 @@ function buildApi({
...
+      isUnmarked: boolean;
+    }): Id {
+      return store.addWriteIn({
+        electionId: loadCurrentElectionIdOrThrow(workspace),
+        castVoteRecordId: input.cvrId,
+        contestId: input.contestId,
+        optionId: input.optionId,
+        side: input.side || 'front', // BMD ballots are always on the front.
+        isUnmarked: input.isUnmarked,
+        machineMarkedText: undefined, // VERIFY if this is correct
```
nikhilb4a, line 740: I could use advice on if this is the correct value for `machineMarkedText`

apps/admin/backend/src/store.ts
```diff
@@ -2137,6 +2288,31 @@ export class Store {
     return rows.map((r) => r.id);
   }
 
+  /**
+   * Gets the write-in IDs for a given cast vote record's contest.
+   */
+  getCvrContestWriteInIds({
```
nikhilb4a, line 2294: This might be too specific of a query, as I could make `cvrId` and `contestId` optional. I just went with this for now since it was my use case, but I am open to advice. Similar

apps/admin/backend/src/store.ts
```diff
@@ -1847,6 +1851,127 @@ export class Store {
     return rows;
   }
 
+  /**
+   * Gets write-in record adjudication queue for a specific contest.
+   */
+  getWriteInAdjudicationCvrQueue({
```
nikhilb4a, line 1857: The following three queries follow the same pattern used in the existing write-in flow, though now we adjudicate on a cvr by cvr basis, rather than individual write-ins.

Relatedly, I expect to add a follow up PR or commit that deletes all the unneeded code for the existing write-in flow, depending on if that is the expected process when deprecating a screen

eventualbuddha:
Before diving into the code I'm playing around with it a bit locally to get a sense of how the feature works. I'll be dropping comments and questions here as I encounter things.

eventualbuddha:
In this state where I go back into the Write-In dropdown it looks like the name I previously entered is still there and editable in the text box, but it's not. I tried Ctrl+A to select it, using the arrow keys, etc.


https://github.com/user-attachments/assets/22d59ff0-21cd-483a-851c-c9bdb21cf38b

eventualbuddha:
When entering the same name as one previously selected the "Add: XXX" entry is still there but appears to just be the same as selecting the name from the other entry. I think we should just remove the "Add: XXX" entry.

https://github.com/user-attachments/assets/8d2235ad-663b-4374-99b4-183c43900591

eventualbuddha:
Should the special entries "Not a mark" and "Add: XXX" be visually distinguished from the regular name entries? It takes more effort than I'd like to understand that they're conceptually different, i.e. you have to read the text.

CleanShot 2025-04-03 at 14 42 53@2x

eventualbuddha:
> Should the special entries "Not a mark" and "Add: XXX" be visually distinguished from the regular name entries? It takes more effort than I'd like to understand that they're conceptually different, i.e. you have to read the text.
> 

An obviously contrived example, but I can add a candidate called "Not a mark" and now I can't easily tell which is the candidate and which is the action:

CleanShot 2025-04-03 at 14 45 28@2x

eventualbuddha:
If we've made changes but not yet clicked "Save & Next" or "Finish", clicking any of "Back", "Skip", or "Close" discards the changes we've made without any warning.

https://github.com/user-attachments/assets/6687b88a-3dc0-4d78-b3cd-54a35acc92d5

eventualbuddha:
Does it make sense to limit how much text can be entered as a write-in name? The UI and our reporting handles it reasonably well, but I don't think there's any real need to allow more than ~200 characters.

CleanShot 2025-04-03 at 14 55 42@2x

eventualbuddha:
When using the write-in combobox toward the bottom of the sidebar the dropdown entries get cut off quite a bit and it's not always clear that there are more entries. If this were a native `<select>` it might decide to put the popup above the trigger rather than below. Or we could scroll the sidebar to bring it more center?

https://github.com/user-attachments/assets/9b122b94-afff-4003-8801-d733dd6f4ab0

eventualbuddha:
The above comments notwithstanding, this feels nice to use! I look forward to seeing even more improvement to the adjudication feature.

nikhilb4a:
Great feedback, thank you. Good to see how you test, these are some good finds. Thumbs up-ed all the ones that feel clear.

> Should the special entries "Not a mark" and "Add: XXX" be visually distinguished from the regular name entries? 
Definitely agree. Drew added this as a low priority task with the intention that we address it as polish once we have all the critical features built. Does that sound reasonable?

> If we've made changes but not yet clicked "Save & Next" or "Finish", clicking any of "Back", "Skip", or "Close" discards the changes we've made without any warning.
The thinking was that these inputs are pretty trivial to re-enter, but would you recommend I add a confirmation modal in?

eventualbuddha:
> Great feedback, thank you. Good to see how you test, these are some good finds. Thumbs up-ed all the ones that feel clear.
> 
> > Should the special entries "Not a mark" and "Add: XXX" be visually distinguished from the regular name entries?
> 
> Definitely agree. Drew added this as a low priority task with the intention that we address it as polish once we have all the critical features built. Does that sound reasonable?

Yeah. Link to the ticket?

> > If we've made changes but not yet clicked "Save & Next" or "Finish", clicking any of "Back", "Skip", or "Close" discards the changes we've made without any warning.
> 
> The thinking was that these inputs are pretty trivial to re-enter, but would you recommend I add a confirmation modal in?

I agree that they're pretty easy to re-enter. The problem for me is it not being clear that my data has been discarded. Given that, I personally would prefer a modal confirming my intention and letting me know that the text I entered will be discarded and what state it will be in after taking the action. However, I defer to @adghayes on making the call.

> > When using the write-in combobox toward the bottom of the sidebar the dropdown entries get cut off quite a bit and it's not always clear that there are more entries. If this were a native  it might decide to put the popup above the trigger rather than below. Or we could scroll the sidebar to bring it more center?
> > 
> > Good point. I previously had this working better and it's regressed a bit. It was opening upwards more aggressively. I'll look into that again, though I wonder if scrolling would actually be preferred 🤔

One other way to do it is to make the scrollbar in the dropdown permanently visible.

adghayes:
I love the testing! @nikhilb4a I'll skip the comments you gave a thumbs up to as signal you have enough direction.

- I created a ticket for the "Not a mark" visual distinction: https://github.com/votingworks/vxsuite/issues/6242, in addition to tickets for the other low priority issues [on the board](https://github.com/orgs/votingworks/projects/23/views/103)
- For the dropdown, it seemed to be working well when I tested but either that was before the regression or I didn't catch it. I'm unsure how exactly that scrolling would work (just more aggressive auto-scrolling initially, or auto-scrolling every time you selected a write-in to center it?) and I'm thinking it might feel weird. I would be happy with it just opening upward.
- I agree with @eventualbuddha's take on the confirmation modal, that it's sketchy that we're discarding changes in cases where the user might easily not be aware. I've created it as a [separate ticket](https://github.com/orgs/votingworks/projects/23/views/103?pane=issue&itemId=105390493&issue=votingworks%7Cvxsuite%7C6246) to track adding a confirmation modal to handle as a follow-up.

apps/admin/frontend/src/components/adjudication_ballot_image_viewer.tsx
```diff
@@ -96,51 +96,44 @@ const BallotImageViewerControls = styled.div<{ isZoomedIn: boolean }>`
...
-          disabled={isZoomedIn}
+          icon={isZoomedIn ? 'ZoomOut' : 'ZoomIn'}
+          onPress={() => setIsZoomedIn(!isZoomedIn)}
           color="neutral"
           fill="tinted"
         >
-          Zoom In
+          Zoom {isZoomedIn ? 'Out' : 'In'}
         </Button>
       </BallotImageViewerControls>
```
eventualbuddha, line 125: Should we just make this be the same height as the header in the sidebar? It's close enough to make it feel misaligned.

Screenshot-VxAdmin-2025-04-07T20:58:01 032Z

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1124 @@
...
+          }
+          default: {
+            /* istanbul ignore next - @preserve */
+            throwIllegalValue(writeIn, 'adjudicationType');
+          }
+        }
+      }
+
+      setIsStateStale(false);
+      setFocusedOptionId('');
```
eventualbuddha, line 723: Consider using `undefined` to mean "no value".

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1124 @@
...
+      <Main flexRow>
+        <BallotPanel>
+          {isFocusedWriteInHmpb && (
+            <BallotZoomImageViewer
+              ballotBounds={firstWriteInImage.ballotCoordinates}
+              key={currentCvrId} // Reset zoom state for each write-in
+              imageUrl={firstWriteInImage.imageUrl}
+              zoomedInBounds={
+                focusedWriteInImage &&
+                'writeInCoordinates' in focusedWriteInImage
```
eventualbuddha, line 840: Consider adding a `type` property to the `WriteInImageView` types so we don't have to property snoop here.

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1124 @@
...
+                rightIcon="Next"
+                style={{ width: '5.5rem' }}
+              >
+                Skip
+              </Button>
+              <Button
+                disabled={!allWriteInsAdjudicated}
+                icon="Done"
+                onPress={saveAndNext}
+                style={{ flexGrow: '1' }}
```
eventualbuddha, line 1107: We've got a bunch of `styled` components above. Add these too, e.g. `SecondaryNavButton` & `PrimaryNavButton` or `BackSkipButton` & `SaveButton`?

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,133 @@
...
+  toggleVote: () => void;
+  value: string;
+  officialCandidateNames: string[];
+  writeInCandidateNames: string[];
+}): JSX.Element {
+  const theme = useTheme();
+  const [inputValue, setInputValue] = useState('');
+  const inputValueLowerCase = inputValue.toLowerCase();
+
+  function onKeyPress(val?: string) {
```
eventualbuddha, line 48: Rename to `onInputChange`?

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
eventualbuddha, line 1: This button/combobox doesn't seem to handle trailing spaces in names well. If you enter a candidate and then enter it again in a subsequent CVR but add a space accidentally, it appears to work but just silently discards that adjudication.


https://github.com/user-attachments/assets/d0e797cf-1995-415f-8ec1-b944f9251dd8

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,133 @@
...
+    .map((name) => ({
+      label: name,
+      value: name,
+    }));
+
+  // 'Add: NEW_CANDIDATE' entry if there is no exact match
+  if (
+    inputValue &&
+    !options.find(
+      (item) => item.label.toLowerCase() === inputValue.toLowerCase()
```
eventualbuddha, line 73:
```suggestion
      (item) => item.label.toLowerCase() === inputValueLowerCase
```

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,133 @@
...
+    )
+  ) {
+    options.push({ label: `Add: ${inputValue}`, value: inputValue });
+  }
+
+  // If value has been entered and it is a new entry, add it the dropdown
+  if (value && !options.find((option) => option.label === value)) {
```
eventualbuddha, line 80: Use `Array::some` to make it clear we're checking for a value matching a predicate, not finding a value and using its truthiness.

```suggestion
  if (value && !options.some((option) => option.label === value)) {
```

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,133 @@
...
+    options.unshift({ label: 'Not a mark', value: 'invalid' });
+  }
+
+  return (
+    <Container style={{ zIndex: isFocused ? 10 : 0 }}>
+      <CheckboxButton
+        isChecked={isSelected}
+        label="Write-in"
+        onChange={toggleVote}
+        style={{ borderRadius: '0.5rem 0.5rem 0 0' }}
```
eventualbuddha, line 94: Define `styled(CheckboxButton)` with this style above?

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,133 @@
...
+                style={{ marginRight: '0.5rem' }}
+              />
+              {isSelected ? 'Adjudicate' : 'Unmarked'} Write-in
+            </React.Fragment>
+          )
+        }
+        style={{
+          backgroundColor: value ? undefined : theme.colors.warningContainer,
+          borderRadius: '0 0 0.5rem 0.5rem',
+        }}
```
eventualbuddha, line 128: Ditto `styled(SearchSelect)`.

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,133 @@
...
+        placeholder={
+          isFocused ? (
+            'Search or add...'
+          ) : (
+            <React.Fragment>
+              <Icons.Warning
+                color="warning"
+                style={{ marginRight: '0.5rem' }}
+              />
+              {isSelected ? 'Adjudicate' : 'Unmarked'} Write-in
```
eventualbuddha, line 121: Consider using the full strings, which makes it easier to translate (if we ever do) and to search the codebase for (even if we don't translate).
```suggestion
              {isSelected ? 'Adjudicate Write-in' : 'Unmarked Write-in'}
```

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,133 @@
...
+  value,
+  officialCandidateNames,
+  writeInCandidateNames,
+}: {
+  caption?: React.ReactNode;
+  isFocused: boolean;
+  isSelected: boolean;
+  hasInvalidEntry: boolean;
+  // newVal can be name of new or existing candidate,
+  // keyword 'invalid' for invalid write-in, or undefined
```
eventualbuddha, line 35: Consider using a different type to encode the fact that "Not a Mark" was chosen. Using strings with special meanings like this is a [well-known anti-pattern](https://www.theverge.com/tldr/2019/8/14/20805543/null-license-plate-california-parking-tickets-violations-void-programming-bug).

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,133 @@
...
+        onFocus={onInputFocus}
+        onInputChange={onKeyPress}
+        onChange={(val) => {
+          setInputValue('');
+          onChange(val);
+        }}
+        value={value}
+        placeholder={
+          isFocused ? (
+            'Search or add...'
```
eventualbuddha, line 114: Nit: use the ellipsis character `…` (option-; on macOS).

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1124 @@
...
+              icon="X"
+              to={routerPaths.writeIns}
+              variant="inverseNeutral"
+            >
+              Close
+            </LinkButton>
+          </BallotHeader>
+          <BallotVoteCount>
+            <MediumText>
+              Votes cast: {voteCount} of {seatCount}
```
eventualbuddha, line 871: Along with `import { format } from '@votingworks/utils';` above, consider using our count formatting to set a good example even though it won't ever matter here.

```suggestion
              Votes cast: {format.count(voteCount)} of {format.count(seatCount)}
```

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1124 @@
...
+                    })}
+                  />
+                );
+              })}
+            </CandidateButtonList>
+          )}
+          <BallotFooter>
+            <BallotMetadata>
+              <SmallText>
+                {scrollIndex + 1} of {numBallots}
```
eventualbuddha, line 1076: Ditto on count formatting, and here it might matter.
```suggestion
                {format.count(scrollIndex + 1)} of {format.count(numBallots ?? 0)}
```

libs/ui/src/checkbox_button.tsx
```diff
@@ -6,6 +6,7 @@ export interface CheckboxButtonProps {
   isChecked: boolean;
   onChange: (isChecked: boolean) => void;
   disabled?: boolean;
+  style?: React.CSSProperties;
```
eventualbuddha, line 9: Consider reverting this if you do the `styled(CheckboxButton)` as I suggested earlier.

libs/ui/src/search_select.tsx
```diff
@@ -69,9 +76,12 @@ interface SearchSelectBaseProps<T extends string = string> {
   options: Array<SelectOption<T>>;
   'aria-label'?: string;
   style?: React.CSSProperties;
-  placeholder?: string;
+  placeholder?: React.ReactNode | string;
```
eventualbuddha, line 79: `React.ReactNode` is a union that includes `string`.
```suggestion
  placeholder?: React.ReactNode;
```

eventualbuddha (commented):
First pass of code review complete. The main issues to discuss:
- triggering multiple mutation from a single action and the lack of atomicity that causes
- not taking more advantage of the type system to limit the use of `assert`s and sentinel values like `''` and `'invalid'`.
The rest is mostly smaller changes for you to consider making.

apps/admin/backend/src/app.ts
adghayes: We do have a flow for BMD ballots through the adjudication flow that @carolinemodic built last year, and we'll have to confirm that it still works after these changes.

nikhilb4a: @adghayes is it possible for there to be a valid UWI with a BMD? This backend logic should only be relevant for UWIs, though we would need to gate it off on the frontend.

Will check on this in our sync, but it seems like UWI wouldn't make sense for BMDs

apps/admin/frontend/src/api.ts
```diff
@@ -357,6 +431,57 @@ export const getWriteInAdjudicationContext = {
...
+    );
+  },
+} as const;
+
+type GetVoteAdjudicationsInput = QueryInput<'getVoteAdjudications'>;
+export const getVoteAdjudications = {
+  queryKey(input?: GetVoteAdjudicationsInput): QueryKey {
+    return input ? ['getVoteAdjudications', input] : ['getVoteAdjudications'];
+  },
+  useQuery(input?: GetVoteAdjudicationsInput, enabled = true) {
```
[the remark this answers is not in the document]
nikhilb4a: Good point, definitely agree. Will update the functions I added

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
nikhilb4a: I'll update the logic to use `normalizedWriteInName` when comparing the input values instead of just lower casing, as [that will incorporate leading/trailing spaces](https://github.com/votingworks/vxsuite/blob/main/apps/admin/frontend/src/utils/write_ins.ts#L1)

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1124 @@
...
+  margin: 0.25rem 0 0.25rem 0.125rem;
+`;
+
+const ContestTitleDiv = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 0;
+`;
+
+const StyledH1 = styled(H1)`
```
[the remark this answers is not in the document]
nikhilb4a: The two properties are the only difference, smaller font-size and margin removed. Are you suggesting I name it more clearly?

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1124 @@
...
+  originalVote,
+  currentVote,
+  isWriteIn,
+  writeInState,
+  writeInRecord,
+}: {
+  originalVote: boolean;
+  currentVote: boolean;
+  isWriteIn: boolean;
+  writeInState?: WriteInOptionState;
```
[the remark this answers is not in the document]
nikhilb4a: Only in this spot will I leave `writeInState` as optional, as it allows the input object to omit the property for 
non write-in captions

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -367,9 +374,9 @@ export function ContestAdjudicationScreen(): JSX.Element {
   const voteCount = Object.values(voteState).filter(Boolean).length;
   const isOvervote = voteCount > seatCount;
 
-  const writeInOptionIds = Array.from({ length: seatCount }).map(
-    (_, idx) => `write-in-${idx}`
-  );
+  const writeInOptionIds = useMemo(() => iter(allContestOptions(contest))
```
eventualbuddha: Is the use of `useMemo` here due to an observed bug or performance issue? If it's just speculative, I'd remove it.

apps/admin/frontend/src/api.ts
```diff
@@ -398,18 +398,26 @@ export const getCvrWriteInImageViews = {
...
+  useQuery(input?: GetCvrWriteInImageViewsInput) {
     const apiClient = useApiClient();
     return useQuery(
       this.queryKey(input),
       () =>
         apiClient.getCvrContestWriteInImageViews({
-          cvrId: input.cvrId,
-          contestId: input.contestId,
+          cvrId: input?.cvrId || '',
+          contestId: input?.contestId || '',
```
eventualbuddha, line 408: nit: prefer `??` unless there's a special need for falling back from a falsy value.

eventualbuddha (commented):
Thanks for making the smaller changes. Looking good so far! I look forward to seeing the API changes next week.

libs/ui/src/search_select.tsx
```diff
@@ -82,24 +82,23 @@ interface SearchSelectBaseProps<T extends string = string> {
...
+  noOptionsMessage?: () => React.ReactNode;
+  minMenuHeight?: number;
 }
 
 export interface SearchSelectMultiProps<T extends string = string>
   extends SearchSelectBaseProps<T> {
   isMulti: true;
   value: T[];
   onChange: (values: T[]) => void;
-  disabled?: boolean;
```
nikhilb4a, line 82: This was a duplicate property as it was already on `SearchSelectBaseProps`

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -310,33 +314,27 @@ export function ContestAdjudicationScreen(): JSX.Element {
...
   const [focusedOptionId, setFocusedOptionId] = useState<string>();
   const history = useHistory();
-  const scrollStateInitialized = cvrQueueIndex !== undefined;
-  const currentCvrId = scrollStateInitialized
-    ? writeInCvrQueueQuery.data?.[cvrQueueIndex]
-    : undefined;
-  const numBallots = writeInCvrQueueQuery.data?.length;
-  const onLastBallot = cvrQueueIndex ? cvrQueueIndex + 1 === numBallots : false;
+  const isQueueReady = cvrQueueIndex !== undefined;
+  const currentCvrId = isQueueReady ? cvrQueueQuery.data?.[cvrQueueIndex] : '';
```
eventualbuddha:
```suggestion
  const currentCvrId = isQueueReady ? cvrQueueQuery.data?.[cvrQueueIndex] : undefined;
```

apps/admin/frontend/src/components/discard_changes_modal.tsx
```diff
@@ -0,0 +1,34 @@
...
+`;
+
+export function DiscardChangesModal({
+  onBack,
+  onDiscard,
+}: {
+  onBack: () => void;
+  onDiscard: () => void;
+}): JSX.Element {
+  const content = 'Your adjudications for this ballot have not been saved.';
```
eventualbuddha, line 16: Inline this?

apps/admin/frontend/src/components/discard_changes_modal.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <Modal
+      title="Unsaved Changes"
+      content={content}
+      actions={
+        <React.Fragment>
+          <EqualWidthButton variant="primary" onPress={onDiscard}>
```
eventualbuddha: Allow pressing Enter/Space to confirm the modal?
```suggestion
          <EqualWidthButton variant="primary" onPress={onDiscard} autoFocus>
```

apps/admin/frontend/src/components/discard_changes_modal.tsx
```diff
@@ -0,0 +1,34 @@
...
+            Discard Changes
+          </EqualWidthButton>
+          <EqualWidthButton variant="neutral" onPress={onBack}>
```
eventualbuddha, line 27: Trigger `onBack` with the `Esc` key?

apps/admin/backend/schema.sql
```diff
@@ -95,6 +95,7 @@ create table write_ins (
   official_candidate_id text,
   write_in_candidate_id varchar(36),
   is_invalid boolean not null default false,
+  is_manually_created boolean not null default false,
```
nikhilb4a, line 98: Also considering naming this `is_undetected`, to indicate we didn't detect the write-in, but manually created it. Does one name seem more in line with our naming conventions?

apps/admin/frontend/src/components/discard_changes_modal.tsx
nikhilb4a: I haven't been able to get the listener to work for the escape key, so I've made a [ticket](https://github.com/votingworks/vxsuite/issues/6323) to fix this.

apps/admin/backend/schema.sql
eventualbuddha: No strong preference myself.

apps/admin/backend/src/app.ts
```diff
@@ -673,21 +655,56 @@ function buildApi({
...
+          ...input,
+          electionId: loadCurrentElectionIdOrThrow(workspace),
+        }) ?? null
+      );
+    },
+
+    getWriteIns(input: {
+      cvrId?: string;
+      contestId?: string;
+    }): WriteInRecord[] {
```
eventualbuddha, line 692:
```suggestion
    getWriteIns(input: { cvrId?: Id; contestId?: ContestId }): WriteInRecord[] {
```

apps/admin/backend/src/app.ts
```diff
@@ -673,21 +655,56 @@ function buildApi({
...
     },
 
-    getWriteInAdjudicationContext(input: {
-      writeInId: string;
-    }): WriteInAdjudicationContext {
-      return getWriteInAdjudicationContext({
+    getCvrContestWriteInImageViews(input: {
+      cvrId: string;
+      contestId: string;
+    }): Promise<WriteInImageView[]> {
```
eventualbuddha, line 703:
```suggestion
    getCvrContestWriteInImageViews(input: {
      cvrId: Id;
      contestId: ContestId;
    }): Promise<WriteInImageView[]> {
```

apps/admin/frontend/src/components/adjudication_ballot_image_viewer.tsx
```diff
@@ -84,63 +84,71 @@ const ZoomedOutBallotImageContainer = styled.div`
...
+  height: 4rem;
   width: 100%;
   padding: 0.5rem;
   gap: 0.5rem;
 `;
 
+function remToPx(rem: number) {
+  const fontSizeResult = safeParseNumber(
+    // trim 'px' from the result
+    getComputedStyle(document.documentElement).fontSize.substring(0, -2)
```
eventualbuddha, line 101:
```suggestion
    getComputedStyle(document.documentElement).fontSize.substring(0, -'px'.length)
```

apps/admin/frontend/src/components/adjudication_ballot_image_viewer.tsx
```diff
@@ -84,63 +84,71 @@ const ZoomedOutBallotImageContainer = styled.div`
...
+  );
+  // fallback to default of 30
+  /* istanbul ignore next - fallback to default of 30 - @preserve */
+  const fontSize = fontSizeResult.isOk() ? fontSizeResult.ok() : 30;
```
eventualbuddha, line 105: There's a lot of duplication here.

```suggestion
  /* istanbul ignore next - @preserve */
  const fontSize = fontSizeResult.ok() ?? 30;
```

eventualbuddha (approved):

apps/admin/backend/schema.sql
nikhilb4a: Changing to `is_undetected` as that seems more idiomatic here

apps/admin/frontend/src/components/adjudication_ballot_image_viewer.tsx
nikhilb4a: I see, I had thought we needed to check `isOk` first

And thanks for the push to write code that is self-descriptive, not needing additional comments. That's a principle that slipped my mind a bit as I was away from programming.

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -0,0 +1,178 @@
...
+  onInputBlur: () => void;
+  onInputFocus: () => void;
+  officialCandidates: Candidate[];
+  writeInCandidates: Candidate[];
+}): JSX.Element {
+  const theme = useTheme();
+  const [inputValue, setInputValue] = useState('');
+  const normalizedInputValue = normalizeWriteInName(inputValue);
+  function onInputChange(val?: string) {
+    return setInputValue(val || '');
```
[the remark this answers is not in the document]
nikhilb4a: Hadn't heard that term 😅 But good point

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1009 @@
...
+        .filterMap((option) => (option.isWriteIn ? option.id : undefined))
+        .toArray(),
+    [contest]
+  );
+
+  // Vote and write-in state for adjudication management
+  const [hasVoteByOptionId, setHasVoteByOptionId] = useState<HasVoteByOptionId>(
+    {}
+  );
+  const initialHasVoteByOptionIdRef = useRef<HasVoteByOptionId | null>(null);
```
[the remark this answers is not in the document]
nikhilb4a: No we don't need `null`, removing

My understanding is that `useRef` is the right hook to use for something like storing an initial value, a variable that should persist across renders and does not need to trigger re-renders, as a state variable would

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1009 @@
...
+    }));
+  }
+  function clearBallotState() {
+    setHasVoteByOptionId({});
+    setWriteInStatusByOptionId({});
+  }
+  const isStateReady = Object.keys(hasVoteByOptionId).length > 0;
+
+  const [focusedOptionId, setFocusedOptionId] = useState<string>();
+  const [shouldAutoscrollUser, setShouldAutoscrollUser] = useState(false);
```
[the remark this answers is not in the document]
nikhilb4a: This does trigger a scroll [here](https://github.com/votingworks/vxsuite/blob/ad1b02a4574c644350440ad48b28421ccc298f7d/apps/admin/frontend/src/screens/contest_adjudication_screen.tsx#L560-L561), scrolling the candidate option list to the bottom so write-ins show

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1009 @@
...
+  ) {
+    setWriteInStatusByOptionId((prev) => ({
+      ...prev,
+      [optionId]: status,
+    }));
+  }
+  function clearBallotState() {
+    setHasVoteByOptionId({});
+    setWriteInStatusByOptionId({});
+  }
```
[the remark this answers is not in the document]
nikhilb4a: Thanks for raising this, because I do agree with you. I made the call to just use updated types and not make the class/custom hook for the sake of time, as I didn't trust I'd be able to code it and tests in a timeline I felt good about right now. But I feel like this would be a good task for me to do once marginal mark adjudication is done, knowing timelines are met. Would it be reasonable for me to create a ticket to do that later?

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -0,0 +1,1009 @@
...
+      </NavigationScreen>
+    );
+  }
+
+  const originalVotes = cvrVoteInfoQuery.data.votes[contestId];
+  const writeIns = writeInsQuery.data;
+  const writeInImages = writeInImagesQuery.data;
+  const writeInCandidates = writeInCandidatesQuery.data;
+  const safeCvrQueueIndex = cvrQueueIndex;
+  const safeCurrentCvrId = currentCvrId;
```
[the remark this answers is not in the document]
nikhilb4a: As in that they are guaranteed to be defined. Since their origin values are guaranteed to be defined now, but functions declared after still don't capture that type knowledge, this seemed like the right way to avoid using asserts in those functions. Is there a better word for this, `definedCvrQueueIndex`?

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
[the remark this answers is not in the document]
eventualbuddha: I'm okay with ticketing and moving on for now.

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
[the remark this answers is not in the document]
eventualbuddha: I might go the other way and call the original ones e.g. `maybeCvrQueueIndex`.

apps/admin/frontend/src/api.ts
```diff
@@ -316,53 +318,54 @@ export const getWriteInCandidates = {
...
-          contestId: input?.contestId,
-        }),
+      input
+        ? () =>
+            apiClient.getWriteInImageViews({
+              cvrId: input.cvrId,
+              contestId: input.contestId,
+            })
+        : /* istanbul ignore next - @preserve */
+          () => fail('input is required'),
```
nikhilb4a, line 356: I've added this for two more of the queries that use this `enabled: !!input` pattern

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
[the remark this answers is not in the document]
nikhilb4a: Ticketed [here](https://github.com/votingworks/vxsuite/issues/6343)

PR #6214, opened 2025-04-01
Revert Election type in VxPollbook to VxSuiteElection and add basic config tests

carolinemodic:
## Overview
Its no longer a product need to support importing a subset of the VxsuiteElection definition so it is easier to just use the same election file format type from VxSuite. This PR reverts back to using VxSuite's type for elections and adds some basic tests helpers for use in integration testing. It adds a basic config test to test that we can configure with a vxsuite election file, along with a basic start of an e2e test. That test is a little unrelated to the rest of the change but helps get us to a higher test coverage. I plan to follow-up by expanding app.test.ts to test more fully the major user flows, this is just a starting point. 

The final commit is revertting a change that was made in the vxpollbook merge to support the fact that pollbook did not import the full election file. 

## Demo Video or Screenshot
N/A

## Testing Plan
Ran tests. Manually configured with pollbook-package in the apps.


apps/pollbook/backend/src/app_config.test.ts
```diff
@@ -0,0 +1,139 @@
...
+async function allowRealTimeToPass() {
+  vi.useRealTimers();
+  await sleep(200);
+  vi.useFakeTimers();
+  vitest.advanceTimersByTime(200);
+}
+
+function mockElectionManagerAuth(
+  auth: DippedSmartCardAuthApi
+  // election: Election
```
kshen0, line 24: delete

apps/pollbook/backend/src/app_config.test.ts
```diff
@@ -0,0 +1,139 @@
...
+) {
+  vi.mocked(auth.getAuthStatus).mockImplementation(() =>
+    Promise.resolve({
+      status: 'logged_in',
+      user: mockElectionManagerUser({
+        // electionKey: constructElectionKey(electionDefinition.election),
```
kshen0, line 30: delete

kshen0 (approved):

Issue #6091, opened 2025-03-10
Automatically downsample seal images and ballot measure images to avoid overly large election packages


arsalansufi:
In terms of what to downsample to, we should maintain print integrity for, say, 300 DPI printing.

For seals, that probably means 300px by 300px is sufficient, as we print them at roughly 1 inch by 1 inch on ballots. For ballot measure images, we may have to do some dynamic calculation given the dimensions of the image on the rendered ballot.

PR #6215, opened 2025-04-01
[VxDesign] Resize large user images before upload

kofi-q:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6091

- Adding a normalization step to image uploads, which shrinks large images to fit within specified bounds
- Adding maximum bounds for seals, signatures, and ballot measure images, based on the largest display/print size we expect.

### TODO
- The ballot renderer renders images with their original width and height, which causes page overflow for large-enough images. We'll need to add max size restrictions there, which would allow us to set the max upload size appropriately for display in VxMark.
- This PR doesn't address potentially large SVG uploads (large image wrapped in an SVG).
- We can probably relax the 5MB restriction a bit, now that we're resizing before upload, but maybe worth addressing the large-SVG case first.
- Would be useful for users to have a way to resize images in the rich text editor - currently have to resize out side the app first.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/b53fa1a5-54d2-4339-b880-a1511220fdb5

## Testing Plan
- Manual for now (chrome, firefox); adding automated browser tests later:
  - Upload images at lower and upper bounds for seal, clerk signature, ballot measure image
  - Verified upload sizes are reduced
  - Tested long and wide images going through ballot rendering

## Checklist

### DIFF
--- a/apps/design/frontend/src/geography_screen.tsx
+++ b/apps/design/frontend/src/geography_screen.tsx
@@ -16,6 +16,7 @@ import {
   TabPanel,
   RouterTabBar,
   Modal,
+  images,
 } from '@votingworks/ui';
 import {
   Switch,
@@ -458,6 +459,13 @@ const ClerkSignatureImageInput = styled(ImageInput)`
   }
 `;
 
+const PDF_PIXELS_PER_INCH = 96;
+const SIGNATURE_IMAGE_NORMALIZE_PARAMS: Readonly<images.NormalizeParams> = {
+  maxHeightPx: 1 * PDF_PIXELS_PER_INCH,
+  minHeightPx: 50,
+  minWidthPx: 100,
+};
+
 function PrecinctForm({
   electionId,
   savedPrecinct,
@@ -698,8 +706,7 @@ function PrecinctForm({
                           }
                           buttonLabel="Upload Signature Image"
                           removeButtonLabel="Remove Signature Image"
-                          minHeightPx={50}
-                          minWidthPx={100}
+                          normalizeParams={SIGNATURE_IMAGE_NORMALIZE_PARAMS}
                         />
                       </div>
                     )}
--- a/apps/design/frontend/src/geography_screen.tsx
+++ b/apps/design/frontend/src/geography_screen.tsx
@@ -460,8 +460,14 @@ const ClerkSignatureImageInput = styled(ImageInput)`
 `;
 
 const PDF_PIXELS_PER_INCH = 96;
+const LETTER_PAGE_WIDTH_INCHES = 8.5;
+
+/** Generously padded. */
+const LETTER_PAGE_CONTENT_WIDTH_INCHES = LETTER_PAGE_WIDTH_INCHES - 2;
+
 const SIGNATURE_IMAGE_NORMALIZE_PARAMS: Readonly<images.NormalizeParams> = {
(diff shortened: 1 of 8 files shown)


libs/ui/src/images/normalize.ts
```diff
@@ -0,0 +1,161 @@
...
+  | { maxHeightPx?: number; maxWidthPx: number }
+);
+
+/**
+ * Normalizes an image, potentially resizing it to fit within the given bounds,
+ * and returns a serialized data URL version of the final image with the same
+ * encoding as the original.
+ *
+ * If minimum dimensions are specified, an error is returned for images that
+ * fall below the given thresholds (see {@link NormlaizeError}).
```
jonahkagan, line 35: Typo: NormlaizeError -> NormalizeError

libs/ui/src/images/normalize.ts
```diff
@@ -0,0 +1,161 @@
...
+          heightPx: canvas.height,
+          widthPx: canvas.width,
+        })
+      );
+    } catch (error) {
+      resolve(err({ code: 'unexpected', error }));
+    }
+  }
+
+  img.addEventListener('load', onLoad, { once: true });
```
jonahkagan, line 111: Does this image and event listener get garbage collected? Or do we need to manually dispose?

apps/design/frontend/src/geography_screen.tsx
```diff
@@ -458,6 +459,13 @@ const ClerkSignatureImageInput = styled(ImageInput)`
   }
 `;
 
+const PDF_PIXELS_PER_INCH = 96;
+const SIGNATURE_IMAGE_NORMALIZE_PARAMS: Readonly<images.NormalizeParams> = {
+  maxHeightPx: 1 * PDF_PIXELS_PER_INCH,
```
jonahkagan, line 464: Why no max width here?

apps/design/frontend/src/rich_text_editor.tsx
```diff
@@ -89,6 +95,29 @@ function ControlButton({
...
+ *
+ * [TODO] The max width should really be the VxMark screen size (1080px), since
+ * that's the largest render surface for these images, but we need to add image
+ * scaling to the ballot renderer first, to limit them to the bounds of the
+ * contest box.
+ */
+const NORMALIZE_PARAMS: Readonly<images.NormalizeParams> = {
+  maxHeightPx: (LETTER_PAGE_HEIGHT_INCHES - 3) * PDF_PIXELS_PER_INCH,
+  maxWidthPx: (LETTER_PAGE_WIDTH_INCHES - 2) * PDF_PIXELS_PER_INCH,
+  minWidthPx: 1 * PDF_PIXELS_PER_INCH,
```
jonahkagan, line 118: Why min width but no min height here?

jonahkagan (approved):
Looks great. I like the abstractions for normalizing images.

apps/design/frontend/src/rich_text_editor.tsx
kofi-q: Didn't think either one is really needed here, so still on the fence about whether a min width is even needed (since the images are rendered in the rich text editor as 1:1 size anyway). Will remove for now, but LMK if you have a preference for whether we set minimums or not

jonahkagan: I think it's probably not necessary, especially since ballots will be proofed

PR #6223, opened 2025-04-01
VxDesign: Disregard district order when generating ballot styles

jonahkagan:
## Overview

When generating ballot styles for precincts, we combine all precincts/splits that share the same districts into a single shared ballot style. However, we were previously looking at an ordered list of the districts for each precinct/split (rather than treating them as an unordered set). That means if two precincts had the same district list in a different order, they wouldn't end up with the same ballot style. This would create more ballot styles than is necessary.
## Demo Video or Screenshot
N/A

## Testing Plan
Added regression test
## Checklist


kofi-q (approved):
Oh, good catch!

PR #6227, opened 2025-04-02
Grout middleware

jonahkagan:
## Overview

Adds middleware to grout with the intention of using it for authentication and logging.

(Currently just a proof of concept - see test case for example usage)

## Checklist

### DIFF
--- a/libs/grout/src/server.ts
+++ b/libs/grout/src/server.ts
@@ -48,7 +48,10 @@ export interface AnyMethods {
 /**
  * Type for a specific API definition returned by createApi.
  */
-export interface Api<Methods extends AnyMethods, Context extends AnyContext> {
+export interface Api<
+  Methods extends AnyMethods,
+  Context extends AnyContext = AnyContext,
+> {
   methods: Methods;
   middlewares?: Array<Middleware<Context>>;
 }
--- a/libs/grout/src/server.ts
+++ b/libs/grout/src/server.ts
@@ -74,6 +74,7 @@ export type inferApiMethods<SomeApi extends AnyApi> = SomeApi extends Api<
 export interface MiddlewareMethodCall<Context extends AnyContext> {
   methodName: string;
   input: unknown;
+  request: Express.Request;
   /**
    * When a given middleware is called, the Context fields may not have been
    * filled out by other middleware yet, so we can't get any guarantees from the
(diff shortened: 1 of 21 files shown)


libs/grout/src/server.ts
```diff
@@ -158,7 +200,25 @@ export function buildRouter(
           );
         }
 
-        const result = await method(input);
+        let context: AnyContext = {};
+        for (const middleware of api.middlewares ?? []) {
+          const result = await middleware({
+            methodName,
```
kofi-q, line 206: IIUC, idea here is that middleware functions will condition logic based on method name?

Part of me wishes we could somehow co-locate API handlers with their attributes (e.g. `noAuth` in an app that requires authenticated endpoints by default), instead of managing those separately in the middleware...but, I do like how non-intrusive this approach is - the other might require a larger change.

kofi-q: (although, maybe an API method could be a regular function, or an object with a handler function and attributes...assuming the potential typings rabbit-hole isn't ridiculous 😂)

libs/grout/src/server.ts
```diff
@@ -158,7 +200,25 @@ export function buildRouter(
           );
         }
 
-        const result = await method(input);
+        let context: AnyContext = {};
+        for (const middleware of api.middlewares ?? []) {
+          const result = await middleware({
```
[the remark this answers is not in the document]
jonahkagan: Oh right, that will definitely be necessary haha

libs/grout/src/server.ts
jonahkagan: I was thinking that middleware functions would be the same for all handlers in an API. If you want differentiated logic, put that in the method itself. `methodName` is just supplied to middleware for use cases like logging.

So for the auth case, either:
- Make two APIs, one with auth middleware and one without auth middleware
- Make the `user` field on the context nullable, and have the auth middleware only populate the user object if authenticated

In either case, I imagine doing authorization with a helper function called from the auth method (e.g. `requireUserOfType({ user: context.user, userType: <some user type> })` or `requireElectionAccess({ user: context.user, electionId })`)

I'm also open to considering more structured approaches for controlling the middleware with attributes (or having middleware be method-specific instead of applying to the whole API), but this seemed like the simplest place to start for our known use cases.

kofi-q: Ah, I see what you're saying - yeah that second option ,with context optionally filled out, sounds good; addresses the concern of keeping the authz logic close to the endpoint logic

kofi-q (approved):
✅

PR #6229, opened 2025-04-02
Test for event receipt states

carolinemodic:
## Overview
Adds backend tests for all possible events a user can test that modifys the voter verifying the end state of the voter object and the printed receipts. Tested modifications in various orders/combinations to test different possible receipt states. 

## Testing Plan
Ran Tests


apps/pollbook/backend/test/id_helpers.ts
```diff
@@ -0,0 +1,13 @@
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
kshen0, line 1: nit: does adding a return type make it so we don't need this exception?

```
{
  next: () => string;
  reset: () => void;
}
```

kshen0 (approved):

carolinemodic:
oops, didn't realize I forgot to merge this

PR #6232, opened 2025-04-03
design/backend: Log API calls

### LINKED ISSUE #6140: VxDesign: Log API method arguments
It’s currently tough to trace behavior in the API logs because they don’t include the request body (where grout puts all of the request parameters), so I can see what methods were called but not which resources were impacted.

jonahkagan:
## Overview

Closes #6140 

Adds logging middleware to the design/backend API to log method arguments. This will allow us to cross reference API logs with alerts to better investigate.

Note that these logs are logged when the request is received before executing the API method handler. I went back and forth over whether it should happen before or after, but we already have logs from Heroku that are recorded after the request with status code and response time, so the before logs should fill in the information gap (namely, the input parameters) that we need for debugging. It might be a bit difficult to have to look at both log lines together to make sense of things, but the advantage of logging before the handler is that the request will be logged even if the app crashes. I figured that would be nice to have while we're working out error handling kinks, and we can easily improve on it later as needed.

## Demo Video or Screenshot
N/A

## Testing Plan
- Manually tested with debug flag
- Added automated test
## Checklist

### DIFF
--- a/apps/design/backend/src/app.test.ts
+++ b/apps/design/backend/src/app.test.ts
@@ -82,7 +82,6 @@ import {
 import { FULL_TEST_DECK_TALLY_REPORT_FILE_NAME } from './test_decks';
 import {
   BallotOrderInfo,
-  BallotStyle,
   ElectionInfo,
   ElectionListing,
   ElectionStatus,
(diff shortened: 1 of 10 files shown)


kofi-q (approved):
🎉

PR #6248, opened 2025-04-07
fix(utils): add overloads for contest types

eventualbuddha:
When we know the contest type we know the contest option type. This change encodes this knowledge in the type system.


nikhilb4a (approved):
🙏 💯

PR #6252, opened 2025-04-08
Consistent bubble spacing across HMPBs

jonahkagan:
## Overview

Consolidate the way we vertically space bubbles in relation to the option label. I noticed this was a bit inconsistent in some of the ballots.

Changing this ended up changing the famous names fixture election's grid layouts slightly, which caused a ballot hash change and a cascade of test snapshot updates.

## Checklist


libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -477,21 +469,13 @@ function BallotMeasureContest({
...
-              <div
-                style={{
-                  display: 'flex',
-                  gap: '0.5rem',
-                  paddingTop: '0.25rem',
-                }}
-              >
-                <strong
-                  style={{ flex: 1, textAlign: 'right', marginTop: '-0.25rem' }}
-                >
```
jonahkagan, line 489: This was the main case where the alignment was being done differently

libs/hmpb/src/ballot_templates/vx_default_ballot_template.tsx
```diff
@@ -323,11 +318,7 @@ function CandidateContest({
                   borderTop: `1px solid ${Colors.DARK_GRAY}`,
                 }}
               >
-                {/* Match line-height of text to align bubble to center of write-in candidate name */}
-                <BubbleWrapper
-                  optionInfo={optionInfo}
-                  style={{ height: '1.25rem' }}
```
jonahkagan, line 329: Note that the 1.25rem here changes to the default 1.2rem in `AlignedBubble`. This is the main change that causes fixture and snapshot updates in the rest of this PR. Visually it's basically impercetible

libs/hmpb/src/ballot_components.tsx
```diff
@@ -233,19 +233,19 @@ export function Bubble({
...
 }): React.ReactElement {
   return (
     <div
       style={{
         display: 'flex',
         alignItems: 'center',
-        ...style,
+        // Match line-height of text to align bubble to center of first line of
+        // option label
+        height: '1.2rem',
```
kofi-q, line 248: Seems brittle - would it be easy to have this value be shared between here and wherever the line height is set?
(I'm guessing the latter isn't likely to change, so feel free to ignore if it's too much of a headache)

jonahkagan: Good idea, updated here: a21181836

PR #6254, opened 2025-04-08
VxPollbook: System admin login

kshen0:
## Overview

Adds system administrator login and the following behavior.

* Election screen
  * configure/unconfigure
* Settings screen
  * Save logs
  * Date & Time

I'll open separate PRs for USB formatting and Secure Hash Validation.

## Demo Video or Screenshot
Screenshot 2025-04-08 at 4 50 12 PM
Screenshot 2025-04-08 at 4 50 20 PM
Screenshot 2025-04-08 at 4 50 24 PM

**Configure and unconfigure election**

https://github.com/user-attachments/assets/f5b44f39-1083-4afa-a704-396ababe7ccc

**Settings Screen**

https://github.com/user-attachments/assets/c11dcb75-ad1e-451a-a046-0231b7cb9ac8

## Testing Plan
- [x] Add automated tests

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/app.ts
+++ b/apps/pollbook/backend/src/app.ts
@@ -8,9 +8,15 @@ import {
 } from '@votingworks/types';
 import { DippedSmartCardAuthMachineState } from '@votingworks/auth';
 import React from 'react';
-import { Exporter, getBatteryInfo } from '@votingworks/backend';
+import {
+  createSystemCallApi,
+  Exporter,
+  getBatteryInfo,
+} from '@votingworks/backend';
 import { generateFileTimeSuffix } from '@votingworks/utils';
 import { stringify } from 'csv-stringify/sync';
+import { UsbDriveStatus } from '@votingworks/usb-drive';
+import { Logger } from '@votingworks/logging';
 import {
   Workspace,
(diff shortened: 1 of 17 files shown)


adghayes:
looks good, thank you for the screenshots and videos!

two superficial asks:
- on the elections tab, remove the `FieldName` for "Election" because it's already in the header and the sidebar
- in the unconfigured case, use "Insert card to unlock." instead of enumerating all the card types, too wordy

kshen0:
> in the unconfigured case, use "Insert card to unlock." instead of enumerating all the card types, too wordy

Initially I was concerned that removing the list of valid card types from this screen would create user confusion if they tried to log in with a poll worker card to an unconfigured machine. But the following screen catches this and notifies the user, so seems fine to simplify the text to just "Insert card to unlock"!

adghayes:
Oh @kshen0 I actually miswrote. I meant that in the configured case, where all cards are acceptable, just insert "Insert card to unlock." I think for the unconfigured case it still makes sense to do "Insert a system administrator or election manager card." or whatever it was before, which would be consistent with VxAdmin. Since this is already merged, we can just adjust in a later PR.

PR #6256, opened 2025-04-09
Add basic debug perf logging for all grout endpoints

carolinemodic:
## Overview
Adds debug lines to the grout API call handling to get automatic performance stats when toggled for specific method names. I made it method specific as the auth polling and other typical practices make showing all the perf logging kind of impossible to filter through. 

## Testing Plan
Ran DEBUG=*perf:getCheckInCounts* pnpm start in pollbook/frontend and saw expected perf lines logged.

### DIFF
--- a/libs/grout/src/server.ts
+++ b/libs/grout/src/server.ts
@@ -5,6 +5,7 @@ import { rootDebug } from './debug';
 import { serialize, deserialize } from './serialization';
 
 const debug = rootDebug.extend('server');
+const perfDebug = debug.extend('perf');
 
 /**
  * A Grout RPC method.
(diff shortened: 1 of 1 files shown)


libs/grout/src/server.ts
```diff
@@ -226,6 +230,7 @@ export function buildRouter(
           );
         }
 
+        const durationStart = Date.now();
```
carolinemodic, line 233: not completely sure if the middleware should be included in the timing but I think so since its still part of the time the request takes.

jonahkagan (approved):
This looks good!

I'm mentally noting this as one potential use case for middleware that wraps the method handler (rather than just runs beforehand), so you could build this as a consumer of grout rather than having to build it into grout.

carolinemodic:
Yeah I was wondering if it should be middleware and then realized the middleware doesn't really support this haha.

PR #6270, opened 2025-04-10
[VxDesign] Enable creation/export features for the 'Vx Demos' org

kofi-q:
## Overview

We've [recently set up](https://votingworks.slack.com/archives/C085YT4798C/p1744203568443159) a `Vx Demos` org to use for external demos and would like to be able to demo the full feature set of VxDesign without displaying the full list of orgs like we do on the `VotingWorks` support account.

Adding feature flag configuration for the demo org, with most features switched on.
Will be adding the org ID env variable in prod.

## Testing Plan
- Updated unit tests 
- Manual test with mock IDs

## Checklist

### DIFF
--- a/apps/design/backend/src/features.ts
+++ b/apps/design/backend/src/features.ts
@@ -1,4 +1,4 @@
-import { sliOrgId, votingWorksOrgId } from './globals';
+import { sliOrgId, votingWorksOrgId, vxDemosOrgId } from './globals';
 import { ElectionRecord } from './store';
 import { User } from './types';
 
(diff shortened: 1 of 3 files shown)


apps/design/backend/src/features.ts
```diff
@@ -134,6 +134,22 @@ export const userFeatureConfigs = {
     BALLOT_LANGUAGE_CONFIG: true,
   },
 
+  demos: {
+    ACCESS_ALL_ORGS: false,
+    SYSTEM_SETTINGS_SCREEN: true,
+    MARGINAL_MARK_THRESHOLD: false,
+    EXPORT_SCREEN: true,
+    CHOOSE_BALLOT_TEMPLATE: false,
```
jonahkagan, line 142: Will ballot template default to VX or NH? Is that what we want? Maybe we should turn this feature on

jonahkagan (approved):
Looks good, though definitely makes me think we need to come up with a different solution for flagging orgs soon since adding env vars isn't gonna scale

PR #6273, opened 2025-04-10
Split long ballot measures across multiple pages (NH template only)

jonahkagan:
## Overview

Instead of throwing an error when a ballot measure is too long to fit on one page, splits it into chunks spanning as many pages as necessary to fit the entire contest description. The contest options are only included on the final chunk.

To figure out where to split the description text to maximize the amount shown on each page, measures at each top-level child node of the description HTML and checks whether it would overflow the page content area.

## Demo Video or Screenshot
https://votingworks.slack.com/archives/C085YT4798C/p1744156183526989?thread_ts=1743724353.282999&cid=C085YT4798C

## Checklist


libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -450,46 +465,65 @@ function BallotMeasureContest({
               tableBorderColor={Colors.DARKER_GRAY}
               tableHeaderBackgroundColor={Colors.LIGHT_GRAY}
             >
-              {electionStrings.contestDescription(contest)}
```
jonahkagan, line 453: I couldn't think of any easy way to make this work with translations, and since NH ballots are mono-lingual for now, I punted. But I feel a bit worried about this in the future

jonahkagan:
@kofi-q I'd appreciate a high-level review of this implementation approach when you have a chance - not urgent

kofi-q (commented):
The approach looks good to me!

Wondered if we could get away with one render per contest chunk, if it's possible to set a max height and use the browser's visibility checks to see where the break happens - but, not sure how much lift that is, vs how much that buys us beyond the fewer renders and maybe not relying on splitting at just paragraph boundaries (thinking of images, tables, etc).
Generally feeling like a rare enough case that this is a good starting point and we can see what issues come up, if any

kofi-q (commented):
Very nice! Assuming that the logic for finding the split point is reliable, this approach looks good!
Much less hackery than I feared and presumably, the extra parsing is still less work than extra page renders

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -529,6 +567,74 @@ function Contest({
...
+    (child) =>
+      child.y - contestMeasurements.y + child.height > dimensions.height
+  );
+  const descriptionHtmlNode = parseHtml(tooLongContest.description);
+  for (const overflowingChild of descriptionHtmlNode.childNodes.slice(
+    firstOverflowingChildIndex
+  )) {
+    descriptionHtmlNode.removeChild(overflowingChild);
+  }
+  const splitIndex = descriptionHtmlNode.toString().length;
```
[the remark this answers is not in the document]
jonahkagan: Seems to be pretty reliable, though I've only done limited testing. Luckily the HTML that we're putting into it is pretty tightly constrained by our editor

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
jonahkagan: Comment added: fee440576

kofi-q (approved):
✅

PR #6285, opened 2025-04-15
refactor(admin): simplify adjudication APIs

eventualbuddha:
_This started as some comments on #6210, but became unwieldy._

There were a few APIs that were made to take optional `input` with optional properties, but the client never actually needed that optionality. Similarly, `getWriteInAdjudicationCvrQueueMetadata` took an optional `contestId` parameter that was never provided. This complicated the Grout API and also the `Store` methods, but without providing actual benefit.

### DIFF
--- a/apps/admin/backend/src/app.ts
+++ b/apps/admin/backend/src/app.ts
@@ -75,7 +75,6 @@ import {
   ScannerBatch,
   WriteInAdjudicationAction,
   WriteInAdjudicationQueueMetadata,
-  WriteInAdjudicationStatus,
   WriteInCandidateRecord,
   WriteInAdjudicationContext,
   WriteInImageView,
@@ -639,27 +638,23 @@ function buildApi({
       adjudicateCvrContest(input, store, logger);
     },
 
-    getVoteAdjudications(
-      input: {
-        contestId?: ContestId;
-        cvrId?: Id;
-      } = {}
-    ): VoteAdjudication[] {
+    getVoteAdjudications(input: {
+      contestId: ContestId;
+      cvrId: Id;
+    }): VoteAdjudication[] {
       return store.getVoteAdjudications({
-        ...input,
         electionId: loadCurrentElectionIdOrThrow(workspace),
+        contestId: input.contestId,
+        cvrId: input.cvrId,
       });
     },
 
-    getWriteInAdjudicationQueueMetadata(
-      input: {
-        contestId?: ContestId;
-        status?: WriteInAdjudicationStatus;
-      } = {}
-    ): WriteInAdjudicationQueueMetadata[] {
+    getWriteInAdjudicationQueueMetadata(input: {
+      contestId: ContestId;
+    }): WriteInAdjudicationQueueMetadata[] {
       return store.getWriteInAdjudicationQueueMetadata({
         electionId: loadCurrentElectionIdOrThrow(workspace),
-        ...input,
+        contestId: input.contestId,
       });
     },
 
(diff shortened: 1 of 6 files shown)


apps/admin/backend/src/app.ts
```diff
@@ -639,27 +638,23 @@ function buildApi({
...
       });
     },
 
-    getWriteInAdjudicationQueueMetadata(
-      input: {
-        contestId?: ContestId;
-        status?: WriteInAdjudicationStatus;
-      } = {}
-    ): WriteInAdjudicationQueueMetadata[] {
+    getWriteInAdjudicationQueueMetadata(input: {
```
nikhilb4a, line 652: Note that this function will be deprecated in favor of `getWriteInAdjudicationCvrQueueMetadata`, as it is used for the previous write-in adjudication flow

apps/admin/backend/src/app.ts
```diff
@@ -711,24 +706,15 @@ function buildApi({
...
       return store.getWriteInAdjudicationCvrQueue({
-        ...input,
         electionId: loadCurrentElectionIdOrThrow(workspace),
+        contestId: input.contestId,
       });
     },
 
-    getWriteInAdjudicationCvrQueueMetadata(
-      input: {
-        contestId?: ContestId;
```
nikhilb4a, line 727: Ah okay. Thanks for pointing this out, as it was one of the things I felt uncertain about (had posted this related comment [here](https://github.com/votingworks/vxsuite/pull/6210/commits/a75f3691cfe25006cb2e8965af84a28c3fd58579#r2023842993) but I think it got lost in the shuffle

I had seen that we had functions like `getWriteInAdjudicationQueueMetadata` take in optional parameters that weren't used, so I wondered if we like to add extra flexibility to our APIs. But, it's good to confirm that that is not the approach we take, and that we should only introduce the complexity when needed.

In that case, would you recommend I also update [this function](https://github.com/votingworks/vxsuite/pull/6210/files#diff-efabe1fa4f8509f3c628056d769de94c43d4fa70776abff99b0529f5afe9b182R745) to require the parameters? The only use case right now requires `cvrId` and `contestId`

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -337,13 +337,13 @@ export function ContestAdjudicationScreen(): JSX.Element {
     currentCvrId ? { cvrId: currentCvrId } : undefined
   );
   const voteAdjudicationsQuery = getVoteAdjudications.useQuery(
-    currentCvrId ? { cvrId: currentCvrId, contestId } : undefined
+    currentCvrId ? { cvrId: currentCvrId, contestId: contest.id } : undefined
```
nikhilb4a, line 340: Is this intentional?

nikhilb4a (approved):
Thank you for pointing this out, as it was something I was hoping to discuss 🙏

apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
eventualbuddha: Yes, because at this point TypeScript sees `contestId` as `ContestId | undefined`. We could keep this and add an `assert`.

apps/admin/backend/src/app.ts
eventualbuddha: Yes, I'd say go ahead and make the change there as well.

apps/admin/backend/src/app.ts
eventualbuddha: Feel free to take over this PR and do what you want with it, including just incorporating the relevant changes to your branch and closing this.

nikhilb4a: Okay thanks, I'll incorporate the changes in a commit later today. I appreciate your help with this

nikhilb4a:
Closing as all relevant changes have been incorporated in https://github.com/votingworks/vxsuite/pull/6210/commits/5ca3fc4cd1682610d9a8b09669cd2a7c20a0a688

PR #6290, opened 2025-04-16
VxAdmin: fix double sided printing (bug)

### LINKED ISSUE #6065: VxAdmin: bug, printing reports double-sided
VxAdmin printing double-sided only makes posting reports difficult.

We used to print single-sided but then when moving PDF generation and printing to the backend, some silly product manager ex-engineer created this regression.

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1740612364215509?thread_ts=1740612364.215509&cid=CJU9MSC6S

adghayes:
## Overview

Closes #6065. When I merged #4583, I didn't maintain the single-sided printing for VxAdmin, which makes it [procedurally harder](https://votingworks.slack.com/archives/CJU9MSC6S/p1740612364215509) to display reports. This change forces the prints to be single-sided.

Added some regression testing, but wasn't sure about the best way to fend against regressions. Future report styles should also probably use these options, but it's not necessarily the case that everything we ever print from VxAdmin will.

## Testing Plan

Tested on hardware and wrote regression tests.


apps/admin/backend/src/globals.ts
```diff
@@ -57,3 +58,10 @@ const DEFAULT_ALLOWED_EXPORT_PATTERNS =
 export const ADMIN_ALLOWED_EXPORT_PATTERNS =
   process.env.ADMIN_ALLOWED_EXPORT_PATTERNS?.split(',') ??
   DEFAULT_ALLOWED_EXPORT_PATTERNS;
+
+/**
+ * Print settings for reports, namely one-sided printing.
+ */
+export const REPORT_PRINT_OPTIONS: PrintOptions = {
+  sides: PrintSides.OneSided,
+};
```
jonahkagan, line 67: Is there any situation in VxSuite where we want to print double-sided? If not, we can just hardcode this into libs/printing instead of making it a parameter, which would solve your problem of making sure it's always single-sided.

Or a weaker version would just be to change the default to single-sided.

jonahkagan (approved):

apps/admin/backend/src/globals.ts
adghayes: This solution works and is much simpler! Tested on VxAdmin and VxPollBook.

adghayes:
test is, I think, a flake

PR #6295, opened 2025-04-16
test(scan): replace some NH non-QR code fixture uses

eventualbuddha:
## Overview

We're trying to get rid of the non-QR code AccuVote ballot support, so we need to remove the fixtures. This switches to using the marked fixtures in `@votingworks/hmpb`.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated tests.


jonahkagan (approved):
✅

Issue #4378, opened 2023-12-11
Formalize precinct splits in Election data model


jonahkagan:
Currently, the VXF Election data type doesn't have a formal concept of precinct splits. Instead, multiple ballot styles for a single precinct are used to implicitly represent a precinct split.

VxDesign uses an extended data model that explicitly models precinct splits. CDF also supports an explicit precinct split type.

We should align VXF with VxDesign to make everything more straightforward.

This would also allow us to show precinct split names instead of ballot style IDs in the UI (e.g. in VxMark when selecting a ballot style for a voter, or in VxAdmin when entering manual tallies).

adghayes:
Need to do an investigation / discussion to see if this is still useful, what it would unlock, and what markets it could help us in.

PR #6298, opened 2025-04-16
Add precinct splits to precinct type

jonahkagan:
## Overview

Task: https://github.com/votingworks/vxsuite/issues/4378

Consolidates the `SplittablePrecinct` type with the `Precinct` type so that the whole repo now uses the same type for precincts. The main impacts of this change:
- Precinct split information is included in precincts
- Precincts reference their associated districts, so that information isn't only encoded in ballot styles now

Elections exported from VxDesign with these additional fields should still be compatible with existing v4 systems in the field. However, the changes to VxSuite will not be backwards compatible with previous election definitions.

This PR makes only the minimal changes required. I plan to follow up with some more improvements/simplifications that this change will unlock.

## Demo Video or Screenshot
N/A

## Testing Plan
Updated automated tests

## Checklist


kofi-q (approved):
✅

PR #6300, opened 2025-04-17
test(auth): upgrade to vitest@3

eventualbuddha:
Upgrades `libs/auth` to vitest@3. @arsalansufi wanted to pull in a fix for https://github.com/evanw/esbuild/issues/3838 to get back to 100% test coverage. The specific case of `switch` case statements does seem to be fixed, though it looks like there are other blockers:

image

Still, this is a step in the right direction and an improvement over the status quo, especially since we can use absolute coverage numbers with vitest@3.


arsalansufi (approved):
Thank you!

PR #6305, opened 2025-04-17
Get libs/auth back up to 100% code coverage

arsalansufi:
## Overview

:100:

Separating this out from the PR for VxPollBook cert support in libs/auth. Just bumps us back up to 100% code coverage in libs/auth as I prepare to add new code and subsequently cover it

### DIFF
--- a/libs/auth/vitest.config.ts
+++ b/libs/auth/vitest.config.ts
@@ -3,10 +3,6 @@ import { defineConfig } from '../../vitest.config.shared.mjs';
 export default defineConfig({
   test: {
     coverage: {
-      thresholds: {
-        lines: -12,
-        branches: -22,
-      },
       exclude: [
         'src/cac/index.ts',
         'src/index.ts',


libs/auth/src/artifact_authentication.ts
```diff
@@ -320,9 +320,11 @@ async function performArtifactSpecificAuthenticationChecks(
  */
 export async function prepareSignatureFile(
   artifact: ArtifactToExport,
-  /* istanbul ignore next - @preserve */
-  config: ArtifactAuthenticationConfig = constructArtifactAuthenticationConfig()
+  configOverride?: ArtifactAuthenticationConfig
 ): Promise<{ fileContents: Buffer; fileName: string }> {
+  const config =
+    configOverride ??
+    /* istanbul ignore next - @preserve */ constructArtifactAuthenticationConfig();
```
arsalansufi, line 327: Our latest test setup (Vitest 3) doesn't seem to register istanbul ignores of default param assignments. Rearranged the code to work around that

libs/auth/src/dipped_smart_card_auth.ts
```diff
@@ -116,6 +116,7 @@ function logAuthEventIfNecessary(
     }
 
     case 'checking_pin': {
+      /* istanbul ignore else - @preserve */
```
arsalansufi, line 119: The else block here is conceptually irrelevant

libs/auth/src/cac/common_access_card_certs.ts
```diff
@@ -108,9 +108,9 @@ export async function parseCert(
   const certFields: { [fieldName: string]: string } = {};
   for (const certField of certFieldsList) {
     const [fieldName, fieldValue] = certField.split(' = ');
-    if (fieldName && fieldValue) {
-      certFields[fieldName] = fieldValue;
-    }
+    assert(fieldName !== undefined, 'Malformed cert subject line');
+    assert(fieldValue !== undefined, 'Malformed cert subject line');
```
arsalansufi, line 112: These values should never be undefined if OpenSSL parsing (above) succeeds

kshen0 (approved):

PR #6333, opened 2025-04-23
fix(logging-utils): don't crash when logging from converter

eventualbuddha:
## Overview

Introducing the `ThreadsafeFunction` indirection caused the call context with `logger.logAsCurrentRole` to be lost, so `this` was `undefined` which caused a crash. This changes the API to just accept a plain function to avoid this problem.

## Demo Video or Screenshot
n/a

## Testing Plan
Added a test that uses the real `Logger`, though that's sort of moot with the API change.


kofi-q (approved):
🎉

Issue #6123, opened 2025-03-21
Replace user-facing ballot style IDs with precinct/split name


arsalansufi:
**Original request:**
I'm sure this is written somewhere but allowing towns to override default ballot style IDs with more helpful identifiers, e.g., town, school, for NH locals instead of the default 1, 2

Issue automatically created from Slack feedback reported by arsalan: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1742415494705239?thread_ts=1742415494.705239&cid=CJU9MSC6S

adghayes:
Another solution here is #4378, per @jonahkagan.

jonahkagan:
Related: https://github.com/votingworks/vxsuite/issues/4971

jonahkagan:
Main user issue to solve here is the manual tally entry being confusing

PR #6341, opened 2025-04-24
VxAdmin: Label ballot styles by their districts/party in reports

jonahkagan:
## Overview

Task: #6123 

In the VxMark ballot selection and VxAdmin manual tally contexts, ballot styles are considered in conjunction with a precinct, so we can use precinct/split names to identify them. In the VxAdmin reporting context, a user may want to report on a single shared ballot style across precincts (e.g. if there are larger geographic areas that contain the precinct). So instead of identifying a ballot style/precinct combo, we use the list of districts associated with the ballot style as its label. Every ballot style has a unique list of districts, so this is the right unique label (along with the ballot style's party) when considering a ballot style across the entire election's geography.

## Demo Video or Screenshot

Filtering by ballot style:
Screenshot 2025-05-05 at 2 37 48 PM

Report grouped by ballot style:
Screenshot 2025-05-05 at 2 41 46 PM

## Testing Plan
Updated existing tests

## Checklist


kofi-q (approved):
✅

Issue #6219, opened 2025-04-01
VxMark: CDF log export taking >30m exceeding default timeout


mattroe:
First attempt to repro - may require more info

VV40ECT-87 in https://docs.google.com/spreadsheets/d/1SPW_whhGUI2butbYSOFBc35QSsBPV6k7x-KqDFF2AFQ/edit?gid=0#gid=0

Note from Arsalan: I know that we expect CDF log exports to take notably longer than standard exports. If we find that a large CDF log export can ultimately finish with more time, the solution path may involve deferring the lockout due to hitting the inactive session time limit until after the operation completes and the modal is dismissed. We've talked about wanting this for long running CVR exports, too.

PR #6342, opened 2025-04-24
[logging] Use updated log conversion utility for CDF log exports

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/6219

Addressing slow CDF log exports by switching to the newly added Rust-based conversion utility.

This change covers plain log file conversions - will need to follow up with a change to support converting gzipped files as well.

## Testing Plan
- Conversion utility tested separately; updated consumer unit tests to verify inputs & output handling
- Manual with mock log data

## Checklist

### DIFF
--- a/libs/backend/src/system_call/export_logs_to_usb.ts
+++ b/libs/backend/src/system_call/export_logs_to_usb.ts
@@ -1,4 +1,10 @@
-import { Result, err, ok, throwIllegalValue } from '@votingworks/basics';
+import {
+  Result,
+  deferred,
+  err,
+  ok,
+  throwIllegalValue,
+} from '@votingworks/basics';
 import { UsbDrive } from '@votingworks/usb-drive';
 import * as fs from 'node:fs/promises';
 import { join } from 'node:path';
@@ -15,6 +21,7 @@ import { pipeline } from 'node:stream/promises';
 import { createGunzip, createGzip } from 'node:zlib';
 import { dirSync } from 'tmp';
 import { generateFileTimeSuffix } from '@votingworks/utils';
+import { convertVxLogToCdf } from '@votingworks/logging-utils';
 import { execFile } from '../exec';
 
 const LOG_DIR = '/var/log/votingworks';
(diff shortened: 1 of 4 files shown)


libs/backend/src/system_call/export_logs_to_usb.ts
```diff
@@ -30,6 +37,30 @@ export type LogsResultType = Result<
...
+    },
+    logger.getSource(),
+    machineId,
+    codeVersion,
+    inputPath,
+    outputPath,
+    (error) => (error ? reject(error) : resolve())
+  );
+
+  return promise;
```
jonahkagan, line 61: Nit: I'm not necessarily opposed to `deferred` but I think you could also just use the built-in `Promise` constructor:

```ts
return new Promise((resolve, reject) => 
   convertVxLogToCdf(
    (eventId, message, disposition) => {
      void logger.logAsCurrentRole(eventId, { message, disposition });
    },
    logger.getSource(),
    machineId,
    codeVersion,
    inputPath,
    outputPath,
    (error) => (error ? reject(error) : resolve())
  )
)
```

Related: Is there a reason `convertVxLogToCdf` takes a callback vs returning a promise?

jonahkagan (approved):
👍 so simple!

libs/backend/src/system_call/export_logs_to_usb.ts
```diff
@@ -145,7 +179,7 @@ async function exportLogsToUsbHelper({
           machineId,
           codeVersion
         );
-      } catch {
+      } catch (error) {
```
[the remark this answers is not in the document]
kofi-q: Oops, thanks - had some debugging logging and forgot to clean up

libs/backend/src/system_call/export_logs_to_usb.ts
kofi-q: Yeah, was thinking the same - I'm guessing the callback pattern was easier to implement on the Rust side, but considering switching it over to promise along with the change to support compressed files

PR #6345, opened 2025-04-25
upgrade vitest in libs/backend

adghayes:
## Overview

I thought to do a little mindless code contribution but actually this turned out to be a headache. I started with `libs/backend`, but because all the backends import test harnesses from `libs/backend`, conflicting package versions led me to upgrade all the backends. But because most of the backends import test utilities from `libs/image-utils`, I had to upgrade those and then other backends that import them.

In upgraded libraries, I've set the negative coverage thresholds to their maximum. The exception is `pollbook/backend`, which appears to have variable code coverage on every run, which seems bad, but is outside the scope of this PR.

### DIFF
--- a/apps/design/backend/src/test_decks.test.ts
+++ b/apps/design/backend/src/test_decks.test.ts
@@ -22,7 +22,7 @@ import {
 } from './test_decks';
 
 vi.setConfig({
-  testTimeout: 30000,
+  testTimeout: 60_000,
 });
 
 let renderer: Renderer;
--- a/apps/pollbook/backend/vitest.config.ts
+++ b/apps/pollbook/backend/vitest.config.ts
@@ -7,8 +7,8 @@ export default defineConfig({
     clearMocks: true,
     coverage: {
       thresholds: {
-        lines: 78,
-        branches: 70,
+        lines: -200,
+        branches: -119,
       },
       exclude: [
         '**/node_modules/**',
--- a/libs/auth/src/test_utils.ts
+++ b/libs/auth/src/test_utils.ts
@@ -5,7 +5,10 @@ import {
 import type { Mocked, vi } from 'vitest';
 
 import { DippedSmartCardAuthApi } from './dipped_smart_card_auth_api';
-import { InsertedSmartCardAuthApi } from './inserted_smart_card_auth_api';
+import {
+  InsertedSmartCardAuthApi,
+  InsertedSmartCardAuthMachineState,
+} from './inserted_smart_card_auth_api';
 
 /**
  * Builds a mock dipped smart card auth instance for application-level tests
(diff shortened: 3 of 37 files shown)


libs/auth/src/test_utils.ts
```diff
@@ -32,8 +35,9 @@ export function buildMockInsertedSmartCardAuth(
   fn: typeof vi.fn
 ): Mocked<InsertedSmartCardAuthApi> {
   return {
-    getAuthStatus: fn().mockResolvedValue(
-      InsertedSmartCardAuthTypes.DEFAULT_AUTH_STATUS
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getAuthStatus: fn((machineState: InsertedSmartCardAuthMachineState) =>
+      Promise.resolve(InsertedSmartCardAuthTypes.DEFAULT_AUTH_STATUS)
```
adghayes, line 40: The upgrade created a very strange test failure. In https://github.com/votingworks/vxsuite/blob/main/apps/mark/backend/src/app.ui_strings.test.ts, the last test started failing because, after the upgrade, the nested [mock resets](https://github.com/votingworks/vxsuite/blob/a12d72dcaa3c534727d1eec4292f1e4c5331eebe/libs/backend/src/ui_strings/ui_strings_api_test_runner.ts#L20) in the UI strings test harnesses were resetting. There are some [breaking changes](https://vitest.dev/guide/migration.html) around spies, but I don't think any of them correspond exactly to what I was seeing. The reset will reset mocks to their original interpretation, which can be passed as an argument to `vi.fn`, which is what I do here. But there may be a larger issue with nested test harnesses resetting mocks in their calling test suites.

apps/design/backend/src/test_decks.test.ts
```diff
@@ -22,7 +22,7 @@ import {
 } from './test_decks';
 
 vi.setConfig({
-  testTimeout: 30000,
+  testTimeout: 60_000,
```
adghayes, line 25: Maybe there was a perf hit as part of the upgrade, because I had to bump these timeouts substantially to make these pass.

apps/pollbook/backend/vitest.config.ts
```diff
@@ -7,8 +7,8 @@ export default defineConfig({
     clearMocks: true,
     coverage: {
       thresholds: {
-        lines: 78,
-        branches: 70,
+        lines: -200,
+        branches: -119,
```
adghayes, line 11: The actual number of uncovered lines varies from run to run, and this is a number which has passed so far 🤷

apps/design/backend/src/test_decks.test.ts
eventualbuddha: I remember something to this effect with the upgrade I did as well. I see there are others having issues as well, and one of the sources of the performance regression was fixed in https://github.com/vitest-dev/vitest/pull/7291. However, that was fixed in v3.0.3. Whether there are other perf issues lurking, I don't know.

eventualbuddha (approved):

PR #6347, opened 2025-04-27
CircleCI Tweaks

adghayes:
There are three commits making three distinct tweaks:
1. We have still been compiling all our rust libraries on the "Setup Dependencies" phase, even though our build phase now does this in a smarter way, only installing when necessary. Removing that step reduces our overall credit usage on CircleCI by about 20%. 
2. We have been caching the cargo packages alongside the node modules, but only referencing the cache with the checksum of the pnpm lock file. I think this means that when the cargo lock file changes, the wrong packages can still be loaded from storage. It probably just reinstalls in a later step, but separating the two caches can allow persisting cargo packages even when pnpm packages change.
3. We weren't pointing to the right result files for playwright tests, so CircleCI wasn't tracking test results / flaky tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,7 @@ jobs:
           environment:
             JEST_JUNIT_OUTPUT_DIR: ./reports/
       - store_test_results:
-          path: apps/admin/integration-testing/reports/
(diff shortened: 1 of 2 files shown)


eventualbuddha (approved):
⚡⚡⚡

PR #6349, opened 2025-04-28
VxAdmin: search for election packages on USB drive up to depth 3

### LINKED ISSUE #5673: VxAdmin: election package auto-search only checks the root of the USB, would be helpful to check more
The auto-search for election packages on VxAdmin often doesn’t find anything, even though it’s right on the USB, and I have to select it manually. In particular it's annoying that when I export an election package from VxAdmin, it doesn't later show up as an option in the menu because it's not at the root.

https://votingworks.slack.com/archives/CJU9MSC6S/p1730848259118389?thread_ts=1730848259.118389&cid=CJU9MSC6S

adghayes:
## Overview

Closes #5673. Instead of only searching the root of the USB, also searches up to a depth of 3, which means that packages exported on VxAdmin will later be detected by an unconfigured VxAdmin. This is mostly to address internal annoyance during testing. There are probably few customer use cases because customers should be using freshly formatted drives, but if they are re-using a drive and do not place the VxDesign file at the root, it would likely be caught now.

### DIFF
--- a/libs/fs/src/list_directory.ts
+++ b/libs/fs/src/list_directory.ts
@@ -54,12 +54,15 @@ export type ListDirectoryError =
 
 /**
  * Get entries for a directory, includes stat information for each entry.
- * Requires that the path be absolute.
+ * Requires that the path be absolute. Recursive up to a specified depth.
+ * Defaults to not recursive, i.e. depth = 1.
  */
 export async function* listDirectory(
-  path: string
+  path: string,
+  depth = 1
 ): AsyncGenerator<Result<FileSystemEntry, ListDirectoryError>> {
   assert(isAbsolute(path));
+  assert(depth >= 1, 'depth must be positive');
 
   try {
     const dir = await fs.opendir(path);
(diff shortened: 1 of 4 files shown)


libs/fs/src/list_directory.ts
```diff
@@ -117,15 +124,5 @@ export async function* listDirectory(
...
-      yield result;
-    } else {
-      const fileEntry = result.ok();
-      if (fileEntry.type === FileSystemEntryType.Directory) {
-        yield* listDirectoryRecursive(fileEntry.path);
-      }
-      yield result;
-    }
-  }
+  yield* listDirectory(path, Infinity);
```
jonahkagan, line 127: I kind of feel like it would make more sense from an API perspective for `listDirectoryRecursive` to take a `depth` argument rather than `listDirectory`. Though I don't feel super strongly

jonahkagan (approved):

libs/fs/src/list_directory.ts
adghayes: I did this so `listRecursive` could be called where it is currently called and not change those files. Since I don't feel strongly about it either, going to just leave it.

PR #6351, opened 2025-04-28
refactor(type): remove `NewType`

eventualbuddha:
## Overview

I conducted a [Slack poll](https://votingworks.slack.com/archives/CEL6D3GAD/p1745366852296489) last week about whether the `NewType` pattern as used in `vxsuite` is considered useful or not. The consensus was that `NewType` is not, but the string aliases for IDs–such as `ContestId`–are. This PR removes `NewType` and deals with the fallout.

## Demo Video or Screenshot
n/a

## Testing Plan
Existing automated tests.


jonahkagan (approved):
🧹

PR #6355, opened 2025-05-02
chore: update TypeScript

eventualbuddha:
## Overview

Updates for [TypeScript 5.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html) and [TypeScript 5.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html).

## Demo Video or Screenshot
n/a

## Testing Plan
Automated testing.

### DIFF
--- a/libs/custom-paper-handler/src/driver/driver.test.ts
+++ b/libs/custom-paper-handler/src/driver/driver.test.ts
@@ -39,6 +39,12 @@ let mockWebUsbDevice: MockWebUsbDevice;
 let paperHandlerWebDevice: WebUSBDevice;
 let paperHandlerDriver: PaperHandlerDriver;
 
+function arrayBufferOf(...numbers: number[]): ArrayBuffer {
+  const arrayBuffer = new ArrayBuffer(numbers.length);
+  Buffer.from(numbers).copy(new Uint8Array(arrayBuffer));
+  return arrayBuffer;
+}
+
 export async function setUpMocksAndDriver(): Promise<void> {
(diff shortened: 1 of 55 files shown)


libs/custom-paper-handler/src/driver/driver.test.ts
```diff
@@ -39,6 +39,12 @@ let mockWebUsbDevice: MockWebUsbDevice;
 let paperHandlerWebDevice: WebUSBDevice;
 let paperHandlerDriver: PaperHandlerDriver;
 
+function arrayBufferOf(...numbers: number[]): ArrayBuffer {
```
jonahkagan, line 42: Should we put something like this in libs/basics for all the cases you needed to cover in this PR?

jonahkagan (approved):
✨

libs/custom-paper-handler/src/driver/driver.test.ts
eventualbuddha: I could see a case for `@votingworks/test-utils`. WDYT?

jonahkagan: I guess I was thinking of the pattern used in the previous commits in this PR:

```ts
   const arrayBuffer = new ArrayBuffer(data.byteLength);
    const arrayBufferView = new Uint8Array(arrayBuffer);
    data.copy(arrayBufferView);
```

which is slightly different from this use case. I think some of those cases are not in test files

eventualbuddha: Added `arrayBufferFrom` to `libs/basics`.

PR #6359, opened 2025-05-02
VxAdmin: fix zero index manual tally bug

### LINKED ISSUE #6357: Manual tally entry showing a contest index of 0
Manual tally entry showing a contest index of 0

Issue automatically created from Slack feedback reported by jonah: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1746140557335489

adghayes:
## Overview

Closes #6357. 

The bug was occurring because we were using `indexOf` to identify the index of the `contest` within the list of `contests`. Because these are objects, this only works if the contests are identical object references. Because the `contest` is coming from the `getContestById` function which is actually caching results for faster lookups, the comparison stops working as soon as the `electionDefinition` cached by React Query is different than the one cached by `getContestById`. I'm not sure how this happens in practice, but you can recreate this bug in dev with the React Query devtools by resetting the `getCurrentElectionMetadata` query while doing manual results.

As far as the fix - the current fix is just a one line fix to address the issue. Any suggestions to do more to avoid this sort of footgun are welcome.

As far as testing - it seems like a regression test here would be hyper-specific so I did not write it, but open to writing one.

### DIFF
--- a/apps/admin/frontend/src/screens/tally/manual_tallies_form_screen.tsx
+++ b/apps/admin/frontend/src/screens/tally/manual_tallies_form_screen.tsx
@@ -1,6 +1,7 @@
 import {
   assert,
   assertDefined,
+  find,
   mapObject,
   throwIllegalValue,
 } from '@votingworks/basics';
@@ -41,7 +42,6 @@ import {
 } from '@votingworks/ui';
 import {
   format,
-  getContestById,
   getBallotStyleGroup,
   areContestResultsValid,
 } from '@votingworks/utils';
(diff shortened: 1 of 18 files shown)


apps/admin/frontend/src/screens/tally/manual_tallies_form_screen.tsx
```diff
@@ -551,7 +551,7 @@ function ContestForm({
 
   const contest = getContestById(electionDefinition, contestId);
```
jonahkagan, line 552: My guess is that this is an accidental use of `getContestById`, thinking that it was one of the election utils from libs/types (which don't have any caching). I think we should replace this with a simple `find(election, (contest) => contest.id === contestId)`. To prevent this footgun in the future, maybe we limit the exporting of the tabulation lookup functions somehow or rename them, since they are probably intended to only be used in very specific circumstances but they are named similarly to the regular election utils.

jonahkagan (approved):
Thanks for refactoring!

PR #6362, opened 2025-05-05
[logging-utils] Add support for compressed log conversion

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/6219

Updating the new Rust-based CDF log converter to support optionally streaming to and from `gz`-compressed log files.
Just having clients pass in a `compressed` boolean flag, since we always know at the call site which mode we want.

## Testing Plan
- Updated logging-utils unit tests to run in both compressed and uncompressed modes.


eventualbuddha (approved):
🦀 🥳 

I assume you'll be migrating this in a subsequent PR:

```ts
  // Create CDF for all compressed vx-logs files
  // [TODO] Add support for gzipped files to `libs/logging-utils` and update
  // this conversion.
  for (const file of files) {
    if (file.match(COMPRESSED_VX_LOGS_NAME_REGEX)) {
      const cdfFileName = file.replace('vx-logs', 'vx-logs.cdf.json');
      await pipeline(
        createReadStream(join(logDir, file)),
        createGunzip(),
        (inputStream: AsyncIterable<string>) =>
          buildCdfLog(logger, inputStream, machineId, codeVersion),
        createGzip(),
        createWriteStream(join(outputDir, cdfFileName))
      );
    }
  }
```

kofi-q:
> I assume you'll be migrating this in a subsequent PR

Yeah, building on this on another to switch that over

PR #6364, opened 2025-05-05
refactor(scan): rebuild ballot metadata decoding with `bitstream-io`

eventualbuddha:
## Overview

_(Changes made as I was just tinkering with stuff during the offsite)_

This is a small step in the direction of unifying the HMP and BMD ballot interpretation by making the decoding available as its own crate using `bitstream-io` instead of `bitter` to support encoding as well as decoding (`bitter` only supports reading).

I went through a little journey with introducing my own `BitDecode` trait, thinking I'd use `bitter` under the hood for that and then switch to something that can do read and write. When I found `bitstream-io`, I realized it already has traits that do what I want so I scrapped mine. You can read commit-by-commit and see the whole journey, or just look at it holistically.

## Demo Video or Screenshot
n/a

## Testing Plan
Added a few tests.


.circleci/config.yml
```diff
@@ -1369,6 +1369,25 @@ jobs:
       - store_test_results:
           path: libs/utils/reports/
 
+  test-crate-ballot-encoder-rs:
```
jonahkagan, line 1372: Where are you imagining eventually building the unified ballot interpreter?

I ask because I'm wondering if there's a need for this to live in a separate lib or if it could just live within libs/ballot-interpreter?

jonahkagan (approved):

.circleci/config.yml
eventualbuddha: I could see it being within `libs/ballot-interpreter`. I'll see about consolidating.

libs/ballot-encoder-rs/src/types.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// ```
+    /// # use ballot_encoder_rs::types::PageNumber;
+    /// let page_one = PageNumber::new_unchecked(1);
+    /// let page_two = PageNumber::new_unchecked(2);
+    ///
+    /// assert!(page_one.is_recto());
+    /// assert!(!page_two.is_recto());
+    /// ```
+    #[must_use]
+    pub const fn is_recto(self) -> bool {
```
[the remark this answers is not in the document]
eventualbuddha: I waffled between "intuitive" and "correct". I'll do as you suggest.

.circleci/config.yml
eventualbuddha: Consolidated in [3aead87](https://github.com/votingworks/vxsuite/pull/6364/commits/3aead871fd695ac9f016b2c5e605ea6094f03bf5).

PR #6368, opened 2025-05-06
Fix flaky test in VirtualKeyboard

kshen0:
## Overview

Might fix the flaky test for VirtualKeyboard: https://app.circleci.com/pipelines/github/votingworks/vxsuite/19751/workflows/db1bdd70-6ece-4fd4-9409-48e0b747159a/jobs/833610

I believe the test assertion happens before the DOM updates on rare occasions. To try to fix this, we use `vi.waitFor` to wait until the expected key is focused.

## Demo Video or Screenshot
N/A

## Testing Plan
Couldn't repro locally so we'll see if this solves the issue in CI.


kofi-q (approved):
✅

Issue #5781, opened 2024-12-20
always export all ballot images in VxCentralScan


adghayes:
Currently VxCentralScan only exports ballot images with write-ins, by default. We need to just export all images for the purposes of adjudication.

adghayes:
- Remove the "Save Backup" button and flow from the settings menu
- The "Save CVRs" button should now save all images _and_ the rejected ballots
- Unconfiguring should no longer block on saving a backup because, well, you can't. It should just block on saving CVRs

PR #6378, opened 2025-05-06
CentralScan: Always export all ballot images

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/5781

Previously `central-scan`'s `exportCastVoteRecordsToUsbDrive` mutation offered a minimal export that excluded most ballot images (those w/o write-ins and rejected ballots). Now we plan to include them all to ensure ballots with marginal marks will be viewable. If performance is hit too significantly, we can optimize this by filtering out images. Conversation [here](https://votingworks.slack.com/archives/CEL6D3GAD/p1746555973582259).

This PR removes the `isMinimalExport` option for cvr exports on `central-scan` and the "Save Backup" UI option, as "Save CVRs" will now always perform the same comprehensive backup that "Save Backup" previously did.

### New (w/o save backup)
https://github.com/user-attachments/assets/ab99b9b3-cb31-4ac5-a386-890982741c55

### Old (w/ save backup)
https://github.com/user-attachments/assets/750e0d3d-87bb-475e-81d6-460ca9393cf2

## Testing Plan

 I've updated the relevant tests in `libs/backend` that test whether the image correct files are exported for accepted and rejected ballots with and without write-ins. Now, all should be exported on each export, just as is the case from `scan` already.

Struggling to manually test scanning ballots locally to test the "Save CVRs" or "Unconfigure Machine" flow block on saving cvrs, but none of that logic is really changed here anyways and I reviewed the automated tests and they accurately cover that logic.

## Checklist

### DIFF
--- a/apps/central-scan/backend/src/app.ts
+++ b/apps/central-scan/backend/src/app.ts
@@ -241,24 +241,20 @@ function buildApi({
       await importer.doZero();
     },
 
-    async exportCastVoteRecordsToUsbDrive(input: {
-      isMinimalExport?: boolean;
-    }): Promise<Result<void, ExportCastVoteRecordsToUsbDriveError>> {
-      const logItem = input.isMinimalExport ? 'cast vote records' : 'backup';
+    async exportCastVoteRecordsToUsbDrive(): Promise<
+      Result<void, ExportCastVoteRecordsToUsbDriveError>
+    > {
+      const logItem = 'all accepted and rejected cast vote records';
       await logger.logAsCurrentRole(LogEventId.ExportCastVoteRecordsInit, {
         message: `Exporting ${logItem}...`,
       });
       const exportResult = await exportCastVoteRecordsToUsbDrive(
         store,
         usbDrive,
-        input.isMinimalExport
-          ? store.forEachAcceptedSheet()
-          : store.forEachSheet(),
-        { scannerType: 'central', isMinimalExport: input.isMinimalExport }
+        store.forEachSheet(),
+        { scannerType: 'central' }
       );
-      if (!input.isMinimalExport) {
-        store.setScannerBackedUp();
-      }
+      store.setScannerBackedUp();
       if (exportResult.isErr()) {
         await logger.logAsCurrentRole(
           LogEventId.ExportCastVoteRecordsComplete,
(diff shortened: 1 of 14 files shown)


apps/central-scan/frontend/src/screens/scan_ballots_screen.tsx
```diff
@@ -247,12 +244,12 @@ export function ScanBallotsScreen({
...
             title={
               <span>
-                <Icons.Warning color="warning" /> Backup Required
+                <Icons.Warning color="warning" /> CVR Backup Required
               </span>
             }
             content={
               <P>
-                Go to <Font weight="semiBold">Settings</Font> and save a backup
+                Click <Font weight="semiBold">Save CVRs</Font> to save a backup
```
nikhilb4a, line 252: @adghayes tagging in case you want to review the copy on this pr

apps/central-scan/frontend/src/screens/settings_screen.tsx
```diff
@@ -68,16 +64,11 @@ export function SettingsScreen({
       </ButtonRow>
       {!canUnconfigure && (
         <Caption>
-          <Icons.Warning color="warning" /> You must save a backup before you
-          can unconfigure this machine.
+          <Icons.Warning color="warning" /> You must save cvrs before you can
+          unconfigure this machine.
```
adghayes, line 68: cvrs -> CVRs

apps/central-scan/frontend/src/screens/scan_ballots_screen.tsx
adghayes: Suggestion, to match the other screen, "You must save CVRs before you can delete all batches." 

The button is right on the screen so I think we don't need to indicate what to click, and that pattern is not super common in `vxsuite` anyway (although I could see an argument for it).

nikhilb4a: Agreed, updated

apps/central-scan/backend/src/app.ts
```diff
@@ -241,24 +241,20 @@ function buildApi({
...
     },
 
-    async exportCastVoteRecordsToUsbDrive(input: {
-      isMinimalExport?: boolean;
-    }): Promise<Result<void, ExportCastVoteRecordsToUsbDriveError>> {
-      const logItem = input.isMinimalExport ? 'cast vote records' : 'backup';
+    async exportCastVoteRecordsToUsbDrive(): Promise<
+      Result<void, ExportCastVoteRecordsToUsbDriveError>
+    > {
+      const logItem = 'all accepted and rejected cast vote records';
```
eventualbuddha, line 247: Inline this?

eventualbuddha (approved):

PR #6380, opened 2025-05-07
Generate scan diagnostic data in the Hardware Test App

eventualbuddha:
## Overview
Adds a basic package to do some analysis of the scanned images during HTA scanning. The analysis is written both as JSON adjacent to the image files and to a CSV. The data is written to the USB drive if one is inserted.

## Demo Video or Screenshot
Some simple plots of the data from a test run, courtesy of @pieguyatx:
image

## Testing Plan
Tested manually for a bit, shared some iterations with #hardware-testing.


nikhilb4a (approved):
Very cool!

Issue #6217, opened 2025-04-01
VxMark: audio on Review Your Votes screen is garbled/staticky


mattroe:
VV40ECT-81 in https://docs.google.com/spreadsheets/d/1SPW_whhGUI2butbYSOFBc35QSsBPV6k7x-KqDFF2AFQ/edit?gid=0#gid=0

mattroe:
@kofi-q see the related VV40ECT-111 issue - let's confirm both are addressed by this

kofi-q:
re: VV40ECT-111 - did they capture an audio sample of that one? Our decreased-speed audio has definitely always been not so great and I'm wondering if that's what they're flagging, or if it was compounded by the perf issues

mattroe:
Wasn't reading it quickly - agree they are flagging the general problem, ugh.

arsalansufi:
Closing this out in favor of the new issue in the Cert Discrepancies project: https://github.com/votingworks/cert-discrepancies/issues/103

PR #6383, opened 2025-05-07
[libs/auth] Fix audio performance issues due to unstable auth data

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/6217

We recently added response diffing logic to polling requests across VxSuite to skip updating the cache (and triggering re-renders) when the data hasn't changed.

On VxMark, we're running into an edge case where the auth object changes on every request, since we generate a new (unused) session expiry date each time and this was causing performance issues on pages with lots of elements and manifesting in glitchy audio playback.

Patching by locking the session expiry to a static future date, since it's unused, to keep the auth status response stable for diffing. Voter sessions have a separate timeout logic path.

## Testing Plan
- Reproduced audio issues in a VM with CPU limiting on the kiosk-browser process - then, verified that there's no audible glitching after this change.
- Will also be testing the before and after on a real BMD to verify - there are still some unrelated, remaining perf issues to work out.

## Checklist

### DIFF
--- a/libs/auth/src/inserted_smart_card_auth.ts
+++ b/libs/auth/src/inserted_smart_card_auth.ts
@@ -157,6 +157,8 @@ function logAuthEvent(
   }
 }
 
+const NEVER = new Date('01/01/9000');
+
 /**
  * The implementation of the inserted smart card auth API
  */
@@ -205,7 +207,7 @@ export class InsertedSmartCardAuth implements InsertedSmartCardAuthApi {
       return {
         status: 'logged_in',
         user: this.cardlessVoterUser,
-        sessionExpiresAt: computeSessionEndTime(machineState),
+        sessionExpiresAt: NEVER,
       };
     }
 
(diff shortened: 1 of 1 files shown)


libs/auth/src/inserted_smart_card_auth.ts
```diff
@@ -205,7 +207,7 @@ export class InsertedSmartCardAuth implements InsertedSmartCardAuthApi {
       return {
         status: 'logged_in',
         user: this.cardlessVoterUser,
-        sessionExpiresAt: computeSessionEndTime(machineState),
+        sessionExpiresAt: NEVER,
```
jonahkagan, line 210: Maybe leave a comment explaining why it's safe to use `NEVER` for this field?

jonahkagan (approved):

libs/auth/src/inserted_smart_card_auth.ts
kofi-q: Ah, good call - will do

arsalansufi:
Nice find!

PR #6389, opened 2025-05-08
Kevin/pollbook move new registration to election manager

kshen0:
## Overview

Moves new registration flow to election manager page instead of poll worker.

## Demo Video or Screenshot
Screenshot 2025-05-07 at 5 41 31 PM

https://github.com/user-attachments/assets/3cd37f19-20c0-4e8d-a617-892fddfe59a6

Poll worker screen now has a nav with only 1 item in it. I think this is ok for now if we plan to add more poll worker functionality later. We can also remove it if we have no plans for that.

Screenshot 2025-05-07 at 5 42 35 PM

## Testing Plan

Updated e2e frontend test 

## Checklist

### DIFF
--- a/apps/pollbook/frontend/src/app_election_manager_screen.test.tsx
+++ b/apps/pollbook/frontend/src/app_election_manager_screen.test.tsx
@@ -1,0 +1,111 @@
+import { expect, test, beforeEach, afterEach, vi } from 'vitest';
+
+import {
+  ValidStreetInfo,
+  VoterRegistrationRequest,
+} from '@votingworks/pollbook-backend';
+import { Election } from '@votingworks/types';
+import { electionFamousNames2021Fixtures } from '@votingworks/fixtures';
+
+import userEvent from '@testing-library/user-event';
+import {
+  ApiMock,
+  createApiMock,
+  createMockVoter,
+} from '../test/mock_api_client';
+import { act, render, screen } from '../test/react_testing_library';
+import { App } from './app';
+import { AUTOMATIC_FLOW_STATE_RESET_DELAY_MS } from './globals';
+import { DEFAULT_QUERY_REFETCH_INTERVAL } from './api';
+
+let apiMock: ApiMock;
+const famousNamesElection: Election =
+  electionFamousNames2021Fixtures.readElection();
+
+beforeEach(() => {
+  vi.useFakeTimers({ shouldAdvanceTime: true });
+  vi.clearAllMocks();
+  apiMock = createApiMock();
+  apiMock.setElection(undefined);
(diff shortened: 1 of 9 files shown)


apps/pollbook/frontend/src/app_election_manager_screen.test.tsx
```diff
@@ -0,0 +1,111 @@
+import { expect, test, beforeEach, afterEach, vi } from 'vitest';
```
carolinemodic, line 1: should this be in election_manager_screen.test or should that be renamed to app_election_manager_screen.test? I guess you have different beforeEach steps here but I'm not really following the distinction between the test files.

apps/pollbook/frontend/src/app_election_manager_screen.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  // Allow printer status query to refresh
+  vi.advanceTimersByTime(DEFAULT_QUERY_REFETCH_INTERVAL);
+
+  await screen.findByText('Voter Registration');
+  const lastNameInput = await screen.findByLabelText('Last Name');
+  userEvent.type(lastNameInput, voter.lastName);
+  const firstNameInput = screen.getByLabelText('First Name');
+  userEvent.type(firstNameInput, voter.firstName);
+  userEvent.type(screen.getByLabelText('Street Number'), '1000');
```
carolinemodic, line 94: could be nice to add a test for the bad address case but I know thats not really related to this PR haha

carolinemodic (approved):
Agree that the 1 nav item on the pollworker screen feels a little weird but that it probably makes sense to leave it while we're in heavy development and clean up later if we don't end up adding back anything else.

apps/pollbook/frontend/src/app_election_manager_screen.test.tsx
kshen0: The `app_<role>_screen` tests predate the individual component tests. I ported this test with as little diff as possible. The main difference is in this test we render the `<App>` component and in `election_manager_screen.test.tsx` we render the `ElectionManagerScreen` component. `app.tsx` is a pretty thin wrapper on these screens, basically just determining auth and rendering appropriately, so there's not much value added for the "e2e" tests. It's a good idea to just move this test into `election_manager_screen.test.tsx`.

kshen0: Something about the component test doesn't work well with fake timers, so going to leave this in its own file unless we have a compelling reason to combine that outweighs the cost of investigating the fake timer issue.

Issue #6163, opened 2025-03-26
VxPollBook: Networked Configuration


adghayes:
Enable loading a configuration from a networked machine. UX and authentication details TBD.

- [x] Initial backend endpoints
- [x] Initial sys admin flows
- [ ] Update configuration to try all possible machines for a given election if the first fails
- [ ] Add EM flows - remove usb flows for EM 
- [ ] Add warning to sys admin screen when there are multiple pollbook package hash values for the same election id

carolinemodic:
TODO: 
- Smarter logic in trying configures from multiple machines, if one fails and there are many we should automatically try the next one 
- Election Manager flows to configure from the network

adghayes:
@carolinemodic is it correct that this and #6188 will be closed once #6420 is merged?

PR #6390, opened 2025-05-08
Election Manager Configuration over the Network

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/6163
Implements the election manager flows for configuration over the network and removes the ability for an election manager to configure from usb. I introduce a polling interval on the backend (I initially did this on the fe but that was pretty fragile and hard to manage) that will automatically configure from the network when an election manager is authenticated and there is another machine (or more then one other machine) on the network with the same election id as the election manager auth has. If there are multiple machines with different pollbook package hashes that results in an error. Copy for all the different error states still somewhat in flux, will do a pass with @adghayes on that. Copy doc linked below for reference.

Copy Doc: https://docs.google.com/document/d/1pmkZO2LN1qCdxMgr06tnvr4M4h2geUEObOoK7xyPz4A/edit?tab=t.ajjhtiscmdt1#heading=h.rfro7nx7g939 

I also added some resiliencey to the configureFromPeer endpoint if the machine is already configured or it is called twice quickly and the sql insert query results in a unique constraint error. I had this happen when I was trying to trigger the configuration from the frontend which shouldn't be a problem anymore but it seems like a generally good idea to be reilient to that race condition anyway. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/2d2c5d4f-0b96-4962-8439-ca9cb5bda1ef

## Testing Plan
Manual testing, added basic frontend/backend coverage


apps/pollbook/backend/src/pollbook_package.ts
```diff
@@ -255,3 +258,100 @@ export function pollUsbDriveForPollbookPackage({
...
+          // Check if we have successfully configured!
+          if (result.isOk()) {
+            workspace.store.setConfigurationStatus(undefined);
+            clearInterval(intervalId); // Stop the polling interval
+            return;
+          }
+        }
+        // If we made it here all machines failed when we tried to configure from them
+        workspace.store.setConfigurationStatus('network-configuration-error');
+        hadConfigurationError = true;
```
kshen0, line 351: does tracking `hadConfigurationError` across polling intervals serve a purpose, or can we just run `clearInterval` and return? Then we wouldn't need to track `hadConfigurationError`

apps/pollbook/backend/src/pollbook_package.ts
```diff
@@ -255,3 +258,100 @@ export function pollUsbDriveForPollbookPackage({
...
+              pollbook.electionBallotHash ===
+                matchingConfiguredPollbooks[0].electionBallotHash &&
+              pollbook.pollbookPackageHash ===
+                matchingConfiguredPollbooks[0].pollbookPackageHash
+          );
+
+        if (configuredPollbooks.length === 0) {
+          workspace.store.setConfigurationStatus('not-found-network');
+          return;
+        }
```
kshen0, line 321: nit: do this check on L305, closer to assignment and avoiding unnecessary list operations?

apps/pollbook/backend/src/pollbook_package.ts
```diff
@@ -255,3 +258,100 @@ export function pollUsbDriveForPollbookPackage({
...
+        if (matchingConfiguredPollbooks.length === 0) {
+          workspace.store.setConfigurationStatus(
+            'network-has-other-configurations'
+          );
+          return;
+        }
```
kshen0, line 327: Makes sense to avoid scope creep and just bail out here. I wonder if it's useful in the future to show a list of pollbooks and their configurations so users can easily identify the mismatched one(s)

apps/pollbook/backend/src/pollbook_package.ts
```diff
@@ -255,3 +258,100 @@ export function pollUsbDriveForPollbookPackage({
...
+                matchingConfiguredPollbooks[0].pollbookPackageHash
+          );
+
+        if (configuredPollbooks.length === 0) {
+          workspace.store.setConfigurationStatus('not-found-network');
+          return;
+        }
+        if (matchingConfiguredPollbooks.length === 0) {
+          workspace.store.setConfigurationStatus(
+            'network-has-other-configurations'
```
kshen0, line 324: At first glance `'network-has-other-configurations'` and `'network-multiple-pollbook-packages'` seem like they would describe the same thing. Is the former also described by "no-configured-pollbook-matching-card-election" or something similar?

kshen0 (changes requested):

apps/pollbook/backend/src/peer_app.ts
```diff
@@ -78,12 +82,15 @@ function buildApi(context: PeerAppContext) {
...
-        store.setElectionAndVoters(
+        const error = store.setElectionAndVoters(
           pollbookPackage.electionDefinition,
           pollbookPackage.packageHash,
           pollbookPackage.validStreets,
           pollbookPackage.voters
         );
+        if (error) {
+          return err(error);
+        }
```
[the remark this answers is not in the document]
carolinemodic: This should not happen frequently, I only saw it in testing when I was triggering the configuration from the frontend and  had a bug making it always trigger twice in a row. Regardless it doesn't break anything to let it result in an error instead of locking.

apps/pollbook/backend/src/pollbook_package.ts
carolinemodic: we can not clear the interval and return as we need to continue to configure the machine, the current implementation clear the error when the user logs out so that you can log out and back in again to try configuring again. If we cleared the interval then this would completely stop and you could never configure through the auto-network configure

apps/pollbook/backend/src/pollbook_package.ts
```diff
@@ -199,13 +198,8 @@ export function pollUsbDriveForPollbookPackage({
...
         );
-        if (
-          !isElectionManagerAuth(authStatus) &&
-          !isSystemAdministratorAuth(authStatus)
-        ) {
-          usbDebug(
-            'Not logged in as election manager or system admin, not configuring'
-          );
+        if (!isSystemAdministratorAuth(authStatus)) {
+          usbDebug('Not logged in as system admin, not configuring');
```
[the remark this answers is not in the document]
carolinemodic: Yes the polling is started when a machine is unconfigured, or booted unconfigured, and only stopped when unconfigured. This interval for the usb drive configuration needs to ignore and not do anything unless we're logged in as a system admin and the other interval for the auto network configuration is only for election managers. Its ultimately a bit confusing of a state diagram but I'm not sure there is a clear way to easily make it better.

apps/pollbook/backend/src/pollbook_package.ts
carolinemodic: that is shown in the sys admin screen

kshen0 (approved):

Issue #6253, opened 2025-04-08
VxMark: hardware testing app v2


mattroe:
- new UI
- pausing
- additional applicable features we added to scan

PR #6392, opened 2025-05-08
refactor: move shared HTA infrastructure to libraries

eventualbuddha:
## Overview

Refs #6253 

Moves `TaskController` and `ElectricalTestingScreenNew` to `libs` to prepare for using them in VxMark. I renamed VxMark's `ElectricalTestingScreen` to have the suffix `New` so we could keep the old one since VxMark still uses it, but I'll replace VxMark's version with the new one and drop the suffix in a follow-up PR.

## Demo Video or Screenshot
n/a, functionality should be unchanged.

## Testing Plan
Tested manually with the scanner, sound, and card reader.


libs/ui/src/electrical_testing_screen_new.tsx
```diff
@@ -0,0 +1,192 @@
...
+            </Caption>
+          )}
+          {body && <Caption style={{ flexGrow: 1 }}>{body}</Caption>}
+          {updatedAt && (
+            <Caption style={{ flexGrow: 0 }}>
+              <ExtraSmall>{formatTimestamp(updatedAt)}</ExtraSmall>
+            </Caption>
+          )}
+        </Column>
+        {typeof isRunning === 'boolean' && onToggleRunning && (
```
nikhilb4a, line 118: Is `isRunning === 'boolean'` an alternative to `isRunning !== undefined` here?

nikhilb4a (approved):
LGTM!

libs/ui/src/electrical_testing_screen_new.tsx
eventualbuddha: Yes, it's effectively the same here.

Issue #6328, opened 2025-04-22
VxMark - Write-in Audio


mattroe:
VV40ECT-105 in https://docs.google.com/spreadsheets/d/1jyOM_rz1kIplJlWjxd_eEbwwEkmM7bQ6iQthLvuJsvE/edit?gid=639045837#gid=639045837

Write-in entries are not read back to the voter on the review screen as entered

PR #6394, opened 2025-05-08
[VxMark] Read out write-in candidate names just are they were typed

kofi-q:
## Overview

Resolves https://github.com/votingworks/vxsuite/issues/6328

Updating the candidate contest and review screens to read out write-in candidate names as the same series of letters that were typed out on the write-in keyboard (we would previously read out "write-in" instead).

### TODO
- Noticing how weird it sounds for the screen reader to say "single quote" - will follow up with a PR to update that to say "apostrophe" instead.

https://github.com/user-attachments/assets/15a8bbcc-44e3-452b-ba5a-01f5e302b211

https://github.com/user-attachments/assets/4f29f962-8ee4-43b5-ba2c-899f607ad84e

## Testing Plan
- Mostly manual for verifying audio - with basic unit tests to make sure expected, valid characters render successfully as individual letters.

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/app_contest_write_in.test.tsx
+++ b/apps/mark-scan/frontend/src/app_contest_write_in.test.tsx
@@ -128,8 +128,9 @@ test('Single Seat Contest with Write In', async () => {
 
   // Review Screen
   await screen.findByText('Review Your Votes');
-  expect(screen.getByText('SAL')).toBeTruthy();
-  expect(screen.getByText(/\(write-in\)/)).toBeTruthy();
+  // Expect one instance rendered for display and one for audio:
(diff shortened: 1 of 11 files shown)


jonahkagan (approved):
Looks good!

I do think it would probably be slightly more usable if the audio spoke "Write-in" before reading out the characters instead of after, but not sure how hard that would be to implement

kofi-q:
> I do think it would probably be slightly more usable if the audio spoke "Write-in" before reading out the characters instead of after, but not sure how hard that would be to implement

+1! Initially started down that path, but it was a bit more excavation than necessary for this PR. Will either follow up later or just open a task to address separately

Issue #6179, opened 2025-03-26
VxPollBook: Use Entire Poll Book Package for Network Hash


adghayes:
Currently we're only hashing the election. We could hit weird problems if different machines are loaded with different street or voter files, and hashing the whole thing could avoid a category of weird corner cases.

PR #6396, opened 2025-05-08
Implement Pollbook Package Hash

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/6179

Following up on https://github.com/votingworks/vxsuite/pull/6370 this PR finishes the implementation of the new pollbook package hash. It updates the networking logic to only "connect" two machines and share events in the ballot hash and pollbook package hash match, rather than checking the election id. In addition we update the frontend to show the first 7 characters each of the ballotHash-pollbookPackageHash as the election ID in the UI, matching the ballotHash-electionPackageHash structure in vxsuite. 

## Demo Video or Screenshot
Screenshot 2025-05-08 at 12 19 09 PM

Screenshot 2025-05-08 at 12 18 59 PM

## Testing Plan
Tested UI was updated as appropriate, tested basic connection logic.

### DIFF
--- a/apps/pollbook/backend/src/app_multi_node.test.ts
+++ b/apps/pollbook/backend/src/app_multi_node.test.ts
@@ -445,6 +445,7 @@ test('connection status is managed properly with many pollbooks', async () => {
     for (const context of pollbookContexts) {
       expect(context.peerWorkspace.store.getNewEvents).not.toHaveBeenCalled();
     }
+    // Set all the pollbooks to the same election package and pollbook package
     for (const context of pollbookContexts) {
       context.workspace.store.setElectionAndVoters(
         electionDefinition,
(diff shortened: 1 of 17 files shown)


apps/pollbook/backend/src/networking.ts
```diff
@@ -102,8 +102,11 @@ export function fetchEventsFromConnectedPollbooks({
               return;
             }
             if (
-              !election ||
-              currentPollbookService.electionId !== election.id
+              !myMachineInformation?.electionBallotHash ||
+              myMachineInformation.electionBallotHash !==
+                currentPollbookService.electionBallotHash ||
```
kshen0, line 107: if `currentPollBookService.electionBallotHash` is guaranteed to not be undefined can this simplify down to 

```
 myMachineInformation?.electionBallotHash !==
                currentPollbookService.electionBallotHash || ...
```

apps/pollbook/backend/src/app_multi_node.test.ts
```diff
@@ -481,6 +482,49 @@ test('connection status is managed properly with many pollbooks', async () => {
...
+        vitest.advanceTimersByTime(NETWORK_POLLING_INTERVAL);
+        for (const context of pollbookContexts) {
+          expect(
+            await context.localApiClient.getDeviceStatuses()
+          ).toMatchObject({
+            network: {
+              isOnline: true,
+              pollbooks: [
+                expect.objectContaining({
+                  status: PollbookConnectionStatus.WrongElection,
```
kshen0, line 510: is the pollbook package still close enough to the concept of an election that `WrongElection` still makes sense in this context?

apps/pollbook/backend/src/networking.ts
kshen0: never mind i guess it's possible that both my machine and the peer machine are unconfigured and therefore have no election ballot hash?

carolinemodic: yeah I want to make sure that the current machine is configured

apps/pollbook/backend/src/app_multi_node.test.ts
carolinemodic: yes

carolinemodic: there is a ticket to improve the network status table so maybe we would want to differentiate between those cases when we get there but I doubt it I think its easier to just have it all in a wrong election bucket and start showing the hashes in that table

PR #6405, opened 2025-05-08
fix(scan): correct `getLastScannedSheet` API

eventualbuddha:
This was broken by #6380 because we started writing JSON files alongside the images. This caused the "look for the last sheet images" logic to break since it would now find 4 files instead of 2. This fixes it by restricting to images before doing anything else.

### DIFF
--- a/apps/scan/backend/src/electrical_testing/app.ts
+++ b/apps/scan/backend/src/electrical_testing/app.ts
@@ -4,6 +4,7 @@ import * as grout from '@votingworks/grout';
 import { SheetOf } from '@votingworks/types';
 import express, { Application } from 'express';
 import { readdir } from 'node:fs/promises';
+import { join } from 'node:path';
 import { getMachineConfig } from '../machine_config';
 import { ElectricalTestingComponent } from '../store';
 import { type ServerContext } from './context';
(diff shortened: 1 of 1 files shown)


apps/scan/backend/src/electrical_testing/app.ts
```diff
@@ -108,10 +109,18 @@ function buildApi({
...
 
     async getLatestScannedSheet(): Promise<SheetOf<string> | null> {
-      const allScannedImageNames = await readdir(workspace.ballotImagesPath);
-      const latestElectricalTestingImageName = iter(allScannedImageNames)
-        .filter((name) => name.startsWith('electrical-testing-'))
-        .max((a, b) => a.localeCompare(b));
+      const usbDriveStatus = await usbDrive.status();
+      const basedir =
+        usbDriveStatus.status === 'mounted'
+          ? join(usbDriveStatus.mountPoint, 'ballot-images')
```
nikhilb4a, line 115: Curious - what does this addition give? Wondering if it's adding production machine support, and previously it was set up for dev workspace

nikhilb4a (approved):

apps/scan/backend/src/electrical_testing/app.ts
eventualbuddha: Another change in #6380 was to write the images to the USB drive instead of the dev workspace if the USB drive is mounted. This updates the code that looks for the images to reflect that.

PR #6412, opened 2025-05-09
[VxDesign] Add Rust installation to deploy container image

kofi-q:
## Overview

The VxDesign deploy pipeline's been broken by a new transitive dependency on Rust tooling (which previously didn't need to be installed).

Adding a Rust installation layer to the Docker image to unblock. This also required removing `mark-scan` from the `dockerignore` list, since the Rust build fails otherwise, when it can't find all the projects referenced in the root `Cargo.toml`

## Testing Plan
- Via a Heroku review app

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,5 +9,4 @@ heroku.yml
 apps/admin
 apps/central-scan
 apps/scan
-apps/mark-scan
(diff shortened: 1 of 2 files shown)


jonahkagan (approved):
Thanks for fixing this!

Issue #5780, opened 2024-12-20
VxAdmin: import all mark thresholds


adghayes:
Currently VxAdmin only imports the tabulated marks (binary yes/no) when importing a CVR. We need to import the full mark thresholds so they can be used for marginal mark adjudication.

nikhilb4a:
@adghayes should this ticket also include importing all ballot images? Thinking so

adghayes:
@nikhilb4a good question - I've been assuming that we would just switch to import all ballot images, because that's where we're headed eventually - an interface where one can adjudicate any ballot. I think we should do it and deal with the perf issues later, because they will be mild for NH. cc @mattroe @arsalansufi, to confirm you're not concerned about this for NH.

nikhilb4a:
Looking at the code, it should be pretty similar to only import the images we need since the logic is already set up to support that pattern with write-ins, so I think we may as well to get that time benefit while we don't need the other images

Using Arsalan's performance measurements from [before](https://votingworks.slack.com/archives/CEL6D3GAD/p1696262663599539), it looks like after [this PR](https://github.com/votingworks/vxsuite/pull/4056) we see a drop from 28s to 2s by omitting the ballot images for 1000 ballots

adghayes:
Wow, I did not know there was a performance impact like that! Long-term, I wonder if we're going to have to do something different on import. I also wonder how much of the perf problem is having to read files into memory to hash them as opposed to copy them to disk.

Anyway, for this milestone, sounds good to just import the ones we need.

As you're thinking about what contests to tag / what images to import, also think about #6371 and how that should be sequenced with this work.

arsalansufi:
Thanks for considering this! Agreed that we should keep the optimization for this upcoming release, though I acknowledge that we have to revisit this eventually for the future release that contains consolidated CVR export on VxAdmin

PR #6415, opened 2025-05-10
Read Mark Scores from Original Snapshots for HMPBs on CVR import

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/5780

To enable marginal mark adjudication, we want to start reading mark scores into VxAdmin for HMPBs. These scores exist on the `Original` CVR snapshots under the field `MarkMetricValue`. This PR updates our flow to include the original snapshots on all hmpb exports, read them on import, and then keep them in the CVR table of the Admin store.

This PR updates fixtures to include original snapshots in cvr exports. It also [updates](https://github.com/votingworks/vxsuite/pull/6415#discussion_r2087585919) BMD fixtures to use the original snapshot instead of modified, keeping in line with our practice in `libs/backend` for actual BMD exports.

Next, I will be able to use these mark scores to surface marginal marks for adjudication in Admin.

## Demo Video or Screenshot

Screenshot 2025-05-09 at 5 41 09 PM
Sample where I logged the mark scores on the client.

## Checklist


libs/fixture-generators/src/generate-cvrs/generate_cvrs.ts
```diff
@@ -283,16 +294,14 @@ export function* generateCvrs({
...
+                CurrentSnapshotId: `${uniqueId}-original`,
+                UniqueId: uniqueId,
                 CVRSnapshot: [
                   {
                     '@type': 'CVR.CVRSnapshot',
-                    '@id': `${castVoteRecordId}-modified`,
+                    '@id': `${uniqueId}-original`,
                     ...buildCVRSnapshotBallotTypeMetadata(ballotType),
-                    Type: CVR.CVRType.Modified,
+                    Type: CVR.CVRType.Original,
```
nikhilb4a, line 304: I made this change after noticing in `libs/backend` we only use the original snapshot for BMDs, but in this fixture-generator we use the `modified` snapshot. I don't think it has mattered, because later on we choose the snapshot via the `CurrentSnapshotId` key, so we find the existing snapshot.

Let me know if you think I should omit this change

libs/utils/src/cast_vote_records.ts
```diff
@@ -38,6 +39,16 @@ export function getCurrentSnapshot(cvr: CVR.CVR): Optional<CVR.CVRSnapshot> {
   );
 }
 
+/**
+ * Returns the original snapshot of a cast vote record, or undefined if none
+ * exists. If undefined, the cast vote record is invalid.
+ */
+export function getOriginalSnapshot(cvr: CVR.CVR): Optional<CVR.CVRSnapshot> {
+  return cvr.CVRSnapshot.find((snapshot) =>
+    snapshot['@id'].includes('-original')
```
eventualbuddha, line 48: Why would we not use the `Type` field? It seems like the `@id` should not be semantically meaningful.

libs/utils/src/cast_vote_records.ts
```diff
@@ -118,6 +129,35 @@ export function convertCastVoteRecordVotesToTabulationVotes(
...
+ */
+export function convertCastVoteRecordMarkMetricsToMarkScores(
+  originalSnapshot: CVR.CVRSnapshot
+): Tabulation.MarkScores {
+  const markScores: Tabulation.MarkScores = {};
+  for (const cvrContest of originalSnapshot.CVRContest) {
+    const contestMarkScores: Record<ContestOptionId, Tabulation.MarkScore> = {};
+    for (const cvrContestSelection of cvrContest.CVRContestSelection) {
+      // We assume every contest selection has only one selection position
+      assert(cvrContestSelection.SelectionPosition.length === 1);
```
eventualbuddha, line 145: Add a message so the error in this case is useful?

libs/utils/src/cast_vote_records.ts
```diff
@@ -118,6 +129,35 @@ export function convertCastVoteRecordVotesToTabulationVotes(
...
+      const selectionPosition = cvrContestSelection.SelectionPosition[0];
+      assert(selectionPosition);
+      const markMetric = safeParseNumber(selectionPosition.MarkMetricValue);
+      /* istanbul ignore next - @preserve */
+      if (markMetric.isErr()) {
+        throw new Error('unable to interpret mark metric value as number');
+      }
+      contestMarkScores[cvrContestSelection.ContestSelectionId] =
```
eventualbuddha, line 153: You could simplify this a bit.
```suggestion
      const markMetric = safeParseNumber(selectionPosition.MarkMetricValue).expect('unable to interpret mark metric value as number');
      contestMarkScores[cvrContestSelection.ContestSelectionId] = markMetric;
```

libs/backend/src/cast_vote_records/import.test.ts
```diff
@@ -226,6 +226,20 @@ test.each<{
     },
     expectedErrorSubType: 'no-current-snapshot',
   },
+  {
+    description: 'no original snapshot',
+    modifications: {
+      castVoteRecordModifier: (castVoteRecord) => ({
+        ...castVoteRecord,
+        CVRSnapshot: castVoteRecord.CVRSnapshot.filter(
+          (ss) => !ss['@id'].includes('original')
```
eventualbuddha, line 235: Again, use the `Type` field?

libs/fixture-generators/src/generate-cvrs/generate_cvrs.ts
```diff
@@ -406,6 +424,57 @@ export function* generateCvrs({
...
+                                  OptionPosition: getOptionPosition({
+                                    contest,
+                                    optionId,
+                                  }),
+                                  SelectionPosition: [
+                                    {
+                                      '@type': 'CVR.SelectionPosition',
+                                      NumberVotes: 1,
+                                      MarkMetricValue: hasIndication
+                                        ? ['0.75']
```
eventualbuddha, line 464: Any reason not to just use `1.0`?

eventualbuddha (approved):

libs/utils/src/cast_vote_records.ts
nikhilb4a: Hm right, definitely should.

Issue #6164, opened 2025-03-26
VxPollBook: Search Resilient to Apostrophes, Hyphens, Spaces


adghayes:
There are lots of slight differences between folks name on their driver's license and in the SVRS database. For example, O'Brien vs. OBRIEN. We should ignore spaces and apostrophes and hyphens.

PR #6417, opened 2025-05-11
[pollbook] handle punctuation and whitespace in search

kshen0:
## Overview
https://github.com/votingworks/vxsuite/issues/6164

Ignore case, apostrophes, dashes, and whitespace in voter name search.

This is `o(n)`. I haven't noticed any slowdown on the test pollbook package with ~6000 voters but we'll most likely want to replace the regexp search with computing and storing a normalized name column in the future. I didn't do it in this PR because

1. This was much faster to write
2. I'm not sure how much is still in flux with pollbook package reading, voter input files, and so on

## Demo Video or Screenshot

https://github.com/user-attachments/assets/3e6b3668-1014-4dbd-9c70-a49e128fa97f

https://github.com/user-attachments/assets/ca0d6568-17fd-4e45-a72d-c141488d1235

https://github.com/user-attachments/assets/10835a69-9ca8-45ca-aa73-592d2203a6cf

## Testing Plan

- Pollbook backend tests
- libs/db tests

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/app.test.ts
+++ b/apps/pollbook/backend/src/app.test.ts
@@ -796,12 +796,30 @@ test('voter search ignores punctuation', async () => {
     }
 
     const testSearchParams = [
+      // Test puncutation and whitespace in input are ignored
+      {
+        firstName: 'george-washington',
+        lastName: 'carver-farmer',
+      },
+      {
+        firstName: "george'washington",
+        lastName: "carver'farmer",
+      },
+      {
+        firstName: 'mar tha',
+        lastName: 'wash ington',
+      },
+      // Test punctuation and whitespace in db column are ignored
       {
         firstName: 'georgewashington',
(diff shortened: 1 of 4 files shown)


carolinemodic (approved):
There is a pollbook-package-large.zip in libs/fixtures for electionFamousNames if you want to test with a larger data set, that is what I have been considering the upper bound of what we need to support

kshen0:
> There is a pollbook-package-large.zip in libs/fixtures for electionFamousNames if you want to test with a larger data set, that is what I have been considering the upper bound of what we need to support

Thanks. Performance seems pretty good (sub-75ms) with the larger set on my VM. The very small sample pollbook package (~low hundreds of voters) is about 10-12ms per search. I'm guessing real hardware will be slightly less performant but 50-75ms gives us good headroom.

Screenshot 2025-05-12 at 3 19 30 PM

Issue #6201, opened 2025-03-28
VxAdmin: Ballot readiness report tweaks


mattroe:

mattroe:
Per [discussion with SLI](https://votingworks.slack.com/archives/C088THUKP2T/p1741714815244049), we just need to bring the readiness report attributes currently produced by VxDesign (J.1-3) into VxAdmin readiness reports and document how J.4 is met procedurally. This is unblocked though.

arsalansufi:
Here are J.1 through J.4 for reference:

1.1.2-J.1 |   | the allowable number of votes in each contest
1.1.2-J.2 |   | the tabulation method for each contest
1.1.2-J.3 |   | the inclusion or exclusion of contests as the result of precinct splits
1.1.2-J.4 |   | samples of all final ballot styles

We only need to bring J.1 through J.3 into the VxAdmin report, i.e., exactly what the VxDesign readiness report includes. J.4 (the PDFs) will be covered via a procedural solution and documentation updates.

PR #6421, opened 2025-05-12
[Admin] Move ballot style readiness report from Design to Admin

kofi-q:
## Overview

Resolves https://github.com/votingworks/vxsuite/issues/6201

Addressing a functional discrepancy from cert testing WRT ballot style readiness report requirements and moving the detailed report from VxAdmin to replace the summary in VxAdmin.

### TODO
- When contests in different districts have the same title, there's no disambiguation between them in the report and they might appear to be duplicates. We could update the format to group contests by district or add district names to each contest, if this is likely to be a common case.
- We recently updated how ballot style IDs are displayed in other parts of the product - unsure, but assuming this report will also need to get updated to match.

## Demo Video or Screenshot
See test snapshots

## Testing Plan
- Manual + automated

## Checklist


libs/ui/src/diagnostics/configuration_section.tsx
```diff
@@ -18,6 +18,7 @@ export interface ConfigurationSectionProps {
   electionPackageHash?: string;
   expectPrecinctSelection?: boolean;
   markThresholds?: MarkThresholds;
+  omitConfigSectionBallotStyles?: boolean;
```
jonahkagan, line 21: Would it be simpler if we combine the ballot styles info into one section that's used across all machines? I.e. remove the ballot style info from the config section and add the new ballot styles section to all readiness reports?

jonahkagan (approved):

libs/ui/src/diagnostics/configuration_section.tsx
kofi-q: Ah, I like the idea of having consistency across the apps for this piece. Thinking through whether that would involve filtering down to only the relevant ballot styles for a precinct when rendering on a single-precinct machine - will add this to an open thread about this task and follow up with a change if needed.

Issue #5984, opened 2025-02-06
VxScan: Properly identify all scanner disconnects


arsalansufi:
In other words, land some fully tested version of this change: https://github.com/votingworks/vxsuite/pull/5955

See #issue-v401-vxscan-blip-after-configuration for more context

adghayes:
Related to #6127.

arsalansufi:
Getting logs this week from NH customers who experienced crashes that we believe were due to scanner disconnects. Those logs may inform final tweaks we make here

arsalansufi:
Collecting logs here: https://drive.google.com/drive/folders/1fCbw7T6D2OPFoNjHd-NUk0UEqxCeiHa3?usp=drive_link

PR #6422, opened 2025-05-12
pdi-scanner: Retry connecting on Other error during connect

jonahkagan:
## Overview

Task: https://github.com/votingworks/vxsuite/issues/5984

We saw a few machines that crashed due to an "Other" error during connection. We haven't been able to reproduce, but it seems like it might have been due to some sort of temporary blip. This PR turns that error into a "disconnected" error, which will trigger reconnection logic in the VxScan state machine. I also added a distinct error message so we can see in the logs if this occurs in the future and evaluate whether the reconnection logic helped.
## Demo Video or Screenshot
N/A
## Testing Plan
Couldn't test since I can't repro. I put in a temporary patch to return the error manually just to exercise the code path and make sure it would trigger auto-reconnecting logic.

## Checklist

### DIFF
--- a/libs/pdi-scanner/src/rust/main.rs
+++ b/libs/pdi-scanner/src/rust/main.rs
@@ -283,6 +283,18 @@ fn main() -> color_eyre::Result<()> {
                             }
                             client = Some(c);
                         }
+                        // We've seen this Other error in the logs a few times, and we think it might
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
Nice, this does feel safe to try

Issue #6325, opened 2025-04-22
VxMark - Unable to Save CDF logs - copy-failed


mattroe:
VV40ECT-92 https://docs.google.com/spreadsheets/d/1jyOM_rz1kIplJlWjxd_eEbwwEkmM7bQ6iQthLvuJsvE/edit?gid=639045837#gid=639045837

arsalansufi:
Arsalan to see if we can get logs

PR #6433, opened 2025-05-13
[libs/backend] Add more detail to log export errors

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/6325

We had a report come out of cert testing of a `copy-failed` error during a logs export, but we don't have granular enough logs to figure a root cause.
Adding the original error as additional log metadata to help with that (as well as upcoming attempts to repro on a real machine).

## Testing Plan
- Manual + updated unit tests

### DIFF
--- a/libs/backend/src/system_call/export_logs_to_usb.ts
+++ b/libs/backend/src/system_call/export_logs_to_usb.ts
@@ -2,6 +2,7 @@ import {
   Result,
   deferred,
   err,
+  extractErrorMessage,
   ok,
   throwIllegalValue,
 } from '@votingworks/basics';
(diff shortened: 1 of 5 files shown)


libs/backend/src/system_call/export_logs_to_usb.ts
```diff
@@ -249,11 +250,31 @@ export async function exportLogsToUsb({
...
+    const error = result.err();
+
+    let cause: string | undefined;
+    if ('cause' in error) {
+      if (error.cause instanceof Error) {
+        cause = error.cause.stack || error.cause.toString();
+      } else {
+        cause = `${error.cause}`;
+      }
+    }
```
jonahkagan, line 263: Consider using `extractErrorMessage` from libs/basics

jonahkagan (approved):

libs/backend/src/system_call/export_logs_to_usb.ts
kofi-q: Oh nice, didn't know that existed - thanks! Looks like it prioritizes the message over the stack trace, while I want the reverse here, but can use it for the fallbacks

Issue #5782, opened 2024-12-20
VxAdmin: Contest Tagging / Tracking


adghayes:
core functionality for this project is being able to work through a queue of adjudication issues (overvote, marginal mark, etc.). To be able to filter for these ballots and contests quickly, I think they're going to need to be cataloged and tagged on import in some way. It won't be efficient enough to essentially re-tabulate all CVRs whenever you want to adjudicate them

an implementation idea starting point - on import, the contests on every ballot are analyzed for all possible adjudication reasons:

- overvote
- undervote
- marginal marks
- write-in

for each contest, we create an entry in a needs_adjudication column that has flags for each of those reasons. we can easily search based on issue and contest ID and join to the CVR table for more advanced filtering.

that table could also track the status of the contest as "unresolved" or "resolved." to simplify the project, let's say that if a contest is resolved for one issue, it is resolved for all issues, and set the UX up accordingly. for example, if you started by adjudicating overvotes and you hit a contest with a write-in, the UX gently forces you to resolve the write-in too. this greatly simplifies the resolved vs. unresolved tracking and avoids complexity that emerges when people resolve different issues in different orders.

PR #6445, opened 2025-05-14
Admin: Tag contests for adjudication

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/5782

## Checklist


apps/admin/backend/src/cast_vote_records.ts
```diff
@@ -186,6 +192,62 @@ export async function listCastVoteRecordExportsOnUsbDrive(
   );
 }
 
+function determineCvrContestTags(input: {
```
nikhilb4a, line 195: simplify this function. what parts can i pull out?

apps/admin/backend/src/cast_vote_records.ts
```diff
@@ -186,6 +192,66 @@ export async function listCastVoteRecordExportsOnUsbDrive(
...
+  cvrId: Id;
+  writeIns: CastVoteRecordWriteIn[];
+  markScores?: MarkScores;
+}): CvrContestTag[] {
+  const { store, cvrId, writeIns, markScores } = input;
+  const electionId = assertDefined(store.getCurrentElectionId());
+  const { adminAdjudicationReasons, markThresholds } =
+    store.getSystemSettings(electionId);
+
+  const tagsByContestId = new Map<ContestId, CvrContestTag>();
```
eventualbuddha, line 210: This `Map` and the associated helper function is a pretty leaky abstraction in this function. The callers of the helper function still call `Map::set` explicitly knowing that that's how future calls to `getOrCreateTag` will get their data. You could use a class to encapsulate that data and behavior and update `determineCvrContestTags` to something like this:

```ts
export function determineCvrContestTags(input: {
  store: Store;
  cvrId: Id;
  writeIns: CastVoteRecordWriteIn[];
  markScores?: MarkScores;
}): CvrContestTag[] {
  const { store, cvrId, writeIns, markScores } = input;
  const electionId = assertDefined(store.getCurrentElectionId());
  const { adminAdjudicationReasons, markThresholds } =
    store.getSystemSettings(electionId);

  const tagsByContestId = new CvrContestTagList(cvrId);

  for (const writeIn of writeIns) {
    const tag = tagsByContestId.getOrCreateTag(writeIn.contestId);
    if (writeIn.isUnmarked) {
      tag.hasUnmarkedWriteIn = true;
    } else {
      tag.hasWriteIn = true;
    }
  }

  if (adminAdjudicationReasons.includes(AdjudicationReason.MarginalMark)) {
    assert(
      markScores !== undefined,
      `mark scores expected with 'MarginalMark'
       adjudication reason set in system settings`
    );
    for (const [contestId, contestMarkScores] of Object.entries(markScores)) {
      for (const optionMarkScore of Object.values(contestMarkScores)) {
        const hasMarginalMark =
          optionMarkScore >= markThresholds.marginal &&
          optionMarkScore <= markThresholds.definite;
        if (hasMarginalMark) {
          const tag = tagsByContestId.getOrCreateTag(contestId);
          tag.hasMarginalMark = true;
          break;
        }
      }
    }
  }

  return tagsByContestId.toArray();
}
```

Here's my implementation of that class:

```ts
class CvrContestTagList {
  constructor(private readonly cvrId: Id) {}

  private readonly byContestId = new Map<ContestId, CvrContestTag>();

  getOrCreateTag(contestId: ContestId): CvrContestTag {
    const existingContestTag = this.byContestId.get(contestId);

    if (existingContestTag) {
      return existingContestTag;
    }

    const newContestTag: CvrContestTag = {
      cvrId: this.cvrId,
      contestId,
      isResolved: false,
    };
    this.byContestId.set(contestId, newContestTag);
    return newContestTag;
  }

  toArray(): CvrContestTag[] {
    return Array.from(this.byContestId.values());
  }
}
```

eventualbuddha (approved):

apps/admin/backend/src/cast_vote_records.ts
eventualbuddha: A plain object is the approach we tend to use in vxsuite, but personally I don't like it when the keys are essentially allowed to be arbitrary strings. For example, had this been a plain object and the contest had ID `toString`, we would find a value in a seemingly empty object! But it wouldn't be a `CvrContestTag`, it'd be a `Function`. Granted, other parts of our system would have failed before it ever got to this point, but … I'd rather not introduce _more_ places for it to fail in that manner.

nikhilb4a: Thanks for that example. And I appreciate that conclusion:

 `Granted, other parts of our system would have failed before it ever got to this point, but … I'd rather not introduce more places for it to fail in that manner.`

I think I initially gravitated towards plain objects too for the sake of staying more consistent with the codebase. Also, I think I had a (subconscious) idea that this had been intentional  with this being an open source project, maybe because it would be good to use simpler code that would be easier for the average person to understand. But, you make a good point that we shouldn't introduce more places for the system to fail as we might eventually choose to clean that all up, because I imagine as we grow we will see more people test the limits of our system.

I feel pretty convinced to start taking this more correct and reliable approach, and will start trying to use custom classes/hooks more often (when appropriate)

PR #6452, opened 2025-05-15
VxScan: update "Ballot Not Counted" messages

### LINKED ISSUE #5674: VxScan: Review "Ballot Not Counted" messages
Strong opinion for consideration: All the "ballot not counted" messages should say, in bold letters first, what the problem was. Like "Multiple Sheets Detected" , or "Ballot Unreadable". "Ballot not counted" is honestly optional, but if it must be on the screen, it should be the subtitle.

Issue automatically created from Slack feedback reported by ben: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1730830339725829?thread_ts=1730830339.725829&cid=CJU9MSC6S

adghayes:
## Overview

Closes #5674. Updates the scan error screens to center the specific error message rather than "Ballot Not Counted." In cases where it's important to see "Ballot Not Counted," that is in the error message and the error message is bolded. We discussed on Slack [in this thread](https://votingworks.slack.com/archives/CJU9MSC6S/p1746672879099619?thread_ts=1730830339.725829&cid=CJU9MSC6S).

## Demo Video or Screenshot

See "Final" column [on Figma](https://www.figma.com/board/YDepdsySTMRXrlWeIPFyZc/Updated-VxScan-Error-Messages?node-id=0-1&p=f&t=J0Q9VVz38A3pE0oD-0). Where this is no screenshot, it is the same as "After." The fact that some images have the election metadata bar at the bottom and others do not is just an artifact of how I got the screenshots, no change in behavior there.


apps/scan/frontend/src/screens/scan_error_screen.tsx
```diff
@@ -26,50 +27,117 @@ export function ScanErrorScreen({
...
+  } = (() => {
+    if (
+      restartRequired ||
+      !error ||
+      error === 'paper_status_timed_out' ||
+      error === 'scanning_timed_out' ||
+      error === 'unexpected_paper_status' ||
+      error === 'unexpected_event' ||
+      error === 'client_error'
+    ) {
```
adghayes, line 53: One of the most confusing parts of this screen was that there were a few different paths to the same or similar errors. `restartRequired` indicates a restart is required, which is associated with a serious of restart required error messages, but it was not clear to me whether `restartRequired` could be triggered with a different error or an undefined error. Without knowing, I erred on the side of the refactor not affecting previous behavior.

apps/scan/frontend/src/screens/scan_error_screen.tsx
```diff
@@ -26,50 +27,117 @@ export function ScanErrorScreen({
...
+    error !== 'double_feed_calibration_timed_out' && // Only used in double feed calibration
+      error !== 'scanner_diagnostic_failed' && // Only used in ScannerDiagnosticScreen
+      error !== 'outfeed_blocked' // Only used in ScanJamScreen
+  );
+
+  const {
+    title,
+    errorMessage,
+    pollWorkerMessage,
+  }: {
```
adghayes, line 40: Since more text on the screen is now variable, I refactored to avoid having many different `switch` statements.

apps/scan/frontend/src/screens/scan_error_screen.tsx
```diff
@@ -26,50 +27,117 @@ export function ScanErrorScreen({
...
+      error === 'client_error'
+    ) {
+      return {
+        title: appStrings.titleScannerError(),
+        errorMessage: (
+          <Font weight="bold">
+            {appStrings.instructionsScannerAskForRestart()}
+          </Font>
+        ),
+        pollWorkerMessage: 'none',
```
adghayes, line 61: The restart screen is unique in that it doesn't have a poll worker message, which is instead baked into the error message. An alternative would be, instead of:

"The ballot was not counted. Ask a poll worker to restart the scanner."

It could be:

"The ballot was not counted. Scan it again after the scanner is restarted."
"Please ask a poll worker for help."

Do you think there is any reason for the restart screen to be special like it is?

apps/scan/frontend/src/screens/scan_error_screen.tsx
```diff
@@ -78,19 +146,21 @@ export function ScanErrorScreen({
...
-        {restartRequired ? (
-          <P>{appStrings.instructionsScannerAskForRestart()}</P>
-        ) : (
-          <Caption>{appStrings.noteAskPollWorkerForHelp()}</Caption>
-        )}
+        <Caption>
+          {pollWorkerMessage === 'needs-help'
+            ? appStrings.instructionsAskForHelp()
+            : pollWorkerMessage === 'might-need-help'
+            ? appStrings.noteAskPollWorkerForHelp()
```
adghayes, line 161: Some issues require poll worker intervention. Others don't, but a voter could still ask for help. I wanted to reflect this in the copy. It seems silly for a voter to "ask a poll worker _if_ they need help" if the problem is something like a mismatched ballot hash they have no control over.

apps/scan/frontend/src/screens/scan_error_screen.tsx
```diff
@@ -26,50 +27,117 @@ export function ScanErrorScreen({
...
+      // interpretation errors
+      case 'vertical_streaks_detected':
+        return {
+          title: appStrings.titleScannerNeedsCleaning(),
+          errorMessage: (
+            <Font weight="bold">
+              {appStrings.warningScannerNeedsCleaning()}
+            </Font>
+          ),
+          pollWorkerMessage: 'needs-help',
```
jonahkagan, line 75: What if we just use the actual app string here instead of a key that gets replaced later? It seems a tad more straightforward

apps/scan/frontend/src/screens/scan_error_screen.tsx
jonahkagan: I think it's a little clearer the way it is now. The line about asking a poll worker for help is so small that I doubt anyone ever reads it.

apps/scan/frontend/src/screens/scan_error_screen.tsx
```diff
@@ -78,19 +146,21 @@ export function ScanErrorScreen({
...
       <FullScreenPromptLayout
-        title={appStrings.titleScannerBallotNotCounted()}
+        title={title}
         image={
           <FullScreenIconWrapper>
             <Icons.Delete color="danger" />
           </FullScreenIconWrapper>
         }
       >
         <P>{errorMessage}</P>
```
jonahkagan, line 156: I think we should always bold this line for consistency and readability.

jonahkagan (approved):

apps/scan/frontend/src/screens/scan_error_screen.tsx
adghayes: 3be03c2c450db11023f0ef0070eed9ff7900baa3

Issue #6360, opened 2025-05-05
Unique identifier RFI response


arsalansufi:
Ben's initial spec: https://docs.google.com/document/d/1YV4TcIhfozLycHcvUeGC7ylwidB7xDaabdOMdjsl6Hw/edit?usp=sharing

Other notes: Given our plan to require central scanning of BMD ballots for the initial cert campaign, we shouldn't need to apply the above to BMD ballots, for now

PR #6455, opened 2025-05-15
VxScan: Encrypt ballot audit IDs

jonahkagan:
## Overview

Task: #6360 

When the `precinctScanEnableBallotAuditIds` system setting is turned on, VxScan will:
- Generate a random secret key on configure and store it in the db
- Encrypt ballot audit IDs from scanned ballot metadata and add them to the CVR

This PR also adds a script to decrypt the ballot audit IDs for all of the CVRs in an export directory, copying them into a new directory and renaming each CVR JSON file with the decrypted audit ID for easy reference during an audit.

## Demo Video or Screenshot
Example encrypted ballot audit ID in a CVR
Screenshot 2025-05-15 at 2 51 35 PM

Example decryption script output from a CVR export with ballots with audit IDs 1, 2, and 3
Screenshot 2025-05-15 at 2 51 02 PM

## Testing Plan
- Manual end to end testing
- Added new automated tests

## Checklist

### DIFF
--- a/apps/scan/backend/src/app.ts
+++ b/apps/scan/backend/src/app.ts
@@ -25,6 +25,7 @@ import {
 import { assert, assertDefined, ok, Result } from '@votingworks/basics';
 import {
   InsertedSmartCardAuthApi,
+  generateRandomKey,
   generateSignedHashValidationQrCodeValue,
 } from '@votingworks/auth';
 import { UsbDrive, UsbDriveStatus } from '@votingworks/usb-drive';
@@ -64,6 +65,8 @@ import {
   testPrintFailureDiagnosticMessage,
 } from './util/diagnostics';
 import { saveReadinessReport } from './printing/readiness_report';
+
+const BALLOT_AUDIT_ID_KEY_LENGTH = 16;
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export function buildApi({
@@ -170,7 +173,7 @@ export function buildApi({
         );
       }
 
-      store.withTransaction(() => {
+      await store.withTransaction(async () => {
         store.setElectionAndJurisdiction({
           electionData: electionDefinition.electionData,
           jurisdiction: authStatus.user.jurisdiction,
@@ -180,6 +183,11 @@ export function buildApi({
           store.setPrecinctSelection(precinctSelection);
         }
         store.setSystemSettings(systemSettings);
+        if (systemSettings.precinctScanEnableBallotAuditIds) {
+          store.setBallotAuditIdSecretKey(
+            await generateRandomKey(BALLOT_AUDIT_ID_KEY_LENGTH)
+          );
+        }
 
         configureUiStrings({
           electionPackage,
--- a/apps/scan/backend/src/app.ts
+++ b/apps/scan/backend/src/app.ts
@@ -25,7 +25,7 @@ import {
 import { assert, assertDefined, ok, Result } from '@votingworks/basics';
 import {
   InsertedSmartCardAuthApi,
-  generateRandomKey,
+  generateRandomAes256Key,
   generateSignedHashValidationQrCodeValue,
 } from '@votingworks/auth';
 import { UsbDrive, UsbDriveStatus } from '@votingworks/usb-drive';
@@ -65,8 +65,6 @@ import {
   testPrintFailureDiagnosticMessage,
 } from './util/diagnostics';
 import { saveReadinessReport } from './printing/readiness_report';
-
-const BALLOT_AUDIT_ID_KEY_LENGTH = 16;
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export function buildApi({
@@ -184,9 +182,7 @@ export function buildApi({
         }
         store.setSystemSettings(systemSettings);
         if (systemSettings.precinctScanEnableBallotAuditIds) {
-          store.setBallotAuditIdSecretKey(
-            await generateRandomKey(BALLOT_AUDIT_ID_KEY_LENGTH)
-          );
+          store.setBallotAuditIdSecretKey(await generateRandomAes256Key());
         }
 
         configureUiStrings({
--- a/apps/scan/backend/src/export.ts
+++ b/apps/scan/backend/src/export.ts
@@ -1,13 +1,63 @@
 import { LogEventId, Logger } from '@votingworks/logging';
-import { ExportCastVoteRecordsToUsbDriveError } from '@votingworks/types';
 import {
+  ExportCastVoteRecordsToUsbDriveError,
+  SystemSettings,
+} from '@votingworks/types';
+import {
+  assert,
+  assertDefined,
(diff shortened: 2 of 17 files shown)


jonahkagan:
Tagged @kofi-q for main review, plus @benadida for crypto review

libs/auth/scripts/src/decrypt_cvr_ballot_audit_ids.ts
```diff
@@ -0,0 +1,87 @@
...
+      path.join(outputDirectory, `${decryptedBallotAuditId}.json`),
+      JSON.stringify(cvrReport, null, 2),
+      'utf-8'
+    );
+  }
+}
+
+/**
+ * A script for checking a Java Card PIN
+ */
```
arsalansufi, line 76: Was just quickly skimming the libs/auth changes and noticed this copy-pasted comment - but the libs/auth changes otherwise look good!

jonahkagan: Thanks for catching that

apps/scan/backend/src/app.ts
```diff
@@ -65,6 +66,8 @@ import {
 } from './util/diagnostics';
 import { saveReadinessReport } from './printing/readiness_report';
 
+const BALLOT_AUDIT_ID_KEY_LENGTH = 16;
```
benadida, line 69: This is the secret key, which should be 256 bits, so 32 bytes.

libs/auth/src/cryptography.ts
```diff
@@ -550,3 +550,52 @@ export async function signMessage({
     : [scriptPath, scriptInput];
   return runCommand(command, { stdin: message });
 }
+
+/**
+ * Generates a random key of the specified length using OpenSSL. The key is
+ * returned in base64 encoding.
+ */
+export async function generateRandomKey(length: number): Promise<string> {
```
benadida, line 558: suggest renaming this `generateRandomAes256Key` and bundling 256 into the function. Separating the logic of key length from algorithm is not useful and in the future could lead to badness.

libs/auth/src/cryptography.ts
```diff
@@ -550,3 +550,52 @@ export async function signMessage({
...
+  return (await openssl(['rand', String(length)])).toString('base64');
+}
+
+/**
+ * Encrypts the given utf8-encoded input string using AES-256-CBC with the
+ * specified key, returning the encrypted data in base64 encoding.
+ */
+export async function encryptAes256(
```
benadida, line 566: We should do this a little bit differently which is going to sound awfully constraining, but that's on purpose to make sure we don't misuse this in the future and that we use this properly right now.

Let's constrain these functions to be *only* about encrypting smallish integers. So the function should be `encryptSmallInteger`, and we should hardwire that the integer must be no more than 64 bits –– technically the approach we'll take will allow for 127 bits, but I don't want to ever reuse this approach for anything that isn't a reasonable integer.

We should then assert that the input is an integer no more than 64 bits long.

Once we do that, the internal encryption should be AES in ECB mode. We should add a comment that explains that we safely use AES in ECB mode because we are only ever encrypting a single block (128 bits or less). PKCS7 padding will automatically be applied, and since we're less than a single block, the output will only ever be a single block. That's what we want. We also want an input, if encrypted twice, to yield the same output. Thus ECB mode.

This approach should never be used to encrypt anything other than small integers once with the goal of ciphertext equality when the plaintexts are the same.

benadida (changes requested):
happy to talk through these cryptography changes which are not intuitively obvious.

libs/auth/src/cryptography.ts
jonahkagan: Suggestion implemented

apps/scan/backend/src/app.ts
jonahkagan: Switched to the EBC cipher per other comment, which made this point moot, as it only requires a key, not an IV/salt, so there's no value from using the password method. We also realized that it's nice to not have to rely on openssl's password->key derivation so that anyone can decrypt audit IDs without necessarily using openssl.

libs/auth/src/cryptography.ts
jonahkagan: Updated here: 0b1b30e4e

jonahkagan:
@benadida I implemented the changes you recommended, please re-review.

apps/scan/backend/src/export.ts
```diff
@@ -1,13 +1,63 @@
...
+
+export async function encryptBallotAuditId(
+  store: Store,
+  systemSettings: SystemSettings,
+  sheet: Sheet
+): Promise<Sheet> {
+  if (
+    !(
+      systemSettings.precinctScanEnableBallotAuditIds &&
+      sheet.type === 'accepted' &&
```
kofi-q, line 31: My understanding is murky here, but wondering why we wouldn't encrypt the audit ID for non-`accepted` CVRs - do those get exported to the USB drive as well? And if they do, is there any concern with those being plain text? (Assuming that even if they're not tied to a tallied CVR, they at least have some record of attempted votes)

libs/auth/src/cryptography.ts
```diff
@@ -550,3 +550,58 @@ export async function signMessage({
...
+  const inputBuffer = Buffer.from(input, 'utf-8');
+  assert(inputBuffer.length <= 16, 'Input must be 128 bits or less');
+  return (
+    await openssl([
+      'enc',
+      '-aes-256-ecb',
+      '-K',
+      key,
+      '-in',
+      Buffer.from(input, 'utf-8'),
```
kofi-q, line 582: Might as well reuse the `inputBuffer` from above now

kofi-q (approved):
Assuming the dust has settled on the crypto pieces, no blockers on my end

apps/scan/backend/src/export.ts
jonahkagan: I don't think rejected ballots get audited, so I didn't think there was any reason to include the audit ID. Also, I don't think all of them would have the audit ID in the metadata anyway depending on how far along they got in interpretation.

libs/auth/src/cryptography.ts
```diff
@@ -550,3 +550,44 @@ export async function signMessage({
...
+/**
+ * Decrypts the given base64-encoded input string using AES-256-ECB with the
+ * specified hex-encoded key, returning the decrypted data as a utf8-encoded
+ * string.
+ */
+export async function decryptAes256(
+  key: string,
+  input: string
+): Promise<string> {
+  const inputBuffer = Buffer.from(input, 'base64');
```
benadida, line 589: worth asserting that `inputBuffer` is just one block, so should be 128 bits exactly.

benadida (approved):
one suggested tweak, but good to go.

libs/auth/src/cryptography.ts
jonahkagan: Good call!

PR #6458, opened 2025-05-16
apps/admin: Bump timeout for flaky integration test

nikhilb4a:
## Overview

I've noticed this test timeout somewhat often and just saw it fail on main. When I last worked on it I noticed it silently retrying sometimes increasing ci times, and seems like it's often enough that we should bump it. It's pretty comprehensive (lots of adjudications, reports) so it feels reasonable to give it this longer timeout.

## Checklist


kofi-q (approved):
Thanks for fixing! Got bitten by this flake a few times yesterday 😅

PR #6469, opened 2025-05-19
fix(scan): scan up to 22" ballots in HWTA

### LINKED ISSUE #6462: VxScan-HWTA: temporarily set 22" for all paper to enable skew testing

eventualbuddha:
## Overview

Closes #6462

Currently we're limited to testing ballots that are 11" long. This unblocks testing longer ballots, but introduces the possibility that we double-feed ballots in testing. We can overcome this later by adding a configuration setting for the ballot length.

## Demo Video or Screenshot
n/a

## Testing Plan
- [x] Test with 11" ballot.
- [x] Test with 14" ballot.
- [x] Test with 22" ballot.


arsalansufi (approved):

pieguyatx (approved):
looks good, for these first tests double-sheets shouldn't be an issue if we're only doing one at a time

PR #6475, opened 2025-05-20
fix(scan): stop resizing scanned images

eventualbuddha:
## Overview
Back before we fully settled on 200dpi we had scanned a fair amount at 300dpi, which made it useful to be able to resize incoming images down to 200dpi. This was true of incoming scans, but was especially true of fixtures. Now, however, we _never_ want to resize the images (only crop the black borders). Ideally this would just be dead code, but in some cases the images may be tall enough due to stretching that the resize logic is triggered. This may have negative impacts both on finding timing marks and scoring bubbles, especially on an already-stretched/skewed image. Since this code is not needed–even for testing–we should remove it and avoid the potential interpretation issues and slowdowns resizing images can incur.

## Demo Video or Screenshot
n/a

## Testing Plan
Automated testing.


libs/ballot-interpreter/src/hmpb-rust/ballot_card.rs
```diff
@@ -260,7 +260,8 @@ pub fn get_matching_paper_info_for_image_size(
     possible_paper_info: &[PaperInfo],
     resize_strategy: ResizeStrategy,
 ) -> Option<PaperInfo> {
-    const THRESHOLD: f32 = 0.05;
+    const WIDTH_ERROR_THRESHOLD: f32 = 0.05;
+    const HEIGHT_ERROR_THRESHOLD: f32 = 0.15;
```
jonahkagan, line 263: How big is the gap between paper sizes? Given this threshold, would an image size that's halfway between two sizes be assigned the larger size, the smaller size, or neither? I.e. is there an overlap of the error margins between paper sizes where there's non-deterministic behavior (or behavior that's only determined by the order in which we test paper sizes)

jonahkagan (approved):
Nice simplification!

libs/ballot-interpreter/src/hmpb-rust/ballot_card.rs
eventualbuddha: Good question! A 15% increase in height over 11" is 12.65" and a 15% decrease in height from 14" is 11.9", so they do overlap with this value. That said, I don't recall seeing _compression_ of scanned images along the scan direction, only _stretching_. So maybe a better approach is to allow -5..=15% height difference, rather than looking at the absolute difference. What do you think?

jonahkagan: I like that idea!

eventualbuddha: [1b90bb7](https://github.com/votingworks/vxsuite/pull/6475/commits/1b90bb7e17920dbe895cc7b7d3401d720b7b1596)

PR #6480, opened 2025-05-21
VxAdmin: Add marginal mark adjudication

nikhilb4a:
## Overview

Closes:
https://github.com/votingworks/vxsuite/issues/6413
https://github.com/votingworks/vxsuite/issues/6244 (side effect of decoupling write-ins and ballot images)

This PR:
1. Adds marginal mark adjudication to contest adjudication
2. Decouples ballot images from write-ins, since now we may need an image for a ballot with a marginal mark and no write-in
3. Updates contest adjudication to support yes/no contests, as propositions are a significant source of marginal marks
4. Updates adjudication copy from "Write-ins" to "Adjudication"

## Demo Video or Screenshot

https://github.com/user-attachments/assets/5b1de5ab-7151-4082-8f9e-d892eec409cf

## Testing Plan

Backend, frontend, and integration tests
QA incoming

## Checklist


apps/admin/backend/src/util/adjudication.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Store } from '../store';
+import { BallotImageView } from '../types';
+import { rootDebug } from './debug';
+
+const debug = rootDebug.extend('adjudication');
+
+/**
+ * Determines the side of the ballot a contest is on.
+ */
+export function getContestSideFromLayout(
```
nikhilb4a, line 17: This file was renamed from `.../write-in.ts` to `.../adjudication.ts`

`getContestSideFromLayout` is a new function. `getBallotImageView` is renamed from `getWriteInImageView`, because we now need ballot images for ballots without write-ins (any contest option can have a marginal mark). It is also slightly refactored as we now provide the `contestOption layouts` instead of the `writeInCoordinates` to enable zooming in on any contest option.

apps/admin/backend/src/store.ts
```diff
@@ -2132,6 +2150,7 @@ export class Store {
         status: 'pending',
         isUnmarked: fromSqliteBool(row.isUnmarked),
         isUndetected: fromSqliteBool(row.isUndetected),
+        machineMarkedText: row.machineMarkedText,
```
nikhilb4a, line 2153: Previously this field was provided to the frontend via the `getWriteInImageAndLayout`, which doesn't make as much sense anymore with the ballot images no longer so tightly coupled to write-ins. Regardless, I think it makes more sense for `machineMarkedText` to be provided to the frontend on the `WriteInRecord` object, so it is now included with write-ins in `getWriteInRecords`.

apps/admin/backend/src/store.ts
```diff
@@ -2156,25 +2175,6 @@ export class Store {
     );
   }
 
-  /**
-   * Gets the write-in IDs for a given cast vote record's contest.
-   */
-  getWriteInIds({ cvrId, contestId }: { cvrId: Id; contestId: Id }): Id[] {
```
nikhilb4a, line 2162: This function is no longer used

PR #6481, opened 2025-05-21
pdi-scanner: Switch from rusb to nusb

jonahkagan:
## Overview

Switch the pdi-scanner from rusb/rusb_async to nusb. This enables us to more easily develop in a VM (resolving the longstanding unknown buffer size issues) and allows for safer control flow patterns. In particular, nusb has a nice abstraction for queueing response buffers for requests, enabling us to transfer incoming image data from the scanner more rapidly. It also uses futures, which means instead of manually polling each USB endpoint in series, we can use `tokio::select!` to wait for whichever one completes.

## Demo Video or Screenshot
N/A

## Testing Plan
- Manually tested basic scanning flows
- Disconnected and reconnected the USB cable at various times to ensure we correctly map error codes to a "disconnected" status (on both VM and a Lenovo Thinkpad)

## Checklist


libs/pdi-scanner/src/rust/scanner.rs
```diff
@@ -87,138 +81,122 @@ impl Scanner {
...
-                            transfer_pool
-                                .poll_endpoint(ENDPOINT_OUT, default_timeout)
-                                .unwrap();
-
-                            host_to_scanner_ack_tx.send(id).unwrap();
-                        }
-                        Err(mpsc::TryRecvError::Empty) => {}
-                        Err(e) => {
-                            tracing::error!("Error receiving outgoing packet: {e}");
+                    completion = in_image_data_queue.next_complete() => {
```
jonahkagan, line 120: Note that I reordered the future match patterns so that we check for image data first, then incoming events, then outgoing messages. I figured that might be a little bit safer (e.g. if the scanner rapidly sends the EndScan event after sending the last image data packet).

libs/pdi-scanner/src/rust/main.rs
```diff
@@ -283,18 +283,6 @@ fn main() -> color_eyre::Result<()> {
...
-                        // to Disconnected to see if our auto-reconnecting logic (in the VxScan state
-                        // machine) can handle it. We map it here rather than in send_error_response
-                        // to keep the mitigation scoped to the connect command, where we saw this
-                        // error occur.
-                        Err(Error::Usb(UsbError::Rusb(rusb::Error::Other))) => {
-                            send_response(Response::Error {
-                                code: ErrorCode::Disconnected,
-                                message: Some("Unknown USB error during connect".to_owned()),
-                            })?;
-                        }
```
jonahkagan, line 297: Having to remove this is a bit of a bummer, but I can't think of a way to keep it, since we don't know the underlying error code to look for.

libs/pdi-scanner/src/rust/scanner.rs
```diff
@@ -77,6 +67,10 @@ impl Scanner {
         /// receive image data.
         const ENDPOINT_IN_IMAGE_DATA: u8 = 0x86;
 
+        /// For receving responses from the scanner (other than image data).
```
eventualbuddha, line 70:
```suggestion
        /// For receiving responses from the scanner (other than image data).
```

Issue #150, opened 2021-01-15
eject USB after loading ballot package in BSD


benadida:
Right now it just keeps the USB mounted. Eject like in the other modals.

PR #6498, opened 2025-05-22
Update Receipt Numbering Scheme

carolinemodic:
## Overview
Previously we attempted to track and print a global queue number on each receipt when an event occurred. This obviously is hard to manage and keep in sync in a distributed system and resulted in duplicate numbers that, while fine, resulted in confusion. We decided it is better to move to a per-machine receipt number that is exposed to the user, and to include a cover sheet in the exported backup with the most recently included receipt number for each machine so that users can easily identify which receipts did not make it into the backup. 

This PR: 
- Removes the tracking of the global receipt number
- We were already keeping a incrementing per-machine queue in the "local event id" so renames that and reuses that as the receipt number
- Weaves the receipt number into the check in object so it can be referenced in the undo check in receipt
- Minor tweaks to styling of receipts and exported paper backup 

In testing this PR I noticed an edge case in the backup. We print a "new registrations" sheet for each letter of the alphabet, including registrations we have in the system. If that spills over onto multiple pages we were not filling in the empty rows on the final page, so the final commit in this PR is a small bug fix to do so. We don't expect new registrations to be common in the system so this is unlikely to occur but it seems safe to just always fill out the last page with empty rows. 

## Demo Video or Screenshot
See snapshot images 
The backup has a new cover page so everything gets bumped down by 1 page making it a little hard to review, but the first page is the primary page with changes. Also Pages 16/17 show the new registrations I added in the test (see the last commit for the added empty rows in those pages).

## Testing Plan
Ran Tests, verified new added receiptNumber test checks are accurate, verified updates to snapshots are as expected.


adghayes:
Exciting!

One benefit of this change that I want to emphasize. Previously, with a global receipt number, we had no way to be confident in what was included in the paper backup. If after 100 receipts, one machine left the network and printed receipts 101 - 110 while the rest of the network printed receipts 101 - 150, the backup checklist would say "Receipt #150" which seems to imply that the 101 - 110 on the partitioned machine were included, even though they might not have been. Now, we can confidently determine what was included on the paper backup and what was not.

@carolinemodic thank you for catching the new registration bug. I agree, good to always print full rows on the last page.

kshen0:
> include a cover sheet in the exported backup with the most recently included receipt number for each machine so that users can easily identify which receipts did not make it into the backup.

I don't understand this part. @adghayes maybe you're alluding to this in your comment but why is it normal for receipts to not make it to the backup?

adghayes:
@kshen0 the problem is if the network partitions. Say there are six machines. But three of them are brought across the far end of a huge auditorium. Now instead of one big network of six machines, there are two networks of three machines. The backup from either of the smaller networks will not include the events from the other smaller network that happened after the time of partition. TL;DR network partitions.

kshen0 (approved):

Issue #6484, opened 2025-05-22
VV40ECT-112 VxScan - Standard audio connectors - SW work


arsalansufi:

arsalansufi:
First PR: https://github.com/votingworks/vxsuite/pull/6474

PR #6499, opened 2025-05-22
[VxScan] Mute scan success/fail sound effects when using headphones

kofi-q:
## Overview

Now that we're enabling screen reader audio over headphones on VxScan, the scan success/fail sounds collide with speech audio and are redundant. Skipping those sounds whenever we detect that headphones are plugged in.

### Without Headphones

https://github.com/user-attachments/assets/ae625cfc-6a7f-4ab4-b0b6-d134e7fbe7f4

### With Headphones Detected

https://github.com/user-attachments/assets/ef53b435-0fb3-4bfc-a557-b727fadc3ecc

## Testing Plan
- Manual & unit tests - broke out state-based audio logic to make testing easier

## Checklist

### DIFF
--- a/apps/scan/frontend/src/screens/voter_screen.tsx
+++ b/apps/scan/frontend/src/screens/voter_screen.tsx
@@ -1,9 +1,7 @@
 import { ElectionDefinition, SystemSettings } from '@votingworks/types';
-import { useQueryChangeListener } from '@votingworks/ui';
 import { assert, throwIllegalValue } from '@votingworks/basics';
 import { getScannerStatus } from '../api';
 import { POLLING_INTERVAL_FOR_SCANNER_STATUS_MS } from '../config/globals';
-import { useSound } from '../utils/use_sound';
 import { InsertBallotScreen } from './insert_ballot_screen';
 import { ScanBusyScreen } from './scan_busy_screen';
 import { ScanErrorScreen } from './scan_error_screen';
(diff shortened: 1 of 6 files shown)


apps/scan/frontend/src/screens/voter_screen.tsx
```diff
@@ -31,40 +30,9 @@ export function VoterScreen({
...
-
-        default: {
-          // No sound
-          break;
-        }
-      }
-    },
+  useScanFeedbackAudio({
+    currentState: scannerStatusQuery.data?.state,
+    isSoundMuted,
```
jonahkagan, line 35: I don't quite see the value in dropping the use of `useQueryChangeListener`, since that abstracted away the slightly tricky logic around detecting and reacting to state changes safely

jonahkagan (approved):

apps/scan/frontend/src/screens/voter_screen.tsx
kofi-q: Ah, wasn't my intention just to drop it. The tests were getting a little overly complicated when I started mocking things out for this view component, since it proxies to many other components - so, I broke out the audio feedback logic into a separate file for easier testing. I initially just copied this as is and realised:
- the tests were simpler if I didn't have to do the scanner status query in `useScanFeedbackAudio` and just passed the state value we need from here, since we have it readily available here anyway
- the logic for `has the state changed?` wasn't all that complex in this case and just comparing to a ref avoids the extra layer of a `useEffect`.

Open to follow up and change it to do the API query in the `useScanFeedbackAudio`, so LMK - sort of just landed on this organically, so not tied to this implementation

Issue #6356, opened 2025-05-02
VxAdmin: manual tally, prevent changing values through scrolling


arsalansufi:
Also if you scroll while focused on an input field it will change the number. Which obviously we don’t want

Issue automatically created from Slack feedback reported by jonah: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1746140359188579

Issue #6129, opened 2025-03-21
VxAdmin: manual tally ballot override wonkiness


arsalansufi:
Noticed something wonky with the total ballot count override option during manual tally entry - almost like the override didn’t register when we made it all the way through. We were editing 110 to 115 and it didn’t seem to register when we backspaced the 0 to enter a 5, but did register when we erased everything. Not describing this super well and need to verify this behavior after the fact. But we did get everything entered in eventually!

Issue automatically created from Slack feedback reported by arsalan: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1741753476099329?thread_ts=1741753476.099329&cid=CJU9MSC6S

adghayes:
@arsalansufi to try to reproduce. Either repro or close.

arsalansufi:
Nice, yup I just arrived at the same conclusion, with a simple fix in the works

PR #6502, opened 2025-05-22
VxAdmin: prevent negative and decimal manual tallies + prevent change on scroll

### LINKED ISSUE #6381: VxAdmin: disallow negative manual tallies

adghayes:
## Overview

Closes #6381 and #6356. In testing negative manual tally values, I noticed decimal values are currently also allowed, which this PR also addresses.

The issue of number inputs changing on scroll is just browser behavior. The two options I could find to turn of it off were a) use a text input or b) use event handlers to stop the default behavior. I preferred a) because it seemed a bit cleaner and lent itself to solving the negative number problem.

## Testing Plan

- updated existing tests
- new test for negative and decimal
- there isn't a way to test the scroll thing because it's browser behavior, not react behavior, that does not exist in the test environment

### DIFF
--- a/apps/admin/frontend/src/screens/tally/manual_tallies_form_screen.tsx
+++ b/apps/admin/frontend/src/screens/tally/manual_tallies_form_screen.tsx
@@ -178,7 +178,7 @@ const StyledInput = styled.input.attrs({ type: 'text' })`
 interface TallyInputProps {
   id: string;
   value: number | EmptyValue;
-  onNewValue: (newValue: number | EmptyValue) => void;
+  onChange: (newValue: number | EmptyValue) => void;
   disabled?: boolean;
   autoFocus?: boolean;
   style?: React.CSSProperties;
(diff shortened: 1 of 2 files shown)


apps/admin/frontend/src/screens/tally/manual_tallies_form_screen.tsx
```diff
@@ -167,16 +168,56 @@ const FormCard = styled(Card)`
...
-  ::-webkit-outer-spin-button {
-    appearance: none;
-    margin: 0;
-  }
 `;
 
+interface TallyInputProps {
+  id: string;
+  value: number | EmptyValue;
+  onNewValue: (newValue: number | EmptyValue) => void;
```
jonahkagan, line 181: nit: Can we call this `onChange` since that's the usual terminology?

jonahkagan (approved):

adghayes:
@arsalansufi the change here may affect the behavior that contributed to #6129, invalidating the need for a fix there. I'm not sure, since I don't understand that issue, but keep that in mind if you try to repro on main.

Issue #6247, opened 2025-04-07
VxScan: VxMark ballot scanning for 11" definitions


mattroe:
Lots of potential directions that may not require a code change, e.g., we could require using 11" BMDB paper when using 11" HMPBs and require bumping up to 14" HMPBs if you want to use 13.25" BMDB paper. We could also require BMDBs to be scanned on VxCentralScan.

Slack discussion: https://votingworks.slack.com/archives/C08LY2VPEJ2/p1744060315262699

mattroe:
@arsalansufi we need to figure out how we want to handle this for NH-Fall with a BMD deployment even though we don't want to change the documented protocol for this cert campaign (that will be centrally scanned).

arsalansufi:
Ah yeah good point - added to the (new) v4.0.2 milestone. Let's discuss potential paths during our 1:1 tomorrow

mattroe:
We may need to put this behind a system setting and block normal scanning of VxMark ballots temporarily depending on the outcome of: https://votingworks.slack.com/archives/CE7LM9LQ2/p1747676839489319?thread_ts=1747676839.489319&cid=CE7LM9LQ2

mattroe:
We will need an estimate of the impact on max ballot content for 11" when we document the system limit. This dovetails with what @kofi-q is working on re: QR code design adjustments for the other VSAP issues.

arsalansufi:
Code change merged ✅ We can tackle system limit updating once other BMDB template changes land (tracked in their own issues)

PR #6514, opened 2025-05-28
Print BMDBs on 11" paper (instead of 13.25") when HMPBs are on letter 

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/6247

VxScan supplies a maximum paper length param to the PDI scanner given the length of HMPBs to prevent a potential ballot accounting edge case where two sheets are scanned back to back so close together that they're treated as one continuous sheet and the first is deposited into the ballot box without being tabulated.

We were always wary of the fact that BMDBs, 13.25" in length, were successfully scanning on VxScan when configured for 11" HMPBs. But this worked during UAT and QA last year, so we moved forward. We only recently discovered that all of our original 13.25" paper stock was just ever so slightly out of spec and shorter than 13.25". With true 13.25" paper, we are in fact seeing some rejections.

The simplest fix for now is to print BMDBs on 11" paper when HMPBs are on letter. VxMark (i.e. the VSAP 150) supports 8" by 11" paper. We will of course have to notify customers of this limitation, but that's fine. Many customers already expect constraints like this around the relationship between HMPB and BMDB size.

Note that I haven't touched the logic for compacting content when it can't fit on a single page. That logic should still work as it tries compacting two steps beyond the starting layout. The starting layout is chosen given the number of contests on the ballot, so technically, we should update that logic to account for 11" vs. 13.25". But the two compacting steps should still cover most cases, and we're already considering layout adjustments for other cert issues so I didn't want to preemptively calculate new baselines.

## Demo Video or Screenshot

<table>
<tr><td>11"</td><td>13.25"</td></tr>
<tr>
<td><kbd>
app-test-ts-printing-bmd-bs-when-hmp-bs-are-on-letter-1-snap
</kbd></td>
<td><kbd>
app-test-ts-printing-bmd-bs-when-hmp-bs-are-on-legal-1-snap
</kbd></td>
</tr>
</table>

## Testing Plan

- [x] Updated automated tests

We've also ordered 8" by 11" paper for upcoming QA.

## Checklist


apps/mark-scan/backend/src/util/render_ballot.tsx
```diff
@@ -33,16 +35,22 @@ export interface RenderBallotProps {
 
 const MACHINE_TYPE: MachineType = 'markScan';
 
-function getPaperDimensions(): PaperDimensions {
-  /* istanbul ignore next - hardware support in flux - @preserve */
-  return getMarkScanBmdModel() === 'bmd-150'
-    ? PAPER_DIMENSIONS['Bmd150']
-    : PAPER_DIMENSIONS['Letter'];
+function getPaperDimensions(election: Election): PaperDimensions {
+  if (getMarkScanBmdModel() === 'bmd-150') {
```
arsalansufi, line 39: I briefly considered removing all the old VSAP 155 logic, but that quickly ballooned so I punted

libs/printing/src/render.tsx
```diff
@@ -28,7 +28,8 @@ export interface PaperDimensions {
 export const PAPER_DIMENSIONS = {
   // Width of exactly 8in results in 1-3 dots of overflow. The overflowing dots print on a line of
   // their own, followed by a mostly blank line. This causes stripes in the printed page.
-  Bmd150: { width: 7.975, height: 13.25 },
+  Custom8x13pt25: { width: 7.975, height: 13.25 },
+  Custom8x11: { width: 7.975, height: 11 },
```
arsalansufi, line 32: Just reemphasizing this point in the PR description:

> Note that I haven't touched the logic for compacting content when it can't fit on a single page. That logic should still work as it tries compacting two steps beyond the starting layout. The starting layout is chosen given the number of contests on the ballot, so technically, we should update that logic to account for 11" vs. 13.25". But the two compacting steps should still cover most cases, and we're already considering layout adjustments for other cert issues so I didn't want to preemptively calculate new baselines.

apps/mark-scan/backend/src/app.test.ts
```diff
@@ -546,96 +548,116 @@ function expectVotesEqual(expected: VotesDict, actual: VotesDict) {
...
-  await expectElectionState({ ballotsPrintedCount: 1 });
-  const pdfData = printBallotSpy.mock.calls[0][0];
-  await expect(pdfData).toMatchPdfSnapshot({ failureThreshold: 0.027 });
-
-  await waitForStatus('presenting_ballot');
-  const interpretation = await apiClient.getInterpretation();
-  assert(interpretation);
-  assert(interpretation.type === 'InterpretedBmdPage');
-  expectVotesEqual(interpretation.votes, mockVotes);
+test.each([
```
arsalansufi, line 551: Review this diff using https://github.com/votingworks/vxsuite/pull/6514/commits/0d69f55afed87f03cc3bbe95ccf931c1f0627973?w=1

kofi-q (approved):
🚢

PR #6516, opened 2025-05-29
fix(types-rs): ignore zero-sized rects in `Rect::union`

eventualbuddha:
## Overview
Previously, `Rect::zero` used with `foldl` to determine an extent of `Rect` instances would always start at the origin. Now it properly starts at the minimum x and y values present in all the non-zero-sized `Rect`s.

## Demo Video or Screenshot
| Before | After |
|-|-|
|CleanShot 2025-05-28 at 20 13 49@2x|image|

## Testing Plan
Tested manually for its effect on debug images. Added a unit test.

### DIFF
--- a/libs/types-rs/src/geometry.rs
+++ b/libs/types-rs/src/geometry.rs
@@ -343,6 +343,14 @@ impl Rect {
 
     // Returns the smallest rectangle that contains both `self` and `other`.
     pub fn union(&self, other: &Self) -> Self {
+        if self.width == 0 || self.height == 0 {
+            return *other;
+        }
+
+        if other.width == 0 || other.height == 0 {
+            return *self;
+        }
+
         let left = self.left.min(other.left);
         let top = self.top.min(other.top);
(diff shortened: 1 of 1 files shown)


jonahkagan:
What should I be seeing in the before/after images? I can't tell the difference

eventualbuddha:
> What should I be seeing in the before/after images? I can't tell the difference

I'm not sure why they ended up different scale factors, but the important part is that the top part of the image no longer has line segments going all the way to `y = 0`, since they end at the bounding box enclosing all the timing marks.

before-after

jonahkagan:
Ah gotcha, thanks!

eventualbuddha:
> Are there any other lurking zero rect issues? It almost seems a little weird to have zero rects representable at all, though I'm sure there's some reason we have them.

I think it was mostly for the `foldl` mentioned above. I think it's okay to merge this, but we can follow up with a change to the approach that makes zero-size rects unrepresentable if you think that's worth it.

Issue #5294, opened 2024-08-22
VxScan: PDI - use grayscale images


arsalansufi:
We currently pull binary images off of the PDI scanner which are much smaller than grayscale images. This was originally to transfer less data over USB to avoid `libusb` issues. We _may_ have resolved that, so perhaps we could just do grayscale. The grayscale images are preferable for adjudication. We'd have to see if we take a scan time performance hit.

adghayes:
@eventualbuddha not prioritizing or scheduling this right this moment, but assigning it to you (rather than @nikhilb4a) so that he doesn't tackle it as he works through issues in the adjudication project.

arsalansufi:
Some relevant product discussion in https://votingworks.slack.com/archives/CEL6D3GAD/p1747846995729299

PR #6522, opened 2025-05-29
ballot-interpreter: Save binarized ballot images

jonahkagan:
## Overview

Task: #5294 

Now that we're switching the PDI scanner from bitonal to grayscale scanning, the ballot images it produces are much larger. In order to preserve the amount of disk space used, we want to binarize them before saving. For reference, the binarized images are ~100KB, while the grayscale images are ~4MB.

Previously, the interpreter passed through whatever ballot images it was given after cropping and normalizing their orientation. This PR changes it to also binarize the image. This will also have the side-effect of binarizing images in VxCentralScan, which will be good for consistency.

## Demo Video or Screenshot
Example binarized image
f18c495f-05e5-41d7-858e-dfe1d868d615-front

## Testing Plan
- Manual testing
- Added a basic unit test
- Ran benchmarks and saw no perf regression
## Checklist

### DIFF
--- a/libs/ballot-interpreter/src/hmpb-rust/interpret.rs
+++ b/libs/ballot-interpreter/src/hmpb-rust/interpret.rs
@@ -831,6 +831,13 @@ mod test {
         }
     }
 
+    fn is_binary_image(image: &GrayImage) -> bool {
+        image
+            .as_raw()
+            .iter()
+            .all(|&pixel| pixel == 0 || pixel == 255)
+    }
+
     #[test]
     fn test_par_map_pair() {
         assert_eq!(par_map_pair(1, 2, |n| n * 2), (2, 4));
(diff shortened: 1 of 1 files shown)


eventualbuddha (approved):

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -688,21 +688,24 @@ pub fn ballot_card(
         })
         .unwrap_or_default();
 
+    let normalized_front_image = imageproc::contrast::threshold(&front_image, front_threshold);
+    let normalized_back_image = imageproc::contrast::threshold(&back_image, back_threshold);
+
```
eventualbuddha, line 693: FYI, I think we should do this in place instead of creating a clone ([PR](https://github.com/votingworks/vxsuite/pull/6523)).

PR #6538, opened 2025-05-31
fix(scan): refine timing mark rects to improve grid finding

eventualbuddha:
There's more to do on this branch, so for now it's a draft PR. High level, I'm trying to better identify timing marks where there's some debris or pen mark or misprint or whatever that makes the timing marks appear larger than they are. This works by essentially [template matching](https://docs.opencv.org/4.x/d4/dc6/tutorial_py_template_matching.html) within the contour bounds for the best matching timing mark spot. If it looks good enough, we keep it.

Things to consider:
- I added code to ensure each border shares the corners with its neighbors (i.e. the set of marks on the right border and the set of marks on the top border both agree on what the top-right mark is). This caused test failures because somehow there were two _nearly_ equal corner candidate marks. I suspect this is due to the timing mark refinement giving slightly different results depending on starting conditions, and maybe there were two overlapping candidate marks that got refined down to almost the same candidate mark?
- I haven't benchmarked this extensively, and it will cost some amount of time. It'd be good to understand how much.
- Part of the motivation for this is to make finding timing marks good enough that we can remove timing mark inference. Even if we never do that, I still think this is a good change.


eventualbuddha:
Superseded by https://github.com/votingworks/vxsuite/pull/6582

PR #6549, opened 2025-06-03
Fix Flaky PB Networking Test

carolinemodic:
## Overview
When we're relying on the networking polling interval on multiple machines to trigger it is somewhat tricky in tests to make sure we're waiting enough time for the condition we're looking for. I tried various timeout settings, using shouldAdvanceTime as true, etc. but the only thing that seemed to make tests reliably pass was adding an increased timeout on all the waitFor calls, so this sets up that pattern in app_multi_node, its not possible to set this with a global config. 

## Demo Video or Screenshot
n/a
## Testing Plan
Used modified CI that runs only this job 25 times, ran that 3x total and saw no failures.

### DIFF
--- a/apps/pollbook/backend/src/app_multi_node.test.ts
+++ b/apps/pollbook/backend/src/app_multi_node.test.ts
@@ -540,7 +540,7 @@ test('connection status is managed properly with many pollbooks', async () => {
 });
 
 // TODO #6544 Fix Flakes and re-enable tests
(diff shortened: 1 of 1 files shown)


kshen0 (approved):
I had similar issues testing state in mark-scan :/
Pollbook tests are still under 5m but if we find many tests rely on 3s time out we might want to consider doing another spike on this

PR #6571, opened 2025-06-05
VxAdmin: Improve SearchSelect behavior when opening upwards

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/6569

The search select hasn't respected the `maxMenuHeight` and it looks like it's because it needs a number representing pixels. 

## Demo Video or Screenshot

New -

Screenshot 2025-06-04 at 3 18 28 PM

Old -

Screenshot 2025-06-04 at 3 18 06 PM

## Testing Plan

Manual and upcoming QA

## Checklist

### DIFF
--- a/libs/ui/src/search_select.tsx
+++ b/libs/ui/src/search_select.tsx
@@ -78,6 +78,7 @@ interface SearchSelectBaseProps<T = string> {
   onInputChange?: (value?: T) => void;
   menuPortalTarget?: HTMLElement;
   minMenuHeight?: number;
+  maxMenuHeight?: number;
   noOptionsMessage?: () => React.ReactNode;
 }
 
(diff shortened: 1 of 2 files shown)


libs/ui/src/search_select.tsx
```diff
@@ -158,10 +160,10 @@ export function SearchSelect<T = string>({
       unstyled
       components={{ DropdownIndicator, MultiValueRemove }}
       className="search-select"
-      maxMenuHeight="50vh"
```
jonahkagan, line 161: Good catch figuring out that this wasn't working. I'm a bit concerned that making it an optional prop means that all of the other places that SearchSelect is used will be vulnerable to overflow (though apparently they already are). Maybe we set a safe default?

Issue #6206, opened 2025-03-28
VxPollBook: Feature to Mark Voter as Dead


adghayes:
Needs scoping.

PR #6574, opened 2025-06-05
Add flow to mark voters as inactive

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/6206

Adds the ability for election managers to mark a voter as inactive. There is no way to undo this action. Once voter who is checked in can not be marked as inactive and an inactive voter can not be checked in. If we end up syncing a check-in event for a voter that has been deactivated (which could happen if the machine where the check-in occured was offline for example) then the check-in will be ignored and not processed once it is synce'd with the mark inactive event. 

A voter who is inactive can not from the user interface have their name or address changed after being inactive. These events can happen and will persist if they occur before the mark inactive event. If a event gets synced with a time after the mark inactive event (which could happen in an offline machine scenario) it will still be processed and applied. 

Inactive voters still appear in EM search but not in PW search. 

Added a new error condition when checking in in case the voter was marked inactive and that event synced just before the check in goes through. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/52960e1c-2e7a-45bf-bd4a-c688f6b5a40a

## Testing Plan
Manual testing of major flows, added tests to store_multi_node for the edge conditions with data syncing. 

## Checklist


adghayes:
LGTM!

kshen0 (approved):

PR #6575, opened 2025-06-05
[VxScan] Add server-side audio player for system sounds

kofi-q:
## Overview

Adding a server-side audio player for the `success/warning/error/alarm` system sounds. For now, just shells out to PulseAudio's `paplay` utility, called directly in dev and called via an app script in prod, to enable running under the `vx-ui` user (which owns the connection to PulseAudio server). App script was added in https://github.com/votingworks/vxsuite-complete-system/pull/448

This isn't used anywhere yet, but plan is to wire it up at startup by:
- Querying for available devices with [`getAudioInfo()`](https://github.com/votingworks/vxsuite/blob/0aaeaec/libs/backend/src/system_call/get_audio_info.ts#L34)
- Initializing the player with the name of the builtin audio output
- Adding the player to the server `Workspace` context to get passed to the API/state machine for use there.

## Testing Plan
- Unit tests for split dev/prod logic and execFile integration
- Manual for playback tests 

## Checklist

  - Pending addition of specific log event IDs for the new action


nikhilb4a (approved):
🥇

PR #6580, opened 2025-06-06
[VxScan] Add preliminary support for controller/keyboard navigation

kofi-q:
## Overview

Copying over the [keyboard navigation handlers](https://github.com/votingworks/vxsuite/blob/main/apps/mark-scan/frontend/src/lib/assistive_technology.ts) from VxMark to VxScan (leaving out the PAT navigation keys, which aren't used on VxScan).

Binding the handlers in the main app and the preview app (for dev testing).

### TODO
- Handling for the volume button on the controller will happen separately in a later PR.
- This change enables triggering action buttons with the left/right keys, but haven't wired up any of the action buttons yet - will do that, wherever it makes sense, in a follow-up.
- Need to verify that the tactile controller emits keyboard events in the same way when running on a prod image. This is the same controller we use on the v2 VxMark, which has an additional set of handlers for browser gamepad events, but it's unclear if that code path is used in prod.
- Also noticed in the preview app that closing a modal doesn't remove the `aria-hidden` attribute from the main view like it should, causing the button elements to then be hidden from the navigation handlers. Need to look into whether this is just a dev thing or if we need a workaround to fix (we've seen this issue in unit tests as well).

## Demo Video or Screenshot

https://github.com/user-attachments/assets/ff969070-1a53-456d-aafc-b3ebac1c77e3

## Testing Plan
- Unit tests + manual verification

## Checklist

### DIFF
--- a/apps/scan/frontend/src/app.tsx
+++ b/apps/scan/frontend/src/app.tsx
@@ -3,6 +3,7 @@ import { BrowserRouter, Route } from 'react-router-dom';
 import { BaseLogger, LogSource } from '@votingworks/logging';
 import { QueryClient } from '@tanstack/react-query';
 import { AppErrorBoundary } from '@votingworks/ui';
+import React from 'react';
 import { AppRoot } from './app_root';
 import { ApiClient, createApiClient, createQueryClient } from './api';
 import { ScanAppBase } from './scan_app_base';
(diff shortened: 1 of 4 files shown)


nikhilb4a (approved):
⌨️ ✅

Issue #6492, opened 2025-05-22
VV40ECT-99 UnderVote appeared in VxScan - SW work


arsalansufi:
Cert discrepancy parent issue: https://github.com/votingworks/cert-discrepancies/issues/113

This is the issue from the con-ops test where a ballot was stretched/skewed and a mark was missed because of timing mark grid misalignment

Issue #6491, opened 2025-05-22
VV40ECT-98 OverVotes Triggered on Scanners - SW work


arsalansufi:
Cert discrepancy parent issue: https://github.com/votingworks/cert-discrepancies/issues/112

This is the issue from the con-ops test where ballots with imprinted IDs over timing marks were rescanned and misinterpreted as having overvotes when they didn't. Completely acceptable for the SW fix to just reject these ballots.

arsalansufi:
Marking this issue done though I know we're still testing at scale. Will open new issues for any follow-ups!

PR #6582, opened 2025-06-06
feat(scan): add a new timing mark finding algorithm 

eventualbuddha:
## Overview

Refs https://github.com/votingworks/vxsuite/issues/6491, https://github.com/votingworks/vxsuite/issues/6492

This algorithm is intended to replace the existing `contours` algorithm, but I'm adding it alongside it for now to allow for a period of testing before switching over.

The new algorithm is named `corners` and works by:
- for each ballot page edge, scanning vertically by pixel column for slice of black that are about timing-mark height
- for each ballot page edge, stitching adjacent slices together into shapes and filtering out ones too far from timing mark size and scoring the shapes as candidate timing marks
- for the left and right ballot page edges, finding candidate timing marks closest to the image corners that have the expected marks in the same row and column indicating that the mark is a corner
- for the left and right ballot page edges, find a set of marks at roughly the expected distances between the top and bottom corners
- for the top and bottom ballot page edges, find a set of marks at roughly the expected distances between the same corners used for the left and right edges

## Demo Video or Screenshot

**Timing mark shapes found by joining vertical slices**
CleanShot 2025-06-06 at 11 48 52@2x

**Corner set of three timing marks**
CleanShot 2025-06-06 at 12 50 38@2x

**Complete timing marks identified**
image

## Testing Plan
- [x] Tested with the existing automated tests plus one more with imprinting over the timing marks.
- [x] Tested with the TRR corpus of VxDesign ballots (details below).

### TRR Corpus

||**`contours` algorithm (new)**|**`corners` algorithm (old)**|
|-|-|-|
|**Total images/pages**|2,472|2,472|
|**Images/pages with defaced timing marks**|34|34|
|**Failed images/pages**|31 defaced, 2 heavily skewed (6.7/500)|27 defaced, 6 heavily skewed (6.7/500)|
|**Succeeded images/pages**|2,439|2,439|

Both the old and the new have the same success rate, but they succeed and fail on different images. Note that these are _not_ fully interpreted; they just had their timing marks detected, despite some of them having the QR code defaced as well.

|**Image/Page**|`contours`|`corners`|
|-|-|-|
|image|✅|❌|
|image|✅|❌|
|image|✅|❌|
|image|✅|❌|
|image|✅|❌|
|image|❌|✅|
|image|❌|✅|
|image|❌|✅|
|image|❌|✅|
|image|❌|✅|

### Defaced Timing Marks

I did not expect to find that the `corners` algorithm would succeed on some of the defaced timing marks, but it does. This may be something we want to prevent, perhaps by requiring a minimum score on the non-corner timing marks. Here's what it finds as the timing mark shapes:

image

## Performance

The difference is small in absolute terms. On my M4 MacBook Pro, I get ~14ms for `contours` and ~1.6ms for `corners` average times on the TRR corpus.


libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/shape_finding/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+                    let start = i.saturating_sub(half);
+                    let end = (i + half + 1).min(values.len());
+                    let mut window = values[start..end].to_vec();
+                    window.sort();
+                    window[window.len() / 2]
+                })
+                .collect()
+        }
+
+        let window_size = (geometry.timing_mark_size.width / 5.0).ceil() as usize;
```
jonahkagan, line 220: How did you determine a good window size? I don't see much guidance in the wikipedia article other than: 
>Median filtering works well for both positive impulses (spikes) and negative impulses (dropouts), so long as a window can be chosen so that the number of entries infected with impulse noise is (almost) always smaller than half of the window size.

Might be nice to leave a little explanatory comment about choosing this parameter.

I'm also not convinced it needs to scale with the width of the timing mark based on that bit from Wikipedia (though obviously since timing mark width is constant for now, it's a moot point)

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/shape_finding/shape_list_builder.rs
```diff
@@ -0,0 +1,189 @@
...
+    type IntoIter = std::vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.into_shapes().into_iter()
+    }
+}
+
+/// Determines whether the given ranges share at least one value.
+fn ranges_overlap(a: &RangeInclusive<u32>, b: &RangeInclusive<u32>) -> bool {
+    a.contains(b.start()) || a.contains(b.end()) || b.contains(a.start()) || b.contains(a.end())
```
jonahkagan, line 82: Nit: I think `|| b.contains(a.start()) || b.contains(a.end())` is redundant

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/shape_finding/shape_list_builder.rs
jonahkagan, line 1: This is probably the trickiest part of the algorithm to visualize and thus understand. Some visual example (either an actual image or an ASCII image in a comment) might help. Not blocking for merging this, just noting

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/corner_finding.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    /// Find all corners of the ballot grid. Searches based on the left and
+    /// right edges rather than the top and bottom ones.
+    pub fn find_all(
+        image_size: Size<u32>,
+        geometry: &Geometry,
+        candidates: &BallotGridCandidateMarks,
+    ) -> Result<Self, Error> {
+        let timing_mark_center_to_center_distance =
+            geometry.timing_mark_size.width + geometry.timing_mark_size.height;
```
jonahkagan, line 82: This is a little confusing. It's not immediately obvious why the timing marks would be vertically spaced based on width.

Also, IIRC the distance between timing marks is currently not consistent across ballot image lengths, since we use flex containers to spread them out and the number of marks we're using doesn't divide evenly. Not sure if that will pose a larger issue to this algorithm. I see there's an error tolerance, but I figured it was worth calling out.

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/border_finding.rs
```diff
@@ -0,0 +1,202 @@
...
+        );
+        assert!(
+            candidate_timing_marks.contains(corners.1),
+            "Ending corner must be a candidate timing mark"
+        );
+
+        let border = border;
+        let (corner1, corner2) = corners;
+        let timing_mark_center_to_center_distance =
+            geometry.timing_mark_size.width + geometry.timing_mark_size.height;
```
jonahkagan, line 155: Same note about risk of this distance being different across page sizes. Also maybe should be factored since if it's being used in multiple places and is a sensitive part of the algorithm.

Also, is the distance the same for left/right columns and top/bottom rows?

jonahkagan (commented):
Very excited about this approach. It's wonderfully simple to reason through. And seems to be performing really well so far!

My only blocking comment is related to how we rely on the expected distance between timing marks. Other than that I feel great about shipping this!

libs/ballot-interpreter/src/hmpb-ts/types.ts
```diff
@@ -124,25 +124,14 @@ export interface ScoredPositionArea {
...
+ * location of bubbles in the grid.
  */
 export interface TimingMarkGrid {
   /** The geometry of the ballot card. */
   geometry: Geometry;
 
-  /** Timing marks found by examining the image. */
-  partialTimingMarks: PartialTimingMarks;
-
   /** Timing marks inferred from the partial timing marks. */
```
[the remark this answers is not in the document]
eventualbuddha: Collapsed `TimingMarkGrid` and `Complete` into `TimingMarks` ([3c91fbd](https://github.com/votingworks/vxsuite/pull/6582/commits/3c91fbd7250793ba302502be66707508382b873a)).

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/shape_finding/shape_list_builder.rs
eventualbuddha: Without that part the function is no longer associative. That is, `ranges_overlap(a, b) == ranges_overlap(b, a)` does not hold for all `a` and `b`. I'll add tests for this.

eventualbuddha: bb29ad95d95eeff471c2311208b872c7ec3d8850

jonahkagan: Oh, I wasn't thinking about ranges that were fully contained within each other, so I was wrong. Thanks!

Are you concerned at all about ranges where the end number is greater than the beginning number? I noticed that those are treated as empty ranges and always return true for overlapping.

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/corner_finding.rs
eventualbuddha: That's a great point. I'll address this by computing the expected distance between timing marks in both dimensions.

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/border_finding.rs
```diff
@@ -0,0 +1,202 @@
...
+            (top_right, bottom_right),
+        )?;
+
+        // Look for the top and bottom borders by finding the appropriate marks
+        // between the corners we used to find the left and right borders. This
+        // ensures that all four borders are congruent.
+        let mut top_candidates = candidates.top.clone();
+        let mut bottom_candidates = candidates.bottom.clone();
+        top_candidates.extend_from_slice(&[*top_left, *top_right]);
+        bottom_candidates.extend_from_slice(&[*bottom_left, *bottom_right]);
```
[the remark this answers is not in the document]
eventualbuddha: I just commented out these lines and tried it on the TRR corpus and got identical results to before doing so. That said, I do have `assert!`s in `GridBorder::find_between_corners` that ensures that the list of candidates contains the corner candidates, which in the top and bottom case come from a different shape search. So it's possible that the corners we're looking between won't be in the list of candidates, and adding them is just a safety measure.

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/border_finding.rs
eventualbuddha: As noted in [7ef20a2](https://github.com/votingworks/vxsuite/pull/6582/commits/7ef20a2e5dd0e393a574effcd81b165edf212cf8), the distances may be different horizontally vs vertically.

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/shape_finding/mod.rs
eventualbuddha: [23c4d81](https://github.com/votingworks/vxsuite/pull/6582/commits/23c4d8183a9e2857efaf2d3b6cb2ea0074abdc28)

libs/ballot-interpreter/src/hmpb-rust/interpret.rs
```diff
@@ -445,14 +445,14 @@ pub fn ballot_card(
         },
     );
 
-    let side_a_grid = side_a_grid_result?;
-    let side_b_grid = side_b_grid_result?;
+    let side_a_grid = side_a_timing_marks_result?;
```
jonahkagan, line 448: Update `side_a_grid` variable name to match the new pattern

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/border_finding.rs
```diff
@@ -37,15 +37,21 @@ impl BallotGridBorders {
     ) -> Result<Self, Error> {
         let (top_left, top_right, bottom_left, bottom_right) = corners.corner_marks();
 
+        let horizontal_timing_mark_center_to_center_distance =
+            geometry.horizontal_timing_mark_center_to_center_pixel_distance();
+        let maximum_timing_mark_center_distance_error = geometry.timing_mark_width_pixels() / 2.0;
```
jonahkagan, line 42: Should the error bound for timing marks be relative to the expected distance between marks, rather than width of a mark?

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/border_finding.rs
```diff
@@ -37,15 +37,21 @@ impl BallotGridBorders {
...
+
         let left = GridBorder::find_between_corners(
-            geometry,
+            horizontal_timing_mark_center_to_center_distance,
```
jonahkagan, line 45: This seems backwards. Shouldn't left/right use vertical distance, not horizontal distance?

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/corner_finding.rs
```diff
@@ -79,7 +79,7 @@ impl BallotGridCorners {
         candidates: &BallotGridCandidateMarks,
     ) -> Result<Self, Error> {
         let timing_mark_center_to_center_distance =
-            geometry.timing_mark_size.width + geometry.timing_mark_size.height;
+            geometry.vertical_timing_mark_center_to_center_pixel_distance();
```
jonahkagan, line 82: Seems like we should use the vertical distance for finding the adjacent marks in the left/right columns and the horizontal distance for finding the adjacent marks in the top/bottom rows, rather than just using the vertical distance for both.

jonahkagan (commented):
Overall, the changes are in the right direction. I think we could still be even more specific with how we use the expected timing mark distances and then potentially lower the related error tolerances.

libs/ballot-interpreter/src/hmpb-rust/timing_marks/corners/border_finding.rs
eventualbuddha: Yeah, that makes sense. I'll fix.

jonahkagan (approved):
💯 Let's ship it!

PR #6592, opened 2025-06-10
[VxScan] Set USB audio as default output at startup

kofi-q:
## Overview

During testing with various USB audio cards, some devices were *not* automatically selected as the default output by the system (seems to only happen for devices that identify as having a headphone output). The tactile controller we've settled on for VxScan is one of those, so this change ensures that we set the USB device as the default output, if detected at startup. This only affects the screen reader audio played in the client/kiosk-browser.

- Triggering a server shutdown if the USB device was detected, but we can't set it as the default.
- Logging if we can't detect the USB device at all - we may eventually want to throw  there as well, but we'll need to support older machines without the USB audio addition for now.

## Testing Plan
- Unit + manual tests

## Checklist

  - (Pending renames for the log event IDs - going to do that all in one go, later, for recently added logs).


nikhilb4a (approved):
🎤 ✅

PR #6593, opened 2025-06-10
auth: Cleanup file stream after artifact authentication failure

jonahkagan:
## Overview

If artifact authentication fails (e.g. when authenticating an election package), we were seeing a bug where we couldn't unmount the USB drive (or mount a newly inserted drive). It turns out that the authentication process opens a stream file handle to the artifact on the USB drive, and as long as there's a file handle open, `umount` will think the target mount point is busy and refuse to unmount it. To resolve this, we destroy the stream if any error occurs.

## Demo Video or Screenshot
N/A

## Testing Plan
- Manual test on dev machine where we saw the bug originally
- I tried to add an automated regression test, but couldn't figure out a good way to detect if the file handle was closed
## Checklist


arsalansufi (approved):
LGTM!

Issue #6554, opened 2025-06-03
VxAdmin: Adjudication - basic keyboard navigation


adghayes:
The target behavior is:
- candidate list is treated as a classic multi-select `listbox`, which means up and down arrows (and home and end, ideally) navigate through the options while tab and tab-shift navigate into and out of the group. 
  - when there is a "Dismiss" button for marginal marks, up and down pass over that too, in order
  - for write-ins, the up and down should pass focus both on the checkbox ("Write-In") but also the write-in search select. BUT the default behavior of the search select should be updated to not open on arrow key presses.  the search select only opens on "Enter" or "Space," at which points arrows do work in the select component. `Esc` exits the dropdown and leaves focus on the unopened search select
  - "Enter" or "Space" toggle options within the list
- tab navigates out of the candidate list to "Save & Next" if enabled, then "Skip" if enabled, then "Back" if enabled, then "Zoom In/Out", then "Close", then back to candidate list.

Issue #6323, opened 2025-04-22
VxAdmin: Adjudication - Allow "Esc" to close Discard Changes modal


nikhilb4a:
The [listener](https://github.com/votingworks/vxsuite/blob/ad1b02a4574c644350440ad48b28421ccc298f7d/apps/admin/frontend/src/components/discard_changes_modal.tsx#L17) works for Enter, but not for Escape. I played around with it for an hour or so and wasn't finding success, so will address later. It may be due to the key entry being intercepted by another listener.

PR #6595, opened 2025-06-11
VxAdmin: Improve adjudication interface keyboard navigation + support marginally marked write-ins

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/6554
https://github.com/votingworks/vxsuite/issues/6323

This PR:
1. Improves keyboard navigation. New behavior: Tab moves user into and out of option list, rather than scrolling within the list. Up and Down arrow keys move user within the list, following `aria-listbox` guidelines. Enter/space toggle list checkbox items. Left/Right arrow keys move user between ballots (Back/Skip). Tab indexes are updated so user can jump straight to the first item requiring adjudication in the list. `TabIndex` is set to `-1` for all inner checkboxes, as we navigate via the containers. - EDIT: This is no longer included in this PR. Some miscellaneous keyboard improvements are still included like autofocusing on modal primary buttons.
2. Adds support for marginally marked write-ins to be properly flagged as marginal marks.
3. Design touchups: Align dropdown items that have an icon, increase hover size of "Dismiss" button
4. Previously write-ins that did not require adjudication were rendered as "ContestOptionButtons". Now they are always rendered as WriteInAdjudicationButtons, though the html is equivalent. The change was made so focus could remain when a write-in was toggled from off to on. Previously focus was lost because the component would change. Now, WriteInAdjudicationButtons also support marginal mark flags, as they can have marginal marks. Because of this, I extracted `MarginalMarkFlag` into its own component, as it can be used by `ContestOptionButton` and `WriteInAdjudicationButton`

## Demo Video or Screenshot

UPDATED - simple navigation, tab only, no arrow keys

https://github.com/user-attachments/assets/d57f6b3d-1887-46d6-8b6c-e81b2944e1f0

Navigating to adjudicate

https://github.com/user-attachments/assets/965c12ab-34a2-41e5-81b3-5bf3532755a5

Navigating between ballots (left/right)

https://github.com/user-attachments/assets/726c77e5-c65d-4f4a-bc72-80a85766009e

## Testing Plan

Automated and QA this week.

## Checklist

### DIFF
--- a/apps/admin/frontend/src/components/marginal_mark_flag.tsx
+++ b/apps/admin/frontend/src/components/marginal_mark_flag.tsx
@@ -1,0 +1,52 @@
+import { Icons, Button } from '@votingworks/ui';
+import styled from 'styled-components';
+
+const Container = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background-color: ${(p) => p.theme.colors.warningContainer};
+  border: ${(p) => p.theme.sizes.bordersRem.thin}rem solid
+    ${(p) => p.theme.colors.outline};
+  border-bottom: 0;
+  border-radius: 0.5rem 0.5rem 0 0;
+  color: ${(p) => p.theme.colors.neutral};
+  font-weight: 500;
+  padding: 0.25rem 0.5rem;
+
+  button {
+    gap: 0.25rem;
+  }
+`;
+
+const IconTextContainer = styled.div`
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+`;
+
+export function MarginalMarkFlag({
+  onDismissFlag,
+}: {
+  onDismissFlag: () => void;
+}): JSX.Element {
+  return (
+    <Container>
+      <IconTextContainer>
+        <Icons.Warning color="warning" />
+        Review marginal mark
+      </IconTextContainer>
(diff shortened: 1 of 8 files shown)


apps/admin/frontend/src/components/marginal_mark_flag.tsx
```diff
@@ -0,0 +1,52 @@
+import { Icons, Button } from '@votingworks/ui';
```
nikhilb4a, line 1: This file was extracted from `ContestOptionButton`

apps/admin/frontend/src/components/write_in_adjudication_button.tsx
```diff
@@ -146,72 +177,90 @@ export const WriteInAdjudicationButton = forwardRef<HTMLDivElement, Props>(
     ];
 
     return (
-      <Container ref={ref} style={{ zIndex: isFocused ? 10 : 0 }}>
-        <RoundedCheckboxButton
+      <Container
```
nikhilb4a, line 180: The diff looks bigger here than it is. `Container` has two new props, `RoundedCheckboxButton` -> `StyledCheckboxButton`. `showSearchSelect` added as now we want to be able to render write-ins without the SearchSelect (when they are in a default state without a vote)

PR #6610, opened 2025-06-11
[VxScan] Add/update IDs for audio-related log events

kofi-q:
## Overview

Follow-up from recent work adding screen reader audio support to VxScan - replacing a few placeholder log event IDs related to system audio.

## Testing Plan
- Updated relevant tests

## Checklist


nikhilb4a (approved):
☑️

PR #6635, opened 2025-06-16
[VxScan] Set calibrated system volume level at startup

kofi-q:
## Overview

Adding a server startup step to set the system volume for the USB audio device to 100%, which, in combination with the in-app volume setting brings the screen reader audio into the VVSG2 default range of 60db - 70db.

## Testing Plan
- Unit tests for `pactl` integration.
- Manual test for volume levels with our trusty VxAustin foam head:
IMG_5007

## Checklist


nikhilb4a (approved):
LGTM! And that foam head with the cutout is amazing and terrifying 🤣

PR #6639, opened 2025-06-17
[pollbook] Include middle name and suffix on `voters` table

kshen0:
## Overview

* Adds `*_middle_name` and `*_suffix` fields to the `voters` table.
* Updates `findVoterWithName` and `searchVoters` to pass those fields. Behavior should be unchanged for both.
* `findVoterWithName` still applies events to the `voter` objects but does not use the events for name matching
* Currently the search frontend does not pass middle name or suffix
 
## Demo Video or Screenshot
Manual test of existing behavior:

https://github.com/user-attachments/assets/88b7e6ea-9c34-48b3-8968-4d8c676efef1

## Testing Plan
- updated tests
- added test that `findVoterWithName` does not find a match if middle name or suffix doesn't match

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/local_store.test.ts
+++ b/apps/pollbook/backend/src/local_store.test.ts
@@ -11,13 +11,7 @@ import { LocalStore } from './local_store';
 import { VoterRegistrationRequest } from './types';
 
 test('findVotersWithName works as expected - voters without name changes', () => {
-  const workspacePath = tmp.dirSync().name;
-  const localStore = LocalStore.fileStore(
-    workspacePath,
-    mockBaseLogger({ fn: vi.fn }),
-    '0000',
-    'test'
-  );
+  const localStore = LocalStore.memoryStore();
   const testElectionDefinition = getTestElectionDefinition();
   const voters = [
     createVoter('10', 'Dylan', `O'Brien`, 'MiD', 'I'),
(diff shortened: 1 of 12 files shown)


apps/pollbook/backend/src/local_store.ts
```diff
@@ -364,56 +371,23 @@ export class LocalStore extends Store {
...
+      `${middleNamePattern}`,
+      `${firstNamePattern}`,
+      `${suffixPattern}`
     ) as Array<{ voter_id: string; voter_data: string }>;
 
     if (voterRows.length === 0) {
       return [];
     }
 
-    // Map voter rows to voter objects
```
kshen0: I realized we still need to apply events on top of `voter` objects even if we're not filtering by the values. It's readded in a later commit

apps/pollbook/backend/src/local_store.test.ts
```diff
@@ -161,18 +161,44 @@ test('findVoterWithName works as expected - voters with name changes', () => {
...
-      lastName: 'Doe',
-      middleName: 'Allen',
-      suffix: 'Sr',
-    })
-  ).toEqual(
-    expect.arrayContaining([
-      expect.objectContaining({ voterId: voters[0].voterId }),
-    ])
-  );
+  await vi.waitFor(() => {
```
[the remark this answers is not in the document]
carolinemodic: hmm.. might be due to using the file store with a tmp db file here, I forget if that was necessary or not (as opposed to a memory store). The file stores are needed when a test is testing that both the peerApp and localApp are syncing with one another since that happens through the db file, but the test is specific to localStore then a memory store should be ok and less likely to have a race condition issue with another test, but also this fix is fine

apps/pollbook/backend/src/store.ts
```diff
@@ -210,7 +213,8 @@ export abstract class Store {
       events
     );
 
-    return updatedVoters[voterId];
+    const value = updatedVoters[voterId];
```
carolinemodic, line 216: nit: revert, I assume this was an accidental leftover change from debugging

carolinemodic (approved):

PR #6642, opened 2025-06-17
[VxScan] Prevent repeated controller events when holding down a button

kofi-q:
## Overview

We initially used `keydown` events in VxMark (v4) to trigger keyboard shortcuts for navigation and audio control, which worked well, since we only emit a single event when holding down buttons on the VSAP controllers. For VxScan, the controller behaves like regular keyboard and emits repeated `keydown` events when held down, triggering repeated UI actions.

Adding a check for repeated keys for audio controls and VxScan navigation controls to satisfy the VVSG 2 requirement `7.2-M – No repetitive activation`.

## Testing Plan
- Existing tests as regression guard - manual for the hardware controller behaviour.


nikhilb4a (approved):
Small thought but lgtm!

PR #6643, opened 2025-06-17
Do not sync events across pollbooks when code version and/or precinct do not match

carolinemodic:
## Overview
Updates pollbook connection logic to only share events when the code version and precinct match, along with the previous blockers. Also in rethinking the potential race conditions and edge cases here added two more stricter checks (see commits 2 and 3 respectively):

- Checks in the same transaction when setting precinct that events do not exist and return an error to the frontend if they do. The frontend is already checking if events exist and disabling the button but its possible one could sneak in from another machine and this will prevent the write from saving if it happens. The UI handles the error by just disabling the button not showing a message, since this is never really expected to happen in the real world I think thats ok. 
- Returns the current pollbook configuration information with events when sending events to another machine. That machine then checks, in the same transaction, that that configuration matches its internal configuration before committing the events to its own db. This could prevent against a scenario where two pollbooks are connected/configured for Precinct 1. Pollbook-2 checks in Bob in Precinct-1, and I change to Precinct-2 right before I recieve that event from Pollbook-2. This will ensure that I realize the precinct mismatch and bail out before saving the Bob check in event. 

My goal in this PR was to get to 100% test coverage in networking.ts and avahi.ts so there are a few semi-unrelated tests added, some istanbul ignores, and some removal of unneeded lines of code (checking for impossible conditions, debugging remnants from the past, etc.) 

## Demo Video or Screenshot
N/A 

## Testing Plan
Ran tests, ensured 100% coverage of networking.ts

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/app_multi_node.test.ts
+++ b/apps/pollbook/backend/src/app_multi_node.test.ts
@@ -234,7 +234,7 @@ test('connection status between two pollbooks is managed properly', async () => 
     // Set the pollbooks for the same election and precinct
     pollbookContext1.workspace.store.setElectionAndVoters(
       electionDefinition,
-      'fake-package-hash',
+      'mock-package-hash',
       testStreets,
       testVoters
     );
@@ -253,7 +253,7 @@ test('connection status between two pollbooks is managed properly', async () => 
 
     pollbookContext2.workspace.store.setElectionAndVoters(
       electionDefinition,
-      'fake-package-hash',
+      'mock-package-hash',
       testStreets,
       testVoters
     );
@@ -511,7 +511,7 @@ test('connection status is managed properly with many pollbooks', async () => {
     for (const context of pollbookContexts) {
       context.workspace.store.setElectionAndVoters(
         electionDefinition,
-        'fake-package-hash',
+        'mock-package-hash',
         testStreets,
         testVoters
       );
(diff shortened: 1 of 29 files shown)


apps/pollbook/backend/src/peer_app.ts
```diff
@@ -26,9 +26,13 @@ function buildApi(context: PeerAppContext) {
...
     getEvents(input: { lastEventSyncedPerNode: Record<string, number> }): {
       events: PollbookEvent[];
+      configurationInformation: PollbookConfigurationInformation;
       hasMore: boolean;
     } {
-      return store.getNewEvents(input.lastEventSyncedPerNode);
+      return {
+        ...store.getNewEvents(input.lastEventSyncedPerNode),
+        configurationInformation: store.getPollbookConfigurationInformation(),
+      };
```
kshen0, line 35: Is it necessary for `getEvents` to return the configuration information and events in the same call? It would be a little clearer if the configuration information was returned in a separate call but I understand this setup if it avoiding race conditions w/ separate calls would be annoying

apps/pollbook/backend/src/app_multi_node.test.ts
```diff
@@ -536,6 +545,514 @@ test('connection status is managed properly with many pollbooks', async () => {
...
+      'fake-package-hash',
+      testStreets,
+      testVoters
+    );
+    pollbookContext2.workspace.store.setElectionAndVoters(
+      electionDefinition,
+      'fake-package-hash',
+      testStreets,
+      testVoters
+    );
```
kshen0, line 1003: Seems like there's some repeated boilerplate between these tests. Can we factor out what is convenient to helper functions?

apps/pollbook/backend/src/peer_app.ts
carolinemodic: I'm trying to reduce the number of calls happening over the network as this is being called constantly.

PR #6648, opened 2025-06-17
[scan] Support both speaker and headphone audio in the HWTA

kofi-q:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6484

- Mirroring the prod app setup to detect and configure both the builtin and USB audio devices.
  - Not hard-failing when audio configuration fails like we do in prod, since it's helpful to just log and degrade functionality in this case.
- Adding controls for playing audio separately through the speaker and/or the headphones (via USB audio).

## Demo Video or Screenshot

scan-hwta-dual-audio

## Testing Plan
- Manual

## Checklist

### DIFF
--- a/apps/scan/backend/src/electrical_testing/app.ts
+++ b/apps/scan/backend/src/electrical_testing/app.ts
@@ -5,10 +5,17 @@ import { SheetOf } from '@votingworks/types';
 import express, { Application } from 'express';
 import { mkdir, readdir } from 'node:fs/promises';
 import { join } from 'node:path';
+import { LogEventId } from '@votingworks/logging';
 import { getMachineConfig } from '../machine_config';
 import { type ServerContext } from './context';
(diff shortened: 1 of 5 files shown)


apps/scan/backend/src/electrical_testing/app.ts
```diff
@@ -104,7 +115,14 @@ function buildApi({
         usbDriveStatus.status === 'mounted'
           ? join(usbDriveStatus.mountPoint, 'ballot-images')
           : workspace.ballotImagesPath;
-      const allFileNames = await readdir(basedir);
+      const allFileNames = await readdir(basedir).catch((error): string[] => {
```
kofi-q, line 118: Unrelated change, but this was failing for me out-the-box, since I didn't have the right directory structure in my mock USB drive

nikhilb4a (approved):
☑️

PR #6660, opened 2025-06-18
Remove deprecated non-Grout APIs

eventualbuddha:
## Overview

Migrates the `central-scan` APIs still used around adjudication to use Grout, then deletes the `@votingworks/api` package since it's no longer needed.

I started on this task because I had to make some changes to the ballot eject screen in `central-scan` and I didn't want to add more tests on top of the old and busted API.

## Demo Video or Screenshot
Screenshot-VxCentralScan-2025-06-18T19:38:39 844Z

## Testing Plan
Automated test updates, plus verified manually that the ballot eject/adjudication screen still works.


apps/central-scan/backend/src/app.ts
```diff
@@ -207,6 +208,52 @@ function buildApi({
       }
     },
 
+    getNextReviewSheet() {
```
jonahkagan, line 211: Could you add return type annotations to these API methods? It's a convention that I think is helpful

jonahkagan (approved):
Finally!! Thank you

Issue #6618, opened 2025-06-13
Marginal mark count logging


arsalansufi:
https://votingworks.slack.com/archives/C026YV9LPGB/p1749822940209069

arsalansufi:
Plan: After import on VxAdmin, log, whether marginal mark adjudication is enabled or not, mark threshold counts in buckets a bit past the definite threshold (i.e. a rough distribution)

PR #6661, opened 2025-06-18
apps/admin, libs/logging: Add log for mark scores distribution on cvr import

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/6618

We want to better understand what thresholds would work well for marginal mark adjudication, so we are introducing a log that includes: 1) total marks (over 0.00 score) and 2) a distribution of low scores (0.2 and below) in 0.01 increments

## Demo Video or Screenshot

Example of a Scan import followed by a CentralScan:

Screenshot 2025-06-18 at 1 09 48 PM

## Testing Plan

Automated tests and reviewed log

## Checklist

### DIFF
--- a/apps/admin/backend/src/util/cast_vote_records.test.ts
+++ b/apps/admin/backend/src/util/cast_vote_records.test.ts
@@ -1,7 +1,12 @@
 import { expect, test } from 'vitest';
 import { electionTwoPartyPrimaryFixtures } from '@votingworks/fixtures';
 import { Tabulation } from '@votingworks/types';
-import { getCastVoteRecordAdjudicationFlags } from './cast_vote_records';
+import {
+  formatMarkScoreDistributionForLog,
+  getCastVoteRecordAdjudicationFlags,
+  MarkScoreDistribution,
+  updateMarkScoreDistributionFromMarkScores,
+} from './cast_vote_records';
 import { CastVoteRecordAdjudicationFlags } from '..';
 
 const electionDefinition =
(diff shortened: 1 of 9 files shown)


apps/admin/backend/src/util/cast_vote_records.ts
```diff
@@ -91,3 +91,50 @@ export class CvrContestTagList {
...
+/**
+ * Formats the score distribution for logging.
+ * i.e. "0.01": 15, "0.02": 100...
+ */
+export function formatMarkScoreDistributionForLog(
+  distribution: Map<number, number>
+): string {
+  return JSON.stringify(
+    Object.fromEntries(
+      [...distribution].map(([start, count]) => [`${start.toFixed(2)}`, count])
```
arsalansufi, line 137: nit: I think the `${}` may be unnecessary here since `.toFixed` already converts to a string

apps/admin/backend/src/util/cast_vote_records.test.ts
```diff
@@ -143,3 +148,48 @@ test('multiple flags', () => {
...
+    distribution: new Map<number, number>(),
+    total: 0,
+  };
+
+  const markScores: Tabulation.MarkScores = {
+    'contest-a': {
+      'option-1': 0.01,
+      'option-2': 0.15,
+      'option-3': 0.0, // should be ignored
+      'option-4': 0.21, // should not go in a bucket, but counts toward total
```
arsalansufi, line 163: I'm not sure whether actual mark scores have fixed-digit values like this but might be good to throw in some "longer" values, e.g., 0.12312341234, to test the rounding logic

arsalansufi (approved):
Yay excited to get this data!

apps/admin/backend/src/util/cast_vote_records.test.ts
arsalansufi: Oh perfect! So long as the test case is representative of reality, I have no issue here

nikhilb4a: >  So long as the test case is representative of reality

Thanks for naming that principle!

PR #6667, opened 2025-06-19
Remove System Settings from Design App

adghayes:
Removes the System Settings feature from the Design app, as it's not needed in this context. System settings functionality remains available in other VxSuite apps where it's actually used. Changes: Removed SystemSettingsScreen component and related tests, removed system settings route from app navigation, and cleaned up unused imports and dependencies. This simplifies the Design app interface by removing functionality that doesn't belong in this specific application.


arsalansufi:
And I do think we should clean up the internal references to the bitonal threshold in a follow-up (doesn't have to be in this release). Opened an issue accordingly: https://github.com/votingworks/vxsuite/issues/6672

Issue #6602, opened 2025-06-11
VxAdmin: do not error/reject invalid imports of results due to manual tallies


arsalansufi:
In walking through the multi-ward aggregation process, one of my manual results were malformed. It failed on import to another VxAdmin, which is by design. But the surfaced error had no details. When I looked in the logs, the error was `Expected an integer value for total ballots but got 17.33333` , which pointed me in the right direction but still left me confused because all the numbers in the actual file are integers, the total ballot is calculated during the conversion.

high-level - If the import fails because the imported file has vote counts that don't add up properly, that information should be surfaced to the user. 

Issue automatically created from Slack feedback reported by drew: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1749599144715159

adghayes:
@mattroe @arsalansufi "malformed" in this case means "invalid."

If you have invalid manual results - e.g. the numbers don't add up - you can still print results reports. But if you go through the aggregation process, you will hit a non-descript error on the aggregating machine on import. If this comes up in the field, we'll know where to direct people. But better error messaging could help cities figure it out for themselves, which is why I was considering adding this small change to 4.0.2.

mattroe:
Unless there's a good reason to reject, I think we should remove the validation for this particular case given that it's likely to happen in NH from manual tallies

mattroe:
This will already be flagged in the manual tally interface as inaccurate so I don't think we need to block.

arsalansufi:
Per in-person discussion: We could also explore blocking invalid exports, though we should understand why we aren't blocking currently

PR #6675, opened 2025-06-23
VxAdmin: Allow importing of invalid manual tallies

jonahkagan:
## Overview

Fixes https://github.com/votingworks/vxsuite/issues/6602

If a user makes a mistake when entering manual tallies and the votes don't add up to match the total number of ballots, we show a warning on-screen that the tallies are invalid. However, we still allow exporting reports with these invalid tallies. If one of these reports is later imported, we currently block the import and raise an error, which is a bit too late to be blocking, since they would have to go back to the original machine to fix the error.

This PR changes the import logic to allow invalid tallies by rounding up the computed total ballot count. This results in the same warning shown on screen for the imported tallies.

## Testing Plan
- Manual test
- Updated automated test
## Checklist


kshen0 (approved):
Code LGTM but more generally, why allow erroneous exports in the first place? Is there generally some tolerance for clerical errors?

PR #6678, opened 2025-06-24
Tweak VxPollBook backend code coverage threshold

arsalansufi:
## Overview

I've been seeing some flakiness on main with the following error:

```
ERROR: Coverage for branches (80%) does not meet global threshold (80.18%)
```

I suspect that there's a test that only sometimes exercises a branch, so I'm adjusting the threshold accordingly.

### DIFF
--- a/apps/pollbook/backend/vitest.config.ts
+++ b/apps/pollbook/backend/vitest.config.ts
@@ -8,7 +8,7 @@ export default defineConfig({
     coverage: {
       thresholds: {
         lines: 85.63,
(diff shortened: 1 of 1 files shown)


arsalansufi:
> I thought I got to the bottom of all the differences but I have several relevant PRs that are blocked on review right now so I'm not really sure the state of main. Once everything is merged I'll do another pass to see if there are other discrepancies.

Ah gotcha that makes sense, thanks for the heads up!

PR #6679, opened 2025-06-24
Disable ballot scale check by default while we test more

arsalansufi:
## Overview

After better understanding the approach that we're taking to detect ballots that have been printed at too low of a scale (which as a reminder can negatively affect interpretation), I'm realizing that I'd like us to test the approach further, both for actual deployments and even for cert. My main concern is that the measurement differences that we're looking at are extremely small.

We're looking at the average distance from the start of one timing mark to the next in the bottom row of timing marks.

<table><tr><td>

</td></tr></table>

Performing some measurements on ballots printed at various scales in the real world, we took the following measurements:

```
100%: 6.03mm
 99%: 5.99mm
 98%: 5.96mm
 97%: 5.88mm
 96%: 5.84mm
 95%: 5.80mm
 94%: 5.77mm
```

Considering that the difference in the measurement on a 100% scale ballot vs. a 98% ballot is `6.03mm - 5.96mm = 0.07mm`, which is apparently the thickness of a relatively thin eyelash 😆, I'm concerned about noise resulting in a high false positive rate. With that small of a measurement, something as simple as how the ink/toner settles/bleeds on the paper, variance across PDI scanners, or even humidity, could have an effect.

We don't have enough time before the v4.0.2 release to QA extensively across printers, paper types, PDI scanners, and environmental conditions, so I'm making the call to disable this check by default in v4.0.2. I'm still leaving all the code in though so that we can continue to test and refine the measurement. I am also personally interested in a "larger" measurement, e.g., considering the width of the entire bottom row of timing marks or the average horizontal distance beteween pairs of left and right timing marks.

I would still like us to land some safeguard before we finish cert, e.g., in v4.0.3.


libs/types/src/system_settings.ts
```diff
@@ -164,10 +164,4 @@ export const DEFAULT_SYSTEM_SETTINGS: SystemSettings = {
   centralScanAdjudicationReasons: [],
   adminAdjudicationReasons: [],
   precinctScanEnableBmdBallotScanning: true,
-
-  // Determined by analyzing ballots printed at scale levels from 94-100%,
-  // then measuring the timing mark grid against the expected size.
-  //
-  // See https://votingworks.slack.com/archives/CEL6D3GAD/p1750140685476599?thread_ts=1750095447.642289&cid=CEL6D3GAD.
-  minimumDetectedScale: 0.985,
```
arsalansufi, line 172: If excluded, we simply don't perform the check

Issue #6653, opened 2025-06-18
VxPollBook: Multi-Ward Support - Accept & Mock City-Style Street Files


adghayes:
Accept street files for cities as well, we have an example here: https://drive.google.com/file/d/1dnWgIFRLE2WnfqwmK4Hf2zgmHCp4-zSO/view?usp=drive_link

The only changes are that 
- the "Town" column is now "City" 
- the "District" column is now "Ward"

Although we don't have a city voter file at this time, I think we can assume it's the same but with "District" replaced by "Ward"

PR #6682, opened 2025-06-25
Accept city-styled voters/street names files and update pollbook package fixtures

carolinemodic:
## Overview
Recommend reviewing by commit. 
https://github.com/votingworks/vxsuite/issues/6653

Updates the pollbook-package fixtures to create two sample pollbook-package in fixtures to cover different testing needs:
- SimpleSinglePrecinctElection - The only single precinct elections in the current election fixtures are the NH accuvote ones which seem weird to use with pollbooks. This makes a very simple single precinct election for testing with pollbooks and uses the existing voters (10k) and street names files with that election in a mock pollbook package. The street names and voters files have columns formatted like we've seen in towns with 1 precinct. 
- MultiPartyPrimaryElection - I then have a sample multi-precinct pollbook package now user electionMultiPartyPrimary instead of famous names. This is because famous names does not name their precincts "Precinct 1", "Precinct 2", etc. which is what Drew and I have decided the convention will be to match up with values of just "1" and "2" as the "Ward" in the voters and street names files. electionMultiPartyPrimary already labels its precincts this way, and we will then have a mock file for when we introduce primary support. The voters/street names files are formatted like we've seen for a city with a "ward" column. 

Updates to file parsing:
- Rename district to precinct and first populate with ward, then district if ward is not defined. I'm a little skeptical district isn't something else and we don't need both but ... this is fine for now. 
- The street name file added has a new column "city" but there was no "town" previously. I am saving this in case we want it but we currently use Postal City / Town to be the "city" on the voter when we register, not sure if we should re-evaluate that. Making a note in the doc for Drew. 

## Demo Video or Screenshot
N/A
## Testing Plan
Ran Tests, used new pollbook package fixtures and looked at DB data. 

## Checklist


adghayes:
> The street name file added has a new column "city" but there was no "town" previously. I am saving this in case we want it but we currently use Postal City / Town to be the "city" on the voter when we register, not sure if we should re-evaluate that. Making a note in the doc for Drew.

@carolinemodic looking at the street file from Milford, there is no "Town" column, but from Hanover, there is a "Town" column. Due to the way we handle the file imports and ignore extraneous fields, this didn't create any problems for the Hanover pilot. So, I think the "City" column can just be ignored. But the fact that we are saving "City" after this PR is fine.

Correct, that we should be using Postal City/Town for the auto-complete / new addresses. Because even in a town election (e.g. Hanover) the postal city/town can vary, it turns out, and we want to use the correct city/town if it's not the election-holding town.

Issue #6156, opened 2025-03-26
VxPollBook: Multi-Ward Support - Check-In Changes


adghayes:
Search page:

The button from the search page should say something other than "Start Check-In" because it's assumed they can't be checked in. Instead "View Details." We should also indicate the ward on the search results.

Check-in page: 

- check-in is disabled
- name change is disabled
- address change is enabled, in the event they are changing their address into the current ward
- callout indicates that their current address is for a different ward, and which ward

PR #6696, opened 2025-07-01
VxPollbook Multi Precinct Check In Restrictions

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/6156

Implements the restrictions for checking in, or making other voter changes, when a voter is not in the correct precinct. Address changes are always permitted as you can change a voter address into the current precinct. Notably as implemented in a previous PR you can NOT change a voter address outside of the precinct, this combined with the fact that you can only sync with machines configured for the same precinct means that we do not have to worry about the voter being configured for the correct precinct on the backend as there is no way for an event to sneak in and change that if the frontend is allowing a check in (or name change/ etc.). I have previously added backend errors for situations like checking in a voter that is flagged as inactive as an event that changes that status CAN sneak in and sync after a UI is displayed. 

In a single precinct election, nothing in the UI is changed and a voter's precinct is not surfaced. 

In a multi precinct election adds the voters precinct to:
- The search results UI 
- The EM voter details screen
- The PW voter confirm screen

If the voter is not in the precinct that the machine is configured for, changing name, flagging inactive, and checking in are not allowed. 

Some notes mostly for @adghayes to review on his return:
- The Ticket mentions a few UI options. This UI is what Matt thought looked the most sensible when we looked at it but I'm assuming you may take a pass. 
- I am sorting the voter search results to put matching precinct voters first before voters from other precincts. This was not mentioned in the ticket but it seems very unlikely you would want to find someone outside of the precinct so I thought this was a better pattern. Matt also suggested just filtering to only voters in the same precinct but then you would need a toggle in case you want to see all. 
- You did not specifically say we should block flagging as inactive on the matching precinct but it is simpler to do so and I don't see why someone would need to do that for a voter in the wrong precinct. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/5bbc5fc7-4102-4d3b-b927-70ecfe07fbbf

## Testing Plan
Manual testing as shown above + similar run through on a single precinct election to verify no changed. 

Added tests for all new flows/branches 

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/app.test.ts
+++ b/apps/pollbook/backend/src/app.test.ts
@@ -1123,3 +1123,263 @@ test('mark a voter inactive', async () => {
     expect((votersAbigail2 as Voter[])[2].isInactive).toEqual(false);
   });
 });
+
+test('voter search results prioritize voters from configured precinct', async () => {
+  await withApp(async ({ localApiClient, workspace, mockPrinterHandler }) => {
+    // Set up election with voters from different precincts
+    const testVoters: Voter[] = [
+      // These voters will be in a different precinct than the configured one
+      {
+        ...createVoter('voter-1', 'Alice', 'Johnson'),
+        precinct: 'precinct-2',
+      },
+      {
+        ...createVoter('voter-2', 'Bob', 'Johnson'),
+        precinct: 'precinct-2',
+      },
+      // These voters will be in the configured precinct
+      {
+        ...createVoter('voter-3', 'Charlie', 'Johnson'),
+        precinct: currentPrecinctId,
+      },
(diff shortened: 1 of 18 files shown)


kshen0 (approved):

adghayes:
@carolinemodic UI looks good at first glance. Sorting call is great too. I do think that flagging a voter as inactive needs to be enabled regardless of precinct.

If a voter in Ward 1 dies, election administrators will want to reflect that death in all wards. Otherwise I could come into Ward 2 and claim to be the deceased voter with an address change. As is, you'd have to update the address with a false address in order to do that.

Issue #6076, opened 2025-03-04
VxDesign: Move base font size changes into ballot templates


jonahkagan:
Instead of changing the base font size for condensed ballots in the rendering layer, set the base font size in the templates themselves.

PR #6708, opened 2025-07-02
hmpb: Refactor to move base style definition into ballot templates

jonahkagan:
## Overview

Fixes https://github.com/votingworks/vxsuite/issues/6076

Having the renderer be responsible for including `BaseStyles` was always a bit of a violation of the abstraction that templates specify content and styling while the renderer is just responsible for rendering it. It was a shortcut I put in place since all templates shared the same base styles.

This started to break down when we needed to parameterize the styles based on ballot props (namely, adding `compact`, which changes font size and line height). In the future, it looks like we'll also need to change the fonts for upcoming NH state ballots.

To get ahead of this, I added a new field to the ballot template to specify the base styles to use. The renderer will still only add this to the document head once at the beginning, but it will make it easier to customize these styles on a per-template basis moving forward. Plus, it patches up the separation of concerns shortcut.

## Demo Video or Screenshot
N/A

## Testing Plan
Existing tests

## Checklist

### DIFF
--- a/libs/hmpb/src/all_bubble_ballot/template.tsx
+++ b/libs/hmpb/src/all_bubble_ballot/template.tsx
@@ -17,6 +17,7 @@ import { RenderScratchpad } from '../renderer';
 import { allBubbleBallotConfig } from './config';
 import { candidateId, contestId } from './election';
 import { Footer } from './footer';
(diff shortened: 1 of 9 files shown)


kofi-q (approved):
💯

PR #6711, opened 2025-07-02
VxDesign: Allow users to control ballot density

### LINKED ISSUE #6075: VxDesign: User-facing ballot density controls
Allow election officials to choose their density setting. We can potentially move back to having 3 density levels like we did in v3.

We can also incorporate some of the learnings from talking with CCD:
- Don’t go below 10pt font for contest content, strive for 11pt or 12pt
- Look for places to condense non-contest content first (e.g. headers, footers)
- Moving yes/no options inline with ballot measure text is fine as long as it’s done consistently across contests (even with longer customizable option labels)

jonahkagan:
## Overview

Closes #6075 

Adds a "Density" control to the Ballot Layout tab of the Proof Ballots screen (which is where we originally had this feature).

Some notes:
- I decided to stick with two density levels (default and compact), rather than the three levels we had previously, since we've already got the two implemented and it's less work to maintain (i.e. test and QA). We haven't totally verified that there's a need for an intermediate level, though if we wanted to implement one, we could probably just set the font size to 11pt and that would work pretty well.
- Currently, cloning an election doesn't copy over the density setting. This could potentially be a bit confusing. I considered adding `compact` to `Election.ballotLayout` to change this behavior. I decided to leave it, though, because I think it's a good thing for ballot accessibility that the density always starts out at default and you have to manually take an action to change it to compact.
 
## Checklist


kofi-q (commented):
My vote's for the radio buttons as well 👍 

Looking good to me! On the fence about the "Density" label, but no strong feelings - could only come up with "Layout", which is unsatisfyingly vague

mattroe:
Happy with simplifying to two options for now and this radio button is exactly where I envisioned we'd implement. Density is the best copy I can think of as well. LGTM!

kofi-q (approved):

Issue #6440, opened 2025-05-14
VxPollBook: Cosmetic Election Manager Voter Page Changes


adghayes:
On the voter details page:
 
- Capitalize button "Undo check-in" as "Undo Check-In"
- Make "Undo Check-In" button non-primary danger coded
- Capitalize "Checked in" header to "Checked In"
- Add "Done" icon before "Checked In" to mirror the search page
- Capitalize "Not checked in" to "Not Checked In"
- Add "Info" icon before "Not Checked In" to also have an icon there
- For the time, the current timestamp seems like it's a unix timestamp that's GMT and not time-zone sensitive. Make it timezone sensitive and have seconds, but not fractional seconds
- Replace "Machine" with "Poll Book"
- Move the "Close" button to the bottom to mirror the poll worker check-in page. People may be using these screens back to back and I think the inconsistency would throw people off.
- There's some spacing weirdness below the "(Not) Checked In" header that is making the card a little too high when "Not Checked In" and adds spacing between the header and "Time"

On the voter search page, previously around the "Undo Check-In" button there was a small indication above the "Undo Check-In" button that had the check-in information. Although this is duplicative _across_ pages, I think that it's valuable to have it on the search page as well so you don't have to click in, especially in a case where someone was accidentally checked-in. Can you revive that indicator to live above the "View Details" button? I'm not sure how exactly it would look, but I think this would be a good one to look at a screenshot together.

<img width="251" alt="Image" src="https://github.com/user-attachments/assets/234331b8-feb3-4fb3-902d-f0aa02b77dba" />

PR #6716, opened 2025-07-03
voter details page style changes

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/6440

Cosmetic changes for election manager voter search UX.

## Demo Video or Screenshot

cc @adghayes for product feedback

**Search**

Screenshot 2025-07-07 at 10 35 33 AM

**Voter checked in**

Screenshot 2025-07-07 at 10 35 38 AM

* Buttons look misaligned relative to each other even though they are correctly aligned to their columns
* We might consider a hamburger menu on these columns as functionality continues to grow and it becomes cumbersome to add more buttons
* "Close" button is actually in the opposite place compared to the pollworker screen, where it's on the bottom left
  * In a footer with only 1 button I think it feels more natural to keep it on the right because I generally associate action with sides of the screen (left for previous, right for next, probably has to do with reading direction)
  * That said, open to feedback if we want to put the "Cancel" button in the exact same spot (on the left)

**Voter not checked in**

Screenshot 2025-07-07 at 10 45 51 AM

## Testing Plan

- N/A display changes only

## Checklist

### DIFF
--- a/apps/pollbook/frontend/src/voter_details_screen.test.tsx
+++ b/apps/pollbook/frontend/src/voter_details_screen.test.tsx
@@ -372,7 +372,7 @@ test('undo check-in', async () => {
   });
   userEvent.click(undoButton);
 
-  await screen.findByText('Not checked in');
+  await screen.findByText('Not Checked In');
 });
 
 test('reprint check-in receipt', async () => {
--- a/apps/pollbook/frontend/src/voter_details_screen.tsx
+++ b/apps/pollbook/frontend/src/voter_details_screen.tsx
@@ -1,5 +1,6 @@
 import {
   Button,
+  ButtonBar,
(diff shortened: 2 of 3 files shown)


adghayes:
@kshen0 thank you for putting up these screenshots!

- I think that keeping the "Close" button on the right hand side makes sense for the reasons you described. My one suggestion here would be to remove the "Delete" icon because it seems out of place in that position.
- The spacing on the information above the "View Details" button on the EM search page looks a bit off. I recommend just re-using the component from the Voter Search page, which is I believe how it used to work, with special casing for the "Not Checked In" as that is not used / necessary on the Voter Search page.

kshen0:
Forgot to mention before review but I updated screenshots with proposed changes

Issue #6472, opened 2025-05-19
VxPollBook: Mailing Addresses!


adghayes:
We need to support displaying _and_ editing mailing addresses in order to meet state requirements, as communicated verbally to date. 

#### Displaying Mailing Addresses

On the check-in page and the voter details page, we should show the mailing address beneath the domicile address. Suggestion:

- [ ] change label for current "Address" to "Domicile Address"
- [ ] when mailing address is non-empty (it will usually be empty), display "Mailing Address" beneath the "Domicile Address"

#### Modifying Mailing Addresses

- [ ] update existing "Update Address" buttons to be "Update Domicile Address" buttons
- [ ] add new buttons "Update Mailing Address" in the two places (voter check-in and voter detail) where you can currently update domicile addresses

PR #6718, opened 2025-07-03
VxPollbook Mailing Address Support

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/6472
Adds Mailing Address support in addition to Domicile Address support (and renames the latter in the UI). Mailing addresses are not subject to the same valid street restrictions as domicle so the UI for editing is slightly different. Since there is a zip5 and zip4 data item I am parsing the zip into both of those if the user adds a -. Street # / Suffix splitting works the same as in domicile address editing. 

A few notes:
- I decided to make mailing address NOT editable when you are in the wrong precinct for a voter. Domicile address IS editable, because you can change to the current precinct, but I don't know why you would need to change the mailing address for a voter with the domicle address of another precinct (unless you are also changing that) so I err'd on the side of disabling. cc @adghayes 
- If the user did not previously have a mailing address and then adds on there isn't an old crossed out one shown, just the new. 
- Similarly in the situation above the receipt will just list "None" as the previous mailing address, see snapshot. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/c32dbca7-a9b2-4c77-8b2a-ba6c2a4f6df0

see snapshots for receipts

## Testing Plan
See video, ran tests.

## Checklist


adghayes:
LGTM! Agree on the call to disable mailing address editing for voter in the wrong precinct.

Looking at the layout of the page, I'm wondering if having both types of addresses and edits to both types of addresses will cause some weird layout change to the voter details page, because the left-side card gets so tall.

PR #6719, opened 2025-07-03
feat(fixtures): add API to make temp files/dirs

eventualbuddha:
## Overview

`vxsuite` has lots of temporary files and directories created in tests, and is inconsistent about cleaning them up. When an effort is made to clean them up, there are often per-package ways of doing so that are effective to varying degrees. This is a step toward unifying tmpfile handling.

The idea is that all tests that need to create temporary files will use these functions to do so rather than using `tmp` (or `tmp-promise`). Why not use `tmp`? It refuses to delete temporary directories recursively on process exit when `setGracefulCleanup` is enabled, instead requiring that you call `dirSync` with `unsafeCleanup: true` and then manually invoke the `removeCallback` method on the returned `DirEntry`. This makes creating temporary files and directories in tests somewhat painful and encourages improper use such as just not cleaning them up or [big hammers like `rm -rf /tmp/mock-print-job-*`, which causes flaky tests](https://github.com/votingworks/vxsuite/pull/6702).

The next step after this is to integrate the `tmpdir` setup/teardown hooks from `@votingworks/fixtures` in any packages that create temporary files/dirs and replace any uses of `fileSync`/`dirSync`/`tmpNameSync` with the functions added here.

## Demo Video or Screenshot

```ts
import { makeTemporaryFile } from '@votingworks/fixtures';

test('a thing', () => {
  // creates a file in `/tmp` and automatically cleans it up on process exit
  const path = makeTemporaryFile({ content: 'some data!' });
  expect(fs.readFileSync(path, 'utf8')).toEqual('some data!');
});
```

## Testing Plan
Added new automated tests.

### DIFF
--- a/libs/fixtures/src/tmpdir.test.ts
+++ b/libs/fixtures/src/tmpdir.test.ts
@@ -1,8 +1,14 @@
+import { Buffer } from 'node:buffer';
+import { existsSync, readFileSync, statSync } from 'node:fs';
 import { afterEach, expect, test } from 'vitest';
 import {
   clearTemporaryRootDir,
   getPathForFile,
   getTemporaryRootDir,
+  makeTemporaryDirectory,
+  makeTemporaryFile,
+  makeTemporaryFileAsync,
+  makeTemporaryPath,
   setTemporaryRootDir,
   setupTemporaryRootDir,
 } from './tmpdir';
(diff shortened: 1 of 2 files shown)


jonahkagan (approved):
Do you think it would be useful to restrict the usage to tests?

eventualbuddha:
> Do you think it would be useful to restrict the usage to tests?

I _think_ this is addressed with the existing `import/no-extraneous-dependencies` ESLint rule, but let me know if you disagree.

CleanShot 2025-07-03 at 14 37 21@2x

Issue #6160, opened 2025-03-26
VxPollBook: Allow Entering Address Fractions


adghayes:
When entering an address number in address updates or new registrations, we currently allow suffixes (e.g. 16B) but we must also allow fractions 1/4, 1/2, and 3/4. Only those. When a fraction is included, it should be parsed into the specific "house fraction" field. 

Users may try to enter fractional addresses as decimals - we should either parse these as fractions or disallow them from doing so (preventing periods) - whichever is easier.

We should also check we're rendering address fractions properly everywhere.

adghayes:
The pattern we should expect is house number, then fraction, then letter. So "16 1/2" or "16A" or "16 1/2A". To simplify, we do not need to accept also "16A 1/2." This is how these rare addresses are typically represented by the postal service.

PR #6720, opened 2025-07-03
Support Address Fractions

carolinemodic:
## Overview
https://github.com/votingworks/vxsuite/issues/6160

Supports address fractions in addition to street number and suffix in the street number field. You must enter a fraction if there is one before a suffix. Decimals like suggested in the ticket are not supported. See address_input_group tests for all valid formats. We were being somewhat inconsistent in address displays so the new format which is standerdized in frontend and backend for domicile and mailing address is : 

[StreetNumber][Fraction with a prepended space if there is a fraction][suffix with no space][space][streetname][apartment number prepended with a space if present]

## Demo Video or Screenshot

https://github.com/user-attachments/assets/f4131263-bf22-4a2e-9ac7-1ab37fb61621

## Testing Plan
See manual testing above. Added extensive tests for the parsing function. made sure to at least have some coverage of showing fraction vs. fraction + suffix vs suffix vs none in various frontend and backend receipt snapshot tests but probably did not cover every possible permutation in every situation. 

## Checklist

### DIFF
--- a/apps/pollbook/frontend/src/shared_components.tsx
+++ b/apps/pollbook/frontend/src/shared_components.tsx
@@ -228,9 +228,7 @@ export function VoterMailingAddress({
       <div>
         {voter.mailingStreetNumber}
         {prependSpaceIfNeeded(voter.mailingHouseFractionNumber)}
-        {voter.mailingSuffix}
-        {}
-        {voter.mailingStreetName}
+        {voter.mailingSuffix} {voter.mailingStreetName}
         {prependSpaceIfNeeded(voter.mailingApartmentUnitNumber)}
       </div>
       {voter.mailingAddressLine2 === '' ? null : (
--- a/apps/pollbook/frontend/src/voter_confirm_screen.test.tsx
+++ b/apps/pollbook/frontend/src/voter_confirm_screen.test.tsx
@@ -348,7 +348,7 @@ test('displays voter with mailing address change - no previous mailing address',
     mailingAddressChange: {
       mailingStreetNumber: '456',
       mailingStreetName: 'OAK AVE',
-      mailingSuffix: '',
+      mailingSuffix: 'B',
       mailingApartmentUnitNumber: '#2B',
       mailingHouseFractionNumber: '',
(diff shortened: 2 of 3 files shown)


apps/pollbook/frontend/src/address_input_group.tsx
```diff
@@ -42,13 +42,66 @@ function findDetailsFromStreetAddress(
...
 } {
-  const match = input.match(/(\d+)(.*)/);
-  const [, streetNumber = '', streetSuffix = ''] = match ?? [];
-  return { streetNumber, streetSuffix };
+  // First, try to match the complete valid pattern
+  // This regex handles multiple cases:
+  // - Street number only: "16"
+  // - Street number + suffix: "16B" or "16 B"
+  // - Street number + fraction: "16 1/2" or "161/2"
+  // - Street number + incomplete fraction: "16 1/"
```
kshen0, line 56: Why do we want to support incomplete fractions?

apps/pollbook/frontend/src/shared_components.tsx
```diff
@@ -9,6 +9,13 @@ import { throwIllegalValue } from '@votingworks/basics';
 import { Election } from '@votingworks/types';
 import { Column } from './layout';
 
+export function prependSpaceIfNeeded(text?: string): string {
+  if (!text) {
+    return '';
+  }
+  return ` ${text}`;
+}
+
```
kshen0, line 18: Can this be used by `receipt_helpers` instead of redefining?

kshen0 (approved):
I think you reuploaded the video from the previous PR but tests look like they cover expected behavior

apps/pollbook/frontend/src/address_input_group.tsx
carolinemodic: because the input has to keep workign as you are in the process of typing it out

carolinemodic:
oops correct video for @adghayes 

https://github.com/user-attachments/assets/bee70c18-036d-4132-ae18-44b8714ad5fb

PR #6728, opened 2025-07-07
VxAdmin: fix extension filtering

adghayes:
## Overview

For our manual imports on VxAdmin, other than the election package, we are not actually passing the right arguments to electron to filter out files of the wrong extension. The incorrect instances were corrected to match the correct instances. Tested manually.


arsalansufi (approved):
Good catch!

PR #6746, opened 2025-07-09
VxDesign: Consolidate org name fields

jonahkagan:
## Overview

Auth0 orgs have two name fields:
- name (a slug, e.g. `"votingworks"`)
- display name (the actual name, e.g. `"VotingWorks"`)

We only need the display name, so instead of tracking both fields in the app, just use a single `name` field, but populate it with the display name.
## Demo Video or Screenshot
N/A

## Testing Plan
Updated existing tests
## Checklist


nikhilb4a (approved):
Nice simplification for our use case

PR #6753, opened 2025-07-10
Rename HWTA feature flag and add new flag for internal functions

arsalansufi:
## Overview

This PR renames the HWTA feature flag and, more notably, adds a new flag for functions only needed internally, e.g., saving images and image analysis output to the USB drive.

We want to keep the HWTA as minimal as possible for external use by test labs is to ensure the greatest chance of success during hardware tests. Actions like printing more often or writing more data to the USB drive than necessary all run risks, e.g., by affecting EMI output or making the hardware more likely to fail.

If this flag becomes cumbersome to maintain, we can definitely revisit. I just need some way to toggle for upcoming hardware testing this month.

## Testing Plan

- [x] Tested manually


apps/scan/backend/src/electrical_testing/tasks/print_and_scan_task.ts
```diff
@@ -70,30 +74,42 @@ export async function runPrintAndScanTask({
...
+            usbDriveStatus.mountPoint,
+            'ballot-images'
+          );
+          await mkdir(usbDriveExportDirectory, { recursive: true });
+
+          await saveSheetImages({
+            sheetId,
+            ballotImagesPath: usbDriveExportDirectory,
+            images: scannerEvent.images,
+          });
```
arsalansufi, line 103: I recognize that, for internal use, we're now both writing to disk and the USB drive if connected. I think that's fine. Made the code and functionality a little bit easier to reason about when set up that way

eventualbuddha (approved):

PR #6755, opened 2025-07-10
VxDesign: Log the outcome of API calls

jonahkagan:
## Overview

Instead of logging the API call when it's received (before the method handler is called), we log the API call after its handled and include the outcome (success or failure) as well as any error that occurred.

## Checklist


jonahkagan:
This approach ended up being fancier than necessary and a bit hard to reason about. Closing in favor of https://github.com/votingworks/vxsuite/pull/6786

PR #6756, opened 2025-07-10
chore: upgrade `pdfjs-dist` to improve PDF compatibility

eventualbuddha:
## Overview

This version of `pdfjs-dist` changed in some important ways:
- It specifically does not want a `Buffer`. Instead, it wants `Uint8Array`. This is annoying insofar as `Buffer` is a subclass of `Uint8Array`, so we do not get compile-time errors from TypeScript but do get runtime errors if we mess it up. I'm not sure how best to deal with this.
- It consumes the `Uint8Array` data because it is moved to the worker in the frontend scenario. For consistency, NodeJS environments work the same way. Therefore we need to copy the array if we want to retain it. Similar to the above, there's no way to have TypeScript flag improper uses of this at compile time. Instead, we get some incorrect behavior when trying to use the now-empty array. The only way to fix this is to pay the price of copying the data every time we pass an array to the `getDocument` function from `pdfjs-dist`.

## Demo Video or Screenshot
There are very small difference in the rasterization of PDF to images, but in practice it's <0.01% of pixels.

Screenshot-VxAdmin-2025-07-10T20:23:11 997Z

## Testing Plan
- [x] Tested the report preview in VxAdmin manually.

### DIFF
--- a/apps/admin/frontend/.eslintignore
+++ b/apps/admin/frontend/.eslintignore
@@ -1,4 +1,5 @@
 /build
+/public
 /config
 /coverage
 /prodserver


jonahkagan (approved):
Thanks for tackling this!

PR #6769, opened 2025-07-10
feat(hmpb): generate timing mark grid-only PDFs

eventualbuddha:
## Overview

We'd like to start testing scanner hardware using paper with a timing mark grid printed on them instead of blank sheets. We could use ballots, but that's overkill. So now we can print paper with only timing marks.

## Demo Video or Screenshot
timing-mark-grid-only-p1

## Testing Plan
Tests verify that the generation code produces PDFs matching the ones checked in.


libs/hmpb/src/ballot_templates/timing_mark_only_template.tsx
jonahkagan, line 1: Small file organization nit: I think we should consider moving this out of `ballot_templates` and keep only real ballot templates in there. This feels more akin to the all bubble ballot as a variant only used for specific test circumstances - maybe those could be grouped in the file structure instead

libs/hmpb/src/generate_fixtures.ts
```diff
@@ -123,6 +127,28 @@ async function generateNhGeneralElectionFixtures(renderer: Renderer) {
...
+  assert(specDir.startsWith(`${outputDir}/`));
+  await rm(specDir, {
+    recursive: true,
+    force: true,
+  });
+  const generated = await timingMarkGridOnlyFixtures.generate(renderer, {
+    paperSize,
+  });
+  const pdfPath = join(outputDir, specPaths.pdf);
+  assert(pdfPath.startsWith(`${outputDir}/`));
```
jonahkagan, line 147: I noticed you've started using `node:assert` instead of the one from `@votingworks/basics`. Any reason why?

jonahkagan (approved):

Issue #6150, opened 2025-03-26
VxPollBook: Primary Support


adghayes:
**Check-In Requirements**

- For a general election, no change in behavior. 
- For party affiliated voters in a primary election, no change in behavior
- For an undeclared voter in a primary election:
  - The check-in page presents additional buttons for "Democratic" vs. "Republican." Something that takes a single click (rather than a dropdown) is faster / better
  - The check-in receipt looks as if they were registered for the party they just chose because legally, they are. For an undeclared voter getting a Republican ballot, their receipt would just show "Republican"

In terms of positioning, I was thinking something like this (ignore copy):

<img width="982" alt="Image" src="https://github.com/user-attachments/assets/8e014c90-e218-4e43-8c65-12a56caaca33" />

Could use both buttons as outlined primary and when one is selected it becomes filled primary. I don't know if we have a design precedent for this elsewhere in VxSuite. Once the button is selected, "Confirm Check-In" is enabled where it was previously disabled.

 **Export Requirements**

The party choice should be saved for every voter, just as the OOS DL information is.

The voter history export should include the ballot choice of the voter in a new column, "Party Choice." For consistency, we just leave this column in for generals as well.

**Statistics Requirements**

See #6191.

adghayes:
noting that we're pivoting the UX to make the party selection on a separate screen, per [discussion on Slack](https://votingworks.slack.com/archives/C07CR6T1TLK/p1752107349529299?thread_ts=1751923184.132559&cid=C07CR6T1TLK)

adghayes:
Closed by #6780, follow-ups tracked by separate tickets.

PR #6780, opened 2025-07-12
Support party selection during primary election check-in flow

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/6150

Adds support for primary elections to pollbook.

* Requires party selection for Undeclared voters in check-in flow for primary elections
  * Voters with a party declared don't need to select party, but their party on file will appear on the details page for the check-in
* Includes party selection column in voter history export
   * For generals the column is still there, but is blank
  
## Demo Video or Screenshot

Screenshot 2025-07-11 at 6 30 12 PM

Screenshot 2025-07-11 at 6 30 16 PM

Screenshot 2025-07-11 at 6 30 20 PM

**Existing check in for a voter with declared party, unchanged**

https://github.com/user-attachments/assets/0112be1d-2c46-494f-998b-c1b884a61e06

**Check in where party must be selected (bonus OOS DL demo)**

https://github.com/user-attachments/assets/9b505ea6-03c3-402b-b714-dccf61fb635c

**Back button**

https://github.com/user-attachments/assets/91a08880-f7c6-4623-8973-dbbea7de88d4

## Testing Plan
- added tests for new flow
- added an e2e test for new pollworker screen

## Checklist


adghayes:
@kshen0 thank you for the multiple videos, flows look on point.

One nit UI request - on the check-in details in the voter details, please use the full party names specifically "Democratic" or "Republican."

apps/pollbook/backend/src/app.ts
```diff
@@ -266,13 +267,36 @@ function buildApi({ context, logger, barcodeScannerClient }: BuildAppParams) {
     async checkInVoter(input: {
       voterId: string;
       identificationMethod: VoterIdentificationMethod;
+      ballotParty: PartyAbbreviation;
```
kshen0, line 270: The behavior as written is that all check-ins require a party, even generals, but we only validate and display those for primaries.

This is a little off semantically - what does it mean to have a party selection in a general? But it was a little easier to require this across the board so that we get type safety and don't have to throw existence checks in a bunch of places. I'm still on the fence and may revisit this later to allow `ballotParty` IFF the election is a primary.

kofi-q (approved):
LGTM! (After product feedback's addressed)

Design note:
Thinking this selection model would be better served by the `RadioGroup` component in `libs/ui` - something feels off about the centred buttons here. Don't feel strongly enough about it to be a blocker though (and can always be a later iteration) but it's an option (cc @adghayes ).

Image

And just in case there's a possibility that any vertical ordering of parties could be ill-received, we could instead use the multi-column layout, maybe fill the whole width of the screen, if that doesn't look too weird:

Image

kshen0:
> @kshen0 thank you for the multiple videos, flows look on point.
> 
> One nit UI request - on the check-in details in the voter details, please use the full party names specifically "Democratic" or "Republican."

Updated:

Screenshot 2025-07-14 at 11 05 50 AM

adghayes:
@kofi-q fair suggestion around the radio group buttons. We'll ship the feature as-is but keep this in mind if we iterate on this page in the future. When we someday go to supporting more than two parties, might make sense to switch at that time.

My main takeaway from your comment is that it's another example of how it would be useful to get design input further upstream.

PR #6786, opened 2025-07-14
VxDesign: Log outcome of API calls

jonahkagan:
## Overview

Instead of logging the API call when it's received (before the method handler is called), we log the API call after its handled and include the outcome (success or failure) as well as any error that occurred.

## Demo Video or Screenshot
Example log line for `getUser` request when user is not logged in
```
[backend:run] 2025-07-14T22:23:22.791Z design-backend:app {
[backend:run]   source: 'vx-design-service',
[backend:run]   eventId: 'api-call',
[backend:run]   eventType: 'application-action',
[backend:run]   user: 'system',
[backend:run]   message: undefined,
[backend:run]   disposition: 'failure',
[backend:run]   methodName: 'getUser',
[backend:run]   input: undefined,
[backend:run]   userAuth0Id: undefined,
[backend:run]   userOrgId: undefined,
[backend:run]   error: 'auth:unauthorized'
[backend:run] }
```

## Testing Plan
- Added new automated tests
- Manual test as shown above

## Checklist

### DIFF
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -163,60 +163,70 @@ export function buildApi({ auth0, logger, workspace, translator }: AppContext) {
     requireOrgAccess(user, electionOrgId);
   }
 
-  const middlewares: Array<grout.Middleware<ApiContext>> = [
-    async function loadUser({ request, context }) {
-      const user = await auth0.userFromRequest(request);
-      if (!user) {
-        throw new AuthError('auth:unauthorized');
-      }
-      return { ...context, user };
-    },
-
-    // TODO(jonah): It might be nice to have a way to log the result of the API
-    // call, not just that the call was received. Potentially, we could pass
-    // middlewares a continuation callback so they could run code before and
-    // after the request handler.
-    async function logApiCall({ methodName, input, context }) {
-      await logger.logAsCurrentRole(
-        LogEventId.ApiCall,
-        {
-          methodName,
-          input: JSON.stringify(input),
-          userAuth0Id: context.user?.auth0Id,
-          userOrgId: context.user?.orgId,
-        },
-        debug
-      );
-    },
-
-    /**
-     * All methods should check authorization. This middleware checks
-     * authorization automatically for methods that have either `electionId` or
-     * `orgId` in their input. If a method doesn't have either of or has some
-     * other reason to handle authorization separately, it must be listed in
-     * `methodsThatHandleAuthThemselves` within this function.
-     */
-    async function checkAuthorization({ methodName, input, context }) {
-      if (input) {
-        assert(context.user);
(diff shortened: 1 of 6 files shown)


libs/grout/src/server.ts
```diff
@@ -114,9 +120,32 @@ export interface MiddlewareMethodCall<Context extends AnyContext> {
...
+) => Partial<Context> | void | Promise<Partial<Context> | void>;
+
+/**
+ * Collection of {@link BeforeMiddleware} and {@link AfterMiddleware} functions
+ * for an API.
+ */
+export interface Middlewares<Context extends AnyContext> {
+  before?: Array<BeforeMiddleware<Context>>;
+  after?: Array<AfterMiddleware<Context>>;
+}
```
nikhilb4a, line 147: I like this API! Very clear.

jonahkagan:
@nikhilb4a I forgot to convert this from a draft before originally requesting your review, so I still need your approval to merge

apps/design/backend/src/app.ts
```diff
@@ -163,60 +163,70 @@ export function buildApi({ auth0, logger, workspace, translator }: AppContext) {
...
+          'getUser',
+          'getUserFeatures',
+          'getAllOrgs',
+        ];
+        assert(methodsThatHandleAuthThemselves.includes(methodName));
+      },
+    ],
+
+    after: [
+      async function logApiCall({ methodName, input, context }, result) {
```
nikhilb4a, line 209: Thinking of a case where the api call results in a fatal crash of some kind for the server, would we lose any log of the call?

nikhilb4a (approved):

apps/design/backend/src/app.ts
jonahkagan: It's a good question.

My thoughts:
- We do wrap pretty much all of the logic for process an API call and sending the request in a `try/catch`, then run the after middlewares in a `finally`, so it would have to be some sort of crash that doesn't propagate up the call stack. Hopefully there aren't many of those, though I think it's certainly possible.
- Even in that case, we still have logging of the request at other levels of the web stack (e.g. the Heroku router). They will have slightly less info (e.g. no user auth data).
- I did consider doing a before and after log, but it felt like it might be overkill. We'd probably want to have a request ID or something like that to tie the two together.

I'm open to considering making some change in this regard if you think it's worthwhile. My gut says this is good enough until we see a counterexample in the wild that proves otherwise.

nikhilb4a: > Even in that case, we still have logging of the request at other levels of the web stack (e.g. the Heroku router). They will have slightly less info (e.g. no user auth data).

I agree with this being good for now, knowing that we have some trace of the request that includes the api method

PR #6796, opened 2025-07-15
[libs/printing] Fix printing with Uint8Array data

kofi-q:
## Overview

Fixing an issue in the print path where `Uint8Array` data is streamed as single-byte chunks when we shell out to `lpr`. This is due to a recent change to generate PDFs as `Uint8Array` data instead of `Buffer` - the latter is [handled differently by Node](https://nodejs.org/docs/latest-v20.x/api/stream.html#streamreadablefromiterable-options) and streamed as a single chunk when using `Readable.from()`. Converting `Uint8Array`s to `Uint8Array[]` before streaming to stdin to mimic that behaviour.

## Testing Plan
- Updated relevant tests to use `Uint8Array` instead of `Buffer`

### DIFF
--- a/libs/printing/src/printer/print.test.ts
+++ b/libs/printing/src/printer/print.test.ts
@@ -1,6 +1,5 @@
 import { beforeEach, expect, test, vi } from 'vitest';
 import { ok } from '@votingworks/basics';
-import { Buffer } from 'node:buffer';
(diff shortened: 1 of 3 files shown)


eventualbuddha (approved):
Thanks for fixing this.

PR #6806, opened 2025-07-16
VxAdmin: clear diagnostics on unconfigure

### LINKED ISSUE #6795: Clear VxAdmin diagnostics results when clearing the election definition for consistency with other machines
https://votingworks.slack.com/archives/C088THUKP2T/p1752604091218289?thread_ts=1752601614.645439&cid=C088THUKP2T

adghayes:
## Overview

Closes #6795. Moves VxAdmin over to the same unconfigure mental model as all the other apps - resets the whole database on unconfigure.

One thing this breaks (which is already broken on the other apps) is disk space tracking. Currently we have an overly involved disk space tracking pattern (my bad) where the store is actually tracking the maximum ever seen disk space available to the app and treating that as the maximum available space. The reason is to avoid the sort of situation where you get a new iPhone and you say "why is 15GB already being used?" But that situation is not really that bad and it gets our applications out of the business of tracking disk space over time.


arsalansufi (approved):
Thanks for tackling! Agreed on the decision to simplify disk space tracking

Issue #6789, opened 2025-07-14
VxMark: pre-printed ballot marking demo


mattroe:
Especially after the recent executive order, most of our target states are expressing a preference for a BMD that prints ballots that match the HMPB instead of QR code summary ballots. We want to be able to demo support for this capability in our upcoming demos to NH, LA, and OK in September. NH in particular wants to print ballots from pre-printed ballot stock. Other states may want to print on demand, but we can start with pre-printed ballot marking support for the purposes of a September demo.

To limit initial scope, the demo only needs to support printing:
- On a single-sheet letter-sized ballot produced in VxDesign
- From a HP 4001dn we've tested the initial integration against
- For one ballot orientation
- Where ballot sheets are loaded before marking into the expected tray (no infeed instructions necessary)

**Printing Variability**
I suspect there may be variability in print margins / offset between HP printers that impacts bubble printing placement. As part of this integration, we should:
1. test if different printers do indeed shift bubble printing placement
2. understand if the variability is large enough that we need to build some future printer calibration/offset setting into this feature

It might just work across printers and any variability is small enough that it doesn't impact mark interpretation!

**Ballot Sheet Insertion / Length**
We'll eventually need to support printing up to 22" ballots (our max support HMPB size). We'll do this via a custom HP tray & firmware: https://votingworks.slack.com/archives/C094XPFCX08/p1752002699311389

For the sake of the demo, we can assume that we will using letter sized ballots in the normal tray. We may want to eventually support inserting single sheets into the other tray as it's potentially easier for voters to interact for single sheet feeding.

PR #6810, opened 2025-07-17
[Mark] Add feature-flagged Print Mode toggle

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/6789

Adding a (feature-flagged for now) toggle for switch between ballot summary prints and bubble marks on pre-printed ballots. This is likely not the final location of the toggle - just needed somewhere to put it while we test.

Will be looking up the toggle state later in the print flow.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/f4361857-d164-4561-88f9-c3c50870edca

## Testing Plan
- Unit tests for added store/Api/client code.

## Checklist

  - Not yet, but will add some as we finalise the feature


apps/mark/backend/src/store.test.ts
```diff
@@ -77,3 +77,46 @@ test('reset clears the database', () => {
...
+    });
+
+    return store;
+  }
+
+  test('get - schema default', () => {
+    expect(newStore().getPrintMode()).toEqual<PrintMode>('summary');
+  });
+
+  test('set and get rountrip', () => {
```
kshen0, line 101: typo: "roundtrip"

apps/mark/backend/src/store.ts
```diff
@@ -24,6 +29,9 @@ import { join } from 'node:path';
 
 const SchemaPath = join(__dirname, '../schema.sql');
 
+// [TODO] Fits better in ./types.ts, but this avoids a dependency cycle for now:
```
kshen0, line 32: Kind of surprising that `types.ts` depends on `store.ts`. Is the goal to move all types to `types.ts` so that it imports nothing and everything imports `types.ts`?

kshen0 (approved):

apps/mark/backend/src/store.ts
kofi-q: > so that it imports nothing

+1 - that feels like a good rule of thumb to adopt. If importing any internal types into `types.ts`, they should be moved there instead

Issue #6762, opened 2025-07-10
VxPollBook: Barcode Scanning - Handle AAMVA "NONE"/"unavl" values in Middle Name


adghayes:
From the AAMVA card design standard:

> Mandatory data elements for which no data exists for a given cardholder are to be encoded with the word "NONE".
In the event data is not available for a mandatory data element, "unavl" is to be encoded. 

The middle name is required, so we should expect to see "NONE" in cases where a person has no middle name. Let's update the barcode parser to count "NONE" _and_ "unavl" as an empty string for the middle name.

Issue #6737, opened 2025-07-08
VxPollBook: fix invisible middle name & suffix search parameter bug


adghayes:
address the problem described by @kshen0 [here](https://votingworks.slack.com/archives/C07CR6T1TLK/p1751064664135039?thread_ts=1750785008.365259&cid=C07CR6T1TLK) with [the solution](https://docs.google.com/document/d/1pmkZO2LN1qCdxMgr06tnvr4M4h2geUEObOoK7xyPz4A/edit?tab=t.gljx81y3nu73#heading=h.op4z0hh61wwl) agreed upon [here](https://votingworks.slack.com/archives/C07CR6T1TLK/p1752013098935379?thread_ts=1752007004.161979&cid=C07CR6T1TLK)

PR #6813, opened 2025-07-18
[pollbook] add intermediate search state for ID scans

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/6737

Adds an intermediate search state to the voter search flow. When middle name or suffix are included in search params (as is only possible when scanning and ID) we show an uneditable input with the full name and an "Edit Search" button. Clicking the button returns the screen to the existing first/last name inputs and deletes the middle name/suffix fields from the search param.

After this change, the editable search inputs will always reflect exactly what is being searched - no more hidden search params.

Also addresses a related bug in the daemon: https://github.com/votingworks/vxsuite/issues/6762
* Expected values of "NONE" or "unavl" are parsed to empty string so they don't show up in the frontend

## Demo Video or Screenshot

**Display-only search input, no results**

Screenshot 2025-07-18 at 12 15 05 PM

**Display-only search input, >0 results**
Screenshot 2025-07-18 at 12 16 13 PM

Recommend watching these with sound on to hear the beep when scans happen.

**Scan with no matches at all**

https://github.com/user-attachments/assets/a0e2d38c-9339-408f-84dd-b39ff3db5d35

**Scan with no matches, but partial matches exist**

https://github.com/user-attachments/assets/5cb79bc1-8755-456e-8bbe-b71a83d327a1

**Scan with exactly one match, partial matches exist**

https://github.com/user-attachments/assets/7095f75c-b66a-40cf-a251-b789d42dd5e1

**Scan with >1 exact match**

https://github.com/user-attachments/assets/e3f6f6cf-cb4c-4339-9c07-1118302a8502

## Testing Plan

- manually tested
- added component tests for intermediate state

## Checklist

### DIFF
--- a/apps/pollbook/frontend/src/voter_search_screen.test.tsx
+++ b/apps/pollbook/frontend/src/voter_search_screen.test.tsx
@@ -147,7 +147,7 @@ test('after an ID scan with "hidden" fields, shows full name and "Edit Search" b
 
   await act(() => vi.advanceTimersByTime(DEFAULT_QUERY_REFETCH_INTERVAL));
 
-  expect(onBarcodeScanMatch).toHaveBeenCalledOnce();
+  await vi.waitFor(() => expect(onBarcodeScanMatch).toHaveBeenCalled());
   expect(onBarcodeScanMatch).toHaveBeenCalledWith(mockVoter);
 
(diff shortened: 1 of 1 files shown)


kofi-q (approved):
🚢

PR #6815, opened 2025-07-18
[hmpb] Add generateMarkOverlay() for marking pre-printed ballots

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/6789

Initial pass at a utility for generating HMPB bubble mark overlay PDFs. Will be using this in VxMark to mark pre-printed ballot sheets.

Currently relies on measurements from within VxSuite code and grid layouts in the election definition - future support for marking 3rd party ballots may require including a bit more information in the election definition, like the location of a corner timing mark, but we currently have everything we need to print on Vx ballots.

### TODO
- Just have a snapshot test in place for now, but planning on an interpretation test as well for a little more coverage.
- Based on what we learn from tying this out on different printers, will be adding support for including calibrated offset adjustments in the inputs to the function.

## Demo Video or Screenshot
Physical prints for the ballot fixture used in the unit test:

| Front | Back |
| -- | -- |
| IMG_5064 Large | IMG_5065 Large |
| IMG_5066 Large | IMG_5067 Large |
| IMG_5068 Large | IMG_5069 Large |

## Testing Plan
- Snapshot and manual test for now

## Checklist


libs/hmpb/src/marking.ts
```diff
@@ -0,0 +1,199 @@
...
+] as const;
+
+const timingMarkCount = [34, 41] as const;
+const timingMarkSize = [
+  TIMING_MARK_DIMENSIONS.width * IN,
+  TIMING_MARK_DIMENSIONS.height * IN,
+] as const;
+
+const bubbleSize = [BUBBLE_WIDTH_PX * PX, BUBBLE_HEIGHT_PX * PX] as const;
+// We grow the mark slightly to help make up for slight variations in printing:
```
kshen0, line 38: Does this overfill the bubble? Is that preferable, either for aesthetics or interpretation, to slightly underfilling the bubble?

libs/hmpb/src/marking.ts
```diff
@@ -0,0 +1,199 @@
...
+
+const writeInFontName = 'Roboto-Bold';
+const writeInFontSizeDefault = 12;
+const writeInFontSizeReduced = 10;
+
+/**
+ * Center of the top-left timing mark.
+ *
+ * [TODO] We may need to move this into the election definition, to support
+ * marking 3rd part ballots
```
kshen0, line 51: nit: "party"

Also, is this something we'll need to manually calibrate for each supported 3rd party ballot?

libs/hmpb/src/marking.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    const mark = markInfo(contestVotes, pos);
+    if (!mark) continue;
+
+    doc.switchToPage(pageNumber - 1); // Pages are 0-indexed in `pdfkit`.
+
+    const bubbleCenter = [
+      gridOrigin[0] + gridSize[0] * (pos.column / (timingMarkCount[0] - 1)),
+      gridOrigin[1] + gridSize[1] * (pos.row / (timingMarkCount[1] - 1)),
+    ];
```
kshen0, line 116: Is bubble center established by convention? Is there an easy way to share bubble render position logic between this renderer and the renderer for the base layer unmarked HMPB?

kofi-q: Oh yeah, fair point - I was following the logic in the [renderer](https://github.com/votingworks/vxsuite/blob/f08c6b60e5abe542642138a980c013964a600707/libs/hmpb/src/render_ballot.tsx#L320). The different coordinate spaces might complicate a bit, but I could see value in unifying more of the logic/calculations across print/mark/interpret where possible. Will see what's feasible in future iterations

kshen0 (approved):

libs/hmpb/src/marking.ts
kshen0: Hmm yeah seems like we could benefit from some shared logic especially since we've been through a few ballot layout reworks already. But agree it's probably too early to nail that down now.

libs/hmpb/src/marking.ts
kshen0: Given our mark thresholds seems like an underfill would be fine, but if overfilling doesn't mess with interpretation I agree it's better aesthetically / inspires more confidence

Issue #6343, opened 2025-04-24
Contest adjudication - implement class/hook to better manage vote and write-in state


nikhilb4a:
See PR conversation [here](https://github.com/votingworks/vxsuite/pull/6210#discussion_r2055012396)

Tentatively planned for clean up phase of adjudication project after marginal mark implementation.

PR #6816, opened 2025-07-21
Suggestions for #6343

eventualbuddha:
Some suggested changes for #6343:
- Small style warning fixes.
- Use `Map` instead of `Record` to make the types a bit stricter and make this new code less susceptible to failing with option IDs like `toString`.


nikhilb4a:
Sorry it took me so long to get to this, but thanks for the suggestion. Are there cases where you use a Record, or do you find yourself using Maps the majority of the time?

nikhilb4a (commented):
Good reminder about the shortfall of Record with certain keys!

Issue #6818, opened 2025-07-21
VxPollBook: disable changing adddress for voters outside of current precinct


adghayes:
This is not a legal flow in NH.

PR #6821, opened 2025-07-21
Update disabled/enabled actions for wrong precinct and inactive voters

carolinemodic:
## Overview
Updates the following
- When a voter is in the wrong precinct we now disable the update address buttons 
- When a voter is flagged inactive fixes a bug that was preventing check ins on the backend. The feature was originally written so that check ins would be blocked, and specific efforts were went to to make sure that even in rare offline syncing scenarios an illegal check in could not sneak in. When Drew then updated the logic to ALLOW for check ins on inactive voters some of those backend checks were missed and we were silently ignoring these checks in, this fixes that. 
- When a voter is flagged as inactive enable all other actions, see discussion: https://votingworks.slack.com/archives/C07CR6T1TLK/p1753135409790919
- 
https://github.com/votingworks/vxsuite/issues/6818

## Demo Video or Screenshot
The view when a voter is both checked in and inactive on the EM screen looks a little weird as there are two headers in the right-hand box. This is an edge case though so in lieu of a better idea I think its ok. 

https://github.com/user-attachments/assets/687497cb-8601-4c9e-974b-0b59cc6396bc

## Testing Plan
Manual testing, updated automated tests.

### DIFF
--- a/apps/pollbook/backend/src/event_helpers.ts
+++ b/apps/pollbook/backend/src/event_helpers.ts
@@ -162,12 +162,6 @@ export function applyPollbookEventsToVoters(
           debug('Voter %s not found', event.voterId);
           continue;
         }
-        // If we receive a check in event for a voter that was marked as inactive it should not be processed.
-        // This can only occur if there are offline machines or in rare race conditions. Inactivated voters can
-        // not be checked in.
-        if (voter.isInactive) {
-          continue;
-        }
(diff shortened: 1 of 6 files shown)


adghayes:
For the inactive and checked-in states, the checked-in state should just override the inactive state. If they were checked in, they aren't inactive after all.

adghayes:
Can you make that tweak @carolinemodic ?

PR #6822, opened 2025-07-22
fix(scan): prevent losing image data in data race

eventualbuddha:
## Overview

We buffer the incoming packets from the scanner and, if a packet cannot be handled by the code currently reading packets, put unhandled packets into a list to be handled later. However, it's possible that the list may include responses to requests that we're no longer able to handle for whatever reason (e.g. a request timed out, but the response did eventually come). Thus, when we make outgoing requests, because we only expect there to be a single request/response communication happening at a time with the scanner, we discard any unhandled/stale responses. We excluded events from this because those are not in response to a request and should not be discarded simply because we're going to make a request. However, this code had a bug and was discarding image data, which led to a situation where we would try to process the scanned image data and see no data to process. This only happened if we happened to try to send a request to the scanner during a scan, which is unlikely to happen in production due to the fairly rigid structure of our scanning state machine.

The fix for this is relatively simple: make sure to keep image data when discarding stale responses. However, in order to test this with the scanner and verify that it was fixed manually, I needed to make other changes to remove the blocking IO code that the "client" side of the client/scanner pair was using. We'd switched over to using async with tokio when switching from rusb to nusb, but the other half was still using blocking IO. Mixing blocking and non-blocking IO must be done carefully and is easy to get wrong, so it's simpler to just use non-blocking IO (i.e. async) everywhere.

## Demo Video or Screenshot
Debug log showing properly handled image data packets.

```
[backend:run]   2025-07-22T00:02:36.956574Z DEBUG pdictl: PACKET: BeginScanEvent
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.956579Z DEBUG pdictl: sending event: ScanStart
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:245
[backend:run]
[backend:run]   2025-07-22T00:02:36.956584Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 1048576 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]
[backend:run] 2025-07-22T00:02:36.957Z pdi-scanner received: { response: 'ok' }
[backend:run] 2025-07-22T00:02:36.957Z pdi-scanner received: { event: 'scanStart' }
[backend:run] {"source":"vx-scan-backend","eventId":"info","eventType":"system-status","user":"system","message":"Setting electrical testing status message","disposition":"na","component":"scanner","statusMessage":"Received event: scanStart"}
[backend:run] {"source":"vx-scan-backend","eventId":"info","eventType":"system-status","user":"system","message":"Setting electrical testing status message","disposition":"na","component":"scanner","statusMessage":"Ejected document to re-scan"}
[backend:run] 2025-07-22T00:02:36.974Z pdi-scanner pdictl stderr:   2025-07-22T00:02:36.956690Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 1048576 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.956793Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 1048576 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.956923Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 1048576 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.957082Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 1048576 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.957272Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 1048576 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.957442Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 1048576 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.957607Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 1048576 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.957750Z DEBUG pdictl: PACKET: ImageData(ImageData { data: 53248 bytes })
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
[backend:run]
[backend:run]   2025-07-22T00:02:36.957764Z DEBUG pdictl: PACKET: EndScanEvent
[backend:run]     at libs/pdi-scanner/src/rust/main.rs:445
```

## Testing Plan
Tested with the real VxScan app and the VxScan HWTA, which exhibited the bug more reliably.

### DIFF
--- a/libs/pdi-scanner/examples/get_test_string.rs
+++ b/libs/pdi-scanner/examples/get_test_string.rs
@@ -4,6 +4,7 @@
 
 use clap::Parser;
 use std::time::Duration;
+use tokio::time::timeout;
 use tracing_subscriber::prelude::*;
 
 use pdi_scanner::connect;
@@ -57,7 +58,7 @@ async fn main() -> color_eyre::Result<()> {
     for n in 1..=config.times {
         println!(
             "{n:02}: get_test_string result: {:?}",
-            client.get_test_string(Duration::from_secs(1))
+            timeout(Duration::from_secs(1), client.get_test_string()).await
         );
     }
 
--- a/libs/pdi-scanner/examples/scan_forever.rs
+++ b/libs/pdi-scanner/examples/scan_forever.rs
@@ -1,19 +1,16 @@
 use clap::Parser;
 use std::{
     path::PathBuf,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
     time::{Duration, Instant},
 };
+use tokio::time::timeout;
 use tracing_subscriber::prelude::*;
 
 use pdi_scanner::{
     connect,
     protocol::{
         image::{RawImageData, Sheet, DEFAULT_IMAGE_WIDTH},
-        packets::Incoming,
+        packets::{ImageData, Incoming},
         types::{
             ClampedPercentage, DoubleFeedDetectionMode, EjectMotion, FeederMode, ScanSideMode,
         },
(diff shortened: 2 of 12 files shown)


libs/pdi-scanner/examples/get_test_string.rs
```diff
@@ -57,7 +58,7 @@ async fn main() -> color_eyre::Result<()> {
     for n in 1..=config.times {
         println!(
             "{n:02}: get_test_string result: {:?}",
-            client.get_test_string(Duration::from_secs(1))
+            timeout(Duration::from_secs(1), client.get_test_string()).await
```
eventualbuddha, line 61: This is an example of one of the benefits of moving to async: we no longer need to take a `Duration` parameter on each of the methods that expect a response. Instead, we can just use `tokio`’s normal timeout mechanism.

libs/pdi-scanner/examples/scan_forever.rs
```diff
@@ -75,98 +72,103 @@ async fn main() -> color_eyre::Result<()> {
...
-        }
-    })?;
     let mut start: Option<Instant> = None;
 
     loop {
-        if !running.load(Ordering::SeqCst) {
-            eprintln!("exiting");
-            break;
-        }
+        tokio::select! {
```
eventualbuddha, line 106: This makes it so we don’t have to busy loop like we did before. Now we simply await multiple `Future`s and `tokio::select!` picks whichever one is ready first. Also means we can drop the `ctrlc` dependency.

libs/pdi-scanner/src/rust/client.rs
```diff
@@ -42,35 +38,12 @@ pub struct ImageCalibrationTables {
     pub back_black: Vec<u8>,
 }
 
-macro_rules! recv {
```
eventualbuddha, line 45: I removed the macros because they were less needed now that the timeout code is removed. It does make the methods a bit more verbose, but hopefully also more easily comprehensible.

libs/pdi-scanner/src/rust/protocol/packets.rs
```diff
@@ -14,7 +14,7 @@ pub(crate) const PACKET_DATA_START: &[u8] = &[0x02];
 pub(crate) const PACKET_DATA_END: &[u8] = &[0x03];
 
 /// All possible commands that can be sent to the scanner.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
eventualbuddha, line 17: Needed for the tests to enable `assert_eq!`.

libs/pdi-scanner/src/rust/protocol/parsers.rs
```diff
@@ -1366,7 +1366,7 @@ mod tests {
     fn test_packet_with_crc() {
         let input = b"\x02D\x03\xb4";
         let (remainder, (tag,)) = packet_with_crc((tag(b"D"),))(input).unwrap();
-        assert_eq!(remainder, []);
+        assert_eq!(remainder, b"");
```
eventualbuddha, line 1369: This started failing due to having multiple implementations that could apply to the other version. This uses a more specific type to avoid the problem.

libs/pdi-scanner/src/rust/client.rs
```diff
@@ -126,57 +79,81 @@ impl<T> Client<T> {
...
+    // delay sending a response to a command (e.g. when the "eject pause"
+    // feature pauses the scanner, it will queue commands received during that
+    // time).
+    fn clear_unhandled_solicited_packets(&mut self) {
+        let mut unhandled_packets = VecDeque::with_capacity(self.unhandled_packets.len());
+        std::mem::swap(&mut unhandled_packets, &mut self.unhandled_packets);
+
+        let (unsolicited_packets, solicited_packets): (Vec<_>, Vec<_>) = unhandled_packets
+            .into_iter()
+            .partition(|packet| packet.message_type().is_unsolicited());
```
eventualbuddha, line 107: Here is the updated code where we filter packets. It’s done a little differently than the previous version in that we only do the filter/partition once now and use `std::mem::swap` to pull the data out of `self`, then put it back below with `VecDeque::extend`.

PR #6832, opened 2025-07-23
[Mark] Add temporary print calibration controls for testing/demo prep

kofi-q:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6789

When we flesh out the ballot marking feature, we're going to have a flow for calibrating individual printers against pre-printed ballots, which will likely involve setting x/y offsets (and maybe, if there's a need, scaling up/down the grid).

We don't need that all in place for demos, but there's a chance we'll need to adjust offsets slightly as we test/prep. Adding some temporary controls for making those adjustments to the Election Manager screen for now. These show up when the "Bubble Marks" print mode is selected (which is only visible when the `REACT_APP_VX_MARK_ENABLE_BALLOT_PRINT_MODE_TOGGLE` feature flag is enabled).

## Demo Video or Screenshot

https://github.com/user-attachments/assets/51a24abc-07d9-4a95-8453-d72278b57c1b

## Testing Plan
- Mostly manual for now  - short-lived component. We'll likely have a proper calibration flow when we build out the full feature, so most of this will be changing.

## Checklist

### DIFF
--- a/apps/mark/backend/schema.sql
+++ b/apps/mark/backend/schema.sql
@@ -12,6 +12,14 @@ create table election (
   created_at timestamp not null default current_timestamp
 );
 
+-- Temporary dev table:
+create table print_calibration (
+  -- enforce singleton table
+  id integer primary key check (id = 1),
+  offset_mm_x real not null,
+  offset_mm_y real not null
+);
+
 create table system_settings (
   -- enforce singleton table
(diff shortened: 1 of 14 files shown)


eventualbuddha (approved):
Thanks for the video!

PR #6833, opened 2025-07-23
clean up avahi services after tests

carolinemodic:
When the pollbooks announce to the network that they exist for discovery by other pollbooks it requires calling and leaving open this process with avahi. The backend index.ts file when running the app either in dev or in production will close this thread on exit but the tests are operating slightly downstream from that logic and need to remember to clean this up at the end of their runs.


carolinemodic:
This appears to clean up the processes for all tests, its making me rethink how the withManyApps tests are working in the first place with this so I will look into that on a seperate branch

Issue #6151, opened 2025-03-26
VxPollBook: Logging


adghayes:
- [ ] check-in
- [ ] address change
- [ ] name change
- [ ] new registration
- [ ] authentication
- [ ] peripheral connection
- [ ] exports
- [ ] network stuff
- [ ] configuration
- [ ] receipt prints
- [ ] barcode scans 
- [ ] undo check-ins
- [ ] absentee mode switching
- [ ] paper backup events

As far as logging voter information, we need to check what's the best here. TODO: @adghayes 

As far as networked events go, I think it makes sense that all events received are logged and there is just some indication in the log that it was on another machine vs. the current.

PR #6845, opened 2025-07-24
VxPollbook Logging

carolinemodic:
## Overview
Adds a set of comprehensive logs addressing the required points in: https://github.com/votingworks/vxsuite/issues/6151 to VxPollbook. 

Adds middleware logging to the local app grout endpoints to cover most user actions, some get endpoints that are polled by the frontend are silenced intentionally, only certain input parameters that do not contain PII are logged. 

Adds additional logging to the following areas:
- Networking - when pollbook connection states change, or when 1 or more event is synced from another node
- Configuration - when loading pollbook package from usb or the network 
- Backup Worker - basic logging when it exports backups
- Print actions - logging already occurs for connection changes (with real hardware, not mocks) 

The barcode scanner already has sufficient logging. I plan to do a final round of testing this on a real image with real hardware sometime next week to make sure all hardware/connection logging I expect works in a real environment. 

Does not yet look into the optional rabet-v items

https://github.com/votingworks/vxsuite/issues/6151

## Demo Video or Screenshot
Non-exhaustive sample logs. Machine 0001 is configured via usb, Machine 0002 is configured via EM log in when on the network with machine 0001. Then some voter actions are performed on both machines. 

[machine0001.log](https://github.com/user-attachments/files/21416583/machine0001.log)
[machine0002.log](https://github.com/user-attachments/files/21416582/machine0002.log)

## Testing Plan
Manual testing and viewing of logs


adghayes:
I created follow-up tickets for the two RABET-V logging tickets:
- #6846 
- #6847

PR #6855, opened 2025-07-28
VxDesign feature flag cleanup

arsalansufi:
## Overview

I initially opened this PR to set `ENABLE_BMD_BALLOT_SCANNING_ON_VXSCAN_OPTION` to `true` for the Vx demos org to enable demoing VxMark ballot scanning on VxScan. But while I was at it, I also reorganized the feature flag configs to have consistent orders across orgs. Made gauging what's enabled vs. not just a little bit easier.


jonahkagan (approved):
Thanks for cleaning up!

PR #6867, opened 2025-07-29
feat(ui): improve `CheckboxGroup` type safety

eventualbuddha:
## Overview
_(This is part 1 of a series of PRs extracted from https://github.com/votingworks/vxsuite/pull/6862)_

Previously, `CheckboxGroup` was unable to enforce that the `options` and `values` parameters matched, treating all values as `string`. In many cases, there is a more appropriately narrow type that can be used. Doing so increases type safety and prevents bugs. Uses the [`NoInfer` utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#noinfertype) to ensure TypeScript only uses the `options` parameter for inferring the generic type `T`.

## Demo Video or Screenshot
CleanShot 2025-07-29 at 06 48 53@2x

## Testing Plan
Added a type-based test.


kofi-q (approved):
💯

PR #6881, opened 2025-07-30
test: update all non-UI tests to vitest v3

eventualbuddha:
## Overview

There seems to be a problem with running `libs/ui` under vitest v3 such that `styled-components`'s `p.theme` is an empty object rather than the expected theme object. I dug into why a bit, but didn't come up with any explanations.

Where relevant, I updated all the coverage thresholds to use negative values (absolute counts instead of percentages). The only code changes were due to [vitest v3 being stricter about matching errors in `toThrow` et al](https://vitest.dev/guide/migration.html#more-strict-error-equality).

## Demo Video or Screenshot
n/a

## Testing Plan
Automated.

### DIFF
--- a/libs/ballot-encoder/package.json
+++ b/libs/ballot-encoder/package.json
@@ -31,14 +31,14 @@
   "devDependencies": {
     "@types/node": "20.17.31",
     "@types/text-encoding": "^0.0.35",
-    "@vitest/coverage-istanbul": "^2.1.8",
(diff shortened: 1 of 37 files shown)


adghayes:
I had the same trouble when I tried to upgrade one of the frontend libraries a few months ago, and the same problem when I tried to update jest long before. It could be something to do with multiple instances of styled components running. But I've never found a solution and it makes me want to get away from styled components.

PR #6883, opened 2025-07-31
ignore hidden directories in VxAdmin election packages search

### LINKED ISSUE #6841: VxAdmin: Ignore hidden files and directories when searching for election packages on USB drives
One more implication to our searching for election packages on USB drives a couple levels deep is that files that have been deleted off the USB drive can still appear. When you delete files off of a USB drive, I believe some operating systems move the files to a hidden trash folder on the drive. I just found myself confused by why some files I'd just deleted were still showing up in the VxAdmin election package configuration UI

Issue automatically created from Slack feedback reported by arsalan: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1750951370526689?thread_ts=1750951370.526689&cid=CJU9MSC6S

adghayes:
## Overview

Closes #6841. Updates the `listDirectory` method underlying VxAdmin's search for election packages on a USB drive to ignore not only hidden files but also hidden directories and their contents.

### DIFF
--- a/libs/fs/src/list_directory.ts
+++ b/libs/fs/src/list_directory.ts
@@ -64,7 +64,7 @@ export interface ListDirectoryOptions {
 
 const DEFAULT_LIST_DIRECTORY_OPTIONS: Required<ListDirectoryOptions> = {
   depth: 1,
-  excludeHidden: true,
+  excludeHidden: false,
 } as const;
 
 /**
--- a/libs/fs/src/list_directory.ts
+++ b/libs/fs/src/list_directory.ts
@@ -54,11 +54,15 @@ export type ListDirectoryError =
 
 /**
  * Options for listing directories.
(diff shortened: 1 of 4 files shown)


libs/fs/src/list_directory.ts
```diff
@@ -52,15 +52,36 @@ export type ListDirectoryError =
   | { type: 'no-entity'; message: string }
   | { type: 'not-directory'; message: string };
 
+/**
+ * Options for listing directories.
+ * - `depth`: How deep to traverse directories. Defaults to 1 (not recursive).
+ * - `includeHidden`: Whether to include hidden files (those starting with a dot or underscore).
```
eventualbuddha, line 58: If you put these on the properties themselves then you get the inline docs while writing/hovering over the properties themselves, which I think is more useful than having them here.

Issue #6888, opened 2025-07-31
VxPollBook: Improve UI when fields are at the maximum content length allowed


carolinemodic:
https://votingworks.slack.com/archives/C07CR6T1TLK/p1753920090838909

PR #6889, opened 2025-07-31
Enforce max character limits for various voter fields

carolinemodic:
## Overview
Adds and enforces maximum character limits for various user-inputted fields around the voter schema. The frontend will prevent more then the required characters from being entered. The backend will automatically truncate the data if it exceeds the length which can happen either when there is long content in the pollbook-package to begin with or if a user hacked around the frontend and send a request to the backend manually. 

Since the limits need to be shared between the backend and frontend I had to move the voter type to libs/types to get around circular dependency import problems, that is seperated into the second commit (the first commit duplicates the type in frontend/backend for ease but that gets cleaned up). 

Cleaning up UI issues on the long content will be tackled in a followup (https://github.com/votingworks/vxsuite/issues/6888)

The backup worker test had a mailing address suffix above the max limit that was getting truncated so I adjusted that to just be under the limit which is why there are some snapshot changes. 

## Demo Video or Screenshot
Tested address fields as well but for the purposes of the demo you can get the idea from just the name update flow: 

https://github.com/user-attachments/assets/fb1ab9aa-32d3-4ef7-a0c7-74dee1cfeaa3

## Testing Plan
Manually tested all frontend inputs affected. Prior to making frontend changes tested backend truncation by creating longer frontend inputs. Backend truncation also covered in test file.


apps/pollbook/backend/src/globals.ts
```diff
@@ -48,3 +48,18 @@ export const MAX_POLLBOOK_PACKAGE_SIZE = 10 * MEGABYTE;
...
+  nameSuffix: 5,
+  streetNumber: 6,
+  streetName: 50,
+  streetSuffix: 4,
+  apartmentUnitNumber: 15,
+  addressLine2: 75,
+  zip5: 5,
+  zip4: 4,
+  cityTown: 50,
+} as const;
```
kshen0: Just curious, how did we come up with these limits?

apps/pollbook/backend/src/types.test.ts
```diff
@@ -0,0 +1,358 @@
...
+      mailingApartmentUnitNumber: 'o'.repeat(
+        VOTER_INPUT_FIELD_LIMITS.apartmentUnitNumber + 10
+      ),
+      mailingAddressLine2: 'p'.repeat(
+        VOTER_INPUT_FIELD_LIMITS.addressLine2 + 10
+      ),
+      mailingCityTown: 'q'.repeat(VOTER_INPUT_FIELD_LIMITS.cityTown + 10),
+      mailingZip5: 'r'.repeat(VOTER_INPUT_FIELD_LIMITS.zip5 + 10),
+      mailingZip4: 's'.repeat(VOTER_INPUT_FIELD_LIMITS.zip4 + 10),
+    } as const;
```
kshen0: Why not just go 1 over the limit for these tests?

kshen0 (approved):

apps/pollbook/backend/src/globals.ts
carolinemodic: I believe @adghayes set them based on the NH internal systems limits

PR #6895, opened 2025-07-31
wip(hwta-scan): stop polling for status

eventualbuddha:
Rather than using grout/HTTP requests at a fixed interval to poll, we now establish a `WebSocket` to watch for changes that the server pushes to us.


eventualbuddha:
Closing for now. May come back to it later, but for now I plan to move forward with a long-polling version.

PR #6909, opened 2025-08-05
VxPollBook: Update Statistics Screen for primaries and other tweaks

carolinemodic:
## Overview
Makes the following changes to the stats screen: 
- Only shows stats for the configured precinct, if set. 
- Makes the y-axis not include fractions
- Removes participation percent
- For a general election adds a party filter to the voter card, filters as expected
- For a primary election adds a party filter to the stats page as a whole, filters whole page
- Primary election, undeclared filter hides throughput and changes "Check Ins" to "Declared Party" 

I went back and forth in working on this change on how much to seperate or combine the code for the primary vs general cases, ultimately landing on seperating it out more. Initially I had both the backend and frontend more combined but that involved both the backend and frontend checking for primary vs. general and doing different logic which felt confusing, this way the primary vs. general logic is handled by the frontend which just calls different backend functions as appropriate. 

## Demo Video or Screenshot

Single Precinct General Election

https://github.com/user-attachments/assets/f90ad3a5-a437-4490-9a7c-81f61ea49a85

Multi Precinct Primary Election
- There is some lag present, this election has 50k voters which is more then we expect to need to support right now. I haven't tested with the number of events at scale but intend to do so as part of upcoming QA, if we want to increase the speeds here I think that adding a precinct column and indexing it for voters would go a long way as right now we are always loading all voters which is never really necessary. 

https://github.com/user-attachments/assets/ac834414-6df5-4854-ab47-29220b0b235d

## Testing Plan
Manual testing as shown above. Added various testing for backend and frontend functionality. 

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/local_store.ts
+++ b/apps/pollbook/backend/src/local_store.ts
@@ -2,6 +2,7 @@ import { BaseLogger } from '@votingworks/logging';
 import { Client as DbClient } from '@votingworks/db';
 import {
   CheckInBallotParty,
+  PartyAbbreviation,
   safeParseJson,
   Voter,
   VoterAddressChange,
@@ -42,6 +43,8 @@ import {
   VoterNameChangeEvent,
   VoterRegistrationEvent,
   VoterSearchParams,
+  PartyFilterAbbreviation,
+  PrimarySummaryStatistics,
 } from './types';
 import {
   applyPollbookEventsToVoters,
@@ -79,6 +82,7 @@ function toPatternMatches(raw: string): string {
 
 interface GetVotersQueryParams {
   matchConfiguredPrecinctId?: boolean;
+  partyFilter?: PartyAbbreviation;
 }
 
 export class LocalStore extends Store {
@@ -619,7 +623,13 @@ export class LocalStore extends Store {
     return { voter: updatedVoter, receiptNumber };
   }
 
-  getThroughputStatistics(throughputInterval: number): ThroughputStat[] {
+  getThroughputStatistics(
+    throughputInterval: number,
+    partyFilter: PartyFilterAbbreviation
+  ): ThroughputStat[] {
+    if (partyFilter === 'UND') {
+      return [];
+    }
     const intervalMs = throughputInterval * 60 * 1000;
     const checkInEventRows = this.client.all(
       `SELECT * FROM event_log WHERE event_type = '${EventType.VoterCheckIn}' ORDER BY physical_time, logical_counter, machine_id`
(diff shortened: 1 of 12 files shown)


carolinemodic:
@kofi-q for design review
@adghayes for product review
@kshen0 for code review

kofi-q (approved):
Design LGTM!

[Non-blocking] if it's low-lift, would be good to vertically center the card titles in the expanded headers:

Image

carolinemodic:
> I think that performance / lag is generally fine for the current state. In the general election case, however, it would be preferable that when the Voter card is loading data, the loading message is just behind that card rather than behind the whole top row, jumping the whole page when first toggling. it's only going to be on first load so it's not a big deal, I'll defer to you based on whether it's low lift or not.

Yeah I noticed this feels weird as well but the problem is that right now its just one backend query serving the information for both of those cards so it ... is loading the data for both cards. I could split it up into different queries/components but I think its ultimately fine for now and we can decide if we want to prioritize that polish once we look more at QA / other things.

adghayes:
Sounds good. Is it trivial to just have that loading section have a consistent height so the throughput chart doesn't jump? If still too annoying, please ignore me...

carolinemodic:
https://github.com/user-attachments/assets/8c072c96-fb00-4e39-878f-de389cc53201

Updated with the suggested polish ^

apps/pollbook/backend/src/local_store.ts
```diff
@@ -676,23 +693,124 @@ export class LocalStore extends Store {
...
-    const totalCheckIns = Object.values(voters).filter((v) => v.checkIn).length;
+    const totalCheckIns = Object.values(votersWithCheckInMatchingParty).length;
 
     return {
       totalVoters,
       totalCheckIns,
       totalAbsenteeCheckIns,
       totalNewRegistrations,
+      totalUndeclaredDemCheckIns: undeclaredDemVoters,
+      totalUndeclaredRepCheckIns: undeclaredRepVoters,
```
kshen0, line 813: It feels a little counterintuitive that when the filter == `ALL` this property reports 0 undeclared REP/DEM check-ins. The behavior for `UND` might be different enough that it makes sense to split this into 2 methods eg `getPrimarySummaryStatistics` where `partyFilter: Omit<PartyFilterAbbreviation, 'UND'>` and `getPrimarySummaryStatisticsForUndeclaredVoters` and have the API coordinate the appropriate call. Thoughts?

adghayes:
thank you for taking the time for the polish

kshen0 (approved):

Issue #4914, opened 2024-06-03
VxDesign: Improve ballot rendering memory management


jonahkagan:
- Cache renderer in VxDesign backend - to do this, we'll need to figure out a way to make sure the renderer pages still get cleaned up after they are used
- Limit number of concurrent documents rendered to limit memory usage - not sure how to do this since we need each ballot document rendered in Chromium in order to compute the election hash, and then we need to add the QR code before exporting to PDF. Maybe there's a way we can serialize the rendered ballot without the QR code and then reload it into Chromium again to add the QR code, but this may slow things down a good amount.

PR #6934, opened 2025-08-07
VxDesign: Reduce memory usage when rendering ballots

jonahkagan:
## Overview

Task: https://github.com/votingworks/vxsuite/issues/4914

The main memory bottleneck in the current code is that we keep a browser page open for each rendered ballot until we can extract the grid layouts, create the election definition, hash it, and inject that hash in the QR code.

I addressed this by splitting the process into phases: 
1. Render each ballot, then extract the grid layout and the HTML content. Close the browser page when done.
2. Use the grid layouts from all ballots to create the election definition (and its ballot hash)
3. Go through each ballot again, creating a new browser page and loading the previously created HTML content and injecting the QR code with the ballot hash

I also noticed that even with this approach, there appeared to be some memory bloat from creating a new browser page for each ballot, even if I cleaned up the page after using it. To fix this, I enforced rendering the ballots serially so we can reuse a single browser page. Making that change seemed to stop the memory bloat.

Moving forward, we can probably tune this approach more to better balance memory and speed by using a pool of browser pages, but I wanted to start with a minimal change that solved the memory issue.

## Demo Video or Screenshot
Some basic benchmarking with the electionGeneral fixture on my dev machine:

**main**
- export election package + ballots: 1GB memory, 13.051s
- export test decks: 2.5 GB memory, 5.527s

**this branch**
- export election package + ballots: 100-200MB, 14.271s
- test decks: 100-200MB, 20.329s

## Testing Plan
- Manual testing and benchmarking
- Updated existing tests

## Checklist


nikhilb4a:
Sorry this slipped by me today and I just noticed it as I'm about to log off for the weekend. I'll get to it first thing Tuesday!

libs/hmpb/src/renderer.tsx
```diff
@@ -15,7 +15,7 @@ export type Page = Pick<
  * Creates a {@link RenderDocument}
  */
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-export function createDocument(page: Page) {
+export function createDocument(page: Page, cleanup: () => void) {
```
nikhilb4a, line 18: Nit: feel free to ignore as it might be a morning review thing, but it took me a second to wrap my head around the document vs renderer cleanup with them being used close together. I could see it being helpful to rename one to make it more intuitive, maybe even keeping the name `releasePage` or `releaseLock` in this case. That being said, I understand if `cleanup` is a pattern you're trying to use.

Edit - After reading the later commits, I see why `document.cleanup()` makes sense. So, if anything I'd consider renaming `renderer.cleanup()` to `.close()` or `.shutdown()`, but fine as is

libs/hmpb/src/render_ballot.tsx
```diff
@@ -574,25 +605,48 @@ export async function renderAllBallotsAndCreateElectionDefinition<
...
+    .zip(ballotContents)
+    .async()
+    .map(async ([props, ballotContent]) => {
+      const document = await renderer.loadDocumentFromContent(ballotContent);
+      const ballotPdf = await renderBallotPdfWithMetadataQrCode(
+        props,
+        document,
+        electionDefinition
+      );
+      document.cleanup();
```
nikhilb4a, line 643: I'm going to ask a silly question to help me grasp what I'm missing..

What would break if we didn't have the `document.cleanup()` and page lock management? It seems like that protects against us rendering two ballots with the same document/page, but I'm missing where there would be any attempt at parallelism. And since there isn't a place we retry after failing to get a lock, I'm not certain the lock capturing is necessary. Sorry if I'm missing obvious but thanks for any clarification

nikhilb4a (approved):
Great memory management improvements! The separation of the rendering into two steps feels very clear. I was struggling a bit to wrap my head around the need for the cleanup, so I asked some questions for clarification.

libs/hmpb/src/renderer.tsx
jonahkagan: Good feedback - I'll rename one of them to clarify the difference

libs/hmpb/src/render_ballot.tsx
jonahkagan: This is basically just enforcing the memory optimization changes from this PR for any future usage. Ideally, the API for the renderer would be structured in a way that you could only use it in a memory-efficient way. However, currently, without the lock+cleanup, somebody could come along and write code that tries to use two documents simultaneously (which is how the previous code worked). So you're right that it's not necessary - it's more of a temporary stopgap safeguard until we can redesign the renderer API.

nikhilb4a: Got it, that makes sense to me as the code's written. It's a good safeguard against misuse for now.

Issue #6977, opened 2025-08-11
if you make a change on machine A while machine B has the statistics page open, the statistic page will not update (at least for the check-in count) until you navigate away and return


arsalansufi:
if you make a change on machine A while machine B has the statistics page open, the statistic page will not update (at least for the check-in count) until you navigate away and return

Issue automatically created from Slack feedback reported by drew: 
https://votingworks.slack.com/archives/C098ZSZ7719/p1754949783738159?thread_ts=1754949783.738159&cid=C098ZSZ7719

PR #7009, opened 2025-08-13
VxPollBook: Frontend should always refetchOnInterval where data can change from underlying syncs

carolinemodic:
## Overview
Followup to https://github.com/votingworks/vxsuite/pull/6848 . Closes https://github.com/votingworks/vxsuite/issues/6977 .  In that PR setting staleTime (later changed to 0) made the behaviour work as I had hoped and auto refresh when needed to get updated data. I'm not sure how this was working in dev but in prod the behavior is that it only refetches on remount despite being "stale" immediately, which makes sense and follows react's documentation. We should always have a refetchInterval set up in queries where the data can change out from under us when syncing with another machine, this PR audits that in all queries and sets it up. 

## Testing Plan
Copied build directory to imaged machine. Loaded each of: statistics screen, EM voter detail page, search voter result list on the pollbook. On another connected pollbook checked in a voter. Saw the hacked pollbook update accordingly. 

## Checklist

### DIFF
--- a/apps/pollbook/frontend/src/app_poll_worker_screen.test.tsx
+++ b/apps/pollbook/frontend/src/app_poll_worker_screen.test.tsx
@@ -16,11 +16,11 @@ import {
   createMockVoter,
 } from '../test/mock_api_client';
 import { AUTOMATIC_FLOW_STATE_RESET_DELAY_MS } from './globals';
-import { DEFAULT_QUERY_REFETCH_INTERVAL } from './api';
 import {
   getMockAamvaDocument,
   getMockExactSearchParams,
 } from '../test/aamva_fixtures';
+import { DEFAULT_QUERY_REFETCH_INTERVAL } from './api';
 
 let apiMock: ApiMock;
 const famousNamesElection: Election =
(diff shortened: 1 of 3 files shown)


apps/pollbook/frontend/src/api.ts
```diff
@@ -178,7 +178,7 @@ export const searchVoters = {
     return useQuery(
       this.queryKey(searchParams),
       () => apiClient.searchVoters({ searchParams }),
-      { refetchOnMount: 'always' }
+      { refetchInterval: DEFAULT_QUERY_REFETCH_INTERVAL }
```
carolinemodic, line 181: the refetchOnMount "always" is already the default behavior when staleTime is 0 so that is not needed here.

apps/pollbook/frontend/src/api.ts
```diff
@@ -64,7 +64,7 @@ export const getPollbookConfigurationInformation = {
       this.queryKey(),
       () => apiClient.getPollbookConfigurationInformation(),
       {
-        refetchInterval: 1000,
+        refetchInterval: DEFAULT_QUERY_REFETCH_INTERVAL,
```
adghayes, line 67: the effect here is that the network polling and the statistics have the same refetch interval, correct? if the performance impact is negligible, seems fine, but I think we could do with a larger interval for statistics (5 or even 10 seconds). I'm not sure how the `getVoter` endpoint is being used to know what's appropriate there.

carolinemodic: the event polling between computers is more frequent: https://github.com/votingworks/vxsuite/blob/main/apps/pollbook/backend/src/globals.ts#L37

PR #7012, opened 2025-08-13
VxPollBook: handle thrown errors in backup worker export

carolinemodic:
## Overview
We always want to continue trying to export the backup so we should wrap the work in a try/catch to make sure a thrown error doesn't break the polling interval. Specifically if you remove a usb while the export is processing it will throw an error. 

## Testing Plan
Removed usb while export was processing, saw the catch get hit and the interval continue, reinsertted usb, saw backup export onto it fine. 

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/backup_worker.ts
+++ b/apps/pollbook/backend/src/backup_worker.ts
@@ -264,14 +264,21 @@ export function start({
     }
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     for await (const _ of setInterval(BACKUP_INTERVAL)) {
-      const result = await exportBackupVoterChecklist(
-        workspace,
-        usbDrive,
-        workspace.logger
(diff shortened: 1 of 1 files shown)


apps/pollbook/backend/src/backup_worker.ts
```diff
@@ -20,7 +20,7 @@ import {
 } from './voter_checklist';
 import { LocalStore } from './local_store';
 
-const BACKUP_INTERVAL = 1_000 * 60; // 1 minute
+const BACKUP_INTERVAL = 1_000 * 10; // 1 minute
```
adghayes, line 23: is this intentional?

carolinemodic: no was just making it faster for testing will revert thanks

adghayes (approved):

Issue #6976, opened 2025-08-11
VxPollBook: if data is already in the search bar, a scanned driver's license does not leap to identify page or update search


arsalansufi:
actually this seems to work from an empty search state but with an existing search in the bar, it's using the existing data instead of the new scan data

Issue automatically created from Slack feedback reported by kevin: 
https://votingworks.slack.com/archives/C098ZSZ7719/p1754949583370569?thread_ts=1754949436.217639&cid=C098ZSZ7719

PR #7016, opened 2025-08-13
VxPollBook: prevent redirecting too early when IDs are scanned in succession

kshen0:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6976

Issue: scanning `ID B` when data from `ID A` is in the search state results in navigating to the check-in screen for `ID A` (but it should be for `ID B`).

* The `VoterSearch` component uses two effects to handle updates to the scanned ID as reported by the backend
* Effect 1: when the scanned ID changes, it populates the search query params with data from the scanned ID
* Effect 2: when there's a scanned ID and exactly 1 ID search result, navigate to the next page. For pollworkers the next page is the check-in screen. For election managers the next page is the voter details page.

The above setup works for scanning from an empty search state or manually entered data because the existing search data doesn't specify `exactMatch: true` and therefore doesn't match the check for Effect 2. But when you

1. Scan an ID for `John Doe`
2. Click "Back" to return from check-in page to search screen
3. Scan a 2nd ID for `Jane Smith`

the "old" search data does match the check for Effect 2. That effect triggers before Effect 1 can finish updating the search params, or the updated search query can finish, or both. Therefore Effect 2 triggers navigation based on data from John Doe's ID, not Jane Smith's.

To fix this I added a check to Effect 2 that ensures the data on the scanned ID matches the data in the current search params.

## Demo Video or Screenshot

Recommend watching with sound so you can hear when scans happen.

**Before**

https://github.com/user-attachments/assets/51a065c8-8c01-4146-8f26-f1be4d86a285

**After**

https://github.com/user-attachments/assets/34b25640-7c1a-40b1-9d1e-ce9a760aa8b7

## Testing Plan
- manually tested

## Checklist

### DIFF
--- a/apps/pollbook/frontend/src/voter_search_screen.tsx
+++ b/apps/pollbook/frontend/src/voter_search_screen.tsx
@@ -102,6 +102,18 @@ export function createEmptySearchParams(
   };
 }
 
+function documentMatchesParams(
+  document: AamvaDocument,
+  searchParams: VoterSearchParams
+) {
+  return (
+    document.firstName === searchParams.firstName &&
+    document.middleName === searchParams.middleName &&
+    document.lastName === searchParams.lastName &&
+    document.nameSuffix === searchParams.suffix
+  );
+}
+
 export function VoterSearch({
   search,
   setSearch,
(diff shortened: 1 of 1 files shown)


adghayes (commented):
Why are there two effects handling this at all? Could they not be one effect, in which these cases would be easier to handle directly?

kshen0:
> @kshen0 could you create a ticket to track the tech improvement work of combining the scanned ID and search results query, to avoid the back and forth and complexity? In the current approach, this seems safe.

https://github.com/votingworks/vxsuite/issues/7039

Issue #6992, opened 2025-08-12
VxPollBook: fix filtering out punctuation in manual search


arsalansufi:
Manual search is not filtering out punctuation, e.g.:

<img width="1154" height="722" alt="Image" src="https://github.com/user-attachments/assets/89423f1d-a274-4e7f-bb1e-541f025c1ef2" />
<img width="1153" height="722" alt="Image" src="https://github.com/user-attachments/assets/32c0f23d-eb07-4015-ad80-6f28616c469b" />

Issue automatically created from Slack feedback reported by pius: 
https://votingworks.slack.com/archives/C098ZSZ7719/p1755020254035369?thread_ts=1755020254.035369&cid=C098ZSZ7719

PR #7018, opened 2025-08-13
VxPollBook: ignore period in voter search

kshen0:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6992

* Period chars were not being respected in the punctuation-agnostic search
* ie. searching for `St. Clair` if a voter's name was `St Clair` would not work, and vice versa

I refactored the regex construction for exact and prefix searching because it wasn't very DRY.

## Demo Video or Screenshot

**Before**

https://github.com/user-attachments/assets/2647d5da-99b4-4903-9867-216b375d6964

**After**

https://github.com/user-attachments/assets/35521d40-21c2-45f4-b004-e19dd162e847

## Testing Plan

- manually tested
- added store test

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/local_store.test.ts
+++ b/apps/pollbook/backend/src/local_store.test.ts
@@ -422,6 +422,73 @@ test('searchVoters returns results sorted by configured precinct first, then by 
   expect(results[3].precinct).toEqual('precinct-2');
 });
 
+test('searchVoters ignores punctuation', () => {
+  const localStore = LocalStore.memoryStore(mockBaseLogger({ fn: vi.fn }));
+  const testElectionDefinition = getTestElectionDefinition();
+
+  const voters = [
+    createVoter('22', 'Charlie', 'Carter', {
+      middleName: '',
+      suffix: '',
+      precinct: 'precinct-1',
+    }),
+    createVoter('23', 'D av-id', "Da.v'is", {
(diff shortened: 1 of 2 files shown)


apps/pollbook/backend/src/local_store.test.ts
```diff
@@ -151,6 +152,33 @@ test('findVotersWithName works as expected - voters without name changes', () =>
...
+      lastName: 'StMichel',
+      middleName: '',
+      suffix: '',
+    })
+  ).toEqual(
+    expect.arrayContaining([
+      expect.objectContaining({ voterId: voters[6].voterId }),
+    ])
+  );
 });
```
adghayes, line 182: The testing here is for `findVotersWithName` but if it's a manual search, we go down the `searchVoters` code path. Do both benefit from this change?

adghayes (commented):
question about the code path but otherwise looks good

apps/pollbook/backend/src/local_store.test.ts
kshen0: Yes, `toPatternStartsWith` is used by `searchVoters` and `toPatternMatches` is used by `findVotersWithName`. Both were refactored to use a shared set of ignored punctuation. But you make a good point that we should also test `searchVoters`.

Issue #6597, opened 2025-06-11
VxAdmin: read card after programming to confirm success


arsalansufi:
Card programming: The card programming methods return successfully if the relevant writes succeed but don't actually confirm the overall state/validity of the card, which means that programming can indicate success even if the card will subsequently be read as invalid - this mostly comes up during QA/dev when we've mixed up certs, etc.

Issue automatically created from Slack feedback reported by arsalan: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1749576780402089

PR #7035, opened 2025-08-14
Properly error when programming cards configured for the incorrect environment

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6597

While working on a loosely related cert discrepancy, I decided to finally close out a classic source of confusion that has come up in almost every QA run.

If you program a card configured for the incorrect environment, i.e., a dev/QA card on a prod machine or a prod card on a dev/QA machine, VxAdmin will indicate that the card has been successfully programmed. But if you then use the card, it'll show up as invalid/blank.

The current card programming logic declares success so long as the card programming commands succeed. It doesn't actually check the overall state of the card to ensure validity. This PR updates the `program` method to verify that the card has been configured for the correct environment before proceeding with programming commands.

## Demo Video or Screenshot

_Before_

Programming a dev-blessed card on a prod-blessed machine or vice versa results in what appears to be a successful program. But removing and reinserting the card reveals it to be invalid/blank.

https://github.com/user-attachments/assets/eed36135-f303-42c2-b16f-1e94c2f125ea

_After_

The program operation now clearly fails :relieved:

https://github.com/user-attachments/assets/b5e520a8-3963-4580-b567-0f5f498fd3fc

## Testing Plan

- [x] Lots of manual testing with actual Java Cards
- [x] Updated automated tests

## Checklist

- [x] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate for any new user actions. ➡️ Relying on existing logging though there's room for improvement here

### DIFF
--- a/libs/auth/src/java_card.test.ts
+++ b/libs/auth/src/java_card.test.ts
@@ -1076,6 +1076,13 @@ test.each<{
         cardType: expectedCardType,
       })
     );
+    await mockCardSignatureRequest(
+      CARD_VX_CERT.PRIVATE_KEY_ID,
+      getTestFilePath({
+        fileType: 'card-vx-private-key.pem',
+        cardType: expectedCardType,
+      })
+    );
     mockCardPinResetRequest(pin);
     mockCardPinVerificationRequest(pin);
     mockCardKeyPairGenerationRequest(
--- a/libs/auth/src/java_card.ts
+++ b/libs/auth/src/java_card.ts
@@ -302,6 +302,7 @@ export class JavaCard implements Card {
       cardVxCert,
       this.vxCertAuthorityCertPath
     );
+    await this.verifyCardPrivateKey(CARD_VX_CERT.PRIVATE_KEY_ID, cardVxCert);
 
     await this.resetPinAndInvalidateCard(pin);
 


arsalansufi:
@jonahkagan Really good q! Here's the high level.

Cards are first blessed by VotingWorks for a given environment. This is before any VxAdmin programming. This results in a first certificate on the card. Let's call it certificate A.

Then when actually programming the card on VxAdmin, VxAdmin issues a second certificate that confers a user identity. This results in a second certificate on the card. Let's call it certificate B.

The `program` function before this PR simply made sure that the card commands to create certificate B worked. Those commands don't interface with certificate A in any way.

Reading the card after programming checks, among many things, that certificate A is valid and actually for the correct environment.

This comment is making me realize that I could alternatively just extract the part of `safeReadCardDetails` that checks certificate A and add that to the start of the `program` function. It just felt clean to use the same read function that we use elsewhere to update the card status at the end of the `program` function.

But I might play around with this alternative.

arsalansufi:
> This comment is making me realize that I could alternatively just extract the part of safeReadCardDetails that checks certificate A and add that to the start of the program function.

Per this comment above, I did in fact end up simplifying things. Edited the PR title and description accordingly. The branch name is a bit misrepresentative now.

Issue #6036, opened 2025-02-21
VxAll: timezone changes are not reflected in reports


arsalansufi:
Timezone changes require reboot to be reflected in reports.

Issue automatically created from Slack feedback reported by drew: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1740159564410169

eventualbuddha:
Does this happen if you change the time, or only if you change the timezone? If the latter, I'm guessing it's due to `TZ` env var not being reset until processes restart.

eventualbuddha:
This seems to support the idea that NodeJS simply doesn’t see the change: https://github.com/nodejs/node/issues/33805. It also gives us a way to fix it by setting `process.env.TZ` ourselves while we set the timezone with the OS.

adghayes:
Note for whoever works on this, that this is reproducible in dev. Look at a tally report and note the timestamp. Now update the timezone and the time to match (e.g. from PST to EST and move the time forward 3 hours). Look at the tally report, and you'll see the timestamp hasn't changed, even though the local time has changed.

PR #7040, opened 2025-08-15
update `set-clock` script to be independent of TZ environment variable

### LINKED ISSUE #6971: i’m assuming this is applicable to all #, but when you change the timezone, the time changes too
i’m assuming this is applicable to all #, but when you change the timezone, the time changes too. this is confusing and not how I expected it to work!

https://votingworks.slack.com/archives/C098ZSZ7719/p1754947474153709?thread_ts=1754947474.153709&cid=C098ZSZ7719

adghayes:
## Overview

Closes #6971. Since the fix for #6036, we set the timezone of the node process explicitly after we update the system time. An unintended consequence of that was that, next time you run our `set-clock` script, there's a `TZ` environment variable and `timedatectl set-time` actually respects the environment variable over the actual system timezone. Therefore, on all subsequent time changes, the time was getting set with respect to the old timezone.

One solution is to move the `process.env.TZ` setting up before the script is called. But I wanted the script to make sense in isolation and work as expected regardless of environment

### Before

https://github.com/user-attachments/assets/77ba62a2-0906-44c2-b07d-117f0c7b34c7

### After

https://github.com/user-attachments/assets/f1159305-a231-46ab-a935-1cee269eab68

## Testing Plan

Manual Testing.

## Checklist

### DIFF
--- a/libs/backend/intermediate-scripts/set-clock
+++ b/libs/backend/intermediate-scripts/set-clock
@@ -30,4 +30,8 @@ else
     timedatectl set-timezone "${1}"
 fi
 
+# If the parent node process has a TZ environment variable set, it will
+# override the timezone set by timedatectl. So we unset it here.
+unset TZ
+
 timedatectl set-time "${2}"


libs/backend/intermediate-scripts/set-clock
```diff
@@ -30,4 +30,8 @@ else
     timedatectl set-timezone "${1}"
 fi
 
+# If the parent node process has a TZ environment variable set, it will
+# override the timezone set by timedatectl. So we unset it here.
+unset TZ
```
kshen0, line 35: I know the only caller sets `process.env.TZ` but what do you think about storing this so we can resassign the old value of `TZ` once the time change is done?

kshen0 (approved):
LGTM; especially appreciate the low impact change and fastest path to consistent behavior.

I think in the long term it's still counterintuitive for changing time zone to _not_ change wall time. The underlying Linux command and most consumer apps do the wall time conversion for you. But given the potential confusion of magically changing wall time only after modal submission I think this is a good stopgap.

adghayes:
Agree on the ideal behavior, just created https://github.com/votingworks/vxsuite/issues/7044 on the triage board to track.

libs/backend/intermediate-scripts/set-clock
adghayes: As I was debugging this, I actually discovered that `TZ` is undefined on startup. I'm not sure where it's caching the timezone, maybe internally in the engine 🤷

Issue #6957, opened 2025-08-11
VxAdmin: Persist card programming error message on screen long enough for user to read when card is preemptively removed


arsalansufi:
Parent cert discrepancy issue: https://github.com/votingworks/cert-discrepancies/issues/199

PR #7048, opened 2025-08-15
Smart cards screen: Persist error message momentarily after card removed

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6957

During cert testing, SLI has encountered a known edge case where, if you remove a card preemptively after initiating programming, an error message can briefly flash before you have time to read it and we return to the card insertion screen. Errors are currently only displayed on the card details screen, which is naturally cleared when a card is removed.

This PR persists the error message for a couple of seconds even after you remove the card and return to the card insertion screen. I didn't try to distinguish between errors due to preemptive removal and, say, unblessed or misconfigured cards. So these extra seconds of display on the card insertion screen happen for all errors, whether they're strictly needed or not. All of these errors should be exceedingly rare in true production environments.

### Before

Notice the brief flash of an error message.

https://github.com/user-attachments/assets/6cca7852-51d2-40e5-96e0-d43988004f85

### After

https://github.com/user-attachments/assets/4d20d742-61d1-42e3-a0c0-b4c20642361f

## Testing Plan

- [x] Tested manually with actual Java Cards and preemptive removals
- [x] Added automated tests

## Checklist

### DIFF
--- a/libs/ui/src/smart_cards_screen.tsx
+++ b/libs/ui/src/smart_cards_screen.tsx
@@ -571,7 +571,7 @@ export function SmartCardsScreen({
     if (card.status === 'no_card') {
       const timeout = setTimeout(() => {
         setActionResult(undefined);
-      }, 3000);
+      }, 4000);
       return () => clearTimeout(timeout);
     }
 
--- a/libs/ui/src/smart_cards_screen.test.tsx
+++ b/libs/ui/src/smart_cards_screen.test.tsx
@@ -826,7 +826,7 @@ test.each([
 
     await screen.findByText(expectedErrorMessage);
 
-    await vi.advanceTimersByTimeAsync(3000);
+    await vi.advanceTimersByTimeAsync(4000);
 
     await vi.waitFor(() => {
       expect(screen.queryByText(expectedErrorMessage)).not.toBeInTheDocument();


libs/ui/src/smart_cards_screen.tsx
```diff
@@ -59,6 +59,7 @@ const CardActions = styled.div`
   padding: 1rem;
   display: flex;
   flex-direction: column;
+  flex-grow: 1;
```
arsalansufi, line 62: This just makes sure the right panel is always full width to prevent flicker like so - notice the error message go from being wider to narrower:

https://github.com/user-attachments/assets/d11ec594-7741-4b0b-a617-2ef4741c4cfd

libs/ui/src/smart_cards_screen.tsx
```diff
@@ -437,7 +445,7 @@ export function CardDetailsAndActions({
       </CardIllustration>
       <div style={{ flexGrow: 1 }}>
         {actionResult && (
-          <div style={{ padding: '1rem' }}>
+          <div style={{ padding: '1rem', paddingBottom: 0 }}>
```
arsalansufi, line 448: Just tweaking the spacing between action callouts and the content below to be a little more consistent with the spacing between the prompt and the buttons on the insert card screen:

| Before | After |
| - | - |
| before | after |

_Insert card screen for reference_

insert-card

libs/ui/src/smart_cards_screen.tsx
```diff
@@ -554,21 +562,53 @@ export function SmartCardsScreen({
...
+  useEffect(() => {
+    // Preemptively removing a card that's being programmed can result in a flash of an error
+    // message. To ensure that the error message can be read, continue to display it on the insert
+    // card screen for a few seconds.
+    if (card.status === 'no_card') {
+      const timeout = setTimeout(() => {
+        setActionResult(undefined);
+      }, 3000);
+      return () => clearTimeout(timeout);
+    }
```
arsalansufi, line 576: I could see a case for removing this `setTimeout` and just letting the error persist indefinitely on the insert card screen, until a card is reinserted. Went a bit back and forth here to be honest. Decided not to split hairs since this should all be exceedingly rare in production and is mostly just a response to a cert discrepancy.

nikhilb4a (approved):
Nice solution!

Issue #6706, opened 2025-07-02
[VxMark] Get to feature parity with VxMarkScan


kofi-q:
We've mostly kept common features between Mark and MarkScan in sync, but a few things have fallen through the cracks. Using MarkScan as a reference to bring Mark up to feature parity with the rest of the precinct system.

Remaining tasks for later, while I focus on pre-printed ballot marking for now:
- [ ] Election Manager screen(s)
  - [x] Alert when switching between Official and Tests modes when polls are open (IIRC, we do this in the other precinct apps) - Determined not needed
  - [x] Add power down button (not present on MarkScan, but matches VxScan
  - [x] Add Signed Hash Validation
  - [x] Add sound mute/unmute control - Determined not needed
  - [ ] avoid reconfigure from USB after unconfigure
- [ ] Virtual keyboard
- [x] Diagnostics screen(s) - punting to MK 15
- [ ] Voter-facing PAT device calibration flow

PR #7058, opened 2025-08-18
VxMark: standardize EM menu

adghayes:
## Overview

While getting screenshots for VxMark, I noticed that the EM screen is still pretty different than the VxMarkScan one which we iterated on before the v4 release. I just went ahead and copied over the style and layout changes, see screenshots below. 

### Before

Screenshot-VxMark-2025-08-18T22:48:06 316Z

### After

Screenshot-VxMark-2025-08-18T22:47:44 979Z

## Testing Plan

Updated the automated tests.

## Checklist


kofi-q (commented):
Thanks for doing this! FWIW, got a [WIP task](https://github.com/votingworks/vxsuite/issues/6706) to address the parity issues and a few changes started in the VxMark demo branch that we're creating images from. On the fence about committing this right now, to avoid conflicts - was also thinking of a direction similar to the parity work on the Poll Worker screens WRT [splitting out common subcomponents](https://github.com/votingworks/vxsuite/tree/main/libs/mark-flow-ui/src/components/poll_worker) to share between Mark and MarkScan and help avoid future

Do we need screenshots for a documentation pass? What's the timeline for that? If it's short, we can go ahead with this and I'lll add in the consolidation later.

adghayes:
oops, @kofi-q, I did not know that this was already in progress.

We need to submit documentation (user manual + TDP equivalent) as part of our application for the NH RFP, which is due early next week, so our goal has been to wrap up the pieces before the end of the week. I do think the documentation should include screenshots, but it doesn't necessarily need to match VxMarkScan. I just saw the disparity and thought to fix. I can just take screenshots of what we have on the branch. If something's going to change this week, I'll wait to take the screenshots until later in the week.

kofi-q:
Ah, good to know! Wasn't aware of the documentation requirement. If we're taking screenshots, we'll probably want to turn off the `REACT_APP_VX_MARK_ENABLE_BALLOT_PRINT_MODE_TOGGLE` feature flag, which is in a purely demo state right now and will eventually be controlled via system setting in VxDesign. A few other things are in flux/in progress too and I doubt they'll be done before end-of-week. Is this documentation just meant to capture the current state, or will it have to match what we intend to ship as a final product? Figuring out if anything needs to be re-prioritised/re-ordered for this

adghayes:
@kofi-q I just checked with Matt and he confirmed there should be screenshots but that we shouldn't be at all concerned about their accuracy, and even using VxMarkScan screenshots are fine. I'll close this so you can tackle this when you get to it without conflicts.

kofi-q:
SG - thanks! And sorry for the weird, in-between state at the moment. You might've found it already, but the demo branch is at [vxmark-2025-08-demo-image](https://github.com/votingworks/vxsuite/tree/vxmark-2025-08-demo-image). Visually, it should only be the logged-in screens that are different from `main` - rest of the changes are functional

PR #7062, opened 2025-08-19
VxPollBook: Refactor offline/online state management and check for bad ip addresses

carolinemodic:
## Overview
This PR makes a few notable changes that are all a little inter-related:

- Updates handling of the self-machine state to be more merged with how we track other connected nodes, we were already sharing the same database but not keeping the data structure updated with a old apiClient. This means we were caching the old apiClient for other pollbooks but not for the self. As part of the merging of the code handling I removed the specific setOnlineStatus function and now you should control with the setPollbookServiceForName method like you do for other pollbook connection updates, it will check for the special case of the machine going offline and set other connections to lost connection in that situation. 
- Specifically checks for a bad ipv4 address from avahi and ignores it, used the previous cached ip (if it exists) instead to handle weird avahi blips more gracefully
- Makes sure for all pollbooks (self, and other nodes) that we update the cached api client if we need a new ip address in avahi for that pollbook service 
- A by product of merging the code paths is that we the cleanupStalePollbooks method that runs will now handle updating the state of the current machine to LostConnection (setting it offline) when we haven't been able to ping ourselves in 10seconds, making that behavior less reactive then if there is one failed ping

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/networking.ts
+++ b/apps/pollbook/backend/src/networking.ts
@@ -1,6 +1,6 @@
 import { execFile } from '@votingworks/backend';
 import * as grout from '@votingworks/grout';
-import { assert, sleep } from '@votingworks/basics';
+import { assert, assertDefined, sleep } from '@votingworks/basics';
 import { LogEventId } from '@votingworks/logging';
 import { rootDebug } from './debug';
 import {
@@ -76,6 +76,10 @@ export function shouldPollbooksShareEvents(
   );
 }
 
+export function getNodeServiceName(machineId: string): string {
+  return `Pollbook-${machineId}`;
+}
+
 export function fetchEventsFromConnectedPollbooks({
   workspace,
 }: PeerAppContext): void {
@@ -100,14 +104,17 @@ export function fetchEventsFromConnectedPollbooks({
           workspace.store.getPollbookConfigurationInformation();
 
         // Maintain a queue of pollbooks to visit, refill and shuffle when empty
-        const pollbookNames = Object.keys(previouslyConnected).filter(
+        const pollbookNamesToVisit = Object.keys(previouslyConnected).filter(
           (name) =>
             previouslyConnected[name].status ===
-            PollbookConnectionStatus.Connected
+              PollbookConnectionStatus.Connected &&
+            // We don't need to fetch events from ourselves
+            previouslyConnected[name].machineId !==
+              myMachineInformation.machineId
         );
         if (pollbookQueue.length === 0) {
           // Shuffle pollbookNames
-          pollbookQueue = pollbookNames
+          pollbookQueue = pollbookNamesToVisit
             .map((name) => ({ name, sort: Math.random() }))
             .sort((a, b) => a.sort - b.sort)
             .map(({ name }) => name);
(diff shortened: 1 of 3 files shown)


apps/pollbook/backend/src/store_multi_node.test.ts
```diff
@@ -157,8 +157,6 @@ test('offline undo with later real time check in', async () => {
   expect(localB.getCheckInCount()).toEqual(2);
 
   // PollbookB goes offline
-  peerB.setOnlineStatus(false);
```
carolinemodic, line 160: the triggering of online/offline in these tests never did anything and was more for readability in understanding the scenario I am imagining, I think the comments do enough to convey that so just removing

apps/pollbook/backend/src/networking.ts
```diff
@@ -261,41 +287,80 @@ export function setupMachineNetworking({
...
           debug(
             'The current service is no longer found on avahi. Setting online status to false'
           );
-          workspace.store.setOnlineStatus(false);
+          workspace.store.setPollbookServiceForName(selfMachineServiceName, {
+            ...selfMachinePollbookService,
+            apiClient: undefined,
+            address: undefined,
+            status: PollbookConnectionStatus.LostConnection,
+          });
```
kshen0, line 300: nit: make a helper for this if we plan to use in several places?

apps/pollbook/backend/src/networking.ts
```diff
@@ -261,41 +287,80 @@ export function setupMachineNetworking({
...
-            currentPollbookService && currentPollbookService.apiClient
-              ? currentPollbookService.apiClient
-              : createPeerApiClientForAddress(`http://${resolvedIp}:${port}`);
+          const machineUrl = reuseApiClient
+            ? assertDefined(currentPollbookService.address)
+            : advertisedServiceUrl;
+
+          const apiClient = reuseApiClient
+            ? assertDefined(currentPollbookService.apiClient)
+            : createPeerApiClientForAddress(machineUrl);
```
kshen0, line 344: wondering if it might be slightly easier to follow if we just structure this as `if (reuseApiClient) { ... } else { ... }` instead of inlining ternary operators multiple times. just my personal preference, not blocking.

apps/pollbook/backend/src/peer_store.ts
```diff
@@ -271,6 +228,31 @@ export class PeerStore extends Store {
...
+      // We are now offline, we should mark all other services as offline.
+      for (const [otherPollbookName, otherPollbookService] of Object.entries(
+        this.connectedPollbooks
+      )) {
+        if (otherPollbookName === avahiServiceName) {
+          continue;
+        }
+        if (
+          CommunicatingPollbookConnectionStatuses.includes(
+            otherPollbookService.status
```
kshen0, line 245: do we want to leave peers in `ShutDown` status? is it more accurate to say they're `LostConnection` (and we don't know if they are still powered down or have been powered up)? or does it not really make a difference?

apps/pollbook/backend/src/networking.ts
carolinemodic: I find ternary operators better when setting variables though so you can use const

PR #7065, opened 2025-08-19
VxDesign: Add ability to upload signature and caption at the election level for NhBallots

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/7056

Previously clerk signature images and captions could only be added to elections/ballots for split precincts. Now for NhBallots, we want the parent election to have signature fields, with the split precincts still overriding if provided.

## Demo Video or Screenshot

**Election missing signature**
https://github.com/user-attachments/assets/19292b32-ffb0-4ccd-884e-b5330fc1e4f7

- Error when viewing ballot with missing signature
- Error when downloading ballots with missing signature

**Election with signature and precinct override**
https://github.com/user-attachments/assets/c4d10015-1953-48ee-bbef-4d79b6de7bc9

## Testing Plan

Automated tests added for new functionality 

Tests updated for:
- Updating election info
- Ballot PDF preview of Precinct split signatures, now it overrides parent election signatures

## Checklist


apps/design/frontend/src/signature_image_input.tsx
```diff
@@ -0,0 +1,44 @@
+import styled from 'styled-components';
+import { images } from '@votingworks/ui';
+import { ImageInput, ImageInputProps } from './image_input';
+
+const StyledImageInput = styled(ImageInput)`
+  img {
+    width: 10rem;
```
nikhilb4a, line 7: This means the precinct split image will now be larger, as it was previously set to 3rem, which was very hard to see. The Seal right above it was 10rem, so now they are consistent.

apps/design/frontend/src/layout.tsx
```diff
@@ -31,16 +31,28 @@ export const FieldName = styled.div`
...
   label: string;
   children: React.ReactNode;
+  isOptional?: boolean;
 }): JSX.Element {
   return (
     <label>
-      <FieldName>{label}</FieldName>
+      <div style={{ display: 'flex', alignItems: 'center', gap: '0.25rem' }}>
+        <FieldName>{label}</FieldName>
+        {isOptional && <FieldTag> — Optional</FieldTag>}
```
nikhilb4a, line 54: Will post in Slack for convo on the right ui design for these as optional fields

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -70,8 +70,8 @@ function Header({
 
   electionTitleOverride,
   electionSealOverride,
-  clerkSignatureImage,
-  clerkSignatureCaption,
+  clerkSignatureImage: clerkSignatureImageOverride,
```
nikhilb4a, line 73: I added these to communicate that these are overrides now - could I get input on if you'd suggest more of a comment or even removing these variable renames? I can see an argument from all the above.

I decided to not rename the fields within the `NhPrecinctSplitOptions` type that provides these props to avoid making a backwards incompatible change that would require a migration. But open to that if we'd prefer

nikhilb4a:
@jonahkagan I've updated the PR to make the signature inputs required. Could I get your opinion on which of the following changes are neccessary:
1. Update the [electionGridLayoutNewHampshireHudson](https://github.com/votingworks/vxsuite/blob/main/libs/fixtures/data/electionGridLayoutNewHampshireHudson/election.json) and the [electionGridLayoutNewHampshireTestBallot](https://github.com/votingworks/vxsuite/tree/main/libs/fixtures/data/electionGridLayoutNewHampshireTestBallot/election.json)
2. When changing from NhBallot to VxDefaultBallot, should we delete the signature fields from the election object? Or is it fine to leave them their unused, so if a user is quickly toggling, we don't drop the fields
3. Should the nh_ballot_template treat the election.signature

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -102,6 +102,13 @@ function Header({
       ? assertDefined(getPartyForBallotStyle({ election, ballotStyleId }))
       : undefined;
 
+  const clerkSignatureImage =
+    clerkSignatureImageOverride ?? election.signatureImage;
+  // clerkSignatureCaptionOverride will be an empty string if it is
+  // entered and then deleted, so treat it as undefined
+  const clerkSignatureCaption =
+    clerkSignatureCaptionOverride || election.signatureCaption;
```
nikhilb4a, line 110: The `election.signatureImage` and `election.signatureCaption` are still treated as optional in this template for backwards compatibility. In the future we could `assert` these fields

nikhilb4a:
@jonah Commits 6 to 12 include the proposed changes:
- Signature is updated to be a nested object
- Signature is required for the NhBallot template
- Helpful error message shown when ballot errors when viewing
- Helpful error message shown when exports fail (including now for contestTooLong)
- Callout shown at ballotTemplate selection to indicate missing required field
- VxDefaultBallot election def exports do not include election.signature data

apps/design/frontend/src/ballot_screen.tsx
```diff
@@ -314,6 +314,30 @@ export function BallotScreen(): JSX.Element | null {
...
+                  >
+                    <Callout color="danger" icon="Danger">
+                      <span>
+                        The election is missing the following required field:
+                        &quot;{err.field}&quot;. Update in{' '}
+                        <Link
+                          to={routes.election(electionId).electionInfo.path}
+                        >
+                          Election Info
+                        </Link>
```
jonahkagan, line 334: I like the inclusion of the link to fix the issue!

I think it's a little early to be abstracting a pattern for missing required fields, since they aren't all updated in the Election Info screen and may need different instructions.  Maybe we just make a single specific error type and user-facing message for missing signatures to start and see if we need a more general solution later? Then the message could be worded a little more directly, e.g. "Missing signature. Upload a signature in Election Info."

apps/design/frontend/src/ballot_screen.tsx
```diff
@@ -314,6 +314,30 @@ export function BallotScreen(): JSX.Element | null {
 
             if (ballotResult.isErr()) {
               const err = ballotResult.err();
+              if (err.error === 'missingRequiredField') {
```
jonahkagan, line 317: Maybe `switch (err.error)` now that we have multiple cases

apps/design/frontend/src/export_screen.tsx
```diff
@@ -176,22 +211,43 @@ export function ExportScreen(): JSX.Element | null {
...
+                      &quot;signature&quot;. Update in{' '}
+                      <Link to={routes.election(electionId).electionInfo.path}>
+                        Election Info
+                      </Link>
+                      .
+                    </span>
+                  </Callout>
+                </Row>
+              )}
+            </React.Fragment>
```
jonahkagan, line 250: I think switching from VX to NH template for an existing election is going to be a rare enough edge case that this warning isn't necessary (given that it's already guarded against in the places where we render the ballot template)

apps/design/frontend/src/export_screen.tsx
```diff
@@ -258,7 +314,7 @@ export function ExportScreen(): JSX.Element | null {
         </P>
         {exportError && (
           <P>
-            <Icons.Danger /> An unexpected error occurred. Please try again.
+            <Icons.Danger /> {exportError}
```
jonahkagan, line 317: Can we switch this to the danger callout pattern while we're here?

apps/design/frontend/src/export_screen.tsx
```diff
@@ -31,18 +32,49 @@ import {
...
+import { Column, FieldName, InputGroup, Row } from './layout';
 import { useTitle } from './hooks/use_title';
 
 const ballotTemplateOptions = {
   VxDefaultBallot: 'VotingWorks Default Ballot',
   NhBallot: 'New Hampshire Ballot',
 } satisfies Record<BallotTemplateId, string>;
 
+function parseExportError(error: string): string {
+  // Try to parse as JSON first, in case it's a stringified BallotLayoutError
```
jonahkagan, line 49: I was surprised to see that this error got stringified and is parseable. I think it's happening by accident (a fallback case of the `extractErrorMessage` function), not by intention, which makes me a bit worried about relying on it.

I found this [comment](https://github.com/votingworks/vxsuite/blob/main/libs/hmpb/src/render_ballot.tsx#L499-L503) (which I think I wrote) that indicates that we throw the `Err` result specifically because we don't think we need to show it to the user.

I actually think there's something to this comment, since we do expect proofing to be completed and ballots to be finalized before clicking export. So maybe the generic message was actually ok, since it really shouldn't be happening in normal use cases.

The reason I'm considering going back to a generic message is because it would be at least some more work to add plumbing to intentionally pass the `Err` result through the task pipeline to the frontend.

What do you think?

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -103,6 +105,16 @@ function Header({
...
 
+  // Signature is guaranteed to exist due to validation in BallotPageFrame
+  assert(election.signature);
+  const clerkSignatureImage =
+    clerkSignatureImageOverride ?? election.signature.image;
+  // clerkSignatureCaptionOverride will be an empty string if it is
+  // entered and then deleted, so treat it as undefined and fallback
+  // the election signature
+  const clerkSignatureCaption =
+    clerkSignatureCaptionOverride || election.signature.caption;
```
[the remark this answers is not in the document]
jonahkagan: I don't think we need to support an empty caption override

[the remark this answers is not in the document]
jonahkagan: If we're doing trimming, I think we should do it at the input form

apps/design/backend/src/worker/generate_election_package_and_ballots.ts
```diff
@@ -106,9 +106,15 @@ export async function generateElectionPackageAndBallots(
 
   const formattedElection = formatElectionForExport(election, ballotStrings);
 
+  // For non-NH ballot templates, remove the signature field as it is unused
+  const electionForTemplate =
+    ballotTemplateId === 'NhBallot'
+      ? formattedElection
+      : { ...formattedElection, signature: undefined };
```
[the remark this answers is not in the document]
nikhilb4a: Hm, that's a good point. 

Realistically, keeping the signature data is mostly for ease-of-life of internal users when switching between ballots templates, since it isn't common to change a template, so I don't want to introduce risk because of it. 

I think removing the field when loading elections sounds simple enough, so I'll go with that, but if I suspect any more edge cases I'd just go with deleting the data when switching templates

apps/design/frontend/src/ballot_screen.tsx
nikhilb4a: Thanks for that feedback, I was feeling uncertain on which path to take, but I do agree with what youv'e said. I'll go with a single specific error type!

apps/design/frontend/src/export_screen.tsx
nikhilb4a: Yeah, I agree with going back to a more generic error message for the points you've raised 

As I think more about it, if we want to communicate why an election package can't be exported because of a missing field, that should probably be done before attempting to start it, not once it's failed, for an error that is static like this and not due to a dynamic calculation like `ContestTooLong`. 

So, if anything, I think we should disable the export buttons with a tooltip/message preventing downloads. But, I don't feel strongly about that yet, as it would then be a different pattern than "View Ballot". We could also disable those buttons too, but I am more partial to how easy to digest the ballot viewer error message is since it's full screen

jonahkagan: Those are fair points! I think we'll need to think more about these kinds of concerns when we expose the export functionality to election officials. Really I'd want them to catch any issues before finalizing.

With this specific error case, I'm not so worried about it though, since end users don't have the ability (or motivation) to change templates.

apps/design/frontend/test/fixtures.ts
```diff
@@ -87,8 +87,7 @@ export function electionInfoFromElection(election: Election): ElectionInfo {
     state: election.state,
     jurisdiction: election.county.name,
     seal: election.seal,
-    signatureImage: election.signatureImage,
-    signatureCaption: election.signatureCaption,
+    signature: election.signature,
```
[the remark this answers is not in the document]
jonahkagan: I think it makes a lot of sense for `Election.signature` to be a nested object, because we want to model the constraint that you either have both image+caption or neither.

`ElectionInfo` only serves this specific form, so I think it makes sense to use the data model that makes the form code as simple as possible.

libs/ballot-interpreter/src/__image_snapshots__/interpret-vx-hmpb-test-ts-hmpb-nh-general-election-letter-marked-ballot-interpretation-1-snap.png
nikhilb4a, line 1: Not certain why only these two snaps had their `electionId` update, do you recommend I look into it further?

Issue #6958, opened 2025-08-11
VxMark: Plugging headphones into the PAT jack registers input


arsalansufi:
Parent cert discrepancy issue: https://github.com/votingworks/cert-discrepancies/issues/200

PR #7083, opened 2025-08-21
VxMarkScan: ignore snp signal for 2 seconds after positive connection status

kshen0:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6958

When headphones are plugged into the PAT jack on the VSAP 150 it registers as a connected PAT device and sends a sip and a puff. In the app this causes a single `Tab` and `Enter` keypress ie. unwanted UI interaction.

To fix this I updated the daemon to not send sip/puff keypresses for 2 seconds after PAT connection (or headphone connection that registers as PAT connection).

## Demo Video or Screenshot

Here's a real sip & puff working as expected on this branch by sending `1` and `2` keypress events when sip or puff happens. Note that missed events are because the jack on this VSAP unit is faulty.

https://github.com/user-attachments/assets/f00e3bfa-58a3-4263-b6d2-f5010ea37462

Broken behavior showing `1` and `2` keypress events when headphones are plugged in.

https://github.com/user-attachments/assets/d332ec35-7a97-4443-910a-0732746be3ee

Fixed behavior

https://github.com/user-attachments/assets/72fc1a03-7b11-4302-8328-be4cf0922fef

## Testing Plan

Tested manually on hardware

### DIFF
--- a/apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
+++ b/apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
@@ -254,6 +254,7 @@ struct CurrentStatus {
     sip: SipAndPuffSignalStatus,
     puff: SipAndPuffSignalStatus,
     sip_puff_device_connected: SipAndPuffDeviceStatus,
+    sip_puff_device_connection_time: Option<Instant>,
 }
 
 struct KeypressSpec {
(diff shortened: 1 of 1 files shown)


apps/mark-scan/fai-100-controller/src/fai_100_controllerd.rs
```diff
@@ -381,14 +387,22 @@ fn handle_status_response(
...
+        .sip_puff_device_connection_time
+        .is_some_and(|t| t.elapsed() > Duration::from_secs(2));
+
     // Only check for sip & puff actions when the device is connected because
-    // sip/puff values are inverted when no device is connected, per comment above.
-    // Even if values were consistent, logically no sip/puff signal can be sent without a
-    // connected device.
+    // sip/puff values are inverted when no device is connected.
+    // Even if values were consistent, logically no sip/puff signal can be sent without a connected device.
     if current_status.sip_puff_device_connected == SipAndPuffDeviceStatus::Connected {
```
kshen0, line 401: I originally made a single `connection_signal_delay_elapsed` check here but the effect was the stray keypresses were only delayed, not ignored. We could increase the delay, but gating just the `send_keystroke` calls while keeping state updates allows us to use a shorter delay (for the benefit of actual snp users) while still solving the problem.

kofi-q (approved):
Ah, nice! Just noticed a similar issue recently while testing the new VxMark - might follow this pattern to resolve that, if we can successfully intercept the HID keyboard events from that device.

PR #7095, opened 2025-08-25
VxPollBook: Expand networking tests primarily to cover more of the offline/online logic

carolinemodic:
## Overview
Adds more testing to the networking tests to test the offline/online behavior, the behavior with ignoring a ipv6 address, and the cleanupStalePollbookServices function running to set machines to LostConnection and/or the self to offline. 

The amount of tests in app_multi_node was getting hairy so I also split out the network configuration tests to their own file and moved shared helpers to a util file. 

## Testing Plan
Ran Tests

## Checklist


carolinemodic:
Updated, the test that existed prior to this PR is still testing a few different things but the transitions its testing are more intertwined so its simpler to just keep that one as is for now I think.

Issue #7070, opened 2025-08-20
VxDesign: Remove Auth0 from the critical path


arsalansufi:
https://votingworks.slack.com/archives/C08AWNY93PB/p1755707156982819
https://votingworks.slack.com/archives/C08AWNY93PB/p1756145961693719

PR #7099, opened 2025-08-26
VxDesign: Improve resilience to Auth0 outages

jonahkagan:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7070

Auth0 is the source of truth for organizations. VxDesign currently loads organizations from the Auth0 API in a few different places:
1. On every API request, load the user's organization name 
2. When listing elections, load each election's organization name
3. When creating/cloning an election, load a list of organizations to select the new election's org (only for users with access to all orgs)

This is problematic if Auth0 has an outage, since it will impede users from using VxDesign (in particular, users that are already logged in and have a stored session cookie). To prevent this, this PR makes a few changes:
1. Stop loading the user's organization name - we weren't actually using it
2. Cache the organizations in the database and load them from there instead as needed

## Demo Video or Screenshot
N/A

## Testing Plan
- Manual testing
- Updated automated tests

## Checklist

### DIFF
--- a/apps/design/backend/migrations/1756159958924_organizations-table.js
+++ b/apps/design/backend/migrations/1756159958924_organizations-table.js
@@ -1,0 +1,45 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const { ManagementClient } = require('auth0');
+const { assertDefined } = require('@votingworks/basics');
+
+/**
+ * @type {import('node-pg-migrate').ColumnDefinitions | undefined}
+ */
+exports.shorthands = undefined;
+
+/**
+ * @param pgm {import('node-pg-migrate').MigrationBuilder}
+ * @param run {() => void | undefined}
+ * @returns {Promise<void>}
+ */
+exports.up = async (pgm) => {
+  pgm.createTable('organizations', {
+    id: { type: 'text', primaryKey: true },
+    name: { type: 'text', notNull: true, unique: true },
+  });
+
+  if (process.env.NODE_ENV === 'production' || process.env.AUTH_ENABLED) {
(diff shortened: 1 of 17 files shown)


apps/design/backend/migrations/1756159958924_organizations-table.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param run {() => void | undefined}
+ * @returns {Promise<void>}
+ */
+exports.up = async (pgm) => {
+  pgm.createTable('organizations', {
+    id: { type: 'text', primaryKey: true },
+    name: { type: 'text', notNull: true, unique: true },
+  });
+
+  if (process.env.NODE_ENV === 'production' || process.env.AUTH_ENABLED) {
```
jonahkagan, line 21: The `pnpm db:reset-dev` command runs all of the migrations, so we want to make sure that still works even if we're not hooked up to Auth0 in dev

apps/design/backend/src/store.ts
```diff
@@ -427,6 +428,45 @@ export class Store {
     return new Store(new Db(logger));
   }
 
+  /**
+   * Takes the organizations from Auth0 (the source of truth) and caches them in
+   * the database, adding/removing/updating records as necessary.
+   */
+  async cacheOrganizations(organizations: Org[]): Promise<void> {
```
nikhilb4a, line 435: nit: Maybe `syncOrganizations` to indicate we also are deleting outdated orgs?

nikhilb4a (approved):
Very nice, really significant resiliency improvement!

kofi-q (approved):
🚢

Issue #6771, opened 2025-07-11
VxMarkScan/Scan/CentralScan: Add UPS diagnostic to confirm battery levels manually


mattroe:
Parent issue: https://github.com/votingworks/cert-discrepancies/issues/166

Summary: Add a section to VxMarkScan/VxScan/VxCentralScan readiness reports that prompts the user to "Confirm UPS is fully charged" and indicates the UPS status once that confirmation is complete. Let's follow existing readiness report patterns.

arsalansufi:
@kshen0 You can skip VxMark! That code is in a partial state that we're gonna revisit later as we bring code in from Kofi's demo branch. For now, you can focus on VxMarkScan, which is all we're certifying right now. Let me update the ticket title and description accordingly

PR #7112, opened 2025-08-27
VxMarkScan: readiness report - UPS instructions

kshen0:
## Overview

Part of https://github.com/votingworks/vxsuite/issues/6771

Ticket says copy should be "Confirm UPS is fully charged". I added a note about connection as well and replaced the "UPS" abbreviation with the full term, but can adjust as needed.

In a future PR for other VxSuite apps I'll probably need to shift things around with the shared component.

## Demo Video or Screenshot

Screenshot 2025-08-26 at 8 46 09 PM

https://github.com/user-attachments/assets/7d4b5d67-e35f-446f-a334-56b76b43fb6b

## Testing Plan

- manually tested all flows
- added tests

## Checklist


libs/ui/src/diagnostics/uninterruptible_power_supply.tsx
```diff
@@ -0,0 +1,35 @@
...
+          Confirm the uninterruptible power supply is connected and fully
+          charged.
+        </P>
+        <ButtonBar style={{ marginTop: '0.5rem' }}>
+          <Button icon="Done" onPress={passTest}>
+            UPS is Fully Charged
+          </Button>
+          <Button icon="Delete" onPress={failTest}>
+            UPS is not Fully Charged
+          </Button>
```
arsalansufi, line 30: nit: For true title case, I think we want `UPS Is Fully Charged` and `UPS Is Not Fully Charged` (because I can never remember exact title case rules I used https://titlecaseconverter.com)

arsalansufi (commented):
Just one nit but otherwise looks great! Avoiding the abbreviation where you have makes sense to me

arsalansufi (approved):
Proactively approving to keep you unblocked once you tweak the casing!

PR #7121, opened 2025-08-27
USB Handling: fix mount wedge state

### LINKED ISSUE #7092: bug: USB detection can become wedged, and require restart, with too much inserting and removing
@adghayes: I've seen this over the years on various machines. It is not specific to VxPollBook, where it was last reported. I was able to reproduce quite easily by inserting and removing the USB drive until it was no longer detected. I think there is a race condition in the state management.

Issue automatically created from Slack feedback reported by pius: 
https://votingworks.slack.com/archives/C098ZSZ7719/p1755885716933569?thread_ts=1755885716.933569&cid=C098ZSZ7719

adghayes:
## Overview

Closes #7092. The bug is that, when we mount a USB drive, we acquire a lock (mutex) in the `libs/usb-drive` code. For ejecting and formatting, the lock is released in a `finally` block. Since our mounting takes place during a `status` call, which we don't want to hang, we `void` the promise and let the lock persist between calls. The lock is only released once we detect the mount point via `lsblk` and count the drive as mounted, which can take a second. If you remove the USB drive after we successfully mounted i.e. `mount.sh` has exited but _before_ the mount point is detected during `lsblk`, the lock is never released and USB drives cannot be mounted and used until the machine is restarted.

The fix is, as part of our `mount` logic, polling `lsblk` until we get a mount and calling the lock release in all cases. We have to decide how long to wait for a mount point, and I settled on 5 seconds based on my experience of it taking up to a few seconds. What if, on a machine like VSAP, `lsblk` somehow takes longer than 5 seconds? Subsequent polling will trigger re-mounts, which will fail, with errors swallowed and logged. Once the original mount is registered by `lsblk`, the mount point will be seen and it will be mounted. Seems safe to me.

The underlying challenge here is the delay with `lsblk`. According to Linux, the drive is mounted and can be safely used after `mount` exits successfully - we don't have to wait for the `lsblk` status. We do wait for the `lsblk` status to avoid having more internal state. The delay in `lsblk` is due to kernel buffering something something, and it brings me back to wondering whether there's a lower level way to detect USB devices that we should be using. I'm pretty sure we could rewrite this to access the system files like `/sys/class/block` and `/proc/mounts` and we'd mount USB drives and detect changes in general a lot faster. But we might want to save that rewrite for when re-consider polling here or have a larger release.

## Testing Plan

Updated testing. Separated getting USB drive data from the system into a separate file and test suite. Handling everything in one file was making mocking in tests really confusing due to a bunch of different kinds of `exec` calls. With polling, it became impossible.

## Checklist

### DIFF
--- a/libs/usb-drive/src/block_devices.test.ts
+++ b/libs/usb-drive/src/block_devices.test.ts
@@ -1,0 +1,188 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { promises as fs, PathLike } from 'node:fs';
+import { BlockDeviceInfo, getUsbDriveDeviceInfo } from './block_devices';
+import { exec } from './exec';
+
+const readdirMock = vi.mocked<(path: PathLike) => Promise<string[]>>(
+  fs.readdir
+);
+const readlinkMock = vi.mocked(fs.readlink);
+const execMock = vi.mocked(exec);
+
+vi.mock(
+  import('node:fs'),
+  async (importActual): Promise<typeof import('node:fs')> => {
+    const actual = await importActual();
+    return {
+      ...actual,
+      promises: {
+        ...actual.promises,
+        readdir: vi.fn().mockRejectedValue(new Error('readdir not mocked')),
+        readlink: vi.fn().mockRejectedValue(new Error('readlink not mocked')),
+      },
+    };
+  }
+);
+
+vi.mock(
+  import('./exec.js'),
+  async (importActual): Promise<typeof import('./exec')> => ({
+    ...(await importActual()),
+    exec: vi.fn().mockRejectedValue(new Error('exec not mocked')),
+  })
+);
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function lsblkOutput(devices: Array<Partial<BlockDeviceInfo>> = []) {
+  return {
+    stdout: JSON.stringify({
+      blockdevices: devices.map((device) => ({
+        name: 'sdb1',
+        mountpoint: '/media/usb-drive-sdb1',
+        fstype: 'vfat',
+        fsver: 'FAT32',
+        label: 'VxUSB-00000',
+        type: 'part',
+        ...device,
+      })),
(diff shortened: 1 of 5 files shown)


jonahkagan:
Ah gotcha, thanks for clarifying. I wonder if we could just keep using the existing polling but add something like this:

```ts
void mount(deviceInfo, logger).catch(releaseActionLock);
setTimeout(() => {
   if (actionLock.current === 'mounting') releaseActionLock();
}, 5_000)
```

adghayes:
That's a good idea. I think the only hole is that the `mounting` lock could be a different one than originally kicked off the timeout if the drive was mounted, removed, and re-inserted. So I think we would want to track the timeout as internal state and cancel it if the thing is actually mounted. Thoughts on that? I think it's viable, although of course could create new state management bugs if not carefully done.

jonahkagan:
Good point. We might want to consider moving to a different model for the lock then, where acquiring the lock returns a one-time-use function to release the lock. I think we already have most of this implemented in the mutex in libs/utils, though it looks like it throws if the release function is called multiple times, so we might have to swallow that error.

adghayes:
I took a look at the `Mutex` implementation - it looks like the value can only be read if you have the lock. In other words, it's a single reader lock. The polling needs to be able to access the lock value to differentiate between mounting and formatting here. I tried the `useTimeout` approach and it seems to fix the issue 🤷. I think options so far here are:

- significant changes to Mutex to enable multiple readers for value
- my initial approach (see first commit on this branch)
- the setTimeout (see latest on this branch)

I actually find the `setTimeout` approach, built on our current approach, harder to reason through because it forces splitting the opening and closing of the lock across different instances of `status`.

adghayes:
What exactly is the problem with the initial approach? Or were you just reaching for something simpler?

jonahkagan:
I was just exploring the solution space. I think I agree with you that the original approach is the clearest. It makes `mount` behave more like you'd expect - it blocks until the USB drive is mounted.

arsalansufi:
Okay that's good to hear! With that in mind, I'm comfortable including it. FYI @pieguyatx that the alpha images that you're currently working with don't yet include this fix

PR #7125, opened 2025-08-28
VxDesign: Reset db migration ordering

nikhilb4a:
## Overview

Other migrations were run since I initially created the migration to add election signatures, so the new migration needs a timestamp/ordering that comes after the most recently added ones. I deleted the migration script and re-added it, so only the file name has changed.

I noticed this because the build failed to deploy on staging, so this PR should fix that.

## Checklist


nikhilb4a:
Yeah definitely, good call. 

It's nice that our Arlo tool handles this internally

nikhilb4a:
So if main has a migration script the branch is missing, the branch is out of date. It's not okay to be out of date if the branch has a new migration script. Sound right?

Issue #7089, opened 2025-08-22
Enable ballot print scale warnings


arsalansufi:
There are a number of cert discrepancies that all revolve around ballot print scale issues. Let's help close them all out by enabling these warnings.

We could add `minimumBallotScale: 0.975` to system settings by default from here on out. We'll have to walk SLI through that addition to their existing election packages.

To avoid requiring SLI to update election packages, we could also enable the warnings no matter what and replace `minimumBallotScale` with `minimumBallotScaleOverride`. Up for discussion.

Issue #6672, opened 2025-06-23
VxScan: remove references to PDI scanner bitonal threshold setting


arsalansufi:
A follow-up to https://github.com/votingworks/vxsuite/pull/6667#issuecomment-2989060333

PR #7128, opened 2025-08-28
Rename minimumDetectedScale minimumDetectedBallotScaleOverride

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7089

This PR revisits how we want to enable our ballot scale warnings in v4.0.3 as it pertains to system settings. Because the mistake of printing at too low of a scale is so easy to make, internally and at SLI, I want this to be enabled by default.

I've accordingly opted for the following approach:
- Rename the `minimumDetectedScale` system setting `minimumDetectedBallotScaleOverride`. (The notable addition to the name is "Override". The addition of "Ballot" is just for clarity.)
- When the `minimumDetectedBallotScaleOverride` system setting isn't specified at all, default to a minimum detected ballot scale of 0.985.
- When the `minimumDetectedBallotScaleOverride` system setting is specified, use that. The check can be entirely disabled by setting `minimumDetectedBallotScaleOverride` to 0 (or really any low value).

The obvious alternative is to have a `minimumDetectedBallotScale` (no "Override") system setting that has to be specified for the check to be enabled. And to disable the check by not specifying a `minimumDetectedBallotScale`.

The reason I'm leaning toward the former is that I don't want SLI to have to edit their existing election packages to see these warnings. Another secondary reason is that I want this on for elections on v4.0.3+ but don't want it accidentally on for elections on v4.0.2, which contains the earlier version of this code that we aren't 100% confident in. With the alternative, we'd have to take care to specify this setting for cert (v4.0.3+) but not for fall elections (v4.0.2), and that feels risky.

While I was at it, I performed some continued partial cleanup of the no longer needed bitonal threshold system setting.

## Testing Plan

Will cover this feature in general during QA. Relied on automated tests for the scope of this PR.

## Checklist


libs/types/src/system_settings.ts
```diff
@@ -148,12 +148,6 @@ export const DEFAULT_MARK_THRESHOLDS: Readonly<MarkThresholds> = {
   writeInTextArea: 0.05,
 };
 
-/**
- * The default bitonal threshold for scanning ballots.
- * See Section 2.1.43 of the PDI PageScan software specification.
- */
-export const DEFAULT_BITONAL_THRESHOLD = 75;
```
arsalansufi, line 155: Still not completely removed from lower levels in the stack but continued the removal while I was passing through (reminder that this is no longer needed because we now scan in grayscale rather than B&W)

apps/central-scan/backend/src/importer.ts
```diff
@@ -224,7 +225,9 @@ export class Importer {
           adjudicationReasons: store.getAdjudicationReasons(),
           markThresholds,
           allowOfficialBallotsInTestMode,
-          minimumDetectedScale,
+          minimumDetectedScale:
+            minimumDetectedBallotScaleOverride ??
+            DEFAULT_MINIMUM_DETECTED_BALLOT_SCALE,
```
arsalansufi, line 230: I considered pushing this logic even further down the stack into the Rust code, putting both the default and override handling there. But decided to keep this change minimal for lack of time.

Handling at this level works pretty well, too, since TS reminds you to convert from the `minimumDetectedBallotScaleOverride` system setting to the `minimumDetectedScale` that the interpreter's public interface expects.

apps/design/frontend/src/system_settings_screen.tsx
```diff
@@ -313,35 +313,36 @@ export function SystemSettingsForm({
                 />
               </InputGroup>
             )}
-            {features.MINIMUM_DETECTED_SCALE_OPTION && (
```
arsalansufi, line 316: I was previously feature flag gating this because I didn't want SLI turning this on in v4.0.2, which as noted earlier has an implementation of this check that we weren't 100% confident in. But now safe to remove the feature flag.

libs/types/src/system_settings.ts
jonahkagan: Thanks for continuing this removal! Should we remove lower as well? (not necessarily in this PR)

apps/central-scan/backend/src/importer.ts
jonahkagan: Seems fine for now!

libs/types/src/system_settings.ts
```diff
@@ -148,11 +149,7 @@ export const DEFAULT_MARK_THRESHOLDS: Readonly<MarkThresholds> = {
...
+export const DEFAULT_MINIMUM_DETECTED_BALLOT_SCALE = 0.975;
```
jonahkagan, line 152: I don't remember how we landed on this value but assuming you got it from somewhere

jonahkagan (approved):

libs/types/src/system_settings.ts
arsalansufi: Yeah I think we should! We have this tech-improvement issue open until we complete that: https://github.com/votingworks/vxsuite/issues/6672

libs/types/src/system_settings.ts
jonahkagan: Sounds good!

PR #7129, opened 2025-08-28
VxScan: UPS diagnostic

kshen0:
## Overview

Part of https://github.com/votingworks/vxsuite/issues/6771

Adds UPS manual diagnostic to VxScan.

Also updates the header on the diagnostic page in VxMark to say "Back to System Administrator Actions" instead of "Back to Poll Worker Actions". I can't remember if we ever linked to diagnostics from the poll worker page but it's linked from the system admin page now.

## Demo Video or Screenshot

The failure message is a little verbose but it seems valuable to be clear that the system does not have any information on UPS battery status.

Screenshot 2025-08-28 at 11 27 15 AM

https://github.com/user-attachments/assets/217c9438-6ba9-490c-b81c-fc13f629c2b6

## Checklist


apps/scan/backend/src/util/diagnostics.ts
```diff
@@ -38,3 +38,8 @@ export const TEST_PRINT_USER_FAIL_REASON =
 
 export const TEST_AUDIO_USER_FAIL_REASON =
   'The user indicated audio playback was not successful.';
+
+export const TEST_UPS_USER_PASS_REASON =
+  'The user confirmed the UPS was connected and fully charged.';
+export const TEST_UPS_USER_FAIL_REASON =
+  'The user indicated the UPS was not connected or not fully charged.';
```
arsalansufi, line 45: :100:

Yeah I think you're right to include that the user indicated this and that the system doesn't actually have access to UPS status

apps/scan/frontend/src/api.ts
```diff
@@ -508,7 +508,7 @@ export const logTestPrintOutcome = {
 
 export const getMostRecentAudioDiagnostic = {
   queryKey(): QueryKey {
-    return ['getMostAudioDiagnostic'];
+    return ['getMostRecentAudioDiagnostic'];
```
arsalansufi, line 511: Nice catch 🔍

apps/scan/frontend/src/components/ups_diagnostic_modal_button.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { DiagnosticOutcome } from '@votingworks/types';
+import { Button, Modal, P } from '@votingworks/ui';
+import * as api from '../api';
+
+export interface UpsDiagnosticModalProps {
+  setVisible: (visible: boolean) => void;
+}
+
+export function UpsDiagnosticModal(
```
arsalansufi, line 10: I see in https://github.com/votingworks/vxsuite/pull/7131 we add a version of this modal to `libs/ui`

Can we use that one here as well just to make sure copy remains consistent across the apps? Maybe that was already your plan and you're just planning to merge the VxCS PR and rebase this one on top or vice versa

libs/ui/src/diagnostics/uninterruptible_power_supply_section.tsx
```diff
@@ -0,0 +1,44 @@
+import { assert } from '@votingworks/basics';
+import { DiagnosticRecord } from '@votingworks/types';
+import { H2, P } from '../typography';
+import { InfoIcon, SuccessIcon, WarningIcon } from './icons';
+
+export interface UpsSectionProps {
+  upsSectionContents?: JSX.Element;
```
arsalansufi, line 7: What is this prop used for? I might be missing something about how the diagnostics pages are wired up

arsalansufi (approved):
One thought around code sharing, otherwise LGTM!

apps/scan/frontend/src/components/ups_diagnostic_modal_button.tsx
arsalansufi: Ah I see now that the VxCS PR is built on top of this PR!

I'm sure you're already aware, but you can change the base branch of the VxCS PR to this PR's branch. Then when you merge this PR, the VxCS PR's base branch _should_ auto-update to be main

kshen0: Yeah sorry I meant to open that PR against this branch. For some reason I always expect github to open a PR for B against A when B branches off A, even though it has no way of knowing that's the user's intent.

libs/ui/src/diagnostics/uninterruptible_power_supply_section.tsx
kshen0: `*Section` diagnostic components are used in 2 places, the backend rendering of the PDF and the frontend rendering of the UI components. It's useful for each place to pass children (or not). In the VxCentralScan PR I pass a button to toggle the modal from the frontend, but not the backend PDF renderer.

In practice, it feels like the diagnostic pages are pretty disjointed across all the apps, so this pattern seems most used in VxCentralScan and least used in VxMarkScan.

PR #7133, opened 2025-08-29
VxDesign: Validate migration ordering

nikhilb4a:
## Overview

[Slack thread](https://votingworks.slack.com/archives/C085YT4798C/p1756397404136719)

Validate the ordering of VxDesign migrations using a script run in CI before backend tests are run. The script can also be run locally with an argument provided (main/branch) to indicate which version of the check to run.

On main - 
- Fail if a new migration was added in the last commit with a timestamp earlier than what was previously on main

On non-main - 
- Fail if a new migration was added to the branch and the branch is missing a migration from origin/main ([example failure](https://app.circleci.com/pipelines/github/votingworks/vxsuite/21961/workflows/dbe596c4-1728-4772-9a7d-07b709f345ef/jobs/964543))
- Fail if a new migration was added with a timestamp before origin/main's newest migration ([example failure](https://app.circleci.com/pipelines/github/votingworks/vxsuite/21968/workflows/9b15448e-6aaf-4846-8d75-b3aab88f265e/jobs/964923))

## Checklist


jonahkagan (approved):
Looks great! Excited to try it out. I have a small worry that we may struggle to maintain a bash script over time but we can see how it goes

nikhilb4a:
That worry makes sense! I'm definitely happy to own maintenance for the foreseeable future, and I won't be offended if we eventually conclude that we want to take a different approach, I appreciate that we are giving this a try and I feel pretty confident in it

Issue #6767, opened 2025-07-10
VxScan/Mark: Use color for all warning iconography


mattroe:
Parent issue: https://github.com/votingworks/cert-discrepancies/issues/105

PR #7137, opened 2025-09-02
Use two-tone warning icon to meet both voter-facing contrast and color requirements

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6767

We've been stuck between a rock and a hard place as it pertains to VVSG2 contrast requirements and VVSG2 color requirements. For the default voter-facing color mode, we need 10:1 contrast ratios. We also need to use yellow/orange for warning iconography. But achieving a 10:1 contrast ratio with yellow/orange on a white background requires darkening it, enough that it becomes brown.

In our initial cert submission, we decided to just use black for warning iconography, with our take on VVSG2 color requirements being that only _if_ you use color does warning iconography have to be yellow/orange. After some back and forth with SLI, we weren't able to get buy-in on that read.

This PR presents an alternative that I think may work. Rather than using the brown color (which SLI may still not consider okay as it's not really yellow/orange), I'm trying out a two-tone icon with a black border/text and a yellow interior. The white background and icon's black border have a 10:1 contrast ratio as do the icon's black border and yellow interior.

### Current icon

current

### Brown icon that meets 10:1 contrast ratio requirement

brown

### Two-tone icon that should also meet the contrast ratio requirement

two-tone

two-tone-2

## Testing Plan

- [x] Tested manually
- [x] Checked color contrast at https://www.audioeye.com/color-contrast-checker

## Checklist

### DIFF
--- a/libs/ui/src/themes/make_theme.ts
+++ b/libs/ui/src/themes/make_theme.ts
@@ -30,6 +30,7 @@ export const TouchscreenPalette = {
   Purple50: '#b591f3',
   Purple80: '#4d2692',
 
+  Orange20: '#ffc457',
   Orange50: '#ec791e',
   Orange80: '#5c3600',
 
@@ -154,6 +155,7 @@ type TouchscreenColorTheme = Pick<
   | 'danger'
   | 'successAccent'
   | 'warningAccent'
+  | 'warningContainer'
 >;
 
 /**
@@ -185,7 +187,7 @@ function expandToFullColorTheme(theme: TouchscreenColorTheme): ColorTheme {
     onDanger: theme.background,
     dangerContainer: theme.background,
 
-    warningContainer: theme.background,
+    warningContainer: theme.warningContainer,
 
     inverseBackground: theme.onBackground,
     onInverse: theme.background,
@@ -206,6 +208,7 @@ export const colorThemes: Record<ColorMode, ColorTheme> = {
     primary: TouchscreenPalette.Gray100,
     danger: TouchscreenPalette.Gray100,
     warningAccent: TouchscreenPalette.Gray100,
+    warningContainer: TouchscreenPalette.Gray100,
     successAccent: TouchscreenPalette.Gray100,
   }),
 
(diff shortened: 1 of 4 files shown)


kshen0 (approved):

libs/ui/src/themes/make_theme.ts
```diff
@@ -224,6 +228,7 @@ export const colorThemes: Record<ColorMode, ColorTheme> = {
     primary: TouchscreenPalette.Purple80,
     danger: TouchscreenPalette.Red80,
     warningAccent: TouchscreenPalette.Gray90,
+    warningContainer: TouchscreenPalette.Orange20,
```
jonahkagan, line 231: Why did you decide to add `warningContainer` instead of just change `warningAccent` and special casing _all_ use cases to add a black border?

jonahkagan: One option to look into might be using the font awesome "regular" style with a background color instead of the "solid" style

arsalansufi: Yeah I thought about this but ran into 2 (surmountable) hurdles in my limited time:

1. Most of the regular style icons (including exclamation triangle) require Font Awesome Pro. I actually bought a subscription for the team accordingly while investigating this issue. We'd just need to take some time to configure our app with the extra icons. Also, if we're using Font Awesome Pro, I think we should just use their provided duotone icons when possible.
2. Applying a background color to a regular style icon doesn't work out of the box but I'm sure can be solved for the different categories of shapes.

shape

arsalansufi: Landed on this for now: https://github.com/votingworks/vxsuite/pull/7137/commits/dd58f6974723d8c66878fc83c7895a1840dd90ac

Took care to make sure that we have no existing cases that'll trigger the assertion check - found just one case that would trigger: https://github.com/votingworks/vxsuite/pull/7137/commits/ba05d5a7acfe4050d611429fd7b0838220c66b23

jonahkagan: Ah ok yeah I see the wrinkles. This seems like a good enough solution for now. Thanks for sharing the additional context!

PR #7172, opened 2025-09-08
DRAFT: [VxQR] Set up reporting endpoint scaffolding in VxDesign

carolinemodic:
## Overview
This PR is currently in DRAFT form. The code is not written to be ready to merge but is meant to provide an example on how I am intending to implement this at a high level. 

The intention here:
 - /report - The endpoint whose URL will get encoded on the VxQR printed QR code. The url will accept query parameters with the reporting data, the endpoint will parse that data, verify the signature, and save the data to postgres. It will also generate a session ID for a cookie and set that cookie, then redirect to a results reporting confirmation page (currently /results, will likely be renamed) 
 - The frontend react App will switch and find any /results routes and return a seperate UI screen for that route without loading user / auth information. 
 - That results confirmation page from the frontend will call a backend GET endpoint which will load the session id from the cookie in order to associate the current session with the submitted results. The results will be returned back to the UI and rendered (not included in this PR) so the user can see the data they submitted. 

The major alternative I can think of to doing this cookie dance is to simply return templated html from the /report endpoint (similar to how we do in the current python backend: https://github.com/votingworks/quickresults-server/blob/main/server.py#L136 ) that includes all of the submitted results. I think its worth figuring out the approach outlined in this PR so that we can keep the frontend code in apps/design/frontend, and use our general react frameworks/libraries/etc. I also wasn't sure the easiest way to plug in templated html to the current express app setup. But I'm not entirely sure on the ideal way to do this so wanted to put up this proof of concept for discussion / alignment. 

A variation of this could be having a similar set up but having a multi-page react app with a different index.html / App.tsx file for the /results path. 

Note that there are two main new UIs we will need for VxQR, this confirmation page is one of them, the other is the aggregated results view. That second view WILL be under the authentication overhead so I imagine will be much more simply implemented in the SPA of apps/design/frontend with grout endpoints as needed for data in the backend. 

## Demo Video or Screenshot
You can test this with the review app, navigating to: /report should automatically redirect you to /results without requiring a log in. 

Review App: https://vxdesign-caro-vxqr-back-mm4ifq.herokuapp.com/ 
Reporting endpoint example: https://vxdesign-caro-vxqr-back-mm4ifq.herokuapp.com/report?p=ajshdkjahd&s=ajdksahd

## Testing Plan
With localhost and review app navigate to /report with and without query parameters, see redirect to /results work as expected with logs of query parameters and cookie session id as expected.

### DIFF
--- a/apps/design/backend/package.json
+++ b/apps/design/backend/package.json
@@ -56,6 +56,7 @@
     "@votingworks/ui": "workspace:*",
     "@votingworks/utils": "workspace:*",
     "auth0": "4.17.0",
+    "cookie-parser": "^1.4.7",
     "csv-parse": "^5.5.0",
     "debug": "4.3.4",
     "dotenv": "16.3.1",
@@ -74,6 +75,7 @@
     "zod": "3.25.42"
   },
   "devDependencies": {
+    "@types/cookie-parser": "^1.4.9",
     "@types/debug": "4.1.8",
     "@types/express": "4.17.14",
     "@types/fs-extra": "11.0.1",
--- a/apps/design/backend/src/app.ts
+++ b/apps/design/backend/src/app.ts
@@ -1,5 +1,6 @@
 import * as grout from '@votingworks/grout';
 import * as Sentry from '@sentry/node';
+import cookieParser from 'cookie-parser';
 import { auth as auth0Middleware, requiresAuth } from 'express-openid-connect';
 import path, { join } from 'node:path';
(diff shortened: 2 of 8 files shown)


carolinemodic:
Closing this while I pursue an attempt to instead grab the QR code URL from the frontend, parse out the query params and pass those through to a grout endpoint that does the "saving" work. While will hopefully be a bit simpler and avoid the need for a redirection / two url paths.

Issue #7160, opened 2025-09-04
VxDesign: rochester special rotation


mattroe:
https://votingworks.slack.com/archives/C08B80X72LE/p1756999056291039

PR #7216, opened 2025-09-16
VxDesign: Move candidate rotation to NH ballot template

jonahkagan:
## Overview

Previously, when using the NH ballot template, we would rotate the candidates according to NH statutes when saving a contest. This was a simple shortcut to ensuring that ballots listed the candidates in the correct rotated order _and_ that the resulting election definition listed candidates in the same order as the ballots (so that VxMark and tally reports would also list them in that order).

As we look ahead towards more complex rotation logic (e.g. rotating candidates differently for each precinct as needed for https://github.com/votingworks/vxsuite/issues/7160), it no longer makes sense to apply rotation on save, since there's no guaranteed single ordering for candidates across the entire election.

This PR moves the candidate rotation logic into the NH ballot template in preparation for more complex logic. For users of VxDesign, this means that in the contest form the candidates will appear in the order entered by the user, while in the ballots, they will be rotated.

In addition, since currently all candidate rotation _does_ still produce a single ordering across the entire election, and the rest of VxSuite uses the candidate order from `Election.contests` (rather than using the candidate order from the ballots specified in `Election.gridLayouts`), we add a temporary workaround to modify the candidate order in `Election.contests` to match that of `Election.gridLayouts` when exporting the election definition (only in cases where there is a single global ordering - looking ahead to cases where that isn't the case).

## Demo Video or Screenshot

https://github.com/user-attachments/assets/e9a6c3a6-492f-4d5f-8103-10eea773fe7f

## Testing Plan
- Relying mainly on existing tests mostly to show that in most cases, candidate order is unaffected
- Updated NH HMPB fixtures/snapshots since rotation logic will now apply anywhere we use the template
- Added a test to ensure that election definition candidate order matches rotated ballot order for NH ballots

## Checklist


libs/hmpb/src/render_ballot.tsx
```diff
@@ -615,8 +644,46 @@ export async function layOutBallotsAndCreateElectionDefinition<
...
+  // Temporary workaround for candidate rotation to ensure that VxMark's voting
+  // flow and tally reports in VxAdmin/VxScan list candidates in the same order
+  // that they appear on the HMPB. (Eventually, we should use the gridLayouts
+  // for that ordering instead of the election contests.)
+  //
+  // For each candidate contest: if all grid layouts have the same
+  // ordering of candidates, change the election definition to also have that
+  // ordering of candidates.
+  const contests = election.contests.map((contest) => {
+    if (template.isAllBubbleBallot) return contest;
```
nikhilb4a, line 656: Nice to see this field have other useful applications

nikhilb4a (approved):
This looks great! I like that the candidate rotation isn't happening as a side effect now at contest save. Feels well placed in the ballot template.

PR #7221, opened 2025-09-17
LA Demo: ballot PDF file parsing fix

adghayes:
## Overview

Allows supporting multi-lingual ballots with hyphens in the name. Not resilient to precincts with hyphens in them yet, will need to revisit for production.

## Testing Plan

- tested on our demo election
- tested on an LA election
- confirmed all the LA elections have precincts with no hyphens


arsalansufi (approved):
Thank you for finding and fixing!

Issue #6859, opened 2025-07-28
[VxMark] Switch from a per-machine toggle to a per-election system setting for controlling bubble vs. summary ballots


arsalansufi:
After discussion with @mattroe, we agreed that this should be set on a per-election basis and propagated via the election package rather than set on a per-machine basis. We don't foresee any situation where, for a given election, a jurisdiction will want some BMDs printing summary ballots while others are printing onto bubble ballots.

PR #7223, opened 2025-09-17
VxDesign: Add BMD print mode setting for Vx users

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/6859

Adding a dev/internal system setting for switching between BMD print modes:
- Summary (existing default): Summary ballot with QR-encoded votes.
- Bubble Marks: Vote marks printed on top of preprinted ballot sheets.
- Full Ballot (not implemented on `main` yet): Full ballot printouts with votes marked.

Will be replacing the dev toggle in VxMark soon and reading from this setting instead.
Copy could probably use some tweaking before launching the feature, but also open to changes on the system setting/flag naming.

## Demo Video or Screenshot

Screenshot 2025-09-17 at 14 59 04

Screenshot 2025-09-17 at 16 20 03

## Testing Plan
- Unit tests for edit/save UX

## Checklist

### DIFF
--- a/libs/types/src/system_settings.ts
+++ b/libs/types/src/system_settings.ts
@@ -117,11 +117,15 @@ export interface SystemSettings {
   readonly bmdPrintMode?: BmdPrintMode;
 }
 
-const PRINT_MODES = ['bubble_marks', 'full_ballot', 'summary'] as const;
+const PRINT_MODES = [
+  'bubble_ballot',
+  'marks_on_preprinted_ballot',
+  'summary',
+] as const;
 
 /**
- * - `bubble_marks`: Bubble marks only, printed on preprinted HMPB sheets.
- * - `full_ballot`: Full HMPBs, printed on blank sheets, with votes marked.
+ * - `bubble_ballot`: Full HMPBs, printed on blank sheets, with votes marked.
+ * - `marks_on_preprinted_ballot`: Bubble marks only, on preprinted HMPB sheets.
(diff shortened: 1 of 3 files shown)


libs/types/src/system_settings.ts
```diff
@@ -109,8 +109,23 @@ export interface SystemSettings {
...
   readonly bmdAllowOvervotes?: boolean;
+
+  /**
+   * The BMD print mode for the election.
+   * See {@link BmdPrintMode}.
+   */
+  readonly bmdPrintMode?: BmdPrintMode;
 }
 
+const PRINT_MODES = ['bubble_marks', 'full_ballot', 'summary'] as const;
```
jonahkagan, line 120: In my head, I've thought of these modes as:
- Summary ballot
- Pre-printed bubble ballot
- On-demand bubble ballot

Maybe that helps with naming the modes?

One idea might be:
- `summary_ballot`
- `bubble_ballot`
- `marks_on_preprinted_ballot`

jonahkagan (approved):

libs/types/src/system_settings.ts
kofi-q: Thanks - yeah, I like the idea of having `bubble_ballot` refer to the full print and giving the pre-printed mode more explicit naming.
Was trying "on-demand" copy earlier and initially couldn't mentally get past that the fact that these all feel "on-demand" in some way, but coming around to it

Issue #7212, opened 2025-09-16
Diagnostics date edge case


arsalansufi:
Parent cert discrepancy: https://github.com/votingworks/cert-discrepancies/issues/217

PR #7227, opened 2025-09-18
Address diagnostics retrieval edge case

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7212

A kinda funny find by SLI. If you set the clock to some time in the future, record a diagnostic result, reset the clock back, and record a second diagnostic result, the second diagnostic result doesn't show. You remain stuck with the first diagnostic result captured in the "future."

We currently retrieve the diagnostic result with the latest timestamp, where the timestamp is the system time at the time of record creation. Because our machines are offline and clocks are set manually, this time may not actually correspond to the latest created record according to real world time. As a simple fix, I'm now retrieving the diagnostic record with the highest sequential ID.

### Before

No matter what I click, the diagnostic result doesn't update.

https://github.com/user-attachments/assets/0fe1eb48-c3cb-4f75-bd52-40afba8ec4b1

### After

Now, even if the current result has a future timestamp, the latest interaction according to real world time wins out.

https://github.com/user-attachments/assets/f7baf57e-3558-4bca-861e-dd2306d1bc74

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

## Checklist

- [ ] ~I have added the "user_facing_change" label to this PR, if relevant, to automate an announcement in #machine-product-updates.~ N/A

### DIFF
--- a/libs/backend/src/diagnostics.ts
+++ b/libs/backend/src/diagnostics.ts
@@ -47,6 +47,9 @@ export function getMostRecentDiagnosticRecord(
   client: Client,
   type: DiagnosticType
 ): DiagnosticRecord | undefined {
+  // Retrieve the latest record using the monotonically increasing ID rather than the timestamp to
+  // ensure that we're pulling by real world time rather than system time, which can be toggled
+  // into the future and then back into the past
   const record = client.one(
     `
       select
@@ -56,7 +59,7 @@ export function getMostRecentDiagnosticRecord(
         timestamp
       from diagnostics
       where type = ?
-      order by timestamp desc
+      order by id desc
       limit 1
     `,
     type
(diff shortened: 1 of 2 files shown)


libs/backend/src/diagnostics.ts
```diff
@@ -56,7 +56,7 @@ export function getMostRecentDiagnosticRecord(
         timestamp
       from diagnostics
       where type = ?
-      order by timestamp desc
+      order by id desc
```
arsalansufi, line 59: On all our machines, the diagnostics table ID is a monotonically increasing integer:

https://github.com/votingworks/vxsuite/blob/576efc196103513036c671c3149fc1822d9926e3/apps/scan/backend/schema.sql#L65

nikhilb4a (approved):
Kind of impressive they found this 😅

Issue #7233, opened 2025-09-19
VxDesign: Improve ballot export resiliency


arsalansufi:
Creating this issue to track at a high-level what Jonah has laid out in https://docs.google.com/document/d/1KNvTYZX2S2Nvc4jxL4TVOthKRGi8gWSeRpLs97_HQw0/edit?usp=sharing

PR #7239, opened 2025-09-22
VxDesign: Render ballots concurrently using a pool of browser pages

jonahkagan:
## Overview

Task: #7233 

Originally, we created a separate browser page for each ballot and rendered them all concurrently using `Promise.all`. However, this lead to untenably high memory usage for large sets of ballots. To fix this, we switched to serial rendering, reusing a single browser page, which limited memory usage at the cost of speed.

In this PR, we take a middle path, rendering ballots concurrently using a limited pool of browser pages. From some basic experimentation, I found that 10 browser pages cost 100-200 MB of extra memory usage but sped things up 4x. Adding more browser pages beyond that seemed to yield diminishing returns for the workloads I was testing: 100 and 500 ballots.

## Demo Video or Screenshot
N/A

## Testing Plan
- Manual tested and measured speed + memory usage for a variety of pool sizes from 2-20
- Added some new automated tests to ensure pages are managed correctly
- Relied on existing automated tests to ensure no logical regressions

## Checklist


kofi-q (approved):
✅

PR #7246, opened 2025-09-24
VxDesign: Sync dev auth orgs to the DB cache

kofi-q:
## Overview

Ran into an issue in dev where the API requests fail after doing a DB reset, due to missing dev orgs in the cache.
Adding the startup org sync call for non-auth-enabled dev as well to make sure they get restored after DB resets.

## Testing Plan
- Manual

## Checklist

### DIFF
--- a/apps/design/backend/src/index.ts
+++ b/apps/design/backend/src/index.ts
@@ -47,22 +47,18 @@ async function main(): Promise<number> {
   const { store } = workspace;
 
   const auth0 = authEnabled() ? Auth0Client.init() : Auth0Client.dev();
-  if (authEnabled()) {
-    try {
-      await store.syncOrganizationsCache(await auth0.allOrgs());
-    } catch (error) {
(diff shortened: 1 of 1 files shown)


apps/design/backend/src/index.ts
```diff
@@ -61,6 +61,8 @@ async function main(): Promise<number> {
       }
       throw error;
     }
+  } else {
+    await store.syncOrganizationsCache(await auth0.allOrgs());
```
jonahkagan, line 65: Should we just remove the `if (authEnabled())` guard on L50? I've been through so many iterations with this that I can't remember why I added it in the first place lol

jonahkagan (approved):

apps/design/backend/src/index.ts
kofi-q: I know the feeling 😅 
+1 to just removing the condition - will do

Issue #7247, opened 2025-09-24
VxQr: Pass through precinct of VxScan went set and only include contest data relevant to that precinct


carolinemodic:

mattroe:
https://github.com/votingworks/vxsuite/issues/7283

Issue #7152, opened 2025-09-04
Optimize CompressedTally format


carolinemodic:
Update CompressedTally to encode only the contests in ballot styles in the precinct the machine is configured for. When configured for all precincts encode the whole election definition. Add precinct index to the encoded format. Create one qr code across parties in a primary scenario. Update how we are encoding the CompressedTally in the qr code to either:
- Instead of json.stringiyfing the array, make a string in a more sensible way with a seperator. Don't base64 encode it, add gzipping. 
- Send CompressedTally numbers as a purely numeric encoded string and take advantage of qr code numeric string optimized encoding (will need to update qrcode.react to v4.1) 

Update parsing logic on the other end to manage those changes and still be able to interpret the results

PR #7248, opened 2025-09-24
VxDesign: Implement basic functionality of VxQr reporting endpoint

carolinemodic:
## Overview
This PR implements the basic core functionality for the VxQr reporting endpoint while setting up for future changes. 
- I refactored some of the encoding/decoding helpers a bit better and made the way we are encoding the CompressedTally less obviously inefficient. There is still more optimization work to tackle in future PRs, namely only reporting results for the precinct configured for, and passing through precinct information in general. #7247 #7152 
- I added a helper to the store to get the saved data, this isn't used anywhere other then tests yet but will be used in the next step of adding a UI to view all aggregated reported results
   -  I plan to make this only show data for the most recent exported ballot hash for the given election. If a user is changing/reexporting a ballot package there are ways this could get confusing, but its unclear how to handle those cases other then just stopping working and not showing results there. Longer term we should probably have a more locked "finalized" or "archived" state that the election needs to be in to enable VxQr usage that will prevent changing of the last exported ballot hash. 
- I intend to do a design / UI pass of the reporting UI in the future but wanted to get the relevant information on screen and make it at least usable on mobile where we expect users to access it. 
- As discussed in Slack to start we are just always going to replace data for the same machine/election/is_live_mode with newer data if seen: https://votingworks.slack.com/archives/C09DD114KGQ/p1758582869863789

Open to suggestion on if the results should be saved in compressed/encoded format or as the json blob of tabulation results. I err'd on the side of compressed just since its less space to store and is easy enough to convert when loading. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/bccbf9dd-7b8d-4239-acbc-cf71a90cf85b

## Testing Plan

Testing encoding QR codes and reading those results and that they were parsed and handled properly. Integration tests to test error cases. 

## Checklist


libs/auth/src/signatures.ts
```diff
@@ -34,3 +34,22 @@ export function constructPrefixedMessage( // eslint-disable-line vx/gts-jsdoc
...
+} {
+  const parts = prefixedMessage.split(MESSAGE_SEPARATOR);
+
+  if (parts.length < 3 || parts[0] !== MESSAGE_FORMAT_VERSION.toString()) {
+    throw new Error('Invalid prefixed message format');
+  }
+
+  return {
+    messageType: parts[1] as string,
+    messagePayload: parts[2] as string,
```
arsalansufi, line 53: A bit of a gotcha but it's been hard to guarantee that the message separator doesn't show up in the message payload, so I might make a tweak here to return all parts index 2 and onward concatenated back together just in case there are more than 2 instances of the message separator, e.g., `parts.slice(2, parts.length).join(MESSAGE_SEPARATOR)`

A recent PR where this came up: https://github.com/votingworks/livecheck-web/pull/4

libs/utils/src/compressed_tallies.ts
```diff
@@ -124,19 +144,56 @@ function getContestTalliesForCompressedContest(
...
+    buffer.byteOffset,
+    buffer.byteLength / 2
+  );
+  const compressedTally: CompressedTally = [];
+  let offset = 0;
+  for (const contest of election.contests) {
+    if (contest.type === 'yesno') {
+      compressedTally.push(
+        Array.from(
+          uint16Array.slice(offset, offset + 5)
```
eventualbuddha, line 163: This and the other `slice` results are completely unchecked and may not return a subarray of the expected length:

```ts
> (new Uint16Array(2)).slice(0, 5)
Uint16Array(2) [ 0, 0 ]
```
I believe we should be able to calculate the expected length upfront and reject it if it doesn't match, then these operations in the loop should be safe.

Also, we can make the `5` less of a magic number and force the build to fail if it gets out of sync by using something like this:

```ts
  const yesNoContestCompressedTallyLength: YesNoContestCompressedTally['length'] = 5;
```

We can do something similar for candidate contests, though it'd be slightly more complicated.

apps/design/backend/migrations/1758581797749_add-quick-results-table.js
```diff
@@ -0,0 +1,34 @@
...
+      signed_at: {
+        type: 'timestamptz',
+        notNull: true,
+      },
+      encoded_compressed_tally: { type: 'text', notNull: true },
+      precinct_id: { type: 'text' }, // Null for non-precinct-specific results
+    },
+    {
+      constraints: {
+        primaryKey: ['ballot_hash', 'machine_id', 'is_live_mode'],
```
eventualbuddha, line 29: Do we want this to be per ballot hash or per election ID? Do we allow multiple elections with the same ballot hash?

apps/design/backend/src/app.ts
```diff
@@ -882,37 +885,46 @@ export function buildUnauthenticatedApi({ logger }: AppContext) {
...
+        const {
+          ballotHash,
+          machineId,
+          isLive,
+          signedTimestamp,
+          encodedCompressedTally,
+        } = decodeQuickResultsMessage(payload);
+
+        const electionRecord =
+          await store.getElectionFromBallotHash(ballotHash);
```
eventualbuddha, line 899: So if I (accidentally or intentionally) create an election in VxDesign with the same election, even if it belongs to a different organization, there's a chance that it's _my_ election that will get incoming results instead of the intended one. Given the assumption that the production VxDesign will be locked down sufficiently and not allow accounts with the general public, it seems likely that this would only happen by accident. However, I could easily see that happening if someone created a duplicate as a backup or something.

apps/design/backend/src/store.ts
```diff
@@ -1931,4 +1938,99 @@ export class Store {
...
+      client.withTransaction(async () => {
+        const { rowCount } = await client.query(
+          `
+            insert into results_reports (
+              ballot_hash,
+              election_id,
+              machine_id,
+              is_live_mode,
+              signed_at,
+              encoded_compressed_tally,
```
eventualbuddha, line 2014: If we're going to be saving the encoded compressed tally (or even if not), does it make sense to include a version number or something in case the schema changes in the future? Or will we be migrating the data on schema changes?

apps/design/backend/src/store.ts
```diff
@@ -1931,4 +1938,99 @@ export class Store {
...
+              precinct_id
+            ) values ($1, $2, $3, $4, $5, $6, $7)
+            on conflict (ballot_hash, machine_id, is_live_mode)
+            do update set
```
eventualbuddha, line 2018: Noting that this does not enforce that the record being updated has the same `election_id` as the record we intended to create.

eventualbuddha (approved):
Overall progress looks good, approving since none of my concerns are blockers for an in-progress feature.

libs/utils/src/compressed_tallies.ts
```diff
@@ -4,55 +4,74 @@ import {
...
 } from '@votingworks/types';
+import { Buffer } from 'node:buffer';
 import { assert, throwIllegalValue, typedAs } from '@votingworks/basics';
+import { ContestResults } from '@votingworks/types/src/tabulation';
+
+export function encodeCompressedTally(
+  compressedTally: CompressedTally
+): string {
+  const flatArray = compressedTally.flat();
+  const uint16Array = new Uint16Array(flatArray);
```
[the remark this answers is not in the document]
carolinemodic: uint16 seems way more then enough space to hold any tally object but I should maybe do a quick safety/check assert of that fact before compressing

libs/utils/src/compressed_tallies.ts
```diff
@@ -124,19 +144,56 @@ function getContestTalliesForCompressedContest(
...
+): CompressedTally {
+  const buffer = Buffer.from(encodedCompressedTally, 'base64url');
+  const uint16Array = new Uint16Array(
+    buffer.buffer,
+    buffer.byteOffset,
+    buffer.byteLength / 2
+  );
+  const compressedTally: CompressedTally = [];
+  let offset = 0;
+  for (const contest of election.contests) {
```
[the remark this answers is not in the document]
carolinemodic: yeah

apps/design/backend/migrations/1758581797749_add-quick-results-table.js
carolinemodic: I went with ballot hash since thats how we will look up the data. My understanding that I checked w/ Jonah is that whenever an election is cloned/duplicated a new ID will generate, since the election id is part of what is in the election json blob that gets hashed that should guarantee unique ballot hashes across elections. And it should guarantee that if two results have the same ballot_hash they also have the same election_id.

apps/design/backend/src/app.ts
carolinemodic: See other comment this will gen a new id and create a new ballot hash

libs/auth/src/signatures.ts
```diff
@@ -50,6 +50,6 @@ export function deconstructPrefixedMessage(prefixedMessage: string): {
 
   return {
     messageType: parts[1] as string,
-    messagePayload: parts[2] as string,
+    messagePayload: parts.slice(2, parts.length).join(MESSAGE_SEPARATOR),
```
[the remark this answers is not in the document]
carolinemodic: Yeah thats fair, this particular encoding step is also used by signed hash validation and I think it makes sense for the format to be the same between shv/vxqr. I'll make a followup task to reasses this but not change for this PR. https://github.com/votingworks/vxsuite/issues/7270

Issue #4980, opened 2024-06-20
Unify underlying BMD/HMPB interpretation infrastructure


eventualbuddha:

arsalansufi:
Noting that we should keep edge cases like https://github.com/votingworks/vxsuite/pull/5602 in mind when we work on this. May be inherently resolved by image cropping

adghayes:
@arsalansufi noticing the status move here - are you thinking that this is something to get into v4.1? IIRC, part of why we didn't fully enable this was to not trigger more testing in the v4.0 campaign. Going into v4.1 is the time, if any.

arsalansufi:
While I would still like to land this eventually, it's dropped in importance as we move away from summary ballots (and even when we do have summary ballots, we scan so few). Also being realistic, we won't be able to land this in time for v4.1.0. But v4.1.1+ maybe. If paired with other interpretation changes, then not a big deal to land this in a v4.1.1+ as we'll be redoing an accuracy test anyway. And as I noted in a different thread, accuracy testing has become less involved over time.

PR #7259, opened 2025-09-25
feat: improve Rust BMD/HMP ballot data encoding support

eventualbuddha:
## Overview

Refs #4980 

Brings full encoding and decoding support for both BMD and HMP ballot QR codes as used in VxSuite to Rust. Rather than having a separate `ballot-encoder-rs` crate, this is implemented as `bitstream-io` traits on the types where they're defined in `types-rs`.

The next steps will be to add some sort of test verifying that the Rust and TypeScript versions are interchangeable and to add the feature flag that controls which version is used in the ballot interpretation pipeline.

## Demo Video or Screenshot
```rust
use types_rs::{coding, hmpb};

let metadata = hmpb::Metadata {
    ballot_hash,
    precinct_id: PrecinctId::from("precinct-1".to_owned()),
    ballot_style_id: BallotStyleId::from("ballot-style-1".to_owned()),
    page_number: PageNumber::new_unchecked(1),
    is_test_mode: true,
    ballot_type: BallotType::Precinct,
    ballot_audit_id: None,
};
let qr_code_data = coding::encode_with(&metadata, &election).unwrap();
```

## Testing Plan
Added doctests, unit tests, and some integration tests. Several of the unit and integration tests use property testing to improve coverage of the input space.

### DIFF
--- a/libs/types-rs/src/hmpb.rs
+++ b/libs/types-rs/src/hmpb.rs
@@ -1,4 +1,4 @@
-use bitstream_io::{FromBitStreamWith, ToBitStream, ToBitStreamWith};
+use bitstream_io::{FromBitStreamWith, ToBitStreamWith};
 use serde::{Deserialize, Serialize};
 
 use crate::{
@@ -192,16 +192,16 @@ impl ToBitStreamWith<'_> for Metadata {
         let precinct_index = election
             .precinct_index(&self.precinct_id)
             .ok_or_else(|| Error::InvalidPrecinctId(self.precinct_id.clone()))?;
-        precinct_index.to_writer(w)?;
+        w.build(&precinct_index)?;
 
         let ballot_style_index = election
             .ballot_style_index(&self.ballot_style_id)
             .ok_or_else(|| Error::InvalidBallotStyleId(self.ballot_style_id.clone()))?;
-        ballot_style_index.to_writer(w)?;
+        w.build(&ballot_style_index)?;
 
-        self.page_number.to_writer(w)?;
+        w.build(&self.page_number)?;
         w.write_bit(self.is_test_mode)?;
-        self.ballot_type.to_writer(w)?;
+        w.build(&self.ballot_type)?;
 
         match self.ballot_audit_id {
             Some(ref ballot_audit_id) => {
@@ -213,7 +213,7 @@ impl ToBitStreamWith<'_> for Metadata {
                 else {
                     return Err(Error::InvalidBallotAuditId(ballot_audit_id.clone()));
                 };
-                ballot_audit_id_length.to_writer(w)?;
+                w.build(&ballot_audit_id_length)?;
                 w.write_bytes(ballot_audit_id.as_bytes())?;
             }
 
@@ -618,8 +618,8 @@ mod test {
         #[test]
         fn test_ballot_audit_id_coding(ballot_audit_id in "[0-9a-z-]{1,100}") {
             let ballot_audit_id_length = BallotAuditIdLength::new(ballot_audit_id.len() as u8).unwrap();
-            let bytes = collect_writes::<coding::Error>(|mut writer| {
-                ballot_audit_id_length.to_writer(&mut writer)?;
+            let bytes = collect_writes::<coding::Error>(|writer| {
+                writer.build(&ballot_audit_id_length)?;
                 Ok(writer.write_bytes(ballot_audit_id.as_bytes())?)
             }).unwrap();
 
--- a/libs/types-rs/src/coding.rs
+++ b/libs/types-rs/src/coding.rs
@@ -164,7 +164,7 @@ where
     T: ToBitStream,
     T::Error: From<io::Error>,
 {
-    collect_writes(|writer| value.to_writer(writer))
+    collect_writes(|writer| writer.build(value))
 }
 
 /// Encode a codable value, with a context, to an array of bytes using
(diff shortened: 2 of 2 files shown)


libs/types-rs/src/bmd/write_in_name.rs
```diff
@@ -0,0 +1,116 @@
...
+        // write length
+        w.write_unsigned_var(Self::BITS, self.0.len() as u32)?;
+
+        // write character indexes
+        for ch in self.0.chars() {
+            let (index, _) = Self::CHARS
+                .char_indices()
+                .find(|(_, c)| *c == ch)
+                .expect("character must be in CHARS");
+            w.write_unsigned_var(Self::BITS, index as u32)?;
```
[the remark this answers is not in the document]
eventualbuddha: I agree this would be a useful performance improvement, but we're only dealing with up to 40 characters per write-in and encoding/decoding write-ins is such a small part of the overall flow that I don't think it's worth it.

libs/types-rs/src/hmpb.rs
```diff
@@ -118,14 +157,65 @@ impl FromBitStreamWith<'_> for Metadata {
...
-const BALLOT_HASH_LENGTH: u32 = 20;
-const HEX_BYTES_PER_CHAR: u32 = 2;
+impl ToBitStreamWith<'_> for Metadata {
+    type Context = Election;
+    type Error = Error;
+
+    fn to_writer<W: bitstream_io::BitWrite + ?Sized>(
+        &self,
+        w: &mut W,
+        context: &Self::Context,
```
jonahkagan, line 184: Maybe s/context/election/ for clarity and also to match the BMD encoding function

libs/types-rs/src/bmd/completed_ballot.rs
```diff
@@ -0,0 +1,186 @@
...
+
+            let contest_vote = match contest_votes[..] {
+                [] => continue,
+                [_, _, ..] => {
+                    return Err(Error::InvalidVotes {
+                        message: "Too many ContestVote records for contest".to_owned(),
+                    })
+                }
+                [contest_vote] => contest_vote,
+            };
```
jonahkagan: Any reasoning behind representing `votes` as a vector rather than a mapping (like we do in the corresponding TS code)? This guard in particular would be unnecessary if there was a structural constraint that each contest could only have one vote record

libs/types-rs/src/bmd/completed_ballot.rs
```diff
@@ -0,0 +1,186 @@
...
+    where
+        Self: Sized,
+    {
+        w.write_bytes(PRELUDE)?;
+        w.write_bytes(&self.ballot_hash)?;
+
+        let precinct_index = election
+            .precinct_index(&self.precinct_id)
+            .ok_or_else(|| Error::InvalidPrecinctId(self.precinct_id.clone()))?;
+        w.build(&precinct_index)?;
```
jonahkagan: What's the difference between `w.build(&precinct_index)` and `precinct_index.to_writer(w)` (which is used in the HMPB encoding function)?

jonahkagan (approved):
Looks great overall! Thanks for writing thorough tests and docs

libs/types-rs/src/bmd/completed_ballot.rs
eventualbuddha: Functionally, these are identical. In fact, if you look at the definition for `BitWriter::build` it just calls `to_writer`:

```rust
    /// Builds and writes complex type
    fn build<T: ToBitStream>(&mut self, build: &T) -> Result<(), T::Error> {
        build.to_writer(self)
    }
```

The main difference is then just a matter of readability and usability. Since `to_writer` is a trait method from `ToBitStream`, you have to have `ToBitStream` in scope to call `to_writer` on a struct that implements it. This isn't a big deal, but is a little bit noisy and mildly confusing (I remember struggling with that when I first learned Rust).

On the other hand, calling `BitWriter::build` means you can avoid the `ToBitStream` import and keeps the thing doing the action, the `BitWriter` in this case, as the central character, so to speak. The same distinction exists with `BitReader::parse` and `FromBitStream::from_reader`. This requires understanding how traits work too, so conceptually it's not really any easier. But I think this form makes it clearer what's happening and is more idiomatic Rust, so I'll switch everything over to use this style.

libs/types-rs/src/election.rs
```diff
@@ -31,6 +31,34 @@ pub struct Election {
...
+    /// Finds the index of the precinct with the given ID, returning `None` if
+    /// no such precinct is found or if the index cannot be represented in the
+    /// number of bits allowed for a [`PrecinctIndex`].
+    #[must_use]
+    pub fn precinct_index(&self, precinct_id: &PrecinctId) -> Option<PrecinctIndex> {
+        self.precincts
+            .iter()
+            .enumerate()
+            .find(|(_, precinct)| &precinct.id == precinct_id)
+            .and_then(|(index, _)| u32::try_from(index).ok())
```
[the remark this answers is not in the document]
eventualbuddha: 066f575e

libs/types-rs/src/hmpb.rs
eventualbuddha: dab346e3

libs/types-rs/src/bmd/completed_ballot.rs
eventualbuddha: d36c8e5d

libs/types-rs/src/bmd/completed_ballot.rs
eventualbuddha: d2a05ee7

PR #7278, opened 2025-09-30
fix(scan): ensure message and newline are written together

eventualbuddha:
## Overview
Without locking for both writes it may be that some other writes happen simultaneously.

## Demo Video or Screenshot
For example, stdout could look like this:
```
{"responsehello world!
":"hello again!
ok"}final hello!


```
This might happen if the writes to stdout made by `serde_json::to_writer` were being made while another thread was writing to stdout.

## Testing Plan
Manually tested to ensure that `pdictl` still works as expected.


jonahkagan (approved):
Good catch!

PR #7317, opened 2025-10-06
[DRAFT] VxDesign: Delete TTS edits when related entities are modified

kofi-q:
Want to get early feedback on the direction here before diving into tests: this is an outline of upcoming changes to keep the `tts_strings` table in sync with modifications made to the original election strings.

Easier to review with whitespace ignored.

General Idea:
- [Not shown in this draft] Seed the UI with the original election strings when there's no existing `tts_strings` entry - add a `tts_strings` entry when an edit is saved. 
- When an entity is deleted, delete all the associated `tts_strings` entries.
- When an entity is updated, diff relevant fields with the original and, for any that have changed, delete associated `tts_strings`.
- This is trickier with `contests` and `precincts`, since we currently delete them wholesale and re-insert. To avoid too much change in the short term, handling those updates client-side for now (sample updates to the contest forms shown here):
  - In the `onChange` hooks, we track which strings have been changed and in `onSubmit`, we send a request to delete the associated `tts_strings` - no diffing here, so just adding and deleting a character will invalidate a TTS edit, but we can refine later.
- Still some product questions to iron out - e.g. thinking it could be useful to have a way of indicating to the user that their edits have reverted/will revert audio edits, but might be rare enough in the upcoming first pass that people are going back and forth between audio edits and election info edits.

Open to other approaches - briefly considered having the TTS edits in each respective entity table, but that has it's own set of complexities and still has the issue of needing to manage the consistency between base string edits and TTS edits.


kofi-q (commented):
Felt good to delete a bunch of code - thanks for your input!
Only potential issue I could think of would be any ballot measure descriptions that get large enough to go past the index size limit, but that's hopefully unlikely with the HTML stripping we have in place for the TTS cache and can reuse here

jonahkagan:
Nice, the new schema looks good to me! Agreed that we're probably safe to punt on the ballot measure description issue for now, since it's the same problem we already have.

PR #7322, opened 2025-10-07
VxMark: support printing longer ballot sizes

adghayes:
## Overview

Currently we can create ballots from VxMark that print marks onto pre-printed ballots. It's been limited to letter and legal size ballots because the typical LaserJet Pro 4001dn supports maximum 14" paper. With a custom tray and custom firmware provided by HP, however, plus some configuration changes, we can print on up to 22" paper. This PR allows those sizes. 

I'm not sure if we can detect the firmware version systematically, which would be nice to only allow this option when it's availabile. The key change is actually not the "engine firmware" but the "formatter firmware" which I think would be even more unlikely to detect. I think on those printers, the job just fails (as it now would crash the app). 

## Demo Video or Screenshot

IMG_3272

## Testing Plan

Tested manually on 22", 19", and 17".

## Checklist

### DIFF
--- a/apps/mark/backend/src/util/print_ballot.test.tsx
+++ b/apps/mark/backend/src/util/print_ballot.test.tsx
@@ -91,33 +91,6 @@ describe(`printMode === "bubble_marks"`, () => {
       ],
     ]);
   });
-
-  const testInvalidSizes = test.each<HmpbBallotPaperSize>([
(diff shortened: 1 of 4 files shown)


adghayes:
Yes, it turns out that you can have paper of any size if you have enough paper and tape!

PR #7325, opened 2025-10-07
VxQr: Design pass on VxScan components including addition to report polls open status

carolinemodic:
## Overview
Takes a pass at the VxScan exposed screens or VxQR based on design feedback with @mattroe and @jonahkagan . We are also going to add a new feature to report and view polls open status so this PR does the VxScan - related components of that to surface the QR code after opening polls and to be able to create a QR code that just has "polls_open" rather than the compressed tally. We could send something more compact than a string but compared to the compressed tally the polls state is trivial in size. 

Scanning the QR code for polls open reporting is not yet supported and will be handled in a future PR. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/c4c8877a-5f4b-4fe8-875c-cbb0ee28441e

## Testing Plan
Tested polls open and polls close flows to view QR code after printing. Tested viewing QR code from poll worker menu. 

## Checklist

### DIFF
--- a/apps/scan/backend/src/app.ts
+++ b/apps/scan/backend/src/app.ts
@@ -7,6 +7,7 @@ import {
   SinglePrecinctSelection,
   DiagnosticRecord,
   DiagnosticOutcome,
+  doesPollsStateSupportLiveReporting,
 } from '@votingworks/types';
 import {
   combineElectionResults,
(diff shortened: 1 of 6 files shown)


jonahkagan (approved):
Design lgtm, didn't review code

PR #7329, opened 2025-10-07
chore: remove unused election property

eventualbuddha:
We currently don't use this in the Rust code, so we'll just remove it for now.


kshen0 (approved):
LGTM assuming `types-rs` fix is easy

PR #7364, opened 2025-10-14
VxDesign: Revamp VxQr frontend for general polls reporting

carolinemodic:
## Overview
Incorporates polls open reporting into the "results" vxqr frontend by rebranding as "live reports. The main page will be a summary of precinct report information with the ability to then view results on a per-precinct or all-precincts basis, when polls closed data is available. 

If there is data for a VxScan configured as "All Precincts" a row "Precinct Not Specified" will be added to the end of the precincts table, otherwise it is omitted. When new data appears the updated row in the precinct table will momentarily highlight. This PR does NOT yet incorporate a success chime, that is planned as a followup. 

The frontend now will show test data whenever there is no live data. As long as live data exists the frontend will switch to "live" mode and only show live data. The clear data (intended only for demos) will clear both live and test data. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/c06f1ef2-5eb2-4f03-a481-b75ea203905e

## Testing Plan
See video for main flow of testing. Added tests for major flows.. 

## Checklist

### DIFF
--- a/apps/design/backend/src/app.results.test.ts
+++ b/apps/design/backend/src/app.results.test.ts
@@ -32,7 +32,7 @@ import {
   testSetupHelpers,
   unzipElectionPackageAndBallots,
 } from '../test/helpers';
-import { Org, User } from './types';
+import { ALL_PRECINCTS_REPORT_KEY, Org, User } from './types';
 import { Workspace } from './workspace';
 
 const mockFeatureFlagger = getFeatureFlagMock();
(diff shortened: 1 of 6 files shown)


apps/design/backend/src/app.results.test.ts
```diff
@@ -631,26 +667,98 @@ test('quick results reporting works for polls open reporting', async () => {
...
+          {
+            machineId: 'mock-02',
+            pollsState: 'polls_open',
+            precinctSelection: singlePrecinctSelectionFor(
+              sampleElectionDefinition.election.precincts[0].id
+            ),
+            signedTimestamp: new Date('2024-05-04T09:00:00Z'),
+          },
+        ],
+        '': [
```
eventualbuddha: I don't really love having sentinel values like this that don't really explain themselves. Perhaps make a constant?

```ts
export const ALL_PRECINCTS_REPORT_KEY = '';
```

Issue #7167, opened 2025-09-08
VxAdmin: Fix overvote write-in tally adjustment for filtered reports


adghayes:
Parent Issue: https://github.com/votingworks/cert-discrepancies/issues/226

VxAdmin has a report option for "Single Precinct Tally Report" which provides a dropdown to select a precinct. It appears that the dropdown is showing all precincts and precinct splits, rather than just precincts. When a precinct split is selected, it causes a crash. We likely need to change the dropdown to only show full precincts rather than precinct splits.

nikhilb4a:
Tested with that election package and haven't been able to recreate the issue yet.

Steps:
1. Loaded election package into Admin
2. Scanned a couple ballots from different precincts in Scan (official mode)
3. Loaded CVRs into Admin, viewed tally report by precinct for split and non-split precinct

Note: as seen in the screen recording, the tally report isn't actually viewable in the UI. @adghayes is this expected/known behavior? When I export the report, it is viewable (shows in screen recording)

https://github.com/user-attachments/assets/c714c052-8fc7-4e44-8787-320ca4ce31c7

I also noticed that in this election, `election.precincts` includes individual entries for split precincts. When I created a new election with a precinct split in VxDesign, there was a single entry for the precinct in `election.precincts`, with split info stored in `precinct.split`. Any idea if we've changed the way we store `precinct splits` since this election package was created?

nikhilb4a:
Also, noting the above recording is on `v4.0.2-rc4`

adghayes:
The un-viewable report is definitely not expected / known behavior - is that specific to the single precinct report? If it's applicable to all reports, I'm guessing something is broken on `main` or there's some issue in your environment as it relates to the recent upgrades to PDF libraries.

nikhilb4a:
It's for all reports, so I think you're right it is probably an environment specific issue. The PDFs look good when exported.

Also, another possibility is that this election package doesn't actually have a split. When I re-upload it to VxDesign 4.0.2 it loads an election with separate precincts for 2a and 2b. So the name of those may imply a split when really they are distinct districts. Or if we are sure it did have splits, then it is definitely of a different format and the likely cause of the issue.

I created a new election (still VxDesign 4.0.2), added a split, scanned a few ballots, and created the below video. The single precinct tally report still works, and in this case the tally report for this Precinct includes tally info from Split District and Split School. So, still not able to reproduce.

[...]

Issue #6014, opened 2025-02-12
VxAdmin: Fix underlying write-in/overvote bug


nikhilb4a:
In https://github.com/votingworks/vxsuite/issues/5656 we introduced a temporary bug fix for write-in/overvotes cases that were not tabulating in tally reports properly. 

We should fix the underlying issue related to using SQL command to pre-process write-in tallies.

PR #7372, opened 2025-10-14
VxAdmin: Update write-in tabulation to be sequential rather than via aggregating SQL query

nikhilb4a:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/6014
Closes https://github.com/votingworks/vxsuite/issues/7167

This PR updates the logic for write-in tabulation, as it has caused a few bugs over time.

Previously we used an approach of aggregating tallies via a SQL query, but this had issues in that it could not take into account whether a write-in was part of an overvote, and thus should be omitted from tallies.

The new approach gets all write-ins, and then streams through them, reviewing their details to determine it's tabulation value. It extends the concept from `includeUnadjudicatedAndInvalidUnmarkedWriteIns`, used in the previous `tabulateWriteInTallies` function to now include overvoted write-ins, renaming the field `includeUnallocablePendingWriteInsAsPending`. The idea is that for the Write-in Adjudication Report, we want to include pending unmarked write-ins and pending write-ins part of overvote in the Pending Tally, to indicate they are still awaiting adjudication. Conversely, we do not want to include them in the Tally Report, as they are strictly not valid tallies at that point.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/06a59acc-07de-4fd3-b4dd-2af290ea025b

Example that shows a myriad of write-in combinations adding up to the correct tallies. It might take a minute to understand if you are new to this. 
Key points:
- Pending UWI contributes to WIA report pending count, not the tally report count
- Pending WI part of overvote contributes to WIA report pending count, not the tally report count
- Adjudicated WI part of overvote contributes to WIA report invalid count, no effect on tally report (as we don't show invalid count)

## Testing Plan

1. Updated automated tests
2. Manual testing by Drew and myself

Tested
- filtered tally report (absentee vs precinct)
- UWI, overvotes, pending + combinations of each

## Checklist

### DIFF
--- a/apps/admin/backend/src/store.ts
+++ b/apps/admin/backend/src/store.ts
@@ -1847,14 +1847,6 @@ export class Store {
   /**
    * Gets write-ins specifically for tabulation, filtered and
    * grouped by cast vote record attributes.
-   *
-   * When using write-in tallies to hydrate tally results, unmarked write-ins
-   * that haven't been adjudicated or are invalid should not be included
-   * because they are not associated with an original or adjudicated mark. Whe
-   * using write-in tallies as a summary of write-in adjudication, they should
-   * be included because they exist in the adjudication flow just as marked
-   * write-ins do. The behavior can be controlled with the
-   * `includeUnadjudicatedAndInvalidUnmarkedWriteIns` flag.
    */
   *getWriteInsForTallies({
     election,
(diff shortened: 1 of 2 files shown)


eventualbuddha (approved):
I don't see any red flags here in terms of the code or likely performance issues as a result of this change. The logic seems fine, though I'm relying on @adghayes's validation more than my own here.

adghayes:
@eventualbuddha I did a pretty close read of the logic and feel good about that, consider it approved on my part. 

Although @arsalansufi this is definitely something we should QA and give more attention to as part of the v4.0.4 release.

arsalansufi:
Added a note to https://github.com/votingworks/vxsuite/issues/7397 re QA!

Issue #7266, opened 2025-09-26
VxDesign: In-Context Audio Proofing Screens


kofi-q:
## Context
Add audio preview/edit buttons to all relevant fields in the existing edit screens, which open up a sub-view for managing the corresponding audio. Applies to the following user-provided strings:
- Election title
- State name
- County name
- Precinct names
- Precinct split names
- District names
- Contest titles, descriptions, and option labels
- Candidate names
- Party names - both short and long names
- Ballot measure option labels

## Prototype/Mock
Available in [Figma](https://www.figma.com/design/g5S6rv7kYH8kfEomZTzZqr/In-Context-Audio-Editing?node-id=1-19&t=BMDIci1Ju11oDgLO-0)

## Tasks
- [x] Shared components for audio link buttons and editor
- [x] Batch editing + audio editing for:
  - [x] Election Info screen
  - [x] Districts screen
  - [x] Precincts screen
  - [x] Parties screen
  - [x] Contests screen
- [x] Election string preview component
  - Out of scope: language switching

PR #7378, opened 2025-10-15
VxDesign: Add client-side audio proofing APIs

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/7266

Adding corresponding client-side APIs in preparation for the audio proofing/editing UX.

## Testing Plan
- None yet - will be testing via the UI components

## Checklist

### DIFF
--- a/apps/design/frontend/src/api.ts
+++ b/apps/design/frontend/src/api.ts
@@ -273,8 +273,8 @@ export const ttsEditsGet = {
     const apiClient = useApiClient();
     return useQuery(
       this.queryKey(params),
-      () => (params.original ? apiClient.ttsEditsGet(params) : null),
-      { ...opts }
+      () => apiClient.ttsEditsGet(params),
+      opts
     );
   },
 } as const;


apps/design/frontend/src/api.ts
```diff
@@ -273,17 +264,75 @@ export const getBallotTemplate = {
...
+/* istanbul ignore next - WIP @preserve */
+export const ttsEditsGet = {
+  queryKey(params: TtsEditKey): QueryKey {
+    return ['ttsEditsGet', params.orgId, params.languageCode, params.original];
+  },
+  useQuery(params: TtsEditKey, opts: { enabled?: boolean } = {}) {
+    const apiClient = useApiClient();
+    return useQuery(
+      this.queryKey(params),
+      () => (params.original ? apiClient.ttsEditsGet(params) : null),
```
jonahkagan: When would `params.original` not be truthy? Just curious - it's sometimes hard to understand these chunks without their usage context

apps/design/frontend/src/api.ts
```diff
@@ -273,17 +264,75 @@ export const getBallotTemplate = {
...
+      { ...opts }
```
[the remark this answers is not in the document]
kofi-q: Thanks for catching! Leftover from tinkering with other options (was trying out the suspense feature in the prototype)

apps/design/frontend/src/api.ts
kofi-q: Also leftover from the demo code (where everything was just getting fetched at the top level and a key may or may not have been selected yet) - somehow missed this query while cleaning up

> it's sometimes hard to understand these chunks without their usage context

LMK if there's anything else I can clarify around this or the upcoming feature in general

PR #7379, opened 2025-10-15
VxDesign: VxQR: Add basic fallback pagination when a QR Code overflows

carolinemodic:
## Overview

See https://votingworks.slack.com/archives/C09DD114KGQ/p1760466734563769 for discussion on switching back to medium error encoding. 

When we see we will overflow a medium encoded qr code (just over 2300 bytes) we instead try to break into multiple sub-sections, failing after 25 attempts. The qr codes will be iterated through in VxScan and VxDesign will not show you your results or save them for the aggregated results view unless all have been scanned. 

We do not expect this to come up in practical use cases but are adding for resiliency. 

## Demo Video or Screenshot

VxScan (tested by artificially lowered the max bytes) 

https://github.com/user-attachments/assets/c3384627-5edb-4639-aaf4-07547407fbe1

VxDesign confirmation view on a partial report 
Screenshot 2025-10-14 at 4 09 39 PM

## Testing Plan
Lowered byte size to just under my current elections encoded size, tested flow end to end. Added tests 

## Checklist

### DIFF
--- a/libs/auth/src/signed_quick_results_reporting.ts
+++ b/libs/auth/src/signed_quick_results_reporting.ts
@@ -332,11 +332,12 @@ export async function generateSignedQuickResultsReportingUrl(
         quickResultsReportingUrl,
         maxQrCodeLength
       );
+      /* istanbul ignore next - @preserve */
       if (!url) break;
       return [url];
     }
+    /* istanbul ignore next - @preserve */
     default: {
-      /* istanbul ignore next - @preserve */
       throwIllegalValue(pollsState);
     }
   }


jonahkagan (commented):
🧑‍🎨 Design approved (modulo copy changes)

libs/utils/src/tabulation/compressed_tallies.ts
```diff
@@ -98,13 +113,15 @@ export function compressAndEncodeTally({
   election,
   results,
   precinctSelection,
+  numParts: numSections,
```
eventualbuddha: Do you want these names to be distinct?

libs/auth/src/signed_quick_results_reporting.ts
```diff
@@ -185,57 +214,89 @@ export async function generateSignedQuickResultsReportingUrl(
...
+  ) {
+    const compressedTallies =
+      pollsState === 'polls_closed_final'
+        ? compressAndEncodeTally({
+            election,
+            results,
+            precinctSelection,
+            numParts: numPartsNeeded,
+          })
+        : ['']; // For polls open reports we don't have results to send.
```
eventualbuddha: This has a fair amount of special casing for the `pollsState` both here and in `encodeQuickResultsMessage`. Both functions need to know that a `compressedTally` that is an empty string gets treated specially. While there are presumably tests that enforce this, it indicates to me that this needs to be decomposed so that we can treat the two cases differently using an `if` or `switch` where the containing `while` is in this function. For example, perhaps `encodeQuickResultsMessage` ought to be unaware of exactly what the message is, so instead of taking a `compressedTally` and `pollsState` it simply takes a `message`. Similarly, this whole loop isn't a loop in the `pollsState !== 'polls_closed_final'` state, which is a bit confusing while reading it.

apps/design/backend/src/app.ts
```diff
@@ -989,6 +991,98 @@ export function buildUnauthenticatedApi({ logger, workspace }: AppContext) {
...
+
+          const partials = await store.fetchPartialPages({
+            ballotHash,
+            machineId,
+            isLive,
+            pollsState,
+          });
+
+          // If we don't yet have all pages, return a minimal OK response.
+          if (partials.length < numPages) {
```
eventualbuddha: What if `numPages > partials.length`?

apps/design/backend/src/store.ts
```diff
@@ -2314,17 +2314,161 @@ export class Store {
...
+      )
+    );
+  }
+
   async deleteQuickReportingResultsForElection(
     electionId: ElectionId
   ): Promise<void> {
-    await this.db.withClient((client) =>
-      client.query(
+    await this.db.withClient(async (client) => {
```
eventualbuddha: These two queries should be in a transaction, right? Or maybe you can even combine the queries with a `;` separator.

PR #7386, opened 2025-10-15
ui: Prevent button contents from wrapping (desktop theme only)

jonahkagan:
## Overview

Disables text and icon wrapping for the `Button` component to prevent weird aesthetics when buttons end up in horizontally constrained settings (e.g. when put next to a long string or in a shrinking table column).

I only modified this style for the desktop theme because I wasn't sure if there were situations where we need wrapping in the precinct system (e.g. when using extra large text). Since font sizes don't change dynamically in the desktop theme, it seemed safe.

That means this change will apply to VxDesign, VxAdmin, and VxCentralScan.

After making this change, I also realized this could cause cards to overflow horizontally, so I fixed that as well.

## Demo Video or Screenshot

Before

https://github.com/user-attachments/assets/1ddce1e7-de97-467b-b777-0f23cd335667

After

https://github.com/user-attachments/assets/ed659f5d-5518-4721-81f7-02a0bcf23779

## Testing Plan
Manual test

## Checklist

### DIFF
--- a/apps/design/backend/src/app.test.ts
+++ b/apps/design/backend/src/app.test.ts
@@ -2991,7 +2991,7 @@ test('Export test decks', async () => {
         total: testDecksTask.task!.progress!.progress,
       });
     },
-    { interval: 500, retries: 10 }
+    { interval: 1000, retries: 10 }
   );
 
   await suppressingConsoleOutput(async () => {


kofi-q (approved):
🎉

Issue #7327, opened 2025-10-07
VxScan alarm work


arsalansufi:
Note that all alarms should only play when polls are open

- [x] Make VxScan USB disconnect alert sound continuous like VxMarkScan alarm
- [ ] Add alarm for removals from second port as well, where accessible interface will be connected (frontend done, system check pending)
- [x] Add alarm for PDI scanner cover opening (might want this one behind a system setting to avoid disruptions during expected cleanings)
- [x] Make all new alarms dismissable via poll worker card insertion
- [ ] Update docs to instruct poll workers to insert card before actions that would set off alarm

PR #7396, opened 2025-10-16
Add system setting for disabling VxScan alarms

arsalansufi:
## Overview

Step 1 for https://github.com/votingworks/vxsuite/issues/7327

We're updating VxScan alarms for the cert campaign and are adding a system setting to make it possible to disable them. The system setting:

> Disables the VxScan alarms triggered when USB drives are removed or the scanner cover is
> opened while polls are open. These alarms can be silenced by inserting a smart card, but
> forgetting to do so before performing a routine action like scanner cleaning could cause
> unnecessary worry in a polling place. We're accordingly giving election officials the option
> to disable these alarms completely.

The updated alarm logic itself will come in a subsequent PR.

I've opted for a system setting that disables rather than enables so that, when omitted, alarms are on. This will ensure that alarms turn on for SLI with their existing election definitions.

## Demo Video or Screenshot

| Before | After |
| :- | :- |
| before | after |

I made a few other small tweaks to the system settings page while I was at it, beyond the new toggle for the new system setting:
* Replaced a reference to "BMD ballot" with "summary ballot" for clarity
* Removed some headings that no longer felt necessary
* Allowed for wrapping of cards when screen width is low

Here's a video of the wrapping:

https://github.com/user-attachments/assets/7ea6aff1-c616-4927-8bcd-c5c12c360baf

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

## Checklist

### DIFF
--- a/apps/design/backend/src/features.ts
+++ b/apps/design/backend/src/features.ts
@@ -115,40 +115,31 @@ export enum ElectionFeature {
 export type UserFeaturesConfig = Partial<Record<UserFeature, boolean>>;
 export type ElectionFeaturesConfig = Partial<Record<ElectionFeature, boolean>>;
 
+const vxUserFeaturesConfig: UserFeaturesConfig = {
+  EXPORT_SCREEN: true,
+  CHOOSE_BALLOT_TEMPLATE: true,
+  EXPORT_TEST_DECKS: true,
+
+  SYSTEM_SETTINGS_SCREEN: true,
+  VXSCAN_BMD_BALLOT_SCANNING_SYSTEM_SETTING: true,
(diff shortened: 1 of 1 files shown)


apps/design/backend/src/features.ts
```diff
@@ -12,53 +12,70 @@ export function isVxOrSliOrg(orgId: string): boolean {
  * officials can do.
  */
 export enum UserFeature {
```
arsalansufi: No functional change in this commit (https://github.com/votingworks/vxsuite/pull/7396/commits/3019d8897503613f20a3e18da152e58f3493e444)

Just rearranging and renaming some feature flags, grouping all feature flags involving the export screen and system settings screen

Issue #7394, opened 2025-10-16
Fix Compressed Tally encode/decode for NH Ballot Template


carolinemodic:
Something about the candidate rotation is breaking tabulation via VxQR when using the NH ballot template.

carolinemodic:
See https://votingworks.slack.com/archives/C085YT4798C/p1760635955801179 for continued discussion. 

The problem is that when exporting in specific case for this template we will rotate the cnadidates on the election.contests object, but not update the rotation known about inside of vxdesign

PR #7398, opened 2025-10-16
VxDesign: Block exporting live report nh ballot elections

carolinemodic:
## Overview
Temporary blunt hammer to block the VxQR tabulation issues with NH Ballot Template elections until we find a better solution. See: https://github.com/votingworks/vxsuite/issues/7394 

Since only the VotingWorks org can enable the feature flag this is only possibly a problem for elections in that org.

## Demo Video or Screenshot

Screenshot 2025-10-16 at 3 26 13 PM

## Testing Plan
Tried to export nh/vxqr election and got the error. Changed template or disabled live reports and succeeded. 

## Checklist

- [x ] I have prefixed my PR title with "VxDesign: ", "VxPollBook: ", or "HWTA: " if my change is specific to one of those products.

### DIFF
--- a/apps/design/frontend/src/export_screen.tsx
+++ b/apps/design/frontend/src/export_screen.tsx
@@ -194,7 +194,6 @@ export function ExportScreen(): JSX.Element | null {
     );
   }
 
-  console.log(exportError);
   return (
     <ElectionNavScreen electionId={electionId}>
       <Header>


apps/design/frontend/src/export_screen.tsx
```diff
@@ -191,6 +194,7 @@ export function ExportScreen(): JSX.Element | null {
     );
   }
 
+  console.log(exportError);
```
jonahkagan: Remove console.log

Issue #7312, opened 2025-10-06
VxDesign: MS SEMS Import


adghayes:
Add the ability to create an election in VxDesign from SEMS files.

PR #7400, opened 2025-10-16
VxDesign: Load and convert MS SEMS elections

jonahkagan:
## Overview

Task: https://github.com/votingworks/vxsuite/issues/7312

Adds the ability to load MS SEMS election files and convert them into a VX election via the "Load Election" button.

The conversion code is based on our historical [SEMS converter code](https://github.com/votingworks/vxsuite/blob/m16-dev-branch/services/converter-ms-sems/converter/SEMSinput.py). The nitty gritty details are explained in the code, but the high level is that the user needs to upload two CSV files that are exported from SEMS, one that has most of the election data, and the other that has a mapping of candidate IDs.

One important detail is that we need to save the SEMS IDs for election entities so that we can later use them to convert election results for import back into SEMS. We do this the same way that we did this previously - by using these SEMS IDs as the internal IDs for each election entity. To ensure they are unique and can serve as db primary keys, we prefix them with the election ID (which is a normal VX ID generated by VxDesign).

This functionality is currently controlled by a new feature flag that is only enabled for the VotingWorks org. I plan to figure out how to expose this for MS orgs (of which there are currently none) in a future PR.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/d636468f-ebf7-4672-bed6-30e7559a35ee

## Testing Plan
Mostly automated testing with sample files. It's a bit hard to ensure that the resulting elections are correct (beyond basic validation). I did some manual inspection of the resulting elections (which are snapshotted). I assume we'll figure out some of the edge cases as we use the feature more, but I'm open to other ideas for testing this ahead of time.

Note that there are still some backend integration tests and frontend tests that I have yet to write.
## Checklist

### DIFF
--- a/apps/design/backend/src/convert_ms_election.ts
+++ b/apps/design/backend/src/convert_ms_election.ts
@@ -146,7 +146,7 @@ export function convertMsElection(
   }).filter((row: string[]) => row.every((cell) => cell !== ''));
 
   // Election file format:
-  // 9 sections, each row starts with 0-8 to idenify which section it's a part of
+  // 9 sections, each row starts with 0-8 to identify which section it belongs to
   function sectionRows(sectionNumber: number): Array<string[]> {
     return electionFileRows.filter((row) => row[0] === String(sectionNumber));
   }


apps/design/backend/src/app.ts
```diff
@@ -304,52 +307,84 @@ export function buildApi(ctx: AppContext) {
...
-              candidate.middleName !== undefined &&
-              candidate.lastName !== undefined
-            ) {
-              return candidate;
-            }
-            return {
-              ...candidate,
-              ...splitCandidateName(candidate.name),
+      const electionResult = ((): Result<Election, Error> => {
+        switch (input.upload.format) {
```
kofi-q: The reused common code paths in this method are so minimal that I think could be clearer to have explicit methods like `loadElectionVxf()`, `loadElectionMsSems()`, etc (or something along those lines), since we know which path we're taking at request-time on the client anyway. Wary of the potential road to large switch-case method with additional special-case fields in the input. WDYT?

jonahkagan: I'll mock up the change and see how it feels

jonahkagan: Ok, here's a sketch of what it looks like: https://github.com/votingworks/vxsuite/pull/7405/files. I think the server side does end up a bit clearer, but the frontend is a bit hairy, mainly because of way the form is weirdly split across two parts of a modal. The selector for which file format to use is part of the modal contents, but the action buttons are a separate prop to the modal. I would like to just separate out the forms into two separate components with separate form state, but I can't think of a simple way to do that. I could swap out the entire modal, but then both modals would need to have the format selector within them. Curious if you have any ideas here or any new thoughts after seeing what this code looks like

apps/design/backend/src/convert_ms_election.ts
```diff
@@ -0,0 +1,393 @@
...
+          candidates,
+        };
+      }
+
+      // Ballot measure
+      case '1': {
+        const [, , contestLabel] = row;
+        const [titlePrefix, ...descriptionLines] = contestText.split(/\\n/);
+        const contestTitle = `${titlePrefix}: ${contestLabel}`;
+        const description = descriptionLines.join('\n').trim();
```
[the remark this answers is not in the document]
jonahkagan: Oh great callout, will look into this, thanks

Issue #7366, opened 2025-10-14
VxMarkScan: PAT configuration wizard not appearing - repro and investigation


arsalansufi:
Parent cert discrepancy: https://github.com/votingworks/cert-discrepancies/issues/263

Issue #7285, opened 2025-10-01
VxMarkScan: Repro and debug input from PAT devices being ignored on write-in screen


arsalansufi:
Parent cert discrepancy: https://github.com/votingworks/cert-discrepancies/issues/258

PR #7401, opened 2025-10-17
Ensure creation of `_pat_connection.status` file

kshen0:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7285
Closes https://github.com/votingworks/vxsuite/issues/7366

The above issues report 2 PAT behaviors related to connection status are broken
1. Using a PAT device for write-ins
2. PAT device usage tutorial on first connection

The BMD 150 PAT input (model FAI 100) communicates whether a PAT device is plugged in ("PAT connection status") by writing to a file at `workspace/_pat_connection.status`. The mark-scan backend uses a client `src/pat-input/connection_status_reader.ts` to open the `.status` file and read its contents. However, the daemon only writes the `.status` file when the status first changes, so the file is not guaranteed to exist at server startup.
 
Both broken behaviors rely on `connection_status_reader`. [Logs](https://votingworks.slack.com/archives/C083TM1GPA8/p1760473256140669?thread_ts=1759253837.741409&cid=C083TM1GPA8) show that `connection_status_reader` can't find `_pat_connection.status` at time of mark-scan server startup. It immediately bails, server startup continues, and `connection_status_reader` never checks for existence of the file again. In this situation all checks for connection status return `NOT_CONNECTED`, leading to the broken behavior reported.

This PR makes 2 changes to fix this

1. Writes the `.status` file on daemon startup
2. Adds retry to backend's PAT connection client if the `.status` file doesn't exist

https://github.com/votingworks/vxsuite-build-system/pull/191 creates the `WORKSPACE` directory at imaging time so #1 won't fail due to absence of that directory.

## Demo Video or Screenshot

**Before**

Note that sip/puff still works because those events are sent as keypresses. Handling these keypresses and translating them to DOM navigation is solely handled by the frontend.

https://github.com/user-attachments/assets/bb8fdc94-6968-4d4e-aa96-40e2a3048934

**After**

https://github.com/user-attachments/assets/d48e5db9-af0d-45d3-97c7-c290ea87a534

## Testing Plan

Manually reprod and tested solution.

### DIFF
--- a/apps/mark-scan/backend/src/pat-input/connection_status_reader.test.ts
+++ b/apps/mark-scan/backend/src/pat-input/connection_status_reader.test.ts
@@ -8,7 +8,11 @@ import {
 import * as fs from 'node:fs/promises';
 import { Buffer } from 'node:buffer';
 import { join } from 'node:path';
-import { PatConnectionStatusReader } from './connection_status_reader';
+import {
+  CONNECTION_RETRY_INTERVAL_MS,
+  CONNECTION_TIMEOUT_MS,
+  PatConnectionStatusReader,
+} from './connection_status_reader';
 import { FAI_100_STATUS_FILENAME } from './constants';
 
 const ASCII_ZERO = 48;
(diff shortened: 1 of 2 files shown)


arsalansufi:
@kshen0 One clarifying question for you:

> A separate PR will ensure the WORKSPACE directory is created on machine configuration

What is the exact gap that you're trying to cover with that comment? The WORKSPACE directory is already created if non-existent on backend server startup here:

https://github.com/votingworks/vxsuite/blob/4c070d884ed40a3f6d5333dab1ce1a8f4c356dd5/apps/mark-scan/backend/src/index.ts#L50

arsalansufi:
Relatedly, I'm seeing a few cross dependencies here:
1. Backend server creates workspace directory
2. Daemon writes file inside that directory
3. Backend server reads that file

I think I see now that "A separate PR will ensure the WORKSPACE directory is created on machine configuration" is referring to the case 1 doesn't happen in time for 2.

This does get me wondering whether there's a simpler solution here 🤔 Again happy to discuss sync!

kshen0:
> What is the exact gap that you're trying to cover? 

We're trying to avoid the situation where
* the workspace is not yet created
* the daemon runs first and tries to create the `.status` file but can't because the workspace dir doesn't exist
* the server runs second and creates the workspace dir, but now it's too late

We could solve this by having the daemon create the workspace if it doesn't exist, but as noted in [Slack](https://votingworks.slack.com/archives/C04RWBCCTCZ/p1760549011944199?thread_ts=1760482726.912079&cid=C04RWBCCTCZ) it's simpler and more consistent with other apps to have `setup-machine.sh` create the workspace dir when the machine is first imaged. "Configuration" is the wrong word for me to use here.

> Relatedly, I'm seeing a few cross dependencies here:
> 
> 1. Backend server creates workspace directory
> 2. Daemon writes file inside that directory
> 3. Backend server reads that file
> 
> I think I see now that "A separate PR will ensure the WORKSPACE directory is created on machine configuration" is referring to the case 1 doesn't happen in time for 2.
> 
> This does get me wondering whether there's a simpler solution here 🤔 Again happy to discuss sync!

The sequence you describe here is correct and what we're trying to avoid.

arsalansufi:
Gotcha! Yeah this clarifies my main confusion, that by "configuration" you really meant image creation

PR #7415, opened 2025-10-21
VxDesign: Store exported election data on export and use in VxQR

carolinemodic:
## Overview
Saves the election when exporting to a new table in VxDesign. We could add it to the current elections table but I opted for a new one so we can keep all exported elections, not just the most recent, although for the VxQR use case I don't need that. 

When retrieving an exported election by ballot hash we assert that the contents in the DB still match the ballot hash. Then parse the election. If the election can not be parsed. we return an error to the VxQR frontend, for now this is fine. If we find breaking changes problematic we could in the future parse out only a subset of the election that we actually need here. 

Re-Enables VxQR for the NH ballot template. 

## Demo Video or Screenshot
n/a

## Testing Plan
Added tests, did an e2e vxqr test with a simple election in the NH ballot template and verified that tabulation interpretted the tally properly and that the vxqr frontends display the tally report in the same rotated order that vxsuite does. 

## Checklist


apps/design/backend/migrations/1761063539949_add-exported-elections-table.js
```diff
@@ -0,0 +1,35 @@
...
+exports.shorthands = undefined;
+
+/**
+ * @param pgm {import('node-pg-migrate').MigrationBuilder}
+ * @param run {() => void | undefined}
+ * @returns {Promise<void> | void}
+ */
+exports.up = (pgm) => {
+  // Create table to store exported election blobs for Live Reports validation
+  pgm.createTable('exported_elections', {
```
jonahkagan: >  We could add it to the current elections table but I opted for a new one so we can keep all exported elections, not just the most recent, although for the VxQR use case I don't need that.

Can you share more about your thinking here? I'm not immediately able to think of a use case for checking an exported election other than the most recent.

apps/design/backend/src/app.ts
```diff
@@ -985,6 +1006,15 @@ export function buildUnauthenticatedApi({ logger, workspace }: AppContext) {
...
 
+        // Get the exported election data and validate it
+        const exportedElectionResult =
+          await store.getExportedElection(ballotHash);
+        if (exportedElectionResult.isErr()) {
+          return exportedElectionResult;
+        }
+        const election = exportedElectionResult.ok();
+
+        // Also get the election record for metadata
```
jonahkagan: It makes me a little nervous that all of these endpoints load both the database election record and also the exported election. Seems like it would be easy to get them mixed up. Is it possible to just use the exported election? E.g. we could have a store method `getLastExportedElection(electionId)`

apps/design/backend/src/app.test.ts
```diff
@@ -3637,3 +3639,177 @@ test('decryptCvrBallotAuditIds', async () => {
     expect(contents.CVR[0]).toEqual(cvr.castVoteRecord);
   }
 });
+
+test('getExportedElectionDefinition returns the exported election including reordering updates on render', async () => {
```
jonahkagan: I think this would make a bit more sense to live in `store.test.ts` because it's not testing the app API. That being said, if it's too much trouble to have it live there, I'm ok with it being here, but maybe we can just change the test title to `store.getExportedElectionDefinition ...` because I was a bit confused at first when I saw it

PR #7417, opened 2025-10-21
Address remaining VxScan log spam

arsalansufi:
## Overview

We've been reducing log spam to address a variety of logging related cert discrepancies, and this PR should address the last holdout on VxScan - successive SCANNING_DISABLED and successive SCANNING_ENABLED logs:

IMG_1726

We now only log the first of these logs in a successive sequence and omit the rest.

Side note: I recognize that preventing spammy logs involves some whack-a-mole. And as I fix this one source, a new one may come up elsewhere. Or the logic that protects against this case may break for some reason. To help us avoid regressions, I'm going to update our QA checklists to review logs for spam.

## Testing Plan

- [x] Tested manually
- [x] Added a regression test that I confirmed fails when the current patch isn't in place

## Checklist

- [ ] ~I have prefixed my PR title with "VxDesign: ", "VxPollBook: ", or "HWTA: " if my change is specific to one of those products.~ N/A
- [ ] ~I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate for any new user actions.~ N/A
- [ ] ~I have added the "user_facing_change" label to this PR, if relevant, to automate an announcement in #machine-product-updates.~ N/A

PR #7424, opened 2025-10-23
Coalesce repeated log lines to reduce log spam while still maintaining fidelity

arsalansufi:
## Overview

An alternative to https://github.com/votingworks/vxsuite/pull/7417, this PR presents a generalized application-level solution for coalescing repeated log lines to reduce log spam while still maintaining fidelity. Log spam has been a factor behind a variety of logging-related during cert, e.g., running out of disk space on VxMarkScan, generating log files larger than the FAT32 4GB file size limit, generating log files too large to be converted to CDF format in our 1GB /tmp directory, etc.

Note that we used to coalesce logs at the system level but turned this off as it often interfered with our log routing/matching rules: https://github.com/votingworks/vxsuite-complete-system/blob/6a65989c73592dbe36826c340b12c023a1a31738/config/rsyslog.conf#L41-L44. Hence my landing on an application-level solution.

503907403-4e719f52-6af4-4ffb-87a7-3c1f6249ea11

The logic can be summarized as follows:

The first time that a log is seen, it'll always be emitted immediately. Exact repeats of that log after will then be collected and emitted when:
* A different log is encountered and the repeat sequence ends
* The repeat buffer collects 100 logs.
* The time between the first repeat and last repeat exceeds 60 seconds.

Sample repeat log:

> {"source":"vx-scan-backend","eventId":"scanner-state-machine-event","eventType":"application-action","user":"system","message":"Event: SCANNING_DISABLED","disposition":"na","eventObject":"{\"type\":\"SCANNING_DISABLED\"}","repeatCount":"58","firstRepeatAt":"2025-10-23T19:32:01.978Z","lastRepeatAt":"2025-10-23T19:32:30.900Z"}

See the tests for many more examples!

Relevant Slack thread: https://votingworks.slack.com/archives/CEL6D3GAD/p1761149528362049

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

## Checklist

### DIFF
--- a/libs/logging/src/base_logger.ts
+++ b/libs/logging/src/base_logger.ts
@@ -28,16 +28,16 @@ export const CLIENT_SIDE_LOG_SOURCES = [
 ];
 
 /**
- * The maximum number of repeated log lines before we emit a log line with repeat details, even if
- * a repeat sequence hasn't ended yet.
+ * The maximum number of repeated log lines before we emit a repeat summary log line, even if a
+ * repeat sequence hasn't ended yet.
  */
-const MAX_REPEAT_COUNT_BEFORE_REPEAT_LOG_EMIT = 100;
+const MAX_REPEAT_COUNT_BEFORE_REPEAT_SUMMARY_LOG_EMIT = 100;
 
 /**
  * The maximum number of seconds between the first and last repeat of a log line before we emit a
- * log line with repeat details, even if a repeat sequence hasn't ended yet.
+ * repeat summary log line, even if a repeat sequence hasn't ended yet.
  */
-const MAX_SECONDS_BETWEEN_FIRST_AND_LAST_REPEAT_BEFORE_REPEAT_LOG_EMIT = 60;
+const MAX_SECONDS_BETWEEN_FIRST_AND_LAST_REPEAT_BEFORE_REPEAT_SUMMARY_LOG_EMIT = 60;
 
 const debug = makeDebug('logger');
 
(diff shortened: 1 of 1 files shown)


libs/logging/src/base_logger.ts
```diff
@@ -25,18 +27,118 @@ export const CLIENT_SIDE_LOG_SOURCES = [
...
+    this.firstRepeatAt = undefined;
+    this.lastRepeatAt = undefined;
+  }
+
+  private recordRepeat(): number {
+    this.repeatCount += 1;
+    if (this.repeatCount === 1) {
+      this.firstRepeatAt = new Date();
+    }
+    this.lastRepeatAt = new Date();
```
jonahkagan: Should these timestamps be the same for the first repetition of a log? Maybe doesn't matter in real world usage

arsalansufi: Oh yeah good catch! They happened to be exactly the same in my testing but lemme make sure they really are the same

libs/logging/src/base_logger.ts
```diff
@@ -48,6 +150,37 @@ export class BaseLogger {
...
+    const logLine = this.prepareLogLine(eventId, user, logData);
+
+    const { isRepeat, repeatCount, secondsBetweenFirstAndLastRepeat } =
+      this.repeatLogTracker.track(logLine);
+    if (!isRepeat) {
+      // If a repeat sequence has ended, emit a log line with repeat details
+      this.repeatLogTracker.flushAndRestartTracking(
+        (l) => this.logInternal(l, outerDebug),
+        logLine
+      );
```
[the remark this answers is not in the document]
arsalansufi: Landed on https://github.com/votingworks/vxsuite/pull/7424/commits/8b7dcb6f19c6eb0624acdd1a811808b1bca12552, which I'm pretty happy with - thanks for the nudge!

PR #7425, opened 2025-10-23
DRAFT: Add ordering information to Ballot Styles

carolinemodic:
This is still very much a work in progress but wanted to get some high level feedback from @jonahkagan on the current approach to apply rotation rules at the time of generating ballot styles and include ordering information on the Ballot Style object. 

I'm primarily looking for feedback on what should live in libs/hmpb vs. apps-design-backend. Technically as discussed this rotation logic isn't only used for hmpbs, but given that thats where we define the rest of template logic, it feels appropriate to co-locate the logic for a specific template in there, as I do here. The entire ballot_styles.ts file in VxDesignBackend could be moved to libs/hmpb other then the dependency on VxDesign's version of the BallotStyle type. The advantage to that would be that in the fixture generation in libs/hmpb and libs/fixture-generator (for election-packages) we could access this logic. Otherwise we need to manually semi-reimplement this logic in the fixture generation (as I do in libs/hmpb here), or same that the "baseElection" needs to have this defined in order for it to come through on the other end. It feels a little weird in my current setup now that rotation logic application is split across vxdesign and libs/hmpb where it was previously being tested for but maybe thats ok. 

I'm also thinking about how if you change the ballot template, we will need to regenerate ballot styles which I don't think we do today. I briefly considered if the layOutBallotsAndElectionDefinition function should regenerate ballot styles to make sure that they are accurate, but this is a little strange when the ballotProps are created before that function is called and exist in relation to what ballot styles exist... 

Recommend looking by commit so you can ignore the fixtures regen.

### DIFF
--- a/libs/types/src/election.test.ts
+++ b/libs/types/src/election.test.ts
@@ -12,6 +12,7 @@ import {
   getContestsFromIds,
   formatBallotHash,
   getDistrictIdsForPartyId,
+  getOrderedContests,
   getPartyAbbreviationByPartyId,
   getPartyFullNameFromBallotStyle,
   getPartyIdsInBallotStyles,
@@ -36,10 +37,12 @@ import {
 import {
   BallotIdSchema,
   HmpbBallotPaperSize,
+  BallotStyle,
   BallotStyleId,
   BallotStyleSchema,
   CandidateContest,
   CandidateSchema,
+  Election,
   ElectionDefinitionSchema,
   PartyIdSchema,
   WriteInIdSchema,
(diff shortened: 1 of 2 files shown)


apps/design/backend/src/ballot_styles.ts
```diff
@@ -1,42 +1,172 @@
...
+  getAllOrderedContestSetsForNhBallot,
+} from '@votingworks/hmpb';
 import { BallotStyle } from './types';
 
+/**
+ * Generates ordered contests for VxDefault ballot template.
+ * This template does not apply any rotation.
+ * Returns a single set with the contests in their original order.
+ */
+function getOrderedContestsWithoutRotation({
```
carolinemodic: defined the ordering for the default template, could be in libs/hmpb colocated with that template but I thought "no rotation" might be a more general case multiple templates use

PR #7433, opened 2025-10-28
pdi-scanner: Add locations to USB error messages

jonahkagan:
## Overview

We've had a number of low-level USB error messages show up in rare cases, but haven't been able to trace them back to a specific function call because only the message is logged ([example](https://votingworks.slack.com/archives/C09MAG9NQSE/p1761149927230019?thread_ts=1761146401.576809&cid=C09MAG9NQSE)). This PR adds a location trace to USB errors so we can at least know where they originate in our codebase (though it won't be a full stack trace into any dependencies). Uses the [locate-error crate](https://crates.io/crates/locate-error).

## Demo Video or Screenshot
I changed `nusb::list_devices` to return an error with the message "test error", then ran pdictl with a `connect` command and got this output:

```
{"response":"error","code":"other","message":"usb error: nusb error: test error (libs/pdi-scanner/src/rust/scanner.rs:31:27) (libs/pdi-scanner/src/rust/lib.rs:13:23)"}
```

Here are the code points in the trace:
- [libs/pdi-scanner/src/rust/scanner.rs:31:27](https://github.com/votingworks/vxsuite/blob/jonah/pdi-scanner-error-trace/libs/pdi-scanner/src/rust/scanner.rs#L31)
- [libs/pdi-scanner/src/rust/lib.rs:13:23](https://github.com/votingworks/vxsuite/blob/jonah/pdi-scanner-error-trace/libs/pdi-scanner/src/rust/lib.rs#L13)

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1174,6 +1174,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "hashbrown"
+version = "0.16.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5419bdc4f6a9207fbeba6d11b604d481addf78ecd10c11ad51e76c2f6482748d"
+
+[[package]]
 name = "heck"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


eventualbuddha:
This seems fine. I'm curious what alternatives you evaluated and what the pros and cons were. Does `tracing` not give what we need? Do we want a more generalized solution than this, or are we okay with limiting it to only USB errors? A tool like [`anyhow`](https://docs.rs/anyhow/latest/anyhow/) (and maybe also `thiserror`?) allows adding context and gathering a stack of error information that may be useful.

jonahkagan:
Closing in favor of https://github.com/votingworks/vxsuite/pull/7442

PR #7436, opened 2025-10-28
VxDesign: Add orderedDisplayCandidatesByContest as optional BallotStyle field for Ballot Rotation

carolinemodic:
## Overview
Implements a new optional field orderedDisplayCandidatesByContest on the BallotStyle object in an Election. In the future we may want to make this required, but in order to make this a non-breaking change for now we will have it as optional and fallback to the ordered as defined by election.contests. The field defines how candidates should be ordered on a given ballot style for candidate contests. Yes/No Contests never need to have ordering specified. The libs/hmpb logic to lay out the bubbles will now use this ordering for determining the order on a given ballot. I added a function to apply rotation in a way that it will be different for different precincts for the unit tests and to generate an election more easily locally by temporarily using it for the default template - we can do this again when needed to create a test election package for SLI. 

I intend to add partyIds to DisplayCandidate in a followup PR for basic initial support of cross-endorsed candidates, leaving it out for now just to avoid adding even more complication to this PR. 

In the future we may have the rotation rules be a user-selectable item, perhaps a subset of rotation rules is defined by the template, since we currently have a 1-to-1 mapping of templates to rotation needed I am punting on that for the time being while we gather more data on how this should work. 

## Demo Video or Screenshot
Example browsing ballots when using the rotateAlphabeticallyByPrecinct (this case causes more ballot styles to be created then we previously have needed):

https://github.com/user-attachments/assets/40d6e576-ced7-433a-b017-67283caea412

## Testing Plan
Ran Tests
Exports NH ballot template election - inspected that election.json rotated candidates in the ballot styles AND on the overall election
Hacked default ballot template to rotateAlphabeticallyByPrecinctName and manually inspected election.json and rendered ballots. 

Note: I'm a little torn on whether to make a new test fixture with the orderings defined (or modified an existing one, maybe primary with precinct splits) for testing consumers, or if I should just modify the election in the tests to add the ordering when I need it, but will address that in future prs. 

## Checklist


libs/types/src/election.ts
```diff
@@ -410,6 +410,16 @@ export const PrecinctsSchema = z
     }
   });
 
+// Represents a bubble option that should be displayed for selection on a ballot.
+export interface DisplayCandidate {
+  id: CandidateId;
+  // TODO(CROSS_ENDORSE): add more partyIds as needed.
+}
```
jonahkagan: Naming nit: `DisplayCandidate` kind of makes me think this should have the info needed to display a candidate. I was thinking of this more as a type that just represents a candidate in an ordering, so maybe `OrderedCandidate` or `CandidateOption` would make more sense. Then the ballot style field could be `orderedCandidatesByContest` or `orderedCandidateOptionsByContest`. I don't feel strongly about this at all though

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -813,7 +804,7 @@ async function BallotPageContent(
   );
   // For now, just one section for candidate contests, one for ballot measures.
   // TODO support arbitrarily defined sections
-  const contests = getContests({ election, ballotStyle });
+  const contests = getOrderedContests({ election, ballotStyle });
```
jonahkagan: I'm realizing that whenever we define contest sections, it's going to have to interact with `getOrderedContests`. Since we aren't planning to have different contest orders across ballot styles, I wonder if we should change the API to be `getOrderedCandidates` and call it for each contest to constrain where rotation applies to as localized an area as possible.

libs/hmpb/src/ballot_styles.ts
```diff
@@ -68,28 +82,48 @@ export function generateBallotStyles(params: {
...
+              precincts,
+              districtIds,
+              precinctsOrSplitIds,
+              electionId,
+            }
+          );
+          // Create ballot styles for each ordered contest set and language config
+          return orderedContestSets.flatMap(
+            ({ orderedContests, precinctsOrSplits: precinctsForSet }) => {
+              ballotStyleIndex += 1;
```
jonahkagan: I wonder if we should generate ballot style IDs that help us see which ballot styles have the same contests with a different rotation? E.g. if `1_en`, `2_en` have the same contests with different rotation, maybe we name them `1.1_en` and `1.2_en` or something

libs/hmpb/src/render_ballot.tsx
```diff
@@ -637,26 +609,30 @@ export async function layOutBallotsAndCreateElectionDefinition<
   const contests = election.contests.map((contest) => {
     if (template.isAllBubbleBallot) return contest;
     if (contest.type !== 'candidate') return contest;
-    const gridLayoutsWithContest = gridLayouts.filter((layout) =>
-      layout.gridPositions.some(
-        (gridPosition) => gridPosition.contestId === contest.id
-      )
+    const ballotStylesWithContest = election.ballotStyles.filter(
```
jonahkagan: Update comment on L601-608 now that this logic is changed

libs/hmpb/src/ballot_rotation.ts
```diff
@@ -0,0 +1,191 @@
...
+ */
+function getDefaultCandidateOrdering({
+  contests,
+  precinctsOrSplitIds,
+  districtIds,
+}: RotationParams): CandidateOrderingSet[] {
+  // Get contests for this ballot style (filtered by district)
+  const ballotStyleContests = contests.filter((contest) =>
+    districtIds.includes(contest.districtId)
+  );
```
jonahkagan: I kind of feel like this should be done higher up in the call stack, since it applies to all ballot rotation. I.e. maybe when generating the ballot style, we should filter which contests apply and then pass that into the rotation logic. We could also ensure that only candidate contests get passed in, which would simplify some of the rotation logic.

libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
carolinemodic: Changing to this for now, if it ends up being easier when I get to the consumers to also have getOrderedContests I'll add it back, but I agree doing it on a contest-level kind of makes more sense now that we aren't ordering contests

PR #7439, opened 2025-10-28
VxDesign: Add support for cross endorsed candidates to rotation data model

carolinemodic:
## Overview
Depends on https://github.com/votingworks/vxsuite/pull/7436

Adds support to the data model for cross-endorsed candidates. You can not currently create a cross-endorsed candidate in VxDesign, either ballot template right now by default would handle that as having ONE bubble position for the candidate if that changed in the future. A future template, like for NH state elections, will need to explictly in its rotation logic return unique DisplayCandidate items per-party if they want seperate bubbles generated, and displayed on the bmd per-party. 

## Demo Video or Screenshot
N/A

## Testing Plan
getOrderedContests handling of cross-endorsement tested through unit tests since we can't actually create these elections easily

## Checklist

### DIFF
--- a/libs/types/src/election.test.ts
+++ b/libs/types/src/election.test.ts
@@ -50,6 +50,7 @@ import {
   DistrictId,
   Precinct,
   BallotStyle,
+  Election,
 } from './election';
 import { safeParse, safeParseJson, unsafeParse } from './generic';
 import {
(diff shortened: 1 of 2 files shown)


carolinemodic:
moved validation logic to when parsing the election

PR #7446, opened 2025-11-03
VxMark: Handle cross-endorsed candidates properly

carolinemodic:
## Overview
Updates the bmd handling of cross-endorsed candidates so that you can select any combination of party-options for a candidate and the appropriate selections will be reflected properly on:
- The review screen
- Summary printed ballot
- Marks printed on a pre-printed ballot 

Printing a full bubble ballot is not yet incorporated into main and not explictly handled or tested. 

Modifying the parties within a Candidate object type (so within the CandidateVote) feels a little icky but there isn't a better way to do this without exploding the refactoring scope. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/6295280d-83a6-40e5-ad08-de5805738a4d

## Testing Plan
Had to hack a few things in VxDesign to create a rotated example election with a cross endorsement and grid layouts but once I had that I tested all voter flows in VxMark and visually inspected both summary ballots and bubble-marking prints to make sure the bubbles looked like they were being printed in the proper alignment for all possible ways of voting for a cross-endorsed candidate. 

## Checklist

### DIFF
--- a/libs/hmpb/src/vote_matching.ts
+++ b/libs/hmpb/src/vote_matching.ts
@@ -1,5 +1,9 @@
 import { assert, deepEqual } from '@votingworks/basics';
-import { Candidate, GridPositionOption } from '@votingworks/types';
+import {
+  Candidate,
+  GridPosition,
+  GridPositionOption,
+} from '@votingworks/types';
 
 /**
  * Matches a vote to a grid position.
@@ -15,11 +19,12 @@ import { Candidate, GridPositionOption } from '@votingworks/types';
 export function voteMatchesGridPosition(
   vote: Candidate,
   gridPos: GridPositionOption,
-  allGridPositions: readonly GridPositionOption[]
+  allGridPositions: readonly GridPosition[]
 ): boolean {
   // Get all grid positions for this contest and candidate ID
   const allCandidateGridPositions = allGridPositions.filter(
-    (gp) => gp.type === 'option' && gp.optionId === vote.id
+    (gp): gp is GridPositionOption =>
+      gp.type === 'option' && gp.optionId === vote.id
   );
   // If there are not multiple grid positions for this candidate, we can
   // directly match on option ID


libs/hmpb/src/vote_matching.ts
```diff
@@ -0,0 +1,59 @@
...
+  CandidateContest,
+  getOrderedCandidatesForContestInBallotStyle,
+  GridPositionOption,
+} from '@votingworks/types';
+
+/**
+ * Matches a vote to a grid position by using positional matching.
+ *
+ * For candidate contests with cross-endorsed candidates, the order of grid
+ * positions maps to the order of OrderedCandidateOptions in the ballot style.
```
jonahkagan: How are we sure that this is true? Do we have any code that enforces that grid positions are in the same order as the ordered candidate options? IIRC, the order of grid positions is currently determined by what order the browser returns the bubble elements in when we are extracting grid layouts from the rendered ballot.

libs/mark-flow-ui/src/components/candidate_contest.tsx
```diff
@@ -87,8 +88,46 @@ const WriteInModalActionsSidebar = styled.div`
...
+function getCandidateChoiceId(
+  candidateId: CandidateId,
+  partyIds?: readonly string[]
+): string {
+  if (!partyIds || partyIds.length === 0) {
+    return candidateId;
+  }
+  // Sort party IDs to ensure consistent ordering
+  return `${candidateId}|${[...partyIds].sort().join(',')}`;
+}
```
jonahkagan: I'm a bit worried that using a string to identify unique candidate options will make it easy to make a mistake where we accidentally use the wrong id (e.g. using `candidate.id` instead). Would it be possible to just pass around the entire candidate object instead? Or maybe we could make a new object that represents an option and use that instead of candidate objects.

E.g.
```ts
type CandidateOption {
  candidate: Candidate,
  equals(otherOption: CandidateOption) {
     // compare candidate.id and candidate.partyIds
  }
}
```

I feel like this edge case is pretty complicated so I'm grasping for something to make it feel safer/more manageable at a type level.

libs/hmpb/src/vote_matching.ts
jonahkagan: Should we consider moving the HMPB bubble position info into the ballot style ordered candidates instead?

carolinemodic:
@jonahkagan this is ready for another look (tomorrow is fine)

libs/mark-flow-ui/src/components/candidate_contest.tsx
```diff
@@ -179,21 +206,22 @@ export function CandidateContest({
...
 
-  function handleUpdateSelection(candidateId: string) {
+  function handleUpdateSelection(candidate: Candidate) {
     /* istanbul ignore else - @preserve */
-    if (candidateId) {
-      const candidate = findCandidateById(vote, candidateId);
-      if (candidate) {
-        if (candidate.isWriteIn) {
-          setWriteInPendingRemoval(candidate);
+    if (candidate) {
```
jonahkagan: Seems like it's not possible for `candidate` to be undefined here, so this `if` is unnecessary now

libs/mark-flow-ui/src/components/candidate_contest.tsx
```diff
@@ -351,19 +380,35 @@ export function CandidateContest({
...
+                !isEquivalentToSelected;
+
               function handleDisabledClick() {
                 handleChangeVoteAlert(candidate);
               }
               let prefixAudioText: ReactNode = null;
               let suffixAudioText: ReactNode = null;
 
-              const numVotesRemaining = contest.seats - vote.length;
+              // Recalculate unique candidates after this selection/deselection
```
jonahkagan: This comment doesn't quite make sense to me since we're not in a selection/deselection event handler, but simply rendering candidates

jonahkagan (approved):
Thanks for taking a second pass on this. It feels a lot clearer/safer to me now!

Are there any downstream consequences to having multiple copies of the same `Candidate` in the resulting vote array?

libs/hmpb/src/marking.ts
```diff
@@ -177,27 +184,48 @@ type MarkInfo =
...
-    if (typeof vote === 'string') {
-      if (vote === gridPos.optionId) return {};
-      continue;
-    }
+    const allGridPositions = layout.gridPositions.filter(
+      (gp): gp is GridPositionOption => gp.type === 'option'
+    );
 
-    if (vote.id === gridPos.optionId) return {};
+    if (voteMatchesGridPosition(candidateVote, gridPos, allGridPositions)) {
```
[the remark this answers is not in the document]
carolinemodic: but I'm simplifying and adjusting the comments around the function some

PR #7472, opened 2025-11-05
VxAdmin: Use ballot style rotation, deduplicating multi-endorsed candidates, for Adjudication flows and Manual Tally entry

carolinemodic:
## Overview
From discussion with Matt/Product we want to rotate the candidates on these screens according to the appropriate ballot style, but we do NOT want to expose the complexity of multi-endorsed candidates. The order of a multi-endorsed candidate will be the placement it first appears. Matt agrees some of the flows here feel non-ideal for multi-endorsed candidates but thinks it is fine and if we get user feedback that its confusing we can revisit in the future. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/b93429c9-1b66-4628-9ddf-ca7bbdf9a715

## Testing Plan
Used fixture for manual testing, verified tallys after adjudication / adding manual tallies. 

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/contest_adjudication_screen.test.tsx
+++ b/apps/admin/frontend/src/screens/contest_adjudication_screen.test.tsx
@@ -1,6 +1,14 @@
 import { afterEach, beforeEach, describe, expect, test } from 'vitest';
-import { readElectionTwoPartyPrimaryDefinition } from '@votingworks/fixtures';
-import { ContestId, ContestOptionId, Id } from '@votingworks/types';
+import {
+  readElectionTwoPartyPrimaryDefinition,
+  electionFamousNames2021Fixtures,
+} from '@votingworks/fixtures';
+import {
+  BallotStyleGroupId,
+  ContestId,
+  ContestOptionId,
+  Id,
+} from '@votingworks/types';
 import userEvent from '@testing-library/user-event';
 import { Route } from 'react-router-dom';
 import type {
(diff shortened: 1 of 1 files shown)


apps/admin/frontend/src/screens/contest_adjudication_screen.tsx
```diff
@@ -317,13 +329,29 @@ export function ContestAdjudicationScreen(): JSX.Element {
 
   const adjudicateCvrContestMutation = adjudicateCvrContest.useMutation();
 
-  const officialOptions = useMemo(
-    () =>
-      isCandidateContest
+  const officialOptions = useMemo(() => {
+    if (!cvrVoteInfoQuery.data) {
+      // Return a placeholder that will be replaced once data loads
```
jonahkagan: Could we remove the useMemo and move this whole thing down to after the guard that checks that all the data has loaded? Then we wouldn't need a placeholder. I'm not sure how important the memoization is for performance, but it doesn't seem like this would be a very expensive calculation since there are only a small number of candidates per contest

carolinemodic: we need this argument for the useContestAdjudicationState hook so I think we need a placeholder regardless

jonahkagan: Oh gotcha. Well I'd probably pull the data loading out into a parent component in that case but that refactor might be more scope than makes sense for this PR

PR #7501, opened 2025-11-10
VxDesign: Support ballot rotation in VXF <-> CDF Conversion 

carolinemodic:
## Overview
Maintains the order defined by CDF's OrderedContent when converting CDF -> VXF. Adds support to define the OrderedContent in CDF in order based on the VXF Election's defined orderedCandidatesByContest on the BallotStyle object. This means that ballot rotation ordering is maintained throughout VXF/CDF conversions.  It is not clear how to represent multi-endorsed candidates where we want multiple Option Positions for a single candidate id in the CDF format, we would need to use a different primary ID for the option positions defined in CDF then just using the candidate id which is a larger and more complex change, so currently CDF format will only support displaying 1 option for multi-endorsed candidates.

## Demo Video or Screenshot
N/A

## Testing Plan
Added rotation to main CDF conversion tests. Mocked VxDesign to rotate candidates alphabetically by precinct name to create a sample rotation election, exported in CDF format, imported to Vxsuite and tested rotation came through in VxMark and VxAdmin as expected. 

## Checklist

### DIFF
--- a/libs/types/src/cdf/ballot-definition/convert.ts
+++ b/libs/types/src/cdf/ballot-definition/convert.ts
@@ -573,8 +573,18 @@ export function convertVxfElectionToCdfBallotDefinition(
                     '@type': 'BallotDefinition.OptionPosition',
                     Sheet: position.sheetNumber,
                     Side: position.side as Cdf.BallotSideType,
+                    // Technically these should be in inches, not grid
+                    // coordinates, since that's the measurement unit
+                    // specified in the ballot format, but grid coordinates
+                    // are what our interpreter uses, and converting to inches
+                    // and back would just add arbitrary confusion.
                     X: position.column,
                     Y: position.row,
+
+                    // It's not clear what the height/width of an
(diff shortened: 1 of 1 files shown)


libs/types/src/cdf/ballot-definition/convert.ts
```diff
@@ -511,50 +645,7 @@ export function convertVxfElectionToCdfBallotDefinition(
...
-                  OptionPosition: [
-                    {
-                      '@type': 'BallotDefinition.OptionPosition',
-                      Sheet: position.sheetNumber,
-                      Side: position.side as Cdf.BallotSideType,
-                      // Technically these should be in inches, not grid
-                      // coordinates, since that's the measurement unit
-                      // specified in the ballot format, but grid coordinates
-                      // are what our interpreter uses, and converting to inches
-                      // and back would just add arbitrary confusion.
```
jonahkagan: I think we should preserve some of this commentary

PR #7507, opened 2025-11-12
VxPrint: Hook up election info bar to backend; improve handling of Unconfigured machine

nikhilb4a:
## Overview

This fills in the TBDs for the election info bar by connecting it to the backend, following the pattern in `admin` of having the API for `getElectionRecord` instead of `getElectionDefinition`, so that way we can include the `ElectionPackageHash`

**Update**
This PR now also improves the handling of the unconfigured machine states for each auth type, motivated by the PR discussion below about the `ScreenWrapper` for unconfigured states.

Updated behavior:

System admins can see the normal Screen Wrapper when unconfigured - 
Screenshot 2025-11-13 at 4 28 25 PM

Election managers see the Screen Wrapper without left nav items, shows `UnconfiguredElectionManagerScreen` - 
Screenshot 2025-11-13 at 4 28 15 PM

Poll Workers cannot log in when unconfigured, instead show `InvalidCardScreen` on card insertion - 
Screenshot 2025-11-13 at 4 28 33 PM

## Demo Video or Screenshot

Before (TBDs in sidebar) - 
Screenshot 2025-11-12 at 1 55 33 PM

After - 
Screenshot 2025-11-12 at 1 54 57 PM

## Checklist

### DIFF
--- a/apps/print/frontend/src/app.tsx
+++ b/apps/print/frontend/src/app.tsx
@@ -9,6 +9,7 @@ import {
   SetupCardReaderPage,
   SystemCallContextProvider,
   UnlockMachineScreen,
+  VendorScreen,
 } from '@votingworks/ui';
 import { BrowserRouter } from 'react-router-dom';
 import { DevDock } from '@votingworks/dev-dock-frontend';
@@ -19,6 +20,7 @@ import {
   isElectionManagerAuth,
   isPollWorkerAuth,
   isSystemAdministratorAuth,
+  isVendorAuth,
 } from '@votingworks/utils';
 import { assert } from '@votingworks/basics';
 import { MachineLockedScreen } from './screens/machine_locked_screen';
@@ -31,7 +33,9 @@ import {
   getAuthStatus,
   getElectionRecord,
   getUsbDriveStatus,
+  logOut,
   systemCallApi,
+  unconfigureMachine,
 } from './api';
 import { ElectionManagerApp } from './election_manager_app';
 import { UnconfiguredElectionManagerScreen } from './screens/unconfigured_election_manager_screen';
@@ -48,6 +52,8 @@ function AppRoot({
   apiClient: ApiClient;
 }): JSX.Element | null {
   const checkPinMutation = checkPin.useMutation();
+  const logOutMutation = logOut.useMutation();
+  const unconfigureMutation = unconfigureMachine.useMutation();
   const getAuthStatusQuery = getAuthStatus.useQuery();
   const getElectionRecordQuery = getElectionRecord.useQuery();
   const getUsbDriveStatusQuery = getUsbDriveStatus.useQuery();
@@ -119,6 +125,17 @@ function AppRoot({
   }
 
   if (authStatus.status === 'logged_in') {
+    if (isVendorAuth(authStatus)) {
+      return (
+        <VendorScreen
+          logOut={logOutMutation.mutate}
+          rebootToVendorMenu={apiClient.rebootToVendorMenu}
+          unconfigureMachine={() => unconfigureMutation.mutateAsync()}
+          isMachineConfigured={electionRecord !== null}
+        />
+      );
+    }
+
     if (isSystemAdministratorAuth(authStatus)) {
       return <SystemAdministratorApp />;
     }


apps/print/frontend/src/components/screen_wrapper.tsx
```diff
@@ -10,20 +10,30 @@ import {
...
-  const getElectionDefinitionQuery = getElectionDefinition.useQuery();
-  const electionDefinition = getElectionDefinitionQuery.data;
+  const getElectionRecordQuery = getElectionRecord.useQuery();
+  const getMachineConfigQuery = getMachineConfig.useQuery();
+
+  if (!getElectionRecordQuery.isSuccess || !getMachineConfigQuery.isSuccess) {
+    return null;
+  }
+
+  const { electionDefinition, electionPackageHash } = assertDefined(
```
kshen0: Will we ever render this wrapper for unconfigured screens, which would result in `getElectionRecordQuery.data` being `null`?

apps/print/frontend/src/election_manager_app.tsx
```diff
@@ -1,16 +1,19 @@
...
 import { PrintScreen } from './screens/print_screen';
 import { TitleBar } from './components/title_bar';
 import { electionManagerRoutes } from './routes';
 import { SettingsScreen } from './screens/settings_screen';
 
 function ElectionManagerElectionScreen(): JSX.Element | null {
-  const electionDefinitionQuery = getElectionDefinition.useQuery();
-  const { election } = assertDefined(electionDefinitionQuery.data);
+  const electionRecordQuery = getElectionRecord.useQuery();
+  if (!electionRecordQuery.isSuccess) return null;
```
kshen0: nit: I generally prefer the `then` clause of an `if` statement to be on its own line wrapped with braces even if trivial

apps/print/frontend/src/components/screen_wrapper.tsx
nikhilb4a: Good call, both good ideas. I like calling it `ConfiguredScreenWrapper` for clarity

apps/print/frontend/src/election_manager_app.tsx
nikhilb4a: Agreed, slipped by!

apps/print/frontend/src/app.tsx
```diff
@@ -71,21 +72,8 @@ function AppRoot({
...
 
-  if (authStatus.status === 'remove_card') {
-    return <RemoveCardScreen productName="VxPrint" />;
-  }
-
-  if (
-    authStatus.status === 'checking_pin' &&
-    ['vendor', 'system_administrator'].includes(authStatus.user.role)
-  ) {
+  if (authStatus.status === 'checking_pin') {
```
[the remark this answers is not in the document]
nikhilb4a: I also added support for the `Vendor` card because it was annoying when I'd accidentally insert it and then need to restart the app to get out of the `Unhandled state` screen 😅

apps/print/frontend/src/components/screen_wrapper.tsx
kshen0: > for EMs should we include the stripped down wrapper, similar to Pollbook? Or should we just show the UnconfiguredScreen?

I think we should do whatever is simplest to implement. The empty sidebar is nice for consistency but I wouldn't go out of your way to build it out or special case anything.

apps/print/frontend/src/app.tsx
```diff
@@ -100,32 +88,53 @@ function AppRoot({
...
+    return (
+      <InvalidCardScreen
+        reasonAndContext={authStatus}
+        recommendedAction={
+          authStatus.reason === 'machine_not_configured'
+            ? 'Use a system administrator or election manager card.'
+            : 'Use a valid card.'
+        }
+        cardInsertionDirection="right"
+      />
```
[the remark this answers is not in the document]
nikhilb4a: Admittedly I'm still wrapping my head around the different auth states, so sorry if I answer incorrectly, but my understanding is that the `machine_not_configured` authStatus.reason only happens when a Poll Worker card is inserted to a machine that isn't configured, so that would be the only case where we'll show `Use a system administrator or election manager card.`

And then `Use a valid card.` will show when a card from another election is used

PR #7524, opened 2025-11-13
ballot-interpreter: Refine timing mark smoothing and scoring

jonahkagan:
## Overview

Refines the smoothing and scoring of timing marks in the `corners` algorithm to make scoring more consistent. There are two main problems tackled here:

**Noise causing nubs/spurs**
Noise (e.g. from imprinter marks) can cause nubs/spurs on the edges of a timing mark. We smooth these out with a median filter. The previous median filter logic considered a window of 8 pixels centered around each pixel in the top/bottom edge of the timing mark. However, at the ends of the edges, it would shrink the window so as not to overflow. This meant that noise at the corners of a timing mark got smoothed less than noise in the middle.

Screenshot 2025-11-13 at 1 52 49 PM

To help with this, I changed the logic to wrap around at the corners to the other end of the edge. My reasoning was that if there was noise at one corner but not the other, wrapping around would let the good corner help smooth out the bad corner. If there's noise at both corners, well, there might not be anything we can do. But specifically in the imprinter case, the noise is generally at one corner, so this is a pragmatic improvement for a known case.

Screenshot 2025-11-13 at 2 06 01 PM

**Scoring variability due to subpixel binarization**
Timing mark edges often don't line up perfectly with pixel boundaries. This causes fuzzy gray pixelation at the edge of the timing mark. If only some of the gray is above the binarization threshold and some is below, this ends up causing an extra partial row of pixels to be included in the timing mark shape. This causes a 1 pixel offset when scoring the timing mark, which can lead to an otherwise perfectly identified timing mark scoring as low as 91%.

Screenshot 2025-11-13 at 2 25 05 PM

 
This isn't inherently a big issue, since we set the acceptable threshold for corner marks at 80%. However, when working on an upcoming change to try to detect timing marks with fold lines through them, I made some of the logic for finding the timing mark shape a bit more relaxed. This meant that shapes could have bigger nubs/spurs in the imprinting case, which ended up causing mark scores below the 80% threshold.

To build in more buffer for those upcoming changes, I wanted to improve the scoring to remove the ~9% variability due to binarization. To tackle this, I added a 1px vertical "jiggle" - moving the expected timing mark rectangle 1px up and down and taking the best score found. This approach has worked well for bubble scoring and seems to work well here too.

After making this change, I was actually able to up the corner mark score threshold to 90%.

Screenshot 2025-11-13 at 4 32 22 PM

## Demo Video or Screenshot
Included above

## Testing Plan
- Existing automated tests pass
- Ran benchmarks, saw no measurable change
- Tested against TRR corpus with debug-timing-marks binary. All expected failures from `main` are still caught. There were 5 new unexpected failures, but I manually confirmed that they are reasonable failures (4 were ballots with pen squiggles through the timing marks, 1 had very high skew).

## Checklist

### DIFF
--- a/apps/scan/backend/test/helpers/pdi_helpers.ts
+++ b/apps/scan/backend/test/helpers/pdi_helpers.ts
@@ -219,7 +219,7 @@ export const ballotImages = {
       Uint8Array.from(await readFile(vxFamousNamesFixtures.blankBallotPath))
     ),
   completeHmpbInvalidScale: async () => {
-    const scale = 0.9;
+    const scale = 0.95;
     const sheet = await pdfToImageSheet(
       Uint8Array.from(await readFile(vxFamousNamesFixtures.blankBallotPath)),
       {


eventualbuddha (commented):
Retroactive approval ✔️

Issue #7426, opened 2025-10-24
VxPollBook: allow single-precinct elections to have the precinct not end in "1"


arsalansufi:
We introduced some validation logic late in the last QA period that prevents a user from loading a malformed poll book package where the wards associated with each voter don’t map to precincts in the election definition. Unfortunately, we didn’t special case single precinct (e.g. town) elections, so they only work if the precinct title ends in “1”. Our testing didn’t capture this because the title of the precinct in our town fixtures, for both automated and manual testing, is “Precinct 1".

This issue has no impact on Claremont’s election, as they are multi-precinct. But it’s a bug we will need to fix for the spring.

It does have an impact on the NHTCTA demo election, because the election definition there can’t be used unmodified on VxPollBook (need to add “1” to the precinct name) so we can’t share cards between VxPollBook and VxSuite. That’s OK, because we would have to re-image anyway to get the poll books to be prod. We’ll just use separate sets of cards.

Issue automatically created from Slack feedback reported by drew: 
https://votingworks.slack.com/archives/C098ZSZ7719/p1761340466532999

PR #7526, opened 2025-11-14
VxPollBook: update multi-precinct import logic

### LINKED ISSUE #7456: VxPollBook: trim zeros in ward identifiers when importing voters and streetNames files
the precinct validation logic bites again! for claremont, my initial poll book package failed on import because in their voter and street name files, the ward is “02” rather than “2" and we’re not trimming those zeros, so we just failed. luckily, I can just massage those files, doing a careful find and replace, without touching the election definition and messing up card compatibility

https://votingworks.slack.com/archives/C098ZSZ7719/p1761592515438229

adghayes:
## Overview

Closes #7456 and #7426. Handles district / ward numbers with leading zeros. Sets up different code paths for single and multi-precinct elections.

## Checklist

### DIFF
--- a/apps/pollbook/backend/src/pollbook_package.test.ts
+++ b/apps/pollbook/backend/src/pollbook_package.test.ts
@@ -81,7 +81,7 @@ describe('parseVotersFromCsvString', () => {
     ).not.toThrow();
   });
 
-  test('throws when ward or district id has no match election precincts', () => {
+  test('throws when ward or district id has no match in election precincts', () => {
     const csvString = `Voter ID,First Name,Last Name,Postal Zip5,Ward
 123,John,Doe,12345,01
 456,Jane,Smith,67890,99`;


apps/pollbook/backend/src/pollbook_package.ts
```diff
@@ -63,9 +63,32 @@ type PollbookPackageParseError =
...
+    }
+  | {
+      type: 'multi-precinct';
+      precinctIds: Record<string, string>;
+    };
+
+function trimZerosFromExternalPrecinctId(externalId: string): string {
+  const externalWardIdParseIntResult = safeParseInt(externalId);
+  if (externalWardIdParseIntResult.isErr()) {
+    throw new Error(`Non-numeric ward or district: ${externalId}`);
```
kshen0: Are ward or district IDs ever alphanumeric?

adghayes: in NH, they are always just numeric. this already doesn't apply to anywhere outside of NH, so I think it's OK to further tune it to NH

Issue #7466, opened 2025-11-05
Post-election investigation: Errors requiring rescans


arsalansufi:
> Ballot Scan Failed: There was a problem scanning your ballot. Please scan it again.

Wards: [Keene Ward 3](https://votingworks.slack.com/archives/C09KF5P7D41/p1762271880751329), [Portsmouth Ward 2](https://votingworks.slack.com/archives/C09KF5P7D41/p1762280754196589)

Keene's errors were likely the result of folds through timing marks, TBD for Portsmouth

arsalansufi:
Keene Ward 3 logs: https://drive.google.com/file/d/1qsGIhH-JA6yfc6MX0bg9VeHjOdT8fECH/view?usp=sharing

PR #7534, opened 2025-11-18
ballot-interpreter: Allow gaps in timing marks due to folds

jonahkagan:
## Overview

Task: https://github.com/votingworks/vxsuite/issues/7466

When a fold goes through a timing mark, it disrupts the black ink and exposes some of the white paper. This causes the timing mark detection algorithm to fail.

Screenshot 2025-11-10 at 11 53 11 AM

To identify potential timing mark shapes, we look at each column of pixels (within a specified inset from the edge of the image) and find contiguous vertical strips of black pixels. We then join adjacent vertical slices to form rectangles.

The problem here is that some of the vertical strips are interrupted by light pixels from the fold, so they end up split into two short strips. These short strips are too short to be included in the rectangles, so they are filtered out.

In the debug image below, we see a few different colored rectangles within a single timing mark. Those are the groups of adjacent vertical strips that didn’t get filtered out. Since they weren’t all adjacent to each other (i.e. there are some interrupted strips in between them), they end up being seen as distinct rectangles rather than being joined into one timing-mark-shaped rectangle.

Screenshot 2025-11-10 at 11 52 21 AM

To fix this, I modified the code that finds black strips to allow for a small gap within each black strip (up to 1/3 of the expected height of a timing mark).

[Investigation thread](https://votingworks.slack.com/archives/CEL6D3GAD/p1762803409415999)

## Demo Video or Screenshot

Screenshot 2025-11-10 at 12 25 36 PM (1)

## Testing Plan
- Added regression unit test
- Ran benchmarks and verified no performance impact
- Ran debug-timing-marks binary against TRR corpus and verified no change on top of #7524 

## Checklist


nikhilb4a (approved):
Nice!

Issue #7519, opened 2025-11-13
VxPrint: polish and wire up `Print` screen


adghayes:

adghayes:
closing this as tracked by #7520 and #7591

PR #7538, opened 2025-11-19
VxPrint: Wire up Print tab - Use election's party and language, find ballotStyleId, track print counts

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/7519

This PR continues wiring up the Print tab, setting up for the Reports tab.
1. Connect Print tab to Election's party and language data
2. Update PrintBallot logic to use the ballot style props selected on the frontend to find the proper ballot style
3. Adds logic to track print counts on print, to be used in Reports

## Demo Video or Screenshot

https://github.com/user-attachments/assets/b9c911e8-b64e-4db7-8dbd-8a7fb1a12d6c

## Checklist

### DIFF
--- a/apps/print/backend/src/app.ts
+++ b/apps/print/backend/src/app.ts
@@ -224,15 +224,25 @@ export function buildApi(ctx: AppContext) {
         count: input.copies,
       });
 
-      // const ballot = store.getBallot(input.ballotPrintId);
+      // const ballot = store.getBallot({
+      //   ballotStyleId,
+      //   precinctId: input.precinctId,
+      //   ballotType: input.ballotType,
+      // });
       // if (!ballot || !ballot.encodedBallot) {
       //   await logger.logAsCurrentRole(LogEventId.PrinterPrintRequest, {
-      //     message: `No ballot found with id ${input.ballotPrintId}`,
+      //     message: 'No ballot found',
+      //     ballotProps: JSON.stringify({
+      //       precinctId: input.precinctId,
+      //       splitId: input.splitId,
+      //       partyId: input.partyId,
+      //       languageCode: input.languageCode,
+      //       ballotType: input.ballotType,
+      //     }),
       //     disposition: 'failure',
       //   });
       //   return;
       // }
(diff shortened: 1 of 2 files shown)


apps/print/backend/src/util/ballot_styles.ts
```diff
@@ -0,0 +1,77 @@
...
+import { getBallotStyleGroupsForPrecinctOrSplit } from '@votingworks/utils';
+
+interface FindBallotStyleArgs {
+  precinctId: PrecinctId;
+  splitId?: Id;
+  languageCode: LanguageCode;
+  partyId?: Id;
+}
+
+export function findBallotStyleId(
```
nikhilb4a: The logic here roughly follows VxMark's ballot style selection [logic](https://github.com/votingworks/vxsuite/blob/jonah/refine-mark-finding/libs/mark-flow-ui/src/components/poll_worker/ballot_style_select.tsx#L51), although we must also consider language and party

apps/print/backend/src/app.ts
```diff
@@ -174,31 +179,73 @@ export function buildApi(ctx: AppContext) {
...
+      //   await logger.logAsCurrentRole(LogEventId.PrinterPrintRequest, {
+      //     message: `No ballot found with id ${input.ballotPrintId}`,
+      //     disposition: 'failure',
+      //   });
+      //   return;
+      // }
+
+      // await printer.print({
+      //   data: Buffer.from(ballot.encodedBallot, 'base64'),
+      // });
```
kshen0: Are you keeping these commented out for the above TODO?

nikhilb4a: Yeah, I was thinking it might be helpful for you if/when you are needing to test printing ballots. That being said what I left previously wasn't super helpful, but I added one more [commit](https://github.com/votingworks/vxsuite/pull/7538/commits/59529a6a786c4123eb3847129c45ae08156dba1a) that makes it so the commented out logic will work when uncommented

PR #7542, opened 2025-11-19
VxPrint: Add Report tab with ballot print counts

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/7521

This PR adds the first pass of the Report tab. UI polish still to come. 

On the backend, this PR updated the `BallotPrintCount` type to include the ballot props needed to fill out the Report table.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/bf90e609-a8bd-4d71-b1d2-27c17fcd6fd9

## Testing Plan

Manual for now, automated tests to be added later

## Checklist

### DIFF
--- a/apps/print/backend/src/util/ballot_styles.ts
+++ b/apps/print/backend/src/util/ballot_styles.ts
@@ -126,7 +126,7 @@ export function addBallotsPropsToPrintCountRow(
     assert(ballotStyle.partyId !== undefined);
     const party = election.parties.find((p) => p.id === ballotStyle.partyId);
     assert(party, `No party found with id ${ballotStyle.partyId}`);
-    ({ name: partyName } = party);
+    partyName = party.name;
   }
 
   return {
--- a/apps/print/frontend/src/screens/report_screen.tsx
+++ b/apps/print/frontend/src/screens/report_screen.tsx
@@ -13,7 +13,7 @@ import {
 } from '@votingworks/ui';
 import { hasSplits } from '@votingworks/types';
 import { format } from '@votingworks/utils';
-import { assertDefined } from '@votingworks/basics';
+import { assert } from '@votingworks/basics';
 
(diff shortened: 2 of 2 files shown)


apps/print/frontend/src/utils.ts
```diff
@@ -1,7 +1,59 @@
+import type { BallotPrintCount } from '@votingworks/print-backend';
 import { Election, LanguageCode } from '@votingworks/types';
 
+function sortLanguages(
```
nikhilb4a: Added this in so language order stays consistent. I noticed this because an election I was testing with had Spanish first in the election definition. It looked a bit od to have the default value of English not be the first radio button, so I added this ordering, following the order in VxDesign.

apps/print/backend/src/util/ballot_styles.ts
```diff
@@ -75,3 +85,53 @@ export function findBallotStyleId(
...
+  const precinctAndSplitsForBallotStyle = getPrecinctsAndSplitsForBallotStyle({
+    election,
+    ballotStyle,
+  });
+
+  const matchingPrecinctOrSplit = assertDefined(
+    precinctAndSplitsForBallotStyle.find(
+      (ps) =>
+        ps.precinct.id === precinct.id &&
+        (precinctHasSplits ? ps.split !== undefined : ps.split === undefined)
```
nikhilb4a: I could use some help sanity checking the logic for the Split case –

So first we get all precinctsAndSplits that use this ballot style. Then, if this precinctHasSplits, we find the precinct/Split that has a defined split, without needing to check the splitId (which I don't have). The assumption here is that only one split per precinct can have a specific ballot style, otherwise they would have the same districtIds which is not allowed by VxDesign ([link](https://github.com/votingworks/vxsuite/blob/main/apps/design/frontend/src/geography_screen.tsx#L621)). Does that sound right to you?

Issue #7531, opened 2025-11-17
Don't let a corrupted image completely block adjudication/tabulation


arsalansufi:
A bandaid in lieu of a larger root cause fix for https://github.com/votingworks/vxsuite/issues/7391

PR #7546, opened 2025-11-21
Don't crash on corrupted images during adjudication, still allow adjudication to continue

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7531

During a demo a couple of months back, the VxAdmin adjudication UI repeatedly crashed (something went wrong) upon reaching a specific ballot, blocking us from completing adjudication and thereby tabulation.

In our investigation, we found that the image of that ballot was corrupted (0 bytes). This further seemed to be the result of a power cord getting pulled right as that ballot was accepted. Note that the issue was not that the copy on the USB drive was corrupted but that the original copy on the machine's disk was never able to fully save. So even a re-export of CVRs to the USB drive wouldn't help us. (And if the issue had just been at the USB drive level, we would've detected that a CVR re-sync was necessary and self-corrected.)

The deeper fix here is to sequence the original image write to disk such that, if the system fails mid-write, we can recover on reboot. This would likely mirror the logic that we have in place to keep the USB drive in sync with the machine's drive. But this is a non-trivial change for what should be an extremely rare case (never before seen in an actual election).

For the time being and for v4.0.4, I've instead opted to make us resilient to corrupted images during adjudication. You can now still proceed through adjudication even if an image is corrupted. Yes this might mean that the corrupted write-in can't be correctly tabulated (you'd have to tabulate it as "unknown"), but at least tabulation as a whole can be completed.

In our most recent election, one town had the power go out, and their UPS didn't kick in. Thankfully, we didn't encounter this edge case, but the report made me want to introduce a quick safeguard for next year's elections.

### Before

https://github.com/user-attachments/assets/0ec638df-97cd-4d56-b85c-e18a034eff4e

### After

unable-to-load-image

## Checklist

### DIFF
--- a/apps/admin/frontend/src/screens/contest_adjudication_screen.test.tsx
+++ b/apps/admin/frontend/src/screens/contest_adjudication_screen.test.tsx
@@ -1259,6 +1259,74 @@ describe('ballot image viewer', () => {
     });
   });
 
+  test('hmp ballot with corrupted image', async () => {
+    const contestId = 'zoo-council-mammal';
+    const cvrIds = ['id-174', 'id-175'];
+    const cvrId = cvrIds[0];
+    const cvrId2 = cvrIds[1];
+    const writeInRecords: WriteInRecord[] = [
+      {
+        status: 'pending',
+        id: '1',
+        cvrId,
+        contestId,
+        electionId,
+        optionId: 'write-in-0',
+      },
+    ];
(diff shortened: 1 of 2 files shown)


arsalansufi:
On second thought, I can find someone else to review next week to ease your review burden just a bit!

kofi-q (approved):
Wow, that's a fun edge case! 😅

PR #7547, opened 2025-11-21
VxDesign: Add TTS plain text editor shared component

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/7266

Adding the plain text editing version of the audio editor (vs the word-based phonetic editor, which will come later). Currently unused, but will be eventually wrapped in a component that takes care of switching between plain text and phonetic editing and displaying a preview of the original string (full context shown in screengrabs below.

### TODO
- Currently assumes the happy path - will need to add any appropriate error handling as the feature gets fleshed out.

### In Context:
vxdesign-tts-test-editor

### Basic Edit States:

https://github.com/user-attachments/assets/3d829c6a-def0-4a87-a4cf-c3709d62d87a

### Layout Edge Cases:

https://github.com/user-attachments/assets/91d86d2c-68d6-4e5a-b846-407e74593000

## Testing Plan
- Unit tests for basic interaction/state changes
- Manual functional/styling spot checks on Chrome, FF, Safari

## Checklist

### DIFF
--- a/apps/design/frontend/src/ballot_audio/tts_text_editor.tsx
+++ b/apps/design/frontend/src/ballot_audio/tts_text_editor.tsx
@@ -1,0 +1,292 @@
+import React from 'react';
+import styled from 'styled-components';
+
+import { assertDefined } from '@votingworks/basics';
+import { TtsEdit } from '@votingworks/types';
+import { Icons, Button, DesktopPalette, Caption, Font } from '@votingworks/ui';
+
+import * as api from '../api';
+import { cssThemedScrollbars } from '../scrollbars';
+
+const Container = styled.div`
+  display: grid;
+  max-height: 100%;
+  grid-template-rows: min-content 1fr;
+  overflow: hidden;
+`;
+
+const Header = styled(Font)`
+  box-shadow: 0 0.25rem 0.5rem rgba(255, 255, 255, 75%);
+  display: block;
+  padding: 0 0 0.5rem;
+  margin: 0;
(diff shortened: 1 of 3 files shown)


apps/design/frontend/src/ballot_audio/tts_text_editor.tsx
```diff
@@ -0,0 +1,292 @@
...
+  const savedEdit = api.ttsEditsGet.useQuery({
+    orgId,
+    languageCode,
+    original,
+  });
+
+  // [TODO] Render a loading state instead.
+  if (!savedEdit.isSuccess) return null;
+
+  return <EditorImpl {...props} savedEdit={savedEdit.data} />;
```
kofi-q: Kind of a weird split, but wanted to only fetch the audio data once the saved edit (if any) has been fetched, so doing this to work around the no-conditional-hooks restriction

nikhilb4a: Might not be better, but I think you could consider exposing the `enabled` prop on the audio fetch query, keeping the query disabled until the savedEdit query is complete. I used it in a somewhat comparable way [here](https://github.com/votingworks/vxsuite/blob/main/apps/admin/frontend/src/api.ts#L393) when I needed to [sequence queries for contest adjudication](https://github.com/votingworks/vxsuite/blob/main/apps/admin/frontend/src/screens/contest_adjudication_screen.tsx#L290)

Issue #7527, opened 2025-11-14
HWTA and main app auto-restart


arsalansufi:
https://votingworks.slack.com/archives/C088CE8UXJ9/p1763044497325019

* Auto-restart the HWTA on something went wrong after 10 seconds
* Auto-restart the main app on something went wrong after 10 minutes (as SLI needs for the HWTA and main app to have matched behavior, but we also prefer a power cycle as the means of recovery)

arsalansufi:
Noting that we didn't end up adding a lower level restart (as in OS/kernel level). We opted to just call `sudo reboot` from the app (by way of kiosk-browser).

Once we do more testing with that change in place, we can see whether something more is necessary to get through HW testing.

PR #7548, opened 2025-11-24
Make various updates to error screens for cert

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7308
Closes https://github.com/votingworks/vxsuite/issues/7527

This PR makes various updates to error screens for cert, per discussions/negotiations with SLI.

First, it adds a note to unknown/unexpected error screens to reach out to VotingWorks if a restart doesn't remedy.

Second, it adds an auto-restart after 10 seconds to the unknown/unexpected error screens for the VxMarkScan and VxScan hardware test apps. We want to hedge our bets for hardware testing, where the system needs to be able to recover from issues without human intervention. Many issues, like crashes in response to shocks, can hypothetically be recovered from with a restart.

Note that SLI has generally asked that we maintain parity between recovery mechanisms in the hardware test apps and the main apps, so we've also added an auto-restart to the main VxMarkScan and VxScan apps but with a longer time delay (10 minutes). Ultimately, a software restart isn't as likely to remedy an issue as a full power down and power up (e.g., unplug and replug) because a software restart doesn't power cycle internal components. The longer delay will make it such that in most cases the user will still manually power down and power up rather than relying on the auto-restart.

## Demo Video or Screenshot

_VxSuite something went wrong screens_

<table>
laptops
<br />
precinct-scanner
<br />
bmd
</table>

_VxScan and VxMarkScan state machine error screens_

<table>
vxscan-state-machine
<br />
vxmarkscan-state-machine
</table>

_HWTA something went wrong screen_

<table>
hwta
</table>

_Card reader not detected screen_

<table>
card-reader-not-detected
</table>

_VxPollBook something went wrong screen_

<table>
poll-book
</table>

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

## Checklist


libs/ui/src/setup_card_reader_page.tsx
```diff
@@ -1,23 +1,15 @@
...
+import { Caption, H1, P } from './typography';
 
+export function SetupCardReaderPage(): JSX.Element {
   return (
     <Screen>
       <Main centerChild>
         <H1>Card Reader Not Detected</H1>
-        <P>{connectMessage}</P>
+        <P>{ERROR_SCREEN_MESSAGES.RESTART}</P>
+        <Caption>{ERROR_SCREEN_MESSAGES.REACH_OUT}</Caption>
```
arsalansufi: SLI specifically flagged this screen as a screen that should include a reach-out message like the unknown error screens since there's no other action the user can take other than rebooting.

apps/pollbook/frontend/src/error_screen.tsx
```diff
@@ -16,6 +22,7 @@ export function ErrorScreen(): JSX.Element {
           <Icons.Danger />
         </FullScreenIconWrapper>
         <H1>Something went wrong</H1>
+        <P>{ERROR_SCREEN_MESSAGES.RESTART}</P>
```
arsalansufi: Didn't add the note to reach out to VotingWorks to VxPollBook since it's not bound by the same certification process and we personally didn't want to have to add this additional copy to these screens

PR #7555, opened 2025-11-25
VxScan: remove double confirmation on adjudication warnings

### LINKED ISSUE #5670: VxScan: remove double confirmation on adjudication warnings
after watching voters in MS cast ballots and being confused by 1) the undervote warning and then 2) the undervote cast as-is double confirm, I want to raise the question — do we think the double confirmation for casting-as-is is necessary/important for second chance voting warnings?

we’ve had it this way in the product for a long time and haven’t questioned it. but we also haven’t had voters experience casting-as-is functions in NH (due to the NH overvote rule) and my experience watching voters this week in MS makes me feel like it goes against voter experience expectations. it also just makes the process of casting in these cases slower.

!subteam^S05828MAZPV in particular for thoughts.

Issue automatically created from Slack feedback reported by roe: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1731101530044379?thread_ts=1731101530.044379&cid=CJU9MSC6S

adghayes:
## Overview

Closes #5670. Previously, when casing an overvoted, undervoted, or blank ballot flagged by the system settings, the user would have to tap "Cast Ballot" and then, in a confirmation modal, tap "Cast Ballot" again. Now they only have to do it once.

### Before

https://github.com/user-attachments/assets/5d23709a-6c71-4006-ae3c-a542a1262bee

### After

https://github.com/user-attachments/assets/dcc1bbbf-ae76-427f-99d1-c9cce68a1a9a

## Testing Plan

Updated automated tests and tried in dev.

## Checklist

### DIFF
--- a/apps/scan/frontend/src/screens/scan_warning_screen.test.tsx
+++ b/apps/scan/frontend/src/screens/scan_warning_screen.test.tsx
@@ -96,7 +96,11 @@ test('overvote', async () => {
     {}
   );
 
-  userEvent.click(screen.getByRole('button', { name: 'Cast Ballot' }));
+  const castBallotButton = screen.getButton('Cast Ballot');
+  const returnBallotButton = screen.getButton('Return Ballot');
+  userEvent.click(castBallotButton);
+  expect(castBallotButton).toBeDisabled();
+  expect(returnBallotButton).toBeDisabled();
 });
 
 test('overvote when casting overvotes is disallowed', async () => {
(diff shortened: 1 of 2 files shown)


adghayes:
I noticed that in the old version, the second "Cast Ballot" button had a check icon, while the main button does not. @kofi-q what do you think of the icon situation?

kofi-q (approved):
LGTM overall - had a question in Slack about whether or not we'd want to preserve the modal for the edge case where the contests aren't listed on the main warning screen, WRT VVSG, but leaning toward that being a non-issue.

No strong feelings on the checkmark icon - more of a nice-to-have, but there's also a chance it won't fit at the XL text size (IIRC there were a few spots where we had to drop icons because of that)

Issue #7513, opened 2025-11-13
VxPrint: add printer status icon to top bar


adghayes:
In VxPollBook, we have a printer icon in the top bar. When it's disconnected, there's a little warning icon. When it's connected, there's no warning icon.

In VxPrint, there's richer status information available from the printer that we'd like to convey:

- printer is completely disconnected
- printer is connected but "stopped" e.g. jam, no paper
- printer is connected and working but toner is low
- printer is all good

On design, brief discussion [here](https://votingworks.slack.com/archives/C09GQB0QB4N/p1763062959862189). The current idea is to put the printer icon, status icon, and short status text all in the top bar. For example:

- printer is completely disconnected
  - icon: disabled, red
  - text:  Not Connected
- printer is stopped
  - icon: danger, red
  - text: Jammed / No Paper / Other
- printer is low on toner
  - icon: warning, yellow
  - text: Toner Low
- printer ready to print or printing
  - icon: done, green
  - text: None

Because the language is no longer icon vs. no icon, for parallelism we probably ought to add an icon for the connected state.

adghayes:
Closed by #7545 and #7565.

PR #7565, opened 2025-12-02
VxPrint: more detailed print status icon

kshen0:
## Overview

https://github.com/votingworks/vxsuite/issues/7513

Open to design nits. `<P>` felt too large for the context so I used `<Caption>`, but both are vertically centered ie. not flush with the bottom of the icon due to line height. We could finagle it if desired.

Adds printer rich status icons and text.

## Demo Video or Screenshot
Screenshot 2025-12-01 at 5 37 44 PM
Screenshot 2025-12-01 at 5 37 51 PM
Screenshot 2025-12-01 at 5 38 01 PM
Screenshot 2025-12-01 at 5 43 55 PM

Jam state (trust that I am jamming and unjamming the printer when the statuses change):
https://github.com/user-attachments/assets/08ff2a92-d61e-4881-b154-b2d58fd33998

## Testing Plan
- automated tests todo

## Checklist

### DIFF
--- a/apps/print/frontend/src/components/toolbar.tsx
+++ b/apps/print/frontend/src/components/toolbar.tsx
@@ -144,7 +144,7 @@ function PrinterStatus({ status }: { status: PrinterStatusType }) {
     );
   }
 
-  return <BasePrinterStatus icon={<Icons.Done color="success" />} />;
+  return <BasePrinterStatus />;
 }
 
 function UsbStatus({ status }: { status: UsbDriveStatus }) {


apps/print/frontend/src/components/toolbar.tsx
```diff
@@ -26,15 +30,125 @@ function useCurrentDate(): Date {
...
+    if (
+      richStatus.stateReasons.find(
+        (reason) =>
+          reason === 'spool-area-full' || reason === 'spool-area-full-report'
+      )
+    ) {
+      return (
+        <BasePrinterStatus
+          icon={<Icons.Danger color="danger" />}
+          labelText="Output Full"
```
nikhilb4a: I'm new to the printer world but wondering if the term "Output Full" is intuitive? Though I'm not sure I can think of anything better. "Tray Full" comes to mind, but not sure that it's an improvement

adghayes:
+1 on moving the time to the right, I'm thinking it should probably be [Printer - USB - Battery - Time].

- The printer has the most information and jumps around the most, would be nice to put it first.
- Battery, I don't think we have here yet, but can we re-use from VxPollBook?
- USB, I'm not sure if it makes sense to warn people when it's disconnected anymore

Re: USB - on VxPollBook, we're trying to guide users toward having the USB inserted. In VxPrint, that doesn't really matter in the same way, so I think the _warning when no USB_ pattern is misplaced. @kshen0 @nikhilb4a do you have other ideas? We could:
a) use a less aggressive icon
b) move to a button instead of an icon that is either "Eject USB" or "No USB", like we have in VxAdmin, but smaller and in the toolbar.
c) no indication, leave it to the relevant modals

I think the move is (b), but I understand I'm increasing scope here (could be a follow-up PR).

apps/print/frontend/src/components/toolbar.tsx
```diff
@@ -26,15 +30,125 @@ function useCurrentDate(): Date {
...
+    }
+
+    // Encountered when a jam on the 1st of 2 pages resulted in a subtle jam entirely
+    // inside the printer. The printer screen readout was helpful in this case, but
+    // due to the vagueness of 'other-error' our user-facing error should be vague as well
+    if (richStatus.stateReasons.find((reason) => reason === 'other-error')) {
+      return (
+        <BasePrinterStatus
+          icon={<Icons.Danger color="danger" />}
+          labelText="See Printer Screen"
```
adghayes: nit: use "Display" rather than "Screen" so you don't think it's within the app

kshen0:
> Re: USB - on VxPollBook, we're trying to guide users toward having the USB inserted. In VxPrint, that doesn't really matter in the same way, so I think the _warning when no USB_ pattern is misplaced. @kshen0 @nikhilb4a do you have other ideas? We could: a) use a less aggressive icon b) move to a button instead of an icon that is either "Eject USB" or "No USB", like we have in VxAdmin, but smaller and in the toolbar. c) no indication, leave it to the relevant modals
> 
> I think the move is (b), but I understand I'm increasing scope here (could be a follow-up PR).

I think B is the best choice as well since the USB drive is only needed for initial configuration and not continuous export. The only drawback I can think of is introducing another pattern (eject from toolbar instead of main screen) which _might_ be confusing. But if we find that's the case, we can add it to a more traditional `Settings` screen or move it entirely.

apps/print/frontend/src/components/toolbar.tsx
```diff
@@ -26,15 +30,125 @@ function useCurrentDate(): Date {
...
+  ) {
+    return (
+      <BasePrinterStatus
+        icon={<Icons.Warning color="inverseWarning" />}
+        labelText="Low Toner"
+      />
+    );
+  }
+
+  return <BasePrinterStatus icon={<Icons.Done color="success" />} />;
```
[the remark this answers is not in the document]
kshen0: From the ticket

> printer ready to print or printing
> icon: done, green
> text: None

But I'm all for consistency. @adghayes what do you think?

PR #7575, opened 2025-12-03
VxPrint: Add Ballots Printed Report

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/7521

Adds a Ballots Printed Report that can be printed or exported. The format and code for the report follow the [Ballot Count Report](https://github.com/votingworks/vxsuite/blob/c8326e389e7f80be5f0413f64bab2a345871ea6f/libs/ui/src/reports/ballot_count_report.tsx#L54).

## Demo Video or Screenshot

Example report:
[ballots-printed-report__2025-12-02_18-06-03.pdf](https://github.com/user-attachments/files/23894125/ballots-printed-report__2025-12-02_18-06-03.pdf)

https://github.com/user-attachments/assets/c544fdf5-ea1e-418a-979a-74204014d259

## Checklist

### DIFF
--- a/apps/print/backend/src/reports/ballots_printed_report.ts
+++ b/apps/print/backend/src/reports/ballots_printed_report.ts
@@ -64,7 +64,7 @@ export async function printBallotsPrintedReport({
   try {
     await printer.print({ data });
     await logger.logAsCurrentRole(LogEventId.ElectionReportPrinted, {
-      message: `User printed a Ballots Printed Report.`,
+      message: 'User printed a Ballots Printed Report.',
       disposition: 'success',
     });
   } catch (error) {


libs/ui/src/reports/report_generated_metadata.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from '@votingworks/types';
 import { LabeledValue, ReportMetadata } from './report_header';
 
-export function AdminReportMetadata({
+export function ReportGeneratedMetadata({
```
nikhilb4a: The component had `Admin` in the name, but nothing `Admin` specific. Renaming to reuse as the header for the Ballots Printed Report

apps/print/frontend/src/utils.ts
```diff
@@ -25,35 +24,3 @@ export function getAvailableLanguages(election: Election): LanguageCode[] {
     .filter((lang) => lang !== undefined)
     .sort(sortLanguages);
 }
-
-// sortBallotPrintCounts sort order: totalCount, precinctOrSplitName, partyName, languageCode
-export function sortBallotPrintCounts(
```
nikhilb4a: Moved sorting to backend store function so ballot print counts are sorted consistently regardless of use (i.e. pdf report vs. ui)

apps/print/backend/src/reports/ballots_printed_report.ts
```diff
@@ -0,0 +1,110 @@
...
+  store,
+}: {
+  printer: Printer;
+  logger: Logger;
+  store: Store;
+}): Promise<void> {
+  const report = buildBallotsPrintedReport({ store });
+
+  try {
+    const data = (await renderToPdf({ document: report })).unsafeUnwrap();
```
kshen0: nit: you can use the `Result` from `await renderToPdf(...)` to check for an error and leave it outside the try/catch.

```
const renderResult = await renderToPdf({ document: report }));
if (renderResult.isErr()) {
  const error = renderResult.err();
  // log error
}
const data = renderResult.ok();
try {
  await printer.print({ data });
  ...
}
```

It's marginally clearer this way because an error in `renderToPdf` won't report as "Error in attempting to print ballots printed report".

PR #7583, opened 2025-12-04
VxPrint: Allow users to Print All Ballot Styles

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/7519

Adds a utility for Election Managers to simplify printing all ballot styles, by language and ballot type. I've added an endpoint that counts the number of distinct ballots so we can indicate to the user how many ballots will be printed.

## Demo Video or Screenshot

https://github.com/user-attachments/assets/b7967a90-090a-44de-b49d-7937f2a31fe5

## Testing Plan

Manual for now, logs to verify

## Checklist

### DIFF
--- a/apps/print/backend/src/app.ts
+++ b/apps/print/backend/src/app.ts
@@ -246,11 +246,10 @@ export function buildApi(ctx: AppContext) {
         return;
       }
 
-      for (let i = 0; i < input.copies; i += 1) {
-        await printer.print({
-          data: Buffer.from(ballot.encodedBallot, 'base64'),
-        });
-      }
+      await printer.print({
+        data: Buffer.from(ballot.encodedBallot, 'base64'),
+        copies: input.copies,
+      });
 
       store.incrementBallotPrintCount({
         precinctId: input.precinctId,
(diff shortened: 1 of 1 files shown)


nikhilb4a:
@adghayes feel free to let me know any styling nits you have. I added a ballot style count to the final primary button so the EM has an idea of how many ballots will be printed.

apps/print/backend/src/app.ts
```diff
@@ -261,6 +273,54 @@ export function buildApi(ctx: AppContext) {
...
+        languageCode: input.languageCode,
+        ballotType: input.ballotType,
+      });
+
+      let totalPrintCount = 0;
+      for (const ballot of ballots) {
+        for (let i = 0; i < input.copiesPerStyle; i += 1) {
+          await printer.print({
+            data: Buffer.from(ballot.encodedBallot, 'base64'),
+          });
```
adghayes: @nikhilb4a I recommend, instead of calling print for every copy, passing the `copies` or equivalent to `printer.print`. it's better for the printer to manage fewer jobs, and the number here could balloon rapidly

apps/print/backend/src/app.ts
```diff
@@ -216,37 +226,38 @@ export function buildApi(ctx: AppContext) {
...
+          }),
+          disposition: 'failure',
+        });
+        return;
+      }
+
+      await printer.print({
+        data: Buffer.from(ballot.encodedBallot, 'base64'),
+        copies: input.copies,
+      });
```
kshen0: I wonder if in a future version we need to read the number of copies printed and update by that amount? If the print fails halfway through due to lack of paper or something else, this number will be incorrect.

nikhilb4a: That's a good point, will note that for discussion

adghayes: It's a problem that has come up in Warren in v2 - they print ballots but due to some malfunction, some of the ballots don't print or are invalid - and their counts end up wrong. The most common cause - lack of paper - doesn't usually mess up the count if paper is then inserted. But there are other things that might!

I believe the eng lift is quite high. The CUPS server, which is easy to communicate with, doesn't really track job status. We'd have to make IPP requests directly to the printer and track some kind of ID for each job.

I think doing that would be cool and tighten up the product, but I'm not sure at what level of product maturity we should actually tackle it.

PR #7604, opened 2025-12-04
ui: Add "option" ARIA role to SearchSelect

jonahkagan:
## Overview

This enables us to more easily select options in frontend tests using `getByRole('option')`.

## Demo Video or Screenshot
N/A

## Testing Plan
Updated unit tests
## Checklist


kofi-q (approved):
LGTM! Surprised these aren't annotated at the library level

Also, maybe a little pedantic, but would say that we shouldn't use testability as motivation for aria roles, since they serve such a specific purpose.

PR #7605, opened 2025-12-05
VxPrint: Allow user to eject usb from toolbar, show battery status

nikhilb4a:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7577

## Demo Video or Screenshot

**Updated after PR Feedback**

https://github.com/user-attachments/assets/685db0cb-0489-472b-9fd7-31fd9215d11d

**USB eject**

https://github.com/user-attachments/assets/8ca866fa-237a-41a5-88e1-810162fea1fb

**Battery percentage** 

Screenshot 2025-12-04 at 4 26 06 PM

## Testing Plan

Manual for now

## Checklist

### DIFF
--- a/apps/print/frontend/src/components/toolbar.tsx
+++ b/apps/print/frontend/src/components/toolbar.tsx
@@ -29,6 +29,11 @@ const ToolbarButton = styled(Button)`
   padding: 0.25rem 0.75rem;
 `;
 
+const StatusCaption = styled(Caption)`
+  font-size: 0.8rem;
+  font-weight: 500;
+`;
+
 function useCurrentDate(): Date {
   const [currentDate, setCurrentDate] = useState(new Date());
 
@@ -53,7 +58,7 @@ function BasePrinterStatus({
     <Row style={{ gap: '0.25rem', justifyContent: 'end' }}>
       <Icons.Print color="inverse" />
       {icon}
-      {labelText && <Caption>{labelText}</Caption>}
+      {labelText && <StatusCaption>{labelText}</StatusCaption>}
     </Row>
   );
 }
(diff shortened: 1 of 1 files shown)


apps/print/frontend/src/components/toolbar.tsx
```diff
@@ -147,29 +159,61 @@ function PrinterStatus({ status }: { status: PrinterStatusType }) {
...
+      {status && !status.discharging && (
+        <Icons.Bolt style={{ fontSize: '0.8em' }} color="inverse" />
+      )}
+      {status && format.percent(status.level)}
+      {status && status.level < 0.25 && status.discharging && (
+        <Icons.Warning color="inverseWarning" />
+      )}
+    </Row>
+  );
+}
```
adghayes: No need for this VxPrint push but noting - we should probably extract this out for the sake of VxPrint and VxPollBook eventually.

apps/print/frontend/src/components/toolbar.tsx
```diff
@@ -147,29 +159,61 @@ function PrinterStatus({ status }: { status: PrinterStatusType }) {
...
+
+type ExtendedUsbDriveStatus = UsbDriveStatus['status'] | 'ejecting';
+const buttonIconAndText: Record<ExtendedUsbDriveStatus, [IconName, string]> = {
+  no_drive: ['Disabled', 'No USB'],
+  error: ['Disabled', 'No USB'],
+  mounted: ['Eject', 'Eject USB'],
+  ejecting: ['Eject', 'Ejecting...'],
+  ejected: ['Disabled', 'No USB'],
```
adghayes: This `ejected` string should be `USB Ejected` rather than `No USB`. To be clear to the user why the inserted USB drive is not usable. And to match behavior of the other `usbcontroller` button.

adghayes (approved):
See comment on updating the text when ejected, but otherwise LGTM.

PR #7624, opened 2025-12-05
VxPrint: Sort using fields from ballotPrintCount

nikhilb4a:
## Overview

Follow up to https://github.com/votingworks/vxsuite/pull/7623/files

Realized that approach takes print counts into ordering. Now, we solely use precinct/party names.

## Testing Plan

Manually tested

## Checklist

### DIFF
--- a/apps/print/backend/src/app.ts
+++ b/apps/print/backend/src/app.ts
@@ -339,30 +339,14 @@ export function buildApi(ctx: AppContext) {
       const isTestMode = store.getTestMode();
       const ballotMode = isTestMode ? 'test' : 'official';
 
-      const ballotPrintCounts = store.getBallotPrintCounts({
-        ballotMode,
-      });
-
-      const ballots = store
-        .getBallots({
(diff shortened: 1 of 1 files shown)


apps/print/backend/src/app.ts
```diff
@@ -349,20 +349,31 @@ export function buildApi(ctx: AppContext) {
...
           );
-          const indexB = ballotPrintCounts.findIndex(
+          const printCountB = ballotPrintCounts.find(
             (count) =>
               count.ballotStyleId === b.ballotStyleId &&
               count.precinctId === b.precinctId
           );
-          return indexA - indexB;
+          assert(printCountA);
+          assert(printCountB);
```
arsalansufi: Any chance a print count might not be populated? Say because a given ballot style hasn't been printed yet?

nikhilb4a: Good q, no it will have a print count of 0!

arsalansufi (approved):

Issue #7653, opened 2025-12-10
VxPrint: code cleanup


nikhilb4a:
1. Review location of `ScreenWrapper`
2. Any remaining Todos

PR #7701, opened 2025-12-15
VxPrint: Close remaining todos

nikhilb4a:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7653

Each commit is one todo:
1. Update comment reg. single language ballot style constraint
2. Update main() to be sync
3. Extract `TestModeCallout` from central-scan and print to libs/ui. Use in design. Remove `themeOverride` prop that was previously only used in design use case. 

## Demo Video or Screenshot

For 3 - 

VxDesign Live Reports before:
Screenshot 2025-12-15 at 2 17 52 PM

VxDesign Live Reports after:
Screenshot 2025-12-15 at 2 17 26 PM

VxPrint, unchanged:
Screenshot 2025-12-15 at 2 25 53 PM

VxCentralScan, unchanged:
Screenshot 2025-12-15 at 2 16 18 PM

## Checklist

### DIFF
--- a/libs/ui/src/test_mode.tsx
+++ b/libs/ui/src/test_mode.tsx
@@ -71,13 +71,13 @@ function TestModeCalloutDesktop({
 }
 
 export function TestModeCallout({
-  forDesktop,
+  viewMode,
   style,
 }: {
-  forDesktop?: boolean;
+  viewMode: 'touch' | 'desktop';
   style?: React.CSSProperties;
 }): JSX.Element {
-  if (forDesktop) {
+  if (viewMode === 'desktop') {
     return <TestModeCalloutDesktop style={style} />;
   }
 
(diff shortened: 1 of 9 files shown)


libs/ui/src/test_mode.tsx
```diff
@@ -44,8 +41,45 @@ export function TestModeCallout({
...
+      <Icons.Warning color="warning" /> Test Ballot Mode
+    </TestModeCardDesktop>
+  );
+}
+
+export function TestModeCallout({
+  forDesktop,
+  style,
+}: {
+  forDesktop?: boolean;
```
kshen0: What do you think about making this an enum/string literal union type instead of boolean (eg. `viewMode: 'desktop' | 'tablet'`) to be explicit about what the non-desktop alternative is?

PR #7707, opened 2025-12-16
HWTA/VxMark: Copy VxScan approach for System vs Headphone Audio Control

carolinemodic:
## Overview
Abstracts framework for managing audio player to use the system or usb output from VxScan to libs/backend and adds it to VxMark. Uses system audio for startup chime and barcode success chime. Added a feature to the HWTA to alternate sounds between the two outputs and add toggles in the UI for each. 

When testing this I found a bug with the barcode scanner not exiting properly in the HWTA that I fixed as well. 

## Demo Video or Screenshot
N/A

## Testing Plan
Tested on VxDev using HWTA and ensuring in the app that the barcode sound comes through the system speakers and the screen reader text is through the headphones.
I plan to make a QA dev image to test this in an image environment after merging. 

## Checklist

### DIFF
--- a/libs/backend/src/audio/get_audio_info_with_retry.test.ts
+++ b/libs/backend/src/audio/get_audio_info_with_retry.test.ts
@@ -5,6 +5,12 @@ import * as audio from '../system_call/get_audio_info';
 
 vi.mock('../system_call/get_audio_info');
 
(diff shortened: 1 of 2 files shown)


kofi-q (approved):
LGTM!

Issue #7666, opened 2025-12-11
VxPrint: test coverage


nikhilb4a:
1. BallotsPrintedReport in libs/ui - Done
2. print/backend - Done
3. print/frontend

PR #7742, opened 2026-01-06
VxPrint: Add backend test coverage

nikhilb4a:
## Overview

https://github.com/votingworks/vxsuite/issues/7666

Full test coverage for print/backend. There were two branches that I was struggling to get coverage for, but they seemed inconsequential. I added tests for most error branches, but added `istanbul ignores` for certain hard to reach branches that were more app setup and not `VxPrint` logic, and pretty much only where I saw other apps also ignored as precedent.

Found Claude Opus 4.5 to be the most helpful test writing assistant. Also tried Gemini 3 and ChatGpt 5.2 with less success

## Demo Video or Screenshot

Screenshot 2026-01-05 at 4 50 05 PM

## Testing Plan

Spider man meme

## Checklist


apps/print/backend/src/util/auth.ts
```diff
@@ -1,12 +1,41 @@
...
+import {
+  isFeatureFlagEnabled,
+  BooleanEnvironmentVariableName,
+  isIntegrationTest,
+} from '@votingworks/utils';
 import { Workspace } from './workspace';
 import { Store } from '../store';
 
+/* istanbul ignore next - @preserve */
+export function getDefaultAuth(logger: BaseLogger): DippedSmartCardAuth {
```
nikhilb4a: Extracted from `server.ts start()` to keep that function a bit more streamlined, following the [code structure](https://github.com/votingworks/vxsuite/blob/main/apps/mark/backend/src/server.ts#L37-L38) in `mark`.

apps/print/backend/src/app.ts
```diff
@@ -281,7 +275,7 @@ export function buildApi(ctx: AppContext) {
         ballotType: input.ballotType,
         ballotMode,
       });
-      if (!ballot || !ballot.encodedBallot) {
+      if (!ballot) {
```
nikhilb4a: Is it safe to drop the `!ballot.encodedBallot` check?

kshen0 (approved):

PR #7763, opened 2026-01-09
VxDesign: Add migration for backfilling top-level signatures on old NH elections (plus improvements to restore_db_backup.sh)

arsalansufi:
## Overview

As we prep for upcoming ballot production, some existing customers will likely be copying their town elections from last year to kickstart election def and ballot prep for this year. Since last year, we've changed where we persist town clerk signature images and captions. As noted in the migration:

> NH clerk signature images and captions were previously stored on precinct split objects. We've
> since introduced top-level fields for this data on election objects. The fields on the precinct
> split object still exist for override purposes however, e.g., when a school district ballot
> needs a school clerk signature rather than a town clerk signature.
>
> For elections using the NH template, ballot rendering errors when these fields are missing on
> the election object, hence the motivation for this migration, which moves this data from the
> relevant precinct split objects to the election objects, where possible.

<table>
error
</table>

This migration will make sure that these existing customers don't encounter rendering errors after copying last years' elections.

While I was at it, I also made an update to the `restore_db_backup.sh` script for local dev with prod data that should make it more useful out of the box.

## Testing Plan

- [x] Bootstrapped my local DB with a prod snapshot and confirmed that the migration is working as intended

## Checklist

### DIFF
--- a/apps/design/frontend/src/precincts_form.tsx
+++ b/apps/design/frontend/src/precincts_form.tsx
@@ -366,7 +366,7 @@ export function PrecinctForm(props: PrecinctFormProps): React.ReactNode {
 
                       {features.PRECINCT_SPLIT_CLERK_SIGNATURE_IMAGE_OVERRIDE && (
                         <div>
-                          <FieldName>Signature Image</FieldName>
+                          <FieldName>Signature Image Override</FieldName>
                           <SignatureImageInput
                             disabled={disabled}
                             value={split.clerkSignatureImage}
@@ -381,7 +381,7 @@ export function PrecinctForm(props: PrecinctFormProps): React.ReactNode {
                       )}
 
                       {features.PRECINCT_SPLIT_CLERK_SIGNATURE_CAPTION_OVERRIDE && (
-                        <InputGroup label="Signature Caption">
+                        <InputGroup label="Signature Caption Override">
                           <input
                             disabled={disabled}
                             type="text"
--- a/apps/design/frontend/src/precincts_screen.test.tsx
+++ b/apps/design/frontend/src/precincts_screen.test.tsx
@@ -289,8 +289,9 @@ test('editing a precinct - adding splits in NH', async () => {
     )
   );
 
-  const split3ClerkSignatureCaption =
-    within(split3Card).getByLabelText('Signature Caption');
+  const split3ClerkSignatureCaption = within(split3Card).getByLabelText(
+    'Signature Caption Override'
+  );
   expect(split3ClerkSignatureCaption).toHaveValue('');
   userEvent.type(
     split3ClerkSignatureCaption,


apps/design/backend/scripts/restore_db_backup.sh
```diff
@@ -1,5 +1,29 @@
...
+fi
+
+DB_BACKUP="${1}"
+
+SCRIPTS_DIRECTORY="$(dirname "${BASH_SOURCE[0]}")"
+"${SCRIPTS_DIRECTORY}/db_reset_dev.sh"
+
+PGPASSWORD=design pg_restore --verbose --clean --no-acl --no-owner -h localhost -U design -d design "${DB_BACKUP}"
+
+# Insert the dev user used for local development when AUTH_ENABLED=FALSE
```
arsalansufi: Previously, the app would hang, unable to find the dev user, after restoring from a prod snapshot. This should make it easier to get up and running with a prod snapshot.

kofi-q (approved):
🚢

apps/design/backend/scripts/restore_db_backup.sh
arsalansufi: Ah yeah so I thought about that but ran into some issues with order of operations.

When bootstrapping a local DB with a prod snapshot, the order of operations is: Run all migrations (which will set up schemas and create the dev user) and then restore from the prod DB snapshot (which will wipe the dev user since the dev user doesn't exist in the prod DB snapshot).

~I have been wondering whether we could change the order: Restore from the prod DB snapshot and then run all the migrations. But I ran into some errors with that approach. I don't recall the exact details but can try that again to properly record what the limitations were (if any).~ Actually makes sense that this wouldn't work since many of the migrations are, I assume, not idempotent.

arsalansufi: > Gotcha, makes sense. Now that I think about it more, I think pg_restore overwrites the database schema set up by the migrations anyway, so it's maybe unnecessary/misleading to run them first.

I _think_ I ran into an issue restoring from a prod snapshot without running the migrations first, but it's possible I had my wires crossed. Agreed conceptually that the pg_restore should render the migrations unnecessary. Will take a closer look and clean up as makes sense

jonahkagan: Thanks!

apps/design/backend/scripts/restore_db_backup.sh
```diff
@@ -1,5 +1,29 @@
...
+
+if [ "${#}" -ne 1 ]; then
+  echo "Usage: $0 <db-backup>" >&2
+  exit 1
+fi
+
+DB_BACKUP="${1}"
+
+SCRIPTS_DIRECTORY="$(dirname "${BASH_SOURCE[0]}")"
+"${SCRIPTS_DIRECTORY}/db_reset_dev.sh"
```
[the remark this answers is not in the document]
arsalansufi: Later reverted this specific addition in https://github.com/votingworks/vxsuite/pull/7768#discussion_r2684226465

Not seeing those occasional errors any more - must've gotten my wires crossed

apps/design/backend/scripts/restore_db_backup.sh
arsalansufi: ✅ https://github.com/votingworks/vxsuite/pull/7768

Issue #7393, opened 2025-10-16
MS: VxDesign Guardrails


adghayes:
For elections uploaded with the MS flow (SEMS conversion), we should prevent making changes that would invalidate the SEMS upload at the end of election night. This could mean avoiding adding or removing any entities or their relationships, while still allowing names to change. This includes:

- _prevent cloning election_ 
- _prevent adding or removing districts_
- _prevent adding or removing precincts or precinct splits_
- _prevent changing associated districts for each precinct_
- prevent adding or removing contests
- prevent changing district for a contest
- prevent adding or removing candidates 
- implied by above, prevent toggling between candidate to ballot measure type
- prevent changing party association of candidates or contests
- (Added) prevent editing candidate names
- (Added) prevent reordering contests
- (Added) prevent editing contest option labels
- (Added) prevent adding or removing parties

While there are some other things that MS shouldn't really have the option to do (turn off write-ins, number of seats), that shouldn't affect the interoperability and would be found in proofing. By limiting it to "restrict e-def structure changes" hopefully this will be reusable for other jurisdictions importing from outside systems.

nikhilb4a:
Also noting we are preventing Reordering Contests

adghayes:
Thank you for catching.

adghayes:
Confirmed!

PR #7765, opened 2026-01-12
VxDesign: Restrict editing election info for MS SEMS elections

nikhilb4a:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7393

Editable Election Info fields should be restricted to:
- title
- state
- jurisdiction

## Demo Video or Screenshot

https://github.com/user-attachments/assets/df1d8e50-901d-42ed-a7a6-f23edc2cd78f

## Testing Plan

Automated test added

## Checklist

### DIFF
--- a/apps/design/frontend/src/election_info_screen.test.tsx
+++ b/apps/design/frontend/src/election_info_screen.test.tsx
@@ -384,6 +384,49 @@ test('edit election disabled when ballots are finalized', async () => {
   expect(editButton).toBeDisabled();
 });
 
+test('elections with external source have limited editable fields', async () => {
+  const electionRecord = generalElectionRecord(jurisdiction.id);
+  electionRecord.ballotTemplateId = 'MsBallot';
+  const { election } = electionRecord;
+  const electionId = election.id;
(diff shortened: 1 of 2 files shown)


nikhilb4a:
Yeah good callout, I was considering that as well since this is one of the few cases of having a mixture of editable and non-editable fields on the same page.

That being said, the [contest form](https://github.com/votingworks/vxsuite/pull/7756) also allows only some fields to be editable, and doesn't have Tooltips. I thought it might end up being a lot of Tooltips otherwise. 

So I maybe lean towards relying on training to explain why some fields are disabled. I think @adghayes previously felt comfortable with that approach too, but will tag him in case he feels differently. Open to input, as adding the tooltips would be simple enough. Do you have a leaning?

nikhilb4a:
Throwing out that a low lift approach could be to add a callout next to the Election Info's "Edit" button indicating that elections loaded from external sources have restricted editing ability, just so there is some in product explanation.
Screenshot 2026-01-12 at 2 02 44 PM

PR #7774, opened 2026-01-13
Infer system settings type from zod schema

jonahkagan:
## Overview

Our general approach has been to define TS types and Zod schemas separately, using the `z.ZodType` annotation to enforce that the schema matches the TS type. However, this approach allows optional fields to be omitted from the Zod schema, which can lead to bugs since Zod will silently omit any parsed fields that are not listed in the schema ([example bug](https://votingworks.slack.com/archives/CEL6D3GAD/p1768261606977009)). This is a particularly high-risk issue at the input boundary for the certified system, since we can't fix this type of bug after releasing.

To guard against this, we switch to the approach recommended by Zod, which is to define the Zod schema as the source of truth and infer a TS type using `z.infer`. I actually don't see any real disadvantage to this approach in general except that it feels a little weird to define the data model mainly through a schema rather than as core language types.

## Demo Video or Screenshot
N/A

## Testing Plan
- Existing tests for correctness
- Confirmed that doc comments still show up in IDE features

## Checklist

### DIFF
--- a/libs/types/src/system_settings.ts
+++ b/libs/types/src/system_settings.ts
@@ -1,6 +1,6 @@
 import { Result } from '@votingworks/basics';
 import { z } from 'zod/v4';
-import { AdjudicationReason, AdjudicationReasonSchema } from './election';
+import { AdjudicationReasonSchema } from './election';
 import {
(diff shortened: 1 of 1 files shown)


arsalansufi (approved):
LGTM!

PR #7777, opened 2026-01-13
VxDesign: Move generation of results reporting url to backend

jonahkagan:
## Overview

For the upcoming default system settings feature, we'll need to be able to set the results reporting URL by default when creating new elections, which happens on the backend. To ensure that the frontend path still matches the URL set on the backend, I added a type assertion to the route definition.

## Demo Video or Screenshot
N/A
## Testing Plan
Updated existing tests

## Checklist

### DIFF
--- a/apps/design/backend/src/types.ts
+++ b/apps/design/backend/src/types.ts
@@ -10,6 +10,7 @@ import {
 import { DateWithoutTime } from '@votingworks/basics';
 import { ContestResults } from '@votingworks/types/src/tabulation';
 import { z } from 'zod/v4';
+import { baseUrl } from './globals';
 
 export const StateCodes = ['DEMO', 'MS', 'NH'] as const;
 export type StateCode = (typeof StateCodes)[number];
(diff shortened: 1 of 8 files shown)


apps/design/backend/src/types.ts
```diff
@@ -167,3 +168,10 @@ export type ResultsReportingError =
   | 'invalid-payload'
   | 'invalid-signature'
   | GetExportedElectionError;
+
+export const RESULTS_REPORTING_PATH = '/report';
+export type ResultsReportingPath = typeof RESULTS_REPORTING_PATH;
+
+export function resultsReportingUrl(): string {
+  return new URL(RESULTS_REPORTING_PATH, baseUrl()).toString();
+}
```
jonahkagan: This is the function I plan to use for default system settings

apps/design/frontend/src/routes.ts
```diff
@@ -13,7 +15,7 @@ import { Route } from '@votingworks/ui';
 export const resultsRoutes = {
   root: {
     title: 'Results Reported',
-    path: '/report',
+    path: typedAs<ResultsReportingPath>('/report'),
```
kofi-q: nice

Issue #6526, opened 2025-05-30
VxDesign: Crash on long ballot measure


adghayes:
I've found that if I create a ballot measure with a lot of text, I can get a "Something Went Wrong." I've been using 5 paragraphs of auto-generated lorem ipsum.

> Lorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos.
> 
> Lorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos.
> 
> Lorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos.
> 
> Lorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos.
> 
[...]

Issue #6220, opened 2025-04-01
VxDesign: rendering fails for (vertically) long ballot measure images


kofi-q:
Added image resizing in https://github.com/votingworks/vxsuite/pull/6215 to keep images fitting within the bounds of a full-width ballot measure contest box.
We define the maximums statically, so they don't account for the length of the ballot measure text, so if it's long enough, it pushes large images into the next page, which causes rendering to fail.

Probably an unlikely case, and maybe will be handled by upcoming work to split contest contents across multiple pages.

arsalansufi:
Ooh nice great finds!

PR #7780, opened 2026-01-14
VxDesign: Surface a user error when ballot measure splitting fails because a child is itself too tall to fit on the page

arsalansufi:
## Overview

Related to https://github.com/votingworks/vxsuite/issues/6526 and https://github.com/votingworks/vxsuite/issues/6220

The NH ballot template has logic to split long ballot measures across multiple pages. This splitting happens at line breaks. So if a ballot measure contains a long paragraph that is itself too tall to fit on the page, the splitting and overall rendering fails. Today, this results in a hang/crash/timeout. After this PR, we'll surface a user error with guidance for the user on how to proceed.

## Demo Video or Screenshot

| Before | After |
| - | - |
| before | after |

## Testing Plan

- [x] Tested manually
- [x] Updated automated tests

## Checklist


libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -809,6 +814,17 @@ async function splitLongBallotMeasureAcrossPages(
       dimensions.height
   );
   assert(firstOverflowingChildIndex !== -1, 'No overflowing child found');
+
+  // If a given child, e.g., paragraph, is itself too tall to fit on the page, we can't proceed and
+  // need the user to try a longer paper size or higher density, or add a line break to their
+  // content.
+  if (firstOverflowingChildIndex === 0) {
```
nikhilb4a: Mind explaining why we are only checking `=== 0`?

nikhilb4a (approved):
Quick question to understand the error case, but nice improvement!

PR #7781, opened 2026-01-14
VxMark Diagnostics Screen

carolinemodic:
## Overview
Adds diagnostics screen for VxMark accessible from sysadmin and election manager pages. Everything EXCEPT the accessible controller diagnostics is implemented - that particular diagnostic is blocked on us deciding on a controller and will be added later. For now we just show connection status. The barcode diagnostic just looks for any successful barcode scan to pass. The print diagnostic is similar to the VxAdmin printer diagnostic but prints a sample summary ballot rather than a sample ballot count report. All other diagnostics are copied from other vxsuite apps. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/8e204df6-7f1f-4acf-a1e1-d9ee163212e3

## Testing Plan
Tested flow above from both EM and sysadmin screens. Tested on real hardware that audio goes through proper (system vs headphone) channel and that all device connection flags work as expected. Tested the barcode reader with real hardware (not pictured in video). 

## Checklist

### DIFF
--- a/apps/mark-scan/frontend/src/pages/diagnostics/diagnostics_screen.test.tsx
+++ b/apps/mark-scan/frontend/src/pages/diagnostics/diagnostics_screen.test.tsx
@@ -125,11 +125,13 @@ test('data from API is passed to screen contents', async () => {
     DiagnosticSectionTitle.PatInput,
     'Available'
   );
-  screen.getByText('Test passed, 3/23/2022, 11:20:00 AM');
-  screen.getByText('Test passed, 3/23/2022, 11:15:00 AM');
-  screen.getByText('Test passed, 3/23/2022, 11:10:00 AM');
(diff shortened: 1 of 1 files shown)


apps/mark/frontend/src/pages/headphone_input_diagnostic_screen.tsx
```diff
@@ -0,0 +1,93 @@
...
+interface HeadphoneInputDiagnosticScreenProps {
+  onComplete: () => void;
+  onCancel: () => void;
+}
+
+export function HeadphoneInputDiagnosticScreen({
+  onComplete,
+  onCancel,
+}: HeadphoneInputDiagnosticScreenProps): JSX.Element {
+  const audioRef = useRef<HTMLAudioElement>(new Audio('/sounds/chime.wav'));
```
eventualbuddha: Do we need to duplicate this `wav` source?

PR #7789, opened 2026-01-15
Update error message for contest ID assertion

eventualbuddha:
This can show up when a client is stale. This new message might tell the user enough to know they should refresh, whereas the old one does not.

Prompted by [this Sentry error](https://votingworks.sentry.io/issues/7192652758/?alert_rule_id=15751537&alert_type=issue&notification_uuid=dc1e5873-e571-4f3f-b240-c647e24b5c0a&project=4508717994016768&referrer=slack).

### DIFF
--- a/apps/design/backend/src/store.ts
+++ b/apps/design/backend/src/store.ts
@@ -1893,7 +1893,7 @@ export class Store {
         ).rows.map((row) => row.id);
         assert(
           contestIds.length === existingContestIds.length,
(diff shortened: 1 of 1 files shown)


jonahkagan:
I don't think we show assertion errors to users, just a generic "Something went wrong" message. I'm definitely open to thinking through how to gracefully handle update requests with stale data, but it might require a more involved solution.

PR #7797, opened 2026-01-20
Update "Vote for X" copy on the NH template

nikhilb4a:
## Overview

[Slack thread
](https://votingworks.slack.com/archives/C085YT4798C/p1768930701027379)

"silly NH ballot design detail from previous NH-SoS guidance that came up during [#state-nh-ballot-design](https://votingworks.slack.com/archives/C08N5UTSDU5) review we said we’d change for next election. Although I’d argue we meet the [statute as written](https://law.justia.com/codes/new-hampshire/title-lxiii/chapter-659/section-659-17/) today, the state believes that the first clause must be the numerical presentation of the number and the second clause must be the written form. You can see how they think about this in this graphic for reference:"

image

## Demo Video or Screenshot

**Before**
Screenshot 2026-01-20 at 1 09 55 PM

**After**
Screenshot 2026-01-20 at 12 34 51 PM

## Testing Plan

Regenerated fixtures, likely will need to update tests in other libs/apps

## Checklist


libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
```diff
@@ -532,13 +532,11 @@ function CandidateContest({
           <ContestTitle>{electionStrings.contestTitle(contest)}</ContestTitle>
         </DualLanguageText>
         <DualLanguageText delimiter="/">
-          <div>{voteForText}</div>
+          <div>
+            {voteForText}
+            {willBeElectedText && <>; {willBeElectedText}</>}
```
nikhilb4a: I've put the `willBeElected` text on the same line as the `voteForText`, given Matt pointed out the `;` is in the statute. That being said, do you think the keeping them on the same line is the right call, or should we maintain the two line format even with the semicolon?

jonahkagan (approved):

PR #7802, opened 2026-01-20
VxDesign: [libs/ui] Enable text selection on non-touch UIs

kofi-q:
## Overview

We disable text selection by default in our global CSS (to prevent text selection during regular usage of our touchscreen machines) and haven't updated it since adding web-based VxDesign. Updating to only disable selection on whether or not we're using a touchscreen theme.

### Desktop

https://github.com/user-attachments/assets/1b6d4f54-aa3d-4988-a8c7-43b59075a709

### Touch

https://github.com/user-attachments/assets/60308833-1474-492a-a83c-7232152f9296

## Testing Plan
- Manual

## Checklist

### DIFF
--- a/libs/ui/src/global_styles.tsx
+++ b/libs/ui/src/global_styles.tsx
@@ -67,7 +67,8 @@ ${NORMALIZE_CSS}
     font-weight: ${(p) => p.theme.sizes.fontWeight.regular};
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
-    user-select: none;
+    user-select: ${(p) =>
+      isTouchscreen(p.theme.screenType) ? 'none' : undefined};
   }
 
   ${(p) => (p.theme.sizeMode === 'print' ? undefined : LEGACY_PRINT_STYLES)}


jonahkagan (approved):
🎉

libs/ui/src/global_styles.tsx
```diff
@@ -67,7 +67,8 @@ ${NORMALIZE_CSS}
     font-weight: ${(p) => p.theme.sizes.fontWeight.regular};
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
-    user-select: none;
+    user-select: ${(p) =>
+      isTouchscreen(p.theme.screenType) ? 'none' : undefined};
```
jonahkagan: I realized after I approved this that I'm not sure there's any reason for a user to select text on any of the deployed machines, since they don't have access to copy/paste. Maybe we should only enable it on VxDesign specifically

kofi-q: Yeah, I was wondering about this too and figured it's fine either way - don't think it's useful on any of the others, but doesn't hurt either (hopefully I'm not missing an edge case). Could imagine an unlikely edge case scenario where you copy paste something in VxAdmin for some of the data entry tasks, maybe?

jonahkagan: Yeah probably doesn't hurt. I think I remember at some point intentionally disabling it to avoid accidental selection that just looked weird, but I don't think it's a big deal.

PR #7793, opened 2026-01-16
VxDesign: Handle error in rich text editor when image is too big

nikhilb4a:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7298

Handles errors when uploading images via the rich text editor. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/25037553-528b-4be2-b9ca-3233587e92e2

## Testing Plan

Automated test added

## Checklist

- [x] I have prefixed my PR title with "VxDesign: ", "VxPollBook: ", or "HWTA: " if my change is specific to one of those products.
- [ ] I have added [logging](https://github.com/votingworks/vxsuite/tree/main/libs/logging) where appropriate for any new user actions.
- [x] I have added the "user_facing_change" label to this PR, if relevant, to automate an announcement in #machine-product-updates.

PR #7803, opened 2026-01-20
VxDesign: Fix rich text editor disabled styling

jonahkagan:
## Overview

When the error message got added in https://github.com/votingworks/vxsuite/pull/7793, it changed the internal structure of the RichTextEditor component (adding a container). This broke the styling rules from RichTextEditorWithAudio, which were using CSS selectors to modify the style of the editor element (assuming it was the top-level element within the editor component).

It made me realize that due to the way we're encapsulating components, we should probably try to avoid modifying component internals via CSS from parent components (which is something I've definitely done before myself).

Details of the fix in commits!

## Demo Video or Screenshot

Before
Screenshot 2026-01-20 at 3 44 16 PM

After
Screenshot 2026-01-20 at 3 52 06 PM

Updated error state
Screenshot 2026-01-20 at 3 53 16 PM

## Testing Plan
Manual test

## Checklist

### DIFF
--- a/apps/design/frontend/src/rich_text_editor.tsx
+++ b/apps/design/frontend/src/rich_text_editor.tsx
@@ -30,7 +30,6 @@ import styled from 'styled-components';
 import React, { useState } from 'react';
 import { Buffer } from 'node:buffer';
 import { ImageInputButton } from './image_input';
-import { Column } from './layout';
 import { NormalizeParams } from './image_normalization';
 
 const ControlGroup = styled.div`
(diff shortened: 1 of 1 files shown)


jonahkagan:
> Honestly makes me want to just remove the file size limit, since we're resizing client-side now, anyway.

Seems worth exploring. Is there any way for an image to still be too big after resizing? E.g. an SVG with a ton of detail could conceivably be too large, but it seems unlikely.

> Also just noticing that we didn't/don't have a way of clearing the error without uploading another image, so it would persist even if you cancel out of edit mode - seems like an edge case, so can be a later fix.

I noticed this too but also had the same thought that it was enough of an edge case to leave be for now.

Issue #7736, opened 2026-01-05
VxDesign: Self-service downloads, gated by VotingWorks support user action


arsalansufi:
We want customers to be able to download election packages and ballots themselves this ballot production run. We also don't want customers to be able access these files until we've completed QA on our end. This will require the following changes:

- [ ] A support-user sign-off action that unblocks downloads by a customer.
- [ ] Ensuring that election packages and ballot PDFs are “locked” after the finalize button is hit. Even if VxDesign is redeployed with a code change, the election packages and ballot PDFs should be reexportable identically so. They should only change if we unfinalize and refinalize. Implementation detail: Jonah and I have discussed how generating election packages and ballots on finalize and then pulling from S3 when using the export button is a potential path forward here.
- [ ] Making it possible to export the election package and ballots separately, not within the wrapper ZIP that they currently come in. We don’t want customers to have to remember to perform an extra unzip before putting election packages on USB drives. What we export from VxDesign should be directly copyable to a USB drive.
- [ ] Hiding test ballot exports since NH doesn’t use test ballots and we don’t want them to accidentally use those.
- [ ] Clearly labeling sample vs. official ballots. Our current file naming scheme might be sufficient but we also might want something more explicit, whether via separate download buttons or subdirectories within the ballots ZIP.

Additional details and discussion in https://docs.google.com/document/d/1GhtlkXgXIGRcztoz1JbGHPAPOdRSw4Isc-yaNfzbFhY/edit?usp=sharing

PR #7806, opened 2026-01-21
VxDesign: Add state feature flags for export settings

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/7736

Adding state feature flags to control whether each of the following is included in exports:
- TTS audio
- Sample ballots
- Test ballots (this will, by extension, eventually be used to toggle test deck exports as well)

As a first pass, sample ballots are only enabled for NH and test ballots and audio are only enabled for non-NH.

Also updating the export task payload to include a `shouldExportTestBallots` flag, mirroring the existing `shouldExportSampleBallots` flag. Used in filtering the ballot props sent to the renderer - will later be used in determining whether or not to export a test ballot archive once we start splitting up exports into separate ZIP files).

### NH export defaults:
export-defaults-nh

### Non-NH export defaults:
export-defaults-non-nh

## Testing Plan
- New & updated tests
- Manual spot checks

## Checklist


apps/design/backend/test/helpers.ts
```diff
@@ -233,22 +233,25 @@ export async function exportElectionPackage({
...
   apiClient: ApiClient;
   electionId: ElectionId;
   fileStorageClient: FileStorageClient;
   workspace: Workspace;
   electionSerializationFormat: ElectionSerializationFormat;
-  shouldExportAudio: boolean;
-  shouldExportSampleBallots: boolean;
+  shouldExportAudio?: boolean;
+  shouldExportSampleBallots?: boolean;
+  shouldExportTestBallots?: boolean;
```
jonahkagan: I'd prefer these stay as required props so that we don't accidentally forget to propagate them somewhere (unless there's a reason to make them optional)

PR #7808, opened 2026-01-21
VxMark: Auto-disable usb ports when printer is disconnected in activated voting state

carolinemodic:
## Overview
When in an activated voting state automatically disable usb ports, require an authenticated user to re-enable usb ports in order to reconnect a printer. The printer must be connected while a user is authenticated or the ports will be disabled again. PWs and EMs have the ability to re-enable ports when disabled but not disable them. 

See for user flow discussion: https://votingworks.slack.com/archives/CEL6D3GAD/p1769017922988249

## Demo Video or Screenshot

https://github.com/user-attachments/assets/292405be-9171-4b49-8b92-7d1af4257212

## Testing Plan
Jerry-rigged VxDev to actually have port disabling functional see above video. 
Ran tests. 

## Checklist


eventualbuddha (approved):
I don't see any issues with the code, but I do have a concern about how this can be tested in dev. You mentioned that you had to make some one-off modifications to vxdev to do the video you posted and test it. It'd be helpful to have that built into vxsuite somehow, or at the very least documented.

Issue #7800, opened 2026-01-20
VxDesign: MS results SEMS conversion errors when report CSV contains manual tallies


arsalansufi:
We perform a strict check on the CSV headers, and missed that, when manual tallies have been entered, we have additional `Manual Votes,Scanned Votes` columns.

https://github.com/votingworks/vxsuite/blob/feb2b6c0b09b73f0704bc263a7e29895e43f7c7f/apps/design/backend/src/convert_ms_results.ts#L24-L32

https://github.com/votingworks/vxsuite/blob/feb2b6c0b09b73f0704bc263a7e29895e43f7c7f/apps/design/backend/src/convert_ms_results.ts#L99-L106

Sample file: [TEST-official-tally-report-by-precinct__2026-01-20_17-08-00.csv](https://github.com/user-attachments/files/24752440/TEST-official-tally-report-by-precinct__2026-01-20_17-08-00.csv)

PR #7811, opened 2026-01-22
VxDesign: MS results conversion: Handle case that Vx CSV includes manual tallies

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7800

Discovered an edge case where the MS results conversion functionality in VxDesign considers a Vx results CSV with manual tallies an invalid input due to unrecognized headers. Updated the code to handle Vx results CSVs both with and without manual tallies.

## Testing Plan

- [x] Tested manually
- [x] Updated automated tests

## Checklist


arsalansufi:
Not that this is a big PR, but to balance review load, assigned to @nikhilb4a instead of @jonahkagan and will likely stick with @nikhilb4a for small changes that I make to chip in with ballot production prep!

Issue #7471, opened 2025-11-05
VxScan: Accessibility odds and ends for cert


arsalansufi:
Parent cert discrepancies: https://github.com/votingworks/cert-discrepancies/issues/222, https://github.com/votingworks/cert-discrepancies/issues/268

This issue could take many different forms depending on the exact accessibility peripherals that we land on. Our selection will be entirely driven by what we believe we can pass hardware testing with. That could be anything from a USB-connected storm controller to a PAT and headphone jack inside the PW door.

Potential tasks:
- [ ] Mapping controller key presses to those that VxScan expects for audio controls
- [ ] Depending on the number of buttons on the accessible input, some number of on-screen audio controls
- [ ] Depending on the audio output, ensuring that scan success/error sounds still emit through the speaker (nothing to do if the accessible audio output is USB based, more to do if the accessible audio output uses the audio jack on the SBC)
- [ ] Tying up alarm work for removal of the accessibility peripheral if possible

Eng plan + Mocks: https://docs.google.com/document/d/1XlaKNvgxXf4nUD5KZcl1KZna5_83yabV-DjWDs0yrIw
Calibration Guide: https://docs.google.com/document/d/16fJqQPNTT_Rf29gkmwQ5W-5wQH5JFzTXrDdjJIqYQfg/edit

PR #7859, opened 2026-01-29
[Scan] Enable PAT device nav on voter settings screen

kofi-q:
## Overview

Moving keyboard nav handling up to the root app component to get coverage on the voter settings screen as well. Was previously only triggering on the main screens, since the voter settings screen is rendered at a top-level route. Probably makes sense to eventually move the settings screen into the `VoterScreen` anyway - just keeping the change scope small for now.

Verified this doesn't interfere with the PAT device tutorial taking priority on initial input, since that intercepts events in the capturing phase.

Also excluding the dev dock from accessible controller navigation, since it often gets in the way/confusing especially when it's off-screen.

### Before:

https://github.com/user-attachments/assets/e490420c-802f-424e-8d17-ee51d4ff3f0c

### After:

https://github.com/user-attachments/assets/9d16fa29-fa3e-406b-9d17-cd54afaa09c8

## Testing Plan
- Added repro test

### DIFF
--- a/libs/dev-dock/frontend/src/dev_dock.test.tsx
+++ b/libs/dev-dock/frontend/src/dev_dock.test.tsx
@@ -122,12 +122,12 @@ test('renders nothing if dev dock is disabled', () => {
   featureFlagMock.disableFeatureFlag(
     BooleanEnvironmentVariableName.ENABLE_DEV_DOCK
   );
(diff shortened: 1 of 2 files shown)


carolinemodic (approved):
thanks for fixing the dev dock thing too! I find that very annoying lol

PR #7860, opened 2026-01-30
VxDesign/VxMark: Allow for multi-page summary ballots 

carolinemodic:
## Overview
When the content on the page of a summary ballot can not be rendered on one page this allows for splitting the ballot up across multiple pages. The page split determination is done dynamically on render so depending on the voting pattern of that particular ballot (maximizing all options on a vote for 10 uses more physical space then leaving a contest blank, for example) a different ballot style may have different numbers of pages in different instances. This PR first addresses the generation of multi-page summary ballots both from VxMark and VxDesign in test deck generation. VxMarkScan can NOT print multi page bmd ballots due to its more complicated flows and is NOT impacted. Then it loosens the relevant system settings for VxMark. I am NOT addressing the QR code size overflow possibility in this PR, the number of write in characters used across contests is still capping the same way VxMarkScan is. Later commits address the scanning and interpretation from VxScan/VxCentralScan and ultimately, the CVR creation/import in VxAdmin and melding with hmpb sheet numbers for sheet by sheet accounting. 

We no longer display summary vs bubble ballot counts in the system, so if there is 1 hmpb sheet 1 scanned, 2 hmpb sheet 2 scanned, 2 bmd sheet 1 scanned, and 1 bmd sheet 2 scanned, the final ballot counts will look like: sheet 1: 3, sheet2: 3. 

Based on Product teams preference ballots are still always printed single sided and expected that way when scanning. 

## Demo Video or Screens
Example multi sheet VxMark Print: 
[sheet-3.pdf](https://github.com/user-attachments/files/25018796/sheet-3.pdf)
[sheet-2.pdf](https://github.com/user-attachments/files/25018798/sheet-2.pdf)
[sheet-1.pdf](https://github.com/user-attachments/files/25018800/sheet-1.pdf)

Example Polls Closed Report with multi-sheet hmpb and multi-sheet bmd (above ballots) scanned:
[print-job-2026-01-29T23:36:03.308Z.pdf](https://github.com/user-attachments/files/25018828/print-job-2026-01-29T23.36.03.308Z.pdf)

Example VxAdmin Report with mix multi-sheet hmpb/summary ballots: 
[TEST-unofficial-full-election-tally-report__2026-02-02_08-23-41.pdf](https://github.com/user-attachments/files/25018897/TEST-unofficial-full-election-tally-report__2026-02-02_08-23-41.pdf)

See snapshots for more examples 

## Testing Plan
Manually tested workflow with medium to large sized elections from VxDesign and test deck creation, creating ballots with different voting patterns in VxMark, scanning in BOTH central scan and vxscan (verifying polls closed report), and importing CVRs to vxadmin and verifying reports. See snapshots for examples of ballots. Tested both with an election that always overflows to multiple pages in test decks and in fully-voted scenarios, and one on the boundary that only overflows when fully voted. 

## Checklist

### DIFF
--- a/libs/ballot-interpreter/src/interpret.ts
+++ b/libs/ballot-interpreter/src/interpret.ts
@@ -561,8 +561,6 @@ function validateInterpretResults(
       sliceBallotHashForEncoding(metadata.ballotHash) !==
       sliceBallotHashForEncoding(ballotHash)
     ) {
-      console.log('expected:', sliceBallotHashForEncoding(ballotHash));
-      console.log('actual:', sliceBallotHashForEncoding(metadata.ballotHash));
       return typedAs<InvalidBallotHashPage>({
         type: 'InvalidBallotHashPage',
         expectedBallotHash: sliceBallotHashForEncoding(ballotHash),


apps/mark/backend/src/util/print_ballot.tsx
```diff
@@ -34,8 +59,74 @@ export async function printBallot(p: PrintBallotProps): Promise<void> {
...
   const { electionDefinition } = assertDefined(store.getElectionRecord());
+  const { election } = electionDefinition;
   const isLiveMode = !store.getTestMode();
 
+  if (!sharedRenderer) {
+    sharedRenderer = new SummaryBallotLayoutRenderer();
+  }
+
+  const uiStringsPackage = store.getUiStringsStore().getAllUiStrings();
+  const pageBreaks = await sharedRenderer.computePageBreaks(
```
eventualbuddha: I'm guessing that the vast majority of the time we'll be returning a single page from this. Could we not combine the rendering with the computation of the page breaks for such a case? The `computes single page for small election` test case on my (pretty fast) dev machine takes 220ms to run this function. On production hardware that will likely be closer to a full second or perhaps more. For the "large election" test it's even worse: 460ms, which could be multiple seconds in production, though that one ends up at 5 pages so can't be optimized in the same way.

Have you tried this on production hardware?

libs/printing/src/summary_ballot_layout.test.tsx
```diff
@@ -0,0 +1,447 @@
...
+    } finally {
+      await renderer.close();
+    }
+    const pdfs = await renderMultiPageBallot(
+      electionDefinition,
+      pageBreaks,
+      votes
+    );
+
+    // Convert PDFs to images and snapshot test each page
```
eventualbuddha: Ditto.

apps/mark/backend/src/util/print_ballot.tsx
eventualbuddha: Nice, seems like a good approach.

carolinemodic:
noting that I also just sanity tested the adjudication flows when adjudicating a write in on a first page and second/third page of a bmd ballot and saw the proper image rendered

Issue #7785, opened 2026-01-15
SHV QR codes can be dense and don't always scan easily on all devices


arsalansufi:
*Problem:* 
SHV QR codes can be dense and don't always scan easily on all devices.

*From:* Vince / Arsalan

*Proposed solution:* 
We can likely decrease QR code density by decreasing the error correction level. Error correction is more important for printed QR codes that can be physically damaged. For a QR code rendered on a high fidelity screen, less relevant.

_Collected by @arsalan_

Issue automatically created from Slack feedback reported by Product Feedback: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1768495626258289

PR #7867, opened 2026-02-01
Decrease density of SHV QR codes by decreasing error correction level

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7785

Our SHV QR codes can be so large/dense that some phones have trouble reading them. Some of this density is attributable to our use of the highest error correction level. This PR decreases the density of SHV QR codes by decreasing that level. For live reporting, we already use level M instead of level H QR codes, so we have existing signal that this is safe.

As part of this PR, I've also decreased the default error correction level for future QR codes.

Copying the code comment in the PR:

>  Our QR codes are generally displayed on high-fidelity screens or printed with high-fidelity
>  printers, so the highest error correction levels are in most cases unnecessary. In fact, for
>  some of our larger QR codes, e.g., SHV QR codes, we've found that the highest levels make the QR
>  codes so large/dense that some phones have trouble reading them.
> 
>  We can override with higher levels when we need to optimize for resilience, e.g., summary
>  ballots printed on lower-fidelity thermal printers. And we can override with lower levels when
>  we need to optimize for space constraints, e.g., HMP ballots.

## Demo Video or Screenshot

| Before | After |
| - | - |
| before | after |

## Testing Plan

- [x] Scanned new less dense SHV QR code manually

## Checklist

- [ ] ~I have added the "user_facing_change" label to this PR, if relevant, to automate an announcement in #machine-product-updates.~ N/A

### DIFF
--- a/libs/ui/src/qrcode.tsx
+++ b/libs/ui/src/qrcode.tsx
@@ -9,7 +9,23 @@ const ResponsiveSvgWrapper = styled.div`
   }
 `;
 
+/**
+ * Using a higher level makes a QR code more resilient to damage or distortion but also requires
+ * a larger/denser QR code.
+ */
 export type QrCodeLevel = 'L' | 'M' | 'Q' | 'H';
+
+/**
+ * Our QR codes are generally displayed on high-fidelity screens or printed with high-fidelity
+ * printers, so the highest error correction levels are in most cases unnecessary. In fact, for
+ * some of our larger QR codes, e.g., SHV QR codes, we've found that the highest levels make the QR
+ * codes so large/dense that some phones have trouble reading them.
+ *
+ * We can override with higher levels when we need to optimize for resilience, e.g., summary
+ * ballots printed on lower-fidelity thermal printers. And we can override with lower levels when
+ * we need to optimize for space constraints, e.g., HMP ballots.
+ */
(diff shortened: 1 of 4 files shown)


libs/ui/src/qrcode.tsx
```diff
@@ -9,15 +9,35 @@ const ResponsiveSvgWrapper = styled.div`
...
 
+/**
+ * Our QR codes are generally displayed on high-fidelity screens or printed with high-fidelity
+ * printers, so the highest error correction levels are in most cases unnecessary. In fact, for
+ * some of our larger QR codes, e.g., SHV QR codes, we've found that the highest levels make the QR
+ * codes so large/dense that some phones have trouble reading them.
+ *
+ * We can override with higher levels when we need to optimize for resilience, e.g., summary
+ * ballots printed on lower-fidelity thermal printers. And we can override with lower levels when
+ * we need to optimize for space constraints, e.g., HMP ballots.
```
arsalansufi: HMP ballots specify L for their QR code. That's unchanged by this PR:

https://github.com/votingworks/vxsuite/blob/3c2def4e85da359f476b8d199e6d9c415446e4d0/libs/hmpb/src/render_ballot.tsx#L457

PR #7906, opened 2026-02-02
VxDesign: Optimize contests screen loading

jonahkagan:
## Overview

After introducing the section headers feature, I noticed that the contest list was loading after the edit panel, causing the elements to jump around. To fix this, I changed the parent component to not render until all queries needed by both subcomponents are loaded.

I also noticed while debugging this issue that there were a lot of duplicate requests (due to the same query being mounted by multiple components on the same page). We solved this in Arlo by adding a 1s stale time (instead of the default of 0), so I implemented that solution here.

## Demo Video or Screenshot

Before

https://github.com/user-attachments/assets/23a523f7-62cd-413f-9bbe-8c3dedf1cbc2

After

https://github.com/user-attachments/assets/7bece207-7251-49ba-ae40-1fc6933ea69d

## Testing Plan
Manual test

## Checklist


kofi-q (approved):
LGTM! Non-blocking note

PR #7924, opened 2026-02-03
VxDesign: Update reorder contests logic to be single query

nikhilb4a:
## Overview

Sentry alert - https://votingworks.slack.com/archives/C08AWNY93PB/p1770130303861249

Our `reorderContests` and `createContests` (using `insertContest`) endpoints both have caused alerts for slow queries when reordering contests, because they make queries for each contests when reordering. This PR updates the logic to have shared helper that reorders contest using a multi-row query.

## Demo Video or Screenshot

No external behavior change

## Testing Plan

Manually tested reordering/adding a contest
Automated tests already in place

## Checklist

### DIFF
--- a/apps/design/backend/src/store.ts
+++ b/apps/design/backend/src/store.ts
@@ -342,6 +342,35 @@ async function insertParty(
   );
 }
 
+async function applyContestOrder(
+  client: Client,
+  electionId: ElectionId,
+  contestIds: string[]
+): Promise<number> {
+  const contestPositions = contestIds.map((_, i) => i);
+  const { rowCount } = await client.query(
+    `
+    with next_order as (
+      select coalesce(max(ballot_order) + 1, 1) as value
+      from contests where election_id = $3
+    )
+    update contests
+    set ballot_order = ordered.position + (select value from next_order)
+    from unnest($1::text[], $2::int[]) as ordered(id, position)
+    where contests.id = ordered.id
+      and contests.election_id = $3
+    `,
+    contestIds,
+    contestPositions,
+    electionId
+  );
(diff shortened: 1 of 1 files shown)


apps/design/backend/src/store.ts
```diff
@@ -1877,18 +1899,15 @@ export class Store {
...
-          assert(rowCount === 1, `Contest not found: ${contestId}`);
-        }
+        const rowCount = await applyContestOrder(
+          client,
+          electionId,
+          contestIds
+        );
+        assert(
+          rowCount === contestIds.length,
+          `${contestIds.length - rowCount} contests not found`
```
nikhilb4a: We lose the missing `contestId` in an error case with this approach, but the trade off seems worth it given that error seems unlikely

nikhilb4a:
https://votingworks.slack.com/archives/C08AWNY93PB/p1770234888628439?thread_ts=1770130303.861249&cid=C08AWNY93PB

"After looking into solutions, Jonah and I decided it probably isn't worth optimizing this query because it isn't having real world effects on users, as the endpoint is completing requests quite quickly since the number of contests per election is bounded. I will look into a change after this election cycle."

For now, we are instead updating our Sentry settings to increase the threshold of request duration for this n+1 alert to be triggered. Of the [5 events](https://votingworks.sentry.io/issues/7239223250/events/?alert_rule_id=15751537&alert_type=issue&notification_uuid=3dea0d5c-d25f-4c5e-8d2d-4f33013f576c&project=4508717994016768&query=&referrer=slack) I looked at, the max duration was ~120ms, so I've set the [minimum threshold to 200ms](https://votingworks.sentry.io/settings/projects/vxdesign/performance/) for now. From there it has the threshold can be increased in 100ms increments.

PR #7936, opened 2026-02-05
feat: add retry streak detection with lower threshold for unreadable ballots

### LINKED ISSUE #7659: Have ballot interpreter use same constant for adjacent shape merging and max cumulative streak width threshold (4px) (and/or other potential refinements)
https://votingworks.slack.com/archives/CEL6D3GAD/p1765460807832929?thread_ts=1765382300.773589&cid=CEL6D3GAD

Other ideas to consider:
* Make the adjacent shape merging constant the max allowed value for the max cumulative streak width threshold. It's not just a good default but also potentially an upper bound on what's safe.
* Alternatively, we could check streaks again with a 1px threshold if we can’t find the timing marks. Basically turning an unreadable ballot error into a cleaning warning. This would allow us to safely up the threshold past the adjacent shape merging constant.

eventualbuddha:
## Overview

Fixes #7659

When timing marks cannot be found on a ballot, retry streak detection with a lower threshold to differentiate between truly unreadable ballots and ballots that are unreadable due to minor streaks below the normal threshold.

This allows the system to surface "scanner needs cleaning" errors instead of generic "unreadable ballot" errors when minor streaks are the root cause, providing better feedback to poll workers.

## Demo Video or Screenshot
Screenshot_2026-02-05_10-35-29

Configuration:
- `maxCumulativeStreakWidth` (default: 5px) - normal streak threshold
- `retryStreakWidthThreshold` (default: 1px) - retry threshold (must be < normal)

Both thresholds are configurable via VxDesign system settings under "Scanner Thresholds" and are applied system-wide to VxScan and VxCentralScan.

## Testing Plan

Tested with new automated unit tests.

## Checklist

### DIFF
--- a/apps/design/frontend/src/system_settings_screen.test.tsx
+++ b/apps/design/frontend/src/system_settings_screen.test.tsx
@@ -754,12 +754,12 @@ test('validates streak width threshold must be less than max cumulative width', 
 
   userEvent.click(screen.getByRole('button', { name: 'Edit' }));
 
-  const maxCumulativeStreakWidthInput = screen.getByRole('spinbutton', {
+  const maxCumulativeStreakWidthInput = screen.getByRole<HTMLInputElement>('spinbutton', {
     name: 'Max Cumulative Streak Width (pixels)',
-  }) ;
-  const retryStreakWidthThresholdInput = screen.getByRole('spinbutton', {
+  });
+  const retryStreakWidthThresholdInput = screen.getByRole<HTMLInputElement>('spinbutton', {
     name: 'Retry Streak Width Threshold (pixels)',
-  }) ;
+  });
 
   // Set max width to 5 and retry threshold to 5 (invalid: should be less than)
   userEvent.clear(maxCumulativeStreakWidthInput);


libs/ballot-interpreter/src/bubble-ballot-rust/interpret.rs
```diff
@@ -359,13 +377,36 @@ pub fn ballot_card(
...
+    let mut timing_marks = match ballot_card.find_timing_marks(options.timing_mark_algorithm) {
+        Ok(marks) => marks,
+        Err(Error::MissingTimingMarks { reason }) => {
+            // If timing marks couldn't be found, retry streak detection with a lower threshold
+            // to differentiate between truly unreadable ballots and ballots with minor streaks.
+            if matches!(
+                options.vertical_streak_detection,
+                VerticalStreakDetection::Enabled
+            ) {
+                let streaks = ballot_card.detect_vertical_streaks();
```
jonahkagan: Do we need to detect the streaks again or can we just use `detected_vertical_streaks`?

jonahkagan (approved):
Excited for this!

PR #7937, opened 2026-02-05
VxDesign: Hide timing marks on sample ballots

jonahkagan:
## Overview

Sample ballots aren't for scanning, so we hide the timing marks to make that even clearer. This came up as we were discussing requirements for MI ballots, but we realized it should apply to all sample ballots.

## Demo Video or Screenshot
[Sample ballot](https://github.com/votingworks/vxsuite/blob/8ba980235f88f0fdba5c66814d86ad646e5b90dd/libs/hmpb/fixtures/vx-famous-names/sample-ballot.pdf)

## Testing Plan
Added sample ballot fixture

## Checklist


libs/hmpb/src/render_ballot.tsx
```diff
@@ -581,7 +584,7 @@ export async function layOutBallotsAndCreateElectionDefinition<
   // All ballots of a given ballot style must have the same grid layout.
   // Changing precinct/ballot type/ballot mode shouldn't matter.
   const layoutsByBallotStyle = iter(ballotLayouts)
-    .map((ballot) => ballot.gridLayout)
+    .filterMap((ballot) => ballot.gridLayout)
```
nikhilb4a: til filtermap!

libs/hmpb/fixtures/vx-famous-names/sample-ballot.pdf
nikhilb4a: The ballot looks so different with no timing marks 😅

libs/hmpb/src/render_ballot.tsx
jonahkagan: Yeah, it's a nice one to have! Wish it were part of the standard array methods

Issue #7877, opened 2026-02-01
VxScan: add early voting mode


adghayes:
1. Add a new “Early Voting” mode to VxScan (vs. the current default “Election Day” mode)
    i. VxScan should display on all relevant screens that it is in early voting mode. We can loosely define “relevant screens” as the screens where the test mode banner already appears.
   ii. It should not be possible to toggle to or from early voting mode while a batch is in progress (a.k.a polls are open).
   iii. When in early voting mode, the default option on the poll worker screen should be “Pause Voting” rather than “Close Polls”.

Issue #7876, opened 2026-02-01
VxSuite: Early Voting Support (Parent Issue)


adghayes:
See https://docs.google.com/document/d/1oK6-pJslgZorXxy1MBV6uYnk2emvgdTpq0eQkqgPd-s/edit?tab=t.vfhcet5dhaih.

PR #7939, opened 2026-02-05
VxScan: add backend, button, and flag for early voting mode

kshen0:
## Overview

Relates to https://github.com/votingworks/vxsuite/issues/7876 and partially addresses https://github.com/votingworks/vxsuite/issues/7877

Adds plumbing for setting "Ballot Casting Mode" (internal name only).

Does not read the value other than the button, which is hidden by feature flag.

## Demo Video or Screenshot

Screenshot 2026-02-05 at 1 31 27 PM

## Testing Plan
- added tests

## Checklist

- [ ] I have added the "user_facing_change" label to this PR, if relevant, to automate an announcement in #machine-product-updates. (Will add label when flag is removed)

### DIFF
--- a/apps/scan/backend/src/store.ts
+++ b/apps/scan/backend/src/store.ts
@@ -336,7 +336,7 @@ export class Store {
     ) as { ballotCastingMode: BallotCastingMode } | undefined;
 
     if (!electionRow) {
-      // ballot casting mode will be the default once an election is defined
+      // election_day will be the default when an election is defined
       return 'election_day';
     }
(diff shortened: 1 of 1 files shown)


adghayes:
looks good!

For consistency with the future, I recommend using "Ballot Casting Mode" rather than "Ballot Casting Period". In the future, we may be adding something similar to VxCentralScan in which case the "period" doesn't change. An annoying change given all you've already written up, but better now than further down the line.

PR #7940, opened 2026-02-05
[Scan/Mark] Add on-screen audio controls to Audio Settings

kofi-q:
## Overview

Adding on-screen volume and rate-of-speech adjustment controls to the voter audio settings screen in VxScan (and by extension, VxMark, since they share the same settings component).

Additionally, updating the font sizes on VxScan to match the current screen size on production hardware. Made it easier to fit everything on the audio settings screen.

### VxScan

https://github.com/user-attachments/assets/1ac7f0d0-8f05-42d2-acba-25884e38d541

#### Voter Instructions

scan-voter-instructions

### VxMark

https://github.com/user-attachments/assets/2d00eb70-8202-412e-9837-9e1d5c9cada6

## Checklist

### DIFF
--- a/libs/ui/src/ui_strings/audio_context.test.tsx
+++ b/libs/ui/src/ui_strings/audio_context.test.tsx
@@ -14,9 +14,8 @@ import {
 } from './audio_volume';
 import {
   DEFAULT_PLAYBACK_RATE,
-  MAX_PLAYBACK_RATE,
-  MIN_PLAYBACK_RATE,
   PLAYBACK_RATES,
+  PlaybackRate,
 } from './audio_playback_rate';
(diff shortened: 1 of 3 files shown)


kofi-q:
K, code and demo video updated - it's a lot of change, so will wait for a second round review when you're back next week

kofi-q:
> Looks good, just to clarify AFAICT you are adding the labels to read the rate of change amount when toggling via the UI, but I should still follow up on that other task I have to make them also read out when we toggle rate via the controller?

Ah yeah, this only covers the on-screen controls - for controller-based rate change, the logic in the screen reader event handler will need updating

PR #7942, opened 2026-02-05
VxMark: Restart audio for your almost done page on left arrow

### LINKED ISSUE #7661: VxMark: provide ability to repeat final screen instructions

carolinemodic:
## Overview
Closes #7661 
On contest pages the user can use left/right arrows to move between pages and repeat the intro text if they desire. There is no way to do this on the post print instructions screen, this makes it so that the "left" arrow which otherwise is unused (since you can not go back) will just repeat the "Your almost done..." text if pressed. I think this intuitively makes sense since you are navigating back to reload the page in a way and it matches how you would repeat text on contest pages but navigating forward and back. 

## Demo Video or Screenshot
N/A

## Testing Plan
Manually tested, verified text is repeated.

## Checklist


kofi-q (approved):
Nice! Was wondering how we'd solve this - this is much cleaner than I'd feared 😆

Issue #5651, opened 2024-12-03
VxMark: add election hash to BMD ballots


adghayes:
very useful for ballot testing and accounting

PR #7967, opened 2026-02-09
VxSuite(ALL): Add shared Claude.MD file

carolinemodic:
## Overview
Adds a basic shared Claude.MD file and symlinks .github/copilot-instructions.md to it. This is just meant to be a starting place. I apparently actually accidently nuked my files last week doing a `pnpm clean-all` so this has not been tested on much work. 

Also adds a pnpm test:run directive to all apps/libraries so that running tests in run rather than watch mode is generally simpler. (in seperate commit) 

Eventually we will likely want to link to other md files with more details on certain things. I imagine the style guide section will need the most iteration and changes as we figure out how to actually tell the tools to write code in the same style that we do. 

Note: Theoretically copilot also ingests Claude.MD but at least once in the past I didn't find it to be reading the instructions so I thought adding a symlink was safe  for redundancy but might not be needed. 

## Demo Video or Screenshot
N/A

## Testing Plan
Ran claude code on my next task (https://github.com/votingworks/vxsuite/issues/5651) and see that it is: understanding where the code for "vxmark frontend" is, running tests in a way it can handle, etc.

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -11,7 +11,7 @@ contains 8 applications and 30+ shared libraries managed with pnpm workspaces.
 apps/           # 8 voting system applications
   admin/        # Election administration
   central-scan/ # Central ballot scanning
-  design/       # Election design (VxDesign)
+  design/       # Election design (VxDesign) — see note below
   mark/         # Ballot marking device
   mark-scan/    # Combined mark + scan device
   pollbook/     # Electronic pollbook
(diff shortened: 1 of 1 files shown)


eventualbuddha (approved):
Seems worth a shot.

CLAUDE.md
```diff
@@ -0,0 +1,338 @@
...
+- **`const` by default** — only use `let` when reassignment is needed
+- **Interfaces over type aliases** — prefer `interface` for object shapes
+- **Array syntax** — use `T[]` not `Array<T>`
+- **File naming** — snake_case for all module files (e.g. `app_root.tsx`,
+  `use_language_controls.ts`)
+- **Constants** — UPPER_CASE for module-level constants
+- **No floating promises** — all promises must be awaited or explicitly voided
+  (`void somePromise()`)
+- **Result types** — use `Result<T, E>` from `@votingworks/basics` for error
+  handling instead of throwing
```
jonahkagan: I would suggest a slightly more nuanced convention around error handling:
- Fail fast for unexpected errors: Unexpected errors should crash the application (e.g. throw an exception) to ensure we detect them early and prevent undefined behavior.
- Expected errors should be handled with the `Result` type: Errors resulting from user behavior or known possible external conditions should be handled using the `Result<T, E>` type from `@votingworks/basics` to ensure that we preset an actionable error message to the user.

jonahkagan (approved):

PR #7977, opened 2026-02-11
[Scan] HWTA: Add minimal headphone calibration

kofi-q:
## Overview

Couple changes to the VxScan HWTA related to the switch to using builtin audio for the screen reader:
- Update the interval sounds to alternate between speaker and headphones - since they both now share the same audio card, they can't be played at the same time.
- Add a headphone calibration button/modal - currently just enables adjusting the gain offset we apply to screen reader audio in order to find a setting that gets us as close to the `100 dB` VVSG2 maximum without going over. 
  - The idea is that HW team can perform the calibration and SW team can update the code accordingly. There's no persistence for now, WRT the calibrated value, but we can expand functionality as needed.

## Demo Video or Screenshot (audio chimes only sped up for the recording)

https://github.com/user-attachments/assets/0363b6e5-e43d-4459-80c2-1725654adf61

## Testing Plan
- Manual

## Checklist

### DIFF
--- a/apps/scan/frontend/src/electrical_testing/app_root.tsx
+++ b/apps/scan/frontend/src/electrical_testing/app_root.tsx
@@ -463,7 +463,8 @@ function UsbDriveControls({
 }
 
 function AudioControls(): JSX.Element {
-  const [enabled, setEnabled] = React.useState(true);
+  const [headphonesOn, setHeadphonesOn] = React.useState(true);
+  const [speakerOn, setSpeakerOn] = React.useState(true);
   const [calibrating, setCalibrating] = React.useState(false);
   const [output, setOutput] = React.useState<'Headphones' | 'Speaker'>(
     'Speaker'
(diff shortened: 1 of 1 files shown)


carolinemodic (approved):
Nice! 
Are we worried at all about needing to set different hardware units to different calibration levels? Since it seems like we will have to update this in-code which will impact all units. For my own knowledge can you point me where we actually configure this in code? 

Once updated in code will this also calibrate the system level in addition to the headphones level? 

For VxMark, I think what the headphones audio is coming through is still TBD but if we switch to a port I presume we will want to add this there, if we end up with the audio through usb is there a way to calibrate that? Is it still useful to have this to calibrate the system level (if its doing that haha) ? 

Seems like you are muting both sounds as opposed to allowing to mute just speaker or just headphones, the latter is how I did it on VxMark, but since you are listing which one is playing at any moment I think its fine as you can tell what the machine thinks its doing.

PR #7979, opened 2026-02-11
VxMark/MarkScan: Add ability to skip to viewing all contests from any contests page 

carolinemodic:
## Overview
Adds a "View All" button to all contests pages, other than hte final contest on the ballot, that when clicked skips to a version of the review screen. When accessed this way the review screen will have a "Back" button, and clicking "Change" on any contest will bring you back to the contests page where you keep the view all / back / next button paradigm. When accessed via "Next" on the final contest the review page behaves as it did prior to this change. 

When in XL mode the first button will be full width and the buttons will be in two rows. 

Note: I'm appending "Contests" to "View All" on the audio-track to make that clearer. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/61fa4102-d399-4cd0-a5f1-6a2181a13a3a

## Testing Plan
Manually tested through Mark and MarkScan as shown above. Ran tests 

## Checklist

### DIFF
--- a/libs/mark-flow-ui/src/pages/contest_page.tsx
+++ b/libs/mark-flow-ui/src/pages/contest_page.tsx
@@ -16,7 +16,7 @@ import {
   Button,
   PageNavigationButtonId,
   AccessibilityMode,
-  AudioOnly,
+  WithAltAudio,
 } from '@votingworks/ui';
 import { assert, throwIllegalValue } from '@votingworks/basics';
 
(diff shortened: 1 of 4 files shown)


libs/mark-flow-ui/src/pages/contest_page.tsx
```diff
@@ -136,16 +137,25 @@ export function ContestPage(props: ContestPageProps): JSX.Element {
...
+      rightIcon={isReviewMode ? 'Next' : 'ListUnordered'}
+      ref={viewAllButtonRef}
     >
-      {appStrings.buttonReview()}
+      {isReviewMode ? (
+        appStrings.buttonReview()
+      ) : (
+        <React.Fragment>
+          {appStrings.buttonViewAll()}
+          <AudioOnly> {appStrings.labelContests()}</AudioOnly>
```
kofi-q: Where possible, we've been trying to avoid audio that involves stitching 2 clips, since they read much better as one continuous clip (exceptions are when we have text in the `Label: Value` format) - would recommend using the `<WithAltAudio>` approach here and having the audio text be a continuous `View all contests` clip instead

PR #8003, opened 2026-02-17
refactor(scan): replace `zbar` with `zedbar`

eventualbuddha:
## Overview

Before this change, VxSuite used two QR code detectors in Rust and two in Node. The Rust ones were the `zbar-rust` crate (links to `libzbar-dev` on the system) and the `rqrr` crate (pure Rust). The Node ones were `@votingworks/qrdetect` (a Node addon bundling the C code for `zbar`) and `quirc` (another C-based QR code reader). The Rust side handles bubble ballots and the Node side handles summary ballots.

This change replaces the two `zbar`-based QR code readers with `zedbar`, my own Rust port of `zbar`. On the Rust side it builds like any other pure-Rust crate. On the Node side we use the `zedbar` NPM package, which is a WebAssembly build of the Rust `zedbar` code.

Ideally we'd consolidate into just `zedbar` and `rqrr` and handle both summary and bubble ballot interpretation on the Rust side, but that's a larger change I didn't want to tackle just yet. This change means we no longer need to keep `@votingworks/qrdetect` around, and we no longer need to install `libzbar-dev`. We're still bundling two copies of `zedbar`, but at least it removes a medium-sized C dependency.

## Demo Video or Screenshot

n/a

## Testing Plan

- [x] Automated tests.
- [x] Performance comparison.
- [x] Re-scan with TRR corpus.


jonahkagan (approved):
Cool!

eventualbuddha:
I had Claude do some analysis and posted it [to the `votingworks/qr-benchmark` repo](https://github.com/votingworks/qr-benchmark). The result summary is below.

`zedbar` is slower than `zbar` (almost 2x), but the absolute numbers we're dealing with are pretty small. They could be in the 10-20ms range on our production machines, but that's still a tiny difference and seems worth the trade off to me. Importantly, `zedbar` and `zbar` have exactly the same accuracy, and they're better than `quirc` and `rqrr` in terms of success rate.

---

Benchmarked 5 QR code readers on **2,472 real ballot PNG images** from a TRR CVR
export. Each image was cropped to bottom-left and top-right corners (width/4
square) matching production detection logic, then tested with each reader.

### Rust Readers (native)

| Reader | Success | Fail | Rate | Mean | Median | Min | Max | p95 |
|--------|--------:|-----:|-----:|-----:|-------:|----:|----:|----:|
| rqrr | 1,968 | 504 | 79.6% | 3.26ms | 2.87ms | 2.44ms | 21.42ms | 4.84ms |
| **zbar** (old) | 2,429 | 43 | 98.3% | 1.71ms | 1.43ms | 1.33ms | 16.77ms | 2.71ms |
| **zedbar** (new) | 2,429 | 43 | 98.3% | 3.08ms | 2.57ms | 2.01ms | 18.77ms | 6.37ms |

### Node Readers (addon/WASM)

| Reader | Success | Fail | Rate | Mean | Median | Min | Max | p95 |
|--------|--------:|-----:|-----:|-----:|-------:|----:|----:|----:|
| quirc | 2,001 | 471 | 80.9% | 1.37ms | 1.32ms | 0.78ms | 8.43ms | 2.28ms |
| **qrdetect** (old) | 2,429 | 43 | 98.3% | 2.57ms | 2.34ms | 1.80ms | 8.33ms | 3.88ms |
| **zedbar-wasm** (new) | 2,429 | 43 | 98.3% | 4.43ms | 4.10ms | 3.30ms | 18.33ms | 6.64ms |

### Discrepancies

| Comparison | Old succeeded, new failed | New succeeded, old failed | Data mismatches |
|------------|:-------------------------:|:-------------------------:|:---------------:|
| zbar vs zedbar (Rust) | 0 | 0 | 0 |
| qrdetect vs zedbar-wasm (Node) | 0 | 0 | 0 |

### Summary

- **Accuracy is identical.** zedbar detects the exact same 2,429/2,472 images as zbar. zedbar-wasm detects the exact same 2,429/2,472 images as qrdetect. Zero discrepancies in either direction, zero data mismatches.
- **Performance is ~1.8x slower.** zedbar median is 2.57ms vs zbar's 1.43ms (Rust native). zedbar-wasm median is 4.10ms vs qrdetect's 2.34ms (Node). Both are well within acceptable bounds for ballot scanning.
- **rqrr and quirc remain weaker.** Both miss ~20% of ballots, confirming why the fallback chain (rqrr then zbar/zedbar) is necessary.

### Methodology

- **Rust binary:** `rqrr 0.7.1`, `zbar-rust 0.0.23`, `zedbar 0.2.1` via `image` crate, parallelized with `rayon`
- **Node script:** `node-quirc 2.3.0`, `@votingworks/qrdetect 1.0.1`, `zedbar 0.2.1` (WASM) via `sharp` for image loading
- Cropping logic replicated from `detect.rs:get_detection_areas` — `crop_size = width/4`, bottom-left then top-right corners
- Timing covers only the detection call, not image loading or cropping

Issue #8012, opened 2026-02-19
VxDesign: Revert special-case image resizing logic for Ashland after ballots are finalized


kofi-q:
To help [reduce the sheet count](https://votingworks.slack.com/archives/C0AA3ESLM4Y/p1771448566151329) for an Ashland ballot, we're adding special-case logic to resize the images for one particularly image-heavy contest. Once everything's settled for this ballot production run, we can revert that change. 
Longer-term fix: https://github.com/votingworks/vxsuite/issues/8011

PR #8014, opened 2026-02-19
VxDesign: Add temporary special case to resize specific contest images

kofi-q:
## Overview

https://votingworks.slack.com/archives/C0AA3ESLM4Y/p1771448566151329

Temporarily adding a special-cased image resize for one [`Ashland, NH` contest](https://design.voting.works/elections/g4usy39oo1fw/contests/h9fzgh2orsim), to help reduce the sheet count for the related ballot style. In order to get the images to stack in pairs on each line, we need to shrink the images to 50% of the available width. Doing this offline and re-uploading requires shrinking them relative to a 96dpi resolution (due to current assumptions in our image upload handling), which causes a bit too much resolution loss.

Will be reverting this after the ballot production run in favour of a broader fix: https://github.com/votingworks/vxsuite/issues/8012

## Demo Video or Screenshot

| Before | After |
| - | - |
| before | after |

## Testing Plan
- Manual

## Checklist

### DIFF
--- a/libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
+++ b/libs/hmpb/src/ballot_templates/nh_ballot_template.tsx
@@ -641,6 +641,16 @@ function BallotMeasureContest({
   compact?: boolean;
   continuesOnNextPage?: boolean;
 }) {
+  // [TODO] Temporary: to be removed after March ballot production run.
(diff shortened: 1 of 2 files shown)


arsalansufi (approved):
LGTM!

PR #8038, opened 2026-02-26
VxDesign: Bring back hiding of timing marks on sample ballots but ensure bubble positions aren't changed

arsalansufi:
## Overview

We recently hid timing marks on sample ballots in https://github.com/votingworks/vxsuite/pull/7937. But then identified an issue that required us to revert that change in https://github.com/votingworks/vxsuite/pull/7988. See the revert PR for more details on the issue.

This PR brings back the hiding of timing marks on sample ballots, reverting the revert and making additional changes on top to ensure that bubble positions for sample ballots are identical to other ballots. This PR also brings back the safeguard that has historically protected against such divergences.

## Demo Video or Screenshot

You can see in the before image the way that we're using the space typically consumed by timing marks whereas in the after image we're keeping that space unused as if timing marks were there.

before-after

## Testing Plan

- [x] Updated automated tests
- [x] Tested manually

## Checklist

### DIFF
--- a/libs/hmpb/src/render_ballot.test.tsx
+++ b/libs/hmpb/src/render_ballot.test.tsx
@@ -27,7 +27,6 @@ import {
 } from 'node-html-parser';
 import {
   allBaseBallotProps,
-  layOutBallotsAndCreateElectionDefinition,
   layOutMinimalBallotsToCreateElectionDefinition,
   renderBallotTemplate,
 } from './render_ballot';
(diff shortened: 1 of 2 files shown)


libs/hmpb/src/render_ballot.test.tsx
```diff
@@ -126,6 +127,33 @@ test('layOutMinimalBallotsToCreateElectionDefinition', async () => {
   expect(electionDefinition).toEqual(fixtureElectionDefinition);
 });
 
+test('sample ballots, despite not having timing marks, have the same bubble positions as official and test ballots', async () => {
```
arsalansufi: Used the Claude mutation-test skill to ensure that this will now fail if sample ballot layout diverges from official ballot layout!

<table>
Image
</table>

Note that this test does assume that the layout safeguard is applied to sample ballots, which it for some time wasn't. Spent a bit of time trying to set up a test that makes a sample ballot have a different layout and ensures that we error, but it's a bit tricky to do so without changing source code. Timeboxed and decided to call it.

PR #8055, opened 2026-03-03
VxDesign: Fix PartiesScreen test flake

kofi-q:
## Overview

[This test](https://app.circleci.com/pipelines/github/votingworks/vxsuite/24311/workflows/93ae9ff5-7e7f-4d10-b91a-025d3769115e/jobs/1095528) was flaking in a similar way to the one that was fixed in https://github.com/votingworks/vxsuite/pull/7946 - looks like it's due to an unrelated query in a nested component (audio button) that gets rendered in later ticks after the top-level screen, which, in a slow-enough CI run, doesn't get kicked off before the test is done (for short tests with only one wait on the initial load). 

Repro'd by adding a synthetic rendering delay in the nested component to mimic slow CI - fixed more globally by switching the mock for the transitive API dependency to an optional repeated. The `audio editing` test is explicitly checking the API interaction with the button's rendering.

## Checklist

### DIFF
--- a/apps/design/frontend/src/parties_screen.test.tsx
+++ b/apps/design/frontend/src/parties_screen.test.tsx
@@ -19,7 +19,7 @@ import {
   user,
 } from '../test/api_helpers';
 import { electionInfoFromElection } from '../test/fixtures';
-import { render, screen, waitFor, within } from '../test/react_testing_library';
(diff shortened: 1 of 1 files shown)


jonahkagan (approved):
Thanks for digging into this again

PR #8066, opened 2026-03-05
docs: spec for dev-dock mock file isolation

eventualbuddha:
## Spec

[specs/0002-dev-dock-isolation.md](https://github.com/votingworks/vxsuite/blob/spec/dev-dock-mock-isolation/specs/0002-dev-dock-isolation.md)

## Summary

Proposes consolidating all dev-dock mock file storage (USB drive, smart card, printers, batch scanner images, dev-dock state) under a single `<repo-root>/.vx-dev-dock/<NODE_ENV>/` directory tree. The `NODE_ENV` subdirectory provides isolation between running dev servers (`development`), unit tests (`test`), and Playwright integration tests (`production`), eliminating worktree collisions and test/dev cross-contamination.

### DIFF
--- a/specs/0002-dev-dock-isolation.md
+++ b/specs/0002-dev-dock-isolation.md
@@ -63,7 +63,7 @@ Unify all dev-dock mock file storage under a single directory tree rooted in the
 repo:
 
 ```
-<repo-root>/.vx-dev-dock/<NODE_ENV>/
+<repo-root>/.mock-state/<NODE_ENV>/
   mock-file-card.json       # smart card
   usb-drive/                # USB drive state + data
   hp-printer/               # HP laser printer state (state.json)
@@ -81,7 +81,7 @@ The `NODE_ENV` subdirectory provides isolation based on environment:
 - `production` — used by integration tests (Playwright runs apps in production
   mode but sets `IS_INTEGRATION_TEST=true`)
 
-The entire `.vx-dev-dock/` directory is ignored by Git.
+The entire `.mock-state/` directory is ignored by Git.
 
 ### Implementation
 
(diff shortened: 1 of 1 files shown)


specs/0002-dev-dock-isolation.md
```diff
@@ -0,0 +1,152 @@
...
+`@votingworks/utils` computes the path. Callers pass their own `REPO_ROOT`
+constant (derived from `__dirname`) rather than having the function walk up the
+directory tree, keeping the logic simple and explicit.
+
+```typescript
+// libs/utils/src/dev_dock.ts
+export function getDevDockMockRootDir(repoRoot: string): string {
+  const nodeEnv = process.env['NODE_ENV'] ?? 'development';
+  return join(repoRoot, '.vx-dev-dock', nodeEnv);
+}
```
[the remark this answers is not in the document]
eventualbuddha: I did not intend for this to be verbatim copy-pasted during implementation, and this is pretty clear from context where `join` comes from.

specs/0002-dev-dock-isolation.md
```diff
@@ -0,0 +1,160 @@
...
+   inside package source trees is surprising and pollutes the working directory
+   with runtime state. The mix of `/tmp/`, `~/`, and in-source paths made the
+   system hard to reason about.
+
+4. **Integration tests running in `production` mode.** Playwright integration
+   tests run the app with `NODE_ENV=production`, so the mock USB drive glob
+   patterns did not include the dev mock path, causing export operations to fail
+   or silently not find the mock drive.
+
+## Proposal
```
jonahkagan: I like this proposal a lot! I think it will be a lot clearer

Makes me also think it would be nice to group all the mock packages together somehow. E.g. instead of `libs/` put them in `libs/mocks` or something

specs/0002-dev-dock-isolation.md
```diff
@@ -0,0 +1,160 @@
...
+
+Unify all dev-dock mock file storage under a single directory tree rooted in the
+repo:
+
+```
+<repo-root>/.vx-dev-dock/<NODE_ENV>/
```
jonahkagan: I kind of conceptualize the mocks as separate from dev dock since they are sometimes used in other contexts (e.g. tests), but this is a quibble about the naming of this storage directory, and I don't really have much better ideas (the best I can think of is `.mock-storage`)

PR #8113, opened 2026-03-16
fix(pdi-scanner): make startup wait for a usable scanner

eventualbuddha:
## Overview

Make `pdi-scanner` explicitly wait for a usable command channel after `connect()` by splitting readiness priming from scan initialization and adding a short bounded handshake retry loop. Testing showed that sending `ESC-K` to enable CRC checks fixed the issue described below. This makes some sense since the "initial commands" we'd send that were modeled after what the official PDI scanner library does included that command.

This works around a host-specific startup issue I saw on bare metal, where the scanner would enumerate and let me claim interface 0 successfully, but the first command after connect could still time out. In the VxSuite VM, the same startup path was already working reliably, though we'd still put some connection retry logic into `pdictl` to deal with a "first connection fails" issue we'd seen before which may or may not be the same bug. The diagnostics pointed to readiness rather than raw USB access, so this PR makes the client actively confirm the command channel is usable before continuing.

The API split keeps the behavior explicit:
- `wait_until_ready()` means "the command channel is usable"
- `initialize_scanning()` means "perform scan-specific setup like feeder mode and calibration-table fetches"

## Demo Video or Screenshot

N/A

## Testing Plan

- [x] Manually tested with the various `libs/pdi-scanner/examples` binaries both in the host and the VM.
- [x] Manually tested running against VxScan, disconnecting and reconnecting the PDI scanner.

## Diagnostic Testing

- `cd libs/pdi-scanner && cargo test --quiet`
- Bare-metal host diagnostics showed:
  - USB enumeration and interface claiming worked on the host
  - `GetTestString` alone timed out immediately after connect
  - `ESC-K` alone produced no response
  - `ESC-K` followed by `GetTestString` succeeded and later register reads succeeded
  - fixed delays were inconsistent, which motivated probe/retry instead of a single sleep
- The updated startup behavior was then exercised against both the bare-metal host and the VxSuite VM

### DIFF
--- a/libs/pdi-scanner/src/rust/main.rs
+++ b/libs/pdi-scanner/src/rust/main.rs
@@ -217,14 +217,11 @@ fn error_to_code_and_message(error: &Error) -> (ErrorCode, Option<String>) {
 async fn initialize_connected_scanner(
     mut client: Client<Scanner>,
 ) -> pdi_scanner::Result<(Client<Scanner>, ImageCalibrationTables)> {
-    match timeout(Duration::from_millis(500), client.wait_until_ready()).await {
-        Ok(Ok(())) => {}
-        Ok(Err(error)) => return Err(error),
-        Err(_) => match timeout(Duration::from_secs(3), client.wait_until_ready()).await {
-            Ok(Ok(())) => {}
-            Ok(Err(error)) => return Err(error),
-            Err(_) => return Err(Error::RecvTimeout),
-        },
+    if timeout(Duration::from_secs(3), client.wait_until_ready())
+        .await
+        .is_err()
+    {
+        return Err(Error::RecvTimeout);
     }
 
     let calibration_tables =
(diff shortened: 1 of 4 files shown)


libs/pdi-scanner/src/rust/main.rs
```diff
@@ -214,6 +214,29 @@ fn error_to_code_and_message(error: &Error) -> (ErrorCode, Option<String>) {
...
+) -> pdi_scanner::Result<(Client<Scanner>, ImageCalibrationTables)> {
+    match timeout(Duration::from_millis(500), client.wait_until_ready()).await {
+        Ok(Ok(())) => {}
+        Ok(Err(error)) => return Err(error),
+        Err(_) => match timeout(Duration::from_secs(3), client.wait_until_ready()).await {
+            Ok(Ok(())) => {}
+            Ok(Err(error)) => return Err(error),
+            Err(_) => return Err(Error::RecvTimeout),
+        },
+    }
```
jonahkagan: It seems a little odd to have `wait_until_ready` have retry logic but then also have another layer of retry logic where we call `wait_until_ready` twice. I think it would be more straightforward to have all of the retry logic that we deem necessary within `wait_until_ready`

eventualbuddha: Agreed. Changed in 080b932d47a0827fbd5c8a68effe80127d570d13.

Issue #7870, opened 2026-02-01
Polling Place Support (Parent Issue)


adghayes:
See https://docs.google.com/document/d/17LG5gI5FXdBzC3kZSaIZeECS8xTUw-0Ljg5Tvbkfe8M/edit?tab=t.aqsx8n1igw9u.

adghayes:
Closed via #8750. Woohoo!

PR #8142, opened 2026-03-18
Add backend DB/API scaffolding for reading/writing Polling Places across VxSuite

kofi-q:
## Overview

Grouping together the more repetitive changes for get/set backend APIs/DB methods across our precinct apps before diving into app-specific changes. Adding
- A `polling_place_id` column to the `election` tables, along with get/set store methods
- Server API get/set methods - in most cases, the existing `PrecinctSelection` is included in the `ElectionState` or `PrecinctScannerConfig` bundle in the get path, so adding the polling place ID there.
- Client API hooks

Not replacing `PrecinctSelection` yet - will replace usage app-by-app to keep things manageable.

## TODO
- Auto-select polling place during configuration if there's only one in the election
- Add a polling-place-specific log event (currently reusing the existing `PrecinctConfigurationChanged` one, but might wait till everything's moved over and just rename that event)

## Testing Plan
- Basic get/set smoke tests mirroring the existing ones for `PrecinctSelection`

### DIFF
--- a/apps/mark-scan/frontend/src/api.ts
+++ b/apps/mark-scan/frontend/src/api.ts
@@ -400,11 +400,25 @@ export const setTestMode = {
   },
 } as const;
 
+// [TODO] Remove after migration to polling places.
 export const setPrecinctSelection = {
   useMutation() {
     const apiClient = useApiClient();
     const queryClient = useQueryClient();
     return useMutation(apiClient.setPrecinctSelection, {
+      async onSuccess() {
+        await queryClient.invalidateQueries(getElectionState.queryKey());
+      },
+    });
+  },
+} as const;
+
+/* istanbul ignore next - WIP - @preserve */
+export const setPollingPlaceId = {
+  useMutation() {
+    const apiClient = useApiClient();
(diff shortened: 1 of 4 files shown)


apps/central-scan/backend/schema.sql
```diff
@@ -4,7 +4,7 @@ create table election (
   election_data text not null,
   election_package_hash text not null,
   jurisdiction text not null,
-  precinct_selection text,
+  precinct_selection text, -- [TODO] Remove after migration to polling places
```
carolinemodic: why does central scan not get polling place id?

kofi-q: I noticed the precinct selection here is effectively a no-op - it always gets [set to `AllPrecincts`](https://github.com/votingworks/vxsuite/blob/90d696d/apps/central-scan/backend/src/importer.ts#L87-L88) and then never queried. CentralScan just hardcodes `AllPrecincts` as input to the interpreter, so where I'm leaning is to update the interpreter to take in a set of precinct/split IDs to use for its validation instead, then CentralScan would just always send the full set from the election

PR #8152, opened 2026-03-19
[VxAdmin] Flesh out client app screens and flows

carolinemodic:
## Overview
Sync's the election on VxAdminClient from the host and uses that for election manager and poll worker auth. 

Adds settings/adjudication/diagnostics screens to each of sysadmin, em and pollworker as appropriate for the role along with basic diagnostics and setting (SHV, format usb, etc.) functionality. 

"Start Adjudication" button is just a placeholder and does nothing. 

## Demo Video or Screenshot

https://github.com/user-attachments/assets/0ad3b1e4-0830-4d9f-bc10-a3c4fed3df0e

## Testing Plan
See above video

## Checklist


libs/ui/src/diagnostics/admin_readiness_report.tsx
```diff
@@ -35,6 +35,27 @@ export function AdminReadinessReportContents(
   );
 }
 
+type ClientReportContentsProps = ConfigurationSectionProps &
```
nikhilb4a: Wondering if this Client version is needed, because I imagine the Clients will also be able to connect to printers for items like this readiness report? Maybe adjudication reports

apps/admin/frontend/src/client/components/client_navigation_screen.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { useContext } from 'react';
```
nikhilb4a: I'm wondering if we should have a separate ClientNavigationScreen or if we should just make the existing one more flexible, as I think they are identical aside from the nav items, which already vary by user type, now they also just vary by machine type. What do you think about that? I can see the argument for the separate screens too, if it's to keep things simple and that it's not too much duplicated code.

PR #8191, opened 2026-03-27
Fix duplicate Write-In rows in grouped tally reports

adghayes:
## Summary

Fixes a bug where tally reports could show two indistinguishable "Write-In" rows for partially adjudicated contests when results are grouped (by party, precinct, batch, scanner, or voting method).

**Root cause:** `getEmptyElectionResults` initializes every write-in contest with a `GENERIC_WRITE_IN_CANDIDATE` entry (name: `"Write-In"`). When write-in adjudication data exists for one group but not another, `modifyElectionResultsWithWriteInSummary` was only called for groups that had write-in records. Groups without any write-ins kept stale `"Write-In"` entries. When results were later combined across groups (e.g. via `combineElectionResults` / `combineCandidateContestResults`), if a stale group was processed before the correctly-modified group, the stale `"Write-In"` name persisted — producing two rows with the same name.

The bug is most likely to surface in primary elections (where party grouping is always on and it's common for one party to have write-ins while another doesn't), but it applies to any grouped tabulation where at least one group has no write-in CVRs at all.

**Fix:** Always call `modifyElectionResultsWithWriteInSummary` for every result group, falling back to an empty write-in summary when no write-in data exists. This ensures stale `GENERIC_WRITE_IN_CANDIDATE` entries are cleaned up in all groups — dropped when there are no pending write-ins, or replaced with `PENDING_WRITE_IN_CANDIDATE` when there are.

## Trade-offs

- Slightly more work per tabulation (runs `modifyElectionResultsWithWriteInSummary` on groups that previously skipped it), but the operation is lightweight — it just iterates contest write-in summaries.
- No behavioral change for ungrouped elections or cases where all groups have write-in data.

## Test plan

- [x] Wrote a failing test reproducing the bug (write-ins on one party only in a primary), verified it failed before the fix, then reverted to keep test suite clean
- [x] All existing tests in `admin/backend` pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/admin/backend/src/app.tally_report_data.test.ts
+++ b/apps/admin/backend/src/app.tally_report_data.test.ts
@@ -902,15 +902,13 @@ test('primary, partial write-in adjudication uses correct unadjudicated label', 
   );
   mockElectionManagerAuth(auth, election);
 
-  // Use Fish party contest (party "1") for write-ins. The Mammal party (party
-  // "0") is initialized first in the expected groups, so when combining results
-  // across parties, the Mammal group's stale generic "Write-In" entry for this
-  // contest (from getEmptyElectionResults) gets processed first, overriding
-  // the correct "Unadjudicated Write-In" name from the Fish group.
+  // Regression test for https://github.com/votingworks/vxsuite/pull/8191
+  // When one party has write-ins and another doesn't, the party without
+  // write-ins must not leave stale "Write-In" entries that produce duplicate
+  // rows on the tally report.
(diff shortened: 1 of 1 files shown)


adghayes:
One of the core problems that maybe we should solve later is that we have two constants - a "Pending Write-In Candidate" and a "Generic Write-In Candidate". They have different names, but the same IDs. I think we should probably have them be fully different, which would avoid this issue and surface things in a more obvious way.

PR #8203, opened 2026-03-30
Add system setting for disabling screen reader audio on VxScan and avoid "Your ballot was counted" getting cut off

arsalansufi:
## Overview

A final batch of changes to VxScan audio logic for v4.0.5 🔈 🎧

First, this PR adds a system setting for disabling screen reader audio on VxScan (and all associated under-the-hood logic). This logic has perf and user experience implications, and at least on my old build, when scanning ballots in quick succession, audio has a tendency to spill over. For example, headphone audio will spill through the speaker (the speaker will partially read out "Your ballot was counted") and vice versa (the success chime will go through the headphones rather than the speaker). The delays added in https://github.com/votingworks/vxsuite/pull/8189 (and adjusted in https://github.com/votingworks/vxsuite/pull/8194) don't always cut it. We'll apply this system setting for customers that don't intend to use headphones with VxScan (likely all customers).

Second, assuming one is using headphones, the "Your ballot was counted" readout is frequently interrupted by the "Insert your ballot" readout when we return to that screen. This doesn't happen so much on our dev laptops, but it's consistent on my actual build running a production image. I went with a super simple fix: disabling the readout on the "Insert your ballot" screen. A user would put on headphones after that screen has been on for some time already so wouldn't hear "Insert your ballot" anyway. And the key audio to hear is adjudication warnings and "Your ballot was counted". Main motivation for this change is avoiding a cert discrepancy.

## Testing Plan

- [x] Added automated tests
- [x] Tested on actual hardware by editing a QA image

## Checklist

- [ ] ~I have added the "user-facing-change" label to this PR, if relevant, to automate an announcement in #machine-product-updates.~

### DIFF
--- a/apps/scan/frontend/src/app.test.tsx
+++ b/apps/scan/frontend/src/app.test.tsx
@@ -1563,6 +1563,46 @@ test('voter help button hidden when relevant system setting is set', async () =>
   ).not.toBeInTheDocument();
 });
 
+test.each([
+  {
+    description: 'present when screen reader audio is enabled',
+    precinctScanDisableScreenReaderAudio: false,
+    isAudioTabPresent: true,
+  },
+  {
+    description: 'not present when screen reader audio is disabled',
+    precinctScanDisableScreenReaderAudio: true,
+    isAudioTabPresent: false,
+  },
+])(
+  'voter settings audio tab - $description',
+  async ({ precinctScanDisableScreenReaderAudio, isAudioTabPresent }) => {
+    apiMock.expectGetConfig({
+      systemSettings: {
+        ...DEFAULT_SYSTEM_SETTINGS,
+        precinctScanDisableScreenReaderAudio,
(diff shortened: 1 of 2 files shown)


apps/scan/backend/src/server.ts
```diff
@@ -69,14 +69,15 @@ export async function start({
   workspace.clearUploads();
 
   const audioCard = await AudioCard.default(NODE_ENV, logger);
-
-  // Screen reader volume levels are calibrated against a maximum system
-  // volume setting:
-  await audioCard.useHeadphones();
   await audioCard.setVolume(100);
```
kofi-q: Thinking this should move to after the `setIsScreenReaderEnabled` call, so that affects the selected output. Should be a no-op in practice regardless anyway, outside of maybe the first boot, but the ordering might be clearer that way

PR #8217, opened 2026-03-30
fix: remove detectDevices USB listeners when the server closes

eventualbuddha:
## Overview

`detectDevices` adds `attach`/`detach` listeners to the global `usb` singleton but never removes them. Every `start()` call adds another pair, so any test file that starts a server repeatedly blows past Node's default listener limit.

`detectDevices` now returns a cleanup function, and each app's `start()` registers it on the server's `close` event. Every shutdown path that already exists — `withApp`'s `finally` in central-scan, `server.close()` in the admin and mark-scan tests — now releases the listeners, so no call sites and no `StartOptions` had to change.

The mocked `listen()` in the central-scan and scan server tests has to return something `close` can be registered on, hence the `EventEmitter` return values.

## Demo Video or Screenshot

Before, on `main`:

```
$ cd apps/central-scan/backend && pnpm test:run src/app.grout.test.ts
(node) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 attach listeners added to [EventEmitter]. MaxListeners is 10.
(node) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 detach listeners added to [EventEmitter]. MaxListeners is 10.

 Test Files  1 passed (1)
      Tests  17 passed (17)
```

After:

```
$ cd apps/central-scan/backend && pnpm test:coverage

 Test Files  17 passed (17)
      Tests  125 passed (125)
```

## Testing Plan

- `libs/backend` — `pnpm test:coverage`: 359 tests pass. `detect_devices.test.ts` gains a case asserting that no further events are logged after cleanup.
- `apps/central-scan/backend` — `pnpm test:coverage`: 125 tests pass, and the full suite is now free of `MaxListenersExceededWarning`.
- `apps/scan/backend`, `apps/admin/backend`, `apps/mark-scan/backend` — `pnpm test:run src/server.test.ts` passes in each.
- `pnpm lint` passes for all five packages.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/admin/backend/src/server.ts
+++ b/apps/admin/backend/src/server.ts
@@ -111,7 +111,7 @@ export async function start(options: StartOptions = {}): Promise<Server> {
   } = options;
 
   debug('starting server...');
-  detectDevices({ logger: baseLogger });
+  const stopDetectingDevices = detectDevices({ logger: baseLogger });
 
   const workspacePath =
     options.workspacePath ?? resolveWorkspacePath(baseLogger);
@@ -274,5 +274,6 @@ export async function start(options: StartOptions = {}): Promise<Server> {
   // (USB_DRIVE_CHANGE_LONG_POLL_TIMEOUT_MS); no socket-level timeout is needed.
   server.timeout = 0;
 
+  server.on('close', stopDetectingDevices);
   return server;
 }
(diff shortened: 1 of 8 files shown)


eventualbuddha:
Reworked and rebased onto current `main`. The earlier `AbortSignal` approach is gone: it added a `signal` option to four apps' `StartOptions` that only one caller ever passed, and the abort handlers needed `/* istanbul ignore next */` because nothing exercised them. Registering the cleanup on the server's `close` event covers every shutdown path that already existed, so no call sites and no options changed, and no coverage suppressions were needed.

Resolved the review comments that the rework made moot — the ones about cleanup only running on abort, the required-vs-optional `signal` field, the `listen()` mock returning `undefined`, and the `detect_devices` test ignoring its cleanup function (it now calls it).

arsalansufi (approved):

PR #8219, opened 2026-03-31
Add (feature-flagged) polling place readiness report section for precinct apps

kofi-q:
## Overview

This is part of the preparation for replacing references to `PrecinctSelection` across the repo with a polling place/ID.

Splitting up the `ConfigurationSection` readiness report component into smaller composable units, to enable conditionally rendering the precinct selection section or a new polling place section, based on feature flag. Also solves the issue of the `ConfigurationSection` having grown over time into a bit of a tangle of conditional renders.

### TODO
- Added a `pollingPlaceTypeName` util - will be following up to replace existing code where we're duplicating that logic.
- There was a missing test file for the `Mark` readiness report - will follow up separately to backfill those.
- The UI for the report screens could use some polish, but will resist the urge for now

### With new Polling Place section (no selection):

readiness-no-polling-place

### With Polling Place selected:

readiness-polling-place-selected

## Testing Plan
- Existing tests as regression guard, new unit tests for added section/conditional render
- Spot checks 

## Checklist

### DIFF
--- a/libs/types/src/polling_places.test.ts
+++ b/libs/types/src/polling_places.test.ts
@@ -21,6 +21,7 @@ import {
   pollingPlaceMembers,
   pollingPlacePrecinctIds,
   pollingPlacesGenerateFromPrecincts,
+  pollingPlaceTypeName,
 } from './polling_places';
 
 test('pollingPlaceBallotStyles', () => {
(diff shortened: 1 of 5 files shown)


libs/ui/src/diagnostics/mark_readiness_report.tsx
```diff
@@ -41,19 +51,38 @@ interface ReportContentsProps
...
+          <PollingPlaceSection
+            election={election}
+            pollingPlaceId={pollingPlaceId}
+          />
+        ) : (
+          <PrecinctSelectionSection
+            election={election}
+            precinctSelection={precinctSelection}
+          />
+        )}
```
kshen0: This branching shows up in a few places. I'm guessing it's not worth a wrapper since we plan to deprecate precinct selections across the board, but just want to confirm that we're not planning on keeping precinct selections at all?

PR #8223, opened 2026-04-01
HWTA(all): Various updates for the next round of hardware testing

arsalansufi:
## Overview

Batching a number of updates for the next round of hardware testing! Changes include the following:

_All HWTAs_

* Crash when the card reader is presumed to be unhealthy to trigger the HWTA auto-restart logic. Reminder that, during HW testing, we have to recover without human intervention. Hitting the power button is considered human intervention, so we make use of automatic restarts on crash. In the HWTA, card reader errors aren't guaranteed to cause a crash. This PR changes that, using a best guess heuristic. More context on the motivation for this change in: https://votingworks.slack.com/archives/C0AA2KHHC6R/p1774817304432839
* Decrease CPU metrics polling interval from 1s to 15s. This metrics polling is relatively cheap, but I just want to
keep CPU activity to a minimum as we go into this next round of HW testing.

_VxMark/VxScan HWTAs_

* Default the headphone volume to 40 since the HWTA doesn't have the same logic as our main app to calibrate relative to a default of 100 (and 100 without any calibration will hurt most ears). This was already in place on the VxScan HWTA. Just added to the VxMark HWTA for consistency.
* Then on both machines, when the headphone calibration modal is opened, up the volume to 100 for correct calibration. Drop back down to 40 after leaving the modal. Came up as a challenge during recent headphone calibration efforts.

_VxScan HWTA_

* Avoid a jarring experience when there's a PDI scanner error with two changes. First, delay within the connection retry loop to avoid spamming the scanner. Second, put full error details behind a button/modal rather than displaying them all on screen. See the jarring experience here hah: https://votingworks.slack.com/archives/C09FVEJJ8UX/p1773944975364059?thread_ts=1773433276.668609&cid=C09FVEJJ8UX

  New experience below (see the details button under **Scanner**):

| | |
| - | - |
| details-0 | details-1 |

* Address a memory leak on the VxScan HWTA that results in a crash after a day or so of scanning.
* Also clean up images on disk to avoid running out of disk space, which has happened at least once in the wild.

## Testing Plan

- [x] Lots of manual testing. Will be doing more before shipping to SLI.

## Checklist

- [ ] ~I have added the "user-facing-change" label to this PR, if relevant, to automate an announcement in #machine-product-updates.~

### DIFF
--- a/apps/mark-scan/frontend/src/electrical_testing/app_root.tsx
+++ b/apps/mark-scan/frontend/src/electrical_testing/app_root.tsx
@@ -6,7 +6,7 @@ import {
   InputControls,
   P,
 } from '@votingworks/ui';
-import React, { useState } from 'react';
+import { useState } from 'react';
 import useInterval from 'use-interval';
 import {
   getCpuMetrics,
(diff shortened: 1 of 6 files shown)


libs/backend/src/card_reader_error_tracker.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  assertHealthy(): void {
+    assert(
+      this.consecutiveErrorsAfterNoCardReaderEvent <
+        CONSECUTIVE_ERRORS_AFTER_NO_CARD_READER_EVENT_BEFORE_CRASHING,
+      'Card reader registered as disconnected and ' +
+        `${this.consecutiveErrorsAfterNoCardReaderEvent} subsequent reads failed`
+    );
+  }
```
arsalansufi: Happy to talk through how I landed on this heuristic. Informed by this video and discussion with JD: https://votingworks.slack.com/archives/C0AA2KHHC6R/p1774817304432839

I initially wanted to just check for consecutive `no_card_reader` errors but needed to broaden when I discovered that it was possible to get a mix of `no_card_reader` and `no_card` (despite a card being inserted).

PR #8227, opened 2026-04-01
[libs/ui] Add a PollingPlacePicker as a ChangePrecinctButton replacement

kofi-q:
## Overview

Adding a polling place picker component to be later used in the Mark, MarkScan, Scan apps. It will replace the existing `PrecinctSelection`-based `ChangePrecinctButton` and follows the same basic flows, with some additions for polling places:
- In the `default` mode (used when polling places can be switched safely without needing to rest polls), it's rendered as a single-select dropdown with the configured polling places listed out.
  - Since polling places can have the same name but different types, in theory, we include a type label when there's more than one type of polling place present.
- In the `confirmation_required` mode, it's rendered as a modal button, with the dropdown rendered in-modal.
- Clients can optionally filter the polling places down to specific types, which solves the VxScan use case where users can switch between "Early Voting" mode (only `early_voting` places shown) and "Election Day" mode (both `election_day` and `absentee` places shown).
  - VxScan will also be explicitly omitting the type labels, even when showing both `election_day` and `absentee` - relevant product discussion [here](https://votingworks.slack.com/archives/CEL6D3GAD/p1774992215069989?thread_ts=1774991244.872799&cid=CEL6D3GAD).

## Demo Video or Screenshot

https://github.com/user-attachments/assets/c40e847e-f273-4f90-9fb9-b19e88fd9710

### Sample in-context usage (not yet implemented):

mark-polling-place-select

## Testing Plan
- Unit tests, mostly mirroring the older `ChangePrecinctButton` tests, with additional cases for the polling-place-specific interactions.

### DIFF
--- a/libs/ui/src/polling_place_picker.test.tsx
+++ b/libs/ui/src/polling_place_picker.test.tsx
@@ -99,30 +99,6 @@ describe('dropdown contents', () => {
       ]);
     });
 
-    test(`[${spec.mode} mode] omits type labels if 'noTypeLabels' is specified`, () => {
-      render(
-        <PollingPlacePicker
-          mode={spec.mode}
-          noTypeLabels
-          places={[earlyVoting1, electionDay1]}
-          selectedId={electionDay1.id}
-          selectPlace={vi.fn(() => Promise.resolve())}
-        />
-      );
-
-      spec.prep?.();
-
-      const dropdown = getDropdown();
-      const selectEl = dropdown.closest('.search-select');
-      expect(selectEl).toHaveTextContent(electionDay1.name);
-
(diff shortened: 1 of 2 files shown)


kshen0:
> VxScan will also be explicitly omitting the type labels, even when showing both election_day and absentee - relevant product discussion [here](https://votingworks.slack.com/archives/CEL6D3GAD/p1774992215069989?thread_ts=1774991244.872799&cid=CEL6D3GAD).

How does this work with polling places that share the same name but different type? If we have Place A and Place B for both election day and absentee, in VxScan will it look like:

* Place A
* Place B
* Place A
* Place B

Where the first 2 are election day and last 2 are absentee?

Issue #7823, opened 2026-01-23
Open Primary Support


adghayes:
See [Michigan Open Primary Product Requirements](https://docs.google.com/document/d/1Pouky21Co-1clTvVw9edTQRUka-PRQNXPvacpmUuFTY/edit?tab=t.nl4y4cykvk33) for details.

arsalansufi:
Just one sub-issue remaining that isn't demo blocking so moving to the v4.1.0 initial submission milestone 👏

PR #8229, opened 2026-04-01
Add isOpenPrimary() helper and open primary fixture

jonahkagan:
🤖 Co-authored with Claude Code

## Overview

Task: #7823

Foundation for open primary support. Adds an `isOpenPrimary()` helper to detect open primaries (primary elections where ballot styles have no `partyId`), schema validation ensuring primaries are uniformly closed or open, and a generic open primary election fixture.

Part of the MI open primaries project — this is the first in a series of PRs.

## Demo Video or Screenshot

N/A — no UI changes.

## Testing Plan

- Unit tests for `isOpenPrimary()` covering general, closed primary, and open primary elections
- Unit test for schema validation rejecting mixed ballot styles
- Fixture integration test verifying `isOpenPrimary()` returns `true` for the new fixture

## Checklist

- [ ] I have added the "user-facing-change" label to this PR, if relevant, to automate an announcement in #machine-product-updates.

### DIFF
--- a/libs/types/src/election.ts
+++ b/libs/types/src/election.ts
@@ -913,6 +913,24 @@ export const ElectionSchema: z.ZodSchema<Election> = z
         }
       }
     }
+
+    if (election.type === 'primary') {
+      const hasBallotStyleWithPartyId = election.ballotStyles.some(
+        (bs) => bs.partyId
+      );
+      const hasBallotStyleWithoutPartyId = election.ballotStyles.some(
+        (bs) => !bs.partyId
+      );
+      if (hasBallotStyleWithPartyId && hasBallotStyleWithoutPartyId) {
+        ctx.issues.push({
+          code: 'custom',
+          path: ['ballotStyles'],
+          message:
+            'Primary election ballot styles must either all have a partyId (closed primary) or all omit partyId (open primary).',
+          input: election,
+        });
+      }
+    }
   });
 export type OptionalElection = Optional<Election>;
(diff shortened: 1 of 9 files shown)


libs/types/src/election.ts
```diff
@@ -913,6 +913,24 @@ export const ElectionSchema: z.ZodSchema<Election> = z
         }
       }
     }
+
+    if (election.type === 'primary') {
```
carolinemodic, line 917: was about to comment on the first commit about this concern but I see you have already addressed haha

carolinemodic (approved):
It seems like the new fixture is a "base" fixture, i.e. it does not have gridLayouts, etc. I assume partly because we might not know how to generate ballots / grid layouts for these elections yet. Are you planning to add this fixture to the generate-election-packages flows in libs/fixture-generators so that the version with gridLayouts and an election package are auto-generated or do we not actually need that for the testing this fixture is being set up for?

jonahkagan:
> It seems like the new fixture is a "base" fixture, i.e. it does not have gridLayouts, etc. I assume partly because we might not know how to generate ballots / grid layouts for these elections yet. Are you planning to add this fixture to the generate-election-packages flows in libs/fixture-generators so that the version with gridLayouts and an election package are auto-generated or do we not actually need that for the testing this fixture is being set up for?

Next PR coming up is for a ballot template. My plan is to generate HMPB fixtures (ballots + election.json) based on this base fixture. Do you think it's also worth adding to fixture-generators? I don't really have a sense of where the fixtures will be used yet, so might just be something for me to keep an eye on.

Issue #8204, opened 2026-03-30
Adjudication: Update contest highlights to include contest header/title


arsalansufi:
Contest highlights currently only include contest options and not contest headers/titles. Note that we need to consider this for all our templates (present and future).

Now that we've tagged and branched off for v4.0.5 (and very few if any changes remain for the v4.0 cert campaign), I'm comfortable with us making a breaking change for this update if necessary.

Issue #5650, opened 2024-12-03
nit: not sure if this is expected behavior or specific to this election package, but using the QA multi-lingual ballot package on VxCentralScan, I’m noticing that the contest highlights on the page with two columns seem to have the width as if it were three columns


arsalansufi:
nit: not sure if this is expected behavior or specific to this election package, but using the QA multi-lingual ballot package on VxCentralScan, I’m noticing that the contest highlights on the page with two columns seem to have the width as if it were three columns. see rightmost highlight

Issue automatically created from Slack feedback reported by drew: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1732325696856269

PR #8233, opened 2026-04-02
Spec: Improve contest and option highlight bounds for adjudication

jonahkagan:
## Overview

🤖 Co-authored with Claude Code

Tasks: #8204, #5650

Planning spec for fixing contest highlight bounds in the adjudication UI.
Contest bounds are currently derived from a single shared option-size
measurement, causing missing headers and wrong dimensions on mixed-column
ballots. Also affects write-in adjudication zoom accuracy.

Proposes measuring per-contest and per-option bounds directly during ballot
rendering, stored as grid coordinate rects. Also proposes restructuring grid
layout data into a sheet/contest/option hierarchy on `BallotStyle`.

Example image of existing highlights:
image

### DIFF
--- a/specs/0006-highlight-entire-contest-box.md
+++ b/specs/0006-highlight-entire-contest-box.md
@@ -1,0 +1,188 @@
+# Highlight Entire Contest Box in Adjudication
+
+**Author:** @jonahkagan
+
+**Status:** `planning`
+
+## Existing Discussion
+
+- [votingworks/vxsuite#8204](https://github.com/votingworks/vxsuite/issues/8204)
+  — contest highlights don't include the header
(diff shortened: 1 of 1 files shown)


eventualbuddha (approved):
This makes sense to me. I'd go with Option 1 as far as the compatibility question is concerned, though I'm not sure there's a huge difference between Options 1 & 2. We'll know when we no longer have customers on v4.0.x either way.

nikhilb4a (approved):
Looking forward to this, especially getting contest titles! Though, I am interested in seeing how full width BMs look in the UI, to Drew's point.

Agreed that option 1 seems like the best if time allows. Otherwise option 2 seems like a fine temporary fallback.

eventualbuddha:
Still relevant? Should this be merged?

jonahkagan:
Sure, we can merge

Issue #7874, opened 2026-02-01
VxScan: allow setting VxScan for a polling place


adghayes:
When set for a polling place, VxScan should accept any ballot style for any of the precincts that belong to the polling place.

PR #8235, opened 2026-04-02
[Scan] EM Screen: Show polling place picker when feature is enabled

kofi-q:
## Overview

https://github.com/votingworks/vxsuite/issues/7874

Currently gated by the `ENABLE_POLLING_PLACES` dev feature, while work's in progress.

When the feature is on:
- Replace the `ChangePrecinctButton` on the Election Manager screen with the `PollingPlacePicker`
- When switching ballot casting mode:
  - Clear any previous polling place selection, since they'll no longer be valid.
  - Reset the polls state, as we do when changing the precinct/polling place selection. The casting mode switch API is guarded by an assertion that we're either in the initial or paused states, so this essentially resets paused polls to initial.

Also moving the ballot casting mode toggle above the polling place picker, since the former now controls what's visible in the latter. Make the visual balance look really off, but I'll need to deal with it for now and revisit later to see if we can refine this a bit.

### TODO
- Now that we're clearing the polling place selection when switching casting mode, it's worth showing a confirmation modal like we do when switching precinct/polling place. Will tackle in a followup.
- Casting mode can currently be switched when ballots have been cast, if in the paused state - want to follow up and add a `ballotsCounted === 0` guard as well, assuming that matches product expectation

## Demo Video or Screenshot

https://github.com/user-attachments/assets/098860d5-27c4-4635-97e0-0cc3a759bc66

## Testing Plan
- Unit tests with the polling places feature enabled
- Manual spot checks

### DIFF
--- a/apps/scan/backend/src/app_config.test.ts
+++ b/apps/scan/backend/src/app_config.test.ts
@@ -14,6 +14,7 @@ import { assertDefined, err, ok } from '@votingworks/basics';
 import {
   mockElectionManagerUser,
   mockSessionExpiresAt,
+  suppressingConsoleOutput,
 } from '@votingworks/test-utils';
 import { mockElectionPackageFileTree } from '@votingworks/backend';
 import { InsertedSmartCardAuthApi } from '@votingworks/auth';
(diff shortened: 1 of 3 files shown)


apps/scan/backend/src/app.ts
kofi-q: Noted in the description, but planning on revisiting this and adding an additional guard here based on whether any ballots have been cast (and a corresponding one in the UI), but LMK if it should technically be safe to switch casting mode when polls are paused with some votes cast (wasn't sure if maybe that covers a machine that's used for early voting, paused, then reopened on election day)

adghayes:
@kofi-q I just skimmed this so may be missing a bunch of context. I've been under the impression that the `EDIT_POLLING_PLACES` would be a VxDesign-only toggle, and all election definitions would have polling places. So the picker in the precinct apps wouldn't have to vary in the offline system based on the flag. Is this a temporary thing or a permanent architecture?

PR #8236, opened 2026-04-03
Add support for 18" and 20" bubble ballots

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/7901

This PR adds support for 18" and 20" bubble ballots for Michigan. Was craving a quick win haha

## Testing Plan

- [x] Updated/augmented automated tests
- [x] Generated and tested an 18" ballot manually using local VxDesign and the dev dock
- [x] Manually tested the 20" all bubble ballot fixture
- [ ] Will complete additional manual tests during the next QA push

## Checklist

### DIFF
--- a/libs/ballot-interpreter/src/bubble-ballot-rust/ballot_card.rs
+++ b/libs/ballot-interpreter/src/bubble-ballot-rust/ballot_card.rs
@@ -961,7 +961,8 @@ pub fn get_matching_paper_info_for_image_size(
         })
         .min_by(|(_, (_, height_error1)), (_, (_, height_error2))| {
             height_error1
-                .partial_cmp(height_error2)
+                .abs()
+                .partial_cmp(&height_error2.abs())
                 .unwrap_or(Ordering::Equal)
         })
         .map(|(paper_info, _)| *paper_info)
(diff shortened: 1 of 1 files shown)


libs/ballot-interpreter/src/bubble-ballot-rust/ballot_card.rs
```diff
@@ -961,7 +961,8 @@ pub fn get_matching_paper_info_for_image_size(
         })
         .min_by(|(_, (_, height_error1)), (_, (_, height_error2))| {
             height_error1
-                .partial_cmp(height_error2)
+                .abs()
+                .partial_cmp(&height_error2.abs())
```
arsalansufi: The most interesting part of this PR. As noted in the commit message:

> We compute the height diffs between the image and each supported
paper size. We filter some paper sizes out as obviously wrong. Then
among the viable paper sizes with reasonable diffs, we pick the min
diff. But the old code was forgeting to apply an absolute value.
>
> This meant that, when looking at a perfect 19" ballot, we'd compute
a negative diff relative to the 20" size that was within the bounds
of reasonable. And we'd mistakenly pick 20" over 19" because the
20" comparison resulted in a small negative diff whereas the 19"
comparison resulted in a diff of 0. Applying absolute value
corrects this.
>
> Added a regression test accordingly

Concrete numbers for clarity, with Claude's help:

> Without .abs(), min_by picks the most negative error. For an 18" (3800px) image:
>   - Custom18 (3600px): error = (3800-3600)/3600 = +0.056                                                                                              
>   - Custom19 (3800px): error = (3800-3800)/3800 = 0.0                                                                                                 
>   - Custom20 (4000px): error = (3800-4000)/4000 = -0.05                                                                                               
>
>   Since -0.05 < 0.0 < +0.056, Custom20 wins — the negative error makes the larger size look like a "better" match.

